PR #4828, opened 2020-09-17
fix(MessageReaction): <reaction>.me not properly updating

Darqam:
Prior to this change, <reaction>me would not properly update when the client would have interacted with the same reaction. This PR aims to fix this issue.

Note that I am unaware how this affects typings (if it all).

**Status**

### DIFF
--- a/src/structures/MessageReaction.js
+++ b/src/structures/MessageReaction.js
@@ -118,14 +118,14 @@ class MessageReaction {
     if (this.partial) return;
     this.users.cache.set(user.id, user);
     if (!this.me || user.id !== this.message.client.user.id || this.count === 0) this.count++;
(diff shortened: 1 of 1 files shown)


tguichaoua:
With your changes, `this.me` is update each time `_add` or `_remove` is called. 

Example:
1. you call `_add` with the client's user: `me` is set to `true`.
2. you call `_add` with a regular user: `me` is set to `false`. (But it's must not be set to `false` here.)

This is the same problem with `_remove`.

Darqam:
I see the issue actually comes from me misunderstanding the docs, `me` is meant to be if the bot *has* reacted with that at any point previously and not if the client was the one reacting in this instance.

This "bug fix" is therefor not one and I'll close the PR.

Issue #4831, opened 2020-09-20
Some channels that extend GuildChannel improperly inherit setTopic


zaida04:
VoiceChannels, CategoryChannels, and other channels that extend GuildChannel inherit the setTopic method, which isn't right because topic related things only exists on TextChannel and NewsChannel. 

A solution would be to just remove it from GuildChannel and put it only on TextChannel, and since NewsChannel inherits from TextChannel, it would inherit the method.

Links: 
https://discord.js.org/#/docs/main/master/class/CategoryChannel?scrollTo=setTopic
https://discord.js.org/#/docs/main/master/class/VoiceChannel?scrollTo=setTopic
https://discord.js.org/#/docs/main/master/class/GuildChannel?scrollTo=setTopic

ImRodry:
This bug is still present to this day and I think it should be fixed before v13 is out. A good way to fix this and future issues would be to create a BaseGuildTextChannel that would hold all the common properties of TextChannel and NewsChannel to hopefully also fix the issue with rateLimitPerUser

iCrawl:
PRs welcome.

monbrey:
As I said in the linked issue, there's a lot more to it than just `setTopic`. If you do want to restructure the channel inheritance, there's probably other methods to look at too like createInvite.

ImRodry:
I was thinking of making a PR but not with the BaseGuildTextChannel class since I'm not entirely sure how to handle that and may end up breaking stuff so I just wanted to move things to the right place. Do you think it's worth making a PR just to do this or should we wait for someone more experienced with the library to do the full restructure?

PR #4832, opened 2020-09-20
fix: Remove setTopic from GuildChannel and apply to TextChannel

zaida04:
#4831 
Remove setTopic from GuildChannel
Add it to TextChannel ~~(or TextBasedChannel??)~~
NewsChannel inherits it from TextChannel.

**Status**

- [X] This PR includes breaking changes (methods removed or renamed, parameters moved or removed)

### DIFF
--- a/src/structures/CategoryChannel.js
+++ b/src/structures/CategoryChannel.js
@@ -7,6 +7,14 @@ const GuildChannel = require('./GuildChannel');
  * @extends {GuildChannel}
  */
 class CategoryChannel extends GuildChannel {
+  _patch(data) {
+    super._patch(data);
+
+    // Category Channels cannot have invites
+    this.createInvite = undefined;
+    this.fetchInvites = undefined;
+  }
+
   /**
    * Channels that are a part of this category
    * @type {Collection<Snowflake, GuildChannel>}
--- a/src/structures/CategoryChannel.js
+++ b/src/structures/CategoryChannel.js
@@ -7,8 +7,8 @@ const GuildChannel = require('./GuildChannel');
  * @extends {GuildChannel}
  */
 class CategoryChannel extends GuildChannel {
-  _patch(data) {
-    super._patch(data);
+  constructor(guild, data) {
+    super(guild, data);
 
     // Category Channels cannot have invites
     this.createInvite = undefined;
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -821,7 +821,6 @@ declare module 'discord.js' {
       options?: { lockPermissions?: boolean; reason?: string },
     ): Promise<this>;
     public setPosition(position: number, options?: { relative?: boolean; reason?: string }): Promise<this>;
-    public setTopic(topic: string, reason?: string): Promise<this>;
     public updateOverwrite(
       userOrRole: RoleResolvable | UserResolvable,
       options: PermissionOverwriteOption,
@@ -1527,6 +1526,7 @@ declare module 'discord.js' {
     ): Promise<Webhook>;
     public setNSFW(nsfw: boolean, reason?: string): Promise<TextChannel>;
     public setRateLimitPerUser(rateLimitPerUser: number, reason?: string): Promise<TextChannel>;
+    public setTopic(topic: string, reason?: string): Promise<TextChannel>;
     public fetchWebhooks(): Promise<Collection<Snowflake, Webhook>>;
   }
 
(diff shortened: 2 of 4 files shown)


monbrey:
If you're fixing incorrect inheritence, I'd also point out that CategoryChannel.createInvite and CategoryChannel.fetchInvites don't work, as categories cannot have invites.

typings/index.d.ts
```diff
@@ -1527,6 +1526,7 @@ declare module 'discord.js' {
     ): Promise<Webhook>;
     public setNSFW(nsfw: boolean, reason?: string): Promise<TextChannel>;
     public setRateLimitPerUser(rateLimitPerUser: number, reason?: string): Promise<TextChannel>;
+    public setTopic(topic: string, reason?: string): Promise<TextChannel>;
```
tguichaoua, line 1529:
```suggestion
    public setTopic(topic: string, reason?: string): Promise<this>;
```
If the methods is called on `NewsChannel` you will have the correct returned type with `this`.

tguichaoua (changes requested):

typings/index.d.ts
zaida04: Fixed. Thank you!

advaith1: Isn't this broken on all methods then?

src/structures/CategoryChannel.js
```diff
@@ -7,6 +7,14 @@ const GuildChannel = require('./GuildChannel');
  * @extends {GuildChannel}
  */
 class CategoryChannel extends GuildChannel {
+  _patch(data) {
+    super._patch(data);
+
+    // Category Channels cannot have invites
+    this.createInvite = undefined;
+    this.fetchInvites = undefined;
```
NotSugden, line 15: typings need updating for this, and perhaps this should be done in the constructor, and not every time the channel is updated

would it perhaps be an idea to use an interface (similar to `TextBasedChannel`) to apply these methods in the first place? idk

zaida04: Gotcha, I took inspiration for that from NewsChannel, which does it in the _patch method

src/structures/CategoryChannel.js
```diff
@@ -7,6 +7,14 @@ const GuildChannel = require('./GuildChannel');
  * @extends {GuildChannel}
  */
 class CategoryChannel extends GuildChannel {
+  constructor(guild, data) {
+    super(guild, data);
+
+    // Category Channels cannot have invites
+    this.createInvite = undefined;
```
vladfrangu, line 14: Is there even any need for this? You moved the base function from GuildChannel to TextChannel, CategoryChannel inherits from GuildChannel, not TextChannel

vladfrangu (changes requested):

src/structures/CategoryChannel.js
zaida04: If you’re referring to my setting of those methods to undefined, this was a solution to another problem brought up about category channels improperly inheriting methods https://github.com/discordjs/discord.js/pull/4832#issuecomment-695605097

vladfrangu: Oops, missed that, you're right

vladfrangu (approved):

iCrawl:
This needs its conflicts resolved.

src/structures/CategoryChannel.js
```diff
@@ -7,6 +7,14 @@ const GuildChannel = require('./GuildChannel');
  * @extends {GuildChannel}
  */
 class CategoryChannel extends GuildChannel {
+  constructor(guild, data) {
```
SpaceEEC, line 10: I don't think these changes belong into a PR with the title `fix: Remove setTopic from GuildChannel and apply to TextChannel`.

Also, I do not think that documented methods should be `undefined`, I'd rather have these behave similar to `PartialGroupDMChannel#delete` and `#fetch`: (Make them throw an error instead)
https://github.com/discordjs/discord.js/blob/9ffcd83027f0fc06d69df21475865ad55138de01/src/structures/PartialGroupDMChannel.js#L37-L43
We could additionally to that add a warning to the docstrings of these methods in the `CategoryChannel` class.

SpaceEEC (changes requested):

src/structures/CategoryChannel.js
zaida04: Wellll, this is consistent with https://github.com/discordjs/discord.js/blob/master/src/structures/NewsChannel.js#L14 where the same thing is done.

SpaceEEC: That is a property, not a method.

kyranet:
There are merge conflicts, please rebase or resolve them, @zaida04.

Furthermore, a solution for @SpaceEEC's requested change awaits. I think that after that has been addressed, the PR will look good to go.

zaida04:
Closing as this PR is on a now deleted fork of djs which means I can no longer push updates to this. Will reopen on new fork.

PR #4833, opened 2020-09-20
Update index.d.ts

bigbizze:
I am not sure if this is generated or not elsewhere but the argument you're passing into the function callback for broadcastEval is "this" where this is an object of type ShardClientUtil.

You could alternatively consider passing in this.client instead of this if that's what you intended.

**Status**

### DIFF
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -1374,7 +1374,7 @@ declare module 'discord.js' {
     public mode: ShardingManagerMode;
     public parentPort: any | null;
     public broadcastEval(script: string): Promise<any[]>;
-    public broadcastEval<T>(fn: (client: Client) => T): Promise<T[]>;
+    public broadcastEval<T>(fn: (client: ShardClientUtil) => T): Promise<T[]>;
     public fetchClientValues(prop: string): Promise<any[]>;
     public respawnAll(shardDelay?: number, respawnDelay?: number, spawnTimeout?: number): Promise<void>;
     public send(message: any): Promise<void>;


almostSouji:
I just tested this and am not quite sure how you got to the conclusion something might be off here? BroadcastEval can be executed in two ways:

1. you pass a string of arbitrary code to be executed in context of the client (the Client of each shard during execution is bound to this)
2. you pass a function that gets executed (the Client of each shard during execution is bound to the first function parameter as well as this)

Accordingly all three variations below will execute and yield the expected results:

```js
client.shard.broadcastEval(() => this.shard.ids[0]);
client.shard.broadcastEval((client) => client.shard.ids[0])
client.shard.broadcastEval("this.shard.ids[0]")
```

Which also correctly fits the two type declarations in place:

https://github.com/discordjs/discord.js/blob/1e63f3756e814d6b1a2a9c17af9c2b28ce37e472/typings/index.d.ts#L1376-L1377

In case I am missing something feel free to re-open and explain what exactly the wrong behavior is, but so far this looks like an invalid typing suggestion to me.

bigbizze:
> you pass a string of arbitrary code to be executed in context of the client (the Client of each shard during execution is bound to this)

Ah, good to know, I wasn't of it being rebound, I probably should have done more investigation prior to making this request.

https://github.com/discordjs/discord.js/blob/4c2308b4c697e5c82e84bf54dc158784d0285b4e/src/sharding/ShardClientUtil.js#L140

Fwiw, I was referring to here, where "this" referring to ShardClientUtil and not Client appears to being passed to the callback given by the user when script is a function.

I must have been having a different issue where the client argument had unexpected properties for a different reason.

In any event, I switched to using a string again since it appears like the function passed is serialized anyway (as expected).

kyranet:
That's a string, and it's run in `client._eval`:

Code that receives the _sEval command:

https://github.com/discordjs/discord.js/blob/5249cf33e56c56e49b62f4438ca46c56802fa9a6/src/sharding/Shard.js#L333-L340

Code that then runs the code:

https://github.com/discordjs/discord.js/blob/4c2308b4c697e5c82e84bf54dc158784d0285b4e/src/sharding/ShardClientUtil.js#L183-L189

`Client#_eval`:

https://github.com/discordjs/discord.js/blob/4c2308b4c697e5c82e84bf54dc158784d0285b4e/src/client/Client.js#L383-L392

If you run `client._eval('this');`, you will receive the client itself, because it's evaluated in Client's context. Same happens here.

PR #4835, opened 2020-09-20
feat(Rest): switch queue to AsyncQueue

### LINKED ISSUE #2496: Incomplete Error Stack Trace Capturing

For months, the error stack traces were incredibly poor in `DiscordAPIError` instances, where in small bots this is not such as problem since the code can be found and fixed easily, this becomes a more serious issue when it comes to massive bots that have dozens of thousands lines of code, where a single error/bug is already very hard to find and debug without a proper error stack.

```prolog
DiscordAPIError: Missing Permissions
     at item.request.gen.end (/.../node_modules/discord.js/src/rest/handlers/RequestHandler.js:79:65)
     at then (/.../node_modules/snekfetch/src/index.js:218:21)
     at process._tickCallback (internal/process/next_tick.js:178:7)
```

Whilst I understand errors should be handled (and they are, those errors are handled by either `try`/`catch` or with `Promise#catch()`), but generally speaking, you can have many functions that interact with the Discord API that may not be handled correctly, this risk, again, becomes bigger as the bot's code becomes larger.

Multipurpose bots (like mine) are greatly affected by this issue, since they make many kinds of requests or abstract complex "modules" separately from what comes to be "pieces" (most of them use a framework) such as commands, events... e.g. starboard, where it interacts with messages listening to the *messageReactionAdd* event, check in the cache if the message was starred already (or check a database, in more complex modules), then send in case the message wasn't starred, or edit it. There are many kinds of things a single module can do for the sake of working correctly.

**I understand** that the Discord API throws generic error codes (such as in this case, `50013` [Missing Permissions]), and discord.js handles those errors correctly, the stack is, however, incomplete. The code `50013`, and the name `Missing Permissions`, are too generic. We also have [`DiscordAPIError#path`](https://discord.js.org/#/docs/main/master/class/DiscordAPIError?scrollTo=path) which, thankfully, tells us the endpoint used so we can recognize which type (reaction, message create, message edit, message delete...) (**but does not tell us which HTTP request has been used!**, [message edit](https://discordapp.com/developers/docs/resources/channel#edit-message) and [message delete](https://discordapp.com/developers/docs/resources/channel#delete-message) have the same URL, but the former is a PATCH and the latter is a DELETE).

The line that throws the error (with an useless stack) is here: https://github.com/discordjs/discord.js/blob/b186472785db6bbe7de37e1a75104d79dd0cc44f/src/rest/handlers/RequestHandler.js#L79.

The error handling has been done in multiple ways, but the following is the only one that ensures an almost-accurate error stack, however, this is ugly, large, and be misleading when chaining multiple API requests:

```javascript
const { DiscordAPIError } = require('discord.js');

// ...
try {
  // Do an API operation without permissions or a malformed
  // API request, e.g. messages with a too-long content, embeds
  // with many fields...
} catch (error) {
  if (error instanceof DiscordAPIError) Error.captureStackTrace(error);
  console.error(error);
}
// ...
```

Which outputs this to console (the test has been done with eval):

```javascript
DiscordAPIError: Missing Permissions
    at eval (eval at eval (/.../commands/System/Admin/eval.js:96:13), <anonymous>:1:7)
    at module.exports.eval (/.../commands/System/Admin/eval.js:96:13)
    at module.exports.timedEval (/.../commands/System/Admin/eval.js:84:9)
    at module.exports.run (/.../commands/System/Admin/eval.js:20:54)
    at module.exports.runCommand (/.../node_modules/klasa/src/monitors/commandHandler.js:89:90)
    at process._tickCallback (internal/process/next_tick.js:178:7)
```

And the error stack above is more helpful, it tells the end developer the line that failed. Unfortunately, I am not very confident with discord.js' REST so I do not know what would be causing the issue, or if it's caused by `snekfetch` (since it's where the stack starts at). But I will be very thankful if this issue gets solved. It is not like it's in the top of my priorities, but I have created this issue to point out two issues:

- The `DiscordAPIError` class does capture the path, but not the method (as mentioned before, message edit and message delete have the same endpoint, but they differ on the method).
- The Error stacks are incomplete and are too hard to debug in large bots since they do not give us the location of our code that caused the issue.

In a channel your bot has no permissions to send a message (to make it throw a `DiscordAPIError`):

```javascript
message.channel.send('Test')
  .catch(console.error);
```

**Further details:**

- discord.js version: master [`b8a9a76`](https://github.com/discordjs/discord.js/commit/b8a9a76cf6d9a1ac59b95913ed9bc7dc7d0f61da)
- node.js version: Node.js 8.11.1 / 9.11.1 / 10.0.0-rc.1 (Happens in any compatible version)
- Operating system: Windows 10 Fall Creators Update, Ubuntu 16.04
- Priority this issue should have: low, this issue can be [monkey patched](https://en.wikipedia.org/wiki/Monkey_patch) (although losing performance for this), and a very refined error handler can display half of the details needed to find the issue.

- [ ] I found this issue while running code on a __user account__
- [x] I have also tested the issue on latest master, commit hash: `b8a9a76cf6d9a1ac59b95913ed9bc7dc7d0f61da` (not latest, but `/src/rest` hasn't changed in 4 months).

### LINKED ISSUE #3665: Reactions getting stuck
Reacting on a message made by my bot hangs. This hang gets resolved by pushing more reacts through the rest queue. I was unable to reproduce this problem but if contacted would be willing to give my code.

Tried reproducing was unable to however my bot is a living example of this problem.
https://cdn.discordapp.com/attachments/334350562659139584/662982934097166336/music_-_Discord_2020-01-04_12-35-54.mp4
^^ Video where you can see the exact problem

**Further details:**
- discord.js version: v12
- Node.js version: v10.16.3 & v12.14.0
- Operating system: Debian 10.2 & Debian 10.1.0
- Priority this issue should have – please be realistic and elaborate if possible:
Not very high as it happens in really rare cases and as how I was unable to reproduce this

Prior to v12 when my bot was still using v11.x this problem was not present

- [x] I have also tested the issue on latest master, commit hash: d2ef02906c5618d883b1adfb348fd248a4db616f

### LINKED ISSUE #4792: Support more complete async stack traces when failures occur during API calls
When errors occur on API calls, it can sometimes be difficult to determine where in my code the call was made.

```txt
DiscordAPIError: Missing Permissions
    at RequestHandler.execute (C:\Users\kevin\Desktop\discord.js-async-example\node_modules\discord.js\src\rest\RequestHandler.js:170:25)
    at processTicksAndRejections (internal/process/task_queues.js:97:5) {
  method: 'post',
  path: '/channels/471091337459007490/messages',
  code: 50013,
  httpStatus: 403
}
```

Take the following example, where I purposely denied the "**Send Messages**" permission in a channel, but send the "**test**" command anyway:

```javascript
const Discord = require('discord.js');
const client = new Discord.Client();

const MY_TOKEN = '{INSERT TOKEN}';

async function sendThisMessagePlease(channel, text) {
    await channel.send(text);
}

(async () => {
    client.on('message', async msg => {
        if (msg.content.toLowerCase() !== 'test') {
            return;
        }

        try {
            await sendThisMessagePlease(msg.channel, 'Hello world');
        } catch (error) {
            console.error(error.stack);
        }
    });

    await client.login(MY_TOKEN);
})();
```
With the "Send Messages" permission missing, the following stack trace is captured:
```txt
DiscordAPIError: Missing Permissions
    at RequestHandler.execute (C:\Users\kevin\Desktop\discord.js-async-example\node_modules\discord.js\src\rest\RequestHandler.js:170:25)
    at processTicksAndRejections (internal/process/task_queues.js:97:5)
```

Notice that this stack trace does not contain any reference to the above source code.

This lack of information seems to be caused by returning a `new Promise` in [RESTManager.js push()](https://github.com/discordjs/discord.js/blob/master/src/rest/RESTManager.js#L38) and according to [this article](https://www.even.li/posts/2020-06-04-async-stack-traces/), the stack trace can be basically lost if you do not use the "async/await" keywords.

I've made some changes to the RESTManager and RequestHandler in my [examples/async-stack-trace](https://github.com/KevinNovak/discord.js/commit/c946acb053eb8ca6314509045a96e19a1f655957) branch. This is NOT production code, but just a way to show what the stack traces **could** look like if async/await was used throughout.

Using the same code example as above, denying the "Send Messages" permission, but using my branch for discord.js, I get a new stack trace with more information:

```txt
DiscordAPIError: Missing Permissions
    at RequestHandler.execute (C:\Users\kevin\Desktop\discord.js-async-example\node_modules\discord.js\src\rest\RequestHandler.js:171:15)
    at processTicksAndRejections (internal/process/task_queues.js:97:5)
    at async sendThisMessagePlease (C:\Users\kevin\Desktop\discord.js-async-example\example-1.js:7:5)
    at async Client.<anonymous> (C:\Users\kevin\Desktop\discord.js-async-example\example-1.js:17:13)
```

Notice that this stack trace is more complete, and I can tell exactly where in my code this error is happening.

kyranet:

Better alternative to #2744.

- Fixes #2496.
- Fixes #4792.
- Fixes #3665.

When running the following code as an example `message.channel.send();`, REST gives us this stack trace:

```javascript
DiscordAPIError: Cannot send an empty message
    at RequestHandler.execute (***/node_modules/discord.js/src/rest/RequestHandler.js:170:25)
    at runMicrotasks (<anonymous>)
    at processTicksAndRejections (internal/process/task_queues.js:93:5)
```

After this patch, it returns this one:

```javascript
DiscordAPIError: Cannot send an empty message
    at RequestHandler.execute (***/node_modules/discord.js/src/rest/RequestHandler.js:157:13)
    at processTicksAndRejections (internal/process/task_queues.js:93:5)
    at async RequestHandler.push (***/node_modules/discord.js/src/rest/RequestHandler.js:39:14)
    at async Object.eval (***/dist/commands/System/Admin/eval.js:73:26)
    at async Object.run (***/dist/commands/System/Admin/eval.js:18:49)
    at async default_1.runCommand (***/dist/monitors/commandHandler.js:57:38)
    ...
```

Which correctly points to where I ran my eval command (same code before and after this patch).

**Typings are unpatched until further decision from the team on whether we should use discord.js v13's core package (once released, will come with its typings), or stick with this implementation in the meantime.**

There are a few modifications needed for this to work, but since the modified properties were inside the library's internals (which are private), I believe this falls under `semver: patch`, and we can ship this in v12.4.

What is better about this compared to the previous system?

- Can be easily tweaked to support retries on `AbortError`, which can fix 2 issues in one go.
- Complete async stack traces (async queue lock implementation, needs Node.js v12, which the library already needs, V8's async stack traces are very new).
- Impossible to dead-lock by design, if an error happened during the handler, it'd hang as it needs to call `reject()`, with this patch, it'd bubble up, and a `try`/`finally` takes care of resuming the queue.

**Status**

### DIFF
--- a/src/rest/AsyncQueue.js
+++ b/src/rest/AsyncQueue.js
@@ -1,0 +1,91 @@
+/**
+ * MIT License
+ *
+ * Copyright (c) 2020 kyranet, discord.js
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in all
+ * copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+
+'use strict';
+
+// TODO(kyranet, vladfrangu): replace this with discord.js v13's core AsyncQueue.
+
+/**
+ * An async queue that preserves the stack and prevents lock-ups
+ */
+class AsyncQueue {
+  constructor() {
+    /**
+     * The promises array
+     * @private
+     */
+    this.promises = [];
+  }
+  /**
+   * The remaining amount of queued promises
+   */
+  get remaining() {
+    return this.promises.length;
+  }
+
+  /**
+   * Waits for last promise and queues a new one
+   * @returns {Promise<void>}
+   * @example
+   * const queue = new AsyncQueue();
+   * async function request(url, options) {
+   *     await queue.wait();
+   *     try {
+   *         const result = await fetch(url, options);
+   *         // Do some operations with 'result'
+   *     } finally {
+   *         // Remove first entry from the queue and resolve for the next entry
+   *         queue.shift();
+   *     }
+   * }
+   *
+   * request(someUrl1, someOptions1); // Will call fetch() immediately
+   * request(someUrl2, someOptions2); // Will call fetch() after the first finished
+   * request(someUrl3, someOptions3); // Will call fetch() after the second finished
+   */
+  wait() {
+    const next = this.promises.length ? this.promises[this.promises.length - 1].promise : Promise.resolve();
+    let resolve;
+    const promise = new Promise(res => {
+      resolve = res;
+    });
+
+    this.promises.push({
+      resolve,
+      promise,
+    });
+
+    return next;
+  }
+
+  /**
+   * Frees the queue's lock for the next item to process
+   */
+  shift() {
+    const deferred = this.promises.shift();
+    if (typeof deferred !== 'undefined') deferred.resolve();
+  }
+}
+
+module.exports = AsyncQueue;
--- a/src/rest/AsyncQueue.js
+++ b/src/rest/AsyncQueue.js
@@ -38,6 +38,7 @@ class AsyncQueue {
      */
     this.promises = [];
   }
+
   /**
    * The remaining amount of queued promises
    * @type {number}
@@ -47,7 +48,7 @@ class AsyncQueue {
   }
 
   /**
-   * Waits for last promise and queues a new one
+   * Waits for last promise and queues a new one.
    * @returns {Promise<void>}
    * @example
    * const queue = new AsyncQueue();
@@ -82,7 +83,7 @@ class AsyncQueue {
   }
 
   /**
-   * Frees the queue's lock for the next item to process
+   * Frees the queue's lock for the next item to process.
    */
   shift() {
     const deferred = this.promises.shift();
--- a/src/rest/AsyncQueue.js
+++ b/src/rest/AsyncQueue.js
@@ -32,13 +32,15 @@
 class AsyncQueue {
   constructor() {
     /**
-     * The promises array
+     * The promises array.
+     * @type {Array<{promise: Promise<void>, resolve: Function}>}
      * @private
      */
     this.promises = [];
   }
   /**
    * The remaining amount of queued promises
+   * @type {number}
    */
   get remaining() {
     return this.promises.length;
--- a/src/rest/AsyncQueue.js
+++ b/src/rest/AsyncQueue.js
@@ -27,7 +27,8 @@
 // TODO(kyranet, vladfrangu): replace this with discord.js v13's core AsyncQueue.
 
 /**
- * An async queue that preserves the stack and prevents lock-ups
+ * An async queue that preserves the stack and prevents lock-ups.
+ * @private
  */
 class AsyncQueue {
   constructor() {
--- a/src/rest/APIRequest.js
+++ b/src/rest/APIRequest.js
@@ -15,6 +15,7 @@ class APIRequest {
     this.method = method;
     this.route = options.route;
     this.options = options;
+    this.retries = 0;
 
     let queryString = '';
     if (options.query) {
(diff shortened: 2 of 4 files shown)


src/rest/AsyncQueue.js
```diff
@@ -0,0 +1,93 @@
...
+class AsyncQueue {
+  constructor() {
+    /**
+     * The promises array.
+     * @type {Array<{promise: Promise<void>, resolve: Function}>}
+     * @private
+     */
+    this.promises = [];
+  }
+  /**
```
NotSugden, line 41:
```suggestion
  }

  /**
```

src/rest/AsyncQueue.js
```diff
@@ -0,0 +1,93 @@
...
+   * The remaining amount of queued promises
+   * @type {number}
+   */
+  get remaining() {
+    return this.promises.length;
+  }
+
+  /**
+   * Waits for last promise and queues a new one
```
NotSugden, line 50:
```suggestion
   * Waits for last promise and queues a new one.
```

src/rest/AsyncQueue.js
```diff
@@ -0,0 +1,93 @@
...
+    this.promises.push({
+      resolve,
+      promise,
+    });
+
+    return next;
+  }
+
+  /**
+   * Frees the queue's lock for the next item to process
```
NotSugden, line 85:
```suggestion
   * Frees the queue's lock for the next item to process.
```

iCrawl (approved):

Papaia (commented):
Shouldn't AsyncQueue be a private class?

vladfrangu (approved):
Unless we want to wait a bit more for v13's rest, LGTM

SpaceEEC (approved):

PR #4837, opened 2020-09-21
feat(GuildManager): add AFK and system channel options in create

cherryblossom000:

This PR adds support for the `afk_channel_id`, `afk_timeout`, and `system_channel_id` parameters in the [create guild](https://discord.com/developers/docs/resources/guild#create-guild-json-params) endpoint by adding the `afkChannelID`, `afkTimeout`, and `systemChannelID` options in `GuildManager#create`.

This PR also fixes a minor bug in `create` when specifying types for the channels due to the channel types not being changed from `'text'`, `'voice'` etc to the corresponding numbers, so Discord would return an error.

**Status**

### DIFF
--- a/src/managers/GuildManager.js
+++ b/src/managers/GuildManager.js
@@ -8,6 +8,7 @@ const GuildMember = require('../structures/GuildMember');
 const Invite = require('../structures/Invite');
 const Role = require('../structures/Role');
 const {
+  ChannelTypes,
   Events,
   VerificationLevels,
   DefaultMessageNotifications,
@@ -129,6 +130,8 @@ class GuildManager extends BaseManager {
    * <warn>This is only available to bots in fewer than 10 guilds.</warn>
    * @param {string} name The name of the guild
    * @param {Object} [options] Options for the creating
+   * @param {number} [options.afkChannelID] The ID of the AFK channel
+   * @param {number} [options.afkTimeout] The AFK timeout in seconds
    * @param {PartialChannelData[]} [options.channels] The channels for this guild
    * @param {DefaultMessageNotifications} [options.defaultMessageNotifications] The default message notifications
    * for the guild
@@ -137,18 +140,22 @@ class GuildManager extends BaseManager {
    * @param {string} [options.region] The region for the server, defaults to the closest one available
    * @param {PartialRoleData[]} [options.roles] The roles for this guild,
    * the first element of this array is used to change properties of the guild's everyone role.
+   * @param {number} [options.systemChannelID] The ID of the system channel
    * @param {VerificationLevel} [options.verificationLevel] The verification level for the guild
    * @returns {Promise<Guild>} The guild that was created
    */
   async create(
     name,
     {
+      afkChannelID,
+      afkTimeout,
       channels = [],
       defaultMessageNotifications,
       explicitContentFilter,
       icon = null,
       region,
       roles = [],
+      systemChannelID,
       verificationLevel,
     } = {},
   ) {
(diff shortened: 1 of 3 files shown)


typings/index.d.ts
```diff
@@ -2577,6 +2574,19 @@ declare module 'discord.js' {
     name?: string;
   }
 
+  interface GuildCreateOptions {
+    afkChannelID?: number;
```
tguichaoua, line 2578:
```suggestion
    afkChannelID?: Snowflake;
```
The type is `snowflake` in Discord's doc.

typings/index.d.ts
```diff
@@ -2577,6 +2574,19 @@ declare module 'discord.js' {
...
+    afkTimeout?: number;
+    channels?: PartialChannelData[];
+    defaultMessageNotifications?: DefaultMessageNotifications | number;
+    explicitContentFilter?: ExplicitContentFilterLevel | number;
+    icon?: BufferResolvable | Base64Resolvable | null;
+    region?: string;
+    roles?: PartialRoleData[];
+    systemChannelID?: number;
```
tguichaoua, line 2586:
```suggestion
    systemChannelID?: Snowflake;
```
The type is snowflake in Discord's doc.

src/managers/GuildManager.js
```diff
@@ -129,6 +130,8 @@ class GuildManager extends BaseManager {
    * <warn>This is only available to bots in fewer than 10 guilds.</warn>
    * @param {string} name The name of the guild
    * @param {Object} [options] Options for the creating
+   * @param {number} [options.afkChannelID] The ID of the AFK channel
```
tguichaoua, line 133:
```suggestion
   * @param {Snowflake} [options.afkChannelID] The ID of the AFK channel
```

src/managers/GuildManager.js
```diff
@@ -137,18 +140,22 @@ class GuildManager extends BaseManager {
    * @param {string} [options.region] The region for the server, defaults to the closest one available
    * @param {PartialRoleData[]} [options.roles] The roles for this guild,
    * the first element of this array is used to change properties of the guild's everyone role.
+   * @param {number} [options.systemChannelID] The ID of the system channel
```
tguichaoua, line 143:
```suggestion
   * @param {Snowflake} [options.systemChannelID] The ID of the system channel
```

tguichaoua (changes requested):

typings/index.d.ts
cherryblossom000: Even though the docs says it's a snowflake, it can accept a number of one of the IDs used in the `channels` (like how `parent_id` works). Should I just use `Snowflake` because that's what the docs says, or should we allow (normal) numbers as well because a `Snowflake` implies that the number as to be in the snowflake format? I also created https://github.com/discord/discord-api-docs/issues/2089 that addresses this.

tguichaoua: Sorry, my bad. I didn't read that:
```
When using the channels parameter, the id field within each channel object may be set to an integer placeholder, and will be replaced by the API upon consumption.
```
So, I think that `number` is the suitable type.

src/managers/GuildManager.js
vladfrangu: This isn't actually wrong per say, as both Snowflakes as int or strings work in the create guild endpoint 👀

vladfrangu (approved):

kyranet (approved):

SpaceEEC (approved):

iCrawl (approved):

PR #3168, opened 2019-03-29
src: Add News and Store channels, and missing guild props

vladfrangu:
**Please describe the changes this PR makes and why it should be merged:**

This PR adds:

- Support for the News and Store Channel types
- Support for TextChannels and NewsChannels being able to change their type
	- Thanks Discord, very cool
- Adds the missing properties in the guild object
	- `max_members`
	- `max_presences`
	- `banner`
	- `description`
	- `vanity_url_code`

**Status**
- [x] Code changes have been tested against the Discord API, or there are no code changes
- [x] I know how to update typings and have done so, or typings don't need updating

**Semantic versioning classification:**  
- [x] This PR changes the library's interface (methods or parameters added)
  - [ ] This PR includes breaking changes (methods removed or renamed, parameters moved or removed)
- [ ] This PR **only** includes non-code changes, like changes to documentation, README, etc.

PR #4845, opened 2020-09-23
refactor: improve the accuracy of docs/improve docs

NotSugden:
This PR changes some code to
- match the documentation, some properties are typed as `?Type` which [indicates nullable](https://jsdoc.app/tags-type.html) but were being set to `undefined`, they are now set to null
  - this also just changes some code in general to prefer null over undefined
- correctly types some properties as nullable (such as `Guild#voice`)
- removes unnessercary `@name` tags from certian declarations, they are only needed if the typedef is not directly above the actual declaration
- changes a couple of `@prop` tags to `@property` for consistency
- changes some classes to have an `@abstract` tag on them, as they are abstract classes

**this PR is a breaking change**

**Status**

  - not fully tested if `null` causes any internal problems with the library (eg `typeof x === 'undefined'), will get round to it


src/structures/Guild.js
```diff
@@ -299,40 +301,44 @@ class Guild extends Base {
...
-     * <info>You will need to fetch the guild using {@link Guild#fetch} if you want to receive this parameter</info>
-     * @type {?number}
-     * @name Guild#maximumMembers
-     */
-    if (typeof data.max_members !== 'undefined') this.maximumMembers = data.max_members || 250000;
+    if (typeof data.max_members !== 'undefined') {
+      /**
+       * The maximum amount of members the guild can have
+       * <info>You will need to fetch the guild using {@link Guild#fetch} if you want to receive this parameter</info>
+       * @type {?number}
```
kyranet, line 308: The field defaulting it to 250000.

```suggestion
       * @type {number}
```

src/structures/Guild.js
```diff
@@ -299,40 +301,44 @@ class Guild extends Base {
...
-     * <info>You will need to fetch the guild using {@link Guild#fetch} if you want to receive this parameter</info>
-     * @type {?number}
-     * @name Guild#maximumPresences
-     */
-    if (typeof data.max_presences !== 'undefined') this.maximumPresences = data.max_presences || 25000;
+    if (typeof data.max_presences !== 'undefined') {
+      /**
+       * The maximum amount of presences the guild can have
+       * <info>You will need to fetch the guild using {@link Guild#fetch} if you want to receive this parameter</info>
+       * @type {?number}
```
kyranet, line 317: The field is defaulting to 25000.

```suggestion
       * @type {number}
```

kyranet (changes requested):

src/structures/Guild.js
NotSugden: i'll remove the if statements surrounding these aswell then ye?

NotSugden: upon further look - both of these are optional on the [API Docs](https://discord.com/developers/docs/resources/guild) however, `max_members` is sent in both gateway `GUILD_CREATE` and REST `GET /guilds/:id` - but im not sure if the docs are wrong here or there is a case where it can be omitted so i marked it as nullable to be safe - its also not marked as null on the docs so there isn't a need to apply a default to max_members - it should always be a number

izexi: The `max_members` field wasn't present within the `GUILD_CREATE` event when it was initially rolled out (see https://github.com/discordjs/discord.js/pull/3168#discussion_r272542448) hence why `Guild#fetch()` was introduced along with the JSDoc info. But that seemed to have changed... and it is now present there, I agree with @NotSugden that they should still probably be marked as nullable for both `maximumMembers` & `maximumPresences` since that's how it is documented on the API docs and just in case Discord decides to make another change.

src/sharding/Shard.js
```diff
@@ -96,7 +96,7 @@ class Shard extends EventEmitter {
      * @type {Function}
      * @private
      */
-    this._exitListener = this._handleExit.bind(this, undefined);
```
vladfrangu, line 99: This may cause issues if reverted, as it's there for a reason (defaulting respawn to this.manager.respawn)... can you test it out?

src/structures/GuildAuditLogs.js
```diff
@@ -84,7 +84,7 @@ const Targets = {
 /**
  * All available actions keyed under their names to their numeric values.
  * @name GuildAuditLogs.Actions
- * @type {AuditLogAction}
+ * @type {Object<string, number>}
```
vladfrangu, line 87: Why were these changed? IIRC these typedefs exist, right?

vladfrangu (changes requested):

src/structures/GuildAuditLogs.js
NotSugden: that was changed because that doc isn't the actual typedef for `AuditLogAction` and it was causing `GuildAuditLogs.Actions` to be typed as `string|number` on the docs, which is incorrect, its an object with keys as strings and numbers for values

src/sharding/Shard.js
NotSugden: i did actually test this, binding the first parameter to undefined still seems to cause the defaulting behavoir
```js
const fn = ((param = true) => param).bind(null, undefined);
console.log(fn()); // true, not undefined
```

kyranet (approved):

vladfrangu (approved):

SpaceEEC (approved):

iCrawl (approved):

PR #3723, opened 2020-01-19
Add support for MessageReaction#remove and MESSAGE_REACTION_REMOVE_EMOJI

RDambrosio016:
**Please describe the changes this PR makes and why it should be merged:**
This pr adds support for the new event MESSAGE_REACTION_REMOVE_EMOJI documented in https://github.com/discordapp/discord-api-docs/pull/1309, it adds a new method to `MessageReaction` to remove it (remove all occurences of an emoji reaction from a message). it adds another event which is triggered when a bot does this too.

**Status**
- [x] Code changes have been tested against the Discord API, or there are no code changes
- [x] I know how to update typings and have done so, or typings don't need updating

**Semantic versioning classification:**  
- [x] This PR changes the library's interface (methods or parameters added)
  - [ ] This PR includes breaking changes (methods removed or renamed, parameters moved or removed)
- [ ] This PR **only** includes non-code changes, like changes to documentation, README, etc.

PR #4861, opened 2020-09-29
refactor: guildAvailable, typingStop from Constants

louistio:

Remove guildAvailable,typingStop from Constants
    
Both of these events no longer exist and should not be exposed in
Constants. Typings have been adjusted to correspond to Constants.Events.

**Status**

I don't think this should be considered a breaking change as these events are not documented and there is no reason why a user would access these constants.


typings/index.d.ts
```diff
@@ -395,15 +394,15 @@ declare module 'discord.js' {
       MESSAGE_REACTION_ADD: 'messageReactionAdd';
       MESSAGE_REACTION_REMOVE: 'messageReactionRemove';
       MESSAGE_REACTION_REMOVE_ALL: 'messageReactionRemoveAll';
+      MESSAGE_REACTION_REMOVE_EMOJI: 'messageReactionRemoveEmoji';
       USER_UPDATE: 'userUpdate';
       PRESENCE_UPDATE: 'presenceUpdate';
       VOICE_STATE_UPDATE: 'voiceStateUpdate';
+      VOICE_SERVER_UPDATE: 'voiceServerUpdate';
```
louistio, line 401: This event was added to constants in #3124 even though it's not used in the lib... I think it should probably be removed, up for discussion

typings/index.d.ts
```diff
@@ -395,15 +394,15 @@ declare module 'discord.js' {
       MESSAGE_REACTION_ADD: 'messageReactionAdd';
       MESSAGE_REACTION_REMOVE: 'messageReactionRemove';
       MESSAGE_REACTION_REMOVE_ALL: 'messageReactionRemoveAll';
+      MESSAGE_REACTION_REMOVE_EMOJI: 'messageReactionRemoveEmoji';
```
louistio, line 397: This seems to have been forgotten in #3723

typings/index.d.ts
```diff
@@ -362,7 +362,6 @@ declare module 'discord.js' {
     Events: {
```
louistio, line 362: Not directly related to these changes, but why is `guildMemberAvailable` [not documented](https://discord.js.org/#/docs/main/stable/class/Client)? It is part of `ClientEvents` and emitted in `PresenceUpdate.js`.

louistio, line 362: Not directly related to these changes, but why is `disconnect` part of `ClientEvents`? It's [not documented](https://discord.js.org/#/docs/main/stable/class/Client) and doesn't seem to be emitted? Which is weird because `ShardClientUtil.js` adds listeners for it (unless `ShardClientUtil#client` is a `Shard` object)

vladfrangu: That event eas long removed iirc.. might need to update things

SpaceEEC: 1. The event was removed with 3f5161eb76b71ab8984e39ff39bf2937b82f5b43, and probably forgotten here.

2. I do not know how we want to handle that in `ShardClientUtil` now.
Using `Client#invalidated` instead could be an option. That however would require some internal refactoring as it currently is not emitted on the initial connecting process for some reason.

SpaceEEC: Only the docstring was removed in f3cad81f5314a3997ac188277e2471ba84cbd771, the code emitting it was not touched. This looks more like an oversight, than an intended removal.

typings/index.d.ts
SpaceEEC: The raw data of this event is already accessible through [`Client#ws`](https://discord.js.org/#/docs/main/stable/class/WebSocketManager) and this is not something you need a lot (e.g. if you use an external audio player).

With that in mind, I think it's fine to just remove this line instead of just forwarding it in the lib.

louistio: You're saying it should be removed from `Constants.js`, correct? Typings should be in sync with the actual file, so I would undo adding this line as well

kyranet:
There are merge conflicts, please rebase or resolve them, @louistio.

louistio:
@kyranet Ok, I've rebased the changes. @SpaceEEC never replied to my question but I went ahead and remove `VOICE_SERVER_UPDATE`, which is what I think they were suggesting.

amishshah:
@louistio could you re-add the VOICE_SERVER_UPDATE event? It's used in an ongoing PR that will add support for @discordjs/voice.

louistio:
@amishshah Sure, I've re-added it 👍

iCrawl (approved):

kyranet (approved):

src/util/Constants.js
```diff
@@ -235,7 +235,6 @@ exports.Events = {
   GUILD_DELETE: 'guildDelete',
   GUILD_UPDATE: 'guildUpdate',
   GUILD_UNAVAILABLE: 'guildUnavailable',
-  GUILD_AVAILABLE: 'guildAvailable',
```
vladfrangu, line 238: Honestly, I still don't get why we removed this event, but keep `guildMemberAvailable` in our code. personally, this event should reappear again, to be able to know when a guild goes from unavailable to available. CC @iCrawl

vladfrangu (changes requested):

SpaceEEC (approved):

src/util/Constants.js
SpaceEEC: Me neither, but I think readding this would belong into a separate PR.
In similar manner could `guildMemberAvailable` be made more reliable, since we add members into the cache from all over the place, not just the two places were we currently emit that event.

PR #4866, opened 2020-09-29
feat(VoiceState): Add `VoiceState.setMuteDeaf`

### LINKED ISSUE #4865: Add `VoiceState.setMuteDeaf` function
The `VoiceState.setMute()` and `VoiceState.setDeaf()` functions have to be called separately, requiring two requests against the rate-limit if trying to mute and deafen a user.

Since they use the same API call, a combined `setMuteDeaf` function would allow both changes to be made in the same request, thus making more efficient use of the rate limit.

My current approach is to check the voice state of users before hand, and only make the request when required:
```Node.js
        const { voice } = await this._guildMember.fetch();

        // Don't waste rate limits on duplicate requests.
        const requests = [];
        if (voice.serverMute !== mute) requests.push(voice.setMute(mute, finalReason));
        if (voice.serverDeaf !== deaf) requests.push(voice.setDeaf(deaf, finalReason));
        await Promise.all(requests)
```

However, this doesn't help in cases where both updates are required.

**Additional context**
I would be more than happy to attempt a pull-request for this feature, if desired.

jftanner:
Fixes #4865 by adding a helper method to mute and deafen users with one API call.

**Status**

### DIFF
--- a/src/structures/VoiceState.js
+++ b/src/structures/VoiceState.js
@@ -148,6 +148,19 @@ class VoiceState extends Base {
   }
 
   /**
+   * Mutes/unmutes and deafens/undeafens the member of this voice state.
+   * @param {boolean} mute Whether or not the member should be muted
+   * @param {boolean} deaf Whether or not the member should be deafened
+   * @param {string} [reason] Reason for muting/unmuting and deafening/undeafening
+   * @returns {Promise<GuildMember>}
+   */
+  setMuteDeaf(mute, deaf, reason) {
+    return this.member
+      ? this.member.edit({ mute, deaf }, reason)
+      : Promise.reject(new Error('VOICE_STATE_UNCACHED_MEMBER'));
+  }
+
+  /**
    * Kicks the member from the voice channel.
    * @param {string} [reason] Reason for kicking member from the channel
    * @returns {Promise<GuildMember>}
--- a/src/structures/VoiceState.js
+++ b/src/structures/VoiceState.js
@@ -149,12 +149,15 @@ class VoiceState extends Base {
 
   /**
    * Mutes/unmutes and deafens/undeafens the member of this voice state.
-   * @param {boolean} mute Whether or not the member should be muted
-   * @param {boolean} deaf Whether or not the member should be deafened
+   * @param {boolean|Object} setting Whether or not the member should be muted and deafened
+   * @param {boolean} setting.mute Whether or not the member should be muted
+   * @param {boolean} setting.mute Whether or not the member should be deafened
    * @param {string} [reason] Reason for muting/unmuting and deafening/undeafening
    * @returns {Promise<GuildMember>}
    */
-  setMuteDeaf(mute, deaf, reason) {
+  setMuteDeaf(setting, reason) {
+    if (typeof setting === 'boolean') setting = { mute: setting, deaf: setting };
+    const { mute, deaf } = setting;
     return this.member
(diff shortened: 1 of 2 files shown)


Jiralite:
Personally, `<VoiceState>.setMuteDeaf()` sounds like it would alter both the mute and deaf properties from one parameter (and the second parameter being the reason). Perhaps it would be better to use an object or a boolean as the first parameter? This would allow for more flexibility:
```JavaScript
<VoiceState>.setMuteDeaf({
  mute: true,
  deaf: true
}, "reason");
```
```JavaScript
// Sets both mute and deaf properties to false
<VoiceState>.setMuteDeaf(false, "reason");
```

jftanner:
Hrm, that's a good point. I'm happy with either interface. The former is likely more useful, but the latter matches the existing helpers.

Given that there _is_ the ability to use `GuildMember.edit()` directly, and this is just a helper function, I'd lean towards the latter. But I'm happy to implement it either way you prefer.

jftanner:
I've made the requested change, so that the method accepts an object or a boolean.

I'm not 100% certain I've got the JSDoc and Types syntax correct, so it'd be worth some extra attention there in your review.

almostSouji:
Why is this necessary? The functionality is there in `<VoiceState>.member.edit({deaf: true, mute: true}, "reason")`. I don't see any benefit here in adding yet another "convenience accessor" or what one might want to call these. We end up further and further diluting the API until you can do everything from everywhere in different ways.

jftanner:
> Why is this necessary?

That's a fair question. I'll repeat my reply from the original issue:
> [...] it's not immediately obvious how to do them simultaneously. (Hence this ticket.) But I also understand the desire to keep a leaner codebase. [...] Perhaps a compromise, and just add a link to GuildMember.edit into the docs for those two helper methods?

almostSouji:
I generally don't see much purpose in all the convenience accessors we already have in the library, but giving even further (now combined) methods like this way seems a step into the wrong direction to me.

jftanner:
I think that's a fair critique. Especially with your point about `<VoiceState>.setMuteDeafChannel(true, true, channel, reason)` being the next logical conclusion.

It's those existing helpers that made it confusing in the first place. Because the `setMute` and `setDeaf` helpers exist, it didn't even occur to me to find a different mechanism in a different model. (Nor the other developer I was talking to, sparking the issue and PR.)

Given that the existing helpers can't be removed without a breaking change, would you be opposed to adding a link in the documentation to help new users find `<GuildMember>.edit`, after they find `<VoiceState>.setMute`? Ultimately, that confusion is the only thing I'm trying to fix.

monbrey:
I don't quite understand the confusion here, since the model of "helper methods" is generally very consistent across the whole library. Helper methods are for single properties. `VoiceState#setMute()`, `GuildChannel#setName()`, `GuildMember#setNickname()`.

When making changes to multiple properties, the edit method is the way to do it. While mute and deaf could be seen as a logical pairing, we definitely don't want to start making multi-helpers for every permutation and combination of possible properties.

And finally, if the syntax for a method like this is to be `<VoiceState>.setMuteDeaf({ mute: true, deaf: true });` it's just a worse version of the same edit syntax but with less properties.

jftanner:
Given the lukewarm response here, and since the functionality _is_ available (if not immediately obvious), I'm going to rescind my PR.

Thanks, everyone, for your time and consideration.

PR #4867, opened 2020-09-30
feat(Message): add messageEditHistoryMaxSize to limit stored msg edits

### LINKED ISSUE #4826: Excessive memory usage for message

First off, I'm not really sure this is actually a bug yet, but I'm opening this so I can track the digging and document things so far.

In my Grafana logging for my bot, I noticed that the memory usage for one of my shards had started to increase, out-of-line from all the other shards. Looking further, I then noticed that this seemed to match the calculated memory usage for cached messages in the shard.

image
image

Memory RSS is generated from process.memoryUsage() so should be accurate, whereas Cached Messages Memory is an approximation using [object-sizeof](https://www.npmjs.com/package/object-sizeof) with an array containing every message in cache. However, this inaccuracy shouldn't matter as this is all about comparison between other shards and other messages.

Alone, these two graphs don't say much, there are probably just more messages in that shard for one reason or another. Alas, there are not. Grafana also has a graph set up for the total number of messages in cache, using similar logic to Cached Messages Memory but without using object-sizeof, and we can see that there are a similar number of messages in the problematic shard to the other shards.

image

So, at this point, its safe to say that something is different about these messages that's causing them to use a load of extra memory. My next step in debugging what was going on was to get that largest message from each shard and just manually confirm that the messages in shard 7 were indeed much larger. This was done by executing the following script in all shards:

```js
const sizeof = require('object-sizeof');
const messages = client.guilds.cache.array().reduce((prev, guild) => prev.concat(guild.channels.cache.array().reduce((prevC, channel) => { if ('messages' in channel) { return prevC.concat(channel.messages.cache.array()); } return prevC; }, [])), []).sort((a, b) => sizeof(b) - sizeof(a));
return [messages.length, sizeof(messages[0]), sizeof(messages[messages.length-1])];
```

With the following results:

```
Shard 0 Result
2404, 3814174, 3096

Shard 1 Result
2103, 96504, 3118

Shard 2 Result
5176, 30996, 3102

Shard 3 Result
3440, 1138488, 3106

Shard 4 Result
1820, 31540, 3128

Shard 5 Result
2514, 28520, 3012

Shard 6 Result
2373, 248242, 3154

Shard 7 Result
4050, 84625818, 3130
```

This confirmed for me that the messages in shard 7, the problematic shard, were definitely much larger than those in the other, nominal shards. Again, object-sizeof likely isn't being completely accurate with the size estimations, but what matters here is that it is consistent for the comparison between shards and messages.

With this confirmed, the next step was to understand what about the messages in this problematic shard were using so much memory. Again, using object-sizeof, I got the largest messages in the shard and calculated the size for a set of attributes that each message has in Discord.js:

```js
const sizeof = require('object-sizeof');
const message = client.guilds.cache.array().reduce((prev, guild) => prev.concat(guild.channels.cache.array().reduce((prevC, channel) => { if ('messages' in channel) { return prevC.concat(channel.messages.cache.array()); } return prevC; }, [])), []).sort((a, b) => sizeof(b) - sizeof(a))[0];

return JSON.stringify(Object.keys(message).reduce((prev, key) => { prev[key] = sizeof(message[key]) / 1024 ** 2; return prev; }, {}), null, 2);
```

With the following result:

```json
{
  "channel": 0.0026187896728515625,
  "deleted": 0.000003814697265625,
  "id": 0.000034332275390625,
  "type": 0.0000133514404296875,
  "system": 0.000003814697265625,
  "content": 0,
  "author": 0.0003662109375,
  "pinned": 0.000003814697265625,
  "tts": 0.000003814697265625,
  "nonce": 0,
  "embeds": 0.0007877349853515625,
  "attachments": 0,
  "createdTimestamp": 0.00000762939453125,
  "editedTimestamp": 0.00000762939453125,
  "reactions": 88.01104545593262,
  "mentions": 0.0001068115234375,
  "webhookID": 0,
  "application": 0,
  "activity": 0,
  "_edits": 107.88620185852051,
  "flags": 0.00002288818359375,
  "reference": 0
}
```

I also ran the same script in another shard (5) where memory usage appeared far more nominal, for comparison:

```json
{
  "channel": 0.0026721954345703125,
  "deleted": 0.000003814697265625,
  "id": 0.000034332275390625,
  "type": 0.0000133514404296875,
  "system": 0.000003814697265625,
  "content": 0,
  "author": 0.0003490447998046875,
  "pinned": 0.000003814697265625,
  "tts": 0.000003814697265625,
  "nonce": 0,
  "embeds": 0.0044078826904296875,
  "attachments": 0,
  "createdTimestamp": 0.00000762939453125,
  "editedTimestamp": 0.00000762939453125,
  "reactions": 0.06316184997558594,
  "mentions": 0.0001068115234375,
  "webhookID": 0,
  "application": 0,
  "activity": 0,
  "_edits": 0.06797599792480469,
  "flags": 0.00002288818359375,
  "reference": 0
}
```

See above.

**Further details:**

- discord.js version: 12.3.1
- Node.js version: 12.18.0
- Operating system: Windows Server 2016 (deal with it)
- Priority this issue should have – please be realistic and elaborate if possible: N/A

**Relevant client options:**

- partials: none
- gateway intents:

```js
(new Intents()).add(
    Intents.FLAGS.GUILDS,
    Intents.FLAGS.GUILD_MEMBERS,
    Intents.FLAGS.GUILD_BANS,
    Intents.FLAGS.GUILD_EMOJIS,
    Intents.FLAGS.GUILD_MESSAGES,
    Intents.FLAGS.GUILD_MESSAGE_REACTIONS,
    Intents.FLAGS.DIRECT_MESSAGES,
)
```
- other:

```js
messageCacheMaxSize: 10,
messageCacheLifetime: 30 * 60,
messageSweepInterval: 5 * 60,
```

MattIPv4:

This fixes #4826 by adding a new client option to set a fixed limit on the edit history stored for each message in the client cache.

This has been tested in production on my bot (where the original memory issue was observed) for roughly a month and seems to be fine.

**Status**

### DIFF
--- a/src/client/Client.js
+++ b/src/client/Client.js
@@ -432,6 +432,13 @@ class Client extends BaseClient {
     if (typeof options.messageSweepInterval !== 'number' || isNaN(options.messageSweepInterval)) {
       throw new TypeError('CLIENT_INVALID_OPTION', 'messageSweepInterval', 'a number');
     }
+    if (
+      typeof options.messageEditHistoryMaxSize !== 'number' ||
+      isNaN(options.messageEditHistoryMaxSize) ||
+      options.messageEditHistoryMaxSize < -1
+    ) {
+      throw new TypeError('CLIENT_INVALID_OPTION', 'messageEditHistoryMaxSize', 'a number greater than or equal to -1');
+    }
     if (typeof options.fetchAllMembers !== 'boolean') {
       throw new TypeError('CLIENT_INVALID_OPTION', 'fetchAllMembers', 'a boolean');
     }
--- a/src/structures/Message.js
+++ b/src/structures/Message.js
@@ -240,8 +240,11 @@ class Message extends Base {
    * @private
    */
   patch(data) {
-    const clone = this._clone();
-    this._edits.unshift(clone);
+    const { messageEditHistoryMaxSize } = this.client.options;
+    if (messageEditHistoryMaxSize !== 0) {
+      const editsLimit = messageEditHistoryMaxSize === -1 ? Infinity : messageEditHistoryMaxSize;
+      if (this._edits.unshift(this._clone()) > editsLimit) this._edits.pop();
+    }
 
     if ('edited_timestamp' in data) this.editedTimestamp = new Date(data.edited_timestamp).getTime();
     if ('content' in data) this.content = data.content;
--- a/src/structures/Message.js
+++ b/src/structures/Message.js
@@ -235,15 +235,17 @@ class Message extends Base {
   }
 
   /**
-   * Updates the message.
+   * Updates the message and returns the old message.
    * @param {Object} data Raw Discord message update data
+   * @returns {Message}
    * @private
    */
   patch(data) {
+    const clone = this._clone();
     const { messageEditHistoryMaxSize } = this.client.options;
     if (messageEditHistoryMaxSize !== 0) {
       const editsLimit = messageEditHistoryMaxSize === -1 ? Infinity : messageEditHistoryMaxSize;
-      if (this._edits.unshift(this._clone()) > editsLimit) this._edits.pop();
+      if (this._edits.unshift(clone) > editsLimit) this._edits.pop();
     }
 
     if ('edited_timestamp' in data) this.editedTimestamp = new Date(data.edited_timestamp).getTime();
@@ -271,6 +273,8 @@ class Message extends Base {
     );
 
     this.flags = new MessageFlags('flags' in data ? data.flags : 0).freeze();
+
+    return clone;
   }
 
   /**
--- a/src/util/Constants.js
+++ b/src/util/Constants.js
@@ -19,6 +19,9 @@ const browser = (exports.browser = typeof window !== 'undefined');
  * sweepable (in seconds, 0 for forever)
  * @property {number} [messageSweepInterval=0] How frequently to remove messages from the cache that are older than
  * the message cache lifetime (in seconds, 0 for never)
+ * @property {number} [messageEditHistoryMaxSize=-1] Maximum number of previous versions to hold for an edited message
+ * (-1 or Infinity for unlimited - don't do this without message sweeping, otherwise memory usage will climb
+ * indefinitely. Setting this to 0 may have undesired effects such as messageUpdate not firing)
  * @property {boolean} [fetchAllMembers=false] Whether to cache all guild members and users upon startup, as well as
  * upon joining a guild (should be avoided whenever possible)
  * @property {DisableMentionType} [disableMentions='none'] Default value for {@link MessageOptions#disableMentions}
@@ -43,6 +46,7 @@ exports.DefaultOptions = {
   messageCacheMaxSize: 200,
   messageCacheLifetime: 0,
   messageSweepInterval: 0,
+  messageEditHistoryMaxSize: -1,
   fetchAllMembers: false,
   disableMentions: 'none',
   partials: [],
--- a/src/util/Constants.js
+++ b/src/util/Constants.js
@@ -20,8 +20,7 @@ const browser = (exports.browser = typeof window !== 'undefined');
  * @property {number} [messageSweepInterval=0] How frequently to remove messages from the cache that are older than
  * the message cache lifetime (in seconds, 0 for never)
  * @property {number} [messageEditHistoryMaxSize=-1] Maximum number of previous versions to hold for an edited message
- * (-1 or Infinity for unlimited - don't do this without message sweeping, otherwise memory usage will climb
- * indefinitely. Setting this to 0 may have undesired effects such as messageUpdate not firing)
+ * (-1 or Infinity for unlimited - don't do this without sweeping, otherwise memory usage may climb indefinitely.)
  * @property {boolean} [fetchAllMembers=false] Whether to cache all guild members and users upon startup, as well as
  * upon joining a guild (should be avoided whenever possible)
  * @property {DisableMentionType} [disableMentions='none'] Default value for {@link MessageOptions#disableMentions}
--- a/src/client/actions/MessageUpdate.js
+++ b/src/client/actions/MessageUpdate.js
@@ -9,9 +9,9 @@ class MessageUpdateAction extends Action {
       const { id, channel_id, guild_id, author, timestamp, type } = data;
       const message = this.getMessage({ id, channel_id, guild_id, author, timestamp, type }, channel);
       if (message) {
-        message.patch(data);
+        const old = message.patch(data);
         return {
-          old: message._edits[0],
+          old,
           updated: message,
         };
       }
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -977,7 +977,7 @@ declare module 'discord.js' {
   export class Message extends Base {
     constructor(client: Client, data: object, channel: TextChannel | DMChannel | NewsChannel);
     private _edits: Message[];
-    private patch(data: object): void;
+    private patch(data: object): Message;
 
     public activity: MessageActivity | null;
     public application: ClientApplication | null;
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -2317,6 +2317,7 @@ declare module 'discord.js' {
     messageCacheMaxSize?: number;
     messageCacheLifetime?: number;
     messageSweepInterval?: number;
+    messageEditHistoryMaxSize?: number;
     fetchAllMembers?: boolean;
     disableMentions?: 'none' | 'all' | 'everyone';
     allowedMentions?: MessageMentionOptions;


src/client/Client.js
```diff
@@ -432,6 +432,9 @@ class Client extends BaseClient {
     if (typeof options.messageSweepInterval !== 'number' || isNaN(options.messageSweepInterval)) {
       throw new TypeError('CLIENT_INVALID_OPTION', 'messageSweepInterval', 'a number');
     }
+    if (typeof options.messageEditHistoryMaxSize !== 'number' || isNaN(options.messageEditHistoryMaxSize)) {
+      throw new TypeError('CLIENT_INVALID_OPTION', 'messageEditHistoryMaxSize', 'a number');
+    }
```
tguichaoua, line 437:
```suggestion
    if (typeof options.messageEditHistoryMaxSize !== 'number' || isNaN(options.messageEditHistoryMaxSize) || options.messageEditHistoryMaxSize < -1) {
      throw new TypeError('CLIENT_INVALID_OPTION', 'messageEditHistoryMaxSize', 'a number');
    }
```
Number lower than -1 could mess with `Message#patch`:
```js
    if (this.client.options.messageEditHistoryMaxSize !== -1) {
      this._edits.splice(this.client.options.messageEditHistoryMaxSize);
    }
```

tguichaoua (changes requested):

src/client/Client.js
MattIPv4: Ah yeah, good call

kyranet (approved):

src/util/Constants.js
```diff
@@ -43,6 +46,7 @@ exports.DefaultOptions = {
   messageCacheMaxSize: 200,
   messageCacheLifetime: 0,
   messageSweepInterval: 0,
+  messageEditHistoryMaxSize: 100,
```
vladfrangu, line 49: Honestly... This imo can be lower - like 50, otherwise lgtm

vladfrangu (approved):

src/util/Constants.js
MattIPv4: I chose 100 to help with ensuring this wouldn't be a breaking change for 99% of use-cases. Considering we're going from Infinity to this by default, I think it'd be wise to stick with a larger number for now, and it can be reduced in the future if folks this it's appropriate?

almostSouji: introducing any sort of default here will make this major, no matter how low or high the number is

MattIPv4: Ah, fair enough. Is it best to default this to -1 one then, or would it be better to set an aggressive default and keep this as a major change?

almostSouji: can surely introduce it with no limit in minor to give people access to use it if they choose to do so by explicitly providing the option (and thus being aware of the change) and change that to be more aggressive for v13 in a separate PR.

src/structures/Message.js
```diff
@@ -242,6 +242,9 @@ class Message extends Base {
   patch(data) {
     const clone = this._clone();
     this._edits.unshift(clone);
+    if (this.client.options.messageEditHistoryMaxSize !== -1) {
+      this._edits.splice(this.client.options.messageEditHistoryMaxSize);
+    }
```
izexi, line 247: The length of the array could be checked before calling `Array#unshift()`, as opposed to unshifting the cloned message and then spilcing it after (when the limit is reached). This would also prevent `_clone()` from being called when the limit is reached.
```suggestion
    const { messageEditHistoryMaxSize } = this.client.options;
    if (messageEditHistoryMaxSize === -1 || this._edits.length < messageEditHistoryMaxSize) {
      const clone = this._clone();
      this._edits.unshift(clone);
    }
```

MattIPv4: clone is unshifted, pushed onto the head of the array. We're slicing from the end of the array. The only time this optimisation would help is when `messageEditHistoryMaxSize` is `0`, which will be pretty rare considering it has some side-effects.

src/util/Constants.js
MattIPv4: Makes sense to me, will set this to -1.

src/structures/Message.js
izexi: It would also help when the `Message#_edits` has reached the length of the set `messageEditHistoryMaxSize`. Currently, when this happens the message will be cloned and unshifted to the start of the array and then the last element would be spliced out (as you mentioned). With the suggested change, `Message#_clone()` => `Message#_edits#unshift()` => `Message#_edits#splice()` wouldn't even be called everytime the message is edited when the limit has been reached. I'm guessing that this would outweigh the cost of an extra comparison, and will most likely for cases where the `messageEditHistoryMaxSize` to a lower number.

MattIPv4: That would be absolutely breaking behaviour. If that were implemented, you'd then be in a situation where you were holding super old history and ignoring newer edits.

izexi: Yes... my apologies I somehow overlooked that, I have updated my comment's suggestion so that it actually updates with incoming new edits and pops off the last element after by checking the returned value of [`Array#unshift()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/unshift) (which is the length of the array) exceeds the `messageEditHistoryMaxSize`. Once again, I'm not fully sure whether this would be more efficient for all cases but I do believe that avoiding the `_clone()`, `unshift()` and `splice()` calls when the limit is reached (especially for lower limits) would be more optimal. Feel free to disregard this suggestion if you disagree.

MattIPv4: I guess there's also a bit of an unintended advantage to using splice vs pop here, in that if a user decided for any reason to update `messageEditHistoryMaxSize` on the fly, the splice implementation would ensure the edit history is capped to the new value, whereas pop may not if the edit history exceeds `messageEditHistoryMaxSize` and the value was decreased.

Imo, this change doesn't provide much benefit and definitely increases the code complexity. However, happy to be correct if folks with more knowledge of JS internals believe this is significantly more optimal instead of calling splice every time.

kyranet: > However, happy to be correct if folks with more knowledge of JS internals believe this is significantly more optimal instead of calling splice every time.

You called, I delivered.

There's a huge overhead on doing `splice` compared to `pop`, the performance differing by a factor of 2.35 on Node.js v14.5.0.

The reasons why are:

- [`pop()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/pop) has no arguments, it always removes one value from the array and returns it.
- [`splice()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/splice) has multiple optional arguments: `array.splice(start[, deleteCount[, item1[, item2[, ...]]]])`.
- [`splice()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/splice) can remove *and* insert, given an amount of elements to remove and the index, which can introduce extra cost.
- [`splice()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/splice) creates an array with all the removed elements, which needs it to iterate for each value in it.

The specification can be used to see how much convoluted splice() is compared to pop():

- [`Array.prototype.pop()`](https://tc39.es/ecma262/#sec-array.prototype.pop)
- [`Array.prototype.splice(start, deleteCount, ...items)`](https://tc39.es/ecma262/#sec-array.prototype.splice)

tl;dr: the former is 13 lines in the spec, the latter is 62 lines, with multiple loops and conditionals.

> **Note**: Loading the page might take a while, I suggest opening pop and then scrolling down.

kyranet: Honestly, I'd go for `pop()`, it reads better as in this context we give it the meaning that "we're pushing a new value at the start and removing the last", similar to a FILO queue. I didn't even know `splice(maxNumberOfElements)` was a valid overload, and its behaviour is not obvious for everyone.

> I guess there's also a bit of an unintended advantage to using splice vs pop here, in that if a user decided for any reason to update `messageEditHistoryMaxSize` on the fly, the splice implementation would ensure the edit history is capped to the new value, whereas pop may not if the edit history exceeds `messageEditHistoryMaxSize` and the value was decreased.

Changing the option doesn't immediately trim all the message caches either, but it will for new messages.

MattIPv4: 👍 TIL there was such a performance difference. pop does indeed make sense then. Will change :)

src/structures/Message.js
```diff
@@ -460,7 +462,10 @@ class Message extends Base {
    * }
    */
   async crosspost() {
-    await this.client.api.channels(this.channel.id).messages(this.id).crosspost.post();
+    await this.client.api
+      .channels(this.channel.id)
+      .messages(this.id)
+      .crosspost.post();
```
kyranet, line 468: Please revert those changes, it's causing lint to fail.

advaith1: I was having the same issue in my PR; looks like eslint inside VSCode automatically changes it to multiple lines and shows errors if it's one line, but the cli fails if it's one line.

MattIPv4: eslint automatically formatted this in WebStorm, wasn't a change I actively made 😂

src/structures/Message.js
```diff
@@ -241,7 +241,9 @@ class Message extends Base {
    */
   patch(data) {
     const clone = this._clone();
-    this._edits.unshift(clone);
+    const { messageEditHistoryMaxSize } = this.client.options;
+    const editsLimit = messageEditHistoryMaxSize === -1 ? Infinity : messageEditHistoryMaxSize;
+    if (this._edits.unshift(clone) > editsLimit) this._edits.pop();
```
BannerBomb, line 246: Just asking, I may be wrong about the idea, but why not skip the caching if the provided max size is `0` that way you aren't pushing the item into the array then instantly popping it back out?

src/structures/Message.js
MattIPv4: Oh wait, this happens in the pre-commit hooks...

izexi (approved):

MattIPv4:
Fixed the issue with messageUpdate not firing by changing Message#patch to return the old message clone, which can then be used when the event is fired. Method is private, so I assume this can remain minor.

SpaceEEC (approved):

src/structures/Message.js
```diff
@@ -235,15 +235,17 @@ class Message extends Base {
   }
 
   /**
-   * Updates the message.
+   * Updates the message and returns the old message.
    * @param {Object} data Raw Discord message update data
+   * @returns {Message}
```
izexi, line 240: The [typings](https://github.com/MattIPv4/discord.js/blob/limit-message-edit-history/typings/index.d.ts#L980) should be updated to reflect this

MattIPv4: TIL privates have typings

MattIPv4: (∩ ͡° ͜ʖ ͡°)⊃━☆ﾟ. * ･ ｡ﾟ,

iCrawl (approved):

PR #4869, opened 2020-09-30
feat(MessageManager): extend API coverage

monbrey:
This is to be the first of a few PRs I plan to make which adds additional methods to the Manager classes. Open to discussion and feedback (and additional testing please!) before I start submitting others.

This allows users to directly handle not only creating and fetching Messages from the Manager, but also editing, crossposting, reacting to, pinning and unpinning messages by ID, allowing management of uncached messages.

MessageManager seemed like the logical place to start with this given it already provided support for deleting by ID, and is one of the two most common areas a structure is likely to be uncached.

Like the existing delete method, these all return Promise<void> rather than returning the updated cached structure, as they're designed to compatible with uncached Messages.

**Status**

### DIFF
--- a/src/managers/MessageManager.js
+++ b/src/managers/MessageManager.js
@@ -115,7 +115,7 @@ class MessageManager extends BaseManager {
 
   /**
    * Edits a message, even if it's not cached.
-   * @param {MessageResolvable} message The message to delete
+   * @param {MessageResolvable} message The message to edit
    * @param {StringResolvable|APIMessage} [content] The new content for the message
    * @param {MessageEditOptions|MessageEmbed} [options] The options to provide
    * @returns {Promise<void>}
@@ -131,6 +131,18 @@ class MessageManager extends BaseManager {
           clone._patch(d);
         }
       });
+    }
+  }
+
+  /**
+   * Publishes a message in an announcement channel to all channels following it, even if it's not cached.
+   * @param {MessageResolvable} message The message to publish
+   * @returns {Promise<void>}
+   */
+  async crosspost(message) {
+    message = this.resolveID(message);
+    if (message) {
+      await this.client.api.channels(this.channel.id).messages(message).crosspost.post();
     }
   }
 
--- a/src/managers/MessageManager.js
+++ b/src/managers/MessageManager.js
@@ -1,6 +1,7 @@
 'use strict';
 
 const BaseManager = require('./BaseManager');
+const APIMessage = require('../structures/APIMessage');
 const Message = require('../structures/Message');
 const Collection = require('../util/Collection');
 const LimitedCollection = require('../util/LimitedCollection');
@@ -111,6 +112,27 @@ class MessageManager extends BaseManager {
    * @param {MessageResolvable} message The message resolvable to resolve
    * @returns {?Snowflake}
    */
+
+  /**
+   * Edits a message, even if it's not cached.
+   * @param {MessageResolvable} message The message to delete
+   * @param {StringResolvable|APIMessage} [content] The new content for the message
+   * @param {MessageEditOptions|MessageEmbed} [options] The options to provide
+   * @returns {Promise<void>}
+   */
+  async edit(message, content, options) {
+    message = this.resolveID(message);
+    if (message) {
+      const { data } =
+        content instanceof APIMessage ? content.resolveData() : APIMessage.create(this, content, options).resolveData();
+      await this.client.api.channels[this.channel.id].messages[message].patch({ data }).then(d => {
+        if (this.cache.has(message)) {
+          const clone = this.cache.get(message)._clone();
+          clone._patch(d);
+        }
+      });
+    }
+  }
 
   /**
    * Deletes a message, even if it's not cached.
--- a/src/managers/MessageManager.js
+++ b/src/managers/MessageManager.js
@@ -123,16 +123,15 @@ class MessageManager extends BaseManager {
    */
   async edit(message, content, options) {
     message = this.resolveID(message);
-    if (message) {
-      const { data } =
-        content instanceof APIMessage ? content.resolveData() : APIMessage.create(this, content, options).resolveData();
-      await this.client.api.channels[this.channel.id].messages[message].patch({ data }).then(d => {
-        if (this.cache.has(message)) {
-          const clone = this.cache.get(message)._clone();
-          clone._patch(d);
-        }
-      });
-    }
+    if (!message) throw new TypeError('INVALID_TYPE', 'message', 'MessageResolvable');
+    const { data } =
+      content instanceof APIMessage ? content.resolveData() : APIMessage.create(this, content, options).resolveData();
+    await this.client.api.channels[this.channel.id].messages[message].patch({ data }).then(d => {
+      if (this.cache.has(message)) {
+        const clone = this.cache.get(message)._clone();
+        clone._patch(d);
+      }
+    });
   }
 
   /**
@@ -142,9 +141,9 @@ class MessageManager extends BaseManager {
    */
   async crosspost(message) {
     message = this.resolveID(message);
-    if (message) {
-      await this.client.api.channels(this.channel.id).messages(message).crosspost.post();
-    }
+    if (!message) throw new TypeError('INVALID_TYPE', 'message', 'MessageResolvable');
+
+    await this.client.api.channels(this.channel.id).messages(message).crosspost.post();
   }
 
   /**
@@ -156,13 +155,13 @@ class MessageManager extends BaseManager {
    */
   async pin(message, options) {
     message = this.resolveID(message);
-    if (message) {
-      await this.client.api
-        .channels(this.channel.id)
-        .pins(message)
-        .put(options)
-        .then(() => this);
-    }
+    if (!message) throw new TypeError('INVALID_TYPE', 'message', 'MessageResolvable');
+
+    await this.client.api
+      .channels(this.channel.id)
+      .pins(message)
+      .put(options)
+      .then(() => this);
   }
 
   /**
@@ -174,13 +173,13 @@ class MessageManager extends BaseManager {
    */
   async unpin(message, options) {
     message = this.resolveID(message);
-    if (message) {
-      await this.client.api
-        .channels(this.channel.id)
-        .pins(message)
-        .delete(options)
-        .then(() => this);
-    }
+    if (!message) throw new TypeError('INVALID_TYPE', 'message', 'MessageResolvable');
+
+    await this.client.api
+      .channels(this.channel.id)
+      .pins(message)
+      .delete(options)
+      .then(() => this);
   }
 
   /**
@@ -208,9 +207,9 @@ class MessageManager extends BaseManager {
    */
   async delete(message, reason) {
     message = this.resolveID(message);
-    if (message) {
-      await this.client.api.channels(this.channel.id).messages(message).delete({ reason });
-    }
+    if (!message) throw new TypeError('INVALID_TYPE', 'message', 'MessageResolvable');
+
+    await this.client.api.channels(this.channel.id).messages(message).delete({ reason });
   }
 
   async _fetchId(messageID, cache, force) {
--- a/src/managers/MessageManager.js
+++ b/src/managers/MessageManager.js
@@ -165,6 +165,24 @@ class MessageManager extends BaseManager {
   }
 
   /**
+   * Unins a message from the channel's pinned messages, even if it's not cached.
+   * @param {MessageResolvable} message The message to unpin
+   * @param {Object} [options] Options for unpinning
+   * @param {string} [options.reason] Reason for unpinning
+   * @returns {Promise<void>}
+   */
+  async unpin(message, options) {
+    message = this.resolveID(message);
+    if (message) {
+      await this.client.api
+        .channels(this.channel.id)
+        .pins(message)
+        .delete(options)
+        .then(() => this);
+    }
+  }
+
+  /**
    * Deletes a message, even if it's not cached.
    * @param {MessageResolvable} message The message to delete
    * @param {string} [reason] Reason for deleting this message, if it does not belong to the client user
--- a/src/managers/MessageManager.js
+++ b/src/managers/MessageManager.js
@@ -147,6 +147,24 @@ class MessageManager extends BaseManager {
   }
 
   /**
+   * Pins a message to the channel's pinned messages, even if it's not cached.
+   * @param {MessageResolvable} message The message to pin
+   * @param {Object} [options] Options for pinning
+   * @param {string} [options.reason] Reason for pinning
+   * @returns {Promise<void>}
+   */
+  async pin(message, options) {
+    message = this.resolveID(message);
+    if (message) {
+      await this.client.api
+        .channels(this.channel.id)
+        .pins(message)
+        .put(options)
+        .then(() => this);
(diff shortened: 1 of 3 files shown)


NotSugden (commented):
just another nit, would it be better to use async functions and await the promises, rather than `.then(() => this)`?

monbrey:
> just another nit, would it be better to use async functions and await the promises, rather than `.then(() => this)`?

Yeah I have no idea what's "better" here - I followed the convention which was already in place from the existing `MessageManager#delete()`. I have no preference for either and will take advice from maintainers.

src/managers/MessageManager.js
```diff
@@ -113,6 +114,95 @@ class MessageManager extends BaseManager {
...
+   */
+  async pin(message, options) {
+    message = this.resolveID(message);
+    if (!message) throw new TypeError('INVALID_TYPE', 'message', 'MessageResolvable');
+
+    await this.client.api
+      .channels(this.channel.id)
+      .pins(message)
+      .put(options)
+      .then(() => this);
```
izexi, line 168: The `.then(...)` here is redundant.
```suggestion
      .put(options);
```

src/managers/MessageManager.js
```diff
@@ -113,6 +114,95 @@ class MessageManager extends BaseManager {
...
+   */
+  async unpin(message, options) {
+    message = this.resolveID(message);
+    if (!message) throw new TypeError('INVALID_TYPE', 'message', 'MessageResolvable');
+
+    await this.client.api
+      .channels(this.channel.id)
+      .pins(message)
+      .delete(options)
+      .then(() => this);
```
izexi, line 186: Same as above.
```suggestion
      .delete(options);
```

src/structures/Message.js
```diff
@@ -419,7 +419,7 @@ class Message extends Base {
   }
 
   /**
-   * Options that can be passed into editMessage.
+   * Options that can be passed into Message#edit.
```
izexi, line 422: A link can be used here to help navigate to the `Message#edit` options on the docs.
```suggestion
   * Options that can be passed into {@link Message#edit}.
```

typings/index.d.ts
```diff
@@ -1982,14 +1982,27 @@ declare module 'discord.js' {
     constructor(channel: TextChannel | DMChannel, iterable?: Iterable<any>);
     public channel: TextBasedChannelFields;
     public cache: Collection<Snowflake, Message>;
+    public crosspost(message: MessageResolvable): Promise<void>;
```
izexi, line 1985: This following suggestions rely on the [`APIMessage`](https://github.com/discordjs/discord.js/pull/4876/files#diff-494415c3af35245ecae73f5d42d8b331R2194-R2281) interface over on #4876.
```suggestion
    public crosspost(message: MessageResolvable): Promise<Message | APIRawMessage>;
```

typings/index.d.ts
```diff
@@ -1982,14 +1982,27 @@ declare module 'discord.js' {
...
+    public delete(message: MessageResolvable, reason?: string): Promise<void>;
+    public edit(
+      message: MessageResolvable,
+      content: APIMessageContentResolvable | MessageEditOptions | MessageEmbed | APIMessage,
+    ): Promise<void>;
```
izexi, line 1990:
```suggestion
    ): Promise<Message | APIRawMessage>;
```

typings/index.d.ts
```diff
@@ -1982,14 +1982,27 @@ declare module 'discord.js' {
...
+    public edit(
+      message: MessageResolvable,
+      content: StringResolvable,
+      options: MessageEditOptions | MessageEmbed,
+    ): Promise<void>;
```
izexi, line 1995:
```suggestion
    ): Promise<Message | APIRawMessage>;
```

izexi (commented):
On the [Edit Message](https://discord.com/developers/docs/resources/channel#edit-message) and [Crosspost Message](https://discord.com/developers/docs/resources/channel#crosspost-message) endpoints the [message](https://discord.com/developers/docs/resources/channel#message-object) is returned as a response on success, it could be be useful for developers to have the according methods to return the raw message data or the [`Message`](https://discord.js.org/#/docs/main/stable/class/Message) itself. This would especially benefit those that don't have such messages cached and may need to access the data after calling the endpoint.

monbrey:
@izexi Thanks for the review! Before I merge those suggestions though there's a few things to discuss.

- Is having inconsistent return values a good thing? I don't really have a side on that.
- Can the `APIRawMessage` data be used to cache a `Message`, and if so should it be in order to resolve the above inconsistency?
- Shouldn't we also have a jsdoc typedef for `APIRawMessage` rather than defining those as `@returns {Message|Object}`

izexi:
@monbrey 
> * Is having inconsistent return values a good thing? I don't really have a side on that.

I don't really have a side on this too, I simply see it as making use of the data returned from the API which is would be a lot more useful than `void` to developers. If we're talking about the union return type, yes this may be a bit of a hassle for Typescript devs since they will need to add in that casting if they're going to make use of the returned value but I don't see this as a huge concern.

---

> * Can the `APIRawMessage` data be used to cache a `Message`, and if so should it be in order to resolve the above inconsistency?

I agree that would be a good idea to solve the inconsistency, but I feel like that somewhat defeats the purpose of attempting to separate the cache from the managers if the data would just be cached within the method itself. Although its not very intuitive if the developer wanted to cache the message that wasn't already cached after calling the method from the manager, they could call `MessageManager#add` passing the returned raw message data to cache the `Message`.

---

> * Shouldn't we also have a jsdoc typedef for `APIRawMessage` rather than defining those as `@returns {Message|Object}`

Yes, that would be ideal but in other places objects are just vaguely typed as `Object` in the JSDoc and it's going to be a bit of a pain to document a `@typedef` for the `APIRawMessage` interface since it has quite a lot of properties and nested properties so that might be a bit overkill.

src/managers/MessageManager.js
```diff
@@ -113,6 +114,95 @@ class MessageManager extends BaseManager {
...
+  }
+
+  /**
+   * Adds a reaction to the message, even if it's not cached.
```
advaith1, line 190:
```suggestion
   * Adds a reaction to a message, even if it's not cached.
```

typings/index.d.ts
kyranet: If this is done, I think it'll be suitable to add an overload, where passing a `Message` instance makes it always return a `Promise<Message>`.

advaith1:
why isn't it imported from `discord-api-types`? iirc this was discussed in another pr but idr what happened

monbrey:
> why isn't it imported from `discord-api-types`?

I think because that's not a dependency of discord.js yet?

src/structures/Message.js
```diff
@@ -439,13 +439,7 @@ class Message extends Base {
...
    */
   edit(content, options) {
-    const { data } =
-      content instanceof APIMessage ? content.resolveData() : APIMessage.create(this, content, options).resolveData();
-    return this.client.api.channels[this.channel.id].messages[this.id].patch({ data }).then(d => {
-      const clone = this._clone();
-      clone._patch(d);
-      return clone;
-    });
+    return this.channel.messages.edit(this.id, content, options).then(() => this);
```
vladfrangu, line 442: The `then` will cause a different output than before (patched clone vs now the same message without the edits

typings/index.d.ts
```diff
@@ -2184,6 +2207,95 @@ declare module 'discord.js' {
 
   type APIMessageContentResolvable = string | number | boolean | bigint | symbol | readonly StringResolvable[];
 
+  interface APIRawMessage {
```
vladfrangu, line 2210: This interface is incomplete.. why don't we use `discord-api-types`? I also don't remember the discussion for that

vladfrangu (changes requested):

typings/index.d.ts
monbrey: I'm happy to use it, just not sure if this PR should be covering that. I copied the interface from the PR izeki referenced in his comments.

src/structures/Message.js
monbrey: Yeah I gotcha now, will resolve this.

SpaceEEC (changes requested):
Also, couldn't `MessageManager#edit` and `MessageManager#crosspost` return message instances instead of objects? Discord seems to respond with a message there according to the docs.

typings/index.d.ts
kyranet: It should use `discord-api-types`, we're within semver-major (v13) changes, and we'll eventually use the library for REST's routes.

src/managers/MessageManager.js
```diff
@@ -113,6 +114,91 @@ class MessageManager extends BaseManager {
...
+      .patch({ data })
+      .then(d => {
+        const existing = this.cache.get(message);
+        if (existing) {
+          const clone = existing._clone();
+          clone._patch(d);
+          return clone;
+        }
+        return this.add(d);
+      });
```
SpaceEEC, line 141: Might as well `await` this. 👀 
```suggestion
    const d = await this.client.api
      .channels(this.channel.id)
      .messages(message)
      .patch({ data });
    const existing = this.cache.get(message);
    if (existing) {
      const clone = existing._clone();
      clone._patch(d);
      return clone;
    }
    return this.add(d);
```

SpaceEEC (changes requested):

iShibi:
Will `MessageManager#reply` be a good addition to this API coverage campaign? We can finally make some good use of that `replyTo` field in `MessageOptions` 😅

advaith1:
you can just use Channel#send though?

iShibi:
Yeah, totally forgot about that :) 👍

iCrawl (approved):

src/structures/Message.js
```diff
@@ -559,23 +544,15 @@ class Message extends Base {
...
-      );
+    return this.channel.messages.react(this.id, emoji).then(
+      () =>
+        this.client.actions.MessageReactionAdd.handle({
+          user: this.client.user,
+          channel: this.channel,
+          message: this,
+          emoji: Util.parseEmoji(emoji),
+        }).reaction,
+    );
```
kyranet, line 555: This should use async/await.

src/managers/MessageManager.js
```diff
@@ -113,6 +114,87 @@ class MessageManager extends BaseManager {
...
+   * @returns {Promise<void>}
+   */
+  async react(message, emoji) {
+    message = this.resolveID(message);
+    if (!message) throw new TypeError('INVALID_TYPE', 'message', 'MessageResolvable');
+
+    emoji = this.client.emojis.resolveIdentifier(emoji);
+    if (!emoji) throw new TypeError('EMOJI_TYPE', 'emoji', 'EmojiIdentifierResolvable');
+
+    // eslint-disable-next-line newline-per-chained-call
```
kyranet, line 194: Any specific reason to ignore a style change? We want to keep everything consistent across the library, so ignoring style rules doesn't seem like a good idea.

cc: @iCrawl

typings/index.d.ts
```diff
@@ -1977,14 +1977,26 @@ declare module 'discord.js' {
...
+    public crosspost(message: MessageResolvable): Promise<Message>;
+    public delete(message: MessageResolvable): Promise<void>;
+    public edit(
+      ...args:
+        | [
+            message: MessageResolvable,
+            content: APIMessageContentResolvable | MessageEditOptions | MessageEmbed | APIMessage,
+          ]
+        | [message: MessageResolvable, content: StringResolvable, options: MessageEditOptions | MessageEmbed]
+    ): Promise<Message>;
```
kyranet, line 1989: Why not make those 2 separate overloads, instead of one using rest and tuple syntax?

kyranet (changes requested):

src/managers/MessageManager.js
monbrey: This particular line was having a format clash between eslint and prettier

typings/index.d.ts
monbrey: Honestly, this came from 7bf12fb and I don't at all remember why

kyranet (approved):

src/managers/MessageManager.js
```diff
@@ -113,6 +114,87 @@ class MessageManager extends BaseManager {
...
+   * @param {MessageResolvable} message The message to pin
+   * @param {Object} [options] Options for pinning
+   * @param {string} [options.reason] Reason for pinning
+   * @returns {Promise<void>}
+   */
+  async pin(message, options) {
+    message = this.resolveID(message);
+    if (!message) throw new TypeError('INVALID_TYPE', 'message', 'MessageResolvable');
+
+    await this.client.api.channels(this.channel.id).pins(message).put(options);
```
vladfrangu, line 164: This... leaving aside the fact idk if reason is supported, you're just passing the options *directly*. That means we can pass in ANY option, like query, headers, data, etc... Might wanna limit this

src/managers/MessageManager.js
```diff
@@ -113,6 +114,87 @@ class MessageManager extends BaseManager {
...
+   * @param {MessageResolvable} message The message to unpin
+   * @param {Object} [options] Options for unpinning
+   * @param {string} [options.reason] Reason for unpinning
+   * @returns {Promise<void>}
+   */
+  async unpin(message, options) {
+    message = this.resolveID(message);
+    if (!message) throw new TypeError('INVALID_TYPE', 'message', 'MessageResolvable');
+
+    await this.client.api.channels(this.channel.id).pins(message).delete(options);
```
vladfrangu, line 178: ^

vladfrangu (changes requested):

src/managers/MessageManager.js
monbrey: Tested and reason isn't supported, fixed by dropping options entirely

src/managers/MessageManager.js
```diff
@@ -113,6 +114,83 @@ class MessageManager extends BaseManager {
...
    */
 
+  /**
+   * Edits a message, even if it's not cached.
+   * @param {MessageResolvable} message The message to edit
+   * @param {StringResolvable|APIMessage} [content] The new content for the message
+   * @param {MessageEditOptions|MessageEmbed} [options] The options to provide
+   * @returns {Promise<Message>}
+   */
+  async edit(message, content, options) {
```
vladfrangu, line 124: Just out of curiosity, if I do `manager.edit('123', { embed: { description: 'Hello world' } })`, what would happen? Will it ~~blend~~ break? If so, shouldn't we instead have managers take in all options in an object?

monbrey: Mobile right now so I cant test directly - looking at the source, this should make it's way through the same path as `Message.edit` and `TextChannel.send`.

That is, `APIMessage.create` to `APIMessage.transformOptions` where it does the `typeof content === object` check.

No preference if we keep this or move to single object as you suggested.

vladfrangu: CC @iCrawl @kyranet ^

kyranet: I personally prefer a single object, it makes a distinction of the content with the ID, which are not to be confused, take:

```js
messages.edit('841298235918516286', { embed: { description: 'Hello world' } });
```

Versus the direct approach:
```js
message.edit({ embed: { description: 'Hello world' } });
```

For somebody who's used to the latter, using the former might confuse them because of the extra argument.

PR #4870, opened 2020-10-01
feat(Message): replace reply functionality with inline replies

almostSouji:

| ⛔ on hold, waiting for implementation of Message#inlineReply. Will move the implementation to Message#reply while also removing Message#inlineReply after merge|
| --- |

<details>
<summary>original:</summary>

This PR removes Message#reply from the library as it is just a convenience shortcut for
```js
<Message>.channel.send(`${<Message>.author}`, content)
```

Before it was argued that the added benefit of adding the reply recipient as parsed mention to the respective options was indeed a considerable ease of use compared to the #send call with all required options to make this work, however Discord has recently had previews for inline replies in development client and rumors of a new message type (19) being the proper implementation after the initial tests have reached us.

While this is not yet officially confirmed in any discord API documentation cleaning the namespace to provide actual reply functionality (should it be available for bots in the first place) or keeping it clear due to the associated meaning of the term "reply" brought forward by aforementioned functionality.

The implementation of the feature itself can then take place in a minor release, while the clearing of the name space (this PR) obviously requires a semver: major, thus me making this PR now, aiming it for v13 together with using the new API versions.

- remove the method from the code base
- remove from APIMessage options
- remove from typings
- remove from boilerplate
- remove from examples
- remove from test files

Should I have missed any caveat please do let me know!

</details>

**Status**

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,7 +76,7 @@ client.on('ready', () => {
 
 client.on('message', msg => {
   if (msg.content === 'ping') {
-    msg.reply('pong');
+    msg.channel.send('pong');
   }
 });
 
--- a/docs/examples/avatars.js
+++ b/docs/examples/avatars.js
@@ -23,7 +23,7 @@ client.on('message', message => {
   // If the message is "what is my avatar"
   if (message.content === 'what is my avatar') {
     // Send the user's avatar URL
-    message.reply(message.author.displayAvatarURL());
+    message.channel.send(message.author.displayAvatarURL());
   }
 });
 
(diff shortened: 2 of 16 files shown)


almostSouji:
This PR just has expanded to also include the inline reply functionality https://github.com/discord/discord-api-docs/pull/2118

We will aim to bring this feature to version 12 in the next minor release as Message#inlineReply (@monbrey has volunteered for this task). Version 13 (this PR, major) will move it to Message#reply once the implementation is set in monbreys PR, putting the implementation part of this on hold.

almostSouji:
Designing around making this compatible with current API functionality in place is a bigger issue than anticipated. https://github.com/discordjs/discord.js/pull/4874 will become major (v13) and include breaking changes instead, fully incorporating this PR.

PR #4872, opened 2020-10-01
feat(GuildMemberManager): extend API coverage

monbrey:
API Coverage part 2: Electric Boogaloo

Refer to #4869 (nice) for the logic behind these changes

**Status**


src/managers/GuildMemberManager.js
```diff
@@ -135,6 +136,48 @@ class GuildMemberManager extends BaseManager {
...
 
+  /**
+   * Edits a member of the guild.
+   * <warn> The user must be a member of the guild </warn>
+   * @param {UserResolvable} user The member to edit
+   * @param {GuildMemberEditData} data The data to edit the member with
+   * @param {string} [reason] Reason for editing this user
+   * @returns {Promise<void>}
+   */
+  async edit(user, data, reason) {
```
NotSugden, line 147: would it make more sense to name this parameter 'member'? since a member is being edited, the same for the other methods

src/managers/GuildMemberManager.js
```diff
@@ -135,6 +136,48 @@ class GuildMemberManager extends BaseManager {
     return this._fetchMany(options);
   }
 
+  /**
+   * Edits a member of the guild.
+   * <warn> The user must be a member of the guild </warn>
```
NotSugden, line 141:
```suggestion
   * <warn>The user must be a member of the guild</warn>
```

src/managers/GuildMemberManager.js
```diff
@@ -191,6 +234,39 @@ class GuildMemberManager extends BaseManager {
     return endpoint.post({ data: body, reason }).then(data => data.pruned);
   }
 
+  /**
+   * Kicks a user from the guild.
+   * <warn> The user must be a member of the guild </warn>
```
NotSugden, line 239:
```suggestion
   * <warn>The user must be a member of the guild</warn>
```

NotSugden (changes requested):

src/managers/GuildMemberManager.js
monbrey: I considered this, but the param being a `UserResolvable` felt awkward with it named `member` so I wasn't sure on that

kyranet (approved):

src/managers/GuildMemberManager.js
```diff
@@ -135,6 +136,48 @@ class GuildMemberManager extends BaseManager {
     return this._fetchMany(options);
   }
 
+  /**
+   * Edits a member of the guild.
+   * <warn>The user must be a member of the guild</warn>
```
vladfrangu, line 141: I don't think this needs to be a warn, I think an info should be more than enough

src/managers/GuildMemberManager.js
```diff
@@ -135,6 +136,48 @@ class GuildMemberManager extends BaseManager {
...
+   * @param {UserResolvable} user The member to edit
+   * @param {GuildMemberEditData} data The data to edit the member with
+   * @param {string} [reason] Reason for editing this user
+   * @returns {Promise<void>}
+   */
+  async edit(user, data, reason) {
+    const id = this.client.users.resolveID(user);
+    if (!id) throw new TypeError('INVALID_TYPE', 'user', 'UserResolvable');
+
+    if (data.channel) {
```
vladfrangu, line 151: This should really not modify the input object as if I pass a frozen data object, it'll throw an error. You probably want to create an _data local obj and add props to it from the input

src/managers/GuildMemberManager.js
```diff
@@ -135,6 +136,48 @@ class GuildMemberManager extends BaseManager {
...
+    let endpoint = this.client.api.guilds(this.guild.id);
+    if (id === this.client.user.id) {
+      const keys = Object.keys(data);
+      if (keys.length === 1 && keys[0] === 'nick') endpoint = endpoint.members('@me').nick;
+      else endpoint = endpoint.members(id);
+    } else {
+      endpoint = endpoint.members(id);
+    }
+    await endpoint.patch({ data, reason }).then(d => {
+      if (this.cache.has(id)) {
```
vladfrangu, line 172: Shouldn't we cache with the new data we get from the member? Assuming, of course, if we get any data..

src/managers/GuildMemberManager.js
```diff
@@ -191,6 +234,39 @@ class GuildMemberManager extends BaseManager {
     return endpoint.post({ data: body, reason }).then(data => data.pruned);
   }
 
+  /**
+   * Kicks a user from the guild.
+   * <warn>The user must be a member of the guild</warn>
```
vladfrangu, line 239: Ditto, could be an info

vladfrangu (changes requested):

src/managers/GuildMemberManager.js
monbrey: Valid point - I'll check if this would be possible. The idea was that when operating on an uncached instance it would remain uncached, but if patch returns a complete member object to us I agree we should add to cache. Will test and implement if so.

src/managers/GuildMemberManager.js
monbrey: Are you referring to the way `data` is modified by resolving variables? While I'd technically agree with the princples you're brining up here, that part of the code is directly lifted from the existing `GuildMember#edit` method. Modifying data objects is done pretty extensively through Discord.js, we don't really follow principle of not mutating data well at all.

Having said that, not really an excuse not to start! I can definitely incorporate this.

kyranet: We can slowly fix that and remove all mutations on the library, or we can do that when we switch to TypeScript.

src/managers/GuildMemberManager.js
monbrey: > PATCH/guilds/{guild.id}/members/{user.id}
Modify attributes of a guild member. Returns a 204 empty response on success.

We do not, so no change here.

src/managers/GuildMemberManager.js
```diff
@@ -135,6 +136,50 @@ class GuildMemberManager extends BaseManager {
...
+      else endpoint = endpoint.members(id);
+    } else {
+      endpoint = endpoint.members(id);
+    }
+    await endpoint.patch({ _data, reason }).then(d => {
+      if (this.cache.has(id)) {
+        const member = this.cache.get(id);
+        const clone = member._clone();
+        d.user = member.user;
+        clone._patch(d);
```
vladfrangu, line 178: So if based on the above discussion we don't get any data, why patching with.. no data?

vladfrangu (changes requested):

src/managers/GuildMemberManager.js
```diff
@@ -135,6 +136,50 @@ class GuildMemberManager extends BaseManager {
...
+    if (_data.roles) _data.roles = _data.roles.map(role => (role instanceof Role ? role.id : role));
+    let endpoint = this.client.api.guilds(this.guild.id);
+    if (id === this.client.user.id) {
+      const keys = Object.keys(_data);
+      if (keys.length === 1 && keys[0] === 'nick') endpoint = endpoint.members('@me').nick;
+      else endpoint = endpoint.members(id);
+    } else {
+      endpoint = endpoint.members(id);
+    }
+    await endpoint.patch({ _data, reason }).then(d => {
```
kyranet, line 173:
```suggestion
    await endpoint.patch({ data: _data, reason }).then(d => {
```

kyranet (dismissed):

src/managers/GuildMemberManager.js
monbrey: Looked into the original again - it actually patches the GuildMember with the _data object, rather than any returned value.

src/managers/GuildMemberManager.js
advaith1: I can see reasons for user but imo member is better, because it is acting on the member, even though it's being found from its user id/object; why can't this be a GuildMemberResolvable though?

monbrey: The GuildMemberResolvable resolution requires the ID to exist in cache, which breaks the uncached support this method is going for.

src/managers/GuildMemberManager.js
```diff
@@ -135,6 +136,50 @@ class GuildMemberManager extends BaseManager {
...
+    } else {
+      endpoint = endpoint.members(id);
+    }
+    await endpoint.patch({ data: _data, reason });
+
+    if (this.cache.has(id)) {
+      const member = this.cache.get(id);
+      const clone = member._clone();
+      _data.user = member.user;
+      clone._patch(_data);
```
vladfrangu, line 179: You..want to return the clone too... Otherwise this code block is useless

vladfrangu (dismissed):

src/managers/GuildMemberManager.js
monbrey: This is one of those areas which I revised on the MessageManager PR - debating whether these methods should *sometimes* return an object if one was cached, and return raw API data or null in other cases.

Thoughts?

vladfrangu: It would be a breaking change to return raw api data... That, and in this endpoints case we get no data apparently? Soo..yeahh idk, it's up to discussion

monbrey: This is a brand new method so it can return whatever we like I would think?

vladfrangu: Yes, however you use the method in GuildMember which would change return types and make it a breaking PR, unless the aforementioned method returns the previous data

monbrey: The `GuildMember#edit` method doesn't rely on the return value of this - it just assumes that the Manager will have updated the cache if it's available, then returns `this`.

https://github.com/discordjs/discord.js/blob/b0c8f740fd97bb0550d4cf623b09748b519861ce/src/structures/GuildMember.js#L292

src/managers/GuildMemberManager.js
```diff
@@ -191,6 +236,39 @@ class GuildMemberManager extends BaseManager {
...
+    const id = this.client.users.resolveID(user);
+    if (!id) return Promise.reject(new TypeError('INVALID_TYPE', 'user', 'UserResolvable'));
+
+    return this.client.api
+      .guilds(this.guild.id)
+      .members(id)
+      .delete({ reason })
+      .then(() => {
+        if (user instanceof GuildMember) return user;
+        const _user = this.client.users.resolve(id);
```
NotSugden, line 263: since `id` is an ID here, would it be better to use `this.client.users.cache.get`?

src/managers/GuildMemberManager.js
```diff
@@ -191,6 +236,39 @@ class GuildMemberManager extends BaseManager {
...
+        if (_user) {
+          const member = this.resolve(_user);
+          return member || _user;
```
NotSugden, line 266:
```suggestion
          return this.resolve(_user) || _user;
```
seems a bit pointless making a variable here since its just going to be returned on the next line,

src/managers/GuildMemberManager.js
```diff
@@ -135,6 +136,50 @@ class GuildMemberManager extends BaseManager {
...
+    // Clone the data object for immutability
+    const _data = { ...data };
+    if (_data.channel) {
+      _data.channel = this.guild.channels.resolve(_data.channel);
+      if (!_data.channel || _data.channel.type !== 'voice') {
+        throw new Error('GUILD_VOICE_CHANNEL_RESOLVE');
+      }
+      _data.channel_id = _data.channel.id;
+      _data.channel = undefined;
+    } else if (data.channel === null) {
```
SpaceEEC, line 160: Shouldn't this also use `_data`?
Or maybe just reassign `{ ...data }` to `data` since I do not think we need a reference to the original object.

src/managers/GuildMemberManager.js
SpaceEEC: This method won't update the cache however, only the clone, which is currently not being returned.

It however might be worth a consideration to actually update the cached entity over a clone since not everybody has access to / enabled the guild members intent to receive that update to actually patch it.

src/managers/GuildMemberManager.js
monbrey: Why does it matter exactly?

src/managers/GuildMemberManager.js
monbrey: This contradicts Vlad's earlier review: https://github.com/discordjs/discord.js/pull/4872#discussion_r498685807

src/managers/GuildMemberManager.js
NotSugden: just seems a bit silly to use `resolve` with lots of useless checks like
`x instanceof GuildMember` and `x instanceof Message` and `x instanceof User` before it tries to get the user directly from the cache

src/managers/GuildMemberManager.js
monbrey: Do we want to update the cached entity with potentially unreliable user input? The API method returns a 204 Empty response on success, so we don't have any Discord-validated data to update with.

src/managers/GuildMemberManager.js
```diff
@@ -194,6 +238,38 @@ class GuildMemberManager extends BaseManager {
...
+      .delete({ reason })
+      .then(() => {
+        if (user instanceof GuildMember) return user;
+        const _user = this.client.users.cache.get(id);
+        if (_user) {
+          return this.resolve(_user) || _user;
+        }
+        return id;
+      });
+  }
```
kyranet, line 271: Use async/await

kyranet (changes requested):

src/managers/GuildMemberManager.js
```diff
@@ -136,6 +137,49 @@ class GuildMemberManager extends BaseManager {
...
+    if (_data.roles) _data.roles = _data.roles.map(role => (role instanceof Role ? role.id : role));
+    let endpoint = this.client.api.guilds(this.guild.id);
+    if (id === this.client.user.id) {
+      const keys = Object.keys(_data);
+      if (keys.length === 1 && keys[0] === 'nick') endpoint = endpoint.members('@me').nick;
+      else endpoint = endpoint.members(id);
+    } else {
+      endpoint = endpoint.members(id);
+    }
+    await endpoint.patch({ data: _data, reason });
```
vladfrangu, line 174: With api v8, this returns a member object. We should probably use it? 👀

Cc @kyranet @iCrawl

src/managers/GuildMemberManager.js
```diff
@@ -136,6 +137,49 @@ class GuildMemberManager extends BaseManager {
...
+
+    if (this.cache.has(id)) {
+      const _member = this.cache.get(id);
```
vladfrangu, line 177: Replace .has/.get combo with .get and non-null check

vladfrangu (changes requested):

src/managers/GuildMemberManager.js
```diff
@@ -209,23 +277,21 @@ class GuildMemberManager extends BaseManager {
...
-        return id;
-      });
+
+    await this.client.api.guilds(this.guild.id).bans[id].put({ data: options });
+
+    if (user instanceof GuildMember) return user;
+    const _user = this.client.users.resolve(id);
+    if (_user) {
+      const member = this.resolve(_user);
+      return member || _user;
```
SpaceEEC, line 292:
```suggestion
      return this.resolve(_user) || _user;
```

src/structures/GuildMember.js
```diff
@@ -281,33 +280,8 @@ class GuildMember extends Base {
...
-      endpoint = endpoint.members(this.id);
-    }
-    await endpoint.patch({ data, reason });
-
-    const clone = this._clone();
-    data.user = this.user;
-    clone._patch(data);
-    return clone;
+  edit(data, reason) {
+    return this.guild.members.edit(this, data, reason).then(() => this);
```
SpaceEEC, line 284: Why is this now returning `this` and no longer the patched clone?

src/managers/GuildMemberManager.js
```diff
@@ -136,6 +137,45 @@ class GuildMemberManager extends BaseManager {
...
+    // Clone the data object for immutability
+    const _data = { ...data };
+    if (_data.channel) {
+      _data.channel = this.guild.channels.resolve(_data.channel);
+      if (!_data.channel || _data.channel.type !== 'voice') {
+        throw new Error('GUILD_VOICE_CHANNEL_RESOLVE');
+      }
+      _data.channel_id = _data.channel.id;
+      _data.channel = undefined;
+    } else if (data.channel === null) {
```
SpaceEEC, line 161: Why is this `data.channel === null` and not `_data.channel === null`?

src/managers/GuildMemberManager.js
```diff
@@ -136,6 +137,45 @@ class GuildMemberManager extends BaseManager {
...
+    if (id === this.client.user.id) {
+      const keys = Object.keys(_data);
+      if (keys.length === 1 && keys[0] === 'nick') endpoint = endpoint.members('@me').nick;
+      else endpoint = endpoint.members(id);
+    } else {
+      endpoint = endpoint.members(id);
+    }
+    const d = await endpoint.patch({ data: _data, reason });
+
+    return this.cache.get(id)?._clone().patch(d) ?? this.add(d, false);
```
SpaceEEC, line 176: `GuildMember#_patch` does not return anything.

```suggestion
    const clone = this.cache.get(id)?._clone();
    clone?.patch(d);
    return clone ?? this.add(d, false);
```

Alternatively you could make `_patch` return `this`.
(Only 2 of 31 of your classes seem to do that though, not that I see an issue with that)

SpaceEEC (changes requested):

src/managers/GuildMemberManager.js
kyranet: Using `??` would be better, though:
```suggestion
      return this.resolve(_user) ?? _user;
```

src/managers/GuildMemberManager.js
monbrey: Fixed in #bbfd002

src/managers/GuildMemberManager.js
monbrey: Oversight when I changed to non-mutating, fixed in #bbfd002

src/structures/GuildMember.js
monbrey: No idea, fixed in #bbfd002

src/managers/GuildMemberManager.js
monbrey: Fixed in #bbfd002

src/managers/GuildMemberManager.js
```diff
@@ -194,6 +236,34 @@ class GuildMemberManager extends BaseManager {
...
+  async kick(user, reason) {
+    const id = this.client.users.resolveID(user);
+    if (!id) return Promise.reject(new TypeError('INVALID_TYPE', 'user', 'UserResolvable'));
+
+    await this.client.api.guilds(this.guild.id).members(id).delete({ reason });
+
+    if (user instanceof GuildMember) return user;
+    const _user = this.client.users.cache.get(id);
+    if (_user) {
+      return this.resolve(_user) || _user;
```
kyranet, line 262:
```suggestion
      return this.resolve(_user) ?? _user;
```

kyranet (changes requested):

src/managers/GuildMemberManager.js
```diff
@@ -194,6 +236,34 @@ class GuildMemberManager extends BaseManager {
...
+    if (!id) return Promise.reject(new TypeError('INVALID_TYPE', 'user', 'UserResolvable'));
+
+    await this.client.api.guilds(this.guild.id).members(id).delete({ reason });
+
+    if (user instanceof GuildMember) return user;
+    const _user = this.client.users.cache.get(id);
+    if (_user) {
+      return this.resolve(_user) ?? _user;
+    }
+    return id;
```
kyranet, line 264:
```suggestion
    return this.resolve(user) ?? this.client.users.resolve(user) ?? id;
```

src/managers/GuildMemberManager.js
```diff
@@ -209,23 +279,20 @@ class GuildMemberManager extends BaseManager {
...
-      });
+
+    await this.client.api.guilds(this.guild.id).bans[id].put({ data: options });
+
+    if (user instanceof GuildMember) return user;
+    const _user = this.client.users.resolve(id);
+    if (_user) {
+      return this.resolve(_user) ?? _user;
+    }
+    return id;
```
kyranet, line 295:
```suggestion
    return this.resolve(user) ?? this.client.users.resolve(user) ?? id;
```

kyranet (changes requested):
I'm not sure if those fit in the maximum amount of columns per line but... I tried those and it worked, not bad for much simplified logic!

kyranet (approved):

iCrawl:
This needs a rebase.

iCrawl (approved):

src/managers/GuildMemberManager.js
```diff
@@ -149,6 +150,47 @@ class GuildMemberManager extends BaseManager {
...
+  /**
+   * Edits a member of the guild.
+   * <info>The user must be a member of the guild</info>
+   * @param {UserResolvable} member The member to edit
+   * @param {GuildMemberEditData} data The data to edit the member with
+   * @param {string} [reason] Reason for editing this user
+   * @returns {Promise<GuildMember>}
+   */
+  async edit(member, data, reason) {
+    const id = this.client.users.resolveID(member);
```
SpaceEEC, line 162:
```suggestion
    const id = this.client.users.resolveID(user);
```

src/managers/GuildMemberManager.js
```diff
@@ -149,6 +150,47 @@ class GuildMemberManager extends BaseManager {
...
 
+  /**
+   * Edits a member of the guild.
+   * <info>The user must be a member of the guild</info>
+   * @param {UserResolvable} member The member to edit
+   * @param {GuildMemberEditData} data The data to edit the member with
+   * @param {string} [reason] Reason for editing this user
+   * @returns {Promise<GuildMember>}
+   */
+  async edit(member, data, reason) {
```
SpaceEEC, line 161:
```suggestion
  async edit(user, data, reason) {
```

src/managers/GuildMemberManager.js
```diff
@@ -149,6 +150,47 @@ class GuildMemberManager extends BaseManager {
     return data.reduce((col, member) => col.set(member.user.id, this.add(member, cache)), new Collection());
   }
 
+  /**
+   * Edits a member of the guild.
+   * <info>The user must be a member of the guild</info>
+   * @param {UserResolvable} member The member to edit
```
SpaceEEC, line 156:
```suggestion
   * @param {UserResolvable} user The member to edit
```

SpaceEEC (changes requested):
Or the other way for `GuildMemberManager#kick`, currently it's inconsistent.

Otherwise LGTM.

src/managers/GuildMemberManager.js
monbrey: Shouldn't I actually make this GuildMemberResolvable?

SpaceEEC: I'm not sure, since those types are identical. But `GuildMemberResolvable` might make more sense here.
The same applies to the `kick` method then too.

vladfrangu (approved):

PR #4874, opened 2020-10-02
feat(Message|TextChannel): Inline replies

monbrey:
Ref: https://github.com/discord/discord-api-docs/pull/2118

This implements the inline reply functionality, both for receiving and sending messages.
It now also ~~steals~~ merges the changes that Souji had made in #4870 to remove Message#reply and instead completely replaces it with inline reply functionality under the same name. It is therefore **semver major**

### Changes

- `MessageOptions#replyTo: MessageResolvable` added
- `MessageOptions#reply` removed
- `Message#reply` modified to use inline-reply functionality
- `APIMessage#resolveData` updated to resolve replyTo parameters
- Typings updated accordingly
- Boilerplate, examples and tests updated thanks to @almostSouji 

**Status**


src/structures/Message.js
```diff
@@ -223,6 +223,17 @@ class Message extends Base {
...
         }
       : null;
+
+    /**
+     * The message this message replies to
+     * @type {?Message}
+     */
+    if ('referenced_message' in data) {
+      this.replyReference =
+        this.channel.messages.get(data.referenced_message.id) || this.channel.messages.add(data.referenced_message);
```
NotSugden, line 233: doesn't this just need to be `this.channel.messages.add(...)` since add will handle if its cached

src/structures/Message.js
```diff
@@ -223,6 +223,17 @@ class Message extends Base {
           messageID: data.message_reference.message_id,
         }
       : null;
+
+    /**
+     * The message this message replies to
+     * @type {?Message}
+     */
```
NotSugden, line 230: this needs to be directly above `this.replyReference = ...` or have an `@name Message#replyReference` tag

NotSugden (changes requested):

src/structures/Message.js
```diff
@@ -422,10 +422,16 @@ class Message extends Base {
     );
   }
 
+  /**
+   * The Message this crosspost/reply/pin add references, if cached
+   * @type {Message}
+   * @readonly
+   */
```
tguichaoua, line 429:
```suggestion
  /**
   * The Message this crosspost/reply/pin add references, if cached
   * @type {?Message}
   * @readonly
   */
```
`referencedMessage` can be `null`.

tguichaoua (changes requested):

src/structures/Message.js
monbrey: Thanks!

advaith1:
we should add a `mention`/`ping` option to mention the author of the original message (`replied_user` boolean in `allowed_mentions`, `true` by default)

the mention is no longer present at the start of the message content when replying with pinging

image

README.md
```diff
@@ -76,7 +76,7 @@ client.on('ready', () => {
 
 client.on('message', msg => {
   if (msg.content === 'ping') {
-    msg.reply('pong');
+    msg.channel.send('pong');
```
vladfrangu, line 79: Why have a lot of `msg.reply` examples been removed, when the method is still available?

vladfrangu (changes requested):

README.md
almostSouji: Likely caused by the merge of my removal, which has been incorporated into this PR. However I also don't feel like we should use inline replies as the standard way to do command replies either, making the replacement still valid imo

vladfrangu: Doesn't mean remove everything... especially from typings

monbrey: The removal of the typings is an oversight, as I was originally planning to name this method something different to maintain it as minor rather than major. Will add them back

src/structures/Message.js
```diff
@@ -418,6 +422,18 @@ class Message extends Base {
...
+  /**
+   * The Message this crosspost/reply/pin add references, if cached
+   * @type {?Message}
+   * @readonly
+   */
+  get referencedMessage() {
+    if (!this.reference) return null;
+    const referenceChannel = this.client.channels.resolve(this.reference.channelID);
+    if (!referenceChannel) return null;
+    return referenceChannel.messages.resolve(this.reference.messageID);
```
NotSugden, line 434: would it be better to just use `.cache.get` rather than resolve here? since they're the ids it seems pretty useless to use resolve

monbrey: Resolve does the same thing anyway, and seemed more consistent with other internal usage

monbrey:
> we should add a `mention`/`ping` option to mention the author of the original message (`replied_user` boolean in `allowed_mentions`, `true` by default)

I'll add the additional property to the `allowedMentions` MessageOption, is there a need to also provide a shortcut to it?

`Message#reply("content", { allowedMentions: { repliedUser: true } })` would be the full implementation.

src/structures/APIMessage.js
```diff
@@ -182,19 +169,12 @@ class APIMessage {
...
-          allowedMentions.users.push(id);
-        }
-      } else {
-        allowedMentions = { users: [id] };
+
+    let message_reference;
+    if (typeof this.options.replyTo !== 'undefined') {
+      const message_id = this.target.messages.resolveID(this.options.replyTo);
+      if (message_id) {
+        message_reference = { message_id };
```
MoartMedia, line 177: I think it would be add channel_id and guild_id.
https://discord.com/developers/docs/resources/channel#message-object-message-reference-structure

monbrey: These fields are not required when sending an inline reply as they can only be to messages in the same channel.

vBread:
Should [`'INLINE_REPLY'`](https://github.com/discordjs/discord.js/pull/4874/files#diff-1c714d0a8a5ad05fd5427eb604e7666f098bf10ca0105e610a43330c4a5d1e1b) be `'REPLY'` to match the [API docs](https://github.com/discord/discord-api-docs/blob/master/docs/resources/Channel.md#message-types)?

monbrey:
> 
> 
> Should [`'INLINE_REPLY'`](https://github.com/discordjs/discord.js/pull/4874/files#diff-1c714d0a8a5ad05fd5427eb604e7666f098bf10ca0105e610a43330c4a5d1e1b) be `'REPLY'` to match the [API docs](https://github.com/discord/discord-api-docs/blob/master/docs/resources/Channel.md#message-types)?

I swear it was `INLINE_REPLY` originally haha, will update

src/structures/APIMessage.js
MoartMedia: if they can only use same channel,
It should be check TargetMessage and replyToMessage in same channel.

monbrey: That's not something Discord.js should be responsible for checking, nor can it without attempting to fetch the message from that channel with the ID provided. This will result in throwing a DiscordAPIError if the message ID is invalid.

Developers will receive alreaey receive a DiscordAPIError error anyway if they attempt to target an invalid message, and can catch that and handle accordingly.

advaith1:
> fix(Message): referenced_message is null, not undefined

it can be either:

>  If the message is a reply but the `referenced_message` field is not present, the backend did not attempt to fetch the message that was being replied to, so its state is unknown. If the field exists but is null, the referenced message was deleted.

monbrey:
> 
> 
> > fix(Message): referenced_message is null, not undefined
> 
> it can be either:
> 
> > If the message is a reply but the `referenced_message` field is not present, the backend did not attempt to fetch the message that was being replied to, so its state is unknown. If the field exists but is null, the referenced message was deleted.

I'm not explicitly checking for null, just falsey, so the fix applied still works

advaith1:
> I'll add the additional property to the `allowedMentions` MessageOption, is there a need to also provide a shortcut to it?
> 
> `Message#reply("content", { allowedMentions: { repliedUser: true } })` would be the full implementation.

imo it feels more natural to have a shortcut because it's different from the rest of `allowedMentions`, not really required through

typings/index.d.ts
```diff
@@ -1012,6 +1012,7 @@ declare module 'discord.js' {
     public webhookID: Snowflake | null;
     public flags: Readonly<MessageFlags>;
     public reference: MessageReference | null;
+    public replyReference: Message | null;
```
SpaceEEC, line 1015:
```suggestion
```

SpaceEEC (changes requested):

1A3Dev:
> > we should add a `mention`/`ping` option to mention the author of the original message (`replied_user` boolean in `allowed_mentions`, `true` by default)
> 
> I'll add the additional property to the `allowedMentions` MessageOption, is there a need to also provide a shortcut to it?
> 
> `Message#reply("content", { allowedMentions: { repliedUser: true } })` would be the full implementation.

`repliedUser` isn't really needed since using `replied_user` works anyway?

1A3Dev:
Does #5027 not make more sense since it doesn't remove the original functionality?

typings/index.d.ts
```diff
@@ -1012,6 +1012,7 @@ declare module 'discord.js' {
     public webhookID: Snowflake | null;
     public flags: Readonly<MessageFlags>;
     public reference: MessageReference | null;
+    public referencedMessage: Message | null;
```
vladfrangu, line 1015: this is a readonly prop, move accordingly

vladfrangu (changes requested):

oddyamill:
привет русским

src/structures/APIMessage.js
```diff
@@ -182,19 +169,19 @@ class APIMessage {
...
-        if (!parsed && !(allowedMentions.users && allowedMentions.users.includes(id))) {
-          if (!allowedMentions.users) allowedMentions.users = [];
-          allowedMentions.users.push(id);
-        }
-      } else {
-        allowedMentions = { users: [id] };
+
+    if (allowedMentions) {
+      allowedMentions.replied_user = allowedMentions.repliedUser;
+      delete allowedMentions.repliedUser;
```
NotSugden, line 175: this would modify the original `allowedMentions` object in the `ClientOptions`, or the object given in the `MessageOptions`, perhaps implement the same `allowedMentions = Util.cloneObject(allowedMentions)` behaviour as before if `allowedMentions && typeof allowedMentions.repliedUser === 'boolean'`

NotSugden (changes requested):

isuckatprogram:
hi

kylerchin:
Is this ready to be merged into master?

monbrey:
> 
> 
> Is this ready to be merged into master?

No

src/structures/APIMessage.js
vladfrangu: Is this resolved now? If so, please mark it as such

kyranet: I believe so, from fbb1c93.

vladfrangu (approved):
Looks fine to me! LGTM 👍

NotSugden:
the only use of the `APIMessage#isUser` getter is removed in this PR, it would make sense if this PR also removed that getter

SpaceEEC (approved):

kyranet (approved):

iCrawl (approved):

isuckatprogram:
pog

PR #4876, opened 2020-10-02
fix(typings): return types for 'Webhook(Client)#send()'

izexi:
As shown below, [`Webhook(Client)#send()`](https://discord.js.org/#/docs/main/stable/class/Webhook?scrollTo=send) would either return a Message (if the client was initiated and the channel was cached) or it would return the raw message response from the API (which has been added as a `APIRawMessage` interface in this PR).
https://github.com/discordjs/discord.js/blob/d2341654fe6ef87c6048bb5bf74c394062f3f618/src/structures/Webhook.js#L162-L164

Cuurently the typings do not reflect this behaviour, this PR fixes that.

I did consider making use of the [`APIMessage`](https://github.com/discordjs/discord-api-types/blob/main/v6/payloads/channel.ts#L53-L81) over on [`discord-api-types`](https://github.com/discordjs/discord-api-types) but that would mean introducing another dependency which be a breaking change for all Typescript developers.

| :warning: This change is breaking for Typescript developers that make use of properties or methods exclusive to a Message and not within the APIRawMessage from the resolved returned value of `Webhook(Client)#send()`. |
| --- |

<details>
<summary>Here's a snippet of how I tested these changes</summary>

```ts
import { expectType } from 'tsd';
import { WebhookRawMessageResponse, Message, WebhookClient } from 'discord.js';

const webhookClient = new WebhookClient('', '');

expectType<Message | WebhookRawMessageResponse>(
  await webhookClient.send('foo')
);

expectType<(Message | WebhookRawMessageResponse)[]>(
  await webhookClient.send('foo', { split: true })
);

expectType<
  Message | WebhookRawMessageResponse | (Message | WebhookRawMessageResponse)[]
>(await webhookClient.send({ embeds: [] }, { split: true }));

expectType<Message | WebhookRawMessageResponse>(
  await webhookClient.send('foo', { embeds: [] })
);

expectType<
  Message | WebhookRawMessageResponse | (Message | WebhookRawMessageResponse)[]
>(await webhookClient.send('foo', { embeds: [], split: true }));
```
</details>

**Status**

### DIFF
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -2193,7 +2193,7 @@ declare module 'discord.js' {
     content: string;
     channel_id: Snowflake;
     author: {
-      bot: true;
+      bot: boolean;
       id: Snowflake;
       username: string;
       avatar: string | null;
@@ -2266,7 +2266,7 @@ declare module 'discord.js' {
       public_flags?: number;
       member?: {
         nick: string | null;
-        roles: string[];
+        roles: Snowflake[];
         joined_at: string;
         premium_since?: string | null;
         deaf: boolean;
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -2193,7 +2193,7 @@ declare module 'discord.js' {
     content: string;
     channel_id: Snowflake;
     author: {
-      bot: boolean;
+      bot?: true;
       id: Snowflake;
       username: string;
       avatar: string | null;
@@ -2255,14 +2255,7 @@ declare module 'discord.js' {
       username: string;
       discriminator: string;
       avatar: string | null;
-      bot?: boolean;
-      system?: boolean;
-      mfa_enabled?: boolean;
-      locale?: string;
-      verified?: boolean;
-      email?: string | null;
-      flags?: number;
-      premium_type?: number;
+      bot?: true;
       public_flags?: number;
       member?: {
         nick: string | null;
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -2097,18 +2097,21 @@ declare module 'discord.js' {
     edit(options: WebhookEditData): Promise<Webhook>;
     send(
       content: APIMessageContentResolvable | (WebhookMessageOptions & { split?: false }) | MessageAdditions,
-    ): Promise<Message>;
-    send(options: WebhookMessageOptions & { split: true | SplitOptions }): Promise<Message[]>;
-    send(options: WebhookMessageOptions | APIMessage): Promise<Message | Message[]>;
+    ): Promise<Message | APIRawMessage>;
+    send(options: WebhookMessageOptions & { split: true | SplitOptions }): Promise<(Message | APIRawMessage)[]>;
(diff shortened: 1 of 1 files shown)


vladfrangu:
I don't see how adding a dependency on `discord-api-types` is breaking for TS developers (note: dependency, not dev/peer dep).. 👀

typings/index.d.ts
```diff
@@ -2184,6 +2187,102 @@ declare module 'discord.js' {
...
+      discriminator: string;
+      avatar: string | null;
+      bot?: boolean;
+      system?: boolean;
+      mfa_enabled?: boolean;
+      locale?: string;
+      verified?: boolean;
+      email?: string | null;
+      flags?: number;
+      premium_type?: number;
```
NotSugden, line 2265: properties like this wont ever be present i dont think, they're only available using oauth2

typings/index.d.ts
```diff
@@ -2184,6 +2187,102 @@ declare module 'discord.js' {
 
   type APIMessageContentResolvable = string | number | boolean | bigint | symbol | readonly StringResolvable[];
 
+  interface APIRawMessage {
+    id: Snowflake;
+    type: number;
+    content: string;
+    channel_id: Snowflake;
+    author: {
+      bot: boolean;
```
NotSugden, line 2196:
```suggestion
      bot?: boolean;
```
this is optional, true if present so if you wanted yo be explicit you could do `bot?: true`

NotSugden (changes requested):
these overloads should be applied to `WebhookClient` directly, as the `Webhook` class is instantiated by internal methods and always has a client, thus always returns a full message

izexi:
> I don't see how adding a dependency on `discord-api-types` is breaking for TS developers (note: dependency, not dev/peer dep).. 👀

@vladfrangu I was assuming that they had to go under `devDependicies` since that's where the other type packages were listed. Although, as mentioned by Sudgen some of the OAuth2 properties wont be present so I could Omit them I guess but I'm not sure about adding it as a dependency just for the `APIMessage`.

--- 
> these overloads should be applied to `WebhookClient` directly, as the `Webhook` class is instantiated by internal methods and always has a client, thus always returns a full message

@NotSugden If the channel isn't within the `ChannelManager#cache` Collection then it would return the raw message, although this isn't possible unless the developer removes the channel manually I believe it should still be accounted for.

iCrawl (approved):

kyranet:
> > I don't see how adding a dependency on `discord-api-types` is breaking for TS developers (note: dependency, not dev/peer dep).. 👀
> 
> @vladfrangu I was assuming that they had to go under `devDependicies` since that's where the other type packages were listed. Although, as mentioned by Sudgen some of the OAuth2 properties wont be present so I could `Omit` them I guess but I'm not sure about adding it as a dependency just for the `APIMessage`.

Since we're releasing this for discord.js v13, I personally don't see a reason to not include `discord-api-types`, we can afford semver-major this time.

iCrawl:
For what its worth: I would like to keep this as is and move forward like this until we have a clear path for modularization.

vladfrangu (approved):

SpaceEEC (approved):

kyranet (approved):

PR #4880, opened 2020-10-03
feat(*): enforce strings

almostSouji:

This PR removes StringResolvable (which was just a nice naming scheme for "any" anyways) and enforces and tests for strings to be passed where before any could be resolved to strings.

We have a lot of issues in support regarding people wondering why certain parts of a message or embed may show "undefined" or "null" due to unexpectedly resolving them to strings ([despite it being documented](https://discord.js.org/#/docs/main/stable/typedef/StringResolvable), as space has pointed out).

This is majorly opinionated and I personally believe we should hand the responsibility of providing message content as a string (and how it might be composed from the respective structures) up to the developer. While I do see that resolving anything to a string has its benefits (especially with structures that do have proper #toString implementations resulting in useful content), I also feel like this change will ultimately result in more verbose and explicit code for users as well as more useful pointers (as the previously resolved `undefined` -> `"undefined"`, to name one example, would now throw with a useful error pointing clearly to the issue)

To achieve this I replaced the `resolveString` method with `verifyString` and let the context provide the error to be thrown in order to provide useful information (which embed or message attribute caused this error) utilizing `RangeError` and the existing error message constant system.

**Input/Feedback requested**

| ❓  For now this only throws with empty string where the API rejects. However while empty strings are valid (the API accepts and processes the request instead of throwing) in several positions in embed structures they do cause unexpected behavior. For example providing `""` as embed author name or footer text will hide the respective icon, despite technically being valid (a zero-width space can circumvent this behavior). We could decide to be opinionated here and generally enforce a non-empty string to be passed where strings are expected in embeds. |
| --- |

**Status**

### DIFF
--- a/src/errors/Messages.js
+++ b/src/errors/Messages.js
@@ -37,6 +37,13 @@ const Messages = {
   COLOR_RANGE: 'Color must be within the range 0 - 16777215 (0xFFFFFF).',
   COLOR_CONVERT: 'Unable to convert color to a number.',
 
+  EMBED_TITLE: 'MessageEmbed title must be a string',
+  EMBED_FIELD_NAME: 'MessageEmbed field names must be non-empty strings.',
+  EMBED_FIELD_VALUE: 'MessageEmbed field values must be non-empty strings.',
+  EMBED_FOOTER_TEXT: 'MessageEmbed footer text must be a string.',
+  EMBED_DESCRIPTION: 'MessageEmbed description must be a string.',
+  EMBED_AUTHOR_NAME: 'MessageEmbed author name must be a string.',
+
   FILE_NOT_FOUND: file => `File could not be found: ${file}`,
 
   USER_NO_DMCHANNEL: 'No DM Channel exists!',
@@ -71,7 +78,8 @@ const Messages = {
   IMAGE_SIZE: size => `Invalid image size: ${size}`,
 
   MESSAGE_BULK_DELETE_TYPE: 'The messages must be an Array, Collection, or number.',
-  MESSAGE_NONCE_TYPE: 'Message nonce must be an integer or a string.',
+  MESSAGE_NONCE_TYPE: 'Message nonce must fit in an unsigned 64-bit integer.',
+  MESSAGE_CONTENT_TYPE: 'Message content must be a non-empty string.',
 
   TYPING_COUNT: 'Count must be at least 1',
 
--- a/src/errors/Messages.js
+++ b/src/errors/Messages.js
@@ -78,7 +78,7 @@ const Messages = {
   IMAGE_SIZE: size => `Invalid image size: ${size}`,
 
   MESSAGE_BULK_DELETE_TYPE: 'The messages must be an Array, Collection, or number.',
-  MESSAGE_NONCE_TYPE: 'Message nonce must fit in an unsigned 64-bit integer.',
+  MESSAGE_NONCE_TYPE: 'Message nonce must be an integer or a string.',
   MESSAGE_CONTENT_TYPE: 'Message content must be a non-empty string.',
 
   TYPING_COUNT: 'Count must be at least 1',
--- a/src/errors/Messages.js
+++ b/src/errors/Messages.js
@@ -37,7 +37,7 @@ const Messages = {
   COLOR_RANGE: 'Color must be within the range 0 - 16777215 (0xFFFFFF).',
   COLOR_CONVERT: 'Unable to convert color to a number.',
 
-  EMBED_TITLE: 'MessageEmbed title must be a string',
+  EMBED_TITLE: 'MessageEmbed title must be a string.',
   EMBED_FIELD_NAME: 'MessageEmbed field names must be non-empty strings.',
   EMBED_FIELD_VALUE: 'MessageEmbed field values must be non-empty strings.',
   EMBED_FOOTER_TEXT: 'MessageEmbed footer text must be a string.',
--- a/src/structures/APIMessage.js
+++ b/src/structures/APIMessage.js
@@ -92,7 +92,7 @@ class APIMessage {
     if (this.options.content === null) {
       content = '';
     } else if (typeof this.options.content !== 'undefined') {
-      content = Util.resolveString(this.options.content);
+      content = Util.verifyString(this.options.content, new RangeError('MESSAGE_CONTENT_TYPE'), false);
     }
 
     if (typeof content !== 'string') return content;
@@ -322,7 +322,7 @@ class APIMessage {
   /**
    * Transforms the user-level arguments into a final options object. Passing a transformed options object alone into
    * this method will keep it the same, allowing for the reuse of the final options object.
-   * @param {StringResolvable} [content] Content to send
+   * @param {string} [content] Content to send
    * @param {MessageOptions|WebhookMessageOptions|MessageAdditions} [options={}] Options to use
    * @param {MessageOptions|WebhookMessageOptions} [extra={}] Extra options to add onto transformed options
    * @param {boolean} [isWebhook=false] Whether or not to use WebhookMessageOptions as the result
@@ -358,7 +358,7 @@ class APIMessage {
   /**
    * Creates an `APIMessage` from user-level arguments.
    * @param {MessageTarget} target Target to send to
-   * @param {StringResolvable} [content] Content to send
+   * @param {string} [content] Content to send
    * @param {MessageOptions|WebhookMessageOptions|MessageAdditions} [options={}] Options to use
    * @param {MessageOptions|WebhookMessageOptions} [extra={}] - Extra options to add onto transformed options
    * @returns {MessageOptions|WebhookMessageOptions}
--- a/src/structures/APIMessage.js
+++ b/src/structures/APIMessage.js
@@ -92,7 +92,7 @@ class APIMessage {
     if (this.options.content === null) {
       content = '';
     } else if (typeof this.options.content !== 'undefined') {
-      content = Util.verifyString(this.options.content, new RangeError('MESSAGE_CONTENT_TYPE'));
+      content = Util.verifyString(this.options.content, RangeError, 'MESSAGE_CONTENT_TYPE');
     }
 
     if (typeof content !== 'string') return content;
--- a/src/structures/APIMessage.js
+++ b/src/structures/APIMessage.js
@@ -92,7 +92,7 @@ class APIMessage {
     if (this.options.content === null) {
       content = '';
     } else if (typeof this.options.content !== 'undefined') {
-      content = Util.verifyString(this.options.content, new RangeError('MESSAGE_CONTENT_TYPE'), false);
+      content = Util.verifyString(this.options.content, new RangeError('MESSAGE_CONTENT_TYPE'));
     }
 
     if (typeof content !== 'string') return content;
--- a/src/structures/APIMessage.js
+++ b/src/structures/APIMessage.js
@@ -92,7 +92,7 @@ class APIMessage {
     if (this.options.content === null) {
       content = '';
     } else if (typeof this.options.content !== 'undefined') {
-      content = Util.verifyString(this.options.content, RangeError, 'MESSAGE_CONTENT_TYPE');
+      content = Util.verifyString(this.options.content, RangeError, 'MESSAGE_CONTENT_TYPE', false);
     }
 
     if (typeof content !== 'string') return content;
--- a/src/structures/MessageEmbed.js
+++ b/src/structures/MessageEmbed.js
@@ -335,8 +335,7 @@ class MessageEmbed {
    * @returns {MessageEmbed}
    */
   setDescription(description) {
-    description = Util.verifyString(description, new RangeError('EMBED_DESCRIPTION'));
-    this.description = description;
+    this.description = Util.verifyString(description, new RangeError('EMBED_DESCRIPTION'));
     return this;
   }
 
@@ -347,8 +346,7 @@ class MessageEmbed {
    * @returns {MessageEmbed}
    */
   setFooter(text, iconURL) {
-    text = Util.verifyString(text, new RangeError('EMBED_FOOTER_TEXT'));
-    this.footer = { text, iconURL };
+    this.footer = { text: Util.verifyString(text, new RangeError('EMBED_FOOTER_TEXT')), iconURL };
     return this;
   }
 
@@ -389,8 +387,7 @@ class MessageEmbed {
    * @returns {MessageEmbed}
    */
   setTitle(title) {
-    title = Util.verifyString(title, new RangeError('EMBED_TITLE'));
-    this.title = title;
+    this.title = Util.verifyString(title, new RangeError('EMBED_TITLE'));
     return this;
   }
 
@@ -443,9 +440,11 @@ class MessageEmbed {
    * @returns {EmbedField}
    */
   static normalizeField(name, value, inline = false) {
-    name = Util.verifyString(name, new RangeError('EMBED_FIELD_NAME'), false);
-    value = Util.verifyString(value, new RangeError('EMBED_FIELD_VALUE'), false);
-    return { name, value, inline };
+    return {
+      name: Util.verifyString(name, new RangeError('EMBED_FIELD_NAME'), false),
+      value: Util.verifyString(value, new RangeError('EMBED_FIELD_VALUE'), false),
+      inline,
+    };
   }
 
   /**
--- a/src/structures/MessageEmbed.js
+++ b/src/structures/MessageEmbed.js
@@ -315,7 +315,7 @@ class MessageEmbed {
    * @returns {MessageEmbed}
    */
   setAuthor(name, iconURL, url) {
-    this.author = { name: Util.verifyString(name, new RangeError('EMBED_AUTHOR_NAME')), iconURL, url };
+    this.author = { name: Util.verifyString(name, RangeError, 'EMBED_AUTHOR_NAME'), iconURL, url };
     return this;
   }
 
@@ -335,7 +335,7 @@ class MessageEmbed {
    * @returns {MessageEmbed}
    */
   setDescription(description) {
-    this.description = Util.verifyString(description, new RangeError('EMBED_DESCRIPTION'));
+    this.description = Util.verifyString(description, RangeError, 'EMBED_DESCRIPTION');
     return this;
   }
 
@@ -346,7 +346,7 @@ class MessageEmbed {
    * @returns {MessageEmbed}
    */
   setFooter(text, iconURL) {
-    this.footer = { text: Util.verifyString(text, new RangeError('EMBED_FOOTER_TEXT')), iconURL };
+    this.footer = { text: Util.verifyString(text, RangeError, 'EMBED_FOOTER_TEXT'), iconURL };
     return this;
   }
 
@@ -387,7 +387,7 @@ class MessageEmbed {
    * @returns {MessageEmbed}
    */
   setTitle(title) {
-    this.title = Util.verifyString(title, new RangeError('EMBED_TITLE'));
+    this.title = Util.verifyString(title, RangeError, 'EMBED_TITLE');
     return this;
   }
 
@@ -441,8 +441,8 @@ class MessageEmbed {
    */
   static normalizeField(name, value, inline = false) {
     return {
-      name: Util.verifyString(name, new RangeError('EMBED_FIELD_NAME'), false),
-      value: Util.verifyString(value, new RangeError('EMBED_FIELD_VALUE'), false),
+      name: Util.verifyString(name, RangeError, 'EMBED_FIELD_NAME', false),
+      value: Util.verifyString(value, RangeError, 'EMBED_FIELD_VALUE', false),
       inline,
     };
   }
--- a/src/structures/MessageEmbed.js
+++ b/src/structures/MessageEmbed.js
@@ -265,8 +265,8 @@ class MessageEmbed {
 
   /**
    * Adds a field to the embed (max 25).
-   * @param {StringResolvable} name The name of this field
-   * @param {StringResolvable} value The value of this field
+   * @param {string} name The name of this field
+   * @param {string} value The value of this field
    * @param {boolean} [inline=false] If this field will be displayed inline
    * @returns {MessageEmbed}
    */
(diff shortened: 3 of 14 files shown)


src/errors/Messages.js
```diff
@@ -32,8 +32,12 @@ const Messages = {
   COLOR_RANGE: 'Color must be within the range 0 - 16777215 (0xFFFFFF).',
   COLOR_CONVERT: 'Unable to convert color to a number.',
 
-  EMBED_FIELD_NAME: 'MessageEmbed field names may not be empty.',
-  EMBED_FIELD_VALUE: 'MessageEmbed field values may not be empty.',
+  EMBED_TITLE: 'MessageEmbed title must be a string',
```
NotSugden, line 35:
```suggestion
  EMBED_TITLE: 'MessageEmbed title must be a string.',
```

Vaporox:
Couldn't this just use the `INVALID_TYPE` error message instead of adding a new message for each MessageEmbed property?

almostSouji:
oriented on the existing

https://github.com/discordjs/discord.js/blob/d2341654fe6ef87c6048bb5bf74c394062f3f618/src/errors/Messages.js#L35-L36

I do think it can be useful to know which part of an embed the error might be from without having to do a lot of backtracing, but I am of course open for change here.

almostSouji:
> Couldn't this just use the `INVALID_TYPE` error message instead of adding a new message for each MessageEmbed property?

I somehow missed we also resolve function input for this, one error message is variable location and must part is indeed sufficient.

Mark has brought up that we currently ignore if a function is passed in any URL field of embeds, which causes the handling further down the line to ignore the input. Explicit string casting makes use of discords Validation and results in 
```
embed.author.icon_url: Not a well formed URL.
```

Applying this seems to fit nicely into the scope of this PR to more strictly enforce strings (and valid input) rather than defaulting to still valid embeds being sent.

src/errors/Messages.js
```diff
@@ -32,8 +32,7 @@ const Messages = {
   COLOR_RANGE: 'Color must be within the range 0 - 16777215 (0xFFFFFF).',
   COLOR_CONVERT: 'Unable to convert color to a number.',
 
-  EMBED_FIELD_NAME: 'MessageEmbed field names may not be empty.',
-  EMBED_FIELD_VALUE: 'MessageEmbed field values may not be empty.',
+  EMBED_INVALID_TYPE: (location, must) => `MessageEmbed ${location} must be ${must}.`,
```
Vaporox, line 35:
```suggestion
  EMBED_INVALID_TYPE: (location, must, an = false) => `MessageEmbed ${location} must be a${an ? 'n' : ''} ${must}.`,
```
Missing the article here

Vaporox:
Next to that, shouldn't it use TypeError instead of RangeError?

src/errors/Messages.js
almostSouji: considered an article, don't think it adds anything of use

almostSouji:
Considered that as well, truth be told I by now have lost any and all intuition as to when which of the two is used. Apparently non-empty string is range but only string would be Type?

In which case it might even make sense to dynamically instantiate the error in the method and just pass the string? I really don't know anymore.

almostSouji:
Re-viewing API behavior it might even be feasible to default to empty strings if a non-string is provided and let the API handle the rest where non-empty field input is required.

Typing wise this won't happen in TS. However justifying type checks in only select areas of the library consistency wise seems to be a bad idea. If type checks are wanted one should really write TS instead of JS.

Casting (not resolving) to string here might be the correct decision following this mindset, reducing this to removal of resolveString and typings (where strings can quite easily be enforced).

src/structures/MessageEmbed.js
```diff
@@ -314,24 +318,26 @@ class MessageEmbed {
...
    * @param {string} [iconURL] The icon URL of the footer
    * @returns {MessageEmbed}
    */
   setFooter(text, iconURL) {
-    text = Util.resolveString(text);
-    this.footer = { text, iconURL, proxyIconURL: undefined };
+    this.footer = {
+      text: Util.verifyString(text, new RangeError('EMBED_INVALID_TYPE', 'footer text', 'string')),
+      iconURL: iconURL ? String(iconURL) : undefined,
+      proxyIconURL: undefined,
```
NotSugden, line 339: can't this line just be removed, there's no point in it

src/util/Util.js
```diff
@@ -371,22 +371,16 @@ class Util {
...
+   * @param {string} data The string resolvable to resolve
+   * @param {Error} [error=Error] The error to throw, should the validation fail
+   * @param {boolean} [allowEmpty=true] Wether an empty string should be allowed
    * @returns {string}
    */
-  static resolveString(data) {
-    if (typeof data === 'string') return data;
-    if (Array.isArray(data)) return data.join('\n');
-    return String(data);
+  static verifyString(data, error = new Error(), allowEmpty = true) {
```
vladfrangu, line 380: Just a small tweak
```suggestion
  static verifyString(data, error = new Error(`Expected a string, got ${data} instead`), allowEmpty = true) {
```

vladfrangu (changes requested):
Fix lints 🙏

iCrawl (approved):

kyranet (approved):

iCrawl:
@almostSouji rebase

src/errors/Messages.js
```diff
@@ -70,7 +69,8 @@ const Messages = {
   IMAGE_SIZE: size => `Invalid image size: ${size}`,
 
   MESSAGE_BULK_DELETE_TYPE: 'The messages must be an Array, Collection, or number.',
-  MESSAGE_NONCE_TYPE: 'Message nonce must be an integer or a string.',
+  MESSAGE_NONCE_TYPE: 'Message nonce must fit in an unsigned 64-bit integer.',
```
vladfrangu, line 72: Wrong rebase here, this message shouldn't be changed

typings/index.d.ts
```diff
@@ -2195,15 +2171,14 @@ declare module 'discord.js' {
...
-  type APIMessageContentResolvable = string | number | boolean | bigint | symbol | readonly StringResolvable[];
-
   interface ApplicationAsset {
     name: string;
     id: Snowflake;
     type: 'BIG' | 'SMALL';
   }
 
-  interface AuditLogChange {
+    type APIMessageContentResolvable = string | number | boolean | bigint | symbol | readonly string[];
```
vladfrangu, line 2180: Is this even needed? I thought the point was to drop all these, and stick to string?

vladfrangu (changes requested):

typings/index.d.ts
almostSouji: should be solved in https://github.com/discordjs/discord.js/pull/4880/commits/c20bf396e987b2f8a8c8eb9cdec471f5ed245ca8 - moved the tests (like primitive symbol and non-options object to expect-error, since these introduced tests are precisely not what's wanted after this PR; removed the newly introduced type APIMessageContentResolvable, since it was just used in edit, where now a string is expected anyways)

typings/index.d.ts
```diff
@@ -2758,22 +2732,22 @@ declare module 'discord.js' {
...
-    name?: string;
+    name: string;
     url?: string;
     iconURL?: string;
     proxyIconURL?: string;
   }
 
   interface MessageEmbedFooter {
-    text?: string;
+    text: string;
```
SpaceEEC, line 2750: Why are these now required? An empty object is technically valid.

src/structures/MessageEmbed.js
```diff
@@ -341,7 +344,7 @@ class MessageEmbed {
    * @returns {MessageEmbed}
    */
   setImage(url) {
-    this.image = { url };
+    this.image = { url: String(url) };
```
SpaceEEC, line 347: Why are these not verified?

src/structures/APIMessage.js
```diff
@@ -82,7 +82,7 @@ class APIMessage {
     if (this.options.content === null) {
       content = '';
     } else if (typeof this.options.content !== 'undefined') {
-      content = Util.resolveString(this.options.content);
+      content = Util.verifyString(this.options.content, new RangeError('MESSAGE_CONTENT_TYPE'), false);
```
SpaceEEC, line 85: I assume disallowing empty strings here is to require a more explicit `null` to clear the message's content?
(When having a file or embed on the message, that is a valid action)

almostSouji: that might actually just be an oversight, will allow empty strings

src/structures/MessageEmbed.js
almostSouji: If I remember my reasoning correctly these  (URLs) are rejected by the API if anything but a valid looking URL is provided, with an appropriate error, leading to no surprising effects (like the mentioned "undefined", "null" and "[Object object]" strings in places where people pass non-string structures.

That being said I could see how verifying the string on our end and then still having the verification of URL format on the API could work in a layered sense.

typings/index.d.ts
almostSouji: oversight, fixed in #92636cc40a59cc98f376eac520a0ddc0dbef700b

iCrawl:
This needs a rebase.

iCrawl:
This needs a rebase.

src/errors/Messages.js
vladfrangu: Psst, Souji.

src/structures/APIMessage.js
vladfrangu: Psst again

src/structures/MessageEmbed.js
```diff
@@ -315,23 +315,23 @@ class MessageEmbed {
...
   /**
    * Sets the description of this embed.
-   * @param {StringResolvable} description The description
+   * @param {string} description The description
    * @returns {MessageEmbed}
    */
   setDescription(description) {
-    description = Util.resolveString(description);
+    description = Util.verifyString(description, new RangeError('EMBED_DESCRIPTION'));
     this.description = description;
```
vladfrangu, line 323: Can't this be inlined?

src/structures/MessageEmbed.js
```diff
@@ -315,23 +315,23 @@ class MessageEmbed {
...
    * Sets the footer of this embed.
-   * @param {StringResolvable} text The text of the footer
+   * @param {string} text The text of the footer
    * @param {string} [iconURL] The icon URL of the footer
    * @returns {MessageEmbed}
    */
   setFooter(text, iconURL) {
-    text = Util.resolveString(text);
+    text = Util.verifyString(text, new RangeError('EMBED_FOOTER_TEXT'));
     this.footer = { text, iconURL };
```
vladfrangu, line 335: Inline like in setAuthor

src/structures/MessageEmbed.js
```diff
@@ -369,11 +369,11 @@ class MessageEmbed {
...
   /**
    * Sets the title of this embed.
-   * @param {StringResolvable} title The title
+   * @param {string} title The title
    * @returns {MessageEmbed}
    */
   setTitle(title) {
-    title = Util.resolveString(title);
+    title = Util.verifyString(title, new RangeError('EMBED_TITLE'));
     this.title = title;
```
vladfrangu, line 377: Inline

src/structures/MessageEmbed.js
```diff
@@ -421,23 +421,21 @@ class MessageEmbed {
...
    * @param {boolean} [inline=false] Set the field to display inline
    * @returns {EmbedField}
    */
   static normalizeField(name, value, inline = false) {
-    name = Util.resolveString(name);
-    if (!name) throw new RangeError('EMBED_FIELD_NAME');
-    value = Util.resolveString(value);
-    if (!value) throw new RangeError('EMBED_FIELD_VALUE');
+    name = Util.verifyString(name, new RangeError('EMBED_FIELD_NAME'), false);
+    value = Util.verifyString(value, new RangeError('EMBED_FIELD_VALUE'), false);
```
vladfrangu, line 431: inline

src/util/Util.js
vladfrangu: Psst.
Just a small tweak
```suggestion
  static verifyString(data, error = new Error(`Expected a string, got ${data} instead`), allowEmpty = true) {
```

typings/index.d.ts
```diff
@@ -91,13 +91,13 @@ declare module 'discord.js' {
 
     public static create(
       target: MessageTarget,
-      content: APIMessageContentResolvable,
+      content: string,
```
vladfrangu, line 94: Content can be null

typings/index.d.ts
```diff
@@ -1043,27 +1043,19 @@ declare module 'discord.js' {
     public edit(
       content: APIMessageContentResolvable | MessageEditOptions | MessageEmbed | APIMessage,
     ): Promise<Message>;
-    public edit(content: StringResolvable, options: MessageEditOptions | MessageEmbed): Promise<Message>;
+    public edit(content: string, options: MessageEditOptions | MessageEmbed): Promise<Message>;
```
vladfrangu, line 1046: content can be null

typings/index.d.ts
vladfrangu: This has been undone

vladfrangu (changes requested):
Your rebase has uh... borked, and a lot of requested changes are now missing! Please shoot an eye at them again <3

kyranet (dismissed):

kyranet:
You won't believe this, @almostSouji, but this needs a rebase *again*

almostSouji:
i hate every single one of you.

src/util/Util.js
```diff
@@ -347,22 +347,22 @@ class Util {
...
-    if (Array.isArray(data)) return data.join('\n');
-    return String(data);
+  static verifyString(
+    data,
+    error = Error,
+    errorMessage = `Expected a string, got ${data} instead.`,
+    allowEmpty = true,
+  ) {
+    if (typeof data !== 'string') throw new error(errorMessage);
+    if (!allowEmpty && data.length === 0) throw new error(errorMessage);
```
vladfrangu, line 364: Won't this catastrophically break? Look at how its used and what parameters it expects

vladfrangu (changes requested):

src/util/Util.js
almostSouji: oopsies! fixed in dec397f i think

vladfrangu (approved):

kyranet (approved):

iCrawl:
This needs a rebase @almostSouji

typings/index.d.ts
```diff
@@ -1759,7 +1755,8 @@ declare module 'discord.js' {
       route: object,
       reason?: string,
     ): Promise<{ id: Snowflake; position: number }[]>;
-    public static splitMessage(text: StringResolvable, options?: SplitOptions): string[];
+    public static splitMessage(text: string, options?: SplitOptions): string[];
+    public static str2ab(str: string): ArrayBuffer;
```
SpaceEEC, line 1759: Probably slipped through a rebase.
```suggestion
```

typings/index.d.ts
```diff
@@ -2451,7 +2443,7 @@ declare module 'discord.js' {
     RESOURCE_OVERLOADED: 130000;
   }
 
-  type APIMessageContentResolvable = string | number | boolean | bigint | symbol | readonly StringResolvable[];
+  type APIMessageContentResolvable = string | number | boolean | bigint | symbol | readonly string[];
```
SpaceEEC, line 2446: Shouldn't this type be gone too?
This will currently compile, but fail at runtime:
`message.reply(Symbol.for("hello there"));`

typings/index.d.ts
```diff
@@ -117,13 +117,13 @@ declare module 'discord.js' {
...
       target: MessageTarget,
-      content: APIMessageContentResolvable,
+      content: string | null,
       options?: undefined,
       extra?: MessageOptions | WebhookMessageOptions,
     ): APIMessage;
     public static create(
       target: MessageTarget,
-      content: StringResolvable,
+      content: string,
```
SpaceEEC, line 126: Shouldn't this and the methods below until inclusive `reply` also allow `null`?

src/structures/APIMessage.js
```diff
@@ -92,7 +92,7 @@ class APIMessage {
     if (this.options.content === null) {
       content = '';
     } else if (typeof this.options.content !== 'undefined') {
-      content = Util.resolveString(this.options.content);
+      content = Util.verifyString(this.options.content, RangeError, 'MESSAGE_CONTENT_TYPE');
```
SpaceEEC, line 95: This currently allows empty strings, contrary to what the error states, did you miss the `false` for the `allowEmpty` parameter?

SpaceEEC (changes requested):

typings/index.d.ts
almostSouji: not too sure what reply you are referring to but I think I applied the requested changes in https://github.com/discordjs/discord.js/pull/4880/commits/872675caab1da1ffb69394ddd8ae280f14798f83

typings/index.d.ts
```diff
@@ -1943,14 +1939,14 @@ declare module 'discord.js' {
     public send(options: WebhookMessageOptions & { split: true | SplitOptions }): Promise<RawMessage[]>;
     public send(options: WebhookMessageOptions | APIMessage): Promise<RawMessage | RawMessage[]>;
     public send(
-      content: StringResolvable,
+      content: string,
```
SpaceEEC, line 1942:
```suggestion
      content: string | null,
```

typings/index.d.ts
```diff
@@ -1943,14 +1939,14 @@ declare module 'discord.js' {
...
       options: (WebhookMessageOptions & { split?: false }) | MessageAdditions,
     ): Promise<RawMessage>;
     public send(
-      content: StringResolvable,
+      content: string,
```
SpaceEEC, line 1946:
```suggestion
      content: string | null,
```

typings/index.d.ts
```diff
@@ -1943,14 +1939,14 @@ declare module 'discord.js' {
...
       options: WebhookMessageOptions & { split: true | SplitOptions },
     ): Promise<RawMessage[]>;
-    public send(content: StringResolvable, options: WebhookMessageOptions): Promise<RawMessage | RawMessage[]>;
+    public send(content: string, options: WebhookMessageOptions): Promise<RawMessage | RawMessage[]>;
```
SpaceEEC, line 1949:
```suggestion
    public send(content: string | null, options: WebhookMessageOptions): Promise<RawMessage | RawMessage[]>;
```

typings/index.d.ts
```diff
@@ -2285,14 +2281,12 @@ declare module 'discord.js' {
...
-    send(
-      content: APIMessageContentResolvable | (MessageOptions & { split?: false }) | MessageAdditions,
-    ): Promise<Message>;
+    send(content: string | (MessageOptions & { split?: false }) | MessageAdditions): Promise<Message>;
     send(options: MessageOptions & { split: true | SplitOptions }): Promise<Message[]>;
     send(options: MessageOptions | APIMessage): Promise<Message | Message[]>;
-    send(content: StringResolvable, options: (MessageOptions & { split?: false }) | MessageAdditions): Promise<Message>;
-    send(content: StringResolvable, options: MessageOptions & { split: true | SplitOptions }): Promise<Message[]>;
-    send(content: StringResolvable, options: MessageOptions): Promise<Message | Message[]>;
+    send(content: string, options: (MessageOptions & { split?: false }) | MessageAdditions): Promise<Message>;
```
SpaceEEC, line 2287:
```suggestion
    send(content: string | null, options: (MessageOptions & { split?: false }) | MessageAdditions): Promise<Message>;
```

typings/index.d.ts
```diff
@@ -2285,14 +2281,12 @@ declare module 'discord.js' {
...
+    send(content: string, options: MessageOptions & { split: true | SplitOptions }): Promise<Message[]>;
```
SpaceEEC, line 2288:
```suggestion
    send(content: string | null, options: MessageOptions & { split: true | SplitOptions }): Promise<Message[]>;
```

typings/index.d.ts
```diff
@@ -2285,14 +2281,12 @@ declare module 'discord.js' {
...
+    send(content: string, options: MessageOptions): Promise<Message | Message[]>;
```
SpaceEEC, line 2289:
```suggestion
    send(content: string | null, options: MessageOptions): Promise<Message | Message[]>;
```

typings/index.d.ts
```diff
@@ -1172,14 +1172,14 @@ declare module 'discord.js' {
     public reply(options: ReplyMessageOptions & { split: true | SplitOptions }): Promise<Message[]>;
     public reply(options: ReplyMessageOptions | APIMessage): Promise<Message | Message[]>;
     public reply(
-      content: StringResolvable,
+      content: string,
```
SpaceEEC, line 1175:
```suggestion
      content: string | null,
```

typings/index.d.ts
```diff
@@ -1172,14 +1172,14 @@ declare module 'discord.js' {
...
       options: (ReplyMessageOptions & { split?: false }) | MessageAdditions,
     ): Promise<Message>;
     public reply(
-      content: StringResolvable,
+      content: string,
```
SpaceEEC, line 1179:
```suggestion
      content: string | null,
```

typings/index.d.ts
```diff
@@ -1172,14 +1172,14 @@ declare module 'discord.js' {
...
       options: ReplyMessageOptions & { split: true | SplitOptions },
     ): Promise<Message[]>;
-    public reply(content: StringResolvable, options: ReplyMessageOptions): Promise<Message | Message[]>;
+    public reply(content: string, options: ReplyMessageOptions): Promise<Message | Message[]>;
```
SpaceEEC, line 1182:
```suggestion
    public reply(content: string | null, options: ReplyMessageOptions): Promise<Message | Message[]>;
```

SpaceEEC (changes requested):

typings/index.d.ts
```diff
@@ -2377,20 +2371,20 @@ declare module 'discord.js' {
...
       options: (WebhookMessageOptions & { split?: false }) | MessageAdditions,
     ): Promise<Message | RawMessage>;
     send(
-      content: StringResolvable,
+      content: string | null,
       options: WebhookMessageOptions & { split: true | SplitOptions },
     ): Promise<(Message | RawMessage)[]>;
     send(
-      content: StringResolvable,
+      content: string,
```
SpaceEEC, line 2387:
```suggestion
      content: string | null,
```

SpaceEEC (changes requested):

typings/index.d.ts
```diff
@@ -1970,25 +1966,25 @@ declare module 'discord.js' {
     public token: string;
     public editMessage(
       message: MessageResolvable,
-      content: APIMessageContentResolvable | APIMessage | MessageEmbed | MessageEmbed[],
+      content: string | APIMessage | MessageEmbed | MessageEmbed[],
```
SpaceEEC, line 1969:
```suggestion
      content: string | null | APIMessage | MessageEmbed | MessageEmbed[],
```

SpaceEEC (changes requested):

SpaceEEC (approved):

PR #4884, opened 2020-10-03
fix(GuildMemberManager): Use actually random nonce in fetch

### LINKED ISSUE #4883: GuildMemberManager.fetch is not reentrant
When calling guild.members.fetch({user: userIds}) a second time before the first call finishes, both Promises return the results of the first fetch.

wasdennnoch:

Fixes #4883
If you call the function mulltiple times immediately after another (within the same millisecond), both request would have the same nonce, resulting in the second request resulting with the data of the first request.
To fix this, make sure to use an actually random nonce when fetching members.

**Status**

### DIFF
--- a/src/managers/GuildMemberManager.js
+++ b/src/managers/GuildMemberManager.js
@@ -264,7 +264,7 @@ class GuildMemberManager extends BaseManager {
     user: user_ids,
     query,
     time = 120e3,
-    nonce = SnowflakeUtil.generate().toString(16),
+    nonce = SnowflakeUtil.generate(),
     force = false,
   } = {}) {
     return new Promise((resolve, reject) => {
--- a/src/managers/GuildMemberManager.js
+++ b/src/managers/GuildMemberManager.js
@@ -263,7 +263,7 @@ class GuildMemberManager extends BaseManager {
     user: user_ids,
     query,
     time = 120e3,
-    nonce = Date.now().toString(16),
+    nonce = Math.random().toString(16),
     force = false,
   } = {}) {
     return new Promise((resolve, reject) => {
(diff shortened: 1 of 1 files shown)


src/managers/GuildMemberManager.js
```diff
@@ -263,7 +263,7 @@ class GuildMemberManager extends BaseManager {
     user: user_ids,
     query,
     time = 120e3,
-    nonce = Date.now().toString(16),
+    nonce = Math.random().toString(16),
```
kyranet, line 266: Sadly, this is [bound to collisions](https://stackoverflow.com/questions/578700/how-trustworthy-is-javascripts-random-implementation-in-various-browsers), which is why we have crypto.

However, we have [`Snowflake.generate()`](https://github.com/discordjs/discord.js/blob/d2341654fe6ef87c6048bb5bf74c394062f3f618/src/util/Snowflake.js#L29-L47), which can generate up to 4095 different IDs in the same millisecond - very unlikely for it to hit.

kyranet (changes requested):

src/managers/GuildMemberManager.js
wasdennnoch: While that is true, how high are the chances of actually having a collision within the short time frame that gulld member requests nonces are valid?

kyranet (approved):

src/managers/GuildMemberManager.js
kyranet: It is actually very low, but I wouldn't like to take chances :)

src/managers/GuildMemberManager.js
```diff
@@ -263,7 +264,7 @@ class GuildMemberManager extends BaseManager {
     user: user_ids,
     query,
     time = 120e3,
-    nonce = Date.now().toString(16),
+    nonce = SnowflakeUtil.generate().toString(16),
```
vladfrangu, line 267:
```suggestion
    nonce = SnowflakeUtil.generate(),
```

vladfrangu (changes requested):

vladfrangu (approved):

iCrawl (approved):

PR #4895, opened 2020-10-06
feat(Message): allow custom emoji format for react

### LINKED ISSUE #4006: message.react does not correctly resolve custom emojis using identifier
**Update, not a bug - feature enhancement.**

Assumed the below should have worked as expected, but is not the case - my bad.

For my use case, I'm retrieving an emoji identifier via `client.emojis.resolveIdentifier` using `<:name:id>`. Passing the resolved identifier value to `message.react` throws the following API error:

Also occurs if you first resolve the GuildEmoji from `guild.emojis.resolve` and pass the object to `client.emojis.resolveIdentifier`.

```
emoji_id: Value "693728847857713192>" is not snowflake.
```

should be able to reproduce this by having the following code in your message listener, replacing **custom_emoji** with a valid custom emoji `<:name:id>`. I just typed this out here for minimal example.

```js
const emojiIdentifier = message.client.emojis.resolveIdentifier(custom_emoji);
if (emojiIdentifier) message.react(emojiIdentifier);
```

**Further details:**

- discord.js version: 12.1.1
- Node.js version: 13.5.0
- Operating system: Windows and OSX
- Priority:
Low , shouldn't be too difficult to fix and it impacts a pretty common / straight forward feature. (I'd pick it up myself, but I don't have the time right now, unfortunately)
- [x ] I have also tested the issue on latest master, commit hash: 6544d2233804f01b08d15a88edf9b5331949353d

almostSouji:

- ⚠resolves #4006

Currently the `"<:name:id>"` and `"<a:name:id>"` emoji "mention" format strings are not valid input to Message#react, as #4006 brought up. This PR aims to support this input by utilizing the libraries [Util.parseEmoji](https://discord.js.org/#/docs/main/stable/class/Util?scrollTo=s-parseEmoji) method.

The length check is required as the parseEmoji function also returns a valid object with nonsensical input. A name will always be present, if the emoji is valid - either by the custom emoji name or the unicode representation.

**Status**

### DIFF
--- a/src/managers/GuildEmojiManager.js
+++ b/src/managers/GuildEmojiManager.js
@@ -123,11 +123,10 @@ class GuildEmojiManager extends BaseManager {
     if (typeof emoji === 'string') {
       const res = parseEmoji(emoji);
       if (res && res.name.length) {
-        return `${res.animated ? 'a:' : ''}${res.name}${res.id ? `:${res.id}` : ''}`;
+        emoji = `${res.animated ? 'a:' : ''}${res.name}${res.id ? `:${res.id}` : ''}`;
       }
-      if (!emoji.includes('%')) {
-        return encodeURIComponent(emoji);
-      }
+      if (!emoji.includes('%')) return encodeURIComponent(emoji);
+      else return emoji;
     }
     return null;
   }
--- a/src/managers/GuildEmojiManager.js
+++ b/src/managers/GuildEmojiManager.js
@@ -6,6 +6,7 @@ const GuildEmoji = require('../structures/GuildEmoji');
 const ReactionEmoji = require('../structures/ReactionEmoji');
 const Collection = require('../util/Collection');
 const DataResolver = require('../util/DataResolver');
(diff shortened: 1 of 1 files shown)


src/managers/GuildEmojiManager.js
```diff
@@ -119,6 +121,10 @@ class GuildEmojiManager extends BaseManager {
     if (emojiResolvable) return emojiResolvable.identifier;
     if (emoji instanceof ReactionEmoji) return emoji.identifier;
     if (typeof emoji === 'string') {
+      const res = parseEmoji(emoji);
+      if (res.name.length) {
```
Papaia, line 125: It might be null:
https://github.com/discordjs/discord.js/blob/77b6a7d5bd7463f9f397179f5897ec0d00706749/src/util/Util.js#L262-L268
```suggestion
      if (res && res.name.length) {
```

kyranet (approved):

vladfrangu (approved):

iCrawl (approved):

PR #4906, opened 2020-10-15
feat: create 'GuildTemplate' & add 'Client#fetchGuildTemplate()'

izexi:
- Ref: https://github.com/discord/discord-api-docs/pull/1712
 - Adds `DataResolver#resolveGuildTemplateCode()` to resolve template URLs
 - Along with `DataResolver#resolveCode()` to reuse code for the above and the `resolveInviteCode` method
 - Creates a `GuildTemplate` structure with a `createGuild` method to add support for the `POST /guilds/templates/:template_code` endpoint
 - Adds `Client#fetchGuildTemplate()` to support for the `GET /guilds/templates/:template_code` endpoint
 - Resolves #4651 

<details>
<summary>Here's a snippet of how I tested these changes</summary>

```js
import { once } from 'events';
import { Client, DataResolver, GuildTemplate } from 'discord.js';
import { TEST_TEMPLATE_CODE, TOKEN } from './constants';

const client = new Client();
let createdGuild;

client.login(TOKEN);

test('DataResolver#resolveGuildTemplateCode()', () => {
  expect(
    DataResolver.resolveGuildTemplateCode(
      'https://discord.com/template/FKvmczH2HyUf'
    )
  ).toBe('FKvmczH2HyUf');

  expect(
    DataResolver.resolveGuildTemplateCode('https://discord.new/FKvmczH2HyUf')
  ).toBe('FKvmczH2HyUf');

  expect(
    DataResolver.resolveGuildTemplateCode('https://discord.invalid/code')
  ).toBe('https://discord.invalid/code');
});


test('Client#fetchGuildTemplate() & GuildTempalte#createGuild()', async () => {
  await once(client, 'ready');

  const template = await client.fetchGuildTemplate(TEST_TEMPLATE_CODE);
  expect(template).toBeInstanceOf(GuildTemplate);

  const createdGuildName = 'foo';
  createdGuild = await template.createGuild(createdGuildName);
  expect(createdGuild.name).toBe(createdGuildName);
});

afterAll(() => createdGuild.delete());
```

</details>

**Status**

- [x] Code changes have been tested against the Discord API, or there are no code changes
- [x] I know how to update typings and have done so, or typings don't need updating

**Semantic versioning classification:**

- [x] This PR changes the library's interface (methods or parameters added)
  - [ ] This PR includes breaking changes (methods removed or renamed, parameters moved or removed)
- [ ] This PR **only** includes non-code changes, like changes to documentation, README, etc.

PR #4907, opened 2020-10-15
feat: add support for guild templates

izexi:
- Ref: https://github.com/discord/discord-api-docs/pull/2144
 - Continuation of #4906 (branched from there and separated this into another PR since the referenced PR of the docs above containing these changes hasn't been merged yet)

<details>
<summary>Here's a snippet of how I tested these changes</summary>

```js
import { once } from 'events';
import { Client, GuildTemplate } from 'discord.js';
import { TEST_GUILD_ID, TOKEN } from './constants';

const client = new Client();
let createdChannel;

client.login(TOKEN);

test('Client#fetchGuildTemplate() & GuildTempalte#createGuild()', async () => {
  await once(client, 'ready');

  const guild = client.guilds.cache.get(TEST_GUILD_ID);

  const template = await guild.createTemplate('bar');
  expect(template).toBeInstanceOf(GuildTemplate);

  await guild
    .fetchTemplates()
    .then((templates) =>
      expect(templates.size === 1 && templates.has(template.code)).toBe(true)
    );

  await template
    .edit({ description: 'baz' })
    .then((t) => expect(t.description).toBe('baz'));

  const createdChannelName = 'qux';
  createdChannel = await guild.channels.create(createdChannelName);
  await template
    .sync()
    .then((t) =>
      expect(
        t.guild.channels.cache.some(({ name }) => name === createdChannelName)
      ).toBe(true)
    );

  await template.delete();
  await expect(
    client.fetchGuildTemplate(template.code).catch(() => null)
  ).resolves.toBe(null);
});

afterAll(() => createdChannel.delete());
```

</details>

**Status**


src/structures/GuildTemplate.js
```diff
@@ -0,0 +1,197 @@
...
+     */
+    this.guild = this.client.guilds.add(data.serialized_source_guild, false);
+
+    /**
+     * Whether this template has unsynced changes
+     * @type {?boolean}
+     */
+    this.unSynced = 'is_dirty' in data ? Boolean(data.is_dirty) : null;
+
+    return this;
```
NotSugden, line 94: _patch usually returns void, it would be inconsistent to change this wouldn't it?

izexi: I don't see any real downsides of having it return the instance itself, I made this change to avoid having to have some code duplication over on `edit/sync()` (`this._patch(data); return this;`). One thing to also note, is that in [`VoiceState#_patch()`](https://github.com/discordjs/discord.js/blob/master/src/structures/VoiceState.js#L71) the `VoiceState` instance is also returned over there.

advaith1:
docs PR has been merged

vladfrangu (approved):

iCrawl (approved):

kyranet (approved):

src/util/Constants.js
```diff
@@ -81,12 +81,14 @@ exports.DefaultOptions = {
...
    * @property {string} [cdn='https://cdn.discordapp.com'] Base url of the CDN
    * @property {string} [invite='https://discord.gg'] Base url of invites
+   * @property {string} [template='https://discord.com/template'] Base url of templates
    */
   http: {
     version: 7,
     api: 'https://discord.com/api',
     cdn: 'https://cdn.discordapp.com',
     invite: 'https://discord.gg',
+    template: 'https://discord.com/template',
```
SpaceEEC, line 91: The client is using this other domain in the ui.
```suggestion
    template: 'https://discord.new',
```

It looks like neither endpoint / domain is documented however, or is this explicitly documented somewhere?

src/structures/GuildTemplate.js
```diff
@@ -0,0 +1,197 @@
...
+
+    /**
+     * The amount of times this template has been used
+     * @type {number}
+     */
+    this.usageCount = data.usage_count;
+
+    /**
+     * The ID of the user that created this template
+     * @type {string}
```
SpaceEEC, line 54:
```suggestion
     * @type {Snowflake}
```

src/structures/GuildTemplate.js
```diff
@@ -0,0 +1,197 @@
...
+    this.updatedAt = new Date(data.updated_at);
+
+    /**
+     * The ID of the guild that this template belongs to
+     * @type {string}
+     */
+    this.guildID = data.source_guild_id;
+
+    /**
+     * The guild that this template belongs to
```
SpaceEEC, line 83:
```suggestion
     * The guild that this template would create
```
Or something like this?

Since this is very much a different guild (as it's the serialized template)

src/structures/GuildTemplate.js
```diff
@@ -0,0 +1,197 @@
...
+     * @type {Guild}
+     */
+    this.guild = this.client.guilds.add(data.serialized_source_guild, false);
```
SpaceEEC, line 86: This will insert the template channels into `Client#channels`, ignoring the cache parameter.

https://github.com/discordjs/discord.js/blob/eaecd0e8b790f218b2c52021b433c0957ec539c3/src/structures/Guild.js#L396-L401

Not too sure how to go about this, we could go with just the raw object. Probably not the most handy thing to work with however, not that I can think of many use cases.

src/structures/GuildTemplate.js
```diff
@@ -0,0 +1,197 @@
...
+     * The user that created this template
+     * @type {User}
+     */
+    this.creator = this.client.users.add(data.creator);
+
+    /**
+     * The time of when this template was created at
+     * @type {Date}
+     */
+    this.createdAt = new Date(data.created_at);
```
SpaceEEC, line 68: There probably should be a `createdTimestamp` getter for this, to stay consistent with the rest of the library.

src/structures/GuildTemplate.js
```diff
@@ -0,0 +1,197 @@
...
+
+    /**
+     * The time of when this template was last synced to the guild
+     * @type {Date}
+     */
+    this.updatedAt = new Date(data.updated_at);
```
SpaceEEC, line 74: There probably should be a `updatedTimestamp` getter for this, to stay consistent with the rest of the library.

src/util/Constants.js
```diff
@@ -81,12 +81,14 @@ exports.DefaultOptions = {
    * @property {string} [api='https://discord.com/api'] Base url of the API
    * @property {string} [cdn='https://cdn.discordapp.com'] Base url of the CDN
    * @property {string} [invite='https://discord.gg'] Base url of invites
+   * @property {string} [template='https://discord.com/template'] Base url of templates
```
SpaceEEC, line 84: That is what the client is using, see my comment below.
```suggestion
   * @property {string} [template='https://discord.new'] Base url of templates
```

typings/index.d.ts
```diff
@@ -852,6 +857,26 @@ declare module 'discord.js' {
...
   }
 
+  export class GuildTemplate extends Base {
+    constructor(client: Client, data: object);
+    public readonly url: string;
+    public code: string;
+    public name: string;
+    public description: string | null;
+    public usageCount: number;
+    public creatorID: string;
```
SpaceEEC, line 867:
```suggestion
    public creatorID: Snowflake;
```

typings/index.d.ts
```diff
@@ -852,6 +857,26 @@ declare module 'discord.js' {
...
+    public creator: User;
+    public createdAt: Date;
+    public updatedAt: Date;
+    public guildID: string;
```
SpaceEEC, line 871:
```suggestion
    public guildID: Snowflake;
```

src/structures/GuildTemplate.js
SpaceEEC: I think it's a bit misleading that `guildID` and `guild` represent two different guilds, it probably is better to rename this field to something that properly reflects that this is the actual template, not "just" the source guild itself.

src/structures/GuildTemplate.js
```diff
@@ -0,0 +1,197 @@
...
+
+    /**
+     * The ID of the guild that this template belongs to
+     * @type {string}
+     */
+    this.guildID = data.source_guild_id;
```
SpaceEEC, line 80: With the field below renamed we could add a getter for the source guild.
Although one could argue to not have that one here to avoid confusion between source guild and serialized source guild.

SpaceEEC (changes requested):
You could also add the error `30031` (A server can only have a single template.).
See: https://github.com/discord/discord-api-docs/issues/2143#issuecomment-708091836

src/util/Constants.js
advaith1: discord.gg isn't documented anywhere either (they both should probably be added to the reference page)

advaith1:
@SpaceEEC that was added in #2144

src/util/Constants.js
izexi: `https://discord.new/:template_code` seems to redirect to `https://discord.com/template/:template_code`, so I thought that maybe Discord was perhaps using the `.new` domain for templates temporarily or something (?) which is why I decided to go with `.com/template`. I could be wrong, but I feel like it would be a safer bet to use this one anyways?

vladfrangu: It's not temporary iirc, just like how the gift domain exists, there's one for templates

src/structures/GuildTemplate.js
izexi: Renamed the prop to `serializedGuild`, also went with just returning the raw object and threw in a link on the JSDoc to the guild structure over on the API docs.

SpaceEEC (approved):

PR #4915, opened 2020-10-19
fix: handle crash when fetching members with invalid IDs

### LINKED ISSUE #4896: bot crashes with uncatchable error from members.fetch
When running `message.guild.members.fetch({ user: [] })` with an invalid snowflake like `a`
ex `message.guild.members.fetch({ user: ['a', '1'] })` it seems to crash the node process without any errors. Even a catch doesn't handle it since there's no error thrown. Then after exactly 120 seconds the bot makes a successful connection to the gateway in which it then throws the error but the node process is still in a crashed state in which the bot never fully reconnected. I do use PM2 so pm2 is configured to restart a crashed node process. Though the process isn't fully crashed. The only thing that's emitted in the console is from the Shard reconnect event right when the bot goes offline.

You can view the video I recorded of it happening here https://streamable.com/f3y0mm once you reach the 11 second mark you can skip the video to the 2 minute mark, everything between that is just untrimmed video of it showing it's still offline. But at the 2:11 mark you can see the bot throws the error but is still offline.

With the debug logs enabled this is all that's emitted
```
[2020-10-08 17:03:12] [WS => Shard 0] [INVALID SESSION] Resumable: false.
[2020-10-08 17:03:12] [SHARD RECONNECTING] Shard 0
```

```js
message.guild.members.fetch({ user: ['a', '1'] })
```

**Further details:**

- discord.js version: master
- Node.js version: v14.12.0
- Operating system: Ubuntu 18.04
- Priority this issue should have – please be realistic and elaborate if possible: medium-high.  Medium since you shouldn't be providing invalid snowflakes anyways and high since it causes a crash/delay that has no way of handling it.

**Relevant client options:**

- partials: none
- gateway intents: none
- other: none

- [x] I have also tested the issue on latest master, commit hash: d234165

TeeSeal:

Resolves #4896 

 Suggestions to improve `Snowflake.maybeSnowflake()` are most welcome.

**Status**


src/structures/Base.js
```diff
@@ -37,6 +38,15 @@ class Base {
   valueOf() {
     return this.id;
   }
+
+  /**
+   * Tells whether a given string may be used as an ID for an instance of this class
```
NotSugden, line 43:
```suggestion
   * Tells whether a given string may be used as an ID for an instance of this class.
```

typings/index.d.ts
```diff
@@ -1422,6 +1423,7 @@ declare module 'discord.js' {
   export class SnowflakeUtil {
     public static deconstruct(snowflake: Snowflake): DeconstructedSnowflake;
     public static generate(timestamp?: number | Date): Snowflake;
+    public static maybeSnoflake(string: string): boolean;
```
NotSugden, line 1426:
```suggestion
    public static maybeSnowflake(string: string): boolean;
```

BrunoS3D (approved):
LGTM 🚀

caio-vinicius (approved):

caio-vinicius:
Please, we need these changes

src/client/websocket/WebSocketManager.js
```diff
@@ -233,17 +233,17 @@ class WebSocketManager extends EventEmitter {
...
 
-      shard.on(ShardEvents.INVALID_SESSION, () => {
+      const reconnectShard = (() => {
         this.client.emit(Events.SHARD_RECONNECTING, shard.id);
-      });
+        this.shardQueue.add(shard);
+        this.reconnect();
+      }).bind(this);
+
+      shard.on(ShardEvents.INVALID_SESSION, reconnectShard);
```
vladfrangu, line 242: Thanks a lot for finding an edge case with WS (again ;;)! However, this isn't the right fix imho. Honestly, if we get an invalid session that's nit resumable, we should just destroy the shard and clean it up, unless other's have any other suggestions 👀

vladfrangu (changes requested):
Other than the open discussion on the ws side, I believe that we shouldn't have to handle invalid IDs ourselves and have to wait `n` seconds if an invalid one is passed in this example... Can someone bring this issue up to Discord? It seems kinda.. broken to hang bots and lose potential payloads in the process :(

src/client/websocket/WebSocketManager.js
TeeSeal: I mean, it is the Discord docs that say that upon receiving this opcode we should resume/reconnect: https://discord.com/developers/docs/topics/opcodes-and-status-codes

image

The way it works currently, is that the connection is lost, the shard is not properly destroyed and the connection is kept intact, while discord has destroyed the session on their end, leaving the shard in a sort of limbo where it keeps sending packets but receives nothing from Discord.

TeeSeal:
> Other than the open discussion on the ws side, I believe that we shouldn't have to handle invalid IDs ourselves and have to wait `n` seconds if an invalid one is passed in this example... Can someone bring this issue up to Discord? It seems kinda.. broken to hang bots and lose potential payloads in the process :(

I completely agree with you. It should not be our responsibility to validate ID format. I just thought that this issue, considering its gravity, needs at least a temporary work around until Discord gets things sorted on their end, to prevent library users from encountering this weird behavior. I honestly don't know how long it usually takes them to fix these kinds of issues.

The other solution here would be to remove the ID validation and just go with the 3 second hang and have the shard reconnect after and maybe kindly ask our library users to pay utmost attention to the IDs they pass into UserManager#fetch. If we go with this, we won't have to make any changes when Discord finally fixes this.

TeeSeal:
Actually, scratch that. It seems while this PR was open Discord has already fixed the problem on their end. I just performed the same tests I did in https://github.com/discordjs/discord.js/issues/4896#issuecomment-707645960 and no longer see the same behavior. Now when providing invalid IDs Discord seems to properly return a not_found response:

```
{
  t: 'GUILD_MEMBERS_CHUNK',
  s: 4,
  op: 0,
  d: {
    not_found: [ null, null ],
    nonce: '1756a7cb8fd',
    members: [],
    guild_id: '117271426867789833',
    chunk_index: 0,
    chunk_count: 1
  }
}
```
Hence, I will remove the ID validation as it doesn't fit very well in the lib. But I still think we should do something about opcode 9 (INVALID_SESSION).

src/client/websocket/WebSocketManager.js
vladfrangu: Can we test this? I mean, the docs want us to close connection, then re-identify/resume... This would keep the conn alive (and just re-identify).. Is there even any way to test this?

TeeSeal:
Ok nevermind again, I asked the person who opened the issue to re-test and they still see the issue, and after they said that so do I. I think I might have used the updated branch when testing the first time and thought that Discord fixed this on their end 😄  Or Discord is being inconsistent in its behavior, but I'd say the former is more probable

src/client/websocket/WebSocketManager.js
TeeSeal: I did test this.

1. Apply only commit 43218347f0ddad180a9fe67e1b08c8065c0928d6
2. Do `someGuild.members.fetch({ user: ['a', '1'] })`
3. Watch the client hang for a few seconds then receive `Invalid Session`

With my changes it seems to reconnect alright after that.

vladfrangu:
Could it be different gateway versions have different outputs?

TeeSeal:
> Could it be different gateway versions have different outputs?

I tried both v6 and v7 and both seem to suffer from this problem. I opened an issue with the Discord API to discuss this and maybe come to a resolution, see what they say. https://github.com/discord/discord-api-docs/issues/2180

BannerBomb:
@TeeSeal They've responded back to the issue report on the discord-api github. https://github.com/discord/discord-api-docs/issues/2180#issuecomment-744131302

TeeSeal:
> @TeeSeal They've responded back to the issue report on the discord-api github. https://github.com/discord/discord-api-docs/issues/2180#issuecomment-744131302

Well, they pretty much said: don't send non-numeric IDs 🤷‍♂️ We'll need someone with more authority to decide whether this is worth handling on the library side. I'm kinda siding with leaving this as is because:
- It is not the library's fault that the user is unable to validate their inputs and sends non-numeric IDs;
- It is not the library's fault that the Discord gateway explodes when receiving such IDs

typings/index.d.ts
```diff
@@ -114,6 +114,7 @@ declare module 'discord.js' {
     public readonly client: Client;
     public toJSON(...props: { [key: string]: boolean | string }[]): object;
     public valueOf(): string;
+    public static isValidID(string: string): boolean;
```
Vaporox, line 117:
```suggestion
    public static isValidID(string: string): string is Snowflake;
```

typings/index.d.ts
```diff
@@ -1422,6 +1423,7 @@ declare module 'discord.js' {
   export class SnowflakeUtil {
     public static deconstruct(snowflake: Snowflake): DeconstructedSnowflake;
     public static generate(timestamp?: number | Date): Snowflake;
+    public static maybeSnowflake(string: string): boolean;
```
Vaporox, line 1426:
```suggestion
    public static maybeSnowflake(string: string): string is Snowflake;
```

Vaporox (commented):
This will be useful in combination with #5157

typings/index.d.ts
```diff
@@ -1422,6 +1423,7 @@ declare module 'discord.js' {
   export class SnowflakeUtil {
     public static deconstruct(snowflake: Snowflake): DeconstructedSnowflake;
     public static generate(timestamp?: number | Date): Snowflake;
+    public static maybeSnowflake(string: string): string is Snowflake;
```
Mesteery, line 1426: Literally `string is string`...
But can be valid after https://github.com/discordjs/discord.js/pull/5157.

typings/index.d.ts
```diff
@@ -114,6 +114,7 @@ declare module 'discord.js' {
     public readonly client: Client;
     public toJSON(...props: { [key: string]: boolean | string }[]): object;
     public valueOf(): string;
+    public static isValidID(string: string): string is Snowflake;
```
Mesteery, line 117: same than https://github.com/discordjs/discord.js/pull/4915#discussion_r552859828.
```suggestion
    public static isValidID(id: string): string is Snowflake;
```

typings/index.d.ts
kyranet: I suggest `value`:
```suggestion
    public static maybeSnowflake(value: string): value is Snowflake;
```

TeeSeal: I mean, I don't really mind changing it but I'd argue this is an improvement. `string` describes exactly what that argument is: any given string. If anything, `value` is even more broad than `string` is. I do agree that there are cases when using `string` as an argument name is no good, but I reckon this is not one of them.

Give me an objective reason `string` is bad and why you think other names are better and I will change it straight away. Subjective thoughts and wishes should have no influence on code.

Mesteery: The `string` argument name, is very confusely with the type.
Moreover, `string` is not really good for an argument name, because it does not describe the argument. `string` can tell anything. `value` is more precise, less confusing and better.

TeeSeal: I agree with the first part: it can be confused with the type. I will update it in regards to that.

Though your second argument really makes no sense to me. `value` is more precise in what way exactly? `string` tells you it's a string, `value` can be anything 😄

kyranet: Not really, because it's typed as a `string`, it doesn't just accept any *value*.

As far as that typing is concerned, it reads as a "value of type string", which sounds better than "string of type string".

Mesteery: For me, it's more precise than `string`. There are maybe better names than `value` (who is better than `string`). `possibleSnowflake` (no ?) ?

TeeSeal: Updated to `value`

typings/index.d.ts
TeeSeal: ditto: https://github.com/discordjs/discord.js/pull/4915#discussion_r552961662

catcosmicice (approved):

kyranet:
Closing as the PR is stale, we don't plan to validate strings, and the issue has been marked as user error.

PR #4917, opened 2020-10-19
fix(GuildEmojiManager): check for guild in methods that use it

Vaporox:

`Client#emojis` and `Guild#emojis` both return a `GuildEmojiManager`, however the current code base does not account for former not having a valid guild property. `client.emojis.create(...)` before resulted in `DiscordAPIError: 405: Method Not Allowed`, this PR aims to fix that by checking if a guild is present before using it for methods and throwing a custom error.

**Consideration**

There's also the option to create two separate classes for `Client#emojis` and `Guild#emojis`, I've decided to take this approach for now but I'm open to suggestions.

**Status**

- [x] Code changes have been tested against the Discord API

### DIFF
--- a/src/errors/Messages.js
+++ b/src/errors/Messages.js
@@ -103,6 +103,8 @@ const Messages = {
   FETCH_GROUP_DM_CHANNEL: "Bots don't have access to Group DM Channels and cannot fetch them",
 
   MEMBER_FETCH_NONCE_LENGTH: 'Nonce length must not exceed 32 characters.',
+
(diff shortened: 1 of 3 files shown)


SpaceEEC:
Wouldn't it be more elegant to move `resolve`, `resolveID`, and `resolveIdentifier` to a new class (maybe just `EmojiManager`?) and have `GuildEmojiManager` extend that?

That way there wouldn't be any `add` or `create` methods to call there to begin with.

Vaporox:
This probably sounds like a better idea than what I did. Should I change that on this branch or open a new PR, since it's a completely different approach?

PR #4935, opened 2020-10-24
fix(RoleManager): fix ID return value, change return type to collection

Vaporox:

1. When using `RoleManager#fetch` with a single ID, [it tries to get the role from `this.cache`](https://github.com/discordjs/discord.js/blob/master/src/managers/RoleManager.js#L58), however that won't work when `cache` is set to `false` and it will return `null` even when the role was fetched correctly. This has ben fixed by getting the role from the collection instead.

2. Unlike any other manager, fetching without passing an ID returns the manager instead of the fetched collection, this has been changed for consistency.

**Status**

- [x] Code changes have been tested against the Discord API

### DIFF
--- a/src/managers/RoleManager.js
+++ b/src/managers/RoleManager.js
@@ -2,6 +2,7 @@
 
 const BaseManager = require('./BaseManager');
 const Role = require('../structures/Role');
+const Collection = require('../util/Collection');
 const Permissions = require('../util/Permissions');
 const { resolveColor } = require('../util/Util');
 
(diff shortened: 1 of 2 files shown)


src/managers/RoleManager.js
```diff
@@ -34,7 +35,7 @@ class RoleManager extends BaseManager {
    * @param {Snowflake} [id] ID or IDs of the role(s)
    * @param {boolean} [cache=true] Whether to cache the new roles objects if it weren't already
    * @param {boolean} [force=false] Whether to skip the cache check and request the API
-   * @returns {Promise<Role|RoleManager>}
+   * @returns {Promise<Role|Collection<Snowflake, Role>>}
```
Extroonie, line 38:
```suggestion
   * @returns {Promise<?Role|Collection<Snowflake, Role>>}
```
it can be nullable here

vladfrangu (approved):

SpaceEEC (approved):

kyranet (approved):

iCrawl (approved):

PR #4944, opened 2020-10-28
docs: document required intents

advaith1:

This PR adds the Gateway Intent required, or the impact of Intents, to descriptions of events and properties.

**Status**

- [x] This PR **only** includes non-code changes, like changes to documentation, README, etc.

### DIFF
--- a/src/client/actions/MessageReactionRemove.js
+++ b/src/client/actions/MessageReactionRemove.js
@@ -31,6 +31,7 @@ class MessageReactionRemove extends Action {
     reaction._remove(user);
     /**
      * Emitted whenever a reaction is removed from a cached message.
+     * Requires the GUILD_MESSAGE_REACTIONS or DIRECT_MESSAGE_REACTIONS intent.
      * @event Client#messageReactionRemove
      * @param {MessageReaction} messageReaction The reaction object
      * @param {User} user The user whose emoji or reaction emoji was removed
--- a/src/client/actions/MessageReactionRemove.js
+++ b/src/client/actions/MessageReactionRemove.js
@@ -31,7 +31,7 @@ class MessageReactionRemove extends Action {
     reaction._remove(user);
     /**
      * Emitted whenever a reaction is removed from a cached message.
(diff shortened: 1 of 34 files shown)


src/client/actions/MessageReactionRemove.js
```diff
@@ -31,6 +31,7 @@ class MessageReactionRemove extends Action {
     reaction._remove(user);
     /**
      * Emitted whenever a reaction is removed from a cached message.
+     * Requires the GUILD_MESSAGE_REACTIONS or DIRECT_MESSAGE_REACTIONS intent.
```
vladfrangu, line 34: Might need to specify which is for what (so people don't expect DIRECT_MESSAGE_* to return things from guilds)

vladfrangu (changes requested):

src/client/actions/MessageReactionRemove.js
advaith1: isn't that kinda common sense though

vladfrangu: You'd be surprised... 😓

Papaia:
I don't think this is something the library has to document, it is documented in the Discord gateway reference, which the Intents class and guide return to.
We might want to document which event is which (`GUILD_MEMBER_ADD` <-> `guildMemberAdd`), but that seems pretty clear.

advaith1:
@Papaia I mainly did this to document it on *properties* because many properties returned by discord.js are actually updated by events which aren't obvious (for example, guild.memberCount doesn't get updated without the members intent); since these are created by discord.js they can't be in Discord docs.

And I also documented them on the events because no point in doing it for some but not others

advaith1:
why was this closed? I thought Papaia had accepted my explanation, and it's important to document how intents can affect properties made by the lib

iCrawl:
We never documented anything related to permissions so far, so I don't really see why we would document intents now.

As long as the permissions don't get documented, we prolly won't document intents. Feels 50/50

advaith1:
Would the intents page in the guide be a good place to put this?

iCrawl:
That sounds like a more reasonable approach to me.

PR #4953, opened 2020-10-29
Add intents to the DefaultOptions

Darqam:
This PR adds the intents key to DefaultOptions in order to be able to access it (if set) through client.options.ws.intents.

undefined was chosen as a value since neither souji nor I were sure what to set it at and "it works". If a better value is needed, that can evidently be changed.

I also state typings do not need to be changed as it seems like this ends up being more of a fix than a modification, `intents` is already present in the wsoptions typings.

**Status**


SpaceEEC:
I can access the intents through `client.options.ws.intents` just fine when specifying them when creating my client instance.

PR #4955, opened 2020-10-29
[BREAKING] refactor(BitFields): require use of flags

Papaia:
Requires the use of BitField flags instead of strings.
```diff
- permissions.has('SEND_MESSAGES')
+ permissions.has(Permissions.FLAGS.SEND_MESSAGES)
```
**Status**


jxpsert:
Why? What's wrong with using strings?
This would break a lot of users' code for essentially no reason.

kyranet:
> Why? What's wrong with using strings?
> This would break a lot of users' code for essentially no reason.

My opinion on this is that there are many ways in which this can go wrong, you can easily check in the Issues tab for issues we have when passing invalid permission strings. This is not an issue when using FLAGS.

This PR (as well as many others) can mean the start of removing "convenience" stuff - `channel.type` being a string instead of the raw number from Discord, for example - and hopefully save a lot of memory. This is something I already tested and saw large improvements.

jxpsert:
Seems fair. Though, this would still break hundreds of bots. Maybe accept both strings and flags?

However, removing the 'convenience' in Channel#type is just unnecessary, if you ask me.

Papaia:
I couldn't have explained it better than Kyra.
More "convenience" removes could be things like [`GuildChannel#viewable`](https://discord.js.org/#/docs/main/master/class/GuildChannel?scrollTo=viewable) which is basically another getter for every instance of a guild channel just for one simple permissions check that can be done by the user.

jxpsert:
Yeah, but checking permissions is easier than having to remember/look up channel type numbers all the time. I'd agree with removing stuff like viewable, which are just shortcuts.

kyranet:
> However, removing the 'convenience' in Channel#type is just unnecessary, if you ask me.

I meant changing `channel.type` so it doesn't look-up from an enum, but rather use the raw number. You can always check if a channel is a text channel by doing `channel.type === ChannelTypes.Text` as opposed to `channel.type === 'text'`.

**Advantages include:**

- If Discord adds a new channel type, we can still handle it, the number is more helpful than `'unknown'` (which hides the real type).
- Less processing at start-up, resulting on a faster boot time.
- Overall faster, it's less expensive to compare integers than it is to compare strings.
- A lot smaller, a string has a size header (64 bits) followed by either 8 or 16 bits per character (in V8 at least), versus a fixed 32 bits.
- If we rename a type, you don't have to update (`'announcement'` -> `'news'` under the same type number, been there, done that).

**Disadvantages include:**

- Harder to debug, a number doesn't really tell you the actual representation.
- Can't really check easily without using a constant.

> If we go for prioritizing constants over convenience strings, then this kind of change should probably be done. We have an issue and that is discord.js using more memory than many other libraries (the bar is set at ~8.4mb for used heap having nothing cached, but grows a lot faster when the library starts caching objects Discord gives to us).

---

Anyways, this conversation is getting off-topic, I'd suggest continuing this in `#library-discussion` from the official server. :)

typings/index.d.ts
```diff
@@ -2215,7 +2215,6 @@ declare module 'discord.js' {
 
   type BitFieldResolvable<T extends string> =
     | RecursiveReadonlyArray<T | number | Readonly<BitField<T>>>
```
NotSugden, line 2217: wouldn't the `T` from here also need removing?

NotSugden:
docs examples need to be changed - for example the example on `Client#generateInvite`

NotSugden:
`Role#setPermissions` example too needs updating too, im sure there are more too

src/client/Client.js
```diff
@@ -360,7 +360,11 @@ class Client extends BaseClient {
    * @returns {Promise<string>}
    * @example
    * client.generateInvite({
-   *   permissions: ['SEND_MESSAGES', 'MANAGE_GUILD', 'MENTION_EVERYONE'],
+   *   permissions: [
+   *     Permsisions.FLAGS.SEND_MESSAGES,
+   *     Permsisions.FLAGS.MANAGE_GUILD,
+   *     Permsisions.FLAGS.MENTION_EVERYONE,
+   *   ],
```
ckohen, line 367:
```suggestion
   *   permissions: [
   *     Permissions.FLAGS.SEND_MESSAGES,
   *     Permissions.FLAGS.MANAGE_GUILD,
   *     Permissions.FLAGS.MENTION_EVERYONE,
   *   ],
```
Correct typos.

Papaia:
Replaced by #5122.

PR #4933, opened 2020-10-24
feat(GuildEmojiManager): implement GuildEmojiManager#fetch

Extroonie:
**Please describe the changes this PR makes and why it should be merged:**
This PR implements API support for the `GET` [guild-emoji](https://discord.com/developers/docs/resources/emoji#get-guild-emoji)/[list-guild-emojis](https://discord.com/developers/docs/resources/emoji#list-guild-emojis) endpoint(s).
**Status**

- [x] Code changes have been tested against the Discord API, or there are no code changes
- [x] I know how to update typings and have done so, or typings don't need updating

**Semantic versioning classification:**

- [x] This PR changes the library's interface (methods or parameters added)
  - [ ] This PR includes breaking changes (methods removed or renamed, parameters moved or removed)
- [ ] This PR **only** includes non-code changes, like changes to documentation, README, etc.

PR #4966, opened 2020-11-01
feat(GuildChannelManager): add 'fetch' method

extroonie:
This PR implements API support for the `GET` [guild-channels](https://discord.com/developers/docs/resources/guild#get-guild-channels) endpoint.
**Status**


typings/index.d.ts
```diff
@@ -1913,6 +1913,8 @@ declare module 'discord.js' {
       name: string,
       options: GuildCreateChannelOptions,
     ): Promise<TextChannel | VoiceChannel | CategoryChannel>;
+    public fetch(id: Snowflake, cache?: boolean, force?: boolean): Promise<GuildChannel | null>;
+    public fetch(id?: Snowflake, cache?: boolean, force?: boolean): Promise<Collection<Snowflake, GuildChannel>>;
```
Papaia, line 1917:
```suggestion
    public fetch(): Promise<Collection<Snowflake, GuildChannel>>;
```

Extroonie: It may behave incorrectly in such cases, though. Suppose I want to fetch all channels and not cache them, that overwrite will apply there.

Papaia: So you expect the users to call `.fetch(undefined, false)`?

Extroonie: That's what you do to not cache the channels. I don't see why shouldn't we support that?

tipakA: Since it's copy paste of that anyway, how does `roles.fetch()` behave there

monbrey: `guild.roles.fetch()` is inherently a hack because it's not possible to fetch an individual role. If this is intended to replicate that same functionality, make sure you refer to #4935 first

Extroonie: Yeah, I'm completely aware of that. I PRed #4933 first which already fixes that. Should be fine 👍

src/managers/GuildChannelManager.js
```diff
@@ -126,6 +127,36 @@ class GuildChannelManager extends BaseManager {
     });
     return this.client.actions.ChannelCreate.handle(data).channel;
   }
+
+  /**
+   * Obtains one or more guild channels from Discord, or the channel cache if they're already available.
+   * @param {Snowflake} [id] ID of the channel
+   * @param {boolean} [cache=true] Whether to cache the new channel objects if it weren't already
+   * @param {boolean} [force=false] Whether to skip the cache check and request the API
+   * @returns {Promise<GuildChannel|Collection<Snowflake, GuildChannel>>}
```
NotSugden, line 136:
```suggestion
   * @returns {Promise<(?GuildChannel)|Collection<Snowflake, GuildChannel>>}
```
the function can return null here

src/managers/GuildChannelManager.js
```diff
@@ -126,6 +127,36 @@ class GuildChannelManager extends BaseManager {
...
+  async fetch(id, cache = true, force = false) {
+    if (id && !force) {
+      const existing = this.cache.get(id);
+      if (existing) return existing;
+    }
+
+    // We cannot fetch a single guild channel, as of this commit's date, Discord API throws with 404
+    const data = await this.client.api.guilds(this.guild.id).channels.get();
+    const channels = new Collection();
+    for (const channel of data) channels.set(channel.id, this.add(channel, cache));
```
NotSugden, line 157:
```suggestion
    for (const channel of data) channels.set(channel.id, this.client.channels.add(channel, this, cache));
```
`GuildChannelManager#add` doesn't add it to the `Client#channels` `ChannelManager` cache

NotSugden (changes requested):

src/managers/GuildChannelManager.js
```diff
@@ -126,6 +127,36 @@ class GuildChannelManager extends BaseManager {
...
+  async fetch(id, cache = true, force = false) {
+    if (id && !force) {
+      const existing = this.cache.get(id);
+      if (existing) return existing;
+    }
+
+    // We cannot fetch a single guild channel, as of this commit's date, Discord API throws with 404
+    const data = await this.client.api.guilds(this.guild.id).channels.get();
+    const channels = new Collection();
+    for (const channel of data) channels.set(channel.id, this.client.channels.add(channel, this, cache));
```
Awoocado, line 157:
```suggestion
    for (const channel of data) channels.set(channel.id, this.client.channels.add(channel, this.guild, cache));
```

Awoocado: `GuildChannelManager#add`'s second parameter should be a `Guild` instead

iCrawl (approved):

kyranet (approved):

vladfrangu (approved):
LGTM, although it's about time we looked into making fetch fuctions take options objects instead of a list of params... Outside the scope of this PR tho 😅

src/managers/GuildChannelManager.js
```diff
@@ -126,6 +127,36 @@ class GuildChannelManager extends BaseManager {
...
+   *   .then(channel => console.log(`The channel name is: ${channel.name}`))
+   *   .catch(console.error);
+   */
+  async fetch(id, cache = true, force = false) {
+    if (id && !force) {
+      const existing = this.cache.get(id);
+      if (existing) return existing;
+    }
+
+    // We cannot fetch a single guild channel, as of this commit's date, Discord API throws with 404
```
SpaceEEC, line 154: What about `GET /channels/:id`?
It seems to return the same data as `GET /guilds/:id/channels` (except it's only the requested channel instead of all of them).

Extroonie: Yeah, that seems to be possible. But however it's quite inconsistent though? the method suggests `GET /guilds/channels/:id` which apparently does not support getting a single ID. Same with `RoleManager#fetch`. But I'll go with it if it looks fine.

iCrawl:
This needs a rebase.

extroonie:
@iCrawl done

iCrawl:
This needs another rebase. @extroonie

extroonie:
@iCrawl done x2

SpaceEEC (approved):

PR #4967, opened 2020-11-01
fix(Guild): fetch member if already in the guild

SpaceEEC:

This PR fixes a bug that caused an internal error to be thrown when trying add a member to a guild using `Guild#addMember`.
The error was only thrown if the member was already part of the guild, but not in the cache.

This is due to Discord responding with `204 No Content if the user is already a member of the guild.` in that case, as stated in [their Documentation](https://canary.discord.com/developers/docs/resources/guild#add-guild-member).

I solved this by checking for this no content response (which is an array or arrabuffer, depending on the environment), and then fetch the member from the guild to keep the documented return value.

**Status**

### DIFF
--- a/src/structures/Guild.js
+++ b/src/structures/Guild.js
@@ -17,6 +17,7 @@ const RoleManager = require('../managers/RoleManager');
 const VoiceStateManager = require('../managers/VoiceStateManager');
 const Collection = require('../util/Collection');
 const {
+  browser,
   ChannelTypes,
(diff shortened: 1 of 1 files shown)


fernandolguevara:
any update of this?

iCrawl (approved):

kyranet (approved):

vladfrangu (approved):

PR #4969, opened 2020-11-01
fix(MessageReaction*Action): correctly cache incoming members and users

SpaceEEC:

This PR fixes a bug involving partials around the `messageReactionAdd` event causing it to emit a partial user with an id with the value `undefined`, as well as caching incoming members and users from said event and `typingStart`.

The above mentioned bug was caused due to `getUser` being called with a raw user object, rather than an object with an `user_id` key, which it actually is expecting.
https://github.com/discordjs/discord.js/blob/b8fd3f65d964f080afb42dd37eb21a13c0d84f8f/src/client/actions/Action.js#L87-L90

---

`MESSAGE_REACTION_ADD`, as well as `TYPING_START`, emit a member and user, which now will be added to the cache, removing the requirement to enable the `USER` partial type for this event to work on previously uncached users. 

If these are missing members and users (due to not originating from a guild), they will be passed long to `getUser` as before.

Those fields are documented, just not yet implemented by us, see here:
https://discord.com/developers/docs/topics/gateway#message-reaction-add-message-reaction-add-event-fields
https://discord.com/developers/docs/topics/gateway#typing-start-typing-start-event-fields

**Status**

### DIFF
--- a/src/client/actions/Action.js
+++ b/src/client/actions/Action.js
@@ -93,8 +93,9 @@ class GenericAction {
     if (data.guild_id && data.member && data.member.user) {
       const guild = this.client.guilds.cache.get(data.guild_id);
       if (guild) {
-        const member = this.getMember(data.member, guild);
-        return member ? member.user : this.getUser(data.member.user);
+        return guild.members.add(data.member).user;
+      } else {
+        return this.client.users.add(data.member.user);
       }
     }
     return this.getUser(data);
--- a/src/client/actions/MessageReactionAdd.js
+++ b/src/client/actions/MessageReactionAdd.js
@@ -8,7 +8,10 @@ const { PartialTypes } = require('../../util/Constants');
 { user_id: 'id',
      message_id: 'id',
      emoji: { name: '�', id: null },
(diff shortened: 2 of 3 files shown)


vladfrangu (approved):

kyranet (approved):

src/client/actions/Action.js
```diff
@@ -93,8 +93,9 @@ class GenericAction {
     if (data.guild_id && data.member && data.member.user) {
       const guild = this.client.guilds.cache.get(data.guild_id);
       if (guild) {
-        const member = this.getMember(data.member, guild);
-        return member ? member.user : this.getUser(data.member.user);
+        return guild.members.add(data.member).user;
+      } else {
+        return this.client.users.add(data.member.user);
```
advaith1, line 98: when would `guild` not exist?

vladfrangu: Empty cache, incomplete cache... many, many ways `guild` may not exist (and we still get the data so we should let you use it 🤷‍♂️)

iCrawl (approved):

PR #4934, opened 2020-10-24
feat: BaseGuildEmojiManager

Vaporox:
**Please describe the changes this PR makes and why it should be merged:**

[Continuing from #4917] This adds a new manager class `BaseGuildEmojiManager` to be used for `Client#emojis`, `GuildEmojiManager` then extends that to provide methods to interact with the API.

**Status**

- [x] Code changes have been tested against the Discord API
- [x] I know how to update typings and have done so

**Semantic versioning classification:**

- [x] This PR changes the library's interface (methods or parameters added)
  - [x] This PR includes breaking changes (methods removed or renamed, parameters moved or removed)
- [ ] This PR **only** includes non-code changes, like changes to documentation, README, etc.

PR #4983, opened 2020-11-07
feat(Client): add ClientEmojiManager

TheNoob27:

Hey, my first Pull Request!
Having the Client#emojis getter which would cycle through every emoji, add it to a collection and all that, just to check for one emoji isn't really ideal. This ClientEmojiManager takes a different approach, and instead iterates through each *guild* and returns an emoji if the guild has it.
I've tested this a few times versus the previous getter, and it seems to be faster most of the time. It's definitely faster if say you want to get the first cached emoji in the bots first cached guild, because it only iterates through one emoji.
Previously, when I was using the old getter, I would often be getting memory leaks because my bot was very reaction focused and the MessageReaction#emoji getter loads every single emoji into a collection every time the getter is called, but after switching to this it seemed to all go away.

I do have to clarify that I have *not* been able to update the typings.

**Status**


NotSugden:
#4934 is taking a better approach to this issue

TheNoob27:
I guess it is kinda out of place, but I decided to make it anyway because
- it's faster, most of the time
- it's more memory (and possibly cpu) friendly

but yeah i'll just leave this open in case anyone ever wanted to come back to this

almostSouji:
#4934 has been merged. If you have improvements to make to the other approach please open a separate PR building from the latest default branch commit.

Issue #4975, opened 2020-11-03
guildMemberUpdate does not utilise partials


MattIPv4:
**Please describe the problem you are having in as much detail as possible:**

With both the `USER` and `GUILD_MEMBER` partial enabled, with the `GUILD_MEMEBRS` intent, the `guildMemberUpdate` event does not fire when an uncached user/member changes their nickname. However, it does fire for a fully cached user/member.

I would expect `guildMemberUpdate` to be able to utilise partials, with only IDs plus the updated value being passed in the objects -- I haven't looked into it much though, so I could be wrong if Discord doesn't send any useful data about the change.

https://github.com/discordjs/discord.js/blob/b8fd3f65d964f080afb42dd37eb21a13c0d84f8f/src/client/websocket/handlers/GUILD_MEMBER_UPDATE.js

**Include a reproducible code sample here, if possible:**

```js
    const client = new Client({
        ws: {
            intents: (new Intents()).add(
                Intents.FLAGS.GUILDS,
                Intents.FLAGS.GUILD_MEMBERS,
            ),
        },
        partials: ['USER', 'GUILD_MEMBER'],
    });
    client.on('guildMemberUpdate', (oldMember, newMember) => console.log(oldMember, newMember));
```

Update the nickname of another user in a guild that the bot is in, without that user sending any message so they are uncached. Observe no logging.

Update the nickname of yourself after having sent a message in the guild. Observe change logged.

**Further details:**

- discord.js version: 12.4.1
- Node.js version: 12.18.1
- Operating system: macOS 10.15.7
- Priority this issue should have – please be realistic and elaborate if possible: P2 - Would be useful considering partials help with memory performance greatly.

**Relevant client options:**

- partials: USER, GUILD_MEMBER
- gateway intents: GUILDS, GUILD_MEMEBRS
- other: none

- [ ] I have also tested the issue on latest master, commit hash:

PR #4986, opened 2020-11-07
fix(GuildMemberUpdate): cache incoming members & use partials if enabled

SpaceEEC:

This PR ensures that the incoming member (update) data is always cached, even if the member was not cached.

If using partials, the guildMemberUpdate event will be emitted with a partial 'old' member.
If not using partials, the guildMemberAvailable event will be emitted.

I also added the guildMemberAvailable event doc string back, which was erroneously removed from the documentation some time ago. (See: f3cad81f5314a3997ac188277e2471ba84cbd771 )

**This is not tested yet.**

**Status**

### DIFF
--- a/src/client/actions/GuildMemberUpdate.js
+++ b/src/client/actions/GuildMemberUpdate.js
@@ -1,0 +1,44 @@
+'use strict';
+
+const Action = require('./Action');
+const { Status, Events } = require('../../util/Constants');
+
+class GuildMemberUpdate extends Action {
+  handle(data, shard) {
+    const { client } = this;
+    const user = client.users.cache.get(data.user.id);
+    if (data.user.username) {
+      if (!user) {
+        client.users.add(data.user);
+      } else if (!user.equals(data.user)) {
+        client.actions.UserUpdate.handle(data.user);
+      }
+    }
+
+    const guild = client.guilds.cache.get(data.guild_id);
+    if (guild) {
+      const member = this.getMember({ user: data.user }, guild);
+      if (member) {
+        const old = member._update(data);
+        /**
+         * Emitted whenever a guild member changes - i.e. new role, removed role, nickname.
+         * Also emitted when the user's details (e.g. username) change.
+         * @event Client#guildMemberUpdate
+         * @param {GuildMember} oldMember The member before the update
+         * @param {GuildMember} newMember The member after the update
+         */
+        if (shard.status === Status.READY) client.emit(Events.GUILD_MEMBER_UPDATE, old, member);
+      } else {
+        const newMember = guild.members.add(data);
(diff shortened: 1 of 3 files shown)


src/client/actions/GuildMemberUpdate.js
```diff
@@ -0,0 +1,44 @@
+'use strict';
+
+const Action = require('./Action');
+const { Status, Events } = require('../../util/Constants');
+
+class GuildMemberUpdate extends Action {
+  handle(data, shard) {
+    const { client } = this;
+    const user = client.users.cache.get(data.user.id);
+    if (data.user.username) {
```
kyranet, line 10: `user` is used only inside the condition:
```suggestion
    if (data.user.username) {
      const user = client.users.cache.get(data.user.id);
```

kyranet (changes requested):

src/client/actions/GuildMemberUpdate.js
```diff
@@ -0,0 +1,44 @@
+'use strict';
+
+const Action = require('./Action');
+const { Status, Events } = require('../../util/Constants');
+
+class GuildMemberUpdate extends Action {
```
NotSugden, line 6:
```suggestion
class GuildMemberUpdateAction extends Action {
```
consistency with the other action classes

src/client/actions/GuildMemberUpdate.js
```diff
@@ -0,0 +1,44 @@
...
+         * @event Client#guildMemberAvailable
+         * @param {GuildMember} member The member that became available
+         */
+        this.client.emit(Events.GUILD_MEMBER_AVAILABLE, newMember);
+      }
+    }
+  }
+}
+
+module.exports = GuildMemberUpdate;
```
NotSugden, line 44:
```suggestion
module.exports = GuildMemberUpdateAction;
```

NotSugden (changes requested):

MattIPv4:
cc #4975

kyranet (approved):

src/client/actions/GuildMemberUpdate.js
```diff
@@ -0,0 +1,44 @@
+'use strict';
+
+const Action = require('./Action');
+const { Status, Events } = require('../../util/Constants');
+
+class GuildMemberUpdateAction extends Action {
```
vladfrangu, line 6: Any...specific reason this was turned into an action?

SpaceEEC: To be able to use `Action#getMember` (on line 20)

vladfrangu (approved):

iCrawl (approved):

PR #4636, opened 2020-07-16
fix(Partials): correctly set properties as nullable

NotSugden:
**Please describe the changes this PR makes and why it should be merged:**
This PR changes some properties to be explicitly set to null, as they are supposed to be, from structures that can be partial

There is also a change from truthy checks to `'x' in data` in the _patch methods of said structures
**Status**

- [x] Code changes have been tested against the Discord API, or there are no code changes
- [x] I know how to update typings and have done so, or typings don't need updating
  - I've done a brief check, but it seems to all be covered by the `Partial*` interfaces

**Semantic versioning classification:**

- [ ] This PR changes the library's interface (methods or parameters added)
  - [ ] This PR includes breaking changes (methods removed or renamed, parameters moved or removed)
- [ ] This PR **only** includes non-code changes, like changes to documentation, README, etc.

PR #4990, opened 2020-11-08
fix(User): only assign to bot initially or if info is actually present

### LINKED ISSUE #4943: fetchAuditLogs() showing bot as false when should be true
On 10/21 I noticed my role logs showing bot entries when I was using if (executor.bot) return;

Iv'e been using the same logs for months and never had this issue. Did some logging and noticed 
GuildAuditLogsEntry {
  targetType: 'USER',
  actionType: 'UPDATE',
  action: 'MEMBER_ROLE_UPDATE',
  reason: 'Reaction Roles - User added or removed their reaction.',
  executor: ClientUser {
    id: '696870234262339614',
    username: 'Sx Bot',
    bot: false,
    discriminator: '8364',
    avatar: '8d9248ee2749a9dcef67ef9432535294',
    flags: UserFlags { bitfield: 65536 },
    lastMessageID: null,
    lastMessageChannelID: null,
    verified: true,
    mfaEnabled: true,
  },
Bot is showing as false when in fact it's true. I tried the fetch on another bot, same issue 
executor: <ref *2> KlasaUser {
    id: '696870234262339614',
    system: false,
    locale: null,
    flags: UserFlags { bitfield: 65536 },
    username: 'Sx Bot',
    bot: false,
    discriminator: '8364',
    avatar: '8d9248ee2749a9dcef67ef9432535294',
    lastMessageID: null,
    lastMessageChannelID: null,

I tested by doing a RAW api request, and it shows true as expected 
 {
      id: '696870234262339614',
      username: 'Sx Bot',
      avatar: '8d9248ee2749a9dcef67ef9432535294',
      discriminator: '8364',
      public_flags: 65536,
      bot: true
    }

### LINKED ISSUE #4985: User#bot erroneously changing into false when fetching audit logs when using user partial

I'm using `messageReactionAdd` event and I'm fetching some entries from the audit log, If it happened to add a reaction at the same time that an Audit log fetching process happening, it masses up something I'm not sure what it is but it loses track if the user is a bot or not in messageReactionAdd event

the fetch code something like 
```js
var fetchedLogs = await newState.guild.fetchAuditLogs({
                limit: 1,
            });
            const Logs = fetchedLogs.entries.first();
```

**note:** my bot add reaction when it starts up

```js
//without fetching
ClientUser {
  id: '594320405523857448',
  username: 'Morino',
  bot: true, // <-- pay attention here
  discriminator: '4849',
  avatar: '84a30bdb5c4b1878a3266354a586707b',
  lastMessageID: '774506721787379793',
  lastMessageChannelID: '771458315795300383',
  verified: true,
  mfaEnabled: true,
  _typing: Map {},
  data: {
    language: 'ar-MA',
    playlist: [],
    _id: 5f9b12e2a354594975f883de,
    id: '594320405523857448',
    __v: 0
  },
  flags: UserFlags { bitfield: 0 }
}

// with fetching
ClientUser {
  id: '594320405523857448',
  username: 'Morino',
  bot: false,  // <-- pay attention here
  discriminator: '4849',
  avatar: '84a30bdb5c4b1878a3266354a586707b',
  lastMessageID: null,
  lastMessageChannelID: null,
  verified: true,
  mfaEnabled: true,
  _typing: Map {},
  data: {
    language: 'ar-MA',
    playlist: [],
    _id: 5f9b12e2a354594975f883de,
    id: '594320405523857448',
    __v: 0
  },
  flags: UserFlags { bitfield: 0 }
}
```

**Further details:**

- discord.js version: 12.3.1
- Node.js version: 14.14.0
- Operating system: Windows 10 pro N version 2004 build 19041.572
- Priority this issue should have – please be realistic and elaborate if possible: not so important

**Relevant client options:**

- partials: 'CHANNEL', 'GUILD_MEMBER', 'MESSAGE', 'REACTION', 'USER'
- gateway intents: none
- other: none

- [x] I have also tested the issue on the latest master, commit hash: 7ec0bd9

SpaceEEC:

Fixes #4985 and fixes #4943.

This PR fixes the bot property of a bot to erroneously changing into `false` when the USER partial was enabled and their partial is requested through something like `Action#getUser` or explicitly through `UserStore#add` with just an `id` property.

This is solved by only initially setting the property, or updating it if the payload actually includes a `bot` property

**Status**

### DIFF
--- a/src/structures/User.js
+++ b/src/structures/User.js
@@ -45,11 +45,13 @@ class User extends Base {
       this.username = null;
     }
 
-    /**
-     * Whether or not the user is a bot
-     * @type {boolean}
-     */
-    this.bot = Boolean(data.bot);
+    if ('bot' in data || typeof this.bot !== 'boolean') {
+      /**
+       * Whether or not the user is a bot
+       * @type {boolean}
+       */
+      this.bot = Boolean(data.bot);
+    }
 
     if ('discriminator' in data) {
       /**


src/structures/User.js
```diff
@@ -45,11 +45,13 @@ class User extends Base {
...
-     * @type {boolean}
-     */
-    this.bot = Boolean(data.bot);
+    if ('bot' in data || typeof this.bot !== 'boolean') {
+      /**
+       * Whether or not the user is a bot
+       * @type {boolean}
+       */
+      this.bot = Boolean(data.bot);
+    }
```
vladfrangu, line 54: Needs an `else this.bot = null`

vladfrangu (changes requested):

src/structures/User.js
SpaceEEC: That would cause the prop to erroneously become `null` instead of `false` then.
Not overwriting `bot` if we have nothing to assign in an update is the intention with this PR.

vladfrangu: It wouldn't cause the prop to erroneously become `null`? It's literally what the properties above and below this one do...

SpaceEEC: >It wouldn't cause the prop to erroneously become `null`?

It would, see below.

> It's literally what the properties above and below this one do...

I do not understand that part, if there is an issue with something, please elaborate.

---

The below part:

Adding that `else ...` would cause this behavior:
```js
// Partial user payload
const data = { id: '123' };

// user props
const that = { bot: true, id: '123', username: 'foo', discriminator: '0815' };

if ('bot' in data || typeof that.bot !== 'boolean') {
    that.bot = Boolean(data.bot);
} else that.bot = null;

console.log(that); // { bot: null /* ... other props, unchanged */ } <- should have stayed true
```
Patching a user with partial data, which is what this issue is about, would cause the property to become `null` while it should not be touched at all.

vladfrangu: Because your code doesn't do the same thing we already to for `username`, `avatar`, `discriminator`, etc...
```js
const data = { id: '123' };

// user props
const that = { bot: true, id: '123', username: 'foo', discriminator: '0815' };

if (Reflect.has(data, 'bot')) {
	that.bot = Boolean(data.bot);
} else if (typeof that.bot !== 'boolean') {
	that.bot = null;
}
```

NotSugden: in this case, `bot` will only be in the `data` if it is true, so (as attempted in #4636) `User#bot` will be `null` if the user is not a bot, so we have to assume false if its not present

vladfrangu: So Discord really doesn't send `bot: false` for normal users... *great* :/
Ignore me then

iCrawl (approved):

vladfrangu (approved):

advaith1:
Shouldn't this also be done for User#system?

PR #4991, opened 2020-11-08
feat(Shard): shard-specific broadcastEval/fetchClientValues + shard Id util

MattIPv4:

Adds the ability to call `broadcastEval` and `fetchClientValues` from a shard to run on different target shard, instead of broadcasting to all and using an `if` in the given script to filter by shard.

Now you can target a shard directly:

```javascript
const { ShardClientUtil } = require('discord.js');
const guild = '433980600391696384';

// Get the shard Id -- `client.shard.constructor.shardIdForGuildId` could also be used
const shard = ShardClientUtil.shardIdForGuildId(guild, 2);

// Get the guild from that shard only
const name = await client.shard.broadcastEval(`this.guilds.cache.get('${guild}').name`, shard);

return [name, shard, client.shardId];
```

We get back `[JetBrains Community, 1, 0]`, so we can see it correctly went to the target shard and found the guild there.

**Status**

### DIFF
--- a/src/sharding/ShardClientUtil.js
+++ b/src/sharding/ShardClientUtil.js
@@ -228,14 +228,14 @@ class ShardClientUtil {
   }
 
   /**
-   * Get the shard Id for a given guild Id.
-   * @param {Snowflake} guildId Snowflake guild Id to get shard Id for
+   * Get the shard ID for a given guild ID.
+   * @param {Snowflake} guildID Snowflake guild ID to get shard ID for
    * @param {number} shardCount Number of shards
    * @returns {number}
    */
-  static shardIdForGuildId(guildId, shardCount) {
-    const shard = Number(BigInt(guildId) >> 22n) % shardCount;
-    if (shard < 0) throw new Error('SHARDING_SHARD_MISCALCULATION', shard, guildId, shardCount);
+  static shardIDForGuildID(guildID, shardCount) {
+    const shard = Number(BigInt(guildID) >> 22n) % shardCount;
+    if (shard < 0) throw new Error('SHARDING_SHARD_MISCALCULATION', shard, guildID, shardCount);
     return shard;
   }
 }
--- a/src/sharding/ShardClientUtil.js
+++ b/src/sharding/ShardClientUtil.js
@@ -96,8 +96,9 @@ class ShardClientUtil {
   }
 
   /**
-   * Fetches a client property value of each shard.
+   * Fetches a client property value of each shard, or a given shard.
    * @param {string} prop Name of the client property to get, using periods for nesting
+   * @param {number} [shard] Shard to fetch property from, all if undefined
    * @returns {Promise<Array<*>>}
    * @example
    * client.shard.fetchClientValues('guilds.cache.size')
@@ -105,19 +106,19 @@ class ShardClientUtil {
    *   .catch(console.error);
    * @see {@link ShardingManager#fetchClientValues}
    */
-  fetchClientValues(prop) {
+  fetchClientValues(prop, shard) {
     return new Promise((resolve, reject) => {
       const parent = this.parentPort || process;
 
       const listener = message => {
-        if (!message || message._sFetchProp !== prop) return;
+        if (!message || message._sFetchProp !== prop || message._sFetchPropShard !== shard) return;
         parent.removeListener('message', listener);
         if (!message._error) resolve(message._result);
         else reject(Util.makeError(message._error));
       };
       parent.on('message', listener);
 
-      this.send({ _sFetchProp: prop }).catch(err => {
+      this.send({ _sFetchProp: prop, _sFetchPropShard: shard }).catch(err => {
         parent.removeListener('message', listener);
         reject(err);
       });
--- a/src/sharding/ShardClientUtil.js
+++ b/src/sharding/ShardClientUtil.js
@@ -125,8 +125,9 @@ class ShardClientUtil {
   }
 
   /**
-   * Evaluates a script or function on all shards, in the context of the {@link Client}s.
+   * Evaluates a script or function on all shards, or a given shard, in the context of the {@link Client}s.
    * @param {string|Function} script JavaScript to run on each shard
+   * @param {number} [shard] Shard to run script on, all if undefined
    * @returns {Promise<Array<*>>} Results of the script execution
    * @example
    * client.shard.broadcastEval('this.guilds.cache.size')
@@ -134,20 +135,20 @@ class ShardClientUtil {
    *   .catch(console.error);
    * @see {@link ShardingManager#broadcastEval}
    */
-  broadcastEval(script) {
+  broadcastEval(script, shard) {
     return new Promise((resolve, reject) => {
       const parent = this.parentPort || process;
       script = typeof script === 'function' ? `(${script})(this)` : script;
 
       const listener = message => {
-        if (!message || message._sEval !== script) return;
+        if (!message || message._sEval !== script || message._sEvalShard !== shard) return;
         parent.removeListener('message', listener);
         if (!message._error) resolve(message._result);
         else reject(Util.makeError(message._error));
       };
       parent.on('message', listener);
 
-      this.send({ _sEval: script }).catch(err => {
+      this.send({ _sEval: script, _sEvalShard: shard }).catch(err => {
         parent.removeListener('message', listener);
         reject(err);
       });
--- a/src/sharding/ShardClientUtil.js
+++ b/src/sharding/ShardClientUtil.js
@@ -100,7 +100,7 @@ class ShardClientUtil {
    * Fetches a client property value of each shard, or a given shard.
    * @param {string} prop Name of the client property to get, using periods for nesting
    * @param {number} [shard] Shard to fetch property from, all if undefined
-   * @returns {Promise<Array<*>>}
+   * @returns {Promise<*>|Promise<Array<*>>}
    * @example
    * client.shard.fetchClientValues('guilds.cache.size')
    *   .then(results => console.log(`${results.reduce((prev, val) => prev + val, 0)} total guilds`))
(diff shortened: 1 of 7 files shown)


src/util/Snowflake.js
```diff
@@ -79,6 +79,14 @@ class SnowflakeUtil {
     });
     return res;
   }
+
+  /**
+   * Discord's epoch value (2015-01-01T00:00:00.000Z).
+   * @returns {number}
```
NotSugden, line 85:
```suggestion
   * @type {number}
   * @readonly
```

NotSugden (changes requested):

src/sharding/ShardingManager.js
```diff
@@ -222,30 +222,48 @@ class ShardingManager extends EventEmitter {
...
+
+  /**
+   * Runs a method with given arguments on all shards, or a given shard.
+   * @param {string} method Method name to run on each shard
+   * @param {Array<*>} args Arguments to pass through to the method call
+   * @param {number} [shard] Shard to run on, all if undefined
+   * @returns {Promise<Array<*>>} Results of the method execution
+   * @private
+   */
+  _performOnShards(method, args, shard) {
```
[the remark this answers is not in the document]
MattIPv4: `shard` is currently an optional arg, hence it is at the end -- do you think it makes sense to make it required (to support this change), and `undefined` can be passed explicitly when needed?

src/sharding/ShardingManager.js
```diff
@@ -222,30 +222,48 @@ class ShardingManager extends EventEmitter {
...
     if (this.shards.size !== this.shardList.length) return Promise.reject(new Error('SHARDING_IN_PROCESS'));
+
+    if (shard !== undefined) {
+      if (this.shards.has(shard)) return Promise.all([this.shards.get(shard)[method](...args)]);
+      return Promise.reject(new Error('SHARDING_SHARD_NOT_FOUND', shard));
+    }
+
     const promises = [];
-    for (const shard of this.shards.values()) promises.push(shard.fetchClientValue(prop));
+    for (const sh of this.shards.values()) promises.push(sh[method](...args));
```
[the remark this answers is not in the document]
MattIPv4: `shard` is a parameter in the method, so can't also be used here, which is why `sh` is used instead.

src/sharding/ShardingManager.js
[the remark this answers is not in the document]
NotSugden: oh, mb

src/sharding/ShardingManager.js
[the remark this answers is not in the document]
NotSugden: mb

src/sharding/ShardClientUtil.js
```diff
@@ -224,6 +227,18 @@ class ShardClientUtil {
...
+   * Get the shard Id for a given guild Id.
+   * @param {Snowflake} guildId Snowflake guild Id to get shard Id for
+   * @param {number} shardCount Number of shards
+   * @returns {number}
+   */
+  static shardIdForGuildId(guildId, shardCount) {
+    // This performs (guild_id >> 22) % num_shards, but with a string Snowflake
+    const timestamp = Snowflake.deconstruct(guildId).timestamp;
+    return (timestamp - Snowflake.epoch) % shardCount;
+  }
```
MattIPv4, line 241: @pizzafox @advaith1 lmk if this implementation makes sense based on what y'all had done

pizzafox: There should be validation here. On mobile, will post what I do in a second.

pizzafox:
```js
static shardIdForGuildId(guildId, shardCount) {
  // This performs (guild_id >> 22) % num_shards, but with a string Snowflake
  const timestamp = Snowflake.deconstruct(guildId).timestamp;
  
  const shard = (timestamp - Snowflake.epoch) % shardCount;
  
  if (calculated >= 0 && calculated < shardCount) {
    return shard;
  }

  throw new Error(`shardCount is incorrect or guildId ${guildID} is invalid, calculated ${calculated}`);
}
```
Something like this, but with the funny errors file Discord.js uses.

pizzafox (changes requested):
Code probably works but I have a few minor suggestions with the `shardIdForGuildId` function.

src/sharding/ShardClientUtil.js
tipakA: Huh? All snowflakes are bigger than MAX_SAFE_INTEGER. Only reason you could not notice is because your older ID ends with 0

MattIPv4: `Snowflake#deconstruct` uses strings for the full snowflake + binary representation, only casting to integers to return the resultant object.

pizzafox: I'm probably misremembering what specifically happened, but it was due to some number being too precise for a double. Wrapping everything in `BigInt` fixed it.

kyranet: It's fine to use this code, we require Node.js v12, and according to [node.green](https://node.green#ES2020-features-BigInt), it's available since v10.8.0.
```javascript
const shard = Number(BigInt(guildId) >> 22n) % shardCount;
```

NotSugden: *keep in mind you will need to change the `ecmaVersion` to `2020` in `.eslintrc.json` if you choose to use the `BigInt(guildId)` approach*

MattIPv4: 👍 Using the BigInt option seems like a nice, clean option. It is the first use of it in the codebase though, so yeah, ecma needs bumping :)

kyranet: > _keep in mind you will need to change the `ecmaVersion` to `2020` in `.eslintrc.json` if you choose to use the `BigInt(guildId)` approach_

It's fine, Node.js v12 runs ECMAScript 2020, which also includes `BigInt`. Besides, this wouldn't be a breaking change at all, just something that should have been done a long time ago.

src/sharding/ShardClientUtil.js
```diff
@@ -224,6 +226,18 @@ class ShardClientUtil {
...
+
+  /**
+   * Get the shard Id for a given guild Id.
+   * @param {Snowflake} guildId Snowflake guild Id to get shard Id for
+   * @param {number} shardCount Number of shards
+   * @returns {number}
+   */
+  static shardIdForGuildId(guildId, shardCount) {
+    const shard = Number(BigInt(guildId) >> 22n) % shardCount;
+    if (shard < 0 || shard >= shardCount) throw new Error('SHARDING_SHARD_MISCALCULATION', guildId, shardCount, shard);
```
kyranet, line 238: Since `shard` is calculated with the modulus of `shardCount`, it can never be higher or equals than it, therefore only negative numbers would trigger this condition:

```suggestion
    if (shard < 0) throw new Error('SHARDING_SHARD_MISCALCULATION', guildId, shardCount, shard);
```

kyranet (changes requested):

src/sharding/ShardClientUtil.js
MattIPv4: Ah, yes

src/sharding/ShardClientUtil.js
NotSugden: yeah, just noting that it needs to be done

src/sharding/ShardClientUtil.js
```diff
@@ -235,9 +234,9 @@ class ShardClientUtil {
    * @returns {number}
    */
   static shardIdForGuildId(guildId, shardCount) {
-    // This performs (guild_id >> 22) % num_shards, but with a string Snowflake
-    const timestamp = Snowflake.deconstruct(guildId).timestamp;
-    return (timestamp - Snowflake.epoch) % shardCount;
+    const shard = Number(BigInt(guildId) >> 22n) % shardCount;
+    if (shard < 0) throw new Error('SHARDING_SHARD_MISCALCULATION', guildId, shardCount, shard);
```
NotSugden, line 238:
```suggestion
    if (shard < 0) throw new Error('SHARDING_SHARD_MISCALCULATION', shard, guildId, shardCount);
```
match parameter order for the error fn

NotSugden (changes requested):

src/sharding/ShardClientUtil.js
MattIPv4: im stupid, ty

pizzafox (approved):
New edits look good.

src/sharding/ShardClientUtil.js
```diff
@@ -224,6 +226,18 @@ class ShardClientUtil {
     }
     return this._singleton;
   }
+
+  /**
+   * Get the shard Id for a given guild Id.
+   * @param {Snowflake} guildId Snowflake guild Id to get shard Id for
```
SpaceEEC, line 232:
```suggestion
   * Get the shard ID for a given guild ID.
   * @param {Snowflake} guildId Snowflake guild ID to get shard ID for
```

SpaceEEC (changes requested):
Otherwise LGTM

src/util/Snowflake.js
```diff
@@ -79,6 +79,15 @@ class SnowflakeUtil {
     });
     return res;
   }
+
+  /**
+   * Discord's epoch value (2015-01-01T00:00:00.000Z).
+   * @type {number}
+   * @readonly
+   */
+  static get epoch() {
```
Gawdl3y, line 88: Should probably be EPOCH since it's a static constant.

Gawdl3y (dismissed):

vladfrangu (changes requested):

vladfrangu (approved):

kyranet (approved):

iCrawl (approved):

SpaceEEC (approved):

iCrawl (approved):

PR #4992, opened 2020-11-08
test(disableMention): use switch statement for commands

pizzafox:

Massively increase performance of the disasbleMention test file.
This is good and also very necessary.

**Status**

### DIFF
--- a/test/disableMentions.js
+++ b/test/disableMentions.js
@@ -41,7 +41,7 @@ client.on('message', message => {
       message.reply(tests[0]);
       break;
     case 'test2':
-      message.reply(tests[0]);
+      message.reply(tests[1]);
       break;
     case 'test3':
       message.reply(tests[2]);
--- a/test/disableMentions.js
+++ b/test/disableMentions.js
@@ -36,9 +36,19 @@ client.on('message', message => {
   // Clean content and log each character
   console.log(Util.cleanContent(args.join(' '), message).split(''));
 
-  if (command === 'test1') message.reply(tests[0]);
-  else if (command === 'test2') message.reply(tests[1]);
-  else if (command === 'test3') message.reply(tests[2]);
+  switch (command) {
+    case 'test1':
+      message.reply(tests[0]);
+      break;
+    case 'test2':
+      message.reply(tests[0]);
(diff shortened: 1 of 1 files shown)


pizzafox:
Why are you reacting with 😕 this is an important PR

tipakA:
Please leave the shitposts out of the repositories. Hactoberfest ended over a week ago, and there is no major difference between switch and if/else if,

pizzafox:
I am well aware Hacktoberfest is over. This change could influence developers who are still learning to use `switch` statements which perform better and are great for solving problems like these.

kyranet:
It's an "informal" test file made quickly to test stuff, nothing that needs to fancy, and definitely not something users read.

Besides, users should use a proper command handler or a framework, not a large switch case.

pizzafox:
devsnek: switch statements can't be O(1) unless the entire domain of values is covered
but either way I would merge that, it's more readable

kyranet:
Readability is subjective. Don't get me wrong, I use `switch`, but I do so very rarely (normally inside helper functions that always `return`). By itself, `switch` comes with a few readability drawbacks:

- It uses a lot of keywords (`switch`, `case:`, `break`/`return`).
- It has a lot of grammar going on, lint rules require braces to be opened (`case: { ... }`) when at least one assignment is made, this creates certain inconsistency and makes the `break;` feel out of place.
- Limitations and even hard-to-predict behaviour inside loops - does a `continue;` or a `break;` inside a loop affect the `switch`, or does it affect the loop itself? How do I enter and exit from a switch in a non-`default` case inside a loop and have it run the code below?
- It increases indentation and makes the code bulkier as it gets very expanded.

And as devsnek said, it's not `O(1)` unless all cases are covered. It is not in this case, therefore it's `O(n)` (where `n` is the amount of cases), sharing the same costs as regular `if` and `if else` chains. Therefore, there's no performance improvement here. And if there is, it'll be noticeable after hundreds of millions of iterations.

monbrey:
`disableMentions` is a candidate for removal anyway, as Discord's `allowed_mentions` API parameter supersedes it in pretty much every way.

Issue #4916, opened 2020-10-19
Message#delete using timeout doesn't verify if the message is already deleted


ala89:
The issue is quite simple, running Message#delete passing a timeout as an option simply runs Message#delete on the said message `timeout` ms later, without checking if the message was already deleted or not. A check should be added to avoid API errors.

abhijith94:
You could just handle the error in the catch block and see if the httpStatus returned is 404 or not.

```
client.on("message", (msg) => {
  if (msg.content == "delete this") {
    msg.delete({ timeout: 10000 }).catch((error) => {
      console.log(error);
    });
  }
});
```

documentation - https://discord.js.org/#/docs/main/stable/class/Message?scrollTo=delete

ala89:
Yeah but this is a monkey patching solution, we don't want to send useless requests to the API

monbrey:
The library should not be responsible for performing checks on the user's behalf.

If this is a concern for you, you can set your own timeout for deleting the message and perform checks/catch errors accordingly.

ElectrifyPro:
The solution for this issue is very simple to write, and there isn't any way to manually disable the timeout that Message#delete creates, so perhaps we should create this enhancement as a convenience. What are your thoughts?

monbrey:
My thoughts are written above

> The library should not be responsible for performing checks on the user's behalf.

The way to "disable" the timeout is not to set one - it's effectively nothing but a glorified helper option for setting your own setTimeout anyway.
```js
setTimeout(() => {
  if(!message.deleted) message.delete()
}, 5000);
```

NotSugden:
you could also extend the `Message` class using [Structures.extend](https://discord.js.org/#/docs/main/stable/class/Structures?scrollTo=s-extend) if you wanted to, this would probably be the best soloution

ala89:
I did extend the structure but I believe most users would except such check. Beginners who set a timeout on a message in a channel that ends up being deleted will never understand where that error is from. But I'll close it if you believe it's not necessary.

almostSouji:
In my personal opinion this points more towards removal of the option, leaving it entirely to the developer how to implement the timeout, with whatever checks and other functionalities in the callback as they deem necessary for their specific use case. We don't provide this on any other structures delete call, just on Message, where it was deemed useful before, due to the high volume of people asking for delays.

[...]

PR #4999, opened 2020-11-09
cleanup(Message): remove options for Message#delete

Vaporox:

Removes options for `Message#delete` (and `MessageManager#delete`):
- `timeout` because that implementation is prone to errors (see e.g. #4916) and inconsistent with others, being the only method which takes a timeout value
- `reason` because messages deleted by bot users won't create audit log entries

**Status**

### DIFF
--- a/src/managers/MessageManager.js
+++ b/src/managers/MessageManager.js
@@ -116,14 +116,13 @@ class MessageManager extends BaseManager {
   /**
    * Deletes a message, even if it's not cached.
    * @param {MessageResolvable} message The message to delete
-   * @param {string} [reason] Reason for deleting this message, if it does not belong to the client user
    * @returns {Promise<void>}
    */
-  async delete(message, reason) {
+  async delete(message) {
     message = this.resolveID(message);
     if (!message) throw new TypeError('INVALID_TYPE', 'message', 'MessageResolvable');
 
-    await this.client.api.channels(this.channel.id).messages(message).delete({ reason });
+    await this.client.api.channels(this.channel.id).messages(message).delete();
   }
 
   async _fetchId(messageID, cache, force) {
--- a/src/structures/Message.js
+++ b/src/structures/Message.js
@@ -563,28 +563,16 @@ class Message extends Base {
 
   /**
    * Deletes the message.
-   * @param {Object} [options] Options
-   * @param {number} [options.timeout=0] How long to wait to delete the message in milliseconds
-   * @param {string} [options.reason] Reason for deleting this message, if it does not belong to the client user
    * @returns {Promise<Message>}
    * @example
    * // Delete a message
-   * message.delete({ timeout: 5000 })
-   *   .then(msg => console.log(`Deleted message from ${msg.author.username} after 5 seconds`))
+   * message.delete()
+   *   .then(msg => console.log(`Deleted message from ${msg.author.username}`))
    *   .catch(console.error);
    */
-  delete(options = {}) {
-    if (typeof options !== 'object') return Promise.reject(new TypeError('INVALID_TYPE', 'options', 'object', true));
-    const { timeout = 0, reason } = options;
-    if (timeout <= 0) {
-      return this.channel.messages.delete(this.id, reason).then(() => this);
-    } else {
-      return new Promise(resolve => {
-        this.client.setTimeout(() => {
-          resolve(this.delete({ reason }));
-        }, timeout);
-      });
-    }
+  async delete() {
+    await this.channel.messages.delete(this.id);
+    return this;
   }
 
   /**
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -1017,7 +1017,7 @@ declare module 'discord.js' {
       options?: AwaitReactionsOptions,
     ): Promise<Collection<Snowflake, MessageReaction>>;
     public createReactionCollector(filter: CollectorFilter, options?: ReactionCollectorOptions): ReactionCollector;
-    public delete(options?: { timeout?: number; reason?: string }): Promise<Message>;
+    public delete(): Promise<Message>;
     public edit(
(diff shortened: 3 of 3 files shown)


anandre:
If the API supports providing a reason for deletion, I say it should be kept in, even if it currently doesn't show up - what if Discord does sort it out and then bot deletions do show up?  We'd need another PR to revert this change.  AFAIK, we keep full API coverage in other spots where it likewise doesn't actually do anything, this should be the same.  Maybe put a note about it on the description about how it currently does nothing, though.

advaith1:
The reason option just uses the standard header for most endpoints, there is no API reason support specific to this endpoint

Vaporox:
What other spots are you thinking of exactly? I can see what you mean, but in my opinion a note about it isn't a better alternative – both need another PR as soon as Discord decides to add logs for that.

iCrawl (approved):

ImRodry:
I don't understand why this pr was approved. The timeout option can generate as many errors as the setTimeout function, since it does essentially the same, and this is the user's own problem since they are the ones responsible for adding a check to see if the message has already been deleted (or if it's delectable)
Although messages deleted by bots are not logged anymore, this might be fixed like @anandre said previously and, if so, it would have to be added again.

anandre:
I only suggested to leave the `reason` parameter.  I'm in favor of removing `timeout` for it to be in line with all of the other `delete` methods, none of which take a `timeout` parameter.

ImRodry:
> I only suggested to leave the `reason` parameter. I'm in favor of removing `timeout` for it to be in line with all of the other `delete` methods, none of which take a `timeout` parameter.

but the reason they provided doesn't make much sense to me. Having the timeout option takes less lines than a setTimeout function and is already implemented, so why remove it? I use it in my code and I would have to change everything if this were to be added

kyranet:
> > I only suggested to leave the `reason` parameter. I'm in favor of removing `timeout` for it to be in line with all of the other `delete` methods, none of which take a `timeout` parameter.
> 
> but the reason they provided doesn't make much sense to me. Having the timeout option takes less lines than a setTimeout function and is already implemented, so why remove it? I use it in my code and I would have to change everything if this were to be added

Part of the reason why comes from your previous message:

> I don't understand why these changes were approved. The timeout option can generate as many errors as the setTimeout function, since it does essentially the same, and this is the user's own problem since they are the ones responsible for adding a check to see if the message has already been deleted (or if it's deletable)

All discord.js methods go to the REST/WS handler immediately, `Message#delete` is the exception, why should we have this inconsistency?

Besides, you say that yourself, the user is responsible for adding a check when a message has already been deleted - problem is, you may do `await message.delete({ timeout: 2000 });` after an `if (message.deletable) { ... }` check (which also accounts for `message.deleted`), and somebody deletes the message earlier, the message has been deleted, but there is no way to stop the timeout in discord.js's code, and you'll get a DiscordAPIError.

Besides, most of the time, when we want to delete a message, we want to delete it right away. Also, consider doing `const { setTimeout } = require('timers/promises');` ([docs](https://nodejs.org/dist/latest-v15.x/docs/api/timers.html#timers_timerspromises_settimeout_delay_value_options)) or `util.promisify(setTimeout)` ([docs](https://nodejs.org/dist/latest-v15.x/docs/api/util.html#util_util_promisify_original)) for simplicity, as you can do:

```js
await sleep(5000); // sleep being setTimeout from timers/promises, or promisify(setTimeout)
if (!message.deleted) await message.delete();
```

ImRodry:
Fair enough @kyranet 
Putting that aside though, why is reason being removed despite all the previous feedback?

tipakA:
Cause it appears that API doesn't take the reason here.

marzeq:
> Cause it appears that API doesn't take the reason here.

It still does, it's just not shown in the audit logs.

kyranet:
> Putting that aside though, why is reason being removed despite all the previous feedback?

Just FYI, the fate of discord.js is not decided by a single person, but by the development team. Crawl decided to approve the removal of the reason field because Discord did not support it for 2 years, I presume they're filtered out because moderation bots can spam the audit logs quite fast, just like they have a lot higher ratelimit when deleting new messages (a separate one from deleting messages that are, per say, a few hours or even years old).

As for me personally, I see why Discord would do that, but I prefer the reason to be kept, even if it's not shown in the clients nowadays, it might in the future.

ImRodry:
> > Putting that aside though, why is reason being removed despite all the previous feedback?
> 
> Just FYI, the fate of discord.js is not decided by a single person, but by the development team. Crawl decided to approve the removal of the reason field because Discord did not support it for 2 years, I presume they're filtered out because moderation bots can spam the audit logs quite fast, just like they have a lot higher ratelimit when deleting new messages (a separate one from deleting messages that are, per say, a few hours or even years old).
> 
> As for me personally, I see why Discord would do that, but I prefer the reason to be kept, even if it's not shown in the clients nowadays, it might in the future, at least for messages that aren't seconds old.

Thank you for this information! I do understand that discord could’ve chosen to hide it but if the API supports it, it should always be kept so that’s good to know.

kyranet:
> Thank you for this information! I do understand that discord could’ve chosen to hide it but if the API supports it, it should always be kept so that’s good to know.

Because the reason is sent as a header (`X-Audit-Log-Reason`) as documented in Discord's [docs](https://discord.com/developers/docs/resources/audit-log#audit-logs):

https://github.com/discordjs/discord.js/blob/2685b960d7bbf47d7884545ad06aec8160907342/src/rest/APIRequest.js#L39

Typically, if a header is not supported, the request doesn't fail, but rather the header is ignored. This is not the same as the JSON data we send.

However, to back keeping `reason` in `Message#delete`, there's an audit log type called `MESSAGE_DELETE` (`72`), even if it's not shown, it means no harm to keep.

SpaceEEC (approved):

PR #5020, opened 2020-11-18
Warn the user in their console if their DJS version is outdated.

Peppy8651:

Almost every day on support servers like the DJS Discord, I see people who forget or are too lazy to update. This can be fine of course, but it can take away time from other people who need it more than some people that forgot to update. Therefore, I made a package for discord.js that uses node-fetch and the package.json to determine the latest version and the current version of the module in the user's node modules when a user uses the ready event.  It fetches the latest discord.js release json from Github and fetches the version of it. The code I put inside `READY.js` checks if these two versions (from Github and the package.json) are the same. If they are not, it will warn the user that the version they are using is not the current or latest version and tell them ways to update using npm install. This would happen when the user inputs `client.on()` with the ready event into their code. I know this will probably not stop people from currently forgetting if they're on V11, however this may prevent these kinds of situations happening with newer versions. Credit goes to [redbrain](https://github.com/fent/node-ytdl-core/pull/779). They made a similar pull request for [node-ytdl-core](https://github.com/fent/node-ytdl-core/) and the idea had they made me want to do it for discord.js since this is a similar situation to what made redbrain make a pull request. 

EDIT: Examples of my console and the code I added functioning are in the examples folder. I should also add that you can turn this function off by going to the package.json and changing `outdatedwarn` to false. By doing this, you should be able to avoid any future warnings.

**Status**

- [] I know how to update typings and have done so, or typings don't need updating

 - [] This PR includes breaking changes (methods removed or renamed, parameters moved or removed)
- [] This PR **only** includes non-code changes, like changes to documentation, README, etc.


Papaia:
Why use `miniget` instead of the already-included `node-fetch`?
A c t u a l l y
Why is this the libs responsibility?

Peppy8651:
That's true. I'll change it so it fetches using node-fetch.

Peppy8651:
Ok, replaced miniget with node-fetch and tested. It still works.

redbrain:
Firstly, congratulations on your first ever PR. Open source software is driven by the community, and I'm glad to see you join it.
An update-checker was crucial for ytdl-core, because every update fixed major issues with the library.
However, the Discord API does not regularly make breaking changes, nor is it important to warn users about old D.JS versions.
Furthermore, that checker was tailored to ytdl specifically, using miniget so as to not introduce more dependencies.
For those reasons I'd advise against this specific addition to D.JS.

Peppy8651:
Yes, however as a lot people see with V11, it is unsupported and their bots break. Plus I changed the dependencies. It now only uses node-fetch.

tipakA:
If you are going to recommend just doing `npm install discord.js@latest` when someone has v11, you are going to cause way more harm than good.

There is a reason v11 is v11 and v12 is v12. Those are not hot-swappable, and updating major will require code revision.

Peppy8651:
I'm trying to prevent any further situations like these in newer versions. I can't undo what has already been done. Imagine in a year or two people are coming to the support server with problems about V12 when we have moved on to V13. Of course, I could add a link to the changelog.

Peppy8651:
Actually, I will change that. It is a pretty bad idea just to recommend npm installing.

Roki100:
i dont like this idea

BannerBomb:
I agree, I don't think discordjs needs to notify users of being outdated since there could be reasons users are outdated such as if they reverted an update temporarily while they are updating their code plus like redbrain said, discordjs doesn't have enough breaking updates to have to consider doing it. If they really wanted to do it they could just deprecate the version of the package [deprecated field](http://doc.codingdict.com/npm-ref/cli/deprecate.html) instead of making a request to the github api. By deprecating a version of the package with npm, when the user attempts to install the package it will show the warning message that was specified.

Peppy8651:
> I agree, I don't think discordjs needs to notify users of being outdated since there could be reasons users are outdated such as if they reverted an update temporarily while they are updating their code plus like redbrain said, discordjs doesn't have enough breaking updates to have to consider doing it. If they really wanted to do it they could just deprecate the version of the package [deprecated field](http://doc.codingdict.com/npm-ref/cli/deprecate.html) instead of making a request to the github api. By deprecating a version of the package with npm, when the user attempts to install the package it will show the warning message that was specified.

Users can turn this off by going into their discord.js module package.json and turning outdatedwarn to false.

BannerBomb:
> 
> 
> > I agree, I don't think discordjs needs to notify users of being outdated since there could be reasons users are outdated such as if they reverted an update temporarily while they are updating their code plus like redbrain said, discordjs doesn't have enough breaking updates to have to consider doing it. If they really wanted to do it they could just deprecate the version of the package [deprecated field](http://doc.codingdict.com/npm-ref/cli/deprecate.html) instead of making a request to the github api. By deprecating a version of the package with npm, when the user attempts to install the package it will show the warning message that was specified.
> 
> Users can turn this off by going into their discord.js module package.json and turning outdatedwarn to false.

When deprecating a package it adds a key to the package.json called `deprecated` that allows defining a custom warning message for why the package was deprecated.
image

Peppy8651:
> > > I agree, I don't think discordjs needs to notify users of being outdated since there could be reasons users are outdated

Ok, that's a good idea for really old versions, but could we maybe have a middle ground? Like maybe instead of warning in the console there would be a reminder for when Discord.js is updated. I know they already do it in the discord server, however not all users that use Discord.js are in the discord server.

vBread:
> > I agree, I don't think discordjs needs to notify users of being outdated since there could be reasons users are outdated such as if they reverted an update temporarily while they are updating their code plus like redbrain said, discordjs doesn't have enough breaking updates to have to consider doing it. If they really wanted to do it they could just deprecate the version of the package [deprecated field](http://doc.codingdict.com/npm-ref/cli/deprecate.html) instead of making a request to the github api. By deprecating a version of the package with npm, when the user attempts to install the package it will show the warning message that was specified.
> 
> Users can turn this function off by going into their discord.js module package.json and turning outdatedwarn to false.

No one's going to purposefully go out of their way to dig in their node_modules and edit the package.json. At this point, it feels like you're making excuses.

Peppy8651:
> > > I agree, I don't think discordjs needs to notify users of being outdated since there could be reasons users are outdated such as if they reverted an update temporarily while they are updating their code plus like redbrain said, discordjs doesn't have enough breaking updates to have to consider doing it. If they really wanted to do it they could just deprecate the version of the package [deprecated field](http://doc.codingdict.com/npm-ref/cli/deprecate.html) instead of making a request to the github api. By deprecating a version of the package with npm, when the user attempts to install the package it will show the warning message that was specified.
> > 
> > 
> > Users can turn this function off by going into their discord.js module package.json and turning outdatedwarn to false.
> 
> No one's going to purposefully go out of their way to dig in their node_modules and edit the package.json. At this point, it feels like you're making excuses.

Hmm, I guess I am becoming somewhat defensive and making excuses. But my point is, Discord.js could be a little clearer when they make an update so then some people can realize if an update is there or not.

tipakA:
image
image

Peppy8651:
???

almostSouji:
Re: being clearer about updates: We follow semantic [versioning](https://semver.org/) to guarantee stable uptime.
Usually that also means that you don't have to update if you don't wish to do so. However some breaking changes that discord has put off for quite some time now (namely a domain change) has caused the latter to change slightly. You will be forced to update to version 12 before the API drops support on `discordapp.com`. This requires a change in library internals which we are not willing to apply.

If you wish to introduce new features you should inform yourself about the state of the libraries you are using (GitHub or the NPM registry entry both list the releases)

Re: screenshots by tipakA:
This is internal discussion about this PR and issues that might arise if your suggestion was applied by us (and one of the reasons why it will not be applied)

---

Thank you for bringing this suggestion to our attention, but we will not be going forward with this. If we choose to go a route it will be the package deprecation on NPM.

PR #5022, opened 2020-11-18
feat(GuildChannel): support conversion between text and news 

monbrey:
Discord.js did not support providing `type` in the GuildChannel#edit payload, this adds support for that and a `setType` helper method on TextChannel and NewsChannel.

The API will throw an error if attempting to convert to anything other than Text or News.

**Status**

### DIFF
--- a/src/structures/TextChannel.js
+++ b/src/structures/TextChannel.js
@@ -87,6 +87,16 @@ class TextChannel extends GuildChannel {
   }
 
   /**
+   * Sets the type of this channel (only conversion between text and news is supported)
+   * @param {string} type The new channel type
+   * @param {string} reason Reason for changing the channel's type
+   * @returns {Promise<GuildChannel>}
+   */
+  setType(type, reason) {
+    return this.edit({ type }, reason);
+  }
+
+  /**
    * Fetches all webhooks for the channel.
    * @returns {Promise<Collection<Snowflake, Webhook>>}
    * @example
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -1188,6 +1188,7 @@ declare module 'discord.js' {
       options?: { avatar?: BufferResolvable | Base64Resolvable; reason?: string },
     ): Promise<Webhook>;
     public setNSFW(nsfw: boolean, reason?: string): Promise<NewsChannel>;
+    public setType(type: string, reason?: string): Promise<GuildChannel>;
     public fetchWebhooks(): Promise<Collection<Snowflake, Webhook>>;
     public addFollower(channel: GuildChannelResolvable, reason?: string): Promise<NewsChannel>;
   }
@@ -1529,6 +1530,7 @@ declare module 'discord.js' {
     ): Promise<Webhook>;
     public setNSFW(nsfw: boolean, reason?: string): Promise<TextChannel>;
     public setRateLimitPerUser(rateLimitPerUser: number, reason?: string): Promise<TextChannel>;
+    public setType(type: string, reason?: string): Promise<GuildChannel>;
     public fetchWebhooks(): Promise<Collection<Snowflake, Webhook>>;
   }
 
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -2232,7 +2232,10 @@ declare module 'discord.js' {
 
   interface ChannelData {
     name?: string;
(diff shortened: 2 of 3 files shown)


src/structures/TextChannel.js
```diff
@@ -86,6 +86,16 @@ class TextChannel extends GuildChannel {
     return this.edit({ nsfw }, reason);
   }
 
+  /**
+   * Sets the type of this channel (only conversion between text and news is supported)
+   * @param {string} type The new channel type
+   * @param {string} reason Reason for changing the channel's type
```
NotSugden, line 92:
```suggestion
   * @param {string} [reason] Reason for changing the channel's type
```
optional

typings/index.d.ts
```diff
@@ -2230,6 +2232,7 @@ declare module 'discord.js' {
 
   interface ChannelData {
     name?: string;
+    type?: string;
```
NotSugden, line 2235: i think theres a type for this somewhere, ChannelType i think

NotSugden (changes requested):

typings/index.d.ts
```diff
@@ -1188,6 +1188,7 @@ declare module 'discord.js' {
       options?: { avatar?: BufferResolvable | Base64Resolvable; reason?: string },
     ): Promise<Webhook>;
     public setNSFW(nsfw: boolean, reason?: string): Promise<NewsChannel>;
+    public setType(type: string, reason?: string): Promise<GuildChannel>;
```
NotSugden, line 1191: could the same ChannelType be used here ?

typings/index.d.ts
```diff
@@ -1529,6 +1530,7 @@ declare module 'discord.js' {
     ): Promise<Webhook>;
     public setNSFW(nsfw: boolean, reason?: string): Promise<TextChannel>;
     public setRateLimitPerUser(rateLimitPerUser: number, reason?: string): Promise<TextChannel>;
+    public setType(type: string, reason?: string): Promise<GuildChannel>;
```
NotSugden, line 1533: here too

typings/index.d.ts
```diff
@@ -2230,6 +2232,10 @@ declare module 'discord.js' {
 
   interface ChannelData {
     name?: string;
+    type?: Exclude<
+      keyof typeof ChannelType | ChannelType,
+      'dm' | 'group' | 'unknown' | ChannelType.dm | ChannelType.group | ChannelType.unknown
+    >;
```
NotSugden, line 2238: the method only accepts the string varient, perhaps maybe `'news' | 'text'` (since they are the only ones that can be changed to at the moment)

NotSugden (changes requested):

typings/index.d.ts
monbrey: I was originally not in favour of restricting the types (responsibility of the user), but I guess this exclusion already partially does that anyway. Will look into how best to do that

iCrawl (approved):

typings/index.d.ts
```diff
@@ -1188,6 +1188,10 @@ declare module 'discord.js' {
       options?: { avatar?: BufferResolvable | Base64Resolvable; reason?: string },
     ): Promise<Webhook>;
     public setNSFW(nsfw: boolean, reason?: string): Promise<NewsChannel>;
+    public setType(
+      type: Exclude<keyof typeof ChannelType, 'dm' | 'group' | 'unknown' | 'category' | 'voice' | 'store'>,
```
vladfrangu, line 1192: Use pick instead of Exclude

vladfrangu (changes requested):

vladfrangu (changes requested):
Can you check the types (that they're right), and update it in TextChannel and ChannelData?

monbrey:
> 
> 
> Can you check the types (that they're right), and update it in TextChannel and ChannelData?

My bad missing the other two places, is there a reason you think the `Pick<typeof ChannelType, 'text' | 'news'>` isn't right though? Just want to make sure I'm not missing something here.

vladfrangu (approved):

faith (approved):

kyranet (approved):

SpaceEEC (approved):

PR #5027, opened 2020-11-20
feat(Message): added replies

1A3Dev:
Adds option for `Message#reply` and `TextBasedChannel#send`:
- `refer` is an object and has the options `message`, `guild` and `channel` corresponding to the ids for the `message_reference` object.

Adds option for `Message#reply`:
- `refer` as a boolean is used to make the reply refer to the original message.

**Status**

### DIFF
--- a/src/structures/APIMessage.js
+++ b/src/structures/APIMessage.js
@@ -208,7 +208,11 @@ class APIMessage {
       avatar_url: avatarURL,
       allowed_mentions: typeof content === 'undefined' ? undefined : allowedMentions,
       flags,
-      message_reference: this.options.message_reference,
+      message_reference: this.options.refer ? {
+        message_id: this.options.refer.message,
+        guild_id: this.options.refer.guild,
+        channel_id: this.options.refer.channel
+      } : null,
     };
     return this;
   }
--- a/src/structures/APIMessage.js
+++ b/src/structures/APIMessage.js
@@ -208,6 +208,7 @@ class APIMessage {
       avatar_url: avatarURL,
       allowed_mentions: typeof content === 'undefined' ? undefined : allowedMentions,
       flags,
+      message_reference: this.options.message_reference,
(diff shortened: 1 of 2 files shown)


NotSugden:
duplicate of #4874

almostSouji:
This feature is being developed in https://github.com/discordjs/discord.js/pull/4874, if you want to suggest changes or additions please do so in the original PR thread.

1A3Dev:
oof I searched reply and replies before i did this and it didn't come up

1A3Dev:
This is way easier though and doesn't replace the original functionality of the reply method (also I searched before I made this and couldn't find any posts so 🤷)

vBread:
The original reply was useless seeing as it was literally just `message.channel.send(<@${id}>, whatever)`

1A3Dev:
> The original reply was useless seeing as it was literally just `message.channel.send(<@${id}>, whatever)`

still lots of people use it and it'd be an unnecessary change when it could just be added as an option like this does.

1A3Dev:
My latest commits fix the small things I missed on it.

tipakA:
> The original reply was useless seeing as it was literally just `message.channel.send(<@${id}>, whatever)`

This is the same thing as `MessagEmbed.addBlankField()` was. Literally just called `MessageEmbed.addField('\u200b', '\u200b', true)`. It was removed for exactly that reason.

If you really need that reply behavior, absolutely nothing stops you from adding a mention manually into the message text and using `send()`. Even better: doing so allows you to have flexibility for where you want the mention to be, where `reply()` forces the mention to be in front of your text and separated from it with a comma.

1A3Dev:
I can see nobody wants this so I'll be the only using it but whatever 🤷‍♂️

1A3Dev:
easier than changing all my code to add the other version of reply.

PR #5034, opened 2020-11-22
fix(Collector): throw an error if a non-function was provided as filter

SpaceEEC:

This PR asserts that the provided filter-function is actually a function, if that fails, an error indicating such is thrown.

**Status**

### DIFF
--- a/src/structures/interfaces/Collector.js
+++ b/src/structures/interfaces/Collector.js
@@ -1,6 +1,7 @@
 'use strict';
 
 const EventEmitter = require('events');
+const { TypeError } = require('../../errors');
 const Collection = require('../../util/Collection');
 const Util = require('../../util/Util');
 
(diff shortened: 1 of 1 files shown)


src/structures/Message.js
```diff
@@ -370,8 +370,8 @@ class Message extends Base {
    *   .catch(console.error);
    */
   awaitReactions(filter, options = {}) {
+    const collector = this.createReactionCollector(filter, options);
     return new Promise((resolve, reject) => {
-      const collector = this.createReactionCollector(filter, options);
```
NotSugden, line 374: i believe the promise will actually be rejected if an error is thrown inside the callback (so long as said callback function is not async), i've given this a test and that does seem to be the case, so it can be kept inside here

src/structures/interfaces/TextBasedChannel.js
```diff
@@ -298,8 +298,8 @@ class TextBasedChannel {
    *   .catch(collected => console.log(`After a minute, only ${collected.size} out of 4 voted.`));
    */
   awaitMessages(filter, options = {}) {
+    const collector = this.createMessageCollector(filter, options);
     return new Promise((resolve, reject) => {
-      const collector = this.createMessageCollector(filter, options);
```
NotSugden, line 302: same as previous comment

kyranet (approved):

iCrawl (approved):

PR #5041, opened 2020-11-24
fix(ClientPresence): add presence to ClientPresence from ClientOptions

Mirdukkk:
this PR adds data about the client's presence to the ClientPresence class from the options that were passed in the client constructor. without this PR, the client cannot display its status or activity in any way, because by default, ClientPresence is changed only using the setStatus, setActivity, and setPresence methods. so if these methods are not used, the ClientPresence class always contains only standard values, although in fact the values may differ from those displayed in the Discord. 
this PR also fixes an issue where the setStatus and setActivity methods overwritten values for each other. that is, if I set the idle status to the client via setStatus, and then use setActivity to set the activity, the idle status will simply disappear. similarly, in the opposite direction.

**Status**

### DIFF
--- a/src/client/Client.js
+++ b/src/client/Client.js
@@ -131,7 +131,7 @@ class Client extends BaseClient {
      * @private
      * @type {ClientPresence}
      */
-    this.presence = new ClientPresence(this);
+    this.presence = new ClientPresence(this, options.presence);
 
     Object.defineProperty(this, 'token', { writable: true });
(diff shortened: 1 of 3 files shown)


Mirdukkk:
hey guys, review my PR please

vladfrangu (approved):

kyranet (approved):

iCrawl (approved):

PR #5047, opened 2020-11-25
fix(MessageReaction): set MessageReaction#me in patch method

### LINKED ISSUE #5046: Incorrect reaction.me value on first messageReactionAdd event trigger
`reaction.me` incorrectly returns false the first time `client.on("messageReactionAdd")` is called, even when `reaction.fetch()` is called.

```typescript
client.on("messageReactionAdd", async (reaction, user) => {
   if (reaction.partial) await reaction.fetch().catch(() => null);
   const reaction2 = reaction.message.reactions.cache.find((r) => reaction.emoji.id === r.emoji.id);
   // reaction2 should be identical to reaction
   console.log(reaction.me);  // FALSE --- (this is incorrect)
   console.log(reaction2.me);  // TRUE
}
```

**Further details:**
I am using the event to watch for a user giving the same reaction as the bot on an old message.  
This issue only occurs the first time this event is triggered. `reaction.me` is the correct value and subsequent times this event is called.  
I tried a few different solutions like fetching the reaction twice, or using setTimeout to wait between the event trigger and the reaction fetch. Neither fixed the issue.

- discord.js version: 12.5.0
- Node.js version: 15.2.1
- Priority this issue should have – please be realistic and elaborate if possible: Low. The code I use for reaction2 can be used to re-assign reaction with the proper value.

**Relevant client options:**

- partials: ["MESSAGE", "REACTION", "USER"],
- gateway intents: ["GUILDS", "GUILD_VOICE_STATES", "GUILD_MESSAGES", "GUILD_MESSAGE_REACTIONS"]
- other: none

monbrey:
Fixes #5046 by setting `MessageReaction#me` in `MessageReaction#_patch` ensuring it gets updated when fetching partials.

**Status**

### DIFF
--- a/src/structures/MessageReaction.js
+++ b/src/structures/MessageReaction.js
@@ -29,12 +29,6 @@ class MessageReaction {
     this.message = message;
 
     /**
-     * Whether the client has given this reaction
-     * @type {boolean}
-     */
-    this.me = data.me;
-
-    /**
      * A manager of the users that have given this reaction
      * @type {ReactionUserManager}
      */
@@ -54,6 +48,12 @@ class MessageReaction {
        */
       this.count = data.count;
     }
+
+    /**
+     * Whether the client has given this reaction
+     * @type {boolean}
+     */
+    this.me = data.me;
   }
 
   /**


iCrawl (approved):

ArrowM:
Thanks buddy :)

ArrowM:
I just updated to 12.5.1. I think your fix for this reversed the issue - but did not fix it. Now `reaction.me` is the correct value the first time the event is triggered, but returns false on every subsequent trigger.

monbrey:
Sounds like something I'd cause 😁 will look into it further

PR #5050, opened 2020-11-26
revert: 5047

monbrey:
Reverts #5047 because I think it introduced more bugs than it solved. 

There's still a very strange issue in here somewhere which results in `MessageReaction#me` being updated when you fetch it... but not until after fetch has returned... just in the cache.

image

**Status**

### DIFF
--- a/src/structures/MessageReaction.js
+++ b/src/structures/MessageReaction.js
@@ -29,6 +29,12 @@ class MessageReaction {
     this.message = message;
 
     /**
+     * Whether the client has given this reaction
+     * @type {boolean}
+     */
+    this.me = data.me;
+
+    /**
      * A manager of the users that have given this reaction
      * @type {ReactionUserManager}
      */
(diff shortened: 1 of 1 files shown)


BannerBomb:
I've took a few hours testing the original issue using the commit [4fcb9ebf300633022e2b9867fa06a586307ff17a](https://github.com/discordjs/discord.js/commit/4fcb9ebf300633022e2b9867fa06a586307ff17a) which is before your PR was merged to fix the issue. I found the issue to be here https://github.com/discordjs/discord.js/blob/4fcb9ebf300633022e2b9867fa06a586307ff17a/src/client/actions/MessageReactionAdd.js#L36-L40. However, the only fix for it that can think of is if you fetch the reactions users before emitting the event. It happens because the message doesn't already have the "me" value set since not all of the reacted users have been cached. This isn't an issue when not using partials because the event isn't fired since it's a partial.

kyranet (approved):

vladfrangu (approved):

SpaceEEC (approved):

PR #5054, opened 2020-11-27
chore(Release): version upgrade

Vaporox:
This change was only applied to the stable branch, not master.

*Side note: stable still has the version set to `12.5.0` in `package-lock.json` FWIW*


iCrawl:
This isn't needed and shouldn't be done in this case. Stable will be rebased onto master to get rid of the double  commits because I cherry-picked.

PR #5065, opened 2020-12-01
cleanup(Client): improve generateInvite()

Vaporox:

- Move `InviteGenerationOptions` typedef from below the class to above the method (consistency)
- Passing a `PermissionResolvable` is deprecated, so it now throws an error if the passed options are not an object
- ~Fetching the application doesn't seem required, so the method doesn't return a promise anymore~
- Only add permissions to the query if they aren't `0`

**Status**

### DIFF
--- a/src/client/Client.js
+++ b/src/client/Client.js
@@ -373,38 +373,46 @@ class Client extends BaseClient {
   }
 
   /**
-   * Generates a link that can be used to invite the bot to a guild.
-   * @param {InviteGenerationOptions|PermissionResolvable} [options] Permissions to request
-   * @returns {Promise<string>}
-   * @example
-   * client.generateInvite({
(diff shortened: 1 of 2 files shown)


NotSugden:
i believe the application was fetched because some older bots actually have different application and bot user IDs (iirc anyway)

Vaporox:
Got a confirmation on DAPI and actually could reproduce on e.g. MEE6 or Dyno, that's unfortunate but I'll revert that then

vladfrangu (approved):

kyranet (approved):

iCrawl (approved):

Vaporox:
Since `application.id` is sent with `READY` now, if anyone wants to implement that feature I'll open another PR to actually remove the promise here

PR #5066, opened 2020-12-01
fix(Message): update getters to take null permissions into account

### LINKED ISSUE #5056: Channel permissionsFor can be null
When using `Channel.permissionsFor`, function can return null but there is no check for that.

I havn't used permissionsFor directly but used to check `Message.deleted` and received an error.
```js
TypeError: Cannot read property 'has' of null
    at Message.get deletable [as deletable] (/home/koyamie/Koya/node_modules/discord.js/src/structures/Message.js:412:69)
```
Example here: https://github.com/discordjs/discord.js/blob/master/src/structures/Message.js#L412

**Further details:**

- discord.js version: latest master commit
- Node.js version: 14.15.1
- Operating system: debian 9
- Priority this issue should have – please be realistic and elaborate if possible: medium

**Relevant client options:**

- partials: ['REACTION','MESSAGE']
- gateway intents: 13999
- other: none

- [x] I have also tested the issue on latest master, commit hash:

zaida04:
#5056, check to make sure return from channel#permissionsFor isn't null before calling .has on it. Open to refactoring to avoid duplicate permissionsFor calls

closes #5056 

**Status**

- [] Code changes have been tested against the Discord API, or there are no code changes

### DIFF
--- a/src/structures/Message.js
+++ b/src/structures/Message.js
@@ -180,9 +180,9 @@ class Message extends Base {
      */
     this.activity = data.activity
       ? {
-        partyID: data.activity.party_id,
-        type: data.activity.type,
-      }
+          partyID: data.activity.party_id,
+          type: data.activity.type,
+        }
       : null;
 
     /**
@@ -218,10 +218,10 @@ class Message extends Base {
      */
     this.reference = data.message_reference
       ? {
-        channelID: data.message_reference.channel_id,
-        guildID: data.message_reference.guild_id,
-        messageID: data.message_reference.message_id,
-      }
+          channelID: data.message_reference.channel_id,
+          guildID: data.message_reference.guild_id,
+          messageID: data.message_reference.message_id,
+        }
       : null;
   }
 
@@ -409,8 +409,8 @@ class Message extends Base {
     return (
       !this.deleted &&
       (this.author.id === this.client.user.id ||
-        ((this.guild && this.channel.permissionsFor(this.client.user)?.has(Permissions.FLAGS.MANAGE_MESSAGES, false)))
-      ));
+        (this.guild && this.channel.permissionsFor(this.client.user)?.has(Permissions.FLAGS.MANAGE_MESSAGES, false)))
+    );
   }
 
   /**
--- a/src/structures/Message.js
+++ b/src/structures/Message.js
@@ -409,7 +409,9 @@ class Message extends Base {
     return (
       !this.deleted &&
       (this.author.id === this.client.user.id ||
-        (this.guild && this.channel.permissionsFor(this.client.user).has(Permissions.FLAGS.MANAGE_MESSAGES, false)))
+        (this.guild &&
+          this.channel.permissionsFor(this.client.user) &&
+          this.channel.permissionsFor(this.client.user).has(Permissions.FLAGS.MANAGE_MESSAGES, false)))
     );
   }
 
--- a/src/structures/Message.js
+++ b/src/structures/Message.js
@@ -406,12 +406,11 @@ class Message extends Base {
    * @readonly
    */
   get deletable() {
+    const permissions = this.guild && this.channel.permissionsFor(this.client.user);
     return (
       !this.deleted &&
       (this.author.id === this.client.user.id ||
-        (this.guild &&
-          this.channel.permissionsFor(this.client.user) &&
-          this.channel.permissionsFor(this.client.user).has(Permissions.FLAGS.MANAGE_MESSAGES, false)))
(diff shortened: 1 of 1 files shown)


src/structures/Message.js
```diff
@@ -409,7 +409,9 @@ class Message extends Base {
     return (
       !this.deleted &&
       (this.author.id === this.client.user.id ||
-        (this.guild && this.channel.permissionsFor(this.client.user).has(Permissions.FLAGS.MANAGE_MESSAGES, false)))
+        (this.guild &&
+          this.channel.permissionsFor(this.client.user) &&
+          this.channel.permissionsFor(this.client.user).has(Permissions.FLAGS.MANAGE_MESSAGES, false)))
```
NotSugden, line 414: why call this twice?
wouldn't it be better to do something like
```js
const permissions = this.guild && this.channel.permissionsFor(this.client.user);
return !permissions || permissions.has(...);```

Comment (kyranet) [reply src/structures/Message.js:414]: If #5067 merges first, this line can be refactored to
```suggestion
          this.channel.permissionsFor(this.client.user)?.has(Permissions.FLAGS.MANAGE_MESSAGES, false) ?? false))
```

iCrawl: The change @kyranet mentioned can be implemented here now.

vladfrangu (approved):

src/structures/Message.js
```diff
@@ -409,7 +409,7 @@ class Message extends Base {
     return (
       !this.deleted &&
       (this.author.id === this.client.user.id ||
-        (this.guild && this.channel.permissionsFor(this.client.user).has(Permissions.FLAGS.MANAGE_MESSAGES, false)))
+        (this.guild && this.channel.permissionsFor(this.client.user)?.has(Permissions.FLAGS.MANAGE_MESSAGES, false)))
```
NotSugden, line 412: this will return `undefined` if there are no permissions returned by `permissionsFor`, or alternatively return `null` if there is no guild, instead of a boolean like its meant to

NotSugden (changes requested):

vladfrangu (changes requested):
Need to handle reported issue first

zaida04:
@vladfrangu resolved

src/structures/Message.js
```diff
@@ -406,10 +406,10 @@ class Message extends Base {
    * @readonly
    */
   get deletable() {
-    return (
+    return Boolean(
       !this.deleted &&
-      (this.author.id === this.client.user.id ||
-        (this.guild && this.channel.permissionsFor(this.client.user).has(Permissions.FLAGS.MANAGE_MESSAGES, false)))
+        (this.author.id === this.client.user.id ||
+          (this.guild && this.channel.permissionsFor(this.client.user)?.has(Permissions.FLAGS.MANAGE_MESSAGES, false))),
```
Vaporox, line 412:
```suggestion
          this.channel.permissionsFor?.(this.client.user)?.has(Permissions.FLAGS.MANAGE_MESSAGES)),
```
needs more optional chaining :cowboy_hat_face:

also `false` shouldnt be passed here because administrators can delete others messages

src/structures/Message.js
```diff
@@ -406,10 +406,10 @@ class Message extends Base {
    * @readonly
    */
   get deletable() {
-    return (
+    return Boolean(
       !this.deleted &&
-      (this.author.id === this.client.user.id ||
-        (this.guild && this.channel.permissionsFor(this.client.user).has(Permissions.FLAGS.MANAGE_MESSAGES, false)))
+        (this.author.id === this.client.user.id ||
+          this.channel.permissionsFor?.(this.client.user)?.has(Permissions.FLAGS.MANAGE_MESSAGES)),
```
vladfrangu, line 412: When...will permissionsFor as a function be null or undefined on a channel where a message is in?
```suggestion
          this.channel.permissionsFor(this.client.user)?.has(Permissions.FLAGS.MANAGE_MESSAGES)),
```

vladfrangu (changes requested):

src/structures/Message.js
NotSugden: it would be undefined in a dm channel

vladfrangu (approved):

iCrawl (approved):

NotSugden:
the same error also happens for `Message#pinnable`, which is right underneath this getter, it would make sense to include it in this PR too

SpaceEEC (approved):

kyranet (approved):

PR #5068, opened 2020-12-01
feat(GuildMember): add 'resetNickname' method

jecsham:

We know that there are alternatives to reset a nicknames, like:
```js
<GuildMember>.setNickname('');
// or
<GuildMember>.setNickname(<GuildMember>.user.username) // not recommended
```
however, not many of us know this and it may take some time for some to realize it, as has been my case. That's why I decided to create this alias.

**Status**

### DIFF
--- a/src/structures/GuildMember.js
+++ b/src/structures/GuildMember.js
@@ -328,6 +328,15 @@ class GuildMember extends Base {
   }
 
   /**
+   * Resets the nickname for this member.
+   * @param {string} [reason] Reason for resetting the nickname
+   * @returns {Promise<GuildMember>}
+   */
+  resetNickname(reason) {
+    return this.edit({ nick: '' }, reason);
+  }
+
+  /**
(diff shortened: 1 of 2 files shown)


anandre:
I would argue that better documentation on the `setNickname` method would be preferable instead of making a helper method to do it.  Passing `null` to the method will reset it.

tl;dr: make a PR to make documentation/example better and show using `null` not a whole new helper method

jecsham:
@anandre I had thought about it, but I think its more intuitive with a helper function. 
With `setNickname` it would still be confusing to figure out. I'm not saying that consulting the documentation is bad, but usually many people uses typescript, so they can see methods and properties that an object has, which saves time.

I understand that the helper functions are usually redundant code, but they make a library much easier to use.

Anyway, if you don't change your mind, @zaida04 started the PR you requested, so you can close this one without problems!

iCrawl:
See #5070

Issue #2894, opened 2018-10-13
Support for additional ffmpeg input parameters on voice.


Deivu:
**Is your feature request related to a problem? Please describe.**
I noticed that Discord.js Voice doesn't support additional input parameters on ffmpeg encoder. Because of this, we cannot really tweak the sound output that much.

**Describe the ideal solution**
This can be implemented on the <Voice Connection>.play() options. Where in we can put the additional params on the ffmpeg along side with the default one.

**Describe alternatives you've considered**
I thought about recreating the readable stream, but that can become a cpu intensive workload because of double processing that happens, where in you can just use the ffmpeg to fine tune the sound.

amishshah:
What additional parameters would you use?

Deivu:
the -af equalizer

PR #5083, opened 2020-12-05
Added an option to provide custom FFMPEG arguments.

VimHax:

Added a new property to the `StreamOptions` object which gives users the ability to provide custom FFMPEG arguments or entirely replace the already existing FFMPEG arguments with their own. This gives more control over to users who are using FFMPEG for streaming.

**Status**

- [ * ] Code changes have been tested against the Discord API, or there are no code changes
- [ * ] I know how to update typings and have done so, or typings don't need updating

- [ * ] This PR changes the library's interface (methods or parameters added)
- [   ] This PR includes breaking changes (methods removed or renamed, parameters moved or removed)
- [   ] This PR **only** includes non-code changes, like changes to documentation, README, etc.

### DIFF
--- a/src/client/voice/player/BasePlayer.js
+++ b/src/client/voice/player/BasePlayer.js
@@ -41,7 +41,8 @@ class BasePlayer extends EventEmitter {
 
     const isStream = input instanceof ReadableStream;
 
-    const args = isStream ? FFMPEG_ARGUMENTS.slice() : ['-i', input, ...FFMPEG_ARGUMENTS];
+    const ARGS = options.replaceArgs ? options.customFFMPEGArgs : [...FFMPEG_ARGUMENTS, ...options.customFFMPEGArgs];
+    const args = isStream ? ARGS.slice() : ['-i', input, ...ARGS];
     if (options.seek) args.unshift('-ss', String(options.seek));
 
     const ffmpeg = new prism.FFmpeg({ args });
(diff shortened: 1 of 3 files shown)


Deivu:
You could just use the opus directly yourself.

Refer to Hydra's comment in #2894

VimHax:
Oh, I didn't know it was so simple to add a transcoder... I've googled a bunch but I couldn't find such a simple solution... I suppose that is the solution then to the problem I attempted to fix, however, this is also just a small and simple addition which you can use for small changes to the parameters, if you want something bigger, you can use a full blown transcoder. For my purpose, using such a transcoder is definitely overkill...

Also what about performance impacts? Wouldn't running a separate transcoder add some overhead versus just adding extra parameters to the already existing encoder?

Deivu:
If you are already passing an opus stream, it won't, besides even you pass for example the ytdl stream to play, it would still do the same work of converting it so and forth behind the scenes, think of it that you are just doing it when you use prism-media directly. 

Refer to https://github.com/discordjs/discord.js/blob/master/src/client/voice/player/BasePlayer.js

Think of it as, you will just do the ffmpeg convertion, and pass it to `playPCMStream` which D.JS does when you pass an unknown stream to it

amishshah:
I agree with everything @Deivu has said above! Customisation such as this is outside the scope of the library, but following the advice Deivu has given above should allow you to do what you're trying to do.

PR #5090, opened 2020-12-09
test(roles|channels): Added unit tests for roles and channels

hm343:
Added tests for the creation, editing, assignment, and deletion of roles. Added tests for the creation, editing, and deletion of channels, both text and voice.

**Status**

### DIFF
--- a/test/channels.js
+++ b/test/channels.js
@@ -1,0 +1,71 @@
+'use strict';
+
+const assert = require('assert');
+const { token, test_guild_id } = require('./auth');
+const { Client } = require('../src');
+
+// Create client
+const client = new Client({});
+
+// Test role create and deletion
+client.on('ready', async () => {
+
+  try {
+
+    // Fetch guild to use
+    const guild = await client.guilds.fetch(test_guild_id);
+
+    // Create a new text channel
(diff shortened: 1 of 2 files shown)


test/channels.js
```diff
@@ -0,0 +1,71 @@
+'use strict';
+
+const assert = require('assert');
+const { token, test_guild_id } = require('./auth');
+const { Client } = require('../src');
+
+// Create client
+const client = new Client({});
```
thetayloredman, line 8: Is there a reason we can't just do `new Client()`?

test/roles.js
```diff
@@ -0,0 +1,99 @@
+'use strict';
+
+const assert = require('assert');
+const { token, test_guild_id, test_user_id } = require('./auth');
+const { Client } = require('../src');
+
+// Create client
+const client = new Client({});
```
thetayloredman, line 8: Is there a reason we can't just do `new Client()`?

iCrawl:
We are probably going to move away from manual tests like this (apart from the fact that we can't really run those automated anyway in this way or fashion).

Thank you for the heads-up in how we could approach this, though 👍

PR #5108, opened 2020-12-12
fix(Message): #system non-zero message types are not guaranteed to be system

ckohen:
As of discord/discord-api-docs#2118 not all non-zero message types are system. 
Type 19 `Reply` is the only current type that benefits from this change.

Note: the reply type will only be recieved once #4879 is implemented.

This PR also introduces SystemMessageTypes so that future message types can be more easily sorted.

**Status**

### DIFF
--- a/src/util/Constants.js
+++ b/src/util/Constants.js
@@ -435,8 +435,8 @@ exports.MessageTypes = [
  * * REPLY
  * @typedef {string} SystemMessageType
  */
-exports.SystemMessageTypes = module.exports.MessageTypes.filter(
-  type => type !== null && type !== 'DEFAULT' && type !== 'REPLY',
+exports.SystemMessageTypes = exports.MessageTypes.filter(
+  type => type && type !== 'DEFAULT' && type !== 'REPLY',
 );
 
 /**
--- a/src/util/Constants.js
+++ b/src/util/Constants.js
@@ -435,9 +435,7 @@ exports.MessageTypes = [
  * * REPLY
  * @typedef {string} SystemMessageType
  */
-exports.SystemMessageTypes = exports.MessageTypes.filter(
-  type => type && type !== 'DEFAULT' && type !== 'REPLY',
-);
+exports.SystemMessageTypes = exports.MessageTypes.filter(type => type && type !== 'DEFAULT' && type !== 'REPLY');
 
 /**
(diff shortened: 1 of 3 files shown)


typings/index.d.ts
```diff
@@ -3148,6 +3148,22 @@ declare module 'discord.js' {
...
+    | 'CHANNEL_ICON_CHANGE'
+    | 'PINS_ADD'
+    | 'GUILD_MEMBER_JOIN'
+    | 'USER_PREMIUM_GUILD_SUBSCRIPTION'
+    | 'USER_PREMIUM_GUILD_SUBSCRIPTION_TIER_1'
+    | 'USER_PREMIUM_GUILD_SUBSCRIPTION_TIER_2'
+    | 'USER_PREMIUM_GUILD_SUBSCRIPTION_TIER_3'
+    | 'CHANNEL_FOLLOW_ADD'
+    | 'GUILD_DISCOVERY_DISQUALIFIED'
+    | 'GUILD_DISCOVERY_REQUALIFIED';
```
Vaporox, line 3165:
```suggestion
  type SystemMessageType = Exclude<MessageType, 'DEFAULT' | 'REPLY'>;
```
maybe this would make more sense? however it's inconsistent with the actual exported constant then

ckohen: This is a good point, the export should be exclusionary as well, making it consistent.

src/util/Constants.js
```diff
@@ -429,6 +429,16 @@ exports.MessageTypes = [
...
 ];
 
+/**
+ * The types of messages that are `System`. The available types are `MessageTypes` excluding:
+ * * DEFAULT
+ * * REPLY
+ * @typedef {string} SystemMessageType
+ */
+exports.SystemMessageTypes = module.exports.MessageTypes.filter(
+  type => type !== null && type !== 'DEFAULT' && type !== 'REPLY',
```
Vaporox, line 439:
```suggestion
exports.SystemMessageTypes = exports.MessageTypes.filter(
  type => type && type !== 'DEFAULT' && type !== 'REPLY',
```
last nitpick

Vaporox:
oh yeah, you need to add it to [`Constants`](https://github.com/discordjs/discord.js/blob/master/typings/index.d.ts#L518) too

vladfrangu (approved):

kyranet (approved):

SpaceEEC (approved):

iCrawl (approved):

PR #5113, opened 2020-12-12
feat(Browser): remove browser 😊

iCrawl:
**Please describe the changes this PR makes and why it should be merged:**

**Status**

- [ ] Code changes have been tested against the Discord API, or there are no code changes
- [x] I know how to update typings and have done so, or typings don't need updating

**Semantic versioning classification:**

- [x] This PR changes the library's interface (methods or parameters added)
  - [x] This PR includes breaking changes (methods removed or renamed, parameters moved or removed)
- [ ] This PR **only** includes non-code changes, like changes to documentation, README, etc.

PR #5130, opened 2020-12-17
refactor(typings): many cool changes

### LINKED ISSUE #5015: Move ChannelType typings to within discord.js module
A clear and concise description of what the problem is. Eg. I'm always frustrated when [...]

The `ChannelType` enum typings are declared at the top of `typings/index.d.ts`, outside of the `discord.js` module. Thus, they appear at the root of all TypeScript projects which have discord.js in their packages. 

Type Definition:
https://github.com/discordjs/discord.js/blob/b8fd3f65d964f080afb42dd37eb21a13c0d84f8f/typings/index.d.ts#L1-L10 

For example, if I import `discord.js` as follows:

```
import * as Discord from "discord.js";
```

then I would expect to access `ChannelType` as such:

```
let myType = Discord.ChannelType.text;
```

Instead, `ChannelType` appears at the root:

```
let myType = ChannelType.text;
```

and gives an error at runtime:

```
ReferenceError: ChannelType is not defined
```

Move `ChannelType` typings into the `discord.js` module.

I need to use this for `GuildCreateChannelOptions` as seen here:
https://github.com/discordjs/discord.js/blob/b8fd3f65d964f080afb42dd37eb21a13c0d84f8f/typings/index.d.ts#L2562-L2565

So for now I just use the `any` type and assign `"text"` or `"voice"` as needed.

**Additional context**

I've noticed a near duplicate inside `Constants`:

https://github.com/discordjs/discord.js/blob/b8fd3f65d964f080afb42dd37eb21a13c0d84f8f/typings/index.d.ts#L556-L564

Not sure what to do with this.

vaporox:

- Remove `declare module` to allow for npm aliases
- Remove `null` as a possible value for `Constants.UserAgent` as of #5113
- Replace `PathLike` with `string`
- Replace `{ [key: X]: Y }` with `Record<X, Y>`
​
- Move `AllowedImageFormat` from region Classes to region Typedefs
- Move `Collection` from region Collections to region Classes
- Move `version` and `Constants` into their own region Constants
​
- Remove overloads for `Collection#flatMap` and `Collection#mapValues` that already exist in `BaseCollection`
- Change the return type of `Collection#toJSON` to be an array which depends on `V`
- Create a type `Flatten<T>` used for `Util#flatten` and `Collection#toJSON`
​
- Update `index.ts` accordingly

Resolves #5015.

**Status**

### DIFF
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -2822,9 +2822,9 @@ export interface MessageOptions {
 export type MessageReactionResolvable = MessageReaction | Snowflake;
 
 export interface MessageReference {
-  channelID: string;
-  guildID: string;
-  messageID: string | null;
+  channelID: Snowflake;
+  guildID: Snowflake;
+  messageID: Snowflake | null;
 }
 
 export type MessageResolvable = Message | Snowflake;
@@ -3088,7 +3088,7 @@ export interface RolePosition {
   position: number;
 }
 
-export type RoleResolvable = Role | string;
+export type RoleResolvable = Role | Snowflake;
 
 export interface RoleTagData {
   botID?: Snowflake;
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -2860,11 +2860,11 @@ export type OverwriteResolvable = PermissionOverwrites | OverwriteData;
 
 export type OverwriteType = 'member' | 'role';
 
-export interface PermissionFlags extends Record<PermissionString, number> {}
+export type PermissionFlags = Record<PermissionString, number>;
 
-export interface PermissionObject extends Record<PermissionString, boolean> {}
+export type PermissionObject = Record<PermissionString, boolean>;
 
-export interface PermissionOverwriteOption extends Partial<Record<PermissionString, boolean | null>> {}
+export type PermissionOverwriteOption = Partial<Record<PermissionString, boolean | null>>;
 
 export type PermissionResolvable = BitFieldResolvable<PermissionString>;
 
@@ -2901,7 +2901,7 @@ export type PermissionString =
   | 'MANAGE_WEBHOOKS'
   | 'MANAGE_EMOJIS';
 
-export interface RecursiveArray<T> extends ReadonlyArray<T | RecursiveArray<T>> {}
+export type RecursiveArray<T> = ReadonlyArray<T | RecursiveArray<T>>;
 
 export type RecursiveReadonlyArray<T> = ReadonlyArray<T | RecursiveReadonlyArray<T>>;
 
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -1,4 +1,2226 @@
-declare enum ChannelType {
+import BaseCollection from '@discordjs/collection';
+import { ChildProcess } from 'child_process';
+import { EventEmitter } from 'events';
+import { Readable, Stream, Writable } from 'stream';
+import * as WebSocket from 'ws';
+
+//#region Classes
+
+export class Activity {
+  constructor(presence: Presence, data?: object);
+  public applicationID: Snowflake | null;
+  public assets: RichPresenceAssets | null;
+  public readonly createdAt: Date;
+  public createdTimestamp: number;
+  public details: string | null;
+  public emoji: Emoji | null;
+  public flags: Readonly<ActivityFlags>;
+  public name: string;
+  public party: {
+    id: string | null;
+    size: [number, number];
+  } | null;
+  public state: string | null;
+  public timestamps: {
+    start: Date | null;
+    end: Date | null;
+  } | null;
+  public type: ActivityType;
+  public url: string | null;
+  public equals(activity: Activity): boolean;
+}
+
+export class ActivityFlags extends BitField<ActivityFlagsString> {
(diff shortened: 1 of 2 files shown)


typings/index.d.ts
```diff
@@ -1,4 +1,2226 @@
...
+}
+
+export interface ChannelPosition {
+  channel: ChannelResolvable;
+  position: number;
+}
+
+export type ChannelResolvable = Channel | Snowflake;
+
+export enum ChannelType {
```
kyranet, line 2223: `ChannelType` shouldn't be exported, it's a private type made for convenience. It also does not exist as a JSDoc but as the description of [`Channel#type`](https://discord.js.org/#/docs/main/stable/class/Channel?scrollTo=type), therefore, it shouldn't be exported.

kyranet (changes requested):

typings/index.d.ts
Vaporox: You cannot declare enums in declaration files without exporting them, so it's either convert to a type (could do that here) or actually create an export in the source (would probably need another `fix:` pr)

vladfrangu: Can you share the error you get by doing `declare enum ChannelType`? That's also what compiled TS code does with enums (it does `export declare enum <name>`)?

Vaporox: It doesn't give an error but you can still import the enum that way

typings/index.d.ts
```diff
@@ -9,3260 +2231,1007 @@ declare enum ChannelType {
...
+}
 
-  type SpeakingString = 'SPEAKING' | 'SOUNDSHARE' | 'PRIORITY_SPEAKING';
+export interface RolePosition {
+  role: RoleResolvable;
+  position: number;
+}
 
-  type StreamType = 'unknown' | 'converted' | 'opus' | 'ogg/opus' | 'webm/opus';
+export type RoleResolvable = Role | string;
```
NotSugden, line 3091:
```suggestion
export type RoleResolvable = Role | Snowflake;
```

typings/index.d.ts
```diff
@@ -9,3260 +2231,1007 @@ declare enum ChannelType {
...
+  | 'DEAFEN_MEMBERS'
+  | 'MOVE_MEMBERS'
+  | 'USE_VAD'
+  | 'CHANGE_NICKNAME'
+  | 'MANAGE_NICKNAMES'
+  | 'MANAGE_ROLES'
+  | 'MANAGE_WEBHOOKS'
+  | 'MANAGE_EMOJIS';
+
+export interface RecursiveArray<T> extends ReadonlyArray<T | RecursiveArray<T>> {}
```
NotSugden, line 2904:
```suggestion
export type RecursiveArray<T> = ReadonlyArray<T | RecursiveArray<T>>;
```

typings/index.d.ts
```diff
@@ -9,3260 +2231,1007 @@ declare enum ChannelType {
...
+
+export type OverwriteResolvable = PermissionOverwrites | OverwriteData;
+
+export type OverwriteType = 'member' | 'role';
+
+export interface PermissionFlags extends Record<PermissionString, number> {}
+
+export interface PermissionObject extends Record<PermissionString, boolean> {}
+
+export interface PermissionOverwriteOption extends Partial<Record<PermissionString, boolean | null>> {}
```
NotSugden, line 2867:
```suggestion
export type PermissionFlags = Record<PermissionString, number>;

export type PermissionObject = Record<PermissionString, boolean>;

export type PermissionOverwriteOption = Partial<Record<PermissionString, boolean | null>>;
```

typings/index.d.ts
```diff
@@ -9,3260 +2231,1007 @@ declare enum ChannelType {
...
+  split?: boolean | SplitOptions;
+  replyTo?: MessageResolvable;
+}
+
+export type MessageReactionResolvable = MessageReaction | Snowflake;
+
+export interface MessageReference {
+  channelID: string;
+  guildID: string;
+  messageID: string | null;
```
NotSugden, line 2827:
```suggestion
  channelID: Snowflake;
  guildID: Snowflake;
  messageID: Snowflake | null;
```

NotSugden (changes requested):
this diff is a bit tricky to read, i believe this branch is a few commits behind too, so possibly needs a rebase

typings/index.d.ts
```diff
@@ -1,4 +1,2226 @@
...
+
+export interface ChannelPosition {
+  channel: ChannelResolvable;
+  position: number;
+}
+
+export type ChannelResolvable = Channel | Snowflake;
+
+export enum ChannelType {
   text = 0,
```
Mesteery, line 2224: An enum starts by default at 0. So there is no need to explicitely indicate the index.

Mesteery (changes requested):

typings/index.d.ts
vaporox: that's only valid for source files, not for declaration files.

vaporox:
Closing this as there are too many conflicts that can't get resolved easily. Might pick it up again towards the end of the v12 lifecycle

PR #5132, opened 2020-12-18
refactor: new node features

NotSugden:
This PR makes use of optional chaining and the nullish coalescing operator in various places, added in node v14,

aswell as a couple logic changes to use `&&` instead of a tenary like `x ? y : null` where x can be null

**Status**

  - untested as of yet
  - i might need some help fully testing this just because changing `||` to `??` may cause some unintended behaviour


BannerBomb:
Shouldn't this be `if (cache) existing._patch?.(data);`. The code is already checking that the variable "existing" isn't undefined and on this line before you made the change was `if (existing._patch &&`
https://github.com/discordjs/discord.js/pull/5132/files#diff-19338f04c86e704e9e4cb898fc9d21411ca4774ecae906a4e3179a306f55974dR25

Shouldn't this be `if (cache) existing?._patch?.(data);`. The code before you made the change was `if (existing && existing._patch &&`.
https://github.com/discordjs/discord.js/pull/5132/files#diff-649365ccd46d128b24504a4bcbb566995be7338be1ca43868ad43ae9eaf802e7R46
Should be 

This is missing a `.delete()` at the end of the api.
https://github.com/discordjs/discord.js/pull/5132/files#diff-0362e882b074f7f48e846fdc34e56a307b11eb067d140ddd0ec4b095cf86ce74R60-R63

All of the below can be ignored if they were intentionally done that way. I just brought them up because the other changes I saw were being made.

This isn't a error, but just a suggestion. Couldn't you use a for loop instead of reduce on these since all that's being done here is setting data to a collection, and for loops are significantly faster at iterations?
It would also make them consistent with fetchInvites, fetchWebhooks, fetchVoiceRegions.
https://github.com/discordjs/discord.js/pull/5132/files#diff-9c03fb4d8c1b03db876641d3900a9c08f5a3a30be8d8a3b9840018507243c284R668-R674
https://github.com/discordjs/discord.js/pull/5132/files#diff-9c03fb4d8c1b03db876641d3900a9c08f5a3a30be8d8a3b9840018507243c284R708
https://github.com/discordjs/discord.js/pull/5132/files#diff-9c03fb4d8c1b03db876641d3900a9c08f5a3a30be8d8a3b9840018507243c284R695-R698

Did you mean to leave these as Promise.reject instead of changing them to throw?
https://github.com/discordjs/discord.js/pull/5132/files#diff-9c03fb4d8c1b03db876641d3900a9c08f5a3a30be8d8a3b9840018507243c284R1383
https://github.com/discordjs/discord.js/pull/5132/files#diff-6931a2b8bde677910786bd21afe3bc29740277d6510de785acfebca7e616ad89R257

Can't you just return the this._patch method here since that method already returns "this" itself?
https://github.com/discordjs/discord.js/pull/5132/files#diff-771d176d28e8d7cd4ba65413e0562da9908ee989169a6d1b1aac819ed472cc40R147-R148
https://github.com/discordjs/discord.js/pull/5132/files#diff-771d176d28e8d7cd4ba65413e0562da9908ee989169a6d1b1aac819ed472cc40R166-R167

NotSugden:
> Can't you just return the this._patch method here since that method already returns "this" itself?

actually, these methods should return `void` to be consistent with all of the other `_patch` methods

> Did you mean to leave these as Promise.reject instead of changing them to throw?

fixed

> This isn't a error, but just a suggestion. Couldn't you use a for loop instead of reduce on these since all that's being done here is setting data to a collection, and for loops are significantly faster at iterations?
It would also make them consistent with fetchInvites, fetchWebhooks, fetchVoiceRegions.
[...]

can you leave review comments on these lines? because the links you provided don't take me to them

> Shouldn't this be if (cache) existing?._patch?.(data);. The code before you made the change was if (existing && existing._patch &&.

i think every structure has `_patch` anyway, but ill add that just to make sure

src/structures/Guild.js
```diff
@@ -653,36 +649,29 @@ class Guild extends Base {
...
-          collection.set(ban.user.id, {
-            reason: ban.reason,
-            user: this.client.users.add(ban.user),
-          });
-          return collection;
-        }, new Collection()),
-      );
+  async fetchBans() {
+    const bans = await this.client.api.guilds(this.id).bans.get();
+    return bans.reduce((collection, ban) => {
```
BannerBomb, line 668: This is the first method I mentioned about using a for loop instead of reduce.

src/structures/Guild.js
```diff
@@ -697,34 +686,26 @@ class Guild extends Base {
...
-          new Collection(),
-        ),
-      );
+  async fetchIntegrations({ includeApplications = false } = {}) {
+    const integrations = await this.client.api.guilds(this.id).integrations.get({
+      query: {
+        include_applications: includeApplications,
+      },
+    });
+    return integrations.reduce(
```
BannerBomb, line 695: and the second method that could be changed into a for loop.

src/structures/Guild.js
```diff
@@ -697,34 +686,26 @@ class Guild extends Base {
...
-  fetchTemplates() {
-    return this.client.api
-      .guilds(this.id)
-      .templates.get()
-      .then(templates =>
-        templates.reduce((col, data) => col.set(data.code, new GuildTemplate(this.client, data)), new Collection()),
-      );
+  async fetchTemplates() {
+    const templates = await this.client.api.guilds(this.id).templates.get();
+    return templates.reduce((col, data) => col.set(data.code, new GuildTemplate(this.client, data)), new Collection());
```
BannerBomb, line 708: and the third method that could be changed into a for loop.

src/structures/Guild.js
kyranet: May I ask why?

And that goes out of this PR's scope, same goes for the other requests.

BannerBomb: I mentioned it because it's not really consistent with the other methods of that class the class is using a mix of for-each and reduce and reduce is slower than using for-each.

kyranet: As of Node.js 12, this pattern of reduce (and many others) get translated to nearly the same code as a for-of loop.

Once it's optimised, the performance will be nearly the same, so it really doesn't matter.

iCrawl (dismissed):
This looks mostly fine, only nit I would have is that you switched some for loops where you set into a collection to `.reduce` but left others intact. Should decide on one pattern over the other, but not mix-match so much.

NotSugden:
ill switch for loops to `reduce`, `map`, etc where possible

src/managers/MessageManager.js
```diff
@@ -135,10 +133,8 @@ class MessageManager extends BaseManager {
   }
 
   async _fetchMany(options = {}, cache) {
-    const data = await this.client.api.channels[this.channel.id].messages.get({ query: options });
-    const messages = new Collection();
-    for (const message of data) messages.set(message.id, this.add(message, cache));
-    return messages;
+    const data = await this.client.api.channels(this.channel.id).messages.get({ query: options });
+    data.reduce((messages, message) => messages.set(message.id, this.add(message, cache)), new Collection());
```
BannerBomb, line 137: This is missing a return.

src/rest/DiscordAPIError.js
```diff
@@ -57,11 +55,10 @@ class DiscordAPIError extends Error {
...
         messages.push(v);
       } else {
-        messages = messages.concat(this.flattenErrors(v, newKey));
+        return messages.concat(this.flattenErrors(v, newKey));
       }
-    }
-
-    return messages;
+      return messages;
+    });
```
BannerBomb, line 61: This is missing an initial value.

src/sharding/ShardingManager.js
```diff
@@ -216,9 +216,12 @@ class ShardingManager extends EventEmitter {
...
    */
   broadcast(message) {
-    const promises = [];
-    for (const shard of this.shards.values()) promises.push(shard.send(message));
-    return Promise.all(promises);
+    return Promise.all(
+      this.shards.reduce((promises, shard) => {
+        promises.push(shard.send(message));
+        return promises;
+      }),
```
BannerBomb, line 223: This is also missing an initial value.

src/structures/Guild.js
```diff
@@ -939,17 +933,15 @@ class Guild extends Base {
...
+    options.roles = options.roles?.reduce((array, role) => {
+      const resolvedRole = this.guild.roles.resolve(role);
+      if (!resolvedRole) {
+        throw new TypeError('INVALID_TYPE', 'options.roles', 'Array or Collection of Roles or Snowflakes', true);
       }
-      options.roles = roles;
-    }
+      array.push(role);
+      return array;
+    });
```
BannerBomb, line 943: Missing an initial value here also.

src/structures/GuildChannel.js
```diff
@@ -282,13 +282,12 @@ class GuildChannel extends Channel {
...
-    for (const member of this.guild.members.cache.values()) {
-      if (this.permissionsFor(member).has('VIEW_CHANNEL', false)) {
+    return this.guild.members.cache.reduce((members, member) => {
+      if (this.permissionsFor(member)?.has(Permissions.FLAGS.VIEW_CHANNEL, false)) {
         members.set(member.id, member);
       }
-    }
-    return members;
+      return members;
+    });
```
BannerBomb, line 290: Also missing an initial value here.

src/sharding/ShardingManager.js
```diff
@@ -221,6 +221,7 @@ class ShardingManager extends EventEmitter {
         promises.push(shard.send(message));
         return promises;
       }),
+      [],
```
BannerBomb: The initial value should be moved up one line :smile:

src/sharding/ShardingManager.js
```diff
@@ -262,9 +266,12 @@ class ShardingManager extends EventEmitter {
...
 
-    const promises = [];
-    for (const sh of this.shards.values()) promises.push(sh[method](...args));
-    return Promise.all(promises);
+    return Promise.all(
+      this.shards.reduce((promises, _shard) => {
+        promises.push(_shard[method](...args));
+        return promises;
+      }),
+    );
```
BannerBomb, line 274: I knew I forgot to mention one yesterday. But here it is. This line is also missing an initial value.

src/client/Client.js
```diff
@@ -38,7 +38,7 @@ class Client extends BaseClient {
     let data = process.env;
     try {
       // Test if worker threads module is present and used
-      data = require('worker_threads').workerData || data;
+      data = require('worker_threads').workerData ?? data;
     } catch {
       // Do nothing
     }
```
kyranet, line 44: I'm pretty sure we can remove the try/catch, worker_threads are available since 10.5.0 and are currently stable (from far before Node.js v14).

@vladfrangu thoughts?

src/client/actions/GuildEmojisUpdate.js
```diff
@@ -5,7 +5,7 @@ const Action = require('./Action');
 class GuildEmojisUpdateAction extends Action {
   handle(data) {
     const guild = this.client.guilds.cache.get(data.guild_id);
-    if (!guild || !guild.emojis) return;
+    if (!guild?.emojis) return;
```
kyranet, line 8: When would [Guild#emojis](https://discord.js.org/#/docs/main/stable/class/Guild?scrollTo=emojis) be falsy?

src/client/voice/networking/VoiceUDPClient.js
```diff
@@ -90,46 +90,45 @@ class VoiceConnectionUDPClient extends EventEmitter {
...
+        this.emit('debug', '[UDP] socket closed');
+      })
+      .once('message', message => {
+        this.emit('debug', `[UDP] message: [${[...message]}] (${message})`);
+        // Stop if the sockets have been deleted because the connection has been closed already
+        if (!this.voiceConnection.sockets.ws) return;
+
+        const packet = parseLocalPacket(message);
+        if (packet.error) {
+          this.emit('debug', `[UDP] ERROR: ${packet.error}`).emit('error', packet.error);
```
kyranet, line 108: [EventListener#emit](https://nodejs.org/dist/latest-v15.x/docs/api/events.html#events_emitter_emit_eventname_args) returns `boolean`.

```suggestion
          this.emit('debug', `[UDP] ERROR: ${packet.error}`);
          this.emit('error', packet.error);
```

src/errors/DJSError.js
```diff
@@ -15,7 +15,7 @@ function makeDiscordjsError(Base) {
     constructor(key, ...args) {
       super(message(key, args));
       this[kCode] = key;
-      if (Error.captureStackTrace) Error.captureStackTrace(this, DiscordjsError);
+      Error.captureStackTrace?.(this, DiscordjsError);
```
kyranet, line 18: The conditional was made because [Error.captureStackTrace](https://nodejs.org/api/errors.html#errors_error_capturestacktrace_targetobject_constructoropt) is a non-standard feature implemented only by V8.

Since we do not longer use webpack, we can remove said conditional.

```suggestion
      Error.captureStackTrace(this, DiscordjsError);
```

src/managers/GuildEmojiManager.js
```diff
@@ -48,23 +48,17 @@ class GuildEmojiManager extends BaseGuildEmojiManager {
...
-        data.roles.push(role.id);
+    data.roles = roles?.reduce((array, role) => {
+      const resolvedRole = this.guild.roles.resolve(role);
+      if (!resolvedRole) {
+        throw new TypeError('INVALID_TYPE', 'options.roles', 'Array or Collection of Roles or Snowflakes', true);
       }
-    }
+      array.push(role);
+      return array;
+    });
```
kyranet, line 58:
```suggestion
    }, []);
```

kyranet (approved):
You can also replace all `EventEmitter#removeListener` with the newer `EventEmitter#off`, since we use `EventEmitter#on` over `EventEmitter#addListener`.

kyranet (changes requested):
You can also replace all `EventEmitter#removeListener` with the newer `EventEmitter#off`, since we use `EventEmitter#on` over `EventEmitter#addListener`.

src/structures/Guild.js
```diff
@@ -977,17 +933,15 @@ class Guild extends Base {
...
     options.access_token = options.accessToken;
-    if (options.roles) {
-      const roles = [];
-      for (let role of options.roles instanceof Collection ? options.roles.values() : options.roles) {
-        role = this.roles.resolve(role);
-        if (!role) {
-          throw new TypeError('INVALID_TYPE', 'options.roles', 'Array or Collection of Roles or Snowflakes', true);
-        }
-        roles.push(role.id);
+    options.roles = options.roles?.reduce((array, role) => {
```
kyranet, line 936:
```suggestion
    options.roles = options.roles?.map(role => {
```

src/structures/Guild.js
```diff
@@ -977,17 +933,15 @@ class Guild extends Base {
...
+      const resolvedRole = this.guild.roles.resolve(role);
+      if (!resolvedRole) {
+        throw new TypeError('INVALID_TYPE', 'options.roles', 'Array or Collection of Roles or Snowflakes', true);
       }
-      options.roles = roles;
-    }
+      array.push(role);
+      return array;
+    }, []);
```
kyranet, line 943:
```suggestion
      return role;
    });
```

src/structures/Channel.js
```diff
@@ -118,7 +116,7 @@ class Channel extends Base {
         channel = new PartialGroupDMChannel(client, data);
       }
     } else {
-      guild = guild || client.guilds.cache.get(data.guild_id);
+      guild = guild ?? client.guilds.cache.get(data.guild_id);
```
kyranet, line 119:
```suggestion
      if (!guild) guild = client.guilds.cache.get(data.guild_id);
```

src/structures/GuildChannel.js
```diff
@@ -281,13 +282,12 @@ class GuildChannel extends Channel {
...
-    for (const member of this.guild.members.cache.values()) {
-      if (this.permissionsFor(member).has('VIEW_CHANNEL', false)) {
+    return this.guild.members.cache.reduce((members, member) => {
+      if (this.permissionsFor(member)?.has(Permissions.FLAGS.VIEW_CHANNEL)) {
         members.set(member.id, member);
       }
-    }
-    return members;
+      return members;
+    }, []);
```
kyranet, line 290:
```suggestion
    }, new Collection());
```

src/structures/GuildChannel.js
```diff
@@ -281,13 +282,12 @@ class GuildChannel extends Channel {
    * @readonly
    */
   get members() {
-    const members = new Collection();
-    for (const member of this.guild.members.cache.values()) {
-      if (this.permissionsFor(member).has('VIEW_CHANNEL', false)) {
+    return this.guild.members.cache.reduce((members, member) => {
+      if (this.permissionsFor(member)?.has(Permissions.FLAGS.VIEW_CHANNEL)) {
```
kyranet, line 286: Since we're passing a GuildMember instance, permissionsFor will always return a non-null value.
```suggestion
      if (this.permissionsFor(member).has(Permissions.FLAGS.VIEW_CHANNEL)) {
```

src/structures/GuildChannel.js
```diff
@@ -567,28 +560,31 @@ class GuildChannel extends Channel {
...
   /**
    * Whether the channel is deletable by the client user
-   * @type {boolean}
+   * @type {?boolean}
    * @readonly
    */
   get deletable() {
-    return this.permissionsFor(this.client.user).has(Permissions.FLAGS.MANAGE_CHANNELS, false);
+    const permissions = this.permissionsFor(this.client.user);
+    return permissions && permissions.has(Permissions.FLAGS.MANAGE_CHANNELS);
```
kyranet, line 568: To be honest, I'm more of a fan of throwing an error here (as `permissionsFor` will only return `undefined` when somebody sweeps their member cache in a non-sensible way, aka not re-adding the client's member) as opposed to just returning `undefined`, but we can keep the current behaviour with:
```suggestion
    return permissions ? permissions.has(Permissions.FLAGS.MANAGE_CHANNELS) : false;
```

Same goes for many other places.

cc: @vladfrangu @iCrawl

src/structures/GuildMember.js
```diff
@@ -172,7 +172,7 @@ class GuildMember extends Base {
    */
   get displayColor() {
     const role = this.roles.color;
-    return (role && role.color) || 0;
+    return role?.color ?? 0;
```
kyranet, line 175: You can do `this.roles.color?.color ?? 0;`

src/structures/GuildMember.js
```diff
@@ -182,7 +182,7 @@ class GuildMember extends Base {
    */
   get displayHexColor() {
     const role = this.roles.color;
-    return (role && role.hexColor) || '#000000';
+    return role?.hexColor ?? '#000000';
```
kyranet, line 185: Same as above: `this.roles.color?.hexColor ?? '#000000';`

src/structures/Invite.js
```diff
@@ -116,11 +116,11 @@ class Invite extends Base {
    * @readonly
    */
   get deletable() {
-    const guild = this.guild;
+    const { guild } = this;
     if (!guild || !this.client.guilds.cache.has(guild.id)) return false;
     if (!guild.me) throw new Error('GUILD_UNCACHED_ME');
```
kyranet, line 121: You can do:

```javascript
const { me } = guild;
if (!me) throw new Error('GUILD_UNCACHED_ME');
```

And then in the line 123, do `permissionsFor(me).has` - since `me` is a GuildMember, permissionsFor will return a non-null.

Also in the line 124, `guild.me.permissions` -> `me.permissions`.

src/structures/Message.js
```diff
@@ -150,7 +150,7 @@ class Message extends Base {
      * @type {ReactionManager}
      */
     this.reactions = new ReactionManager(this);
-    if (data.reactions && data.reactions.length > 0) {
+    if (data.reactions?.length > 0) {
```
kyranet, line 153: Do we need to check the length? for-of loops short-circuit in this case.
```suggestion
    if (data.reactions) {
```

src/structures/Webhook.js
```diff
@@ -157,9 +157,8 @@ class Webhook {
         auth: false,
       })
       .then(d => {
```
kyranet, line 159: You can refactor this to use async/await.

src/util/BitField.js
```diff
@@ -99,12 +93,13 @@ class BitField {
...
+   * @returns {Object<string, boolean>}
    */
   serialize(...hasParams) {
-    const serialized = {};
-    for (const [flag, bit] of Object.entries(this.constructor.FLAGS)) serialized[flag] = this.has(bit, ...hasParams);
-    return serialized;
+    return Object.entries(this.constructor.FLAGS).reduce((serialized, [flag, bit]) => {
+      serialized[flag] = this.has(bit, ...hasParams);
+      return serialized;
+    }, {});
```
kyranet, line 102: [`Object.fromEntries`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/fromEntries) can also be used.

src/util/Constants.js
```diff
@@ -680,7 +680,8 @@ exports.WebhookTypes = [
 ];
 
 function keyMirror(arr) {
-  let tmp = Object.create(null);
-  for (const value of arr) tmp[value] = value;
-  return tmp;
+  return arr.reduce((obj, value) => {
+    obj[value] = value;
+    return obj;
+  }, {});
```
kyranet, line 686: It was intentional to use `Object.create(null)`:
image

`Object.create(null)` creates an empty object, without inherited properties nor methods, which is what we use for enums.

```suggestion
  }, Object.create(null));
```

src/util/Util.js
```diff
@@ -236,18 +236,20 @@ class Util {
...
-      .then(data => data.shards * (1000 / guildsPerShard));
+    });
+    if (response.ok) {
+      const { shards } = await response.json();
+      return shards * (1000 / guildsPerShard);
+    } else if (response.status === 401) {
+      throw new DiscordError('TOKEN_INVALID');
+    } else {
+      throw response;
+    }
```
kyranet, line 252: No useless else-return:
```suggestion
    }

    throw response.status === 401 ? new DiscordError('TOKEN_INVALID') : response;
```

kyranet (changes requested):
There are many cases where Array#reduce was used to fill an array of the same size, Array#map should be used instead.

I also found a lot of inconsistencies from existing code (not from this PR) regarding permissions:

In most cases, we assume `permissionsFor(this.client.user)` will always return a non-null value, which it indeed will, more than 99.99% of cases (when the GUILD intent is used (which discord.js needs to operate), Discord always sends the client's member, as such, `guild.me` is always available, even without the members or the presence intents).

The only case at which it can be null, is when a user sweeps discord.js's caches to save memory, (a practice that was discouraged, but was quite common regardless) but did not account for re-adding `guild.me` back, which lack breaks most of the library.

We need to consolidate this and decide what should we do in this regard, since in regular conditions this code will always work, it's the user's interference in discord.js's cache what makes it break.

Returning (not `null`, but `undefined`) in the getters is something I don't really agree with, and should be discussed.

src/client/Client.js
vladfrangu: Yeah, we can totally remove the try/catch, due to the reasons you already specified

src/rest/DiscordAPIError.js
```diff
@@ -57,11 +55,10 @@ class DiscordAPIError extends Error {
       } else if (typeof v === 'string') {
         messages.push(v);
       } else {
-        messages = messages.concat(this.flattenErrors(v, newKey));
+        return messages.concat(this.flattenErrors(v, newKey));
```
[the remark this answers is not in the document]
NotSugden: concat is the correct usage here no? using push here would be like
```js
string[].push(string[]) // string[][]```
but a `string[]` is what we want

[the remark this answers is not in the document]
Comment (kyranet) [reply src/rest/DiscordAPIError.js:58]: Edited to make it like this:
```ts
string[].push(...string[]) // string[]
```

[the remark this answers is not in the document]
NotSugden: ah i see

src/structures/GuildChannel.js
NotSugden: i think there should be some error handling here for when the permissions aren't returned, maybe a custom error like `UNCACHED_MEMBER` idk

kyranet: Sure, it's not an error everyone will see, only those who sweep caches.

HarmoGlace:
In https://github.com/discordjs/discord.js/blob/stable/src/util/DataResolver.js#L42 you can use optional chaining

NotSugden:
> In https://github.com/discordjs/discord.js/blob/stable/src/util/DataResolver.js#L42 you can use optional chaining

its already being used there

HarmoGlace:
I don't see it in your PR 🤔

`return match?.[0] ?? data`

advaith1:
https://github.com/discordjs/discord.js/pull/5132/files#diff-919bb374deac8ecaa94a5d6f70403ba5e0287487076c056028491697f58d4606R40

src/rest/APIRequest.js
```diff
@@ -6,7 +6,9 @@ const AbortController = require('abort-controller');
 const fetch = require('node-fetch');
 const { UserAgent } = require('../util/Constants');
 
-if (https.Agent) var agent = new https.Agent({ keepAlive: true });
+let agent;
+
+if (https.Agent) agent = new https.Agent({ keepAlive: true });
```
vaporox, line 11:
```suggestion
const agent = new https.Agent({ keepAlive: true });
```
When would this class not exist?

The import could also be refactored to `const { Agent } = require('https');`

kyranet: Most likely a webpack remnant, which we removed a few months ago.

src/structures/Message.js
```diff
@@ -558,24 +553,17 @@ class Message extends Base {
...
-      .then(
-        () =>
-          this.client.actions.MessageReactionAdd.handle({
-            user: this.client.user,
-            channel: this.channel,
-            message: this,
-            emoji: Util.parseEmoji(emoji),
-          }).reaction,
-      );
+    await this.client.api.channels(this.channel.id).messages(this.id).reactions(emoji)['@me'].put();
```
vaporox, line 560:
```suggestion
    await this.client.api.channels(this.channel.id).messages(this.id).reactions(emoji, '@me').put();
```

iCrawl:
This needs a rebase.

src/client/voice/networking/VoiceUDPClient.js
```diff
@@ -90,46 +90,46 @@ class VoiceConnectionUDPClient extends EventEmitter {
...
-      if (!this.voiceConnection.sockets.ws) return;
-
-      const packet = parseLocalPacket(message);
-      if (packet.error) {
-        this.emit('debug', `[UDP] ERROR: ${packet.error}`);
-        this.emit('error', packet.error);
-        return;
-      }
+    const socket = (this.socket = udp.createSocket('udp4'))
+      .on('error', e => {
```
vladfrangu, line 94: Honestly, I'm not a fan of inlinig these with the assigning/creation of the socket... Maybe make it `socket.on().on()` afterwards

src/managers/GuildEmojiRoleManager.js
```diff
@@ -71,7 +71,10 @@ class GuildEmojiRoleManager {
   remove(roleOrRoles) {
     if (roleOrRoles instanceof Collection) return this.remove(roleOrRoles.keyArray());
     if (!Array.isArray(roleOrRoles)) return this.remove([roleOrRoles]);
-    roleOrRoles = roleOrRoles.map(r => this.guild.roles.resolveID(r));
+    roleOrRoles = roleOrRoles.map(role => {
+      const resolvedRole = this.guild.roles.resolve(role);
+      return resolvedRole && resolvedRole.id;
```
vladfrangu, line 76:
```suggestion
      return resolvedRole?.id;
```

src/managers/GuildEmojiRoleManager.js
```diff
@@ -71,7 +71,10 @@ class GuildEmojiRoleManager {
...
+    });
```
kyranet, line 77: Any reason for this change?

kyranet (changes requested):

src/sharding/Shard.js
```diff
@@ -132,15 +132,15 @@ class Shard extends EventEmitter {
...
+    if (spawnTimeout === -1 || spawnTimeout === Infinity) return this.process ?? this.worker;
     await new Promise((resolve, reject) => {
       const cleanup = () => {
         clearTimeout(spawnTimeoutTimer);
-        this.off('ready', onReady);
-        this.off('disconnect', onDisconnect);
-        this.off('death', onDeath);
+        this.removeListener('ready', onReady)
+          .removeListener('disconnect', onDisconnect)
+          .removeListener('death', onDeath);
```
vladfrangu, line 143: nit: Really no need to rename `off` from `removeListener` since they're the same, but its fine!

src/structures/APIMessage.js
```diff
@@ -93,11 +93,11 @@ class APIMessage {
 
     if (content) {
       if (isCode) {
-        const codeName = typeof this.options.code === 'string' ? this.options.code : '';
+        const codeName = this.options.code ?? '';
```
vladfrangu, line 96: Please revert this.

src/structures/APIMessage.js
```diff
@@ -148,13 +148,10 @@ class APIMessage {
...
       // eslint-disable-next-line eqeqeq
-      flags = this.options.flags != null ? new MessageFlags(this.options.flags).bitfield : this.target.flags.bitfield;
+      flags = (this.options.flags != null ? new MessageFlags(this.options.flags) : this.target.flags).bitfield;
     }
 
-    let allowedMentions =
-      typeof this.options.allowedMentions === 'undefined'
-        ? this.target.client.options.allowedMentions
-        : this.options.allowedMentions;
+    let allowedMentions = this.target.client.options.allowedMentions ?? this.options.allowedMentions;
```
vladfrangu, line 154: Invert this

src/structures/APIMessage.js
```diff
@@ -277,7 +268,7 @@ class APIMessage {
       name = findName(attachment);
     } else {
       attachment = fileLike.attachment;
-      name = fileLike.name || findName(attachment);
+      name = fileLike.name ?? findName(attachment);
```
vladfrangu, line 271: This is better off as a `||` to prevent empty strings

src/structures/Guild.js
```diff
@@ -144,22 +201,38 @@ class Guild extends Base {
...
-     * The full amount of members in this guild
-     * @type {number}
+     * Whether embedded images are enabled on this guild
+     * @type {boolean}
+     * @deprecated
      */
-    this.memberCount = data.member_count || this.memberCount;
+    this.embedEnabled = data.embed_enabled;
+
+    if (!this.emojis) {
```
vladfrangu, line 217: Why is this not always present, like channels or roles or members?

src/structures/Guild.js
```diff
@@ -892,21 +862,19 @@ class Guild extends Base {
...
-      ['audit-logs'].get({
-        query: {
-          before: options.before,
-          limit: options.limit,
-          user_id: this.client.users.resolveID(options.user),
-          action_type: options.type,
-        },
-      })
-      .then(data => GuildAuditLogs.build(this, data));
+    const data = await this.client.api.guilds(this.id)['audit-logs'].get({
```
vladfrangu, line 869:
```suggestion
    const data = await this.client.api.guilds(this.id, 'audit-logs').get({
```

src/structures/Guild.js
```diff
@@ -927,17 +895,14 @@ class Guild extends Base {
...
-        roles.push(role.id);
+    options.roles = options.roles?.map(role => {
+      const resolvedRole = this.guild.roles.resolve(role);
+      if (!resolvedRole) {
+        throw new TypeError('INVALID_TYPE', 'options.roles', 'Array or Collection of Roles or Snowflakes', true);
       }
-      options.roles = roles;
-    }
+      return resolvedRole.id;
+    }, []);
```
vladfrangu, line 904: ?

```suggestion
    });
```

src/structures/GuildChannel.js
```diff
@@ -366,12 +358,12 @@ class GuildChannel extends Channel {
 
     const newData = await this.client.api.channels(this.id).patch({
       data: {
-        name: (data.name || this.name).trim(),
-        type: ChannelTypes[data.type?.toUpperCase()],
+        name: (data.name ?? this.name).trim(),
+        type: ChannelTypes[(data.type ?? this.type).toUpperCase()],
```
vladfrangu, line 362: Reverterino this line.
```suggestion
        type: ChannelTypes[data.type?.toUpperCase()],
```

src/structures/GuildChannel.js
```diff
@@ -414,8 +406,7 @@ class GuildChannel extends Channel {
   setParent(channel, { lockPermissions = true, reason } = {}) {
     return this.edit(
       {
-        // eslint-disable-next-line no-prototype-builtins
-        parentID: channel !== null ? (channel.hasOwnProperty('id') ? channel.id : channel) : null,
+        parentID: channel !== null ? ('id' in channel ? channel.id : channel) : null,
```
vladfrangu, line 409: This is the only time I recall seeing us doing `hasOwnProperty` / `in` for this.. Don't we usually `channel.id ?? channel`?

src/structures/GuildChannel.js
vladfrangu: I agree with Kyra

src/structures/MessageMentions.js
```diff
@@ -136,12 +133,11 @@ class MessageMentions {
...
-    this._members = new Collection();
-    this.users.forEach(user => {
+    return this.users.reduce((members, user) => {
       const member = this.guild.members.resolve(user);
-      if (member) this._members.set(member.user.id, member);
-    });
-    return this._members;
+      if (member) members.set(member.id, member);
+      return members;
+    }, (this._members = new Collection()));
```
vladfrangu, line 140: I see no benefit in this... if anything making it less readable.. 

CC @iCrawl @kyranet thoughts?

src/structures/Presence.js
```diff
@@ -75,24 +75,24 @@ class Presence {
...
 
     if (data.activities) {
       /**
        * The activities of this presence
        * @type {Activity[]}
        */
       this.activities = data.activities.map(activity => new Activity(this, activity));
-    } else if (data.activity || data.game) {
-      this.activities = [new Activity(this, data.game || data.activity)];
+    } else if (data.activity ?? data.game) {
```
vladfrangu, line 94:
```suggestion
    } else if (data.game ?? data.activity) {
```

src/structures/interfaces/TextBasedChannel.js
```diff
@@ -161,9 +161,8 @@ class TextBasedChannel {
     }
 
     const { data, files } = await apiMessage.resolveFiles();
-    return this.client.api.channels[this.id].messages
-      .post({ data, files })
-      .then(d => this.client.actions.MessageCreate.handle(d).message);
+    const d = await this.client.api.channels[this.id].messages.post({ data, files });
```
vladfrangu, line 164:
```suggestion
    const d = await this.client.api.channels(this.id).messages.post({ data, files });
```

src/util/BitField.js
```diff
@@ -71,10 +71,7 @@ class BitField {
    * @returns {BitField} These bits or new BitField if the instance is frozen.
    */
   add(...bits) {
-    let total = this.constructor.defaultBit;
-    for (const bit of bits) {
-      total |= this.constructor.resolve(bit);
-    }
+    const total = bits.reduce((acc, bit) => acc | this.constructor.resolve(bit), 0);
```
vladfrangu, line 74:
```suggestion
    const total = bits.reduce((acc, bit) => acc | this.constructor.resolve(bit), this.constructor.defaultBit);
```

src/util/BitField.js
```diff
@@ -86,10 +83,7 @@ class BitField {
    * @returns {BitField} These bits or new BitField if the instance is frozen.
    */
   remove(...bits) {
-    let total = this.constructor.defaultBit;
-    for (const bit of bits) {
-      total |= this.constructor.resolve(bit);
-    }
+    const total = bits.reduce((acc, bit) => acc | this.constructor.resolve(bit), 0);
```
vladfrangu, line 86:
```suggestion
    const total = bits.reduce((acc, bit) => acc | this.constructor.resolve(bit), this.constructor.defaultBit);
```

vladfrangu (dismissed):

iCrawl:
If this is still wanted, this needs a rebase and the changes applied.

src/managers/GuildEmojiRoleManager.js
NotSugden: it was changed like this to be consistent with the `resolve` usage, there were some usages of `resolve` and `resolveID`

src/managers/GuildEmojiRoleManager.js
NotSugden: that would return `undefined | string`, `null | string` is the wanted return type here

src/structures/Guild.js
NotSugden: on the inital call its not present

src/sharding/Shard.js
kyranet: If that were so, I'd believe it was for old Node.js compatibility (`off` is rather a new feature) and/or consistency.

Since this PR changes everything, I believe it's right to change the remaining `removeListener` to `off`, for the same reason we use `on` over `addListener` (or was it `createListener`?)

src/structures/Presence.js
NotSugden: wasn't `game` removed entirely with v8?

src/structures/Guild.js
NotSugden: on the initial call its not present

src/sharding/Shard.js
NotSugden: the reason crawl gave at the time was incase it was ever removed, [see discord conversation](https://discord.com/channels/222078108977594368/222197033908436994/682992005348982811)

iCrawl: Yeah, I think we can safely use `off` by now though, it's safe to stay.

src/managers/GuildEmojiRoleManager.js
vladfrangu: Then slap a `?? null` on it :D

src/structures/MessageMentions.js
iCrawl: No big opinion on this. Can stay like this (changed) for what it's worth.

src/managers/GuildEmojiRoleManager.js
iCrawl: I think it can reasonably stay like it is now (changed).

src/client/actions/Action.js
```diff
@@ -49,14 +49,14 @@ class GenericAction {
...
+    const id = data.message_id ?? data.id;
     return (
-      data.message ||
+      data.message ??
       this.getPayload(
         {
           id,
           channel_id: channel.id,
-          guild_id: data.guild_id || (channel.guild ? channel.guild.id : null),
+          guild_id: data.guild_id ?? (channel.guild ? channel.guild.id : null),
```
DiscordCatto, line 59:
```js
guild_id: data.guild_id ?? channel.guild?.id ?? null,
```
Change it like the others.

src/managers/PresenceManager.js
```diff
@@ -40,7 +40,7 @@ class PresenceManager extends BaseManager {
     const presenceResolvable = super.resolve(presence);
     if (presenceResolvable) return presenceResolvable;
     const UserResolvable = this.client.users.resolveID(presence);
-    return super.resolve(UserResolvable) || null;
+    return super.resolve(UserResolvable) ?? null;
```
ckohen, line 43: `super.resolve` should already return null if this would evaluate to null if I'm not mistaken, why not  just remove it completely.

NotSugden:
refactored this PR to just add the usage of the nullish coercing operator and optional chaining

src/client/Client.js
```diff
@@ -38,7 +38,7 @@ class Client extends BaseClient {
     let data = process.env;
     try {
       // Test if worker threads module is present and used
-      data = require('worker_threads').workerData || data;
+      data = require('worker_threads').workerData ?? data;
```
kyranet, line 41: worker_threads is always available, you can just do `const data = workerData ?? process.env;`.

src/managers/ChannelManager.js
```diff
@@ -22,10 +22,10 @@ class ChannelManager extends BaseManager {
...
     const existing = this.cache.get(data.id);
     if (existing) {
-      if (existing._patch && cache) existing._patch(data);
-      if (guild) guild.channels?.add(existing);
-      if (ThreadChannelTypes.includes(existing.type) && typeof existing.parent?.threads !== 'undefined') {
-        existing.parent.threads.add(existing);
+      if (cache) existing._patch(data);
+      guild?.channels?.add(existing);
+      if (ThreadChannelTypes.includes(existing.type)) {
+        existing.parent?.threads.add(existing);
```
kyranet, line 28: The original code checks for the existence of `threads` too:
```suggestion
        existing.parent?.threads?.add(existing);
```

Unless, of course, that it was to check the parent. cc: @ckohen

src/structures/Message.js
```diff
@@ -133,13 +133,13 @@ class Message extends Base {
...
      */
-    this.embeds = (data.embeds || []).map(e => new Embed(e, true));
+    this.embeds = data.embeds?.map(e => new Embed(e, true)) ?? [];
 
     /**
      * A list of MessageActionRows in the message
      * @type {MessageActionRow[]}
      */
-    this.components = (data.components ?? []).map(c => BaseMessageComponent.create(c, this.client));
+    this.components = data.components.map(c => BaseMessageComponent.create(c, this.client)) ?? [];
```
kyranet, line 142:
```suggestion
    this.components = data.components?.map(c => BaseMessageComponent.create(c, this.client)) ?? [];
```

src/structures/Message.js
```diff
@@ -326,16 +322,19 @@ class Message extends Base {
       this.attachments = new Collection(this.attachments);
     }
 
+    this.embeds = data.embeds?.map(e => new Embed(e, true)) ?? this.embeds.slice();
+    this.components = data.components.map(c => BaseMessageComponent.create(c, this.client)) ?? this.components.slice();
```
kyranet, line 326:
```suggestion
    this.components = data.components?.map(c => BaseMessageComponent.create(c, this.client)) ?? this.components.slice();
```

src/structures/MessageActionRow.js
```diff
@@ -45,7 +45,7 @@ class MessageActionRow extends BaseMessageComponent {
      * The components in this action row
      * @type {MessageActionRowComponent[]}
      */
-    this.components = (data.components ?? []).map(c => BaseMessageComponent.create(c, null, true));
+    this.components = data.components.map(c => BaseMessageComponent.create(c, null, true)) ?? [];
```
kyranet, line 48:
```suggestion
    this.components = data.components?.map(c => BaseMessageComponent.create(c, null, true)) ?? [];
```

src/structures/MessageCollector.js
```diff
@@ -115,7 +115,7 @@ class MessageCollector extends Collector {
    * @returns {void}
    */
   _handleGuildDeletion(guild) {
-    if (this.channel.guild && guild.id === this.channel.guild.id) {
+    if (guild.id === this.channel.guild.id) {
```
kyranet, line 118:
```suggestion
    if (guild.id === this.channel.guild?.id) {
```

src/structures/Presence.js
```diff
@@ -75,24 +75,24 @@ class Presence {
...
+      this.activities = [new Activity(this, data.game ?? data.activity)];
```
kyranet, line 95: The access order in here isn't the same as of line above.

src/structures/Webhook.js
```diff
@@ -180,9 +180,8 @@ class Webhook {
...
-        const channel = this.client.channels ? this.client.channels.cache.get(d.channel_id) : undefined;
-        if (!channel) return d;
-        return channel.messages.add(d, false);
+        const channel = this.client.channels?.cache.get(d.channel_id);
+        return channel ? channel.messages.add(d, false) : d;
```
kyranet, line 184:
```suggestion
        return channel?.messages.add(d, false) ?? d;
```

kyranet (dismissed):

src/managers/ChannelManager.js
ckohen: Theoretically its impossible to have a parent channel that doesn't have a ThreadManager, but in case threads can ever be in other channel types in the future (before we get around to adding ThreadManagers to them) it checks.

src/structures/Message.js
```diff
@@ -326,16 +322,19 @@ class Message extends Base {
...
-      'mentions' in data ? data.mentions : this.mentions.users,
-      'mention_roles' in data ? data.mention_roles : this.mentions.roles,
-      'mention_everyone' in data ? data.mention_everyone : this.mentions.everyone,
-      'mention_channels' in data ? data.mention_channels : this.mentions.crosspostedChannels,
-      'referenced_message' in data ? data.referenced_message.author : this.mentions.repliedUser,
+      data.mentions ?? this.mentions.users,
+      data.mention_roles ?? this.mentions.roles,
+      data.mention_everyone ?? this.mentions.everyone,
+      data.mention_channels ?? this.mentions.crosspostedChannels,
+      data.referenced_message.author ?? this.mentions.repliedUser,
```
kyranet, line 334: `data.referenced_message` is nullable, according to #5929:
```suggestion
      data.referenced_message?.author ?? this.mentions.repliedUser,
```

src/structures/Presence.js
NotSugden: tbh, wasn't `activity` and `game` removed on v8? can't this block just be removed?

kyranet: cc: @vladfrangu

src/structures/Guild.js
```diff
@@ -459,7 +456,7 @@ class Guild extends AnonymousGuild {
    * @readonly
    */
   get afkChannel() {
-    return this.client.channels.cache.get(this.afkChannelID) || null;
+    return this.client.channels.cache.get(this.afkChannelID) ?? null;
```
SpaceEEC, line 459: Maybe `this.client.channels.resolve` all of these?

src/structures/GuildChannel.js
```diff
@@ -85,7 +85,7 @@ class GuildChannel extends Channel {
    * @readonly
    */
   get parent() {
-    return this.guild.channels.cache.get(this.parentID) || null;
+    return this.guild.channels.cache.get(this.parentID) ?? null;
```
SpaceEEC, line 88: Same here

src/structures/GuildMember.js
```diff
@@ -121,8 +121,7 @@ class GuildMember extends Base {
    * @readonly
    */
   get lastMessage() {
-    const channel = this.guild.channels.cache.get(this.lastMessageChannelID);
-    return (channel && channel.messages.cache.get(this.lastMessageID)) || null;
+    return this.guild.channels.cache.get(this.lastMessageChannelID)?.messages.cache.get(this.lastMessageID) ?? null;
```
SpaceEEC, line 124: Same here

src/structures/GuildTemplate.js
```diff
@@ -202,7 +202,7 @@ class GuildTemplate extends Base {
    * @readonly
    */
   get guild() {
-    return this.client.guilds.cache.get(this.guildID) || null;
+    return this.client.guilds.cache.get(this.guildID) ?? null;
```
SpaceEEC, line 205: Same here

src/structures/User.js
```diff
@@ -134,8 +134,7 @@ class User extends Base {
    * @readonly
    */
   get lastMessage() {
-    const channel = this.client.channels.cache.get(this.lastMessageChannelID);
-    return (channel && channel.messages.cache.get(this.lastMessageID)) || null;
+    return this.client.channels.cache.get(this.lastMessageChannelID)?.messages.cache.get(this.lastMessageID) ?? null;
```
SpaceEEC, line 137: This one too?

src/structures/VoiceState.js
```diff
@@ -88,7 +88,7 @@ class VoiceState extends Base {
    * @readonly
    */
   get member() {
-    return this.guild.members.cache.get(this.id) || null;
+    return this.guild.members.cache.get(this.id) ?? null;
```
SpaceEEC, line 91: Same here and below

src/structures/Webhook.js
```diff
@@ -180,9 +180,8 @@ class Webhook {
         auth: false,
       })
       .then(d => {
-        const channel = this.client.channels ? this.client.channels.cache.get(d.channel_id) : undefined;
-        if (!channel) return d;
-        return channel.messages.add(d, false);
+        const channel = this.client.channels?.cache.get(d.channel_id);
+        return channel?.messages.add(d, false) ?? d;
```
SpaceEEC, line 184: Why not combine those?

src/structures/interfaces/TextBasedChannel.js
```diff
@@ -39,7 +39,7 @@ class TextBasedChannel {
    * @readonly
    */
   get lastMessage() {
-    return this.messages.cache.get(this.lastMessageID) || null;
+    return this.messages.cache.get(this.lastMessageID) ?? null;
```
SpaceEEC, line 42: This one too?

src/structures/Presence.js
SpaceEEC: I don't seem to be receiving it anymore, it's also no longer in the [Discord Developer Documentation](https://discord.com/developers/docs/topics/gateway#presence-update).

SpaceEEC (changes requested):
Nullish coalescing operator 🙏

src/structures/Webhook.js
NotSugden: oversight

kyranet: More than an oversight, it's related to the previous code, which was `channels ? channels.messages.add(d, false) : d`. Just needed a second refactor pass 😄

iCrawl:
This needs a rebase.

src/managers/ChannelManager.js
```diff
@@ -22,10 +22,10 @@ class ChannelManager extends BaseManager {
   add(data, guild, cache = true) {
     const existing = this.cache.get(data.id);
     if (existing) {
-      if (existing._patch && cache) existing._patch(data);
-      if (guild) guild.channels?.add(existing);
-      if (ThreadChannelTypes.includes(existing.type) && typeof existing.parent?.threads !== 'undefined') {
-        existing.parent.threads.add(existing);
+      if (cache) existing._patch(data);
```
vladfrangu, line 25: This line isn't the same (tho one could ask `when is _patch undefined`)

src/structures/ClientPresence.js
```diff
@@ -31,9 +31,9 @@ class ClientPresence extends Presence {
   _parse({ status, since, afk, activities }) {
     const data = {
       activities: [],
-      afk: typeof afk === 'boolean' ? afk : false,
+      afk: afk ?? false,
```
vladfrangu, line 34: Should keep that as it was before

src/structures/GuildAuditLogs.js
```diff
@@ -356,7 +356,7 @@ class GuildAuditLogsEntry {
      * Specific property changes
      * @type {AuditLogChange[]}
```
vladfrangu, line 357: Well this jsdoc comment is wrong

src/structures/Presence.js
```diff
@@ -104,14 +98,14 @@ class Presence {
...
-    this.clientStatus = data.client_status || null;
+    this.clientStatus = data.client_status ?? null;
 
     return this;
   }
 
   _clone() {
     const clone = Object.assign(Object.create(this), this);
-    if (this.activities) clone.activities = this.activities.map(activity => activity._clone());
+    clone.activities = this.activities.map(activity => activity._clone());
```
vladfrangu, line 108:
```suggestion
    clone.activities = this.activities?.map(activity => activity._clone());
```

vladfrangu (changes requested):

src/structures/Presence.js
NotSugden: i don't think `this.activities` can ever be undefined

src/managers/ChannelManager.js
NotSugden: seeing as `Channel` extends `Base` and `Base` has an `_patch` method, never so isn't it still the same?

vladfrangu (approved):

iCrawl:
This needs a rebase.

vladfrangu (dismissed):
https://github.com/discordjs/discord.js/pull/5132#discussion_r659947226 and https://github.com/discordjs/discord.js/pull/5132#discussion_r659948414, not to mention that, and you won't believe it, this needs another rebase

NotSugden:
reeeebased

src/structures/GuildAuditLogs.js
```diff
@@ -354,9 +354,9 @@ class GuildAuditLogsEntry {
 
     /**
      * Specific property changes
-     * @type {AuditLogChange[]}
+     * @type {?(AuditLogChange[])}
```
vladfrangu, line 357: make sure typings match now (so they're also marked as nullable)

vladfrangu (changes requested):
Small thing, otherwise LGTM

vladfrangu (approved):

src/structures/Guild.js
```diff
@@ -462,7 +459,7 @@ class Guild extends AnonymousGuild {
    * @readonly
    */
   get afkChannel() {
-    return this.client.channels.cache.get(this.afkChannelID) || null;
+    return this.client.channels.resolve(this.afkChannelID) ?? null;
```
kyranet, line 462:
```suggestion
    return this.client.channels.resolve(this.afkChannelID);
```

src/structures/Guild.js
```diff
@@ -471,7 +468,7 @@ class Guild extends AnonymousGuild {
    * @readonly
    */
   get systemChannel() {
-    return this.client.channels.cache.get(this.systemChannelID) || null;
+    return this.client.channels.resolve(this.systemChannelID) ?? null;
```
kyranet, line 471:
```suggestion
    return this.client.channels.resolve(this.systemChannelID);
```

src/structures/Guild.js
```diff
@@ -480,7 +477,7 @@ class Guild extends AnonymousGuild {
    * @readonly
    */
   get widgetChannel() {
-    return this.client.channels.cache.get(this.widgetChannelID) || null;
+    return this.client.channels.resolve(this.widgetChannelID) ?? null;
```
kyranet, line 480:
```suggestion
    return this.client.channels.resolve(this.widgetChannelID);
```

src/structures/Guild.js
```diff
@@ -489,7 +486,7 @@ class Guild extends AnonymousGuild {
    * @readonly
    */
   get rulesChannel() {
-    return this.client.channels.cache.get(this.rulesChannelID) || null;
+    return this.client.channels.resolve(this.rulesChannelID) ?? null;
```
kyranet, line 489:
```suggestion
    return this.client.channels.resolve(this.rulesChannelID);
```

src/structures/Guild.js
```diff
@@ -498,7 +495,7 @@ class Guild extends AnonymousGuild {
    * @readonly
    */
   get publicUpdatesChannel() {
-    return this.client.channels.cache.get(this.publicUpdatesChannelID) || null;
+    return this.client.channels.resolve(this.publicUpdatesChannelID) ?? null;
```
kyranet, line 498:
```suggestion
    return this.client.channels.resolve(this.publicUpdatesChannelID);
```

src/managers/BaseManager.js
```diff
@@ -42,12 +42,12 @@ class BaseManager {
   }
 
   add(data, cache = true, { id, extras = [] } = {}) {
-    const existing = this.cache.get(id || data.id);
-    if (existing && existing._patch && cache) existing._patch(data);
+    const existing = this.cache.get(id ?? data.id);
+    if (cache) existing?._patch(data);
```
kyranet, line 46: Not sure if there are structures with a manager that lack of a _patch method, but...
```suggestion
    if (cache) existing?._patch?.(data);
```

src/structures/GuildChannel.js
```diff
@@ -85,7 +85,7 @@ class GuildChannel extends Channel {
    * @readonly
    */
   get parent() {
-    return this.guild.channels.cache.get(this.parentID) || null;
+    return this.guild.channels.resolve(this.parentID) ?? null;
```
kyranet, line 88:
```suggestion
    return this.guild.channels.resolve(this.parentID);
```

src/structures/GuildChannel.js
```diff
@@ -118,8 +118,8 @@ class GuildChannel extends Channel {
 
       // Compare overwrites
       return (
-        channelVal !== undefined &&
-        parentVal !== undefined &&
+        channelVal &&
+        parentVal &&
```
kyranet, line 122: With this change, the method returns `undefined | boolean`.

src/structures/GuildTemplate.js
```diff
@@ -202,7 +202,7 @@ class GuildTemplate extends Base {
    * @readonly
    */
   get guild() {
-    return this.client.guilds.cache.get(this.guildID) || null;
+    return this.client.guilds.resolve(this.guildID) ?? null;
```
kyranet, line 205:
```suggestion
    return this.client.guilds.resolve(this.guildID);
```

src/structures/Message.js
```diff
@@ -347,7 +346,7 @@ class Message extends Base {
    * @readonly
    */
   get member() {
-    return this.guild ? this.guild.members.resolve(this.author) || null : null;
+    return this.guild?.members.resolve(this.author);
```
kyranet, line 349:
```suggestion
    return this.guild?.members.resolve(this.author) ?? null;
```

src/structures/Presence.js
```diff
@@ -66,7 +66,7 @@ class Presence {
    * @readonly
    */
   get user() {
-    return this.client.users.cache.get(this.userID) || null;
+    return this.client.users.cache.get(this.userID) ?? null;
```
kyranet, line 69:
```suggestion
    return this.client.users.resolve(this.userID);
```

src/structures/Presence.js
```diff
@@ -75,27 +75,21 @@ class Presence {
    * @readonly
    */
   get member() {
-    return this.guild.members.cache.get(this.userID) || null;
+    return this.guild.members.cache.get(this.userID) ?? null;
```
kyranet, line 78:
```suggestion
    return this.guild.members.resolve(this.userID);
```

src/structures/interfaces/TextBasedChannel.js
```diff
@@ -39,7 +39,7 @@ class TextBasedChannel {
    * @readonly
    */
   get lastMessage() {
-    return this.messages.cache.get(this.lastMessageID) || null;
+    return this.messages.resolve(this.lastMessageID) ?? null;
```
kyranet, line 42:
```suggestion
    return this.messages.resolve(this.lastMessageID);
```

kyranet (changes requested):

src/managers/BaseManager.js
NotSugden: all structures extend base which has an _patch method on it so ?

src/structures/GuildAuditLogs.js
NotSugden: typings match

kyranet (approved):

iCrawl (approved):

src/structures/GuildMember.js
```diff
@@ -121,8 +121,7 @@ class GuildMember extends Base {
    * @readonly
    */
   get lastMessage() {
-    const channel = this.guild.channels.cache.get(this.lastMessageChannelID);
-    return (channel && channel.messages.cache.get(this.lastMessageID)) || null;
+    return this.guild.channels.resolve(this.lastMessageChannelID)?.messages.resolve(this.lastMessageID) ?? null;
```
SpaceEEC, line 124: That's redundant:
```suggestion
    return this.guild.channels.resolve(this.lastMessageChannelID)?.messages.resolve(this.lastMessageID);
```

src/structures/Message.js
```diff
@@ -347,7 +346,7 @@ class Message extends Base {
    * @readonly
    */
   get member() {
-    return this.guild ? this.guild.members.resolve(this.author) || null : null;
+    return this.guild?.members.resolve(this.author) ?? null;
```
SpaceEEC, line 349: That's redundant:
```suggestion
    return this.guild?.members.resolve(this.author);
```

src/structures/User.js
```diff
@@ -141,8 +141,7 @@ class User extends Base {
    * @readonly
    */
   get lastMessage() {
-    const channel = this.client.channels.cache.get(this.lastMessageChannelID);
-    return (channel && channel.messages.cache.get(this.lastMessageID)) || null;
+    return this.client.channels.resolve(this.lastMessageChannelID)?.messages.resolve(this.lastMessageID) ?? null;
```
SpaceEEC, line 144: That's redundant:
```suggestion
    return this.client.channels.resolve(this.lastMessageChannelID)?.messages.resolve(this.lastMessageID);
```

SpaceEEC (changes requested):
Three tiny nits left, otherwise LGTM.

src/structures/GuildMember.js
DTrombett: Am I wrong or with the optional chaining operator it's possible that this returns undefined?

SpaceEEC: No, you are actually correct. 😅

SpaceEEC (approved):

PR #5142, opened 2020-12-22
Support for interactionCreate event.

MatteZ02:

Add support for interactionCreate event so us small people can still use slash-commands ourselves and get an event response from djs. All this includes is addition for the event.

**Status**

- [✅] Code changes have been tested against the Discord API, or there are no code changes
- [✅] I know how to update typings and have done so, or typings don't need updating

- [✅] This PR changes the library's interface (methods or parameters added)
- [❌] This PR includes breaking changes (methods removed or renamed, parameters moved or removed)
- [❌] This PR **only** includes non-code changes, like changes to documentation, README, etc.

### DIFF
--- a/src/client/actions/ActionsManager.js
+++ b/src/client/actions/ActionsManager.js
@@ -37,6 +37,7 @@ class ActionsManager {
     this.register(require('./GuildIntegrationsUpdate'));
     this.register(require('./WebhooksUpdate'));
     this.register(require('./TypingStart'));
+    this.register(require('./InteractionCreate'))
   }
 
   register(Action) {
--- a/src/client/actions/InteractionCreate.js
+++ b/src/client/actions/InteractionCreate.js
@@ -12,6 +12,7 @@ const parseContent = options => {
 
 class InteractionCreateAction extends Action {
   async handle(data) {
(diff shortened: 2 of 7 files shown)


Extroonie:
#5106 is already made which implements the full support though?

MatteZ02:
I was just wondering as full support will be added in some version of v13. So why not just add support for the event before thats released.

Extroonie:
> I was just wondering as full support will be added in some version of v13. So why not just add support for the event before thats released.

This will still need #4879 (support for gateway v8) which is indeed major and therefore needs v13. It also does not seem like there are any releases soon other than v13 though.

MatteZ02:
> This will still need #4879 (support for gateway v8) which is indeed major and therefore needs v13. It also does not seem like there are any releases soon other than v13 though.

Interesting as i personally didn't change the gateway version on my fork and it works just fine with the event. Not sure what thats about.

zaida04:
While Interactions are reported to work on v6, v6 is deprecated.

advaith1:
neither this pr nor the full slash command pr needs v8 but v13 is gonna be the next version anyway; no need for this pr

PR #5144, opened 2020-12-22
feat(Guild): add api methods for membership screening

ckohen:
Ref: discord/discord-api-docs#2396
A similar PR is coming to do the same with the Guild Welcome Screen

**Status**

### DIFF
--- a/src/structures/Guild.js
+++ b/src/structures/Guild.js
@@ -1487,7 +1487,7 @@ class Guild extends Base {
       throw new Error('COMMUNITY');
     }
     const fields = memberScreen.formFields?.map(field => ({
-      field_type: field.field_type,
+      field_type: field.fieldType,
       label: field.label,
       values: field.values,
       required: field.required,
--- a/src/structures/Guild.js
+++ b/src/structures/Guild.js
@@ -1005,7 +1005,7 @@ class Guild extends Base {
     if (!this.features.includes('COMMUNITY')) {
       throw new Error('COMMUNITY');
     }
-    const data = await this.client.api.guilds(this.id, 'verification-level').get();
+    const data = await this.client.api.guilds(this.id, 'member-verification').get();
     let formFields = [];
     for (const field of data.form_fields) {
       formFields.push({
--- a/src/structures/Guild.js
+++ b/src/structures/Guild.js
@@ -487,15 +487,6 @@ class Guild extends Base {
   }
 
   /**
-   * Whether Membership Screening is enabled on this guild
-   * @type {boolean}
-   * @readonly
-   */
-  get membershipScreeningEnabled() {
-    return this.features.includes('MEMBER_VERIFICATION_GATE_ENABLED');
-  }
-
-  /**
    * If this guild is partnered
    * @type {boolean}
    * @readonly
@@ -993,7 +984,10 @@ class Guild extends Base {
    * @example
    * // Fetches the guild membership screening options
    * guild.fetchMembershipScreening()
-   *   .then(memberScreen => console.log(`Membership Screening is ${memberScreen.enabled ? 'enabled' : 'disabled'}`))
+   *   .then(memberScreen => {
+   *     const requiredFields = memberScreen.formFields.filter(field => field.required);
+   *     console.log(`Screening has ${memberScreen.formFields.length} steps, ${requiredFields.length} are required`);
+   *   })
    *   .catch(console.error);
    */
   async fetchMembershipScreening() {
@@ -1002,7 +996,7 @@ class Guild extends Base {
     }
     const data = await this.client.api.guilds(this.id)['member-verification'].get();
     return {
-      enabled: this.membershipScreeningEnabled,
+      enabled: this.features.includes('MEMBER_VERIFICATION_GATE_ENABLED'),
       description: data.description,
       formFields: data.form_fields.map(field => ({
         fieldType: field.field_type,
@@ -1504,7 +1498,7 @@ class Guild extends Base {
     if (typeof memberScreen.formFields !== 'undefined') data.form_fields = JSON.stringify(fields);
     const res = await this.client.api.guilds(this.id)['member-verification'].patch({ data });
     return {
-      enabled: memberScreen.enabled,
+      enabled: memberScreen.enabled ?? this.features.includes('MEMBER_VERIFICATION_GATE_ENABLED'),
       description: res.description,
       formFields: res.form_fields.map(field => ({
         fieldType: field.field_type,
--- a/src/structures/Guild.js
+++ b/src/structures/Guild.js
@@ -492,7 +492,7 @@ class Guild extends Base {
    * @readonly
    */
   get membershipScreeningEnabled() {
-    return this.features.includes('MEMBERSHIP_VERIFICATION_GATE_ENABLED');
+    return this.features.includes('MEMBER_VERIFICATION_GATE_ENABLED');
   }
 
   /**
--- a/src/structures/Guild.js
+++ b/src/structures/Guild.js
@@ -1480,7 +1480,7 @@ class Guild extends Base {
    * A `Partial` object is a representation of any existing object.
    * This object contains between 1 and all of the original objects parameters.
    * This is true regardless of whether the parameters are optional in the base object.
-   * @interface Partial
+   * @typedef {Object} Partial
    */
 
   /**
--- a/src/structures/Guild.js
+++ b/src/structures/Guild.js
@@ -1495,27 +1495,29 @@ class Guild extends Base {
       throw new Error('COMMUNITY');
     }
     let form_fields = [];
-    for (const field of memberScreen.formFields) {
-      form_fields.push({
-        field_type: field.field_type,
-        label: field.label,
-        values: field.values,
-        required: field.required,
-      });
+    if (typeof memberScreen.formFields !== 'undefined') {
+      for (const field of memberScreen.formFields) {
+        form_fields.push({
+          field_type: field.field_type,
+          label: field.label,
+          values: field.values,
+          required: field.required,
+        });
+      }
     }
+    let data = {};
+    if (typeof memberScreen.enabled !== 'undefined') data.enabled = memberScreen.enabled;
+    if (typeof memberScreen.description !== 'undefined') data.description = memberScreen.description;
+    if (typeof memberScreen.formFields !== 'undefined') data.form_fields = JSON.stringify(form_fields);
     return this.client.api
       .guilds(this.id, 'member-verification')
       .patch({
-        data: {
-          enabled: memberScreen.enabled,
-          form_fields: JSON.stringify(form_fields),
-          description: memberScreen.description,
-        },
+        data: data,
       })
-      .then(data => {
+      .then(res => {
         this.membershipScreeningEnabled = memberScreen.enabled;
         let formFields = [];
-        for (const field of data.form_fields) {
+        for (const field of res.form_fields) {
           formFields.push({
             fieldType: field.field_type,
             label: field.label,
@@ -1525,7 +1527,7 @@ class Guild extends Base {
         }
         return {
           enabled: this.membershipScreeningEnabled,
-          description: data.description,
+          description: res.description,
           formFields: formFields,
         };
       });
--- a/src/structures/Guild.js
+++ b/src/structures/Guild.js
@@ -170,15 +170,15 @@ class Guild extends Base {
      * * DISCOVERABLE
      * * FEATURABLE
      * * INVITE_SPLASH
+     * * MEMBER_VERIFICATION_GATE_ENABLED
      * * NEWS
      * * PARTNERED
+     * * PREVIEW_ENABLED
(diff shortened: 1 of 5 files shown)


src/structures/Guild.js
```diff
@@ -265,6 +267,12 @@ class Guild extends Base {
       this.embedChannelID = data.embed_channel_id;
     }
 
+    /**
+     * Whether Membership Screening is enabled on this guild
+     * @type {boolean}
+     */
+    this.membershipScreeningEnabled = this.features.includes('MEMBERSHIP_VERIFICATION_GATE_ENABLED');
```
NotSugden, line 274: this should be a getter

src/structures/Guild.js
```diff
@@ -959,6 +967,61 @@ class Guild extends Base {
       .then(data => GuildAuditLogs.build(this, data));
   }
 
+  /**
+   * Data for a field in Membership Screening
+   * @typedef {Object} GuildMembershipScreeningField
+   * @property {MembershipScreeningType} fieldType The type of the field
+   * @property {string} label The title of the field
+   * @property {string[]?} values The list of values in the field
```
NotSugden, line 975:
```suggestion
   * @property {string[]} [values] The list of values in the field
```

src/structures/Guild.js
```diff
@@ -959,6 +967,61 @@ class Guild extends Base {
...
+        fieldType: field.field_type,
+        label: field.label,
+        values: field.values,
+        required: field.required,
+      });
+    }
+    return {
+      enabled: this.membershipScreeningEnabled,
+      description: data.description,
+      formFields: formFields,
```
NotSugden, line 1021: this could just be `formFields: data.form_fields.map(...);`

src/structures/Guild.js
```diff
@@ -1422,6 +1485,54 @@ class Guild extends Base {
...
+    if (typeof memberScreen.formFields !== 'undefined') {
+      for (const field of memberScreen.formFields) {
+        form_fields.push({
+          field_type: field.field_type,
+          label: field.label,
+          values: field.values,
+          required: field.required,
+        });
+      }
+    }
```
NotSugden, line 1507: once again, this can be perhaps
```js
const fields = memberScreen.formFields?.map(...);
```

src/structures/Guild.js
```diff
@@ -1422,6 +1485,54 @@ class Guild extends Base {
...
+    let data = {};
```
NotSugden, line 1508:
```suggestion
    const data = {};
```

src/structures/Guild.js
```diff
@@ -1422,6 +1485,54 @@ class Guild extends Base {
...
+    if (typeof memberScreen.enabled !== 'undefined') data.enabled = memberScreen.enabled;
+    if (typeof memberScreen.description !== 'undefined') data.description = memberScreen.description;
+    if (typeof memberScreen.formFields !== 'undefined') data.form_fields = JSON.stringify(form_fields);
+    return this.client.api
+      .guilds(this.id, 'member-verification')
+      .patch({
+        data: data,
+      })
```
NotSugden, line 1516:
```suggestion
      .patch({ data })
```

src/structures/Guild.js
```diff
@@ -1422,6 +1485,54 @@ class Guild extends Base {
...
+      .then(res => {
```
NotSugden, line 1517: could this use async/await rather than a `.then`

src/structures/Guild.js
```diff
@@ -1422,6 +1485,54 @@ class Guild extends Base {
...
+            fieldType: field.field_type,
+            label: field.label,
+            values: field.values,
+            required: field.required,
+          });
+        }
+        return {
+          enabled: this.membershipScreeningEnabled,
+          description: res.description,
+          formFields: formFields,
```
NotSugden, line 1531: seems this code is duplicated, in any case the same as before this can use the `map` function

typings/index.d.ts
```diff
@@ -2664,6 +2670,25 @@ declare module 'discord.js' {
...
+    description: string;
+    enabled: boolean;
+    formFields: GuildMembershipScreeningField[];
+  }
+
+  interface GuildMembershipScreeningData {
+    description?: string;
+    enabled?: boolean;
+    formFields?: GuildMembershipScreeningField[];
+  }
```
NotSugden, line 2683:
```suggestion
  type GuildMembershipScreeningData = Partial<GuildMembershipScreening>;
```
either this or just directly use `Partial<GuildMembershipScreening>`

NotSugden (changes requested):

src/structures/Guild.js
ckohen: I noticed that there was a few instances of this necessary, would it be desirable to have a private function that transforms this data and call that in each location instead?

src/structures/Guild.js
```diff
@@ -1422,6 +1476,38 @@ class Guild extends Base {
...
+    const fields = memberScreen.formFields?.map(field => ({
+      field_type: field.field_type,
+      label: field.label,
+      values: field.values,
+      required: field.required,
+    }));
+    const data = {};
+    if (typeof memberScreen.enabled !== 'undefined') data.enabled = memberScreen.enabled;
+    if (typeof memberScreen.description !== 'undefined') data.description = memberScreen.description;
+    if (typeof memberScreen.formFields !== 'undefined') data.form_fields = JSON.stringify(fields);
```
NotSugden, line 1497:
```suggestion
    if (typeof memberScreen.formFields !== 'undefined') data.form_fields = fields;
```
i don't think this is meant to be stringified here

NotSugden (changes requested):

src/structures/Guild.js
ckohen: In my testing (and according to the referenced docs) it has to be stringified. This may be an accident in the API, but this currently accurately reflects what the API is expecting.

NotSugden: oops, misread docs mb

src/structures/Guild.js
```diff
@@ -177,6 +177,8 @@ class Guild extends Base {
      * * VERIFIED
      * * VIP_REGIONS
      * * WELCOME_SCREEN_ENABLED
+     * * MEMBER_VERIFICATION_GATE_ENABLED
+     * * PREVIEW_ENABLED
```
tipakA, line 181: Shouldn't these be sorted alphabetically just like the rest was? Same goes for the typings file.

ckohen: Guess I put them in the order discord docs have them in without noticing they're alphabetical here. Fixed!

src/structures/Guild.js
```diff
@@ -1422,6 +1476,38 @@ class Guild extends Base {
       .then(() => this);
   }
 
+  /**
+   * Edits the guild's membership screening form.
+   * @param {Partial<GuildMembershipScreening>} memberScreen The membership screening data for the guild
```
NotSugden, line 1481: could there be something that explains what `Partial<x>` is on the docs? 
image

NotSugden (changes requested):

almostSouji:
update: upstream  https://github.com/discord/discord-api-docs/pull/2396 has been merged and finalized

src/structures/Guild.js
```diff
@@ -484,6 +486,15 @@ class Guild extends Base {
...
 
+  /**
+   * Whether Membership Screening is enabled on this guild
+   * @type {boolean}
+   * @readonly
+   */
+  get membershipScreeningEnabled() {
+    return this.features.includes('MEMBER_VERIFICATION_GATE_ENABLED');
+  }
+
```
advaith1, line 497: this is not needed imo; .partnered and .verified were hacks on checking other features (added before the PARTNERED and VERIFIED features iirc) and should be removed too in a separate pr; `guild.membershipScreeningEnabled` is barely an improvement over `guild.features.has('MEMBER_VERIFICATION_GATE_ENABLED')`

src/util/Constants.js
```diff
@@ -512,6 +512,13 @@ exports.ExplicitContentFilterLevels = ['DISABLED', 'MEMBERS_WITHOUT_ROLES', 'ALL
  */
 exports.VerificationLevels = ['NONE', 'LOW', 'MEDIUM', 'HIGH', 'VERY_HIGH'];
 
+/**
+ * The type of membership screening. Here are the available types:
+ * * TERMS
+ * @typedef {string} MembershipScreeningType
+ */
+exports.MembershipScreeningType = ['TERMS'];
```
advaith1, line 520: can we add the name here?

maybe
```suggestion
/**
 * The type of membership screening. Here are the available types:
 * * TERMS: Server Rules
 * @typedef {string} MembershipScreeningType
 */
exports.MembershipScreeningType = ['TERMS'];
```

src/structures/Guild.js
```diff
@@ -959,6 +970,49 @@ class Guild extends Base {
...
+   * @example
+   * // Fetches the guild membership screening options
+   * guild.fetchMembershipScreening()
+   *   .then(memberScreen => console.log(`Membership Screening is ${memberScreen.enabled ? 'enabled' : 'disabled'}`))
+   *   .catch(console.error);
+   */
+  async fetchMembershipScreening() {
+    if (!this.features.includes('COMMUNITY')) {
+      throw new Error('COMMUNITY');
+    }
```
advaith1, line 1002: I don't see this in https://github.com/discordjs/discord.js/blob/master/src/errors/Messages.js, and it doesn't seem to be added in this PR?

src/util/Constants.js
ckohen: Good idea, i'll put `TERMS` in a code block though to make it a little clearer

src/structures/Guild.js
ckohen: yeah, i see your point, especially for v13 where we are trying to remove these simple wrappers, will remove

src/structures/Guild.js
```diff
@@ -959,6 +970,49 @@ class Guild extends Base {
...
+
+  /**
+   * Fetches the guild Membership Screening data.
+   * @returns {Promise<GuildMembershipScreening>}
+   * @example
+   * // Fetches the guild membership screening options
+   * guild.fetchMembershipScreening()
+   *   .then(memberScreen => console.log(`Membership Screening is ${memberScreen.enabled ? 'enabled' : 'disabled'}`))
+   *   .catch(console.error);
+   */
```
advaith1, line 998: maybe use an example that actually requires fetching it (can just check the feature for this, and users shouldn't be fetching it for this)

src/structures/Guild.js
ckohen: Good point, I hadn't realized that Error's were extended when I wrote this, am aware now, will fix in next push

iCrawl (approved):

kyranet (approved):

vladfrangu (approved):

advaith1 (approved):

src/structures/Guild.js
```diff
@@ -1422,6 +1470,45 @@ class Guild extends Base {
       .then(() => this);
   }
 
+  /**
+   * A `Partial` object is a representation of any existing object.
+   * This object contains between 1 and all of the original objects parameters.
```
SpaceEEC, line 1475: Since you actually can edit with an empty object.
```suggestion
   * This object contains between 0 and all of the original objects parameters.
```

SpaceEEC (changes requested):

src/structures/Guild.js
```diff
@@ -959,6 +961,52 @@ class Guild extends Base {
...
+   *     const requiredFields = memberScreen.formFields.filter(field => field.required);
+   *     console.log(`Screening has ${memberScreen.formFields.length} steps, ${requiredFields.length} are required`);
+   *   })
+   *   .catch(console.error);
+   */
+  async fetchMembershipScreening() {
+    if (!this.features.includes('COMMUNITY')) {
+      throw new Error('COMMUNITY');
+    }
+    const data = await this.client.api.guilds(this.id)['member-verification'].get();
```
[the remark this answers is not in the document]
ckohen: This may be a bug with how the API is currently handling this, see https://github.com/discord/discord-api-docs/issues/2530
A check is still a good idea though, will add.

ckohen:
Looks like the associated API methods for this are being changed, I'll draft this PR until we get the updated ones.
Ref: https://github.com/discord/discord-api-docs/pull/2547

iCrawl:
If this is still wanted (upstream PR has been merged) this should be undrafted and rebased.

advaith1:
@iCrawl the docs were removed pending breaking changes, and Discord hasn't given any information since then

iCrawl:
Right, I didn't check, I assumed that would be the "new" PR. Thats fine then.

src/structures/Guild.js
NotSugden, line 997:
```suggestion
    const data = await this.client.api.guilds(this.id, 'member-verification').get();
```

ckohen: I'll address this when I can resume work on this PR.

PR #5147, opened 2020-12-23
feat(Application): application flags

vaporox:

﻿Adds support for the new `application.flags` property as of discord/discord-api-docs#2295. Since this is only sent with `READY` and not with `Get Current Application`, some parts of the lib needs a bit of restructuring:

- New property `client.application` – `null` when instantiating, a partial application as soon as `READY` gets emitted
- New property `application.flags` (obviously)
- New property `application.partial`
- `client.fetchApplication()` -> `client.application.fetch()`
- `client.generateInvite()`: `Promise<string>` -> `string` (since `id` is sent with `READY` too)
- Make now nullable properties actually nullable
- A bit of refactoring

**Status**


NotSugden (changes requested):
there should probably be info/warn tags on the description of `Client#generateInvite()` and the `Client#application` property saying they can only be used/present if the client is logged in

src/client/websocket/handlers/READY.js
```diff
@@ -1,21 +1,25 @@
...
+    client.user = new ClientUser(client, data.user);
+    client.users.cache.set(client.user.id, client.user);
   }
 
   for (const guild of data.guilds) {
     guild.shardID = shard.id;
     client.guilds.add(guild);
   }
 
+  if (!client.application) {
```
NotSugden, line 20: why not `if ('application' in data) {`?

typings/index.d.ts
```diff
@@ -251,8 +251,11 @@ declare module 'discord.js' {
     public botPublic: boolean | null;
     public botRequireCodeGrant: boolean | null;
     public cover: string | null;
+    public flags: number;
     public owner: User | Team | null;
     public rpcOrigins: string[];
+    public readonly partial: boolean;
```
NotSugden, line 257: this property should be moved up so its sorted alphabetically

src/client/websocket/handlers/READY.js
Vaporox: actually, since discord only sends one ready dispatch per session, is there even any check needed?

NotSugden: i would suppose not

Vaporox: after finding https://github.com/discordjs/discord.js/commit/1bafa4b86bcedb62b2fb349dda06f0867273481b with git blame, i think it's best to just apply the same behavior here, maybe the flags change between two identify calls ¯\\\_(ツ)\_/¯

Vaporox:
> there should probably be info/warn tags on the description of `Client#generateInvite()` and the `Client#application` property saying they can only be used/present if the client is logged in

I went with an error instead of adding it to the docs since that's what most methods do

src/structures/ClientApplication.js
```diff
@@ -11,35 +11,64 @@ class ClientApplication extends Application {
...
+     * @type {number}
+     */
+    this.flags = data.flags ?? this.flags;
+
     /**
      * The app's cover image
      * @type {?string}
      */
-    this.cover = data.cover_image || null;
+    this.cover = data.cover_image ?? this.cover ?? null;
```
NotSugden, line 24: i don't really like this, it would be better to keep it consistent with the other methods and set a value only if `'cover_image' in data`, the same goes for the rest of the properties

NotSugden (changes requested):

src/structures/ClientApplication.js
Vaporox: sounds reasonable to me, how would i indicate that in jsdoc? add `{type|undefined}` everywhere or is it better to just leave a note at the class doc of `ClientApplication`

maichi77 (approved):

src/structures/ClientApplication.js
```diff
@@ -5,41 +5,69 @@ const Application = require('./interfaces/Application');
...
  */
 class ClientApplication extends Application {
   _patch(data) {
     super._patch(data);
 
+    /**
+     * The flags this application has
+     * @type {number}
+     */
+    if ('flags' in data) this.flags = data.flags;
```
NotSugden, line 20:
```suggestion
    if ('flags' in data) {
      /**
       * The flags this application has
       * @type {number}
       */
      this.flags = data.flags;
    }
```
others also need changing, this is so the docstring is parsed correctly

NotSugden (changes requested):

maichi77 (approved):

iCrawl (approved):

kyranet (approved):

typings/index.d.ts
```diff
@@ -248,11 +248,14 @@ declare module 'discord.js' {
...
-    public cover: string | null;
-    public owner: User | Team | null;
+    public botPublic?: boolean;
+    public botRequireCodeGrant?: boolean;
+    public cover?: string;
+    public flags: number;
+    public owner?: User | Team;
+    public readonly partial: boolean;
     public rpcOrigins: string[];
+    public fetch(): Promise<ClientApplication>;
```
vladfrangu, line 258: Technically, we know that fetch call will return all props.. do we wanna `Required<ClientApplication>`?

vladfrangu (changes requested):

src/structures/ClientApplication.js
```diff
@@ -5,41 +5,82 @@ const Application = require('./interfaces/Application');
...
+    if ('team' in data) {
+      /**
+       * The owner of this OAuth application
+       * @type {User|Team}
+       */
+      this.owner = new Team(this.client, data.team);
+    } else if ('owner' in data) {
+      this.owner = this.client.users.add(data.owner);
+    }
+  }
```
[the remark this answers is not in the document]
vaporox: Was nullable before, decided to make them optional instead to be consistent with other partial structures (see https://github.com/discordjs/discord.js/pull/5147#discussion_r548765022)
I decided to add a notice at the top of the `ClientApplication` class instead of adding it to the jsdoc type so that it doesn't show undefined for integration applications, but I can change that if it's preferred

src/structures/ClientApplication.js
```diff
@@ -5,41 +5,82 @@ const Application = require('./interfaces/Application');
...
+
+    if ('bot_public' in data) {
+      /**
+       * If this app's bot is public
+       * @type {boolean}
+       */
+      this.botPublic = data.bot_public;
+    }
+
+    if ('team' in data) {
```
SpaceEEC, line 56: Discord sends `team: null`, if the application is not owned by a team.
This is causing the Team constructor (or rather Team's _patch method) to throw an error.

```suggestion
    // Discord explicitly sends null for this field, if not owned by a team.
    if (data.team) {
```
Or something like this checking for truthiness.

SpaceEEC (changes requested):

src/structures/ClientApplication.js
```diff
@@ -5,41 +5,83 @@ const Application = require('./interfaces/Application');
 
 /**
  * Represents a Client OAuth2 Application.
+ * <info>The only guaranteed properties are `id` and `flags`. To get the entire
+ * application, you need to fetch it with `application.fetch()` first.</info>
```
SpaceEEC, line 9:
```suggestion
 * application, you need to fetch it with {@link ClientApplication#fetch} first.</info>
```

src/structures/interfaces/Application.js
```diff
@@ -23,23 +23,29 @@ class Application extends Base {
...
-
-    /**
-     * The app's icon hash
-     * @type {string}
-     */
-    this.icon = data.icon;
+    if ('name' in data) {
+      /**
+       * The name of the app
+       * @type {string}
```
vladfrangu, line 29: All of these are now `?string`

vladfrangu (changes requested):

src/structures/interfaces/Application.js
vaporox: Doesn't `?type` only mean `type|null`? In this case the properties are optional, not nullable

SpaceEEC (approved):

advaith1:
`flags` should be a BitField but unfortunately, the flags themselves haven't been documented yet and most of them aren't known

vladfrangu:
> `flags` should be a BitField but unfortunately, the flags themselves haven't been documented yet and most of them aren't known

Yeah, and I don't know if they should be documented - I'll ask later. But, flag wise, the main ones I know off are related to the intents switches

advaith1:
bitfield flags:

MANAGED_EMOJI = 4,
GROUP_DM_CREATE = 16,
RPC_HAS_CONNECTED = 2048,
GATEWAY_PRESENCE = 4096,
GATEWAY_PRESENCE_LIMITED = 8192,
GATEWAY_GUILD_MEMBERS = 16384,
GATEWAY_GUILD_MEMBERS_LIMITED = 32768,
VERIFICATION_PENDING_GUILD_LIMIT = 65536,
EMBEDDED = 131072

vaporox:
where are those taken from? :eyes:

advaith1:
the client

src/structures/ClientApplication.js
```diff
@@ -2,44 +2,87 @@
...
-
-    /**
-     * The owner of this OAuth application
-     * @type {?User|Team}
-     */
-    this.owner = data.team ? new Team(this.client, data.team) : data.owner ? this.client.users.add(data.owner) : null;
+    if ('flags' in data) {
+      /**
+       * The flags this application has
+       * @type {ApplicationFlags}
```
vladfrangu, line 20: Aren't almost all of these... possibly `?` now

vladfrangu (changes requested):

src/structures/ClientApplication.js
```diff
@@ -11,35 +12,64 @@ class ClientApplication extends Application {
   _patch(data) {
     super._patch(data);
 
+    /**
+     * The flags this application has
+     * @type {ApplicationFlags}
+     */
+    this.flags = this.flags ?? new ApplicationFlags(data.flags);
```
vladfrangu, line 19: This..won't patch flags if they're changed..

src/structures/ClientApplication.js
```diff
@@ -11,35 +12,64 @@ class ClientApplication extends Application {
...
+
     /**
      * The app's cover image
      * @type {?string}
      */
-    this.cover = data.cover_image || null;
+    this.cover = data.cover_image ?? this.cover ?? null;
```
vladfrangu, line 25: This won't update it if the cover is removed

vladfrangu (changes requested):

src/structures/ClientApplication.js
vaporox: There's no way to differenciate between an application being partial or just not having a cover image

vladfrangu: `Reflect.has(data, 'cover_image') ? data.cover_image : this.cover ?? null`

vaporox: But that would set it to null when receiving a partial application

vladfrangu: > But that would set it to null when receiving a partial application

No.. it'd set it to this.cover or null... Aka previous value or null if non existent

vaporox: Oh yea my bad, but then there's no difference to the current approach because the property is optional, not nullable ([docs](https://discord.com/developers/docs/topics/oauth2#application-object))

vladfrangu: There is a difference. If I remove my cover image, d.js will still keep it instead of nulling it out...

vaporox: Same thing as with your suggestion. `data.cover_image` is virtually the same as `Reflect.has(data, 'cover_image')` because again, the property is not nullable

vladfrangu: Pretty sure applications *can* not have cover images. Reflect.has checks if the property is present, no matter its value. ?? triggers for null or undefined, of which the cover image can be...correct me if I'm wrong tho

vladfrangu (approved):

PR #5150, opened 2020-12-25
Add type definition for BaseClient#api

Flam3rboy:
Even if BaseClient#api is private, some libraries use it and it simplifies the use of the RouteBuilder proxy for the discord.js devs.
This PR adds type definition for BaseClient#api and has been tested.
Before that ``object`` was specified for BaseClient#api, which could lead to tsc errors. 

**Status**

### DIFF
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -157,7 +157,7 @@ declare module 'discord.js' {
 
   type RouteBuilderReturn = RouteBuilder &
     RouteBuilderMethods & {
-      (...args: any[]): RouteBuilderReturn;
+      (...args: string[]): RouteBuilderReturn;
     };
 
   interface RouteBuilder {
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -122,7 +122,7 @@ declare module 'discord.js' {
     private _timeouts: Set<NodeJS.Timeout>;
     private _intervals: Set<NodeJS.Timeout>;
     private _immediates: Set<NodeJS.Immediate>;
-    private readonly api: object;
+    private readonly api: RouteBuilder;
     private rest: object;
     private decrementMaxListeners(): void;
     private incrementMaxListeners(): void;
@@ -137,6 +137,34 @@ declare module 'discord.js' {
     public setImmediate(fn: (...args: any[]) => void, ...args: any[]): NodeJS.Immediate;
     public toJSON(...props: { [key: string]: boolean | string }[]): object;
   }
+  
+	interface RouteBuilderMethodOptions {
+		query?: any;
+		versioned?: boolean;
+		auth?: boolean;
+		reason?: string;
+		headers?: {
+			[key: string]: string;
+		};
+		data: any;
+	}
+
+	interface RouteBuilderMethods {
+		get(opts?: RouteBuilderMethodOptions): Promise<any>;
+		post(opts?: RouteBuilderMethodOptions): Promise<any>;
+		delete(opts?: RouteBuilderMethodOptions): Promise<any>;
+		patch(opts?: RouteBuilderMethodOptions): Promise<any>;
+		put(opts?: RouteBuilderMethodOptions): Promise<any>;
+	}
+
+	type RouteBuilderReturn = RouteBuilder &
+		RouteBuilderMethods & {
+			(...args: any[]): RouteBuilderReturn;
+		};
+
+	interface RouteBuilder {
+		[key: string]: RouteBuilderReturn;
+	}
 
   export class BaseGuildEmoji extends Emoji {
     constructor(client: Client, data: object, guild: Guild);
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -143,9 +143,7 @@ declare module 'discord.js' {
     versioned?: boolean;
     auth?: boolean;
     reason?: string;
-    headers?: {
-      [key: string]: string;
-    };
+    headers?: Record<string, string>;
     data: any;
   }
 
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -160,9 +160,7 @@ declare module 'discord.js' {
       (...args: string[]): RouteBuilderReturn;
     };
 
-  interface RouteBuilder {
-    [key: string]: RouteBuilderReturn;
-  }
+  type RouteBuilder = Record<string, RouteBuilderReturn>;
 
   export class BaseGuildEmoji extends Emoji {
     constructor(client: Client, data: object, guild: Guild);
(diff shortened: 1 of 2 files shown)


Vaporox:
since `api` is a private field, you need to `@ts-ignore` it anyway, no? so i don't think the type should give you problems here

and even then, because it's private other libraries shouldn't be using it in the first place

Flam3rboy:
actually you don't need ``// @ts-ignore`` for every .api use, if you change the typing like so:
```ts
declare module "discord.js" {
	// @ts-ignore
	interface Client {
	    api: RouteBuilder;
	}
```

But nevertheless it makes it easier for discord.js maintainers.

Vaporox:
how would it make the maintenance of a js library easier?

Flam3rboy:
That they have suggestions for the HTTP methods with corresponding parameters.

Vaporox:
then you need to update the jsdoc, not typings

Flam3rboy:
oh forgot those ones 👍

typings/index.d.ts
```diff
@@ -138,6 +138,34 @@ declare module 'discord.js' {
...
   }
 
+  interface RouteBuilderMethodOptions {
+    query?: any;
+    versioned?: boolean;
+    auth?: boolean;
+    reason?: string;
+    headers?: {
+      [key: string]: string;
+    };
```
NotSugden, line 148:
```suggestion
    headers?: Record<string, string>;
```

typings/index.d.ts
```diff
@@ -138,6 +138,34 @@ declare module 'discord.js' {
     public toJSON(...props: { [key: string]: boolean | string }[]): object;
   }
 
+  interface RouteBuilderMethodOptions {
+    query?: any;
```
NotSugden, line 142: could there be a more specific type than `any`, afaik this accepts `URLSSearchParams` or a `Record<string, string | string[]>`

typings/index.d.ts
```diff
@@ -138,6 +138,34 @@ declare module 'discord.js' {
...
+    versioned?: boolean;
+    auth?: boolean;
+    reason?: string;
+    headers?: {
+      [key: string]: string;
+    };
+    data: any;
```
NotSugden, line 149:
```suggestion
    data: Record<string, unknown>;
```

typings/index.d.ts
```diff
@@ -138,6 +138,34 @@ declare module 'discord.js' {
...
+  }
+
+  interface RouteBuilderMethods {
+    get(opts?: RouteBuilderMethodOptions): Promise<any>;
+    post(opts?: RouteBuilderMethodOptions): Promise<any>;
+    delete(opts?: RouteBuilderMethodOptions): Promise<any>;
+    patch(opts?: RouteBuilderMethodOptions): Promise<any>;
+    put(opts?: RouteBuilderMethodOptions): Promise<any>;
```
NotSugden, line 157: maybe a type parameter for the return type?
```ts
method<T = never>(opts: ...): Promise<T>;```

File: typings/index.d.ts [hunk 5]
```diff
@@ -138,6 +138,34 @@ declare module 'discord.js' {
...
+  }
+
+  type RouteBuilderReturn = RouteBuilder &
+    RouteBuilderMethods & {
+      (...args: any[]): RouteBuilderReturn;
```
NotSugden, line 162:
```suggestion
      (...args: string[]): RouteBuilderReturn;
```

typings/index.d.ts
```diff
@@ -138,6 +138,34 @@ declare module 'discord.js' {
...
+    };
+
+  interface RouteBuilder {
+    [key: string]: RouteBuilderReturn;
+  }
```
NotSugden, line 167:
```suggestion
  type RouteBuilder = Record<string, RouteBuilderReturn>;
```

NotSugden (changes requested):

typings/index.d.ts
Flam3rboy: but the request can return any data so 🤷‍♂️

typings/index.d.ts
Flam3rboy: fails with:
> Error: typings/index.d.ts(158,8): error TS2456: Type alias 'RouteBuilderReturn' circularly references itself.

typings/index.d.ts
NotSugden: thats the point, you can specify it as a type parameter
```ts
const data = await client.api.some.path.get<User>();
```

although i guess
```ts
const data: User = await client.api.some.path.get();
```
would work the same

iCrawl: The latter is def a more favorable pattern.

Extroonie:
While this can still be a workaround for the time being, but I'd however suggest this as a milestone for v14 which is indeed a TS rewrite and a whole reorganisation of the current structure. We can (probably) find a more better way to improve the REST support.

kyranet:
> While this can still be a workaround for the time being, but I'd however suggest this as a milestone for v14 which is indeed a TS rewrite and a whole reorganisation of the current structure. We can (probably) find a more better way to improve the REST support.

I don't think discord.js v14 will not use a proxy, instead, it'll use a routes namespace with methods that format routes and said data is passed directly to the REST handler, so to some point, `Client#api` behave very differently past that time, as this would make it take a raw URI path as opposed to building that path with a proxy.

cc: @vladfrangu

typings/index.d.ts
kyranet: TypeScript encourages the latter's pattern - returning `unknown` (here that'd be `Promise<unknown>`), and then leaving the user to cast it.

```typescript
const data = await client.api.some.path.get() as User;
```

advaith1:
Since I often use client.api (and typescript), this would be useful but it would be even better if it wasn't marked as private in the typings

vladfrangu:
> Since I often use client.api (and typescript), this would be useful but it would be even better if it wasn't marked as private in the typings

`client.api` is really not a public API tho, and it should remain marked as private.. If you really want to use it and bypass TS bitching, either `@ts-ignore` it or use object access notations `client['api']` (don't forget that this assumes you know what you're doing). In the future, we hope to not even require you access this raw, but provide functions for most things that won't require full structures, but this, again, remains to be seen

---

> > While this can still be a workaround for the time being, but I'd however suggest this as a milestone for v14 which is indeed a TS rewrite and a whole reorganisation of the current structure. We can (probably) find a more better way to improve the REST support.
> 
> I don't think discord.js v14 will use a proxy, instead, it'll use a routes namespace with methods that format routes and said data is passed directly to the REST handler, so to some point, `Client#api` behave very differently past that time (or will probably not longer exist, superseded/replaced by `Client#rest` since we're also going modular), as this would make it take a raw URI path as opposed to building that path with a proxy.
> 
> cc: @vladfrangu

You're correct, in the future, we'll move to the `Routes` object (provided by [discord-api-types](https://github.com/discordjs/discord-api-types)), and technically, so can you. Remains to be seen tho. Also, not really fair to call it `v14`, more so `-next`

PR #5154, opened 2020-12-27
chore(ClientOptions): reduce default edit history to 10

MattIPv4:

As we're now in breaking/major change territory with v13, I'd like to propose we reduce the default limit to a message's edit history to 10 to avoid unexpected and potentially rather confusing memory usage for some users. When we added this feature originally, it was added as non-major so the default limit was set to be unlimited, a no-op for existing users, but now we can update it.

**Status**


kyranet:
**Hot take**: we could remove `Message#_edits` (and `Message#edits`) altogether. It doesn't make any sense to store edits of a Discord structure when we don't for any other structure - just like we did with `timeout` in `Message#delete` with #4999 ;)

PR #5157, opened 2020-12-27
Improve typesafety of Snowflake

pizzafox:

Since `Snowflake`s are always ` number in string form, this updates the type to use a [template literal type](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-1.html#template-literal-types), which will grant TypeScript users safe usage of the type.

**Status**

### DIFF
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -3128,7 +3128,7 @@ declare module 'discord.js' {
 
   type ShardingManagerMode = 'process' | 'worker';
 
-  type Snowflake = string;
+  type Snowflake = `${number}`;
 
   interface SplitOptions {
     maxLength?: number;


Vaporox (changes requested):
the `typescript` dependency needs to be bumped to 4.1

kyranet (approved):
That... works... I wasn't aware `${number}` was a thing:

image

Only thing though, it allows decimals, but it's still better than nothing. Playground [here](https://www.typescriptlang.org/play?#code/C4TwDgpgBAygdgewO4DMA2BDA1tAvFAAwBIBvOAVwFsAjCAJwF8CBuAKFYGME4BnYKDAC5YiVJhxR8AcgxS2XXv2rD4ydNjxQpARgAMAJgNzO3PlA4rR6idL2H9AOgAs+uUA).

**Edit:** Playground for [`${bigint}`](https://www.typescriptlang.org/play?#code/C4TwDgpgBAygdgewO4DMA2BDA1tAvFAAwBIBvAIwEsBzCuYAXwIG4AoFgYwTgGdgoMAXLESpMOKPgDkGSa048+ZIfGTpseKJICMABgBM+2Ry68o7ZSLXipug3oB0AFj2ygA) for 20cbac9, this solved decimals inside numbers, and seems to forbid the `n` suffix (which BigInts use).

Once CI's errors are solved, this is good to go.

pizzafox:
Fixed by doing `${bigint}`!

typings/index.d.ts
```diff
@@ -1216,7 +1216,7 @@ declare module 'discord.js' {
     public equals(presence: Presence): boolean;
   }
 
-  export class ReactionCollector extends Collector<Snowflake, MessageReaction> {
+  export class ReactionCollector extends Collector<Snowflake| string, MessageReaction> {
```
pizzafox, line 1219: I am totally not sure if this is the right move here, can someone with more knowledge verify?

Vaporox (approved):

typings/index.d.ts
kyranet: This is correct, `Collector<K, V>` defines those two methods:
https://github.com/discordjs/discord.js/blob/6a7745353207290303c1fd62967f7ee6c67c754a/typings/index.d.ts#L301-L302

Which `ReactionCollector` overrides, returning `Snowflake | string`:
https://github.com/discordjs/discord.js/blob/6a7745353207290303c1fd62967f7ee6c67c754a/typings/index.d.ts#L1232-L1233

The patch is correct, because ReactionCollector handles both snowflakes (guild emojis, aka `'773916319895781427'` for example) and emojis (twemojis/unicode, aka `'👍'`). The type here is right as the former is a `Snowflake` and the latter a `string`.

Fyko (approved):

NotSugden (changes requested):
https://github.com/pizzafox/discord.js/blob/cdb6646c9da782461d09a4823924a037aea9f38c/typings/index.d.ts#L2853
this type and possibly others need to be updated to `string | Snowflake`

a MessageReactionResolvable can also be `name:id` which does not match \`${bigint}\`

SpaceEEC (approved):

catcosmicice (approved):

Issue #5139, opened 2020-12-21
Client#userUpdate does not emit despite proper partials enabled


WOZARDLOZARD:
**Please describe the problem you are having in as much detail as possible:**

The userUpdate event for the `Client` does not emit when a username is updated, even though I have enabled all partials in case the user is uncached and all intents except `GUILD_PRESENCES`. 

**Include a reproducible code sample here, if possible:**

```js
const { Client } = require('discord.js');

const client = new Client({
    partials: ["MESSAGE", "CHANNEL", "REACTION", "GUILD_MEMBER", "USER"],
    ws: { intents: ["DIRECT_MESSAGES", "DIRECT_MESSAGE_REACTIONS", "GUILDS", "GUILD_BANS", "GUILD_EMOJIS", 
    "GUILD_INVITES", "GUILD_MEMBERS", "GUILD_MESSAGES", "GUILD_MESSAGE_REACTIONS", 
    "GUILD_VOICE_STATES"] },
});

client.on('userUpdate', (oldUser, newUser) => {
    console.log("Received " + newUser.id);
    // Nothing gets logged when I change my username
});
```

**Further details:**

- discord.js version: 12.5.1
- Node.js version: 14.15.2
- Operating system: MacOS Big Sur 11.1
- Priority this issue should have – please be realistic and elaborate if possible: Medium. While I managed to work around this by using the `guildMemberUpdate` event instead, some people might not have the `GUILD_MEMBERS` privileged intent enabled and thus requires this event to receive the necessary information. 

**Relevant client options:**

- partials: MESSAGE, CHANNEL, REACTION, GUILD_MEMBER, USER
- gateway intents: All intents except GUILD_PRESENCES
- other: none

- [ ] I have also tested the issue on latest master, commit hash:

ckohen:
The gateway `USER_UPDATE` event only fires for the client user. As indicated by the [docs](https://discord.js.org/?source=post_page---------------------------#/docs/main/master/class/Client?scrollTo=e-userUpdate), discord.js fires the `userUpdate` event when `GUILD_MEMBER_UPDATE` is received as well, however, this only occurs if the user is already cached.

[...]

Issue #4800, opened 2020-09-06
guildMemberUpdate new member and old member are the same object


Linker-123:
I was making audit logs system for my bot and encountered a problem the **oldMember.user** object is the same is **newMember.user** object.
i tried doing `console.log(oldMember.user.username + '|' + newMember.user.username)` and the results both were the new username of the user!

NotSugden:
i don't think `guildMemberUpdate` fires for `username` changes, are you perhaps looking for [`GuildMember#nickname`](https://discord.js.org/#/docs/main/stable/class/GuildMember?scrollTo=nickname)?

[`GuildMember#User`](https://discord.js.org/#/docs/main/stable/class/GuildMember?scrollTo=user) is a `User` object, which is different from a `GuildMember` object, it represents the user as a whole, not just for one specific guild

Crazyshooting:
> ~i don't think `guildMemberUpdate` fires for `username` changes~ it does fire, but still, are you perhaps looking for [GuildMember#nickname](https://discord.js.org/#/docs/main/stable/class/GuildMember?scrollTo=nickname)?
> 
> [GuildMember#User](https://discord.js.org/#/docs/main/stable/class/GuildMember?scrollTo=user) is a `User` object, which is different from a `GuildMember` object, it represents the user as a whole, not just for one specific guild

no i was looking for the username change

NotSugden:
ah, seem to have found the issue
the GUILD_MEMBER_UPDATE handler https://github.com/discordjs/discord.js/blob/dfd63bdb6b73c39bd071689561f42dce3827fc14/src/client/websocket/handlers/GUILD_MEMBER_UPDATE.js#L16) calls `GuildMember#_update` which calls the `_patch` method, and updates the user on the member https://github.com/discordjs/discord.js/blob/dfd63bdb6b73c39bd071689561f42dce3827fc14/src/structures/GuildMember.js#L79

almostSouji:
I have stumbled across this issue before but forgot to make a trackable issue for it, thanks for bringing it up again! I will try to shortly summarize and explain what i found out:

### Upfront conclusion

Discord sends both PRESENCE_UPDATE as well as GUILD_MEMBER_UPDATE per guild the bot shares with the member if the underlying user information is changed.

- On the PU we determine the change and apply it to the cache
- On the GMU the underlying user is already changed from the PU before and we can not emit the old user with the old member

[...]

PR #5161, opened 2020-12-29
chore(Client): cleanup user update

ckohen:
This removes internal handler calls to `Client#userUpdate` to align the event with the way the API emits it.
This is especially important now as the behavior of this event could drastically change depending on your intents.
It also did not respect the `USER` partial because `ClientUser` is always expected to be cached and this is what the event was built around.

Notes:
- `guildMemberUpdate` and `presenceUpdate` may seem to fire with no changes, this is due to a user update, this behavior has not changed
- A consideration was placed on making the user object represent the change, however, given all guilds that the bot shares with the member being updated will emit this event, it would become inconsistent at scale.

close: #5139 

**Status**


src/client/actions/GuildMemberUpdate.js
```diff
@@ -23,6 +15,7 @@ class GuildMemberUpdateAction extends Action {
         /**
          * Emitted whenever a guild member changes - i.e. new role, removed role, nickname.
          * Also emitted when the user's details (e.g. username) change.
+         * Even when emitted from a user detail change, the user object will be identical.
```
NotSugden, line 18: could there not be some logic to have these not identical?

ckohen: > A consideration was placed on making the user object represent the change, however, given all guilds that the bot shares with the member being updated will emit this event, it would become inconsistent at scale.

There could be logic, however, this would only work for the first guildMemberUpdate / presenceUpdate for any given change to a user. Thus a majority of events would still end up having this issue and it becomes inconsistent.

iCrawl:
Needs conflicts resolved.

SpaceEEC:
With this PR it would be impossible to detect what and that a user changed, or am I misunderstanding this?

ckohen:
You would be able to detect that a user changed, just not what has changed about said user.

However, given that (and this is why it took me so long to respond), I have been thinking hard about a way to make presenceUpdate and guildMemberUpdate reference the old user.

I would propose the following solution for that:
* create a secondary UserManager as `client#_oldUsers`
* store the old user in this cache before updating, this would occur in `GuildMember#_clone` or `PresenceUpdate`
* modify `GuildMember#_clone` to set the cloned `GuildMember#user` to the oldUser
* add `Presence#_isOld` that is checked when using the `Presence#user` getter
* automatically remove users from `client#_oldUsers` after 2 minutes (perhaps this is a client option)
  * it is important to note that any further user updates that would occur before this removal would be classified under a single user update. E.g. two username updates within these 2 minutes would emit twice, but the old `GuildMember#user` object would be identical. 

I think this is a better compromise than the current proposed changes while still aligning better with discords actual event structure. Thoughts?

ckohen:
Ok, reading up on #4800 gave me some insight, would it be desiraable to keep emitting userUpdate and simply guard against emitting GMU and PU when no other parameters have changed?

ckohen:
Will open a new PR with the changes in the last comment per discord.

PR #5170, opened 2020-12-31
fix(GuildChannel): check for community required channels in GuildChannel#deletable

vBread:
Currently, `GuildChannel#deletable` will return true regardless if the channel is required for a community (rules, public updates) and will throw a `50074: Cannot delete a channel required for Community guilds`. This PR adds an inequality check between the `GuildChannel#id` and the `Guild#rulesChannelID` or `Guild#publicUpdatesChannelID`

--

**Status**

### DIFF
--- a/src/structures/GuildChannel.js
+++ b/src/structures/GuildChannel.js
@@ -572,8 +572,8 @@ class GuildChannel extends Channel {
    */
   get deletable() {
     return (
-        this.permissionsFor(this.client.user).has(Permissions.FLAGS.MANAGE_CHANNELS, false) && 
-        (this.guild.rulesChannelID !== this.id || this.guild.systemChannelID !== this.id)
+      this.permissionsFor(this.client.user).has(Permissions.FLAGS.MANAGE_CHANNELS, false) && 
+      (this.guild.rulesChannelID !== this.id || this.guild.publicUpdatesChannelID !== this.id)
     );
   }
 
--- a/src/structures/GuildChannel.js
+++ b/src/structures/GuildChannel.js
@@ -572,7 +572,7 @@ class GuildChannel extends Channel {
    */
   get deletable() {
     return (
-      this.permissionsFor(this.client.user).has(Permissions.FLAGS.MANAGE_CHANNELS, false) && 
+      this.permissionsFor(this.client.user).has(Permissions.FLAGS.MANAGE_CHANNELS, false) &&
       (this.guild.rulesChannelID !== this.id || this.guild.publicUpdatesChannelID !== this.id)
     );
   }
--- a/src/structures/GuildChannel.js
+++ b/src/structures/GuildChannel.js
@@ -573,7 +573,8 @@ class GuildChannel extends Channel {
   get deletable() {
     return (
       this.permissionsFor(this.client.user).has(Permissions.FLAGS.MANAGE_CHANNELS, false) &&
-      (this.guild.rulesChannelID !== this.id || this.guild.publicUpdatesChannelID !== this.id)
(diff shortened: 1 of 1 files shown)


src/structures/GuildChannel.js
```diff
@@ -571,7 +571,10 @@ class GuildChannel extends Channel {
    * @readonly
    */
   get deletable() {
-    return this.permissionsFor(this.client.user).has(Permissions.FLAGS.MANAGE_CHANNELS, false);
+    return (
+        this.permissionsFor(this.client.user).has(Permissions.FLAGS.MANAGE_CHANNELS, false) && 
+        (this.guild.rulesChannelID !== this.id || this.guild.systemChannelID !== this.id)
```
NotSugden, line 576: the guild system channel is deletable, its only the rules channel that you can't delete

NotSugden (changes requested):
style changes also needed

src/structures/GuildChannel.js
NotSugden: the `publicUpdatesChannelID` also can't be deleted

src/structures/GuildChannel.js
```diff
@@ -571,7 +571,10 @@ class GuildChannel extends Channel {
    * @readonly
    */
   get deletable() {
-    return this.permissionsFor(this.client.user).has(Permissions.FLAGS.MANAGE_CHANNELS, false);
+    return (
+      this.permissionsFor(this.client.user).has(Permissions.FLAGS.MANAGE_CHANNELS, false) &&
+      (this.guild.rulesChannelID !== this.id || this.guild.publicUpdatesChannelID !== this.id)
```
Mesteery, line 576:
```suggestion
      ![this.guild.rulesChannelID, this.guild.publicUpdatesChannelID].includes(this.id))
```

Mesteery (changes requested):

src/structures/GuildChannel.js
Mesteery: Why this reaction, @NotSugden ?

kyranet: It's not exactly more readable per-se, array includes reads better when we want to check more than 2 values. Nevertheless, I don't think we do checks this way in this repository, besides, `Array#includes` has the disadvantage that it doesn't short-circuit, nor it does look consistent when followed with other conditions (`&&`, `||`, or `??`).

Mesteery: It's in case there are more values, after all, it's possible.
> Array#includes has the disadvantage that it doesn't short-circuit, nor it does look consistent when followed with other conditions (&&, ||, or ??).
Yes, i know.

vBread: > It's in case there are more values, after all, it's possible.

There aren't, and if more are added, we'll refactor then. There's no point now, though, because we know there are only 2.

SpaceEEC, line 576: If the channel to delete is _either_ the rules channel or the public updates channel is enough.
It doesn't need to be _both_ 👀 
```suggestion
      this.guild.rulesChannelID !== this.id &&
      this.guild.publicUpdatesChannelID !== this.id
```

SpaceEEC (changes requested):

src/structures/GuildChannel.js
vBread: oops, oversight

vladfrangu (approved):

kyranet (approved):

SpaceEEC (approved):

iCrawl (approved):

PR #5177, opened 2021-01-02
Change messageReactionAdd and messageReactionRemove to return member instead of user

ImRodry:
**Please describe the changes this PR makes and why it should be merged:** I changed the "user" variable in both of these events to "member" so that people can easily check for the member's roles/other data inside that same guild

**Status**

  - [X] This PR includes breaking changes (methods removed or renamed, parameters moved or removed)

### DIFF
--- a/src/client/actions/MessageReactionAdd.js
+++ b/src/client/actions/MessageReactionAdd.js
@@ -20,8 +20,13 @@ class MessageReactionAdd extends Action {
 
     const client = this.client;
     const guild = client.guilds.cache.get(data.guild_id);
-    const member = this.getMember({ user: data.member.user }, guild);
+    let member = null
+    if (guild) member = this.getMember({ user: data.member.user }, guild);
+    else member = this.getUserFromMember(data)
     if (!member) return false;
+    let me = false
+    if (member.user) me = member.user.id === this.client.user.id
+    else me = member.id === this.client.user.id
 
     // Verify channel
     const channel = this.getChannel(data);
@@ -38,7 +43,7 @@ class MessageReactionAdd extends Action {
     const reaction = message.reactions.add({
       emoji: data.emoji,
       count: message.partial ? null : 0,
-      me: member.user.id === this.client.user.id,
+      me: me,
     });
     if (!reaction) return false;
     reaction._add(member);
@@ -46,7 +51,7 @@ class MessageReactionAdd extends Action {
      * Emitted whenever a reaction is added to a cached message.
      * @event Client#messageReactionAdd
      * @param {MessageReaction} messageReaction The reaction object
-     * @param {Member} member The guild member that applied the guild or reaction emoji
+     * @param {GuildMember} member The guild member that applied the guild or reaction emoji
      */
     this.client.emit(Events.MESSAGE_REACTION_ADD, reaction, member);
 
--- a/src/client/actions/MessageReactionAdd.js
+++ b/src/client/actions/MessageReactionAdd.js
@@ -20,13 +20,13 @@ class MessageReactionAdd extends Action {
 
     const client = this.client;
     const guild = client.guilds.cache.get(data.guild_id);
-    let member = null
+    let member = null;
     if (guild) member = this.getMember({ user: data.member.user }, guild);
-    else member = this.getUserFromMember(data)
+    else member = this.getUserFromMember(data);
     if (!member) return false;
-    let me = false
-    if (member.user) me = member.user.id === this.client.user.id
-    else me = member.id === this.client.user.id
+    let me = false;
+    if (member.user) me = member.user.id === this.client.user.id;
+    else me = member.id === this.client.user.id;
 
     // Verify channel
     const channel = this.getChannel(data);
--- a/src/client/actions/MessageReactionAdd.js
+++ b/src/client/actions/MessageReactionAdd.js
@@ -18,8 +18,7 @@ class MessageReactionAdd extends Action {
   handle(data) {
     if (!data.emoji) return false;
 
-    const client = this.client;
-    const guild = client.guilds.cache.get(data.guild_id);
+    const guild = this.client.guilds.cache.get(data.guild_id);
     let member = null;
     if (guild) member = this.getMember({ user: data.member.user }, guild);
     else member = this.getUserFromMember(data);
@@ -51,7 +50,7 @@ class MessageReactionAdd extends Action {
      * Emitted whenever a reaction is added to a cached message.
      * @event Client#messageReactionAdd
      * @param {MessageReaction} messageReaction The reaction object
-     * @param {GuildMember} member The guild member that applied the guild or reaction emoji
+     * @param {Member} member The member that applied the guild or reaction emoji
      */
     this.client.emit(Events.MESSAGE_REACTION_ADD, reaction, member);
 
--- a/src/client/actions/MessageReactionAdd.js
+++ b/src/client/actions/MessageReactionAdd.js
@@ -50,7 +50,7 @@ class MessageReactionAdd extends Action {
      * Emitted whenever a reaction is added to a cached message.
      * @event Client#messageReactionAdd
      * @param {MessageReaction} messageReaction The reaction object
-     * @param {Member} member The member that applied the guild or reaction emoji
+     * @param {GuildMember} member The member that applied the guild or reaction emoji
      */
     this.client.emit(Events.MESSAGE_REACTION_ADD, reaction, member);
 
--- a/src/client/actions/MessageReactionAdd.js
+++ b/src/client/actions/MessageReactionAdd.js
@@ -18,7 +18,9 @@ class MessageReactionAdd extends Action {
   handle(data) {
     if (!data.emoji) return false;
 
-    const member = this.getMember(data);
+    const client = this.client;
+    const guild = client.guilds.cache.get(data.guild_id);
+    const member = this.getMember({ user: data.member.user }, guild);
     if (!member) return false;
 
     // Verify channel
--- a/src/client/actions/MessageReactionAdd.js
+++ b/src/client/actions/MessageReactionAdd.js
@@ -18,8 +18,8 @@ class MessageReactionAdd extends Action {
   handle(data) {
     if (!data.emoji) return false;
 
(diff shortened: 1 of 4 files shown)


advaith1:
there is no member in DMs

ImRodry:
> there is no member in DMs

good point, I'll make it return user if there is no member then

src/client/actions/MessageReactionAdd.js
```diff
@@ -18,8 +18,10 @@ class MessageReactionAdd extends Action {
   handle(data) {
     if (!data.emoji) return false;
 
-    const user = this.getUserFromMember(data);
-    if (!user) return false;
+    const client = this.client;
+    const guild = client.guilds.cache.get(data.guild_id);
```
vBread, line 22: There's no point in creating a separate variable for the client
```suggestion
    const guild = this.client.guilds.cache.get(data.guild_id);
```

src/client/actions/MessageReactionAdd.js
```diff
@@ -32,23 +34,23 @@ class MessageReactionAdd extends Action {
...
     });
     if (!reaction) return false;
-    reaction._add(user);
+    reaction._add(member);
     /**
      * Emitted whenever a reaction is added to a cached message.
      * @event Client#messageReactionAdd
      * @param {MessageReaction} messageReaction The reaction object
-     * @param {User} user The user that applied the guild or reaction emoji
+     * @param {Member} member The guild member that applied the guild or reaction emoji
```
vBread:
```suggestion
     * @param {Member} member The member that applied the guild or reaction emoji
```

src/client/actions/MessageReactionRemove.js
```diff
@@ -27,18 +29,18 @@ class MessageReactionRemove extends Action {
...
+    const reaction = this.getReaction(data, message, member);
     if (!reaction) return false;
-    reaction._remove(user);
+    reaction._remove(member);
     /**
      * Emitted whenever a reaction is removed from a cached message.
      * @event Client#messageReactionRemove
      * @param {MessageReaction} messageReaction The reaction object
-     * @param {User} user The user whose emoji or reaction emoji was removed
+     * @param {Member} member The guild member whose emoji or reaction emoji was removed
```
vBread:
```suggestion
     * @param {Member} member The member whose emoji or reaction emoji was removed
```

src/client/actions/MessageReactionRemove.js
```diff
@@ -15,8 +15,10 @@ class MessageReactionRemove extends Action {
   handle(data) {
     if (!data.emoji) return false;
 
-    const user = this.getUser(data);
-    if (!user) return false;
+    const client = this.client;
+    const guild = client.guilds.cache.get(data.guild_id);
```
vBread, line 19:
```suggestion
    const guild = this.client.guilds.cache.get(data.guild_id);
```

vBread (changes requested):

src/client/actions/MessageReactionAdd.js
```diff
@@ -18,8 +18,14 @@ class MessageReactionAdd extends Action {
   handle(data) {
     if (!data.emoji) return false;
 
-    const user = this.getUserFromMember(data);
-    if (!user) return false;
+    const guild = this.client.guilds.cache.get(data.guild_id);
+    let member = null;
+    if (guild) member = this.getMember({ user: data.member.user }, guild);
+    else member = this.getUserFromMember(data);
```
monbrey, line 24: This results in the event emitting with unreliable types - developers would now need to check if its emitting a User or a GuildMember. This is less desirable imo.

The event should probably emit a third optional `member` in addition to the guaranteed user.

src/client/actions/MessageReactionAdd.js
```diff
@@ -32,23 +38,23 @@ class MessageReactionAdd extends Action {
...
     });
     if (!reaction) return false;
-    reaction._add(user);
+    reaction._add(member);
     /**
      * Emitted whenever a reaction is added to a cached message.
      * @event Client#messageReactionAdd
      * @param {MessageReaction} messageReaction The reaction object
-     * @param {User} user The user that applied the guild or reaction emoji
+     * @param {Member} member The member that applied the guild or reaction emoji
```
monbrey, line 53:
```suggestion
     * @param {GuildMember} member The member that applied the guild or reaction emoji
```
`Member` isnt a type

test/random.js
```diff
@@ -217,14 +217,14 @@ client.on('message', msg => {
   }
 });
 
-client.on('messageReactionAdd', (reaction, user) => {
+client.on('messageReactionAdd', (reaction, member) => {
```
monbrey, line 220: As above, `member` is not guaranteed and isn't reliable typing.

test/random.js
```diff
@@ -217,14 +217,14 @@ client.on('message', msg => {
...
   if (reaction.message.channel.id !== '222086648706498562') return;
-  reaction.message.channel.send(`${user.username} added reaction ${reaction.emoji}, count is now ${reaction.count}`);
+  reaction.message.channel.send(`${member.user.username} added reaction ${reaction.emoji}, count is now ${reaction.count}`);
 });
 
-client.on('messageReactionRemove', (reaction, user) => {
+client.on('messageReactionRemove', (reaction, member) => {
```
monbrey, line 225: Same here

src/client/actions/MessageReactionAdd.js
```diff
@@ -18,8 +18,14 @@ class MessageReactionAdd extends Action {
...
-    const user = this.getUserFromMember(data);
-    if (!user) return false;
+    const guild = this.client.guilds.cache.get(data.guild_id);
+    let member = null;
+    if (guild) member = this.getMember({ user: data.member.user }, guild);
+    else member = this.getUserFromMember(data);
+    if (!member) return false;
+    let me = false;
+    if (member.user) me = member.user.id === this.client.user.id;
+    else me = member.id === this.client.user.id;
```
NotSugden, line 28:
```suggestion
    const user = this.getUserFromMember(data);
    if (!user) return false;
    const guild = this.client.guilds.cache.get(data.guild_id);
    const member = guild && this.getMember(data, guild);
```

src/client/actions/MessageReactionAdd.js
```diff
@@ -32,23 +38,23 @@ class MessageReactionAdd extends Action {
     // Verify reaction
     if (message.partial && !this.client.options.partials.includes(PartialTypes.REACTION)) return false;
     const existing = message.reactions.cache.get(data.emoji.id || data.emoji.name);
-    if (existing && existing.users.cache.has(user.id)) return { message, reaction: existing, user };
+    if (existing && existing.users.cache.has(member.user.id)) return { message, reaction: existing, member };
```
NotSugden, line 41:
```suggestion
    if (existing && existing.users.cache.has(user.id)) return { message, reaction: existing, member };
```

src/client/actions/MessageReactionAdd.js
```diff
@@ -32,23 +38,23 @@ class MessageReactionAdd extends Action {
...
     const reaction = message.reactions.add({
       emoji: data.emoji,
       count: message.partial ? null : 0,
-      me: user.id === this.client.user.id,
+      me: me,
```
NotSugden, line 45:
```suggestion
      me: user.id === this.client.user.id,
```

src/client/actions/MessageReactionAdd.js
```diff
@@ -32,23 +38,23 @@ class MessageReactionAdd extends Action {
...
     });
     if (!reaction) return false;
-    reaction._add(user);
+    reaction._add(member);
```
NotSugden, line 48:
```suggestion
    reaction._add(user);
```

src/client/actions/MessageReactionAdd.js
NotSugden: also needs to be marked as optional
```diff
- {GuildMember} member
+ {GuildMember} [member]
```

NotSugden (changes requested):
same changes for remove event & typings need updating

iCrawl:
Sorry but this looks like a lot of jank for me. 

Looking at this code its almost easier to just call `guild.members.get(user.id)` yourself.

src/client/actions/MessageReactionAdd.js
monbrey: Its only optional if it gets moved to a third param (which it should imo). At the moment based on how its written, the typings are actually `{User|GuildMember} member`

NotSugden: oops, i misread the diff and read 3 parameters lol

src/client/actions/MessageReactionAdd.js
ImRodry: They can do that by checking if the event was triggered in or outside of DMs since it only emits user when it's not in a guild. I  don't think it would be too hard to do honestly.

src/client/actions/MessageReactionAdd.js
ImRodry: I actually had it like this but vBread suggested I should change that, I'll change it back

anandre:
I agree with Crawl. As far as I know, the library doesn't emit other events with structures that the API isn't providing. Adding this would be inconsistent with other parts of the library and adding unnecessary overhead to the library.

ImRodry:
> I agree with Crawl. As far as I know, the library doesn't emit other events with structures that the API isn't providing. Adding this would be inconsistent with other parts of the library and adding unnecessary overhead to the library.

when I was testing this I quickly realized that it wouldn't be as easy to do as I thought but I believe it would still make sense to change this since it would be quite handy for most people to be able to check the member's information on the server without having to use additional processes. If you guys don't think the same way, we can close this and move on, not that big of a deal

advaith1:
@anandre the gateway does include `member` in Message Reaction Add; that's the same as message.member in messageCreate

src/client/actions/MessageReactionAdd.js
vBread: No, I suggested you change the description, not the type

ImRodry:
Clearly this didn't work out as expected. Thought it was a good idea but since you guys don't agree, I'll be closing this now. Thank you for your input though!

PR #5180, opened 2021-01-05
docs(BaseManager): fix a typo in #resolveID description

iShibi:

This PR adds an extra check in the `BaseManager#resolveID` method to make sure that it doesn't return the `idOrInstance` parameter back when it is a non-numeric string. I am aware of the fact that the method only expects a string in the case of a `Snowflake` and passing any other type of string is the fault of the user, but I think in case a user does pass a string, which is not a snowflake the method should return `null` instead of the string that was passed.

I had a chat about this in the discord server with @kyranet and that is from where I got the `regex` used in the PR. It was him who advised me to use regex instead of what I was doing in a somewhat related project. Using `{17, 19}` instead of `{17, 18}` in the regex was @almostSouji's idea for keeping the regex future proof.

While working on this method I noticed a typo in the description of the method, so I fixed that too :)

**Status**

- [X] This PR **only** includes non-code changes, like changes to documentation, README, etc.

### DIFF
--- a/src/managers/BaseManager.js
+++ b/src/managers/BaseManager.js
@@ -69,7 +69,7 @@ class BaseManager {
    */
   resolveID(idOrInstance) {
     if (idOrInstance instanceof this.holds) return idOrInstance.id;
-    if (typeof idOrInstance === 'string' && /^\d{17,19}$/.test(idOrInstance)) return idOrInstance;
+    if (typeof idOrInstance === 'string') return idOrInstance;
     return null;
   }
 
--- a/src/managers/BaseManager.js
+++ b/src/managers/BaseManager.js
@@ -63,13 +63,13 @@ class BaseManager {
   }
 
   /**
-   * Resolves a data entry to a instance ID.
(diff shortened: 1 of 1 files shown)


kyranet:
As I mentioned in the support guild, I'm against this, `resolveID` is used in many places both internally and publicly, adding a RegExp check would slowdown all the internals.

If you're calling this on arbitrary input, you should sanitize your user's input.

With #5157, the type safety of everything using the `Snowflake` type will be increased, and this won't be necessary.

iShibi:
Okay, I understand it. We will be getting the check benefit at the cost of speed, which isn't good, given that `resolveID` is used in a lot of places in the library. Anyway, so should I create a new PR for the typo or is that fine too?

iShibi:
I'm sorry for that question, I'm kind of new to this 😅. Made a new commit with just the fixed typo, Thanks @Extroonie :)

kyranet (approved):

SpaceEEC (approved):

iCrawl (approved):

PR #5203, opened 2021-01-11
Consider Cloudflare Ratelimits when hitting 429

anna-rmrf:
We can get 429s from Cloudflare where the retryAfter is actually in seconds and not milliseconds which would result in an issue especially for large bots. I faced the issue myself and had to see how it can be identified.
When it's a cloudflare rate limit, the VIA header is missing and the body message has the "You are being blocked from accessing our API temporarily due to exceeding our rate limits frequently....". A safe solution would be awaiting in seconds and not milliseconds when this behavior is noticed. This fix did it for me.

**Status**

- [1] Code changes have been tested against the Discord API
- [1] I know how to update typings and have done so, or typings don't need updating

- [0] This PR changes the library's interface (methods or parameters added)
- [0] This PR includes breaking changes (methods removed or renamed, parameters moved or removed)
- [0] This PR **only** includes non-code changes, like changes to documentation, README, etc.

### DIFF
--- a/src/rest/RequestHandler.js
+++ b/src/rest/RequestHandler.js
@@ -4,172 +4,170 @@ const AsyncQueue = require('./AsyncQueue');
 const DiscordAPIError = require('./DiscordAPIError');
 const HTTPError = require('./HTTPError');
 const {
-    Events: {
-        RATE_LIMIT
-    },
+  Events: { RATE_LIMIT },
 } = require('../util/Constants');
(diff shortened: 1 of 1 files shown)


advaith1:
Discord sends it in seconds now so when d.js is updated to use API v8 (in the next update, v13), this will no longer be an issue.

anna-rmrf:
@advaith1 Ahh awesome, thanks for the clarification!

PR #5215, opened 2021-01-14
feat(Client): add InviteGenerationOptions#additionalScopes

ckohen:
This PR adds the ability to generate an invite link with additional scopes. 
It does not accept all scopes as valid due to whitelist requirements and that combining those scopes with the bot scopes is highly unlikely

**Status**

### DIFF
--- a/src/client/Client.js
+++ b/src/client/Client.js
@@ -423,7 +423,7 @@ class Client extends BaseClient {
         }
       });
       scopes.unshift('bot');
-      query.set('scope', scopes.join('+'));
+      query.set('scope', scopes.join(' '));
     }
 
     return `${this.options.http.api}${this.api.oauth2.authorize}?${query}`;
--- a/src/client/Client.js
+++ b/src/client/Client.js
@@ -417,11 +417,9 @@ class Client extends BaseClient {
 
     if (options.additionalScopes) {
       const scopes = options.additionalScopes;
-      scopes.forEach(scope => {
-        if (!InviteScopes.includes(scope)) {
-          throw new TypeError('INVALID_TYPE', 'additionalScopes', 'Array of Invite Scopes', true);
-        }
-      });
+      if (scopes.some(scope => !InviteScopes.includes(scope))) {
+        throw new TypeError('INVALID_TYPE', 'additionalScopes', 'Array of Invite Scopes', true);
+      }
       scopes.unshift('bot');
       query.set('scope', scopes.join(' '));
     }
--- a/src/client/Client.js
+++ b/src/client/Client.js
@@ -17,7 +17,7 @@ const Invite = require('../structures/Invite');
 const VoiceRegion = require('../structures/VoiceRegion');
 const Webhook = require('../structures/Webhook');
 const Collection = require('../util/Collection');
-const { Events, DefaultOptions } = require('../util/Constants');
+const { Events, DefaultOptions, InviteScopes } = require('../util/Constants');
 const DataResolver = require('../util/DataResolver');
 const Intents = require('../util/Intents');
 const Permissions = require('../util/Permissions');
@@ -377,6 +377,7 @@ class Client extends BaseClient {
    * @property {PermissionResolvable} [permissions] Permissions to request
    * @property {GuildResolvable} [guild] Guild to preselect
    * @property {boolean} [disableGuildSelect] Whether to disable the guild selection
+   * @property {InviteScope[]} additionalScopes
    */
 
   /**
@@ -412,6 +413,17 @@ class Client extends BaseClient {
       const guildID = this.guilds.resolveID(options.guild);
       if (!guildID) throw new TypeError('INVALID_TYPE', 'options.guild', 'GuildResolvable');
       query.set('guild_id', guildID);
+    }
+
+    if (options.additionalScopes) {
+      const scopes = options.additionalScopes;
+      scopes.forEach(scope => {
+        if (!InviteScopes.includes(scope)) {
+          throw new TypeError('INVALID_TYPE', 'additionalScopes', 'Array of Invite Scopes', true);
+        }
+      });
+      scopes.unshift('bot');
+      query.set('scope', scopes.join('+'));
     }
 
     return `${this.options.http.api}${this.api.oauth2.authorize}?${query}`;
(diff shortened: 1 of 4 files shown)


src/client/Client.js
```diff
@@ -414,6 +415,17 @@ class Client extends BaseClient {
       query.set('guild_id', guildID);
     }
 
+    if (options.additionalScopes) {
+      const scopes = options.additionalScopes;
+      scopes.forEach(scope => {
+        if (!InviteScopes.includes(scope)) {
+          throw new TypeError('INVALID_TYPE', 'additionalScopes', 'Array of Invite Scopes', true);
+        }
+      });
```
NotSugden, line 424: could this use `if (scopes.some(...)) { ... }` instead?

NotSugden (changes requested):

src/client/Client.js
```diff
@@ -414,6 +415,15 @@ class Client extends BaseClient {
       query.set('guild_id', guildID);
     }
 
+    if (options.additionalScopes) {
+      const scopes = options.additionalScopes;
+      if (scopes.some(scope => !InviteScopes.includes(scope))) {
```
izexi, line 420: A sanity check to ensure that `options.additionalScopes` is an array could be added here which will correspond with the thrown error below that.
```suggestion
      if (!Array.isArray(scopes) || scopes.some(scope => !InviteScopes.includes(scope))) {
```

src/client/Client.js
```diff
@@ -414,6 +415,15 @@ class Client extends BaseClient {
...
+        throw new TypeError('INVALID_TYPE', 'additionalScopes', 'Array of Invite Scopes', true);
+      }
+      scopes.unshift('bot');
+      query.set('scope', scopes.join(' '));
```
izexi, line 424: Not really a fan of using `Array#unshift()` here since it could end up mutating unintended arrays.
```suggestion
      query.set('scope', ['bot', ...scopes].join(' '));
```

src/client/Client.js
ckohen: oh wow, I can't believe I didn't already include this 🤦

iCrawl (approved):

src/client/Client.js
```diff
@@ -414,6 +415,14 @@ class Client extends BaseClient {
       query.set('guild_id', guildID);
     }
 
+    if (options.additionalScopes) {
+      const scopes = options.additionalScopes;
+      if (!Array.isArray(scopes) || scopes.some(scope => !InviteScopes.includes(scope))) {
+        throw new TypeError('INVALID_TYPE', 'additionalScopes', 'Array of Invite Scopes', true);
+      }
```
kyranet, line 422: Do we not want to throw an error with the invalid scope in question?

ckohen: In general, the library does not emit the cause of an error when checking within Arrays / Collections (erroring with similar messages as seen here). I would like to change the INVALID_TYPE error to be able to include a more descriptive error when dealing with said Arrays / Collections in general, but that is outside the scope of this PR.

If you think this should use a different error type that would solve this problem, let me know.

kyranet: In fact #5228 implemented the problematic bit in BitField, this is more a suggestion to adopt features we started implementing with v13, rather than keeping with what we did in the previous versions (v12 and lower).

ckohen: I see, I would think this would need a general refactor (this PR can add the error) to use `PARAMETER_INVALID` or something similar and perform the checks in different steps. I would love this overall to help neatly narrow down where errors occurred

SpaceEEC (approved):

src/client/Client.js
```diff
@@ -414,6 +415,14 @@ class Client extends BaseClient {
...
+      query.set('scope', ['bot', ...scopes].join(' '));
```
Mesteery, line 423:
```suggestion
      query.set('scope', `bot ${scopes.join(' ')}`);
```
I think it's better, because it avoids unnecessary iteration, it's not necessarily more readable, it's not necessarily less readable either.
It's still an interesting micro-optimization.

src/util/Constants.js
```diff
@@ -378,6 +378,34 @@ exports.WSEvents = keyMirror([
   'WEBHOOKS_UPDATE',
 ]);
 
+/**
+ * A valid scope to request when generating an invite link.
+ * <warn>Scopes that require whitelist are not considered valid for this generator</warn>
+ * * `applications.builds.read`: allows reading build data for a users applications
+ * * `applications.commands`: allows this bot to create commands in the server
+ * * `connections`: makes the endpoint for getting a users connections available
```
Mesteery, line 386: `applications.entitlements` and `applications.store.update` are missing.

Mesteery (changes requested):

src/client/Client.js
```diff
@@ -414,6 +415,17 @@ class Client extends BaseClient {
...
     }
 
+    if (options.additionalScopes) {
+      const scopes = options.additionalScopes;
+      if (!Array.isArray(scopes)) {
+        throw new TypeError('INVALID_TYPE', 'additionalScopes', 'Array of Invite Scopes', true);
+      } else if (scopes.some(scope => !InviteScopes.includes(scope))) {
+        const invalidScope = scopes.find(scope => !InviteScopes.includes(scope));
+        throw new TypeError('INVALID_ELEMENT', 'Array', 'additionalScopes', invalidScope);
+      }
```
kyranet, line 425: You can avoid a second iteration, since the previous `if` always throws when executed, the `else` branch is useless, which allows us to make a new `if`, and thus, we can do the following to avoid a second iteration:
```suggestion
      }
      const invalidScope = scopes.find(scope => !InviteScopes.includes(scope));
      if (invalidScope) {
        throw new TypeError('INVALID_ELEMENT', 'Array', 'additionalScopes', invalidScope);
      }
```

kyranet (changes requested):

src/client/Client.js
ckohen: I literally thought about this, then didn't do it 🤦,  thanks!

kyranet (approved):

vladfrangu (approved):

PR #5251, opened 2021-01-23
fix(GuildChannel): regression on default channel type

### LINKED ISSUE #5273: Error using GuildChannel.setTopic()
I have a bot which changes topic, but today after update to latest commit from master (5ac3b57f9bd53d1c20549a70942b023826f6f726) I have an Error:
```js
type: Value "text" is not int.
    at RequestHandler.execute (C:\Git\bot-main\node_modules\discord.js\src\rest\RequestHandler.js:152:13)
    at processTicksAndRejections (internal/process/task_queues.js:93:5)
    at async RequestHandler.push (C:\Git\bot-main\node_modules\discord.js\src\rest\RequestHandler.js:37:14)
    at async TextChannel.edit (C:\Git\bot-main\node_modules\discord.js\src\structures\GuildChannel.js:357:21) {
  method: 'patch',
  path: '/channels/769946367970902066',
  code: 50035,
  httpStatus: 400
}
```
I also tested on latest version 12.5.1 and topic is changing **without Error**.
I think this is after this commit: 5ac3b57f9bd53d1c20549a70942b023826f6f726

```js
client.channels.cache.get('769946367970902066').setTopic('Test').then(console.log).catch(console.error)
```

**Further details:**

- discord.js version: commit 5ac3b57f9bd53d1c20549a70942b023826f6f726
- Node.js version: 14.15.3
- Operating system: Debian 10 and Windows 10
- Priority this issue should have – please be realistic and elaborate if possible:
I can't change TextChannel topic.

**Relevant client options:**

- partials: none
- gateway intents: NON_PRIVILEGED
- other: none

I have also tested the issue on latest master, commit hash: 5ac3b57f9bd53d1c20549a70942b023826f6f726

almostSouji:

⚠️ closes #5273

The merge of #5022 has caused a regression, which this PR aims to fix.

When no channel is provided in data the PR defaults to `this.type`. This is however not the required numerical for the channel type the API expects resulting in

```
type: Value "text" is not int.
```

~~The other case (type is present in data) chooses to do the defaulting through the ChannelTypes enum and uppercasing the expected text value, which I have accordingly replicated in the defaulting case.~~

**Status**

### DIFF
--- a/src/structures/GuildChannel.js
+++ b/src/structures/GuildChannel.js
@@ -357,7 +357,7 @@ class GuildChannel extends Channel {
     const newData = await this.client.api.channels(this.id).patch({
       data: {
         name: (data.name || this.name).trim(),
-        type: data.type ? ChannelTypes[data.type.toUpperCase()] : this.type,
+        type: data.type ? ChannelTypes[data.type.toUpperCase()] : ChannelTypes[this.type.toUpperCase()],
         topic: data.topic,
         nsfw: data.nsfw,
         bitrate: data.bitrate || this.bitrate,
--- a/src/structures/GuildChannel.js
+++ b/src/structures/GuildChannel.js
@@ -357,7 +357,7 @@ class GuildChannel extends Channel {
     const newData = await this.client.api.channels(this.id).patch({
       data: {
         name: (data.name || this.name).trim(),
-        type: data.type ? ChannelTypes[data.type.toUpperCase()] : ChannelTypes[this.type.toUpperCase()],
+        type: ChannelTypes[data.type?.toUpperCase()],
         topic: data.topic,
         nsfw: data.nsfw,
         bitrate: data.bitrate || this.bitrate,


src/structures/GuildChannel.js
```diff
@@ -357,7 +357,7 @@ class GuildChannel extends Channel {
     const newData = await this.client.api.channels(this.id).patch({
       data: {
         name: (data.name || this.name).trim(),
-        type: data.type ? ChannelTypes[data.type.toUpperCase()] : this.type,
+        type: data.type ? ChannelTypes[data.type.toUpperCase()] : ChannelTypes[this.type.toUpperCase()],
```
vladfrangu, line 360: Or we could just send undefined if we don't patch it..? Thoughts?

Koyamie:
Shouldn't we still use `this.type` if no new type is sent to the data object ?

I was thinking about this:
```js
type: ChannelTypes[(data.type || this.type).toUpperCase()],
```

almostSouji:
Not required, has no benefit: <https://discord.com/developers/docs/resources/channel#modify-channel>
> All JSON parameters are optional.

Aka: you just need to pass what you actually want to edit

Koyamie:
> Not required, has no benefit: https://discord.com/developers/docs/resources/channel#modify-channel
> 
> > All JSON parameters are optional.
> 
> Aka: you just need to pass what you actually want to edit, meaning undefined (which this will otherwise result in) is valid

Alright, thanks 👍

Koyamie (approved):

iCrawl (approved):

src/structures/GuildChannel.js
```diff
@@ -357,7 +357,7 @@ class GuildChannel extends Channel {
     const newData = await this.client.api.channels(this.id).patch({
       data: {
         name: (data.name || this.name).trim(),
-        type: data.type ? ChannelTypes[data.type.toUpperCase()] : this.type,
+        type: ChannelTypes[data.type?.toUpperCase()],
```
kyranet, line 360: Wouldn't this trigger `ChannelTypes[undefined]` when type is not specified? If you wanted to pass a falsy value (including `0`), you could do:

```suggestion
        type: data.type ? ChannelTypes[data.type.toUpperCase()] : data.type,
```

Defaulting to whichever `data.type`'s value is when it's not truthy.

kyranet (changes requested):

src/structures/GuildChannel.js
vladfrangu: It doesn't matter that it'd call `ChannelTypes[undefined]`, since that, in turn, returns `undefined`, thus not being sent to Discord. You could argue we don't support numbers as types, but that should come if we ever replace string types with numbers

vladfrangu (approved):

kyranet (approved):

vladfrangu (approved):

SpaceEEC (approved):

almeidx:
You mentioned this pull request instead of the https://github.com/discordjs/discord.js/issues/5273 issue in the pull request description

Wilco2998:
it would be better if instead of using uppercased channel type name to use instead an integer that discord uses to represent a type https://discord.com/developers/docs/resources/channel#channel-object-channel-types

this how it looks for my own quick fix but off course for the library it would be better to map each type to an integer
afbeelding

almostSouji:
> it would be better if instead of using uppercased channel type name to use instead an integer

This is out of scope of this regression fix - we might do so going forward, as mentioned by vlad above.
>  You could argue we don't support numbers as types, but that should come if we ever replace string types with numbers

ZeProf2Code (approved):

thehairy (approved):

PR #5254, opened 2021-01-24
feat: expose RESTManager & create rest typings

Fyko:
This PR creates typings for all REST-related classes and exposes RESTManager to the user.
Example:
```ts
import { Client, Intents, RESTManager } from 'discord.js';
// intents are redundant for this example but required per #4879
const client = new Client({ intents: Intents.FLAGS.GUILDS });
const manager = new RESTManager(client, 'Bearer', 'the bearer token');
manager.api.users('@me').get().then(console.dir)
```

The JSDoc I wrote is definitely lacking and likely riddled with mistakes. Your reviews are greatly appreciated 😜 

**Status**

### DIFF
--- a/src/rest/RESTManager.js
+++ b/src/rest/RESTManager.js
@@ -7,13 +7,64 @@ const { Error } = require('../errors');
 const Collection = require('../util/Collection');
 const { Endpoints } = require('../util/Constants');
 
+/**
+ * Manages REST requests to the Discord API
+ */
 class RESTManager {
-  constructor(client, tokenPrefix = 'Bot') {
+  /**
+   *
+   * @param {Client} client The Discord client
+   * @param {?string} tokenPrefix The token prefix for the Authorization header
+   * @param {?string} token The token to perform requests with
+   */
+  constructor(client, tokenPrefix = 'Bot', token = client.token) {
+    /**
+     * The client who's configuration & utilities to use
+     * @type {Client}
+     * @readonly
+     * @private
+     */
     this.client = client;
+
+    /**
+     * The token used to perform requests
+     * @type {string}
+     * @readonly
+     * @private
+     */
+    this.token = token;
+
+    /**
+     * All current Requesthandlers
+     * @type {Collection<string, RequestHandler>}
+     * @readonly
+     * @private
+     */
     this.handlers = new Collection();
+
+    /**
+     * The token prefix for the Authorization header
+     * @type {string}
+     * @private
+     * @readonly
+     */
     this.tokenPrefix = tokenPrefix;
+
+    /**
+     * Whether or not to include the API version in requests
+     * @type {boolean}
+     * @private
+     * @readonly
+     */
     this.versioned = true;
+
+    /**
+     * The current global timeout, if any.
+     * @type {Promise<void>|null}
+     * @private
+     */
     this.globalTimeout = null;
+
     if (client.options.restSweepInterval > 0) {
       client.setInterval(() => {
         this.handlers.sweep(handler => handler._inactive);
@@ -21,20 +72,56 @@ class RESTManager {
     }
   }
 
+  /**
+   * Returns the route builder for crafting and eventually performing API requests.
+   * @example
+   * const manager = new RESTManager(client, 'Bearer', 'the bearer token');
+   * manager.api.users('@me').get().then(console.log);
+   * @example
+   * // returns an array of the authorized user's guilds
+   * manager.api.users('@me').guilds.get().then(console.log);
+   * @type {Object}
+   */
   get api() {
     return routeBuilder(this);
   }
 
+  /**
+   * Creates the Authorizaation header value
+   * @returns {string}
+   */
   getAuth() {
-    const token = this.client.token || this.client.accessToken;
-    if (token) return `${this.tokenPrefix} ${token}`;
+    if (this.token) return `${this.tokenPrefix} ${this.token}`;
     throw new Error('TOKEN_MISSING');
   }
 
+  /**
+   * Returns various functions for creating CDN links
+   * @type {Object}
+   */
   get cdn() {
     return Endpoints.CDN(this.client.options.http.cdn);
   }
 
+  /**
+   * Options for performing a REST request
+   * @typedef RequestOptions
+   * @type {Object}
+   * @property {URLSearchParams|object} [query] The querystring for this request
+   * @property {boolean} [versioned] Whether or not to include the API version in request url
+   * @property {boolean} [auth] Whether or not this request requires authorization
+   * @property {string} [reason] The value for the X-Audit-Log-Reason header the guild audit log
+   * @property {Object} [headers] The headers for the request. Content-Type and Authorization are handled internally
+   * @property {Object} [data] The request body
+   */
+
+  /**
+   * Performs a request
+   * @param {string} method The HTTP method to perform
+   * @param {string} url The url to request
+   * @param {?RequestOptions} options The options for the request
+   * @returns {Collection<string, *> | Buffer}
+   */
   request(method, url, options = {}) {
     const apiRequest = new APIRequest(this, method, url, options);
     let handler = this.handlers.get(apiRequest.route);
@@ -47,6 +134,10 @@ class RESTManager {
     return handler.push(apiRequest);
   }
 
+  /**
+   * The base url for API requests
+   * @type {string}
+   */
   get endpoint() {
     return this.client.options.http.api;
   }
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -1,5 +1,3 @@
-import { URLSearchParams } from 'url';
-
 declare enum ChannelType {
   text = 0,
   dm = 1,
@@ -18,6 +16,7 @@ declare module 'discord.js' {
   import { PathLike } from 'fs';
   import { Response } from 'node-fetch';
   import { Readable, Stream, Writable } from 'stream';
+  import { URLSearchParams } from 'url';
   import * as WebSocket from 'ws';
 
   export const version: string;
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -1,3 +1,5 @@
+import { URLSearchParams } from 'url';
+
 declare enum ChannelType {
   text = 0,
   dm = 1,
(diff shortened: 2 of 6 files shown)


src/rest/RESTManager.js
```diff
@@ -7,34 +7,121 @@ const { Error } = require('../errors');
...
+/**
+ * Manages REST requests to the Discord API
+ */
 class RESTManager {
-  constructor(client, tokenPrefix = 'Bot') {
+  /**
+   *
+   * @param {Client} client The Discord client
+   * @param {?string} tokenPrefix The token prefix for the Authorization header
+   * @param {?string} token The token to perform requests with
```
papaia, line 18:
```suggestion
   * @param {?string} [tokenPrefix='Bot'] The token prefix for the Authorization header
   * @param {?string} [token=client.token] The token to perform requests with
```

src/rest/RESTManager.js
```diff
@@ -7,34 +7,121 @@ const { Error } = require('../errors');
...
+   * // returns an array of the authorized user's guilds
+   * manager.api.users('@me').guilds.get().then(console.log);
+   * @type {Object}
+   */
   get api() {
     return routeBuilder(this);
   }
 
+  /**
+   * Creates the Authorizaation header value
```
papaia, line 90:
```suggestion
   * Creates the Authorization header value
```

typings/index.d.ts
```diff
@@ -14,11 +14,85 @@ declare module 'discord.js' {
...
+  import { URLSearchParams } from 'url';
   import * as WebSocket from 'ws';
 
   export const version: string;
 
+  //#region Rest
+
+  class AsyncQueue {
+    // tslint:disable-next-line:ban-types
+    private readonly promises: { promise: Promise<void>; resolve: Function }[];
```
papaia, line 28:
```suggestion
    private readonly promises: { promise: Promise<void>; resolve: (value?: any) => void }[];
```

src/rest/RESTManager.js
```diff
@@ -7,34 +7,121 @@ const { Error } = require('../errors');
...
+     */
     this.client = client;
+
+    /**
+     * The token used to perform requests
+     * @type {string}
+     * @readonly
+     * @private
+     */
+    this.token = token;
```
papaia, line 35:
```suggestion
    Object.defineProperty(this, 'token', { value: token, writable: true });
```

src/rest/RESTManager.js
```diff
@@ -7,34 +7,121 @@ const { Error } = require('../errors');
...
-  constructor(client, tokenPrefix = 'Bot') {
+  /**
+   *
+   * @param {Client} client The Discord client
+   * @param {?string} tokenPrefix The token prefix for the Authorization header
+   * @param {?string} token The token to perform requests with
+   */
+  constructor(client, tokenPrefix = 'Bot', token = client.token) {
+    /**
+     * The client who's configuration & utilities to use
```
papaia, line 22:
```suggestion
     * The client whose configuration & utilities to use
```

src/rest/RESTManager.js
```diff
@@ -7,34 +7,121 @@ const { Error } = require('../errors');
...
+     * @type {Client}
```
papaia, line 23: Could this possibly be a `BaseClient`?

src/rest/RESTManager.js
```diff
@@ -7,34 +7,121 @@ const { Error } = require('../errors');
...
+    /**
+     * The token used to perform requests
+     * @type {string}
+     * @readonly
+     * @private
+     */
+    this.token = token;
+
+    /**
+     * All current Requesthandlers
```
papaia, line 38:
```suggestion
     * All current request handlers
```

src/rest/RESTManager.js
```diff
@@ -7,34 +7,121 @@ const { Error } = require('../errors');
...
   getAuth() {
-    const token = this.client.token || this.client.accessToken;
-    if (token) return `${this.tokenPrefix} ${token}`;
+    if (this.token) return `${this.tokenPrefix} ${this.token}`;
     throw new Error('TOKEN_MISSING');
   }
 
+  /**
+   * Returns various functions for creating CDN links
+   * @type {Object}
```
papaia, line 100:
```suggestion
   * @type {object}
```

papaia (changes requested):

typings/index.d.ts
kyranet: From the original TypeScript code, there's this interface:

```typescript
interface InternalAsyncQueueDeferredPromise {
	resolve(): void;
	promise: Promise<void>;
}
```

And `AsyncQueue#promises` is `InternalAsyncQueueDeferredPromise[]`. Therefore:

```suggestion
    private readonly promises: { promise: Promise<void>; resolve(): void }[];
```

typings/index.d.ts
```diff
@@ -14,11 +14,85 @@ declare module 'discord.js' {
...
+  }
+
+  type RouteBuilder = Record<string, RouteBuilderMethods & ((...args: string[]) => RouteBuilderMethods)>;
+
+  export class RESTManager {
+    public constructor(client: Client, tokenPrefix?: string, token?: string);
+    public readonly client: Client;
+    private readonly handlers: Collection<string, RequestHandler>;
+    private readonly tokenPrefix: string;
+    private readonly versoined: boolean;
```
kyranet, line 72:
```suggestion
    private readonly versioned: boolean;
```

typings/index.d.ts
```diff
@@ -14,11 +14,85 @@ declare module 'discord.js' {
...
+    public globalTimeout: Promise<void> | null;
+    public cdn: string;
+    public set endpoint(endpoint: string);
+
+    public api: RouteBuilder;
+    public getAuth(): string;
+    public request(method: string, url: string, options: RequestOptions): Collection<string, unknown> | Buffer;
```
kyranet, line 79: Wait... in which case does `RESTManager#request` return a Collection?

typings/index.d.ts
```diff
@@ -14,11 +14,85 @@ declare module 'discord.js' {
...
+    public push(): Collection<string, unknown> | Buffer;
+  }
+
+  interface RequestOptions {
+    query?: URLSearchParams | Record<string, string | string[]>;
+    versioned?: boolean;
+    auth?: boolean;
+    reason?: string;
+    headers?: Record<string, string>;
+    data?: Record<string, unknown>;
```
kyranet, line 54: Those are very troublesome, for example:

```typescript
interface Headers {
	Authorization: string;
}

const headers: Headers = { Authorization: 'Hello World' };

const options: RequestOptions = { headers };
```

Causes this error:

```typescript
Type 'Headers' is not assignable to type 'Record<string, string>'.
  Index signature is missing in type 'Headers'.(2322)
```

To properly fix it, the `{}` type (albeit discouraged by a TS ESlint rule) accepts *any* non-null object, which is what we want.

kyranet (changes requested):

iCrawl:
Merging this would publicly expose the whole API since you marked neither the class nor the methods private:
https://github.com/discordjs/discord.js/pull/5150#issuecomment-751306243

typings/index.d.ts
Fyko: That was meant to be Record<string, unknown>

src/rest/RESTManager.js
Fyko: Per https://jsdoc.app/tags-type.html, `object` is reserved for the type of typedefs

typings/index.d.ts
Fyko: Resolved in https://github.com/discordjs/discord.js/pull/5254/commits/e5ce43d328229f40dce8d92af40b21596ae50d98

Fyko:
> Merging this would publicly expose the whole API since you marked neither the class nor the methods private:
> [#5150 (comment)](https://github.com/discordjs/discord.js/pull/5150#issuecomment-751306243)

That's why I made this PR: to expose the RESTManager so it can be used for performing requests with a Bearer token.
I assumed that was the intended purpose considering `RESTManager#getAuth` checked `Client#accessToken` but the class itself was never documented nor exported
https://github.com/discordjs/discord.js/blob/9ffcd83027f0fc06d69df21475865ad55138de01/src/rest/RESTManager.js#L29

What would you like changed to comply with keeping it private at the Client scope?
re: @iCrawl

src/rest/RESTManager.js
```diff
@@ -7,34 +7,121 @@ const { Error } = require('../errors');
...
+   * @property {Object} [headers] The headers for the request. Content-Type and Authorization are handled internally
+   * @property {Object} [data] The request body
+   */
+
+  /**
+   * Performs a request
+   * @param {string} method The HTTP method to perform
+   * @param {string} url The url to request
+   * @param {?RequestOptions} options The options for the request
+   * @returns {Collection<string, *> | Buffer}
```
NotSugden, line 123:
```suggestion
   * @returns {Collection<string, *>|Buffer}
```

Fyko: I didn't add whitespace there for the sake of consistency

NotSugden: yeah, the suggestion is removing the whitespace?

Fyko: i guess i forgot how to read, my bad

typings/index.d.ts
```diff
@@ -14,11 +14,85 @@ declare module 'discord.js' {
...
+
+  interface RouteBuilderMethods {
+    get(opts?: RequestOptions): Promise<unknown>;
+    post(opts?: RequestOptions): Promise<unknown>;
+    delete(opts?: RequestOptions): Promise<unknown>;
+    patch(opts?: RequestOptions): Promise<unknown>;
+    put(opts?: RequestOptions): Promise<unknown>;
+  }
+
+  type RouteBuilder = Record<string, RouteBuilderMethods & ((...args: string[]) => RouteBuilderMethods)>;
```
tguichaoua, line 65:
```suggestion
type HttpMethod = 'get' | 'post' | 'delete' | 'patch' | 'put';

export type RouteBuilder = Record<
    HttpMethod,
    <T>(options?: RequestOptions) => Promise<T>
> &
    {
        [k in string]: RouteBuilder;
    } &
    ((...args: string[]) => RouteBuilder);
```
With your implementation you cannot do more that 1 call (eg `api.foo.goo.get()`).
Also I added a generic type for http methods but I don't know if it's a good idea or not.

[playground](https://www.typescriptlang.org/play?#code/JYOwLgpgTgZghgYwgAgEoQI4FcIGcwDyADmMAPYi7IDeyAvgLABQzYAnkSgBJhhECyEMAAsyAE2QBeZAHIA5kJnIAPrKJl8S1TLEQANkIha1cMAmHGZRLGBkBuZswgAPdVDDJ2nNGRsQAQljAerpQUmgQCGRQYgA8zMiJyDx8giLiADQJSbEAKgB8ABRkJOSUAPwAXBHYeISlFLgAlFL5yAAKUGQAtsC4EHn5zG0AZNmJ1ONJiQDaANbIoMj4UKByALrVqL6QgcGhDkzTdMhjR0mFhQB0N3BQcrjVK2sz6y2Sbdt+eyHQTYfMAweLp+LY7AJBX5haTUE5wKhwEBsAFMEGQK4wMhkK5yLE4oSFf6OJgAejJ5IplKplOJoEgsEQKC+u0hoTSojEVEm50SCjAxRIuCqNRw+GIpEaTWqnR6fQGWBAcxAZAA7iB8odpup8AKwEKtphRfUJZQpR0ur1+rEFUrVerNUldECILr9SK6uKys1pRa5dbFcq1RqpkRTOZXcL0LUxQ1TT7ZVabYH7SGbBGDdHjV6zTLLfKA3bg0xGCwmF4UAB9ZkQ-bQcLoKIxWLPEByDI+b6s6Ds8RUEbIS43K53B5PMCrVuvd6fcE-NlCDnNIuAoTIOBEYDVKuzrvQmhwhFIlHr4AYvG47F8wl2ZBkmWcdxsWQXpRiMh4ZDKjwuPoeCieDgUBkas527Bde1OAdrlue5HmWccXjeVoOxZWsoB7TkmhkIA)

Fyko: Good catch! As for the generic type, see https://github.com/discordjs/discord.js/pull/5150#discussion_r548905773.

iCrawl:
> What would you like changed to comply with keeping it private at the Client scope?

I would honestly not want it to be exposed at all. Having it be private till this day and continued is a very deliberate decision on our part, again referring back to Vlads comment.

Fyko:
understandable, thanks for the transparency

PR #5256, opened 2021-01-24
refactor(APIRouter): require square brackets for endpoint paths

papaia:
Requires consistent use of `[]` for paths when using the route builder.
```diff
- api.users('@me').get()
+ api.users['@me'].get()

- api.guilds(guild.id).emojis(emoji).get()
+ api.guilds[guild.id].emojis[emoji].get()
```
This also protects the option that in the future routes will have `get`, `post`, or any other HTTP method in them:
```js
// old
api.a.b.c.get.toString() // options => manager.request(...
api.a.b.c.get() // GET /a/b/c
// new
api.a.b.c.get.toString() // '/a/b/c/get'
api.a.b.c.get() // GET /a/b/c
```
#### Notes
- **The goal of this PR is to enforce one style over the other - literally throw when not used "incorrectly". This goes farther than simply changing all occurrences to one style. We could stick to the other style (`()`) and manually make sure new PRs don't add `[]` by mistake, but that's not really enforcing it.**
 
### DIFF
--- a/src/errors/Messages.js
+++ b/src/errors/Messages.js
@@ -105,6 +105,8 @@ const Messages = {
   FETCH_GROUP_DM_CHANNEL: "Bots don't have access to Group DM Channels and cannot fetch them",
 
   MEMBER_FETCH_NONCE_LENGTH: 'Nonce length must not exceed 32 characters.',
+
+  API_ROUTER_INVALID_CALL: 'API router function calls must use HTTP verbs.',
 };
 
 for (const [name, message] of Object.entries(Messages)) register(name, message);
--- a/src/structures/interfaces/TextBasedChannel.js
+++ b/src/structures/interfaces/TextBasedChannel.js
@@ -321,7 +321,7 @@ class TextBasedChannel {
       }
       if (messageIDs.length === 0) return new Collection();
       if (messageIDs.length === 1) {
-        await this.client.api.channels(this.id).messages(messageIDs[0]).delete();
+        await this.client.api.channels[this.id].messages(messageIDs[0]).delete();
         const message = this.client.actions.MessageDelete.getMessage(
           {
             message_id: messageIDs[0],
--- a/src/structures/interfaces/TextBasedChannel.js
+++ b/src/structures/interfaces/TextBasedChannel.js
@@ -321,7 +321,7 @@ class TextBasedChannel {
       }
       if (messageIDs.length === 0) return new Collection();
       if (messageIDs.length === 1) {
-        await this.client.api.channels[this.id].messages(messageIDs[0]).delete();
+        await this.client.api.channels[this.id].messages[messageIDs[0]].delete();
         const message = this.client.actions.MessageDelete.getMessage(
           {
             message_id: messageIDs[0],
--- a/src/client/Client.js
+++ b/src/client/Client.js
@@ -265,10 +265,7 @@ class Client extends BaseClient {
    */
   fetchGuildTemplate(template) {
     const code = DataResolver.resolveGuildTemplateCode(template);
-    return this.api.guilds
-      .templates(code)
-      .get()
-      .then(data => new GuildTemplate(this, data));
+    return this.api.guilds.templates[code].get().then(data => new GuildTemplate(this, data));
   }
 
   /**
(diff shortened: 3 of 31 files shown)


src/structures/interfaces/TextBasedChannel.js
```diff
@@ -321,7 +321,7 @@ class TextBasedChannel {
       }
       if (messageIDs.length === 0) return new Collection();
       if (messageIDs.length === 1) {
-        await this.client.api.channels(this.id).messages(messageIDs[0]).delete();
+        await this.client.api.channels[this.id].messages(messageIDs[0]).delete();
```
almostSouji, line 324: If you want to enforce square brackets in the library (as proposed in #5256 which seems to be included as given here?) you need to update the latter here. 

Though I'm personally not sure why that's necessary, to begin with.

ToledoSDL (approved):

almostSouji:
I generally agree with consistency, so either switching to `()` or `[]` entirely is probably a good call. I'm not yet sure which I personally prefer yet. Using `[]` eliminates functionality in passing multiple parameters at once `api.webhooks('id', 'token')` would need to be `api.webhooks[id][token]`. 

In the end, it's a style choice.

papaia:
Not really.
If you switch to using only `()`, you make it sloppy because you will then call `api.gateway().bot()`, which 1) makes the paths with no params longer and way less readable, and 2) confusing as to what is a request and what is the path.
If we're talking consistency here I believe that `[id][token]` instead of `(id, token)` is a "compromise" we're willing to take to make it clear what is the route and where the request happens.

vladfrangu:
> because you will then call `api.gateway().bot().get()`

The `apply` trap is different from the `get` trap, that request would remain formatted as `api.gateway.bot.get()`, unless this PR changes that too...

---

Personally, I prefer sticking with `()`, but I know others don't agree. I'm all for consistency tho, so whichever one is the chosen one is fine by me.

almostSouji:
No, that example is just just wrong. 

Both the `()` and the `[]` approach are for parts of the path that have variable input. `api.guilds("12345678927384").channels("129387378734").get()` vs. `api.guilds["12345678927384"].channels["129387378734"].get()`. 

The example you proposed above would be `api.gateway.bot.get()` no matter which of the two approaches is taken.

> If we're talking consistency here I believe that [id][token] instead of (id, token) is a "compromise" we're willing to take to make it clear what is the route and where the request happens.

We (discord.js, internals, staff, and core developers) haven't made any decision on this so far. We might even retire the route builder in its current form before v13 is released.

We do appreciate that you are trying to be a driving force in terms of progress and consistency, but please wait for decisions to be made before taking this proposal as something that will definitely be applied and even requesting changes on other PRs regarding it.

papaia:
> The example you proposed above would be `api.gateway.bot.get()` no matter which of the two approaches is taken.

No. This PR currently allows both `[]` and `()` like in the past, with the added possibility of `.get.something` without erroring, and the reason for that is to account for there are places where I didn't replace the `()`.
The end goal is to change this if block:
```js
if (!methods.includes(method)) {
  route.push(...args.filter(x => x != null)); // eslint-disable-line eqeqeq
  return new Proxy(noop, handler);
}
```
To error when `apply` is called if the last in the array is not an http method, so calling anything but `.get()`, `.post()`, etc. would not work - enforcing the consistency by blocking the alternative.
Simply updating all occurrences of brackets to `()` would be consistent, but you could still use `[]` - You're not enforcing the consistency.

src/errors/Messages.js
```diff
@@ -105,6 +105,8 @@ const Messages = {
   FETCH_GROUP_DM_CHANNEL: "Bots don't have access to Group DM Channels and cannot fetch them",
 
   MEMBER_FETCH_NONCE_LENGTH: 'Nonce length must not exceed 32 characters.',
+
+  API_ROUTER_INVALID_CALL: 'API router function calls must use HTTP verbs.',
```
papaia, line 109: For anyone reviewing this, please suggest better wording for this error message.

iCrawl:
There have been so many changes, and not really any developer was on board with this PR, so I am going to close this.

PR #5269, opened 2021-01-27
chore(*): streamline issue and PR templates and redirects

almostSouji:

- GitHub uses checkboxes as todo lists only. We currently use them as checkboxes for versioning classification, which intersects badly with GitHub native features (milestone progress, progress notifications, the progress bar in PR overview), to solve the issue this PR moves classification and optional lines into a comment at the bottom (see next point) to be moved out if applicable.
- Fix grammar and typo in various places, streamline wording to be more concise and compact
- Remove comments from the top. The "please ask questions on discord.js" is now covered with a button in the issue view. The comment is not stripped when posted to our webhook, blocking any and all useful information in favor of those lines.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -6,12 +6,6 @@ title: ''
 labels: 's: unverified, type: bug'
 assignees: ''
 ---
-<!--
-If you need help with discord.js installation or usage, please go to the discord.js Discord server instead:
-  https://discord.gg/bRCvFy9
-This issue tracker is only for bug reports and enhancement suggestions.
-You won't receive any basic help here.
--->
 
 **Please describe the problem you are having in as much detail as possible:**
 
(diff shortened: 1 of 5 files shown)


.github/SUPPORT.md
```diff
@@ -1,7 +1,7 @@
 # Seeking support?
 
-We're sorry, we only use this issue tracker for bugs in the library itself and feature requests for it. We are not able to provide general support or answser questions on the issue tracker.
+Sorry, we only use this issue tracker for bug reports and feature request. We are not able to provide general support or answer questions in the form of GitHub issues.
```
iCrawl, line 3: I don't think we need to word this as an apology.

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -6,12 +6,6 @@ title: ''
 labels: 's: unverified, type: bug'
 assignees: ''
 ---
-<!--
```
iCrawl, line 9: I'm not sure if we should remove this comment.
Maybe rephrase it a bit, but overall this surely catches some people who want to create issues when theres a question, even if we still have some that entirely ignore it.

almostSouji: I feel like the redirect solves this problem better than the comment does. This renders our GH hook description quite useless, if not removed
image

Additionally, we recently have converted questions that weren't super trivial to discussions, which might also be a valid approach going forward

iCrawl: You are right in the assumption that the normal workflow would be:

"New issue" -> read what it says there -> evaluate whether to create an issue or go to the server -> create issue/go to the server.

But in most cases, I'm sure a lot of people just skip all the middle steps and just go directly to the first "bug report" because they cannot differentiate between a bug report and an issue on their side.

This is solely a further mitigation method on our part.

almostSouji: I see that point, yeah, I'll put a short one-liner in place in hopes to deter some of the more persistent cases without sacrificing the entire GH webhook to it 👍

iCrawl (approved):

PR #5278, opened 2021-02-01
feat(GuildMemberManager): add kick method

iShibi:

Currently, `GuildMember#kick` is the only way to kick a user from the guild. Although most of the times, the member we want to kick can be obtained either from mentions or cache but there can be some cases where we only have the id of an uncached member. To kick that uncached member we have to fetch the member first and then call `#kick` on it. This unnecessary API call can be avoided if we expose the `Remove Guild Member` endpoint via `GuildMemberManager#kick`.


vaporox:
This is covered by #4872 already

iShibi:
I should make checking open PRs beforehand a habit, made the same mistake twice in a day. It's closing time again then. 😔

PR #5283, opened 2021-02-01
feat(WebSocketManager): let identify throw on depleted limits

### LINKED ISSUE #5282: Throw when identify limit is fully depleted instead of delaying

Client stalls indefinitely when identify limit is depleted. To find the source of the misbehavior users currently have to go great length (adding a listener for the `"debug"` event) to find out what's stalling their bot

Instead of just preventing going over the limit, throw an error

(maybe optionally restore the old behavior with explicit opt-in, the default should be throwing - noone wants to wait for 24h for a successful login, let's be real)

**Additional context**

almostSouji:

⚠️ closes #5282
✔️  ~~untested~~ now tested by @vladfrangu 

### DIFF
--- a/src/client/websocket/WebSocketManager.js
+++ b/src/client/websocket/WebSocketManager.js
@@ -93,16 +93,6 @@ class WebSocketManager extends EventEmitter {
      * @private
      */
     this.reconnecting = false;
-
-    /**
-     * The current session limit of the client
-     * @private
-     * @type {?Object}
-     * @property {number} total Total number of identifies available
-     * @property {number} remaining Number of identifies remaining
-     * @property {number} reset_after Number of milliseconds after which the limit resets
-     */
-    this.sessionStartLimit = null;
   }
 
   /**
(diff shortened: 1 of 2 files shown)


vladfrangu:
This doesn't seem to be a major change at all? We could argue the previous behavior was unintended, and this is a "bug fix"

Could we at least keep the log that logs the limit on connect?

iCrawl:
We can absolutely not argue that, and considering there is no v12 release anymore it's not really relevant.

Wilco2998:
could this also be applied to more of the functionality of the library like when changing something about a role

monbrey:
> 
> 
> could this also be applied to more of the functionality of the library like when changing something about a role>
> I got a ratelimit when trying to change a role color too many times

When developing code that respects rate limits and works with them, the queuing behaviour is far more useful than throwing a rejection. The identify limit is the exception to this.

When writing code that completely *disrepects* Discord's rate limits with actions they specifically discourage, like rainbow roles, you should be reconsidering your bots features.

vladfrangu (changes requested):
Can you please add the debug log back at least? 🙏

vladfrangu (approved):
Pls ignore my previous review, my brain is currently lagging behind a LOT. LGTM

iCrawl (approved):

kyranet (approved):

SpaceEEC:
`WebSocketManager#sessionStartLimit` seems to be only assigned to once when calling `connect`, which does not seem to be called again.
If I'm not wrong with that, that property could probably be removed, as it will just go stale.

PR #5164, opened 2020-12-29
feat(*): remove ALL unnecessary getters

zaida04:
This WIP pr is meant to remove ALL unnecessary getters like Message#editable or Message#deletable and etc. These getters do nothing but put more responsibility on the lib and pollute structures with getters that do things that can easily be done by the end user. This might take a bit 🤣.

All help is appreciated in this, if I miss some getters or if I mistakenly remove a necessary getter.

**Status**

- [X] Code changes have been tested against the Discord API, or there are no code changes
- [X] I know how to update typings and have done so, or typings don't need updating

**Semantic versioning classification:**

- [X] This PR changes the library's interface (methods or parameters added)
  - [X] This PR includes breaking changes (methods removed or renamed, parameters moved or removed)
- [ ] This PR **only** includes non-code changes, like changes to documentation, README, etc.

PR #5287, opened 2021-02-02
feat: remove permission getters attempt no.2, electric boogaloo

zaida04:
After further discussion in the Discord Server, I have decided to revive https://github.com/discordjs/discord.js/pull/5164 through this PR. I believe I prematurely closed that PR. This move is bound to be a controversial one, so more discussion is welcome on this topic. 

Before you go ahead and downvote the PR, be sure to understand that this PR only aims to cleanup useless getters that shouldn't be the responsibility of the lib to provide. 

## What will this PR remove?
- getters that are simple wrappers over easily doable permission checks/id checks

## What *won't* this PR remove?
- getters that are relied on heavily by other internal structures.
- getters that do a level of checking higher than simple user-doable checks

## Getters Removal List
- [X] Message#editable
- [x] Message#pinnable
- [ ] GuildChannel#viewable (by extension this will remove StoreChannel#viewable, TextChannel#viewable, VoiceChannel#viewable, NewsChannel#viewable)
- [x] GuildChannel#deletable (by extension this will remove StoreChannel#deletable, TextChannel#deletable, VoiceChannel#deletable, NewsChannel#deletable)
- [ ] VoiceChannel#editable
- [x] VoiceChannel#speakable
- [x] GuildMember#bannable
- [x] GuildMember#kickable

⚠️Before you suggest getters to add to the removal list, ensure that these getters **aren't** used heavily elsewhere. Don't want to break lib functionality do we now 👀 

-----

### DIFF
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -796,7 +796,6 @@ declare module 'discord.js' {
 
   export class GuildMember extends PartialTextBasedChannel(Base) {
     constructor(client: Client, data: object, guild: Guild);
-    public readonly bannable: boolean;
     public deleted: boolean;
     public readonly displayColor: number;
     public readonly displayHexColor: string;
@@ -806,7 +805,6 @@ declare module 'discord.js' {
     public pending: boolean;
     public readonly joinedAt: Date | null;
     public joinedTimestamp: number | null;
-    public readonly kickable: boolean;
     public lastMessageChannelID: Snowflake | null;
     public readonly manageable: boolean;
     public nickname: string | null;
@@ -3091,19 +3089,16 @@ declare module 'discord.js' {
   interface PartialGuildMember
     extends Partialize<
       GuildMember,
-      | 'bannable'
       | 'displayColor'
       | 'displayHexColor'
       | 'displayName'
       | 'guild'
-      | 'kickable'
       | 'permissions'
       | 'roles'
       | 'manageable'
       | 'presence'
       | 'voice'
     > {
-    readonly bannable: boolean;
     readonly displayColor: number;
     readonly displayHexColor: string;
     readonly displayName: string;
@@ -3111,7 +3106,6 @@ declare module 'discord.js' {
     readonly manageable: boolean;
     joinedAt: null;
     joinedTimestamp: null;
-    readonly kickable: boolean;
     readonly permissions: GuildMember['permissions'];
     readonly presence: GuildMember['presence'];
     readonly roles: GuildMember['roles'];
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -962,7 +962,6 @@ declare module 'discord.js' {
     public createdTimestamp: number;
     public readonly deletable: boolean;
     public deleted: boolean;
-    public readonly editable: boolean;
     public readonly editedAt: Date | null;
     public editedTimestamp: number | null;
     public embeds: MessageEmbed[];
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -3128,7 +3128,6 @@ declare module 'discord.js' {
       | 'channel'
       | 'deletable'
       | 'crosspostable'
-      | 'editable'
       | 'mentions'
       | 'url'
       | 'flags'
@@ -3138,7 +3137,6 @@ declare module 'discord.js' {
     channel: Message['channel'];
     readonly deletable: boolean;
     readonly crosspostable: boolean;
-    readonly editable: boolean;
     embeds: Message['embeds'];
     flags: Message['flags'];
     mentions: Message['mentions'];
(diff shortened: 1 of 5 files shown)


alii:
> Before you go ahead and downvote the PR, be sure to understand that this PR only aims to cleanup useless getters that shouldn't be the responsibility of the lib to provide.

Out of curiosity, why is it _not_ the lib's responsibility to provide. Sure, it's not a property coming from Discord's API directly, but they are useful and save a lot of time when using the library. Removing them makes the library less user friendly and considering the number of beginners that write with D.js, it would mean they are writing more boilerplate that makes the API more complex to use.

Sure, they are redoable, but imho they are far from useless.

robertt:
> Before you go ahead and downvote the PR, be sure to understand that this PR only aims to cleanup useless getters that shouldn't be the responsibility of the lib to provide.

Using this argument, why does the library provide an embed builder if it can be done without?

zaida04:
> > Before you go ahead and downvote the PR, be sure to understand that this PR only aims to cleanup useless getters that shouldn't be the responsibility of the lib to provide.
> 
> Using this argument, why does the library provide an embed builder if it can be done without?

The point of this PR is to fix consistency. We don't have getters such as GuildChannel#sendable and it feels bloated to have getters that wrap around one line checks. The embed builder is *not* a getter that wraps around a one line permission check, and is far more robust than any of the getters being targeted in this PR. To try to compare them to each other when they both serve different purposes is incorrect in the context of this PR.

zaida04:
> > Before you go ahead and downvote the PR, be sure to understand that this PR only aims to cleanup useless getters that shouldn't be the responsibility of the lib to provide.
> 
> Out of curiosity, why is it _not_ the lib's responsibility to provide. Sure, it's not a property coming from Discord's API directly, but they are useful and save a lot of time when using the library. Removing them makes the library less user friendly and considering the number of beginners that write with D.js, it would mean they are writing more boilerplate that makes the API more complex to use.
> 
> Sure, they are redoable, but imho they are far from useless.

Far from useless? If you think the user can't do something like `this.author.id === this.client.user.id;` as seen in message#editable, then there is a problem. Notice how I'm not targeting any getters that have anything more complex than a one line check, so your point about user friendliness doesn't really make much sense. Users should be able to understand how dAPI permissions work, and by abstracting them in this manner it'll only serve to confuse them later on. Beginners shouldn't rely on getters to hold their hand and do the thinking for them.

typings/index.d.ts
```diff
@@ -1610,10 +1605,8 @@ declare module 'discord.js' {
   export class VoiceChannel extends GuildChannel {
     constructor(guild: Guild, data?: object);
     public bitrate: number;
-    public readonly editable: boolean;
```
Wittiest, line 1613: Note: At the moment, your PR **does not** remove `editable` from src/structures/VoiceChannel.js.

alii:
> Far from useless? If you think the user can't do something like `this.author.id === this.client.user.id;` as seen in message#editable, then there is a problem. Notice how I'm not targeting any getters that have anything more complex than a one line check, so your point about user friendliness doesn't really make much sense. Users should be able to understand how dAPI permissions work, and by abstracting them in this manner it'll only serve to confuse them later on. Beginners shouldn't rely on getters to hold their hand and do the thinking for them.

I'm not convinced. One of the reasons d.js is so popular is the fact that it's easy to use and does a lot of the work for you. Learning how things work under the hood is great, but it should still be left to the library. Writing the following code is much more verbose than a simple getter. Sure `.editable` is very small and the simple equality check can be implemented, but all of a sudden the code is no longer using English terminology and it becomes harder to read.

Take the following
```ts
if (message.editable) {
    await message.edit("Hey, listen!").catch(() => console.warn("Uh oh..."));
}
```

It's readable, declarative, and the developer can easily see that this codeblock will only run if the message is editable. This is the original reason for the implementation of this getter.

Now take the example you are proposing
```ts
if (message.author.id === client.user.id) {
    await message.edit("Hey, listen!").catch(() => console.warn("Uh oh..."));
}
```

All of a sudden, the readability is lost and it takes sec to figure out what this codeblock does and what the equality check is for.

Now, `.editable` is a very bad example, but I wanted to raise it since you did too. What if we did GuildChannel#deletable

```ts
if (channel.deletable) {
    // Great, we can delete it
}
```
Versus the overly verbose proposal
```ts
if (channel.permissionsFor(client.user).has(Permissions.FLAGS.MANAGE_CHANNELS, false)) {
    // Great, we can delete it
}
```

The latter is much more difficult to read, especially for those starting out. It's a lot of boilerplate for no real reason. I can scan over the first `.deletable` and see **exactly** what the `if` is checking for. With the latter it takes a while to deduce what is happening. What is the last `false`? Why `MANAGE_CHANNELS`? etc...

The `.permissionsFor` API still exists for people who would like to use it and opt-in to the more laborious methods, but I can't see why it would be favourable. I also agree with Robert's point of having an embed builder. It's there because it's useful and abstracts a lot of boilerplate code.

Finally, I've seen a few people mention performance in the Discord.js server. I agree with @Fyko's take that can be found here: https://discord.com/channels/222078108977594368/222197033908436994/806224983055532062

PR #5289, opened 2021-02-02
Added user.asMention() method

tobyguelly:
When I used to code with discord.js I always looked for a simple way to create a user mention.
The toString method does not cover all usages, so I thought a user.asMention() method would be great to create a simple mention instead of manually adding pointing brackets for getting a mention string.
It should be added because it is simple to use and simple to find.


almostSouji:
<https://discord.js.org/#/docs/main/stable/class/User?scrollTo=toString>

PR #5290, opened 2021-02-02
Added asMention Getter

tobyguelly:
When I was coding with discord.js, I was looking for a ``asMention`` getter or method for User objects.
The Getter I created is a simple alternative way for creating a mention of a user, which is also easier to find than the ``.toString()`` method.
In my opinion this should be used to make code when creating a mention for a user better readable, and understandable.
This also fits in any situation, and can be used instead of getting the ID and putting an @ and pointing brackets around it.


almostSouji:
duplicate of #5289

PR #5296, opened 2021-02-05
feat(Message): add ReplyMessageOptions for #reply

iShibi:

1) The `options` parameter for `Message#reply` has a `replyTo` field. This field later gets set to `this` inside the method. This begs the question that why even expose the `replyTo` option to the end-user at all in this case. This PR creates a new type `ReplyMessageOptions` that replaces `MessageOptions` parameter for `Message#reply` method. This new `type` for `options` has all the fields that are in `MessageOptions` except the `replyTo` field.

2) Docs show the default value for `repliedUser` in `MessageMentionOptions` as `none`, but `Message#reply` does ping the author of the referenced message by default. The PR also fixes that.


typings/index.d.ts
```diff
@@ -3189,6 +3189,8 @@ declare module 'discord.js' {
     maxUsers?: number;
   }
 
+  interface ReplyMessageOptions extends Omit<MessageOptions, 'replyTo'> {}
```
NotSugden, line 3192:
```suggestion
  type ReplyMessageOptions = Omit<MessageOptions, 'replyTo'>;
```

iShibi: I'm quite new to TS. Will this change make any (substantial) difference or is it just a preference thing?

advaith1: it's the correct way of doing it but shouldn't make a difference afaik

advaith1: however, we might wanna keep the `interface` code since we'll be adding support for `fail_if_not_exists` which is only for replies, so the {} won't be empty anymore

iShibi: #5298 @ckohen added the new field for `fail_if_not_exist` in `MessageOptions`. Thus, `ReplyMessageOptions` will inherit that field too. So, should we accept this change now?

iShibi: This thing resolved itself with the latest commit 6e1b85f

iCrawl (approved):

vladfrangu (approved):

iShibi:
Just wanna say that when this PR was first created it only removed the `replyTo` field. Then, Discord released the `fail_if_not_exists` option, which made it dependent on #5298. As, a result this PR won't work on its own and will need #5298 to be merged first. This will also help in overwriting the doc changes that were needed in that PR but are redundant because of the changes made here.

kyranet (approved):

SpaceEEC (approved):

iCrawl:
You'll need to rebase this now @iShibi

iShibi:
Rebase done. I hope I did it correctly 🤞

PR #5298, opened 2021-02-06
refactor(MessageOptions): move replyTo to reply#messageReference and add failIfNotExists

ckohen:

Add support for new `fail_if_not_exists` parameter in `message_reference`
ref: https://github.com/discord/discord-api-docs/pull/2572

This also changes the replyTo parameter to be more in line with the API

### DIFF
--- a/src/structures/interfaces/TextBasedChannel.js
+++ b/src/structures/interfaces/TextBasedChannel.js
@@ -64,6 +64,8 @@ class TextBasedChannel {
    * @property {boolean|SplitOptions} [split=false] Whether or not the message should be split into multiple messages if
    * it exceeds the character limit. If an object is provided, these are the options for splitting the message
    * @property {MessageResolvable} [replyTo] The message to reply to (must be in the same channel and not system)
+   * @property {boolean} [replyIfDeleted=false] whether to create a standard message instead of erroring if
+   * the message referenced in replyTo is deleted (Ignored when replyTo is not set)
    */
 
   /**
--- a/src/structures/interfaces/TextBasedChannel.js
+++ b/src/structures/interfaces/TextBasedChannel.js
@@ -102,7 +102,7 @@ class TextBasedChannel {
    * Options for sending a message with a reply.
    * @typedef {Object} ReplyOptions
    * @param {MessageResolvable} messageReference The message to reply to (must be in the same channel and not system)
-   * @param {boolean} [failIfNotExists=true] whether to error if the referenced message
+   * @param {boolean} [failIfNotExists=true] Whether to error if the referenced message
    * does not exist (creates a standard message in this case when false)
    */
 
--- a/src/structures/interfaces/TextBasedChannel.js
+++ b/src/structures/interfaces/TextBasedChannel.js
@@ -102,8 +102,8 @@ class TextBasedChannel {
    * Options for sending a message with a reply.
    * @typedef {Object} ReplyOptions
    * @param {MessageResolvable} messageReference The message to reply to (must be in the same channel and not system)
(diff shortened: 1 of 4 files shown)


iShibi:
Nice 👍, I was just thinking that shouldn't the field name emphasize more on the `error` part. The API calls it `fail_if_not_exists`, which makes it clear that something will fail. Even though our description of this field states that it will error, the name doesn't sound like an error will occur.

iShibi:
I would suggest the name `errorIfDeleted` for this field. The reason being that:
1) It emphasizes the error part and conveys the meaning of the field better.
2) It will be in line with the API. Right now we are flipping the booleans and because of the field name, our default for this field is opposite of the API's default. The `errorIfDeleted` can be documented to be `true` by default just like `fail_if_not_exists`.

ckohen:
My only issue with `errorIfDeleted` is that it doesn't indicate it only applies to replies and is a confusing name when not in that context. I will consider this though

vaporox:
`deleted` shouldn't be included in the name at all since passing an invalid id would also have different behavior depending on the parameter. I think that `errorOnInvalidReply` or similar would maybe be a better name here

almostSouji:
So far we've always decided to stick with the API names for options, I'd personally vote for this specific one to not become an exception.

ckohen:
Hmm, I would like to stick with API names as there is clearly a lot of thought in creating these names, but that poses the problem that the library doesn't expose message_reference whatsoever. This means that `MessageOpitons#failIfNotExists` has no indication that it is reply specific other than the jsdoc string. 

Maybe this needs to wait until #5296  lands and have that name? Even then I still don't like that it has no indication of being part of message_reference.

monbrey:
> 
> 
> Hmm, I would like to stick with API names as there is clearly a lot of thought in creating these names, but that poses the problem that the library doesn't expose message_reference whatsoever. This means that `MessageOpitons#failIfNotExists` has no indication that it is reply specific other than the jsdoc string.
> 
> Maybe this needs to wait until #5296 lands and have that name? Even then I still don't like that it has no indication of being part of message_reference.

Replies haven't been released in a stable release yet. If you want to rename `MessageOptions#replyTo` to `messageReference` I don't see the issue with that.

ckohen:
Yep, Souji and I talked about it in the discord, and this is going to end up being a refactor + add

SpaceEEC (changes requested):

src/structures/interfaces/TextBasedChannel.js
```diff
@@ -98,6 +98,14 @@ class TextBasedChannel {
    * @property {string} [append=''] Text to append to every piece except the last
    */
 
+  /**
+   * Options for sending a message with a reply.
+   * @typedef {Object} ReplyOptions
+   * @param {MessageResolvable} messageReference The message to reply to (must be in the same channel and not system)
+   * @param {boolean} [failIfNotExists=true] whether to error if the referenced message
```
iShibi, line 105:
```suggestion
   * @param {boolean} [failIfNotExists=true] Whether to error if the referenced message
```
jsdoc descriptions of function params start with an uppercase letter everywhere in the lib.

ckohen: 🤦 thanks!

vladfrangu (approved):

kyranet (approved):

iShibi:
⚠️ Typings for `WebhookMessageOptions` will need another update (see #5476) once this PR merges.

iCrawl (approved):

SpaceEEC (approved):

PR #5312, opened 2021-02-10
fix(GuildMember): correctly check for premium_since

NotSugden:
This PR fixes a bug introduced by #4636, `GuildMember#premiumSinceTimestamp` was being set to 0 every time rather than null

### DIFF
--- a/src/structures/GuildMember.js
+++ b/src/structures/GuildMember.js
@@ -85,7 +85,7 @@ class GuildMember extends Base {
 
     if ('nick' in data) this.nickname = data.nick;
     if ('joined_at' in data) this.joinedTimestamp = new Date(data.joined_at).getTime();
-    if ('premium_since' in data) this.premiumSinceTimestamp = new Date(data.premium_since).getTime();
+    if (data.premium_since) this.premiumSinceTimestamp = new Date(data.premium_since).getTime();
     if ('roles' in data) this._roles = data.roles;
     this.pending = data.pending ?? false;
   }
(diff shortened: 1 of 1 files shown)


Fyko (approved):

papaia (approved):

kyranet (approved):

SpaceEEC:
How does Discord propagate this property changing to `null`? (i.e. when the member stops boosting)

kyranet (approved):

vladfrangu (approved):

NotSugden:
> How does Discord propagate this property changing to `null`? (i.e. when the member stops boosting)

not sure, I would assume though that `premium_since` would be null so i've updated that to take that into account, no way of really testing this though

SpaceEEC (approved):
Let's hope for the best then.

src/structures/GuildMember.js
```diff
@@ -85,7 +85,9 @@ class GuildMember extends Base {
 
     if ('nick' in data) this.nickname = data.nick;
     if ('joined_at' in data) this.joinedTimestamp = new Date(data.joined_at).getTime();
-    if ('premium_since' in data) this.premiumSinceTimestamp = new Date(data.premium_since).getTime();
+    if ('premium_since' in data) {
+      this.premiumSinceTimestamp = data.premium_since === null ? null : new Date(data.premium_since).getTime();
+    }
```
vaporox, line 90:
```suggestion
    if ('premium_since' in data) this.premiumSinceTimestamp = data.premium_since && new Date(data.premium_since).getTime();
```

iCrawl (approved):

PR #5318, opened 2021-02-12
feat: PermissionOverwriteManager

extroonie:
This PR introduces `PermissionOverwriteManager`, to manage guild channel permission overwrites and store their cache. This was brought up as an approach to https://github.com/discordjs/discord.js/pull/5234#issuecomment-764658517. It also removes the existing permission overwrite methods from `GuildChannel`, it is, therefore, **semver major**. Some of the logic is ~~stolen~~ taken from #5489 and merged here as well.


src/managers/PermissionOverwriteManager.js
```diff
@@ -0,0 +1,137 @@
...
+   *   SEND_MESSAGES: false
+   * })
+   *   .then(channel => console.log(channel.permissionOverwrites.cache.get(message.author.id)))
+   *   .catch(console.error);
+   */
+  async edit(userOrRole, options, reason) {
+    userOrRole = this.channel.guild.roles.resolve(userOrRole) ?? this.client.users.resolve(userOrRole);
+    if (!userOrRole) throw new TypeError('INVALID_TYPE', 'parameter', 'User nor a Role');
+
+    const type = userOrRole instanceof Role ? 'role' : 'member';
```
NotSugden, line 74: use OverwriteTypes.role & OverwriteTypes.member

src/managers/PermissionOverwriteManager.js
```diff
@@ -0,0 +1,137 @@
...
+   *   SEND_MESSAGES: false
+   * })
+   *   .then(channel => console.log(channel.permissionOverwrites.cache.get(message.author.id)))
+   *   .catch(console.error);
+   */
+  async create(userOrRole, options, reason) {
+    userOrRole = this.channel.guild.roles.resolve(userOrRole) ?? this.client.users.resolve(userOrRole);
+    if (!userOrRole) throw new TypeError('INVALID_TYPE', 'parameter', 'User nor a Role');
+
+    const type = userOrRole instanceof Role ? 'role' : 'member';
```
NotSugden, line 109: same here

typings/index.d.ts
```diff
@@ -3014,6 +3036,8 @@ declare module 'discord.js' {
 
   type PermissionResolvable = BitFieldResolvable<PermissionString, bigint>;
 
+  type PermissionOverwriteResolvable = Snowflake | PermissionOverwrites;
```
NotSugden, line 3039: why not also add `GuildMember`, `User`, and `Role` types here too (could also use the respective `Resolvable` types too i suppose

NotSugden (changes requested):

typings/index.d.ts
extroonie: out of the scope of this pr, perhaps a separate one (?)

NotSugden: this type is added in this PR, it would make sense to resolve a `PermissionOverwrite` from the types above, so why not add that functionality right off the bat ?

extroonie: oh nvm, i misread what you said, disregard. will do, good catch, thanks!

src/managers/PermissionOverwriteManager.js
```diff
@@ -0,0 +1,138 @@
...
+const PermissionOverwrites = require('../structures/PermissionOverwrites');
+const Role = require('../structures/Role');
+const Collection = require('../util/Collection');
+const { OverwriteTypes } = require('../util/Constants');
+
+/**
+ * Manages API methods for guild channel permission overwrites and stores their cache.
+ * @extends {BaseManager}
+ */
+class PermissionOverwriteManager extends BaseManager {
```
tguichaoua, line 14: `PermissionOverwriteManager#set`, `edit`, `create` and `delete` should return `this` to allow chaining.

tguichaoua (changes requested):

src/managers/PermissionOverwriteManager.js
extroonie: It's not consistent; and chaining can still be done anyway, so I don't really think so that'd be necessary.

tguichaoua: When these methods were defined in `GuildChannel`, it was relevant to return the `this`, to allow chaining. But since they are now defined in the manager, it don't make sense to return the `GuildChannel` object. At least they should return `void` or a more relevant value.

extroonie: Right, but neither of the API requests to `/channels/:id/permissions` (except `POST`) return permission overwrites - which would be even more consistent if the API did so. And regarding to your point, neither of the managers have this behavior: `channel.permissionOverwrites.create(...).delete(..).set(..)`, so I'm kind of against this inconsistency. Although, returning `GuildChannel` reasonably makes sense and is quite relevant.

src/managers/PermissionOverwriteManager.js
```diff
@@ -0,0 +1,138 @@
...
+   *      id: message.author.id,
+   *      deny: [Permissions.FLAGS.VIEW_CHANNEL],
+   *   },
+   * ], 'Needed to change permissions');
+   */
+  async set(overwrites, reason) {
+    if (!Array.isArray(overwrites) && !(overwrites instanceof Collection)) {
+      throw new TypeError('INVALID_TYPE', 'overwrites', 'Array or Collection of Permission Overwrites', true);
+    }
+    await this.channel.edit({ permissionOverwrites: overwrites, reason });
```
vladfrangu, line 53: Doesn't this call return the channel anyways?

src/managers/PermissionOverwriteManager.js
```diff
@@ -0,0 +1,138 @@
...
+   * @returns {Promise<GuildChannel>}
+   * @example
+   * message.channel.permissionOverwrites.set([
+   *   {
+   *      id: message.author.id,
+   *      deny: [Permissions.FLAGS.VIEW_CHANNEL],
+   *   },
+   * ], 'Needed to change permissions');
+   */
+  async set(overwrites, reason) {
```
vladfrangu, line 49: Personally, I'd be more of a fan of calling this `overwrite` than `set`, since we're replacing everything. CC @iCrawl @kyranet tho

src/managers/PermissionOverwriteManager.js
```diff
@@ -0,0 +1,138 @@
...
+   *   SEND_MESSAGES: false
+   * })
+   *   .then(channel => console.log(channel.permissionOverwrites.cache.get(message.author.id)))
+   *   .catch(console.error);
+   */
+  async edit(userOrRole, options, reason) {
+    userOrRole = this.channel.guild.roles.resolve(userOrRole) ?? this.client.users.resolve(userOrRole);
+    if (!userOrRole) throw new TypeError('INVALID_TYPE', 'parameter', 'User nor a Role');
+
+    const type = userOrRole instanceof Role ? OverwriteTypes.role : OverwriteTypes.member;
```
vladfrangu, line 75: Why not let us specify the type too, and support user/role IDs?

src/managers/PermissionOverwriteManager.js
```diff
@@ -0,0 +1,138 @@
...
+    if (existing) {
+      await this.client.api
+        .channels(this.channel.id)
+        .permissions(userOrRole.id)
+        .put({
+          data: { id: userOrRole.id, type, allow, deny },
+          reason,
+        });
+      return this.channel;
+    }
```
vladfrangu, line 88: This whole check is redundant, and is 1:1 with what `this.create` would do.
In fact, this whole *function* is a duplicate of the create function. Might as well make `edit` call `create`, or remove `edit` and rename `create` to `upsert` or similar

src/managers/PermissionOverwriteManager.js
```diff
@@ -0,0 +1,138 @@
...
+   *   SEND_MESSAGES: false
+   * })
+   *   .then(channel => console.log(channel.permissionOverwrites.cache.get(message.author.id)))
+   *   .catch(console.error);
+   */
+  async create(userOrRole, options, reason) {
+    userOrRole = this.channel.guild.roles.resolve(userOrRole) ?? this.client.users.resolve(userOrRole);
+    if (!userOrRole) throw new TypeError('INVALID_TYPE', 'parameter', 'User nor a Role');
+
+    const type = userOrRole instanceof Role ? OverwriteTypes.role : OverwriteTypes.member;
```
vladfrangu, line 110: ditto as above

src/managers/PermissionOverwriteManager.js
```diff
@@ -0,0 +1,138 @@
...
+  }
+
+  /**
+   * Deletes permission overwrites for a user or role in this channel
+   * @param {UserResolvable|RoleResolvable} userOrRole The user or role to delete
+   * @param {string} [reason] The reason for deleting the overwrite
+   * @returns {GuildChannel}
+   */
+  async delete(userOrRole, reason) {
+    userOrRole = this.channel.guild.roles.resolve(userOrRole) ?? this.client.users.resolve(userOrRole);
```
vladfrangu, line 130: Should use IDs, not full objects. That way, for example, we can pass in an ID from a db without needing to first fetch the user

typings/index.d.ts
```diff
@@ -2058,6 +2058,34 @@ declare module 'discord.js' {
     public delete(message: MessageResolvable): Promise<void>;
   }
 
+  // Hacky workaround because changing the signature of an overridden method errors
+  class OverridableManager<V, K, R = any> extends BaseManager<V, K, R> {
```
vladfrangu, line 2062: Is this even used anywhere?

typings/index.d.ts
```diff
@@ -1217,14 +1217,14 @@ declare module 'discord.js' {
     public iconURL(options?: ImageURLOptions): string | null;
   }
 
-  export class PermissionOverwrites {
-    constructor(guildChannel: GuildChannel, data?: object);
+  export class PermissionOverwrites extends Base {
```
vladfrangu, line 1220: Not related to this line, but to this file, update GuildChannel#permissionOverwrites types too

vladfrangu (changes requested):

typings/index.d.ts
extroonie: Done.

src/managers/PermissionOverwriteManager.js
```diff
@@ -0,0 +1,149 @@
...
+    let { type, reason } = overwriteOptions;
+    if (typeof type !== 'number') {
+      userOrRole = this.channel.guild.roles.resolve(userOrRole) ?? this.client.users.resolve(userOrRole);
+      if (!userOrRole) return Promise.reject(new TypeError('INVALID_TYPE', 'parameter', 'User nor a Role'));
+      type = userOrRole instanceof Role ? OverwriteTypes.role : OverwriteTypes.member;
+    }
+
+    const { allow, deny } = existing
+      ? PermissionOverwrites.resolveOverwriteOptions(options)
+      : PermissionOverwrites.resolveOverwriteOptions(options, existing);
```
vladfrangu, line 83:
```suggestion
    const { allow, deny } = PermissionOverwrites.resolveOverwriteOptions(options, existing);
```

src/managers/PermissionOverwriteManager.js
```diff
@@ -0,0 +1,149 @@
...
+   *   SEND_MESSAGES: false
+   * })
+   *   .then(channel => console.log(channel.permissionOverwrites.cache.get(message.author.id)))
+   *   .catch(console.error);
+   */
+  edit(userOrRole, options, overwriteOptions) {
+    userOrRole = this.channel.guild.roles.resolveID(userOrRole) ?? this.client.users.resolveID(userOrRole);
+    const existing = this.cache.get(userOrRole);
+    if (existing) return this.upsert(userOrRole, options, overwriteOptions, existing);
+    return this.create(userOrRole, options, overwriteOptions);
```
vladfrangu, line 131:
```suggestion
    return this.upsert(userOrRole, options, overwriteOptions, existing);
```

vladfrangu (changes requested):

src/structures/PermissionOverwrites.js
```diff
@@ -48,33 +51,20 @@ class PermissionOverwrites {
...
-        data: {
-          id: this.id,
-          type: OverwriteTypes[this.type],
-          allow,
-          deny,
-        },
-        reason,
-      });
+  async edit(options, reason) {
+    await this.channel.permissionOverwrites.edit(this.id, options, reason);
```
vladfrangu, line 67: Time to skip checks! Call permissionOverwrites.upsert with the `id, options, { type: this.type (but number), reason }, this`

vladfrangu (changes requested):

src/managers/PermissionOverwriteManager.js
extroonie: Discussed in the Discord server and it was rather more consistent with `set` (e.g. `GuildMember#roles#set`, etc.)

vladfrangu (approved):

kyranet (approved):

iCrawl:
This needs another rebase.

extroonie:
@iCrawl done

iCrawl:
This needs a rebase.

extroonie:
@iCrawl done x2

vladfrangu:
You wouldn't get this from any other guy, but this needs another rebase 🎶

extroonie:
@vladfrangu done x3

src/structures/GuildChannel.js
```diff
@@ -67,14 +66,14 @@ class GuildChannel extends Channel {
       this.parentID = data.parent_id;
     }
 
+    /**
+     * A manager of permission overwrites that belong to this channel
+     * @type {PermissionOverwriteManager}
+     */
+    this.permissionOverwrites = new PermissionOverwriteManager(this);
```
SpaceEEC, line 73: Why is this being moved outside of the guard?
This will cause all overwrites to be deleted whenever a channel is patched with partial data.
See: https://github.com/discordjs/discord.js/issues/5605

src/structures/PermissionOverwrites.js
```diff
@@ -8,15 +9,17 @@ const Permissions = require('../util/Permissions');
 /**
  * Represents a permission overwrite for a role or member in a guild channel.
  */
```
SpaceEEC, line 11:
```suggestion
 * @extends {Base}
 */
```

typings/index.d.ts
```diff
@@ -2295,6 +2281,34 @@ declare module 'discord.js' {
     public unpin(message: MessageResolvable): Promise<void>;
   }
 
+  export class PermissionOverwriteManager extends BaseManager<Snowflake, PermissionOverwrites, PermissionOverwriteResolvable> {
+    constructor(client: Client, iterable?: Iterable<any>);
+    public set(
+      overwrites: readonly OverwriteResolvable[] | Collection<Snowflake, OverwriteResolvable>,
+      reason?: string,
+    ): Promise<GuildChannel>;
+    public upsert(
```
SpaceEEC, line 2290: See my comment above.

SpaceEEC (dismissed):

src/structures/GuildChannel.js
extroonie: That is intended to be moved outside the guard as this is a manager now, and now with the latest commit, I moved it to the constructor.

DraftProducts (approved):

src/structures/GuildChannel.js
```diff
@@ -37,7 +36,11 @@ class GuildChannel extends Channel {
     this.guild = guild;
 
     this.parentID = this.parentID ?? null;
-    this.permissionOverwrites = this.permissionOverwrites ?? new Collection();
+    /**
+     * A manager of permission overwrites that belong to this channel
+     * @type {PermissionOverwriteManager}
+     */
+    this.permissionOverwrites = new PermissionOverwriteManager(this);
```
SpaceEEC, line 43: This line will be executed _after_ the base class called `_patch`, causing said method to throw since the permission overwrite manager isn't set yet.

If you rebase this PR onto master, you can pass a third parameter (`immediatePatch`) to the super call causing the base class not to call `_patch`, allowing you to call it here yourself after the manager has been set.

src/structures/GuildChannel.js
```diff
@@ -68,13 +71,8 @@ class GuildChannel extends Channel {
     }
 
     if ('permission_overwrites' in data) {
-      /**
-       * A map of permission overwrites in this channel for roles and users
-       * @type {Collection<Snowflake, PermissionOverwrites>}
-       */
-      this.permissionOverwrites = new Collection();
```
SpaceEEC, line 75: I can't exactly test this thesis now due the issue pointed out by my above comment, but the manager's cache should be `clear`-ed here to actually remove removed overwrites.

SpaceEEC (changes requested):

kyranet:
This needs a rebase + Space's requested changes addressed.

SpaceEEC (approved):

PR #5321, opened 2021-02-13
fix(Role): pass Permissions class, not the bitfield

vaporox:

The `toJSON()` method from the `Permissions` class returns the bitfield as a string, however the bitfield is passed directly here, resulting in `TypeError: Do not know how to serialize a BigInt`

- Typings don't need updating

### DIFF
--- a/src/structures/Role.js
+++ b/src/structures/Role.js
@@ -197,8 +197,6 @@ class Role extends Base {
    *   .catch(console.error);
    */
   async edit(data, reason) {
-    if (typeof data.permissions !== 'undefined') data.permissions = new Permissions(data.permissions);
-    else data.permissions = this.permissions;
     if (typeof data.position !== 'undefined') {
       await Util.setPosition(
         this,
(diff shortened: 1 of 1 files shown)


vladfrangu (approved):

src/structures/Role.js
```diff
@@ -197,8 +197,8 @@ class Role extends Base {
    *   .catch(console.error);
    */
   async edit(data, reason) {
-    if (typeof data.permissions !== 'undefined') data.permissions = Permissions.resolve(data.permissions);
-    else data.permissions = this.permissions.bitfield;
+    if (typeof data.permissions !== 'undefined') data.permissions = new Permissions(data.permissions);
+    else data.permissions = this.permissions;
```
izexi, line 201: Might as well remove the `if`/`else` and pass this directly within the `data` object [below](https://github.com/vaporox/discord.js/blob/fix-edit/src/structures/Role.js#L223) (this would also avoid any unwanted mutation)
```diff
- permissions: data.permissions,
+ permissions: typeof data.permissions !== 'undefined' ? new Permissions(data.permissions) : this.permissions
```

izexi (approved):

kyranet (approved):

SpaceEEC (approved):

iCrawl (approved):

PR #5324, opened 2021-02-14
fix(GuildManager): add missing toString() on Permission#resolve fns

Fyko:
This PR adds missing `toString` functions on various `Permissions#resolve(...)` functions within GuildManager#create.
This should be treated with high severity.


SpaceEEC (approved):

kyranet (approved):
An alternative to this would have been doing `new Permissions()` since its `toJSON()` serializes the data correctly, but this also works.

iCrawl (approved):

Issue #95, opened 2015-12-12
With every new sound being streamed, a new FFMPEG process is made.


XNBlank:
Everytime playFile() is called, a new process for FFMPEG is created, and old ones are not destroyed.

hydrabolt:
Should be fixed in 4d438672061f1a9202418c740e5a86d7f2769622

hydrabolt:
Was actually fixed in https://github.com/hydrabolt/discord.js/commit/c2cf1badfdf2d88e90e7b0e59bf684bd913627c0

PR #5331, opened 2021-02-16
line  #95

gishman:
should be       if (guild.members) {

guild.members can be empty and throws and uncaught exception

### DIFF
--- a/src/client/actions/Action.js
+++ b/src/client/actions/Action.js
@@ -92,7 +92,7 @@ class GenericAction {
   getUserFromMember(data) {
     if (data.guild_id && data.member && data.member.user) {
       const guild = this.client.guilds.cache.get(data.guild_id);
-      if (guild) {
+           if (guild.members) {
         return guild.members.add(data.member).user;
       } else {
         return this.client.users.add(data.member.user);


MattIPv4:
The `members` property is defined here, during the creation of the instance of `Guild`. I'm not really sure how you can end up in a situation where it is undefined, unless you're modifying the structure itself?

https://github.com/discordjs/discord.js/blob/ee5bc1a5c4e218ea1f16992694a2bff86c86d2d2/src/structures/Guild.js#L50

Your change in its current form will also lead to undefined errors, as `guild` here may be undefined if the guild wasn't found in the cache. If there is actually a reason to be checking the presence of the `members` property, you'll want to use the optional chaining operator, so we can safely check for the presence of the `guild` and `guild.members` at the same time: `if (guild?.members) {`

gishman:
Ok, but I kept getting this error once a typing even started on my bot launch

gishman:
> The `members` property is defined here, during the creation of the instance of `Guild`. I'm not really sure how you can end up in a situation where it is undefined, unless you're modifying the structure itself?
> 
> https://github.com/discordjs/discord.js/blob/ee5bc1a5c4e218ea1f16992694a2bff86c86d2d2/src/structures/Guild.js#L50
> 
> Your change in its current form will also lead to undefined errors, as `guild` here may be undefined if the guild wasn't found in the cache. If there is actually a reason to be checking the presence of the `members` property, you'll want to use the optional chaining operator, so we can safely check for the presence of the `guild` and `guild.members` at the same time: `if (guild?.members) {`

I'm not calling anything, but when the Typing event is fired, it throws this error

PR #5332, opened 2021-02-16
docs(Message): add example for message.reply

Rexogamer:
This PR adds an example for `message.reply`.

### DIFF
--- a/src/structures/Message.js
+++ b/src/structures/Message.js
@@ -577,7 +577,7 @@ class Message extends Base {
    * @example
    * // Reply to a message
    * message.reply('This is a reply!')
-   *   .then(console.log(`Replied to message "${message.content}"`))
+   *   .then(() => console.log(`Replied to message "${message.content}"`))
    *   .catch(console.error);
    */
   reply(content, options) {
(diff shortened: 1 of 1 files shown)


src/structures/Message.js
```diff
@@ -574,6 +574,11 @@ class Message extends Base {
    * @param {MessageOptions|MessageAdditions} [options] The additional options to provide
    * @param {MessageResolvable} [options.replyTo=this] The message to reply to
    * @returns {Promise<Message|Message[]>}
+   * @example
+   * // Reply to a message
+   * message.reply('This is a reply!')
+   *   .then(msg => console.log(`Replied to message "${msg.content}"`))
```
Rexogamer, line 580: Is this the right info to log?

advaith1: this would log `Replied to message "This is a reply!"`, is that intentional?

Rexogamer: That was what I was intending, although I wasn't sure if it was the right thing to log

Rexogamer: Fixed in ad778e8e046f2abec3bb1a90ad20a695359551e0

src/structures/Message.js
```diff
@@ -574,6 +574,11 @@ class Message extends Base {
    * @param {MessageOptions|MessageAdditions} [options] The additional options to provide
    * @param {MessageResolvable} [options.replyTo=this] The message to reply to
    * @returns {Promise<Message|Message[]>}
+   * @example
+   * // Reply to a message
+   * message.reply('This is a reply!')
+   *   .then(console.log(`Replied to message "${message.content}"`))
```
papaia, line 580:
```suggestion
   *   .then(() => console.log(`Replied to message "${message.content}"`))
```

SpaceEEC (approved):

PR #5345, opened 2021-02-22
fix: emit non-cached MessageDelete events

ChipWolf:

This change will ensure the `MessageDelete` event will fire in the event a message has not been cached.
This effectively makes `message` nullable in the `MessageDelete` context.

### DIFF
--- a/src/client/actions/MessageDelete.js
+++ b/src/client/actions/MessageDelete.js
@@ -13,13 +13,13 @@ class MessageDeleteAction extends Action {
       if (message) {
         channel.messages.cache.delete(message.id);
         message.deleted = true;
-        /**
-         * Emitted whenever a message is deleted.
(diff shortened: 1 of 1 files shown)


NotSugden:
this can already be achieved by enabling the `MESSAGE` partial, and you actually get an ID for the deleted message then unlike this

almostSouji:
Resources regarding partials:
- https://discord.js.org/#/docs/main/stable/topics/partials
- https://discordjs.guide/popular-topics/partials.html
- https://discordjs.guide/popular-topics/reactions.html#listening-for-reactions-on-old-messages

ChipWolf:
Found [it](https://github.com/discordjs/discord.js/blob/master/docs/topics/partials.md), thanks.

PR #5362, opened 2021-03-01
fix(Constants.Colors): change hex value of white to resolve on desktop

vBread:
This PR changes the value of `Colors.White` from `0xFFFFFF` to `0xFEFEFE` so it renders properly on desktop (mobile doesn't have this issue).

Screen Shot 2021-02-28 at 4 01 24 PM

### DIFF
--- a/src/util/Constants.js
+++ b/src/util/Constants.js
@@ -486,7 +486,7 @@ exports.ClientApplicationAssetTypes = {
 
 exports.Colors = {
   DEFAULT: 0x000000,
-  WHITE: 0xffffff,
+  WHITE: 0xfefefe,
   AQUA: 0x1abc9c,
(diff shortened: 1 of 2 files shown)


almostSouji:
Personally not in favor of making white not actually white just because discord decides to be weird, relevant comment in a similar issue:
https://github.com/discordjs/discord.js/pull/3841#issuecomment-600413827

vBread:
Would it be better to add some sort of note to the docs then or is it not that important?

TheNoob27:
> making white not actually white

`#fefefe` or `#fffffe` is practically still white, it's just not _white_ white. The difference isn't that noticeable, so a simple small change like this to make `Colors.WHITE` actually useful to more people shouldn't really be a bad thing

PR #5365, opened 2021-03-02
feat(Guild): add Guild#displayIcon() method

vBread:
This PR adds a `displayIcon()` method to the Guild object for convenience, similar to [`User#displayAvatarURL()`](https://discord.js.org/#/docs/main/stable/class/User?scrollTo=displayAvatarURL).

### DIFF
--- a/src/structures/Guild.js
+++ b/src/structures/Guild.js
@@ -525,6 +525,16 @@ class Guild extends Base {
   }
 
   /**
+   * The URL to this guild's icon if it has one.
+   * Otherwise the guild's acronym will be returned.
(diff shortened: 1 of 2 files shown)


monbrey:
I think this is a pretty misleading method because it _isn't_ similar to `User#displayAvatarURL()`.
This implies that a default URL can be obtained for the acronym and used in places a real icon can be, which isn't the case at all.

Not sure where this method would be useful.

vBread:
> This implies that a default URL can be obtained for the acronym and used in places a real icon can be.

Fair; didn't actually realize how that could be interpreted. Thinking about the actual use cases now are also pretty niche... I'll close this then 👍

Issue #5350, opened 2021-02-23
GuildDelete event triggers every time bot starts


DomDom3333:
**Please describe the problem you are having in as much detail as possible:**
Every time I start the bot, the guildDelete event triggers. However while the guild object is valid, the name and member count is undefined. This results in an error.

**Include a reproducible code sample here, if possible:**
Can't be reproduced without this specific Token. May be related to Discord itself. Starting the bot with a different token fixes the issue.

```js
bot.on("guildDelete", guild => {
    LOG.info(`Got booted from a Server!!!!!!!!\n- Name: ${guild.name}\n- Member Count: ${guild.memberCount}\nI'm now in ${bot.guilds.cache.size} Servers!`);
    Utils.contactAdmins(`Got booted from a Server!!!!!!!!\n- Name: ${guild.name}\n- Member Count: ${guild.memberCount}\nI'm now in ${bot.guilds.cache.size} Servers!`);
})
```

**Further details:**

- discord.js version: 12.4.1
- Node.js version: 12.18.1
- Operating system: Win10
- Priority this issue should have – please be realistic and elaborate if possible: Low-Medium

**Relevant client options:**

- partials: none
- gateway intents: none
- other: none

Albert-J04:
I am seeing the same running on Ubuntu 18.04

almostSouji:
12.4.1 is no longer supported, please check 12.5.1 for the same issue.
If you can reproduce it add a listener for the websocket event and provide the output:

```js
client.ws.on("GUILD_DELETE", console.log);
```

MidSpike:
**Can reproduce** on `12.5.1`

Output that @almostSouji requested:
```js
{ id: 'censored_guild_id' } 0
```

Good thing that I keep track of guild add / removes!

The guild in question actually removed my bot around 2021-02-24...

However every time that I start the bot, **only** that guild will have a `GuildDelete` trigger for it.

almostSouji:
I actually did not expect that to log anything. Since that's the case this behavior is coming directly from Discord and there is nothing we can do about it. Discord devs are aware but currently unsure about why it happens.

Natemo6348:
If you want to prevent it from, say sending the set message on guildDelete like I have, you could add something like 
```
if (guild.name === undefined) return;
```

Ortovoxx:
I also get the same error. This should be fixed with #5367 

> Good thing that I keep track of guild add / removes!
> The guild in question actually removed my bot around 2021-02-24...

[...]

PR #5367, opened 2021-03-02
Added check for if guild is unavailable on guild object

Ortovoxx:
**guildDelete event fires on startup due to an unavailable guild but not the guildUnavailable event**

There is a guild in my bot which is unavailable ( sitenote: I am pretty certain that this my bot user left this guild about a week or so ago ) and everytime the client starts up the `guildDelete` event fires and the `guildUnavailable` event does not fire.

I looked into it and it appears that the `GuildDeleteAction` event fires when this occurs but the data object that is sent to it just contains the guild id property: `{ id: '813428000925417513' }` with no `unavailable` property like [documented](https://discord.com/developers/docs/resources/guild#unavailable-guild-object)

However it appears that the guild does have the `available` property found on all guilds so I added an extra check for this property when deciding if to fire the `guildUnavailable` or `guildDelete` event.

I believe that this is an edge case on discords end as I mentioned earlier that I don't think that my bot is in this guild anymore. It has also been unavailable for about a week now.

Adding this change to my local library seemed to fix it. Let me know if I am missing anything here I hope my first PR has been useful!

edit: This will fix the behavior in #5350 which is a Discord bug 

### DIFF
--- a/src/client/actions/GuildDelete.js
+++ b/src/client/actions/GuildDelete.js
@@ -18,7 +18,7 @@ class GuildDeleteAction extends Action {
         if (channel.type === 'text') channel.stopTyping(true);
       }
 
-      if (data.unavailable) {
+      if (data.unavailable || !guild.available) {
         // Guild is unavailable
         guild.available = false;
 


almostSouji:
Not sure yet if this should need to be handled in the library. Discord devs are aware of it happening but have so far not been able to figure out when it occurs. With how the said statement was phrased I'm not yet certain if this should be handled on the library level at all.

Ortovoxx:
Okay that makes sense. Hopefully it get fixed soon.

Ortovoxx:
Reopening because we need to handle in the library and ignore soft deleted guilds. See [this comment](https://github.com/discordjs/discord.js/issues/5350#issuecomment-793726018)

DaliborTrampota:
https://github.com/discordjs/discord.js/issues/5430 hello i have the same issue i  guess

vladfrangu (approved):
LGTM

kyranet (approved):

vladfrangu (changes requested):
What space said

iCrawl:
This PR has been pretty stale with no feedback from the author, hence I am going to close this.

Please reopen with the suggested changes if this is still needed/wanted.

PR #5371, opened 2021-03-04
refactor: from .then() to async/await

alaahouerbi:
Refactoring of .then() into an async/await

Code changes have been tested against the Discord API, or there are no code changes


monbrey:
I would review the fact that you used `var`, but I don't see the point of this PR changing one single instance of `then` in one single file.

If you want to do a piece of work that sweeps across the whole lib to maintain this consistency, sure.

And please use `let / const` if you do.

NotSugden:
#5132 already covers this

PR #5384, opened 2021-03-10
feat: move internal regular expressions to static properties

vaporox:

This moves the regular expressions used in the `DataResolver` class to static properties for external use.

### DIFF
--- a/src/structures/GuildTemplate.js
+++ b/src/structures/GuildTemplate.js
@@ -222,4 +222,10 @@ class GuildTemplate extends Base {
   }
 }
 
+/**
+ * Regular expression that globally matches guild template links
+ * @type {RegExp}
+ */
+GuildTemplate.GUILD_TEMPLATES_PATTERN = /discord(?:app)?\.(?:com\/template|new)\/([\w-]{2,255})/gi;
+
 module.exports = GuildTemplate;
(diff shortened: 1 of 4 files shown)


iCrawl (approved):

vladfrangu (approved):

kyranet (approved):

SpaceEEC:
Using a static global regex with [`RegExp#exec`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/exec) will store a ` lastIndex` on the object which messes with further calls to `exec` on different strings.

We'd either need to clone the regexp before scanning the string or use `String#matchAll`  which seems to be supported on node v14 ([node.green with string matchAll](https://node.green/#ES2020-features-String-prototype-matchAll)) and does that internally. (See [mdn on Regexp.exec() and matchAll()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/matchAll#regexp.exec_and_matchall))

I didn't really test either, so feel free to correct me if I am wrong.

vaporox:
@SpaceEEC I was able to reproduce the problem you mentioned, tested it and it should be fixed now

SpaceEEC (approved):

PR #5385, opened 2021-03-11
docs(GuildChannel): #members add cached spec, include voice

almostSouji:
`GuildChannel#members` currently reads

> A collection of members that can see this channel, mapped by their ID.

- This is not true for `voice` type channels, where it instead reflects the members in the channel (which can differ from members being able to view the channel)
- This does not specify that only cached members are considered (due to it being a convenience wrapper around a `Guild#members` cache filter)

### DIFF
--- a/src/structures/GuildChannel.js
+++ b/src/structures/GuildChannel.js
@@ -287,7 +287,7 @@ class GuildChannel extends Channel {
   }
 
   /**
-   * A collection of members of this channel, mapped by their ID.
+   * A collection of cached members of this channel, mapped by their ID.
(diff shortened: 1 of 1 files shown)


SpaceEEC:
Is it intended that this PR does not resolve the second point in your list?
I imagine that adding the word `cached` in the first sentence would do that.

almostSouji:
That was indeed an oversight, fixed

SpaceEEC (approved):

kyranet (approved):

iCrawl (approved):

PR #5390, opened 2021-03-11
Fixed jsdoc for Role#tags docs showing "optional" field.

BannerBomb:
In the documentation [Role#tags](https://discord.js.org/#/docs/main/master/class/Role?scrollTo=tags) is documented as a method where the "optional" field is displayed while all other object-based properties in the documentations do not. For example [Presence#clientStatus](https://discord.js.org/#/docs/main/master/class/Presence?scrollTo=clientStatus) which is how Role#tags should look.

### DIFF
--- a/src/structures/Role.js
+++ b/src/structures/Role.js
@@ -86,9 +86,9 @@ class Role extends Base {
     /**
      * The tags this role has
      * @type {?Object}
-     * @property {Snowflake} [botID] The id of the bot this role belongs to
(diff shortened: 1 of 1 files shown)


NotSugden:
this should stay how it is no?

https://jsdoc.app/tags-type.html in any case `?type` means that it is `type` or `null`, these properties will just be undefined if they don't exist and not null

BannerBomb:
Ah, I didn't know that. I always thought `?type` meant "null or undefined" and `[param name]` was used for specifying optional method parameters.

PR #5402, opened 2021-03-14
feat(Voice): implement support for @discordjs/voice

amishshah:

This PR allows for creating adapters that can be used by [@discordjs/voice](https://github.com/discordjs/voice/). This PR also removes the existing voice codebase as it will no longer be maintained.

The only voice code that will remain in discord.js will be voice states, which aren't specifically part of the Discord Voice API.

## What is @discordjs/voice?

[@discordjs/voice](https://github.com/discordjs/voice/) is a stand-alone library that implements the Discord Voice API. It aims to offer better stability, performance, and scalability than the existing voice client within Discord.js v12.

Discord.js v13 will remove support for the Voice API, and this functionality can be replaced by installing @discordjs/voice (`npm install @discordjs/voice`) as an add-on.

## Example usage:
```js
const { joinVoiceChannel } = require('@discordjs/voice');

const connection = joinVoiceChannel({
  channelId: channel.id,
  guildId: channel.guild.id,
  adapterCreator: channel.guild.voiceAdapterCreator,
});
```

## How do I update my code to use @discordjs/voice?

@discordjs/voice is nearing its release. If you want to get a head start on migrating your voice code to the library, you should take a look at this PR which shows how to create a music bot with @discordjs/voice and TypeScript - https://github.com/discordjs/voice/pull/107


src/client/voice/ClientVoiceManager.js
```diff
@@ -41,6 +75,27 @@ class ClientVoiceManager {
     return broadcast;
   }
 
+  /**
+   * Creates a voice adapter for a given channel, to be used by @discordjs/voice.
+   * @param {VoiceChannel} channel The voice channel to create an adapter for.
+   * @returns {DiscordGatewayAdapterCreator}
+   * @private
```
advaith1, line 82: why is this private?

typings/index.d.ts
```diff
@@ -1621,6 +1621,7 @@ declare module 'discord.js' {
     public type: 'voice';
     public userLimit: number;
     public join(): Promise<VoiceConnection>;
+    public createVoiceAdapter(): object;
```
advaith1, line 1624: would it be possible to import the type from `@discordjs/voice` if it is installed?

vladfrangu (changes requested):

typings/index.d.ts
vladfrangu: There's no conditional type imports. We can import it directly and it'll default to `any` if not present (and you use `skipLibCheck` in your tsconfig)

src/client/voice/ClientVoiceManager.js
amishshah: It is preferred to use `channel.createVoiceAdapter` instead.

advaith1: But the library is moving to structure methods mainly calling public methods on the Manager so users can do either way; this is similar to the Manager methods and there doesn't seem to be any disadvantage to users using this

typings/index.d.ts
didinele: > There's no conditional type imports. We can import it directly and it'll default to `any` if not present (and you use `skipLibCheck` in your tsconfig)

This is already the case with `@types/ws` being a dev dep. I see no issue doing the same with `@discordjs/voice`?

src/client/voice/ClientVoiceManager.js
amishshah: The code has been refactored to remove 2 declarations of this

amishshah:
I'm converting this back to a draft as I want to fix https://github.com/discordjs/voice/issues/78 before adapters are added directly to Discord.js.

src/client/voice/ClientVoiceManager.js
```diff
@@ -29,6 +35,36 @@ class ClientVoiceManager {
...
+        const guild = client.guilds.cache.get(guildID);
+        if (guild?.shardID === id) {
+          adapter.destroy();
+          this.adapters.delete(guildID);
+        }
+      }
+    };
+
+    client.on('shardDisconnect', (event, id) => onShardDisconnect(id));
+    client.on('shardReconnecting', id => onShardDisconnect(id));
```
vladfrangu, line 67: So... on reconnecting we destroy it entirely...will it reconnect / recreate after it successfully reconnects?

amishshah: That's up to the developer (if they notice that a particular shard has reconnected, they can respawn any disconnected voice connections if they want to)

vladfrangu (approved):

amishshah:
Converting to a draft, need to update the adapter as well as deprecate current Discord.js voice code.

src/structures/BaseGuildVoiceChannel.js
```diff
@@ -66,30 +66,6 @@ class BaseGuildVoiceChannel extends GuildChannel {
...
-    return this.client.voice.joinChannel(this);
-  }
-
-  /**
-   * Leaves this voice-based channel.
-   * @example
-   * // Leave a voice-based channel
-   * channel.leave();
-   */
-  leave() {
```
kyranet, line 88: Without join and leave, how are developers supposed to join/leave voice channels?

amishshah: That is done completely externally now, e.g. with @discordjs/voice

typings/index.d.ts
```diff
@@ -866,6 +855,7 @@ declare module 'discord.js' {
     public setWidget(widget: GuildWidgetData, reason?: string): Promise<Guild>;
     public splashURL(options?: ImageURLOptions): string | null;
     public toJSON(): unknown;
+    public readonly voiceAdapterCreator: DiscordGatewayAdapterCreator;
```
SpaceEEC, line 858: This should be above the methods. 👀

src/client/voice/ClientVoiceManager.js
```diff
@@ -19,97 +16,29 @@ class ClientVoiceManager {
...
-     * @type {Collection<Snowflake, VoiceConnection>}
-     */
-    this.connections = new Collection();
-
-    /**
-     * Active voice broadcasts that have been created
-     * @type {VoiceBroadcast[]}
+     * Maps guild IDs to voice adapters created for use with @discordjs/voice.
+     * @type {Map<Snowflake, Object>}
+     * @readonly
```
SpaceEEC, line 21: We only `@readonly` properties if they are actually read-only.
```suggestion
```

src/client/voice/ClientVoiceManager.js
```diff
@@ -19,97 +16,29 @@ class ClientVoiceManager {
...
      */
-    this.broadcasts = [];
-  }
+    this.adapters = new Map();
```
SpaceEEC, line 23: This does not seem to be in the typings 👀

SpaceEEC (changes requested):

kyranet (approved):
~~I'm not sure if the lack of `@discordjs/voice` in devDeps is intentional, but otherwise~~ LGTM.

Edit: I just noticed the package.json change 😅

vladfrangu (approved):

iCrawl (approved):

SpaceEEC (approved):

PR #5403, opened 2021-03-14
docs(TextChannel): warning about setRateLimitPerUser NewsChannel

ImRodry:
This adds a warning to the docs denoting that setRateLimitPerUser is not available on NewsChannels

### DIFF
--- a/src/structures/TextChannel.js
+++ b/src/structures/TextChannel.js
@@ -68,6 +68,7 @@ class TextChannel extends GuildChannel {
 
   /**
    * Sets the rate limit per user for this channel.
+   * <warn>It is not currently possible to set the rate limit per user on a `NewsChannel`.</warn>
    * @param {number} rateLimitPerUser The new ratelimit in seconds
    * @param {string} [reason] Reason for changing the channel's ratelimits
    * @returns {Promise<TextChannel>}
--- a/src/structures/TextChannel.js
+++ b/src/structures/TextChannel.js
@@ -53,6 +53,7 @@ class TextChannel extends GuildChannel {
 
     /**
      * The ratelimit per user for this channel in seconds
+     * <warn>It is not currently possible to set a rate limit per user on a `NewsChannel`.</warn>
      * @type {number}
      */
     this.rateLimitPerUser = data.rate_limit_per_user || 0;
--- a/src/structures/NewsChannel.js
+++ b/src/structures/NewsChannel.js
@@ -1,131 +1,18 @@
 'use strict';
 
-const GuildChannel = require('./GuildChannel');
-const Webhook = require('./Webhook');
-const TextBasedChannel = require('./interfaces/TextBasedChannel');
+const TextChannel = require('./TextChannel');
 const { Error } = require('../errors');
-const MessageManager = require('../managers/MessageManager');
-const Collection = require('../util/Collection');
-const DataResolver = require('../util/DataResolver');
 
 /**
  * Represents a guild news channel on Discord.
- * @extends {GuildChannel}
+ * @extends {TextChannel}
  */
-class NewsChannel extends GuildChannel {
-  /**
-   * @param {Guild} guild The guild the news channel is part of
-   * @param {Object} data The data for the news channel
-   */
-  constructor(guild, data) {
-    super(guild, data);
-    /**
-     * A manager of the messages sent to this channel
-     * @type {MessageManager}
-     */
-    this.messages = new MessageManager(this);
-
-    /**
-     * If the guild considers this channel NSFW
-     * @type {boolean}
-     * @readonly
-     */
-    this.nsfw = Boolean(data.nsfw);
-    this._typing = new Map();
-  }
-
+class NewsChannel extends TextChannel {
   _patch(data) {
     super._patch(data);
 
-    /**
-     * The topic of the text channel
-     * @type {?string}
-     */
-    this.topic = data.topic;
-
-    if (typeof data.nsfw !== 'undefined') this.nsfw = Boolean(data.nsfw);
-
-    /**
-     * The ID of the last message sent in this channel, if one was sent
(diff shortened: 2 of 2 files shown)


vladfrangu:
Copy-pasting the entire text channel just to remove one property doesn't seem like the ideal solution for this at all.. If anything, the docs should explain that NewsChannels lack the property through some way

ImRodry:
> Copy-pasting the entire text channel just to remove one property doesn't seem like the ideal solution for this at all.. If anything, the docs should explain that NewsChannels lack the property through some way

From what I understood the docs take everything from the code itself so, if NewsChannel extends TextChannel, every property will have the exact same description. I know copy-pasting is not good practice but I think it's necessary in this case due to the already existent inconsistencies in the NewsChannel typings.

ProFireDev (changes requested):
"Copy-pasting the entire text channel just to remove one property doesn't seem like the ideal solution for this at all.. If anything, the docs should explain that NewsChannels lack the property through some way"

this is not the best way of doing it, as such it shouldn't be merged.. i would say that its best if this PR were to get closed, and then maybe you or someone else can comeback with a better solution.

ImRodry:
> "Copy-pasting the entire text channel just to remove one property doesn't seem like the ideal solution for this at all.. If anything, the docs should explain that NewsChannels lack the property through some way"
> 
> this is not the best way of doing it, as such it shouldn't be merged.. i would say that its best if this PR were to get closed, and then maybe you or someone else can comeback with a better solution.

Do you know what solution would be better? I'd be glad to do that if there was one, I just couldn't find any other way to completely remove a property from a class in the docs without doing this :/

ProFireDev:
> > "Copy-pasting the entire text channel just to remove one property doesn't seem like the ideal solution for this at all.. If anything, the docs should explain that NewsChannels lack the property through some way"
> > this is not the best way of doing it, as such it shouldn't be merged.. i would say that its best if this PR were to get closed, and then maybe you or someone else can comeback with a better solution.
> 
> Do you know what solution would be better? I'd be glad to do that if there was one, I just couldn't find any other way to completely remove a property from a class in the docs without doing this :/

ill have to read though the code more and try and figure it out but the following was on to something.

> Copy-pasting the entire text channel just to remove one property doesn't seem like the ideal solution for this at all.. If anything, the docs should explain that NewsChannels lack the property through some way

I think updating the docs would go a long way to make this more clear, but I would encourage you to ask in the discord too. they might be able to help you out better or come up with a solution.

[](https://discord.com/invite/bRCvFy9)

ImRodry:
I initially asked a support agent there and this is the solution he gave me so I'm not really sure who else to ask :/
I think ideally the docs would need an update to allow for properties to be removed instead of blindly taking it from the code.

ProFireDev:
> I initially asked a support agent there and this is the solution he gave me so I'm not really sure who else to ask :/
> I think ideally the docs would need an update to allow for properties to be removed instead of blindly taking it from the code.

@almostSouji do you think this would be possible? 

You seem to be the one that contributed to the guide most recently, so I'm assuming you would have a say in something like this.

As for @ImRodry  if the docs won't get updated, I would suggest closing this PR and opening a new issue instead.

ImRodry:
Sure thing, I'll wait for that guy's reply and if the docs can't be updated I'll do that, thank you for the feedback!

catcosmicice (approved):

kyranet (changes requested):
Since it seems you're waiting for the team's consensus, Vlad's approach at https://github.com/discordjs/discord.js/pull/5403#issuecomment-798998122 is desirable.

ImRodry:
@kyranet how should we do that though?

kyranet:
> @kyranet how should we do that though?

The issue is about the docs not explaining the lack of the property, so instead of duplicating a lot of code for documentation purposes, you can add some documentation to explain why the property is missing, and that'll solve the issue you described in this PR.

ImRodry:
Well ideally we would remove the property entirely but a warning can also work. Would it be possible to update the docs to somehow detect that a property was removed from a class?

ImRodry:
@kyranet tried a different and simpler approach, let me know what you think!

src/structures/TextChannel.js
```diff
@@ -53,6 +53,7 @@ class TextChannel extends GuildChannel {
 
     /**
      * The ratelimit per user for this channel in seconds
+     * <warn>It is not currently possible to set a rate limit per user on a `NewsChannel`.</warn>
```
kyranet, line 56: I'm unsure about this warn being in TextChannel's docs, though.

kyranet (commented):
Problem with the docs as they're written right now is that we get an out-of-scope warning: the behaviour affects `NewsChannel` exclusively, but is present in `TextChannel`.

src/structures/TextChannel.js
ImRodry: Well it will default to 0 but I figured it’d make sense to say it doesn’t exist. Want me to remove it?

ImRodry:
> Problem with the docs as they're written right now is that we get an out-of-scope warning: the behaviour affects `NewsChannel` exclusively, but is present in `TextChannel`.

Yeah but if we added that to a JSDoc on NewsChannel it would be overwritten by the one on TextChannel. You can see this on the line I linked above

src/structures/TextChannel.js
SpaceEEC: It will actually be set to undefined for some wild reason, so this warn might justified.
https://github.com/discordjs/discord.js/blob/2eac84296b448907213680690ec766bb5fbe5990/src/structures/NewsChannel.js#L15

ImRodry: Well perfect then. Is it ready to be merged?

iCrawl (approved):

SpaceEEC (approved):

PR #4619, opened 2020-07-07
feat(Guild): discovery splash

advaith1:
**Please describe the changes this PR makes and why it should be merged:**
Adds support for reading and setting a guild's Discovery splash
Looks like it got into GuildPreview but was never added to Guild
The CDN stuff was added in #3965

**Status**

- [x] Code changes have been tested against the Discord API, or there are no code changes
- [x] I know how to update typings and have done so, or typings don't need updating

**Semantic versioning classification:**

- [x] This PR changes the library's interface (methods or parameters added)
  - [ ] This PR includes breaking changes (methods removed or renamed, parameters moved or removed)
- [ ] This PR **only** includes non-code changes, like changes to documentation, README, etc.

PR #5419, opened 2021-03-19
fix(Guild/GuildChannel): methods reason arg usage

BannerBomb:
I fixed an issue where the following methods were passing the reason argument as a property to the edit method instead of a parameter.
Most of these methods have had it that way for 3 years in which I don't know if y'all will want to fix it at this point. But I brought it up in case.

`Guild#setIcon` -> was introduced in #1975 (3 years ago)
`Guild#setSplash` -> was introduced in #1975 (3 years ago)
`Guild#setBanner` -> was introduced in #3364 (2 years ago)
`Guild#setDiscoverySplash` -> was introduced in #4619 (7 months ago)

`GuildChannel#overwritePermissions` -> was introduced in #3853 (1 year ago)

### DIFF
--- a/src/structures/Guild.js
+++ b/src/structures/Guild.js
@@ -1163,7 +1163,7 @@ class Guild extends Base {
    *  .catch(console.error);
    */
   async setIcon(icon, reason) {
-    return this.edit({ icon: await DataResolver.resolveImage(icon), reason });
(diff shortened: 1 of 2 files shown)


ProFireDev (approved):
looks  good, approved for review

thehairy (approved):

SpaceEEC (approved):

iCrawl (approved):

PR #5425, opened 2021-03-21
added send method to class Channel

Dastan-dotcom:
**Added send method to the types**

without this typescript would give an error if we tried to write this code:
`const channel = await client.channels.fetch('<id>')`
`ch.send('<message>')`

even though the code works fine, yet typescript shows an error `Property 'send' does not exist on type 'Channel'.`

adding this line of code to the Channel class fixes this issue.


zaida04:
This PR is invalid. The issue you bring up can easily be solved by typecasting the channel.

PR #5438, opened 2021-03-26
Updated Readme.md

AestraDev:
• Arranged for better look .
• Added  command handling in examples  for more useful and advanced examples.
• write full form of some short forms

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,11 +21,11 @@
 
 - [About](#about)
 - [Installation](#installation)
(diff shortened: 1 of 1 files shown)


iCrawl:
Making sub-bulletpoints to bulletpoints should not happen here, this is intentional.

The basic example from the website does not need to be on the readme, when going ahead and using the project one would open the documentation page anyway as a first step.

Thank you for the contribution.

PR #5442, opened 2021-03-27
Add ! to all User/GuildMember mentions

ImRodry:
Whenever mentioning a user in a message, the mention will always have the ! no matter if you write it or not (which you can observe through message.content), therefore this fixes that inconsistency with the toString method. The ! doesn't seem to have any practical use from what I was able to observe in my tests so adding this should not break anything

### DIFF
--- a/src/structures/DMChannel.js
+++ b/src/structures/DMChannel.js
@@ -73,7 +73,7 @@ class DMChannel extends Channel {
    * DMChannel object.
    * @returns {string}
    * @example
-   * // Logs: Hello from <@123456789012345678>!
+   * // Logs: Hello from <@!123456789012345678>!
    * console.log(`Hello from ${channel}!`);
    */
   toString() {
--- a/src/structures/GuildMember.js
+++ b/src/structures/GuildMember.js
@@ -378,11 +378,11 @@ class GuildMember extends Base {
    * When concatenated with a string, this automatically returns the user's mention instead of the GuildMember object.
    * @returns {string}
    * @example
-   * // Logs: Hello from <@123456789012345678>!
+   * // Logs: Hello from <@!123456789012345678>!
    * console.log(`Hello from ${member}!`);
    */
   toString() {
-    return `<@${this.nickname ? '!' : ''}${this.user.id}>`;
+    return `<@!${this.user.id}>`;
   }
 
   toJSON() {
--- a/src/structures/MessageMentions.js
+++ b/src/structures/MessageMentions.js
@@ -205,7 +205,7 @@ class MessageMentions {
 MessageMentions.EVERYONE_PATTERN = /@(everyone|here)/g;
 
 /**
- * Regular expression that globally matches user mentions like `<@81440962496172032>`
+ * Regular expression that globally matches user mentions like `<@!81440962496172032>`
  * @type {RegExp}
  */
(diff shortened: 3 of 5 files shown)


advaith1:
the ! is intended to indicate that the user has a nickname (as the docs say) for client optimizations but the clients do not always do that due to bugs

ImRodry:
> the ! is intended to indicate that the user has a nickname (as the docs say) for client optimizations but the clients do not always do that due to bugs

Can you show me an example where the ! makes a difference? I tried tagging a user with a nickname both with and without the ! and the result was the same. I also did that on a dm channel, where there are no nicknames, and the mention had the ! so that's why I assume it wasn't related to that.

Jiralite:
> I tried tagging a user with a nickname both with and without the ! and the result was the same.
This is because you are on the beta chat box. Turn it off and the `!` will disappear. The current implementation is fine!

ImRodry:
> > I tried tagging a user with a nickname both with and without the ! and the result was the same.
> 
> This is because you are on the beta chat box. Turn it off and the `!` will disappear. The current implementation is fine!

Nope, I disabled that, tagged a user with a nickname with and without the ! and both had the nickname. The difference was only visible from message.content, not from the user's perspective.

advaith1:
its checked for notifications iirc, not sure if it currently does though due to the client bugs

anyway this code shouldn't be changed

ImRodry:
Well I tried mentioning myself in both ways and I got a notification on both instances.
To be honest, I don't know the original purpose of the ! in mentions since it always seems to be irrelevant but I believe this should be merged because the current implementation is inconsistent with the most likely scenario which is when a user mentions another user by writing their name and pressing tab. If you do this with or without the text bot beta, the resulting mention will have the !

advaith1:
> inconsistent with the most likely scenario which is when a user mentions another user by writing their name and pressing tab

well as I said, that is a client bug according to devs and may be fixed eventually

ImRodry:
But what exactly is the intended behavior? Mentions obtained by writing user names have always had the ! to my knowledge

advaith1:
The intended behavior is that the ! is present when the member has a nickname in the server

https://discord.com/developers/docs/reference#message-formatting

ImRodry:
Yeah fair enough then. If I undid the change to GuildMember.js and kept the rest would this be merged?

Jiralite:
> Mentions obtained by writing user names have always had the ! to my knowledge

This is not true (and still is not true) as I said: this is because you are on the beta chat box. Turn it off and the ! will disappear (to stress, not for nicknames).

ImRodry:
> > Mentions obtained by writing user names have always had the ! to my knowledge
> 
> This is not true (and still is not true) as I said: this is because you are on the beta chat box. Turn it off and the `!` will disappear (to stress, not for nicknames).

I did disable it and tested by writing the user’s name after the @ and it did have the !, check it yourself

Jiralite:
Screenshot 2021-03-28 at 05 50 58

ImRodry:
That is indeed very odd because I always get the mention with ! when writing the user’s name 🤔

almostSouji:
Thank you for attempting to fix what seemed to be an inconsistency, however, the current behavior is wanted and this doesn't really fix or provide benefit to anything.

ImRodry:
Fair enough, thanks for the feedback!

PR #5447, opened 2021-03-29
Make ClientEvents, CollectorFilter and Message types generic.

VimHax:

I've added some generic type parameters for the following types for the respective reasons stated below.

-> `CollectorFilter` - I feel that this type would benefit from being able to define exactly what parameters will be passed in to the function. Not only does this leave less room for mistakes but it also makes it so users don't have to work with the type `any` as much when they're using TypeScript.
-> `ClientEvents` - This type has been given a generic parameter which dictates which events will emit partial data. This is useful as redundant checking for partial data won't be required if the user has not enabled the respective partial in the `ClientOptions`.
-> `Message` - The generic parameter for this type describes whether or not the message was sent in a guild. When you're working with messages you often have to check whether or not `Message#guild` is null even when you're pretty certain it isn't, inserting `Message#guild as Guild` everywhere is also much less elegant. This parameter can also be used to add more type information to the return types of some functions, such as `Message#reply` which will always return a `Message` which was sent in a guild if the original message was sent in a guild.


iCrawl:
This needs a rebase.

VimHax:
I merged the changes into master, will that do?

typings/index.d.ts
```diff
@@ -958,15 +958,15 @@ declare module 'discord.js' {
     public toString(): string;
   }
 
-  export class Message extends Base {
-    constructor(client: Client, data: object, channel: TextChannel | DMChannel | NewsChannel);
+  export class Message<G extends boolean = boolean> extends Base {
```
NotSugden, line 961: maybe could rename the type parameter to like `Guild` or smth, G feels like a weird name

VimHax: But wouldn't that clash with the `Guild` type?

NotSugden: true, the point was to have a better name for it tho

VimHax: I made the name `SentInGuild` and changed some other names too... Hopefully that has improved the situation.

VimHax:
Should've made separate PRs for each of the features I implemented.

PR #4766, opened 2020-08-31
issue when providing a lowercase color name

BannerBomb:
**Please describe the changes this PR makes and why it should be merged:**
The changes made are to change the color string to upper case since that's how they are written in the Constants file to fix an issue that when trying to use lower case names it will output NaN instead of the color.

An example of what I mean
image

**Status**

- [x] Code changes have been tested against the Discord API, or there are no code changes
- [ ] I know how to update typings and have done so, or typings don't need updating

**Semantic versioning classification:**

- [ ] This PR changes the library's interface (methods or parameters added)
  - [ ] This PR includes breaking changes (methods removed or renamed, parameters moved or removed)
- [ ] This PR **only** includes non-code changes, like changes to documentation, README, etc.

PR #5448, opened 2021-03-30
feat: interactions

vaporox:

Attempt #​2! Most of the logic is taken from #5106, but with a giant refactor, most notably the interaction client class being removed and an application command manager class being added. To see a quick example of how this branch is structured, have a look [here](https://github.com/vaporox/discord.js/blob/feat-interactions/docs/examples/commands.md).

Additionally, this PR also adds the new `USE_APPLICATION_COMMANDS` permission, the `APPLICATION_COMMAND` message type and the `EPHEMERAL` and `LOADING` message flags.


src/structures/ApplicationCommand.js
```diff
@@ -0,0 +1,150 @@
...
+   * Transforms an {@link ApplicationCommandOption} object into something that can be used with the API.
+   * @param {ApplicationCommandOption} option The option to transform
+   * @returns {Object}
+   * @private
+   */
+  static transformOption(option) {
+    return {
+      ...option,
+      type: ApplicationCommandOptionTypes[option.type],
+      options: option.options?.map(this.transformOption),
```
NotSugden, line 145:
```suggestion
      options: option.options?.map(opt => this.transformOption(opt)),
```
or
```suggestion
      options: option.options?.map(this.transformOption.bind(this)),
```

src/structures/ApplicationCommand.js
```diff
@@ -0,0 +1,150 @@
...
+  /**
+   * Edits this application command.
+   * @param {ApplicationCommandData} data The data to update the command with
+   * @returns {ApplicationCommand}
+   */
+  async edit(data) {
+    const raw = {};
+    if (data.name) raw.name = data.name;
+    if (data.description) raw.description = data.description;
+    if (data.options) raw.options = data.options.map(this.constructor.transformOption);
```
NotSugden, line 85: same as previous

src/structures/ApplicationCommand.js
```diff
@@ -0,0 +1,150 @@
...
+     * The description of this command
+     * @type {string}
+     */
+    this.description = data.description;
+
+    /**
+     * The options of this command
+     * @type {ApplicationCommandOption[]}
+     */
+    this.options = data.options?.map(this.constructor.transformOption) ?? [];
```
NotSugden, line 47: again

src/structures/CommandInteraction.js
```diff
@@ -0,0 +1,146 @@
...
+     * The options passed to the command.
+     * @type {CommandInteractionOption[]}
+     */
+    this.options = data.data.options?.map(o => this.constructor.transformOption(o, data.data.resolved)) ?? [];
+
+    /**
+     * An associated webhook client, can be used to create deferred replies
+     * @type {WebhookClient}
+     */
+    this.webhook = new WebhookClient(this.client.application.id, this.token);
```
NotSugden, line 39: logic in `Client#destroy()` and on the `GUILD_DELETE` event that calls the `destroy()` method on all these `WebhookClient`s needs to be added

NotSugden (changes requested):

src/structures/ApplicationCommand.js
vaporox: This doesn't seem to be required: [Playground](https://www.typescriptlang.org/play?#code/MYGwhgzhAEByCuBbARgUwE4wN4Chr+gG0ByYAewDsIAXdeYas9YgXQC5pqBPAB1TIBmcJGkwBuPAURgefACYB+DhREZCLCZPzSeACgCU0XAROcAFgEsIAOh3zoAXiIBGADTQATO4DMLWzN1qSxtyKlp6RnRrMDk5ABUAdzJ9CRMAXxwtaBowagtgaBj4pN0VRGVVdENjU3x0VGp4dApoMugAak9UggyMnFBIGDjUGiMsstEIDgQUDAhNE1CaOgYmAzHa-CCrawm5x1bUBOFZ8SzezKXqThHrpwoj6GGaCWpb3cqbHQNNJbIQVDWEBkADmgXee0w-lkqDkKRwQA)

I also just tested creating / fetching application commands, and it worked fine without this change

vaporox: oh nevermind, the error comes when it calls itself recursively, i'll just change it to the class name

src/managers/ApplicationCommandManager.js
```diff
@@ -0,0 +1,104 @@
...
+    const data = await this.commandPath.get();
+    return data.reduce((coll, command) => coll.set(command.id, this.add(command, cache)), new Collection());
+  }
+
+  /**
+   * Creates an application command.
+   * @param {ApplicationCommandData} command The command
+   * @returns {Promise<ApplicationCommand>}
+   */
+  async create(command) {
```
Fyko, line 71: Functions such as `ApplicationCommandManager#edit` and `ApplicationCommandManager#delete` would be quite lovely, similar to #4869 and #4872. I would suggest an upsert function however from reading the docs posting [already does that](https://discord.com/developers/docs/interactions/slash-commands#create-global-application-command).

src/structures/CommandInteraction.js
```diff
@@ -0,0 +1,149 @@
...
+const Interaction = require('./Interaction');
+const WebhookClient = require('../client/WebhookClient');
+const { ApplicationCommandOptionTypes, InteractionResponseTypes } = require('../util/Constants');
+const MessageFlags = require('../util/MessageFlags');
+
+/**
+ * Represents a command interaction.
+ * @extends {Interaction}
+ */
+class CommandInteraction extends Interaction {
```
cactysman, line 13: Personally I don't quite like how command-related info is split across three keys.
Instead of having `commandID`, `commandName` and `options`, why not group that under a key like `commandInfo`?

monbrey: The current naming fits better with other conventions in the library.

Besides, I would argue that most of the time you as a user, you would be using the `CommandInteraction#command` getter, which internally uses the `CommandInteraction#commandID` to return the full `ApplicationCommand` object

cactysman: My thought was that the key name `options` would seem a bit too generic, but if that goes with the conventions, then that's alright.

I assumed `Interactions` would also reflect other types of interactions, but if that's only the slash commands, that would not be an issue.

vaporox: While I agree that `commandID` and `commandName` *could* be grouped together, I don't think it would make much sense, considering that you have the `command` getter in case you have fetched the command(s) before. As for `options`, this should stay on the top level, otherwise it could be confused with `ApplicationCommand#options` which has a different structure than `CommandInteraction#options` does.

src/structures/ApplicationCommand.js
```diff
@@ -0,0 +1,135 @@
...
+  get createdAt() {
+    return new Date(this.createdTimestamp);
+  }
+
+  /**
+   * The manager that this command belongs to
+   * @type {ApplicationCommandManager}
+   */
+  get manager() {
+    return this.guild?.commands ?? this.client.application.commands;
```
NotSugden, line 73:
```suggestion
    return (this.guild ?? this.client.application).commands;
```

src/structures/CommandInteraction.js
```diff
@@ -0,0 +1,149 @@
...
+    this.webhook = new WebhookClient(this.client.application.id, this.token);
+  }
+
+  /**
+   * The invoked application command, if it was fetched before
+   * @type {?ApplicationCommand}
+   */
+  get command() {
+    const id = this.commandID;
+    return this.guild?.commands.cache.get(id) ?? this.client.application.commands.cache.get(id) ?? null;
```
NotSugden, line 48:
```suggestion
    return (this.guild ?? this.client.application).commands.cache.get(this.commandID) ?? null;
```

src/structures/CommandInteraction.js
```diff
@@ -0,0 +1,149 @@
...
+    }
+
+    if (option.type === ApplicationCommandOptionTypes.USER) {
+      const user = resolved.users[option.value];
+      result.user = this.client.users.add(user);
+
+      const member = resolved.members?.[option.value];
+      if (member) {
+        member.user = user;
+        result.member = this.guild?.members.add(member) ?? null;
```
NotSugden, line 131: is there a need to use optional chaining here? a member would only be provided if there is a guild

src/structures/CommandInteraction.js
```diff
@@ -0,0 +1,149 @@
...
+    }
+
+    if (option.type === ApplicationCommandOptionTypes.CHANNEL) {
+      const channel = resolved.channels[option.value];
+      result.channel = this.client.channels.add(channel, this.guild);
+    }
+
+    if (option.type === ApplicationCommandOptionTypes.ROLE) {
+      const role = resolved.roles[option.value];
+      result.role = this.guild?.roles.add(role) ?? null;
```
NotSugden, line 142: the same goes for here too

typings/index.d.ts
```diff
@@ -538,6 +592,9 @@ declare module 'discord.js' {
     DefaultMessageNotifications: DefaultMessageNotifications[];
     VerificationLevels: VerificationLevel[];
     MembershipStates: 'INVITED' | 'ACCEPTED';
+    ApplicationCommandOptionTypes: ApplicationCommandOptionTypes;
+    InteractionTypes: InteractionTypes;
+    InteractionResponseTypes: InteractionResponseTypes;
```
NotSugden, line 597:
```suggestion
    ApplicationCommandOptionTypes: typeof ApplicationCommandOptionTypes;
    InteractionTypes: typeof InteractionTypes;
    InteractionResponseTypes: typeof InteractionResponseTypes;
```
it would be `typeof x` here right? without `typeof` these would just be typed as numbers when they aren't numbers

NotSugden (changes requested):

src/structures/CommandInteraction.js
vaporox: An interaction from a global command can still have an associated guild, so it would return null in that case

src/structures/CommandInteraction.js
vaporox: This is a fail-safe in case the guild is not cached

src/structures/CommandInteraction.js
NotSugden: you're right, mb

src/util/Constants.js
```diff
@@ -431,6 +434,7 @@ exports.InviteScopes = [
  * * GUILD_DISCOVERY_DISQUALIFIED
  * * GUILD_DISCOVERY_REQUALIFIED
  * * REPLY
+ * * APPLICATION_COMMAND
  * @typedef {string} MessageType
  */
 exports.MessageTypes = [
```
NotSugden, line 440: below this is the `SystemMessageTypes` export, that needs to be updated to exclude the new type

NotSugden (changes requested):

iCrawl:
This needs a rebase.

src/util/Constants.js
```diff
@@ -454,15 +459,21 @@ exports.MessageTypes = [
...
  * The types of messages that are `System`. The available types are `MessageTypes` excluding:
  * * DEFAULT
  * * REPLY
+ * * APPLICATION_COMMAND
+ * * LOADING
  * @typedef {string} SystemMessageType
  */
-exports.SystemMessageTypes = exports.MessageTypes.filter(type => type && type !== 'DEFAULT' && type !== 'REPLY');
+exports.SystemMessageTypes = exports.MessageTypes.filter(
+  type => type && type !== 'DEFAULT' && type !== 'REPLY' && type !== 'APPLICATION_COMMAND' && type !== 'LOADING',
```
advaith1, line 475: LOADING is a message flag, not a message type

```suggestion
 ];
 
 /**
  * The types of messages that are `System`. The available types are `MessageTypes` excluding:
  * * DEFAULT
  * * REPLY
  * * APPLICATION_COMMAND
  * @typedef {string} SystemMessageType
  */
exports.SystemMessageTypes = exports.MessageTypes.filter(type => type && type !== 'DEFAULT' && type !== 'REPLY');
exports.SystemMessageTypes = exports.MessageTypes.filter(
  type => type && type !== 'DEFAULT' && type !== 'REPLY' && type !== 'APPLICATION_COMMAND',
```

src/util/MessageFlags.js
```diff
@@ -28,6 +28,7 @@ class MessageFlags extends BitField {}
  * * `SUPPRESS_EMBEDS`
  * * `SOURCE_MESSAGE_DELETED`
  * * `URGENT`
+ * * `EPHEMERAL`
```
advaith1, line 31:
```suggestion
 * * `EPHEMERAL`
 * * `LOADING`
```

src/util/MessageFlags.js
```diff
@@ -37,6 +38,7 @@ MessageFlags.FLAGS = {
   SUPPRESS_EMBEDS: 1 << 2,
   SOURCE_MESSAGE_DELETED: 1 << 3,
   URGENT: 1 << 4,
+  EPHEMERAL: 1 << 6,
```
advaith1, line 41:
```suggestion
  EPHEMERAL: 1 << 6,
  LOADING: 1 << 7,
```

typings/index.d.ts
```diff
@@ -3046,7 +3191,9 @@ declare module 'discord.js' {
     | 'CHANNEL_FOLLOW_ADD'
     | 'GUILD_DISCOVERY_DISQUALIFIED'
     | 'GUILD_DISCOVERY_REQUALIFIED'
-    | 'REPLY';
+    | 'REPLY'
+    | 'APPLICATION_COMMAND'
+    | 'LOADING';
```
advaith1, line 3196:
```suggestion
    | 'APPLICATION_COMMAND';
```

typings/index.d.ts
```diff
@@ -2995,7 +3134,13 @@ declare module 'discord.js' {
...
   }
 
-  type MessageFlagsString = 'CROSSPOSTED' | 'IS_CROSSPOST' | 'SUPPRESS_EMBEDS' | 'SOURCE_MESSAGE_DELETED' | 'URGENT';
+  type MessageFlagsString =
+    | 'CROSSPOSTED'
+    | 'IS_CROSSPOST'
+    | 'SUPPRESS_EMBEDS'
+    | 'SOURCE_MESSAGE_DELETED'
+    | 'URGENT'
+    | 'EPHEMERAL';
```
advaith1, line 3143:
```suggestion
    | 'EPHEMERAL'
    | 'LOADING';
```

typings/index.d.ts
```diff
@@ -1,3 +1,14 @@
+declare enum ApplicationCommandOptionTypes {
+  SUB_COMMAND = 1,
+  SUB_COMMAND_GROUP = 2,
+  STRING = 3,
+  INTEGER = 4,
+  BOOLEAN = 5,
+  USER = 6,
+  CHANNEL = 7,
+  ROLE = 8,
```
advaith1, line 9:
```suggestion
  SUB_COMMAND_GROUP,
  STRING,
  INTEGER,
  BOOLEAN,
  USER,
  CHANNEL,
  ROLE,
```

or could just import these enums from discord-api-types

typings/index.d.ts
```diff
@@ -9,6 +20,17 @@ declare enum ChannelType {
   unknown = 7,
 }
 
+declare enum InteractionResponseTypes {
+  PONG = 1,
+  CHANNEL_MESSAGE_WITH_SOURCE = 4,
+  DEFERRED_CHANNEL_MESSAGE_WITH_SOURCE = 5,
```
advaith1, line 26:
```suggestion
  DEFERRED_CHANNEL_MESSAGE_WITH_SOURCE,
```

typings/index.d.ts
```diff
@@ -9,6 +20,17 @@ declare enum ChannelType {
...
+}
+
+declare enum InteractionTypes {
+  PING = 1,
+  APPLICATION_COMMAND = 2,
```
advaith1, line 31:
```suggestion
  APPLICATION_COMMAND,
```

src/util/Constants.js
```diff
@@ -431,6 +434,8 @@ exports.InviteScopes = [
  * * GUILD_DISCOVERY_DISQUALIFIED
  * * GUILD_DISCOVERY_REQUALIFIED
  * * REPLY
+ * * APPLICATION_COMMAND
+ * * LOADING
```
advaith1, line 438:
```suggestion
```

src/util/Constants.js
```diff
@@ -454,15 +459,21 @@ exports.MessageTypes = [
   null,
   null,
   'REPLY',
+  'APPLICATION_COMMAND',
+  'LOADING',
```
advaith1, line 463:
```suggestion
```

advaith1 (changes requested):

typings/index.d.ts
```diff
@@ -3330,7 +3478,7 @@ declare module 'discord.js' {
 
   type SystemChannelFlagsResolvable = BitFieldResolvable<SystemChannelFlagsString, number>;
 
-  type SystemMessageType = Exclude<MessageType, 'DEFAULT' | 'REPLY'>;
+  type SystemMessageType = Exclude<MessageType, 'DEFAULT' | 'REPLY' | 'APPLICATION_COMMAND' | 'LOADING'>;
```
advaith1, line 3481:
```suggestion
  type SystemMessageType = Exclude<MessageType, 'DEFAULT' | 'REPLY' | 'APPLICATION_COMMAND'>;
```

missed this one

could just revert the "more message types" commit instead

advaith1 (changes requested):

monbrey:
Can I suggest that when `CommandInteraction#defer()` is used, it sets a flag like `CommandInteraction#deferred`?

This would allow other sections of code such as an error handler to dynamically determine how to respond based on interaction state.

ckohen:
> Can I suggest that when `CommandInteraction#defer()` is used, it sets a flag like `CommandInteraction#deferred`?

To add to this, it would be great if when using `CommandInteraction#reply()` and this flag is set it returns the message. I know this makes typings a little more complicated, so to add to that and solve the typings issue, if the flag has not been set, GET and return `@original` within this call as well (PATCH with no data for now)

monbrey:
> To add to this, it would be great if when using `CommandInteraction#reply()` and this flag is set it returns the message.

The `reply()` method will API error if the interaction response has already been deferred - one or the other must be the initial acknowledgement.

`editReply()` is the one that can be used to update a deferred response via the webhook - which I agree should return the Message if it can.

cactysman:
> `editReply()` is the one [...] which I agree should return the Message if it can.
Yes please.
Let's try to stick to one way of dealing with messages instead of making it unnecessarily complicated with multiple different methods.

ckohen:
> The `reply()` method will API error if the interaction response has already been deferred - one or the other must be the initial acknowledgement.

If there were error checking as you suggest to determine what method to use then it wouldn't api error because it would edit instead? I would intuitively use reply() after a defer because I...haven't replied yet. (I may also be biasing based on what I have experienced from sneks PR)

> `editReply()` is the one that can be used to update a deferred response via the webhook - which I agree should return the Message if it can.

This already does return the message, granted only if the channel is cached.

<details>
  <summary>Basically what I was proposing is this</summary>

```js
 async reply(content, options) {
    const apiMessage = content instanceof APIMessage ? content : APIMessage.create(this, content, options);
    const { data } = await apiMessage.resolveData().resolveFiles();

    if(this.deffered) return this.editReply(apiMessage);
    await this.client.api.interactions(this.id, this.token).callback.post({
      data: {
        type: InteractionResponseTypes.CHANNEL_MESSAGE_WITH_SOURCE,
        data,
      },
    });
    const raw = this.client.api.interactions(this.client.applicatoin.id, this.token).message('@original').get();
    return this.channel?.messages.add(raw) ?? null;
}
```

Currently the second to last line would actually end in `.patch({ data: {})`
</details>

monbrey:
> If there were error checking as you suggest to determine what method to use then it wouldn't api error because it would edit instead?

No, I don't believe the library should internally decide which endpoint to use. `defer()` and `reply()` are two different acknowledgement types. It's not up to the library to protect the user from API errors.

All in asking for is a `deferred` property which will assist the user in protecting themselves, in a similar manner to your example.

src/structures/CommandInteraction.js
```diff
@@ -0,0 +1,170 @@
...
+
+    if (option.type === ApplicationCommandOptionTypes.CHANNEL) {
+      const channel = resolved.channels[option.value];
+      result.channel = this.client.channels.add(channel, this.guild);
+    }
+
+    if (option.type === ApplicationCommandOptionTypes.ROLE) {
+      const role = resolved.roles[option.value];
+      result.role = this.guild.roles.add(role);
+    }
```
MrZillaGold, line 164: Is it better to use `switch`?

vaporox: If we used `switch` here, there would need to be either an `eslint-disable` or a block for every case due to `no-case-declarations`. This would just make it unnecessarily complex.

src/client/websocket/handlers/INTERACTION_CREATE.js
```diff
@@ -0,0 +1,19 @@
...
+
+    /**
+     * Emitted when an interaction is created.
+     * @event Client#interaction
+     * @param {Interaction} interaction The interaction which was created
+     */
+    return client.emit(Events.INTERACTION_CREATE, interaction);
+  }
+
+  return client.emit('debug', `[INTERACTION] Received interaction with unknown type: ${data.type}`);
```
stu43005, line 18:
```suggestion
  return client.emit(Events.DEBUG, `[INTERACTION] Received interaction with unknown type: ${data.type}`);
```

ckohen:
> All in asking for is a `deferred` property which will assist the user in protecting themselves, in a similar manner to your example.

Ah I see, you meant error checking on the users end, not the libs.

I suppose the other half of my point would be fulfilled by a getReply method. Although it would be nice to have reply return directly.

Also worthy of note, I feel that CommandInteraction should be an extendable structure

advaith1:
> I suppose the other half of my point would be fulfilled by a getReply method. Although it would be nice to have reply return directly.

It would be nice if Discord returned the message object, but I don't think d.js should always make an extra API request automatically when it is not always needed. A getReply method should definitely be added though.

vaporox:
I added a `fetchReply()` method, however you need the channel so that `.messages.add()` can be called. This means that it will also return `null` for DMs in most cases since API v8 doesn't give DM channels anymore. I'm not sure if there is a better option than using partial DM channels here because i want to avoid returning the raw message when possible.

src/structures/CommandInteraction.js
```diff
@@ -0,0 +1,194 @@
...
+    });
+    this.replied = true;
+  }
+
+  /**
+   * Fetches the initial reply to this interaction.
+   * @returns {Promise<?Message>}
+   */
+  async fetchReply() {
+    const raw = await this.client.api.webhooks(this.client.application.id, this.token).messages('@original').get();
```
ThaTiemsz, line 112:
```suggestion
    const raw = await this.client.api.webhooks(this.client.application.id, this.token).messages('@original').patch({data: {}});
```
GET endpoint doesn't exist for webhook messages, you'll have to send a PATCH request with an empty body which will return the message object.

ThaTiemsz (changes requested):

src/structures/CommandInteraction.js
```diff
@@ -0,0 +1,195 @@
...
+   * Transforms an option received from the API.
+   * @param {Object} option The received option
+   * @param {Object} resolved The resolved interaction data
+   * @returns {CommandInteractionOption}
+   * @private
+   */
+  transformOption(option, resolved) {
+    const result = {
+      name: option.name,
+      type: ApplicationCommandOptionTypes[option.type],
```
trezy, line 159: For the sake of DX, it would make sense to add more flexible validation to the option type. For example, setting the `type` to a lowercase string of `string` causes the field to be unset, which isn't valid:
code

That could be solved easily by adding `option.type.toUpperCase()` here.

I'd also expect this to allow me to use the ApplicationCommandOptionType values directly.

```suggestion
      type: (typeof option.type === 'number')
        ? option.type
        : ApplicationCommandOptionTypes[option.type.toUpperCase()],
```

src/util/Constants.js
```diff
@@ -454,15 +458,19 @@ exports.MessageTypes = [
...
 /**
  * The types of messages that are `System`. The available types are `MessageTypes` excluding:
  * * DEFAULT
  * * REPLY
+ * * APPLICATION_COMMAND
  * @typedef {string} SystemMessageType
  */
-exports.SystemMessageTypes = exports.MessageTypes.filter(type => type && type !== 'DEFAULT' && type !== 'REPLY');
+exports.SystemMessageTypes = exports.MessageTypes.filter(
+  type => type && type !== 'DEFAULT' && type !== 'REPLY' && type !== 'APPLICATION_COMMAND',
```
advaith1, line 472:
```suggestion
  type => type && !['DEFAULT', 'REPLY', 'APPLICATION_COMMAND'].includes(type)
```

ckohen:
Still looking for CommandInteraction to be extendable

monbrey:
Slash Command Permissions documentation is now merged:
https://github.com/discord/discord-api-docs/pull/2737

src/structures/CommandInteraction.js
```diff
@@ -0,0 +1,195 @@
...
+     * Whether this interaction has already been replied to
+     * @type {boolean}
+     */
+    this.replied = false;
+
+    /**
+     * An associated webhook client, can be used to create deferred replies
+     * @type {WebhookClient}
+     */
+    this.webhook = new WebhookClient(data.application_id, data.token);
```
advaith1, line 51:
```suggestion
    this.webhook = new WebhookClient(data.application_id, data.token, this.client.options);
```
passes along the ClientOptions; I've had issues where I didn't do this and I was debugging it for a while because I forgot the WebhookClient is separate

also maybe rename this from `webhook` to `followup` or something?

src/managers/ApplicationCommandManager.js
```diff
@@ -0,0 +1,192 @@
...
+   * @returns {Promise<ApplicationCommand>}
+   */
+  async edit(command, data) {
+    const id = this.resolveID(command);
+    if (!id) throw new TypeError('INVALID_TYPE', 'command', 'ApplicationCommandResolvable');
+
+    const raw = {};
+    if (data.name) raw.name = data.name;
+    if (data.description) raw.description = data.description;
+    if (data.options) raw.options = data.options.map(ApplicationCommand.transformOption);
```
monbrey, line 105:
```suggestion
    if (data.options) raw.options = data.options.map(ApplicationCommand.transformOption);
    if (data.defaultPermission) raw.default_permission = data.defaultPermission;
```

src/structures/CommandInteraction.js
vaporox: It's also used to edit/delete the interaction reply, so `webhook` is a more fitting name

src/structures/CommandInteraction.js
vaporox: I implemented the ability to pass numbers as types, however I think that if you pass a string, it should still be capitalized (see also e.g. #4766)

src/managers/ApplicationCommandManager.js
```diff
@@ -0,0 +1,201 @@
...
+      ...command,
+      options: command.options?.map(ApplicationCommand.transformOption),
+      default_permission: command.defaultPermission,
+    };
+  }
+
+  /**
+   * Transforms an {@link ApplicationCommandPermissionData} object into something that can be used with the API.
+   * @param {ApplicationCommandPermissionData} permissions The permissions to transform
+   * @param {boolean} [received] Whether this permissions have been received from Discord
```
ThaTiemsz, line 186:
```suggestion
   * @param {boolean} [received] Whether these permissions have been received from Discord
```

src/client/websocket/handlers/INTERACTION_CREATE.js
```diff
@@ -0,0 +1,25 @@
+'use strict';
+
+const { Events, InteractionTypes } = require('../../../util/Constants');
+let Structures;
+
+module.exports = (client, { d: data }) => {
+  if (data.type === InteractionTypes.APPLICATION_COMMAND) {
+    if (data.guild_id && !client.guilds.cache.has(data.guild_id)) return;
```
ckohen, line 8: This will prevent interactions from guilds without a bot. The types for Interaction#guildId and Interaction#guild already allow the guild to be null if it is not cached.

src/structures/CommandInteraction.js
```diff
@@ -0,0 +1,195 @@
...
+    this.replied = true;
+  }
+
+  /**
+   * Fetches the initial reply to this interaction.
+   * @returns {Promise<?Message>}
+   */
+  async fetchReply() {
+    const applicationID = this.client.application.id;
+    const raw = await this.client.api.webhooks(applicationID, this.token).messages('@original').patch({ data: {} });
```
advaith1, line 113:
```suggestion
    const raw = await this.client.api.webhooks(applicationID, this.token).messages('@original').get();
```
pog its out

src/structures/CommandInteraction.js
KevinNovak: Maybe there could be like an alias method, `interaction.followup()` that just does `interaction.webhook.send()`

cmd430:
When creating a command with sub-commands that have options I'm getting a TypeError `Cannot read property 'transformOption' of undefined`, everything is fine when the sub-commands do not have options, examples below

This was working fine as of https://github.com/discordjs/discord.js/pull/5448/commits/f2217d30147017dde54f538b6f854471eef4046e

```js
// Not Working, Throws TypeError: Cannot read property 'transformOption' of undefined
{
  name: 'dev',
  description: 'Command to test stuff',
  options: [
    {
      name: 'test_sub_command',
      description: 'a test sub-command',
      type: 'SUB_COMMAND'
    },
    {
      name: 'test_sub_command_with_arg',
      description: 'a test sub-command with an argument',
      type: 'SUB_COMMAND',
      options: [
        {
          name: 'input',
          type: 'STRING',
          description: 'a string input',
          required: true
        }
      ]
    }
  ]
}
```

```js
// Working, No Errors
{
  name: 'dev',
  description: 'Command to test stuff',
  options: [
    {
      name: 'test_sub_command',
      description: 'a test sub-command',
      type: 'SUB_COMMAND'
    },
    {
      name: 'test_sub_command_with_arg',
      description: 'a test sub-command with an argument',
      type: 'SUB_COMMAND',
    }
  ]
}
```

src/structures/Interaction.js
```diff
@@ -0,0 +1,105 @@
...
+     * The user which sent this interaction
+     * @type {User}
+     */
+    this.user = this.client.users.add(data.user ?? data.member.user);
+
+    /**
+     * If this interaction was sent in a guild, the member which sent it
+     * @type {?GuildMember}
+     */
+    this.member = data.member ? this.guild?.members.add(data.member) ?? null : null;
```
vladfrangu, line 57: Why not support raw data here?

vladfrangu (changes requested):

src/structures/CommandInteraction.js
```diff
@@ -0,0 +1,194 @@
...
+   */
+
+  /**
+   * Transforms an option received from the API.
+   * @param {Object} option The received option
+   * @param {Object} resolved The resolved interaction data
+   * @returns {CommandInteractionOption}
+   * @private
+   */
+  transformOption(option, resolved) {
```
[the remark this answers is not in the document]
ObscuritySRL: `src\structures\CommandInteraction.js:175:40` throws a `TypeError` when providing a valid user ID of someone who is _not_ in that guild. Eg:
```
TypeError: Cannot read property '475059745749991424' of undefined
    at CommandInteraction.transformOption (C:\Users\ObscuritySRL\projects\test-bot\node_modules\discord.js\src\structures\CommandInteraction.js:175:40)
```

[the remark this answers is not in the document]
vaporox: It's not static because it needs access to `this.client` and `this.guild`

The other two problems are a mistake on my part, I'll get to them later

src/structures/Interaction.js
vaporox: I'm personally not a fan of forwarding raw objects to the user. It would help if the guild member and message classes both just had a `guildID` property and a `guild` getter but i suppose that's not acievable before -next, so using the raw member instead is probably the better idea.

vladfrangu: Why is it not doable before -next? it's a new thing, surely we can do it 👀 Plus, we already do raw data for webhooks

vaporox: I'll add support for raw data as soon as #5539 gets merged

CyberFlameGO (approved):

vladfrangu (approved):
Looks good to me

kyranet (approved):

src/managers/ApplicationCommandManager.js
```diff
@@ -0,0 +1,201 @@
...
+    this.cache.delete(id);
+    return cached ?? null;
+  }
+
+  /**
+   * Fetches the permissions for one or multiple commands.
+   * @param {ApplicationCommandResolvable} [command] The command to get the permissions from
+   * @returns {Promise<ApplicationCommandPermissions[]|Collection<Snowflake, ApplicationCommandPermissions[]>>}
+   */
+  async fetchPermissions(command) {
```
HarmoGlace, line 133: This should be optional as you can fetch all command permissions in a guild.

src/managers/ApplicationCommandManager.js
```diff
@@ -0,0 +1,201 @@
...
+    );
+  }
+
+  /**
+   * Sets the permissions for a command.
+   * @param {ApplicationCommandResolvable} command The command to edit the permissions for
+   * @param {ApplicationCommandPermissions[]} permissions The new permissions for the command
+   * @returns {Promise<?ApplicationCommand>}
+   */
+  async setPermissions(command, permissions) {
```
HarmoGlace, line 159: It should also accept an array, as you can edit several application command permissions with one call.

src/structures/Interaction.js
```diff
@@ -0,0 +1,105 @@
...
+
+const Base = require('./Base');
+const { InteractionTypes } = require('../util/Constants');
+const SnowflakeUtil = require('../util/SnowflakeUtil');
+
+/**
+ * Represents an interaction.
+ * @extends {Base}
+ */
+class Interaction extends Base {
```
HarmoGlace, line 11: There is also the application_id and the version fields in the interaction payload

HarmoGlace (changes requested):

src/structures/Interaction.js
vaporox: `application_id` isn't needed for gateway interactions because `this.client.application.id` should always be present

`version` is documented as "[always `1`](https://discord.com/developers/docs/interactions/slash-commands#interaction)", so I don't see why it should be added either

cactysman (approved):

src/structures/Interaction.js
HarmoGlace: The version field should be implemented, the library purpose is to support all api features right? In the future it might be usefull.

vaporox: If it ever becomes useful, it can be added to the interaction class later

docs/examples/commands.md
```diff
@@ -0,0 +1,52 @@
...
+    required: true,
+  }],
+};
+
+client.once('ready', () => {
+  // Creating a global command
+  client.application.commands.create(commandData);
+
+  // Creating a guild-specific command
+  client.guilds.cache.get('id').commands.create(commandData);
```
Fyko, line 27: Would it be possible to also impl this on the manager?

docs/examples/commands.md
```diff
@@ -0,0 +1,52 @@
...
+});
+```
+
+And that's it! As soon as your client gets ready, it will register the `echo` command.
```
Fyko, line 31:
```suggestion
And that's it! As soon as your client is ready, it will register the `echo` command.
```
Using similar terminology to the guide

src/structures/CommandInteraction.js
```diff
@@ -0,0 +1,195 @@
...
+
+  /**
+   * Edits the initial reply to this interaction.
+   * @param {string|APIMessage|MessageEmbed|MessageEmbed[]} content The new content for the message
+   * @param {WebhookEditMessageOptions} [options] The options to provide
+   * @returns {Promise<Message|Object>}
+   */
+  async editReply(content, options) {
+    const raw = await this.webhook.editMessage('@original', content, options);
+    return this.channel?.messages.add(raw) ?? raw;
```
izexi, line 126:
```suggestion
    return this.webhook.editMessage('@original', content, options);
```
`Webhook#editMessage()` already handles this, see [here](https://github.com/discordjs/discord.js/blob/master/src/structures/Webhook.js#L246-L254) for reference.

izexi (commented):
As mentioned on https://github.com/discordjs/discord.js/pull/5530#issuecomment-820249519: documenting `'@orginal'` as a possible param for `Webhook#editMessage/deleteMessage()` (and `fetchMessage` after the referenced PR gets merged) would be more suited for the scope of this PR.

This should also be reflected on the typings since after #5157 gets merged, `'@orginal'` wouldn't comply with `Snowflake` as a param anymore.

src/structures/CommandInteraction.js
vaporox: Taken from the description:

> Returns the raw message data if the webhook was instantiated as a WebhookClient or if the channel is uncached, otherwise a Message will be returned

And since `interaction.webhook` is a WebhookClient, this should stay here

docs/examples/commands.md
vaporox: What exactly do you mean? `Guild#commands` is a manager as well

src/structures/CommandInteraction.js
izexi: Ah yeah sorry mb, I overlooked that. Nevertheless, this should still probably also play it safe and follow the same pattern by returning a patched clone to avoid race conditions with the gateway event (which could result in both params of `Client:messageUpdate` to be the updated message)

ckohen: on that note, why is it a WebhookClient? I do not see a benefit to having a separate client if it does not need, and there is a downside that is ratelimit handling being separated out (i.e. global ratelimits if hit are rediscovered on interactions).

And since we're on that note, the jsdoc references to WebhookClient#deleteMessage and WebhookClient#editMessage will not work, that class does not document the properties.

central419:
excited for this, is there a pre-release available to use?

tipakA:
The PR itself could count as one, you can install from the branch this PR is on.

cactysman:
> excited for this, is there a pre-release available to use?

```bash
npm i discordjs/discord.js#5ef3e933d4d9bdd8221c6fe85a48c30d07b01d8d
```

vaporox:
No, you shouldn't install from a specific commit because there will still be more commits to this PR, install it from the feature branch `npm i vaporox/discord.js#feat-interactions`

HarmoGlace:
You should stick to a commit, as they may be some futures changes which could cause a breakage in your code, as it is still not merged, and update once you tested it. Note that you should not use it in production currently.

AR-Student824:
can you document this a bit?

iShibi:
> 
> 
> can you document this a bit?

They have documented all of it along with the code. If you meant how you to use the interactions then either:
1) See this [example](https://github.com/vaporox/discord.js/blob/feat-interactions/docs/examples/commands.md) which they linked above for a quick intro.
2) Or follow [this PR](https://github.com/discordjs/guide/pull/638) that adds tutorial for interactions. You can see the preview of that WIP PR [here](https://deploy-preview-638--discordjs-guide.netlify.app/interactions/registering-slash-commands.html).

ChosenQuill:
Might have found a potential bug. When replying to the interaction with a message embed, options, it sends a message of the object as ``[object Object]`` instead of the actual embed.
 Example:
```ts
        interaction.reply(
            new MessageEmbed({
                title: "Success",
                color: "GREEN"
            }),
            {ephemeral: true}
        );
```

I think it might have to do something with ``APIMessage#transformOptions`` replacing the options with content (or not in this case) during ``APIMessage#create``. Just a guess.

ignis05:
In my experience this only happens if ephemeral is set to true. I'm not sure if discord supports ephemeral embeds, it might be an issue on discord's side.

advaith1:
Discord does not support embeds or attachments in ephemeral messages

vaporox:
@ChosenQuill That's not a bug, it's simply an overload not supported by the library. Same holds true for `channel.send()` and similar. If you want both a message embed and options, use one object and pass the embed as `options.embed`.

typings/index.d.ts
```diff
@@ -355,6 +406,28 @@ declare module 'discord.js' {
...
+    public replied: boolean;
+    public webhook: WebhookClient;
+    public defer(ephemeral?: boolean): Promise<void>;
+    public deleteReply(): Promise<void>;
+    public editReply(
+      content: string | APIMessage | WebhookEditMessageOptions | MessageEmbed | MessageEmbed[],
+      options?: WebhookEditMessageOptions,
+    ): Promise<Message | RawMessage>;
+    public fetchReply(): Promise<Message | RawMessage>;
+    public reply(
```
NotSugden, line 424: should add some more overloads for this, as for instance `reply(MessageEmbed, InteractionReplyOptions)` or `reply(APIMessage, InteractionReplyoptions)` etc is not valid, which is accepted by the overload

NotSugden (changes requested):

src/structures/CommandInteraction.js
vaporox: Fixed the jsdoc references, the reason why there's a webhook client is so that the methods do not need to be re-implemented. While that'd be easily doable for `editReply()` and `deleteReply()`, the client can also be used for followup messages which would mean even more methods on the `CommandInteraction` class.

ChosenQuill:
Found the issue. [discord/discord-api-doc#2318](https://github.com/discord/discord-api-docs/issues/2318). Surprisingly, Discord still yet to support it even though they included it in various docs. Maybe a typings change might help for now. Thanks for the clarification.

Dragoteryx:
The `channelID` property of interactions has been defined as `string | null` in the typings, however it should be `string` as command interactions always have that property defined. (there is no way to send a command interaction without typing the command *somewhere*)

This also applies to the `channel` property, although if the command has been used in a DM channel it should fetch it first.

monbrey:
> The `channelID` property of interactions has been defined as `string | null` in the typings, however it should be `string` as command interactions always have that property defined. (there is no way to send a command interaction without typing the command _somewhere_)

Following Discord's typings it's listed as optional: <https://discord.com/developers/docs/interactions/slash-commands#interaction> therefore `string | null` is correct per API spec, as they also haven't marked it being always present on ApplicationCommand interaction types.

> This also applies to the `channel` property, although if the command has been used in a DM channel it should fetch it first.

Discord.js does not automatically fetch things for you - we have Partials to handle emitting event for uncached structures, allowing the user to choose to fetch it themselves.

src/util/Constants.js
```diff
@@ -731,6 +739,62 @@ exports.WebhookTypes = [
...
+  null,
+  'SUB_COMMAND',
+  'SUB_COMMAND_GROUP',
+  'STRING',
+  'INTEGER',
+  'BOOLEAN',
+  'USER',
+  'CHANNEL',
+  'ROLE',
+]);
```
advaith1, line 764:
```suggestion
/**
 * The type of an {@link ApplicationCommandOption} object:
 * * SUB_COMMAND
 * * SUB_COMMAND_GROUP
 * * STRING
 * * INTEGER
 * * BOOLEAN
 * * USER
 * * CHANNEL
 * * ROLE
 * * MENTIONABLE
 * @typedef {string} ApplicationCommandOptionType
 */
exports.ApplicationCommandOptionTypes = createEnum([
  null,
  'SUB_COMMAND',
  'SUB_COMMAND_GROUP',
  'STRING',
  'INTEGER',
  'BOOLEAN',
  'USER',
  'CHANNEL',
  'ROLE',
  'MENTIONABLE',
]);
```

typings/index.d.ts
```diff
@@ -1,3 +1,19 @@
+declare enum ApplicationCommandOptionTypes {
+  SUB_COMMAND = 1,
+  SUB_COMMAND_GROUP = 2,
+  STRING = 3,
+  INTEGER = 4,
+  BOOLEAN = 5,
+  USER = 6,
+  CHANNEL = 7,
+  ROLE = 8,
+}
```
advaith1, line 10:
```suggestion
declare enum ApplicationCommandOptionTypes {
  SUB_COMMAND = 1,
  SUB_COMMAND_GROUP = 2,
  STRING = 3,
  INTEGER = 4,
  BOOLEAN = 5,
  USER = 6,
  CHANNEL = 7,
  ROLE = 8,
  MENTIONABLE = 9,
}
```
https://github.com/discord/discord-api-docs/pull/2853

ckohen (commented):
`CommandInteractionr#reply` should take options.embeds, not options.embed
Also, APIMessage should check isWebhook || isInteraction for setting the embeds array

src/structures/CommandInteraction.js
```diff
@@ -0,0 +1,195 @@
...
+        data,
+      },
+      files,
+    });
+    this.replied = true;
+  }
+
+  /**
+   * Fetches the initial reply to this interaction.
+   * @returns {Promise<Message|Object>}
```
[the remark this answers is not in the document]
iCrawl: The mentioned PR has been merged.

src/structures/CommandInteraction.js
```diff
@@ -0,0 +1,195 @@
...
+   */
+  async fetchReply() {
+    const raw = await this.client.api.webhooks(this.applicationID, this.token).messages('@original').get();
+    return this.channel?.messages.add(raw) ?? raw;
```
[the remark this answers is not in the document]
iCrawl: The mentioned PR has been merged.

iCrawl:
This needs a rebase.

zeel01:
Relevant information regarding slash command permissions:
https://github.com/discord/discord-api-docs/blob/master/docs/Change_Log.md#slash-command-permissions

SNvMK:
Embeds are supported in ephemeral messages. have u mentioned it here?

src/structures/CommandInteraction.js
```diff
@@ -0,0 +1,222 @@
...
+        },
+      },
+    });
+    this.deferred = true;
+  }
+
+  /**
+   * Options for a reply to an interaction.
+   * @property {boolean} [ephemeral] Whether the reply should be ephemeral
+   * @typedef {WebhookMessageOptions} InteractionReplyOptions
```
iShibi, line 95:
```suggestion
   * Options for a reply to an interaction.
   * @typedef {WebhookMessageOptions} InteractionReplyOptions
   * @property {boolean} [ephemeral] Whether the reply should be ephemeral
```
Shouldn't the typedef be above the property?

iShibi:
Is there a reason why `username` and `avatarURL` fields weren't removed from `WebhookMessageOptions` when creating `InteractionReplyOptions`?

src/structures/Interaction.js
```diff
@@ -0,0 +1,117 @@
...
+   */
+  get guild() {
+    return this.client.guilds.cache.get(this.guildID) ?? null;
+  }
+
+  /**
+   * Indicates whether this interaction is a command interaction.
+   * @returns {boolean}
+   */
+  isCommand() {
```
amishshah, line 112: I think this would be better as a getter

amishshah (changes requested):
Just one change that I've spotted :smile: rest looks great

src/structures/Interaction.js
vaporox: Agreed, however this method acts as a type guard:

https://github.com/discordjs/discord.js/blob/69dfce33a972305aa0d1b807ce92b75b9129c32a/typings/index.d.ts#L1065

And those aren't availbale for getters

amishshah (approved):

iShibi:
Typings for `InteractionReplyOptions` look good now but the jsdoc for it still need to be fixed. Since, jsdoc doesn't have any `omit` feature, I believe the only way to do this will be to create a whole new typedef without using `WebhookMessageOptions`. Same thing was done for `WebhookMessageOptions`.

src/managers/ApplicationCommandManager.js
```diff
@@ -0,0 +1,307 @@
...
+   */
+  async edit(command, data) {
+    const id = this.resolveID(command);
+    if (!id) throw new TypeError('INVALID_TYPE', 'command', 'ApplicationCommandResolvable');
+
+    const raw = {};
+    if (data.name) raw.name = data.name;
+    if (data.description) raw.description = data.description;
+    if (data.options) raw.options = data.options.map(o => ApplicationCommand.transformOption(o));
+    if (data.defaultPermission) raw.default_permission = data.defaultPermission;
```
vladfrangu, line 146: This will fail if you pass defaultPermission as `false`

src/managers/ApplicationCommandManager.js
```diff
@@ -0,0 +1,307 @@
...
+   *   {
+   *     id: '876543210987654321',
+   *     type: 'USER',
+   *     permission: false,
+   *   },
+   * ])
+   *   .then(console.log)
+   *   .catch(console.error);
+   * @example
+   * // Set the permissions for all commands
```
vladfrangu, line 233:
```suggestion
   * // Set the permissions for multiple commands
```

typings/index.d.ts
```diff
@@ -1,3 +1,20 @@
+declare enum ApplicationCommandOptionTypes {
```
vladfrangu, line 1: Why don't we import the enums from discord-api-types?

vladfrangu (changes requested):

src/managers/ApplicationCommandManager.js
kyranet: Wouldn't it be better to drop the `if`'s? There are cases in which we might want to pass undefined/null, and undefined properties don't get serialized in JSON either way.

vaporox: > Wouldn't it be better to drop the if's? There are cases in which we might want to pass undefined/null, and undefined properties don't get serialized in JSON either way.

Sounds good to me, however I'll keep creating a new object from the provided data to filter out invalid properties

src/managers/ApplicationCommandManager.js
vaporox: I chose the current wording to emphasize that it also clears the permissions of the commands that weren't edited. Not sure which version is more fitting here

typings/index.d.ts
vaporox: Probably because of consistency (or because it hasnt been a dependency yet when this was added), I'll change that

src/client/websocket/handlers/INTERACTION_CREATE.js
```diff
@@ -0,0 +1,23 @@
...
+    const CommandInteraction = Structures.get('CommandInteraction');
+
+    const interaction = new CommandInteraction(client, data);
+
+    /**
+     * Emitted when an interaction is created.
+     * @event Client#interaction
+     * @param {Interaction} interaction The interaction which was created
+     */
+    client.emit(Events.INTERACTION_CREATE, interaction);
```
VoltrexMaster, line 18:
```suggestion
    const Interaction = new CommandInteraction(client, data);

    /**
     * Emitted when an interaction is created.
     * @event Client#interaction
     * @param {Interaction} interaction The interaction which was created
     */
    client.emit(Events.INTERACTION_CREATE, Interaction);
```
Variable `interaction` is not declared with an uppercase `i` making the parameter type in the JSDoc appear as `any`.

NotSugden: the type in the JSDoc should really be `CommandInteraction` and the variable name should stay as it is

ckohen: It won't always be a `CommandInteraction` though, so it should stay as is, e.g. `ButtonInteraction` will be coming soon

VoltrexMaster: I think that's better yea, edited.

vaporox: Basically what @ckohen just said. There will be more types of interactions in the future, so `Interaction` is more fitting

VoltrexMaster: @ckohen Both would be different interaction types though, so we could add them as
`@param {CommandInteraction | ButtonInteraction} interaction`
in the future when its out. Or just wait and add both at the same time if that's what you're referring to.

ckohen: All Interactions will extend the Interaction class, this is valid jsdoc and is how other extended classes (e.g. channelCreate event) are documented

VoltrexMaster: Ah I see, then seems good as it is.

trulyronak:
Curious — what still needs to be done to get slash commands in, and how can outside developers in the community help? Really looking forward to being able to use discord.js for slash commands, and would be interested in helping out

monbrey:
> 
> 
> Curious — what still needs to be done to get slash commands in, and how can outside developers in the community help? Really looking forward to being able to use discord.js for slash commands, and would be interested in helping out

There's nothing specific that needs to be done that wouldn't already be in the comments here. The PR is pending reviews and approvals from the maintainers.

We've also had to keep up with ongoing changes from Discord - the API has only reached a state of stability quite recently.

src/rest/RESTManager.js
```diff
@@ -18,9 +18,10 @@ class RESTManager {
     this.globalReset = null;
     this.globalDelay = null;
     if (client.options.restSweepInterval > 0) {
-      client.setInterval(() => {
+      const interval = client.setInterval(() => {
```
SpaceEEC, line 21: Couldn't you just call `unref` on the returned interval?
Since the variable isn't used otherwise.

SpaceEEC (approved):

src/rest/RESTManager.js
vladfrangu: Eh, it's more readable this way... 🤷‍♂️

typings/index.d.ts
```diff
@@ -21,21 +21,41 @@ declare enum ChannelTypes {
...
 declare enum OverwriteTypes {
   role = 0,
   member = 1,
 }
 
+declare enum StickerFormatTypes {
+  PNG = 1,
+  APNG = 2,
+  LOTTIE = 3,
+}
```
vladfrangu, line 44: Import from dapi-typessss

vaporox: Wouldn't that be out of scope? There's probably a lot more that can be imported

vladfrangu: InteractionResponseTypes and InteractionTypes tho?

vaporox: Those are pascal case in discord-api-types instead of fully upper case, so not the same thing

vladfrangu (approved):

iCrawl (approved):

neolectron:
🎉

gruselhaus:
Wohoooo 🎉

Dcai169:
Any eta on when this will be part of a release?

kyranet:
> Any eta on when this will be part of a release?

Soon:tm:, we'll go thru all the `semver: major` PRs and review/merge those, then we'll release v13.

EikelFlo:
If someone wanted to use it as I do npm i discordjs/discord.js#master works just fine

AnImperialMan:
> If someone wanted to use it as I do npm i discordjs/discord.js#master works just fine

you can just use npm i discordjs/discord.js because master is default branch

docs/examples/commands.md
```diff
@@ -0,0 +1,52 @@
...
+
+```js
+client.on('interaction', interaction => {
+  // If the interaction isn't a slash command, return
+  if (!interaction.isCommand()) return;
+
+  // Check if it is the correct command
+  if (interaction.commandName === 'echo') {
+    // Get the input of the user
+    const input = interaction.options[0].value;
```
dound, line 45: This should be `const input = interaction.options.get('input').value`

monbrey: This PR was already merged some time ago, it makes no sense to continue reviewing it.
This line has already been changed by #5705.

dound: Thanks for the heads up! I stumbled on this PR from Google and had a heck of time tracking down better documentation since this work hasn't made it to a stable release just yet. I'll check out #5705 and work from there. Thank you monbrey!

iShibi: That file along with other docs were removed in f8703e3e59839d7380795ecf719d5bac691860a4. So, there is nothing to change now, thanks for reporting the issue anyway 👍

PR #5451, opened 2021-03-31
Update Ready.js

NSBCypher:


kyranet:
Thank you for your contribution, but we do not accept Pull Requests to v11, and that code affects user accounts, which are against Discord's Terms of Service.

PR #5457, opened 2021-03-31
fix(InviteDelete): guild can be missing

almostSouji:

The recent addition of stage channels has brought up a bug in our handling of the `INVITE_DELETE`. We currently only return if both the channel as well as the guild can not be found in the cache, however, according to [upstream documentation
](https://discord.com/developers/docs/topics/gateway#invite-delete) `guild_id?` can be missing.

We should instead return if the channel can not be found, no matter if the guild can be. 

Trivia: came to light, since currently stage channels never enter the cache to begin with, meaning it can never be retrieved here.


iCrawl:
🤙

iCrawl (approved):

infinitytmbots:
The issue still persists even after updating the library

PR #5458, opened 2021-04-01
feat(typings): explicitly type PremiumTier and Collectors

Vendicated:

CollectionFilter is now generic. All collectors are now properly typed. For example the filter for `channel.awaitMessages` is now `(arg0: Message) => boolean`

PremiumTier is now typed explicitly instead of simply as number

The only case in which these changes are breaking is if the user did something wrong in which case their code would break anyway

### DIFF
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -294,7 +294,7 @@ declare module 'discord.js' {
   }
 
   export abstract class Collector<K, V> extends EventEmitter {
-    constructor(client: Client, filter: CollectorFilter, options?: CollectorOptions);
+    constructor(client: Client, filter: CollectorFilter<[V]>, options?: CollectorOptions);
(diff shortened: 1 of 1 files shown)


Fyko:
This PR seems to be a duplicate of or superseded by #5447

kyranet:
> This PR seems to be a duplicate of or superseded by #5447

#5447 implements many other changes, but doesn't strict type `PremiumTier`, which this PR also does.

Vendicated:
should I remove my other changes and only leave the premiumtier change?

SpaceEEC (approved):

kyranet (approved):

vladfrangu (approved):
🥰🥰

iCrawl (approved):

PR #5459, opened 2021-04-01
feat(MessageEmbed): remove normalizeField validation

anandre:

`normalizeField` was made with a custom error to throw, this PR removes that so that a proper API error with path will be thrown instead.  This makes the error message more informative.

### DIFF
--- a/src/errors/Messages.js
+++ b/src/errors/Messages.js
@@ -37,9 +37,6 @@ const Messages = {
   COLOR_RANGE: 'Color must be within the range 0 - 16777215 (0xFFFFFF).',
   COLOR_CONVERT: 'Unable to convert color to a number.',
 
-  EMBED_FIELD_NAME: 'MessageEmbed field names may not be empty.',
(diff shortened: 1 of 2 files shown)


vladfrangu (approved):

kyranet (approved):

SpaceEEC:
How is the location of where you are sending the message more informative than where you are adding the malformed field?

kyranet:
> How is the location of where you are sending the message more informative than where you are adding the malformed field?

We don't really check values in many other places, so why should we do so here?

kyranet (changes requested):
You should remove those two constants, as their only references are removed with this PR:

https://github.com/anandre/discord.js/blob/b70c8bde81bb2b29fa8d08d76a830e68211e270b/src/errors/Messages.js#L40-L41

anandre:
Removed.

kyranet:
Just one last thing so ESLint passes... this line needs to be removed, now that it's unused:

image

kyranet (approved):

PR #5466, opened 2021-04-01
docs(MessageEmbed): MessageEmbedOptions typedef

ImRodry:
This PR adds the MessageEmbedOptions typedef to the docs and updates the MessageEmbed data param to link to this typedef instead of a general object. This has no impact in the code and its purpose it to only improve the docs for MessageEmbed. I went off of the interface found in https://github.com/discordjs/discord.js/blob/master/typings/index.d.ts#L2922 but please let me know if there is anything that needs changing before this is merged.

### DIFF
--- a/src/structures/MessageEmbed.js
+++ b/src/structures/MessageEmbed.js
@@ -22,7 +22,7 @@ class MessageEmbed {
    * @property {string} [url] The URL of this embed
    * @property {Date|number} [timestamp] The timestamp of this embed
    * @property {ColorResolvable} [color] The color of this embed
-   * @property {EmbedFieldData} [fields] The fields of this embed
+   * @property {EmbedFieldData[]} [fields] The fields of this embed
    * @property {Array<FileOptions|string|MessageAttachment>} [files] The files of this embed
    * @property {Partial<MessageEmbedAuthor>} [author] The author of this embed
    * @property {Partial<MessageEmbedThumbnail>} [thumbnail] The thumbnail of this embed
--- a/src/structures/MessageEmbed.js
+++ b/src/structures/MessageEmbed.js
@@ -11,7 +11,24 @@ class MessageEmbed {
    * @name MessageEmbed
    * @kind constructor
    * @memberof MessageEmbed
-   * @param {MessageEmbed|Object} [data={}] MessageEmbed to clone or raw embed data
+   * @param {MessageEmbed|MessageEmbedOptions} [data={}] MessageEmbed to clone or raw embed data
+   */
+
+  /**
+   * Represents the possible options for a MessageEmbed
+   * @typedef {Object} MessageEmbedOptions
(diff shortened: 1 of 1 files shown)


src/structures/MessageEmbed.js
```diff
@@ -11,7 +11,24 @@ class MessageEmbed {
...
+
+  /**
+   * Represents the possible options for a MessageEmbed
+   * @typedef {Object} MessageEmbedOptions
+   * @property {string} [title] The title of this embed
+   * @property {string} [description] The description of this embed
+   * @property {string} [url] The URL of this embed
+   * @property {Date|number} [timestamp] The timestamp of this embed
+   * @property {ColorResolvable} [color] The color of this embed
+   * @property {EmbedFieldData} [fields] The fields of this embed
```
NotSugden, line 25:
```suggestion
   * @property {EmbedFieldData[]} [fields] The fields of this embed
```

NotSugden (changes requested):

src/structures/MessageEmbed.js
ImRodry: I missed this, my bad!

src/structures/MessageEmbed.js
```diff
@@ -11,7 +11,24 @@ class MessageEmbed {
    * @name MessageEmbed
    * @kind constructor
    * @memberof MessageEmbed
-   * @param {MessageEmbed|Object} [data={}] MessageEmbed to clone or raw embed data
+   * @param {MessageEmbed|MessageEmbedOptions} [data={}] MessageEmbed to clone or raw embed data
```
ImRodry, line 14:
```suggestion
   * @param {MessageEmbed|MessageEmbedOptions} [data] MessageEmbed to clone or raw embed data
```
I believe there is no need to include a default empty object since this parameter is already optional. I wasn't sure whether you would want this to be changed so I didn't initially but please let me know if I should add this to the PR

vladfrangu: It is needed, since it says the value is optional, and by default its an empty object

vladfrangu (approved):

kyranet (approved):

src/structures/MessageEmbed.js
ImRodry: @vladfrangu the value is optional because of the [], it doesn’t need a default empty object and this was done multiple times across the library.

ImRodry: It is in fact very inconsistent. You can also check this where data is optional and there is no default value because it doesn’t need one since it isn’t changing anything https://github.com/discordjs/discord.js/blob/stable/src/structures/MessageAttachment.js#L12

SpaceEEC (approved):

iCrawl (approved):

PR #5467, opened 2021-04-02
fix(Voice): disconnect if voice channel not cached

amishshah:

Resolves an issue where bots will hard crash if they are moved from a voice channel into a stage channel. As stage channels are not currently cached by the library, the bot will unrecoverably crash when trying to access the channel.

This fix will cause a voice connection to be disconnected if it is updated with a channel that is not cached.


almostSouji (approved):
Confirmed issue and this fixing it

iCrawl (approved):

PR #5476, opened 2021-04-04
fix: typings for 'WebhookMessageOptions'

izexi:
Adds the missing `username` and `avatarURL` props and removes the `replyTo` prop from `WebhookMessageOptions` interface.


csuvajit:
I think you're missing `split` and `code`

izexi:
> I think you're missing `split` and `code`

They exist on the [`MessageOptions`](https://github.com/izexi/discord.js/blob/fix-WebhookMessageOptions/typings/index.d.ts#L3015-L3016) interface, so they're already being picked from there

kyranet (approved):

iCrawl (approved):

PR #5487, opened 2021-04-06
Updated docs for clarity

Warix3:

Added clarity to avoid confusion. I personally confused this variable to mean if the client is the one currently reacting to the message.

### DIFF
--- a/src/structures/MessageReaction.js
+++ b/src/structures/MessageReaction.js
@@ -29,7 +29,7 @@ class MessageReaction {
     this.message = message;
 
     /**
-     * Whether the client has given this reaction
+     * Whether the client has already given this reaction
      * @type {boolean}
      */
     this.me = data.me;


Fyko:
I see no issue with the current description. The anecdote you provide ("I personally confuse...") is actually correct. `MessageReaction#me` means the reaction was performed by the client.

Warix3:
@Fyko yes but it means that the action has been performed once in the past, it doesn't refer to the current reaction.
The way it's worded can be confusing, especially when used with event listener for reactions, it can seem like it's referring to the currently added reaction

almostSouji:
"has given" already denotes that this means "if the client user is in the set of users that have reacted" rather than "if the user reacting is the client user", which would unambiguously note, that the client user reacted in this reaction event instance.

Warix3:
@almostSouji That's technically true but if you read it in the context of an event, then everything has happened in the past and it could be interpreted as the current reaction and not "some reaction in the past". I think it's better to use here "has had reacted" or "has already reacted" because we have a double past in the context of events (messageReactionAdd and remove).

PR #5489, opened 2021-04-06
feat(GuildChannel): make createOverwrite and updateOverwrite not dependent on cache

iShibi:

The above two methods need `userOrRole` to be in cache, so that they can be resolved in order to get the `type` of overwrite. This in some edge cases where the provided `userOrRole` is id of an uncached structure will throw an [error](https://discord.com/channels/222078108977594368/223867697312694272/828965813217918996). This PR adds a new optional parameter `type` for these two methods. Which can be used to manually set the type of the overwrite, making the methods not rely on the cache just to get the `type`.

### DIFF
--- a/src/structures/GuildChannel.js
+++ b/src/structures/GuildChannel.js
@@ -218,6 +218,7 @@ class GuildChannel extends Channel {
    * @param {RoleResolvable|UserResolvable} userOrRole The user or role to update
    * @param {PermissionOverwriteOptions} options The options for the update
    * @param {string} [reason] Reason for creating/editing this overwrite
+   * @param {string} [type] The type of overwrite: `member` or `role`
    * @returns {Promise<GuildChannel>}
    * @example
    * // Update or Create permission overwrites for a message author
@@ -227,15 +228,19 @@ class GuildChannel extends Channel {
    *   .then(channel => console.log(channel.permissionOverwrites.get(message.author.id)))
    *   .catch(console.error);
    */
-  async updateOverwrite(userOrRole, options, reason) {
-    userOrRole = this.guild.roles.resolve(userOrRole) || this.client.users.resolve(userOrRole);
-    if (!userOrRole) return Promise.reject(new TypeError('INVALID_TYPE', 'parameter', 'User nor a Role'));
+  async updateOverwrite(userOrRole, options, reason, type) {
+    let userOrRoleID = this.guild.roles.resolveID(userOrRole) || this.client.users.resolveID(userOrRole);
+    if (!type) {
+      userOrRole = this.guild.roles.resolve(userOrRole) || this.client.users.resolve(userOrRole);
+      if (!userOrRole) return Promise.reject(new TypeError('INVALID_TYPE', 'parameter', 'User nor a Role'));
+      userOrRoleID = userOrRole.id;
+    }
 
-    const existing = this.permissionOverwrites.get(userOrRole.id);
+    const existing = this.permissionOverwrites.get(userOrRoleID);
     if (existing) {
       await existing.update(options, reason);
     } else {
-      await this.createOverwrite(userOrRole, options, reason);
+      await this.createOverwrite(userOrRole, options, reason, type);
     }
     return this;
   }
@@ -245,6 +250,7 @@ class GuildChannel extends Channel {
    * @param {RoleResolvable|UserResolvable} userOrRole The user or role to update
    * @param {PermissionOverwriteOptions} options The options for the update
    * @param {string} [reason] Reason for creating/editing this overwrite
+   * @param {string} [type] The type of overwrite: `member` or `role`
    * @returns {Promise<GuildChannel>}
    * @example
    * // Create or Replace permission overwrites for a message author
@@ -254,19 +260,22 @@ class GuildChannel extends Channel {
    *   .then(channel => console.log(channel.permissionOverwrites.get(message.author.id)))
    *   .catch(console.error);
    */
-  createOverwrite(userOrRole, options, reason) {
-    userOrRole = this.guild.roles.resolve(userOrRole) || this.client.users.resolve(userOrRole);
-    if (!userOrRole) return Promise.reject(new TypeError('INVALID_TYPE', 'parameter', 'User nor a Role'));
-
-    const type = userOrRole instanceof Role ? OverwriteTypes.role : OverwriteTypes.member;
+  createOverwrite(userOrRole, options, reason, type) {
+    let userOrRoleID = this.guild.roles.resolveID(userOrRole) || this.client.users.resolveID(userOrRole);
+    if (!type) {
+      userOrRole = this.guild.roles.resolve(userOrRole) || this.client.users.resolve(userOrRole);
+      if (!userOrRole) return Promise.reject(new TypeError('INVALID_TYPE', 'parameter', 'User nor a Role'));
+      userOrRoleID = userOrRole.id;
+      type = userOrRole instanceof Role ? OverwriteTypes.role : OverwriteTypes.member;
+    }
     const { allow, deny } = PermissionOverwrites.resolveOverwriteOptions(options);
 
     return this.client.api
       .channels(this.id)
-      .permissions(userOrRole.id)
+      .permissions(userOrRoleID)
       .put({
         data: {
-          id: userOrRole.id,
+          id: userOrRoleID,
           type,
           allow,
           deny,


src/structures/GuildChannel.js
```diff
@@ -227,15 +228,19 @@ class GuildChannel extends Channel {
...
-  async updateOverwrite(userOrRole, options, reason) {
-    userOrRole = this.guild.roles.resolve(userOrRole) || this.client.users.resolve(userOrRole);
-    if (!userOrRole) return Promise.reject(new TypeError('INVALID_TYPE', 'parameter', 'User nor a Role'));
+  async updateOverwrite(userOrRole, options, reason, type) {
+    let userOrRoleID = this.guild.roles.resolveID(userOrRole) || this.client.users.resolveID(userOrRole);
+    if (!type) {
+      userOrRole = this.guild.roles.resolve(userOrRole) || this.client.users.resolve(userOrRole);
+      if (!userOrRole) return Promise.reject(new TypeError('INVALID_TYPE', 'parameter', 'User nor a Role'));
+      userOrRoleID = userOrRole.id;
+    }
```
NotSugden, line 237: the same check is ran in `createOverwrite` so whats the ppint in adding it here aswell

NotSugden (changes requested):

src/structures/GuildChannel.js
```diff
@@ -227,15 +228,19 @@ class GuildChannel extends Channel {
    *   .then(channel => console.log(channel.permissionOverwrites.get(message.author.id)))
    *   .catch(console.error);
    */
-  async updateOverwrite(userOrRole, options, reason) {
-    userOrRole = this.guild.roles.resolve(userOrRole) || this.client.users.resolve(userOrRole);
-    if (!userOrRole) return Promise.reject(new TypeError('INVALID_TYPE', 'parameter', 'User nor a Role'));
+  async updateOverwrite(userOrRole, options, reason, type) {
```
NotSugden, line 231: might it also actually be worth combining both `reason` and `type` into a single parameter w/ an object? because both are optional

NotSugden (changes requested):
typings need updating too

src/structures/GuildChannel.js
iShibi: Removed it in 319c312

src/structures/GuildChannel.js
iShibi: Sure, just need some time to come up with a good name and a night's sleep :) 👍

ckohen:
I'd just like to point out that this fixes 2 of 8 methods that have this issue, most of the other ones take a role which is more likely to be cached, but there are still edge cases. I really like the way this was handled, not sure if this PR would be a good place to make changes to the other methods as well.

For reference, you can see these other methods in https://github.com/discordjs/discord.js/discussions/5242

src/structures/GuildChannel.js
iShibi: Added the suggested changes in 0bacaa8. Just want to make sure they are okay before I add the typings for them.

src/structures/GuildChannel.js
```diff
@@ -213,11 +213,19 @@ class GuildChannel extends Channel {
     return this;
   }
 
+  /**
+   * Extra information about the overwrite
+   * @typedef {Object} ExtraOverwriteInfo
+   * @property {string} [reason] Reason for creating/editing this overwrite
+   * @property {string} [type] The type of overwrite, either `member` or `role`. Use this to bypass automatic
```
SpaceEEC, line 220: While a string still seems to work (for now), [the documentation specifies an integer](https://discord.com/developers/docs/resources/channel#edit-channel-permissions).

```suggestion
   * @property {number} [type] The type of overwrite, either `0` for a role or `1` for a member. Use this to bypass automatic
```

SpaceEEC (dismissed):

src/structures/GuildChannel.js
```diff
@@ -213,11 +213,19 @@ class GuildChannel extends Channel {
     return this;
   }
 
+  /**
+   * Extra information about the overwrite
+   * @typedef {Object} ExtraOverwriteInfo
+   * @property {string} [reason] Reason for creating/editing this overwrite
+   * @property {number} [type] The type of overwrite, either `0` for a role or `1` for a member. Use this to bypass automatic
+   * resolution of type that results in an error for uncached structure
```
kyranet, line 221:
```suggestion
   * @property {number} [type] The type of overwrite, either `0` for a role or `1` for a member. Use this to bypass
   * automatic resolution of type that results in an error for uncached structure
```

kyranet (changes requested):
Just one change to make CI pass

kyranet (approved):

src/structures/GuildChannel.js
```diff
@@ -254,19 +261,23 @@ class GuildChannel extends Channel {
...
    */
-  createOverwrite(userOrRole, options, reason) {
-    userOrRole = this.guild.roles.resolve(userOrRole) || this.client.users.resolve(userOrRole);
-    if (!userOrRole) return Promise.reject(new TypeError('INVALID_TYPE', 'parameter', 'User nor a Role'));
-
-    const type = userOrRole instanceof Role ? OverwriteTypes.role : OverwriteTypes.member;
+  createOverwrite(userOrRole, options, extraInfo = {}) {
+    let userOrRoleID = this.guild.roles.resolveID(userOrRole) || this.client.users.resolveID(userOrRole);
+    let { type, reason } = extraInfo;
+    if (!type) {
```
vladfrangu, line 267: `type == 0` will make `!type` return true.

vladfrangu (dismissed):

src/structures/GuildChannel.js
iShibi: aah, didn't notice that after Space suggested using number in place of string. I'll fix it 😅

src/structures/GuildChannel.js
```diff
@@ -254,19 +261,23 @@ class GuildChannel extends Channel {
...
    */
-  createOverwrite(userOrRole, options, reason) {
-    userOrRole = this.guild.roles.resolve(userOrRole) || this.client.users.resolve(userOrRole);
-    if (!userOrRole) return Promise.reject(new TypeError('INVALID_TYPE', 'parameter', 'User nor a Role'));
-
-    const type = userOrRole instanceof Role ? OverwriteTypes.role : OverwriteTypes.member;
+  createOverwrite(userOrRole, options, extraInfo = {}) {
+    let userOrRoleID = this.guild.roles.resolveID(userOrRole) || this.client.users.resolveID(userOrRole);
+    let { type, reason } = extraInfo;
+    if (!(type === 0 || type === 1)) {
```
kyranet, line 267:
```suggestion
    if (typeof type !== 'number') {
```

kyranet (dismissed):

src/structures/GuildChannel.js
iShibi: This will let any number to pass which will result in an API error that only value 0 or 1 is allowed.

kyranet: In case that happens, Discord will just throw an error.

Similarly, if Discord were to add more types, this method wouldn't need changes (as long as we provide the ID and the new type), so it's future-proofed.

iCrawl (approved):

vladfrangu:
This PR clashes with https://github.com/discordjs/discord.js/pull/5318 (if the mentioned PR gets merged first, this will need a full rebase)... @iCrawl how should we handle this?

iCrawl:
Depends on what makes more sense to merge. Does a whole new manager really make sense here, except to abstract things more thoroughly?

src/structures/GuildChannel.js
```diff
@@ -213,11 +213,19 @@ class GuildChannel extends Channel {
     return this;
   }
 
+  /**
+   * Extra information about the overwrite
+   * @typedef {Object} ExtraOverwriteInfo
```
kyranet, line 218:
```suggestion
   * @typedef {Object} GuildChannelOverwriteOptions
```

src/structures/GuildChannel.js
```diff
@@ -213,11 +213,19 @@ class GuildChannel extends Channel {
...
+   * @property {number} [type] The type of overwrite, either `0` for a role or `1` for a member. Use this to bypass
+   * automatic resolution of type that results in an error for uncached structure
+   */
+
   /**
    * Updates permission overwrites for a user or role in this channel, or creates an entry if not already present.
    * @param {RoleResolvable|UserResolvable} userOrRole The user or role to update
    * @param {PermissionOverwriteOptions} options The options for the update
-   * @param {string} [reason] Reason for creating/editing this overwrite
+   * @param {ExtraOverwriteInfo} [extraInfo] The extra information for the update
```
kyranet, line 228:
```suggestion
   * @param {GuildChannelOverwriteOptions} [options] The extra information for the update
```

kyranet: Same changes should go to the code and the `createOverwrite` method.

src/structures/GuildChannel.js
kyranet: I'm not sure about this name, it's used in both `createOverwrite` and `updateOverwrite`.

cc: @vladfrangu @iCrawl

kyranet (changes requested):

src/structures/GuildChannel.js
SpaceEEC: 👍 for `GuildChannelOverwriteOptions`

SpaceEEC (approved):

kyranet (approved):

PR #5495, opened 2021-04-07
fix(Resolvables): valid resolvables throw error when uncached

iShibi:

answers/fixes: #5242

This PR is an extension of #5489 (unlike that PR. this one doesn't add any new parameter) and aims to fix errors that arise due IDs that belong to structures which aren't cached. This happens because the library tries to check for the validity of the resolvable by using `<Manager>#resolve`, this doesn't work in case of the resolvable being uncached. All of this can be fixed by using `<Manager>#resolveID` instead. This way we let the API do the validity test. I've tested **most** of the changes and the errors returned by API do a good job in letting the user know what's wrong.

⚠️ *The only change that I haven't tested belongs to `Guild#addMember` due to it being related to OAuth2. Even though it should work, It would be nice if someone can test it :)*

- Code changes **(except one)** have been tested against the Discord API, or there are no code changes

### DIFF
--- a/src/managers/GuildEmojiManager.js
+++ b/src/managers/GuildEmojiManager.js
@@ -51,7 +51,7 @@ class GuildEmojiManager extends BaseGuildEmojiManager {
     if (roles) {
       data.roles = [];
       for (let role of roles instanceof Collection ? roles.values() : roles) {
-        let roleID = this.guild.roles.resolveID(role);
+        const roleID = this.guild.roles.resolveID(role);
         if (!roleID) {
           return Promise.reject(
             new TypeError('INVALID_TYPE', 'options.roles', 'Array or Collection of Roles or Snowflakes', true),
--- a/src/managers/GuildEmojiManager.js
+++ b/src/managers/GuildEmojiManager.js
@@ -51,13 +51,13 @@ class GuildEmojiManager extends BaseGuildEmojiManager {
     if (roles) {
       data.roles = [];
       for (let role of roles instanceof Collection ? roles.values() : roles) {
-        role = this.guild.roles.resolve(role);
-        if (!role) {
+        let roleID = this.guild.roles.resolveID(role);
+        if (!roleID) {
           return Promise.reject(
             new TypeError('INVALID_TYPE', 'options.roles', 'Array or Collection of Roles or Snowflakes', true),
           );
         }
-        data.roles.push(role.id);
+        data.roles.push(roleID);
       }
     }
 
--- a/src/structures/GuildMember.js
+++ b/src/structures/GuildMember.js
@@ -283,7 +283,7 @@ class GuildMember extends Base {
    */
   async edit(data, reason) {
     if (data.channel) {
-      let voiceChannelID = this.guild.channels.resolveID(data.channel);
+      const voiceChannelID = this.guild.channels.resolveID(data.channel);
       if (!voiceChannelID) {
         throw new Error('GUILD_VOICE_CHANNEL_RESOLVE');
       }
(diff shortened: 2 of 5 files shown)


src/managers/GuildEmojiManager.js
```diff
@@ -51,13 +51,13 @@ class GuildEmojiManager extends BaseGuildEmojiManager {
     if (roles) {
       data.roles = [];
       for (let role of roles instanceof Collection ? roles.values() : roles) {
-        role = this.guild.roles.resolve(role);
-        if (!role) {
+        let roleID = this.guild.roles.resolveID(role);
```
NotSugden, line 54:
```suggestion
        const roleID = this.guild.roles.resolveID(role);
```

src/structures/GuildMember.js
```diff
@@ -283,11 +283,11 @@ class GuildMember extends Base {
    */
   async edit(data, reason) {
     if (data.channel) {
-      data.channel = this.guild.channels.resolve(data.channel);
-      if (!data.channel || data.channel.type !== 'voice') {
+      let voiceChannelID = this.guild.channels.resolveID(data.channel);
```
NotSugden, line 286:
```suggestion
      const voiceChannelID = this.guild.channels.resolveID(data.channel);
```
should there also be some check if the channel is voice-based if its cached ?

NotSugden (changes requested):

src/structures/GuildMember.js
iShibi: We could but these are some points that I had in my mind related to this:
1) The check will need another `resolve`, that might slow down the method a little bit.
2) Right now if the channel passed isn't a VC (cached or uncached), it will throw the same error that is returned by the API. While after adding the check, the errors will be **slightly** different depending on the channel passed was cached (lib error) or uncached (API error), which seems a little inconsistent from a user's perspective.

NotSugden: wouldnt nessercerily require another resolve

something like

```js
const voiceChannel = this.guild.channels.cache.get(voiceChannelID);
if (voiceChannel && voiceChannel.type !== 'voice') { ... }```

Comment (iShibi) [reply src/structures/GuildMember.js]: The second point still holds true tho. I feel like letting the API do the work here seems more consistent and also saves us from making that `get` call (which I know isn't that big of a deal but still 😅).

**Comment by ckohen** [comment]
Since #5314 is changing how most of these error drastically I changed those there since vaporox suggested it there as well.

File: src/structures/GuildMember.js [hunk 3]
```diff
@@ -283,11 +283,11 @@ class GuildMember extends Base {
    */
   async edit(data, reason) {
     if (data.channel) {
-      data.channel = this.guild.channels.resolve(data.channel);
-      if (!data.channel || data.channel.type !== 'voice') {
+      const voiceChannelID = this.guild.channels.resolveID(data.channel);
+      if (!voiceChannelID) {
```
SpaceEEC, line 287: Would probably be nice to keep the type check for cached channels.
Not sure if there is some elegant solution for that though.

SpaceEEC (changes requested):

src/structures/GuildMember.js
iShibi: Aight, It's 2-1 now, so I added your suggestion. Check it out and see if there is anything that I forgot :)👍

src/structures/GuildMember.js
iShibi: Added it back in b3ed332. Not sure how elegant it is tho😅

SpaceEEC (approved):

kyranet (approved):

vladfrangu (approved):

iCrawl (approved):

PR #5317, opened 2021-02-12
feat: make changes to PresenceData typings and docs

iShibi:
**Please describe the changes this PR makes and why it should be merged:**

Fixes: #5316 
Also, for some reason, jsdoc description of `ActivityOptions` was missing the `name` property.

**Status and versioning classification:**
- Code changes have been tested against the Discord API, or there are no code changes
- I know how to update typings and have done so, or typings don't need updating

PR #5498, opened 2021-04-08
fix(PresenceData): update docs to reflect activity > activities refactor

MattIPv4:

In #4879 all the presence logic was updated to replace the single `activity` object with an array in `activities`, but the typings & JSDoc were not updated to reflect this. This PR updates those to be accurate.


vaporox:
#5317 already does that

MattIPv4:
Oh, wow, that's further back than I'd looked/expected -- rather annoying to have broken typings & docs in master :(

PR #5504, opened 2021-04-08
feat(GuildManager): allow system channel flags in create

almostSouji:

Allow `systemChannelFlags` to be passed during guild creation as per https://github.com/discord/discord-api-docs/pull/1764

### DIFF
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -2761,6 +2761,7 @@ declare module 'discord.js' {
     roles?: PartialRoleData[];
     systemChannelID?: number;
     verificationLevel?: VerificationLevel | number;
+    systemChannelFlags?: SystemChannelFlagsResolvable;
   }
 
   interface GuildWidget {
--- a/src/managers/GuildManager.js
+++ b/src/managers/GuildManager.js
@@ -16,6 +16,7 @@ const {
 } = require('../util/Constants');
 const DataResolver = require('../util/DataResolver');
 const Permissions = require('../util/Permissions');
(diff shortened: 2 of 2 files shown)


vladfrangu (approved):

typings/index.d.ts
```diff
@@ -2761,6 +2761,7 @@ declare module 'discord.js' {
     roles?: PartialRoleData[];
     systemChannelID?: number;
     verificationLevel?: VerificationLevel | number;
+    systemChannelFlags?: SystemChannelFlagsResolvable;
```
kyranet, line 2764: Please move this property above verificationLevel so they're sorted alphabetically.

kyranet (changes requested):

kyranet (approved):

SpaceEEC:
I'm unable to get this working, the system channel flags are always 0 in the response body / guild create payload.

almostSouji:
- the data we receive from `POST /guilds` has the correct flags set (3)
- the data we receive from the `GUILD_CREATE` websocket event has no flags set (0)
- getting data after the fact from `GET /guilds/:id` also has the correct flags set (3)
- the setting is applied, but the client seems to encounter that issue as well (the toggles show that the messages are enabled, they are however not sent)

So in conclusion that seems to be a discord bug and nothing we can handle on our end (if the GC arrives earlier we use that data in the #create response)

iCrawl (approved):

SpaceEEC (approved):
We'll see if Discord can resolve this.

PR #5505, opened 2021-04-08
feat(Guild): allow description and features in edit

almostSouji:

- `Guild#edit` now allows modification of features (`COMMUNITY`, to enable community features)
- `Guild#edit` now allows the discovery description to be modified

as per https://github.com/discord/discord-api-docs/pull/1764

### DIFF
--- a/src/structures/Guild.js
+++ b/src/structures/Guild.js
@@ -926,6 +926,7 @@ class Guild extends Base {
 
   /**
    * The data for editing a guild.
+   * <info>Only the `COMMUNITY` feature can be modified</info>
    * @typedef {Object} GuildEditData
    * @property {string} [name] The name of the guild
    * @property {string} [region] The region of the guild
@@ -944,6 +945,8 @@ class Guild extends Base {
    * @property {ChannelResolvable} [rulesChannel] The rules channel of the guild
    * @property {ChannelResolvable} [publicUpdatesChannel] The community updates channel of the guild
    * @property {string} [preferredLocale] The preferred locale of the guild
+   * @property {string} [description] The discovery description of the guild
+   * @property {Features[]} [features] The features of the guild
    */
 
   /**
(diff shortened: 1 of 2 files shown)


src/structures/Guild.js
```diff
@@ -926,6 +926,7 @@ class Guild extends Base {
 
   /**
    * The data for editing a guild.
+   * <info>Only the `COMMUNITY` feature can be modified</info>
```
vladfrangu, line 929: Do we want to document this? It'd mean we have to update it every time Discord adds new ones (also pretty sure it's NOT the only one)

vladfrangu (changes requested):

src/structures/Guild.js
almostSouji: every other feature is a privilege you can't self-grant or disable so far

typings/index.d.ts
```diff
@@ -2786,6 +2786,8 @@ declare module 'discord.js' {
     rulesChannel?: ChannelResolvable;
     publicUpdatesChannel?: ChannelResolvable;
     preferredLocale?: string;
+    description?: string;
```
kyranet, line 2789: Those properties should be sorted alphabetically

kyranet (changes requested):

SpaceEEC:
I'm unable to set a description (just silently fails on Discord's end), probably due to not being on discovery.

Further enabling community only works when the payload also includes a rules and public updates channel (does not matter if you previously set them).
Otherwise you are getting the very helpful and totally not misleading error `This feature has been temporarily disabled` (The code is `40006`, if somebody is interested).

I don't think there is much we can do about either though.

SpaceEEC (approved):

typings/index.d.ts
vladfrangu, line 2789: Can't you also set descriptions to null?

almostSouji: it does! fixed

kyranet (approved):

iCrawl (approved):

catcosmicice (approved):

iCrawl (approved):

PR #5507, opened 2021-04-08
feat(Guild): setChannelPositions parent, lockPermissions keys

almostSouji:

- `Guild#setChannelPositions` takes the options `parent` and `lockPermissions` as per https://github.com/discord/discord-api-docs/pull/1776
- This makes the `position` key optional, however is still only a patch level PR, since the old approach (position being mandatory) can still be used, making this non-breaking.
- c732669d49713efa707fe5df9731028f40ffbc2e adds CategoryChannelResolvable for this PR only, #5521 to keep track of and apply the type to the rest of the library

### DIFF
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -2425,7 +2425,7 @@ declare module 'discord.js' {
   interface ChannelPosition {
     channel: ChannelResolvable;
     lockPermissions?: boolean;
-    parent?: ChannelResolvable;
+    parent?: CategoryChannelResolvable;
     position?: number;
   }
 
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -2424,9 +2424,9 @@ declare module 'discord.js' {
 
   interface ChannelPosition {
     channel: ChannelResolvable;
-    position?: number;
-    parent?: ChannelResolvable;
     lockPermissions?: boolean;
+    parent?: ChannelResolvable;
+    position?: number;
   }
 
   type ChannelResolvable = Channel | Snowflake;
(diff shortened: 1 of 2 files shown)


typings/index.d.ts
```diff
@@ -2422,7 +2424,9 @@ declare module 'discord.js' {
 
   interface ChannelPosition {
     channel: ChannelResolvable;
-    position: number;
+    position?: number;
+    parent?: ChannelResolvable;
+    lockPermissions?: boolean;
```
kyranet, line 2429: Properties should be sorted alphabetically:
```suggestion
    lockPermissions?: boolean;
    parent?: ChannelResolvable;
    position?: number;
```

kyranet (changes requested):

kyranet (approved):

iCrawl (approved):

typings/index.d.ts
```diff
@@ -2422,7 +2424,9 @@ declare module 'discord.js' {
 
   interface ChannelPosition {
     channel: ChannelResolvable;
-    position: number;
+    lockPermissions?: boolean;
+    parent?: ChannelResolvable;
```
vladfrangu, line 2428: Why isn't this `CategoryChannelResolvable`?

almostSouji: no such thing exists, all current instances use ChannelResolvable, CategoryResolvable should be introduced after all the stuff that'd need it lands in one central PR, imo

almostSouji: Resolved in c732669d49713efa707fe5df9731028f40ffbc2e for this PR only, #5521 to keep track of and apply the type to the rest of the library

SpaceEEC (approved):

iCrawl (approved):

PR #5508, opened 2021-04-08
feat(Webhook): sourceGuild, sourceChannel, improve owner

almostSouji:

Webhooks can have a `source_guild` and a `source_channel` as per https://github.com/discord/discord-api-docs/pull/1837

- `owner `user data is fully available, meaning we can always construct a user here, and even cache it
- `source_guild` data is not fully available, by not caching but constructing a Guild instance we can leverage some guild methods (invites use the same approach to construct "partial" guilds - very empty guilds)
- `source_channel` for guilds the bot is not on, only consists of name and id, not enough to construct a channel from, meaning the data object default is required here

**Edit: Object fallbacks are needed for WebhookClient**

### DIFF
--- a/src/structures/Webhook.js
+++ b/src/structures/Webhook.js
@@ -67,9 +67,9 @@ class Webhook {
 
     /**
      * The owner of the webhook
-     * @type {?User}
+     * @type {?User|Object}
      */
-    this.owner = data.user ? this.client.users?.add(data.user, true) : null;
+    this.owner = data.user ? this.client.users?.add(data.user) ?? data.user : null;
 
     /**
      * The source guild of the webhook
--- a/src/structures/Webhook.js
+++ b/src/structures/Webhook.js
@@ -73,9 +73,11 @@ class Webhook {
 
     /**
      * The source guild of the webhook
-     * @type {?Guild}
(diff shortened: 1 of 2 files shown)


vaporox:
The purpose of assigning `data.user` directly was for webhook clients that do not have a user manager available, with this change it would just always return `undefined`

kyranet (approved):

src/structures/Webhook.js
```diff
@@ -65,15 +65,23 @@ class Webhook {
...
+     * The owner of the webhook
+     * @type {?User|Object}
+     */
+    this.owner = data.user ? this.client.users?.add(data.user) ?? data.user : null;
+
+    /**
+     * The source guild of the webhook
+     * @type {?Guild}
+     */
+    this.sourceGuild = data.source_guild ? this.client.guilds?.add(data.source_guild, false) : null;
```
vladfrangu, line 78: Should also default to `?? data.source_guild`

vladfrangu (changes requested):

vladfrangu (approved):

iCrawl (approved):

PR #5511, opened 2021-04-09
refactor(RequestHandler): use x-ratelimit-reset-after when present

ckohen:

As of https://github.com/discord/discord-api-docs/issues/2702 the x-ratelimit-reset-after header should be present on routes that have ratelimits. I have verified this to be the case, and that it indeed does have how long until the current reset in seconds, to ms precision.

This PR adds handling for this header, preferring it when it is present. It retains old handling in the event that this header is not present.
It prefers this header due to this method not requiring calculation of the API offset, which is a looser timing due to reliance on the second-precise date header.

This theoretically makes using a lower restTimeOffset more viable as well. I tested on the reactions route with this set to 0 and had no issues.

⚠️ Not all api routes have been tested, a majority of common routes have been.

- Code changes have been ⚠️ mostly ⚠️ tested against the Discord API, or there are no code changes

### DIFF
--- a/src/rest/RequestHandler.js
+++ b/src/rest/RequestHandler.js
@@ -17,7 +17,11 @@ function getAPIOffset(serverDate) {
   return new Date(serverDate).getTime() - Date.now();
 }
 
-function calculateReset(reset, serverDate) {
+function calculateReset(reset, resetAfter, serverDate) {
+  // Use direct reset time when available, server date becomes irrelevant in this case
+  if (resetAfter) {
+    return Date.now() + Number(resetAfter) * 1000;
+  }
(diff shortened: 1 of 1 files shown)


vladfrangu (approved):
Can you also PR this to https://github.com/discordjs/discord.js-next 's REST? 🙏

SpaceEEC (approved):

iCrawl (approved):

ckohen:
> Can you also PR this to https://github.com/discordjs/discord.js-next 's REST? 🙏

@vladfrangu What did you want me to PR? -next already uses this header actually.

PR #5524, opened 2021-04-13
feat(BaseManager): remove stale data in cache

VimHax:

Caches are most effective when the data being stored in them are data which are being required quite often by the client, otherwise, the data stored in the cache will just be hogging up memory without any good justification.

I've implemented a new method for the `BaseManager` class, `poll(lifetime: number): void`, which will remove data in that manager's cache which has been stale for longer than the lifetime provided. The staleness is kept track of by a new property introduced to all the data being stored in the cache, `lastHit: Date`, every time some data gets accessed and the cache gets hit this value will be updated to the current time. When being polled the manager compares the current time to the time stored in the `lastHit` property to determine whether or not the data has outlived it's lifetime. The polling of the managers is managed by the `Client` which owns the manager. A new property has been introduced to the `ClientOptions` object which allows you to set the lifetimes for the data. Just like the implementation for message sweeping, the polling also uses an interval to call the poll method on all the managers.

My implementation has room for optimization, such as not keeping track of the `lastHit` property when the polling has been disabled (which is the case by default)

The idea for this is to keep only the most relevant data in the cache to maximize the cache effectiveness and the relevancy can be controlled by the end-user as they wish.

### DIFF
--- a/src/client/Client.js
+++ b/src/client/Client.js
@@ -569,7 +569,7 @@ class Client extends BaseClient {
     if (typeof options.retryLimit !== 'number' || isNaN(options.retryLimit)) {
       throw new TypeError('CLIENT_INVALID_OPTION', 'retryLimit', 'a number');
     }
-    if (typeof x !== 'undefined') {
+    if (typeof options.cacheData !== 'undefined') {
       if (typeof options.cacheData !== 'object') {
         throw new TypeError('CLIENT_INVALID_OPTION', 'cacheData', 'an object or undefined');
       }
--- a/src/client/Client.js
+++ b/src/client/Client.js
@@ -166,6 +166,19 @@ class Client extends BaseClient {
     if (this.options.messageSweepInterval > 0) {
       this.setInterval(this.sweepMessages.bind(this), this.options.messageSweepInterval * 1000);
     }
+
+    if (this.options.cacheData !== undefined) {
+      let lifetimes = this.options.cacheData.dataLifetime;
+      if (typeof lifetimes === 'number') {
+        lifetimes = {
+          user: lifetimes,
+          channel: lifetimes,
+          guild: lifetimes,
+          emoji: lifetimes,
+        };
+      }
+      this.setInterval(this.sweepCaches.bind(this, lifetimes), this.options.cacheData.pollInterval * 1000);
+    }
   }
 
   /**
@@ -352,6 +365,72 @@ class Client extends BaseClient {
   }
 
   /**
+   * Removes all the stale data in the caches older than the provided lifetimes.
+   * @param {Lifetimes} lifetime Data that has been stale for longer than
+   * this will be removed.
+   * @returns {number} -1 if the lifetime is invalid and 1 on success
+   * @example
+   * // Remove all the stale cached user data older than 1800 seconds
+   * client.sweepCaches({ user: 1800 * 1000 }); // x 1000 to convert to seconds
+   */
+  sweepCaches(lifetime) {
+    if (typeof lifetime !== 'object') {
+      throw new TypeError('INVALID_TYPE', 'lifetime', 'Lifetimes');
+    }
+
+    const { user, channel, guild, emoji } = lifetime;
+    if (user !== undefined && (typeof user !== 'number' || isNaN(user))) {
+      throw new TypeError('INVALID_TYPE', 'lifetime.user', 'number or undefined');
+    }
+    if (channel !== undefined && (typeof channel !== 'number' || isNaN(channel))) {
+      throw new TypeError('INVALID_TYPE', 'lifetime.channel', 'number or undefined');
+    }
+    if (guild !== undefined && (typeof guild !== 'number' || isNaN(guild))) {
+      throw new TypeError('INVALID_TYPE', 'lifetime.guild', 'number or undefined');
+    }
+    if (emoji !== undefined && (typeof emoji !== 'number' || isNaN(emoji))) {
+      throw new TypeError('INVALID_TYPE', 'lifetime.emoji', 'number or undefined');
+    }
+
+    this.emit(Events.DEBUG, 'Cache sweep initiated, polling all the caches');
+
+    if (user !== undefined) {
+      if (user <= 0) {
+        this.emit(Events.DEBUG, "Didn't sweep user cache - lifetime is unlimited");
+      } else {
+        this.users.poll(user);
+      }
+    }
+
+    if (channel !== undefined) {
+      if (channel <= 0) {
+        this.emit(Events.DEBUG, "Didn't sweep channel cache - lifetime is unlimited");
+      } else {
+        this.channels.poll(channel);
+      }
+    }
+
+    if (guild !== undefined) {
+      if (guild <= 0) {
+        this.emit(Events.DEBUG, "Didn't sweep guild cache - lifetime is unlimited");
+      } else {
+        this.guilds.poll(guild);
+      }
+    }
+
+    if (emoji !== undefined) {
+      if (emoji <= 0) {
+        this.emit(Events.DEBUG, "Didn't sweep emoji cache - lifetime is unlimited");
+      } else {
+        this.emojis.poll(emoji);
+      }
+    }
+
+    this.emit(Events.DEBUG, `Swept all the stale data older than ${lifetime / 1000} seconds`);
+    return 1;
+  }
+
+  /**
    * Obtains a guild preview from Discord, available for all guilds the bot is in and all Discoverable guilds.
    * @param {GuildResolvable} guild The guild to fetch the preview for
    * @returns {Promise<GuildPreview>}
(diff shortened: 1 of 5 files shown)


src/client/Client.js
```diff
@@ -351,6 +364,72 @@ class Client extends BaseClient {
...
+    if (emoji !== undefined) {
+      if (emoji <= 0) {
+        this.emit(Events.DEBUG, "Didn't sweep emoji cache - lifetime is unlimited");
+      } else {
+        this.emojis.poll(emoji);
+      }
+    }
+
+    this.emit(Events.DEBUG, `Swept all the stale data older than ${lifetime / 1000} seconds`);
+    return 1;
```
NotSugden, line 430: why return 1 ?

src/client/Client.js
```diff
@@ -492,6 +571,46 @@ class Client extends BaseClient {
...
     }
+    if (options.cacheData !== undefined) {
+      if (typeof options.cacheData !== 'object') {
+        throw new TypeError('CLIENT_INVALID_OPTION', 'The cacheData', 'an object or undefined');
+      }
+      if (typeof options.cacheData.pollInterval !== 'number' || isNaN(options.cacheData.pollInterval)) {
+        throw new TypeError('CLIENT_INVALID_OPTION', 'The cacheData.pollInterval', 'a number');
+      }
+      if (typeof options.cacheData.dataLifetime === 'number') {
+        if (isNaN(options.cacheData.dataLifetime)) {
```
NotSugden, line 582: this could be converted into a single if statement

src/client/Client.js
```diff
@@ -492,6 +571,46 @@ class Client extends BaseClient {
...
+          throw new TypeError('CLIENT_INVALID_OPTION', 'The cacheData.dataLifetime.guild', 'a number or undefined');
+        }
+        if (
+          options.cacheData.dataLifetime.emoji !== undefined &&
+          typeof options.cacheData.dataLifetime.emoji !== 'number'
+        ) {
+          throw new TypeError('CLIENT_INVALID_OPTION', 'The cacheData.dataLifetime.emoji', 'a number or undefined');
+        }
+      } else {
+        throw new TypeError('CLIENT_INVALID_OPTION', 'The cacheData.dataLifetime', 'a number or Lifetimes object');
```
NotSugden, line 611:
```suggestion
        throw new TypeError('CLIENT_INVALID_OPTION', 'cacheData.dataLifetime', 'a number or Lifetimes object');
```
the same for all these occourances (i can spell)

src/managers/BaseManager.js
```diff
@@ -43,11 +43,14 @@ class BaseManager {
 
   add(data, cache = true, { id, extras = [] } = {}) {
     const existing = this.cache.get(id || data.id);
-    if (existing && existing._patch && cache) existing._patch(data);
-    if (existing) return existing;
+    if (existing && existing.entry._patch && cache) existing.entry._patch(data);
```
NotSugden, line 46:
```suggestion
    if (existing?.entry._patch && cache) existing.entry._patch(data);
```

src/managers/BaseManager.js
```diff
@@ -73,6 +76,15 @@ class BaseManager {
     return null;
   }
 
+  /**
+   * Removes all cache entries which have outlived their lifetimes.
+   * @param {number} lifetime The maximum duration (in milliseconds) stale data should live.
+   */
+  poll(lifetime) {
+    const now = new Date();
```
NotSugden, line 84:
```suggestion
    const now = Date.now();
```

NotSugden (changes requested):
the new option also needs to be added to the jsdoc typedef

i also notice `x !== undefined` is used here - granted it makes the code longer but to keep consistency perhaps `typeof x !== 'undefined'` should be used instead ?

src/client/Client.js
VimHax: Whoops! I copied the `Client#sweepMessages` method and modified it for `Client#sweepCaches` but at some point forgot to remove the unnecessary return values, fixed it.

src/client/Client.js
VimHax: Fixed it

VimHax:
Added all the suggestions you mentioned, @NotSugden :)

src/client/Client.js
```diff
@@ -492,6 +569,34 @@ class Client extends BaseClient {
     if (typeof options.retryLimit !== 'number' || isNaN(options.retryLimit)) {
       throw new TypeError('CLIENT_INVALID_OPTION', 'retryLimit', 'a number');
     }
+    if (typeof x !== 'undefined') {
```
BannerBomb, line 572: `x` isn't being defined anywhere else that I can see.

VimHax: :facepalm: I copied the first bit, from Sugden's post and forgot to change the variable :D This is the type of thing TS would've caught like immediately... oh well, I fixed it

src/client/Client.js
```diff
@@ -351,6 +364,70 @@ class Client extends BaseClient {
...
+
+    if (typeof emoji !== 'undefined') {
+      if (emoji <= 0) {
+        this.emit(Events.DEBUG, "Didn't sweep emoji cache - lifetime is unlimited");
+      } else {
+        this.emojis.poll(emoji);
+      }
+    }
+
+    this.emit(Events.DEBUG, `Swept all the stale data older than ${lifetime / 1000} seconds`);
```
BannerBomb, line 428: The lifetime variable would be an object here not a number, based on these lines https://github.com/discordjs/discord.js/blob/c142629e219212206dc647f76a39305b21285901/src/client/Client.js#L376-L380

VimHax: Fixed it (yet another error which TS would've caught)

BannerBomb:
https://github.com/discordjs/discord.js/blob/198e79dc11a6faa03dbaa0ee3bd2dd8328b8a5c9/src/managers/BaseManager.js#L53
I might be overthinking the way the BaseManager works exactly, but with this change wouldn't the way you get from the cache change? Because the way the cache is set up is to have snowflakes mapped to its class. But here you changed it to map a snowflake to an object with the keys entry and lastHit. Currently, to get a value from the cache you would use `cache.get('snowflake here')`, but wouldn't this change it where you'd need to use `cache.get('snowflake here').entry`?

VimHax:
> https://github.com/discordjs/discord.js/blob/198e79dc11a6faa03dbaa0ee3bd2dd8328b8a5c9/src/managers/BaseManager.js#L53
> 
> I might be overthinking the way the BaseManager works exactly, but with this change wouldn't the way you get from the cache change? Because the way the cache is set up is to have snowflakes mapped to its class. But here you changed it to map a snowflake to an object with the keys entry and lastHit. Currently, to get a value from the cache you would use `cache.get('snowflake here')`, but wouldn't this change it where you'd need to use `cache.get('snowflake here').entry`?

That's true actually, I didn't consider that problem. I think a simple way to get around this is to just have a dedicated private method for getting something from the cache and always use that method, another problem which is relevant to this is that there are other parts of the code directly calling `cache.get` which doesn't update the `lastHit` property without additional code, so using a method here would also solve that problem too... Another aproach could be to maybe make a new `Cache` class which extends the `Collection` class but keeps track of the cache hits internally, exposing the same API to the outside as the `Collection` class but just with one extra method, `Cache#poll` which would clear the cache.

I'm not exactly sure which of the above solutions is better (the latter seems far more elegant though), and maybe there's a nice solution to it which I didn't think about at all yet...

VimHax:
Okay having thought more about it, I think the latter is definitely the better option here, there are many places where `cache.get` gets called directly and replacing all of those instances with one method is much uglier than just changing the underlying behavior of the Collection itself, plus it also allows re-usability.

VimHax:
I've also added a feature to the dedicated `Cache` class, max size, this can be used to further fine tune the caches (basically limiting their size if they get too big)... Currently that feature hasn't been integrated (cannot be changed by the end user) and I will integrate that when everything else has been ironed out. The new `Cache` class is meant to be very general purpose so I have added wrappers for some of the methods of the `Collection` class where the `Cache` needs to do some stuff, like update the cache hit collection, in order to satisfy the invariants.

ckohen:
I really appreciate what you are trying to do here, however, there are certain things that djs just always expects to be cached right now and break core functionality without that data being cached.
For example, you will never get a `channelUpdate` event if that channel is not in the cache
Most (if not all) of the guild based events will not emit if the guild is not cached.

Some consideration for this needs to be taken. Currently certain events bypass these checks by using "partials". While I don't love this method of handling uncached data, it is a decent stop-gap solution IMO

VimHax:
> I really appreciate what you are trying to do here, however, there are certain things that djs just always expects to be cached right now and break core functionality without that data being cached.
> For example, you will never get a `channelUpdate` event if that channel is not in the cache
> Most (if not all) of the guild based events will not emit if the guild is not cached.
> 
> Some consideration for this needs to be taken. Currently certain events bypass these checks by using "partials". While I don't love this method of handling uncached data, it is a decent stop-gap solution IMO

Hm, what if there was like a feature to pin data to the cache, as in, when pinned the data won't be removed from the cache during a poll unless unpinned. This way you can only keep the exact guilds/channels etc on the cache as you wish.

monbrey:
The point being made is that discord.js is designed, at its core, to rely on the populated caches received from the identify payload.

The changes you're proposing are a good design, but I think they belong instead as enhancements on something like the `LimitedCollection` which is currently used by the MessageManager. There are a lot of areas that the usage of the LimitedCollection could be expanded such as Users, GuildMembers, Presences etc.

Doing anything that impacts the caching of Guilds, Channels, Roles etc will pretty much break the way the majority of users expect the library to work.

kyranet:
I'd also like to point out that while this PR is supposed to decrease memory usage, it'll ironically increase for many other things - a timestamp is an integer longer than 32 bits (otherwise it'd have a precision of just 25 days), at this point, the JS engine makes them a double, which is 64 bits. There are also more problems:

## Memory Usage

When dealing with small sets of data, this wouldn't matter much (100k entries = 0.76MB overhead), but it becomes considerable when there are very small structures (`GuildEmoji`, `GuildMember`...) which are rather very numerous, and the change would cause them to substantially increase their size.

However, since you have moved the last hit cache to its own collection, that overhead becomes at least twice as large: now it's not just storing a value, but it's also storing a key and a value, and perhaps a hash, for each entry.

### Memory Duplication

There are many users who sweep the cache, discriminately (e.g. for `User`, checking `lastMessageID`) or indiscriminately (regardless of their properties or last use), however, if you delete an entry in, say the `client.users.cache` collection, you're not deleting it everywhere else, there are many places which can and will hold references to the user you just deleted, for example, `ReactionUserManager` (`message.reactions.cache.get(...).users`).

When an entry is deleted from the main collection, discord.js loses data (and therefore will not emit their respective `*Update` events) and will create new instances of it, so while you may delete 1M users in one go with this sweeper, you'd still have 1M users in memory, indefinitely, why? Because `GuildMember#user` is a property, not a getter!

https://github.com/discordjs/discord.js/blob/51551f544b80d7d27ab0b315da01dfc560b2c115/src/structures/GuildMember.js#L71-L78

Of course, there are other places where discord.js caches `User` instances outside the `UserManager`, and in the example mentioned before, considering you sweep members, you could easily be over 50k users which have reacted to a message, and those would take your memory space until they're completely removed, as well as any other reference a user may hold.

Those externally referenced objects would cause a very large memory growth, as you duplicate objects non-stop and make them stale quickly.

## CPU Usage

This overhead also comes with the problem it'd call `Date.now()` a lot of times at start-up, affecting the performance which is already bad enough for internally sharded bots even for new high-end server CPUs.

And not only that, every operation we do in those caches is doubled.

Finally, `Client#sweepCaches` is potentially able to freeze the bot's main thread for quite some time, depending on how many objects there are, and specially for very large bots. Those event loop stalls cause several issues in the bot and discord.js, from the WebSocket not being able to send a heartbeat or receive an ack, to the bot itself not replying to messages for a period of time. This has always been the issue with cache sweepers.

## Stale `lastHit`

`lastHit` is also flawed, how would you know the last time I have read the data? I could hold a User instance in a constant that loops for a long time and still read its properties (and this wouldn't update the property in the manager), or saved temporarily inside a custom structure, this makes `lastHit` itself stale, and relying on a property that can go stale, to determine whether something is stale, is a no-go.

Proxies are the only way to solve this, but they make property access over 100 times slower (640M ops/s vs 4M ops/s in Node.js v12 on a FX8320 @ 3.80GHz), so it's out of the question as that change would 100% CPU bots with at least 3k guilds.

## Conclusions

In a large bot, you're imposing this overhead in many places, including `Message#reactions` (which is a `ReactionManager`), moving this to the `LimitedCollection` Monbrey mentioned would be more adequate.

Now... if you ask my opinion, this should be an user-side enhancement and not something included in the core library. If something should be done in core, that should be using a strategy pattern, instead of something that causes a very large impact for everyone regardless of whether or not they want to automatically remove on lifetimes.

VimHax:
When I made the PR I never expected DJS to depend on cache so heavily, I use a similar implementation of caches for a bot and for that I used the cache in a controlled manner (I like treat the cache as if it's from Redis, so I don't keep the data around for long), but now I can very much see why that won't work here for DJS. So I think then it's then very fair to actually reduce the scope of this just to `LimitedCollection` like you and Monbrey said. I think I can completely get rid the overhead for users who don't care by just opting out of using the `Cache` class controlled by an option in `ClientOptions`.

VimHax:
> The point being made is that discord.js is designed, at its core, to rely on the populated caches received from the identify payload.
> 
> The changes you're proposing are a good design, but I think they belong instead as enhancements on something like the `LimitedCollection` which is currently used by the MessageManager. There are a lot of areas that the usage of the LimitedCollection could be expanded such as Users, GuildMembers, Presences etc.
> 
> Doing anything that impacts the caching of Guilds, Channels, Roles etc will pretty much break the way the majority of users expect the library to work.

What are all the managers that can make use of `LimitedCollection` without breaking the other parts of the library? I'm very new to the codebase so I'm not entirely sure where doing this lifetime stuff can work and cannot work. (Sorry if I'm being annoying) Also I'm going to close this PR and open another one where I will continue by implementing this lifetime behavior to all the managers which can have it. I basically need to revert all the changes I made to this PR so it's kind of pointless continuing this PR.

PR #5525, opened 2021-04-13
feat(Guild): nsfw guilds

almostSouji:

As per https://github.com/discord/discord-api-docs/pull/2805 guilds can now be designated as not safe for work. <https://support.discord.com/hc/en-us/articles/1500005389362-NSFW-Server-Designation>

- At this time PATCH and POST endpoints seem to not support a `nsfw` property.
- The heavily guarded assignment is used to avoid partials guilds messing up the property, as https://github.com/discordjs/discord.js/pull/4990 pointed out for users


kyranet (approved):

src/structures/Guild.js
```diff
@@ -101,6 +101,14 @@ class Guild extends Base {
...
      */
     this.shardID = data.shardID;
+
+    if ('nsfw' in data || typeof this.nsfw !== 'boolean') {
+      /**
+       * Whether the guild is designated as not safe for work
+       * @type {boolean}
+       */
+      this.nsfw = Boolean(data.nsfw);
+    }
```
advaith1, line 111: Seems like this logic is from User#bot where Discord will only send the property when it's `true`? Discord always sends `nsfw` for full guild objects.

```suggestion
    if ('nsfw' in data) {
      /**
       * Whether the guild is designated as not safe for work
       * @type {boolean}
       */
      this.nsfw = data.nsfw;
    }
```

almostSouji: do they? we've had that conversation before and essentially all booleans can act like that

advaith1: well it's documented as required and it's present when it's false; afaik this handling is only needed when it's documented as optional

SpaceEEC (approved):

iCrawl (approved):

PR #5530, opened 2021-04-14
feat(Webhook): add 'fetchMessage' method

izexi:
Ref: https://github.com/discord/discord-api-docs/pull/2812
~~This PR will be marked as draft until the referenced PR above has been merged.~~

### DIFF
--- a/src/structures/Webhook.js
+++ b/src/structures/Webhook.js
@@ -227,6 +227,18 @@ class Webhook {
   }
 
   /**
+   * Gets a message that was sent by this webhook.
+   * @param {Snowflake} message The ID of the message to fetch
+   * @param {boolean} [cache=true] Whether to cache the message
+   * @returns {Promise<Message|Object>} Returns the raw message data if the webhook was instantiated as a
+   * {@link WebhookClient} or if the channel is uncached, otherwise a {@link Message} will be returned
+   */
+  async fetchMessage(message, cache = true) {
+    const data = await this.client.api.webhooks(this.id, this.token).messages(message).get();
+    return this.client.channels?.cache.get(data.channel_id)?.messages.add(data, cache) ?? data;
+  }
+
+  /**
    * Edits a message that was sent by this webhook.
    * @param {MessageResolvable} message The message to edit
    * @param {StringResolvable|APIMessage} [content] The new content for the message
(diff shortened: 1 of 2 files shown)


ThaTiemsz:
has been merged now in https://github.com/discord/discord-api-docs/commit/94d57ea3ed3b2c42cbbb488c9f93d2bc81706cff. On a side note, shouldn't the jsdocs/typings for the method also accept a string since `@original` is also accepted as parameter?

src/structures/Webhook.js
```diff
@@ -226,6 +226,18 @@ class Webhook {
     return this;
   }
 
+  /**
+   * Gets a message that was sent by this webhook.
+   * @param {Snowflake} message The ID of the message to fetch
```
ckohen, line 231:
```suggestion
   * @param {MessageResolvable} message The ID of the message to fetch
```

Mirror this from the similar methods (and possibly add `@original` to the typedef for MessageResolvable, that I'm not sure on though).

izexi:
> has been merged now in [discord/discord-api-docs@94d57ea](https://github.com/discord/discord-api-docs/commit/94d57ea3ed3b2c42cbbb488c9f93d2bc81706cff). On a side note, shouldn't the jsdocs/typings for the method also accept a string since `@original` is also accepted as parameter?

Although `Snowflake` is currently typed as a `string` and passing `@original` would work, I do agree that it would be nicer to have a union type that also accepts `@original` as a string literal.

I was going to comment on this to be added in #5448 after this PR has been merged, since I think its more within the scope of that PR (and the same will need to be done on `Webhook#editMessage/deleteMessage()`).

src/structures/Webhook.js
izexi: I did consider using a `MessageResolvable` type at first but [`MessageManager#fetch()`](https://github.com/izexi/discord.js/blob/feat/Webhook%23fetchMessage()/typings/index.d.ts#L2043) accepts a `Snowflake` as the parameter so I went ahead and used that here as well to be consistent with that. Which similar methods are you referring to here?

ckohen: I misread actually, but I was referencing editMessage and deleteMessage. You are directly passing the message token here so MessageResolvable is actually incorrect 🤦

catcosmicice (approved):

kyranet (approved):

iCrawl (approved):

SpaceEEC (approved):

PR #5538, opened 2021-04-17
fix

Levoamphetamine:

### DIFF
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -36,6 +36,35 @@ declare module 'discord.js' {
 
   export const version: string;
 
+  enum languageOptions {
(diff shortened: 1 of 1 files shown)


almostSouji:
Where does this data supposedly come from? I can't find it on any activity. It's also quite doubtful we'd ever get a timer from the API (and there is no code that'd actually cause one)

Levoamphetamine:
> Where does this data supposedly come from? I can't find it on any activity. It's also quite doubtful we'd ever get a timer from the API (and there is no code that'd actually cause one)

Accidentally made a pull request, its just types for my own fork so typescript doesnt get pissed

PR #5544, opened 2021-04-19
Update WebSocketShard.js

meister03:
Causing a Error and a bot crash, because this.connection is null, which causes a Type Error on accessing this.connection.url
This should be a small fix, moreover, we have to find , why this.connection is null

Its just a small typerror, when this.connection is null. Its a mid term solution, so bots does not crashes.

replaced:
```diff
 - this.connection.url
 +  this.connection ? this.connection.url : "NO CONNECTION URL DEFINED";
```

### DIFF
--- a/src/client/websocket/WebSocketShard.js
+++ b/src/client/websocket/WebSocketShard.js
@@ -265,7 +265,7 @@ class WebSocketShard extends EventEmitter {
    * @private
    */
   onOpen() {
-    this.debug(`[CONNECTED] ${(this.connection? this.connection.url : "CONNECTION URL NOT FOUND")} in ${Date.now() - this.connectedAt}ms`);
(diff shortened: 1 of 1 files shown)


ChiefChippy2:
and how do you expect the bot to function without this anyways

VoltrexMaster:
The client won't work without the connection URL so its better to let it crash instead of mindlessly handling it.

meister03:
> The client won't work without the connection URL so its better to let it crash instead of mindlessly handling it.
Then I could add a throw new Error(), than rather people let wondering, why it crashed.....

vBread:
That won't make a difference, the error will still throw in the same place.

meister03:
> That won't make a difference, the error will still throw in the same place.

true, we have to find the issue, where it causes that this connection is null

PR #5552, opened 2021-04-22
Changed homepage url

LightnerDev:
i changed the homepage url in package.json
I think https://discord.js.org is better than the repo as the homepage

what do i write here?


iCrawl:
I disagree with the change. This is mainly displayed on npmjs anyway and directly going to the source seems more sensible to me than first visiting the website.

PR #5559, opened 2021-04-23
fix(ShardingManager): client error event cannot be emitted

meister03:
When a Error happens during sending, it can not overwrite err.message. So that a error happens: "TypeError: Cannot set property message of  which has only a getter"

The Error can be reproduced with having the Sharding Manager on Worker Mode and than broadcastEvaling `this.guilds.cache`,
the upper scenario works fine for child processes, bc they support many types, but worker threads does not support functions and some other types. So a error comes:  `DOMException [DataCloneError]: #<Object> could not be cloned.`
This Error can just be logged out when you merge the commit, bc there is a error itself on the _respond function, which should be fixed with the commit. I checked it, multiply times...

### DIFF
--- a/src/sharding/ShardClientUtil.js
+++ b/src/sharding/ShardClientUtil.js
@@ -199,8 +199,7 @@ class ShardClientUtil {
    */
   _respond(type, message) {
     this.send(message).catch(err => {
-      let error = {err}
-      error.message = `Error when sending ${type} response to master process: ${err.message}`;
+      let error = new Error(`Error when sending ${type} response to master process: ${err.message}`);
       /**
        * Emitted when the client encounters an error.
        * @event Client#error
--- a/src/sharding/ShardClientUtil.js
+++ b/src/sharding/ShardClientUtil.js
@@ -199,7 +199,7 @@ class ShardClientUtil {
    */
   _respond(type, message) {
     this.send(message).catch(err => {
-      let error = new Error(`Error when sending ${type} response to master process: ${err.message}`);
+      const error = new Error(`Error when sending ${type} response to master process: ${err.message}`);
       /**
        * Emitted when the client encounters an error.
        * @event Client#error
--- a/src/sharding/ShardClientUtil.js
+++ b/src/sharding/ShardClientUtil.js
@@ -200,6 +200,7 @@ class ShardClientUtil {
   _respond(type, message) {
     this.send(message).catch(err => {
       const error = new Error(`Error when sending ${type} response to master process: ${err.message}`);
+      error.stack = err.stack;
       /**
        * Emitted when the client encounters an error.
        * @event Client#error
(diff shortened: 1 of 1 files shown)


BannerBomb:
I honestly don't agree with this fix. Because you are just creating an object with the key "err" with the value of the error class then adding a "message" key to it. Instead of emitting an error class to the error event (which is what it's supposed to receive), you are now emitting an object. The TypeError happens because the DOMException class has a read-only message property which might need some handling for that. With the #<Object> could not be cloned error, I don't think you can send collections/maps to workers. You would need to serialize it to a transferable data structure like an object.

meister03:
Should I do Object assign?
and copy the error on a new object?
because the `err.message` is not writeable

src/sharding/ShardClientUtil.js
```diff
@@ -199,13 +199,14 @@ class ShardClientUtil {
    */
   _respond(type, message) {
     this.send(message).catch(err => {
-      err.message = `Error when sending ${type} response to master process: ${err.message}`;
+      let error = {err}
+      error.message = `Error when sending ${type} response to master process: ${err.message}`;
```
monbrey, line 203:
```suggestion
      let error = new Error(`Error when sending ${type} response to master process: ${err.message}`);
```
Create a new instance of Error with the new message.

meister03: 👍🏻 good, now people will know the error, when some happens

src/sharding/ShardClientUtil.js
```diff
@@ -199,13 +199,13 @@ class ShardClientUtil {
    */
   _respond(type, message) {
     this.send(message).catch(err => {
-      err.message = `Error when sending ${type} response to master process: ${err.message}`;
+      let error = new Error(`Error when sending ${type} response to master process: ${err.message}`);
```
vaporox, line 202:
```suggestion
      const error = new Error(`Error when sending ${type} response to master process: ${err.message}`);
```

src/sharding/ShardClientUtil.js
```diff
@@ -199,13 +199,13 @@ class ShardClientUtil {
    */
   _respond(type, message) {
     this.send(message).catch(err => {
-      err.message = `Error when sending ${type} response to master process: ${err.message}`;
+      const error = new Error(`Error when sending ${type} response to master process: ${err.message}`);
```
vladfrangu, line 202:
```suggestion
      const error = new Error(`Error when sending ${type} response to master process: ${err.message}`);
      error.stack = err.stack;
```

vladfrangu (changes requested):

kyranet:
Any update on this @meister03? Vlad's request is still pending.

meister03:
> Any update on this @meister03? Vlad's request is still pending.
Which update? If the code works?

kyranet:
Vlad left a requested change above, which is still left without a reply.

meister03:
> Vlad left a requested change above, which is still left without a reply.

 Forgot to test it, and it works

kyranet (approved):

iCrawl (approved):

vladfrangu (approved):

SpaceEEC (approved):

PR #5563, opened 2021-04-27
fix(GuildMemberRoleManager): fix remove method when passing collection/array

GamingGeek:
When passing a collection of roles or an array of RoleResolvable, it gets mapped into an array of ids and then filtered but checking if the array contains the role and not the id.

This PR checks if the array contains the role id rather than role instance, allowing GuildMemberRoleManager#remove to work with a collection or array like it should.


GamingGeek:
Nevermind, I just went back to PRs and saw I missed a PR that fixes the same issue.

PR #5565, opened 2021-04-28
fix(BitField): throw an error if bit to resolve is undefined

Keke27210:

Currently, `BitField.has(undefined)` or `BitField.has(<BitField>.FLAGS.ANY_NON_SPECIFIED_FLAG)` returns true.

### DIFF
--- a/src/util/BitField.js
+++ b/src/util/BitField.js
@@ -41,9 +41,9 @@ class BitField {
    * @returns {boolean}
    */
   has(bit) {
-    if (typeof bit === 'undefined') return false;
     if (Array.isArray(bit)) return bit.every(p => this.has(p));
     bit = this.constructor.resolve(bit);
+    if (bit === this.constructor.defaultBit) return false;
     return (this.bitfield & bit) === bit;
   }
 
--- a/src/util/BitField.js
+++ b/src/util/BitField.js
@@ -38,11 +38,9 @@ class BitField {
   /**
    * Checks whether the bitfield has a bit, or multiple bits.
    * @param {BitFieldResolvable} bit Bit(s) to check for
-   * @param {...*} hasParams Additional parameters for the has method, if any
    * @returns {boolean}
    */
-  has(bit, ...hasParams) {
-    if (Array.isArray(bit)) return bit.every(p => this.has(p, ...hasParams));
+  has(bit) {
     bit = this.constructor.resolve(bit);
     return (this.bitfield & bit) === bit;
   }
--- a/src/util/BitField.js
+++ b/src/util/BitField.js
@@ -43,7 +43,6 @@ class BitField {
   has(bit) {
     if (Array.isArray(bit)) return bit.every(p => this.has(p));
     bit = this.constructor.resolve(bit);
-    if (bit === this.constructor.defaultBit) return false;
     return (this.bitfield & bit) === bit;
   }
 
@@ -144,7 +143,6 @@ class BitField {
    */
   static resolve(bit) {
     const { defaultBit } = this;
-    if (typeof bit === 'undefined') return defaultBit;
     if (typeof defaultBit === typeof bit && bit >= defaultBit) return bit;
     if (bit instanceof BitField) return bit.bitfield;
     if (Array.isArray(bit)) return bit.map(p => this.resolve(p)).reduce((prev, p) => prev | p, defaultBit);
--- a/src/util/BitField.js
+++ b/src/util/BitField.js
@@ -41,6 +41,7 @@ class BitField {
    * @returns {boolean}
    */
   has(bit) {
+    if (typeof bit === 'undefined') return false;
     if (Array.isArray(bit)) return bit.every(p => this.has(p));
     bit = this.constructor.resolve(bit);
     return (this.bitfield & bit) === bit;
--- a/src/util/BitField.js
+++ b/src/util/BitField.js
@@ -38,10 +38,11 @@ class BitField {
   /**
    * Checks whether the bitfield has a bit, or multiple bits.
    * @param {BitFieldResolvable} bit Bit(s) to check for
+   * @param {...*} hasParams Additional parameters for the has method, if any
    * @returns {boolean}
    */
-  has(bit) {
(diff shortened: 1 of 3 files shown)


1chiSensei (approved):
lgtm

src/util/BitField.js
```diff
@@ -41,6 +41,7 @@ class BitField {
    * @returns {boolean}
    */
   has(bit) {
+    if (typeof bit === 'undefined') return false;
     if (Array.isArray(bit)) return bit.every(p => this.has(p));
     bit = this.constructor.resolve(bit);
```
vladfrangu, line 46:
```suggestion
    if (Array.isArray(bit)) return bit.every(p => this.has(p));
    bit = this.constructor.resolve(bit);
    if (bit === this.constructor.defaultBit) return false;
```

vladfrangu (changes requested):

src/util/BitField.js
Keke27210: I hadn't thought of it that way, but it seems to be a more generic fix 👍

vladfrangu (approved):

kyranet (approved):

SpaceEEC:
I don't really think this is what we want.

Taking an example where this would yield an unexpected result: A command handler.

In our example no permissions would be required, maybe it's a ping command.
The handler wants to catch invalid permissions early, so it's resolving the required permissions (defaulting to an empty array) to `0n`.
Because it's clever it's saving the number instead of the array. (Saving a few cpu cycles, performance and stuff)
Then when a member wants to run said command the handler checks for them using effectively `message.member.permissions.has(0n)` and gets `false`, thus denying the user to execute said command.
(Except when they are admin, but that's unrelated to this issue)

As result you'd now need to change your command handler to use something like `requiredPermissions === Permissions.defaultBit || message.member.permissions.has(requiredPermissions)`.

`bitField.has(0)` should return `true`, since there are no bits not set in the bit field.

---

Another inconsistency here is that apparently identical values would yield different results
Here: `bitField.has(0)` with `false` and `bitField.has([])` with `true` (Both would return `0` if passed to `BitField.resolve`)

i.e. `bitField.has(BitField.resolve(n)) === bitField.has(n)` should always return true (or both throw an error)

---

Maybe this issue here be approached by editing `BitField.resolve`, although I am pretty sure we internally rely on `undefined` resolving to `BitField.defaultBit` everywhere.

Keke27210:
> I don't really think this is what we want.
> 
> Taking an example where this would yield an unexpected result: A command handler.
> 
> In our example no permissions would be required, maybe it's a ping command.
> The handler wants to catch invalid permissions early, so it's resolving the required permissions (defaulting to an empty array) to `0n`.
> Because it's clever it's saving the number instead of the array. (Saving a few cpu cycles, performance and stuff)
> Then when a member wants to run said command the handler checks for them using effectively `message.member.permissions.has(0n)` and gets `false`, thus denying the user to execute said command.
> (Except when they are admin, but that's unrelated to this issue)
> 
> As result you'd now need to change your command handler to use something like `requiredPermissions === Permissions.defaultBit || message.member.permissions.has(requiredPermissions)`.
> 
> `bitField.has(0)` should return `true`, since there are no bits not set in the bit field.
> 
> Another inconsistency here is that apparently identical values would yield different results
> Here: `bitField.has(0)` with `false` and `bitField.has([])` with `true` (Both would return `0` if passed to `BitField.resolve`)
> 
> i.e. `bitField.has(BitField.resolve(n)) === bitField.has(n)` should always return true (or both throw an error)
> 
> Maybe this issue here be approached by editing `BitField.resolve`, although I am pretty sure we internally rely on `undefined` resolving to `BitField.defaultBit` everywhere.

I'm not sure what to think anymore... Is this a "non-issue"? Or should we go back to the basic change where we check if bit is undefined (and incidentally if the array is empty, if one is provided).

kyranet (dismissed):
Please address the issue explained by @SpaceEEC, as this current code contains a critical bug that will make it behave in unexpected ways.

Keke27210:
> Please address the issue explained by @SpaceEEC, as this current code contains a critical bug that will make it behave in unexpected ways.

My main proposal was to return `false` whenever an unknown/undefined bit was given to the method (even tho I wasn't checking for possible undefined bits in an array). Vlad then requested to check, instead, if bit was equal to defaultBit after calling `resolve` since when undefined, the defaultBit is returned, leading to the noticed issue.
Now I see the problem in that case, but that does not tell me what would an acceptable solution be.

- Should `<BitField>.constructor.resolve` throw an error when `typeof bit === undefined` ?
- Should we silently ignore undefined bit(s) and do not count them ? (making `<BitField>.has(undefined) === false`)
- Should we keep the current behavior and therefore, close this PR.

Keke27210:
The first solution is probably the best one, I mean we should not give any answer if we are not able to give one.
- Reverted changes to `BitField#has`
- Updated static `resolve` method so if `typeof bit === undefined` it passes through all if and throws an error

Keke27210:
Need advice for the three last commits.
Needed so `<Permissions>.has([..., undefined, ...])` throws an error.

src/util/BitField.js
```diff
@@ -38,10 +38,11 @@ class BitField {
   /**
    * Checks whether the bitfield has a bit, or multiple bits.
    * @param {BitFieldResolvable} bit Bit(s) to check for
+   * @param {...*} hasParams Additional parameters for the has method, if any
```
SpaceEEC, line 41: I'm not sure what exactly this solves, but I'd rather drop it and the recursion that requires it.
`BitField.resolve` handles arrays just fine.

SpaceEEC (changes requested):
Should be good to go otherwise.

src/util/BitField.js
Keke27210: Without this, it used to return true whenever they had admin no matter if you actually passed true to checkAdmin.
I still do not really understand the issue here, anyway as you said `<BitField>.constructor.resolve` handles arrays well already.

SpaceEEC (approved):

iCrawl (approved):

kyranet (approved):

Issue #2721, opened 2018-08-09
Let SplitOptions#char accept an array


SimplyLinn:
**Is your feature request related to a problem? Please describe.**
If you want to split a message, you have to pick one specific character to split on. I suggest enhancing the split function to allow you to specify more than one character and split based on all of them in descending order of priority

**Describe the solution you'd like**
Letting `SplitOptions#char` be an array as well as a string. If it's an array, we first try to split the message using the first element in that array, if we fail to construct a series of messages with only that split, we retry splitting the messages that are too long with the second element, etc. If we fail to split the message at all, we throw an error just as we do now.

**Describe alternatives you've considered**
Uhm, the alternative would be to not implement this? It's just a suggestion, and I'm mainly posting here to gather responses, if it's something we want I'll implement it myself and make a PR

**Additional context**
imagine this array being put in:
`['\n', ' ', '']`

That would result in the message being split on first new lines, and if we fail to split the message using only new lines, we try to split using spaces, and if even that fails, we split EVERYTHING, and forcibly split the message wherever is needed. This array would make it so the split never fails. That's useful if, say, we want to make sure that this message being sent no matter what.

I feel this would also make a good default value, but that's probably more debatable than the feature as a whole

1Computer1:
Why not a regex like String#split?

NbOpposite:
Ah, yes, that wouldake even more sense!

NbOpposite:
Actually, hmm... I'm unsure if you can get the prioritizing of splitting I was talking about without a very complex regex, which would make the point of having an easy to use split function moot to begin with. I agree we should accept regex-splitting as well, but not to replace the priority array.

Simply specifying `/\n| /g` as a regex would not prioritize splitting on newlines, it would split on both newlines and spaces at once and remove this whole "use next index as fallback" philosophy behind the array

1Computer1:
Why not both? Perhaps allow it to take `string | regexp | Array<string | regexp>`.

NbOpposite:
Yeah, I might have been unclear but that was my intention from the beginning when I said I agreed we should be supporting regex in combination with the priority array.

cfanoulis:
Uhm, doesn't `String.split();` work with multiple arguments? Isn't that the behaviour you are looking for?

PR #5566, opened 2021-04-28
feat(Util): allow array for StringOptions' char

u9g:

Allow StringOptions to accept a string array for char

- https://github.com/discordjs/discord.js/issues/2721


src/util/Util.js
```diff
@@ -64,6 +64,7 @@ class Util {
   static splitMessage(text, { maxLength = 2000, char = '\n', prepend = '', append = '' } = {}) {
     text = Util.resolveString(text);
     if (text.length <= maxLength) return [text];
+    if (Array.isArray(char)) char = char.filter(splitChar => text.includes(splitChar))[0];
```
VoltrexMaster, line 67:
```suggestion
    if (Array.isArray(char)) char = char.find(splitChar => text.includes(splitChar));
```
No need to use `Array.prototype.filter()` just to get the first result.

u9g: Fixed, thanks

vladfrangu (approved):

kyranet (approved):

SpaceEEC:
I'm not sure if this approach (only using the first provided char that is in the content of the message) is good, I'd prefer something described in the linked issue that actually utilizes (if necessary) all chars in the array.
This also needs to be documented.

kyranet:
@u9g any update on this? You have a pending request from Space.

u9g:
> I'm not sure if this approach (only using the first provided char that is in the content of the message) is good, I'd prefer something described in the linked issue that actually utilizes (if necessary) all chars in the array.
> This also needs to be documented.

Documented and added support for arrays and regex expressions


> @u9g any update on this? You have a pending request from Space.

Just committed :)

SpaceEEC (dismissed):

src/util/Util.js
```diff
@@ -63,9 +63,22 @@ class Util {
    */
   static splitMessage(text, { maxLength = 2000, char = '\n', prepend = '', append = '' } = {}) {
     text = Util.verifyString(text, RangeError, 'MESSAGE_CONTENT_TYPE', false);
-    if (text.length <= maxLength) return [text];
-    const splitText = text.split(char);
-    if (splitText.some(chunk => chunk.length > maxLength)) throw new RangeError('SPLIT_MAX_LEN');
+    const hasElemLongerThanMaxLength = (arr, len) => arr.some(elem => elem.length > len);
```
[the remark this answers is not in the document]
u9g: I mean array can be replaced with splitText too if I swap this line with the line above it.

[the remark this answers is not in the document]
SpaceEEC: Maybe just get rid of that function completely then?
- Check the length of `text` before wrapping it in an array. (lines 67-68)
- Inline the function in line 70 and 81.

src/structures/interfaces/TextBasedChannel.js
```diff
@@ -99,7 +99,8 @@ class TextBasedChannel {
    * Options for splitting a message.
    * @typedef {Object} SplitOptions
    * @property {number} [maxLength=2000] Maximum character length per message piece
-   * @property {string} [char='\n'] Character to split the message with
+   * @property {string | string[] | RegExp | RegExp[]} [char='\n'] Character(s) or Regex(s) to split the message with,
```
papaia, line 102:
```suggestion
   * @property {string|string[]|RegExp|RegExp[]} [char='\n'] Character(s) or Regex(s) to split the message with,
```

kyranet: Consistency with the rest of the library, mostly.

u9g: sure, just curious, added

src/util/Util.js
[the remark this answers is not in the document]
u9g: Done! added in cd094a5

src/util/Util.js
```diff
@@ -63,9 +63,20 @@ class Util {
    */
   static splitMessage(text, { maxLength = 2000, char = '\n', prepend = '', append = '' } = {}) {
     text = Util.verifyString(text, RangeError, 'MESSAGE_CONTENT_TYPE', false);
-    if (text.length <= maxLength) return [text];
-    const splitText = text.split(char);
-    if (splitText.some(chunk => chunk.length > maxLength)) throw new RangeError('SPLIT_MAX_LEN');
+    if (text.length > maxLength) return [text];
```
SpaceEEC, line 66:
```suggestion
    if (text.length > maxLength) return [text];
    let splitText = [text];
```

SpaceEEC (changes requested):

SpaceEEC (approved):

PR #5567, opened 2021-04-28
fix(Message): reply method; 

GoldenCredit101:
this PR will check if there is a replyTo option exist in reply method; if there is, it would reply to message according to replyTo property pass into the reply method by end user if not it would default to current message (this). 


monbrey:
I don't think this is the direction we want to take on this. It's been discussed internally by the maintainers, and I think the agreement was that `Message#reply` should not accept `replyTo` at all.

`TextChannel#send` is where the replyTo option can be used correctly.

PR #5570, opened 2021-04-30
feat: api v9 and threads

ckohen:
This is a very preliminary attempt at adding threads: https://github.com/discord/discord-api-docs/pull/2855

~~⚠️ Disclaimer: as the docs are only semi-finalized and still subject to change, this will remain a draft until they are finalized and threads are actually released and can be test. This is primarily to track progress on the implementation and to ensure all efforts to implement threads can be focused in one place.~~

Notes:
1. As of now, threads are included in the channel caches for the client and guild alongside in their respective text channels threads cache. This is because threads are channels at heart, and allows access to said thread channels as you would any normal channel.
2. Although threads are present in the channel cache, there is a separate event structure in the API, and such those events are mirrored as well.
3. A minor warning about thread channels in the channel cache: usually discord.js expects ALL channels to be cached. With threads, if the client does not have access to the parent channel, then it does not know the thread exists, so the channels can be uncached.

Feel free to make suggestions for major changes in this code, I am open to implementing this any different way.


iCrawl:
Can't wait to see this merged v15

ckohen:
Lol yeah, looking forward to that

src/client/actions/ChannelCreate.js
```diff
@@ -14,7 +14,7 @@ class ChannelCreateAction extends Action {
        * @event Client#channelCreate
        * @param {GuildChannel} channel The channel that was created
        */
-      client.emit(Events.CHANNEL_CREATE, channel);
+      client.emit(Events.THREAD_DELETE, channel);
```
Keke27210, line 17: I *think* there is a tiny problem here unless I missed something.

src/structures/ThreadChannel.js
```diff
@@ -0,0 +1,410 @@
...
+  }
+
+  /**
+   * Gets the overall set of permissions for a member or role in this threads' parent, taking into account overwrites.
+   * @param {GuildMemberResolvable|RoleResolvable} memberOrRole The member or role to obtain the overall permissions for
+   * @returns {?Readonly<Permissions>}
+   */
+  permissionsFor(memberOrRole) {
+    return this.parent?.permissionsFor(memberOrRole);
+  }
```
ImRodry, line 208: How are you taking overwrites into account if you're just checking for the parent permissions? Am I missing something?

src/structures/ThreadChannel.js
```diff
@@ -0,0 +1,410 @@
...
+   *   .catch(console.error);
+   */
+  setArchived(archived, reason) {
+    return this.edit({ archived }, reason);
+  }
+
+  /**
+   * Sets the duration before the channel is automatically archived.
+   * @param {number} autoArchiveDuration How long in minutes before the thread is automatically archived,
+   * one of `60`, `1440`, `4320`, or `10080`
```
ImRodry, line 260:
```suggestion
   * @param {number} autoArchiveDuration How long, in minutes, before the thread is automatically archived.
   * Either `60`, `1440` (1 day), `4320` (3 days), or `10080` (7 days)
```

src/structures/Message.js
```diff
@@ -210,7 +221,7 @@ class Message extends Base {
     this.flags = new MessageFlags(data.flags).freeze();
 
     /**
-     * Reference data sent in a crossposted message or inline reply.
+     * Reference data sent in a message that contains ids identifying a previous message
```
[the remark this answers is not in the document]
ckohen: IDs refers to guild, channel, message ID, not multiple message IDs

src/structures/ThreadChannel.js
ImRodry: How will you take that exception into account then?

ckohen: that's what the other permission bits are for, also, `ThreadChannel#joinable` does the checks needed for sending messages if you really wanted to use it

src/managers/ThreadManager.js
```diff
@@ -0,0 +1,218 @@
...
+
+  /**
+   * Creates a new thread in the channel.
+   * @param {string} name The name of the new Thread
+   * @param {number} autoArchiveDuration How long in minutes before the thread is automatically archived,
+   * one of `60`, `1440`, `4320`, or `10080`
+   * @param {Object} options Options
+   * @param {MessageResolvable} [options.startMessage] The message to start a public or news thread from,
+   * creates a private thread if not provided
+   * @param {string} [options.reason] Reason for creating the thread
```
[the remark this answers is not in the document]
ImRodry: Yeah I know but wouldn’t it be better to make it an actual typedef? Not sure since the two options are used somewhat inconsistently throughout the library

src/managers/ThreadMemberManager.js
```diff
@@ -0,0 +1,98 @@
...
+    if (existing && existing._patch && cache) existing._patch(data);
+    if (existing) return existing;
+
+    const member = new ThreadMember(this.thread, data);
+    if (cache) this.cache.set(data.user_id, member);
+    return member;
+  }
+
+  /**
+   * Data that resolves to give a ThreaddMember object. This can be:
```
GamingGeek, line 38:
```suggestion
   * Data that resolves to give a ThreadMember object. This can be:
```

src/structures/ThreadChannel.js
```diff
@@ -0,0 +1,407 @@
...
+   * Adds a member to this thread.
+   * @param {UserResolvable} member The member to add
+   * @param {string} [reason] The reason for adding this member
+   * @returns {Promise<Snowflake>}
+   */
+  addMember(member, reason) {
+    const id = this.client.users.resolveID(member);
+    if (!id) return Promise.reject(new TypeError('INVALID_TYPE', 'member', 'UserResolvable'));
+    return this.client.api
+      .channels(this.thread.id, 'thread-members', id)
```
GamingGeek, line 174:
```suggestion
      .channels(this.id, 'thread-members', id)
```

src/structures/ThreadChannel.js
```diff
@@ -0,0 +1,407 @@
...
+      .put({ reason })
```
GamingGeek, line 175:
```suggestion
      .post({ reason })
```

ckohen: By the time this PR is merged, POST will not work, so I'll leave it.

vladfrangu:
You wouldn't get this from any other guy, but this needs a rebase <3

src/structures/ThreadMember.js
```diff
@@ -0,0 +1,95 @@
...
+    /**
+     * The flags for this thread member
+     * @type {ThreadMemberFlags}
+     */
+    this.flags = new ThreadMemberFlags(data.flags);
+  }
+
+  /**
+   * The guild member that this thread member instance represents
+   * @type {GuildMember}
```
vladfrangu, line 51:
```suggestion
   * @type {?GuildMember}
```

src/structures/ThreadMember.js
```diff
@@ -0,0 +1,95 @@
...
+  }
+
+  _patch(data) {
+    this.joinedTimestamp = new Date(data.join_timestamp).getTime();
+
+    /**
+     * The flags for this thread member
+     * @type {ThreadMemberFlags}
+     */
+    this.flags = new ThreadMemberFlags(data.flags);
```
vladfrangu, line 46: Freeze these

src/structures/ThreadMember.js
```diff
@@ -0,0 +1,95 @@
...
+   * @type {?Date}
+   * @readonly
+   */
+  get joinedAt() {
+    return this.joinedTimestamp ? new Date(this.joinedTimestamp) : null;
+  }
+
+  /**
+   * The user that this thread member instance represents
+   * @type {User}
```
vladfrangu, line 69:
```suggestion
   * @type {?User}
```

vladfrangu (changes requested):

src/client/actions/ThreadDelete.js
```diff
@@ -0,0 +1,34 @@
+'use strict';
+
+const Action = require('./Action');
+const { Events } = require('../../util/Constants');
+
+class ThreadDeleteAction extends Action {
+  constructor(client) {
+    super(client);
+    this.deleted = new Map();
```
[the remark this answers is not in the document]
ckohen: Not really sure why this is here, copied it from ChannelDelete where it must have been left residually.

src/client/actions/ThreadListSync.js
```diff
@@ -0,0 +1,63 @@
...
+
+      Object.entries(data.members).forEach(([threadID, rawMember]) => {
+        const thread = client.channels.cache.get(threadID);
+        if (thread) {
+          thread.members.add(rawMember);
+        }
+      });
+
+      /**
+       * Emitted whenever the client user gains access to a text or news channel
```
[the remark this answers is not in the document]
ckohen: What isn't accurate about this?

src/client/actions/ThreadListSync.js
```diff
@@ -0,0 +1,63 @@
...
+      };
+    }
+
+    return {};
+  }
+
+  removeStale(channel) {
+    if (channel.threads) {
+      channel.threads.cache.forEach(thread => {
+        if (!thread.archived) {
```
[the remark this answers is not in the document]
ckohen: Good point to bring up, I'm unsure of how this should be handled. Yes it is probably a good idea to do that, but does that mean archived should be treated like deleted when getting a THREAD_UPDATE? Getting THREAD_LIST_SYNC doesn't happen all that often (and only when previously access was denied) so maybe only do it here?

src/managers/ThreadManager.js
```diff
@@ -0,0 +1,226 @@
...
+  /**
+   * The cache of this Manager
+   * @type {Collection<Snowflake, ThreadChannel>}
+   * @name ThreadManager#cache
+   */
+
+  add(thread) {
+    const existing = this.cache.get(thread.id);
+    if (existing) return existing;
+    this.cache.set(thread.id, thread);
```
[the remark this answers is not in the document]
ckohen: Yes, it gets called by ChannelManager once the channel is created, as with GuildChannelManager

src/client/actions/ThreadListSync.js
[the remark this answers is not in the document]
vladfrangu: they're not text or news channels, they're thread channels :P

src/structures/Message.js
```diff
@@ -101,6 +101,16 @@ class Message extends Base {
       this.pinned = null;
     }
 
+    if ('thread' in data) {
+      /**
+       * The thread started by this message
+       * @type {?ThreadChannel}
+       */
+      this.thread = this.client.channels.add(data.thread);
+    } else if (!(this.thread instanceof require('./ThreadChannel'))) {
```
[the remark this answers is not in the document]
ckohen: Not quite, but will make it the same as author `else if (!this.thread)`. This also made me realize I forgot to but this into the *other* patch method 🤦

src/client/actions/ThreadListSync.js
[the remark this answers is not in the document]
ckohen: But the event is emitted when you gain access to the parent channel, being a text or news channel, even if there are no threads.

src/structures/ThreadChannel.js
```diff
@@ -0,0 +1,400 @@
...
+  /**
+   * Adds a member to this thread.
+   * @param {UserResolvable} member The member to add
+   * @param {string} [reason] The reason for adding this member
+   * @returns {Promise<Snowflake>}
+   */
+  addMember(member, reason) {
+    const id = this.client.users.resolveID(member);
+    if (!id) return Promise.reject(new TypeError('INVALID_TYPE', 'member', 'UserResolvable'));
+    return this.client.api
```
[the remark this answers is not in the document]
ckohen: I took inspiration for this from Guild#addMember where we can't do members#add because that is an existing method on managers. I think the decision there was that members#addMember felt weird?

src/structures/ThreadChannel.js
```diff
@@ -0,0 +1,400 @@
...
+     */
+    this.archiveTimestamp = data.thread_metadata.archive_timestamp
+      ? new Date(data.thread_metadata.archive_timestamp).getTime()
+      : null;
+
+    /**
+     * The approximate count of messages in this thread
+     * <info> This value will not count above 50 even when there are more than 50 messages
+     * If you need an approximate value higher than this, use ThreadChannel#messages.cache.size</info>
+     * @type {?number}
```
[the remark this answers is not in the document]
ckohen: Not sure, I haven't seen a null come through, but I have no idea what gets sent after the threshold is reached, it could just stay as 50 or be null.

src/client/actions/ThreadListSync.js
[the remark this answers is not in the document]
ckohen: Oh, and yes, this removes active threads because all active threads are being synced.

src/structures/ThreadChannel.js
```diff
@@ -0,0 +1,400 @@
...
+   * Whether the thread is unarchivable by the client user
+   * @type {boolean}
+   * @readonly
+   */
+  get unarchivable() {
+    return (
+      this.archived &&
+      (this.locked
+        ? this.permissionsFor(this.client.user)?.has(Permissions.FLAGS.MANAGE_THREADS, false)
+        : this.permissionsFor(this.client.user)?.has(Permissions.FLAGS.SEND_MESSAGES, false) &&
```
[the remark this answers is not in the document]
ckohen: Again, not sure with this, similar checks do this across the library, I'll remove them  in the next commit but leave this unresolved for now.

src/managers/ThreadManager.js
```diff
@@ -0,0 +1,226 @@
...
+   *     reason: 'Needed a separate thread for food',
+   *   })
+   *   .then(console.log)
+   *   .catch(console.error);
+   * // Create a new private thread
+   * channel.threads.create('mod-talk', { autoArchiveDuration: 60, reason: 'Needed a separate thread for moderation' })
+   *   .then(console.log)
+   *   .catch(console.error);
+   */
+  async create(name, autoArchiveDuration, { startMessage, reason } = {}) {
```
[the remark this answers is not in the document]
iCrawl: Yes, as it currently stands the example and the actual params are wrong anyway as I see this.

src/client/actions/ThreadListSync.js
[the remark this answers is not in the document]
vladfrangu: > But the event is emitted when you gain access to the parent channel, being a text or news channel, even if there are no threads.

The event is WHAT?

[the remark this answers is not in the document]
ajpalkovic: > even if there are no threads.

rly? I think every place that sends it is checking if there are threads, how'd you reproduce it!

src/managers/ChannelManager.js
```diff
@@ -24,6 +24,9 @@ class ChannelManager extends BaseManager {
     if (existing) {
       if (existing._patch && cache) existing._patch(data);
       if (guild) guild.channels.add(existing);
+      if (ThreadChannelTypes.includes(existing.type) && typeof existing.parent?.threads !== 'undefined') {
+        existing.parent.threads.add(existing);
```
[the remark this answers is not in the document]
kyranet: Just one small doubt about this: can we create 50 thread channels alongside 250 (maximum) regular channels?

If so, I think this change would affect bots that check for whether the channel size is below the maximum before creating a new one.

src/client/actions/ThreadListSync.js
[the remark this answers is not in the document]
ckohen: Hmm, cannot reproduce with no threads anymore. Although, it does seem weird to not sync threads when there were previously threads. E.g. if you have active threads, lose access to a channel, thread is marked archived, gain access to the channel. Since there is no `THREAD_DELETE` fired here and we obviously don't get `THREAD_UPDATE` while not having access to the channel, the cache is now out of date and doesn't get updated (in this case, removing the now archived threads

src/managers/ChannelManager.js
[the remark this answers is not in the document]
ckohen: The line above adds it to the guild channel cache,. Itts accessible via the client channel cache, the guild channel cache and the parent channels threads cache.

Thread channels are independently counted yes, I think that's why this is considered semver major currently. I could add a getter to GuildChannelManager like GuildChannelManager#hasMaxChannels maybe?

src/structures/ThreadChannel.js
```diff
@@ -0,0 +1,394 @@
...
+
+  /**
+   * Whether the thread is joinable by the client user
+   * @type {boolean}
+   * @readonly
+   */
+  get joinable() {
+    return (
+      !this.archived &&
+      this.permissionsFor(this.client.user)?.any([
```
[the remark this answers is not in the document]
ckohen: Ah...hadn't thought about joining a read only thread, if you have manage_threads but not send_messages you can be readonly in a private thread right?

I was using this as a check for sending as well, but that will change with this.

src/structures/ThreadChannel.js
```diff
@@ -0,0 +1,394 @@
...
+   * @type {boolean}
+   * @readonly
+   */
+  get unarchivable() {
+    return (
+      this.archived &&
+      (this.locked
+        ? this.permissionsFor(this.client.user)?.has(Permissions.FLAGS.MANAGE_THREADS)
+        : this.permissionsFor(this.client.user)?.has(Permissions.FLAGS.SEND_MESSAGES) &&
+          this.members.has(this.client.user.id))
```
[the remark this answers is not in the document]
ckohen: What part of the requirement?

src/structures/ThreadChannel.js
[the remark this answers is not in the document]
ajpalkovic: Yeah, that's correct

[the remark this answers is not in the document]
ajpalkovic: https://www.notion.so/Thread-Permissions-add684b087ea437890b7676a66ccf3da

src/structures/ThreadChannel.js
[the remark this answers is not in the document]
ajpalkovic: I removed the need to be a member to have write access to a thread, it was silly

src/managers/ThreadManager.js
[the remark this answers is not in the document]
ckohen: Addressed in 1c403d8 let me know if it needs more

src/managers/ChannelManager.js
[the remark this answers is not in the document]
ckohen: Added a `GuildChannelManager#channelCountWithoutThreads` in 1c403d8 does that work?

src/structures/Message.js
```diff
@@ -672,6 +683,21 @@ class Message extends Base {
     return this.channel.send(data);
   }
 
+  /**
+   * Create a new public thread from this message
+   * @see ThreadManager#create
+   * @param {string} name The name ofthe new Thread
```
iCrawl, line 689:
```suggestion
   * @param {string} name The name of the new Thread
```

src/structures/ThreadChannel.js
```diff
@@ -0,0 +1,393 @@
...
+   * @type {Date}
+   * @readonly
+   */
+  get archivedAt() {
+    return new Date(this.archiveTimestamp);
+  }
+
+  /**
+   * The parent channel of this thread
+   * @type {?NewsChannel|TextChannel}
```
iCrawl, line 151:
```suggestion
   * @type {?(NewsChannel|TextChannel)}
```

Our docgen is very picky with those apprently.

src/structures/ThreadChannel.js
```diff
@@ -0,0 +1,393 @@
...
+   * @param {boolean} archived Whether the thread is archived
+   * @param {string} [reason] Reason for archiving or unarchiving
+   * @returns {Promise<ThreadChannel>}
+   * @example
+   * // Set the thread to archived
+   * thread.setArchived(true)
+   *   .then(newThread => console.log(`Thread is now ${newThread.archived ? 'archived' : 'active'}`))
+   *   .catch(console.error);
+   */
+  setArchived(archived, reason) {
```
iCrawl, line 241: Shouldn't this default to true?
Just opening this for discussion.

src/structures/ThreadChannel.js
```diff
@@ -0,0 +1,393 @@
...
+   * @param {boolean} locked Whether the thread is locked
+   * @param {string} [reason] Reason for archiving or unarchiving
+   * @returns {Promise<ThreadChannel>}
+   * @example
+   * // Set the thread to locked
+   * thread.setLocked(true)
+   *   .then(newThread => console.log(`Thread is now ${newThread.locked ? 'locked' : 'unlocked'}`))
+   *   .catch(console.error);
+   */
+  setLocked(locked, reason) {
```
iCrawl, line 273: Same as above.

src/managers/ThreadManager.js
```diff
@@ -0,0 +1,233 @@
...
+
+  /**
+   * Obtains a thread from Discord, or the channel cache if it's already available.
+   * @param {ThreadChannelResolvable|FetchThreadsOptions} [options] If a ThreadChannelResolvable, the thread to fetch.
+   * If undefined, fetches all active threads.
+   * If an Object, fetches the specified threads.
+   * @param {boolean} [cache=true] Whether to cache the new thread object if it isn't already
+   * @param {boolean} [force=false] Whether to skip the cache check and request the API,
+   * only applies when fetching a single thread
+   * @returns {Promise<?ThreadChannel|FetchedThreads>}
```
iCrawl, line 134:
```suggestion
   * @returns {Promise<?(ThreadChannel|FetchedThreads)>}
```

iCrawl (changes requested):

src/structures/ThreadChannel.js
ckohen: Hmm, I do kind of like that, there are a couple other places where it would have been a similar decision that don't currently. I.e. `VoiceState#setMute`.

I do think this differs slightly from that though, especially since the likelihood of unarchiving this way seems quite low anyways.

iCrawl: Yes, you are correct. But I would rather change `VoiceState#setMute` too.

amishshah: I'm not opposed to having `VoiceState#setMute`, `VoiceState#setDeaf` etc. default to `true`.

amishshah (changes requested):
I'm happy with the implementation! Just found a few typos in the documentation, and a little confusion with `ThreadMemberFlags`. The [threads documentation](https://github.com/discord/discord-api-docs/blob/master/docs/topics/Threads.md) states:

> Currently the only flags are for notification settings

Are you deciding to not include these flags since they're not any use to the bot? Just wanted to clarify :+1:

src/util/ThreadMemberFlags.js
```diff
@@ -0,0 +1,30 @@
...
+ */
+
+/**
+ * Bitfield of the packed bits
+ * @type {number}
+ * @name ThreadMemberFlags#bitfield
+ */
+
+/**
+ * Numeric thread member flags. All available properties:
```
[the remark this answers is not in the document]
ckohen: There is at least 1 flag currently, I just don't know what it is and haven't figured it out. I would like to include before this gets merged though, so if anyone can figure it out (or maybe AJ can tell us).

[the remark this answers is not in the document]
ajpalkovic: The flags are: ```
export const ThreadMemberFlags = {
  HAS_INTERACTED: 1 << 0,
  ALL_MESSAGES: 1 << 1,
  ONLY_MENTIONS: 1 << 2,
  NO_MESSAGES: 1 << 3,
};
```

But they aren't documented in the official docs, and I don't plan to document them either. Mainly because I don't think any of them are really relevant for bots right now, they're user notification settings really, so there's no use for it. This is more preserving the option for us to add useful booleans to the thread member in the future.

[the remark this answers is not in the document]
Comment (ckohen) [reply src/util/ThreadMemberFlags.js:25]: Would you suggest not including the flags property in the thread member object then?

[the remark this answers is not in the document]
Comment (ajpalkovic) [reply src/util/ThreadMemberFlags.js:25]: That's hard for me to answer, I don't know your audience and user base, it's definitely a potential option though, I wouldn't say no to it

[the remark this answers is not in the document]
Comment (amishshah) [reply src/util/ThreadMemberFlags.js:25]: If they aren't documented officially then I'm happy for them not to be included here either 👍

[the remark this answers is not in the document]
Comment (iCrawl) [reply src/util/ThreadMemberFlags.js:25]: Overall we try to cover the public API (so usually only documented stuff), and don't allow user bots in any way shape or form.

File: src/structures/ThreadChannel.js (hunk 26 above)
Comment (ckohen) [reply src/structures/ThreadChannel.js:241]: I'm going to do these ones here because they are new, the voicestate ones should be a separate PR

File: src/util/ThreadMemberFlags.js [hunk 30]
```diff
@@ -0,0 +1,30 @@
...
+ */
+
+/**
+ * Bitfield of the packed bits
+ * @type {number}
+ * @name ThreadMemberFlags#bitfield
+ */
+
+/**
+ * Numeric thread member flags. There are currently no bitflags relevant to bts for this.
```
amishshah, line 25:
```suggestion
 * Numeric thread member flags. There are currently no bitflags relevant to bots for this.
```

amishshah (changes requested):

amishshah (approved):

iCrawl:
This needs a rebase.

src/client/actions/ThreadMemberUpdate.js
```diff
@@ -0,0 +1,30 @@
...
+        return { newMember };
+      }
+      const old = member._update(data);
+      /**
+       * Emitted whenever the client user's thread member is updated.
+       * @event Client#threadMemberUpdate
+       * @param {ThreadMember} oldMember The member before the update
+       * @param {ThreadMember} newMember The member after the update
+       */
+      client.emit(Events.THREAD_MEMBER_UPDATE, old, member);
```
vladfrangu, line 24: shouldn't you return the newMember here too?

src/managers/GuildChannelManager.js
```diff
@@ -21,6 +21,19 @@ class GuildChannelManager extends BaseManager {
...
+  /**
+   * The number of channels in this managers cache excluding thread channels
+   * that do not count towards a guild's maximum channels restriction.
+   * @type {number}
+   * @readonly
+   */
+  get channelCountWithoutThreads() {
+    return this.cache.reduce((acc, channel) => {
+      if (ThreadChannelTypes.includes(channel.type)) return acc;
+      return acc++;
```
vladfrangu, line 33: Won't this always return with 1 less? 
```suggestion
      return ++acc;
```

src/structures/Message.js
```diff
@@ -672,6 +683,21 @@ class Message extends Base {
...
+   * @param {string} name The name of the new Thread
+   * @param {ThreadAutoArchiveDuration} autoArchiveDuration How long before the thread is automatically archived
+   * @param {string} [reason] Reason for creating the thread
+   * @returns {Promise<ThreadChannel>}
+   */
+  startThread(name, autoArchiveDuration, reason) {
+    if (!['text', 'news'].includes(this.channel.type)) {
+      return Promise.reject(new Error('MESSAGE_THREAD_PARENT'));
+    }
+    return this.channel.threads.create(name, autoArchiveDuration, { startMessage: this, reason });
```
vladfrangu, line 698: Womp, this call is wrong

src/structures/TextChannel.js
```diff
@@ -73,11 +80,29 @@ class TextChannel extends GuildChannel {
       this.lastPinTimestamp = data.last_pin_timestamp ? new Date(data.last_pin_timestamp).getTime() : null;
     }
 
+    if ('default_auto_archive_duration' in data) {
+      /**
+       * The default auto archive duration for newly created threads in this channel
+       * @type {ThreadAutoArchiveDuration?}
```
vladfrangu, line 86: Isn't this technically nullable or possibly undefined

src/structures/ThreadChannel.js
```diff
@@ -0,0 +1,397 @@
...
+     * The timestamp the thread was last archived or unarchived at
+     * @type {?number}
+     */
+    this.archiveTimestamp = data.thread_metadata.archive_timestamp
+      ? new Date(data.thread_metadata.archive_timestamp).getTime()
+      : null;
+
+    /**
+     * The approximate count of messages in this thread
+     * <info> This value will not count above 50 even when there are more than 50 messages
```
vladfrangu, line 112:
```suggestion
     * <info>This value will not count above 50 even when there are more than 50 messages
```

src/structures/ThreadChannel.js
```diff
@@ -0,0 +1,397 @@
...
+     * If you need an approximate value higher than this, use ThreadChannel#messages.cache.size</info>
+     * @type {number}
+     */
+    this.messageCount = data.message_count;
+
+    /**
+     * The approximate count of users in this thread
+     * <info> This value will not count above 50 even when there are more than 50 members</info>
```
vladfrangu, line 120:
```suggestion
     * <info>This value will not count above 50 even when there are more than 50 members</info>
```

vladfrangu (dismissed):

src/managers/GuildChannelManager.js
kyranet: Actually, that'd always return 0, come think of it.

ckohen: Yeah, always returns 0 🤦

src/structures/Message.js
ckohen: 🤦

src/client/actions/ThreadMemberUpdate.js
ckohen: newMember is only for if the member wasn't already cached, so the event is not passed on to the user

src/structures/TextChannel.js
ckohen: ? at the end usually means undefined, looks like the site renders it as nullable though. definitely possible to be undefined, but its not nullable I think

iCrawl: I think our docgen only supports `?Type` properly, and `Type?` yields the same because of this.

But I would argue doing `?Type` is reasonable here, even if its not null, indicating nullish?

src/client/actions/ThreadListSync.js
```diff
@@ -0,0 +1,62 @@
...
+const Action = require('./Action');
+const Collection = require('../../util/Collection');
+const { Events } = require('../../util/Constants');
+
+class ThreadListSyncAction extends Action {
+  handle(data) {
+    const client = this.client;
+
+    const guild = client.guilds.cache.get(data.guild_id);
+    if (guild) {
```
synthghost, line 12: Might I suggest changing this line to:
```js
if (!guild) return {};
```
That way the next thirty or so lines don't need to be indented.

src/managers/ThreadManager.js
```diff
@@ -0,0 +1,233 @@
...
+   * @param {boolean} [force=false] Whether to skip the cache check and request the API,
+   * only applies when fetching a single thread
+   * @returns {Promise<?(ThreadChannel|FetchedThreads)>}
+   * @example
+   * // Fetch a thread by its id
+   * channel.threads.fetch('831955138126104859')
+   *   .then(channel => console.log(channel.name))
+   *   .catch(console.error);
+   */
+  fetch(options, cache = true, force = false) {
```
SpaceEEC, line 141: With #5824 being merged, this should also use `BaseFetchOptions`.
It looks like this PR is already rebased on-top of a commit including that PR's changes. :eyes:

SpaceEEC (changes requested):
Just a few nits

iCrawl (approved):

typings/index.d.ts
```diff
@@ -1203,6 +1217,11 @@ declare module 'discord.js' {
     public removeAttachments(): Promise<Message>;
     public reply(options: string | APIMessage | (ReplyMessageOptions & { split?: false })): Promise<Message>;
     public reply(options: APIMessage | (ReplyMessageOptions & { split: true | SplitOptions })): Promise<Message[]>;
+    public startThread(
+      name: 'string',
```
noftaly, line 1221:
```suggestion
      name: string,
```
is this normal ?

ckohen: oops, ty, nice catch

src/managers/ThreadManager.js
```diff
@@ -0,0 +1,232 @@
...
+   * Creates a new thread in the channel.
+   * @param {ThreadCreateOptions} [options] Options
+   * @returns {Promise<ThreadChannel>}
+   * @example
+   * // Create a new public thread
+   * channel.threads
+   *   .create({
+   *     name: 'food-talk'
+   *     autoArchiveDuration: 60,
+   *     startMessage: channel.lastMessageId,
```
vladfrangu, line 90:
```suggestion
   *     startMessage: channel.lastMessageID,
```

vladfrangu (changes requested):

src/managers/ThreadManager.js
```diff
@@ -0,0 +1,232 @@
...
+   */
+  async fetchArchived({ type = 'public', fetchAll = false, before, limit } = {}, cache = true) {
+    let path = this.client.api.channels(this.channel.id);
+    if (type === 'private' && fetchAll) {
+      path = path.users('@me');
+    }
+    let timestamp;
+    let id;
+    if (typeof before !== 'undefined') {
+      if (before instanceof ThreadChannel || /[0-9]{16,19}/g.test(String(before))) {
```
vladfrangu, line 191: Should probably make the regex stricter?
```suggestion
      if (before instanceof ThreadChannel || /^\d{16,19}$/.test(String(before))) {
```

ckohen: yes yes yes

src/managers/ThreadManager.js
```diff
@@ -0,0 +1,232 @@
...
+    const raw = await this.client.api.channels(this.channel.id).threads.active.get();
+    return this._mapThreads(raw, cache);
+  }
+
+  _mapThreads(rawThreads, cache) {
+    const threads = new Collection();
+    rawThreads.threads.forEach(raw => {
+      const thread = this.client.channels.add(raw, null, cache);
+      threads.set(thread.id, thread);
+    });
```
kyranet, line 223: This should use reduce

src/managers/ThreadManager.js
```diff
@@ -0,0 +1,232 @@
...
+   */
+  async fetchArchived({ type = 'public', fetchAll = false, before, limit } = {}, cache = true) {
+    let path = this.client.api.channels(this.channel.id);
+    if (type === 'private' && fetchAll) {
+      path = path.users('@me');
+    }
+    let timestamp;
+    let id;
+    if (typeof before !== 'undefined') {
+      if (before instanceof ThreadChannel || /^\d{16,19}$/.test(String(before))) {
```
kyranet, line 191: Is there a reason why you would like to support string dates?

src/managers/ThreadMemberManager.js
```diff
@@ -0,0 +1,98 @@
...
+   * @returns {Promise<Collection<Snowflake, ThreadMember>>}
+   */
+  async fetch(cache = true) {
+    const raw = await this.client.api.channels(this.thread.id, 'thread-members').get();
+    const members = new Collection();
+    raw.forEach(rawMember => {
+      const member = this.add(rawMember, cache);
+      members.set(member.id, member);
+    });
+    return members;
```
kyranet, line 94: This should use reduce

src/structures/ThreadChannel.js
```diff
@@ -0,0 +1,394 @@
...
+    return this.members.cache.mapValues(member => member.guildMember);
+  }
+
+  /**
+   * The time the thread was last archived or unarchived at
+   * @type {Date}
+   * @readonly
+   */
+  get archivedAt() {
+    return new Date(this.archiveTimestamp);
```
kyranet, line 143: `archiveTimestamp` is nullable, so it should be checked here (and also update the type tag)

typings/index.d.ts
```diff
@@ -1769,18 +1794,83 @@ declare module 'discord.js' {
...
     public setNSFW(nsfw: boolean, reason?: string): Promise<TextChannel>;
     public setRateLimitPerUser(rateLimitPerUser: number, reason?: string): Promise<TextChannel>;
     public setType(type: Pick<typeof ChannelType, 'text' | 'news'>, reason?: string): Promise<GuildChannel>;
     public fetchWebhooks(): Promise<Collection<Snowflake, Webhook>>;
   }
 
+  export class ThreadChannel extends TextBasedChannel(Channel) {
+    constructor(guild: Guild, data?: object);
+    public archived: boolean;
+    public readonly archivedAt: Date;
```
kyranet, line 1818:
```suggestion
    public readonly archivedAt: Date | null;
```

kyranet (dismissed):

src/structures/ThreadMember.js
```diff
@@ -0,0 +1,95 @@
...
+    return this.joinedTimestamp ? new Date(this.joinedTimestamp) : null;
+  }
+
+  /**
+   * The user that this thread member instance represents
+   * @type {?User}
+   * @readonly
+   */
+  get user() {
+    return this.client.users.cache.get(this.id) ?? null;
```
SpaceEEC, line 73:
```suggestion
    return this.client.users.resolve(this.id);
 ```

File: src/structures/ThreadMember.js [hunk 48]
```diff
@@ -0,0 +1,95 @@
...
+    this.flags = new ThreadMemberFlags(data.flags).freeze();
+  }
+
+  /**
+   * The guild member that this thread member instance represents
+   * @type {?GuildMember}
+   * @readonly
+   */
+  get guildMember() {
+    return this.thread.guild.members.cache.get(this.id) ?? null;
```
SpaceEEC, line 55:
```suggestion
    return this.thread.guild.members.resolve(this.id);
```

SpaceEEC (dismissed):
I only have these two nits:

src/managers/ThreadManager.js
ckohen: The API uses an ISO8601 date for full archived thread pagination instead of snowflakes, wanted to make that doable here without threads in cache.

src/managers/ThreadMemberManager.js
```diff
@@ -0,0 +1,96 @@
...
+
+  /**
+   * The cache of this Manager
+   * @type {Collection<Snowflake, ThreadMember>}
+   * @name ThreadMemberManager#cache
+   */
+
+  add(data, cache = true) {
+    const existing = this.cache.get(data.user_id);
+    if (existing && existing._patch && cache) existing._patch(data);
```
NotSugden, line 29:
```suggestion
    if (cache) existing?._patch?.(data);
```
```suggestion
    if (cache && existing?._patch) existing._patch(data);
```
iirc _patch should be present on all structures so could even just cut out the method truthiness check

src/client/actions/ThreadListSync.js
```diff
@@ -0,0 +1,68 @@
...
+      syncedThreads.set(thread.id, thread);
+    });
+
+    data.members.forEach(rawMember => {
+      // Discord sends the thread id as id in this object
+      const thread = client.channels.cache.get(rawMember.id);
+      if (thread) {
+        thread.members.add(rawMember);
+      }
+    });
```
vladfrangu, line 45: We iterate over the thread members TWICE, one assumes it can be an array or an object, the other that it's definitely an array. Can we do one iteration after you "sync threads"? Would probably achieve the exact same thing

src/client/actions/ThreadListSync.js
```diff
@@ -0,0 +1,68 @@
...
+    if (!data.channels_ids) {
+      guild.channels.cache.forEach(channel => {
+        this.removeStale(channel);
+      });
+    } else {
+      data.channel_ids.forEach(id => {
+        const channel = client.channels.resolve(id);
+        if (channel) this.removeStale(channel);
+      });
+    }
```
vladfrangu, line 23: Worlds smallest nit but can you inverse the if checks? so `if data.channel_ids: removeStaleFromThat(); else: removeAllStaleFromGuild()`

src/structures/ThreadChannel.js
```diff
@@ -0,0 +1,394 @@
...
+    this.messageCount = data.message_count;
+
+    /**
+     * The approximate count of users in this thread
+     * <info>This value will not count above 50 even when there are more than 50 members</info>
+     * @type {number}
+     */
+    this.memberCount = data.member_count;
+
+    if (data.member && this.client.user?.id) this.members.add({ user_id: this.client.user.id, ...data.member });
```
NotSugden, line 124:
```suggestion
    if (data.member && this.client.user) this.members.add({ user_id: this.client.user.id, ...data.member });
```
don't see why the id specifically wouldn't be truthy

src/managers/ThreadManager.js
```diff
@@ -0,0 +1,231 @@
...
+   * // Create a new private thread
+   * channel.threads
+   *   .create({ name: 'mod-talk', autoArchiveDuration: 60, reason: 'Needed a separate thread for moderation' })
+   *   .then(console.log)
+   *   .catch(console.error);
+   */
+  async create({ name, autoArchiveDuration, startMessage, reason } = {}) {
+    let path = this.client.api.channels(this.channel.id);
+    if (startMessage) {
+      startMessage = this.channel.messages.resolveID(startMessage);
```
vladfrangu, line 104: This should probably ensure it's not null and throw if it is. Also, if you could avoid mutation that'd be sweet too!

src/managers/ThreadManager.js
```diff
@@ -0,0 +1,231 @@
...
+    });
+
+    return this.client.actions.ThreadCreate.handle(data).thread;
+  }
+
+  /**
+   * The options for fetching multiple threads, the properties are mutually exclusive
+   * @typedef {Object} FetchThreadsOptions
+   * @property {FetchArchivedThreadOptions} [archived] The options used to fetch archived threads
+   * @property {boolean} [active] When true, fetches active threads. If archived is set, this is ignored!
```
vladfrangu, line 123:
```suggestion
   * @property {boolean} [active] When true, fetches active threads.<warn>If `archived` is set, this is ignored!</warn>
```

src/structures/ThreadChannel.js
```diff
@@ -0,0 +1,394 @@
...
+    return this.members.remove('@me').then(() => this);
+  }
+
+  /**
+   * Gets the overall set of permissions for a member or role in this threads' parent, taking into account overwrites.
+   * @param {GuildMemberResolvable|RoleResolvable} memberOrRole The member or role to obtain the overall permissions for
+   * @returns {?Readonly<Permissions>}
+   */
+  permissionsFor(memberOrRole) {
+    return this.parent?.permissionsFor(memberOrRole);
```
NotSugden, line 192:
```suggestion
    return this.parent?.permissionsFor(memberOrRole) ?? null;
```

`x?.y();` could return undefined

src/managers/ThreadManager.js
```diff
@@ -0,0 +1,231 @@
...
+    const raw = await this.client.api.channels(this.channel.id).threads.active.get();
+    return this._mapThreads(raw, cache);
+  }
+
+  _mapThreads(rawThreads, cache) {
+    const threads = rawThreads.threads.reduce((col, raw) => {
+      const thread = this.client.channels.add(raw, null, cache);
+      return col.set(thread.id, thread);
+    }, new Collection());
+    rawThreads.members.forEach(rawMember => this.client.channels.cache.get(rawMember.id)?.members.add(rawMember));
```
vladfrangu, line 223: Can't you tap into the `threads` collection you just created, since that ensures you have the thread made?

src/managers/ThreadMemberManager.js
```diff
@@ -0,0 +1,96 @@
...
+   */
+  remove(id, reason) {
+    return this.client.api
+      .channels(this.thread.id, 'thread-members', id)
+      .delete({ reason })
+      .then(() => id);
+  }
+
+  /**
+   * Fetches member(s) for the thread from Discord, requires `GUILD_MEMBERS` gateway intent.
```
vladfrangu, line 83: Does this actually need the `GUILD_MEMBERS` intent? Are we talking strictly having the `Guild Members Intent` enabled in the dev portal?

src/util/ThreadMemberFlags.js
```diff
@@ -0,0 +1,30 @@
...
+
+/**
+ * Bitfield of the packed bits
+ * @type {number}
+ * @name ThreadMemberFlags#bitfield
+ */
+
+/**
+ * Numeric thread member flags. There are currently no bitflags relevant to bots for this.
+ * @type {Object}
```
NotSugden, line 26:
```suggestion
 * @type {Object<string, number>}
```

src/structures/ThreadChannel.js
```diff
@@ -0,0 +1,394 @@
...
+      !this.archived &&
+      this.permissionsFor(this.client.user)?.has(
+        this.type === 'private_thread' ? Permissions.FLAGS.MANAGE_THREADS : Permissions.FLAGS.VIEW_CHANNEL,
+        false,
+      )
+    );
+  }
+
+  /**
+   * Whether the thread is manageable by the client user, for deleting or editing rateLimitPerUser or locked.
```
vladfrangu, line 324: Is the `, for deleting...` needed, since we'd have to update it if/when new properties are added

src/structures/ThreadChannel.js
```diff
@@ -0,0 +1,394 @@
...
+   * @param {number} rateLimitPerUser The new ratelimit in seconds
+   * @param {string} [reason] Reason for changing the thread's ratelimits
+   * @returns {Promise<ThreadChannel>}
+   */
+  setRateLimitPerUser(rateLimitPerUser, reason) {
+    return this.edit({ rateLimitPerUser }, reason);
+  }
+
+  /**
+   * Whether the thread is editable by the client user (name, archived, autoArchiveDuration)
```
vladfrangu, line 300: As above/below, whichever way GH arranges things

vladfrangu (changes requested):

src/util/ThreadMemberFlags.js
iCrawl: Is this actually valid JSDoc?

SpaceEEC: It at least seems to work as intended:
image

vladfrangu: > Is this actually valid JSDoc?

Yes, although the correct type would be `Record`. You can even do `import('').Type` and it'd work (but our site doesn't support that)

NotSugden: https://jsdoc.app/tags-type.html

src/client/actions/ThreadListSync.js
SpaceEEC: Looks like this accidentally sneaked in at this commit: https://github.com/discordjs/discord.js/pull/5570/commits/1d7455b9abffaf78d0be6d87285395183b42e4ed

SpaceEEC (approved):

src/managers/ThreadMemberManager.js
ckohen: This is probably a pattern from BaseManager that got copied into all managers, since ThreadMember obviously has _patch I'll remove the truthiness check here.

src/client/actions/ThreadListSync.js
ckohen: ahh, messed up the rebase a bit I suppose

src/managers/ThreadMemberManager.js
ckohen: This requires the intent to be enabled in the dev portal (or via intents request). Hopefully you'll be using it for gateway events too, but I'll clear that up

src/structures/ThreadChannel.js
ckohen: The names don't exactly describe what they do, to me these are both management type things so I listed what you can actually do with each of them, can remove if desired.

src/managers/ThreadManager.js
ckohen: I'll try this out, not sure if this will inline where it seems, but worth a shot

src/managers/ThreadManager.js
ckohen: nice call

src/client/actions/ThreadListSync.js
```diff
@@ -0,0 +1,60 @@
...
+      guild.channels.cache.forEach(channel => {
+        this.removeStale(channel);
+      });
+    }
+
+    const syncedThreads = new Collection();
+    data.threads.forEach(rawThread => {
+      const thread = client.channels.add(rawThread);
+      syncedThreads.set(thread.id, thread);
+    });
```
kyranet, line 29: Please use for-of

src/client/actions/ThreadListSync.js
```diff
@@ -0,0 +1,60 @@
...
+
+    if (data.channels_ids) {
+      data.channel_ids.forEach(id => {
+        const channel = client.channels.resolve(id);
+        if (channel) this.removeStale(channel);
+      });
+    } else {
+      guild.channels.cache.forEach(channel => {
+        this.removeStale(channel);
+      });
```
kyranet, line 22: Please use for-of

src/client/actions/ThreadListSync.js
```diff
@@ -0,0 +1,60 @@
...
+    const client = this.client;
+
+    const guild = client.guilds.cache.get(data.guild_id);
+    if (!guild) return {};
+
+    if (data.channels_ids) {
+      data.channel_ids.forEach(id => {
+        const channel = client.channels.resolve(id);
+        if (channel) this.removeStale(channel);
+      });
```
kyranet, line 18: Please use for-of

src/client/actions/ThreadListSync.js
```diff
@@ -0,0 +1,60 @@
...
+    }
+
+    const syncedThreads = new Collection();
+    data.threads.forEach(rawThread => {
+      const thread = client.channels.add(rawThread);
+      syncedThreads.set(thread.id, thread);
+    });
+
+    Object.values(data.members).forEach(rawMember => {
```
kyranet, line 31: You can either make a ternary with a for-of (`Array.isArray(data.members) ? data.members.entries() : Object.entries(data.members)`... or just a for-of loop

src/managers/ThreadManager.js
```diff
@@ -0,0 +1,232 @@
...
+    const raw = await this.client.api.channels(this.channel.id).threads.active.get();
+    return this._mapThreads(raw, cache);
+  }
+
+  _mapThreads(rawThreads, cache) {
+    const threads = rawThreads.threads.reduce((col, raw) => {
+      const thread = this.client.channels.add(raw, null, cache);
+      return col.set(thread.id, thread);
+    }, new Collection());
+    rawThreads.members.forEach(rawMember => threads.get(rawMember.id)?.members.add(rawMember));
```
kyranet, line 224: Please use for-of

kyranet (changes requested):

src/client/actions/ThreadListSync.js
ckohen: I ended up switching to reduce for this instead.

src/structures/ThreadChannel.js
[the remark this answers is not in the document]
ckohen: Now resolved with ad8d16c

ckohen:
ad8d16c refactors internals a bit, please let me know if I missed a spot where `add` should become `_add`. I think I got em all though

src/client/actions/ThreadListSync.js
```diff
@@ -0,0 +1,59 @@
...
+    const guild = client.guilds.cache.get(data.guild_id);
+    if (!guild) return {};
+
+    if (data.channels_ids) {
+      for (const id of data.channel_ids) {
+        const channel = client.channels.resolve(id);
+        if (channel) this.removeStale(channel);
+      }
+    } else {
+      for (const [, channel] of guild.channels.cache) {
```
kyranet, line 20:
```suggestion
      for (const channel of guild.channels.cache.values()) {
```

kyranet (changes requested):
Just one last thing, then LGTM.

kyranet (approved):

iCrawl:
You Already Know What That Means

kyranet:
@iCrawl 
what-are-you-waiting-for

vladfrangu:
Are you ready to... *thread* lightly?

iCrawl:
This needs a rebase.

ckohen:
Rebased and added the new thing from api docs pr

src/managers/ThreadManager.js
```diff
@@ -0,0 +1,244 @@
...
+   * @param {FetchArchivedThreadOptions} [options] The options to use when fetch archived threads
+   * @param {boolean} [cache=true] Whether to cache the new thread objects if they aren't already
+   * @returns {Promise<FetchedThreads>}
+   */
+  async fetchArchived({ type = 'public', fetchAll = false, before, limit } = {}, cache = true) {
+    let path = this.client.api.channels(this.channel.id);
+    if (type === 'private' && fetchAll) {
+      path = path.users('@me');
+    }
+    let timestamp;
```
NotSugden, line 200:
```suggestion
    let timestamp, id;
```

iCrawl:
```suggestion
    let timestamp;
    let id;
```

ajpalkovic:
Woohoo, congrats!
Thanks y'all :)

sharknix:
Lets go! Now I dont have to use private channels for my ticketing system

PR #5571, opened 2021-04-30
fix(MessageEmbed): null = field: undefined;

ay-valid:

if the messageembed, field is null/undefined then:

set text: undefined -> before: not always working/error in console.


Keke27210:
It is **up to you** to make sure you pass a string to value as it expects. It's not library's job to check that.

PR #5577, opened 2021-05-01
feat(Message): replace referencedMessage with fetchReference

monbrey:
This replaces the cache-reliant `Message#referencedMessage` with a more reliable way to resolve the Message
This follows precedent set by #5480 

The method with throw an error if `Message#reference` is null, or if the channel to fetch the message from is not in the client's cache, as is likely to be the case when the message is a crosspost rather than a reply.

This seemed like the best way to handle it without getting into raw APIs calls and partial channels and all that mess.

### DIFF
--- a/src/structures/Message.js
+++ b/src/structures/Message.js
@@ -409,12 +409,13 @@ class Message extends Base {
    * Fetches the Message this crosspost/reply/pin-add references, if available to the client
    * @returns {Promise<Message>}
    */
-  fetchReference() {
+  async fetchReference() {
     if (!this.reference) throw new Error('MESSAGE_REFERENCE_MISSING');
     const { channelID, messageID } = this.reference;
     const channel = this.client.channels.resolve(channelID);
     if (!channel) throw new Error('GUILD_CHANNEL_RESOLVE');
-    return channel.messages.fetch(messageID);
+    const message = await channel.messages.fetch(messageID);
+    return message;
   }
 
   /**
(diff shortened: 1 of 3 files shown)


src/structures/Message.js
```diff
@@ -406,15 +406,15 @@ class Message extends Base {
...
-   * @readonly
+   * Fetches the Message this crosspost/reply/pin-add references, if available to the client
+   * @returns {Promise<Message>}
    */
-  get referencedMessage() {
-    if (!this.reference) return null;
-    const referenceChannel = this.client.channels.resolve(this.reference.channelID);
-    if (!referenceChannel) return null;
-    return referenceChannel.messages.resolve(this.reference.messageID);
+  fetchReference() {
```
kyranet, line 412: This function needs to be marked as async (or use `return Promise.reject()`) so errors can be handled with `.catch()` handlers.
```suggestion
  async fetchReference() {
```

kyranet (changes requested):

src/structures/Message.js
monbrey: Should this also await on the last line? Linter will complain about an unused async otherwise.

ckohen: linter will complain about a useless await if you add it after a return. I think returen Promise.reject is a better option

kyranet: I'm pretty sure the rule checks for the `await` and `throw` keywords, as `async` changes their behaviour.

iCrawl (approved):

kyranet (approved):

SpaceEEC (approved):

PR #5586, opened 2021-05-04
feat(APIRequest): support setting global headers in HTTPOptions

advaith1:
Adds support for setting ClientOptions#http#headers to an object of headers which will be set on all API requests.

This is useful for using the canary API, setting a global audit log reason, and maybe other stuff in the future

The unrelated typings changes seem to have been automatically done by commit hooks

(commitlint didn't like naming the scope `APIRequest`)

### DIFF
--- a/src/rest/APIRequest.js
+++ b/src/rest/APIRequest.js
@@ -33,7 +33,7 @@ class APIRequest {
         ? this.client.options.http.api
         : `${this.client.options.http.api}/v${this.client.options.http.version}`;
     const url = API + this.path;
-    let headers = this.client.options.http.headers ?? {};
+    let headers = { ...this.client.options.http.headers };
 
     if (this.options.auth !== false) headers.Authorization = this.rest.getAuth();
     if (this.options.reason) headers['X-Audit-Log-Reason'] = encodeURIComponent(this.options.reason);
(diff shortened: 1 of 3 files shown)


kyranet (approved):

src/rest/APIRequest.js
```diff
@@ -33,7 +33,7 @@ class APIRequest {
         ? this.client.options.http.api
         : `${this.client.options.http.api}/v${this.client.options.http.version}`;
     const url = API + this.path;
-    let headers = {};
+    let headers = this.client.options.http.headers ?? {};
```
kyranet, line 36: Actually, now that I think of it, doesn't this need to be cloned? Otherwise, after the first run, it'll be mutated and the options will leak into next requests, e.g. the reason header.

kyranet (dismissed):

src/rest/APIRequest.js
vladfrangu:
```suggestion
    let headers = { ...this.client.options.http.headers ?? {} };
```

iCrawl (approved):

kyranet (approved):

SpaceEEC (approved):

PR #5588, opened 2021-05-05
/src/errors/Messages.js

Matse2005:
Make it clear that the error comes directly from discord.js and not from a third party package

Code changes have been tested against the Discord API, or there are no code changes


VoltrexMaster:
There's no need to clear anything up, error stacktraces exist for a reason, they make it clear that the error is thrown from discord.js.

Hugovidafe:
There's an example about what u can see right now
image
If u see after /node-modules/ you can see that error is coming from discord.js

PR #5598, opened 2021-05-08
fix(Client) added a forgotten keyword "this"

Syjalo:

The keyword "this" was forgotten. Without it, the error occurred:
```
this.presence = new ClientPresence(this, options.presence || 'online');
                                                 ^
TypeError: Cannot read property 'presence' of undefined
```

### DIFF
--- a/src/client/Client.js
+++ b/src/client/Client.js
@@ -129,7 +129,7 @@ class Client extends BaseClient {
      * @private
      * @type {ClientPresence}
      */
-    this.presence = new ClientPresence(this, options.presence);
+    this.presence = new ClientPresence(this, this.options.presence);
 
     Object.defineProperty(this, 'token', { writable: true });
     if (!this.token && 'DISCORD_TOKEN' in process.env) {


iShibi:
The error you've shown here isn't possible unless you're doing something of your own with the lib. In stable `options` is an empty object by default and in master it is neccessary to pass `options` which if you don't will throw missing intents error first before this one.

Syjalo:
> The error you've shown here isn't possible unless you're doing something of your own with the lib. In stable `options` is an empty object by default and in master it is neccessary to pass `options` which if you don't will throw missing intents error first before this one.

Without the keyword "this", options gets data from constructor. The data in constructor is empty, hence property "presence" cannot be read.
With the keyword "this", options gets data from class "BaseClient", where it was assigned default values, including for property "presence". So, only with the keyword "this" property "presence" can be read without an error.

iShibi:
Yes, `this.options` will never throw this error, but here is the thing: The error says that `options` is `undefined`, which is only possible if you didn't provide anything as `ClientOptions` when initializing the `Client` class. Which is what I'm saying will throw the `CLIENT_MISSING_INTENTS` error the moment you run the program. It will never get the chance to reach to the above point in code in order to throw the error you're facing.

Syjalo:
> Yes, `this.options` will never throw this error, but here is the thing: The error says that `options` is `undefined`, which is only possible if you didn't provide anything as `ClientOptions` when initializing the `Client` class. Which is what I'm saying will throw the `CLIENT_MISSING_INTENTS` error the moment you run the program. It will never get the chance to reach to the above point in code in order to throw the error you're facing.

Yes, but I think throw the error `CLIENT_MISSING_INTENTS` should be solved in another PR. This PR solves the error with property `presence`.

iShibi:
That error is thrown by the lib itself and the only way to fix it is by passing the valid intents and not a PR. What I'm not able to understand is that how you were able to get this error instead of that one.

Syjalo:
> That error is thrown by the lib itself and the only way to fix it is by passing the valid intents and not a PR. What I'm not able to understand is that how you were able to get this error instead of that one.

Ah, I see what you mean. I just noticed that this property takes data from the `constructor`, and so that the error `CLIENT_MISSING_INTENTS ` does not occur, I just removed it for a while. I didn't think about the fact that `options` is not empty because of discord.js uses a new version of Discord API. But I still think it's better to use `this`.

iCrawl (approved):

PR #5602, opened 2021-05-09
fix(Types): make event listeners accept async callbacks

kyranet:

When using `client.on` with a callback that returns a `Promise`, eslint will then complain if the `no-misused-promises` (from `@typescript-eslint` and various plugins) is enabled, as `Promise<?>` does not match `void`.

By making the return type of the callbacks `unknown`, we can now return any data, including promises. 

### DIFF
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -311,25 +311,25 @@ declare module 'discord.js' {
     public sweepMessages(lifetime?: number): number;
     public toJSON(): object;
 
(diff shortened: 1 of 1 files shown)


vladfrangu (approved):
I dunno... the type would be `void | Promise<void>` at best... I'm personally fine with this, but I don't know what the rest of the team thinks

zajrik (approved):

iCrawl (approved):

PR #5613, opened 2021-05-11
fix(ApplicationCommandManager): limit permission methods to guilds

vaporox:

These endpoints are not available for global commands.

### DIFF
--- a/src/errors/Messages.js
+++ b/src/errors/Messages.js
@@ -112,7 +112,8 @@ const Messages = {
 
   MEMBER_FETCH_NONCE_LENGTH: 'Nonce length must not exceed 32 characters.',
 
-  GLOBAL_COMMAND: "Permissions for global commands may only be fetched or modified from a guild's application command manager.",
+  GLOBAL_COMMAND:
+    "Permissions for global commands may only be fetched or modified from a guild's application command manager.",
 
   INTERACTION_ALREADY_REPLIED: 'This interaction has already been deferred or replied to.',
 };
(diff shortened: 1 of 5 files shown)


SpaceEEC (approved):
Maybe also add a  `<warn>`ing in the `ApplicationCommand` methods docstrings?

iCrawl (approved):

src/errors/Messages.js
```diff
@@ -112,6 +112,8 @@ const Messages = {
 
   MEMBER_FETCH_NONCE_LENGTH: 'Nonce length must not exceed 32 characters.',
 
+  GLOBAL_COMMAND: 'This method cannot be used on global application commands.',
```
vladfrangu, line 115: Personally, this should explain in some way that you should use a guild's command manager instead for fetching the permissions

vladfrangu (changes requested):
CC @iCrawl @kyranet

src/errors/Messages.js
vaporox: Do you have a suggestion for that?

vladfrangu: "Permissions may only be fetched or modified by calling a guild's application command manager's functions" or similar?

src/errors/Messages.js
```diff
@@ -112,6 +112,9 @@ const Messages = {
 
   MEMBER_FETCH_NONCE_LENGTH: 'Nonce length must not exceed 32 characters.',
 
+  GLOBAL_COMMAND:
```
kyranet, line 115: This key name doesn't really represent its content since it's about global command's permissions rather than global commands themselves:
```suggestion
  GLOBAL_COMMAND_PERMISSIONS:
```

kyranet (changes requested):

kyranet (approved):

PR #5618, opened 2021-05-13
feat(CommandInteraction): ephemeral followup messages

monbrey:
This adds a new method `CommandInteraction#followUp` for sending ephemeral followup messages (if the option is provided).

This would not be possible via `CommandInteraction#webhook#send` as regular Webhook messages cannot be ephemeral. The methods should otherwise be equivalent for non-ephemeral messages.

### DIFF
--- a/src/structures/CommandInteraction.js
+++ b/src/structures/CommandInteraction.js
@@ -188,12 +188,12 @@ class CommandInteraction extends Interaction {
    */
 
   /**
-   * Send a followup message to this interaction.
+   * Send a follow-up message to this interaction.
    * @param {string|APIMessage|MessageAdditions} content The content for the reply
    * @param {InteractionReplyOptions} [options] Additional options for the reply
    * @returns {Promise<Message|Object>}
    */
-  async followup(content, options) {
+  async followUp(content, options) {
     const apiMessage = content instanceof APIMessage ? content : APIMessage.create(this, content, options);
     const { data, files } = await apiMessage.resolveData().resolveFiles();
 
--- a/src/structures/CommandInteraction.js
+++ b/src/structures/CommandInteraction.js
@@ -191,16 +191,18 @@ class CommandInteraction extends Interaction {
    * Send a followup message to this interaction.
    * @param {string|APIMessage|MessageAdditions} content The content for the reply
    * @param {InteractionReplyOptions} [options] Additional options for the reply
-   * @returns {Promise<void>}
+   * @returns {Promise<Message|Object>}
    */
   async followup(content, options) {
     const apiMessage = content instanceof APIMessage ? content : APIMessage.create(this, content, options);
     const { data, files } = await apiMessage.resolveData().resolveFiles();
 
-    await this.client.api.webhooks(this.applicationID, this.token).post({
+    const raw = await this.client.api.webhooks(this.applicationID, this.token).post({
       data,
       files,
     });
+
+    return this.channel?.messages.add(raw) ?? raw;
   }
 
   /**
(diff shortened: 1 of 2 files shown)


src/structures/CommandInteraction.js
```diff
@@ -187,6 +187,22 @@ class CommandInteraction extends Interaction {
    * @property {Role|Object} [role] The resolved role
    */
 
+  /**
+   * Send a followup message to this interaction.
+   * @param {string|APIMessage|MessageAdditions} content The content for the reply
+   * @param {InteractionReplyOptions} [options] Additional options for the reply
+   * @returns {Promise<void>}
```
ckohen, line 194: Why not return the created message?

monbrey: I assumed for the same reason none of the other methods do. Will check what it returns.

vladfrangu (approved):

src/structures/CommandInteraction.js
```diff
@@ -187,6 +187,24 @@ class CommandInteraction extends Interaction {
    * @property {Role|Object} [role] The resolved role
    */
 
+  /**
+   * Send a followup message to this interaction.
+   * @param {string|APIMessage|MessageAdditions} content The content for the reply
+   * @param {InteractionReplyOptions} [options] Additional options for the reply
+   * @returns {Promise<Message|Object>}
+   */
+  async followup(content, options) {
```
kyranet, line 196: According to [Grammarly](https://www.grammarly.com/blog/follow-up-followup/), this is wrong, the verb is to `follow up` and the act is a `follow-up`, both of which are two words. As such, this should be renamed to `followUp` following camelCase conventions.

image

```suggestion
  async followUp(content, options) {
```

src/structures/CommandInteraction.js
```diff
@@ -187,6 +187,24 @@ class CommandInteraction extends Interaction {
    * @property {Role|Object} [role] The resolved role
    */
 
+  /**
+   * Send a followup message to this interaction.
```
kyranet, line 191:
```suggestion
   * Send a follow-up message to this interaction.
```

typings/index.d.ts
```diff
@@ -425,6 +425,10 @@ declare module 'discord.js' {
     ): Promise<Message | RawMessage>;
     public editReply(content: string, options?: WebhookEditMessageOptions): Promise<Message | RawMessage>;
     public fetchReply(): Promise<Message | RawMessage>;
+    public followup(
```
kyranet, line 428:
```suggestion
    public followUp(
```

kyranet (changes requested):

src/structures/CommandInteraction.js
monbrey: I agree 100% buuuuuut

Screenshot_20210513-182414_Chrome

kyranet: That deserves a PR on their end (plus, you know how many typos people find in the discord-api-docs repo).

typings/index.d.ts
```diff
@@ -425,6 +425,10 @@ declare module 'discord.js' {
...
+      content: string | APIMessage | InteractionReplyOptions | MessageAdditions,
+    ): Promise<Message | RawMessage>;
+    public followup(content: string, options?: InteractionReplyOptions): Promise<Message | RawMessage>;
```
monbrey, line 431:
```suggestion
    public followUp(content: string, options?: InteractionReplyOptions): Promise<Message | RawMessage>;
```

kyranet (approved):

SpaceEEC (approved):

iCrawl (approved):

PR #5030, opened 2020-11-22
Add new wrapper Widget

ThumusLive:
**Please describe the changes this PR makes and why it should be merged:**
Hi, I make a method to request the widget of a guild. It's really simple.
I think it should be merged because it's a important feature of discord.
**Status**

- [x] Code changes have been tested against the Discord API, or there are no code changes
- [ ] I know how to update typings and have done so, or typings don't need updating

**Semantic versioning classification:**

- [x] This PR changes the library's interface (methods or parameters added)
  - [ ] This PR includes breaking changes (methods removed or renamed, parameters moved or removed)
- [ ] This PR **only** includes non-code changes, like changes to documentation, README, etc.

PR #5619, opened 2021-05-13
feat(Widget): wrapper for widget.json

ThumusLive:
Okay, I'm remaking a new pr because my last was really terrible + i've deleted the original (#5030)

My goal is to add Widget in discord.js
Widget is a part of Discord and it's not fully featured

Thanks to @ckohen @advaith1 @SpaceEEC @almostSouji @vladfrangu @NotSugden @kyranet @Mesteery for helping me in the last pull request


typings/index.d.ts
```diff
@@ -3661,6 +3674,30 @@ declare module 'discord.js' {
     $device?: string;
   }
 
+  interface WidgetMember {
```
ThumusLive, line 3677: I've made this typedef because these members aren't real, the id is just a number (1, 2, 3, etc...) and I don't think an new class is needed

src/client/Client.js
```diff
@@ -365,6 +366,20 @@ class Client extends BaseClient {
...
+   * @param {GuildResolvable} guild The guild to fetch the widget for
+   * @returns {Promise<Widget>}
+   */
+  fetchWidget(guild) {
+    const id = this.guilds.resolveID(guild);
+    if (!id) throw new TypeError('INVALID_TYPE', 'guild', 'GuildResolvable');
+    return this.api
+      .guilds(id, 'widget.json')
+      .get()
+      .then(data => new Widget(this, data));
```
monbrey, line 380: I believe we've started to prefer async over then where possible, e.g.
```suggestion
  async fetchWidget(guild) {
    const id = this.guilds.resolveID(guild);
    if (!id) return Promise.reject(new TypeError('INVALID_TYPE', 'guild', 'GuildResolvable'));
    const data = await this.api.guilds(id, 'widget.json').get();
    return new Widget(this, data);
```

typings/index.d.ts
```diff
@@ -3661,6 +3674,30 @@ declare module 'discord.js' {
...
+    self_mute?: boolean;
+    suppress?: boolean;
+    channel_id?: string;
+    avatar_url: string;
+    activity?: {
+      name: string;
+    };
+  }
+
+  interface WidgetChannel {
```
ThumusLive, line 3695: This interface here is a regular channel but the channel has the possibility to not be already cached by the client (and possibly can't be cached)

src/structures/Widget.js
```diff
@@ -0,0 +1,93 @@
...
+  /**
+   * Builds the widget with the provided data.
+   * @param {*} data The raw data of the widget
+   * @private
+   */
+  _patch(data) {
+    /**
+     * When a widget is disable, data will always send a code. When it's enable, data doesn't have a code.
+     */
+    if ('code' in data) {
```
monbrey, line 27: I don't really understand why you're creating a Widget class for a JSON error. I'd suggest throwing a `WIDGET_DISABLED` error or something instead of constructing this class at all.

src/structures/Widget.js
```diff
@@ -0,0 +1,93 @@
...
+       * The error message.
+       * @type {?string}
+       */
+      this.message = data.message;
+
+      /**
+       * Check if the guild has the widget enabled or not.
+       * @type {boolean}
+       */
+      this.enabled = false;
```
ThumusLive, line 44: "enabled" is the only thing who is everytime set, it's to help people for checking if the guild has the widget enabled or not

src/structures/Widget.js
```diff
@@ -0,0 +1,93 @@
...
+
+      /**
+       * The name of the guild.
+       * @type {?string}
+       */
+      this.name = data.name;
+
+      /**
+       * The invite of the guild.
+       * @type {?string | ?null}
```
monbrey, line 60: Why would it be an optional null if not a string?

src/structures/Widget.js
```diff
@@ -0,0 +1,93 @@
...
+       */
+      this.instant_invite = data.instant_invite;
+
+      /**
+       * The list of channels in the guild.
+       * @type {Collection<string, WidgetChannel>}
```
monbrey, line 66: Does it not receive enough data to construct a regular Channel here?

src/structures/Widget.js
```diff
@@ -0,0 +1,93 @@
...
+       */
+      this.channels = new Collection();
+      for (const channel of data.channels) {
+        this.channels.set(channel.id, channel);
+      }
+
+      /**
+       * The list of members in the guild.
+       * @type {Collection<string, WidgetMember>}
```
monbrey, line 75: Same question regarding WidgetMember - can we not construct a regular GuildMember? Otherwise it appears we're just setting raw objects into these Collections - maybe we can reference `discord-api-types`?

typings/index.d.ts
```diff
@@ -2029,6 +2029,19 @@ declare module 'discord.js' {
...
+  export class Widget extends Base {
+    constructor(client: Client, data: object);
+    private _patch(data: object): void;
+    public enabled: boolean;
+    public code?: number;
+    public message?: string;
+    public id?: string;
+    public instant_invite?: string | null;
+    public channels?: Collection<string, WidgetChannel>;
+    public members?: Collection<string, WidgetMember>;
```
monbrey, line 2041: See above comments regarding these types

typings/index.d.ts
```diff
@@ -3661,6 +3674,30 @@ declare module 'discord.js' {
...
+      name: string;
+    };
+  }
+
+  interface WidgetChannel {
+    id: string;
+    name: string;
+    position: number;
+  }
+
```
monbrey, line 3700: See above comments regarding these types

src/structures/Widget.js
ThumusLive: I don't know at all how to make it :(
But yeah it might be a good idea

monbrey: I think the rest of the constructor (in the `else` part) is fine. But if the Widget is disabled, check that back in the `fetchWidget` function and reject the Promise instead of returning a Widget

src/structures/Widget.js
ThumusLive: That's weird I know but I need to explain
When the raw object comes, invite can appears or not due to the error message and the code
image
But "instant_invite" can be null too when the widget is enabled :
image

monbrey: Yeah but as I said in my other comment, that first screenshot is a JSON error, not a widget. You shouldn't create one at all if you get an error code.

src/structures/Widget.js
ThumusLive: It seems not and the raw data just have 3 parameters : id, position and name

src/structures/Widget.js
ThumusLive: Here it's clear that we can't, id is just a enum, 1 to the number of members

ThumusLive:
It's a mess in the logs but I've added in client an Promise.reject + clean up the if condition in the Widget class

src/client/Client.js
```diff
@@ -365,6 +366,19 @@ class Client extends BaseClient {
...
   }
 
+  /**
+   * Obtains the widget of a guild from Discord, available for all guilds.
+   * @param {GuildResolvable} guild The guild to fetch the widget for
+   * @returns {Promise<Widget>}
+   */
+  async fetchWidget(guild) {
+    const id = this.guilds.resolveID(guild);
+    if (!id) return Promise.reject(new TypeError('INVALID_TYPE', 'guild', 'GuildResolvable'));
```
PyroTechniac, line 376: These are in an async method, so you can just throw the errors

PyroTechniac (changes requested):

src/client/Client.js
monbrey: I don't remember the exact difference, but I know there *is* a difference between the two. Possibly relating to async stack traces?

I think it was @kyranet I discussed it with

ThumusLive: If something is wrong I can just revert it.
Btw I've seen this throw Error in Client#login (who is async)

ckohen: > I don't remember the exact difference, but I know there is a difference between the two. Possibly relating to async stack traces or being able to .catch the result.

I believe the difference should be unnoticeable in most instances, not sure about performance wise. But, AFAIK, throwing in an async function causes the Promise created by being an async function to reject, whereas `return Promise.reject` resolves the internal async Promise with a rejected Promise, which will still fall through to the catch because that is how Promise chains work.

monbrey: Yeah, youre right, this can be a throw. Its when you return a Promise (eg the .then it was previously) from a non-async function that this is required.

src/client/Client.js
```diff
@@ -365,6 +366,19 @@ class Client extends BaseClient {
       .then(data => new GuildPreview(this, data));
   }
 
+  /**
+   * Obtains the widget of a guild from Discord, available for all guilds.
```
vladfrangu, line 370: Not true, guilds must have the widget enabled if I recall correctly 👀

src/structures/Widget.js
```diff
@@ -0,0 +1,68 @@
...
+
+    /**
+     * The name of the guild.
+     * @type {string}
+     */
+    this.name = data.name;
+
+    /**
+     * The invite of the guild.
+     * @type {string | null}
```
vladfrangu, line 38: Not how we do jsdocs for it

src/structures/Widget.js
```diff
@@ -0,0 +1,68 @@
...
+
+const Base = require('./Base');
+const Collection = require('../util/Collection');
+
+/**
+ * Represents a Widget.
+ */
+class Widget extends Base {
+  /**
+   * @param {Client} client - The instantiating client
```
vladfrangu, line 11: Do we do the `-` anywhere else?

src/structures/Widget.js
```diff
@@ -0,0 +1,68 @@
...
+     * The name of the guild.
+     * @type {string}
+     */
+    this.name = data.name;
+
+    /**
+     * The invite of the guild.
+     * @type {string | null}
+     */
+    this.instant_invite = data.instant_invite;
```
vladfrangu, line 40: Properties are camelCased in discord.js

vladfrangu (dismissed):

src/structures/Widget.js
kyranet: We do in `@discordjs/voice`, but I don't recall this in the base d.js repository.

ThumusLive:
I have add a new function, exactly the same as the Client#fetchWidget but in the Widget to update the widget

typings/index.d.ts
```diff
@@ -2029,6 +2029,17 @@ declare module 'discord.js' {
     public once(event: string, listener: (...args: any[]) => void): this;
   }
 
+  export class Widget extends Base {
+    constructor(client: Client, data: object);
+    private _patch(data: object): void;
+    public fetch(): Promise<Widget>;
+    public id: string;
+    public instant_invite: string | null;
```
BannerBomb, line 2037:
```suggestion
    public instantInvite: string | null;
```

typings/index.d.ts
```diff
@@ -2029,6 +2029,17 @@ declare module 'discord.js' {
...
+    public channels: Collection<string, WidgetChannel>;
+    public members: Collection<string, WidgetMember>;
+    public presence_count: number;
```
BannerBomb, line 2040:
```suggestion
    public presence_count: number;
    public presenceCount: number;
```

src/structures/Widget.js
```diff
@@ -0,0 +1,79 @@
...
+  }
+  /**
+   * Builds the widget with the provided data.
+   * @param {*} data The raw data of the widget
+   * @private
+   */
+  _patch(data) {
+    /**
+     * The id of the guild.
+     * @type {string}
```
BannerBomb, line 26: shouldn't all the ids be typed as "Snowflake" instead of "string"?

ThumusLive: Done in d3e90ba

ThumusLive:
I have done a test but that change anything I guess ? I've just link my local to my github project

src/structures/Widget.js
```diff
@@ -0,0 +1,79 @@
...
+
+    /**
+     * The invite of the guild.
+     * @type {?string}
+     */
+    this.instantInvite = data.instant_invite;
+
+    /**
+     * The list of channels in the guild.
+     * @type {Collection<string, WidgetChannel>}
```
BannerBomb, line 44:
```suggestion
     * @type {Collection<Snowflake, WidgetChannel>}
```

src/structures/Widget.js
```diff
@@ -0,0 +1,79 @@
...
+     */
+    this.channels = new Collection();
+    for (const channel of data.channels) {
+      this.channels.set(channel.id, channel);
+    }
+
+    /**
+     * The list of members in the guild.
+     * @type {Collection<string, WidgetMember>}
```
BannerBomb, line 53:
```suggestion
     * @type {Collection<Snowflake, WidgetMember>}
```

typings/index.d.ts
```diff
@@ -2029,6 +2029,17 @@ declare module 'discord.js' {
     public once(event: string, listener: (...args: any[]) => void): this;
   }
 
+  export class Widget extends Base {
+    constructor(client: Client, data: object);
+    private _patch(data: object): void;
+    public fetch(): Promise<Widget>;
+    public id: string;
+    public instantInvite: string | null;
+    public channels: Collection<string, WidgetChannel>;
```
BannerBomb, line 2038:
```suggestion
    public channels: Collection<Snowflake, WidgetChannel>;
```

typings/index.d.ts
```diff
@@ -2029,6 +2029,17 @@ declare module 'discord.js' {
...
+    public members: Collection<string, WidgetMember>;
```
BannerBomb, line 2039:
```suggestion
    public members: Collection<Snowflake, WidgetMember>;
```

typings/index.d.ts
```diff
@@ -3661,6 +3672,30 @@ declare module 'discord.js' {
     $device?: string;
   }
 
+  interface WidgetMember {
+    id: string;
```
BannerBomb, line 3676:
```suggestion
    id: Snowflake;
```

typings/index.d.ts
```diff
@@ -3661,6 +3672,30 @@ declare module 'discord.js' {
...
+    username: string;
+    discriminator: string;
+    avatar: null;
+    status: PresenceStatus;
+    deaf?: boolean;
+    mute?: boolean;
+    self_deaf?: boolean;
+    self_mute?: boolean;
+    suppress?: boolean;
+    channel_id?: string;
```
BannerBomb, line 3686:
```suggestion
    channel_id?: Snowflake;
```

typings/index.d.ts
```diff
@@ -3661,6 +3672,30 @@ declare module 'discord.js' {
...
+    avatar_url: string;
+    activity?: {
+      name: string;
+    };
+  }
+
+  interface WidgetChannel {
+    id: string;
```
BannerBomb, line 3694:
```suggestion
    id: Snowflake;
```

typings/index.d.ts
```diff
@@ -2029,6 +2029,17 @@ declare module 'discord.js' {
     public once(event: string, listener: (...args: any[]) => void): this;
   }
 
+  export class Widget extends Base {
+    constructor(client: Client, data: object);
+    private _patch(data: object): void;
+    public fetch(): Promise<Widget>;
+    public id: string;
```
BannerBomb, line 2036:
```suggestion
    public id: Snowflake;
```

src/structures/Widget.js
ThumusLive: It's not a Snowflake for members, it's just a number

typings/index.d.ts
ThumusLive: Same issue

typings/index.d.ts
ThumusLive: Same issue

src/client/Client.js
```diff
@@ -365,6 +366,19 @@ class Client extends BaseClient {
...
+  /**
+   * Obtains the widget of a guild from Discord, available for guilds with the widget enabled.
+   * @param {GuildResolvable} guild The guild to fetch the widget for
+   * @returns {Promise<Widget>}
+   */
+  async fetchWidget(guild) {
+    const id = this.guilds.resolveID(guild);
+    if (!id) throw new TypeError('INVALID_TYPE', 'guild', 'GuildResolvable');
+    const data = await this.api.guilds(id, 'widget.json').get();
+    if (data.code) throw new Error('WIDGET_DISABLED');
```
SpaceEEC, line 378: This will never evaluate to true, since a disabled widget won't result in a response with a successful response code.
```suggestion
```

src/structures/Widget.js
```diff
@@ -0,0 +1,79 @@
...
+     * The list of channels in the guild.
+     * @type {Collection<Snowflake, WidgetChannel>}
+     */
+    this.channels = new Collection();
+    for (const channel of data.channels) {
+      this.channels.set(channel.id, channel);
+    }
+
+    /**
+     * The list of members in the guild.
```
SpaceEEC, line 52: Maybe note down that the keys won't be user ids, but just arbitrary numbers?

src/structures/Widget.js
```diff
@@ -0,0 +1,79 @@
...
+    this.presenceCount = data.presence_count;
+  }
+
+  /**
+   * Update the Widget.
+   * @returns {Promise<Widget>}
+   */
+  async fetch() {
+    const data = await this.client.api.guilds(this.id, 'widget.json').get();
+    if (data.code) throw new Error('WIDGET_DISABLED');
```
SpaceEEC, line 73: Same thing applies here too.
```suggestion
```

SpaceEEC (changes requested):

src/client/Client.js
ThumusLive: I don't think so, I have tried last time and the error was triggered, I'll retry it and I'll update this message later

iCrawl:
This needs another rebase.

ThumusLive:
Wdym by an other rebase ?

iCrawl:
You have conflicts, they need to be resolved.

ThumusLive:
Okay but how can I resolve this conflict ? First time I've seen this

src/structures/Widget.js
```diff
@@ -0,0 +1,79 @@
...
+    }
+
+    /**
+     * The list of members in the guild.
+     * These strings are just arbitrary numbers, they aren't Snowflakes.
+     * @type {Collection<string, WidgetMember>}
+     */
+    this.members = new Collection();
+    for (const member of data.members) {
+      this.members.set(member.id, member);
```
kyranet, line 58: I just noticed this is raw data from Discord which is exposed to the user, and as such, users would require to use `snake_case` for some properties (`self_deaf`, `self_mute`...) when reading them.

Shouldn't this have its own structure?

kyranet (changes requested):

typings/index.d.ts
```diff
@@ -2029,6 +2029,17 @@ declare module 'discord.js' {
...
   }
 
+  export class Widget extends Base {
+    constructor(client: Client, data: object);
+    private _patch(data: object): void;
+    public fetch(): Promise<Widget>;
+    public id: Snowflake;
+    public instantInvite: string | null;
+    public channels: Collection<Snowflake, WidgetChannel>;
+    public members: Collection<string, WidgetMember>;
```
kyranet, line 2039: `WidgetMember` is not typed, please add the class definition.

src/structures/WidgetMember.js
```diff
@@ -0,0 +1,93 @@
...
+    /**
+     * If the user is in a VoiceChannel, it's the id of the VoiceChannel of the member.
+     * @type {?Snowflake}
+     */
+    this.channelID = data.channel_id;
+    /**
+     * The avatarUrl of the member.
+     * @type {string}
+     */
+    this.avatarUrl = data.avatar_url;
```
kyranet, line 84: We use `URL` (see: User#avatarURL, User#displayAvatarURL...):
```suggestion
    this.avatarURL = data.avatar_url;
```

src/structures/WidgetMember.js
```diff
@@ -0,0 +1,93 @@
...
+   * Builds the WidgetMember with the provided data.
+   * @param {*} data The raw data of the WidgetMember
+   * @private
+   */
+  _patch(data) {
+    /**
+     * The id of the user. It's an arbitrary number.
+     * @type {string}
+     */
+    this.id = data.id;
```
kyranet, line 27: I don't think this is supposed to change, please move it to the constructor.

src/structures/WidgetMember.js
```diff
@@ -0,0 +1,93 @@
...
+     */
+    this.channelID = data.channel_id;
+    /**
+     * The avatarUrl of the member.
+     * @type {string}
+     */
+    this.avatarUrl = data.avatar_url;
+    /**
+     * The activity of the member.
+     * @type {?WidgetActivity}
```
kyranet, line 87: The `WidgetActivity` typedef does not exist. You can use this [existing typedef](https://github.com/discordjs/discord.js/blob/51551f544b80d7d27ab0b315da01dfc560b2c115/src/structures/Presence.js#L8-L13) as a reference for writing this one.

kyranet (changes requested):

src/structures/WidgetMember.js
ThumusLive: The actual problem is it just contains a name, that's it

kyranet: Then you write it as
```js
/**
 * Activity sent in a {@link WidgetMember}.
 * @typedef {Object} WidgetActivity
 * @property {string} name The name of the activity
 */
```

Somewhere in this code.

src/index.js
```diff
@@ -115,6 +115,7 @@ module.exports = {
   VoiceRegion: require('./structures/VoiceRegion'),
   VoiceState: require('./structures/VoiceState'),
   Webhook: require('./structures/Webhook'),
+  Widget: require('./structures/Widget'),
```
SpaceEEC, line 118: WidgetMember should also be exported:
```suggestion
  Widget: require('./structures/Widget'),
  WidgetMember: require('./structures/WidgetMember'),
```

SpaceEEC (changes requested):
Otherwise LGTM

src/structures/WidgetMember.js
```diff
@@ -0,0 +1,90 @@
...
+     * @type {string}
+     */
+    this.username = data.username;
+
+    /**
+     * The discriminator of the member.
+     * @type {string}
+     */
+    this.discriminator = data.discriminator;
+    /**
```
kyranet, line 37:
```suggestion
    this.discriminator = data.discriminator;

    /**
```

src/structures/WidgetMember.js
```diff
@@ -0,0 +1,90 @@
...
+     * The avatar of the member.
+     * @type {?string}
+     */
+    this.avatar = data.avatar;
+    /**
```
kyranet, line 42:
```suggestion
    this.avatar = data.avatar;

    /**
```

src/structures/WidgetMember.js
```diff
@@ -0,0 +1,90 @@
...
+     * If the user is in a VoiceChannel, it's the id of the VoiceChannel of the member.
+     * @type {?Snowflake}
+     */
+    this.channelID = data.channel_id;
+    /**
+     * The avatarUrl of the member.
+     * @type {string}
+     */
+    this.avatarURL = data.avatar_url;
+    /**
```
kyranet, line 82:
```suggestion
    this.avatarURL = data.avatar_url;

    /**
```

src/structures/WidgetMember.js
```diff
@@ -0,0 +1,90 @@
...
+     * If the user is in a VoiceChannel, it's the supress state of the member.
+     * @type {?boolean}
+     */
+    this.suppress = data.suppress;
+    /**
+     * If the user is in a VoiceChannel, it's the id of the VoiceChannel of the member.
+     * @type {?Snowflake}
+     */
+    this.channelID = data.channel_id;
+    /**
```
kyranet, line 77:
```suggestion
    this.channelID = data.channel_id;

    /**
```

src/structures/WidgetMember.js
```diff
@@ -0,0 +1,90 @@
...
+     * If the user is in a VoiceChannel, it's the selfMute state of the member.
+     * @type {?boolean}
+     */
+    this.selfMute = data.self_mute;
+    /**
+     * If the user is in a VoiceChannel, it's the supress state of the member.
+     * @type {?boolean}
+     */
+    this.suppress = data.suppress;
+    /**
```
kyranet, line 72:
```suggestion
    this.suppress = data.suppress;

    /**
```

src/structures/WidgetMember.js
```diff
@@ -0,0 +1,90 @@
...
+     * If the user is in a VoiceChannel, it's the selfDeaf state of the member.
+     * @type {?boolean}
+     */
+    this.selfDeaf = data.self_deaf;
+    /**
+     * If the user is in a VoiceChannel, it's the selfMute state of the member.
+     * @type {?boolean}
+     */
+    this.selfMute = data.self_mute;
+    /**
```
kyranet, line 67:
```suggestion
    this.selfMute = data.self_mute;

    /**
```

src/structures/WidgetMember.js
```diff
@@ -0,0 +1,90 @@
...
+     * If the user is in a VoiceChannel, it's the mute state of the member.
+     * @type {?boolean}
+     */
+    this.mute = data.mute;
+    /**
+     * If the user is in a VoiceChannel, it's the selfDeaf state of the member.
+     * @type {?boolean}
+     */
+    this.selfDeaf = data.self_deaf;
+    /**
```
kyranet, line 62:
```suggestion
    this.selfDeaf = data.self_deaf;

    /**
```

src/structures/WidgetMember.js
```diff
@@ -0,0 +1,90 @@
...
+     * If the user is in a VoiceChannel, it's the deaf state of the member.
+     * @type {?boolean}
+     */
+    this.deaf = data.deaf;
+    /**
+     * If the user is in a VoiceChannel, it's the mute state of the member.
+     * @type {?boolean}
+     */
+    this.mute = data.mute;
+    /**
```
kyranet, line 57:
```suggestion
    this.mute = data.mute;

    /**
```

src/structures/WidgetMember.js
```diff
@@ -0,0 +1,90 @@
...
+     * The status of the member.
+     * @type {PresenceStatus}
+     */
+    this.status = data.status;
+    /**
+     * If the user is in a VoiceChannel, it's the deaf state of the member.
+     * @type {?boolean}
+     */
+    this.deaf = data.deaf;
+    /**
```
kyranet, line 52:
```suggestion
    this.deaf = data.deaf;

    /**
```

src/structures/WidgetMember.js
```diff
@@ -0,0 +1,90 @@
...
+     * The avatar of the member.
+     * @type {?string}
+     */
+    this.avatar = data.avatar;
+    /**
+     * The status of the member.
+     * @type {PresenceStatus}
+     */
+    this.status = data.status;
+    /**
```
kyranet, line 47:
```suggestion
    this.status = data.status;

    /**
```

src/structures/WidgetMember.js
```diff
@@ -0,0 +1,90 @@
...
+   * @typedef {Object} WidgetActivity
+   * @property {string} name The name of the activity
+   */
+  /**
+   * @param {Client} client - The instantiating client
+   * @param {Object} data - The raw data
+   */
+  constructor(client, data) {
+    super(client);
+    /**
```
kyranet, line 20:
```suggestion
    super(client);

    /**
```

src/structures/WidgetMember.js
```diff
@@ -0,0 +1,90 @@
...
+ */
+class WidgetMember extends Base {
+  /**
+   * Activity sent in a {@link WidgetMember}.
+   * @typedef {Object} WidgetActivity
+   * @property {string} name The name of the activity
+   */
+  /**
+   * @param {Client} client - The instantiating client
+   * @param {Object} data - The raw data
```
kyranet, line 16:
```suggestion
   * @param {Client} client The instantiating client
   * @param {Object} data The raw data
```

src/structures/WidgetMember.js
```diff
@@ -0,0 +1,90 @@
...
+/**
+ * Represents a WidgetMember.
+ */
+class WidgetMember extends Base {
+  /**
+   * Activity sent in a {@link WidgetMember}.
+   * @typedef {Object} WidgetActivity
+   * @property {string} name The name of the activity
+   */
+  /**
```
kyranet, line 14:
```suggestion
   */

  /**
```

src/structures/Widget.js
```diff
@@ -0,0 +1,80 @@
...
+class Widget extends Base {
+  /**
+   * @param {Client} client - The instantiating client
+   * @param {Object} data - The raw data
+   */
+  constructor(client, data) {
+    super(client);
+    this._patch(data);
+  }
+  /**
```
kyranet, line 19:
```suggestion
  }

  /**
```

kyranet (changes requested):
Just some formatting things for consistency, otherwise LGTM

kyranet (approved):

SpaceEEC (approved):

src/structures/WidgetMember.js
```diff
@@ -0,0 +1,102 @@
...
+    this.suppress = data.suppress;
+
+    /**
+     * If the user is in a VoiceChannel, it's the id of the VoiceChannel of the member.
+     * @type {?Snowflake}
+     */
+    this.channelID = data.channel_id;
+
+    /**
+     * The avatarUrl of the member.
```
vladfrangu, line 89: Correct the comment, separate avatar and URL

src/structures/WidgetMember.js
```diff
@@ -0,0 +1,102 @@
...
+    this.selfMute = data.self_mute;
+
+    /**
+     * If the user is in a VoiceChannel, it's the supress state of the member.
+     * @type {?boolean}
+     */
+    this.suppress = data.suppress;
+
+    /**
+     * If the user is in a VoiceChannel, it's the id of the VoiceChannel of the member.
```
vladfrangu, line 83: `The id of the voice channel the member is in, if any`

src/structures/WidgetMember.js
```diff
@@ -0,0 +1,102 @@
...
+    this.avatar = data.avatar;
+
+    /**
+     * The status of the member.
+     * @type {PresenceStatus}
+     */
+    this.status = data.status;
+
+    /**
+     * If the user is in a VoiceChannel, it's the deaf state of the member.
```
vladfrangu, line 53: `If the member is server deafened` assuming this is a server deaf

src/structures/WidgetMember.js
```diff
@@ -0,0 +1,102 @@
...
+     * @type {?boolean}
+     */
+    this.deaf = data.deaf;
+
+    /**
+     * If the user is in a VoiceChannel, it's the mute state of the member.
```
vladfrangu, line 59: `If the member is server muted`

src/structures/WidgetMember.js
```diff
@@ -0,0 +1,102 @@
...
+     * @type {?boolean}
+     */
+    this.mute = data.mute;
+
+    /**
+     * If the user is in a VoiceChannel, it's the selfDeaf state of the member.
```
vladfrangu, line 65: `If the member is self deafened`

src/structures/WidgetMember.js
```diff
@@ -0,0 +1,102 @@
...
+     * @type {?boolean}
+     */
+    this.selfDeaf = data.self_deaf;
+
+    /**
+     * If the user is in a VoiceChannel, it's the selfMute state of the member.
```
vladfrangu, line 71: `If the member is self muted`

src/structures/WidgetMember.js
```diff
@@ -0,0 +1,102 @@
...
+     * @type {?boolean}
+     */
+    this.selfMute = data.self_mute;
+
+    /**
+     * If the user is in a VoiceChannel, it's the supress state of the member.
```
vladfrangu, line 77: `If the member is suppressed`

vladfrangu (changes requested):

src/structures/WidgetMember.js
```diff
@@ -0,0 +1,102 @@
...
+  /**
+   * @param {Client} client The instantiating client
+   * @param {Object} data The raw data
+   */
+  constructor(client, data) {
+    super(client);
+
+    /**
+     * The id of the user. It's an arbitrary number.
+     * @type {string}
```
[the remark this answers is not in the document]
ThumusLive: Because discord return a string with a number in it, but i'll fix it with a parseInt

src/structures/Widget.js
```diff
@@ -0,0 +1,81 @@
...
+     */
+    this.channels = new Collection();
+    for (const channel of data.channels) {
+      this.channels.set(channel.id, channel);
+    }
+
+    /**
+     * The list of members in the guild.
+     * These strings are just arbitrary numbers, they aren't Snowflakes.
+     * @type {Collection<string, WidgetMember>}
```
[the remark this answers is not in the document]
ThumusLive: Fixed too

src/structures/WidgetMember.js
```diff
@@ -21,9 +21,9 @@ class WidgetMember extends Base {
 
     /**
      * The id of the user. It's an arbitrary number.
-     * @type {string}
+     * @type {number}
      */
-    this.id = data.id;
+    this.id = parseInt(data.id);
```
vladfrangu, line 26: You can keep this as a string! Sorry, was mostly a question about it 👍

src/structures/Widget.js
```diff
@@ -57,7 +57,7 @@ class Widget extends Base {
      */
     this.members = new Collection();
     for (const member of data.members) {
-      this.members.set(member.id, new WidgetMember(this.client, member));
+      this.members.set(parseInt(member.id), new WidgetMember(this.client, member));
```
vladfrangu, line 60: You can keep this as a string! Sorry, was mostly a question about it 👍

vladfrangu (changes requested):

src/structures/WidgetMember.js
ThumusLive: Oh my bad ! Don't worry x) I'll revert it

src/structures/Widget.js
ThumusLive: Same here x)

ThumusLive:
Before commit it, do we need add a Guild#fetchWidget like I said upper ?

src/structures/Widget.js
```diff
@@ -0,0 +1,81 @@
...
+    }
+
+    /**
+     * The list of members in the guild.
+     * These strings are just arbitrary numbers, they aren't Snowflakes.
+     * @type {Collection<string, WidgetMember>}
+     */
+    this.members = new Collection();
+    for (const member of data.members) {
+      this.members.set(parseInt(member.id), new WidgetMember(this.client, member));
```
vladfrangu, line 60: Forgot to remove the parseInt

src/structures/WidgetMember.js
vladfrangu: Drop the parseInt, update JSdoc too to mark it as a string

typings/index.d.ts
```diff
@@ -2029,6 +2030,34 @@ declare module 'discord.js' {
...
+    public id: Snowflake;
+    public instantInvite?: string;
+    public channels: Collection<Snowflake, WidgetChannel>;
+    public members: Collection<string, WidgetMember>;
+    public presenceCount: number;
+  }
+
+  export class WidgetMember extends Base {
+    constructor(client: Client, data: object);
+    public id: string;
```
vladfrangu, line 2046: Technically also a Snowflake

vladfrangu (changes requested):

typings/index.d.ts
ThumusLive: Not really, a Snowflake is more a date and an user who nerver change
Here, the id is just a string with an arbitrary number in it and it can change

iCrawl (approved):

PR #5620, opened 2021-05-13
fix(ApplicationCommand): remove undefined values from transformOption

cmd430:
Removes undefined keys returned by `<ApplicationCommand>.transformOption`

This allows for comparisons against what Discord returns and what you have on disk for a command for example using `require('util').isDeepStrictEqual` so you can determine if a command has changed on disk since the last check

I also believe this makes more sense than having some properties explicitly set as undefined and belive it follows suit with the rest of the library

### DIFF
--- a/src/structures/ApplicationCommand.js
+++ b/src/structures/ApplicationCommand.js
@@ -204,7 +204,7 @@ class ApplicationCommand extends Base {
    * @private
    */
   static transformOption(option, received) {
-    return {
+    const transformOption = {
       type: typeof option.type === 'number' && !received ? option.type : ApplicationCommandOptionTypes[option.type],
       name: option.name,
       description: option.description,
@@ -212,6 +212,12 @@ class ApplicationCommand extends Base {
       choices: option.choices,
       options: option.options?.map(o => this.transformOption(o)),
     };
+
+    for (const prop in transformOption) {
+      if (transformOption[prop] === undefined) delete transformOption[prop];
+    }
+
+    return transformOption;
   }
 }
 


kyranet:
You can just do `JSON.parse(JSON.stringify())` so data is serialized correctly (that also strips `undefined` values away), this change comes with a very large performance impact just for testing benefits, even though it doesn't have any effect in the end result.

Also, don't testing frameworks have an option to ignore undefined values? 🤔

cmd430:
[util.isDeepStrictEqual](https://nodejs.org/api/util.html#util_util_isdeepstrictequal_val1_val2) is a node built in and fails if the key is explicitly set as undefined in one of the objects it is comparing

[This method is also significantly faster than `JSON.parse(JSON.stringify())`](https://jsben.ch/64VdO) and I do not believe the performance of this is significant enough to matter

more so `JSON.parse(JSON.stringify())` is just bad practice that can lead to issues

kyranet:
> [util.isDeepStrictEqual](https://nodejs.org/api/util.html#util_util_isdeepstrictequal_val1_val2) is a node built in and fails if the key is explicitly set as undefined in one of the objects it is comparing

Alternative, either set those keys as `undefined`, or make a tool that strips those in the unit tests.

> [This method is also significantly faster than `JSON.parse(JSON.stringify())`](https://jsben.ch/64VdO) and I do not believe the performance of this is significant enough to matter
> 
> more so `JSON.parse(JSON.stringify())` is just bad practice that can lead to issues

That's not to be used in this library, but in your test suite. We can't just strip `undefined` in the entire library, that's a can of worms you're about to open that would impact every bot's performance for no benefit.

Also, I believe some test suites such as Jest have methods to check whether two objects are equal while ignoring undefined values (aka not a strict check).

ckohen:
If you wanted to add this to the library, I would propose to add an `#isEqual()` method as some other classes do (granted this is a slightly different case because you would be comparing an instantiated class and an object instead of two instantiated classes). That way you can pass the object that you would pass to any of the creator / editor functions and get an equal

cmd430:
> that's not to be used in this library, but in your test suite.

This is not for a test suite this is for checking if i need to edit a command or not when starting the bot (using slash commands defined in files) 

> And unlike your loop, it also works in nested objects.

This loop works fine in nested objects because of the recursive nature of this function

> If you wanted to add this to the library, I would propose to instead add an #isEqual() method 

I will have a look at adding it this way then

kyranet:
> If you wanted to add this to the library, I would propose to instead add an `#isEqual()` method as some other classes do (granted this is a slightly different case because you would be comparing an instantiated class and an object instead of two instantiated classes). That way you can pass the object that you would pass to any of the creator / editor functions and get an equal

That'll be a little complicated, though, because as you said, they're plain objects. However, this is a more correct approach, although not sure if they will be accepted (specially as non-strict equality), please ask in `#library-discussion` from the [discord.js official server](https://discord.gg/djs) beforehand.

By the way @cmd430, you might want to change your strategy on how you check for object equality, because in future versions of discord.js, we'll most likely send a lot more `undefined`s in other places to prevent [object shape changes](https://mathiasbynens.be/notes/shapes-ics) and de-optimizations in V8's inline cache, which would potentially help us making discord.js's internals either monomorphic or polymorphic (which come with huge performance benefits) over megamorphic (which are flagged to not be optimized). Calling `delete` also does effectively counter-act those performance benefits and bring worse performance.

kyranet:
Re-posting from our conversation in the Discord server:

> We concluded a structure called `ApplicationCommandOption` (as opposed to a plain object with a typedef) would be a sensible way to approach this

> And it'd also refactor away `ApplicationCommand.transformOption` (they'd be transformed inside `ApplicationCommandOption`'s constructor) as well as allow us to add `isEqual` in them for fast deep comparison

> The idea behind checking for equals recursively is so we can dynamically build a framework which updates the registered application commands on Discord automatically, without us needing to write migrations or pushing all commands in every log in, which can be rather inefficient

[Source](https://discord.com/channels/222078108977594368/682166281826598932/842332937528803358)

---

This is what @cmd430 intended to achieve with this PR, and it's something I approve.

kyranet:
Closing as there was no activity in this PR. Please open a new one for the new set of changes once you're available :)

kyranet:
Status update, #6414 has been opened with the discussed changes 👍🏼

PR #5624, opened 2021-05-13
chore(Colors): use new Discord colors

Dev-CasperTheGhost:
**Please describe the changes this PR makes and why it should be merged:**

This PR updates the `Colors` constant to use the new [Discord colors](https://cdn.discordapp.com/attachments/613427779723067411/842438442880073748/2021_Palette_.png). 

**Status and versioning classification:**

- Code changes have been tested against the Discord API, or there are no code changes
- I know how to update typings and have done so, or typings don't need updating

<details>
	<summary><strong>Edit: example of color in an embed</strong></summary>

image

</details>

PR #5626, opened 2021-05-14
changed some colors

CertifiedRice:
This Pull request updates the colors in constants.js to make it look better. I have selected some colors that look a bit not so nice. So I changed them.


monbrey:
#5624 will change these colours to meet Discord's new branding guidelines.

If you personally don't like the colours, you have no obligation to use them and can pass your own hexcodes to embeds/roles/wherever else.

CertifiedRice:
ok acceptable

PR #5633, opened 2021-05-14
Session restoring during boot/login

Zhincore:
This PR adds parameter `sessionID` to `Client.login` allowing Discord.js to restore previous session(s).

Snippets showing my use of this feature (supports one shard, but the PR itself should be able to handle multiple):
```typescript
client.once("ready", () => {
  fs.writeFileSync("cache/sessionID", client.ws.shards.get(0)["sessionID"]);
});

let sessionID: string;
if (fs.existsSync("cache/sessionID",)) {
  sessionID = fs.readFileSync("cache/sessionID").toString();
}
client.login(process.env.DISCORD_TOKEN, sessionID);
```
This feature is really useful mainly during development, when you're restarting the bot often, or most probably if you have for example nodemon restart the bot for you, you might eventually hit the IDENTIFY rate limit, this doesn't happen with session restoring.
Passing invalid/expired sessionID(s) slows down the boot, partially because d.js waits for 5 seconds before creating new session, but passing `null` or not passing `sessionID` parameter at all has no impact on logging-in.

### DIFF
--- a/src/client/Client.js
+++ b/src/client/Client.js
@@ -201,7 +201,7 @@ class Client extends BaseClient {
    * @param {string|string[]} [sessionID] ID(s) of previous session(s) to restory
    * @returns {Promise<string>} Token of the account used
    * @example
-   * client.login('my token');
+   * client.login('my token', 'aaaaaaaaaa');
    */
   async login(token = this.token, sessionID) {
     if (!token || typeof token !== 'string') throw new Error('TOKEN_INVALID');
--- a/src/client/Client.js
+++ b/src/client/Client.js
@@ -201,7 +201,7 @@ class Client extends BaseClient {
    * @param {string|string[]} [sessionID] ID(s) of previous session(s) to restory
    * @returns {Promise<string>} Token of the account used
    * @example
-   * client.login('my token', 'aaaaaaaaaa');
+   * client.login('my token');
    */
   async login(token = this.token, sessionID) {
     if (!token || typeof token !== 'string') throw new Error('TOKEN_INVALID');
--- a/src/client/Client.js
+++ b/src/client/Client.js
@@ -198,11 +198,12 @@ class Client extends BaseClient {
   /**
    * Logs the client in, establishing a websocket connection to Discord.
    * @param {string} [token=this.token] Token of the account to log in with
+   * @param {string|string[]} [sessionID] ID(s) of previous session(s) to restory
    * @returns {Promise<string>} Token of the account used
    * @example
    * client.login('my token');
    */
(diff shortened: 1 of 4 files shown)


src/client/Client.js
```diff
@@ -198,11 +198,12 @@ class Client extends BaseClient {
   /**
    * Logs the client in, establishing a websocket connection to Discord.
    * @param {string} [token=this.token] Token of the account to log in with
+   * @param {string|string[]} [sessionID] ID(s) of previous session(s) to restory
```
ImRodry, line 201:
```suggestion
   * @param {string|string[]} [sessionID] ID(s) of previous session(s) to restore
```

typings/index.d.ts
```diff
@@ -218,7 +218,7 @@ declare module 'discord.js' {
     public fetchVoiceRegions(): Promise<Collection<string, VoiceRegion>>;
     public fetchWebhook(id: Snowflake, token?: string): Promise<Webhook>;
     public generateInvite(options?: InviteGenerationOptions | PermissionResolvable): Promise<string>;
-    public login(token?: string): Promise<string>;
+    public login(token?: string, sessionID?: string|string[]): Promise<string>;
```
ImRodry, line 221: Wouldn't you just pass one sessionID instead of an array of session IDs to login?

Zhincore: From what I understand, each shard has it's own session. So passing an array allows you to assign each of your shards with a session ID

src/client/websocket/WebSocketShard.js
```diff
@@ -19,7 +19,7 @@ if (!browser) {
  * Represents a Shard's WebSocket connection
  */
 class WebSocketShard extends EventEmitter {
-  constructor(manager, id) {
+  constructor(manager, id, sessionIDs) {
```
Keke27210, line 22:
```suggestion
  constructor(manager, id, sessionID) {
```
That parameter only holds a single session ID.

src/client/websocket/WebSocketShard.js
```diff
@@ -59,7 +59,7 @@ class WebSocketShard extends EventEmitter {
      * @type {?string}
      * @private
      */
-    this.sessionID = null;
+    this.sessionID = sessionIDs;
```
Keke27210, line 62:
```suggestion
    this.sessionID = sessionID ?? null;
```
See above + should fallback to `null` if none was provided.

Zhincore:
D.js doesn't fetch required data (like client.user) when resuming instead of normal login, this has to be addressed somehow

PR #5646, opened 2021-05-19
feat(Message): message reaction listener

khiromoshito:
This adds `Message#catchReactions` method to listen to individual message reactions. Compared to `Message#awaitReactions`, this only needs one callback function (with similar parameters as the `filter` param) as argument, which is fired *every time* a reaction is added to the message.

* Code changes have been tested against the Discord API, or there are no code changes
* I know how to update typings and have done so
* This PR changes the library's interface (methods or parameters added)

### DIFF
--- a/src/structures/Message.js
+++ b/src/structures/Message.js
@@ -407,6 +407,36 @@ class Message extends Base {
     });
   }
 
+  /** Listens to individual user reactions. In contrast to one-time callback by awaitReactions,
+   *  this provides a callback fired every time a new reaction is added to a message.
+   * 
+   *  @param {Function} callback
+   *  @example
+   *  // Attaches a reaction listener to a message
+   *  message.catchReactions((reaction, user)=>{
+   *    console.log(`${user.tag} reacted ${reaction.emoji.name}`);
+   *  });
+   */
+  catchReactions(callback) {
+
+    /** Since the filter is called twice for every reaction,
+    * a counter will be used to disregard odd indexes
+    * which represent a duplicate from the preceding even index
+    */
+    let index = 0;
+
+    // This utilizes awaitReactions and simplifies it into a single callback
+    this.awaitReactions(
+      // awaitReaction's filter function is called every time a reaction
+      // is checked for collection. This will be attached to the callback.
+      (reaction, user)=>{
+        if(index%2==0) callback(reaction, user);
+        index++;
+
+        return true;
+    }, {max: Infinity});
+  }
+
   /**
    * Whether the message is editable by the client user
    * @type {boolean}
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -1149,6 +1149,9 @@ declare module 'discord.js' {
       filter: CollectorFilter<[MessageReaction, User]>,
       options?: AwaitReactionsOptions,
     ): Promise<Collection<Snowflake, MessageReaction>>;
+    public catchReactions(
+      filter: MessageReactionsCallback<[MessageReaction, User]>
+    ): void;
     public createReactionCollector(
       filter: CollectorFilter<[MessageReaction, User]>,
       options?: ReactionCollectorOptions,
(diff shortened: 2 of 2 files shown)


vBread:
Forgive my ignorance if I'm missing something, but I fail to see how this is any different from [`awaitReactions`](https://discord.js.org/#/docs/main/stable/class/Message?scrollTo=awaitReactions)/[`createReactionCollector`](https://discord.js.org/#/docs/main/stable/class/Message?scrollTo=createReactionCollector).

khiromoshito:
> Forgive my ignorance if I'm missing something, but I fail to see how this is any different from [`awaitReactions`](https://discord.js.org/#/docs/main/stable/class/Message?scrollTo=awaitReactions)/[`createReactionCollector`](https://discord.js.org/#/docs/main/stable/class/Message?scrollTo=createReactionCollector).

Reaction collectors, as what I have observed, may collect reactions, but will not callback the collection unless it ends (either by hitting max or ending timeout).
 For example, implementing Message#awaitReactions with a max of 10 reactions will not fire the promise callback unless the collection already has 10 reactions. If simply applying a max of 1, this may instantly callback a single reaction, but since the collection has ended, the developer has to re-implement the awaitReactions method to listen for reactions again.

The method in this pull request provides a simpler interface for listening to individual reactions to a message, without anticipations and much of the boilerplate.

tipakA:
What's so different here than using a `ReactionCollector`? Not only do you get an event the moment something gets collected, you can also decide to stop collecting at any point with one method call, which in turn returns `collected` collection. In the `collect` event, you are free to do anything you want with the collected item.

What's the intended use here?

Collectors are useful when you need a (probably) variable amount of items to collect, and need to act upon each collection individually, probably sending / editing messages upon each (like questionnaries or pagination);

`awaitReactions` / `awaitMessages` are useful when you just need a single or multiple answers, and only need the data when you hit a limit, be it time limit, or max items. (like a prompt for missing or invalid command parameter)

If you need to collect items for longer periods of time, there are always respective events, you can just compare IDs

khiromoshito:
This method is for handling reactions as individual actions, and not as a collection of reactions after a limit. There can be circumstances that this method can be of use, like `react-to-get-a-role` where it expects reactions by reaction as they are added (and not dependent of hitting a maximum nor timeout just to report all those collected reactions). There is indeed a way for this to be solved using the `awaitReactions`, but the aforementioned method is a simpler approach to that.

I also observed how this reactions thing has been a struggle for me and some of my discordjs colleagues on handling reactions. One used infinite loop to re-register reaction collection just to record every single reactions as an action of itself (with `awaitReactions` of max `1` to instantly callback this reaction). Others find it a boilerplate to listen to the callback concerning collections because of how long of a code it is to take to just record reactions individually. 

This new method does not invalidate `awaitReactions` and similar methods, but instead only provides a shorthand for the above-mentioned reasons.

ckohen:
It sounds like your use case would be better satisfied with `client.on('messageReactionAdd')`. Furthermore, reactions were never intended by discord to be used in this way, so there are things coming in the near future that complete obsolete this approach and any need for bots to listen for reactions in this manner.

monbrey:
I understand your use-case, but what you've done is wrap a callback around a Promisified event, when it seems what you really wanted in the first place was just the un-promisified event.

`Message#createReactionCollector` with no `max` option specified meets this use-case, and allows you to handle individual reactions with the `ReactionCollector.on("collect", callback)` event.

I don't see any need for yet another layer of abstraction beyond that.

khiromoshito:
> I understand your use-case, but what you've done is wrap a callback around a Promisified event, when it seems what you really wanted in the first place was just the un-promisified event.
> 
> `Message#createReactionCollector` with no `max` option specified meets this use-case, and allows you to handle individual reactions with the `ReactionCollector.on("collect", callback)` event.
> 
> I don't see any need for yet another layer of abstraction beyond that.

This is a lifesaver answer. Thank you for clearing things up. This pull request shall now be closed.

PR #5655, opened 2021-05-20
cleanup(Client): guard emitting GMU and PU on user updates

ckohen:

Attempt 2!

This is the continuation of #5161 in hopefully a much more accepted fashion.

GMU and PU are currently emitted regardless of the content of the update. Discord also send these events when the underlying user is updated. Since the library abstracts this away into `userUpdate` many PU and GMU events would contain the same exact data in both the old and new objects of their respective events. This simply guards against emitting the event when the objects are identical.

To note: GMU and PU are emitted by discord once for **every** guild the bot shares with a user normally, with this change, you will only get **1** `userUpdate` event, not 1 per guild.

I have not made any docs changes other than documenting `GuildMember#equals()` because the current docs still feel accurate. Feel free to suggest changes if you disagree!

### DIFF
--- a/src/client/actions/GuildMemberUpdate.js
+++ b/src/client/actions/GuildMemberUpdate.js
@@ -27,7 +27,7 @@ class GuildMemberUpdateAction extends Action {
          * @param {GuildMember} oldMember The member before the update
          * @param {GuildMember} newMember The member after the update
          */
-        if (shard.status === Status.READY) client.emit(Events.GUILD_MEMBER_UPDATE, old, member);
+        if (shard.status === Status.READY && !member.equals(old)) client.emit(Events.GUILD_MEMBER_UPDATE, old, member);
       } else {
         const newMember = guild.members.add(data);
         /**
--- a/src/client/actions/PresenceUpdate.js
+++ b/src/client/actions/PresenceUpdate.js
@@ -28,7 +28,7 @@ class PresenceUpdateAction extends Action {
       this.client.emit(Events.GUILD_MEMBER_AVAILABLE, member);
     }
     guild.presences.add(Object.assign(data, { guild }));
-    if (member && this.client.listenerCount(Events.PRESENCE_UPDATE)) {
+    if (member && this.client.listenerCount(Events.PRESENCE_UPDATE) && !member.presence.equals(oldPresence)) {
       /**
        * Emitted whenever a guild member's presence (e.g. status, activity) is changed.
        * @event Client#presenceUpdate
--- a/src/structures/GuildMember.js
+++ b/src/structures/GuildMember.js
@@ -347,6 +347,29 @@ class GuildMember extends Base {
   }
 
   /**
+   * Whether this guild member equals another guild member. It compares all properties, so for most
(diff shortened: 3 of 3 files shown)


src/client/actions/GuildMemberUpdate.js
```diff
@@ -27,7 +27,7 @@ class GuildMemberUpdateAction extends Action {
          * @param {GuildMember} oldMember The member before the update
          * @param {GuildMember} newMember The member after the update
          */
-        if (shard.status === Status.READY) client.emit(Events.GUILD_MEMBER_UPDATE, old, member);
+        if (!member.equals(old) && shard.status === Status.READY) client.emit(Events.GUILD_MEMBER_UPDATE, old, member);
```
kyranet, line 30: Swap the order of the checks, the last one is a lot faster and can discard a much larger chain of operations.

src/client/actions/PresenceUpdate.js
```diff
@@ -28,7 +28,7 @@ class PresenceUpdateAction extends Action {
       this.client.emit(Events.GUILD_MEMBER_AVAILABLE, member);
     }
     guild.presences.add(Object.assign(data, { guild }));
-    if (member && this.client.listenerCount(Events.PRESENCE_UPDATE)) {
+    if (member && !member.presence.equals(oldPresence) && this.client.listenerCount(Events.PRESENCE_UPDATE)) {
```
kyranet, line 31: Same here, place the equals check after the `listenerCount` one.

src/structures/GuildMember.js
```diff
@@ -346,6 +346,30 @@ class GuildMember extends Base {
     return this.guild.members.fetch({ user: this.id, cache: true, force });
   }
 
+  /**
+   * Whether this guild member equals another guild member. It compares all properties, so for most
+   * comparison it is advisiable to just compare `member.id === member2.id` as it is significantly faster
```
kyranet, line 351:
```suggestion
   * comparison it is advisable to just compare `member.id === member2.id` as it is significantly faster
```

src/structures/GuildMember.js
```diff
@@ -346,6 +346,30 @@ class GuildMember extends Base {
...
+   * and is often what most users need.
+   * @param {GuildMember} member the member to compare with
```
kyranet, line 353:
```suggestion
   * @param {GuildMember} member The member to compare with
```

src/structures/GuildMember.js
```diff
@@ -346,6 +346,30 @@ class GuildMember extends Base {
...
+   * @returns {boolean}
+   */
+  equals(member) {
+    return (
+      member &&
+      member instanceof this.constructor &&
```
kyranet, line 359: The truthy check is useless as the `instanceof` will always return `false` for nullish values.
```suggestion
      member instanceof this.constructor &&
```

kyranet (changes requested):

vladfrangu (approved):

kyranet (approved):

iCrawl (approved):

SpaceEEC (approved):

Issue #5449, opened 2021-03-31
Add client option to disable automatic queueing of requests if rate limit is hit and to reject instead


Vendicated:
**Is your feature request related to a problem? Please describe.**
My bot creates a lot of emotes, which eventually leads to hitting the emoji create rate limit. This rate limit is insanely long from experience (up to an hour) resulting in the `GuildEmojiManager.create` Promise to take an hour to resolve. This is super annoying and forces me to use `Promise.race` to time out after a while in order to not get stuck

**Describe the ideal solution**
- new Client Option disableRatelimitQueue, which takes an array of routes, so in my case `[ "emojis" ]`
- On disabled routes, requests will no longer be queued, instead reject the Promise with info on when the ratelimit ends

I could implement this if approved

**Describe alternatives you've considered**
As mentioned above I use Promise.race to time out after a while

**Additional context**
Related to https://github.com/discordjs/discord.js/issues/2885

tandpfun:
Another use case: #5473

doinkythederp:
Adding onto the `disableRatelimitQueue` suggestion, maybe it could take a filter function?
```js
new Client({ disableRatelimitQueue: function(info) { 
    if (isVeryImportant(info)) {
      return false; // Is critical, put on the queue!
    } else {
      return true; // Not critical, just cancel.
    }
  }
});
```

PR #5659, opened 2021-05-21
feat(Rest): optional ratelimit errors

Vendicated:
Adds new client option disableRateLimitQueue which takes either an array of routes or `(ratelimitData) => boolean`.
If a rate limit is hit, the RequestHandler first checks whether this rate limit should be thrown (route starts with one of the array values / function returns true) and if so throws an error instead of queueing the request

This closes https://github.com/discordjs/discord.js/issues/5449

### DIFF
--- a/src/client/Client.js
+++ b/src/client/Client.js
@@ -493,10 +493,10 @@ class Client extends BaseClient {
       throw new TypeError('CLIENT_INVALID_OPTION', 'retryLimit', 'a number');
     }
     if (
-      typeof options.disableRateLimitQueue !== 'undefined' &&
-      !(typeof options.disableRateLimitQueue === 'function' || Array.isArray(options.disableRateLimitQueue))
+      typeof options.rejectOnRateLimit !== 'undefined' &&
+      !(typeof options.rejectOnRateLimit === 'function' || Array.isArray(options.rejectOnRateLimit))
     ) {
-      throw new TypeError('CLIENT_INVALID_OPTION', 'disableRateLimitQueue', 'an array or a function');
+      throw new TypeError('CLIENT_INVALID_OPTION', 'rejectOnRateLimit', 'an array or a function');
     }
   }
 }
--- a/src/client/Client.js
+++ b/src/client/Client.js
@@ -492,6 +492,12 @@ class Client extends BaseClient {
     if (typeof options.retryLimit !== 'number' || isNaN(options.retryLimit)) {
       throw new TypeError('CLIENT_INVALID_OPTION', 'retryLimit', 'a number');
     }
+    if (
+      typeof options.disableRateLimitQueue !== 'undefined' &&
+      !(typeof options.disableRateLimitQueue === 'function' || Array.isArray(options.disableRateLimitQueue))
+    ) {
+      throw new TypeError('CLIENT_INVALID_OPTION', 'disableRateLimitQueue', 'an array or a function');
+    }
   }
 }
 
--- a/src/rest/RequestHandler.js
+++ b/src/rest/RequestHandler.js
@@ -83,22 +83,22 @@ class RequestHandler {
    */
   async onRateLimit(request, limit, timeout, isGlobal) {
     const { options } = this.manager.client;
-    if (options.disableRateLimitQueue) {
-      const rateLimitData = {
-        timeout,
-        limit,
-        method: request.method,
-        path: request.path,
-        route: request.route,
-        global: isGlobal,
-      };
-      const shouldThrow =
-        typeof options.disableRateLimitQueue === 'function'
-          ? await options.disableRateLimitQueue(rateLimitData)
-          : options.disableRateLimitQueue.some(route => rateLimitData.route.startsWith(route.toLowerCase()));
-      if (shouldThrow) {
-        throw new RateLimitError(rateLimitData);
-      }
+    if (!options.rejectOnRateLimit) return;
+
+    const rateLimitData = {
+      timeout,
+      limit,
+      method: request.method,
+      path: request.path,
+      route: request.route,
+      global: isGlobal,
+    };
+    const shouldThrow =
+      typeof options.rejectOnRateLimit === 'function'
+        ? await options.rejectOnRateLimit(rateLimitData)
+        : options.rejectOnRateLimit.some(route => rateLimitData.route.startsWith(route.toLowerCase()));
+    if (shouldThrow) {
+      throw new RateLimitError(rateLimitData);
     }
   }
 
--- a/src/rest/RequestHandler.js
+++ b/src/rest/RequestHandler.js
@@ -143,7 +143,7 @@ class RequestHandler {
         });
       }
 
-      // Determine whether RateLimitError should be thrown
+      // Determine whether a RateLimitError should be thrown
       await this.onRateLimit(request, limit, timeout, isGlobal); // eslint-disable-line no-await-in-loop
 
       if (isGlobal) {
--- a/src/rest/RequestHandler.js
+++ b/src/rest/RequestHandler.js
@@ -143,9 +143,6 @@ class RequestHandler {
         });
       }
 
-      // Determine whether a RateLimitError should be thrown
-      await this.onRateLimit(request, limit, timeout, isGlobal); // eslint-disable-line no-await-in-loop
-
       if (isGlobal) {
         // If this is the first task to reach the global timeout, set the global delay
         if (!this.manager.globalDelay) {
@@ -156,6 +153,9 @@ class RequestHandler {
       } else {
         delayPromise = Util.delayFor(timeout);
       }
+
+      // Determine whether a RateLimitError should be thrown
+      await this.onRateLimit(request, limit, timeout, isGlobal); // eslint-disable-line no-await-in-loop
 
       // Wait for the timeout to expire in order to avoid an actual 429
       await delayPromise; // eslint-disable-line no-await-in-loop
--- a/src/rest/RequestHandler.js
+++ b/src/rest/RequestHandler.js
@@ -95,7 +95,7 @@ class RequestHandler {
       const shouldThrow =
         typeof options.disableRateLimitQueue === 'function'
           ? await options.disableRateLimitQueue(rateLimitData)
-          : options.disableRateLimitQueue.some(route => rateLimitData.route.startsWith(`/${route}/`));
+          : options.disableRateLimitQueue.some(route => rateLimitData.route.startsWith(route.toLowerCase()));
       if (shouldThrow) {
         throw new RateLimitError(rateLimitData);
       }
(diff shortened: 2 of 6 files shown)


src/rest/RequestHandler.js
```diff
@@ -77,6 +78,28 @@ class RequestHandler {
...
+    if (this.manager.client.options.disableRateLimitQueue) {
+      const rateLimitData = {
+        timeout,
+        limit,
+        method: request.method,
+        path: request.path,
+        route: request.route,
+        global: isGlobal,
+      };
+      const shouldThrow =
```
NotSugden, line 91: can't do a multiline suggestion for whatever reason

```js
const { options } = this manager.client;
const shouldThrow = 
  (typeof options.disableRateLimitQueue === 'function' && options.disableRateLimitQueue(rateLimitData)) ||
  options.disableRateLimitQueue.some(...);
```

typings/index.d.ts
```diff
@@ -3489,6 +3488,13 @@ declare module 'discord.js' {
     global: boolean;
   }
 
+  // tslint:disable-next-line:no-empty-interface - Merge RateLimitData interface to not have to type it again
+  interface RateLimitError extends RateLimitData {}
+  export class RateLimitError extends Error {
```
NotSugden, line 3493: this is a class so it has to go up, i think theres a better way to do this too, `extends Constructable<Error & RateLimitData>` afaik

NotSugden (commented):
the onRateLimit function should also be added to typings

typings/index.d.ts
Vendicated: I wasn't sure where to put this, so I just put it below the existing RateLimitData. what would be a more appropriate place for it?

Vendicated:
I have no idea how to look at the docs in my browser so they are untested

typings/index.d.ts
Vendicated: Moved it below HttpError now. Constructable indeed works but then it'd be pretty awkward syntax
```ts
export type RateLimitError = Constructable<Error & RateLimitData & { 
  constructor(data: RateLimitData);
  name: 'RateLimitError';
}>
    ```

File: typings/index.d.ts [hunk 3]
```diff
@@ -2659,6 +2664,7 @@ declare module 'discord.js' {
     intents: BitFieldResolvable<IntentsString, number>;
     ws?: WebSocketOptions;
     http?: HTTPOptions;
+    disableRateLimitQueue?: string[] | ((data: RateLimitData) => boolean | Promise<boolean>);
```
doinkythederp, line 2667: It might be helpful to only allow certain values in the array - people wouldn't have to leave their editor to know what could be entered here. It could also help with small mistakes in the value entered (e.g. "guild" vs. "guilds").

src/rest/RequestHandler.js
```diff
@@ -100,6 +128,9 @@ class RequestHandler {
         // Set the variables based on the route-specific rate limit
         limit = this.limit;
         timeout = this.reset + this.manager.client.options.restTimeOffset - Date.now();
+
+        await this.onRateLimit(request, limit, timeout, isGlobal); // eslint-disable-line no-await-in-loop
```
vladfrangu, line 132: Could you move these two calls after the event emitting? So, after that if check, run this?

src/rest/RequestHandler.js
```diff
@@ -223,8 +254,22 @@ class RequestHandler {
     if (res.status >= 400 && res.status < 500) {
       // Handle ratelimited requests
       if (res.status === 429) {
-        // A ratelimit was hit - this should never happen
+        // A ratelimit was hit - this should only happen if client is already rate limited when connecting
```
vladfrangu, line 257: Changing this comment is not accurate, 429s should never happen. At MOST I'd say changing it to `this could happen`...

src/util/Constants.js
```diff
@@ -3,6 +3,24 @@
 const Package = (exports.Package = require('../../package.json'));
 const { Error, RangeError } = require('../errors');
 
+/**
+ * Rate limit data
+ * @typedef {Object} RateLimitData
+ * @property {number} timeout Time until this rate limit ends, in ms
+ * @property {number} limit The maximum amount of requests of this end point
```
vladfrangu, line 10:
```suggestion
 * @property {number} limit The maximum amount of requests of this endpoint
```

src/rest/RequestHandler.js
```diff
@@ -77,6 +78,30 @@ class RequestHandler {
...
+        limit,
+        method: request.method,
+        path: request.path,
+        route: request.route,
+        global: isGlobal,
+      };
+      const shouldThrow =
+        typeof options.disableRateLimitQueue === 'function'
+          ? await options.disableRateLimitQueue(rateLimitData)
+          : options.disableRateLimitQueue.some(route => rateLimitData.route.startsWith(`/${route}/`));
```
vladfrangu, line 98: So, the intended use case is specifying, say `channels` or `channels/:id` to the option? If so, hard coding it to be `/${route}/` is very limiting. Should simply be a `startsWith(route.toLowerCase())` (since routes are lowercased), to let us be very explicit about it (say you want to throw on `/gateway/bot` (also, the last `/` is invalid for all routes)

vladfrangu (dismissed):

src/rest/RequestHandler.js
Vendicated: really? I reached this very point when the client was already rate limited when connecting so I figured this was the reason one might get there

vladfrangu: The only ratelimit you *could* get is on `/gateway/bot` when identifying too fast

typings/index.d.ts
Vendicated: there are a loooot of possible routes, so I don't think this would be viable and would end up being more limiting than helpful. it would also not allow including ids

src/rest/RequestHandler.js
Vendicated: Well I do hit 429s. This is with the current master branch

image

code:
```js
'use strict';

const Discord = require('..');

  intents: Discord.Intents.NON_PRIVILEGED,
});

client.once('ready', async () => {
  const chan = client.channels.cache.get('800884361733865522');
  while (true) {
    console.log('Editing channel...');
    await chan.edit({ name: Math.random().toString(36) });
  }
});

client.login('TOKEN'');
```

vladfrangu: Wait, hold on, you're telling me that on master, this code causes it to reply with a node-fetch Response? Are you sure?

Vendicated: no, I added `console.log(res)` right after the `if (res.status === 429) {` (the if statement this thread is under), that is why it's logging the node fetch response. The rest of the djs code is the current master branch, the console.log is the only addition I made

vladfrangu: Oh, that error is not unexpected sadly. Discord has a very strict sub-limit for channel name and topic changes, which can explain why you hit a 429 so fast. :(

Vendicated: so 429s should generally not happen, just on some specific routes that have these strict date limits? Is there not a way to catch these too and avoid rats limits or are the headers inaccurate here?

I will revert the comment, but it should be changed to describe these edge cases then

vladfrangu: These sub limits are:
- Not listed in the headers
- Separate from the headers
- Requests that don't trigger these sub limits can still be sent

Avoiding these is currently something we don't have, as we'd have to implement an edge case for each sub limit there is/will be. Implementation, if any, is to be done :D
Personally, I don't think we should change the comment to mention that, as it's invisible for most users anyways

src/util/Constants.js
```diff
@@ -34,6 +52,9 @@ const { Error, RangeError } = require('../errors');
  * (or 0 for never)
  * @property {number} [restGlobalRateLimit=0] How many requests to allow sending per second (0 for unlimited, 50 for
  * the standard global limit used by Discord)
+ * @property {string[]|RateLimitQueueFilter} [disableRateLimitQueue] Decides how rate limits (429) should be handled.
+ * If this option is an array containing the route of the request or a function returning true, a {@link RateLimitError}
```
SpaceEEC, line 56: I think this could also mention that array elements can be route prefixes too.
e.g. `/channels` matches `/channels/222197033908436994/messages`

src/util/Constants.js
```diff
@@ -34,6 +52,9 @@ const { Error, RangeError } = require('../errors');
  * (or 0 for never)
  * @property {number} [restGlobalRateLimit=0] How many requests to allow sending per second (0 for unlimited, 50 for
  * the standard global limit used by Discord)
+ * @property {string[]|RateLimitQueueFilter} [disableRateLimitQueue] Decides how rate limits (429) should be handled.
```
SpaceEEC, line 55: This isn't just invoked when a 429 was received, but also when djs preemptively throttles to avoid that.
This docstring doesn't really make this clear, which might be confusing.

src/rest/RequestHandler.js
```diff
@@ -125,6 +143,20 @@ class RequestHandler {
         });
       }
 
+      // Determine whether RateLimitError should be thrown
```
SpaceEEC, line 146:
```suggestion
      // Determine whether a RateLimitError should be thrown
```

SpaceEEC (changes requested):

Vendicated:
Is this better?

SpaceEEC (approved):
Yes

src/rest/RequestHandler.js
```diff
@@ -77,6 +78,30 @@ class RequestHandler {
     });
   }
 
+  /*
+   * Determines whether the request should be queued or whether a RateLimitError should be thrown
+   */
+  async onRateLimit(request, limit, timeout, isGlobal) {
+    const { options } = this.manager.client;
+    if (options.disableRateLimitQueue) {
```
vladfrangu, line 86: Extra super duper minor nitpick but... Can this be more like
```js
if (!options.disableRateLimitQueue) return null;
...rest of code here...
```

src/client/Client.js
```diff
@@ -492,6 +492,12 @@ class Client extends BaseClient {
     if (typeof options.retryLimit !== 'number' || isNaN(options.retryLimit)) {
       throw new TypeError('CLIENT_INVALID_OPTION', 'retryLimit', 'a number');
     }
+    if (
+      typeof options.disableRateLimitQueue !== 'undefined' &&
```
vladfrangu, line 496: I think this name is misleading, maybe something like `rejectOnRateLimit` or similar might be better?

vladfrangu (changes requested):

src/client/Client.js
Vendicated: yeah im not too happy with the current name either but i didnt come up with anything better. I think rejectOnRateLimit is nice, or maybe shouldRejectOnRateLimit?

iCrawl: `rejectOnRateLimit` is fine.

iCrawl:
@Vendicated if you could apply those changes requested we can move forward with merging this.

Vendicated:
sorry, I forgot 😅

kyranet (approved):

src/rest/RequestHandler.js
```diff
@@ -125,6 +143,20 @@ class RequestHandler {
         });
       }
 
+      // Determine whether a RateLimitError should be thrown
+      await this.onRateLimit(request, limit, timeout, isGlobal); // eslint-disable-line no-await-in-loop
```
vladfrangu, line 147: Here's a small issue. If we throw on a global ratelimit, before preventing every other request from going, we could hit lots of 429s real fast. This should be moved before the await to the delayPromise 👍

vladfrangu (changes requested):

src/rest/RequestHandler.js
Vendicated: i'm not sure if i understood correctly, you mean that line 153 (`this.manager.globalDelay = this.globalDelayFor(timeout);`) should go before this line?

vladfrangu: No, these two lines should go before current line 160 and 161

Vendicated: yeah thats what i meant. Good catch

PR #5664, opened 2021-05-23
Fix #5611 

### LINKED ISSUE #5611: The shardCount option does not allow the value 1
I am sharding my discord bot and while testing I only wanted to get the basics setup first so I was only going to use 1 shard.
In the client option I defined the option: shardCount: 1
If I try to run the code after that I get the error: TypeError [CLIENT_INVALID_OPTION]: The shardCount option must be a number greater than or equal to 1
I looked into the problem a bit and found that when the shardCount option is set to 1 and I place a breakpoint at line 437 (where the error originates from) in /client/Client.js the debugger says the shardCount option is NaN.
If I change the shardCount option to 2 it will show the expected result again.
I believe the problem lies somewhere in the if at line 52 of that file.
Before the if is run this.option.shardCount is 1 but after the if the value has been set to NaN.
The debuger also says that data.SHARD_COUNT is "auto".
So on line 54 this.options.shardCount is set to a number with the value of a string which returns NaN.

```js
// File: index.js
const { ShardingManager } = require("discord.js");
const managerOptions = {
    token: "",
    respawn: false
};
const manager = new ShardingManager("./bot.js", managerOptions);
const shard = manager.createShard(0);
shard.spawn();
```

```js
// File: bot.js
const discord = require("discord.js");
const client = new discord.Client({ shardCount: 1 });
client.login();
```

**Further details:**

- discord.js version: 12.5.3
- Node.js version: v12.18.4
- Operating system: Windows 10 (64 bit)
- Priority this issue should have – please be realistic and elaborate if possible: Low

**Relevant client options:**

- partials: none
- gateway intents: none
- other: shardCount is 1

DumbGameMaker:
This commit fixes #5611 by correctly defining options.shardCount to the input.

### DIFF
--- a/src/client/Client.js
+++ b/src/client/Client.js
@@ -456,6 +456,7 @@ class Client extends BaseClient {
       options.intents = Intents.resolve(options.intents);
     }
     if (typeof options.shardCount !== 'number' || isNaN(options.shardCount) || options.shardCount < 1) {
+      console.warn('WARN: shardCount is set to an invalid integer. ');
       options.shardCount = 1;
     }
     if (options.shards && !(options.shards === 'auto' || Array.isArray(options.shards))) {
--- a/src/client/Client.js
+++ b/src/client/Client.js
@@ -455,7 +455,7 @@ class Client extends BaseClient {
     } else {
       options.intents = Intents.resolve(options.intents);
     }
-    if (typeof options.shardCount !== 'number' || isNaN(options.shardCount) || options.shardCount <= 1) {
+    if (typeof options.shardCount !== 'number' || isNaN(options.shardCount) || options.shardCount < 1) {
       options.shardCount = 1;
     }
     if (options.shards && !(options.shards === 'auto' || Array.isArray(options.shards))) {
--- a/src/client/Client.js
+++ b/src/client/Client.js
@@ -455,8 +455,8 @@ class Client extends BaseClient {
     } else {
       options.intents = Intents.resolve(options.intents);
     }
-    if (typeof options.shardCount !== 'number' || isNaN(options.shardCount) || options.shardCount < 1) {
-      throw new TypeError('CLIENT_INVALID_OPTION', 'shardCount', 'a number greater than or equal to 1');
+    if (typeof options.shardCount !== 'number' || isNaN(options.shardCount) || options.shardCount <= 1) {
+      options.shardCount = 1;
     }
     if (options.shards && !(options.shards === 'auto' || Array.isArray(options.shards))) {
       throw new TypeError('CLIENT_INVALID_OPTION', 'shards', "'auto', a number or array of numbers");
--- a/src/client/Client.js
+++ b/src/client/Client.js
@@ -48,12 +48,14 @@ class Client extends BaseClient {
       }
     }
 
-    if (this.options.shardCount === DefaultOptions.shardCount) {
+    if (this.options.shardCount !== DefaultOptions.shardCount) {
       if ('SHARD_COUNT' in data) {
         this.options.shardCount = Number(data.SHARD_COUNT);
       } else if (Array.isArray(this.options.shards)) {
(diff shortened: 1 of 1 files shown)


MattIPv4 (changes requested):
👋 This does not seem to be a fix for the issue, but rather a workaround. (Is there even an issue? It seems like someone just incorrectly using the library).

The library absolutely should be throwing an error if the user provides an invalid value for this option, like other options do and like this one did before.

If this silently defaults to `1` then a user could very easily end up with a confusing situation in which they're passing an invalid value that they think to be valid, and the library not telling them.

DumbGameMaker:
> 👋 This does not seem to be a fix for the issue, but rather a workaround. (Is there even an issue? It seems like someone just incorrectly using the library).
> 
> The library absolutely should be throwing an error if the user provides an invalid value for this option, like other options do and like this one did before.
> 
> If this silently defaults to `1` then a user could very easily end up with a confusing situation in which they're passing an invalid value that they think to be valid, and the library not telling them.

would it be better if I had it give a warning saying that the input was invalid, and it would default to 1?

tipakA:
As it was said, this is more of a workaround than actual fix. Better idea would be to actually find the core issue as to why the value gets invalid instead of ignoring that and changing the value anyway.

DumbGameMaker:
> As it was said, this is more of a workaround than actual fix. Better idea would be to actually find the core issue as to why the value gets invalid instead of ignoring that and changing the value anyway.

ah ok, I'll do some more debugging

DumbGameMaker:
I think the issue is that it only sets options.shardCount IF it is equal to the default.

DumbGameMaker:
I edited this so that instead of working around the issue, it will correctly define options.shardCount to the input.

SpaceEEC:
I don't think these changes make sense.
Overwriting the shard count passed to the client seems counter-intuitive, if it was specified, the developer apparently knows better than the library.
To add onto this things like `new Client({ shards: [0,1,2] });` or something like my test code below stop working due to this.

Both would probably work again if you provided _anything_ as shard count, e.g. `new Client({ shards: [0,1,2], shardCount: Symbol.for("magic \u2728") });` .

---

My test code:
```js
const manager = new Discord.ShardingManager(path.join(__dirname, 'shard.js'), {
    totalShards: 3,
    token: 'token goes here',
});

manager.spawn();
```
```js
    intents: [Discord.Intents.FLAGS.GUILDS, Discord.Intents.FLAGS.GUILD_MESSAGES],
});

client.on('debug', console.log);

client.login();
```
The behavior I was expecting:
Shards [0,3], [1,3], [2,3] to be spawned.

The behavior I was witnessing:
Shards [0,1], [1,1], ??? were spawned, it stopped at the second because the configuration was invalid, I assume the third would have been [2,1].

This is working fine on [a73a5cf91498cf7b08cea85753ad481c194ec089] (master as of writing this).

---

I assume the fix for #5611 lies in `ShardingManager` or in `Shard`, since it seems like the recommended shard count is not fetch from Discord when going for that approach. (Why not just use `ShardingManager#spawn`?)
But I haven't really looked into that (yet).

kyranet:
Classifying the PR as `semver: major` as it changes the library's behaviour.

@DumbGameMaker if you update the PR to do what Space suggested, I'll lower it to `semver: patch`.

SliceCraft:
> I think the issue is that it only sets options.shardCount IF it is equal to the default.

That is correct, if I change the shardCount to 1 it will display that the shardCount has to be 1 or higher.
If you look at the issue I made you can see that what the code is trying to do is set the shardCount option to a number with the value 'auto' which causes the option to change to NaN.
I don't understand why data.SHARD_COUNT is 'auto', which is why I made a bug report.

kyranet:
Closing due to the lack of activity from the PR's author and the fact that this isn't really a bug (thanks Space for checking it thoroughly).

PR #5314, opened 2021-02-10
refactor(*): return the invalid element when erroring from Array

ckohen:
**Please describe the changes this PR makes and why it should be merged:**

This PR uses the new error from #5215 to provide more detailed errors when erroring from an Array or Collection.

⚠️ This has not been fully tested
⚠️ This changes some methods to use Promise.reject instead of throw or vice-versa

**Status and versioning classification:**

- I know how to update typings and have done so, or typings don't need updating
- This PR includes breaking changes (methods removed or renamed, parameters moved or removed)

PR #5666, opened 2021-05-24
fix(GuildEmojiRoleManager): bug in #remove

iShibi:

This PR fixes a bug in `GuildEmojiRoleManager#remove`, where it wasn't removing role(s) from the emoji. This bug was a result of an [oversight](https://github.com/ckohen/discord.js/blob/217ea9d162762fa18f4d225c7e99131b0eeaef24/src/managers/GuildEmojiRoleManager.js#L86) in #5314. Since, `#keyArray` is being used there on `this._roles` collection, the `role` is already an ID, doing `role.id` returns `undefined` which results in `newRoles` containing IDs of the roles the emoji already has. This means no filtration happened and Discord got role IDs that were already on the emoji. Therefore, the role a user wants to remove from the emoji doesn't get removed.

ℹ️: Even tho the fix could be done by just changing `role.id` to `role`, I did some refactoring by renaming some identifiers in the `#remove` method so that oversight like this doesn't happen again. I noticed that some other methods also need this refactorization in `GuildEmojiRoleManager` file. I didn't do it in order to keep the initial review of the PR easy. Let me know If I should do it or not in this PR.

### DIFF
--- a/src/managers/GuildEmojiRoleManager.js
+++ b/src/managers/GuildEmojiRoleManager.js
@@ -74,16 +74,16 @@ class GuildEmojiRoleManager {
   remove(roleOrRoles) {
     if (!Array.isArray(roleOrRoles) && !(roleOrRoles instanceof Collection)) roleOrRoles = [roleOrRoles];
 
-    const resolvedRoles = [];
+    const resolvedRoleIDs = [];
     for (const role of roleOrRoles.values()) {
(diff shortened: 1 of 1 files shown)


ckohen:
I was asked to leave the name resolvedRoles in the original PR, not 100% sure why, may have been a consistency thing. I'm all for the change though.

SpaceEEC (approved):

iCrawl (approved):

PR #5670, opened 2021-05-25
ShardClient

ValidMelvin:
Just changed something at the SharClientUtils because of an error that kept occuring.


Keke27210:
Not sure about how is this going to fix anything since `process.parentPort` is not a thing
What error made you do that?

SpaceEEC:
Since there is no explanation about why this is being changed or how this is solving what problem, as well as `parentPort` seemingly not being documented in [`process' documentation`](https://nodejs.org/api/process.html), I'm going to close this PR.

If you have further information about your issue and how this (or something else) is solving it, feel free to reopen this PR.

PR #5679, opened 2021-05-26
fix(ApplicationCommandOptionData): options property should be itself

SpaceEEC:

`ApplicationCommandOptionData`'s `options` property should be `ApplicationCommandOptionData` itself to allow the usage of the `ApplicationCommandOptionTypes` enum.

### DIFF
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -2471,7 +2471,7 @@ declare module 'discord.js' {
     description: string;
     required?: boolean;
     choices?: ApplicationCommandOptionChoice[];
-    options?: ApplicationCommandOption[];
+    options?: ApplicationCommandOptionData[];
   }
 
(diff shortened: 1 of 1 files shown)


typings/index.d.ts
```diff
@@ -2471,7 +2471,7 @@ declare module 'discord.js' {
     description: string;
     required?: boolean;
     choices?: ApplicationCommandOptionChoice[];
-    options?: ApplicationCommandOption[];
+    options?: ApplicationCommandOptionData[];
```
vaporox, line 2474:
```suggestion
    options?: this[];
```
This will keep the correct type for the extending `ApplicationCommandOption` interface

vladfrangu (approved):
LGTM; I don't know why we make it be `this` when augmentation works the same tho

iCrawl (approved):

PR #5681, opened 2021-05-27
fix(GuildMemberManager): allow moving members to any non-text channel

### LINKED ISSUE #5673: Cannot move members to a stage channel
If you try to move a member from a voice channel to a stage channel with a bot, it will throw a `GUILD_VOICE_CHANNEL_RESOLVE` error. It seems like this is neglect rather than intention. It's caused by this in `GuildMemberManager.edit`:
```js
      if (!_data.channel || _data.channel.type !== 'voice') {
        throw new Error('GUILD_VOICE_CHANNEL_RESOLVE');
      }
```

**Further details:**

- discord.js version: master branch (commit `0ca200f322f8cb21d0d3e5f6a70cc408d4f97ad1` from 24. May)
- Node.js version: 14
- Operating system: Win10
- Priority this issue should have – please be realistic and elaborate if possible: Medium, it's not crucial, but it's definitely causing issues and unnecessary extra code for clients.

AndyClausen:

This will allow moving members to channels with type `stage`. Closes #5673 🙂 

### DIFF
--- a/src/managers/GuildMemberManager.js
+++ b/src/managers/GuildMemberManager.js
@@ -2,6 +2,7 @@
 
 const BaseManager = require('./BaseManager');
 const { Error, TypeError, RangeError } = require('../errors');
+const BaseGuildVoiceChannel = require('../structures/BaseGuildVoiceChannel');
 const GuildMember = require('../structures/GuildMember');
 const Role = require('../structures/Role');
 const Collection = require('../util/Collection');
@@ -166,7 +167,7 @@ class GuildMemberManager extends BaseManager {
     const _data = { ...data };
     if (_data.channel) {
       _data.channel = this.guild.channels.resolve(_data.channel);
-      if (!_data.channel || _data.channel.type !== 'voice') {
+      if (!(_data.channel instanceof BaseGuildVoiceChannel)) {
         throw new Error('GUILD_VOICE_CHANNEL_RESOLVE');
       }
       _data.channel_id = _data.channel.id;


tipakA:
Technically, this way category channels are going to be an issue, since they return `false` from `isText()`
~~Is this a call for `isVoice()`?~~ 👀

AndyClausen:
> Technically, this way category channels are going to be an issue, since they return `false` from `isText()`
> ~Is this a call for `isVoice()`?~ 👀

Touché. I could change this to just add `stage` to the list of checks, but if an `isVoice` method is going to be added, I'd much rather wait for that 😅

PraveshKunwar:
> Technically, this way category channels are going to be an issue, since they return `false` from `isText()`
> ~Is this a call for `isVoice()`?~ 👀

would an isVoice method be needed?

ThaTiemsz:
> > Technically, this way category channels are going to be an issue, since they return `false` from `isText()`
> > ~Is this a call for `isVoice()`?~ 👀
> 
> Touché. I could change this to just add `stage` to the list of checks, but if an `isVoice` method is going to be added, I'd much rather wait for that 😅

Store channels will also be an issue.

tipakA:
Well, yes, i wasn't really trying to provide an extensive list of every single channel type this check will not work with, but simply give an example of why `isText()` isn't exactly the way here.

And my mention of `isVoice()` was made to avoid this issue in future in case discord would make new voice channel type. I'm definitely not saying whether it should be added or not, since it's a overall niche use case.

ckohen:
Since all applicable voice channels (dm channels technically have voice, but they don't count and not for bots anyways) extend `BaseGuildVoiceChannel`, and any future channel you could move people in a guild should probably follow suit, you can use `instance of BaseGuildVoiceChannel` I think.

AndyClausen:
I'll try that out tomorrow.

Now that I have you all here, should I rather make a new commit or amend the existing when it's such small changes? I've heard many different opinions from other repos.

AndyClausen:
Tested with `instanceof`, seems to work. Rebased and amended my commit.

vladfrangu (approved):

src/managers/GuildMemberManager.js
```diff
@@ -166,7 +167,7 @@ class GuildMemberManager extends BaseManager {
     const _data = { ...data };
     if (_data.channel) {
       _data.channel = this.guild.channels.resolve(_data.channel);
-      if (!_data.channel || _data.channel.type !== 'voice') {
+      if (!_data.channel || !(_data.channel instanceof BaseGuildVoiceChannel)) {
```
kyranet, line 170: `null instanceof Class` returns `false`, so the falsy check is unneeded.

```suggestion
      if (!(_data.channel instanceof BaseGuildVoiceChannel)) {
```

kyranet (changes requested):

kyranet (approved):

vladfrangu (approved):

SpaceEEC (approved):

iCrawl (approved):

PR #5683, opened 2021-05-27
refactor(DJSError): remove redundant check for captureStackTrace

iShibi:

Since, webpack isn't supported anymore, the check for `Error.captureStackTrace` is redundant. This PR removes it.

### DIFF
--- a/src/errors/DJSError.js
+++ b/src/errors/DJSError.js
@@ -15,7 +15,7 @@ function makeDiscordjsError(Base) {
     constructor(key, ...args) {
       super(message(key, args));
       this[kCode] = key;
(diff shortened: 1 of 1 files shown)


kyranet:
I think we should keep the check: [`Error.captureStackTrace`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error#static_methods) is a non-standard V8 extension, and as such, it's not subject to JavaScript's retro-compatibility, so it can be removed at any time with or without previous notice.

iShibi:
That will be an extremely rare situation, still I'll choose to be on the safe side. Closing 👍

PR #5690, opened 2021-05-28
fix(CommandInteraction): channel type should be text based channels

zLupa:

This PR change the Channel type from `CommandInteraction` to `TextChannel`, `DMChannel`, and `NewsChannel` because Slash Commands can't be used in another channel type than text

### DIFF
--- a/src/structures/CommandInteraction.js
+++ b/src/structures/CommandInteraction.js
@@ -16,6 +16,13 @@ class CommandInteraction extends Interaction {
     super(client, data);
 
     /**
+     * The channel this interaction was sent in
+     * @type {?TextChannel|NewsChannel|DMChannel}
+     * @name CommandInteraction#channel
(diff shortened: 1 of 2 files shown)


almostSouji:
Yes, they can, in other text-based channels (at the moment that being NewsChannel, soon also thread)

zLupa:
> Yes, they can, in other text-based channels (at the moment that being NewsChannel, soon also thread)

Yep, you're right, I will update to TextChannel, DMChannel, and NewsChannel.
Seems that now is right

iCrawl (approved):

SpaceEEC (approved):

vladfrangu (approved):

iShibi:
I'm late here but shouldn't the docs need to be updated too?

kyranet (approved):

PR #5692, opened 2021-05-28
feat(MessageSelectMenu): droppybois

### LINKED ISSUE #5753: MessageButton#emoji throws invalid form body error
While passing in options directly when creating a new button, including an emoji value throws an 'Invalid form body error' (full error provided below). Tested against both a Unicode emoji and a valid guild emoji identifier string.

Using .setEmoji() instead with the same input works just fine.

```js
 const row = new Discord.MessageActionRow()
            .addComponents([
                new Discord.MessageButton({
                    customID: 'test',
                    emoji: '🥟',
                    style: 'SECONDARY'
                })
            ])

        message.channel.send('test', {
            components: [row]
        })
```

Error:
```
components[0].components[0].emoji: Only dictionaries may be used in a ModelType
    at RequestHandler.execute (C:\Users\me\Documents\Visual Studio Code\test_bot\node_modules\discord.js\src\rest\RequestHandler.js:243:13)
    at processTicksAndRejections (internal/process/task_queues.js:95:5)
    at async RequestHandler.push (C:\Users\me\Documents\Visual Studio Code\test_bot\node_modules\discord.js\src\rest\RequestHandler.js:49:14)
```

**Further details:**

- discord.js version: master - commit #5674
- Node.js version: 14.17.0
- Operating system: Windows
- Priority this issue should have – please be realistic and elaborate if possible: Moderate/low, since this caters to a very specific use case and the method outlined in the draft guide works just fine.

**Relevant client options:**

- partials: none
- gateway intents: none
- other: none

monbrey:
### Select Menus are still in private beta. This PR will not function if you do not have access.

**Changelog**
 - New class `MessageSelectMenu`
 - New types `MessageSelectMenuOptions`, `MessageSelectOption`, `MessageSelectOptionData`
 - New utility method `resolvePartialEmoji` for resolving emoji without a Client
 - Improved emoji resolving in `MessageButton`, also applied to `MessageSelectOption` (fixes #5753)
 - Improved consistency of documentation across all component classes

### DIFF
--- a/src/structures/Interaction.js
+++ b/src/structures/Interaction.js
@@ -1,7 +1,7 @@
 'use strict';
 
 const Base = require('./Base');
-const { InteractionTypes } = require('../util/Constants');
+const { InteractionTypes, MessageComponentTypes } = require('../util/Constants');
 const SnowflakeUtil = require('../util/SnowflakeUtil');
 
 /**
@@ -126,7 +126,10 @@ class Interaction extends Base {
    * @returns {boolean}
    */
   isButton() {
-    return InteractionTypes[this.type] === InteractionTypes.MESSAGE_COMPONENT && this.componentType === 'BUTTON';
+    return (
+      InteractionTypes[this.type] === InteractionTypes.MESSAGE_COMPONENT &&
+      MessageComponentTypes[this.componentType] === MessageComponentTypes.BUTTON
+    );
   }
 
   /**
@@ -134,7 +137,10 @@ class Interaction extends Base {
    * @returns {boolean}
    */
   isSelectMenu() {
-    return InteractionTypes[this.type] === InteractionTypes.MESSAGE_COMPONENT && this.componentType === 'SELECT_MENU';
+    return (
+      InteractionTypes[this.type] === InteractionTypes.MESSAGE_COMPONENT &&
+      MessageComponentTypes[this.componentType] === MessageComponentTypes.SELECT_MENU
+    );
   }
 }
 
--- a/src/structures/Interaction.js
+++ b/src/structures/Interaction.js
@@ -114,7 +114,7 @@ class Interaction extends Base {
   }
 
   /**
-   * Indicates whether this interaction is a component interaction.
+   * Indicates whether this interaction is a message component interaction.
    * @returns {boolean}
    */
   isMessageComponent() {
@@ -127,6 +127,14 @@ class Interaction extends Base {
    */
   isButton() {
     return InteractionTypes[this.type] === InteractionTypes.MESSAGE_COMPONENT && this.componentType === 'BUTTON';
+  }
+
+  /**
+   * Indicates whether this interaction is a select menu interaction.
+   * @returns {boolean}
+   */
+  isSelectMenu() {
+    return InteractionTypes[this.type] === InteractionTypes.MESSAGE_COMPONENT && this.componentType === 'SELECT_MENU';
   }
 }
 
--- a/src/structures/MessageComponentInteraction.js
+++ b/src/structures/MessageComponentInteraction.js
@@ -55,6 +55,12 @@ class MessageComponentInteraction extends Interaction {
      * @type {InteractionWebhook}
      */
     this.webhook = new InteractionWebhook(this.client, this.applicationID, this.token);
+
+    /**
+     * The values selected in a MessageSelectMenu interaction
+     * @type {string[]}
+     */
+    this.values = this.componentType === 'SELECT_MENU' ? data.data.values : null;
   }
 
   /**
--- a/src/structures/MessageComponentInteraction.js
+++ b/src/structures/MessageComponentInteraction.js
@@ -57,7 +57,7 @@ class MessageComponentInteraction extends Interaction {
     this.webhook = new InteractionWebhook(this.client, this.applicationID, this.token);
 
     /**
-     * The values selected in a MessageSelectMenu interaction
+     * The values selected, if the component which was interacted with was a select menu
      * @type {string[]}
      */
     this.values = this.componentType === 'SELECT_MENU' ? data.data.values : null;
--- a/src/structures/MessageComponentInteraction.js
+++ b/src/structures/MessageComponentInteraction.js
@@ -55,12 +55,6 @@ class MessageComponentInteraction extends Interaction {
      * @type {InteractionWebhook}
      */
     this.webhook = new InteractionWebhook(this.client, this.applicationID, this.token);
-
-    /**
-     * The values selected, if the component which was interacted with was a select menu
-     * @type {string[]}
-     */
-    this.values = this.componentType === 'SELECT_MENU' ? data.data.values : null;
   }
 
   /**
--- a/src/structures/MessageSelectMenu.js
+++ b/src/structures/MessageSelectMenu.js
@@ -1,6 +1,7 @@
 'use strict';
 
 const BaseMessageComponent = require('./BaseMessageComponent');
+const { MessageComponentTypes } = require('../util/Constants');
 const Util = require('../util/Util');
 
 class MessageSelectMenu extends BaseMessageComponent {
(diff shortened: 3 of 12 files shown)


KuryKat (approved):
great changes, it will be great to be able to use it soon! :D👍

BanTheNons:
Is it currently ready for use?

src/structures/MessageSelectMenu.js
```diff
@@ -0,0 +1,184 @@
...
+  /**
+   * Transforms this select menu to a plain object
+   * @returns {Object} The raw data of this select menu
+   */
+  toJSON() {
+    return {
+      custom_id: this.customID,
+      placeholder: this.placeholder,
+      min_values: this.minValues,
+      max_values: this.maxValues ?? this.minValues ? this.options.length : undefined,
```
almostSouji, line 152:
```suggestion
      max_values: this.maxValues ?? (this.minValues ? this.options.length : undefined),
```
fixes a bug resulting in max being ignored due to operator precedence

almostSouji (changes requested):

src/util/Util.js
```diff
@@ -267,6 +267,21 @@ class Util {
...
+   * @returns {Object} Object with possible `animated`, `name`, and `id` properties
+   * @private
+   */
+  static resolvePartialEmoji(emoji) {
+    if (typeof emoji === 'undefined') return null;
+    if (typeof emoji === 'string') {
+      return emoji.length ? (/^\d{17,19}$/.test(emoji) ? { id: emoji } : Util.parseEmoji(emoji)) : null;
+    }
+    const { id, name, animated } = emoji;
+    return { id, name, animated };
```
vladfrangu, line 282: Kinda redundant, no?

src/util/Constants.js
```diff
@@ -828,9 +828,10 @@ exports.InteractionResponseTypes = createEnum([
  * The type of a message component
  * * ACTION_ROW
  * * BUTTON
+ * * SELECT_MENU
```
vladfrangu, line 831: Is this what Discord calls them?

vladfrangu (changes requested):

src/util/Constants.js
monbrey: Yeah, "Select Menu" is the term used in the docs provided

src/util/Util.js
monbrey: Somewhat - this was done to match the typing of what I've said it returns, and what will be stored in the #emoji prop of components.

ckohen: I don't think its redundant, it also prevents returning non emoji props. However, I might suggest throwing an error if both id and name are undefined in between these lines, as in that case you do not have a valid emoji.

src/structures/MessageComponentInteraction.js
```diff
@@ -49,6 +49,12 @@ class MessageComponentInteraction extends Interaction {
      * @type {InteractionWebhook}
      */
     this.webhook = new InteractionWebhook(this.client, this.applicationID, this.token);
+
+    /**
+     * The values selected, if the component which was interacted with was a select menu
+     * @type {string[]}
+     */
+    this.values = this.componentType === 'SELECT_MENU' ? data.data.values : null;
```
kyranet, line 57: If the property only exists when it's a SELECT_MENU, why not make a `SelectMenuMessageComponentInteaction` class?

The line 33 would need to be refactored so we don't resolve the component type twice, but I find it more ideal to define the properties that exist only in different component types, similar to why we have TextChannel, VoiceChannel, DMChannel, etc.

kyranet (commented):
Just a small thing and ckohen's suggestion in the previous review, everything else LGTM.

src/structures/MessageComponentInteraction.js
monbrey: I don't entirely disagree, but I'm definitely dropping the generic "MessageComponent" part of the naming convention if we do make these... well, not generic. `ButtonInteraction` and `SelectMenuInteraction` would fine imo.

Following from that change, what collectors do we build into Message and TextChannel? Do those also get separated?

kyranet:
This needs a rebase.

typings/index.d.ts
```diff
@@ -1449,6 +1451,29 @@ declare module 'discord.js' {
...
+    public setDisabled(disabled: boolean): this;
+    public setMaxValues(maxValues: number): this;
+    public setMinValues(minValues: number): this;
+    public setPlaceholder(placeholder: string): this;
+    public spliceOptions(
+      index: number,
+      deleteCount: number,
+      ...options: MessageSelectOptionData[] | MessageSelectOptionData[][]
+    ): this;
+    public toJSON(): object;
```
kyranet, line 1474:
```suggestion
    public toJSON(): unknown;
```

kyranet (changes requested):
Just one thing, otherwise LGTM:

kyranet (approved):

iCrawl:
This needs a rebase.

vladfrangu (approved):

SpaceEEC (approved):

iCrawl (approved):

iCrawl:
This needs a rebase.

src/structures/Interaction.js
```diff
@@ -128,6 +128,14 @@ class Interaction extends Base {
   isButton() {
     return InteractionTypes[this.type] === InteractionTypes.MESSAGE_COMPONENT && this.componentType === 'BUTTON';
   }
+
+  /**
+   * Indicates whether this interaction is a select menu interaction.
+   * @returns {boolean}
+   */
+  isSelectMenu() {
+    return InteractionTypes[this.type] === InteractionTypes.MESSAGE_COMPONENT && this.componentType === 'SELECT_MENU';
```
NotSugden, line 137: would make sense to also use the `MessageComponentTypes` enum

typings/index.d.ts
```diff
@@ -3459,6 +3496,31 @@ declare module 'discord.js' {
...
   type MessageResolvable = Message | Snowflake;
 
+  interface MessageSelectMenuOptions extends BaseMessageComponentOptions {
+    customID?: string;
+    disabled?: boolean;
+    maxValues?: number;
+    minValues?: number;
+    options?: MessageSelectOptionData[];
+    placeholder?: string;
+  }
```
advaith1, line 3506: why are these all marked as optional? looks like BaseMessageComponentOptions and MessageButtonOptions have the same issue
also, should probably enforce the `type` on the specific component options

```suggestion
  interface MessageSelectMenuOptions extends BaseMessageComponentOptions {
    customID: string;
    disabled?: boolean;
    maxValues?: number;
    minValues?: number;
    options: MessageSelectOptionData[];
    placeholder?: string;
  }
```

monbrey: Because it's not required to pass any of these to the constructors for a MessageButton or MessageSelectMenu

advaith1: These options interfaces can also be used directly when sending a message, in which case the fields are required; the constructors should use Partial<...Options>

monbrey: Fair point, I've been considering similar things - arguably if passing options to a constructor it *should* enforce all the required params, though in that instance `type` can be omitted, while for raw formats in messages it can't be. I wont commit this change just yet, will review the types more broadly later.

iCrawl:
This needs a rebase.

typings/index.d.ts
monbrey: Yeah, I'm going to leave this out of this PR, since if I'm going to make typings changes like this it should be consistent with Buttons too. Will do separately for both after this merges.

bleuthoot-sven:
Looks like Selection Menus have been added to the documentation. I assume they have been released now:

https://discord.com/developers/docs/interactions/message-components#select-menus

advaith1:
yes, it was announced yesterday:

image

monbrey:
This was also merged over a week ago so not sure why it matters to this PR that docs were released.

PR #5695, opened 2021-05-28
feat(ApplicationCommand): methods to make managing command permissions easier

Sasuke-7:

Added new methods to make it easier to manage the permissions of a command.


monbrey:
Unfortunately Discord doesn't provide an endpoint for this. [Edit Application Command Permissions](https://discord.com/developers/docs/interactions/slash-commands#edit-application-command-permissions) expects a full set of `ApplicationCommandPermissions` every time.

For discord.js to support this, the library would need to perform internal fetching of the command permissions and calculate the delta - internal fetching goes against design.

Sasuke-7:
> Unfortunately Discord doesn't provide an endpoint for this. [Edit Application Command Permissions](https://discord.com/developers/docs/interactions/slash-commands#edit-application-command-permissions) expects a full set of `ApplicationCommandPermissions` every time.
> 
> For discord.js to support this, the library would need to perform internal fetching of the command permissions and calculate the delta, as you've done, but internal fetching goes against design.
> 
> Having said that, I do _like_ this method so if the maintainers want to support this design, it'd be a thumbs-up from me. However, it should then be expanded to support more use-cases, such as both adding and removing permissions, and should also be on the `ApplicationCommand` class.

Thanks for the support, I'll expand it and add support for new functions and add it to the `ApplicationCommand` class, really thx ><

src/managers/GuildApplicationCommandManager.js
```diff
@@ -119,6 +120,87 @@ class GuildApplicationCommandManager extends ApplicationCommandManager {
...
+   * guild.commands.addPermissions('123456789012345678', [
+   *   {
+   *     id: guild.roles.everyone.id,
+   *     type: 'ROLE',
+   *     permission: false
+   *   },
+   * ])
+   *   .then(console.log)
+   *   .catch(console.error);
+   */
```
monbrey, line 139: Does this example actually work? I don't think Discord designed it to allow `@everyone` overrides, that's what `defaultPermission` is for.

src/managers/GuildApplicationCommandManager.js
```diff
@@ -119,6 +120,87 @@ class GuildApplicationCommandManager extends ApplicationCommandManager {
...
+  async addPermissions(command, permissions) {
+    const id = this.resolveID(command);
+    if (!id) throw new TypeError('INVALID_TYPE', 'command', 'ApplicationCommandResolvable');
+
+    const perms = await this.fetchPermissions(id).catch(() => []);
```
monbrey, line 144: I know this is to handle the 404 when no permissions are set, but you should probably check the error code or something first - that's not the *only* reason a Promise could reject.

src/managers/GuildApplicationCommandManager.js
```diff
@@ -119,6 +120,87 @@ class GuildApplicationCommandManager extends ApplicationCommandManager {
...
+        permissions.push(perm);
+      }
+    }
+
+    const data = await this.commandPath(id).permissions.put({
+      data: { permissions: permissions.map(perm => this.constructor.transformPermissions(perm)) },
+    });
+
+    return data.permissions.map(perm => this.constructor.transformPermissions(perm, true));
+  }
```
monbrey, line 156:
```suggestion
    return this.setPermissions(id, permissions);
```
Just call the existing method from here after calculating the new permission set, no need to duplicate code.

src/managers/GuildApplicationCommandManager.js
```diff
@@ -119,6 +120,87 @@ class GuildApplicationCommandManager extends ApplicationCommandManager {
...
+  async removePermissions(command, userOrRole) {
+    const id = this.resolveID(command);
+    if (!id) throw new TypeError('INVALID_TYPE', 'command', 'ApplicationCommandResolvable');
+
+    let permissions = await this.fetchPermissions(id).catch(() => []);
+    if (Array.isArray(userOrRole)) {
+      userOrRole = userOrRole.map(x => {
+        if (typeof x === 'string') return x;
+        if (x instanceof Role) return x.id;
+        return this.client.users.resolveID(x);
```
monbrey, line 184:
```suggestion
        return this.client.users.resolveID(x) ?? this.guild.roles.resolveID(x);
```

src/managers/GuildApplicationCommandManager.js
```diff
@@ -119,6 +120,87 @@ class GuildApplicationCommandManager extends ApplicationCommandManager {
...
+      });
+
+      permissions = permissions.filter(perm => !userOrRole.includes(perm.id));
+    } else {
+      if (typeof userOrRole !== 'string') {
+        if (userOrRole instanceof Role) userOrRole = userOrRole.id;
+        userOrRole = this.client.users.resolveID(userOrRole);
```
monbrey, line 191:
```suggestion
        userOrRole = this.client.users.resolveID(userOrRole) ?? this.guild.roles.resolveID(userOrRole);
```

src/managers/GuildApplicationCommandManager.js
```diff
@@ -1,6 +1,7 @@
 'use strict';
 
 const ApplicationCommandManager = require('./ApplicationCommandManager');
+const Role = require('../structures/Role');
```
monbrey, line 4: No longer required with below changes.

src/managers/GuildApplicationCommandManager.js
```diff
@@ -119,6 +120,87 @@ class GuildApplicationCommandManager extends ApplicationCommandManager {
...
+      }
+
+      permissions = permissions.filter(perm => perm.id !== userOrRole);
+    }
+
+    const data = await this.commandPath(id).permissions.put({
+      data: { permissions: permissions.map(perm => this.constructor.transformPermissions(perm)) },
+    });
+
+    return data.permissions.map(perm => this.constructor.transformPermissions(perm, true));
```
monbrey, line 201:
```suggestion
    return this.setPermissions(id, permissions);
```
As above

src/managers/GuildApplicationCommandManager.js
```diff
@@ -119,6 +120,87 @@ class GuildApplicationCommandManager extends ApplicationCommandManager {
...
+   * // Remove multiple roles from this command
+   * guild.commands.removePermissions('123456789012345678', [ '876543210123456789', '765432101234567890' ])
+   *    .then(console.log)
+   *    .catch(console.error);
+   */
+  async removePermissions(command, userOrRole) {
+    const id = this.resolveID(command);
+    if (!id) throw new TypeError('INVALID_TYPE', 'command', 'ApplicationCommandResolvable');
+
+    let permissions = await this.fetchPermissions(id).catch(() => []);
```
monbrey, line 179: As above

monbrey (changes requested):

typings/index.d.ts
```diff
@@ -2117,6 +2121,14 @@ declare module 'discord.js' {
...
       permissions: GuildApplicationCommandPermissionData[],
     ): Promise<Collection<Snowflake, ApplicationCommandPermissions[]>>;
+    public addPermissions(
+      command: ApplicationCommandResolvable,
+      permissions: ApplicationCommandPermissionData[],
+    ): Promise<ApplicationCommandPermissions[]>;
+    public removePermissions(
+      command: ApplicationCommandResolvable,
+      userOrRole: UserResolvable | RoleResolvable | UserResolvable[] | RoleResolvable[],
+    ): Promise<ApplicationCommandPermissionData[]>;
```
JKLorenzo, line 2131:
```suggestion
    ): Promise<ApplicationCommandPermissions[]>;
```

JKLorenzo (changes requested):

src/structures/ApplicationCommand.js
```diff
@@ -183,6 +183,50 @@ class ApplicationCommand extends Base {
...
+   */
+  addPermissions(permissions) {
+    if (!this.guild) throw new Error('GLOBAL_COMMAND_PERMISSIONS');
+    return this.manager.addPermissions(this, permissions);
+  }
+
+  /**
+   * Remove permissions from a command.
+   * <warn>This is only available for guild application commands.</warn>
+   * @param {UserResolvable | RoleResolvable | UserResolvable[] | RoleResolvable[]} userOrRole The user(s) and role(s)
```
vladfrangu, line 211:
```suggestion
   * @param {UserResolvable|RoleResolvable|UserResolvable[]|RoleResolvable[]} userOrRole The user(s) and role(s)
```

src/managers/GuildApplicationCommandManager.js
```diff
@@ -119,6 +119,80 @@ class GuildApplicationCommandManager extends ApplicationCommandManager {
...
+      }
+    }
+
+    return this.setPermissions(id, permissions);
+  }
+
+  /**
+   * Remove permissions from a command.
+   * @param {ApplicationCommandResolvable} command The command to edit the permissions for
+   * @param {UserResolvable | RoleResolvable | UserResolvable[] | RoleResolvable[]} userOrRole The user(s) and role(s)
```
vladfrangu, line 160:
```suggestion
   * @param {UserResolvable|RoleResolvable|UserResolvable[]|RoleResolvable[]} userOrRole The user(s) and role(s)
```

vladfrangu (changes requested):
Worlds smallest nitpicks

src/managers/GuildApplicationCommandManager.js
```diff
@@ -119,6 +119,80 @@ class GuildApplicationCommandManager extends ApplicationCommandManager {
...
+    if (!id) throw new TypeError('INVALID_TYPE', 'command', 'ApplicationCommandResolvable');
+
+    const perms = await this.fetchPermissions(id).catch(err => {
+      if (err.httpStatus !== 404) throw err;
+      return [];
+    });
+
+    for (const perm of perms) {
+      if (!permissions.find(x => x.id === perm.id)) {
+        permissions.push(perm);
```
kyranet, line 150: I don't agree we should be mutating the source array, if somebody passes `Object.freeze([])`, this is bound to throw.

src/managers/GuildApplicationCommandManager.js
```diff
@@ -119,6 +119,80 @@ class GuildApplicationCommandManager extends ApplicationCommandManager {
...
+
+  /**
+   * Remove permissions from a command.
+   * @param {ApplicationCommandResolvable} command The command to edit the permissions for
+   * @param {UserResolvable|RoleResolvable|UserResolvable[]|RoleResolvable[]} userOrRole The user(s) and role(s)
+   * to remove from the command permissions
+   * @returns {Promise<ApplicationCommandPermissions[]>}
+   * @example
+   * // Remove a user permission from this command
+   * guild.commands.addPermissions('123456789012345678', '876543210123456789')
```
kyranet, line 165:
```suggestion
   * guild.commands.removePermissions('123456789012345678', '876543210123456789')
```

src/managers/GuildApplicationCommandManager.js
```diff
@@ -119,6 +119,80 @@ class GuildApplicationCommandManager extends ApplicationCommandManager {
...
+      if (err.httpStatus !== 404) throw err;
+      return [];
+    });
+
+    if (Array.isArray(userOrRole)) {
+      userOrRole = userOrRole.map(x => this.client.users.resolveID(x) ?? this.guild.roles.resolveID(x));
+      permissions = permissions.filter(perm => !userOrRole.includes(perm.id));
+    } else {
+      if (typeof userOrRole !== 'string') {
+        userOrRole = this.client.users.resolveID(userOrRole) ?? this.guild.roles.resolveID(userOrRole);
```
kyranet, line 188: Might want to guard against this returning `null`.

src/managers/GuildApplicationCommandManager.js
```diff
@@ -119,6 +119,80 @@ class GuildApplicationCommandManager extends ApplicationCommandManager {
...
+    const id = this.resolveID(command);
+    if (!id) throw new TypeError('INVALID_TYPE', 'command', 'ApplicationCommandResolvable');
+
+    let permissions = await this.fetchPermissions(id).catch(err => {
+      if (err.httpStatus !== 404) throw err;
+      return [];
+    });
+
+    if (Array.isArray(userOrRole)) {
+      userOrRole = userOrRole.map(x => this.client.users.resolveID(x) ?? this.guild.roles.resolveID(x));
```
kyranet, line 184: Might want to guard against any of the entries returning `null`.

kyranet (changes requested):

src/managers/GuildApplicationCommandManager.js
```diff
@@ -119,6 +119,86 @@ class GuildApplicationCommandManager extends ApplicationCommandManager {
...
+
+    let permissions = await this.fetchPermissions(id).catch(err => {
+      if (err.httpStatus !== 404) throw err;
+      return [];
+    });
+
+    if (Array.isArray(userOrRole)) {
+      userOrRole = userOrRole
+        .map(x => this.client.users.resolveID(x) ?? this.guild.roles.resolveID(x))
+        .filter(x => !!x);
```
kyranet, line 187: I believe this should throw if an ID were to not be resolveable, as opposed to silently failing, similar to what [BitField does](https://github.com/discordjs/discord.js/blob/master/src/util/BitField.js#L151).

kyranet (changes requested):

src/managers/GuildApplicationCommandManager.js
```diff
@@ -119,6 +120,90 @@ class GuildApplicationCommandManager extends ApplicationCommandManager {
...
+  async addPermissions(command, permissions) {
+    const id = this.resolveID(command);
+    if (!id) throw new TypeError('INVALID_TYPE', 'command', 'ApplicationCommandResolvable');
+
+    const perms = await this.fetchPermissions(id).catch(err => {
+      if (err.httpStatus !== 404) throw err;
+      return [];
+    });
+
+    const newPermissions = [...permissions];
```
NotSugden, line 149:
```suggestion
    const newPermissions = permissions.slice();
```
probably not much of a difference but i remember this being used somewhere

src/managers/GuildApplicationCommandManager.js
```diff
@@ -119,6 +120,90 @@ class GuildApplicationCommandManager extends ApplicationCommandManager {
...
+    for (const perm of perms) {
+      if (!newPermissions.find(x => x.id === perm.id)) {
```
NotSugden, line 151:
```suggestion
      if (!newPermissions.some(x => x.id === perm.id)) {
```

src/managers/GuildApplicationCommandManager.js
```diff
@@ -119,6 +120,90 @@ class GuildApplicationCommandManager extends ApplicationCommandManager {
...
+    if (!id) throw new TypeError('INVALID_TYPE', 'command', 'ApplicationCommandResolvable');
+
+    let permissions = await this.fetchPermissions(id).catch(err => {
+      if (err.httpStatus !== 404) throw err;
+      return [];
+    });
+
+    if (Array.isArray(userOrRole)) {
+      userOrRole = userOrRole.map(x => {
+        if (!this.client.users.resolveID(x) && !this.guild.roles.resolveID(x)) {
```
NotSugden, line 187: perhaps create variables for these?

src/managers/GuildApplicationCommandManager.js
```diff
@@ -119,6 +120,90 @@ class GuildApplicationCommandManager extends ApplicationCommandManager {
...
+      if (err.httpStatus !== 404) throw err;
+      return [];
+    });
+
+    if (Array.isArray(userOrRole)) {
+      const ids = userOrRole.map(x => {
+        if (!this.client.users.resolveID(x) && !this.guild.roles.resolveID(x)) {
+          throw new TypeError('INVALID_ELEMENT', 'Array', 'userOrRole', x);
+        }
+        return this.client.users.resolveID(x) ?? this.guild.roles.resolveID(x);
```
kyranet, line 190:
```suggestion
        const resolved = this.client.users.resolveID(x) ?? this.guild.roles.resolveID(x);
        if (resolved) return resolved;
        throw new TypeError('INVALID_ELEMENT', 'Array', 'userOrRole', x);
```

src/managers/GuildApplicationCommandManager.js
```diff
@@ -119,6 +120,90 @@ class GuildApplicationCommandManager extends ApplicationCommandManager {
...
+      });
+
+      permissions = permissions.filter(perm => !ids.includes(perm.id));
+    } else {
+      if (typeof userOrRole !== 'string') {
+        userOrRole = this.client.users.resolveID(userOrRole) && !this.guild.roles.resolveID(userOrRole);
```
kyranet, line 196:
```suggestion
        userOrRole = this.client.users.resolveID(userOrRole) ?? this.guild.roles.resolveID(userOrRole);
```

Also, the typeof check can be a little useless, since `resolveID` short-circuits them.

kyranet (changes requested):
As a side note, please do not dismiss our requested changes without resolving them.

typings/index.d.ts
kyranet: @Sasuke-7 this should be the last requested change, and then the PR will look good to me.

kyranet (changes requested):

kyranet (approved):

kyranet:
Why was this closed?

src/managers/GuildApplicationCommandManager.js
```diff
@@ -119,6 +120,119 @@ class GuildApplicationCommandManager extends ApplicationCommandManager {
...
+      return [];
+    });
+
+    if (typeof userOrRole !== 'string') {
+      userOrRole = this.client.users.resolveID(userOrRole) ?? this.guild.roles.resolveID(userOrRole);
+      if (!userOrRole) {
+        throw new TypeError('INVALID_TYPE', 'userOrRole', 'UserResolvable or RoleResolvable');
+      }
+    }
+    return !!permissions.find(perm => perm.id === userOrRole)?.permission;
```
ckohen, line 233: I believe this should not return the permission, instead returning simply whether or not the id exists in the current permission overwrites. Otherwise how could you tell the difference between the permission not being set (thus true) and it being set false when `default_permission: true`

vladfrangu (approved):

iCrawl:
Superseeded by #5897

PR #5696, opened 2021-05-29
feat(GuildMember): add guild avatars

ImRodry:
This PR adds server avatars to the GuildMember object.

discord-api-docs PR:
- https://github.com/discord/discord-api-docs/pull/3081

### DIFF
--- a/src/structures/GuildMember.js
+++ b/src/structures/GuildMember.js
@@ -71,6 +71,15 @@ class GuildMember extends Base {
     }
 
     if ('nick' in data) this.nickname = data.nick;
+    if ('avatar' in data) {
+      /**
+       * The guild member's avatar hash
+       * @type {?string}
+       */
+      this.avatar = data.avatar;
+    } else if (typeof this.avatar !== 'string') {
+      this.avatar = null;
+    }
     if ('joined_at' in data) this.joinedTimestamp = new Date(data.joined_at).getTime();
     if ('premium_since' in data) {
       this.premiumSinceTimestamp = data.premium_since ? new Date(data.premium_since).getTime() : null;
@@ -110,6 +119,26 @@ class GuildMember extends Base {
    */
   get voice() {
     return this.guild.voiceStates.cache.get(this.id) ?? new VoiceState(this.guild, { user_id: this.id });
+  }
+
+  /**
+   * A link to the member's guild avatar.
+   * @param {ImageURLOptions} [options={}] Options for the Image URL
+   * @returns {?string}
+   */
+  avatarURL({ format, size, dynamic } = {}) {
+    if (!this.avatar) return null;
+    return this.client.rest.cdn.GuildMemberAvatar(this.guild.id, this.id, this.avatar, format, size, dynamic);
+  }
+
+  /**
+   * A link to the member's guild avatar if they have one.
+   * Otherwise, a link to their {@link User#displayAvatarURL} will be returned.
+   * @param {ImageURLOptions} [options={}] Options for the Image URL
+   * @returns {string}
+   */
+  displayAvatarURL(options) {
+    return this.avatarURL(options) ?? this.user.displayAvatarURL(options);
   }
 
   /**
@@ -323,6 +352,7 @@ class GuildMember extends Base {
       this.guild.id === member.guild.id &&
       this.joinedTimestamp === member.joinedTimestamp &&
       this.nickname === member.nickname &&
+      this.avatar === member.avatar &&
       this.pending === member.pending &&
       (this._roles === member._roles ||
         (this._roles.length === member._roles.length && this._roles.every((role, i) => role === member._roles[i])))
@@ -341,12 +371,15 @@ class GuildMember extends Base {
   }
 
   toJSON() {
-    return super.toJSON({
+    const json = super.toJSON({
       guild: 'guildId',
       user: 'userId',
       displayName: true,
       roles: true,
     });
+    json.avatarURL = this.avatarURL();
+    json.displayAvatarURL = this.displayAvatarURL();
+    return json;
   }
 
   // These are here only for documentation purposes - they are implemented by TextBasedChannel
--- a/src/util/Constants.js
+++ b/src/util/Constants.js
@@ -49,6 +49,10 @@ exports.Endpoints = {
         if (dynamic && hash.startsWith('a_')) format = 'gif';
         return makeImageUrl(`${root}/avatars/${userId}/${hash}`, { format, size });
       },
+      GuildMemberAvatar: (guildId, memberId, hash, format = 'webp', size, dynamic = false) => {
+        if (dynamic) format = hash.startsWith('a_') ? 'gif' : format;
+        return makeImageUrl(`${root}/guilds/${guildId}/users/${memberId}/avatars/${hash}`, { format, size });
+      },
       Banner: (id, hash, format, size, dynamic = false) => {
         if (dynamic && hash.startsWith('a_')) format = 'gif';
         return makeImageUrl(`${root}/banners/${id}/${hash}`, { format, size });
--- a/src/util/Constants.js
+++ b/src/util/Constants.js
@@ -50,7 +50,7 @@ exports.Endpoints = {
         return makeImageUrl(`${root}/avatars/${userId}/${hash}`, { format, size });
       },
       GuildMemberAvatar: (guildId, memberId, hash, format = 'webp', size, dynamic = false) => {
-        if (dynamic) format = hash.startsWith('a_') ? 'gif' : format;
+        if (dynamic && hash.startsWith('a_')) format = 'gif';
         return makeImageUrl(`${root}/guilds/${guildId}/users/${memberId}/avatars/${hash}`, { format, size });
       },
       Banner: (id, hash, format, size, dynamic = false) => {
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -2296,6 +2296,14 @@ export const Constants: {
         size: AllowedImageSize,
         dynamic: boolean,
       ) => string;
+      GuildMemberAvatar: (
+        guildId: Snowflake,
+        memberId: Snowflake,
+        hash: string,
+        format?: DynamicImageFormat,
+        size?: AllowedImageSize,
+        dynamic?: boolean,
+      ) => string;
       Icon: (
         guildId: Snowflake,
         hash: string,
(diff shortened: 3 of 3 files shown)


kyranet (approved):

ImRodry:
I would add a way to set the server avatar but this feature seems to still be in development and I can't get it to work although the code shouldn't be much different from ClientUser.setAvatar(). If I can get it to work before this PR is merged I'll probably add a ClientGuildMember class for that and add that function, otherwise it'll need to be added later on

src/structures/GuildMember.js
```diff
@@ -136,6 +145,27 @@ class GuildMember extends Base {
...
+  avatarURL({ format, size, dynamic } = {}) {
+    if (!this.avatar) return null;
+    return this.client.rest.cdn.Avatar(this.id, this.avatar, format, size, dynamic);
+  }
+
+  /**
+   * A link to the member's server avatar if they have one.
+   * Otherwise a link to their user display avatar will be returned.
+
+   * @param {ImageURLOptions} [options={}] Options for the Image URL
```
iShibi, line 162:
```suggestion
   * Otherwise a link to their user display avatar will be returned.
   * @param {ImageURLOptions} [options={}] Options for the Image URL
```
Remove the empty line here

iShibi (changes requested):

src/structures/GuildMember.js
```diff
@@ -136,6 +145,26 @@ class GuildMember extends Base {
...
   }
 
+  /**
+   * A link to the member's server avatar.
+   * @param {ImageURLOptions} [options={}] Options for the Image URL
+   * @returns {?string}
+   */
+  avatarURL({ format, size, dynamic } = {}) {
+    if (!this.avatar) return null;
+    return this.client.rest.cdn.Avatar(this.id, this.avatar, format, size, dynamic);
```
GamingGeek, line 155: This won't work, server avatars use a different URL, which you'll need to add to `src/util/Constants.js`

`/guilds/:guildID/users/:userID/avatars/:hash`

ImRodry: You're absolutely right, I missed this. Can you give me a guildId, a userId and a hash so I can test?

GamingGeek: https://cdn.discordapp.com/guilds/692925064135311390/users/90339695967350784/avatars/67fcd6af88d76888b8525c5c3684b6b2.png?size=2048

GamingGeek (approved):
LGTM

DiscordCatto (changes requested):
`GuildAvatar` needs to be added to the typings file, Constants -> Endpoints -> CDN

src/util/Constants.js
```diff
@@ -132,6 +132,10 @@ exports.Endpoints = {
         if (dynamic) format = hash.startsWith('a_') ? 'gif' : format;
         return makeImageUrl(`${root}/avatars/${userID}/${hash}`, { format, size });
       },
+      GuildAvatar: (guildId, memberId, hash, format = 'webp', size, dynamic = false) => {
```
kyranet, line 135: Since it's a GuildMember avatar, and not a GuildAvatar, this should be named GuildMemberAvatar.

kyranet (changes requested):

src/util/Constants.js
ImRodry: is there anything else I should rename?

kyranet: Not really, just the references in the code and typings.

kyranet (approved):

iCrawl:
This needs a rebase.

ImRodry:
Done!

src/util/Constants.js
```diff
@@ -132,6 +132,10 @@ exports.Endpoints = {
         if (dynamic) format = hash.startsWith('a_') ? 'gif' : format;
         return makeImageUrl(`${root}/avatars/${userID}/${hash}`, { format, size });
       },
+      GuildMemberAvatar: (guildId, memberId, hash, format = 'webp', size, dynamic = false) => {
```
vladfrangu, line 135: Perhaps making these IDs like above for consistency sake?

ImRodry: You mean using ID instead of Id? I believe it was like that when I made the PR or at least I saw that somewhere else but I can change it of course.

vladfrangu:
LGTM, however let's wait for Discord first to finalize this 👍

catcosmicice (approved):

iCrawl (approved):

ImRodry:
@vladfrangu this feature has started rolling out to a percentage of regular users as of a few days ago, can we merge the PR now?

typings/index.d.ts
```diff
@@ -979,6 +1013,8 @@ declare module 'discord.js' {
     public readonly roles: GuildMemberRoleManager;
     public user: User;
     public readonly voice: VoiceState;
+    public avatarURL(options?: ImageURLOptions & { dynamic?: boolean }): string | null;
+    public displayAvatarURL(options?: ImageURLOptions & { dynamic?: boolean }): string;
```
iShibi, line 1017:
```suggestion
    public avatarURL(options?: ImageURLOptions): string | null;
    public displayAvatarURL(options?: ImageURLOptions): string;
```
This isn't needed after #5805

ImRodry: Will fix
Btw are the types Snowflake | number correct for user and guildIds? I see they're used on the Avatar function so I just went with them but it doesn't make much sense right?

src/structures/GuildMember.js
```diff
@@ -83,6 +83,15 @@ class GuildMember extends Base {
...
     if ('nick' in data) this.nickname = data.nick;
+    if ('avatar' in data) {
+      /**
+       * The ID of the member's guild avatar
+       * @type {?string}
+       */
+      this.avatar = data.avatar;
+    } else if (typeof this.avatar !== 'string') {
+      this.avatar = null;
+    }
```
iShibi, line 94:
```suggestion
    if ('avatar' in data) {
      /**
       * The ID of the member's guild avatar
       * @type {?string}
       */
      this.avatar = data.avatar;
    }
```
If `avatar` field is present in the response, then it will be either `null` or `string`. So, the `else if` check seems redundant here. Also, why were you checking for `this.avatar` in it anyway?

almeidx (changes requested):
[`GuildMember#equals`](https://github.com/discordjs/discord.js/blob/d9df258d00bef54353818f62423ec91a5f52da9e/src/structures/GuildMember.js#L385) should be updated to support this new property

src/structures/GuildMember.js
SpaceEEC: I think the goal is to initialize the property with null if the member has no guild avatar.

iShibi: hmm, I see. Wouldn't it be a good idea to have it initialized to `null` in constructor then?

ImRodry: I used the same checks that were used in User#avatar so that they’re consistent with each other but I can change that if you think it’s necessary

vladfrangu (approved):

ImRodry:
Rebased.

src/util/Constants.js
```diff
@@ -7,10 +7,10 @@ exports.UserAgent = `DiscordBot (${Package.homepage.split('#')[0]}, ${Package.ve
 
 exports.WSCodes = {
```
GoldenAngel2, line 8: @ImRodry The `WSCodes` names below needs to be reverted.

GoldenAngel2 (changes requested):

src/util/Constants.js
ImRodry: oops I did a mass replace and didn't notice those, will fix!

src/structures/GuildMember.js
```diff
@@ -112,6 +121,26 @@ class GuildMember extends Base {
...
+  }
+
+  /**
+   * A link to the member's guild avatar if they have one.
+   * Otherwise, a link to their {@link User#displayAvatarURL} will be returned.
+   * @param {ImageURLOptions} [options={}] Options for the Image URL
+   * @returns {string}
+   */
+  displayAvatarURL(options) {
+    return this.avatarURL(options) || this.user.displayAvatarURL(options);
```
NotSugden, line 141:
```suggestion
    return this.avatarURL(options) ?? this.user.displayAvatarURL(options);
```

NotSugden (changes requested):

PuneetGopinath:
Waiting for this

kyranet:
+1'ing the PR won't make it land faster, this is still an experiment and can change at any moment, @PuneetGopinath

PuneetGopinath:
> +1'ing the PR won't make it land faster 

I was just telling to everyone that I like it

> this is still an experiment on Discord's and can change at any moment

Ohk. I didn't know that.

src/util/Constants.js
```diff
@@ -49,6 +49,10 @@ exports.Endpoints = {
         if (dynamic) format = hash.startsWith('a_') ? 'gif' : format;
         return makeImageUrl(`${root}/avatars/${userId}/${hash}`, { format, size });
       },
+      GuildMemberAvatar: (guildId, memberId, hash, format = 'webp', size, dynamic = false) => {
+        if (dynamic) format = hash.startsWith('a_') ? 'gif' : format;
+        return makeImageUrl(`${root}/guilds/${guildId}/users/${memberId}/avatars/${hash}`, { format, size });
```
AustrianDev, line 54: Wouldn't the endpoint be `members` instead of `users` ?

ImRodry: Nope, refer to @GamingGeek’s review above: https://cdn.discordapp.com/guilds/692925064135311390/users/90339695967350784/avatars/67fcd6af88d76888b8525c5c3684b6b2.png?size=2048

xHyroM (approved):

Mysterious-Dev (approved):

ImRodry:
Upstream PR has finally been merged!

SpaceEEC (approved):

src/util/Constants.js
```diff
@@ -49,6 +49,10 @@ exports.Endpoints = {
         if (dynamic && hash.startsWith('a_')) format = 'gif';
         return makeImageUrl(`${root}/avatars/${userId}/${hash}`, { format, size });
       },
+      GuildMemberAvatar: (guildId, memberId, hash, format = 'webp', size, dynamic = false) => {
+        if (dynamic) format = hash.startsWith('a_') ? 'gif' : format;
```
GoldenAngel2, line 53:
```suggestion
        if (dynamic && hash.startsWith('a_')) format = 'gif';
```
So it matches the others.

GoldenAngel2 (changes requested):

src/util/Constants.js
ImRodry: Right, forgot I changed that

Sant268 (approved):

PR #5701, opened 2021-05-29
feat(Rest): show the data that is sent to Discord when an errors occurs 

vladfrangu:

Now you know what Discord means by your embed being wrong... or that color being invalid...Or just in general what we sent that caused the error you received

  - Above is a stretch, because this is internals

### DIFF
--- a/src/rest/DiscordAPIError.js
+++ b/src/rest/DiscordAPIError.js
@@ -5,7 +5,7 @@
  * @extends Error
  */
 class DiscordAPIError extends Error {
-  constructor(path, error, method, status) {
+  constructor(error, status, request) {
     super();
     const flattened = this.constructor.flattenErrors(error.errors || error).join('\n');
     this.name = 'DiscordAPIError';
@@ -15,13 +15,13 @@ class DiscordAPIError extends Error {
      * The HTTP method used for the request
      * @type {string}
      */
-    this.method = method;
+    this.method = request.method;
 
     /**
      * The path of the request relative to the HTTP endpoint
      * @type {string}
      */
-    this.path = path;
+    this.path = request.path;
 
     /**
      * HTTP error code returned by Discord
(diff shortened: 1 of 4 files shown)


iCrawl (approved):

src/rest/DiscordAPIError.js
```diff
@@ -34,6 +34,15 @@ class DiscordAPIError extends Error {
      * @type {number}
      */
     this.httpStatus = status;
+
+    /**
+     * The data associated with the request that caused this error
+     * @type {HTTPErrorData}
+     */
+    this.requestData = {
+      jsonData: request.options.data,
```
kyranet, line 43: Why not just call it `data`?

kyranet (changes requested):

src/rest/DiscordAPIError.js
vladfrangu: Because the entire thing is considered data, because if I repeat data one more time I will cry, and mostly to explicitly say that this is JSON data, not form data or such... If Crawl also wants it changed, I will alter accordingly 👍

kyranet: Thenn make the thing `error.data.body`, as opposed to duplicating data twice (`error.requestData.jsonData`)?

kyranet (approved):

src/structures/APIMessage.js
```diff
@@ -192,7 +193,7 @@ class APIMessage {
       tts,
       nonce,
       embed: this.options.embed === null ? null : embeds[0],
-      embeds,
+      embeds: isWebhook ? embeds : undefined,
```
ckohen, line 196: This needs isInteraction as well, kinda starting to feel like isWebhook should just include isInteraction

ckohen (changes requested):

src/structures/APIMessage.js
```diff
@@ -125,6 +125,7 @@ class APIMessage {
    */
   resolveData() {
     if (this.data) return this;
+    const isWebhook = this.isWebhook;
```
ImRodry, line 128: Why make this a constant?

iCrawl: Computing it 2x is more expensive.

src/structures/APIMessage.js
SpaceEEC: As per above: This currently breaks `interaction.reply({ embeds: [{description: "1"}, {description: "2"}] });`

SpaceEEC (dismissed):

SpaceEEC (approved):

PR #5710, opened 2021-05-30
ci(publish-dev): improve resolving of next version

Favna:

This removes the somewhat-hacky-looking regex and uses `jq` instead to parse the `package.json` file.

Tested the output with WSL:

image

Also tested in a private repo to validate the command in GitHub actions:

image

---

Original description:

~~While not strictly required in [Semantic Versioning rule 9](semver.org/#spec-item-9), based on their examples it seems strongly recommended to first add a hyphenated specification for pre-releases, ergo I added `-dev` before the Git SHA.~~

~~The addition of `--short=7` strips the Git SHA to the first 7 characters, which are the same as you would see when browsing commits on GitHub, making it a bit easier to match versions to their respective Git commit. AFAIK the first 7 characters are still pretty much always unique.~~

### DIFF
--- a/.github/workflows/publish-dev.yml
+++ b/.github/workflows/publish-dev.yml
@@ -21,7 +21,7 @@ jobs:
 
       - name: Publish
         run: |
-          npm version --git-tag-version=false $(cat package.json | grep -oP '"version": "\K[^"]+(?=")').$(git rev-parse --verify HEAD)
+          npm version --git-tag-version=false $(cat package.json | grep -oP '"version": "\K[^"]+(?=")')-dev.$(git rev-parse --verify --short=7 HEAD)
           npm publish --tag dev || true
         env:
           NODE_AUTH_TOKEN: ${{ secrets.NPM_PUBLISH_TOKEN }}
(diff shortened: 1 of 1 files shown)


Favna:
Converted to draft as I want to give refactoring to using `jq` a shot first as recommended in Discord by @kyranet

Nytelife26 (approved):
LGTM

kyranet (approved):

iCrawl:
> While not strictly required in Semantic Versioning rule 9, based on their examples it seems strongly recommended to first add a hyphenated specification for pre-releases, ergo I added -dev before the Git SHA.

This was purposefully left out, and will result in `13.0.0-dev-dev.SHA` now with your changes.

> The addition of --short=7 strips the Git SHA to the first 7 characters, which are the same as you would see when browsing commits on GitHub, making it a bit easier to match versions to their respective Git commit. AFAIK the first 7 characters are still pretty much always unique.

The reason I use the full hash is because its easier to remember the last few than the first 7 when giving support to people, and copy pasting a commit into GitHub this way is a lot easier too.

iCrawl (approved):
👍

vladfrangu (approved):
If Crawl's fine with it, I'm fine with it

PR #5716, opened 2021-05-31
feat: easier guards for channelUpdate

almostSouji:

Below guard fails on `channelUpdate` client events, however works flawlessly on `channelDelete`. By adapting the typings to `DMChannel | GuildChannel` the types are much easier guardable but still encompass all channel types.

```ts
if (channel instanceof DMChannel || !channel.isText()) return;
```

### DIFF
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -2603,7 +2603,7 @@ declare module 'discord.js' {
     applicationCommandUpdate: [oldCommand: ApplicationCommand | null, newCommand: ApplicationCommand];
     channelCreate: [channel: GuildChannel];
     channelDelete: [channel: DMChannel | GuildChannel];
-    channelPinsUpdate: [channel: Channel | PartialDMChannel, date: Date];
+    channelPinsUpdate: [channel: TextChannel | NewsChannel | DMChannel | PartialDMChannel, date: Date];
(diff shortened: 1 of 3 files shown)


typings/index.d.ts
```diff
@@ -2604,7 +2604,7 @@ declare module 'discord.js' {
     channelCreate: [channel: GuildChannel];
     channelDelete: [channel: DMChannel | GuildChannel];
     channelPinsUpdate: [channel: Channel | PartialDMChannel, date: Date];
```
iShibi, line 2606:
```suggestion
    channelPinsUpdate: [channel: TextChannel | NewsChannel | PartialDMChannel, date: Date];
```
While we are at it, I found some more typing issues of the same nature. This is one of them, the other one is the `typingStart` event but github won't let me pick that.

almostSouji: superseded by internal discussion applied in 8143f56 769489a

iCrawl (approved):

SpaceEEC (approved):

typings/index.d.ts
iShibi: I see. btw the jsdoc description for these two events are a little inconsistent. Like for `channelPinsUpdate` it says `DMChannel` or `TextChannel` while for `typingStart` it is generic `channel`. As both of these events have same typings they should have same jsdoc too

PR #5717, opened 2021-05-31
types: use Snowflake type

iCrawl:

### DIFF
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -245,7 +245,7 @@ declare module 'discord.js' {
 
   export class BaseGuildEmoji extends Emoji {
     constructor(client: Client, data: unknown, guild: Guild);
-    private _roles: string[];
+    private _roles: Snowflake[];
 
     public available: boolean | null;
     public readonly createdAt: Date;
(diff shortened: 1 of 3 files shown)


SpaceEEC (approved):

typings/index.d.ts
```diff
@@ -80,7 +81,7 @@ declare module 'discord.js' {
   //#region Classes
 
   export class Activity {
-    constructor(presence: Presence, data?: object);
+    constructor(presence: Presence, data?: unknown);
```
vladfrangu, line 84: I mean at this point we might as well go deeper and use the typings here, no? We don't need to `unknown` them

iCrawl: We need to merge this regardless because of the Snowflake mismatch we already have when someone is using master.

But yes, this opens the possibility of filling this out further (not really to anyone's benefit though since... you don't really instantiate those yourself).

vladfrangu (approved):

kyranet (approved):

typings/index.d.ts
```diff
@@ -243,7 +244,7 @@ declare module 'discord.js' {
   }
 
   export class BaseGuildEmoji extends Emoji {
-    constructor(client: Client, data: object, guild: Guild);
+    constructor(client: Client, data: unknown, guild: Guild);
     private _roles: string[];
```
vaporox, line 248:
```suggestion
    private _roles: Snowflake[];
```

PR #5720, opened 2021-05-31
refactor(GuildPreviewEmoji): make roles an array

vaporox:

This PR changes the type of `GuildPreviewEmoji#roles` from `Set<Snowflake>` to `Snowflake[]` because pretty much no other getter/method returns a set.

### DIFF
--- a/src/structures/GuildPreviewEmoji.js
+++ b/src/structures/GuildPreviewEmoji.js
@@ -13,13 +13,14 @@ class GuildPreviewEmoji extends BaseGuildEmoji {
    * @name GuildPreviewEmoji#guild
    */
 
-  /**
-   * Set of roles this emoji is active for
-   * @type {Set<Snowflake>}
-   * @readonly
-   */
-  get roles() {
-    return new Set(this._roles);
+  constructor(client, data, guild) {
+    super(client, data, guild);
+
+    /**
+     * The roles this emoji is active for
+     * @type {Snowflake[]}
+     */
+    this.roles = data.roles;
   }
 }
 
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -248,7 +248,7 @@ declare module 'discord.js' {
   }
 
   export class BaseGuildEmoji extends Emoji {
-    constructor(client: Client, data: unknown, guild: Guild);
+    constructor(client: Client, data: unknown, guild: Guild | GuildPreview);
     public available: boolean | null;
     public readonly createdAt: Date;
     public readonly createdTimestamp: number;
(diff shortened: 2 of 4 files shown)


typings/index.d.ts
```diff
@@ -1023,7 +1023,7 @@ declare module 'discord.js' {
   export class GuildPreviewEmoji extends BaseGuildEmoji {
     constructor(client: Client, data: object, guild: GuildPreview);
     public guild: GuildPreview;
-    public readonly roles: Set<Snowflake>;
+    public readonly roles: Snowflake[];
```
kyranet, line 1026: It might be sensible to mark the array as read-only to discourage mutations to it.
```suggestion
    public readonly roles: readonly Snowflake[];
```

kyranet (changes requested):

src/structures/GuildPreviewEmoji.js
```diff
@@ -14,12 +14,12 @@ class GuildPreviewEmoji extends BaseGuildEmoji {
...
   /**
-   * Set of roles this emoji is active for
-   * @type {Set<Snowflake>}
+   * The roles this emoji is active for
+   * @type {Snowflake[]}
    * @readonly
    */
   get roles() {
-    return new Set(this._roles);
+    return this._roles;
```
vladfrangu, line 22: I say we should return a slice... So people don't modify the value

vladfrangu (changes requested):
This and kyras comment

src/structures/GuildPreviewEmoji.js
vaporox: I think I actually prefer cloning the array too, but why should it still be kept as readonly then? Wouldn't that just make working with it unnecessarily harder?

vladfrangu: `Guild#features` is also not a getter, but a property. Since this is a getter, it's readonly (can only read, not write technically), but because we don't return the slice of the array it's "readonly but only to the class, not overall" which truly makes this confusing. I suggest we bring this in line with `Guild#features`

vaporox: > `Guild#features` is also not a getter, but a property. Since this is a getter, it's readonly (can only read, not write technically)

The property on the class is readonly, yes, but it's also marked as that with the current changes:
```ts
public readonly roles: Snowflake[];
          ^
```
Adding another `readonly` there would mean that it is frozen or cannot be modified because of other reasons, which isn't really true. If it returned a clone instead, there wouldn't be any problems with the user modifying it so I don't see the point of it there

<br>

> I suggest we bring this in line with `Guild#features`, if possible

Bring in line how exactly? Sorry if I'm not following

vaporox:
I decided to take a completely different approach and just move `_roles` from `BaseGuildEmoji` to `GuildEmoji` so that it can just be a normal property instead of a getter. Not sure if this better than the previous approach but I think that it makes more sense like this.

kyranet (approved):

vladfrangu (approved):

SpaceEEC (approved):

iCrawl (approved):

PR #5722, opened 2021-06-01
types(Snowflake): export the type

monbrey:
This re-exports discord-api-types Snowflake type so it can be imported from discord.js directly.
This will help improve the user experience for Typescript developers encountering the `${bigint}` type.


kakarot-dev:
so do we import snowflake? And how do we use it 🧐

iCrawl (approved):

PR #5723, opened 2021-06-01
fix: interfaces not importing due to re-export of Snowflake

iShibi:

This PR tries to fix the issue caused by re-export of `Snowflake` in #5722. The re-export of Snowflake made all the interfaces and types of discord.js non-importable. I tried to find a correct way of re-exporting types in a declaration file but couldn't find anything that helped. So, instead of importing the `Snowflake` type from `discord-api-types`, I defined the type in the lib itself just like other types. This kept the typings support that `Snowflake` introduced while also made it and other types importable again.

Thanks, @ToasterSticks for noticing a issue...again 😆

### DIFF
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -69,7 +69,6 @@ declare module 'discord.js' {
     APIMessage as RawMessage,
     APIOverwrite as RawOverwrite,
     APIRole as RawRole,
-    Snowflake,
   } from 'discord-api-types/v8';
   import { EventEmitter } from 'events';
   import { PathLike } from 'fs';
@@ -77,8 +76,6 @@ declare module 'discord.js' {
   import * as WebSocket from 'ws';
 
   export const version: string;
-
-  export { Snowflake };
 
   //#region Classes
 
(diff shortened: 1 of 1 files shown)


es183923 (approved):

kyranet:
What about doing this instead?
```typescript
export { Snowflake } from 'discord-api-types/v8';
```

Alongside the import, it's a pattern I used in my early days and it always worked.

iShibi:
I did try that (but only outside the `declare module`) while I was looking for possible fixes but doing so was somehow making the other typings modules like `@discordjs/collection` and `ws` not importable in the declaration file. Doing so inside the `declare module` alongside the import isn't showing this error but it also doesn't fix the issue. The only fix that I can see until the rewrite of declaration file is exporting all the interfaces and types explicitly.

monbrey:
Can't apply suggestions on the deleted lines, but
```js
    Snowflake as APISnowflake, // from api-types
    
    type Snowflake = APISnowflake; // inside module
```
This seems to resolve it for me

iCrawl (approved):

PR #5729, opened 2021-06-02
fix(ApplicationCommands): allow managing commands for uncached guilds

ckohen:

The way the library currently handles global and guild application commands is via a manager that exists on the application or guild respectively. This poses a problem for applications that have guilds authorized only with `applications.commands` as the guild is not cached and therefore does not have a guild application command manager.

This PR addresses this issue by allowing the global application command manger to take `guildID` in all of its methods. It also moves the permissions methods from `GuildApplicationCommandManager` to `ApplicationCommandManager` for the same reason.

It also introduces a new typedef that will be used in an upcoming PR for more consistency in managers. 🚀 

-  Application Commands fetched from `ApplicationCommandManager` never have `guild` even when they are guild commands. I haven't been able to come up with a super good way to document this, suggestions welcome.

### DIFF
--- a/src/managers/ApplicationCommandManager.js
+++ b/src/managers/ApplicationCommandManager.js
@@ -1,9 +1,10 @@
 'use strict';
 
 const BaseManager = require('./BaseManager');
-const { TypeError } = require('../errors');
+const { Error, TypeError } = require('../errors');
 const ApplicationCommand = require('../structures/ApplicationCommand');
 const Collection = require('../util/Collection');
+const { ApplicationCommandPermissionTypes } = require('../util/Constants');
 
 /**
  * Manages API methods for application commands and stores their cache.
@@ -26,14 +27,15 @@ class ApplicationCommandManager extends BaseManager {
 
   /**
    * The APIRouter path to the commands
-   * @type {Object}
-   * @readonly
+   * @param {Snowflake} [options.id] ID of the application command
+   * @param {Snowflake} [options.guildID] ID of the guild to use in the path, for uncached guilds
+   * @returns {Object}
    * @private
    */
-  get commandPath() {
+  commandPath({ id, guildID } = {}) {
     let path = this.client.api.applications(this.client.application.id);
-    if (this.guild) path = path.guilds(this.guild.id);
-    return path.commands;
+    if (this.guild || guildID) path = path.guilds(this.guild?.id ?? guildID);
+    return id ? path.commands(id) : path.commands;
   }
 
   /**
@@ -44,10 +46,22 @@ class ApplicationCommandManager extends BaseManager {
    */
 
   /**
+   * Options used to fetch data from discord
+   * @typedef {Object} BaseFetchOptions
+   * @property {boolean} [cache=true] Whether to cache the fetched data if it wasn't already
+   * @property {boolean} [force=false] Whether to skip the cache check and request the API
+   */
+
+  /**
+   * Options used to fetch Application Commands from discord
+   * @typedef {BaseFetchOptions} FetchApplicationCommandOptions
+   * @property {Snowflake} [guildID] ID of the guild to fetch commands for, for when the guild is not cached
+   */
+
+  /**
    * Obtains one or multiple application commands from Discord, or the cache if it's already available.
    * @param {Snowflake} [id] ID of the application command
-   * @param {boolean} [cache=true] Whether to cache the new application commands if they weren't already
-   * @param {boolean} [force=false] Whether to skip the cache check and request the API
+   * @param {FetchApplicationCommandOptions} [options] Additional options for this fetch
    * @returns {Promise<ApplicationCommand|Collection<Snowflake, ApplicationCommand>>}
    * @example
    * // Fetch a single command
@@ -60,23 +74,28 @@ class ApplicationCommandManager extends BaseManager {
    *   .then(commands => console.log(`Fetched ${commands.size} commands`))
    *   .catch(console.error);
    */
-  async fetch(id, cache = true, force = false) {
+  async fetch(id, { guildID, cache = true, force = false } = {}) {
+    if (typeof id === 'object') {
+      ({ guildID, cache = true, force = false } = id);
+      id = undefined;
+    }
     if (id) {
       if (!force) {
         const existing = this.cache.get(id);
         if (existing) return existing;
       }
-      const command = await this.commandPath(id).get();
+      const command = await this.commandPath({ id, guildID }).get();
       return this.add(command, cache);
     }
 
-    const data = await this.commandPath.get();
+    const data = await this.commandPath({ guildID }).get();
     return data.reduce((coll, command) => coll.set(command.id, this.add(command, cache)), new Collection());
   }
 
   /**
    * Creates an application command.
    * @param {ApplicationCommandData} command The command
+   * @param {Snowflake} [guildID] ID of a guild to create this command in, for when the guild is not cached
    * @returns {Promise<ApplicationCommand>}
    * @example
    * // Create a new command
@@ -87,8 +106,8 @@ class ApplicationCommandManager extends BaseManager {
    *   .then(console.log)
    *   .catch(console.error);
    */
-  async create(command) {
-    const data = await this.commandPath.post({
+  async create(command, guildID) {
+    const data = await this.commandPath({ guildID }).post({
       data: this.constructor.transformCommand(command),
     });
     return this.add(data);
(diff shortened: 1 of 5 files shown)


src/managers/ApplicationCommandManager.js
```diff
@@ -177,6 +199,139 @@ class ApplicationCommandManager extends BaseManager {
       default_permission: command.defaultPermission,
     };
   }
+
+  /**
+   * Fetches the permissions for one or multiple commands.
+   * <warn>When calling this on ApplicationCommandManager, guildID is required.
+   * To fetch all permissions for an uncached guild use `fetchPermissions(undefined, '123456789012345678')`</warn>
```
ckohen, line 206: I just wanted to make a note here about passing `undefined`. I don't love this pattern, but as Monbrey said in the discord, I don't love having to pass { command: '123456789012345678' } either.

monbrey: If the guild ID is the required one, put it first?

ckohen: when called from a GuildApplicationCommandManager its not required, although I suppose I could just use super in that class.

ckohen: Okay, I tried to do this, and it simply doesn't work as wanted.
Putting guildID as the first parameter means implementing `fetchPermissions` and `setPermissions` in `GuildApplicationCommandManager` which we can do in js, but because guildID is not a parameter there, typescript really doesn't like it, suggesting this is a bad pattern in general.

ckohen: Any other suggestions for this (or a suggestoin for how to make typescript happy with different method signatures)?

ckohen: `override` still doesn't let you remove / repurpose parameters in the signature though from what I can tell.

SpaceEEC (changes requested):

src/managers/ApplicationCommandManager.js
SpaceEEC: It's not nice, but you could allow either just a command resolvable, or an object with it and the guild.

ckohen: There is precedent for this elsewhere in the library actulaly, so it works, will do.

ckohen: A CommandResolvable can't be accepted as the only parameter because this.resolve (when providing an id) can't resolve guild application commands, which are required when not passing guildID so that we can get command.guild.id instead. And typescript will complain if I try to make it ApplicationCommand | { guildID: Snowflake | command?: ApplciationCommandResolvable } and then override that to ApplicationCommandResolvable in the subclass

iCrawl:
This needs a rebase.

src/managers/ApplicationCommandManager.js
zakuciael: Excuse a stranger, but if I've understood correctly the goal here is to somehow put the `guildId` parameter to the `fetchPermissions` method (possibly to more methods) and at the same time satisfy the TypeScript compiler in the `GuildApplicationCommandManager`, right?

So I've created a small "demo" of my approach on the TS Playground [here](https://www.typescriptlang.org/play?#code/PTAEGUBcFcCNQIIAckBsCWBjAhpdB7AOwGF8BbM7QgE1E1WwGdGAoepxxFDHPI0ilVoBvFqHGgkcHqEgBTRpABcoRQCd0hAOYBuFmImYi66Jkj41ACgCUoURIeyAFukYA6eYtABeUACI-PQcAXxZQlkgATyQ5CEJ8AHcAMwYAa1jfdU1dfSiYrjQsXAISckoaACUFfFQAN2xYVAyCnmL+MqFQAB84xJTsdL082IBxaHRUamRC3hKBcuoqxhr6xubhUC1xyYBJahVweOS0uQAaOg6aFWnWvlLBSuq6hqbQYKHo2IB5JDvsVAAPAAVc4AaVAcgAHvIaJx0pF8ElQECAHw+UAABSwqQBGOwajw-2BKLBaIAZKAvmR0JBgaS9GwGMwWkU7vMhABZKjYLRyNR2AziKSNLCgJJySCYJwYvnU5glRiWLSofCwf5fNRjCbUdk0AD8KkFjhurLml0WT1Wrx6RocWsmJtm7QeFuWzzW3VAtokPz+gPtU24pudCyWKxeZ38RhdfhRRusKlq+HQIm94hAoGo+FU5AlLm0oCcfLkRtC4XYzIDjra9wWXMIPL5EOhclhLKdtc53N5-PsEmFMnFkulstcjAVlmjCwN7Zrutd4bWCdASZTAscEgzWZzZDz2ULxdLYX0RkIXmVqv+9cb-N8hDkCVnbPN157NgZF7VqFffLcQ6lMpqHK47GO+LCfle3a-v+I5AWOE5+AAjAATAAzH41h6BB35QWof4SgBo7yqB96PtWz4ujYmHgSqX4-nhMGAcBE4bFs2p7CoSFoX45xTkInEoehbzUdh9H4cOTHwaBrHbNQHH+IJPEXC6KikU+ZqUbYwQibRkENj24mEXBxFnpYMnsfsCnccJDKnuesn0eialVkGHbzvRYEsBmADqFipOBDm4YZsHMaBImBfp0EESFUmmVx6HhdqYmMURIGmWp5EaQsVEMhmAAiRCQKACR+QFSVBSlxlpYq5m7JZ8VKXxVxWUJ2lYRFN7BZJJk1ZssnyQ1vHmqpD7qSGQg2DZZWTMl0XddVZl9RZAnWW1QA) and with this solution you can pass all combinations: **no params**, **only command** and **guild id + previous combinations**. I've based it on the @SpaceEEC suggestion to put either a resolvable or an object with `guildId` and the command.

(I know that I wrote it entirely in TypeScript, and it may or may not be 100% applicable but maybe at the very least it will notch you guys in the right direction, hope I've helped)

src/managers/ApplicationCommandManager.js
```diff
@@ -177,6 +204,139 @@ class ApplicationCommandManager extends BaseManager {
...
+   * @param {ApplicationCommandResolvable} [command] The command to get the permissions from
+   * @param {Snowflake} [guildID] ID of the guild to get the permissions for,
+   * for global commands and when the guild is not cached
```
SpaceEEC, line 214: Looks like I missed two instances of this:
```suggestion
   * ignored when using a {@link GuildApplicationCommandManager}
```

src/managers/ApplicationCommandManager.js
```diff
@@ -177,6 +204,139 @@ class ApplicationCommandManager extends BaseManager {
...
+
+  /**
+   * Sets the permissions for a command.
+   * <warn>When calling this on ApplicationCommandManager, guildID is required.
+   * To set multiple permissions for an uncached guild use `setPermissions(permissions, '123456789012345678')`</warn>
+   * @param {ApplicationCommandResolvable|GuildApplicationCommandPermissionData[]} command The command to edit the
+   * permissions for, or an array of guild application command permissions to set the permissions of all commands
+   * @param {ApplicationCommandPermissionData[]} [permissions] The new permissions for the command
+   * @param {Snowflake} [guildID] ID of the guild to get the permissions for,
+   * for global commands and when the guild is not cached
```
SpaceEEC, line 263: And number two:
```suggestion
   * ignored when using a {@link GuildApplicationCommandManager}
```

src/structures/ApplicationCommand.js
```diff
@@ -148,23 +147,24 @@ class ApplicationCommand extends Base {
 
   /**
    * Fetches the permissions for this command.
-   * <warn>This is only available for guild application commands.</warn>
+   * <warn>You must specify guildID on global commands and commands on uncached guilds, otherwise it is ignored.</warn>
+   * @param {Snowflake} [guildID] ID for the guild to fetch permissions for if this is a global command
```
SpaceEEC, line 151: I don't think you can get a global command / command for an arbitrary (potentially uncached) guild without the `ApplicationCommandManager`, right?

```suggestion
   * <warn>Required if this application command is handled by a {@link ApplicationCommandManager}, otherwise ignored.</warn>
   * @param {Snowflake} [guildID] ID for the guild to fetch permissions for
```

If this suggestion is more confusing or maybe not as precise, the current change can probably stay.
The `uncached` part just irks me a bit. 😅 Maybe it could be replaced with `arbitrary`?

src/structures/ApplicationCommand.js
```diff
@@ -148,23 +147,24 @@ class ApplicationCommand extends Base {
...
-    if (!this.guild) throw new Error('GLOBAL_COMMAND_PERMISSIONS');
-    return this.manager.fetchPermissions(this);
+  fetchPermissions(guildID) {
+    return this.manager.fetchPermissions(this, guildID);
   }
 
   /**
    * Sets the permissions for this command.
-   * <warn>This is only available for guild application commands.</warn>
+   * <warn>You must specify guildID on global commands and commands on uncached guilds, otherwise it is ignored.</warn>
```
SpaceEEC, line 165: Similar here then.

SpaceEEC (dismissed):
Otherwise LGTM

iCrawl:
This needs a rebase on top of the changes requested.

SpaceEEC (approved):

iCrawl (approved):

PR #5705, opened 2021-05-30
feat(CommandInteraction): make options a collection

iShibi:
**Please describe the changes this PR makes and why it should be merged:**

Currently, `CommandInteraction#options` is an Array. This PR changes it to a Collection. As [discussed](https://discord.com/channels/222078108977594368/682166281826598932/848333889986101269) in the support server, `options` of a `CommandInteraction` have unique names. Thus, having a Collection with names as keys is much better than using an array in this case.

**Status and versioning classification:**
- Code changes have been tested against the Discord API, or there are no code changes
- I know how to update typings and have done so, or typings don't need updating
- This PR changes the library's interface (methods or parameters added)
- This PR includes breaking changes (methods removed or renamed, parameters moved or removed)

PR #5734, opened 2021-06-02
fix(CommandInteraction): cmds with no options throw error

iShibi:

*cough* So, I totally forgot about slash cmds with no options in #5705. The `options` param in `CommandInteraction#_createOptionsCollection` would be `undefined` for such cmds and thus iterating over it will throw error. This PR fixes the issue by returning an empty collection when `options` is undefined. I did think of making `CommandInteraction#options` null at first:
```js
this.options = data.data?.options ? this._createOptionsCollection(data.data.options, data.data.resolved) : null;
```
but then I saw that before #5705, it was set to an empty array for cmds that have no options, so I went with the empty collection fix.

### DIFF
--- a/src/structures/CommandInteraction.js
+++ b/src/structures/CommandInteraction.js
@@ -261,6 +261,7 @@ class CommandInteraction extends Interaction {
    */
   _createOptionsCollection(options, resolved) {
     const optionsCollection = new Collection();
+    if (typeof options === 'undefined') return optionsCollection;
     for (const option of options) {
       optionsCollection.set(option.name, this.transformOption(option, resolved));
     }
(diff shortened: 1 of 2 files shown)


iShibi:
The typings for `CommandInterationOption#options` were also not updated, fixed it in the above commit. No, idea what's up with the changes that linter made tho

Yan-Jobs (approved):
This will fix an error who is throw when an interaction is received.

Yan-Jobs:
This will fix an error who is throw when an interaction is received.

Yan-Jobs (commented):
This will fix an error who is throw when an interaction is received.

kyranet (approved):

SpaceEEC (approved):

PR #5737, opened 2021-06-03
types(*): change remaining string IDs to Snowflakes

ImRodry:
After #5717 was merged, a few ids kept the `string` type, despite them actually being Snowflakes. This PR changes all remaining string ids to Snowflake in the situations where I could verify that they were actually Snowflakes. See the discussion [here](https://canary.discord.com/channels/222078108977594368/824411059443204127/849945366219587604)

### DIFF
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -438,7 +438,7 @@ declare module 'discord.js' {
   export class CommandInteraction extends Interaction {
     public readonly command: ApplicationCommand | null;
     public channel: TextChannel | DMChannel | NewsChannel;
-    public commandID: string;
+    public commandID: Snowflake;
     public commandName: string;
     public deferred: boolean;
     public options: Collection<string, CommandInteractionOption>;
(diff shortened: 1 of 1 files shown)


typings/index.d.ts
```diff
@@ -3111,7 +3111,7 @@ declare module 'discord.js' {
   }
 
   interface IntegrationAccount {
-    id: string;
+    id: string | Snowflake;
```
ImRodry, line 3114: I didn't make this a Snowflake because the API docs still classify it as a string, however, it can be a Snowflake so I added the type. Docs for reference: https://discord.com/developers/docs/resources/guild#integration-account-object

ImRodry:
This also fixes a bug that occurred when passing an id from a partial user to a method that required a snowflake, so it should get the bugfix label
Something like this:
```ts
client.on("messageReactionAdd", (reaction, user) => {
  client.users.cache.get(user.id)
})```

typings/index.d.ts
kyranet: I think this should be reverted to just `string`, to keep consistency with Discord's docs.

Besides, `string | Snowflake` is just `string`, since the latter is a sub-type of the former.

ImRodry: Would it hurt to have it there? It would just serve as an indication saying that this id can sometimes be a snowflake. I totally understand the consistency argument though and I can change it for that reason.

kyranet: After a quick chat with the team, the consensus is to leave this as is, so I'll approve now, thanks for your contribution :)

kyranet (approved):

SpaceEEC (approved):

typings/index.d.ts
ImRodry: Alright, you're welcome!

iCrawl (approved):

PR #5742, opened 2021-06-03
fix(Typings): add missing typings for `HttpError` -> `requestData`

Dev-CasperTheGhost:

This PR adds missing typings from #5701


typings/index.d.ts
```diff
@@ -3083,6 +3085,16 @@ declare module 'discord.js' {
     cache?: boolean;
   }
 
+  interface HTTPAttachmentData {
+    name: string;
+    file: Buffer;
+  }
+
+  interface HTTPErrorData {
+    json: any;
```
SpaceEEC, line 3094: I assume we'd prefer `unknown` here.

vladfrangu (dismissed):
The constructors need to be updated

Dev-CasperTheGhost:
> The constructors need to be updated

Is there any `request` type or should I create one? @vladfrangu

Dev-CasperTheGhost:
> The constructors need to be updated

Done

typings/index.d.ts
```diff
@@ -3083,6 +3085,20 @@ declare module 'discord.js' {
     cache?: boolean;
   }
 
+  interface HTTPAttachmentData {
+    name: string;
+    file: Buffer;
```
SpaceEEC, line 3090: This can also be a stream:
```suggestion
    file: Buffer | Stream;
```
The source for that is a bit hidden:
https://github.com/discordjs/discord.js/blob/99e2ca408b0ead894c2d91bc2902c392a3635fb0/src/util/DataResolver.js#L106-L111

typings/index.d.ts
```diff
@@ -3083,6 +3085,20 @@ declare module 'discord.js' {
     cache?: boolean;
   }
 
+  interface HTTPAttachmentData {
+    name: string;
```
SpaceEEC, line 3089: This is missing a property:
```suggestion
    attachment: string;
    name: string;
```
See: https://github.com/discordjs/discord.js/blob/99e2ca408b0ead894c2d91bc2902c392a3635fb0/src/structures/APIMessage.js#L303

iCrawl (approved):

typings/index.d.ts
```diff
@@ -703,13 +703,14 @@ declare module 'discord.js' {
   }
 
   export class DiscordAPIError extends Error {
-    constructor(path: string, error: unknown, method: string, httpStatus: number);
+    constructor(error: unknown, status: number, request: HTTPRequest);
```
SpaceEEC, line 706: I don't think there is much point in documenting parts of our APIRequest class:
```suggestion
    constructor(error: unknown, status: number, request: unknown);
```

vladfrangu (approved):

typings/index.d.ts
```diff
@@ -1045,11 +1046,12 @@ declare module 'discord.js' {
   }
 
   export class HTTPError extends Error {
-    constructor(message: string, name: string, code: number, method: string, path: string);
+    constructor(message: string, name: string, code: number, request: HTTPRequest);
```
SpaceEEC, line 1049:
```suggestion
    constructor(message: string, name: string, code: number, request: unknown);
```

typings/index.d.ts
```diff
@@ -3083,6 +3085,20 @@ declare module 'discord.js' {
...
+
+  interface HTTPErrorData {
+    json: unknown;
+    files: HTTPAttachmentData[];
+  }
+
+  interface HTTPRequest {
+    options: HTTPErrorData;
+  }
+
```
SpaceEEC, line 3101:
```suggestion
```

SpaceEEC (changes requested):

src/rest/HTTPError.js
```diff
@@ -42,8 +42,9 @@ class HTTPError extends Error {
     /**
      * The attachment data that is sent to Discord
      * @typedef {Object} HTTPAttachmentData
+     * @property {string} attachment The attachment
```
Dev-CasperTheGhost, line 45: Probably should have a better description, wasn't sure what to put 😅.

SpaceEEC: It looks like that's the source for the attachment data (so it could be a stream or buffer too).
So maybe something like this? (Would also need updating in the typings again 😅 )
```suggestion
     * @property {string|Buffer|Stream} attachment The source of this attachment data
```

---

Assigned from `fileLike` (what was passed as file to djs to send)
https://github.com/discordjs/discord.js/blob/99e2ca408b0ead894c2d91bc2902c392a3635fb0/src/structures/APIMessage.js#L295-L298

Dev-CasperTheGhost: Done 😄

SpaceEEC (approved):
Thank you for this PR!

kyranet (approved):

PR #5752, opened 2021-06-04
types: extend MessageComponentInteractionCollector correctly

monbrey:
Fixes a bug in the typings where MessageComponentInteractionCollector didn't pass type to the base Collector correctly.

### DIFF
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -1374,17 +1374,23 @@ declare module 'discord.js' {
 
     public collect(interaction: Interaction): Snowflake;
     public dispose(interaction: Interaction): Snowflake;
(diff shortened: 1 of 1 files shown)


SpaceEEC (dismissed):
You also need to change the type in the `on` and `once` overloads.
https://github.com/discordjs/discord.js/blob/3dd139359276f19cf0818cce4c0697cf5356b377/typings/index.d.ts#L1377-L1389

kyranet (approved):

SpaceEEC (approved):

ReelJohn:
What is my over all objective and can real or unity be a part

PR #5756, opened 2021-06-05
feat(Sharding*): contexts for broadcastEval

amishshah:

This allows developers to write safer eval code by allowing them to pass functions and contexts to `broadcastEval` in both `ShardClientUtil` and `ShardingManager`.

This avoids directly injecting malicious code into your eval scripts. Instead, you can treat your functions as templates (which you can make sure are safe) which are then injected with a context later on.

For example, the following code is vulnerable (taken and adapted from https://github.com/oadpoaw/discordjs-bot-rce). It's important to recognise that this isn't a security issue with discord.js -- using eval is inherently dangerous, and this is the clearest case of passing unsanitised input to eval.

```js
function findUser(userId) {
  return this.users.cache.get(userId);
}

const unsafe = `');process.exit();a=('`; // causes the shard to exit
const results = await client.shard.broadcastEval(`(${findUser}).call(this, '${unsafe}')`);
```

While this isn't a security issue with discord.js, we can definitely make it much easier to write safer code. The following code snippet is not vulnerable.

```js
function findUser(client, { userId }) {
  return this.users.cache.get(userId);
}

const unsafe = `');process.exit();a=('`;
const results = await client.shard.broadcastEval(findUser, { context: { userId: unsafe }});
```

It's important to note that the context you pass (e.g. in the above example, the context is `{ userId: unsafe }` must be JSON-serializable. This means you can't pass complex data types in the context directly.

This PR removes support for calling `broadcastEval` with strings to enforce better practice with security.

### DIFF
--- a/src/sharding/ShardClientUtil.js
+++ b/src/sharding/ShardClientUtil.js
@@ -139,6 +139,10 @@ class ShardClientUtil {
   broadcastEval(script, options = {}) {
     return new Promise((resolve, reject) => {
       const parent = this.parentPort || process;
+      if (typeof script !== 'function') {
+        reject(new TypeError('SHARDING_INVALID_EVAL_BROADCAST'));
+        return;
+      }
       script = `(${script})(this, ${JSON.stringify(options.context)})`;
 
       const listener = message => {
--- a/src/sharding/ShardClientUtil.js
+++ b/src/sharding/ShardClientUtil.js
@@ -129,6 +129,7 @@ class ShardClientUtil {
    * Evaluates a script or function on all shards, or a given shard, in the context of the {@link Client}s.
    * @param {string|Function} script JavaScript to run on each shard
    * @param {number} [shard] Shard to run script on, all if undefined
+   * @param {Object} [parameterList={}] The JSON-stringifiable parameters to call the function with
    * @returns {Promise<*>|Promise<Array<*>>} Results of the script execution
    * @example
    * client.shard.broadcastEval('this.guilds.cache.size')
@@ -136,10 +137,10 @@ class ShardClientUtil {
    *   .catch(console.error);
    * @see {@link ShardingManager#broadcastEval}
    */
-  broadcastEval(script, shard) {
+  broadcastEval(script, shard, parameterList = {}) {
     return new Promise((resolve, reject) => {
       const parent = this.parentPort || process;
-      script = typeof script === 'function' ? `(${script})(this)` : script;
+      script = typeof script === 'function' ? `(${script})(this, ${JSON.stringify(parameterList)})` : script;
 
       const listener = message => {
         if (!message || message._sEval !== script || message._sEvalShard !== shard) return;
--- a/src/sharding/ShardClientUtil.js
+++ b/src/sharding/ShardClientUtil.js
@@ -127,11 +127,11 @@ class ShardClientUtil {
 
   /**
    * Evaluates a script or function on all shards, or a given shard, in the context of the {@link Client}s.
-   * @param {string|Function} script JavaScript to run on each shard
+   * @param {Function} script JavaScript to run on each shard
    * @param {BroadcastEvalOptions} [options={}] The options for the broadcast
    * @returns {Promise<*>|Promise<Array<*>>} Results of the script execution
    * @example
-   * client.shard.broadcastEval('this.guilds.cache.size')
+   * client.shard.broadcastEval(client => client.guilds.cache.size)
    *   .then(results => console.log(`${results.reduce((prev, val) => prev + val, 0)} total guilds`))
    *   .catch(console.error);
    * @see {@link ShardingManager#broadcastEval}
@@ -139,7 +139,7 @@ class ShardClientUtil {
   broadcastEval(script, options = {}) {
     return new Promise((resolve, reject) => {
       const parent = this.parentPort || process;
-      script = typeof script === 'function' ? `(${script})(this, ${JSON.stringify(options.context)})` : script;
+      script = `(${script})(this, ${JSON.stringify(options.context)})`;
 
       const listener = message => {
         if (!message || message._sEval !== script || message._sEvalShard !== options.shard) return;
--- a/src/sharding/ShardClientUtil.js
+++ b/src/sharding/ShardClientUtil.js
@@ -128,8 +128,7 @@ class ShardClientUtil {
   /**
    * Evaluates a script or function on all shards, or a given shard, in the context of the {@link Client}s.
    * @param {string|Function} script JavaScript to run on each shard
-   * @param {number} [shard] Shard to run script on, all if undefined
-   * @param {Object} [parameterList={}] The JSON-stringifiable parameters to call the function with
+   * @param {BroadcastEvalOptions} [options={}] The options for the broadcast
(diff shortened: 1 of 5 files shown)


src/sharding/ShardClientUtil.js
```diff
@@ -128,28 +128,28 @@ class ShardClientUtil {
...
    *   .then(results => console.log(`${results.reduce((prev, val) => prev + val, 0)} total guilds`))
    *   .catch(console.error);
    * @see {@link ShardingManager#broadcastEval}
    */
-  broadcastEval(script, shard) {
+  broadcastEval(script, options = {}) {
     return new Promise((resolve, reject) => {
       const parent = this.parentPort || process;
-      script = typeof script === 'function' ? `(${script})(this)` : script;
+      script = typeof script === 'function' ? `(${script})(this, ${JSON.stringify(options.context ?? {})})` : script;
```
kyranet, line 142: `null` and `undefined` (both being falsy values) defaulting to an empty object (being truthy) might cause some headaches in the developer's side, and `JSON.stringify` allows both as a value, so why not let them be passed?
```suggestion
      script = typeof script === 'function' ? `(${script})(this, ${JSON.stringify(options.context)})` : script;
```

src/sharding/ShardingManager.js
```diff
@@ -226,14 +226,25 @@ class ShardingManager extends EventEmitter {
...
+   * @param {string|Function} script JavaScript to run on each shard
+   * @param {BroadcastEvalOptions} [options={}] The options for the broadcast
    * @returns {Promise<*>|Promise<Array<*>>} Results of the script execution
    */
-  broadcastEval(script, shard) {
-    return this._performOnShards('eval', [script], shard);
+  broadcastEval(script, options = {}) {
+    return this._performOnShards(
+      'eval',
+      [typeof script === 'function' ? `(${script})(this, ${JSON.stringify(options.context ?? {})})` : script],
```
kyranet, line 245: Same as above:
```suggestion
      [typeof script === 'function' ? `(${script})(this, ${JSON.stringify(options.context)})` : script],
```

kyranet (changes requested):

iCrawl (approved):

vladfrangu (approved):

kyranet (approved):

SpaceEEC:
Is there something speaking against requiring a function to be passed?
I fear this otherwise might go unnoticed / continues to be an issue with snippets / examples floating around.

amishshah:
> Is there something speaking against requiring a function to be passed?
> I fear this otherwise might go unnoticed / continues to be an issue with snippets / examples floating around.

Updated the PR with this suggestion! :smile:

vladfrangu (approved):

kyranet (approved):

src/sharding/ShardClientUtil.js
```diff
@@ -127,29 +127,29 @@ class ShardClientUtil {
...
    * @returns {Promise<*>|Promise<Array<*>>} Results of the script execution
    * @example
-   * client.shard.broadcastEval('this.guilds.cache.size')
+   * client.shard.broadcastEval(client => client.guilds.cache.size)
    *   .then(results => console.log(`${results.reduce((prev, val) => prev + val, 0)} total guilds`))
    *   .catch(console.error);
    * @see {@link ShardingManager#broadcastEval}
    */
-  broadcastEval(script, shard) {
+  broadcastEval(script, options = {}) {
```
SpaceEEC, line 139: I think it would be beneficial to validate that `script` is actually a function.
Otherwise you get some confusing / cryptic errors when attempting to pass strings like before.

src/sharding/ShardingManager.js
```diff
@@ -226,14 +226,21 @@ class ShardingManager extends EventEmitter {
...
    * Evaluates a script on all shards, or a given shard, in the context of the {@link Client}s.
-   * @param {string} script JavaScript to run on each shard
-   * @param {number} [shard] Shard to run on, all if undefined
+   * @param {Function} script JavaScript to run on each shard
+   * @param {BroadcastEvalOptions} [options={}] The options for the broadcast
    * @returns {Promise<*>|Promise<Array<*>>} Results of the script execution
    */
-  broadcastEval(script, shard) {
-    return this._performOnShards('eval', [script], shard);
+  broadcastEval(script, options = {}) {
```
SpaceEEC, line 242: Here too

SpaceEEC (changes requested):

src/sharding/ShardClientUtil.js
kyranet: Should we also check for whether or not the function's body is native code? Such as `Number.isNaN` or `cb.bind(null)` both stringify to `function NameOfFunction() { [native code] }`.

src/sharding/ShardingManager.js
```diff
@@ -226,14 +226,22 @@ class ShardingManager extends EventEmitter {
...
+    if (typeof script !== 'function') throw new TypeError('SHARDING_INVALID_EVAL_BROADCAST');
```
SpaceEEC, line 243: Since the method is supposed to return a promise:
```suggestion
    if (typeof script !== 'function') return Promise.reject(new TypeError('SHARDING_INVALID_EVAL_BROADCAST'));
```

SpaceEEC (dismissed):

typings/index.d.ts
```diff
@@ -1695,10 +1695,8 @@ declare module 'discord.js' {
     public readonly ids: number[];
     public mode: ShardingManagerMode;
     public parentPort: any | null;
-    public broadcastEval(script: string): Promise<any[]>;
-    public broadcastEval(script: string, shard: number): Promise<any>;
-    public broadcastEval<T>(fn: (client: Client) => T): Promise<T[]>;
-    public broadcastEval<T>(fn: (client: Client) => T, shard: number): Promise<T>;
+    public broadcastEval<T, P>(fn: (client: Client, context: P) => T, { shard: undefined, context: P }: BroadcastEvalOptions): Promise<T[]>;
```
SpaceEEC, line 1698: Since the parameter is optional:
```suggestion
    public broadcastEval<T, P>(fn: (client: Client, context: P) => T, { shard: undefined, context: P }?: BroadcastEvalOptions): Promise<T[]>;
```

typings/index.d.ts
```diff
@@ -1721,8 +1719,8 @@ declare module 'discord.js' {
     public totalShards: number | 'auto';
     public shardList: number[] | 'auto';
     public broadcast(message: any): Promise<Shard[]>;
-    public broadcastEval(script: string): Promise<any[]>;
-    public broadcastEval(script: string, shard: number): Promise<any>;
+    public broadcastEval<T, P>(fn: (client: Client, context: P) => T, { shard: undefined, context: P }: BroadcastEvalOptions): Promise<T[]>;
```
SpaceEEC, line 1722: Same here:
```suggestion
    public broadcastEval<T, P>(fn: (client: Client, context: P) => T, { shard: undefined, context: P }?: BroadcastEvalOptions): Promise<T[]>;
```

SpaceEEC (changes requested):
One more thing

SpaceEEC (approved):

PR #5762, opened 2021-06-05
fix(Interaction): add missing types and fix docs lists

### LINKED ISSUE #5751: Interaction#componentType
**Please describe the problem you are having in as much detail as possible:** The guide says to check `if (!interaction.isMessageComponent() && interaction.componentType !== 'BUTTON') return;` while listening for buttons, however there are no typings for `Interaction#componentType`

**Include a reproducible code sample here, if possible:** image

```js
// Place your code here
```

**Further details:**

- discord.js version: Dev Branch cbd7f2b9aa44a9240947ed716d0e72257ac499f7 
- Node.js version: 14.0
- Operating system: Ubuntu 20
- Priority this issue should have – please be realistic and elaborate if possible: Medium, as it may be required to distinguish between buttons and (?)

**Relevant client options:**

- partials: All
- gateway intents: All
- other: none

 cbd7f2b9aa44a9240947ed716d0e72257ac499f7

ImRodry:
This fixes #5751 as well as a bug in two typedefs that had their possible types listed incorrectly.
Decided to make this smaller PR to fix that issue as it can be quite annoying for users that are just now starting to try MessageComponents to have to wait for #5692 to be merged, as this will clearly take a while

### DIFF
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -1328,6 +1328,7 @@ declare module 'discord.js' {
   }
 
   export class MessageComponentInteraction extends Interaction {
+    public componentType: MessageComponentType | MessageComponentTypes;
     public customID: string;
     public deferred: boolean;
     public message: Message | RawMessage;
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -1328,7 +1328,7 @@ declare module 'discord.js' {
   }
 
   export class MessageComponentInteraction extends Interaction {
-    public componentType: MessageComponentType | MessageComponentTypes;
+    public componentType: MessageComponentType;
     public customID: string;
     public deferred: boolean;
     public message: Message | RawMessage;
--- a/src/util/Constants.js
+++ b/src/util/Constants.js
@@ -826,19 +826,19 @@ exports.InteractionResponseTypes = createEnum([
 
 /**
  * The type of a message component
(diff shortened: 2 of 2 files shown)


typings/index.d.ts
```diff
@@ -1328,6 +1328,7 @@ declare module 'discord.js' {
   }
 
   export class MessageComponentInteraction extends Interaction {
+    public componentType: MessageComponentType | MessageComponentTypes;
```
SpaceEEC, line 1331: This should always be the string, not integer:
```suggestion
    public componentType: MessageComponentType;
```
See:
https://github.com/discordjs/discord.js/blob/69bb5f911091e433ad3241d02b5beaca6426d433/src/structures/MessageComponentInteraction.js#L33
https://github.com/discordjs/discord.js/blob/69bb5f911091e433ad3241d02b5beaca6426d433/src/structures/MessageComponentInteraction.js#L60-L62

SpaceEEC (changes requested):

typings/index.d.ts
ImRodry: Ah I saw that but forgot to change it, my bad!

SpaceEEC (approved):

iCrawl (approved):

PR #5763, opened 2021-06-05
refactor(Interactions): move Structures import out of switch block

nordiago:

This PR moves two instances of a `Structures` import out of the `switch (data.type)` block in `INTERACTION_CREATE`, where they are assigned via unnecessary conditionals.

- Typings don't need updating

### DIFF
--- a/src/client/websocket/handlers/INTERACTION_CREATE.js
+++ b/src/client/websocket/handlers/INTERACTION_CREATE.js
@@ -1,22 +1,18 @@
 'use strict';
 
 const { Events, InteractionTypes } = require('../../../util/Constants');
-let Structures;
+const Structures = require('../../../util/Structures');
 
 module.exports = (client, { d: data }) => {
   let interaction;
   switch (data.type) {
     case InteractionTypes.APPLICATION_COMMAND: {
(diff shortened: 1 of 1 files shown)


monbrey:
Did you test this? The structures import was already inside the condition from the original Slash Commands PR, as is often done to prevent circular requires. I maintained the same location when adding the `MESSAGE_COMPONENT` conditional.

nordiago:
@monbrey I did test this, yes, and both commands and component interactions continued to work as before.

I don't _believe_ a circular dependency is likely in this case, but I could certainly have missed something. Still, it's probably not necessary to have two conditionals. Perhaps just one above the `switch` statement?

vladfrangu:
> Did you test this? The structures import was already inside the condition from the original Slash Commands PR, as is often done to prevent circular requires. I maintained the same location when adding the `MESSAGE_COMPONENT` conditional.

There's no way for this file to create a circular dependency with Structures. Then again, testing needed, but this PR LGTM rn

iCrawl (approved):

SpaceEEC (approved):

kyranet (approved):

vladfrangu (approved):

PR #5766, opened 2021-06-06
refactor(Guild): remove region from guild and manager methods

almostSouji:

Remove guild based region from guild data, methods and associated manager as per <https://github.com/discord/discord-api-docs/pull/3001> (guild based regions are no longer settable through the client and equal a noop in v8, will have to be removed with v9 so we might as well go ahead and do so prior)

⚠️ by nature of removal this is a breaking change

### DIFF
--- a/src/structures/Guild.js
+++ b/src/structures/Guild.js
@@ -6,7 +6,6 @@ const GuildPreview = require('./GuildPreview');
 const GuildTemplate = require('./GuildTemplate');
 const Integration = require('./Integration');
 const Invite = require('./Invite');
-const VoiceRegion = require('./VoiceRegion');
 const Webhook = require('./Webhook');
 const { Error, TypeError } = require('../errors');
 const GuildApplicationCommandManager = require('../managers/GuildApplicationCommandManager');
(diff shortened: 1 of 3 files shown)


iCrawl (approved):

kyranet (approved):

ImRodry:
Shouldn't these just be deprecated instead? Guilds that set their region previously will still have it in the guild object from the API so you could add a deprecation warning and remove it on the next version instead

src/structures/Guild.js
```diff
@@ -699,21 +692,6 @@ class Guild extends BaseGuild {
       });
   }
 
-  /**
-   * Fetches available voice regions.
-   * @returns {Promise<Collection<string, VoiceRegion>>}
-   */
-  fetchVoiceRegions() {
```
vladfrangu, line 706: This still returns data in api v9... Should we remove it?

ckohen: AFAIK, v9 isn't technically finalized, I think it should be removed now though so it doesn't sit as deprecated until v14.

SpaceEEC (approved):

src/structures/Guild.js
vladfrangu: That doesn't mean we should remove this function at all. We don't even know if it WILL be removed in API v9 (and considering VIP regions exist, my gut says this will remain), the only thing we know is removed is guild-level voice regions. This function should personally stay

PR #5767, opened 2021-06-06
fix(*): add missing imports for custom errors

iShibi:

Since, custom errors have same names as native errors, they get left un-imported sometimes. Which results in errors that don't have any message in it. This PR fixes it by importing these errors in places where they are being used but were not imported.

### DIFF
--- a/src/client/websocket/WebSocketManager.js
+++ b/src/client/websocket/WebSocketManager.js
@@ -3,7 +3,7 @@
 const EventEmitter = require('events');
 const WebSocketShard = require('./WebSocketShard');
 const PacketHandlers = require('./handlers');
-const { Error: DJSError } = require('../../errors');
+const { Error } = require('../../errors');
 const Collection = require('../../util/Collection');
 const { Events, ShardEvents, Status, WSCodes, WSEvents } = require('../../util/Constants');
 const Util = require('../../util/Util');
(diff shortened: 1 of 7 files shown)


SpaceEEC (dismissed):
I found one more like these:
https://github.com/discordjs/discord.js/blob/dec191aa1e4f22690285ca06c6eee7e6086b2930/src/client/websocket/WebSocketManager.js#L6
and
https://github.com/discordjs/discord.js/blob/dec191aa1e4f22690285ca06c6eee7e6086b2930/src/client/websocket/WebSocketManager.js#L316

Considering we don't seem to be using the global `Error` anywhere in there, the imported class shouldn't be renamed and all occurrences of `DJSError` should be changed to `Error`.

ckohen:
I don't actually see that error key in the messages file. Given it's only used to get the stack, it doesn't strictly need to be a DJSError but I so reason for it not to be. So just make sure to add the corresponding message

iShibi:
Done. btw I took an easy way out by just copying the same message. 😆

kyranet (approved):

SpaceEEC (approved):

iCrawl (approved):

vladfrangu:
> I don't actually see that error key in the messages file. Given it's only used to get the stack, it doesn't strictly need to be a DJSError but I see no reason for it not to be. So just make sure to add the corresponding message

Because there is no key for that error, in fact it could even be empty for all it matters :D

vladfrangu (approved):

Issue #5526, opened 2021-04-13
New Stage channels are causing Discord.js to crash on v11.6.4


icanovi:
**Discord bots using Discord.js version 11.6.4 are throwing a "TypeError: Cannot read property 'id' of undefined" on line 81 of the ClientDataManager.js file. New Stage channels seems to be the cause of it.**

**Include a reproducible code sample here, if possible:**

```js
1|npm -- run start:node  | /app/discord-bot/node_modules/discord.js/src/client/ClientDataManager.js:81
1|npm -- run start:node  |         guild.channels.set(channel.id, channel);
1|npm -- run start:node  |                                    ^
1|npm -- run start:node  | TypeError: Cannot read property 'id' of undefined
1|npm -- run start:node  |     at ClientDataManager.newChannel (/app/node_modules/discord.js/src/client/ClientDataManager.js:81:36)
1|npm -- run start:node  |     at Guild.setup (/app/node_modules/discord.js/src/structures/Guild.js:307:68)
1|npm -- run start:node  |     at GuildCreateHandler.handle (/app/node_modules/discord.js/src/client/websocket/packets/handlers/GuildCreate.js:12:15)
1|npm -- run start:node  |     at WebSocketPacketManager.handle (/app/node_modules/discord.js/src/client/websocket/packets/WebSocketPacketManager.js:108:65)
1|npm -- run start:node  |     at WebSocketConnection.onPacket (/app/node_modules/discord.js/src/client/websocket/WebSocketConnection.js:336:35)
1|npm -- run start:node  |     at WebSocketConnection.onMessage (/app/node_modules/discord.js/src/client/websocket/WebSocketConnection.js:299:17)
1|npm -- run start:node  |     at WebSocket.onMessage (/app/node_modules/ws/lib/event-target.js:120:16)
1|npm -- run start:node  |     at emitOne (events.js:116:13)
1|npm -- run start:node  |     at WebSocket.emit (events.js:211:7)
1|npm -- run start:node  |     at Receiver.receiverOnMessage (/app/node_modules/ws/lib/websocket.js:789:20
```

**Further details:**

- discord.js version: v11.6.4
- Node.js version: v8.11.2
- Operating system: Ubuntu 16.04.2 LTS
- Priority this issue should have – please be realistic and elaborate if possible:Medium

**Relevant client options:**

- partials: none
- gateway intents: none
- other: none

papaia:
Discord v11 is going out of support, this will not get help.
Update to v12.

[...]

PR #5775, opened 2021-06-06
add default case for new channel type: text

PixEye:
data.type may get a new value of 13. Set default new channel type to: text (it was blocking the connection of my bot)

I think the new channel type value (13) comes from the server side.

### DIFF
--- a/src/client/ClientDataManager.js
+++ b/src/client/ClientDataManager.js
@@ -76,6 +76,9 @@ class ClientDataManager {
           case Constants.ChannelTypes.STORE:
             channel = new StoreChannel(guild, data);
             break;
+          default:
+            // console.warn('Unexpected data type:', data.type); // could be 13 for example
+            channel = new TextChannel(guild, data);
         }
 
         guild.channels.set(channel.id, channel);


almostSouji:
- making assumptions about new channel types is a dangerous game to play (see stage channels).
- this targets v11 which we no longer support or develop

PixEye:
This patch is related to the **fatal** error: https://github.com/discordjs/discord.js/issues/5526
Yes, it is about your "old" v11 but I started my bot on October 2019 (less than 2 years ago) and I do not have time to change all the required things to update to your discord.js v12,... I've coded this bot on my spare time and it helps a lot of people (about 1300 registered users + many ponctual users). I am really disapointed at the moment...

kyranet:
You had over a year to migrate from v11 to v12, so that's kind of a bad excuse. If you desire to get further support, please use the [support server](https://discord.gg/djs).

Furthermore, also check [the guide](https://discordjs.guide/additional-info/changes-in-v12.html) for help with the migration.

Reiterating what Souji said, we do not support v11 and will not release new patches, so any change here will be pretty much dead in water, and the entire version is prone to completely die at any moment, as soon as the v6 gateway or discordapp.com API is killed, whichever happens first, and then you'll have a bot that will work as soon as you upgrade to v12, so better do that as soon as possible.

Finally, please refrain from using aggressive language as it can result on a violation of our [Code of Conduct](https://github.com/discordjs/discord.js/blob/master/.github/CODE_OF_CONDUCT.md) and may result on an user block in this repository or organization.

PixEye:
Thank you for the link to the (big) guide.
Sorry if you think I used "aggressive language", did I? I am French, my English is not perfect. 😟

PR #5781, opened 2021-06-08
feat(GuildAuditLogs): make #target a channel for channel related logs

iShibi:

Currently, for logs related to `CHANNEL_*` events, `GuildAuditLogsEntry#target` is an object containing only the id of the related channel. This PR turns this `target` prop into a channel using the data in the `changes` object. There are also some documentation related changes. 

### DIFF
--- a/src/structures/GuildAuditLogs.js
+++ b/src/structures/GuildAuditLogs.js
@@ -1,5 +1,6 @@
 'use strict';
 
+const Channel = require('./Channel');
 const Integration = require('./Integration');
 const Webhook = require('./Webhook');
 const Collection = require('../util/Collection');
@@ -79,7 +80,7 @@ const Targets = {
  * * INTEGRATION_CREATE: 80
  * * INTEGRATION_UPDATE: 81
  * * INTEGRATION_DELETE: 82
- * @typedef {?number|string} AuditLogAction
+ * @typedef {?(number|string)} AuditLogAction
  */
 
 /**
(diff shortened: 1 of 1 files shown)


src/structures/GuildAuditLogs.js
```diff
@@ -480,8 +483,22 @@ class GuildAuditLogsEntry {
           ),
           guild,
         );
+    } else if (targetType === Targets.CHANNEL) {
+      this.target =
+        guild.channels.cache.get(data.target_id) ||
+        Channel.create(
```
SpaceEEC, line 489: This will also insert the channel into the cache of the guild, I don't think this is desirable: 
https://github.com/discordjs/discord.js/blob/32d8c8ba940b31ab577c3de518fef382fe33a60d/src/structures/Channel.js#L156

How much info is available here?

iShibi: For `CHANNEL_UPDATE` it's only `name` but for the other two events, it's:

uR1MafnvnC

Agreed, for the delete event caching the channel on create isn't a good idea. Maybe, we can add an extra boolean param to `Channel#create` to control caching. It will be true by default so it won't break anything that depends on the `Channel#create` method.

iShibi: Thanks, for the thumbs up kyra (I saw it after the commit 😸) but now I feel like all the above effort is not worth for just getting an object for a deleted channel with partial properties. So, I removed the channel creation part. Now, it is just an object with whatever data the api sends in `changes`. For cached channel it will still be a complete channel object.

kyranet (approved):

vladfrangu (approved):

iCrawl (approved):

src/structures/GuildAuditLogs.js
```diff
@@ -349,7 +351,7 @@ class GuildAuditLogsEntry {
 
     /**
      * Any extra data from the entry
-     * @type {?Object|Role|GuildMember}
+     * @type {?(Object|Role|GuildMember|Channel)}
```
SpaceEEC, line 354: This doesn't look like it could be a Channel to me:
```suggestion
     * @type {?(Object|Role|GuildMember)}
```

SpaceEEC (dismissed):

src/structures/GuildAuditLogs.js
iShibi: hmm, right. It can only be a property of the generic object at the most. Thanks 👍

SpaceEEC (approved):

PR #5782, opened 2021-06-08
If a url has a space its ok.

AMTitan:
So if the embed URL has a space it will take out the space.


iCrawl:
It's up to the user to handle this correctly.

PR #5784, opened 2021-06-08
docs(GuildMemberManager): fix docs for #search()

vaporox:

The docs for `GuildMemberManager#search()` are using `@property` instead of `@param`, causing it to not show up on the docs site. To fix this, new typedefs have been added to be more consistent with the typings.

- There are no code changes
- Typings don't need updating

### DIFF
--- a/src/managers/GuildBanManager.js
+++ b/src/managers/GuildBanManager.js
@@ -121,11 +121,16 @@ class GuildBanManager extends BaseManager {
   }
 
   /**
+   * Options for banning a user.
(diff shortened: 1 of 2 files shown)


kyranet (approved):

iShibi:
I've been working on a similar PR, the only overlapping change is `BanOptions`. I'll revert that since it has been addressed here.

iCrawl (approved):

SpaceEEC (approved):

PR #5785, opened 2021-06-08
refactor(*): make typedefs for all options params

iShibi:

This PR starts the work of creating typedefs for `options` params of all the methods that don't have one. I've created a few so far but there are still few more to add. Since, #5784 added some new typedefs too, I'll rebase this PR once that one is merged to remove duplicate changes. This PR also tries to add more helpful description of a property in places where the old one is a little vague.

ℹ️ The names for the new typedefs are taken from their respective interfaces in the typings.

### DIFF
--- a/src/managers/GuildChannelManager.js
+++ b/src/managers/GuildChannelManager.js
@@ -61,7 +61,7 @@ class GuildChannelManager extends BaseManager {
 
   /**
    * Options used to create a new channel in a guild.
-   * @typedef {Object} GuildCreateChannelOptions
+   * @typedef {Object} GuildChannelCreateOptions
    * @property {string} [type='text'] The type of the new channel, either `text`, `voice`, `category`, `news`,
    * `store`, or `stage`
    * @property {string} [topic] The topic for the new channel
@@ -79,7 +79,7 @@ class GuildChannelManager extends BaseManager {
   /**
    * Creates a new channel in the guild.
    * @param {string} name The name of the new channel
-   * @param {GuildCreateChannelOptions} [options={}] Options for creating the new channel
+   * @param {GuildChannelCreateOptions} [options={}] Options for creating the new channel
    * @returns {Promise<GuildChannel>}
    * @example
    * // Create a new text channel
--- a/src/managers/GuildChannelManager.js
+++ b/src/managers/GuildChannelManager.js
@@ -60,21 +60,26 @@ class GuildChannelManager extends BaseManager {
    */
 
   /**
+   * Options used to create a new channel in a guild.
+   * @typedef {Object} GuildCreateChannelOptions
+   * @property {string} [type='text'] The type of the new channel, either `text`, `voice`, `category`, `news`,
+   * `store`, or `stage`
+   * @property {string} [topic] The topic for the new channel
+   * @property {boolean} [nsfw] Whether the new channel is nsfw
+   * @property {number} [bitrate] Bitrate of the new channel in bits (only voice)
+   * @property {number} [userLimit] Maximum amount of users allowed in the new channel (only voice)
+   * @property {ChannelResolvable} [parent] Parent of the new channel
+   * @property {OverwriteResolvable[]|Collection<Snowflake, OverwriteResolvable>} [permissionOverwrites]
+   * Permission overwrites of the new channel
+   * @property {number} [position] Position of the new channel
+   * @property {number} [rateLimitPerUser] The ratelimit per user for the new channel
+   * @property {string} [reason] Reason for creating the new channel
+   */
+
+  /**
    * Creates a new channel in the guild.
    * @param {string} name The name of the new channel
-   * @param {Object} [options] Options
-   * @param {string} [options.type='text'] The type of the new channel, either `text`, `voice`, `category`, `news`,
-   * `store`, or `stage`
-   * @param {string} [options.topic] The topic for the new channel
-   * @param {boolean} [options.nsfw] Whether the new channel is nsfw
-   * @param {number} [options.bitrate] Bitrate of the new channel in bits (only voice)
-   * @param {number} [options.userLimit] Maximum amount of users allowed in the new channel (only voice)
-   * @param {ChannelResolvable} [options.parent] Parent of the new channel
-   * @param {OverwriteResolvable[]|Collection<Snowflake, OverwriteResolvable>} [options.permissionOverwrites]
-   * Permission overwrites of the new channel
-   * @param {number} [options.position] Position of the new channel
-   * @param {number} [options.rateLimitPerUser] The ratelimit per user for the channel
-   * @param {string} [options.reason] Reason for creating the channel
+   * @param {GuildCreateChannelOptions} [options={}] Options for creating the new channel
    * @returns {Promise<GuildChannel>}
    * @example
    * // Create a new text channel
--- a/src/managers/GuildManager.js
+++ b/src/managers/GuildManager.js
@@ -129,22 +129,28 @@ class GuildManager extends BaseManager {
   }
 
   /**
+   * Options used to create a guild.
+   * @typedef {Object} GuildCreateOptions
+   * @param {number} [afkChannelID] The ID of the AFK channel
+   * @param {number} [afkTimeout] The AFK timeout in seconds
+   * @param {PartialChannelData[]} [channels] The channels for this guild
+   * @param {DefaultMessageNotifications} [defaultMessageNotifications] The default message notifications
+   * for the guild
+   * @param {ExplicitContentFilterLevel} [explicitContentFilter] The explicit content filter level for the guild
+   * @param {BufferResolvable|Base64Resolvable} [icon=null] The icon for the guild
+   * @param {string} [region] The region for the server, defaults to the closest one available
+   * @param {PartialRoleData[]} [roles] The roles for this guild,
+   * the first element of this array is used to change properties of the guild's everyone role.
+   * @param {number} [systemChannelID] The ID of the system channel
+   * @param {SystemChannelFlagsResolvable} [systemChannelFlags] The flags of the system channel
+   * @param {VerificationLevel} [verificationLevel] The verification level for the guild
+   */
+
+  /**
    * Creates a guild.
    * <warn>This is only available to bots in fewer than 10 guilds.</warn>
    * @param {string} name The name of the guild
-   * @param {Object} [options] Options for the creating
-   * @param {number} [options.afkChannelID] The ID of the AFK channel
-   * @param {number} [options.afkTimeout] The AFK timeout in seconds
-   * @param {PartialChannelData[]} [options.channels] The channels for this guild
-   * @param {DefaultMessageNotifications} [options.defaultMessageNotifications] The default message notifications
-   * for the guild
-   * @param {ExplicitContentFilterLevel} [options.explicitContentFilter] The explicit content filter level for the guild
-   * @param {BufferResolvable|Base64Resolvable} [options.icon=null] The icon for the guild
-   * @param {PartialRoleData[]} [options.roles] The roles for this guild,
-   * the first element of this array is used to change properties of the guild's everyone role.
-   * @param {number} [options.systemChannelID] The ID of the system channel
-   * @param {SystemChannelFlagsResolvable} [options.systemChannelFlags] The flags of the system channel
-   * @param {VerificationLevel} [options.verificationLevel] The verification level for the guild
+   * @param {GuildCreateOptions} [options] Options for creating the guild
    * @returns {Promise<Guild>} The guild that was created
    */
   async create(
--- a/src/managers/GuildManager.js
+++ b/src/managers/GuildManager.js
@@ -50,7 +50,7 @@ class GuildManager extends BaseManager {
   /**
    * Partial data for a Role.
    * @typedef {Object} PartialRoleData
-   * @property {number} [id] The ID for this role, used to set channel overrides,
+   * @property {Snowflake} [id] The ID for this role, used to set channel overrides,
    * this is a placeholder and will be replaced by the API after consumption
    * @property {string} [name] The name of the role
    * @property {ColorResolvable} [color] The color of the role, either a hex string or a base 10 number
@@ -63,7 +63,7 @@ class GuildManager extends BaseManager {
   /**
    * Partial overwrite data.
    * @typedef {Object} PartialOverwriteData
-   * @property {number|Snowflake} id The Role or User ID for this overwrite
+   * @property {Snowflake} id The Role or User ID for this overwrite
    * @property {string} [type] The type of this overwrite
    * @property {PermissionResolvable} [allow] The permissions to allow
    * @property {PermissionResolvable} [deny] The permissions to deny
@@ -72,16 +72,16 @@ class GuildManager extends BaseManager {
   /**
    * Partial data for a Channel.
    * @typedef {Object} PartialChannelData
-   * @property {number} [id] The ID for this channel, used to set its parent,
+   * @property {Snowflake} [id] The ID for this channel, used to set its parent,
    * this is a placeholder and will be replaced by the API after consumption
-   * @property {number} [parentID] The parent ID for this channel
+   * @property {Snowflake} [parentID] The parent ID for this channel
    * @property {string} [type] The type of the channel
    * @property {string} name The name of the channel
    * @property {string} [topic] The topic of the text channel
    * @property {boolean} [nsfw] Whether the channel is NSFW
    * @property {number} [bitrate] The bitrate of the voice channel
    * @property {number} [userLimit] The user limit of the channel
-   * @property {PartialOverwriteData} [permissionOverwrites]
+   * @property {PartialOverwriteData[]} [permissionOverwrites]
    * Overwrites of the channel
    * @property {number} [rateLimitPerUser] The rate limit per user of the channel in seconds
    */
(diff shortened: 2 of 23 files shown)


src/managers/GuildChannelManager.js
```diff
@@ -59,21 +59,26 @@ class GuildChannelManager extends BaseManager {
...
-   * @param {string} [options.reason] Reason for creating the channel
+   * @property {number} [position] Position of the new channel
+   * @property {number} [rateLimitPerUser] The ratelimit per user for the new channel
+   * @property {string} [reason] Reason for creating the new channel
+   */
+
+  /**
+   * Creates a new channel in the guild.
+   * @param {string} name The name of the new channel
+   * @param {GuildCreateChannelOptions} [options={}] Options for creating the new channel
```
ImRodry, line 81: I think having [options = {}] is redundant because the [] indicates that the parameter is optional and saying that an empty object is assumed is already implicit when the parameter is omitted. I know this is done many times across the lib but I still think it's worth the change (maybe change others too, would need feedback from the maintainers)

src/managers/GuildManager.js
```diff
@@ -128,24 +128,29 @@ class GuildManager extends BaseManager {
...
+   * @param {DefaultMessageNotifications} [defaultMessageNotifications] The default message notifications
+   * for the guild
+   * @param {ExplicitContentFilterLevel} [explicitContentFilter] The explicit content filter level for the guild
+   * @param {BufferResolvable|Base64Resolvable} [icon=null] The icon for the guild
+   * @param {string} [region] The region for the server, defaults to the closest one available
+   * @param {PartialRoleData[]} [roles] The roles for this guild,
+   * the first element of this array is used to change properties of the guild's everyone role.
+   * @param {number} [systemChannelID] The ID of the system channel
+   * @param {SystemChannelFlagsResolvable} [systemChannelFlags] The flags of the system channel
+   * @param {VerificationLevel} [verificationLevel] The verification level for the guild
```
ImRodry, line 146: typedefs have `property`'s, not `param`'s

ImRodry (commented):
I know this is still a work in progress but I'm interested to see progress on this, also here are some changes I think should be made.

src/managers/GuildManager.js
iShibi: This is what happens when you work at the time when you should be sleeping, will fix soon 👍

ImRodry: xD all good

src/managers/GuildChannelManager.js
iShibi: Yes, the `[]` states that the param is optional. Doing `[options = {}]` shows `{}` under default in the docs which will not be there if we remove the `= {}` part. I think this is better because it distinct itself from optional param that are not objects.

ImRodry: Yeah I know how that's displayed but if the options parameter is already marked as an object (by the typedef), adding {} wouldn't add much information imo

src/structures/GuildChannel.js
```diff
@@ -492,19 +502,24 @@ class GuildChannel extends Channel {
    */
 
   /**
-   * Creates an invite to this guild channel.
-   * @param {Object} [options={}] Options for the invite
-   * @param {boolean} [options.temporary=false] Whether members that joined via the invite should be automatically
+   * Options used to create an invite to a guild.
```
ImRodry, line 505:
```suggestion
   * Options used to create an invite to a guild channel.
```

src/structures/GuildChannel.js
```diff
@@ -554,22 +569,15 @@ class GuildChannel extends Channel {
     return invites;
   }
 
-  /* eslint-disable max-len */
+  /**
+   * Options used to clone a guild channel.
+   * @typedef {GuildCreateChannelOptions} GuildChannelCloneOptions
```
ImRodry, line 574: Is it clone or create? And if it’s clone, where did the other properties go?

src/structures/MessageMentions.js
```diff
@@ -161,14 +161,19 @@ class MessageMentions {
     return this._channels;
   }
 
+  /**
+   * Options used to check a mention.
```
ImRodry, line 165:
```suggestion
   * Options used to check for a mention.
```

src/structures/GuildChannel.js
iShibi: Clone. Those are inherited from the `GuildCreateChannelOptions` typedef.

ImRodry: But this is that typedef, am I missing something?

iShibi: This is `GuildChannelCloneOptions` typedef which inherits props from the typedef that is inside the `{}` on the left and is defined [here](https://github.com/discordjs/discord.js/blob/9dd3a9a018e0c13ba448f6e579bc866d68c75dc1/src/managers/GuildChannelManager.js#L63)

ImRodry: Ohhh yeah right my bad lol

typings/index.d.ts
```diff
@@ -1049,7 +1049,7 @@ declare module 'discord.js' {
   }
 
   // tslint:disable-next-line:no-empty-interface - Merge RateLimitData into RateLimitError to not have to type it again
-  interface RateLimitError extends RateLimitData { }
+  interface RateLimitError extends RateLimitData {}
```
ImRodry: Watch out for this change (and others caused by vscode's default formatter) but why don't you address the comment while you're at it?

iShibi: The linter fixes all the formating during commit. That comment isn't mine.

iShibi:
I think I've done all the neccessary changes, so I'll mark this as ready for review. This PR along with #5805 should make the docs and typings a little bit more clean 🥳

typings/index.d.ts
ImRodry: I know the comment isn’t yours but you could address it I guess
But those formatting issues are caused by vscode’s default formatter, just use prettier

ImRodry: Seems like they were fixed in a later commit, nvm

iShibi: The current formating is the correct one as it is same in the master. You commented on it just after rebase when this change was introduced. Since, no commit happened during rebase the linter couldn't have fixed it. It fixed it in the later commits, that's why this suggestion is showing as outdated.

I don't know what you want me to address about the comment. Could you be a little more specific?

ImRodry: Ah alright my bad then
Since the comment says “Merge RateLimitData into RateLimitError to not have to type it again” you could address that since you’re touching interfaces and typings, it’d be in line with the theme of the PR imo

iCrawl (approved):

src/managers/GuildManager.js
```diff
@@ -128,23 +128,29 @@ class GuildManager extends BaseManager {
...
+   * @typedef {Object} GuildCreateOptions
+   * @property {number} [afkChannelID] The ID of the AFK channel
+   * @property {number} [afkTimeout] The AFK timeout in seconds
+   * @property {PartialChannelData[]} [channels] The channels for this guild
+   * @property {DefaultMessageNotifications} [defaultMessageNotifications] The default message notifications
+   * for the guild
+   * @property {ExplicitContentFilterLevel} [explicitContentFilter] The explicit content filter level for the guild
+   * @property {BufferResolvable|Base64Resolvable} [icon=null] The icon for the guild
+   * @property {string} [region] The region for the server, defaults to the closest one available
+   * @property {PartialRoleData[]} [roles] The roles for this guild,
```
kyranet, line 142:
```suggestion
   * @property {PartialRoleData[]} [roles=[]] The roles for this guild,
```

src/managers/GuildManager.js
```diff
@@ -128,23 +128,29 @@ class GuildManager extends BaseManager {
     return super.resolveID(guild);
   }
 
+  /**
+   * Options used to create a guild.
+   * @typedef {Object} GuildCreateOptions
+   * @property {number} [afkChannelID] The ID of the AFK channel
+   * @property {number} [afkTimeout] The AFK timeout in seconds
+   * @property {PartialChannelData[]} [channels] The channels for this guild
```
kyranet, line 136:
```suggestion
   * @property {PartialChannelData[]} [channels=[]] The channels for this guild
```

src/structures/MessageMentions.js
```diff
@@ -161,14 +161,19 @@ class MessageMentions {
     return this._channels;
   }
 
+  /**
+   * Options used to check for a mention.
+   * @typedef {Object} MessageMentionsHasOptions
+   * @property {boolean} [ignoreDirect=false] - Whether to ignore direct mentions to the item
+   * @property {boolean} [ignoreRoles=false] - Whether to ignore role mentions to a guild member
+   * @property {boolean} [ignoreEveryone=false] - Whether to ignore everyone/here mentions
```
kyranet, line 169:
```suggestion
   * @property {boolean} [ignoreDirect=false] Whether to ignore direct mentions to the item
   * @property {boolean} [ignoreRoles=false] Whether to ignore role mentions to a guild member
   * @property {boolean} [ignoreEveryone=false] Whether to ignore everyone/here mentions
```

kyranet (changes requested):
I'm unsure if jsdocs forbid `[field=[]]`, but I figured I'd point out just in case.

kyranet (approved):

src/managers/GuildManager.js
```diff
@@ -128,23 +128,29 @@ class GuildManager extends BaseManager {
     return super.resolveID(guild);
   }
 
+  /**
+   * Options used to create a guild.
+   * @typedef {Object} GuildCreateOptions
+   * @property {number} [afkChannelID] The ID of the AFK channel
```
vladfrangu, line 134: This can be a Snowflake too

src/managers/GuildManager.js
```diff
@@ -128,23 +128,29 @@ class GuildManager extends BaseManager {
...
+   * @property {number} [afkTimeout] The AFK timeout in seconds
+   * @property {PartialChannelData[]} [channels=[]] The channels for this guild
```
vladfrangu, line 136: Just so I'm sure, is this defined?

src/managers/GuildManager.js
```diff
@@ -128,23 +128,29 @@ class GuildManager extends BaseManager {
...
+   * @property {DefaultMessageNotifications} [defaultMessageNotifications] The default message notifications
+   * for the guild
+   * @property {ExplicitContentFilterLevel} [explicitContentFilter] The explicit content filter level for the guild
+   * @property {BufferResolvable|Base64Resolvable} [icon=null] The icon for the guild
+   * @property {string} [region] The region for the server, defaults to the closest one available
```
vladfrangu, line 141: n/a anymore

src/managers/GuildManager.js
```diff
@@ -128,23 +128,29 @@ class GuildManager extends BaseManager {
...
+   * @property {PartialRoleData[]} [roles=[]] The roles for this guild,
```
vladfrangu, line 142: As PartialChannelData, just to be sure, is this defined?

src/managers/GuildManager.js
```diff
@@ -128,23 +128,29 @@ class GuildManager extends BaseManager {
...
+   * the first element of this array is used to change properties of the guild's everyone role.
+   * @property {number} [systemChannelID] The ID of the system channel
```
vladfrangu, line 144: Can be a snowflake too

src/managers/GuildMemberManager.js
```diff
@@ -198,19 +198,19 @@ class GuildMemberManager extends BaseManager {
...
 
   /**
-   * Options for pruning guild members.
+   * Options used for pruning guild members.
    * @typedef {Object} GuildPruneMembersOptions
-   * @property {number} [days] Number of days of inactivity required to kick
-   * @property {boolean} [dry] Get number of users that will be kicked, without actually kicking them
-   * @property {boolean} [count] Whether or not to return the number of users that have been kicked.
+   * @property {number} [days=7] Number of days of inactivity required to kick
+   * @property {boolean} [dry=false] Get number of users that will be kicked, without actually kicking them
```
vladfrangu, line 204:
```suggestion
   * @property {boolean} [dry=false] Get the number of users that will be kicked, without actually kicking them
```

src/rest/RequestHandler.js
```diff
@@ -125,13 +125,7 @@ class RequestHandler {
...
          * Emitted when the client hits a rate limit while making a request
          * @event Client#rateLimit
-         * @param {Object} rateLimitInfo Object containing the rate limit info
-         * @param {number} rateLimitInfo.timeout Timeout in ms
-         * @param {number} rateLimitInfo.limit Number of requests that can be made to this endpoint
-         * @param {string} rateLimitInfo.method HTTP method used for request that triggered this event
-         * @param {string} rateLimitInfo.path Path used for request that triggered this event
-         * @param {string} rateLimitInfo.route Route used for request that triggered this event
-         * @param {boolean} rateLimitInfo.global Whether the rate limit that was reached was the global limit
+         * @param {RateLimitData} rateLimitData Object containing the rate limit info
```
vladfrangu, line 128: Make sure these have the same properties 🙏

src/rest/RequestHandler.js
```diff
@@ -231,12 +225,17 @@ class RequestHandler {
...
         invalidCount % this.manager.client.options.invalidRequestWarningInterval === 0;
       if (emitInvalid) {
+        /**
+         * @typedef {Object} InvalidRequestWarningData
+         * @property {number} count Number of invalid requests that have been made in the window
+         * @property {number} remainingTime Time in ms remaining before the count resets
+         */
+
         /**
          * Emitted periodically when the process sends invalid messages to let users avoid the
```
vladfrangu, line 235:
```suggestion
         * Emitted periodically when the process sends invalid requests to let users avoid the
```

vladfrangu (changes requested):

src/managers/GuildManager.js
iShibi: hmm, now that you pointed it out, I don't get why it is a number in the first place. Should be changed to `Snowflake` in typings too I think.

src/managers/GuildManager.js
iShibi: Same as above, how come no one noticed it for so long 😆

src/managers/GuildManager.js
iShibi: Yup, it's defined [here](https://github.com/discordjs/discord.js/blob/da6f21a3fca084f27d115754fed5fff8091cce77/src/managers/GuildManager.js#L74). Same, issue there tho, the id props are documented as `number`.

src/managers/GuildManager.js
iShibi: Yup, [here](https://github.com/discordjs/discord.js/blob/da6f21a3fca084f27d115754fed5fff8091cce77/src/managers/GuildManager.js#L52)

src/rest/RequestHandler.js
iShibi: yup. [RateLimitData](https://github.com/discordjs/discord.js/blob/da6f21a3fca084f27d115754fed5fff8091cce77/src/util/Constants.js#L8) defined in Constants, has exactly the same props.

iShibi:
Changed, `number` to `Snowflake` for IDs. Also, `PartialChannelData#permissionOverwrites` is an array but wasn't documented as one, fixed that too. Added `PartialOverwriteData` interface for the same in typings.

iCrawl:
This needs a rebase.

iShibi:
Rebased

src/managers/GuildManager.js
```diff
@@ -50,7 +50,7 @@ class GuildManager extends BaseManager {
   /**
    * Partial data for a Role.
    * @typedef {Object} PartialRoleData
-   * @property {number} [id] The ID for this role, used to set channel overrides,
+   * @property {Snowflake} [id] The ID for this role, used to set channel overrides,
```
vladfrangu, line 53: I said this can also be a number for a reason! You can specify a role as `{ id: 1, color: 0xffffff }` or `{ id: "8000000000000000000", role: 0x000000 }`. The snowflake type is ``type Snowflake = `${bigint}`;``, which is a string that can be a bigint

Ditto for the rest

vladfrangu (changes requested):

src/managers/GuildManager.js
iShibi: You said, "it can be a `Snowflake` too", so I thought `number` is wrong type and changed it to `Snowflake` only, which is what we have as a type for most IDs. Fixed it in the latest commit.

vladfrangu (approved):

src/managers/GuildChannelManager.js
```diff
@@ -60,21 +60,26 @@ class GuildChannelManager extends BaseManager {
    */
 
   /**
-   * Creates a new channel in the guild.
-   * @param {string} name The name of the new channel
-   * @param {Object} [options] Options
-   * @param {string} [options.type='text'] The type of the new channel, either `text`, `voice`, `category`, `news`,
+   * Options used to create a new channel in a guild.
+   * @typedef {Object} GuildCreateChannelOptions
```
SpaceEEC, line 64: Maybe name this `GuildChannelCreateOptions` similar to the emoji one below?

src/structures/GuildChannel.js
```diff
@@ -529,19 +539,24 @@ class GuildChannel extends Channel {
    */
 
   /**
-   * Creates an invite to this guild channel.
-   * @param {Object} [options={}] Options for the invite
-   * @param {boolean} [options.temporary=false] Whether members that joined via the invite should be automatically
+   * Options used to create an invite to a guild channel.
+   * @typedef {Object} InviteOptions
```
SpaceEEC, line 543: Maybe name this `CreateInviteOptions`?

src/managers/GuildChannelManager.js
iShibi: Done

SpaceEEC (approved):

PR #5786, opened 2021-06-09
Update APIRequest.js

BlackFenix2:
I am changing the require statement for node-fetch to include the default import. 
from:
```js
const fetch = require('node-fetch');
```

to:
```js
const fetch = require('node-fetch').default;
```

this will allow discord.js to be consumed by node.js apps with bundlers such as webpack.

### DIFF
--- a/src/rest/APIRequest.js
+++ b/src/rest/APIRequest.js
@@ -3,7 +3,7 @@
 const https = require('https');
 const FormData = require('@discordjs/form-data');
 const AbortController = require('abort-controller');
-const fetch = require('node-fetch');
+const fetch = require('node-fetch').default;
 const { UserAgent } = require('../util/Constants');
 
 if (https.Agent) var agent = new https.Agent({ keepAlive: true });


kyranet:
We don't even support Webpack, nor Discord supports web builds due to CORS, so I don't know what's the point.

BlackFenix2:
I am trying to consume this app in a node.js server, this is the intended use of this package. this is not a PR to support client-side implementations. 

the issue is that node-fetch is not required properly when bundled through webpack https://github.com/node-fetch/node-fetch/issues/450, i have introduced the workaround which is the solution to the node-fetch issue. 

this is the error message i get when i try to make any API Request from bundled discord.js
```js
error discord:  HTTPError [TypeError]: fetch is not a function
```

i have tested this change both using the discord.js library as-is without any bundling/minification and bundled with webpack. 

sorry i didn't include more details at first, let me know if you need anything else.

iCrawl:
This library does not support webpack at all really.

You will definitely encounter other issues once you get past this one, hence why we dropped webpack support all together.

PR #5787, opened 2021-06-09
feat: general component improvements

monbrey:
By request, this PR extract several of the general improvements I had discovered while working on #5692 into a separate PR.

- Subclasses for `MessageComponentInteraction` per component type
- Support for parsing arrays of components as rows in raw options
- Improved raw emoji resolution in component builders
- Typings fixes
- Documentation improvements and consistency

### DIFF
--- a/src/structures/APIMessage.js
+++ b/src/structures/APIMessage.js
@@ -3,6 +3,7 @@
 const BaseMessageComponent = require('./BaseMessageComponent');
 const MessageEmbed = require('./MessageEmbed');
 const { RangeError } = require('../errors');
+const { MessageComponentTypes } = require('../util/Constants');
 const DataResolver = require('../util/DataResolver');
 const MessageFlags = require('../util/MessageFlags');
 const Util = require('../util/Util');
@@ -153,7 +154,9 @@ class APIMessage {
     const embeds = embedLikes.map(e => new MessageEmbed(e).toJSON());
 
     const components = this.options.components?.map(c =>
-      BaseMessageComponent.create(Array.isArray(c) ? { type: 1, components: c } : c).toJSON(),
+      BaseMessageComponent.create(
+        Array.isArray(c) ? { type: MessageComponentTypes.ACTION_ROW, components: c } : c,
+      ).toJSON(),
     );
 
     let username;
--- a/src/structures/APIMessage.js
+++ b/src/structures/APIMessage.js
@@ -152,7 +152,9 @@ class APIMessage {
     }
     const embeds = embedLikes.map(e => new MessageEmbed(e).toJSON());
 
-    const components = this.options.components?.map(c => BaseMessageComponent.create(c).toJSON());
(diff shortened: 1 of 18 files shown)


vladfrangu (approved):

iCrawl (approved):

kyranet (approved):

src/structures/APIMessage.js
```diff
@@ -153,7 +153,9 @@ class APIMessage {
     }
     const embeds = embedLikes.map(e => new MessageEmbed(e).toJSON());
 
-    const components = this.options.components?.map(c => BaseMessageComponent.create(c).toJSON());
+    const components = this.options.components?.map(c =>
+      BaseMessageComponent.create(Array.isArray(c) ? { type: 1, components: c } : c).toJSON(),
```
SpaceEEC, line 157: You could import and use `MessageComponentTypes.ACTION_ROW` from the constants file here instead of this magic number.

src/util/Util.js
```diff
@@ -267,6 +267,22 @@ class Util {
     return { animated: Boolean(m[1]), name: m[2], id: m[3] || null };
   }
 
+  /**
+   * Resolves a partial emoji object from an {@link EmojiIdentifierResolvable}, without checking a Client.
+   * @param {EmojiIdentifierResolvable} emoji Emoji identifier to resolve
+   * @returns {Object} Object with possible `animated`, `name`, and `id` properties
```
SpaceEEC, line 273: Since it's returning `null` when resolving fails.
```suggestion
   * @returns {?RawEmoji} Object with possible `animated`, `name`, and `id` properties
```
Shouldn't this also use the new typdef? (The description is redundant then too)

typings/index.d.ts
```diff
@@ -1958,6 +1963,7 @@ declare module 'discord.js' {
     public static moveElementInArray(array: any[], element: any, newIndex: number, offset?: boolean): number;
     public static parseEmoji(text: string): { animated: boolean; name: string; id: Snowflake | null } | null;
     public static resolveColor(color: ColorResolvable): number;
+    public static resolvePartialEmoji(emoji: EmojiIdentifierResolvable): Partial<RawEmoji>;
```
SpaceEEC, line 1966:
```suggestion
    public static resolvePartialEmoji(emoji: EmojiIdentifierResolvable): Partial<RawEmoji> | null;
```

src/util/Util.js
```diff
@@ -267,6 +267,22 @@ class Util {
...
+   * @private
+   */
+  static resolvePartialEmoji(emoji) {
+    if (typeof emoji === 'undefined') return null;
```
SpaceEEC, line 277: Is it intended that passing `undefined` returns `null` but passing `null` throws a destructuring error?

SpaceEEC (changes requested):

iCrawl:
This needs a rebase.

SpaceEEC (approved):

PR #5792, opened 2021-06-09
feat(*): document and support embeds field in message create endpoint

adrifcastr:

The message create endpoint now supports up to 10 embeds just like the webhook endpoint, these changes reflect that addition within the typings and APIMessage#create function.

### DIFF
--- a/src/structures/APIMessage.js
+++ b/src/structures/APIMessage.js
@@ -198,7 +198,7 @@ class APIMessage {
       content,
       tts,
       nonce,
-      embeds,
+      embeds: this.options.embeds ? embeds : undefined,
       components,
       username,
       avatar_url: avatarURL,
--- a/src/structures/APIMessage.js
+++ b/src/structures/APIMessage.js
@@ -147,9 +147,7 @@ class APIMessage {
     if (this.options.embeds) {
       embedLikes.push(...this.options.embeds);
     }
-    else if (this.options.embed) {
-      embedLikes.push(this.options.embed);
-    }
+
     const embeds = embedLikes.map(e => new MessageEmbed(e).toJSON());
 
     const components = this.options.components?.map(c =>
@@ -221,6 +219,22 @@ class APIMessage {
    */
   async resolveFiles() {
     if (this.files) return this;
+    
+    const embedLikes = [];
+
+    if (this.options.embeds) {
+      embedLikes.push(...this.options.embeds);
+    }
+
+    const fileLikes = [];
+    if (this.options.files) {
+      fileLikes.push(...this.options.files);
+    }
+    for (const embed of embedLikes) {
+      if (embed.files) {
+        fileLikes.push(...embed.files);
+      }
+    }
 
     this.files = await Promise.all(this.options.files?.map(file => this.constructor.resolveFile(file)) ?? []);
     return this;
@@ -300,10 +314,9 @@ class APIMessage {
    * @returns {MessageOptions|WebhookMessageOptions}
    */
   static create(target, options, extra = {}) {
-    return new this(
-      target,
-      typeof options !== 'object' || options === null ? { content: options, ...extra } : { ...options, ...extra },
-    );
+    if (typeof options !== 'object' || options === null) return new this(target, { content: options, ...extra });
+    if (options.embed) options.embeds = [options.embed];
+    return new this(target, { ...options, ...extra });
   }
 }
 
--- a/src/structures/APIMessage.js
+++ b/src/structures/APIMessage.js
@@ -312,9 +312,10 @@ class APIMessage {
    * @returns {MessageOptions|WebhookMessageOptions}
    */
   static create(target, options, extra = {}) {
-    if (typeof options !== 'object' || options === null) return new this(target, { content: options, ...extra });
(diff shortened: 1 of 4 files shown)


kyranet (approved):

ckohen:
[Upstream](https://github.com/discord/discord-api-docs/pull/3105) has been merged 🚀 changes are live.

kyranet (commented):
Since Discord seems to plan the removal of the `embed` field (in favour of `embeds`), should we refactor the code and account for less overloads? I see this as a great opportunity to remove some conditional checks made for webhooks (and interactions).

As for the `embed` option, should we keep it as an alias for `embeds: [embed]`? Since `{ embed }` is much nicer and shorter to write (and we already have an alias for `content`, so why not?).

DiscordCatto (changes requested):
[src/structures/APIMessage.js#L147](https://github.com/discordjs/discord.js/blob/master/src/structures/APIMessage.js#L147) needs to be updated aswell, so this works properly.

adrifcastr:
> [src/structures/APIMessage.js#L147](https://github.com/discordjs/discord.js/blob/master/src/structures/APIMessage.js#L147) needs to be updated aswell, so this works properly. 

I am aware, please refer to the initial comment in italics, I somewhat mistimed opening the PR yesterday

adrifcastr:
> Since Discord seems to plan the removal of the `embed` field (in favour of `embeds`), should we refactor the code and account for less overloads? I see this as a great opportunity to remove some conditional checks made for webhooks (and interactions).
> 
> As for the `embed` option, should we keep it as an alias for `embeds: [embed]`? Since `{ embed }` is much nicer and shorter to write (and we already have an alias for `content`, so why not?).

Changes adressed, tested & rebased :)

csuvajit:
[src/structures/APIMessage.js#L203](https://github.com/discordjs/discord.js/blob/master/src/structures/APIMessage.js#L203) needs some changes here.

iCrawl (approved):

vladfrangu (approved):

typings/index.d.ts
```diff
@@ -3729,7 +3749,6 @@ declare module 'discord.js' {
   interface WebhookMessageOptions extends Omit<MessageOptions, 'embed' | 'reply'> {
```
ckohen, line 3749:
```suggestion
  interface WebhookMessageOptions extends Omit<MessageOptions, 'reply'> {
```

ckohen (changes requested):
https://github.com/discordjs/discord.js/blob/9e46178510f71c1e612d6b9561604ebd22833d2d/src/structures/interfaces/TextBasedChannel.js#L73 needs updating.

adrifcastr:
> TextBasedChannel.js#L73 needs updating.

Done, thank you

src/structures/APIMessage.js
```diff
@@ -202,8 +198,7 @@ class APIMessage {
       content,
       tts,
       nonce,
-      embed: !isWebhookLike ? (this.options.embed === null ? null : embeds[0]) : undefined,
-      embeds: isWebhookLike ? embeds : undefined,
+      embeds,
```
ckohen, line 201:
```suggestion
      embeds: embeds.length ? embeds : undefined
```
This prevents stripping embeds when not provided

ToasterSticks: Isn't that how it worked before this change? It could only be removed by passing `null`.

monbrey: And this ensures it will continue to work that way.

ckohen: Made me realize that sometimes we want to send an empty array though...fixed the comment to address that

monbrey: Should it be `typeof this.options.embeds === 'undefined'`? This will treat explicitly passing `null` as falsey

ckohen: `null` isn't actually a valid type to pass currently, if that wants to be supported then yes it should be typeof

SpaceEEC (approved):

src/structures/APIMessage.js
```diff
@@ -202,8 +198,7 @@ class APIMessage {
       content,
       tts,
       nonce,
-      embed: !isWebhookLike ? (this.options.embed === null ? null : embeds[0]) : undefined,
-      embeds: isWebhookLike ? embeds : undefined,
+      embeds: this.options.embeds ? embeds : undefined,
```
vaporox, line 201:
```suggestion
      embeds: this.options.embeds?.map(embed => new MessageEmbed(embed).toJSON()),
```
With this, the additional logic above is also not needed anymore

src/structures/interfaces/TextBasedChannel.js
```diff
@@ -70,7 +70,7 @@ class TextBasedChannel {
   /**
    * Options provided when sending or editing a message.
    * @typedef {BaseMessageOptions} MessageOptions
-   * @property {MessageEmbed|Object} [embed] An embed for the message
+   * @property {MessageEmbed[]|Object[]} [embeds] The embeds for the message
```
vaporox, line 73:
```suggestion
   * @property {(MessageEmbed|Object)[]} [embeds] The embeds for the message
```

PR #5758, opened 2021-06-05
refactor: enforce single param on sending/editing methods

adrifcastr:
**Please describe the changes this PR makes and why it should be merged:**
This PR refactors the methods across the the lib's sending/editing methods to avoid inconsistent overloads by removing multiple param options and enforcing single param usage as laid out in #5771.

(and let's pretend that there isn't a typo in the branch name)
**Status and versioning classification:**
- Code changes have been tested against the Discord API, or there are no code changes
- I pretend to know how to update typings and have done so, or typings don't need updating
- This PR includes breaking changes (methods removed or renamed, parameters moved or removed)

PR #5793, opened 2021-06-09
fix(ApiMessage): only pass objects as options directly

ckohen:

#5758 Introduced a small bug with throwing on invalid content, this fixes that.

### DIFF
--- a/src/structures/APIMessage.js
+++ b/src/structures/APIMessage.js
@@ -320,7 +320,7 @@ class APIMessage {
    * @returns {MessageOptions|WebhookMessageOptions}
    */
   static create(target, options, extra = {}) {
-    if (typeof options === 'string') return new this(target, { content: options, ...extra });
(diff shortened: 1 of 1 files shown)


src/structures/APIMessage.js
```diff
@@ -320,7 +320,7 @@ class APIMessage {
    * @returns {MessageOptions|WebhookMessageOptions}
    */
   static create(target, options, extra = {}) {
-    if (typeof options === 'string') return new this(target, { content: options, ...extra });
+    if (typeof options !== 'object' || options === null) return new this(target, { content: options, ...extra });
     else return new this(target, { ...options, ...extra });
```
kyranet, line 324: What about something like this? I'm unsure about the format Prettier will pick, so this might cause a format error:
```suggestion
    return new this(target, typeof options === 'object' || options !== null ?
    	{ ...options, ...extra } :
    	{ content: options, ...extra });
```

kyranet (approved):

iCrawl (approved):

PR #5795, opened 2021-06-10
fix(BitField): ensure missing returns an array of strings

SpaceEEC:

Currently `BitField#missing` assumes the passed bits are, if an array, an array of strings, this is not necessarily correct.
For example
```js
const missing = new Permissions([Permissions.FLAGS.SEND_MESSAGES, Permissions.FLAGS.VIEW_CHANNEL])
    .missing([Permissions.FLAGS.VIEW_CHANNEL, Permissions.FLAGS.BAN_MEMBERS]);
console.log(missing); // [ 4n ] instead of [ "BAN_MEMBERS" ]
```

This PR ensures that `BitField#missing` always returns an array of strings.

### DIFF
--- a/src/util/BitField.js
+++ b/src/util/BitField.js
@@ -52,8 +52,7 @@ class BitField {
    * @returns {string[]}
    */
   missing(bits, ...hasParams) {
-    if (!Array.isArray(bits)) bits = new this.constructor(bits).toArray(false);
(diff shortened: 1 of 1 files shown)


kyranet (approved):
This code just looks so much more efficient, I like it.

iCrawl (approved):

vladfrangu (approved):

PR #5796, opened 2021-06-10
fix: don't patch missing properties from partial payloads

### LINKED ISSUE #5605: Some values of an object extending Channel loses its value after using slash commands
I was testing the new slash command feature released about a week ago and found out an issue where there are data loss after using slash commands.
One of the Moderators in the Support Server told me it is something to do with how Discord resolves the object.

```js
console.log(args[0] === "channel_id") //Text: true, Slash: true
const channel = message.guild.channels.cache.get(args[0]);
const { parentID } = channel; 
console.log(parentID) //Text: "parent_id" Slash: null
```
**Workaround**
Getting the channel using a snowflake

**Further details:**

- discord.js version: 12.5.0 (git+ssh://git@github.com/discordjs/discord.js.git#0e40f9b86826ba50aa3840807fb86e1bce6b1c3d)
- Node.js version: v14.15.3
- Operating system: Windows 10 Enterprise
- Priority this issue should have – please be realistic and elaborate if possible: Medium

**Relevant client options:**

- partials: Channel, Message, Reaction, User
- gateway intents: All
- other: N/A

Here is the screenshot for the proof:
image

SpaceEEC:

This PR resolves #5605 by only updating properties of Channel and subclasses when they are in the payload.
In similar vein, if a user option was used, their last message properties would reset to null. Since this is something we provide on our own, there is no reason for those to be in `_patch`.

### DIFF
--- a/src/structures/GuildChannel.js
+++ b/src/structures/GuildChannel.js
@@ -35,37 +35,47 @@ class GuildChannel extends Channel {
      * @type {Guild}
      */
     this.guild = guild;
+
+    this.parentID = null;
   }
 
   _patch(data) {
     super._patch(data);
 
-    /**
-     * The name of the guild channel
-     * @type {string}
-     */
-    this.name = data.name;
+    if ('name' in data) {
+      /**
+       * The name of the guild channel
+       * @type {string}
+       */
+      this.name = data.name;
+    }
 
-    /**
-     * The raw position of the channel from discord
-     * @type {number}
-     */
-    this.rawPosition = data.position;
+    if ('position' in data) {
+      /**
+       * The raw position of the channel from discord
+       * @type {number}
+       */
+      this.rawPosition = data.position;
+    }
 
-    /**
-     * The ID of the category parent of this channel
-     * @type {?Snowflake}
-     */
-    this.parentID = data.parent_id || null;
+    if ('parent_id' in data) {
+      /**
+       * The ID of the category parent of this channel
+       * @type {?Snowflake}
+       */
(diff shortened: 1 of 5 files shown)


iCrawl (approved):

src/structures/GuildChannel.js
```diff
@@ -35,37 +35,47 @@ class GuildChannel extends Channel {
...
-    this.permissionOverwrites = new Collection();
-    if (data.permission_overwrites) {
-      for (const overwrite of data.permission_overwrites) {
-        this.permissionOverwrites.set(overwrite.id, new PermissionOverwrites(this, overwrite));
+    if ('permission_overwrites' in data) {
+      /**
+       * A map of permission overwrites in this channel for roles and users
+       * @type {Collection<Snowflake, PermissionOverwrites>}
+       */
+      this.permissionOverwrites = new Collection();
```
vladfrangu, line 74: This will make it be possibly undefined now, if at any point we create a channel with partial data

vladfrangu (changes requested):

kyranet (approved):

src/structures/GuildChannel.js
```diff
@@ -35,37 +35,48 @@ class GuildChannel extends Channel {
...
+      if (data.permission_overwrites) {
```
vladfrangu, line 76: This if check is redundant considering there's already an `in` check

vladfrangu (changes requested):

vladfrangu (approved):

PR #5801, opened 2021-06-10
feat(RequestHandler): emit more info when a rate limit was hit

SpaceEEC:

This PR adds more info into the string that gets emitted when a request gets a 429 response.
Being added is:
- Whether the rate limit is global
- The HTTP method / verb
- The path of the request
- The limit the rate limit bucket has
- The timeout to wait for
- Whether this is a sublimit (and if so, how long to wait)

I kept it similar to how the WebSocket emits debug events, since putting all of that into one line wasn't feasible:
https://github.com/discordjs/discord.js/blob/a1f763ee75a7d906bbe727f616e33206dcc6792c/src/client/websocket/WebSocketShard.js#L243-L247

### DIFF
--- a/src/rest/RequestHandler.js
+++ b/src/rest/RequestHandler.js
@@ -5,7 +5,7 @@ const DiscordAPIError = require('./DiscordAPIError');
 const HTTPError = require('./HTTPError');
 const RateLimitError = require('./RateLimitError');
 const {
-  Events: { RATE_LIMIT, INVALID_REQUEST_WARNING },
+  Events: { DEBUG, RATE_LIMIT, INVALID_REQUEST_WARNING },
 } = require('../util/Constants');
 const Util = require('../util/Util');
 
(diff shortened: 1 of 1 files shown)


src/rest/RequestHandler.js
```diff
@@ -269,6 +266,19 @@ class RequestHandler {
...
+        this.manager.client.emit(
+          DEBUG,
+          `Hit a 429 while executing a request.
+    Global  : ${isGlobal}
+    Method  : ${request.method}
+    Path    : ${request.path}
+    Route   : ${request.route}
+    Limit   : ${limit}
+    Timeout : ${timeout}ms
+    Sublimit: ${sublimitTimeout ? `${sublimitTimeout}ms` : false}`,
```
kyranet, line 279: Shouldn't we show `None`/`none` instead of just `false`?
```suggestion
    Sublimit: ${sublimitTimeout ? `${sublimitTimeout}ms` : 'None'}`,
```

kyranet (approved):

vladfrangu (approved):

iCrawl (approved):

PR #5802, opened 2021-06-10
types(Extendable): add ButtonInteraction

ThatAnonyG:
While working on a TypeScript project with the master branch you will get an error when trying to extend a ButtonInteraction structure. The ButtonInteraction class is being used in the `INTERACTION_CREATE` event handler as `Structure.get("ButtonInteraction");` which implies that it can be extended. So I added the ButtonInteraction class under the Extendable type. Also removed the MessageComponentInteraction from the JSDoc since it was not extendable either.

### DIFF
--- a/src/util/Structures.js
+++ b/src/util/Structures.js
@@ -112,7 +112,6 @@ const structures = {
   Role: require('../structures/Role'),
   User: require('../structures/User'),
   CommandInteraction: require('../structures/CommandInteraction'),
-  MessageComponentInteraction: require('../structures/MessageComponentInteraction'),
   ButtonInteraction: require('../structures/ButtonInteraction'),
 };
 
--- a/src/util/Structures.js
+++ b/src/util/Structures.js
@@ -20,7 +20,6 @@
  * * **`Role`**
  * * **`User`**
  * * **`CommandInteraction`**
- * * **`MessageComponentInteraction`**
  * * **`ButtonInteraction`**
  * @typedef {string} ExtendableStructure
  */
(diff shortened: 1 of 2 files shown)


src/util/Structures.js
```diff
@@ -20,7 +20,6 @@
  * * **`Role`**
  * * **`User`**
  * * **`CommandInteraction`**
- * * **`MessageComponentInteraction`**
```
SpaceEEC, line 23: You also need to remove this line then:
https://github.com/discordjs/discord.js/blob/a56fa6dd7b5386bc4cbc44bc4e88f109e3419bb5/src/util/Structures.js#L115

SpaceEEC (dismissed):

src/util/Structures.js
ThatAnonyG: It has been removed. Thanks for the review.

monbrey:
imo `MessageComponentInteraction` should have been kept and added to the `Extendable` section, not removed.

ThatAnonyG:
> imo `MessageComponentInteraction` should have been kept and added to the `Extendable` section, not removed.

It is not being used currently so I think it would cause some confusion for the end-users. The devs can add it back when they will actually use the extended MessageComponentInteraction classes.

monbrey:
It is being used - ButtonInteraction and future component interactions will extend from it.

That makes it the logical structure to extend if you are adding properties that should exist on all component interactions.

vladfrangu (approved):

SpaceEEC (approved):

PR #5807, opened 2021-06-11
fix(ApiMessage): prevent sending unintended empty embeds

ckohen:

With discords new embeds change for bots, always sending embeds causes some strange behavior. e.g. editing content only results in removal of the embed.

This is a patch for v12 only that fixes this behavior.


vladfrangu (approved):

iCrawl:
Version 12 is not getting another release before v13 drops.

PR #5815, opened 2021-06-11
fix(TextBasedChannel): allow passing an APIMessage with split

almeidx:
Previously, replying to a message (or passing an APIMessage to a `TextBasedChannel#send`) with `split` enabled threw a `DiscordAPIError: Invalid Form Body`.

Example: 

```js
message.reply({ content: '<message content>', split: true });
```

Error:
```
content: Could not interpret "['```\n<message content>\n```']" as string.
```

### DIFF
--- a/src/structures/Webhook.js
+++ b/src/structures/Webhook.js
@@ -158,9 +158,10 @@ class Webhook {
       apiMessage = options.resolveData();
     } else {
       apiMessage = APIMessage.create(this, options).resolveData();
-      if (Array.isArray(apiMessage.data.content)) {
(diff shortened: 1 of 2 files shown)


SpaceEEC (changes requested):
There is another occurrence of this pattern in webhook, can you include that fix there too?
https://github.com/discordjs/discord.js/blob/e3e466d3e56145110ba0cfcba7b3c594267d36fb/src/structures/Webhook.js#L157-L164

iCrawl (approved):

vladfrangu (approved):

SpaceEEC (approved):

kyranet (approved):

PR #5816, opened 2021-06-11
chore(npm): ignore unnecessary files

almeidx:
These files are not necessary for a production environment.

I am unsure about the [`.tern-project`](https://github.com/discordjs/discord.js/blob/master/.tern-project) file. Should it be ignored as well?


kyranet:
Alternatively, you can change the npmignore with a [`"files"`](https://docs.npmjs.com/cli/v7/configuring-npm/package-json#files) field, so files aren't included unless explicitly added, as opposed to having to add new files/directories to the file as we add them.

kyranet (approved):

iCrawl (approved):

PR #5821, opened 2021-06-11
fix(GuildChannel): improve empty overwrite handling for permissionsLocked

### LINKED ISSUE #5819: GuildChannel#permissionsLocked does not handle empty overwrites

When looking at adding `GuildChannel#permissionsLocked` to my serverlog functionality, I discovered that it was behaving as I'd expect it to. Upon digging further, it appeared that in my situation there was an overwrite present in the D.js text channel object for the `@everyone` role that had no permissions allowed or denied, and that this overwrite simply wasn't present in the D.js parent category channel object.

This was the snippet of code that I discovered this via, where `613327370807672833` is the guild ID that I was testing in:

```js
newChannel.permissionOverwrites.forEach(value => {
                const parent = newChannel.parent.permissionOverwrites.get(value.id);
                console.log(value, parent);
...
```

```
PermissionOverwrites {
  id: '613327370807672833',
  type: 'role',
  deny: Permissions { bitfield: 0n },
  allow: Permissions { bitfield: 0n }
} undefined
```

As the overwrite is empty on the text channel, and not present on the category channel, they are functionally the same but the logic inside `GuildChannel#permissionsLocked` does not handle this situation. Fetching the parent category directly from Discord does not change this, the empty overwrite is still missing.

See snippet above, not sure how easy to reproduce, though it happens reliability for me locally with text channels until D.js gets an edit payload for the parent category.

**Further details:**

- discord.js version: 13 dev
- Node.js version: 14.15.4
- Operating system: macOS 10.15
- Priority this issue should have – please be realistic and elaborate if possible: P2

**Relevant client options:**

- partials: channel
- gateway intents: guilds
- other: none

- I have also tested the issue on latest master, commit hash: `e3e466d3e56145110ba0cfcba7b3c594267d36fb`

MattIPv4:

Fixes #5819 by handling the edge case where there is an empty overwrite present on a channel and simply not present on the parent, or vice-versa.

(I should note that this fix only worked when I was actively fetching channel parents on each channel update from the API, but I think that's just a side-effect/requirement of the channels partial now)

### DIFF
--- a/src/structures/GuildChannel.js
+++ b/src/structures/GuildChannel.js
@@ -95,13 +95,27 @@ class GuildChannel extends Channel {
    */
   get permissionsLocked() {
     if (!this.parent) return null;
-    if (this.permissionOverwrites.size !== this.parent.permissionOverwrites.size) return false;
-    return this.permissionOverwrites.every((value, key) => {
-      const testVal = this.parent.permissionOverwrites.get(key);
+
+    // Get all overwrites
+    const overwriteIds = new Set([...this.permissionOverwrites.keys(), ...this.parent.permissionOverwrites.keys()]);
+
+    // Compare all overwrites
+    return [...overwriteIds].every(key => {
+      const channelVal = this.permissionOverwrites.get(key);
+      const parentVal = this.parent.permissionOverwrites.get(key);
+
+      // Handle empty overwrite
+      // eslint-disable-next-line eqeqeq
+      if (channelVal === undefined && parentVal.deny.bitfield == 0 && parentVal.allow.bitfield == 0) return true;
+      // eslint-disable-next-line eqeqeq
+      if (parentVal === undefined && channelVal.deny.bitfield == 0 && channelVal.allow.bitfield == 0) return true;
+
+      // Compare overwrites
       return (
-        testVal !== undefined &&
-        testVal.deny.bitfield === value.deny.bitfield &&
-        testVal.allow.bitfield === value.allow.bitfield
+        channelVal !== undefined &&
+        parentVal !== undefined &&
+        channelVal.deny.bitfield === parentVal.deny.bitfield &&
+        channelVal.allow.bitfield === parentVal.allow.bitfield
       );
     });
   }
--- a/src/structures/GuildChannel.js
+++ b/src/structures/GuildChannel.js
@@ -5,6 +5,7 @@ const Invite = require('./Invite');
 const PermissionOverwrites = require('./PermissionOverwrites');
 const Role = require('./Role');
 const { Error, TypeError } = require('../errors');
+const BitField = require('../util/BitField');
 const Collection = require('../util/Collection');
 const { ChannelTypes } = require('../util/Constants');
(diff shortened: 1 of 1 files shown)


src/structures/GuildChannel.js
```diff
@@ -95,13 +95,27 @@ class GuildChannel extends Channel {
...
+    const overwriteIds = new Set([...this.permissionOverwrites.keys(), ...this.parent.permissionOverwrites.keys()]);
+
+    // Compare all overwrites
+    return [...overwriteIds].every(key => {
+      const channelVal = this.permissionOverwrites.get(key);
+      const parentVal = this.parent.permissionOverwrites.get(key);
+
+      // Handle empty overwrite
+      // eslint-disable-next-line eqeqeq
+      if (channelVal === undefined && parentVal.deny.bitfield == 0 && parentVal.allow.bitfield == 0) return true;
```
SpaceEEC, line 109: Is there a reason you are not using `=== 0n`?

MattIPv4: Hadn't initially worked when I wrote this fix, but will take another look 👍

MattIPv4: No idea what I did before, looks to be consistently `0n` now -- preference on `0n` vs `BitField.defaultBit`?

kyranet: We should use constants (`BitField.defaultBit`) over "magic numbers", so my preference is the latter.

Plus, in the case we somehow change the type, this would be one refactor less to do.

kyranet: Unresolving this because I noticed something:

https://github.com/discordjs/discord.js/blob/e3e466d3e56145110ba0cfcba7b3c594267d36fb/src/util/BitField.js#L164

The constant we want to compare with is `Permissions.defaultBit`, as `allow` and `deny` from `PermissionsOverwrites` are Permissions instances:

https://github.com/discordjs/discord.js/blob/e3e466d3e56145110ba0cfcba7b3c594267d36fb/src/util/Permissions.js#L141

src/structures/GuildChannel.js
```diff
@@ -95,13 +96,33 @@ class GuildChannel extends Channel {
...
+      const parentVal = this.parent.permissionOverwrites.get(key);
+
+      // Handle empty overwrite
+      if (
+        (channelVal === undefined &&
+          parentVal.deny.bitfield === BitField.defaultBit &&
+          parentVal.allow.bitfield === BitField.defaultBit) ||
+        (parentVal === undefined &&
+          channelVal.deny.bitfield === BitField.defaultBit &&
+          channelVal.allow.bitfield === BitField.defaultBit)
```
vladfrangu, line 115: BitField.defaultBit is 0
Permissions.defaultBit is 0n

This will never match with `===`, although APPARENTLY will with `==` (thanks JS.)

vladfrangu (changes requested):

kyranet (approved):
We could avoid a second iteration by changing `[...overwriteIds].every` to a for-of loop, but otherwise this is fine.

PR #5823, opened 2021-06-11
fix(GuildChannel): don't force parentID/permissionOverwrites to empty on create

MattIPv4:

In #5796 a change was made that forced `parentID` to be null and `permissionOverwrites` to be an empty collection on all GuildChannels when they are created. This PR changes that logic to only apply those defaults if they're not set by the initial patch during the super call.

This should resolve the issue of channels firing an update event with parentID changing when they're patched by an event for the first time since the client starts.

Kudos to [muchnameless](https://discord.com/channels/222078108977594368/682166281826598932/853036819264569364) for spotting this before I could.

### DIFF
--- a/src/structures/GuildChannel.js
+++ b/src/structures/GuildChannel.js
@@ -36,8 +36,8 @@ class GuildChannel extends Channel {
      */
     this.guild = guild;
 
-    this.parentID = this.parentID || null;
-    this.permissionOverwrites = this.permissionOverwrites || new Collection();
+    this.parentID = this.parentID ?? null;
+    this.permissionOverwrites = this.permissionOverwrites ?? new Collection();
   }
 
   _patch(data) {
--- a/src/structures/GuildChannel.js
+++ b/src/structures/GuildChannel.js
@@ -36,8 +36,8 @@ class GuildChannel extends Channel {
      */
     this.guild = guild;
 
-    this.parentID = null;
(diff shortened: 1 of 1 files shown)


src/structures/GuildChannel.js
```diff
@@ -36,8 +36,8 @@ class GuildChannel extends Channel {
      */
     this.guild = guild;
 
-    this.parentID = null;
-    this.permissionOverwrites = new Collection();
+    this.parentID = this.parentID || null;
+    this.permissionOverwrites = this.permissionOverwrites || new Collection();
```
kyranet, line 40:
```suggestion
    this.parentID ??= null;
    this.permissionOverwrites ??= new Collection();
```

kyranet (changes requested):
Otherwise LGTM.

kyranet (approved):

src/structures/GuildChannel.js
MattIPv4: I'd also originally written `||=` when I made this change locally and eslint got angry at me -- Node version bump when:tm:?

kyranet: We'll probably bump Node.js at the end of the year (around December) for Node.js v16 to become LTS and give users a grace period to upgrade if they didn't by that time.

iCrawl (approved):

SpaceEEC (approved):

PR #5830, opened 2021-06-12
feat(Interactions): auto-fetch replies where possible

monbrey:
Stemming from internal discussion, this is one of two opposing PRs which provide a way to automatically fetch the reply/deferral/update to an interaction, assuming it's not ephemeral.

This PR does the fetching automatically no matter what, returning a `Promise<Message>` where possible.

Note: I had to return `undefined` manually here instead of the blank `return` due to linting, unsure if this matters or if this still counts as `void`.

(Testing cross-PR closing) - Closes #5831 

**Status and versioning classification:**
- Code changes have been tested against the Discord API, or there are no code changes
- I know how to update typings and have done so, or typings don't need updating
- This PR includes breaking changes (methods removed or renamed, parameters moved or removed)

PR #5831, opened 2021-06-12
feat(Interactions): option to auto-fetch replies

monbrey:
Stemming from internal discussion, this is one of two opposing PRs which provide a way to automatically fetch the reply/deferral/update to an interaction, assuming it's not ephemeral.

This PR adds an option to the methods, allowing the user to enable auto-fetching. The default maintains the existing behaviour. I've written it such that attempting to auto-fetch an ephemeral reply, or auto-fetch an update to an ephemeral message throws an error - this can be changed of course.

(Testing cross-PR closing) - Closes #5830

### DIFF
--- a/src/structures/interfaces/InteractionResponses.js
+++ b/src/structures/interfaces/InteractionResponses.js
@@ -25,7 +25,7 @@ class InteractionResponses {
   /**
    * Defers the reply to this interaction.
    * @param {InteractionDeferOptions} [options] Options for deferring the reply to this interaction
-   * @returns {Promise<void>}
+   * @returns {Promise<Message|void>}
    * @example
    * // Defer the reply to this interaction
    * interaction.defer()
@@ -37,18 +37,21 @@ class InteractionResponses {
    *   .then(console.log)
    *   .catch(console.error);
    */
-  async defer({ ephemeral } = {}) {
+  async defer(options = {}) {
     if (this.deferred || this.replied) throw new Error('INTERACTION_ALREADY_REPLIED');
-    this.ephemeral = ephemeral ?? false;
+    if (options.fetchReply && options.ephemeral) throw new Error('INTERACTION_FETCH_EPHEMERAL');
+    this.ephemeral = options.ephemeral ?? false;
     await this.client.api.interactions(this.id, this.token).callback.post({
       data: {
         type: InteractionResponseTypes.DEFERRED_CHANNEL_MESSAGE_WITH_SOURCE,
         data: {
-          flags: ephemeral ? MessageFlags.FLAGS.EPHEMERAL : undefined,
+          flags: options.ephemeral ? MessageFlags.FLAGS.EPHEMERAL : undefined,
         },
       },
     });
     this.deferred = true;
+
+    return options.fetchReply ? this.fetchReply() : undefined;
   }
 
   /**
@@ -70,6 +73,7 @@ class InteractionResponses {
    */
   async reply(options) {
     if (this.deferred || this.replied) throw new Error('INTERACTION_ALREADY_REPLIED');
+    if (options.fetchReply && options.ephemeral) throw new Error('INTERACTION_FETCH_EPHEMERAL');
     this.ephemeral = options.ephemeral ?? false;
 
     let messagePayload;
@@ -86,6 +90,8 @@ class InteractionResponses {
       files,
     });
     this.replied = true;
+
+    return options.fetchReply ? this.fetchReply() : undefined;
   }
 
   /**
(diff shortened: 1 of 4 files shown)


src/structures/interfaces/InteractionResponses.js
```diff
@@ -140,28 +145,34 @@ class InteractionResponses {
...
    * @example
    * // Defer updating and reset the component's loading state
    * interaction.deferUpdate()
    *   .then(console.log)
    *   .catch(console.error);
    */
-  async deferUpdate() {
+  async deferUpdate(options = {}) {
     if (this.deferred || this.replied) throw new Error('INTERACTION_ALREADY_REPLIED');
+    if (options.fetchReply && (this.message.flags & MessageFlags.FLAGS.EPHEMERAL) === MessageFlags.FLAGS.EPHEMERAL) {
```
vladfrangu, line 159: Is `this.message`
1. A raw message
2. A d.js Message
3. could be either

if it's number 3, you should create an instance of MessageFlags instead to avoid breakage if `this.message` isn't RawMessage

src/structures/interfaces/InteractionResponses.js
```diff
@@ -174,6 +185,10 @@ class InteractionResponses {
   async update(options) {
     if (this.deferred || this.replied) throw new Error('INTERACTION_ALREADY_REPLIED');
 
+    if (options.fetchReply && (this.message.flags & MessageFlags.FLAGS.EPHEMERAL) === MessageFlags.FLAGS.EPHEMERAL) {
```
vladfrangu, line 188: Ditto

vladfrangu (dismissed):

src/structures/interfaces/InteractionResponses.js
monbrey: It's 3 - happy to proceed with that option, though this worked for both djs and raw messages

iCrawl:
This needs a rebase.

vladfrangu (approved):

src/structures/interfaces/InteractionResponses.js
```diff
@@ -85,6 +88,8 @@ class InteractionResponses {
       files,
     });
     this.replied = true;
+
+    return options.fetchReply ? this.fetchReply() : undefined;
```
NotSugden, line 92:
```suggestion
    if (options.fetchReply) return this.fetchReply();
```
looks a bit better but probably doesnt make a difference either way

src/structures/interfaces/InteractionResponses.js
```diff
@@ -140,28 +145,34 @@ class InteractionResponses {
...
+      throw new Error('INTERACTION_FETCH_EPHEMERAL');
+    }
     await this.client.api.interactions(this.id, this.token).callback.post({
       data: {
         type: InteractionResponseTypes.DEFERRED_MESSAGE_UPDATE,
       },
     });
     this.deferred = true;
+
+    return options.fetchReply ? this.fetchReply() : undefined;
```
NotSugden, line 169: same as previous

src/structures/interfaces/InteractionResponses.js
```diff
@@ -188,6 +203,8 @@ class InteractionResponses {
       files,
     });
     this.replied = true;
+
+    return options.fetchReply ? this.fetchReply() : undefined;
```
NotSugden, line 207: again

src/structures/interfaces/InteractionResponses.js
monbrey: This causes eslint's `consistent-return` rule to fail

NotSugden: oh :angry:

kyranet (approved):

iCrawl:
This needs a rebase.

monbrey:
Rebased

iCrawl:
This needs a rebase.

iCrawl (approved):

SpaceEEC:
I'm not sure what the benefit of this PR really is. (Probably because I missed the internal discussion about this)
Are we trying to "fix" Discord not responding with message data?

---

We merged and will probably still merge a bunch of PRs making overloads fewer and simpler.
This PR however is adding another and breaks passing object with an unknown fetch reply status, e.g.
```ts
const options: InteractionReplyOptions = this.getOptions(); // might have fetchReply, but might not
const maybeMessage = await interaction.reply(options); // option is not assignable to either overload
```
To fix this yet another overload is needed that returns `Message | RawMessage | void`.

---

Alternatively to this PR developers could do something like this: (Or anything equivalent to this)
```ts
const message = await interaction.reply("hello there").then(() => interaction.fetchReply())
```

monbrey:
Yeah @SpaceEEC, fixing Discord not responding with any data is exactly the intention :D

I don't really have any strong preference for this feature existing or not - a lot of users are already doing the manual fetching. If it's generally agreed to still not be in line with how the lib should be developed (internal fetches, less overloads etc) lets just close it.

kyranet:
If we want to keep code and overloads simple, we can simply not merge this PR and tell people to call that method themselves. There are multiple solutions for the issue at hand:

- We fetch for the reply with an HTTP call, has the problem that it makes the methods take longer and thus the bot becomes irresponsive for a brief period of time (the major use-case for this seems to be interaction collectors, after all).
- We collect a message from `MESSAGE_CREATE` that matches all the filters, then return it. It has the problems that it's bound to the `bot` scope (which is optional with `applications.commands`) and high CPU usage as it needs to filter many messages, specially in large internal sharding.
- We rewrite collectors to not require the response's message ID. I do not know what other use-cases do users have for fetching the reply.
- We educate our users about stateless and how they can use `custom_id` to avoid collectors, this is the solution I'll 100% go for (100 bytes is more than plenty of space if you use binary data), but I also understand that it's a very advanced and complex topic many users won't understand.

If there are more use-cases for this feature, please let me know and maybe we can think of an alternative solution.

iCrawl:
This needs a rebase.

vladfrangu (approved):

SpaceEEC (approved):

PR #5837, opened 2021-06-12
types(GuildApplicationCommandManager): allow undefined command

papsavas:
I have updated a certain `fetchPermissions()` overload with command ~~set as optional~~ removed so typescript won't complain when the method is called without parameters.
~~Also, included null type to `command`~~ removed too, no use case for it

### DIFF
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -2118,10 +2118,7 @@ declare module 'discord.js' {
     public edit(command: ApplicationCommandResolvable, data: ApplicationCommandData): Promise<ApplicationCommand>;
     public fetch(id: Snowflake, options?: BaseFetchOptions): Promise<ApplicationCommand>;
     public fetch(id?: Snowflake, options?: BaseFetchOptions): Promise<Collection<Snowflake, ApplicationCommand>>;
-    public fetchPermissions(
-      command?: ApplicationCommandResolvable | null,
-    ): Promise<Collection<Snowflake, ApplicationCommandPermissions[]>>;
-    public fetchPermissions(command: ApplicationCommandResolvable): Promise<ApplicationCommandPermissions[]>;
+    public fetchPermissions(command?: ApplicationCommandResolvable | null): Promise<ApplicationCommandPermissions[]>;
     public set(commands: ApplicationCommandData[]): Promise<Collection<Snowflake, ApplicationCommand>>;
     public setPermissions(
       command: ApplicationCommandResolvable,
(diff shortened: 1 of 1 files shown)


typings/index.d.ts
```diff
@@ -2118,7 +2118,7 @@ declare module 'discord.js' {
     public edit(command: ApplicationCommandResolvable, data: ApplicationCommandData): Promise<ApplicationCommand>;
     public fetch(id: Snowflake, options?: BaseFetchOptions): Promise<ApplicationCommand>;
     public fetch(id?: Snowflake, options?: BaseFetchOptions): Promise<Collection<Snowflake, ApplicationCommand>>;
-    public fetchPermissions(command: undefined): Promise<Collection<Snowflake, ApplicationCommandPermissions[]>>;
+    public fetchPermissions(command?: null): Promise<Collection<Snowflake, ApplicationCommandPermissions[]>>;
     public fetchPermissions(command: ApplicationCommandResolvable): Promise<ApplicationCommandPermissions[]>;
```
almeidx, line 2122:
```suggestion
    public fetchPermissions(): Promise<Collection<Snowflake, ApplicationCommandPermissions[]>>;
    public fetchPermissions(command: ApplicationCommandResolvable): Promise<ApplicationCommandPermissions[]>;
```

almeidx (changes requested):

typings/index.d.ts
papsavas: not sure if this will allow passing null values but then again cant really think of a reason it should

almeidx: Why should it allow passing null values?

papsavas: it was suggested by @vaporox but i believe passing null should not mean "fetch me all command perms instead"

almeidx: The argument is just listed as optional on the JSDoc of the function implementation, the type isn't `ApplicationCommandResolvable|null`

https://github.com/discordjs/discord.js/blob/02693bc02f45980d8165820a103220f0027b96b7/src/managers/ApplicationCommandManager.js#L212

vaporox: Completely removing the parameter here would also make sense, however for the `ApplicationCommandManager#fetchPermissions()` overload at https://github.com/discordjs/discord.js/blob/master/typings/index.d.ts#L2078-L2081 (which needs to be updated too), removing it is not possible because it is followed by another parameter, so I thought it would make sense to stay consistent between the two methods.

almeidx: Perhaps making the function take an "options" parameter would be a better approach than the current one

papsavas: Yeah, it also comes from another context since it concerns the application command manager rather than the guild based one such as my updated overload, so there is no need for a null ApplicationCommandResolvable scenario

almeidx (approved):

vladfrangu (approved):

iCrawl (approved):

kyranet (approved):

SpaceEEC (approved):

PR #5838, opened 2021-06-12
fix(ApplicationCommand): default option.required to false

ImRodry:
The data sent by Discord for the `required` field in ApplicationCommandOption is undefined if the option is not required. This PR fixes this so that it defaults to false instead to make it easier to compare with local commands for command updaters.

### DIFF
--- a/src/structures/ApplicationCommand.js
+++ b/src/structures/ApplicationCommand.js
@@ -210,11 +210,13 @@ class ApplicationCommand extends Base {
    * @private
    */
   static transformOption(option, received) {
+    const stringType = typeof option.type === 'string' ? option.type : ApplicationCommandOptionTypes[option.type];
     return {
       type: typeof option.type === 'number' && !received ? option.type : ApplicationCommandOptionTypes[option.type],
       name: option.name,
       description: option.description,
-      required: option.required ?? false,
+      required:
+        option.required ?? (stringType === 'SUB_COMMAND' || stringType === 'SUB_COMMAND_GROUP') ? undefined : false,
       choices: option.choices,
       options: option.options?.map(o => this.transformOption(o, received)),
     };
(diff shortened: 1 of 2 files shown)


kyranet (approved):

src/structures/ApplicationCommand.js
```diff
@@ -214,7 +214,7 @@ class ApplicationCommand extends Base {
       type: typeof option.type === 'number' && !received ? option.type : ApplicationCommandOptionTypes[option.type],
       name: option.name,
       description: option.description,
-      required: option.required,
+      required: option.required ?? false,
```
iCrawl, line 217: I'm not sure this is needed, because once you stringify this the required key will be dropped, so it would correct as require by discord be optional and not be sent in the payload.

ImRodry: But this data isn’t sent anywhere, it’s the data that you get when you fetch the commands from Discord. If that’s not what you mean then I’m not sure I understand

iCrawl: Right, I was under the assumption the transform is also used before stringifying it and sending it to discord 👍

iCrawl (approved):

src/structures/ApplicationCommand.js
ImRodry: And it is which is why you have the received boolean but that only affects the type value

ImRodry:
This didn’t cross my mind when making the PR but SUB_COMMAND and SUB_COMMAND_GROUP types cannot have the required value set so I’ll need to do some adjusting to catch those cases and set it to undefined then

ImRodry:
Oopsie

vladfrangu (approved):

SpaceEEC (approved):

PR #5840, opened 2021-06-13
feat(MessageComponentInteraction): component getter

### LINKED ISSUE #5828: property/method to concisely reference originating component of a component interaction
I'd like to update message components dynamically in response to interactions. The most obvious, broadly applicable use here would be setting a message component's `disabled` boolean in response to some change in the interaction messsage's content. The way I see it: if a button's to remain statically `disabled: true`, chances are good that it shouldn't have been rendered in the first place. And if most such buttons are being created with the intention of updating them, my gut feeling is that it should to be easy to reference said buttons in response to incoming interactions.

But so far, the only way I can see to reference the originating component from an interaction object is with something like: `ComponentInteraction.message.components.map(actionRow => actionRow.components.find(component => component.customID === ComponentInteraction.customID))`. To me, that feels more difficult than it ought to be ought to be for the "stateful" components described by Discord's API documentation:

Screen Shot 2021-06-12 at 12 43 19 AM

Referencing the originating `MessageComponent` of an interaction through a property or method, e.g. `ComponentInteraction.component` or `ComponentInteraction.getSrc()`.

A long reference chain. Admittedly, it's not the end of the world, but it's inconvenient and not especially easy to read.

monbrey:
Closes #5828

This currently uses generic Object/unknown typings for a raw message component, pending https://github.com/discordjs/discord-api-types/pull/132 being merged.

### DIFF
--- a/src/structures/MessageComponentInteraction.js
+++ b/src/structures/MessageComponentInteraction.js
@@ -59,7 +59,7 @@ class MessageComponentInteraction extends Interaction {
 
   /**
    * The component which was interacted with
-   * @type {?MessageActionRowComponent|Object}
+   * @type {?(MessageActionRowComponent|Object)}
    * @readonly
    */
   get component() {
--- a/src/structures/MessageComponentInteraction.js
+++ b/src/structures/MessageComponentInteraction.js
@@ -66,7 +66,7 @@ class MessageComponentInteraction extends Interaction {
     return this.message.components
       .map(row => row.components)
       .flat()
-      .find(component => (component.customID ?? component.custom_id) === this.customID);
+      .find(component => (component.customID ?? component.custom_id) === this.customID) ?? null;
   }
 
   /**
--- a/src/structures/MessageComponentInteraction.js
+++ b/src/structures/MessageComponentInteraction.js
@@ -59,14 +59,14 @@ class MessageComponentInteraction extends Interaction {
 
   /**
    * The component which was interacted with
-   * @type {?MessageActionRowComponent}
+   * @type {?MessageActionRowComponent|Object}
    * @readonly
    */
   get component() {
     return this.message.components
       .map(row => row.components)
       .flat()
-      .find(component => component.customID === this.customID);
+      .find(component => (component.customID ?? component.custom_id) === this.customID);
   }
 
   /**
--- a/src/structures/MessageComponentInteraction.js
+++ b/src/structures/MessageComponentInteraction.js
@@ -58,6 +58,18 @@ class MessageComponentInteraction extends Interaction {
   }
 
   /**
+   * The component which was interacted with
+   * @type {?MessageActionRowComponent}
+   * @readonly
+   */
+  get component() {
+    return this.message.components
+      .map(row => row.components)
+      .flat()
+      .find(component => component.customID === this.customID);
+  }
+
+  /**
    * Resolves the type of a MessageComponent
    * @param {MessageComponentTypeResolvable} type The type to resolve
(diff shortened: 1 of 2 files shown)


src/structures/MessageComponentInteraction.js
```diff
@@ -51,6 +51,18 @@ class MessageComponentInteraction extends Interaction {
     this.webhook = new InteractionWebhook(this.client, this.applicationID, this.token);
   }
 
+  /**
+   * The component which was interacted with
+   * @type {?MessageActionRowComponent|Object}
```
vladfrangu, line 56:
```suggestion
   * @type {?(MessageActionRowComponent|Object)}
```

src/structures/MessageComponentInteraction.js
```diff
@@ -51,6 +51,18 @@ class MessageComponentInteraction extends Interaction {
...
+   * @readonly
+   */
+  get component() {
+    return this.message.components
+      .map(row => row.components)
+      .flat()
```
vladfrangu, line 62: Is `flat` defaulted to Infinity?

vladfrangu (changes requested):

src/structures/MessageComponentInteraction.js
monbrey: Default depth is 1, though not sure it makes a difference here based on the data the API gives us.

vladfrangu:
The required types PR is merged and deployed as a `next` version, but not as a release yet! I want to restructure some things in there, so... for now, should we target `@next`? CC @iCrawl @kyranet

SpaceEEC (changes requested):
Otherwise LGTM

iCrawl:
Lint CI is failing here.

src/structures/MessageComponentInteraction.js
```diff
@@ -51,6 +51,18 @@ class MessageComponentInteraction extends Interaction {
...
 
+  /**
+   * The component which was interacted with
+   * @type {?(MessageActionRowComponent|Object)}
+   * @readonly
+   */
+  get component() {
+    return this.message.components
+      .map(row => row.components)
+      .flat()
```
vaporox, line 62: 🧹

```suggestion
      .flatMap(row => row.components)
```

src/structures/MessageComponentInteraction.js
```diff
@@ -51,6 +51,18 @@ class MessageComponentInteraction extends Interaction {
...
+      .find(component => (component.customID ?? component.custom_id) === this.customID) ?? null;
```
DTrombett, line 63: This should fix ESLint maybe?
```suggestion
    return (
      this.message.components
        .map(row => row.components)
        .flat()
        .find(component => (component.customID ?? component.custom_id) === this.customID) ?? null
    );
```

SpaceEEC: `?? null` should be used on the return value of `Array#find` here to ensure that the getter returns `null` and not `undefined` if the component wasn't found.

Moving it into the predicate callback does nothing as `boolean ?? null` will never return `null` (and `false` is perfectly fine to return there).

iCrawl: @DTrombett Trying to manually fix ESLint is not what should happen here, we have hooks for that for a reason.

iCrawl:
This needs a rebase too.

iCrawl (approved):

vladfrangu (approved):

kyranet (approved):

PR #5841, opened 2021-06-13
fix(Guild): setChannelPositions null parenting

monbrey:
This fixes a couple of issues introduced in #5507 where not providing a parent to `Guild#setChannelPositions` would resolve to `null`, clearing the parent instead of leaving it unchanged.

This also prevented the method from being used to set more than one channel position, due to `DiscordAPIError: Only one channel can have a parent_id modified at a time`.

Additionally, the typings are updated to accept `null` in order to move the channel out of the parent category.

### DIFF
--- a/src/structures/Guild.js
+++ b/src/structures/Guild.js
@@ -1141,7 +1141,7 @@ class Guild extends BaseGuild {
       id: this.client.channels.resolveID(r.channel),
       position: r.position,
       lock_permissions: r.lockPermissions,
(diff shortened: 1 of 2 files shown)


kyranet (approved):
Nice catch on the parenting issue ^^

iCrawl (approved):

vladfrangu (approved):

SpaceEEC (approved):

PR #5843, opened 2021-06-13
feat(*): use enums for consistency and speed

iShibi:

Okay, there are a lot of related changes here:

1) Use enums instead of arrays. This brings consistency and speed (`indexOf` is slow).
2) Some enums (that were arrays earlier) had same name for the enums themselves and the type of elements they hold.
3) `WebhookTypes` enum (ex-array) was missing from typings for `Constants`
4) Some methods that used arrays as enums earlier wouldn't let user to pass `number` type.
5) `DefaultMessageNotifcations` is now `DefaultMessageNotificationLevels` to keep the naming consistent with other enums in djs (discord uses 'level' in the name for enums in documentaion).
6) `DefaultMessageNotifcations` types were not same as what the api has documented ( eg. `ALL` instead of `ALL_MESSAGES`)
6) Although, the raw activity data's `name` property's value is `Custom_Status`, the `type` for it is documented as `Custom`. So, the activity type `CUSTOM_STATUS` is now `CUSTOM`.  

- This PR includes breaking changes

### DIFF
--- a/src/managers/GuildManager.js
+++ b/src/managers/GuildManager.js
@@ -171,17 +171,11 @@ class GuildManager extends BaseManager {
     if (typeof verificationLevel === 'string') {
       verificationLevel = VerificationLevels[verificationLevel];
     }
-    if (defaultMessageNotifications) {
-      defaultMessageNotifications =
-        typeof defaultMessageNotifications === 'number'
-          ? defaultMessageNotifications
-          : DefaultMessageNotificationLevels[defaultMessageNotifications];
+    if (typeof defaultMessageNotifications === 'string') {
+      defaultMessageNotifications = DefaultMessageNotificationLevels[defaultMessageNotifications];
     }
-    if (explicitContentFilter) {
-      explicitContentFilter =
-        typeof explicitContentFilter === 'number'
-          ? explicitContentFilter
-          : ExplicitContentFilterLevels[explicitContentFilter];
+    if (typeof explicitContentFilter === 'string') {
+      explicitContentFilter = ExplicitContentFilterLevels[explicitContentFilter];
     }
     for (const channel of channels) {
       if (channel.type) channel.type = ChannelTypes[channel.type.toUpperCase()];
--- a/src/managers/GuildManager.js
+++ b/src/managers/GuildManager.js
@@ -13,7 +13,7 @@ const {
   ChannelTypes,
   Events,
   VerificationLevels,
-  DefaultMessageNotifications,
+  DefaultMessageNotificationLevels,
   ExplicitContentFilterLevels,
 } = require('../util/Constants');
 const DataResolver = require('../util/DataResolver');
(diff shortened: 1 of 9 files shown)


src/managers/GuildManager.js
```diff
@@ -163,14 +164,21 @@ class GuildManager extends BaseManager {
     } = {},
   ) {
     icon = await DataResolver.resolveImage(icon);
-    if (typeof verificationLevel !== 'undefined' && typeof verificationLevel !== 'number') {
-      verificationLevel = VerificationLevels.indexOf(verificationLevel);
+    if (verificationLevel) {
+      verificationLevel =
+        typeof verificationLevel === 'number' ? verificationLevel : VerificationLevels[verificationLevel];
```
kyranet, line 169: You could probably change this to:
```suggestion
    if (typeof verificationLevel === 'string') {
      verificationLevel = VerificationLevels[verificationLevel];
```

And same for the others, those values take either a string or a number, one is untouched, and the other is transformed.

kyranet: Same goes fo the other checks.

kyranet (approved):

kyranet (changes requested):
Sorry, wrong button... ^

kyranet (approved):

src/util/Constants.js
```diff
@@ -708,38 +708,28 @@ exports.APIErrors = {
...
 
 /**
- * The value set for a guild's default message notifications, e.g. `ALL`. Here are the available types:
- * * ALL
- * * MENTIONS
- * @typedef {string} DefaultMessageNotifications
+ * The value set for a guild's default message notifications, e.g. `ALL_MESSAGES`. Here are the available types:
+ * * ALL_MESSAGES
+ * * ONLY_MENTIONS
+ * @typedef {string} DefaultMessageNotificationLevel
```
vladfrangu, line 714: Why is this renamed, when the property has the same name

src/util/Constants.js
```diff
@@ -708,38 +708,28 @@ exports.APIErrors = {
...
  */
-exports.DefaultMessageNotifications = ['ALL', 'MENTIONS'];
+exports.DefaultMessageNotificationLevels = createEnum(['ALL_MESSAGES', 'ONLY_MENTIONS']);
 
 /**
  * The value set for a team members's membership state:
  * * INVITED
  * * ACCEPTED
- * @typedef {string} MembershipStates
+ * @typedef {string} MembershipState
```
vladfrangu, line 722: Why is this renamed, when the property has the same name

src/util/Constants.js
```diff
@@ -708,38 +708,28 @@ exports.APIErrors = {
...
-  'ACCEPTED',
-];
+exports.MembershipStates = createEnum([null, 'INVITED', 'ACCEPTED']);
 
 /**
  * The value set for a webhook's type:
  * * Incoming
  * * Channel Follower
- * @typedef {string} WebhookTypes
+ * @typedef {string} WebhookType
```
vladfrangu, line 730: Why is this renamed, when the property has the same name

src/structures/Guild.js
```diff
@@ -850,17 +850,17 @@ class Guild extends BaseGuild {
     if (data.splash) _data.splash = data.splash;
     if (data.discoverySplash) _data.discovery_splash = data.discoverySplash;
     if (data.banner) _data.banner = data.banner;
-    if (typeof data.explicitContentFilter !== 'undefined') {
+    if (data.explicitContentFilter) {
```
vladfrangu, line 853: This won't match if the filter is 0

src/structures/Guild.js
```diff
@@ -832,11 +832,11 @@ class Guild extends BaseGuild {
   edit(data, reason) {
     const _data = {};
     if (data.name) _data.name = data.name;
-    if (typeof data.verificationLevel !== 'undefined') {
+    if (data.verificationLevel) {
```
vladfrangu, line 835: This won't match if the level is 0

src/structures/Guild.js
```diff
@@ -850,17 +850,17 @@ class Guild extends BaseGuild {
...
-    if (typeof data.explicitContentFilter !== 'undefined') {
+    if (data.explicitContentFilter) {
       _data.explicit_content_filter =
         typeof data.explicitContentFilter === 'number'
           ? data.explicitContentFilter
-          : ExplicitContentFilterLevels.indexOf(data.explicitContentFilter);
+          : ExplicitContentFilterLevels[data.explicitContentFilter];
     }
-    if (typeof data.defaultMessageNotifications !== 'undefined') {
+    if (data.defaultMessageNotifications) {
```
vladfrangu, line 859: This won't match if the level is 0

vladfrangu (changes requested):

src/util/Constants.js
iShibi: That is how all the enums are documented. The enum is named plural and the type of elements it stores have the singular name. Eg. enum: `NSFWLevels`, the type it stores: `NSFWLevel`

vladfrangu: 👍

src/structures/Guild.js
iShibi: yeah, right. I'll fix it 😸

SpaceEEC (approved):

iCrawl (approved):

iCrawl:
This needs a rebase.

iShibi:
Rebased

Issue #5845, opened 2021-06-13
Presence updates on some activities are skipped


JMTK:
**Please describe the problem you are having in as much detail as possible:**
When playing League of Legends, I was trying to get presence updates for the state updates(In Queue, In Lobby, In Champion Select), however the Presence Update never fired. So I put a few breakpoints in and noticed that it skips over the emit because it finds the 2 presences equal.

**Include a reproducible code sample here, if possible:**
```js
//PresenceUpdate.js
if (member && this.client.listenerCount(Events.PRESENCE_UPDATE) && !member.presence.equals(oldPresence)) {

//Print JSON.stringify(oldPresence, null, 2)
{
  "userID": "82988288729681920",
  "guild": "86908735326281728",
  "status": "online",
  "activities": [
    {
      "id": "custom",
      "name": "Custom Status",
      "type": "CUSTOM_STATUS",
      "url": null,
      "details": null,
      "state": "hello",
      "applicationID": null,
      "timestamps": null,
      "platform": null,
      "party": null,
      "assets": null,
      "flags": 0,
      "emoji": null,
      "sessionID": null,
      "buttons": [],
      "createdTimestamp": 1623594409944
    },
    {
      "id": "dc42c02ce69f4cfc",
      "name": "League of Legends",
      "type": "PLAYING",
      "url": null,
      "details": null,
      "state": "In Lobby",
      "applicationID": "401518684763586560",
      "timestamps": null,
      "platform": null,
      "party": null,
      "flags": 0,
      "emoji": null,
      "sessionID": null,
      "buttons": [],
      "createdTimestamp": 1623595600463
    }
  ]
}

//Print JSON.stringify(member.presence, null, 2)
{
  "userID": "82988288729681920",
  "guild": "86908735326281728",
  "status": "online",
  "activities": [
    {
      "id": "custom",
      "name": "Custom Status",
      "type": "CUSTOM_STATUS",
      "url": null,
      "details": null,
      "state": "hello",
      "applicationID": null,
      "timestamps": null,
      "platform": null,
      "party": null,
      "assets": null,
      "flags": 0,
      "emoji": null,
      "sessionID": null,
      "buttons": [],
      "createdTimestamp": 1623594409944
    },
    {
      "id": "dc42c02ce69f4cfc",
      "name": "League of Legends",
      "type": "PLAYING",
      "url": null,
      "details": "Summoner's Rift",
      "state": "In Champion Select",
      "applicationID": "401518684763586560",
[...]

PR #5846, opened 2021-06-13
fix(presenceUpdate): fire when only state/details change on an activity

JMTK:
Add support for firing presence update events when the only thing that changes is an activity's state or details.

Fixes https://github.com/discordjs/discord.js/issues/5845

### DIFF
--- a/src/structures/Presence.js
+++ b/src/structures/Presence.js
@@ -275,11 +275,12 @@ class Activity {
   equals(activity) {
     return (
       this === activity ||
-      (activity && this.name === activity.name &&
-       this.type === activity.type &&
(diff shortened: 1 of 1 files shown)


vladfrangu (changes requested):
Can you fix lint? Otherwise, if details isn't an object (in which case you'll need to check its keys), LGTM

JMTK:
Sorry about that! There was an extra space that the lint was complaining about that I had missed. Should be good now!

iCrawl (approved):

kyranet (approved):

SpaceEEC (approved):

PR #5847, opened 2021-06-13
fix(Message): editedTimestamp defaulting to 0

ImRodry:
Since edited_timestamp is always present in the message data from Discord (see https://discord.com/developers/docs/resources/channel#message-object), the editedTimestamp property for messages that have never been edited would be 0. This fixes that so that it's null when it hasn't been edited, and the timestamp for edited messages, just like the data from Discord.

### DIFF
--- a/src/structures/Message.js
+++ b/src/structures/Message.js
@@ -163,7 +163,7 @@ class Message extends Base {
      * The timestamp the message was last edited at (if applicable)
      * @type {?number}
      */
-    this.editedTimestamp = 'edited_timestamp' in data ? new Date(data.edited_timestamp).getTime() : null;
+    this.editedTimestamp = data.edited_timestamp ? new Date(data.edited_timestamp).getTime() : null;
 
     /**
      * A manager of the reactions belonging to this message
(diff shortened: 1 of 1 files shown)


src/structures/Message.js
```diff
@@ -290,7 +290,7 @@ class Message extends Base {
   patch(data) {
     const clone = this._clone();
 
-    if ('edited_timestamp' in data) this.editedTimestamp = new Date(data.edited_timestamp).getTime();
+    if (data.edited_timestamp !== null) this.editedTimestamp = new Date(data.edited_timestamp).getTime();
```
vladfrangu, line 293: Since undefined is !== null, won't this possibly become NaN now?

vladfrangu (dismissed):

src/structures/Message.js
ImRodry: It will never be undefined though as the docs suggest

SpaceEEC: Patching or creating a message with partial data may not contain that property at all.

ImRodry: Fair enough, I'll change it to !data.edited_timestamp

kyranet (approved):

SpaceEEC (approved):

iCrawl (approved):

PR #5848, opened 2021-06-13
fix(ApplicationCommand): default option.required fix

muchnameless:
#5838 introduced a bug that made it impossible to set the required option to `true` due to operator precedence, this PR fixes that

### DIFF
--- a/src/structures/ApplicationCommand.js
+++ b/src/structures/ApplicationCommand.js
@@ -216,7 +216,7 @@ class ApplicationCommand extends Base {
       name: option.name,
       description: option.description,
       required:
-        option.required ?? (stringType === 'SUB_COMMAND' || stringType === 'SUB_COMMAND_GROUP') ? undefined : false,
+        option.required ?? ((stringType === 'SUB_COMMAND' || stringType === 'SUB_COMMAND_GROUP') ? undefined : false),
       choices: option.choices,
       options: option.options?.map(o => this.transformOption(o, received)),
     };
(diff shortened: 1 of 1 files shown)


src/structures/ApplicationCommand.js
```diff
@@ -216,7 +216,7 @@ class ApplicationCommand extends Base {
       name: option.name,
       description: option.description,
       required:
-        option.required ?? (stringType === 'SUB_COMMAND' || stringType === 'SUB_COMMAND_GROUP') ? undefined : false,
+        option.required ?? (stringType === 'SUB_COMMAND' || stringType === 'SUB_COMMAND_GROUP' ? undefined : false),
```
kyranet, line 219: It probably meant to do this?
```suggestion
        option.required ?? ((stringType === 'SUB_COMMAND' || stringType === 'SUB_COMMAND_GROUP') ? undefined : false),
```

kyranet (changes requested):
Nice catch! Just one thing:

src/structures/ApplicationCommand.js
muchnameless: this is what I had at first but prettier removed the inner brackets

kyranet (approved):

src/structures/ApplicationCommand.js
kyranet: That means they're not required, thanks for checking!

ImRodry:
Oopsie! Tbh I got quite confused when husky asked me to add those brackets and I was scared it would break something but thought it was ok since no one mentioned it. Nice catch!

iCrawl (approved):

PR #5849, opened 2021-06-13
add optionalDependencies 

colenh:
This adds optionalDependencies to the package.json. This allows for easier addition of the optional packages listed in the readme file. 

To install without these optional packages you use: `npm install --no-optional discord.js` *or* `yarn add discord.js --ignore-optional`

To install with the optional packages you would run the command like usual.

### DIFF
--- a/package.json
+++ b/package.json
@@ -77,6 +77,13 @@
     "tslint": "^6.1.3",
     "typescript": "^4.3.2"
   },
+  "optionalDependencies": {
+    "zlib-sync": "^0.1.7",
+    "discord/erlpack": "^0.1.3",
+    "bufferutil": "^4.0.3",
+    "utf-8-validate": "^5.0.5",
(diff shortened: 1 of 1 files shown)


kyranet:
`optionalDependencies` doesn't have the behaviour we want (opt-in, rather than opt-out), and the vast majority of users don't understand what each package is for - they'll just run the install script without the flag (as they do for basically all other dependencies), which would then install all those dependencies (and some are troublesome, as they enable certain parts which can cause unintended side effects, such as Erlpack).

Secondly, you can't opt-out for specific dependencies, this is all-or-none, and even if it did allow you, it's a lot easier to just install the specific dependencies yourself.

Finally, we used `peerDependencies` for it's opt-in behaviour, and since they became opt-out, we removed them completely from the package.

colenh:
Makes a lot of sense, thank you.

PR #5869, opened 2021-06-16
Update README.md

snpbot:
pp


kyranet:
This change is not necessary and turns a valid code example into an invalid one.

PR #5877, opened 2021-06-19
Options are meant to be optional at least for Client.

LaterIdiot:
This PR provides a patch to the typings in which the Client constructor now makes options optional in the typings.

### DIFF
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -370,7 +370,7 @@ declare module 'discord.js' {
   }
 
   export class Client extends BaseClient {
-    constructor(options: ClientOptions);
+    constructor(options?: ClientOptions);
     private actions: unknown;
     private _eval(script: string): any;
     private _validateOptions(options: ClientOptions): void;


kyranet:
They're not optional because if an empty object is passed, the constructor will always throw an error:

https://github.com/discordjs/discord.js/blob/807ea2d3c197c7c6556bd1c894114932f6657d49/src/client/Client.js#L467

This method is called in:

https://github.com/discordjs/discord.js/blob/807ea2d3c197c7c6556bd1c894114932f6657d49/src/client/Client.js#L76

You must always specify `options.intents`, that's an intended breaking change.

LaterIdiot:
> They're not optional because if an empty object is passed, the constructor will always throw an error:
> 
> https://github.com/discordjs/discord.js/blob/807ea2d3c197c7c6556bd1c894114932f6657d49/src/client/Client.js#L467
> 
> This method is called in:
> 
> https://github.com/discordjs/discord.js/blob/807ea2d3c197c7c6556bd1c894114932f6657d49/src/client/Client.js#L76
> 
> You must always specify `options.intents`, that's an intended breaking change.

I meant to say that the argument for 'options' is optional not the options themselves.

kyranet:
`new Client()` without arguments produces this:

image

And if I do this change:

```diff
- const client = new Client();
+ const client = new Client({ intents:['GUILDS'] });
```

Then the bot starts working correctly.

LaterIdiot:
> `new Client()` without arguments produces this:
> 
> image
> 
> And if I do this change:
> 
> ```diff
> - const client = new Client();
> + const client = new Client({ intents:['GUILDS'] });
> ```
> 
> Then the bot starts working correctly.

Yea you are right I made a mistake of not testing after installing discord.js v13 sorry for this PR. I didn't even check the documentation for the master branch which it clearly stated that it's not optional but necessary.

PR #5880, opened 2021-06-19
Resolving subcommand and subcommand group for CommandInteractions

jaydamani:

Added CommandInteraction#subcommand and CommandInteraction#group

### DIFF
--- a/src/structures/CommandInteraction.js
+++ b/src/structures/CommandInteraction.js
@@ -99,7 +99,7 @@ class CommandInteraction extends Interaction {
       type: ApplicationCommandOptionTypes[option.type],
     };
 
-    if (!('value' in option)) return
+    if (!('value' in option)) return result
     result.value = option.value;
 
     const user = resolved?.users?.[option.value];
@@ -129,18 +129,18 @@ class CommandInteraction extends Interaction {
 
     if (options?.[0].type === 2) {
       /**
-			 * @type {string}
-			 */
+       * @type {string}
+       */
       this.group = options[0].name
       options = options[0].options
-		}
+    }
     if (options?.[0].type === 1) {
       /**
-			 * @type {string}
-			 */
+       * @type {string}
+       */
       this.subcommand = options[0].name
       options = options[0].options
-		}
+    }
 
     if (typeof options === 'undefined') return optionsCollection;
     for (const option of options) {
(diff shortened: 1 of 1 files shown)


monbrey:
I don't see how this is necessary - there are no issues with sub command or sub command group options.

Why do you believe this change is needed?

jaydamani:
With this, lib would do the parsing and figure out the subcommand / subcommand group while options is just a map of arguments provided by the user. Not necessary but could be kinda handy.

monbrey:
It's not "parsing" anything though. Looks to me like you just hoist the options up a level, but then don't do any of the existing handling on them.

Just seems worse to me.

monbrey:
@jaydamani Hey, apologies, I think I completely misread the diff on this change. I've just tested your proposed changes and I think this could be beneficial. Do you want to reopen this for reviews?

jaydamani:
No problem, sorry for the delay ~~I should check my mail more often~~, kinda my bad as well, since I couldn't explain it properly

src/structures/CommandInteraction.js
```diff
@@ -128,6 +126,22 @@ class CommandInteraction extends Interaction {
    */
   _createOptionsCollection(options, resolved) {
     const optionsCollection = new Collection();
+
+    if (options?.[0].type === 2) {
+      /**
+       * @type {string}
+       */
+      this.group = options[0].name;
```
NotSugden, line 134: isn"t there an enum for this?

src/structures/CommandInteraction.js
```diff
@@ -128,6 +126,22 @@ class CommandInteraction extends Interaction {
...
+      options = options[0].options;
+    }
+    if (options?.[0].type === 1) {
+      /**
+       * @type {string}
```
NotSugden, line 139: ...? description?

src/structures/CommandInteraction.js
```diff
@@ -128,6 +126,22 @@ class CommandInteraction extends Interaction {
    */
   _createOptionsCollection(options, resolved) {
     const optionsCollection = new Collection();
+
+    if (options?.[0].type === 2) {
+      /**
+       * @type {string}
```
NotSugden, line 132: ^

NotSugden (changes requested):

iCrawl:
This will be handled in #6107

PR #5886, opened 2021-06-20
fix(Message): flags not being parsed on some edits

monbrey:
One of those rabbit hole bugs that made me change a few things.

To start with, `Message#suppressEmbeds` was broken. This is because:

1. Internally that method calls `Message#edit({ flags: 4 })`.
2. That method then calls `MessageManager#edit(this.id, options)`
3. `MessageManager#edit` constructs an `APIMessage` with `this` as the target.
4. `APIMessage#resolveData` only resolves `flags` if the `APIMessage#target` is a `Message` or `Interaction`.
https://github.com/discordjs/discord.js/blob/master/src/structures/APIMessage.js#L159-L164

So, I decided that:

1. `Message#edit` should pass `this` rather than `this.id`.
2. `MessageManager#edit` should pass the whole message to `APIMessage#target` if provided with one.
3. `APIMessage` should support `MessageManager` as a target also, for when making cacheless edits by ID.

### DIFF
--- a/src/managers/MessageManager.js
+++ b/src/managers/MessageManager.js
@@ -131,8 +131,9 @@ class MessageManager extends BaseManager {
       .resolveFiles();
     const d = await this.client.api.channels[this.channel.id].messages[messageID].patch({ data, files });
 
-    if (this.cache.has(messageID)) {
-      const clone = this.cache.get(messageID)._clone();
+    const existing = this.cache.get(messageID);
+    if (existing) {
+      const clone = existing._clone();
       clone._patch(d);
       return clone;
     }
--- a/src/managers/MessageManager.js
+++ b/src/managers/MessageManager.js
@@ -120,16 +120,19 @@ class MessageManager extends BaseManager {
    * @returns {Promise<Message>}
    */
   async edit(message, options) {
-    message = this.resolveID(message);
-    if (!message) throw new TypeError('INVALID_TYPE', 'message', 'MessageResolvable');
+    const messageID = this.resolveID(message);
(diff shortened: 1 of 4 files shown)


src/managers/MessageManager.js
```diff
@@ -120,16 +120,19 @@ class MessageManager extends BaseManager {
    * @returns {Promise<Message>}
    */
   async edit(message, options) {
-    message = this.resolveID(message);
```
iCrawl, line 123: No more mutation, hooray 🧹

iCrawl (approved):

src/managers/MessageManager.js
```diff
@@ -120,16 +120,19 @@ class MessageManager extends BaseManager {
...
+      : APIMessage.create(message instanceof Message ? message : this, options)
+    )
       .resolveData()
       .resolveFiles();
-    const d = await this.client.api.channels[this.channel.id].messages[message].patch({ data, files });
+    const d = await this.client.api.channels[this.channel.id].messages[messageID].patch({ data, files });
 
-    if (this.cache.has(message)) {
-      const clone = this.cache.get(message)._clone();
+    if (this.cache.has(messageID)) {
```
vladfrangu, line 134: Nit: instead of has/get, do one get and check that in the `if` check

vladfrangu (changes requested):

SpaceEEC (approved):

vladfrangu (approved):

kyranet (approved):

PR #5889, opened 2021-06-20
feat(Managers): ✨ Add GuildInviteManager

DraftProducts:

The InviteManager was missing, this new Manager comes with some extra features
- Invites generation from Guild based on the first accessible channel
- Invite deletion outside of an Invite instance
- Fetch a single invite with cache filling to avoid unnecessary future requests (force option is available)


typings/index.d.ts
```diff
@@ -2340,6 +2341,16 @@ declare module 'discord.js' {
     public remove(user: UserResolvable, reason?: string): Promise<User>;
   }
 
+  export class GuildInvitesManager extends BaseManager<Snowflake, Invite, InviteResolvable> {
```
iShibi, line 2344:
```suggestion
  export class GuildInviteManager extends BaseManager<Snowflake, Invite, InviteResolvable> {
```

typings/index.d.ts
```diff
@@ -795,6 +795,7 @@ declare module 'discord.js' {
     public approximatePresenceCount: number | null;
     public available: boolean;
     public bans: GuildBanManager;
+    public invites: GuildInvitesManager;
```
iShibi, line 798:
```suggestion
    public invites: GuildInviteManager;
```

src/errors/Messages.js
```diff
@@ -105,6 +107,10 @@ const Messages = {
 
   VANITY_URL: 'This guild does not have the VANITY_URL feature enabled.',
 
+  INVITE_RESOLVE_CODE: `Couldn't resolve the code to fetch the invite.`,
+
+  UNKNOWN_INVITE: `Unknown invite.`,
```
DraftProducts, line 112: I'm not sure about using errors this way, this type of error is usually an error returned by the Discord API but no endpoint was created to fetch a unique invite so I took the freedom to create an error if the invitation code did not exist.

iShibi (changes requested):

typings/index.d.ts
DraftProducts: Thx

src/errors/Messages.js
```diff
@@ -81,6 +81,8 @@ const Messages = {
 
   PRUNE_DAYS_TYPE: 'Days must be a number',
 
+  GUILD_CHANNEL_INVITE: 'Could not find a available channel to create a invite.',
```
vladfrangu, line 84: Is my english rusty? I don't know if this is correct or not, so booping @iCrawl 

```suggestion
  GUILD_CHANNEL_INVITE: 'Could not find an available channel to create a invite.',
```

src/managers/GuildInviteManager.js
```diff
@@ -0,0 +1,179 @@
...
+   * guild.invites.fetch({ code: 'DrzKVU3', cache: false })
+   *   .then(console.log)
+   *   .catch(console.error);
+   */
+  fetch(options) {
+    if (!options) return this._fetchMany();
+    const code = DataResolver.resolveInviteCode(options);
+    if (code) return this._fetchSingle({ code, cache: true });
+    if (options.code) {
+      options.code = DataResolver.resolveInviteCode(options);
```
vladfrangu, line 109: Should avoid mutating, in case someone ever sends a frozen object for the options

src/managers/GuildInviteManager.js
```diff
@@ -0,0 +1,179 @@
...
+    return this._fetchSingle(options);
+  }
+
+  async _fetchSingle({ code, cache, force = false }) {
+    if (!force) {
+      const existing = this.cache.get(code);
+      if (existing) return existing;
+    }
+
+    const invites = await this._fetchMany(cache);
```
vladfrangu, line 124: Couldn't this use the `/invites` endpoint? Is the data received different between these endpoints?

src/managers/GuildInviteManager.js
```diff
@@ -0,0 +1,179 @@
...
+    return invite;
+  }
+
+  async _fetchMany(cache) {
+    const data = await this.client.api.guilds(this.guild.id).invites.get();
+    return data.reduce((col, invite) => col.set(invite.code, this.add(invite, cache)), new Collection());
+  }
+
+  /**
+   * Create a invite to the guild from a random channel.
```
vladfrangu, line 136:
```suggestion
   * Create a invite to the guild from the first available channel.
```

src/managers/GuildInviteManager.js
```diff
@@ -0,0 +1,179 @@
...
+   *   .catch(console.error);
+   */
+  create(options = null, channel = null) {
+    if (!channel && (options instanceof GuildChannel || typeof options === 'string')) {
+      channel = options;
+      options = null;
+    }
+    if (!channel) {
+      channel = this.guild.channels.cache.find(c => c.permissionsFor(this.guild.me).has('CREATE_INSTANT_INVITE'));
+      if (!channel) throw new Error('GUILD_CHANNEL_INVITE');
```
vladfrangu, line 158: Either make this entire function async, or wrap these in return Promise.reject, as right now these are sync errors

vladfrangu (changes requested):
At this point, couldn't we get a GuildChannelInviteManager too? /s 
Jokes aside, here are some things I'd like to bring up

src/errors/Messages.js
NotSugden: > create an invite

apart from that looks fine

src/managers/GuildInviteManager.js
DraftProducts: The goal is to reduce the number of requests made by loaning the Discord API, I think it is more interesting to go through this endpoint, the cache is automatically filled, we will avoid future requests.

typings/index.d.ts
```diff
@@ -2340,6 +2341,16 @@ declare module 'discord.js' {
     public remove(user: UserResolvable, reason?: string): Promise<User>;
   }
 
+  export class GuildInviteManager extends BaseManager<Snowflake, Invite, InviteResolvable> {
+    constructor(guild: Guild, iterable?: Iterable<any>);
+    public guild: Guild;
+    public create(options?: InviteOptions, channel?: ChannelResolvable): Promise<Invite>;
```
iShibi, line 2347:
```suggestion
    public create(options?: CreateInviteOptions, channel?: ChannelResolvable): Promise<Invite>;
```

src/errors/Messages.js
iShibi: I would suggest changing `UNKNOWN_INVITE` to `INVITE_NOT_FOUND` or something. Having exactly same key as an `APIError` isn't a good idea. Also, while you're on it, can you add some more content to the error message that provides a little more info about the error?

src/managers/GuildInviteManager.js
vladfrangu: You could still fetch that endpoint and then cache its result. Would also mean you can remove the APIError-esque UNKNOWN_INVITE code 👀

src/errors/Messages.js
DraftProducts: I think that it will be a nice change 👍

src/managers/GuildInviteManager.js
DraftProducts: The goal is to know if a invite exist in the guild.
A Error rejection on a fetch method wouldn't it be better than just returning null ?

vladfrangu: Ah, right, they might still pass in a valid code, but not for the guild. Good point 👍

DraftProducts (commented):
I think everything is good if you don't have other suggestions for changes 👍

src/errors/Messages.js
```diff
@@ -105,6 +107,10 @@ const Messages = {
 
   VANITY_URL: 'This guild does not have the VANITY_URL feature enabled.',
 
+  INVITE_RESOLVE_CODE: `Couldn't resolve the code to fetch the invite.`,
+
+  INVITE_NOT_FOUND: `Could not find the requested invite.`,
```
NotSugden, line 112:
```suggestion
  INVITE_RESOLVE_CODE: 'Couldn't resolve the code to fetch the invite.',

  INVITE_NOT_FOUND: 'Could not find the requested invite.',
```

src/managers/GuildInviteManager.js
```diff
@@ -0,0 +1,179 @@
...
+    return invite;
+  }
+
+  async _fetchMany(cache) {
+    const data = await this.client.api.guilds(this.guild.id).invites.get();
+    return data.reduce((col, invite) => col.set(invite.code, this.add(invite, cache)), new Collection());
+  }
+
+  /**
+   * Create a invite to the guild from the first available channel.
```
vladfrangu, line 136:
```suggestion
   * Create a invite to the guild from the provided or the first available channel.
```

src/managers/GuildInviteManager.js
```diff
@@ -0,0 +1,179 @@
...
+   * @param {CreateInviteOptions|ChannelResolvable} [options=null] The options for creating the invite
+   * @param {ChannelResolvable} [channel=null] The options for creating the invite
```
vladfrangu, line 138: Can this be turned into an object all the times? So you'd do `invites.create({ channel: '123', maxUses: 21 })`?

vladfrangu (dismissed):

src/managers/GuildInviteManager.js
DraftProducts: Wouldn't it be more interesting to use `channelId` instead of `channel` ?

DraftProducts: Would it be an idea to be able to directly specify the channel id ?

DraftProducts: Would it be an idea to be able to directly specify the channel id?

src/managers/GuildInviteManager.js
```diff
@@ -0,0 +1,181 @@
...
+
+  /**
+   * Deletes a invite.
+   * @param {InviteResolvable} invite The invite to delete
+   * @param {string} [reason] Reason for deleting the invite
+   * @returns {Promise<Invite>}
+   */
+  delete(invite, reason) {
+    const code = DataResolver.resolveInviteCode(invite);
+    return this.client.api.invites[code].delete({ reason }).then(() => this);
```
vladfrangu, line 177: Consistency 
```suggestion
    return this.client.api.invites(code).delete({ reason }).then(() => this);
```

vladfrangu (dismissed):

typings/index.d.ts
```diff
@@ -3015,6 +3025,18 @@ declare module 'discord.js' {
     cache: boolean;
   }
 
+  interface CreateChannelInviteOptions extends CreateInviteOptions {
+    channel: GuildChannel
```
kyranet, line 3029:
```suggestion
    channel: GuildChannel;
```

typings/index.d.ts
```diff
@@ -3015,6 +3025,18 @@ declare module 'discord.js' {
...
+  }
+
+  interface FetchInviteOptions extends BaseFetchOptions {
+    code: string
```
kyranet, line 3033:
```suggestion
    code: string;
```

src/managers/GuildInviteManager.js
```diff
@@ -0,0 +1,184 @@
...
+
+  async _fetchMany(cache) {
+    const data = await this.client.api.guilds(this.guild.id).invites.get();
+    return data.reduce((col, invite) => col.set(invite.code, this.add(invite, cache)), new Collection());
+  }
+
+  /**
+   * @typedef {Object} CreateChannelInviteOptions
+   * @property {GuildChannel} [channel] The optional channel to define where to create the invite.
+   * @extends {CreateInviteOptions}
```
kyranet, line 137:
```suggestion
   * @typedef {CreateInviteOptions} CreateChannelInviteOptions
   * @property {GuildChannel} [channel] The optional channel to define where to create the invite.
```

kyranet (changes requested):

src/managers/GuildInviteManager.js
```diff
@@ -0,0 +1,183 @@
...
+    if (!channel) return Promise.reject(new Error('GUILD_CHANNEL_RESOLVE'));
+
+    return channel.createInvite(options);
+  }
+
+  /**
+   * Deletes a invite.
+   * @param {InviteResolvable} invite The invite to delete
+   * @param {string} [reason] Reason for deleting the invite
+   * @returns {Promise<Invite>}
```
vladfrangu, line 172: Wrong return type

src/managers/GuildInviteManager.js
```diff
@@ -0,0 +1,183 @@
...
+
+  /**
+   * @typedef {CreateInviteOptions} CreateChannelInviteOptions
+   * @property {GuildChannel} [channel] The optional channel to define where to create the invite.
+   */
+
+  /**
+   * Create a invite to the guild from the provided or the first available channel.
+   * @param {CreateChannelInviteOptions} [options={}] The options for creating the invite from a optional channel.
+   * @returns {Promise<Invite?>}
```
vladfrangu, line 142: Whats the `?` on the Invite for?

vladfrangu (dismissed):

src/managers/GuildInviteManager.js
DraftProducts: Your right, I think that a void Promise will be better!

src/managers/GuildInviteManager.js
```diff
@@ -0,0 +1,181 @@
...
+   *   .catch(console.error)
+   * @example
+   * // Fetch a single invite without caching
+   * guild.invites.fetch({ code: 'DrzKVU3', cache: false })
+   *   .then(console.log)
+   *   .catch(console.error);
+   */
+  fetch(options) {
+    if (!options) return this._fetchMany();
+    const code = DataResolver.resolveInviteCode(options);
```
NotSugden, line 105: won't this error if options is not a string

src/managers/GuildInviteManager.js
```diff
@@ -0,0 +1,181 @@
...
+   *   .catch(console.error);
+   */
+  create(options = {}) {
+    let channel;
+
+    if (!options.channel) {
+      channel = this.guild.channels.cache.find(c => c.permissionsFor(this.guild.me).has('CREATE_INSTANT_INVITE'));
+      if (!channel) return Promise.reject(new Error('GUILD_CHANNEL_INVITE'));
+    }
+
```
NotSugden, line 161: maybe put this and the line below in an else block from the above if statement, because the channel is already resolved

src/managers/GuildInviteManager.js
```diff
@@ -0,0 +1,181 @@
...
+  }
+
+  async _fetchMany(cache) {
+    const data = await this.client.api.guilds(this.guild.id).invites.get();
+    return data.reduce((col, invite) => col.set(invite.code, this.add(invite, cache)), new Collection());
+  }
+
+  /**
+   * @typedef {CreateInviteOptions} CreateChannelInviteOptions
+   * @property {GuildChannel} [channel] The optional channel to define where to create the invite.
```
NotSugden, line 136:
```suggestion
   * @property {GuildChannelResolvable} [channel] The optional channel to define where to create the invite.
```

typings/index.d.ts
```diff
@@ -3015,6 +3025,18 @@ declare module 'discord.js' {
     cache: boolean;
   }
 
+  interface CreateChannelInviteOptions extends CreateInviteOptions {
+    channel: GuildChannel;
```
NotSugden, line 3029:
```suggestion
    channel: GuildChannelResolvable;
```

NotSugden (changes requested):

src/managers/GuildInviteManager.js
DraftProducts: You're right!

src/managers/GuildInviteManager.js
DraftProducts: Yep!

src/managers/GuildInviteManager.js
```diff
@@ -0,0 +1,187 @@
...
+   *   .then(console.log)
+   *   .catch(console.error);
+   * @example
+   * // Fetch all invites from a guild without caching
+   * guild.invites.fetch({ cache: false })
+   *   .then(console.log)
+   *   .catch(console.error);
+   * @example
+   * // Fetch a single invite
+   * guild.invites.fetch('DrzKVU3')
```
NotSugden, line 90:
```suggestion
   * guild.invites.fetch('bRCvFy9')
```

src/managers/GuildInviteManager.js
```diff
@@ -0,0 +1,187 @@
...
+   *   .then(console.log)
+   *   .catch(console.error);
+   * @example
+   * // Fetch a single invite without checking cache
+   * guild.invites.fetch({ code: 'DrzKVU3', force: true })
```
NotSugden, line 95:
```suggestion
   * guild.invites.fetch({ code: 'bRCvFy9', force: true })
```

src/managers/GuildInviteManager.js
```diff
@@ -0,0 +1,187 @@
...
+   *   .then(console.log)
+   *   .catch(console.error)
+   * @example
+   * // Fetch a single invite without caching
+   * guild.invites.fetch({ code: 'DrzKVU3', cache: false })
```
NotSugden, line 100:
```suggestion
   * guild.invites.fetch({ code: 'bRCvFy9', cache: false })
```

NotSugden (commented):
discord.js invite

src/managers/GuildInviteManager.js
```diff
@@ -0,0 +1,187 @@
...
+    let channel;
+
+    if (options.channel) {
+      channel = this.guild.channels.resolve(options.channel);
+      if (!channel) return Promise.reject(new Error('GUILD_CHANNEL_RESOLVE'));
+    } else {
+      channel = this.guild.channels.cache.find(c =>
+        c.permissionsFor(this.guild.me).has(Permissions.FLAGS.CREATE_INSTANT_INVITE),
+      );
+      if (!channel) return Promise.reject(new Error('GUILD_CHANNEL_INVITE'));
```
SpaceEEC, line 168: Picking a more or less arbitrary channel to create an invite to is a bit out-of-place with how the library operates.

Arbitrary as in that the order of the channels in this collection is undefined (technically insertion order by how Discord happens to send them to use and channels get created while the bot is running and not establishing a new session).

---

The developer might want to create one on a text channel, but this could also be creating one on a voice channel. (And vice-versa)
They might be expecting the top-most channel to be picked, they might want to pick a "general" or "welcome" channel, if one exist. There probably are a few more common things about this, which I can't think of right now.

---

This behaviour also makes you lose on error-handling, you couldn't distinguish between those, without checking yourself too:
- Was there no channel at all?
- No channel of the correct type?
- No channel with permissions?
- Do we not have permissions to create invites guild-wide?

---

We shouldn't make those assumptions and let the developer handle that themselves by making the `channel` parameter mandatory.

SpaceEEC (changes requested):
Is there any benefit (except backwards-compatability) in keeping `Guild#fetchInvites`?
If not, that method should be removed with this PR.

src/managers/GuildInviteManager.js
DraftProducts: > Is there any benefit (except backward-compatibility) in keeping `Guild#fetchInvites`?
> If not, that method should be removed with this PR.

Like `Guild#fetchBans` when adding the `GuildBanManager` I agree that it should be removed 👍

DraftProducts:
> Is there any benefit (except backward-compatibility) in keeping `Guild#fetchInvites`?
> If not, that method should be removed with this PR.

Like `Guild#fetchBans` when adding the `GuildBanManager` I agree that it should be removed 👍

src/managers/GuildInviteManager.js
DraftProducts: > Arbitrary as in that the order of the channels in this collection is undefined (technically insertion order by how Discord happens to send them to use and channels get created while the bot is running and not establishing a new session).

We could sort them from the `GuildChannel#rawPosition` 

> The developer might want to create one on a text channel, but this could also be creating one on a voice channel. (And vice-versa)

The primary interest of this functionality is to reach a targeted server.
  - When we want to join a voice channel, we target the voice channel in the functionality that links the user to the room.
  - When you simply want to join a targeted server, the goal is to get to the server, regardless of the channel, but not by joining a voice room.

This is also why Discord itself has integrated an interface that takes the first textual room available to invite a user from your direct messages (without even entering an error when this is not possible).

image
image

> This behavior also makes you lose on error-handling, you couldn't distinguish between those, without checking yourself too:
> 
> - Was there no channel at all?
> - No channel of the correct type?
> - No channel with permissions?
> - Do we not have permissions to create invites guild-wide?

I think we should work on these points to make this possible properly.

> Picking a more or less arbitrary channel to create an invite to is a bit out-of-place with how the library operates.

I think this feature aims not to pick a random channel but rather to get an invite to the targeted server. And at present and with the means offered by the Discord API, this is not possible otherwise. As shown above, this is how Discord does it. 

I think that this library should cover an action as simple as getting the invitation from a server that the bot is part of instead of leaving it to its developer given the difficulty that this represents and all the "hidden" and "complex" aspects that this includes.

I wouldn't object to removing it if you think it has no place in the library but I still wanted to explain my opinion.

iCrawl: We already figured that was the case, but we still believe firmly in our stance here.

DraftProducts: Okay 👍

iCrawl: Maybe in the future we can revisit this, but as it currently stands we'd rather move forward without it, instead of blocking the whole PR because of it 👍

DraftProducts: Should we roll back to independent channel params?

iCrawl: cc @SpaceEEC

SpaceEEC: I think that's similar to how we do this in `MessageManager` and `GuildMemberManager`, so I'd be in favor of separate parameters.

src/managers/GuildInviteManager.js
```diff
@@ -0,0 +1,187 @@
...
+   *   .then(console.log)
+   *   .catch(console.error);
+   */
+  create(options = {}) {
+    let channel;
+
+    if (options.channel) {
+      channel = this.guild.channels.resolve(options.channel);
+      if (!channel) return Promise.reject(new Error('GUILD_CHANNEL_RESOLVE'));
+    } else {
```
kyranet, line 164: Everything here needs to go, channel must be a required parameter.

src/managers/GuildInviteManager.js
```diff
@@ -0,0 +1,187 @@
...
+      channel = this.guild.channels.cache.find(c =>
+        c.permissionsFor(this.guild.me).has(Permissions.FLAGS.CREATE_INSTANT_INVITE),
+      );
+      if (!channel) return Promise.reject(new Error('GUILD_CHANNEL_INVITE'));
+    }
+
+    return channel.createInvite(options);
```
kyranet, line 171: GuildChannel should call GuildInviteManager, not the other way around. Please change this to the appropriate code and change `channel.createInvite` to `this.guild.invites.create(options)`.

See other managers for reference.

src/managers/GuildInviteManager.js
```diff
@@ -0,0 +1,187 @@
...
+   * // Create an invite to a selected channel
+   * guild.invites.create({ channel: '599942732013764608' })
+   *   .then(console.log)
+   *   .catch(console.error);
+   */
+  create(options = {}) {
+    let channel;
+
+    if (options.channel) {
+      channel = this.guild.channels.resolve(options.channel);
```
kyranet, line 162: This should use `resolveID` to not depend on cache.

kyranet (dismissed):
Team's consensus is to make channel a required parameter.

And likewise all other managers, manager methods should operate independently of cache where possible.

iCrawl (approved):

DraftProducts:
ℹ️ I moved changes to one commit, changes are listed in the description

src/managers/GuildInviteManager.js
```diff
@@ -0,0 +1,185 @@
...
+   * guild.invites.fetch({ code: 'bRCvFy9', force: true })
+   *   .then(console.log)
+   *   .catch(console.error)
+   * @example
+   * // Fetch a single invite without caching
+   * guild.invites.fetch({ code: 'bRCvFy9', cache: false })
+   *   .then(console.log)
+   *   .catch(console.error);
+   */
+  fetch(options) {
```
DraftProducts, line 103: I was thinking about this fetch method options 🤔 
We don't have any channel parameter, if we add one to options, we can call this method from `GuildChannel#fetchInvites` and cache received invites from `GET /channels/{channel.id}/invites`.

src/managers/GuildInviteManager.js
```diff
@@ -0,0 +1,185 @@
...
+    }
+    if (!options.code) {
+      if ('cache' in options) return this._fetchMany(options.cache);
+      return Promise.reject(new Error('INVITE_RESOLVE_CODE'));
+    }
+    return this._fetchSingle({
+      ...options,
+      code: DataResolver.resolveInviteCode(options.code),
+    });
+  }
```
DraftProducts, line 118: I was thinking about this fetch method options 🤔 
We don't have any channel parameter, if we add one to options, we can call this method from `GuildChannel#fetchInvites` and cache received invites from `GET /channels/{channel.id}/invites`.

src/managers/GuildInviteManager.js
```diff
@@ -0,0 +1,185 @@
...
+   * // Create an invite to a selected channel
+   * guild.invites.create('599942732013764608')
+   *   .then(console.log)
+   *   .catch(console.error);
+   */
+  async create(
+    channel,
+    { temporary = false, maxAge = 86400, maxUses = 0, unique, targetUser, targetApplication, targetType, reason } = {},
+  ) {
+    if (typeof options !== 'object') throw new TypeError('INVALID_TYPE', 'options', 'object', true);
```
kyranet, line 152: This always throws as `options` is always undefined.

If you ask why ESLint doesn't complain, it's for the same reason we can do `typeof window === 'undefined'` and `typeof process === 'undefined'` to determine the environment a script runs in.

src/managers/GuildInviteManager.js
kyranet: `{ channelID: Snowflake } | { code: Snowflake }` sounds good to me, perhaps remove the string overload to prevent ambiguity.

Just make sure to define those two overloads separately since they call different methods and return different data... and that they can't be used together in the types.

src/managers/GuildInviteManager.js
kyranet, line 103: This should use async/await.

kyranet (changes requested):

vladfrangu (approved):
Other than kyra's comments, LGTM

src/errors/Messages.js
```diff
@@ -81,6 +81,8 @@ const Messages = {
 
   PRUNE_DAYS_TYPE: 'Days must be a number',
 
+  GUILD_CHANNEL_INVITE: 'Could not find an available channel to create an invite.',
+
```
SpaceEEC, line 85: Since this is no longer used 🧹 
```suggestion
```

src/managers/GuildInviteManager.js
```diff
@@ -0,0 +1,185 @@
...
+        target_application_id: targetApplication?.id ?? targetApplication?.applicationID ?? targetApplication,
+        target_type: targetType,
+      },
+      reason,
+    });
+    return new Invite(this.client, invite);
+  }
+
+  /**
+   * Deletes a invite.
```
SpaceEEC, line 173:
```suggestion
   * Deletes an invite.
```

src/managers/GuildInviteManager.js
```diff
@@ -0,0 +1,185 @@
...
+   * @type {Collection<Snowflake, Invite>}
+   * @name GuildInviteManager#cache
+   */
+
+  add(data, cache) {
+    return super.add(data, cache, { id: data.code, extras: [this.guild] });
+  }
+
+  /**
+   * Data that resolves to give a Invite object. This can be:
```
SpaceEEC, line 35:
```suggestion
   * Data that resolves to give an Invite object. This can be:
```

src/managers/GuildInviteManager.js
```diff
@@ -0,0 +1,185 @@
...
+   * * An invite code
+   * * An invite URL
+   * @typedef {string} InviteResolvable
+   */
+
+  /**
+   * Resolves a InviteResolvable to a Invite object.
```
SpaceEEC, line 42:
```suggestion
   * Resolves an InviteResolvable to an Invite object.
```

src/managers/GuildInviteManager.js
```diff
@@ -0,0 +1,185 @@
...
+   * @method resolve
+   * @memberof GuildInviteManager
+   * @instance
+   * @param {InviteResolvable} invite The invite resolvable to resolve
+   * @returns {?Invite}
+   */
+
+  /**
+   * Resolves a InviteResolvable to a invite code string.
```
SpaceEEC, line 51:
```suggestion
   * Resolves an InviteResolvable to an invite code string.
```

src/managers/GuildInviteManager.js
```diff
@@ -0,0 +1,185 @@
...
+    return invite;
+  }
+
+  async _fetchMany(cache) {
+    const data = await this.client.api.guilds(this.guild.id).invites.get();
+    return data.reduce((col, invite) => col.set(invite.code, this.add(invite, cache)), new Collection());
+  }
+
+  /**
+   * Create a invite to the guild from the provided channel.
```
SpaceEEC, line 138:
```suggestion
   * Create an invite to the guild from the provided channel.
```

SpaceEEC (changes requested):

src/managers/GuildInviteManager.js
DraftProducts: I can take care of it but I would need help with typing.
I did not understand how to separate them into two "overloads"

src/managers/GuildInviteManager.js
DraftProducts: There isn't any await in this function, why should it be needed?

kyranet: The function is full of promise rejections that happen by using `return Promise.reject(new <>Error())` and that could be a lot cleaner by writing `throw new <>Error()` instead.

Also, in case something ever throws a synchronous error, such as reading a property of undefined, it'd be a promise rejection instead of a regular exception.

src/managers/GuildInviteManager.js
kyranet: Something like the following:
```typescript
public fetch(options: { ..., channelID: Snowflake }): Promise<Collection<Snowflake, Invite>>;
public fetch(options: string | { ..., code: string }): Promise<Invite>;
```

src/managers/GuildInviteManager.js
```diff
@@ -0,0 +1,183 @@
...
+  /**
+   * Options used to fetch all invites from a guild.
+   * @typedef {Object} FetchInvitesOptions
+   * @property {boolean} cache Whether or not to cache the fetched invites
+   */
+
+  /**
+   * Fetches invite(s) from Discord.
+   * @param {InviteResolvable|FetchInviteOptions|FetchInvitesOptions} [options] Options for fetching guild invite(s)
+   * @returns {Promise<Invite>|Promise<Collection<Snowflake, Invite>>}
```
kyranet, line 76:
```suggestion
   * @returns {Promise<Invite|Collection<Snowflake, Invite>>}
```

kyranet (changes requested):
Just one thing I noticed:

src/managers/GuildInviteManager.js
DraftProducts: Is it not just possible to edit `FetchInvitesOptions` with optional channelID?
The returned value is the same 🤔 
```ts
interface FetchInvitesOptions {
  channelID?: Snowflake
  cache: boolean;
}
```

src/managers/GuildInviteManager.js
DraftProducts: ESLint is not likely to be against if there is any await?

kyranet: It shouldn't if there are `throw`s, but who knows.

src/managers/GuildInviteManager.js
kyranet: Sure thing, as long as the second overload takes strictly either a string (invite code), or an object with a required code.

Also, is cache not optional?

DraftProducts: > Also, is cache not optional?
Yep, I change it 👍

src/managers/GuildInviteManager.js
DraftProducts: https://eslint.org/docs/rules/require-await#when-not-to-use-it

DraftProducts (commented):
I merge master to resolve conflicts with breaking changes of `Guild#fetchInvites()`

SpaceEEC (changes requested):

src/structures/GuildChannel.js
```diff
@@ -463,46 +462,17 @@ class GuildChannel extends Channel {
...
    */
-  async fetchInvites() {
-    const inviteItems = await this.client.api.channels(this.id).invites.get();
-    const invites = new Collection();
-    for (const inviteItem of inviteItems) {
-      const invite = new Invite(this.client, inviteItem);
-      invites.set(invite.code, invite);
-    }
-    return invites;
+  fetchInvites() {
```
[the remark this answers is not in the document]
DraftProducts: Yes!

SpaceEEC (approved):

iCrawl:
This needs a rebase.

kyranet (approved):
Just the rebase, otherwise LGTM.

src/managers/GuildInviteManager.js
```diff
@@ -0,0 +1,199 @@
...
+   * @method resolve
+   * @memberof GuildInviteManager
+   * @instance
+   * @param {InviteResolvable} invite The invite resolvable to resolve
+   * @returns {?Invite}
+   */
+
+  /**
+   * Resolves an InviteResolvable to an invite code string.
+   * @method resolveID
```
iCrawl, line 52:
```suggestion
   * @method resolveId
```

src/managers/GuildInviteManager.js
```diff
@@ -0,0 +1,199 @@
...
+  fetch(options) {
+    if (!options) return this._fetchMany();
+    if (typeof options === 'string') {
+      const code = DataResolver.resolveInviteCode(options);
+      if (!code) return Promise.reject(new Error('INVITE_RESOLVE_CODE'));
+      return this._fetchSingle({ code, cache: true });
+    }
+    if (!options.code) {
+      if (options.channelID) {
+        const id = this.guild.channels.resolveID(options.channelID);
```
iCrawl, line 117:
```suggestion
        const id = this.guild.channels.resolveId(options.channelId);
```

src/managers/GuildInviteManager.js
```diff
@@ -0,0 +1,199 @@
...
+   */
+  fetch(options) {
+    if (!options) return this._fetchMany();
+    if (typeof options === 'string') {
+      const code = DataResolver.resolveInviteCode(options);
+      if (!code) return Promise.reject(new Error('INVITE_RESOLVE_CODE'));
+      return this._fetchSingle({ code, cache: true });
+    }
+    if (!options.code) {
+      if (options.channelID) {
```
iCrawl, line 116:
```suggestion
      if (options.channelId) {
```

src/managers/GuildInviteManager.js
```diff
@@ -0,0 +1,199 @@
...
+   * // Create an invite to a selected channel
+   * guild.invites.create('599942732013764608')
+   *   .then(console.log)
+   *   .catch(console.error);
+   */
+  async create(
+    channel,
+    { temporary = false, maxAge = 86400, maxUses = 0, unique, targetUser, targetApplication, targetType, reason } = {},
+  ) {
+    const id = this.guild.channels.resolveID(channel);
```
iCrawl, line 168:
```suggestion
    const id = this.guild.channels.resolveId(channel);
```

src/managers/GuildInviteManager.js
```diff
@@ -0,0 +1,199 @@
...
+    if (!id) throw new Error('GUILD_CHANNEL_RESOLVE');
+
+    const invite = await this.client.api.channels(id).invites.post({
+      data: {
+        temporary,
+        max_age: maxAge,
+        max_uses: maxUses,
+        unique,
+        target_user_id: this.client.users.resolveID(targetUser),
```
iCrawl, line 177:
```suggestion
        target_user_id: this.client.users.resolveId(targetUser),
```

src/managers/GuildInviteManager.js
```diff
@@ -0,0 +1,199 @@
...
+        target_application_id: targetApplication?.id ?? targetApplication?.applicationID ?? targetApplication,
```
iCrawl, line 178:
```suggestion
        target_application_id: targetApplication?.id ?? targetApplication?.applicationId ?? targetApplication,
```

iCrawl (changes requested):
`ID` -> `Id` changes required.

DraftProducts:
Arf I was already working on it 👀

iCrawl (approved):

advaith1:
was this actually tested? it seems to not actually add the Guild#invites property, breaking all the places which now try to access it 😬

PR #5895, opened 2021-06-21
types(MessageOptions): remove split?: false overload

monbrey:
Having overloads on MessageOptions for both `{ split?: false }` and `{ split: true | SplitOptions }` was causing TS to error - these  types were seemingly incompatible with either each other, or the base `split?: boolean | SplitOptions` typing.
As a result `MessageOptions` could not be passed to `send` methods.

I don't believe we need the `false` state, because it's only the presence of `true | SplitOptions` that would actually change the return type. I could be wrong though, but this does remove the error.

### DIFF
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -1181,9 +1181,7 @@ declare module 'discord.js' {
   export class InteractionWebhook extends PartialWebhookMixin() {
     constructor(client: Client, id: Snowflake, token: string);
     public token: string;
-    public send(
-      options: string | APIMessage | (InteractionReplyOptions & { split?: false }),
-    ): Promise<Message | RawMessage>;
+    public send(options: string | APIMessage | InteractionReplyOptions): Promise<Message | RawMessage>;
     public send(
       options: APIMessage | (InteractionReplyOptions & { split: true | SplitOptions }),
     ): Promise<(Message | RawMessage)[]>;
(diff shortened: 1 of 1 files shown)


monbrey:
...but now one of the tests fails, so I'm going to re-assess

monbrey:
Yeah this is beyond my understanding of types after all, closing

suneettipirneni:
I was wondering if this is still being considered, I've often had to make type alias to avoid having to type `MessageOptions & { split?: false }`. It would be awesome if this typing was consolidated.

iCrawl (approved):

kyranet (approved):

vladfrangu (approved):

SpaceEEC:
I think the more correct way to go about this is to keep the original overloads and make the new ones return `Message | Message[]` (or `RawMessage | RawMessage[]`, or both), since those overloads don't know whether `split` is present or not.

The `index.ts` (the one without .d`), should be updated to include "tests" for those, so we hopefully don't break it straight away again.

~~Simpler send overloads anyone?~~

iCrawl:
This needs a rebase.

monbrey:
> I think the more correct way to go about this is to keep the original overloads and make the new ones return `Message | Message[]` (or `RawMessage | RawMessage[]`, or both), since those overloads don't know whether `split` is present or not.

We do though, don't we? The first overload confirms that if `split` is either `true` or `SplitOptions` then it returns a Message[]. The only other values TypeScript could allow is either explicit false, or leaving it out entirely.

ImRodry:
This will no longer be needed if #5918 is merged

monbrey:
> This will no longer be needed if #5918 is merged

We know

PR #5897, opened 2021-06-21
feat(Managers): new ApplicationCommandPermissionsManager

ckohen:

Replaces and closes #5695 as the PR seems to be abandoned.  This takes a good chunk of the code from that PR, but moves existing code and new code into a new manager while addressing requested changes.

Benefits of this approach:
Each command has its own way to access all of its permissions properties alongside the guild and global versions
Every function takes an object, making the somewhat confusing required vs optional parameters a little easier to understand.
Typescript users *hopefully* know exactly what they need to use the manager due to some hefty overloads and type passing.
(I did a lot of things that I don't know whether will work or not, but they should 🤞...I hope )

~~⚠️ This is untested as of now, testing in progress.~~

### DIFF
--- a/src/managers/ApplicationCommandPermissionsManager.js
+++ b/src/managers/ApplicationCommandPermissionsManager.js
@@ -2,7 +2,7 @@
 
 const { Error, TypeError } = require('../errors');
 const Collection = require('../util/Collection');
-const { ApplicationCommandPermissionTypes } = require('../util/Constants');
+const { ApplicationCommandPermissionTypes, APIErrors } = require('../util/Constants');
 
 /**
  * Manages API methods for permissions of Application Commands.
@@ -122,7 +122,7 @@ class ApplicationCommandPermissionsManager {
    * <warn>One of `command` AND `permissions`, OR `fullPermissions` is required.
    * `fullPermissions` is not a valid option when passing to a manager where `commandID` is non-null</warn>
    * @typedef {BaseApplicationCommandPermissionsOptions} SetApplicationCommandPermissionsOptions
-   * @param {ApplicationCommandPermissionsData[]} [permissions] The new permissions for the command
+   * @param {ApplicationCommandPermissionData[]} [permissions] The new permissions for the command
    * @param {GuildApplicationCommandPermissionData[]} [fullPermissions] The new permissions for all commands
    * in a guild <warn>When this parameter is set, permissions and command are ignored</warn>
    */
@@ -162,7 +162,7 @@ class ApplicationCommandPermissionsManager {
 
     if (commandID) {
       if (!Array.isArray(permissions)) {
-        throw new TypeError('INVALID_TYPE', 'permissions', 'Array of ApplicationCommandPermissionsData', true);
+        throw new TypeError('INVALID_TYPE', 'permissions', 'Array of ApplicationCommandPermissionData', true);
       }
       const data = await this.permissionsPath(guildID, commandID).put({
         data: { permissions: permissions.map(perm => this.constructor.transformPermissions(perm)) },
@@ -199,7 +199,7 @@ class ApplicationCommandPermissionsManager {
    * Options used to add permissions to a command
    * <warn>The `command` parameter is not optional when the managers `commandID` is `null`</warn>
    * @typedef {BaseApplicationCommandPermissionsOptions} AddApplicationCommandPermissionsOptions
-   * @param {ApplicationCommandPermissionsData[]} permissions The permissions to add to the command
+   * @param {ApplicationCommandPermissionData[]} permissions The permissions to add to the command
    */
 
   /**
@@ -222,14 +222,14 @@ class ApplicationCommandPermissionsManager {
     const { guildID, commandID } = this._validateOptions(guild, command);
     if (!commandID) throw new TypeError('INVALID_TYPE', 'command', 'ApplicationCommandResolvable');
     if (!Array.isArray(permissions)) {
-      throw new TypeError('INVALID_TYPE', 'permissions', 'Array of ApplicationCommandPermissionsData', true);
+      throw new TypeError('INVALID_TYPE', 'permissions', 'Array of ApplicationCommandPermissionData', true);
     }
 
     let existing = [];
     try {
       existing = await this.fetch({ guild: guildID, command: commandID });
     } catch (error) {
-      if (error.code !== 404) throw error;
+      if (error.code !== APIErrors.UNKNOWN_APPLICATION_COMMAND_PERMISSIONS) throw error;
     }
 
     const newPermissions = permissions.slice();
@@ -277,7 +277,7 @@ class ApplicationCommandPermissionsManager {
     try {
       existing = await this.fetch({ guild: guildID, command: commandID });
     } catch (error) {
-      if (error.code !== 404) throw error;
+      if (error.code !== APIErrors.UNKNOWN_APPLICATION_COMMAND_PERMISSIONS) throw error;
     }
 
     if (!users && !roles) throw new TypeError('INVALID_TYPE', 'users OR roles', 'Array or Resolvable', true);
@@ -351,9 +351,10 @@ class ApplicationCommandPermissionsManager {
     try {
       existing = await this.fetch({ guild: guildID, command: commandID });
     } catch (error) {
-      if (error.code !== 404) throw error;
+      if (error.code !== APIErrors.UNKNOWN_APPLICATION_COMMAND_PERMISSIONS) throw error;
     }
 
+    if (!permissionID) throw new TypeError('INVALID_TYPE', 'permissionsID', 'UserResolvable or RoleResolvable');
     let resolvedID = permissionID;
     if (typeof permissionID !== 'string') {
       resolvedID = this.client.users.resolveID(permissionID);
@@ -370,7 +371,7 @@ class ApplicationCommandPermissionsManager {
 
   _validateOptions(guild, command) {
     const guildID = this.guildID ?? this.client.guilds.resolveID(guild);
-    if (!guildID) return Promise.reject(new Error('GLOBAL_COMMAND_PERMISSIONS'));
+    if (!guildID) throw new Error('GLOBAL_COMMAND_PERMISSIONS');
     let commandID = this.commandID;
     if (command && !commandID) {
       commandID = this.manager.resolveID?.(command);
@@ -381,7 +382,7 @@ class ApplicationCommandPermissionsManager {
         commandID = this.client.application?.commands.resolveID(command);
       }
       if (!commandID) {
-        return Promise.reject(new TypeError('INVALID_TYPE', 'command', 'ApplicationCommandResolvable', true));
+        throw new TypeError('INVALID_TYPE', 'command', 'ApplicationCommandResolvable', true);
       }
     }
     return { guildID, commandID };
--- a/src/managers/ApplicationCommandPermissionsManager.js
+++ b/src/managers/ApplicationCommandPermissionsManager.js
@@ -264,7 +264,7 @@ class ApplicationCommandPermissionsManager {
    * @example
    * // Remove multiple roles from this command
    * guild.commands.permissions.remove({
-   *   command: '123456789012345678', roles: [ '876543210123456789', '765432101234567890' ]
+   *   command: '123456789012345678', roles: ['876543210123456789', '765432101234567890']
    * })
    *    .then(console.log)
    *    .catch(console.error);
--- a/src/managers/ApplicationCommandPermissionsManager.js
+++ b/src/managers/ApplicationCommandPermissionsManager.js
@@ -1,0 +1,414 @@
+'use strict';
+
+const { Error, TypeError } = require('../errors');
+const Collection = require('../util/Collection');
+const { ApplicationCommandPermissionTypes } = require('../util/Constants');
+
+/**
+ * Manages API methods for permissions of Application Commands.
+ */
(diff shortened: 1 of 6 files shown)


src/managers/ApplicationCommandPermissionsManager.js
```diff
@@ -0,0 +1,409 @@
...
+      if (!resolvedID) {
+        throw new TypeError('INVALID_TYPE', 'permissionID', 'UserResolvable or RoleResolvable');
+      }
+    }
+    return existing.some(perm => perm.id === resolvedID);
+  }
+
+  _validateOptions(guild, command) {
+    const guildID = this.guildID ?? this.client.guilds.resolveID(guild);
+    if (!guildID) return Promise.reject(new Error('GLOBAL_COMMAND_PERMISSIONS'));
```
SpaceEEC, line 373: You do not await the result of this method, so the rejected promise will just be discarded due to destructuring assignment being used (and probably cause an unhandled promise rejection).

Since this method is only called from within async functions, throwing here should work fine:
```suggestion
    if (!guildID) throw new Error('GLOBAL_COMMAND_PERMISSIONS');
```

src/managers/ApplicationCommandPermissionsManager.js
```diff
@@ -0,0 +1,409 @@
...
+    if (command && !commandID) {
+      commandID = this.manager.resolveID?.(command);
+      if (!commandID && this.guild) {
+        commandID = this.guild.commands.resolveID(command);
+      }
+      if (!commandID) {
+        commandID = this.client.application?.commands.resolveID(command);
+      }
+      if (!commandID) {
+        return Promise.reject(new TypeError('INVALID_TYPE', 'command', 'ApplicationCommandResolvable', true));
```
SpaceEEC, line 384:
```suggestion
        throw new TypeError('INVALID_TYPE', 'command', 'ApplicationCommandResolvable', true);
```

SpaceEEC (changes requested):

src/managers/ApplicationCommandPermissionsManager.js
ckohen: Yep, discovered this during testing...which is now done

src/managers/ApplicationCommandPermissionsManager.js
```diff
@@ -0,0 +1,410 @@
...
+   * @returns {Promise<ApplicationCommandPermissions[]>}
+   * @example
+   * // Remove a user permission from this command
+   * guild.commands.permissions.remove({ command: '123456789012345678', users: '876543210123456789' })
+   *   .then(console.log)
+   *   .catch(console.error);
+   * @example
+   * // Remove multiple roles from this command
+   * guild.commands.permissions.remove({
+   *   command: '123456789012345678', roles: [ '876543210123456789', '765432101234567890' ]
```
1chiSensei, line 267:
```suggestion
   *   command: '123456789012345678', roles: ['876543210123456789', '765432101234567890']
```

1chiSensei (changes requested):

iCrawl (approved):

src/managers/ApplicationCommandPermissionsManager.js
```diff
@@ -0,0 +1,410 @@
...
+ * Manages API methods for permissions of Application Commands.
+ */
+class ApplicationCommandPermissionsManager {
+  constructor(manager) {
+    /**
+     * The manager or command that this manager belongs to
+     * @type {ApplicationCommandManager|ApplicationCommand}
+     */
+    this.manager = manager;
+    /**
```
kyranet, line 17: Missing spaces here and in the fields below:
```suggestion
    this.manager = manager;

    /**
```

src/managers/ApplicationCommandPermissionsManager.js
```diff
@@ -0,0 +1,410 @@
...
+
+  /**
+   * The APIRouter path to the commands
+   * @param {Snowflake} guildID ID of the guild to use in the path,
+   * @param {Snowflake} [commandID] ID of the application command
+   * @returns {Object}
+   * @private
+   */
+  permissionsPath(guildID, commandID) {
+    let path = this.client.api.applications(this.client.application.id).guilds(guildID);
```
kyranet, line 49: How ESLint didn't complain, beats me:
```suggestion
    const path = this.client.api.applications(this.client.application.id).guilds(guildID);
```

src/managers/ApplicationCommandPermissionsManager.js
```diff
@@ -0,0 +1,410 @@
...
+        ),
+      new Collection(),
+    );
+  }
+
+  /**
+   * Data used for overwriting the permissions for all application commands in a guild.
+   * @typedef {Object} GuildApplicationCommandPermissionData
+   * @prop {Snowflake} id The ID of the command
+   * @prop {ApplicationCommandPermissionData[]} permissions The permissions for this command
```
kyranet, line 117:
```suggestion
   * @property {Snowflake} id The ID of the command
   * @property {ApplicationCommandPermissionData[]} permissions The permissions for this command
```

src/structures/ApplicationCommand.js
```diff
@@ -126,62 +140,7 @@ class ApplicationCommand extends Base {
...
-   *     type: 'USER',
-   *     permission: false,
-   *   },
-   * ])
-   *   .then(console.log)
-   *   .catch(console.error);
-   */
-  setPermissions(permissions, guildID) {
-    return this.manager.setPermissions(this, permissions, guildID);
+    return this.manager.delete(this, this.guildID ?? undefined);
```
kyranet, line 143: Why is this defaulting to undefined?

typings/index.d.ts
```diff
@@ -2184,47 +2186,85 @@ declare module 'discord.js' {
...
-      guildID: Snowflake,
+    public set(
+      commands: ApplicationCommandData[],
+      guildID?: Snowflake,
+    ): Promise<Collection<Snowflake, ApplicationCommandType>>;
+    private static transformCommand(command: ApplicationCommandData): unknown;
+  }
+
+  export class ApplicationCommandPermissionsManager<BaseOptions, FetchSingleOptions, GuildType, CommandIDType> {
+    constructor(public manager: ApplicationCommandManager | GuildApplicationCommandManager | ApplicationCommand);
```
kyranet, line 2234: Please declare the field outside for consistency with the other definitions.

kyranet (changes requested):

src/managers/ApplicationCommandPermissionsManager.js
```diff
@@ -0,0 +1,410 @@
...
+  /**
+   * The APIRouter path to the commands
+   * @param {Snowflake} guildID ID of the guild to use in the path,
+   * @param {Snowflake} [commandID] ID of the application command
+   * @returns {Object}
+   * @private
+   */
+  permissionsPath(guildID, commandID) {
+    let path = this.client.api.applications(this.client.application.id).guilds(guildID);
+    return (commandID ? path.commands(commandID) : path.commands).permissions;
```
vladfrangu, line 50: This ternary isn't needed 
image
```suggestion
    return path.commands(commandID).permissions;
```

src/managers/ApplicationCommandPermissionsManager.js
```diff
@@ -0,0 +1,410 @@
...
+    if (!commandID) throw new TypeError('INVALID_TYPE', 'command', 'ApplicationCommandResolvable');
+
+    let existing = [];
+    try {
+      existing = await this.fetch({ guild: guildID, command: commandID });
+    } catch (error) {
+      if (error.code !== APIErrors.UNKNOWN_APPLICATION_COMMAND_PERMISSIONS) throw error;
+    }
+
+    if (!permissionID) throw new TypeError('INVALID_TYPE', 'permissionsID', 'UserResolvable or RoleResolvable');
```
vladfrangu, line 357: Move before API call

src/managers/ApplicationCommandPermissionsManager.js
```diff
@@ -0,0 +1,410 @@
...
+    let resolvedID = permissionID;
+    if (typeof permissionID !== 'string') {
+      resolvedID = this.client.users.resolveID(permissionID);
+      if (!resolvedID) {
+        if (!this.guild) throw new Error('GUILD_UNCACHED_ROLE_RESOLVE');
```
vladfrangu, line 362: Isn't this error needed across the board? Since permissions are guild-only

vladfrangu (changes requested):

iCrawl:
This needs a rebase.

src/managers/ApplicationCommandPermissionsManager.js
ckohen: I think this eslint rule is actually missing from config 🙂

src/structures/ApplicationCommand.js
ckohen: Wow, that diff is messed up 😆 
Anyways, it was because I thought I had made a specific check for undefined, but I didn't, passing null is fine.

src/managers/ApplicationCommandPermissionsManager.js
ckohen: No, *theoretically* it should be impossible to hit this error as you shouldn't be able to pass a Role object from an uncached guild unless you manually construct it I guess.

The missing guild / guild id is  handled in `_validateOptions`, the first thing it does, and that's the first thing called in every function.

kyranet (approved):

src/util/Constants.js
```diff
@@ -602,8 +602,11 @@ exports.VerificationLevels = createEnum(['NONE', 'LOW', 'MEDIUM', 'HIGH', 'VERY_
  * * UNKNOWN_REDISTRIBUTABLE
  * * UNKNOWN_GIFT_CODE
  * * UNKNOWN_GUILD_TEMPLATE
+<<<<<<< HEAD
  * * UNKNOWN_INTERACTION
  * * UNKNOWN_APPLICATION_COMMAND
+=======
+>>>>>>> 70b04688 (fix: fix bugs found in runtime)
```
vladfrangu, line 609: Hello there git rebase

vladfrangu (changes requested):

1chiSensei (approved):

src/util/Constants.js
ckohen: I am very confused how I didn't get an error when rebasing...git you okay?

iCrawl:
This needs a rebase.

src/util/Constants.js
```diff
@@ -637,8 +637,11 @@ exports.VerificationLevels = createEnum(['NONE', 'LOW', 'MEDIUM', 'HIGH', 'VERY_
...
```
vladfrangu, line 644: Git rebase is still here, lurking away

vladfrangu (changes requested):

src/util/Constants.js
ckohen: I...whh...how?  I literally remember removing those exact likens in this rebase 🤔

kyranet: Sounds like the rebase needs a rebase 😃

ckohen: well this PR no longer changes Constants at all 😆 , so hopefully that's a good sign that rebase is gone

typings/index.d.ts
```diff
@@ -2310,47 +2312,86 @@ declare module 'discord.js' {
...
-    ApplicationCommandResolvable
-  > {
+  export class ApplicationCommandManager<
+    ApplicationCommandType = ApplicationCommand<{ guild: GuildResolvable }>,
+  > extends BaseManager<Snowflake, ApplicationCommandType, ApplicationCommandResolvable> {
     constructor(client: Client, iterable?: Iterable<any>);
+    public permissions: ApplicationCommandPermissionsManager<
+      { guild: GuildResolvable; command?: ApplicationCommandResolvable },
+      { guild: GuildResolvable; command: ApplicationCommandResolvable },
+      null,
```
SpaceEEC, line 2322: This is currently breaking assignability from the deriving class (`GuildApplicationCommandManager`).

While this change is technically less specific, compiling djs without errors is something we probably want 😅 
```suggestion
      Guild | null,
```

SpaceEEC (changes requested):

typings/index.d.ts
ckohen: Fixed it in a slightly different way since there was actually an error in the first set of lines too

typings/index.d.ts
```diff
@@ -2292,47 +2294,88 @@ declare module 'discord.js' {
...
+        permissions: ApplicationCommandPermissionData[];
+      },
+    ): Promise<ApplicationCommandPermissions[]>;
+    public set(
+      options: BaseOptions & {
+        fullPermissions: GuildApplicationCommandPermissionData[];
+      },
     ): Promise<Collection<Snowflake, ApplicationCommandPermissions[]>>;
-    private static transformCommand(command: ApplicationCommandData): unknown;
+    private permissionsPath(guildID: Snowflake, commandID: Snowflake): unknown;
```
vladfrangu, line 2378:
```suggestion
    private permissionsPath(guildID: Snowflake, commandID?: Snowflake): unknown;
```

vladfrangu (changes requested):

ckohen:
Sorry about the commit msg on the last commit, did it from mobile. If we hit another rebase I'll fix it

vladfrangu (approved):

SpaceEEC (approved):

PR #5900, opened 2021-06-21
Type WebhookMessageOptions.avatarURL as string | null | undefined

Bluenix2:
Without this change you would not be able to directly pass `client.application?.iconURL()` into `WebhookMessageOptions.avatarURL`. Observe the following error:
```
Type 'string | null | undefined' is not assignable to type 'string | undefined'.
  Type 'null' is not assignable to type 'string | undefined'.
```

- This is only changes typings, it does not cause any breaking changes

### DIFF
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -3971,7 +3971,7 @@ declare module 'discord.js' {
 
   interface WebhookMessageOptions extends Omit<MessageOptions, 'reply'> {
     username?: string;
-    avatarURL?: string;
+    avatarURL?: string | null;
   }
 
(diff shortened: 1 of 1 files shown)


DTrombett:
Why not just use `client.application?.iconURL() ?? undefined`

Bluenix2:
That's what I currently do, but it's more a work-around.

iShibi:
The `WebhookMessageOptions#avatarURL` only takes an optional `string` param. Adding `null` to its typings doesn't reflect what the method actually expects. You can fix this issue by putting a typecheck in place for `client.application?.iconURL()` before passing it.

PR #5903, opened 2021-06-22
refactor(Collector): make filter an option

vaporox:

This makes the filter for collectors optional because you shouldn't need to pass one. Currently, you need to do `createCollector(() => true, { ... })` to achieve the same result.

### DIFF
--- a/src/structures/interfaces/Collector.js
+++ b/src/structures/interfaces/Collector.js
@@ -78,7 +78,7 @@ class Collector extends EventEmitter {
     this._idletimeout = null;
 
     if (typeof this.filter !== 'function') {
-      throw new TypeError('INVALID_TYPE', 'filter', 'function');
+      throw new TypeError('INVALID_TYPE', 'options.filter', 'function');
     }
 
     this.handleCollect = this.handleCollect.bind(this);
--- a/src/structures/interfaces/Collector.js
+++ b/src/structures/interfaces/Collector.js
@@ -91,6 +91,7 @@ class Collector extends EventEmitter {
   /**
    * Call this to handle an event as a collectable element. Accepts any event data as parameters.
    * @param {...*} args The arguments emitted by the listener
+   * @returns {Promise<void>}
    * @emits Collector#collect
    */
   async handleCollect(...args) {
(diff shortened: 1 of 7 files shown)


kyranet (approved):

src/structures/interfaces/Collector.js
```diff
@@ -37,12 +37,6 @@ class Collector extends EventEmitter {
      */
     Object.defineProperty(this, 'client', { value: client });
 
-    /**
-     * The filter applied to this collector
-     * @type {CollectorFilter}
-     */
-    this.filter = filter;
-
```
monbrey, line 45: What's the reasoning behind removing this property? Although it's a very small edge case, wouldn't this prevent someone from changing the filter dynamically?

Could it be kept as `this.filter = options.filter ?? () => true`, removing the need to check for its existence later?

vladfrangu: Technically, you can do `collector.options.filter = () => true;`, however this will throw if someone freezes their options object and then proceeds to want to alter them. But I'm in favor of bringing this back

SpaceEEC (changes requested):
I'd like to keep the validation for the filter (if provided), as that's a not so straight-forward error with a not so helpful stacktrace otherwise. 😅

src/structures/interfaces/Collector.js
```diff
@@ -37,6 +38,13 @@ class Collector extends EventEmitter {
      */
     Object.defineProperty(this, 'client', { value: client });
 
+    /**
+     * The filter applied to this collector
+     * @type {CollectorFilter}
+     * @returns {boolean|Promise<boolean>}
```
vladfrangu, line 44: This...is not handled everywhere if it's a promise..

vladfrangu (changes requested):

src/structures/interfaces/Collector.js
vaporox: It's always been like that, I just copied it over from the `CollectorFilter` typedef to make eslint happy. If it's not too far out of scope I can add the missing handling in this PR

iCrawl: It's fine to put it into this PR.

iCrawl (approved):

src/structures/interfaces/Collector.js
```diff
@@ -75,7 +77,7 @@ class Collector extends EventEmitter {
      */
     this._idletimeout = null;
 
-    if (typeof filter !== 'function') {
+    if (typeof this.filter !== 'function') {
       throw new TypeError('INVALID_TYPE', 'filter', 'function');
```
SpaceEEC, line 81:
```suggestion
      throw new TypeError('INVALID_TYPE', 'options.filter', 'function');
```

src/structures/interfaces/TextBasedChannel.js
```diff
@@ -351,9 +347,9 @@ class TextBasedChannel {
    *   .then(interaction => console.log(`${interaction.customID} was clicked!`))
    *   .catch(console.error);
    */
-  awaitMessageComponentInteraction(filter, { time } = {}) {
+  awaitMessageComponentInteraction(options = {}) {
```
SpaceEEC, line 350: The example for this method also needs to be updated. :eyes:

SpaceEEC (changes requested):
Otherwise LGTM

PR #5905, opened 2021-06-22
feat(MessageMentions): add repliedUser

vaporox:

This PR adds a `repliedUser` property to the message mentions so that users can more quickly see which user was replied to.

### DIFF
--- a/src/structures/Message.js
+++ b/src/structures/Message.js
@@ -180,7 +180,14 @@ class Message extends Base {
      * All valid mentions that the message contains
      * @type {MessageMentions}
      */
-    this.mentions = new Mentions(this, data.mentions, data.mention_roles, data.mention_everyone, data.mention_channels);
+    this.mentions = new Mentions(
+      this,
+      data.mentions,
+      data.mention_roles,
+      data.mention_everyone,
+      data.mention_channels,
+      data.referenced_message?.author,
+    );
 
     /**
      * ID of the webhook that sent the message, if applicable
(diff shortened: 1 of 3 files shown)


kyranet (approved):

src/structures/MessageMentions.js
```diff
@@ -125,6 +125,12 @@ class MessageMentions {
     } else {
       this.crosspostedChannels = new Collection();
     }
+
+    /**
+     * The author of the message that this message is a reply to
+     * @type {?User}
+     */
+    this.repliedUser = repliedUser ? this.client.users.add(repliedUser) : this.repliedUser ?? null;
```
NotSugden, line 133:
```suggestion
    this.repliedUser = repliedUser ? this.client.users.add(repliedUser) : null;
```
thats never assigned before so its just a useless check

iCrawl (approved):

src/structures/MessageMentions.js
monbrey: It protects it against message edits which might not include that property though.

NotSugden: this assignment is in the constructor for `MessageMentions`, so `this.repliedUser ?? null` would always just return null here no? so why not just directly set it to null

src/structures/Message.js
```diff
@@ -314,6 +321,7 @@ class Message extends Base {
       'mention_roles' in data ? data.mention_roles : this.mentions.roles,
       'mention_everyone' in data ? data.mention_everyone : this.mentions.everyone,
       'mention_channels' in data ? data.mention_channels : this.mentions.crosspostedChannels,
+      data.referenced_message?.author,
```
vladfrangu, line 324: this should do the same as above

vladfrangu (changes requested):
This as well as the point NotSugden said

vladfrangu (approved):

src/structures/MessageMentions.js
```diff
@@ -125,6 +125,12 @@ class MessageMentions {
     } else {
       this.crosspostedChannels = new Collection();
     }
+
+    /**
+     * The author of the message that this message is a reply to
```
ImRodry, line 130: This sounds a bit odd imo
```suggestion
     * The author of the message that this message was replied to
```

SpaceEEC (approved):

PR #5906, opened 2021-06-22
feat(InteractionCreate): move to an Action handler

monbrey:
Ignoring the much larger issues about Websocket and Action handlers still being separate, this moves the INTERACTION_CREATE logic over to an Action handler.

This allows `Action#getChannel` to be used for caching a partial DMChannel, preventing `Interaction#channel` from returning null with the CHANNEL partial enabled

### DIFF
--- a/src/client/actions/InteractionCreate.js
+++ b/src/client/actions/InteractionCreate.js
@@ -7,6 +7,8 @@ const Structures = require('../../util/Structures');
 class InteractionCreateAction extends Action {
   handle(data) {
     const client = this.client;
+
+    // Resolve and cache partial channels for Interaction#channel getter
     this.getChannel(data);
 
     let InteractionType;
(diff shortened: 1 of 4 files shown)


iCrawl (approved):

kyranet (approved):

src/client/actions/InteractionCreate.js
```diff
@@ -0,0 +1,44 @@
+'use strict';
+
+const Action = require('./Action');
+const { Events, InteractionTypes, MessageComponentTypes } = require('../../util/Constants');
+const Structures = require('../../util/Structures');
+
+class InteractionCreateAction extends Action {
+  handle(data) {
+    const client = this.client;
+    this.getChannel(data);
```
vladfrangu, line 10: Does this...cache the channel as partial? Can we at least add a comment so whoever touches this file knows that this line isn't here just because? 🙏

vladfrangu (dismissed):
Meant to request changes, ^

SpaceEEC (approved):

ImRodry:
The typings for Interaction#channel need to be updated to add null (seems like this was never added for some odd reason)

vladfrangu (approved):

PR #5910, opened 2021-06-23
revert(BitField): ⏪ Bring back-compatibility after BitField serialization

DraftProducts:

In #4879, some modifications were made to permissions and bitfield methods.
Especially on the type of bitfields that moves from **number** to **BigInt**.

This move has some impact on serialization and deserialization.
Take the example of a simple `JSON.stringify` and `JSON.parse`.
The method `toJSON()` (provided by BitField class) will be called and the type (BigInt in the case of Permissions) will be transformed into a string:
> `8n` => `"8"`

In the past, this type changing was not needed because a number remains a number.
In the actual Bitfield class if you try to pass permissions that went through serialization and deserialization you get the beautiful error `BITFIELD_INVALID`.
So a simple role or channel creation bring you to a crash of your action.

This PR is here to provide compatibility between Bigint stringified and BitFields.

### DIFF
--- a/src/util/BitField.js
+++ b/src/util/BitField.js
@@ -147,7 +147,7 @@ class BitField {
     if (Array.isArray(bit)) return bit.map(p => this.resolve(p)).reduce((prev, p) => prev | p, defaultBit);
     if (typeof bit === 'string') {
       if (typeof this.FLAGS[bit] !== 'undefined') return this.FLAGS[bit];
-      if (!isNaN(bit)) return BigInt(bit);
+      if (!isNaN(bit)) return typeof defaultBit === 'bigint' ? BigInt(bit) : Number(bit);
     }
     throw new RangeError('BITFIELD_INVALID', bit);
   }
--- a/src/util/BitField.js
+++ b/src/util/BitField.js
@@ -129,9 +129,10 @@ class BitField {
   /**
    * Data that can be resolved to give a bitfield. This can be:
    * * A bit number (this can be a number literal or a value taken from {@link BitField.FLAGS})
+   * * A string bit number
    * * An instance of BitField
    * * An Array of BitFieldResolvable
-   * @typedef {number|bigint|BitField|BitFieldResolvable[]} BitFieldResolvable
+   * @typedef {number|string|bigint|BitField|BitFieldResolvable[]} BitFieldResolvable
    */
 
   /**
@@ -144,7 +145,10 @@ class BitField {
     if (typeof defaultBit === typeof bit && bit >= defaultBit) return bit;
     if (bit instanceof BitField) return bit.bitfield;
     if (Array.isArray(bit)) return bit.map(p => this.resolve(p)).reduce((prev, p) => prev | p, defaultBit);
-    if (typeof bit === 'string' && typeof this.FLAGS[bit] !== 'undefined') return this.FLAGS[bit];
+    if (typeof bit === 'string') {
+      if (typeof this.FLAGS[bit] !== 'undefined') return this.FLAGS[bit];
+      if (!isNaN(bit)) return BigInt(bit);
+    }
     throw new RangeError('BITFIELD_INVALID', bit);
   }
 }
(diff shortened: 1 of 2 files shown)


src/util/BitField.js
```diff
@@ -144,7 +145,10 @@ class BitField {
     if (typeof defaultBit === typeof bit && bit >= defaultBit) return bit;
     if (bit instanceof BitField) return bit.bitfield;
     if (Array.isArray(bit)) return bit.map(p => this.resolve(p)).reduce((prev, p) => prev | p, defaultBit);
-    if (typeof bit === 'string' && typeof this.FLAGS[bit] !== 'undefined') return this.FLAGS[bit];
+    if (typeof bit === 'string') {
+      if (typeof this.FLAGS[bit] !== 'undefined') return this.FLAGS[bit];
+      if (!isNaN(bit)) return BigInt(bit);
```
vladfrangu, line 150: This isn't accurate as not every bitfield is a bigint. It should be done on the class that needs it, no?

vladfrangu (dismissed):

src/util/BitField.js
DraftProducts: In the typedoc it is explicitly said that the value can be a BigInt, but your right, this is not the case for all bitFields.
There would therefore be two solutions:
  - Cast the different types according to an `instanceof` of` defaultBit`.
  - Extends all static `resolve` functions in all classes that use a different type than the default in the `BitField` class.
 
I would prefer the first one.
The utility of the `defaultBit` value is also to tell to ` BitField` from its extended class what the type of the bitfield is.
I would find it off-putting to have to modify the base function of each class when it can be done once in the `resolve`.

DraftProducts: Something like this I guess:
```js
if (!isNaN(bit)) return (typeof defaultBit === 'bigint' ? BigInt : Number)(bit);
// or
if (!isNaN(bit)) return typeof defaultBit === 'bigint' ? BigInt(bit) : Number(bit);
```

DraftProducts: I checked all BitField extended classes of discord.js and it seems that only Permissions use bigint.
So it should not be painful to extend `resolve` once.
The two options would be good I guess.

SpaceEEC: Since we already check the type of the default bit in `toJSON` it's probably fine to go with the first option.

typings/index.d.ts
```diff
@@ -3616,7 +3616,7 @@ declare module 'discord.js' {
 
   interface PermissionOverwriteOptions extends Partial<Record<PermissionString, boolean | null>> {}
 
-  type PermissionResolvable = BitFieldResolvable<PermissionString, bigint>;
+  type PermissionResolvable = BitFieldResolvable<PermissionString, string, bigint>;
```
SpaceEEC, line 3619: `BitFieldResolvable` only accepts two type parameters, the second one likely was intended to be `string | bigint`.

That change however would allow `Permissions.resolve("ADD_REAACTIONS")` to compile (and then throw in runtime), which is not wanted behavior.

Since the changes in this PR apply to all `BitField`s, I think we should make a change in `BitFieldResolvable` directly.
To avoid the above loss in typesafety, instead of adding `string` we should add `` `${bigint}` `` there. (Once in the union directly, once in the union of `RecursiveReadonlyArray`.)

SpaceEEC (changes requested):

typings/index.d.ts
DraftProducts: Okay
I wasn't sure about my typings
I will remove the second argument.
I think that some help for the rest would be necessary, I have not yet grasped all that this implies 🙏

DraftProducts: Something like that ?
```ts
  type BitFieldResolvable<T extends string, N extends number | bigint> =
    | RecursiveReadonlyArray<T | N | `${bigint}` | Readonly<BitField<T, N>>>
    | T
    | N
    | Readonly<BitField<T, N>>;
```

SpaceEEC: Almost, you also need to add the `${bigint}` part in the outmost union.
```ts
...
| T
| N
| `${bigint}`
...
```

SpaceEEC (approved):

iCrawl (approved):

vladfrangu (approved):

kyranet (approved):

PR #5918, opened 2021-06-24
refactor: removed `code` and `split` options

kyranet:

:broom: clean-up service here, removing old remnants from old versions of discord.js

:warning: Removed `APIMessage#split`
:warning: Removed `MessageEditOptions#code`
:warning: Removed `BaseMessageOptions#code`
:warning: Removed `BaseMessageOptions#split`

- **Why?**
  No other method in the entire library does two requests to the same path. This method was an exception for many years, and is something that should always been done by the user.
- **How do I restore this feature?**
  You can still use the `Util.splitMessage` utility, and then with given array, send all the messages as you desire.
  As for `code`, simply wrap the content into \`\`\`js\n${content}\`\`\`​. We're releasing an utility later in the future too, so you can do `codeblock('js', content)`.

### DIFF
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -225,10 +225,9 @@ declare module 'discord.js' {
     ): APIMessage;
     public static resolveFile(fileLike: BufferResolvable | Stream | FileOptions | MessageAttachment): Promise<unknown>;
 
-    public makeContent(): string | string[] | undefined;
+    public makeContent(): string | undefined;
     public resolveData(): this;
     public resolveFiles(): Promise<this>;
-    public split(): APIMessage[];
   }
 
   export abstract class Application {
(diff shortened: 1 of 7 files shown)


typings/index.d.ts
```diff
@@ -1183,12 +1182,7 @@ declare module 'discord.js' {
   export class InteractionWebhook extends PartialWebhookMixin() {
     constructor(client: Client, id: Snowflake, token: string);
     public token: string;
-    public send(
-      options: string | APIMessage | (InteractionReplyOptions & { split?: false }),
-    ): Promise<Message | RawMessage>;
-    public send(
-      options: APIMessage | (InteractionReplyOptions & { split: true | SplitOptions }),
-    ): Promise<(Message | RawMessage)[]>;
+    public send(options: string | APIMessage | InteractionReplyOptions)): Promise<Message | RawMessage>;
```
DTrombett, line 1185: Isn't there a bracket more?
```suggestion
    public send(options: string | APIMessage | InteractionReplyOptions): Promise<Message | RawMessage>;
```

iCrawl (approved):

typings/index.ts
```diff
@@ -29,7 +29,6 @@ client.on('messageReactionRemoveAll', async message => {
 
 // These are to check that stuff is the right type
```
vaporox, line 30: Very, very important
```suggestion
// This is to check that stuff is the right type
```

SpaceEEC (dismissed):
You missed one instance of splitting:
https://github.com/kyranet/discord.js/blob/dc02352247f67e290aefef1387cb768ffe92f61c/src/structures/Webhook.js#L169-L171

vladfrangu (approved):
🧹🧹

SpaceEEC (approved):

PR #5923, opened 2021-06-24
fix(User): fix bot and system properties being incorrect in some cases

ImRodry:
Due to the system property only being present in user objects when the user is a system user, it would end up always being null instead of false like it should be. This PR fixes that so that it's true for system users, false for non-system users and null for partial users (regardless of whether they're system or not, since we don't know that)
Another issue in the User class was that the bot property would always be false for partial users due to the check `typeof this.bot !== 'boolean'` which would return true since this.bot was set to null previously. This fixes that so that it's null for partial users, true for bots and false for regular users.

⚠️ I did test these changes but only by creating partial objects myself since I couldn't find a way to easily obtain an actual partial user object. Please let me know if there's something that could go wrong that I didn't pick up in my tests.

### DIFF
--- a/src/structures/User.js
+++ b/src/structures/User.js
@@ -59,13 +59,13 @@ class User extends Base {
       this.username = null;
     }
 
-    if ('bot' in data || typeof this.bot === 'boolean') {
+    if ('bot' in data) {
       /**
        * Whether or not the user is a bot
        * @type {?boolean}
        */
-      this.bot = Boolean(data.bot ?? this.bot);
-    } else if (!this.partial) {
+      this.bot = Boolean(data.bot);
+    } else if (!this.partial && typeof this.bot !== 'boolean') {
       this.bot = false;
     }
 
@@ -89,13 +89,13 @@ class User extends Base {
       this.avatar = null;
     }
 
-    if ('system' in data || typeof this.system === 'boolean') {
+    if ('system' in data) {
       /**
        * Whether the user is an Official Discord System user (part of the urgent message system)
        * @type {?boolean}
        */
-      this.system = Boolean(data.system ?? this.system);
-    } else if (!this.partial) {
+      this.system = Boolean(data.system);
+    } else if (!this.partial && typeof this.system !== 'boolean') {
       this.system = false;
     }
 
--- a/src/structures/User.js
+++ b/src/structures/User.js
@@ -27,7 +27,10 @@ class User extends Base {
      */
     this.id = data.id;
 
+    this.bot = null;
+
     this.system = null;
+
     this.flags = null;
 
     /**
@@ -56,12 +59,14 @@ class User extends Base {
       this.username = null;
     }
 
-    if ('bot' in data || typeof this.bot !== 'boolean') {
+    if ('bot' in data) {
       /**
        * Whether or not the user is a bot
        * @type {boolean}
        */
       this.bot = Boolean(data.bot);
+    } else if (!this.partial) {
+      this.bot = false;
     }
 
     if ('discriminator' in data) {
@@ -90,6 +95,8 @@ class User extends Base {
        * @type {?boolean}
        */
       this.system = Boolean(data.system);
+    } else if (!this.partial) {
+      this.system = false;
     }
 
     if ('public_flags' in data) {
--- a/src/structures/User.js
+++ b/src/structures/User.js
@@ -62,7 +62,7 @@ class User extends Base {
     if ('bot' in data) {
       /**
        * Whether or not the user is a bot
-       * @type {boolean}
+       * @type {?boolean}
        */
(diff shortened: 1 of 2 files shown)


SpaceEEC (dismissed):
When patching an already cached user with partial data (happens in audit logs and wherever we'd get a partial user in the future), the `bot` and `system` property would be assigned `false`.

Can be reproduced by the following steps:
- Enable the `USER` partial
- Add/remove a role from/to a bot
- Fetch audit log
- Inspect `User#bot` of that bot and find `false`

Same thing should apply to `system`, but I don't have a system user to assign a role to at hand 😄 

The ideal behavior of `structure._patch({})` for potentially partial structures would be no changes. (Although this is not yet the case for all instances, but we are slowly getting there.)

ImRodry:
> When patching an already cached user with partial data (happens in audit logs and wherever we'd get a partial user in the future), the `bot` and `system` property would be assigned `false`.
> 
> Can be reproduced by the following steps:
> 
> * Enable the `USER` partial
> * Add/remove a role from/to a bot
> * Fetch audit log
> * Inspect `User#bot` of that bot and find `false`
> 
> Same thing should apply to `system`, but I don't have a system user to assign a role to at hand 😄
> 
> The ideal behavior of `structure._patch({})` for potentially partial structures would be no changes. (Although this is not yet the case for all instances, but we are slowly getting there.)

I tried to reproduce this and got the same behavior but that's because the user is not classified as a partial since it's username is a string (at least it was in my case).
I believe this happened because bot was not present in the data used to create the user but the partial property returned false so it's _technically_ the expected behavior. I believe partial should have more checks than just the type of the user's username to prevent situations like this.

ImRodry:
The previously mentioned issue should be fixed as I made it so that it takes both the received data and the existing data for the user into account. So if one isn't present, the other one will be kept instead of being overwritten with false for non-partial users. Please let me know if there are any other issues with this approach

typings/index.d.ts
```diff
@@ -1868,7 +1868,7 @@ declare module 'discord.js' {
     public lastMessageID: Snowflake | null;
     public readonly partial: false;
     public readonly presence: Presence;
-    public system: boolean | null;
+    public system: boolean;
```
vladfrangu, line 1871: Can't it still be null?

vladfrangu (changes requested):

typings/index.d.ts
ImRodry: Only if the object is a partial, otherwise it will always be a boolean

DTrombett: Shouldn't this be `boolean` in jsdoc too? There it's still `?boolean`.

ImRodry: No, because the JSDoc can’t tell if the object is a partial or not, and the typings can

src/structures/User.js
```diff
@@ -56,12 +59,14 @@ class User extends Base {
...
-    if ('bot' in data || typeof this.bot !== 'boolean') {
+    if ('bot' in data || typeof this.bot === 'boolean') {
       /**
        * Whether or not the user is a bot
-       * @type {boolean}
+       * @type {?boolean}
        */
-      this.bot = Boolean(data.bot);
+      this.bot = Boolean(data.bot ?? this.bot);
+    } else if (!this.partial) {
```
NotSugden, line 68: wouldn't it make more sense to
`else if (!this.partial && typeof this.bot !== 'boolean')` instead

src/structures/User.js
```diff
@@ -84,12 +89,14 @@ class User extends Base {
...
-    if ('system' in data) {
+    if ('system' in data || typeof this.system === 'boolean') {
       /**
        * Whether the user is an Official Discord System user (part of the urgent message system)
        * @type {?boolean}
        */
-      this.system = Boolean(data.system);
+      this.system = Boolean(data.system ?? this.system);
+    } else if (!this.partial) {
+      this.system = false;
```
NotSugden, line 99: same here

src/structures/User.js
ImRodry: Would this replace the typeof this.bot check above or is it just an additional check?

NotSugden: it would replace it

ImRodry: Technically both work, why do you prefer this approach?

NotSugden: because its essentially just adding `this.bot = Boolean(this.bot)` for no reason, just makes more sense

ImRodry: There are instances where data.bot does have a value, but I understand what you're saying so I'll change that

SpaceEEC (approved):

iCrawl (approved):

kyranet (approved):

PR #5928, opened 2021-06-26
types(ThreadMemberManager): fix incorrect usage

ckohen:

Made another oopsie


iCrawl (approved):

kyranet (commented):
Doesn't the usage of `implements` instead of `extends` also work in here?

ckohen:
I could do that yes, but that also requires duplication of all properties / methods in BaseManager (aside from add of course) in ThreadMemberManager because implements doesn't inherently copy those.

kyranet:
Oh, then it's fine.

kyranet (approved):

PR #5929, opened 2021-06-26
fix(Message): Make author of referenced message nullable

SkyFlags:
After #5905 PR, `data.referenced_message` property has been added to Message class (optional, non-nullable). Actually, however, the property is also nullable. This PR makes that nullable, and fixes issue caused by non-nullable property.

p.s. My english is not good. please understand if you find any wrong grammar..

### DIFF
--- a/src/structures/Message.js
+++ b/src/structures/Message.js
@@ -332,7 +332,7 @@ class Message extends Base {
       'mention_roles' in data ? data.mention_roles : this.mentions.roles,
       'mention_everyone' in data ? data.mention_everyone : this.mentions.everyone,
       'mention_channels' in data ? data.mention_channels : this.mentions.crosspostedChannels,
-      'referenced_message' in data ? data.referenced_message?.author ?? null : this.mentions.repliedUser,
+      data.referenced_message ? data.referenced_message.author : this.mentions.repliedUser,
     );
 
     this.flags = new MessageFlags('flags' in data ? data.flags : 0).freeze();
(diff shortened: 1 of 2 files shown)


SpaceEEC (approved):

iCrawl (approved):

kyranet (approved):

iShibi:
Why pass `null` tho? The referenced message was deleted but this doesn't mean that the current message is not a reply to the author of that deleted referenced message anymore.

src/structures/Message.js
```diff
@@ -332,7 +332,7 @@ class Message extends Base {
       'mention_roles' in data ? data.mention_roles : this.mentions.roles,
       'mention_everyone' in data ? data.mention_everyone : this.mentions.everyone,
       'mention_channels' in data ? data.mention_channels : this.mentions.crosspostedChannels,
-      'referenced_message' in data ? data.referenced_message.author : this.mentions.repliedUser,
+      data.referenced_message ? data.referenced_message.author : this.mentions.repliedUser,
```
vladfrangu, line 335: 🧹 
```suggestion
      data.referenced_message?.author ?? this.mentions.repliedUser,
```

vladfrangu (changes requested):

src/structures/Message.js
SkyFlags: Updated. It's my first experience opening a PR. Please understand my mistake 😂

vladfrangu (approved):

PR #5930, opened 2021-06-26
feat(User): support user banners

Koyamie:
With the recent releases of user banners for everyone, Discord confirmed they will add the banner image hash to the user object.
See https://github.com/discord/discord-api-docs/issues/3095#issuecomment-868817186

This PR supports the banner field.

Do not merge until the field is accessible via the API.

### DIFF
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -570,7 +570,7 @@ declare module 'discord.js' {
           format: 'default' | AllowedImageFormat,
           size: number,
         ) => string;
-        Banner: (guildID: Snowflake | number, hash: string, format: AllowedImageFormat, size: number) => string;
+        Banner: (id: Snowflake | number, hash: string, format: AllowedImageFormat, size: number) => string;
         Icon: (
           userID: Snowflake | number,
           hash: string,
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -569,13 +569,21 @@ declare module 'discord.js' {
           hash: string,
           format: 'default' | AllowedImageFormat,
           size: number,
+          dynamic: boolean,
         ) => string;
-        Banner: (id: Snowflake | number, hash: string, format: AllowedImageFormat, size: number) => string;
+        Banner: (
+          id: Snowflake | number,
+          hash: string,
+          format: AllowedImageFormat,
+          size: number,
+          dynamic: boolean,
+        ) => string;
         Icon: (
           userID: Snowflake | number,
           hash: string,
           format: 'default' | AllowedImageFormat,
           size: number,
+          dynamic: boolean,
         ) => string;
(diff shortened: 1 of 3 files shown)


DTrombett:
Typings for `User` class need to be updated.

Koyamie:
I am not familiar with typings but this should do the job? Tell me if I missed something

DTrombett:
They should be ok.

GoldenAngel2:
`User.equals` needs to be updated

typings/index.d.ts
```diff
@@ -570,7 +570,7 @@ declare module 'discord.js' {
           format: 'default' | AllowedImageFormat,
           size: number,
         ) => string;
-        Banner: (guildID: Snowflake | number, hash: string, format: AllowedImageFormat, size: number) => string;
+        Banner: (id: Snowflake | number, hash: string, format: AllowedImageFormat, size: number) => string;
```
GoldenAngel2, line 573: This needs to be changed to 
```js
Banner: (id: Snowflake | number, hash: string, format: AllowedImageFormat, size: number, dynamic: boolean) => string;
```

GoldenAngel2 (changes requested):

typings/index.d.ts
Koyamie: I was wondering if so and looked at the other functions, dynamic does not appear. Isn't it because dynamic is declared here ? 
https://github.com/discordjs/discord.js/blob/e8134d558d4ad5da4a37b0b120be14a15bf70b61/typings/index.d.ts#L3512

GoldenAngel2: Hmm I guess it doesn't need to added there. just from the actual function it needed to be there. But I guess not lol 👍

DraftProducts (approved):

typings/index.d.ts
SpaceEEC: `dynamic` missing is probably an oversight, I can't think of any reason it shouldn't be there.

Koyamie: That was my first thought, I've added the 3 missing `dynamic`.

u9g:
In what situation would the users be equal but the banners be unequal that makes the equal function fail?

DTrombett:
That's the same as other properties. I guess it's there to check two users object that may refer to the same user.

advaith1:
`banner` and `banner_color` are documented in https://github.com/discord/discord-api-docs/pull/3448 and are currently accessible via the canary API (in ClientOptions: `http: {headers: {'x-debug-options': 'canary'}}` or `http: {api: 'https://canary.discord.com/api'}`)

Koyamie:
I'll need someone else to take care of that PR because I won't be available for the next 2 weeks and Discord seems to be slowly releasing the banner on the API.

iCrawl:
@advaith1 will be taking over this in a new PR.

PR #5933, opened 2021-06-26
fix(Sharding): strict type context and return

kyranet:

Right now, `ShardClientUtil#broadcastEval` and `ShardingManager#broadcastEval`'s return types are the same as the callback's, as well as the context passed, however, the context data is serialized, and so is the return type.

This PR implements `Serialized`, which fixes this issue by converting rich types into their serialized counterpart.

### DIFF
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -133,7 +133,7 @@ declare enum WebhookTypes {
   'Channel Follower' = 2,
 }
 
-type Awaited<T> = T | Promise<T>;
+type Awaited<T> = T | PromiseLike<T>;
 
 declare module 'discord.js' {
   import BaseCollection from '@discordjs/collection';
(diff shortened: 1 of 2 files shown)


kyranet:
I'm unsure how to properly type the returns because of the worker mode:

https://github.com/discordjs/discord.js/blob/985d4d6a438fc9123264f6a1c600f34fccc1825f/src/sharding/ShardClientUtil.js#L92-L95

[`process.send`](https://nodejs.org/api/process.html#process_process_send_message_sendhandle_options_callback) seems to JSON serialize values, however, the worker mode supports sending a lot richer data, as this is powered by V8's serder instead.

iCrawl:
This needs a rebase.

iCrawl:
Can't we just go with this for now, despite the worker mode and potentially wait for an issue in case anyone is using this to send more complicated data?

LosTigeros:
What about executing functions that would return a Promise on the shard?
Can't [this](https://github.com/discordjs/discord.js/blob/master/src/sharding/ShardingManager.js#L285) be like:
```js
promises.push(Promise.resolve(sh[method](...args)));
```
This would work for normal functions and PromiseLike ones so `broadcastEval` `script` argument could be either `T` or `Promise<T>`

kyranet:
Yes, we can go for this for now, and if we wanted to make both consistent, we can always use [V8's serialization API](https://nodejs.org/dist/latest-v16.x/docs/api/v8.html#v8_serialization_api), which is what worker threads use, @iCrawl.

Nice catch @LosTigeros! I changed the return of the callbacks from `T` to `Awaited<T>`, so in case it's a promise, the promise's type is unwrapped. Also added tests to make sure they work as expected, see 52ec31f :)

iCrawl (approved):

SpaceEEC (approved):

amishshah (approved):

vladfrangu (approved):

PR #5935, opened 2021-06-26
feat(Speaking): adds Speaking class and other "speaking" properties

PraveshKunwar:

This pr updates some of the typings for the "Speaking class" and other "speaking properties". Here they are:

Added Speaking class (typings): https://discord.js.org/#/docs/main/stable/class/Speaking
Added VoiceState.speaking property (typing): https://discord.js.org/#/docs/main/stable/class/VoiceState?scrollTo=speaking
Added guildMemberSpeaking event (typing): https://discord.js.org/#/docs/main/stable/class/Client?scrollTo=e-guildMemberSpeaking

THIS DOES NOT ADD:

VoiceConnection class or VoiceConnection.speaking (event and property)
BroadDispatcher class or BroadDispatcher.speaking
VoiceWebSocket#startSpeaking

### DIFF
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -1809,6 +1809,11 @@ declare module 'discord.js' {
     public static readonly EPOCH: number;
   }
 
+  export class Speaking extends Bitfield<SpeakingString> {
+    public static FLAGS: Record<SpeakingString, number>;
+    public static resolve(bit?: BitFieldResolvable<SpeakingString, number>): number;
+  }
+
   export class StageChannel extends BaseGuildVoiceChannel {
     public topic: string | null;
     public type: 'stage';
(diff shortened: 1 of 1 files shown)


amishshah:
We've removed support for voice in v13 - the documentation you're referencing is for v12.

We are recommending users look at [@discordjs/voice](https://github.com/discordjs/voice) instead, a separate add-on that provides a more stable voice implementation.

While we are keeping voice states in v13 (since they're not exclusive to the Voice API), the speaking property in particular can only be derived with an active voice connection. This is therefore not possible to achieve in discord.js v13, and is an issue concerned with the new voice library instead.

PraveshKunwar:
Sounds good 👍

PR #5936, opened 2021-06-27
feat(InteractionCollector): reworked to be more generic

monbrey:
Another half-baked idea that comes out of late night conversations and early morning coding.

This refactors the `MessageComponentInteractionCollector` into a generic `InteractionCollector` that can be used to collect any sort of interaction (yes, even Slash Commands).

It also removes the need for it to be bound to any particular message or channel, though it still can be should you want to do that. These are still separate options to the filter as it allows the Collector to also listen to the relevant DELETE events and cancel itself accordingly. It will use these based on the most specific option first, so Message > Channel > Guild.

Draft for now - I'd still like reviews and opinions from other people. I intend to add some checking to the constructor to account for options that don't make sense together, e.g. setting a `componentType` option if `interactionType` is set for slash commands.

**Status and versioning classification**


src/structures/InteractionCollector.js
```diff
@@ -17,26 +23,49 @@ const { Events } = require('../util/Constants');
...
      * @type {?Message}
      */
-    this.message = source instanceof require('./Message') ? source : null;
+    this.message = options.message ?? null;
 
     /**
-     * The source channel from which to collect message component interactions
-     * @type {TextChannel|DMChannel|NewsChannel}
+     * The channel from which to collect interactions, if provided
+     * @type {?TextChannel|DMChannel|NewsChannel}
```
almeidx, line 42:
```suggestion
     * @type {?(TextChannel|DMChannel|NewsChannel)}
```

almeidx (changes requested):

src/structures/InteractionCollector.js
NotSugden: isn't it the same type either way

AustrianDev: Doesn't TextBasedChannel work?

monbrey: TextBasedChannel isn't a jsdoc type, its an interface applied to other classes.

typings/index.d.ts
```diff
@@ -1183,6 +1183,34 @@ declare module 'discord.js' {
...
 
+  export class InteractionCollector<T extends Interaction> extends Collector<Snowflake, T> {
+    constructor(client: Client, options?: InteractionCollectorOptions<T>);
+    private _handleMessageDeletion(message: Message): void;
+    private _handleChannelDeletion(channel: GuildChannel): void;
+    private _handleGuildDeletion(guild: Guild): void;
+
+    public channel: TextChannel | NewsChannel | DMChannel | null;
+    public componentType: MessageComponentType | null;
+    public empty(): void;
```
SpaceEEC, line 1194: Must have slipped through before, but methods should go below properties.

SpaceEEC (dismissed):

src/structures/InteractionCollector.js
```diff
@@ -51,23 +80,26 @@ class MessageComponentInteractionCollector extends Collector {
...
+      this.client.on(Events.MESSAGE_DELETE, this._handleMessageDeletion);
+    }
+    if (this.channel) {
+      this._handleChannelDeletion = this._handleChannelDeletion.bind(this);
+      this.client.on(Events.CHANNEL_DELETE, this._handleChannelDeletion);
+    }
+    if (this.guild) {
+      this._handleGuildDeletion = this._handleGuildDeletion.bind(this);
+      this.client.on(Events.GUILD_DELETE, this._handleGuildDeletion);
+    }
```
vladfrangu, line 95: Small looks nit, alongside a request to increment the max listeners per client.on call :D

```suggestion
    if (this.message) {
      this._handleMessageDeletion = this._handleMessageDeletion.bind(this);
      this.client.on(Events.MESSAGE_DELETE, this._handleMessageDeletion);
    }

    if (this.channel) {
      this._handleChannelDeletion = this._handleChannelDeletion.bind(this);
      this.client.on(Events.CHANNEL_DELETE, this._handleChannelDeletion);
    }

    if (this.guild) {
      this._handleGuildDeletion = this._handleGuildDeletion.bind(this);
      this.client.on(Events.GUILD_DELETE, this._handleGuildDeletion);
    }
```

vladfrangu (dismissed):

typings/index.d.ts
```diff
@@ -1183,6 +1183,34 @@ declare module 'discord.js' {
...
+    public guild: Guild | null;
+    public interactionType: InteractionType | null;
+    public message: Message | null;
+    public options: InteractionCollectorOptions<T>;
+    public total: number;
+    public users: Collection<Snowflake, User>;
+
+    public collect(interaction: Interaction): Snowflake;
+    public dispose(interaction: Interaction): Snowflake;
+    public on(event: 'collect' | 'dispose', listener: (interaction: T) => Awaited<void>): this;
```
NotSugden, line 1205:
```suggestion
    public on(event: 'collect' | 'dispose', listener: (interaction: T) => void): this;
```
void works just the same, no real reason to use `Awaited<void>` over just `void` ?

DTrombett: It's the same as `Client#on()`, which uses Awaited too

monbrey: Pretty sure the reason it was switched to `Awaited<void>` is because otherwise you can't have async listeners.

DTrombett: Yeah just found #5602

NotSugden: didn't get any errors on [ts playground](https://www.typescriptlang.org/play?#code/MYewdgzgLgBAZgVzMGBeGAKOYBcmAeeARiCADYCmAhmAJRoB8MAbiAJYAm9qT2GUAJwQVaAbgBQ4xMgxUIATxncmAb3EwNMARSgIBYGACJoAtmADmhiQF8xk6cAwZlMNZq069BsBQDuMAAUBEABbNggKAB4TM3MGDG0IcmYKRg8kshSMY0FYw1o7W1EgA) 🤷 strange

almeidx: > didn't get any errors on [ts playground](https://www.typescriptlang.org/play?#code/MYewdgzgLgBAZgVzMAXDAFHMb0A80BGIIANgKYCGYAlDALwB8MAbiAJYAmtjL7H9GLDnwwipSjXpNWnbkyzooAJwRlqAbgBQmxMnQUIATz1yYAb00wrMJWSgIlYGACJoStmADmzrQF8N2rrA6OimFtY2dg5OYGQA7jAACkogALZsEGQAPG4engzothCkzGRSkcUkpeiuynnO1AH+6kA)  strange

From the pull request @DTrombett linked
> When using `client.on` with a callback that returns a `Promise`, eslint will then complain if the `no-misused-promises` (from `@typescript-eslint` and various plugins) is enabled, as `Promise<?>` does not match `void`.

NotSugden: ah, makes sense

src/structures/InteractionCollector.js
monbrey: I don't think it's necessary to do it per call? The maxListeners property determines how many listeners can be bound to each event, not in total. This is only binding one listener to each event, so incrementing once should be sufficient.

I'll move the increment call to the top though.

kyranet (approved):

iCrawl:
This needs a rebase.

SpaceEEC (changes requested):

vladfrangu (approved):

vladfrangu (approved):

PR #5937, opened 2021-06-27
fix(Collector): docs and types

vaporox:

This PR is smaller than it sounds but I'm explaining every change in detail here because I don't want to open yet another fix PR lol

Docs:
- The return type of `MessageComponentInteractionCollector#collect()` was changed from `?(Snowflake|string)` to just `?Snowflake` because it only returns either null or the interaction ID
- The return type of `ReactionCollector#collect()` was changed from `Promise<Snowflake|string>` to `Promise<?(Snowflake|string)>` because it can return null as well
- The return type of `Collector#collect()` was changed from... whatever that was before to `?(*|Promise<?*>)` so that extended classes can specify whether it returns a promise or not

Types:
- `Collector` got another type parameter for additional filter parameters
- `Collector#collect()` can now return a promise, which should be specified on extended classes as explained above
- All `collect()` and `dispose()` methods got null added to their return type (probably an oversight)
- A `user` parameter was added to `ReactionCollector#collect()` (probably another oversight)

Miscellaneous:
- An `await` keyword was added in `Collector#handleCollect()` because `this.collect()` is asynchronous for reaction collectors

### DIFF
--- a/src/structures/MessageComponentInteractionCollector.js
+++ b/src/structures/MessageComponentInteractionCollector.js
@@ -82,7 +82,7 @@ class MessageComponentInteractionCollector extends Collector {
   /**
    * Handles an incoming interaction for possible collection.
    * @param {Interaction} interaction The interaction to possibly collect
-   * @returns {?(Snowflake|string)}
+   * @returns {?Snowflake}
    * @private
    */
   collect(interaction) {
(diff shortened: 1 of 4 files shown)


monbrey:
#5936 might impact this, at least in regards to the `MessageComponentInteractionCollector` (which no longer exists)

iCrawl (approved):

kyranet (approved):

vladfrangu (approved):

PR #5939, opened 2021-06-27
fix(Message): #patch throws if referenced message was deleted

iShibi:

This PR fixes an issue where patching a message will throw error if the message it replied to has been deleted. This happens because the lib only checks for the presence of `referenced_message` and not whether its `null` or not.

Thanks, @toastersticks for the 🕵️ work


iShibi:
Closing because #5929

PR #5948, opened 2021-06-28
feat(CategoryChannel): ✨ Return name with 'CategoryChannel#toString'

DraftProducts:

Currently, the stringification leads to a mention of vocal and textual salons but not for categories.
image

The mention simply indicates the name of the category and prefixes it with a `#` which is the prefix of the textual channels.

The objective of this PR would be to extend the `CategoryChannel#toString` method to return the name of the channel.

### DIFF
--- a/src/structures/CategoryChannel.js
+++ b/src/structures/CategoryChannel.js
@@ -17,6 +17,17 @@ class CategoryChannel extends GuildChannel {
   }
 
   /**
+   * When concatenated with a string, this automatically returns the channel's name instead of the Category object.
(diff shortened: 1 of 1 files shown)


almostSouji:
`<#id>` is still the valid mention format for category channels. The prefix is simply how discord chooses to display it, as they choose to embed "button" features for text-like and voice channels.

changing this would introduce an inconsistency as we'd no longer use the client's mention representation for a specific type of channel.

DraftProducts:
Sure, ok 👍

PR #5951, opened 2021-06-28
feat(StageChannel): add createStageInstance method & use better naming convention

iShibi:

This PR adds a new `createStageInstance` method on `StageChannel` that lets the user create a stage instance associated to it. It also refactors the options typedef to keep it consistent with rest of the lib:

1) `CreateStageInstanceOptions` is now `StageInstanceCreateOptions`
2) The `channel` field is a seperate param now instead of being inside the options object

⚠️ As pointed out by @SpaceEEC, the term **"instance"** makes it sound like it is an instance of a Stage Channel which it isn't. This PR removes this ambiguity by using the complete term i.e. **"Stage Instance"** in all places. 

- This PR changes the library's interface
- This PR includes breaking changes

### DIFF
--- a/src/managers/StageInstanceManager.js
+++ b/src/managers/StageInstanceManager.js
@@ -35,7 +35,7 @@ class StageInstanceManager extends BaseManager {
 
   /**
    * Creates a new stage instance.
-   * @param {StageChannel|Snowflake} channel The stage channel whose instance is to be created
+   * @param {StageChannel|Snowflake} channel The stage channel to associate the created instance to
    * @param {StageInstanceCreateOptions} options The options to create the stage instance
    * @returns {Promise<StageInstance>}
    * @example
--- a/src/managers/StageInstanceManager.js
+++ b/src/managers/StageInstanceManager.js
@@ -35,7 +35,7 @@ class StageInstanceManager extends BaseManager {
 
   /**
    * Creates a new stage instance.
-   * @param {StageChannel|Snowflake} channel The stage channel to associate the created instance to
+   * @param {StageChannel|Snowflake} channel The stage channel to associate the created stage instance to
    * @param {StageInstanceCreateOptions} options The options to create the stage instance
    * @returns {Promise<StageInstance>}
    * @example
@@ -68,7 +68,7 @@ class StageInstanceManager extends BaseManager {
 
   /**
    * Fetches the stage instance associated with a stage channel, if it exists.
-   * @param {StageChannel|Snowflake} channel The stage channel whose instance is to be fetched
+   * @param {StageChannel|Snowflake} channel The stage channel whose associated stage instance is to be fetched
    * @param {BaseFetchOptions} [options] Additional options for this fetch
    * @returns {Promise<StageInstance>}
    * @example
@@ -99,7 +99,7 @@ class StageInstanceManager extends BaseManager {
 
   /**
    * Edits an existing stage instance.
-   * @param {StageChannel|Snowflake} channel The stage channel whose instance is to be edited
+   * @param {StageChannel|Snowflake} channel The stage channel whose associated stage instance is to be edited
    * @param {StageInstanceEditOptions} options The options to edit the stage instance
    * @returns {Promise<StageInstance>}
    * @example
@@ -135,7 +135,7 @@ class StageInstanceManager extends BaseManager {
 
   /**
    * Deletes an existing stage instance.
-   * @param {StageChannel|Snowflake} channel The stage channel whose instance is to be deleted
+   * @param {StageChannel|Snowflake} channel The stage channel whose associated stage instance is to be deleted
    * @returns {Promise<void>}
    */
   async delete(channel) {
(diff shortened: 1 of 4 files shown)


kyranet (commented):
I don't agree with the breaking change as the object still has a required property (`topic`).

If it's because `StageChannel#createStageInstance` (probably should be `createInstance`), you can do `create({ ...options, channel: this.id })`.

iShibi:
It will be the same situation like `MessageOptions` again, where `messageReference` was getting set to `this.id` and hence `ReplyMessageOptions` was created. Unlike there, where we couldn't make `messageReference` a seperate param, it is much more intuitive to have `channel` as a first param to `stageInstances#create` here.

`createInstance` sounds good tho, will rename it.

src/managers/StageInstanceManager.js
```diff
@@ -28,31 +28,30 @@ class StageInstanceManager extends BaseManager {
...
-   * @property {StageChannel|Snowflake} channel The stage channel whose instance is to be created
+   * @typedef {Object} StageInstanceCreateOptions
    * @property {string} topic The topic of the stage instance
    * @property {PrivacyLevel|number} [privacyLevel] The privacy level of the stage instance
    */
 
   /**
    * Creates a new stage instance.
-   * @param {CreateStageInstanceOptions} options The options to create the stage instance
+   * @param {StageChannel|Snowflake} channel The stage channel whose instance is to be created
```
SpaceEEC, line 38: See my comment below:
```suggestion
   * @param {StageChannel|Snowflake} channel The stage channel to associate the created instance to
```

SpaceEEC (dismissed):

src/structures/StageChannel.js
```diff
@@ -28,6 +28,15 @@ class StageChannel extends BaseGuildVoiceChannel {
     return this.guild.stageInstances.cache.find(stageInstance => stageInstance.channelID === this.id) ?? null;
   }
 
+  /**
+   * Creates an instance of this stage channel.
```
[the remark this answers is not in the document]
iShibi: Prior to this, the word `instance` was "supposed" to be short for `stage instance` but now that you pointed it out, using the short term does allow opportunities for making some wrong assumptions here. **"Stage Instance of a Stage Channel"** is much unambiguous than **"Instance of a Stage Channel"**. Can you add "stage" before the "instance" in these two suggestions so I can commit them now. I'll make necessary changes in other places. Thanks for noticing this 😸

iShibi:
Refactored some stuff to reflect the new naming convention for things related to stage instances. Added more information covering this change in the description above and changed the title.

*"instance of stage channel" What was I thinking pepehands*

src/managers/StageInstanceManager.js
```diff
@@ -28,31 +28,30 @@ class StageInstanceManager extends BaseManager {
...
    *  topic: 'A very creative topic',
    *  privacyLevel: 'GUILD_ONLY'
    * })
    *  .then(stageInstance => console.log(stageInstance))
    *  .catch(console.error);
    */
-  async create(options) {
-    if (typeof options !== 'object') throw new TypeError('INVALID_TYPE', 'options', 'object', true);
-    let { channel, topic, privacyLevel } = options;
+  async create(channel, options) {
```
vladfrangu, line 50: Same as kyra, I do not understand why channel is now outside of the options object, and personally I'd prefer it be a full object. Thoughts @discordjs/the-big-3?

vladfrangu (dismissed):

src/managers/StageInstanceManager.js
kyranet: The unified approach is consistent with [`RoleManager`](https://discord.js.org/#/docs/main/master/class/RoleManager?scrollTo=create), [`ThreadManager`](https://discord.js.org/#/docs/main/master/class/ThreadManager?scrollTo=create), and a few more places. Although I think in some places, `create` is split in two, such as [`GuildManager`](https://discord.js.org/#/docs/main/master/class/GuildManager?scrollTo=create), [`GuildChannelManager`](https://discord.js.org/#/docs/main/master/class/GuildChannelManager?scrollTo=create), and [`GuildBanManager`](https://discord.js.org/#/docs/main/master/class/GuildBanManager?scrollTo=create).

There's also one with three arguments: [`GuildEmojiManager`](https://discord.js.org/#/docs/main/master/class/GuildEmojiManager?scrollTo=create).

Something needs to be done in a later PR to merge those in objects and increase the consistency across the library, to be honest.

iShibi: I seperated `channel` because then the typedef `StageInstanceCreateOptions` can be reused in `StageChannel#createStageInstance`. The other option which Kyra suggested was to keep the `channel` but that would render it useless for the said method as it eventually gets set to `this.id` inside it. `ReplyMessageOptions` was created to address exactly this for `Message#reply`. Are we going to go back to the same thing again?

If we do want to have a single param here then I think we can create a new typedef for `options` param in `StageChannel#createStageInstance` that will not have the `channel` property.

kyranet: We discussed this internally, @iCrawl says to keep them separated, and if we want to, we can merge arguments in a later PR.

kyranet (approved):

SpaceEEC (approved):

iCrawl (approved):

PR #5955, opened 2021-06-28
feat(Interaction): add guild guard

### LINKED ISSUE #5809: Type guard for interactions in guilds
Many of an interaction's properties rely on each other, to get correct types the following condition needs to be used:
```ts
interaction.guild && interaction.guildID && interaction.member
```

And for CommandInteraction I have found that this additional condition is necessary:
```ts
interaction.guild && interaction.guildID && interaction.member && interaction.channel instanceof GuildChannel
```

Something like `inGuild()` which would be a shortcut for the above if statements.

Leave it up to the user, which is fully possible but perhaps not the most prettiest to repeat.

papsavas:

closes #5809

This PR creates an `inGuild()` method ~~both~~ for `Interaction` ~~and `CommandInteraction`~~  checking if an interaction is used in a guild. Feel free to add suggestions, such as naming or implementation ,as i compeletely followed the ones from the issue

### DIFF
--- a/src/structures/Interaction.js
+++ b/src/structures/Interaction.js
@@ -106,6 +106,14 @@ class Interaction extends Base {
   }
 
   /**
+   * Indicates whether this interaction is received from a guild.
+   * @returns {boolean}
+   */
+  inGuild() {
+    return this.guild && this.guildID && this.member;
+  }
+
+  /**
    * Indicates whether this interaction is a command interaction.
    * @returns {boolean}
    */
--- a/src/structures/Interaction.js
+++ b/src/structures/Interaction.js
@@ -110,7 +110,7 @@ class Interaction extends Base {
    * @returns {boolean}
    */
   inGuild() {
-    return this.guildID && this.member;
+    return Boolean(this.guildID && this.member);
   }
 
   /**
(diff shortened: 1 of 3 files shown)


monbrey:
I like the idea, but I'm not sure it should be relying on two other getters for the validation. `guildID` and `member` are the only ones you receive data for in the constructor.

`channel` and `guild` both assume this is a bot user with the `GUILDS` intent enabled, which means this could return false negatives if those caches aren't complete.

papsavas:
@monbrey in that case i suppose there is no need for an inherited method in `CommandInteraction`, just one `<Interaction>#isGuild()`

src/structures/Interaction.js
```diff
@@ -105,6 +105,14 @@ class Interaction extends Base {
     return this.client.guilds.cache.get(this.guildID) ?? null;
   }
 
+  /**
+   * Indicates whether this interaction is received from a guild.
+   * @returns {boolean}
+   */
+  inGuild() {
+    return this.guildID && this.member;
```
NotSugden, line 113:
```suggestion
    return Boolean(this.guildID && this.member);
```
so it actually returns a boolean

kyranet (approved):

iCrawl (approved):

SpaceEEC (approved):

vladfrangu (approved):

PR #5959, opened 2021-06-29
chore(Threads): general fixup and catch up on features

ckohen:

Fixes included: 
- Typings and docs corrections for `GuildChannelManager`, oversight from #5570 
- Docs example for thread creation made more clear
- Fixes subtle bug with `GuildChannelManager#resolve` due to `ThreadChannel`s not actually extending `GuildChannel`
- Changes `GuildMember#permissionsIn` to use `channel.permissionsFor` instead of `channel.memberPermissions` as `TreadChannel`s do not have this method and it is "private"

Catchup:
- Add new guild features for threads and necessary warnings
- Add new API Error that comes with boost locked thread features

This is unfortunately necessary 😢
I'm making the breaking change to `ThreadManager#create` that will probably be made to the API in its  next version.


src/util/Constants.js
```diff
@@ -764,6 +764,7 @@ exports.APIErrors = {
   ANNOUNCEMENT_EDIT_LIMIT_EXCEEDED: 20022,
   CHANNEL_HIT_WRITE_RATELIMIT: 20028,
   CONTENT_NOT_ALLOWED: 20031,
+  GUILD_PREMIUM_LEVEL_TOO_LOW: 20035,
```
kyranet, line 767: | ⚠️ Pending on https://github.com/discord/discord-api-docs/pull/3189 to be merged before we can include this change in master. |
| --- |

kyranet (approved):

src/managers/ThreadManager.js
```diff
@@ -91,14 +91,19 @@ class ThreadManager extends BaseManager {
    *   .create({
    *     name: 'food-talk'
```
papsavas, line 92: missed a `,` here

ckohen: Thanks, seems that was always missing 🤦‍♂️

src/managers/GuildChannelManager.js
```diff
@@ -50,27 +51,30 @@ class GuildChannelManager extends BaseManager {
...
    */
 
   /**
    * Resolves a GuildChannelResolvable to a Channel object.
-   * @method resolve
-   * @memberof GuildChannelManager
-   * @instance
    * @param {GuildChannelResolvable} channel The GuildChannel resolvable to resolve
-   * @returns {?GuildChannel}
+   * @returns {?GuildChannel|ThreadChannel}
```
vladfrangu, line 62:
```suggestion
   * @returns {?(GuildChannel|ThreadChannel)}
```

src/managers/GuildChannelManager.js
```diff
@@ -50,27 +51,30 @@ class GuildChannelManager extends BaseManager {
...
    */
+  resolve(channel) {
+    if (channel instanceof ThreadChannel) return super.resolve(channel.id);
```
vladfrangu, line 65: Is this...needed?

vladfrangu (changes requested):

src/managers/GuildChannelManager.js
ckohen: Yes, Threads don't extend GuildChannel because they are so different, but they do exist in the guild channel cache, so we resolve by id.

vladfrangu (approved):

iCrawl (approved):

src/util/Constants.js
kyranet: Upstream PR landed! 🎉

SpaceEEC (approved):

PR #5967, opened 2021-06-29
ci: add UNIX timestamp to version number

ImRodry:
Changed the git sha to only include the first 7 characters and added a UNIX timestamp in seconds. This should prevent issues with `npm update` and dependabot being stuck on older versions.

### DIFF
--- a/.github/workflows/publish-dev.yml
+++ b/.github/workflows/publish-dev.yml
@@ -22,7 +22,7 @@ jobs:
 
       - name: Publish
         run: |
-          npm version --git-tag-version=false $(jq --raw-output '.version' package.json).$(git rev-parse --short=7 --verify HEAD).$(date +%s)
+          npm version --git-tag-version=false $(jq --raw-output '.version' package.json).$(git rev-parse --short HEAD).$(date +%s)
           npm publish --tag dev || true
         env:
           NODE_AUTH_TOKEN: ${{ secrets.NPM_PUBLISH_TOKEN }}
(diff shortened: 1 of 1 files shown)


.github/workflows/publish-dev.yml
```diff
@@ -22,7 +22,7 @@ jobs:
 
       - name: Publish
         run: |
-          npm version --git-tag-version=false $(jq --raw-output '.version' package.json).$(git rev-parse --verify HEAD)
+          npm version --git-tag-version=false $(jq --raw-output '.version' package.json).$(git rev-parse --short=7 --verify HEAD).$(date +%s)
```
vaporox, line 25:
```suggestion
          npm version --git-tag-version=false $(jq --raw-output '.version' package.json).$(git rev-parse --short HEAD).$(date +%s)
```
This should be sufficient

iCrawl: I'd like to keep `--verify` and the explicit 7 doesn't hurt no one.

ImRodry: Crawl asked for SHA 7 and removing =7 gives SHA 8. Also why remove --verify?

vaporox: Might be possible that newer git versions have increased the default, in this case that should be fine. Taken from `man git-rev-parse`, `--short` implicitly enables `--verify` so there's no need to add it too.

ImRodry:
Forgot to include a screenshot but this is what it will look like (with the change above)
image

iCrawl (approved):

PR #5971, opened 2021-06-29
docs(ThreadManager): fix FetchedThreads typedef and startMessage

iShibi:

This PR correctly defines the `FetchedThreads` typedef.

It also changes the description for `ThreadCreateOptions#startMessage`. Thread created without a message does default to private if no type was provided but we have `type` set to `public_thread` as default in the lib. So, it is not going to be a private thread if no `startMessage` was provided.

### DIFF
--- a/src/managers/ThreadManager.js
+++ b/src/managers/ThreadManager.js
@@ -74,8 +74,7 @@ class ThreadManager extends BaseManager {
    * @typedef {Object} ThreadCreateOptions
    * @property {string} name The name of the new Thread
    * @property {ThreadAutoArchiveDuration} autoArchiveDuration How long before the thread is automatically archived
-   * @property {MessageResolvable} [startMessage] The message to start a public or news thread from,
-   * creates a private thread if not provided
+   * @property {MessageResolvable} [startMessage] The message to start a thread from
    * @property {ThreadChannelType|number} [type='public_thread'] The type of thread to create
    * <warn>When creating threads in a `news` channel this is ignored and is always `news_thread`</warn>
    * @param {string} [reason] Reason for creating the thread
@@ -186,7 +185,7 @@ class ThreadManager extends BaseManager {
 
   /**
    * The data returned from a thread fetch that returns multiple threads.
-   * @typedef {FetchedThreads}
+   * @typedef {Object} FetchedThreads
    * @property {Collection<Snowflake, ThreadChannel>} threads The threads fetched, with any members returned
    * @property {?boolean} hasMore Whether there are potentially additional threads that require a subsequent call
    */


iCrawl (approved):

src/managers/ThreadManager.js
```diff
@@ -74,8 +74,7 @@ class ThreadManager extends BaseManager {
    * @typedef {Object} ThreadCreateOptions
    * @property {string} name The name of the new Thread
    * @property {ThreadAutoArchiveDuration} autoArchiveDuration How long before the thread is automatically archived
-   * @property {MessageResolvable} [startMessage] The message to start a public or news thread from,
-   * creates a private thread if not provided
+   * @property {MessageResolvable} [startMessage] The message to start a thread from
    * @property {ThreadChannelType|number} [type='public_thread'] The type of thread to create
```
iShibi, line 78: Since, we don't literally provide the method with `public_thread` as default for `type` param, shouldn't this be left as `[type]` and the fact that `type` gets default to `public_thread` internally should be written in the description of this param instead? There are other instances of this behaviour in the lib and this is exactly what we have done for them.

iCrawl: It should probably be kept because `v9` of the API still defaults to a private thread, so we diverge from this.

ckohen: Technically the default is `publiic_thread` | `news_thread` depending on the parent channel type, but it never lets you create any other type of thread in a news channel (basically, passing `news_thread` is unnecessary). It *could* inline default this in the code, but it was unnecessary.

iShibi: Yeah, but those defaults are in the API, by internally I meant in this `ThreadManager#create` method. We don't pass any value to `type` param. It defaults to `news_thread` if it was news channel, `public_thread` if it wasn't and on the API side to `private_thread` in v9 (might change as crawl said). That's why I was suggesting to have `[type]` instead of `[type='public_thread`] as that's what is true according to the definition of this method. And we can describe all this weird defaulting behaviour in the description of the `type` param instead.

ckohen: None of these defaults are API side, the only default that exists is private. that will api error in news channels (and when the guild doesn't have the necessary feature). The warning covers news threads. You could quite literally actually default this as documented and nothing would change.

iShibi: Yeah my bad about calling all of them API defaults in the starting, I meant only the private one. But still the question is why are we documenting it as `[type='public_thread']` when the default for `type` happens inside the method and not in the parameters. We have discussed this earlier in the server and the stance was to not document a default like this where the default is set inside the method. For example the methods for getting icon url, they have defaults that are set internally which is why the fields like `dynamic` and `format` in `ImageURLOptions` don't have any default value documented but the internal defaulting behaviour of being getting set to `false` & `webp` respectively is stated in their descriptions.

ckohen: If you have a better way to handle the double default, by all means!

vladfrangu (approved):

kyranet (approved):

PR #5979, opened 2021-06-30
types(ApplicationCommandPermissionsManager): fix types

ckohen:

Fixes a subtle typings issue with the perms manager. There does remain the issue that set{ fullPermissions }) is invalid when used for an ApplicationCommands permissions manager, but I have no idea how to fix that.

### DIFF
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -2405,7 +2405,7 @@ declare module 'discord.js' {
     ): Promise<ApplicationCommandType>;
     public fetch(
       id: Snowflake,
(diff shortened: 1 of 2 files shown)


iCrawl (approved):

kyranet (approved):

vladfrangu (approved):

SpaceEEC (approved):
Might be worth to add some "tests" for this in the `index.ts` file, since I fear something will sooner or later be broken here. 😅

ckohen:
Okay, "tests" added. Kinda feels like we may want separate files for these cause thats a lot of extra lines, so if thats desired I'll move them.

PR #5981, opened 2021-06-30
Added more collection methods to Collection interface

TetroVolt:
These are common methods which are used in a regular JS project. It would be nice to also have them defined in the typescript typings for those of use using discordjs with typescript. 

I am aware the source types are really defined here: https://github.com/discordjs/collection/blob/main/src/index.ts


kyranet:
I'm not aware of what the problem is, but types are inherited, so we don't need to define the sub-class's methods nor fields.

PR #5983, opened 2021-06-30
feat(RoleManager): added `edit` method, alias `Role#edit`

### LINKED ISSUE #5634: Editing a role can result in a Missing Permissions error even when just wanting to edit simple values (Like name or colour)

When editing a role, such as directly through `Role#edit()` or with a method like `Role#setHoist()`, since they use the same `edit` method, if you want to edit something simple about that role, such as the name, colour, etc, the bot may end up returning a `Missing Permissions` error if the bot doesn't have *at least* the same permissions as the role it's trying to edit, since the `edit` method is sending the role's own permissions as part of the edit with no way to overwrite that functionality as far as I can tell.

- Solution 1. `edit` only sends the values that are necessary relative to the edit the developer (or a method using `edit`) wants to make.
- Solution 2. `edit` allows the developer to overwrite the permissions that the method internally sets, such as setting `permissions` to `null` to have it not specify anything for that field.
- Solution 3. `edit` sends everything but the `permissions` field unless specified when calling it, since the other fields don't need any special permission requirements to be edited.

- Alt. 1. If the functionality is used in a command (as it is in my case), it could be specified to the user that they need to give a role the bot has all the same permissions as the role they are trying to edit (Or maybe give the bot the same role whilst they are trying to edit it). These both seem like they would create more work for the user unnecessarily, and I have had one user choose not to use the relevant command because of the permissions requirement.
- Alt. 2. A request *could* be made directly to the API, but I'm sure this would be unrecommended and would likely cause issues for the library functionality too if not done correctly.

**Additional context**
N/A

kyranet:

Fixes #5634

Also implements `RoleManager#edit` to be consistent with many other managers.

- [x] Updated typings.
- [x] Tested `RoleManager#edit`.
- [x] Tested `Role#edit`.

### DIFF
--- a/src/managers/RoleManager.js
+++ b/src/managers/RoleManager.js
@@ -184,7 +184,7 @@ class RoleManager extends BaseManager {
       mentionable: data.mentionable,
     };
 
-    const d = this.client.api.guilds[this.guild.id].roles[id].patch({ data: _data, reason });
+    const d = this.client.api.guilds(this.guild.id).roles(id).patch({ data: _data, reason });
 
     const clone = this.cache.get(id)?._clone();
     clone?._patch(d);
--- a/src/managers/RoleManager.js
+++ b/src/managers/RoleManager.js
@@ -1,10 +1,11 @@
 'use strict';
 
 const BaseManager = require('./BaseManager');
+const { TypeError } = require('../errors');
 const Role = require('../structures/Role');
 const Collection = require('../util/Collection');
 const Permissions = require('../util/Permissions');
-const { resolveColor } = require('../util/Util');
+const { resolveColor, setPosition } = require('../util/Util');
 
 /**
  * Manages API methods for roles and stores their cache.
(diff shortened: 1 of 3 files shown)


src/managers/RoleManager.js
```diff
@@ -143,6 +144,53 @@ class RoleManager extends BaseManager {
...
+
+    const _data = {
+      name: data.name,
+      color: typeof data.color === 'undefined' ? undefined : resolveColor(data.color),
+      hoist: data.hoist,
+      permissions: typeof data.permissions === 'undefined' ? undefined : new Permissions(data.permissions),
+      mentionable: data.mentionable,
+    };
+
+    const d = this.client.api.guilds[this.guild.id].roles[id].patch({ data: _data, reason });
```
monbrey, line 187:
```suggestion
    const d = this.client.api.guilds(this.guild.id).roles(id).patch({ data: _data, reason });
```
Small nit but iirc it was decided `()` was preferred over `[]` at some point so this is for consistency.

iCrawl (approved):

src/managers/RoleManager.js
```diff
@@ -143,6 +144,53 @@ class RoleManager extends BaseManager {
...
+   *   .then(updated => console.log(`Edited role name to ${updated.name}`))
+   *   .catch(console.error);
+   */
+  async edit(role, data, reason) {
+    const id = this.resolveID(role);
+    if (!id) throw new TypeError('INVALID_TYPE', 'role', 'RoleResolvable');
+
+    if (typeof data.position === 'number') {
+      const updatedRoles = await setPosition(
+        this,
```
SpaceEEC, line 165: The first parameter here should be a role instance, otherwise this will just patch the roles without changing the position of the role in question.

SpaceEEC (changes requested):

src/managers/RoleManager.js
kyranet: Nice catch! I'm unsure why this worked just fine when I tested it. Fixed in b1a5b52.

SpaceEEC (approved):

vladfrangu (approved):

PR #5988, opened 2021-07-01
types: few missing thread channels

monbrey:
A few fixes for places that ThreadChannel should be a valid type.


ckohen:
`MessageComponentInteractionCollector` also needs updating

kyranet (approved):

iCrawl (approved):

PR #5989, opened 2021-07-01
types(Interactions): fix function overload types

thehackerboi69github:
Fixes issues with `Interaction#reply` and `MessageComponentInteraction#update`, due to the overload types the return type would default to `Promise<Message | APIMessage>` if only a string was specified as the content.


monbrey:
You make a valid point, but I don't think this is correct. `string | MessagePayload` should be removed from the overload that returns a `Promise<Message | APIMessage>`

kyranet (approved):

SpaceEEC (approved):

iCrawl (approved):

PR #5991, opened 2021-07-01
refactor: improve typings and docs related to threads

iShibi:

1) This PR improves the typings by:

- Restricting values for `type` in `ThreadManager#create` depending on what type of channel the manager belongs to. This will provide better and valid autocompletion
- Adding interfaces where they were needed to make the typings look clean and tidy
- Fixing a TS bug in `ThreadManager#fetch` where it will never suggest `force` field for `cacheOptions` (the fact that it gets ignored in some specific circumstances has been documented too)

2) This PR also improves the documentation for threads by:
- Adding typedef for `ThreadEditData`
- Adding more detailed information where it was required
- Doing a little bit of sentence restructuring here and there
- Addressing caveats

### DIFF
--- a/src/structures/ThreadChannel.js
+++ b/src/structures/ThreadChannel.js
@@ -26,7 +26,7 @@ class ThreadChannel extends Channel {
     this.guild = guild;
 
     /**
-     * A manager of the messages set to this thread
+     * A manager of the messages sent to this thread
      * @type {MessageManager}
      */
     this.messages = new MessageManager(this);
@@ -52,7 +52,7 @@ class ThreadChannel extends Channel {
 
     if ('parent_id' in data) {
       /**
-       * The ID of the parent channel to this thread
+       * The ID of the parent channel of this thread
        * @type {Snowflake}
        */
       this.parentID = data.parent_id;
@@ -66,20 +66,21 @@ class ThreadChannel extends Channel {
       this.locked = data.thread_metadata.locked ?? false;
 
       /**
-       * Whether the thread is active (false) or archived (true)
+       * The id of the member who created this thread
        * @type {boolean}
        */
       this.archived = data.thread_metadata.archived;
 
       /**
-       * How long in minutes after recent activity before the thread is automatically archived
+       * The amount of time (in minutes) after which the thread will automatically archive in case of no recent activity
        * @type {number}
        */
       this.autoArchiveDuration = data.thread_metadata.auto_archive_duration;
 
       /**
        * The timestamp when the thread's archive status was last changed
-       * <info>If the thread was never archived or unarchived, this is set when it's created</info>
+       * <info>If the thread was never archived or unarchived, this is the timestamp at which the thread was
+       * created</info>
        * @type {number}
        */
       this.archiveTimestamp = new Date(data.thread_metadata.archive_timestamp).getTime();
@@ -111,7 +112,7 @@ class ThreadChannel extends Channel {
 
     if ('rate_limit_per_user' in data) {
       /**
-       * The ratelimit per user for this thread in seconds
+       * The ratelimit per user for this thread (in seconds)
        * @type {number}
        */
       this.rateLimitPerUser = data.rate_limit_per_user ?? 0;
@@ -120,8 +121,8 @@ class ThreadChannel extends Channel {
     if ('message_count' in data) {
       /**
        * The approximate count of messages in this thread
-       * <info>This value will not count above 50 even when there are more than 50 messages
-       * If you need an approximate value higher than this, use ThreadChannel#messages.cache.size</info>
+       * <info>This stops counting at 50. If you need an approximate value higher than that, use
+       * `ThreadChannel#messages.cache.size`</info>
        * @type {number}
        */
       this.messageCount = data.message_count;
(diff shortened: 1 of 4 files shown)


kyranet:
You probably can't believe this, but as soon as the PR was created, a rebase was needed 😄

typings/index.d.ts
```diff
@@ -2661,21 +2661,12 @@ declare module 'discord.js' {
...
-      reason?: string;
-    }): Promise<ThreadChannel>;
-    public fetch(options: ThreadChannelResolvable, cacheOptions?: BaseFetchOptions): Promise<ThreadChannel | null>;
-    public fetch(
-      options?: { archived?: FetchArchivedThreadOptions; active?: boolean },
-      cacheOptions?: { cache?: boolean },
-    ): Promise<FetchedThreads>;
+    public create(options: ThreadCreateOptions<AllowedThreadType>): Promise<ThreadChannel>;
+    public fetch(options?: ThreadChannelResolvable, cacheOptions?: BaseFetchOptions): Promise<ThreadChannel | null>;
+    public fetch(options?: FetchThreadsOptions, cacheOptions?: BaseFetchOptions): Promise<FetchedThreads>;
```
ckohen, line 2669: I don't understand why this was changed, force is never a valid option for this overload, it should be suggested. For the previous overload it is valid.

iShibi: It was merging those two and hence wasn't suggesting the `force` field even when `options` is a `ThreadChannelResolvable`. It isn't `invalid` tho, it is more of a **"redundant"** param when options is `FetchThreadsOptions` as it gets ignored.

ckohen: huh, cause `ThreadChannelResolvable` somehow equates to `FetchThreadsOptions` ??? ts you okay? whatever, if it works it works.

iCrawl (approved):

kyranet (approved):

src/structures/ThreadChannel.js
```diff
@@ -273,12 +282,12 @@ class ThreadChannel extends Channel {
...
 
   /**
-   * Sets a new name for the thread.
+   * Sets a new name for this thread.
    * @param {string} name The new name for the thread
    * @param {string} [reason] Reason for changing the thread's name
    * @returns {Promise<ThreadChannel>}
    * @example
-   * // Set a new thread name
+   * // Change thread's name
```
SpaceEEC, line 290: I think this would be better with a `the`.
```suggestion
   * // Change the thread's name
```

typings/index.d.ts
```diff
@@ -2650,21 +2650,12 @@ declare module 'discord.js' {
...
-      type?: ThreadChannelType | number;
-      reason?: string;
-    }): Promise<ThreadChannel>;
-    public fetch(options: ThreadChannelResolvable, cacheOptions?: BaseFetchOptions): Promise<ThreadChannel | null>;
-    public fetch(
-      options?: { archived?: FetchArchivedThreadOptions; active?: boolean },
-      cacheOptions?: { cache?: boolean },
-    ): Promise<FetchedThreads>;
+    public create(options: ThreadCreateOptions<AllowedThreadType>): Promise<ThreadChannel>;
+    public fetch(options?: ThreadChannelResolvable, cacheOptions?: BaseFetchOptions): Promise<ThreadChannel | null>;
```
SpaceEEC, line 2657: Shouldn't this parameter be required?
```suggestion
    public fetch(options: ThreadChannelResolvable, cacheOptions?: BaseFetchOptions): Promise<ThreadChannel | null>;
```

`channel.threads.fetch();` seems to be calling `channel.threads.fetchActive(true)` internally, which would match the second overload.

typings/index.d.ts
SpaceEEC: With the above change, this could be changed back to `{ cache?: boolean }`.
At least in my testing it seems to work properly with suggesting `cache` and rejecting `{ force: true }`.

SpaceEEC (changes requested):

vladfrangu (approved):
Other than space's comments, LGTM!

typings/index.d.ts
iShibi: It rejecting `force` for `FetchThreadsOptions` wasn't the issue, the problem is that it doesn't suggest `force` for `ThreadChannelResolvable` too which makes it **"seems"** like `force` isn't a valid param. It does allow it to compile so it's not that big of an issue, just a minor inconvenience. Making both of them `BaseFetchOptions` solves the issue and since `force` does nothing for the second overload, it's not a problem even if the user passes it.

SpaceEEC: I'm unable to reproduce that behavior, I'm seeing the suggestions just fine.

iShibi: Weird, doesn't seem to work for me. Using `{ cache?: boolean }` for second overload makes it **not** suggest `force` for the first one too:
Code_bNJDLuQZwJ

iShibi: I just realized that we can solve this issue by using Conditional Types. We can merge the overloads and let conditional typings do its work:
Code_TyyPLLLIsa
I tested it and it works nicely. It suggests both `cache` & `force` when `options` is `ThreadChannelResolvable` and only `cache` when it is `FetchThreadsOptions`.

SpaceEEC: I don't know why it's not showing up for you 🤷 
image

iCrawl: I don't think the `.type` check actually guards, maybe thats why.

iShibi: > 
> 
> I don't think the `.type` check actually guards, maybe thats why.

It doesn't work for me even when I write the exact same code that Space showed above.
Code_CL55hrVv5M

iShibi: Seems like it's just me 🙇‍♂️. Updated the second overload to only suggest `cache` for `cacheOptions`.

SpaceEEC (approved):

PR #5992, opened 2021-07-01
ci: deprecate old dev releases on publish

ImRodry:
This PR adds a step to the publish dev CI that automatically deprecates all dev versions before publishing a new one so that you don't have to manually do that yourselves every 12 hours. Here are the npm docs on [npm-deprecate](https://docs.npmjs.com/cli/v7/commands/npm-deprecate) and you can test the version range [here](https://semver.npmjs.com/)

### DIFF
--- a/.github/workflows/publish-dev.yml
+++ b/.github/workflows/publish-dev.yml
@@ -22,7 +22,7 @@ jobs:
         run: npm ci --ignore-scripts --no-optional
 
       - name: Deprecate old versions
-        run: npm deprecate discord.js@"~13.0.0-dev" "No longer supported" || true
+        run: npm deprecate discord.js@"~13.0.0-dev" "no longer supported" || true
         env:
           NODE_AUTH_TOKEN: ${{ secrets.NPM_PUBLISH_TOKEN }}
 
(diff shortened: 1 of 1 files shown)


.github/workflows/publish-dev.yml
```diff
@@ -21,6 +21,11 @@ jobs:
         # Remove --no-optional after Node 16 upgrade
         run: npm ci --ignore-scripts --no-optional
 
+      - name: Deprecate old versions
+        run: npm deprecate discord.js@"~13.0.0-dev" "No longer supported" || true
```
vladfrangu, line 25: Can't something like this be done, so it auto deprecates any valid dev version from the package.json?

```suggestion
        run: npm deprecate discord.js@"~$(jq --raw-output '.version' package.json)" "No longer supported" || true
```

vladfrangu (dismissed):

.github/workflows/publish-dev.yml
ImRodry: I thought this wasn’t gonna be used after v13 but if it will then yeah I think that can work

vladfrangu: Oh, gotcha

iCrawl, line 25:
```suggestion
        run: npm deprecate discord.js@"~13.0.0-dev" "no longer supported" || true
```

iCrawl: For now, yes.

iCrawl (approved):

PR #5993, opened 2021-07-01
types(VoiceState): fix optional params

PraveshKunwar:

https://discord.js.org/#/docs/main/master/class/VoiceState

1. VoiceState.selfVideo can be null.
2. VoiceState.setChannel(channel, reason): Channel can be optional, as default is none. (according to docs on master)

### DIFF
--- a/src/structures/VoiceState.js
+++ b/src/structures/VoiceState.js
@@ -149,7 +149,7 @@ class VoiceState extends Base {
 
   /**
    * Moves the member to a different channel, or disconnects them from the one they're in.
-   * @param {ChannelResolvable|null} [channel] Channel to move the member to, or `null` if you want to disconnect them
+   * @param {ChannelResolvable|null} channel Channel to move the member to, or `null` if you want to disconnect them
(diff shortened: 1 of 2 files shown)


SpaceEEC (dismissed):
I'd say that the docs should be updated to mark the channel parameter of `VoiceState#setChannel` to not be optional instead, as omitting it results in no update being made.

PraveshKunwar:
Yeah I was a bit confused on why the channel param was set to optional

ImRodry:
@PraveshKunwar you should update the JSDoc to indicate it’s not optional

PraveshKunwar:
> @PraveshKunwar you should update the JSDoc to indicate it’s not optional
> 
> Also, according to Discord’s docs, selfVideo will never be null so maybe that should be updated?

ok I'll take a further look.

iCrawl (approved):

kyranet (approved):

vladfrangu (approved):

PR #5999, opened 2021-07-01
feat(InteractionCollector): reworked to be more generic

monbrey:
Accidentally pushed a delete to remote on the old PR #5936 so.... yeaaaaaah.

Another half-baked idea that comes out of late night conversations and early morning coding.

This refactors the `MessageComponentInteractionCollector` into a generic `InteractionCollector` that can be used to collect any sort of interaction (yes, even Slash Commands).

It also removes the need for it to be bound to any particular message or channel, though it still can be should you want to do that. These are still separate options to the filter as it allows the Collector to also listen to the relevant DELETE events and cancel itself accordingly. It will use these based on the most specific option first, so Message > Channel > Guild.

Draft for now - I'd still like reviews and opinions from other people. I intend to add some checking to the constructor to account for options that don't make sense together, e.g. setting a `componentType` option if `interactionType` is set for slash commands.

**Status and versioning classification**


src/structures/InteractionCollector.js
```diff
@@ -3,40 +3,69 @@
...
+    this.message = options.message ?? null;
 
     /**
-     * The source channel from which to collect message component interactions
-     * @type {TextChannel|DMChannel|NewsChannel}
+     * The channel from which to collect interactions, if provided
+     * @type {?TextChannel|DMChannel|NewsChannel}
      */
-    this.channel = this.message?.channel ?? source;
+    this.channel = this.message ? this.message.channel : options.channel ?? null;
```
kyranet, line 44:
```suggestion
    this.channel = this.message?.channel ?? options.channel ?? null;
```

src/structures/InteractionCollector.js
```diff
@@ -3,40 +3,69 @@
...
+
+    /**
+     * The guild from which to collect interactions, if provided
+     * @type {?Guild}
+     */
+    this.guild = this.channel && this.channel.guild ? this.channel.guild : options.guild ?? null;
```
kyranet, line 50:
```suggestion
    this.guild = this.channel?.guild ?? options.guild ?? null;
```

src/structures/Message.js
```diff
@@ -473,9 +486,15 @@ class Message extends Base {
...
    *   .catch(console.error);
    */
-  awaitMessageComponentInteraction(options = {}) {
+  awaitMessageComponent(options = {}) {
+    const _options = {
+      ...options,
+      interactionType: InteractionTypes.MESSAGE_COMPONENT,
+      max: 1,
+      message: this,
+    };
```
kyranet, line 495: Because the other properties are always overridden in `createMessageComponentCollector`, we don't need to duplicate them again.
```suggestion
    const _options = { ...options, max: 1 };
```

You can also inline it like it was before.

src/structures/interfaces/TextBasedChannel.js
```diff
@@ -329,9 +334,14 @@ class TextBasedChannel {
    *   .then(interaction => console.log(`${interaction.customID} was clicked!`))
    *   .catch(console.error);
    */
-  awaitMessageComponentInteraction(options = {}) {
+  awaitMessageComponent(options = {}) {
+    const _options = {
+      ...options,
+      channel: this,
+      interactionType: InteractionTypes.MESSAGE_COMPONENT,
+    };
```
kyranet, line 342: 🧹 This can go and pass `options` directly, as the two properties are assigned in it.

kyranet (changes requested):

vladfrangu (approved):
Other than kyra's comments, LGTM

src/structures/interfaces/TextBasedChannel.js
monbrey, line 342:
```suggestion
    const _options = { ...options, max: 1 };
```

monbrey: It's actually missing `max: 1` but I wouldnt have noticed without this comment, so thanks.

kyranet (approved):

iCrawl (approved):

src/structures/InteractionCollector.js
```diff
@@ -3,40 +3,69 @@
...
+     * The the type of interaction to collect
+     * @type {?InteractionType}
+     */
+    this.interactionType =
+      typeof options.interactionType === 'number'
+        ? InteractionTypes[options.interactionType]
+        : options.interactionType ?? null;
+
+    /**
+     * The the type of compoonent to collect
```
SpaceEEC, line 62:
```suggestion
     * The the type of component to collect
```

SpaceEEC (changes requested):
Otherwise LGTM

SpaceEEC (approved):

iCrawl:
This needs a rebase.

PR #6003, opened 2021-07-02
types(*): create mention types

DTrombett:
This PR adds more specific typings for mentions like the example below:
```ts
type UserMention = `<@${Snowflake}>`;
```

### DIFF
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -439,7 +439,7 @@ declare module 'discord.js' {
     public fetch(force?: boolean): Promise<Channel>;
     public isText(): this is TextChannel | DMChannel | NewsChannel | ThreadChannel;
     public isThread(): this is ThreadChannel;
-    public toString(): string;
+    public toString(): ChannelMention;
   }
 
   export class Client extends BaseClient {
@@ -1103,7 +1103,7 @@ declare module 'discord.js' {
     public permissionsIn(channel: GuildChannelResolvable): Readonly<Permissions>;
     public setNickname(nickname: string | null, reason?: string): Promise<GuildMember>;
     public toJSON(): unknown;
-    public toString(): string;
+    public toString(): MemberMention;
     public valueOf(): string;
   }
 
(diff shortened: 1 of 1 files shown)


typings/index.d.ts
```diff
@@ -3028,6 +3028,11 @@ declare module 'discord.js' {
     around?: Snowflake;
   }
 
+  type ChannelMention = `<#${Snowflake}>`;
+  type UserMention = `<@${Snowflake}>`;
+  type MemberMention = UserMention | `<@!${Snowflake}>`;
+  type RoleMention = '@everyone' | `<@&${Snowflake}>`;
```
Dev-CasperTheGhost, line 3034:
```suggestion
  type RoleMention = '@everyone' | '@here' | `<@&${Snowflake}>`;
```

DTrombett: `@here` is not a role, while `@everyone` is the role that everyone has.
image

Dev-CasperTheGhost: Ah yeah, true.

DTrombett:
Sorry for the empty commit but there was a problem with npm.

typings/index.d.ts
advaith1: the `@everyone` role is actually unrelated to the `@everyone` ping; actually mentioning the `@everyone` role outputs `@@everyone` and doesn't ping anyone.

would probably be best to remove that if/return tbh

DTrombett: Well maybe, but I didn't change anything on that method. This is just to add types and `Role.toString()` can return `@everyone` too so I added it...

ImRodry, line 3034: I thought these were meant to be sorted alphabetically but at least leave an empty line between them

DTrombett: I can sort them alphabetically but I didn't add a line because I saw some other types without it... I'll check this later.

ImRodry: Then those should probably have it cuz there's an empty line between almost everything in this file. And yeah you should sort them alphabetically within the typedefs region

kyranet (approved):

iCrawl (approved):

vladfrangu (approved):

PR #3696, opened 2020-01-15
feat: remove datastores and implement Managers

RDambrosio016:
**Please describe the changes this PR makes and why it should be merged:**
This PR completely removes DataStores, and it implements Managers. The reasons for this change are highlighted in #3485, and this pr mostly implements it.

## Testing

Due to the inherent breaking nature of this PR, it is not completely testable by one person. if you would like to aid in testing this pr, simply install the branch with:
 `npm i git://github.com/RDambrosio016/discord.js.git#remove-datastores`. If you encounter any weird behavior or errors, please either comment below, or ping me on the discord server (`@BorgerKing#1856`).

The main focus of the PR is completely separating API and cache for data models.

This allows us to:

- Easily swap out API methods.
- Easily extend Managers.
- Easily swap out cache types.
- Separate API and cache methods so as not to have weird overwrites on various managers.

## Future plans

This pr easily allows us to give users control over caching strategies through dependency injection. Users will extend collection and provide it to a helper class similar to `Structures` which the managers will use to get the right collection. In addition to this, bans will be made into their own manager, this will centralize the methods so as not to have them in various managers and structures.

## Misc changes

- Managers of any kind are no longer iterable.
- Managers do not have a `Symbol.species` of `Collection` anymore.
- The `BaseManager` class is `abstract` unlike `DataStore`.
- Neither the `BaseManager` class, nor standalone managers extend `Collection`.
- Collection methods are no longer overwritten with API methods due to the separation of purposes.
- Multiple properties on stores have been documented in managers, e.g. `MessageManager#channel`.
- `MessageStore#remove()` has been renamed to `MessageManager#delete()`

## Managers!

Abstract:

`DataStore` => `BaseManager`

Extends `BaseManager`:

`GuildChannelStore` => `GuildChannelManager`
`GuildEmojiStore` => `GuildEmojiManager`
`GuildMemberStore` => `GuildMemberManager`
`MessageStore` => `MessageManager`
`PresenceStore` => `PresenceManager`
`ReactionStore` => `ReactionManager`
`GuildStore` => `GuildManager`
`ReactionUserStore` => `ReactionUserManager`
`UserStore` => `UserManager`  
`RoleStore` => `RoleManager`

Standalone (`cache` is a getter):

`GuildMemberRoleStore` => `GuildMemberRoleManager`
`GuildEmojiRoleStore` => `GuildEmojiRoleManager`

## todo

- ~~Rethink if Standalone stores should/could be reworked to extend `BaseManager`.~~
[...]

PR #6013, opened 2021-07-03
feat(Managers): add customizable caching for managers

1Computer1:

- All managers are actually managers now:
  - Hierarchy is `BaseManager :> DataManager :> CachedManager`.
  - `BaseManager` has no cache, `DataManager` has a cache getter and the resolve methods, `CachedManager` is the whole thing.

- Customizable cache types for `CachedManager`s:
  - Anything implementing the `Collection` interface can be used.
  - You can disable caching now, you're welcome.
  - Example:
```js
// Using a helper method that does reasonable stuff.
const client = new Client({
    MessageManager: 200, // This is default.
    PresenceManager: 0, // Add more class names here.
  }),
});

// Just do whatever.
const client = new Client({
  makeCache: (manager) => {
    if (manager.name === 'MessageManager') return new LimitedCollection(0);
    return new Coolection();
  }
});
```

- Breaking changes in summary:
  - The new managers' hierarchy and methods and constructors.
  - `ReactionUserManager`'s constructor, cus it was weird.
  - Moved options related stuff out of `Constants` into `Options`.
  - Removed `messageCacheMaxSize`.

- Things I wanna do but too much work/infeasible with current designs:
  - Caches with async methods. *Do this in the rewrite.*
  - More cache types, e.g. one that sweeps based on a predicate (replacing the message sweeping). *Need more design work.*
  - Rename the manager `add` to something better, how about `addToCache`? *Someone do this please.*


src/util/Options.js
```diff
@@ -0,0 +1,154 @@
...
+ * @property {string} [invite='https://discord.gg'] Base url of invites
+ * @property {string} [template='https://discord.new'] Base url of templates
+ * @property {Object} [headers] Additional headers to send for all API requests
+ */
+
+/**
+ * Contains various utilities for client options.
+ */
+class ClientOptionsUtil extends null {
+  constructor() {
```
ckohen, line 93: Unneeded when extending null

src/util/Options.js
```diff
@@ -0,0 +1,154 @@
...
+    throw new Error(`The ${this.constructor.name} class may not be instantiated.`);
+  }
+
+  /**
+   * The default client options.
+   * @returns {ClientOptions}
+   */
+  static createDefaultOptions() {
```
ckohen, line 101: IMO this should be a getter

src/managers/DataManager.js
```diff
@@ -0,0 +1,64 @@
...
+    Object.defineProperty(this, 'holds', { value: Structures.get(holds.name) ?? holds });
+  }
+
+  /**
+   * The cache of items for this manager.
+   * @type {Collection}
+   * @abstract
+   */
+  get cache() {
+    throw new Error(`Getter 'cache' not implemented for ${this.constructor.name}`);
```
ckohen, line 34: Use a djs custom error string (function type)

src/managers/DataManager.js
```diff
@@ -0,0 +1,64 @@
...
+  }
+
+  /**
+   * Resolves a data entry to a data Object.
+   * @param {string|Object} idOrInstance The id or instance of something in this Manager
+   * @returns {?Object} An instance from this Manager
+   */
+  resolve(idOrInstance) {
+    if (idOrInstance instanceof this.holds) return idOrInstance;
+    if (typeof idOrInstance === 'string') return this.cache.get(idOrInstance) ?? null;
```
ckohen, line 44: this.cache isn't always available, I think resolve needs to go on CachedManager, but resolveID could stay here.

src/managers/DataManager.js
```diff
@@ -0,0 +1,64 @@
...
+   * @returns {?Snowflake}
+   */
+  resolveID(idOrInstance) {
+    if (idOrInstance instanceof this.holds) return idOrInstance.id;
+    if (typeof idOrInstance === 'string') return idOrInstance;
+    return null;
+  }
+
+  valueOf() {
+    return this.cache;
```
ckohen, line 60: this.cache doesn't exist on this stucture, are you sure you want it to error when calling valueOf()?

src/managers/DataManager.js
```diff
@@ -0,0 +1,64 @@
...
+ */
+class DataManager extends BaseManager {
+  constructor(client, holds) {
+    super(client);
+
+    if (!Structures) Structures = require('../util/Structures');
+
+    /**
+     * The data structure belonging to this manager.
+     * @name BaseManager#holds
```
ckohen, line 20: needs updating

src/managers/DataManager.js
```diff
@@ -0,0 +1,64 @@
+'use strict';
+
+const BaseManager = require('./BaseManager');
+
+let Structures;
+
+/**
+ * Manager that holds data structures.
```
ckohen, line 8: This feels misleading, it doesn't necessarily hold these data structures, they seem to be anything that accesses data held in a different cache, presenting it in a different (limited) way.

test/tester2000.js
```diff
@@ -0,0 +1,57 @@
...
+    let res;
+    try {
+      res = eval(message.content);
+      if (typeof res !== 'string') res = require('util').inspect(res);
+    } catch (err) {
+      // eslint-disable-next-line no-console
+      console.error(err.stack);
+      res = err.message;
+    }
+    message.channel.send(res, { code: 'js' });
```
ckohen, line 36: This no longer works.

ckohen (changes requested):

src/managers/DataManager.js
1Computer1: It's an abstract class; it on the implementor to make it exist.

src/managers/DataManager.js
1Computer1: ^

test/tester2000.js
1Computer1: Bleh, tell whoever made tester1000 about it.

src/managers/DataManager.js
ckohen: Apparently I can't read 🤦

src/managers/DataManager.js
1Computer1: Docs could be clearer yea, but this isn't just for those weird ones. `CachedManager` extends this, too.

ckohen: But...that has its own description (also that says it "holds a cache" which...feels wrong too.

src/util/Options.js
1Computer1: Nah, you're not "getting" any data. It just is what it is.

src/managers/DataManager.js
1Computer1: Does this sound better?

ckohen: yeah, if people actually care about what these managers do (since each manager that extends them has its own description) they should be able to understand that.

src/util/Options.js
ckohen: I'd argue options.default is better than options.createDefaultOptions()

1Computer1: I believe getters should be used for properties, not glorified lazy evaluation. The method call makes that part obvious.

ckohen: Is this not just a glorified property? Heck you could even store this in a static property if you wanted. There isn't any evaluation going on in this function, only within the returned object.

1Computer1: A property of what? And there is evaluation, in particular the requires are nested inside to avoid a circular dependency,

ckohen: Static properties exist, and that evaluation is done in a function that happens to be called here, which is no different from this, which you are clearly fine witht:
```js
Object.defineProperty(this, '_cache', { value: this.client.options.makeCache(this.constructor, this.holds) });
```

1Computer1: No, look at the arrows. The methods return a function and that gets evaluated there. The requires are not.

src/managers/ReactionUserManager.js
```diff
@@ -1,16 +1,17 @@
...
  * Manages API methods for users who reacted to a reaction and stores their cache.
- * @extends {BaseManager}
+ * @extends {CachedManager}
  */
-class ReactionUserManager extends BaseManager {
-  constructor(client, iterable, reaction) {
-    super(client, iterable, { name: 'User' });
+class ReactionUserManager extends CachedManager {
+  constructor(reaction, iterable) {
+    super(reaction.client, { name: 'User' }, iterable);
```
kyranet, line 13: Why isn't `User` used?

src/managers/VoiceStateManager.js
```diff
@@ -1,14 +1,15 @@
...
 /**
  * Manages API methods for VoiceStates and stores their cache.
- * @extends {BaseManager}
+ * @extends {CachedManager}
  */
-class VoiceStateManager extends BaseManager {
+class VoiceStateManager extends CachedManager {
   constructor(guild, iterable) {
-    super(guild.client, iterable, { name: 'VoiceState' });
+    super(guild.client, { name: 'VoiceState' }, iterable);
```
kyranet, line 11: Same here

src/client/Client.js
```diff
@@ -468,8 +470,8 @@ class Client extends BaseClient {
       throw new TypeError('CLIENT_INVALID_OPTION', 'shards', "'auto', a number or array of numbers");
     }
     if (options.shards && !options.shards.length) throw new RangeError('CLIENT_INVALID_PROVIDED_SHARDS');
-    if (typeof options.messageCacheMaxSize !== 'number' || isNaN(options.messageCacheMaxSize)) {
-      throw new TypeError('CLIENT_INVALID_OPTION', 'messageCacheMaxSize', 'a number');
+    if (typeof options.makeCache !== 'function') {
```
kyranet, line 473: Otherwise this throws when it's unset:
```suggestion
    if (typeof options.makeCache !== 'undefined && typeof options.makeCache !== 'function') {
```

kyranet (dismissed):

src/client/Client.js
ckohen: It should throw when unset, it needs to be set (either via defaults or user) otherwise caching wouldn't work at all (probably hard crash)

kyranet: Wait, nevermind, the options are merged beforehand.

src/managers/ReactionUserManager.js
SpaceEEC: In both cases the changing PR does not explain why this change was made:
- https://github.com/discordjs/discord.js/pull/4616
- https://github.com/discordjs/discord.js/pull/3696

There doesn't seem to be a circular require issue and nothing seems to break when using the appropriate classes instead of this dummy object.

I suppose it's fine to supply the classes again then, although this is probably outside of the scope of this PR.

src/util/Options.js
```diff
@@ -0,0 +1,169 @@
...
+  static cacheEverything() {
+    const Collection = require('./Collection');
+    return () => new Collection();
+  }
+
+  /**
+   * Create a cache factory that always caches nothing.
+   * @returns {CacheFactory}
+   */
+  static cacheNothing() {
```
SpaceEEC, line 163: I'm not too sure how to feel about this.
This is going to "break" pretty much everything except dms.

Maybe at least add a warning about that?

SpaceEEC (dismissed):

src/util/Options.js
```diff
@@ -0,0 +1,169 @@
...
+  /**
+   * The default client options.
+   * @returns {ClientOptions}
+   */
+  static createDefault() {
+    return {
+      shardCount: 1,
+      makeCache: this.cacheWithLimits({ MessageManager: 200 }),
+      messageCacheLifetime: 0,
+      messageSweepInterval: 0,
```
vladfrangu, line 102: Not an issue with this PR, but in the future we should probably look into implementing a sweeper on all cache types?

src/util/Options.js
```diff
@@ -0,0 +1,169 @@
...
+        cdn: 'https://cdn.discordapp.com',
+        invite: 'https://discord.gg',
+        template: 'https://discord.new',
+      },
+    };
+  }
+
+  /**
+   * Create a cache factory using predefined limits.
+   * @param {Partial<Record<string, number>>} [limits={}] Limits for structures.
```
vladfrangu, line 134: I don't think `Partial` is supported by our docs site? CC @iCrawl

typings/index.d.ts
```diff
@@ -524,6 +524,13 @@ declare module 'discord.js' {
     public setUsername(username: string): Promise<this>;
   }
 
+  export class Options {
+    public static createDefaultOptions(): ClientOptions;
+    public static cacheWithLimits(limits?: Partial<Record<string, number>>): CacheFactory;
```
vladfrangu, line 529: Isn't Partial redundant?

vladfrangu (dismissed):

iCrawl:
This needs a rebase.

src/util/Options.js
iCrawl: It would just be a string and not link to anything.

src/managers/CachedManager.js
```diff
@@ -0,0 +1,43 @@
...
+  /**
+   * The cache of items for this manager.
+   * @type {Collection}
+   * @abstract
+   */
+  get cache() {
+    return this._cache;
+  }
+
+  add(data, cache = true, { id, extras = [] } = {}) {
```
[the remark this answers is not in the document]
iCrawl: IMHO yes, but not sure how comp feels about it.

src/util/Options.js
ckohen: It doesn't, I was adding the type in #5144 but that PR is going nowhere fast. (MessageEmbed already attempts to use it)

All I did was add this typedef
```js
 /**
   * A `Partial` object is a representation of any existing object.
   * This object contains between 0 and all of the original objects parameters.
   * This is true regardless of whether the parameters are optional in the base object.
   * @typedef {Object} Partial
   */
```

src/managers/CachedManager.js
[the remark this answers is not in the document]
kyranet: Users should never call Manager#add unless they know what they're doing, so yes, it should be _add.

src/util/Options.js
kyranet: I guess you can backport that part into a new PR, so we don't have to wait for it.

src/util/Options.js
1Computer1: If it breaks too much for reasonable use we can just remove it, let people figure it out.

src/managers/CachedManager.js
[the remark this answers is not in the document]
1Computer1: Yes, outside the scope of this PR. Better to do once the dust settles.

src/util/Options.js
1Computer1: Yep, we had some talks about more collection types and ways to compose them ergonomically. Future PR.

iCrawl:
This needs a rebase.

iCrawl (approved):

src/managers/CachedManager.js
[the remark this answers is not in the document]
ckohen: I'll do a followup to this PR then, maybe even before this gets merged (obviouslly will draft if so)

src/util/Options.js
ckohen: Sounds good, will do

kyranet (approved):

SpaceEEC (approved):
It might be worth a consideration to make `LimitedCollection` public again, otherwise you'd have to reimplement it yourself if `cacheWithLimits` doesn't cut it for somereason.

BONEE4:
eita bixo

PR #6014, opened 2021-07-03
feat(Formatters): added new URL utilities and docs

kyranet:

This PR brings two new formatters from https://github.com/discordjs/builders/pull/4:

- `hideLinkEmbed`
- `hyperlink`

And alongside them, it also adds docs so they're visible in the docs website. ~~, however, something I'm unsure of, is whether `@namespace` works.~~

~~cc: @iCrawl~~


src/util/Formatters.js
```diff
@@ -13,13 +15,134 @@ const {
...
-exports.italic = italic;
-exports.quote = quote;
-exports.strikethrough = strikethrough;
-exports.time = time;
-exports.TimestampStyles = TimestampStyles;
-exports.underscore = underscore;
+/**
+ * Contains various Discord-specific functions for formatting messages.
+ */
+
```
SpaceEEC, line 21: We don't seem to have an empty line between the class and its jsdoc anywhere else
```suggestion
```

SpaceEEC (changes requested):
Otherwise LGTM

src/util/Formatters.js
kyranet: That's an oversight when I was trying multiple things (see previous commits), thanks for catching this one 😅

SpaceEEC (approved):

vladfrangu (approved):

iCrawl (approved):

PR #6024, opened 2021-07-03
feat(Client): enforce passing scopes to generateInvite

monbrey:
Previously this method would automatically include the bot scope - it's now required that the user provide scopes in the options, including one of either bot or applications.commands.

### DIFF
--- a/src/client/Client.js
+++ b/src/client/Client.js
@@ -379,7 +379,7 @@ class Client extends BaseClient {
    * @property {PermissionResolvable} [permissions] Permissions to request
    * @property {GuildResolvable} [guild] Guild to preselect
    * @property {boolean} [disableGuildSelect] Whether to disable the guild selection
-   * @property {InviteScope[]} [additionalScopes] Whether any additional scopes should be requested
+   * @property {InviteScope[]} [scopes=['bot']] Scopes that should be requested
    */
 
   /**
@@ -393,6 +393,7 @@ class Client extends BaseClient {
    *     Permissions.FLAGS.MANAGE_GUILD,
    *     Permissions.FLAGS.MENTION_EVERYONE,
    *   ],
+   *   scope: ['bot'],
    * });
    * console.log(`Generated bot invite link: ${link}`);
    */
@@ -420,16 +421,18 @@ class Client extends BaseClient {
       query.set('guild_id', guildID);
     }
 
-    if (options.additionalScopes) {
-      const scopes = options.additionalScopes;
+    if (options.scopes) {
+      const scopes = options.scopes;
       if (!Array.isArray(scopes)) {
-        throw new TypeError('INVALID_TYPE', 'additionalScopes', 'Array of Invite Scopes', true);
+        throw new TypeError('INVALID_TYPE', 'scopes', 'Array of Invite Scopes', true);
       }
       const invalidScope = scopes.find(scope => !InviteScopes.includes(scope));
       if (invalidScope) {
-        throw new TypeError('INVALID_ELEMENT', 'Array', 'additionalScopes', invalidScope);
+        throw new TypeError('INVALID_ELEMENT', 'Array', 'scopes', invalidScope);
       }
-      query.set('scope', ['bot', ...scopes].join(' '));
+      query.set('scope', [...scopes].join(' '));
+    } else {
+      query.set('scope', 'bot');
     }
 
     return `${this.options.http.api}${this.api.oauth2.authorize}?${query}`;
--- a/src/client/Client.js
+++ b/src/client/Client.js
@@ -425,7 +425,7 @@ class Client extends BaseClient {
     if (invalidScope) {
       throw new TypeError('INVALID_ELEMENT', 'Array', 'scopes', invalidScope);
     }
-    query.set('scope', [...scopes].join(' '));
+    query.set('scope', scopes.join(' '));
 
     if (options.permissions && options.scopes) {
       const permissions = Permissions.resolve(options.permissions);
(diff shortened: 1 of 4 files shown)


iCrawl:
This needs a rebase.

src/client/Client.js
```diff
@@ -402,10 +408,26 @@ class Client extends BaseClient {
 
     const query = new URLSearchParams({
       client_id: this.application.id,
-      scope: 'bot',
     });
 
-    if (options.permissions) {
+    const scopes = options.scopes;
```
kyranet, line 413:
```suggestion
    const {scopes }= options;
```

src/client/Client.js
```diff
@@ -402,10 +408,26 @@ class Client extends BaseClient {
...
+    if (typeof scopes === 'undefined') {
+      throw new TypeError('INVITE_MISSING_SCOPES');
+    }
+    if (!Array.isArray(scopes)) {
+      throw new TypeError('INVALID_TYPE', 'scopes', 'Array of Invite Scopes', true);
+    }
+    const requiredScope = scopes.find(scope => scope === 'bot' || scope === 'applications.commands');
+    if (!requiredScope) {
```
kyranet, line 421:
```suggestion
    if (scopes.every(scope => scope !== 'bot' && scope !== 'applications.commands')) {
```

kyranet: I'd suggest moving those hardcoded scopes into an array or something, so we can extend them more easily in the future.

src/client/Client.js
```diff
@@ -402,10 +408,26 @@ class Client extends BaseClient {
...
+      throw new TypeError('INVITE_MISSING_SCOPES');
+    }
+    const invalidScope = scopes.find(scope => !InviteScopes.includes(scope));
+    if (invalidScope) {
```
kyranet, line 425:
```suggestion
    if (scopes.some(scope => !InviteScopes.includes(scope))) {
```

src/client/Client.js
```diff
@@ -402,10 +408,26 @@ class Client extends BaseClient {
...
+      throw new TypeError('INVALID_ELEMENT', 'Array', 'scopes', invalidScope);
+    }
+    query.set('scope', [...scopes].join(' '));
```
kyranet, line 428: Unless you're supporting enumerators, which would make it throw in the previous lines, this isn't necessary:
```suggestion
    query.set('scope', scopes.join(' '));
```

src/errors/Messages.js
```diff
@@ -128,6 +128,8 @@ const Messages = {
   INTERACTION_EPHEMERAL_REPLIED: 'Ephemeral responses cannot be fetched or deleted.',
   INTERACTION_FETCH_EPHEMERAL: 'Ephemeral responses cannot be fetched.',
 
+  INVITE_MISSING_SCOPES: 'A valid scope must be provided for the invite',
```
kyranet, line 131:
```suggestion
  INVITE_MISSING_SCOPES: 'At least one valid scope must be provided for the invite',
```

typings/index.d.ts
```diff
@@ -3725,7 +3725,7 @@ declare module 'discord.js' {
     permissions?: PermissionResolvable;
     guild?: GuildResolvable;
     disableGuildSelect?: boolean;
-    additionalScopes?: InviteScope[];
+    scopes?: InviteScope[];
```
kyranet, line 3728:
```suggestion
    scopes: InviteScope[];
```

kyranet (changes requested):

src/client/Client.js
monbrey: This comment seems counter to your suggestion though...

src/client/Client.js
monbrey: Will do this but with correct spacing

src/client/Client.js
monbrey: Can't make this change - the TypeError returns `invalidScope` to the user

src/client/Client.js
```diff
@@ -402,10 +408,25 @@ class Client extends BaseClient {
...
+    if (!scopes.some(scope => ['bot', 'applications.commands'].includes(scope))) {
+      throw new TypeError('INVITE_MISSING_SCOPES');
+    }
+    const invalidScope = scopes.find(scope => !InviteScopes.includes(scope));
+    if (invalidScope) {
+      throw new TypeError('INVALID_ELEMENT', 'Array', 'scopes', invalidScope);
+    }
+    query.set('scope', scopes.join(' '));
+
+    if (options.permissions && options.scopes) {
```
SpaceEEC, line 429: Isn't that redundant? Line 417 asserts that it's an array and no array is falsy.
```suggestion
    if (options.permissions) {
```

SpaceEEC (dismissed):
Otherwise LGTM

src/client/Client.js
monbrey: Yeah this is a carry-over from when I thinking about having default scopes.

Should we only bother with permissions if the bot scope is included? It doesn't matter if they're in the URL, Discord just ignores them for applications.commands.

kyranet (approved):
Other than Space's review, LGTM

src/client/Client.js
vladfrangu: Imo, we can just add it, don't have to check the presence of `bot` ourselves

CC @discordjs/the-big-3

vladfrangu (approved):
Other than requested change, LGTM

src/client/Client.js
kyranet: We don't have to, it could be that in a near future, Discord adds support for permissions in slash commands.

iCrawl (approved):

Issue #4929, opened 2020-10-23
Allow ChannelManager#fetch without a guild


GamingGeek:
**Please describe the problem you are having in as much detail as possible:**
When using ChannelManager#fetch for a channel that is not cached and is not in any guilds on the current cluster, I get null rather than the channel

![image](https://user-images.githubusercontent.com/29981918/97032532-e4cf5e00-1559-11eb-9dca-c171a6d83deb.png)

Using `client.api` works fine

![image](https://user-images.githubusercontent.com/29981918/97032786-40015080-155a-11eb-8848-fef38e64b0a7.png)

 

**Include a reproducible code sample here, if possible:**

```js
client.channels.fetch("channelID", false, true) // the channelID likely needs to be a channel accessible to the bot but not accessible to any guilds in the shards of the current process
```

**Further details:**

- discord.js version: 12.4.0
- Node.js version: 14.14.0
- Operating system: Ubuntu Server 18.04
- Priority this issue should have – please be realistic and elaborate if possible: Medium

**Relevant client options:**

- partials: ["REACTION", "MESSAGE", "CHANNEL", "GUILD_MEMBER", "USER"]
- gateway intents: 1607
- other: shardCount is 4 and the shards on the cluster with the issue are [2, 3] whereas the channel in question is on shard 0

- [x] I have also tested the issue on latest master, commit hash: 250c3ae3c1f025c092383e4661bbb386e89a3c23

NotSugden:
is the channel you're trying to fetch from a guild which is not on the current shard? that seems to be the only explanation for this https://github.com/discordjs/discord.js/blob/94c9cc2300d08eba40642f256fbcf9c68487773b/src/managers/ChannelManager.js#L32-L35 here is the only way `ChannelManager#fetch` would return a promise resolving with null

GamingGeek:
@NotSugden It's fetching from the Rest API which is agnostic of shards/clusters meaning it should still work. Using [the same method ChannelManager#fetch uses](https://github.com/discordjs/discord.js/blob/master/src/managers/ChannelManager.js#L91) to fetch the channel from the API works as seen in [this image](https://user-images.githubusercontent.com/29981918/97032786-40015080-155a-11eb-8848-fef38e64b0a7.png) so I'm not quite sure what's happening here.

wasdennnoch:
The issue lies in `Channel#create` where it only creates a d.js channel instance if the parent guild is known:
[...]

PR #6034, opened 2021-07-04
feat: allow channels from uncached guilds to be returned from fetch

ckohen:

This PR allows fetching channels from guilds that do not exist in the client cache (applicable mostly when sharding, but also possible to have guilds uncached via intents limiting or the new collection based limiting)

A big red flag warning here: channels obviously rely on guild internally a lot, however I did not want to change types when the majority of times the values will be present. I would expect people using this option to know exactly what they are doing and which properties can be accessed.

This also fixes up fetching single channels from the guild channel manager, making it cleaner and throwing when trying to fetch a channel from a  different guild (this is the only breaking change)

closes: #4929

### DIFF
--- a/src/errors/Messages.js
+++ b/src/errors/Messages.js
@@ -90,6 +90,7 @@ const Messages = {
   GUILD_CHANNEL_RESOLVE: 'Could not resolve channel to a guild channel.',
   GUILD_VOICE_CHANNEL_RESOLVE: 'Could not resolve channel to a guild voice channel.',
   GUILD_CHANNEL_ORPHAN: 'Could not find a parent to this guild channel.',
+  GUILD_CHANNEL_UNOWNED: 'The fetched channel does not belong to this managers guild',
   GUILD_OWNED: 'Guild is owned by the client.',
   GUILD_MEMBERS_TIMEOUT: "Members didn't arrive in time.",
   GUILD_UNCACHED_ME: 'The client user as a member of this guild is uncached.',
--- a/src/errors/Messages.js
+++ b/src/errors/Messages.js
@@ -90,7 +90,7 @@ const Messages = {
   GUILD_CHANNEL_RESOLVE: 'Could not resolve channel to a guild channel.',
   GUILD_VOICE_CHANNEL_RESOLVE: 'Could not resolve channel to a guild voice channel.',
   GUILD_CHANNEL_ORPHAN: 'Could not find a parent to this guild channel.',
-  GUILD_CHANNEL_UNOWNED: 'The fetched channel does not belong to this managers guild',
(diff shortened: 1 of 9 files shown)


src/errors/Messages.js
```diff
@@ -90,6 +90,7 @@ const Messages = {
   GUILD_CHANNEL_RESOLVE: 'Could not resolve channel to a guild channel.',
   GUILD_VOICE_CHANNEL_RESOLVE: 'Could not resolve channel to a guild voice channel.',
   GUILD_CHANNEL_ORPHAN: 'Could not find a parent to this guild channel.',
+  GUILD_CHANNEL_UNOWNED: 'The fetched channel does not belong to this managers guild',
```
kyranet, line 93:
```suggestion
  GUILD_CHANNEL_UNOWNED: "The fetched channel does not belong to this manager's guild.",
```

kyranet (changes requested):
Just one thing, LGTM otherwise

src/errors/Messages.js
ImRodry: Why change the string type?

monbrey: Because it contains a `'`. This is how eslint/prettier handle's that.

ImRodry: Ah I thought it'd just be \\'

kyranet (approved):

typings/index.d.ts
```diff
@@ -1016,13 +1016,14 @@ declare module 'discord.js' {
...
   export class GuildChannel extends Channel {
-    constructor(guild: Guild, data?: unknown);
+    constructor(guild: Guild, data?: unknown, client?: Client);
     private memberPermissions(member: GuildMember): Readonly<Permissions>;
     private rolePermissions(role: Role): Readonly<Permissions>;
 
     public readonly calculatedPosition: number;
     public readonly deletable: boolean;
     public guild: Guild;
+    public guildID: Snowflake;
```
iCrawl, line 1026:
```suggestion
    public guildId: Snowflake;
```

iCrawl (changes requested):
`ID` -> `Id` changes required.

iCrawl:
Additionally this needs a rebase.

vladfrangu (approved):
Other than requested changes, LGTM

iCrawl (approved):

iCrawl:
This needs a rebase.

SpaceEEC (approved):
> I would expect people using this option to know exactly what they are doing and which properties can be accessed.

We'll see about that 😄

iCrawl:
This needs a rebase.

PR #6035, opened 2021-07-04
refactor(Channel): change channel types to UPPER_CASE

ImRodry:
This PR changes channel types from snake_case to UPPER_CASE to be more consistent with other types across the library and because we didn't have enough breaking changes already! The names for the channel types were taken from Discord's API documentation: https://discord.com/developers/docs/resources/channel#channel-object-channel-types


ImRodry:
and the second I create this there's a merge conflict!

src/client/actions/TypingStart.js
```diff
@@ -2,7 +2,14 @@
 
 const Action = require('./Action');
 const { Events } = require('../../util/Constants');
-const textBasedChannelTypes = ['dm', 'text', 'news', 'news_thread', 'public_thread', 'private_thread'];
+const textBasedChannelTypes = [
```
ImRodry, line 5: A question I had when changing this is: isn't a GROUP_DM channel text-based? I didn't add it because I wasn't sure but I feel like it would make sense

src/managers/GuildChannelManager.js
```diff
@@ -79,8 +79,8 @@ class GuildChannelManager extends CachedManager {
   /**
    * Options used to create a new channel in a guild.
    * @typedef {Object} GuildChannelCreateOptions
-   * @property {string} [type='text'] The type of the new channel, either `text`, `voice`, `category`, `news`,
-   * `store`, or `stage`
+   * @property {string} [type='GUILD_TEXT'] The type of the new channel, either `GUILD_TEXT`, `GUILD_VOICE`,
```
ckohen, line 82: While you're at it, this should take a number, I'd much like to be able to pass in ChannelTypes.GUILD_TEXT

src/client/actions/TypingStart.js
monbrey: Bots can't actually interact with those (e.g. send a message to one) so it makes no sense for them to be text-based as it doesn't have any text-based methods. The only reference to them that exists in the lib is a partial for the data from invites *to* group DMs.

ImRodry: Ah alright, that makes sense

src/client/actions/GuildDelete.js
```diff
@@ -15,7 +15,7 @@ class GuildDeleteAction extends Action {
     let guild = client.guilds.cache.get(data.id);
     if (guild) {
       for (const channel of guild.channels.cache.values()) {
-        if (channel.type === 'text') channel.stopTyping(true);
+        if (channel.type === 'GUILD_TEXT') channel.stopTyping(true);
```
ckohen, line 18: This should probably be changed to include all text based channels, `TextBasedChannelTypes.includes(channel.type)` I believe

src/client/actions/MessageReactionAdd.js
```diff
@@ -23,7 +23,7 @@ class MessageReactionAdd extends Action {
 
     // Verify channel
     const channel = this.getChannel(data);
-    if (!channel || channel.type === 'voice') return false;
+    if (!channel || channel.type === 'GUILD_VOICE') return false;
```
ckohen, line 26: Ditto but reversed, and same for all the other checks

src/client/actions/TypingStart.js
ckohen: This should be a global constant at this point (in the Constants file) so you can use it in other places as mentioned.

ckohen (changes requested):

src/managers/GuildChannelManager.js
ImRodry: I don't think that's possible due to this line:
https://github.com/discordjs/discord.js/blob/ab0b3b9a07f5e4987e4f25e41b2a007f2db06322/src/managers/GuildChannelManager.js#L131
Unless you want me to change that as well?

iCrawl:
This needs a rebase.

ImRodry:
Rebased!

src/client/actions/GuildDelete.js
```diff
@@ -15,7 +15,7 @@ class GuildDeleteAction extends Action {
     let guild = client.guilds.cache.get(data.id);
     if (guild) {
       for (const channel of guild.channels.cache.values()) {
-        if (channel.type === 'text') channel.stopTyping(true);
+        if (TextBasedChannelTypes[channel.type]) channel.stopTyping(true);
```
NotSugden, line 18:
```suggestion
        if (channel.type in TextBasedChannelTypes) channel.stopTyping(true);
```

NotSugden (commented):
same thing in other places ? if `channel.type` is a string then `TextBasedChanneltypes['GUILD_TEXT']` is 0 making it falsy ?

iCrawl:
This needs a rebase.

iCrawl (approved):

iCrawl:
This needs a new rebase.

ImRodry:
Rebased
Also prettier decided it had to format some stuff in the typings file so that also went in the commit

typings/enums.d.ts
```diff
@@ -28,32 +28,18 @@ export enum ApplicationCommandPermissionTypes {
 }
 
 export enum ChannelType {
```
vaporox, line 30: Shouldn't this be called `ChannelTypes` to be consistent with other enums?

ImRodry: There were two enums: ChannelType and ChannelTypes and I had to remove one so I went with that but it can be renamed yeah. 
While we’re at it, should I also change the only other enum with a singular name to plural (InviteTargetType)? @discordjs/the-big-3

ImRodry: Addressed in the latest commit

kyranet (approved):

src/util/Constants.js
```diff
@@ -415,31 +415,73 @@ exports.SystemMessageTypes = exports.MessageTypes.filter(
...
-exports.ThreadChannelTypes = ['news_thread', 'public_thread', 'private_thread'];
+exports.ThreadChannelTypes = ['GUILD_NEWS_THREAD', 'GUILD_PUBLIC_THREAD', 'GUILD_PRIVATE_THREAD'];
+
+/**
+ * The types of channels that are voice-based. The available types are:
+ * * GUILD_VOICE
+ * * GUILD_STAGE_VOICE
+ * @typedef {string} VoiceBasedChannelType
+ */
+exports.VoiceBasedChannelTypes = createEnum(['GUILD_VOICE', 'GUILD_STAGE_VOICE']);
```
SpaceEEC, line 484: This is missing in the typings, I'm also not sure if this really should be an enum, since their values are incorrect.
Maybe just go with an array like with `ThreadChannelTypes`?

src/util/Constants.js
```diff
@@ -415,31 +415,73 @@ exports.SystemMessageTypes = exports.MessageTypes.filter(
...
+ * The types of channels that are text-based. The available types are:
+ * * DM
+ * * GUILD_TEXT
+ * * GUILD_NEWS
+ * * GUILD_NEWS_THREAD
+ * * GUILD_PUBLIC_THREAD
+ * * GUILD_PRIVATE_THREAD
+ * @typedef {string} TextBasedChannelType
+ */
+exports.TextBasedChannelTypes = createEnum([
```
SpaceEEC, line 460: This is missing in the typings, I'm also not sure if this really should be an enum, since their values are incorrect.
Maybe just go with an array like with `ThreadChannelTypes`?

SpaceEEC (dismissed):

src/util/Constants.js
ImRodry: I didn’t know this had to be in the typings but sure, I’ll add it. As for the enum/array it doesn’t really make a difference since these are only used to check if the channel type is inside that enum but for the sake of consistency with thread channel types I can change that too. Same for the other comment

iCrawl:
This needs a rebase.

SpaceEEC (approved):

src/util/Util.js
```diff
@@ -582,7 +582,7 @@ class Util extends null {
     str = str
       .replace(/<@!?[0-9]+>/g, input => {
         const id = input.replace(/<|!|>|@/g, '');
-        if (channel.type === 'dm') {
+        if (channel.type === 'DM') {
```
vladfrangu, line 585: Should use the enums?

vladfrangu (dismissed):

src/util/Util.js
iCrawl: Different PR 👍

ImRodry: Well I guess not

PR #6036, opened 2021-07-04
refactor: change `xID` to `xId`

kyranet:

Did somebody order a pizza? Well, sorry to break it for you, it's another dish of breaking changes.

| ℹ️ This Pull Request does not break every single bot using discord.js, and will for the most part, break those that use `resolveID`. |
| --- |

Barely any bot works untouched when switching from v12 to v13 ~~and Sapphire is still unaffected, even by this~~.

Basically, the gist of this PR is that all instances of `ID` have been changed to `Id`, that includes `resolveID` -> `resolveId`, `messageID` -> `messageId`, `channelID` -> `channelId`, `guildID` -> `guildId`, etc. And this has been in our backlog for around 2 years, so this change was coming out sooner or later.

### DIFF
--- a/src/managers/MessageManager.js
+++ b/src/managers/MessageManager.js
@@ -36,16 +36,16 @@ class MessageManager extends CachedManager {
    * `after` are mutually exclusive. All the parameters are optional.
    * @typedef {Object} ChannelLogsQueryOptions
    * @property {number} [limit=50] Number of messages to acquire
-   * @property {Snowflake} [before] ID of a message to get the messages that were posted before it
-   * @property {Snowflake} [after] ID of a message to get the messages that were posted after it
-   * @property {Snowflake} [around] ID of a message to get the messages that were posted around it
+   * @property {Snowflake} [before] THe message's id to get the messages that were posted before it
+   * @property {Snowflake} [after] THe message's id to get the messages that were posted after it
+   * @property {Snowflake} [around] THe message's id to get the messages that were posted around it
    */
 
   /**
    * Gets a message, or messages, from this channel.
    * <info>The returned Collection does not contain reaction users of the messages if they were not cached.
    * Those need to be fetched separately in such a case.</info>
-   * @param {Snowflake|ChannelLogsQueryOptions} [message] The ID of the message to fetch, or query parameters.
+   * @param {Snowflake|ChannelLogsQueryOptions} [message] The id of the message to fetch, or query parameters.
    * @param {BaseFetchOptions} [options] Additional options for this fetch
    * @returns {Promise<Message>|Promise<Collection<Snowflake, Message>>}
    * @example
@@ -59,7 +59,7 @@ class MessageManager extends CachedManager {
    *   .then(messages => console.log(`Received ${messages.size} messages`))
    *   .catch(console.error);
    * @example
-   * // Get messages and filter by user ID
+   * // Get messages and filter by user id
    * channel.messages.fetch()
    *   .then(messages => console.log(`${messages.filter(m => m.author.id === '84484653687267328').size} messages`))
    *   .catch(console.error);
@@ -96,7 +96,7 @@ class MessageManager extends CachedManager {
    */
 
   /**
-   * Resolves a MessageResolvable to a Message object.
+   * Resolves a {@link MessageResolvable} to a {@link Message} object.
    * @method resolve
    * @memberof MessageManager
    * @instance
@@ -105,8 +105,8 @@ class MessageManager extends CachedManager {
    */
 
   /**
-   * Resolves a MessageResolvable to a Message ID string.
-   * @method resolveID
+   * Resolves a {@link MessageResolvable} to a {@link Message} id.
+   * @method resolveId
    * @memberof MessageManager
    * @instance
    * @param {MessageResolvable} message The message resolvable to resolve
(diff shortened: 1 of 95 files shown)


src/managers/MessageManager.js
```diff
@@ -36,16 +36,16 @@ class MessageManager extends CachedManager {
    * `after` are mutually exclusive. All the parameters are optional.
    * @typedef {Object} ChannelLogsQueryOptions
    * @property {number} [limit=50] Number of messages to acquire
-   * @property {Snowflake} [before] ID of a message to get the messages that were posted before it
-   * @property {Snowflake} [after] ID of a message to get the messages that were posted after it
-   * @property {Snowflake} [around] ID of a message to get the messages that were posted around it
+   * @property {Snowflake} [before] THe message's id to get the messages that were posted before it
+   * @property {Snowflake} [after] THe message's id to get the messages that were posted after it
+   * @property {Snowflake} [around] THe message's id to get the messages that were posted around it
```
iCrawl, line 41:
```suggestion
   * @property {Snowflake} [before] The message's id to get the messages that were posted before it
   * @property {Snowflake} [after] The message's id to get the messages that were posted after it
   * @property {Snowflake} [around] The message's id to get the messages that were posted around it
```

iShibi: I think we can document it like this: "Get the messages that were posted around this message id" and similar for the other two.
There are other places in the lib where we have done it like this to avoid the weird sentence structure.

iCrawl (approved):

ImRodry (commented):
Just a few things I'd like to see clarified. Also if you're doing this for consistency with casing across the library, why not change URL -> Url?

kyranet:
`URL` stands for `Uniform Resource Locator`, it's an acronym, which go always uppercase.

DaStormer:
Isn't ID an abbreviation? So, following camelCase, shouldn't it be capitalized (`xID`)?

Reference: [Microsoft Abbreviation Guidelines](https://docs.microsoft.com/en-us/previous-versions/dotnet/netframework-1.1/141e06ef(v=vs.71)?redirectedfrom=MSDN)

The word `id`, which following camelCase would be `xId`, has a different meaning.

kyranet:
I don't see where in the website says that it should be `ID` over `Id`. Also quite funny you quote MSFT's guidelines when the 2 biggest .NET library for Discord uses `Id`, as you can see in [Discord.NET's `MessageReference`](https://github.com/discord-net/Discord.Net/blob/dev/src/Discord.Net.Core/Entities/Messages/MessageReference.cs) and [DSharpPlus's `DiscordMessageReference`](https://github.com/DSharpPlus/DSharpPlus/blob/master/DSharpPlus/Entities/Channel/Message/DiscordMessageReference.cs). Only library I don't see using `Id` (and rather uses `ID`) is [Remora.Discord](https://github.com/Nihlus/Remora.Discord/blob/master/Backend/Remora.Discord.API/API/Objects/Messages/MessageReference.cs).

monbrey:
The guidelines seem to tell us not to use an abbreviation at all. The final sentence "If you must use abbreviations, use camel case for abbreviations that consist of more than two characters, even if this contradicts the standard abbreviation of the word."

Admittedly this abbreviation is not more than two characters.

But we can do what we want regardless.

Drahcirius:
https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/capitalization-conventions?redirectedfrom=MSDN#capitalizing-compound-words-and-common-terms

TheNoob27:
I know it's way too late but I'm just seeing this now and I figured I'd mention that it's more than just being the same as other discord bot libraries.

What about people with databases? They've probably stored a few structures as userID or guildID by now. The toJSON method obviously used to use ID instead of Id, and people have definitely used that for storing or sending json. So having the keys change could cause some confusion.

Changing all instances of a function like resolveID to resolveId is one thing (although quite inconsistent), but then there's other parts like the short message references and toJSON objects.

Plus, even if no bot works untouched with discord.js v13, that doesn't mean you should make that a _promise_.

kyranet:
> What about people with databases? They've probably stored a few structures as userID or guildID by now. The toJSON method obviously used to use ID instead of Id, and people have definitely used that for storing or sending json. So having the keys change could cause some confusion.

The same as #6189 - your fault for not sanitizing values and instead relying on whatever discord.js gives you (which by the way, is extremely inefficient storage wise).

If I were you and I stored data from Discord, I'd make sure to pass each relevant column/property that I'd need when reading it back, as opposed to passing a dynamic set of properties with a dynamic set of values, since that gives a lot more control and makes migrations across Discord and Discord.js versions a lot easier when they somehow change or get removed.

TheNoob27:
Perhaps I didn't do a great job at explaining the problem I see with this change.

I definitely don't just throw random objects into my database, although I'm not sure if I can speak for everyone.
However, I do store specific keys, the ones that I need, and most of them are IDs. So obviously, seeing as discord.js uses ID, it would only make sense to do so too, right?
So now you've saved millions of documents with ID but you've gotta switch up your code to use Id in some places and ID in other places because of a new discord.js version.
If discord.js used Id throughout its lifetime, I wouldn't care or have a problem, but now I have to deal with saved keys as channelID or messageID while the library I use switches to channelId and messageId.

kyranet:
If you don't just throw random objects into your database, then you're perhaps sanitizing it to some point. Here, the solution is very simple, map `Id` keys to `ID` (e.g. `{ channelID: data.channelId }`) or migrate the data to use `Id`.

PR #6041, opened 2021-07-04
types(Voice): move types to the library's definitions

kyranet:

At first, we used module augmentation so the types could be linked with the actual library's types, however, that ended up not working correctly (and I suspect it's because the `type` definition, which unlike `interface`s, they can't be augmented).

I moved the definitions into discord.js and prefixed them with `Internal` to not conflict with the /voice's definitions, as well as marked them as `@internal` with a description via TSDocs.

### DIFF
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -152,25 +152,6 @@ declare enum WebhookTypes {
 
 type Awaited<T> = T | PromiseLike<T>;
 
-declare module '@discordjs/voice' {
-  import { GatewayVoiceServerUpdateDispatchData, GatewayVoiceStateUpdateDispatchData } from 'discord-api-types/v8';
-
-  export interface DiscordGatewayAdapterLibraryMethods {
-    onVoiceServerUpdate(data: GatewayVoiceServerUpdateDispatchData): void;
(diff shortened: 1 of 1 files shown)


typings/index.d.ts
```diff
@@ -4496,4 +4478,30 @@ declare module 'discord.js' {
...
+    destroy(): void;
+  }
+
+  /**
+   * @internal Use `DiscordGatewayAdapterCreator` from `@discordjs/voice` instead.
+   */
+  type InternalDiscordGatewayAdapterCreator = (
+    methods: InternalDiscordGatewayAdapterLibraryMethods,
+  ) => InternalDiscordGatewayAdapterImplementerMethods;
+  //#endregion
```
iCrawl, line 4506:
```suggestion

  //#endregion
```

typings/index.d.ts
```diff
@@ -4496,4 +4478,30 @@ declare module 'discord.js' {
     : { [K in keyof T]: Serialized<T[K]> };
 
   //#endregion
+
+  //#region voice
```
iCrawl, line 4482:
```suggestion
  //#region voice

```

SpaceEEC (approved):

iCrawl (approved):

PR #6042, opened 2021-07-04
types(Options): should extend null

BannerBomb:
The Options class extends null in the javascript file, but not in the typings.
https://github.com/discordjs/discord.js/blob/185e37602b0fabf6f06a02886128aead9239c5d9/src/util/Options.js#L92
https://github.com/discordjs/discord.js/blob/185e37602b0fabf6f06a02886128aead9239c5d9/typings/index.d.ts#L512

### DIFF
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -509,7 +509,7 @@ declare module 'discord.js' {
     public setUsername(username: string): Promise<this>;
   }
 
-  export class Options {
+  export class Options extends null {
     public static createDefaultOptions(): ClientOptions;
     public static cacheWithLimits(limits?: Record<string, number>): CacheFactory;
     public static cacheEverything(): CacheFactory;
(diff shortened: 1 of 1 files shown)


typings/index.d.ts
```diff
@@ -509,7 +509,7 @@ declare module 'discord.js' {
     public setUsername(username: string): Promise<this>;
   }
 
-  export class Options {
+  export class Options extends null {
```
kyranet, line 512: To disallow constructing from TS:
```suggestion
  export class Options extends null {
    private constructor();
```

kyranet (changes requested):
Otherwise LGTM

kyranet (approved):

SpaceEEC (approved):

iCrawl (approved):

PR #6043, opened 2021-07-04
feat(LimitedCollection): export LimitedCollection

### LINKED ISSUE #4274: LimitedCollection
There are scenarios where I would actually want to use the Limited Collection class, but you cannot import the LimitedCollection class from discord.js, unless you do something like the following: 
```js
const LimitedCollection = require(require.resolve("discord.js").replace("index.js", "util/LimitedCollection.js"))
```
In the discord.js docs, it's listed as Private, but it says that you can initiate the class with 
```js
new Discord.LimitedCollection(maxSize, iterable);
``` 
but this isn't true as its not in the `src/index.js` file and doesn't get imported when you require the library.

I suppose it would make sense to simply add it to the `src/index.js` file, or even add it in the `@discordjs/collection` package in some way.

I guess it would be easy to recreate it or just use the alternative way of getting the class but it'd be so much more simple if it was added to the index file and/or made public.

PanSzelescik:

Closes #4274, I don't know why it isn't exported

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -21,6 +21,7 @@ module.exports = {
   DiscordAPIError: require('./rest/DiscordAPIError'),
   HTTPError: require('./rest/HTTPError'),
   RateLimitError: require('./rest/RateLimitError'),
+  LimitedCollection: require('./util/LimitedCollection'),
   MessageFlags: require('./util/MessageFlags'),
   Intents: require('./util/Intents'),
   Options: require('./util/Options'),
--- a/src/util/LimitedCollection.js
+++ b/src/util/LimitedCollection.js
@@ -8,7 +8,6 @@ const Collection = require('./Collection.js');
  * @extends {Collection}
  * @param {number} [maxSize=0] The maximum size of the Collection
  * @param {Iterable} [iterable=null] Optional entries passed to the Map constructor.
- * @private
  */
 class LimitedCollection extends Collection {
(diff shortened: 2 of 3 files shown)


vladfrangu (approved):

kyranet (approved):

ckohen:
Please remove the @private docstring in the file

PanSzelescik:
> Please remove the `@private` docstring in the file

Removed

iCrawl (approved):

PanSzelescik:
Should I add typings in [Collections region](https://github.com/discordjs/discord.js/blob/fe5d56c9b11b3e5f05933c6d746237b9f353b392/typings/index.d.ts#L2295)?

SpaceEEC (approved):
There is still the issue with implicitly creating new instances of `LimitedCollection` through its methods, e.g. `clone`, `filter`.
TS will say those are `LimitedCollection`s, but they actually are `Collections`.
(Creating new `LimitedCollection`s is not possible without modifying `Collection` and thinking of a way to derive new arbitrary child instances)

I don't know of a way to make TS aware of `Symbol.species` and overwriting all methods to specify that actually a `Collection` is returned is error prone and tedious.

So I suppose 🤷

SpaceEEC:
Yeah, moving the declaration for `LimitedCollection` would be nice, @PanSzelescik.

iCrawl:
This needs a rebase.

PR #6057, opened 2021-07-05
types(*): revert incorrect unknown type changes

ImRodry:
This PR changes the types for all arguments from spread functions from `unknown` to `any`, introduced in #6047. This meant that  any conditions that relied on these functions would return false and break certain things (like Partials)

### DIFF
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -173,9 +173,9 @@ export class BaseClient extends EventEmitter {
   public clearTimeout(timeout: NodeJS.Timeout): void;
   public clearImmediate(timeout: NodeJS.Immediate): void;
   public destroy(): void;
-  public setInterval(fn: (...args: unknown[]) => void, delay: number, ...args: unknown[]): NodeJS.Timeout;
-  public setTimeout(fn: (...args: unknown[]) => void, delay: number, ...args: unknown[]): NodeJS.Timeout;
-  public setImmediate(fn: (...args: unknown[]) => void, ...args: unknown[]): NodeJS.Immediate;
+  public setInterval<T extends any[]>(fn: (...args: T) => Awaited<void>, delay: number, ...args: T): NodeJS.Timeout;
+  public setTimeout<T extends any[]>(fn: (...args: T) => Awaited<void>, delay: number, ...args: T): NodeJS.Timeout;
+  public setImmediate<T extends any[]>(fn: (...args: T) => Awaited<void>, ...args: T): NodeJS.Immediate;
   public toJSON(...props: Record<string, boolean | string>[]): unknown;
 }
 
@@ -303,13 +303,13 @@ export class Client extends BaseClient {
   public on<K extends keyof ClientEvents>(event: K, listener: (...args: ClientEvents[K]) => Awaited<void>): this;
   public on<S extends string | symbol>(
     event: Exclude<S, keyof ClientEvents>,
-    listener: (...args: unknown[]) => Awaited<void>,
+    listener: (...args: any[]) => Awaited<void>,
   ): this;
 
   public once<K extends keyof ClientEvents>(event: K, listener: (...args: ClientEvents[K]) => Awaited<void>): this;
   public once<S extends string | symbol>(
     event: Exclude<S, keyof ClientEvents>,
-    listener: (...args: unknown[]) => Awaited<void>,
+    listener: (...args: any[]) => Awaited<void>,
   ): this;
 
   public emit<K extends keyof ClientEvents>(event: K, ...args: ClientEvents[K]): boolean;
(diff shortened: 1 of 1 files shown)


DTrombett (commented):
Why have all `unknown`s been replaced with `any`?

typings/index.d.ts
```diff
@@ -173,9 +173,9 @@ export class BaseClient extends EventEmitter {
   public clearTimeout(timeout: NodeJS.Timeout): void;
   public clearImmediate(timeout: NodeJS.Immediate): void;
   public destroy(): void;
-  public setInterval(fn: (...args: unknown[]) => void, delay: number, ...args: unknown[]): NodeJS.Timeout;
-  public setTimeout(fn: (...args: unknown[]) => void, delay: number, ...args: unknown[]): NodeJS.Timeout;
-  public setImmediate(fn: (...args: unknown[]) => void, ...args: unknown[]): NodeJS.Immediate;
+  public setInterval<T extends any[]>(fn: (...args: T) => void, delay: number, ...args: T): NodeJS.Timeout;
+  public setTimeout<T extends any[]>(fn: (...args: T) => void, delay: number, ...args: T): NodeJS.Timeout;
+  public setImmediate<T extends any[]>(fn: (...args: T) => void, ...args: T): NodeJS.Immediate;
```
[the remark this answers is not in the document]
ImRodry: isn't that unrelated to the PR though? I have no idea why that's not there

DTrombett:
Yeah, but... I don't think it's too much different...

ImRodry:
It is, it fixes the bug I explained

typings/index.d.ts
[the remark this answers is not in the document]
DTrombett: Well I think this is the right opportunity to add them 😂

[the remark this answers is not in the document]
ImRodry: I'll let the big bois decide @discordjs/the-big-3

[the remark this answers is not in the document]
iCrawl: Should def add that.

typings/index.d.ts
```diff
@@ -173,9 +173,9 @@ export class BaseClient extends EventEmitter {
   public clearTimeout(timeout: NodeJS.Timeout): void;
   public clearImmediate(timeout: NodeJS.Immediate): void;
   public destroy(): void;
-  public setInterval(fn: (...args: unknown[]) => void, delay: number, ...args: unknown[]): NodeJS.Timeout;
-  public setTimeout(fn: (...args: unknown[]) => void, delay: number, ...args: unknown[]): NodeJS.Timeout;
-  public setImmediate(fn: (...args: unknown[]) => void, ...args: unknown[]): NodeJS.Immediate;
+  public setInterval<T extends any[]>(fn: (...args: T) => Awaited<void>, delay: number, ...args: T): NodeJS.Timeout;
```
vladfrangu, line 176: If you replace the T with `T extends unknown[]`.. don't you achieve the same? since TS will (hopefully) infer it

typings/index.d.ts
```diff
@@ -303,13 +303,13 @@ export class Client extends BaseClient {
   public on<K extends keyof ClientEvents>(event: K, listener: (...args: ClientEvents[K]) => Awaited<void>): this;
   public on<S extends string | symbol>(
     event: Exclude<S, keyof ClientEvents>,
-    listener: (...args: unknown[]) => Awaited<void>,
+    listener: (...args: any[]) => Awaited<void>,
```
vladfrangu, line 306: This should remain as unknown, same with once and off

typings/index.d.ts
```diff
@@ -4309,7 +4309,7 @@ export type WSEventType =
   | 'STAGE_INSTANCE_UPDATE'
   | 'STAGE_INSTANCE_DELETE';
 
-export type Serialized<T> = T extends symbol | bigint | (() => unknown)
+export type Serialized<T> = T extends symbol | bigint | (() => any)
```
vladfrangu, line 4312: Same, should be unknown, or the function should be turned into a generic one? CC @kyranet

vladfrangu (changes requested):

typings/index.d.ts
ImRodry: I’m not sure since I’m not experienced with this syntax so I’ll tag @vaporox and @1Computer1

typings/index.d.ts
ImRodry: Why?

vladfrangu: To force you to cast your data or check its type

typings/index.d.ts
kyranet: Doesn't matter because we're not using the output.

kyranet (approved):

vladfrangu (approved):

iCrawl:
This needs a rebase.

ImRodry:
Rebased

PR #6058, opened 2021-07-05
feat(Collector): better types for events

vaporox:

This changes the type of the collector event parameters from `unknown` to something actually helpful.

- There are no code changes

### DIFF
--- a/typings/index.ts
+++ b/typings/index.ts
@@ -494,12 +494,12 @@ notPropertyOf(guildMember, 'lastMessage');
 notPropertyOf(guildMember, 'lastMessageId');
 
 // Test collector event parameters
-declare const mc: MessageCollector;
-mc.on('collect', (...args) => {
+declare const messageCollector: MessageCollector;
+messageCollector.on('collect', (...args) => {
   assertType<[Message]>(args);
 });
 
-declare const rc: ReactionCollector;
-rc.on('dispose', (...args) => {
+declare const reactionCollector: ReactionCollector;
+reactionCollector.on('dispose', (...args) => {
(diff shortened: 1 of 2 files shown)


iCrawl (approved):

vladfrangu (approved):

typings/index.ts
```diff
@@ -489,3 +492,14 @@ notPropertyOf(user, 'lastMessage');
 notPropertyOf(user, 'lastMessageId');
 notPropertyOf(guildMember, 'lastMessage');
 notPropertyOf(guildMember, 'lastMessageId');
+
+// Test collector event parameters
+declare const mc: MessageCollector;
+mc.on('collect', (...args) => {
```
kyranet, line 498:
```suggestion
declare const messageCollector: MessageCollector;
messageCollector.on('collect', (...args) => {
```

typings/index.ts
```diff
@@ -489,3 +492,14 @@ notPropertyOf(user, 'lastMessage');
...
+  assertType<[Message]>(args);
+});
+
+declare const rc: ReactionCollector;
+rc.on('dispose', (...args) => {
```
kyranet, line 503:
```suggestion
declare const reactionCollector: ReactionCollector;
reactionCollector.on('dispose', (...args) => {
```

kyranet (changes requested):
IMO we should declare the names of the constants the same as their types to avoid name conflicts. We can use short names for local non-virtual variables, such as those within an event callback.

Other than this, LGTM.

kyranet (approved):

PR #6060, opened 2021-07-05
refactor(Managers): rename add to _add

1Computer1:

Cus names conflict. Have fun reviewing. It's still a mess.

### DIFF
--- a/src/client/actions/Action.js
+++ b/src/client/actions/Action.js
@@ -26,7 +26,7 @@ class GenericAction {
   getPayload(data, manager, id, partialType, cache) {
     const existing = manager.cache.get(id);
     if (!existing && this.client.options.partials.includes(partialType)) {
-      return manager.add(data, cache);
+      return manager._add(data, cache);
     }
     return existing;
   }
@@ -93,9 +93,9 @@ class GenericAction {
     if (data.guild_id && data.member?.user) {
       const guild = this.client.guilds.cache.get(data.guild_id);
       if (guild) {
-        return guild.members.add(data.member).user;
+        return guild.members._add(data.member).user;
       } else {
-        return this.client.users.add(data.member.user);
+        return this.client.users._add(data.member.user);
       }
     }
     return this.getUser(data);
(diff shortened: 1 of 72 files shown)


ImRodry:
While you're at it, why not rename patch to _patch?

1Computer1:
Someone else can do that.

iCrawl:
Didn't you also say that for this PR.

1Computer1:
💢

vladfrangu:
> While you're at it, why not rename patch to _patch? Or maybe even rename all private properties and methods to add a _ at the beginning

If you're talking about Message#patch, it'd need a proper rename, since it's the function called from MESSAGE_UPDATE 👀, something like `_messageUpdatePatch` or such

ImRodry:
Hm I’m not sure that’d be necessary since there are other _patch functions in other classes, would just be for the sake of consistency and doesn’t need such a big name

ckohen:
Message also already has a _patch

iShibi:
```ts
export interface ThreadMemberManager
extends Omit<CachedManager<Snowflake, ThreadMember, ThreadMemberResolvable>, 'add'> {}
```
In the typings, the `CachedManager` class doesn't have the `add` method anymore, thus there is no need to omit it [here](https://github.com/discordjs/discord.js/blob/7cdf63352a7801de4cfb11bad2f5340c0c71e72c/typings/index.d.ts#L2496) for creating the `ThreadMemberManager` interface. That line can be removed completely and the `ThreadMemberManager` class can extend it just like the others.

*Bonus: no lack of new line will be felt, if there is no line to add a new line after* 😼

iCrawl:
This needs a rebase.

iCrawl (approved):

ckohen:
Since this PR is touching it, `ThreadMemberManager#_add` can be cleaned up a bit (should've done it better to begin with, sorry). Basically remove everything inside it and replace with `return super._add(data, cache, { id: data.user_id });`. If that feels out of scope, I'll do it in another PR.

kyranet:
It's better to do another PR for it, @ckohen

SpaceEEC (dismissed):
You missed one instance :eyes:
https://github.com/discordjs/discord.js/blob/755c180659c125532fe6f8e33e6c3b56e275311b/src/managers/GuildChannelManager.js#L172

vladfrangu (approved):
Noice

SpaceEEC (approved):

kyranet (approved):

PR #6063, opened 2021-07-06
refactor(ApplicationCommandManager): remove unused assignment

kyranet:

The `force` parameter is used only in line 88:

https://github.com/discordjs/discord.js/blob/2e078e44883c1ef5f85ef973d61a305ce2a34251/src/managers/ApplicationCommandManager.js#L88

Which is always unreachable for line 86, where this assignment is at.

### DIFF
--- a/src/managers/ApplicationCommandManager.js
+++ b/src/managers/ApplicationCommandManager.js
@@ -83,7 +83,7 @@ class ApplicationCommandManager extends CachedManager {
    */
   async fetch(id, { guildId, cache = true, force = false } = {}) {
     if (typeof id === 'object') {
-      ({ guildId, cache = true, force = false } = id);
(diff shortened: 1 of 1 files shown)


vladfrangu (approved):

vladfrangu:
Premature clicking of the green button aside, isn't this invalid? It'd default `force` to false if undefined in id, but keep it as true if specified.

Example:

```js
fetch({ guildId: '123', cache: false }); // force is set to false by default, since it's undefined

fetch({ guildId: '123', cache: false, force: true }); // Force is set to true
```

vladfrangu (approved):
It's too early for this, LGTM

iCrawl (approved):

SpaceEEC (approved):

PR #6066, opened 2021-07-06
types(Partial): add partial reactions

DTrombett:
This PR adds the type `AllowedPartials` for structures that can be partial. It also adds the support for partial reactions (the only partial structure without types) and changes how the `Partialize` type works (it should be only used by djs), so that it can easily extend partials structures.

Let me know if there are places where a reaction can be partial, other than in the event.

*Tests included!*


DTrombett:
So these are new changes:
Moved the `partial` getter from DMChannel to Channel so that people can use `<Channel>.partial` even if the channel is not a dm (Shouldn't change a lot for non-typescript users).
Fixed `cleanContent` being not nullable in `Message` for docs.
Added all the necessary tests and fixed some properties being nullable in typings even if they're not when the Structure is not partial.

src/structures/Channel.js
```diff
@@ -74,6 +74,15 @@ class Channel extends Base {
     return new Date(this.createdTimestamp);
   }
 
+  /**
+   * Whether this Channel is a partial
+   * @type {boolean}
+   * @readonly
+   */
+  get partial() {
+    return this instanceof DMChannel && typeof this.lastMessageId === 'undefined';
```
SpaceEEC, line 83: This getter could just always return false and have dm channel (the only one that can be partial currently), do this check:
```suggestion
    return false;
```

src/structures/DMChannel.js
```diff
@@ -50,15 +52,6 @@ class DMChannel extends Channel {
     this.lastPinTimestamp = data.last_pin_timestamp ? new Date(data.last_pin_timestamp).getTime() : null;
   }
 
-  /**
```
SpaceEEC, line 53: This method should be restored.

SpaceEEC (changes requested):

src/structures/Channel.js
DTrombett: Should I add in the jsdoc a note to say that this is always false if it's not a DMChannel?

vladfrangu (approved):

src/structures/Channel.js
```diff
@@ -74,6 +74,16 @@ class Channel extends Base {
     return new Date(this.createdTimestamp);
   }
 
+  /**
+   * Whether this Channel is a partial
+   * Note that this is always false if the channel is not a DM
```
iCrawl, line 79:
```suggestion
   * <info>This is always false outside of DM channels.</info>
```

DTrombett: Added 👍

DTrombett:
I'll fix last thing tomorrow.

DTrombett:
Should be done! LMK if there's something to change.

SpaceEEC (approved):

kyranet (approved):

kyranet:
This needs a rebase, btw.

DTrombett:
I'll do it tomorrow

DTrombett:
Done!

iCrawl (approved):

PR #6071, opened 2021-07-07
refactor(Message): combine patch methods

ckohen:

Two patch methods, both private, second often gets missed during changes. No more!

### DIFF
--- a/src/structures/Message.js
+++ b/src/structures/Message.js
@@ -121,81 +121,93 @@ class Message extends Base {
       this.tts = null;
     }
 
-    /**
-     * A random number or string used for checking message delivery
-     * <warn>This is only received after the message was sent successfully, and
-     * lost if re-fetched</warn>
-     * @type {?string}
-     */
-    this.nonce = partial ? this.nonce : 'nonce' in data ? data.nonce : null;
-
-    /**
-     * A list of embeds in the message - e.g. YouTube Player
-     * @type {MessageEmbed[]}
-     */
-    this.embeds = data.embeds?.map(e => new Embed(e, true)) ?? (partial ? this.embeds.slice() : []);
-
-    /**
-     * A list of MessageActionRows in the message
-     * @type {MessageActionRow[]}
-     */
-    this.components =
-      data.components?.map(c => BaseMessageComponent.create(c, this.client)) ??
-      (partial ? this.components.slice() : []);
-
-    /**
-     * A collection of attachments in the message - e.g. Pictures - mapped by their ids
-     * @type {Collection<Snowflake, MessageAttachment>}
-     */
-    this.attachments =
-      partial && !data.attachments && this.attachments ? new Collection(this.attachments) : new Collection();
-    if (data.attachments) {
-      for (const attachment of data.attachments) {
-        this.attachments.set(attachment.id, new MessageAttachment(attachment.url, attachment.filename, attachment));
-      }
+    if (!partial) {
+      /**
+       * A random number or string used for checking message delivery
+       * <warn>This is only received after the message was sent successfully, and
+       * lost if re-fetched</warn>
+       * @type {?string}
+       */
+      this.nonce = 'nonce' in data ? data.nonce : null;
     }
 
-    /**
-     * A collection of (partial) stickers in the message
-     * @type {Collection<Snowflake, Sticker>}
-     */
-    this.stickers = new Collection(
-      (data.sticker_items ?? data.stickers)?.map(s => [s.id, new Sticker(this.client, s)]),
-    );
-
-    /**
-     * The timestamp the message was sent at
-     * @type {number}
-     */
-    this.createdTimestamp = SnowflakeUtil.deconstruct(this.id).timestamp;
-
-    /**
-     * The timestamp the message was last edited at (if applicable)
-     * @type {?number}
-     */
-    this.editedTimestamp = data.edited_timestamp ? new Date(data.edited_timestamp).getTime() : null;
-
-    /**
-     * A manager of the reactions belonging to this message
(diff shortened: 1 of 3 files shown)


src/structures/Message.js
```diff
@@ -238,7 +259,7 @@ class Message extends Base {
      * Flags that are applied to the message
      * @type {Readonly<MessageFlags>}
      */
-    this.flags = new MessageFlags(data.flags).freeze();
+    this.flags = new MessageFlags(partial ? data.flags ?? 0 : data.flags).freeze();
```
vladfrangu, line 262: Isn't this technically wrong? What if a message has flags, we get a MESSAGE_UPDATE for god knows what, and the flags don't change? Won't this make the cached flags invalid with what they actually are?

src/structures/Message.js
```diff
@@ -298,47 +328,6 @@ class Message extends Base {
...
-  /**
-   * Updates the message and returns the old message.
-   * @param {APIMessage} data Raw Discord message update data
-   * @returns {Message}
-   * @private
-   */
-  patch(data) {
-    const clone = this._clone();
-
-    if (data.edited_timestamp) this.editedTimestamp = new Date(data.edited_timestamp).getTime();
```
vladfrangu, line 310: This isn't set anywhere anymore, unless I missed it

src/structures/Message.js
```diff
@@ -127,25 +127,27 @@ class Message extends Base {
...
+    this.components =
+      data.components?.map(c => BaseMessageComponent.create(c, this.client)) ??
+      (partial ? this.components.slice() : []);
 
     /**
      * A collection of attachments in the message - e.g. Pictures - mapped by their ids
      * @type {Collection<Snowflake, MessageAttachment>}
      */
-    this.attachments = new Collection();
+    this.attachments = partial && this.attachments ? new Collection(this.attachments) : new Collection();
```
vladfrangu, line 150: This doesn't seem to do the same thing anymore?

Previously, if `attachments` was present, it'd clear the collection and fill it in with the new attachments, otherwise cloning the existing attachments.

Now, if we get a partial update, we clone the previous attachments and append to them / overwrite some of them.

Am I reading the code wrong?

src/structures/Message.js
```diff
@@ -179,54 +181,73 @@ class Message extends Base {
...
-      data.mentions,
-      data.mention_roles,
-      data.mention_everyone,
-      data.mention_channels,
-      data.referenced_message?.author,
-    );
+    if (partial) {
+      this.mentions = new Mentions(
+        this,
+        data.mentions ?? this.mentions.users,
```
vladfrangu, line 194: Can't this now throw an error for an uncached message (with partials enabled)? Since `this.mentions` is never defined

vladfrangu (changes requested):

iCrawl:
This needs a rebase.

src/structures/Message.js
ckohen: This seems to be errored logic in the original as well, I'll replace 0 with this.flags

src/structures/Message.js
ckohen: It is, just not in the diff since I didn't need to change it
https://github.com/discordjs/discord.js/pull/6071/files#R178

src/structures/Message.js
ckohen: Nope, I wrote it wrong, supposed to be !data.attachments 🤦

src/structures/Message.js
ckohen: No, the only way this could cause errors is if the _patch call in the constructor sets partial to true, so don't do that!

kyranet (approved):

src/structures/Message.js
vladfrangu: But the MessageUpdateAction literally may create a new Message object and call _patch with partial=true...so, are you sure this isn't an issue?

ckohen: You can't call _patch(data, true) on an uninstantiated message. In creating the new Message object via _add, it calls _patch(data), not partial, then patches as partial later.

iCrawl (approved):

vladfrangu (approved):

src/structures/Message.js
```diff
@@ -127,25 +127,28 @@ class Message extends Base {
      * lost if re-fetched</warn>
      * @type {?string}
      */
-    this.nonce = 'nonce' in data ? data.nonce : null;
+    this.nonce = partial ? this.nonce : 'nonce' in data ? data.nonce : null;
```
NotSugden, line 130: `if (!partial)`?

src/structures/Message.js
```diff
@@ -179,54 +182,73 @@ class Message extends Base {
...
+        data.referenced_message?.author,
+      );
+    }
 
     /**
      * The id of the webhook that sent the message, if applicable
      * @type {?Snowflake}
      */
-    this.webhookId = data.webhook_id ?? null;
+    this.webhookId = partial && this.webhookId ? this.webhookId : data.webhook_id ?? null;
```
NotSugden, line 220: same here ?

NotSugden (commented):
i feel like if statements make more sense than assigning the property to itself

src/structures/Message.js
ckohen: Sure, will use if statements to make it clear

ckohen:
Ended up doing it for everything, makes the code overall much easier to read.

vladfrangu:
As a wise, young man once said, this needs a rebase

SpaceEEC (approved):
I'm not sure what's major about this, but well 😅

PR #6065, opened 2021-07-06
refactor: rename `Constants.OPCode` to `Opcode`, moved objects to enums

kyranet:
**Please describe the changes this PR makes and why it should be merged:**

Renamed `Constants.OPCode` for consistency with recent changes, also moved enum-like objects to the enums file and added tests to ensure they don't receive properties that shouldn't exist.

**Status and versioning classification:**

- Code changes have been tested against the Discord API, or there are no code changes
- I know how to update typings and have done so, or typings don't need updating
- This PR includes breaking changes (methods removed or renamed, parameters moved or removed)

PR #6078, opened 2021-07-08
refactor: change const enums to interfaces

DTrombett:

This PR converts all `const enum`s to `interface`s and move them to the main `index.d.ts` file so that TS users can use index access with variables without receiving the error `A const enum member can only be accessed using a string literal`.


SpaceEEC (changes requested):
Doing this brings back this issue: https://github.com/discordjs/discord.js/pull/6065#discussion_r664458288

Those enums should either be objects in the declarations or enum equivalents in the constants file (would require a small modification to the `createEnum` function)

DTrombett:
Mmh should we change them to interfaces or types?

vladfrangu:
If we go with space's suggestion, I'm more leaning towards the latter

SpaceEEC:
I think we'd use interfaces there, if we go that route.

DTrombett:
So I should move them to the main index.d.ts file and convert them to interfaces?

DTrombett:
Converted all const enums to interfaces. LMK if it's ok.

SpaceEEC (approved):

iCrawl (approved):

kyranet (approved):
A very large downside of this is that we can't use them like enums anymore in types, which forces us to do `typeof ConstantsEvents['MESSAGE_CREATE']` over `ConstantsEvents.MESSAGE_CREATE`, but sure.

vladfrangu (approved):

PR #6079, opened 2021-07-08
docs(InteractionResponses): reply docs example

anandre:

The `reply` method returns `Promise<void>`, so `.then(console.log)` is pointless.  This replaces the log with a simple string to acknowledge success.  I left other methods that potentially resolve into `void` (`defer`, `followUp`, etc) alone.

### DIFF
--- a/src/structures/interfaces/InteractionResponses.js
+++ b/src/structures/interfaces/InteractionResponses.js
@@ -63,7 +63,7 @@ class InteractionResponses {
    * const embed = new MessageEmbed().setDescription('Pong!');
    *
    * interaction.reply({ embeds: [embed] })
-   *   .then('Reply sent.')
+   *   .then(() => console.log('Reply sent.'))
    *   .catch(console.error);
    * @example
    * // Create an ephemeral reply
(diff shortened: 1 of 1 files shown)


src/structures/interfaces/InteractionResponses.js
```diff
@@ -63,7 +63,7 @@ class InteractionResponses {
    * const embed = new MessageEmbed().setDescription('Pong!');
    *
    * interaction.reply({ embeds: [embed] })
-   *   .then(console.log)
+   *   .then('Reply sent.')
```
DTrombett, line 66: This will throw a big error...
```suggestion
   *   .then(() => console.log('Reply sent.'))
```

DTrombett (changes requested):

DTrombett (approved):

vaporox:
Shouldn't at least the example at line 71 be changed too?

SpaceEEC (approved):
There are a few more logs like these in this file, if you want to include them. :eyes:

iCrawl (approved):

PR #6081, opened 2021-07-09
revert: support for nested arrays of components, fix error handling

monbrey:
As nice as this was for the user interface, as per internal discussion it was considered not future proof should there ever be a different type of Row component.

This removes support for sending a nested array of components, e.g. no more:
`message.channel.send({ content, components: [[button]] })`

I also noticed that the way component types were being checked didn't work in some scenarios - this simplifies that process to ensure `client` is passed where it should be a debug event, and isn't everywhere else that it should throw.

### DIFF
--- a/src/structures/BaseMessageComponent.js
+++ b/src/structures/BaseMessageComponent.js
@@ -57,7 +57,7 @@ class BaseMessageComponent {
    * @returns {?MessageComponent}
    * @private
    */
-  static create(data, client, skipValidation = false) {
+  static create(data, client) {
     let component;
     let type = data.type;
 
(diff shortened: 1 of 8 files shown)


iCrawl (approved):

src/structures/BaseMessageComponent.js
```diff
@@ -57,7 +57,7 @@ class BaseMessageComponent {
    * @returns {?MessageComponent}
    * @private
    */
-  static create(data, client, skipValidation = false) {
+  static create(data, client) {
```
Jiralite, line 60: With the removal of `skipValidation`, I believe the `@param` for this should also go on line 56, right?

monbrey: Yeah thanks, I always miss docs stuff

vladfrangu (approved):

ImRodry:
😭

z0rt20:
this is a much easier expression, thanks so much!

kyranet (approved):

ImRodry:
Are you guys sure there will even be another type of Row? What advantages would this even give compared to the current rows? This PR literally just makes our life 10x harder than it was before :/

PR #6084, opened 2021-07-09
refactor(InteractionCollector): only keep Ids of objects

Bluenix2:
Updates the typings for InteractionCollectorOptions so that a message sent by a webhook can be passed to it.


HarmoGlace:
The typins are incorrect as it might be an ephemeral message too.

Bluenix2:
Do you mean that the typings *are still incorrect*, or that my change is incorrect?

My specific use-case is that I use webhooks, looking at the implementation of InteractionCollector it only uses the ID of the message, which APIMessage has.

monbrey:
Yeah, not sure what ephemerals have to do with this - there's no unique typings for those. This seems fine.

HarmoGlace:
Well ephemeral messages only have the `id` and `flags` (ephemeral) properties, so it cannot be a valid Message - It should have its own type/be considered as a partial message.

monbrey:
The more I think about this - this isn't strictly correct. The implementation of `InteractionCollector` isn't fully designed to support raw messages.

I dont think you should just slap another typing on it without making changes to handle the fact that `channel` and `guild` properties won't exist, and therefore can't bind delete listeners.

Bluenix2:
> I dont think you should just slap another typing on it without making changes to handle the fact that `channel` and `guild` properties won't exist, and therefore can't bind delete listeners.
@monbrey 

Correction: The library should actually handle this perfectly fine. `this.message.channel` will be undefined

https://github.com/discordjs/discord.js/blob/d6c43a50bd608c233919e6386683c126cf821346/src/structures/InteractionCollector.js#L44

https://github.com/discordjs/discord.js/blob/d6c43a50bd608c233919e6386683c126cf821346/src/structures/InteractionCollector.js#L50

monbrey:
> > I dont think you should just slap another typing on it without making changes to handle the fact that `channel` and `guild` properties won't exist, and therefore can't bind delete listeners.
> > @monbrey
> 
> Correction: The library should actually handle this perfectly fine. `this.message.channel` will be undefined
> 
> https://github.com/discordjs/discord.js/blob/d6c43a50bd608c233919e6386683c126cf821346/src/structures/InteractionCollector.js#L44
> 
> https://github.com/discordjs/discord.js/blob/d6c43a50bd608c233919e6386683c126cf821346/src/structures/InteractionCollector.js#L50

That's exactly the problem I'm referring to. The reason for this check is to get the channel from the message if one was provided, instead of relying on the other options which could be incorrect.

The fact that it won't cause a fatal error doesn't means it's still functioning as intended.

Bluenix2:
@monbrey I think I get what you mean now. Does 15837bf39ce69be6ef4f6f71dcf3897695d0015a solve what you are talking about?

src/structures/InteractionCollector.js
```diff
@@ -33,21 +33,28 @@ class InteractionCollector extends Collector {
 
     /**
      * The message from which to collect interactions, if provided
-     * @type {?Message}
+     * @type {?Snowflake}
      */
-    this.message = options.message ?? null;
+    this.messageId = options.message?.id ?? options.message ?? null;
```
NotSugden, line 38: might it be worth having some sort of `resolveId` function for resolving the message ID?

src/structures/InteractionCollector.js
```diff
@@ -33,21 +33,28 @@ class InteractionCollector extends Collector {
...
 
     /**
      * The channel from which to collect interactions, if provided
-     * @type {?(TextChannel|DMChannel|NewsChannel)}
+     * @type {?Snowflake}
      */
-    this.channel = this.message?.channel ?? options.channel ?? null;
+    this.channelId =
+      options.message?.channel?.id ?? options.message?.channel_id ?? options.channel?.id ?? options.channel ?? null;
```
NotSugden, line 45: same here, perhaps using `this.client.channels.resolveId` but would require some modifications to accept raw api data so unsure

NotSugden (commented):
etc

src/structures/InteractionCollector.js
Bluenix2: It's quite horribly chained I agree, using `resolveId` will help a bit I suppose.
```js
this.channelId = this.client.channels.resolveId(options.message?.channel) ?? options.message.channel_id ?? this.client.channels.resolveId(options.channel) ?? null;
```

monbrey:
No, this still introduces the same flaws in the design. I was originally considering the same things - why not just accept Snowflakes/Resolvables for these parameters?

The answer is because there's really no way to guarantee three different Snowflakes are remotely related, if provided for all of message/channel/guild in options. This is why the original code preferred the Message object over any additionally provided options.

If this is the redesign you want to go with then yes, `resolveId` would be the right approach, and the options should just accept resolvables, not unions.

src/structures/InteractionCollector.js
```diff
@@ -33,21 +33,28 @@ class InteractionCollector extends Collector {
 
     /**
      * The message from which to collect interactions, if provided
-     * @type {?Message}
+     * @type {?Snowflake}
      */
-    this.message = options.message ?? null;
+    this.messageId = options.message?.id;
```
monbrey, line 38:
```suggestion
    this.messageId = options.message?.id ?? null;
```
Set it to null if not provided (as per typings), don't leave it undefined.

monbrey (changes requested):
Just one fix, otherwise LGTM

iCrawl:
This needs a rebase.

iCrawl (approved):

kyranet (approved):

SpaceEEC (approved):

PR #6085, opened 2021-07-09
types(PartialWebhookFields): add APIMessage to deleteMessage union

Bluenix2:

Without this change the following code does not compile:
```ts
const client = new Client();

(async function() {
  const webhook = new Webhook(client, {id: '123', token: '456abc'});
  const msg = await webhook.send({ content: 'Test' });

  await webhook.deleteMessage(msg);  // Argument of type 'Message | APIMessage' is not assignable to parameter of type 'MessageResolvable | "@original"'.
})();
```

### DIFF
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -2447,7 +2447,7 @@ export function WebhookMixin<T>(Base?: Constructable<T>): Constructable<T & Webh
 export interface PartialWebhookFields {
   id: Snowflake;
   readonly url: string;
-  deleteMessage(message: MessageResolvable | '@original'): Promise<void>;
+  deleteMessage(message: MessageResolvable | APIMessage | '@original'): Promise<void>;
   editMessage(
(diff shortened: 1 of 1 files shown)


NotSugden:
but an `APIMessage` isn't a `MessageResolvable`

Bluenix2:
Ah, do you recommend I change the following line instead?

https://github.com/discordjs/discord.js/blob/610773d3a8a5ffeb96dab2d75ee13dc5f5bd0a2a/typings/index.d.ts#L2450

So that it's: `deleteMessage(message: MessageResolvable | APIMessage | '@original'): Promise<void>;`

NotSugden:
yeah that might be the better option

SpaceEEC (approved):

iCrawl (approved):

kyranet (approved):

vladfrangu (approved):

PR #6094, opened 2021-07-11
types(Interaction): change inGuild return type

DTrombett:

This PR removes the `inGuild` method in `Interaction`, because, as discussed in the [Discord server](https://canary.discord.com/channels/222078108977594368/682166281826598932/863179005350903858), it's used only in this structure (not in any other `Structure` like Message or `Channel`) and doesn't provide any type guard for TS users so it's useless, as shown below:
```ts
client.on("interactionCreate", (interaction) => {
    if (interaction.inGuild()) console.log(interaction.client.guilds.fetch(interaction.guildId));
})
```
```ts
client.on("interactionCreate", (interaction) => {
    if (interaction.guildId) console.log(interaction.client.guilds.fetch(interaction.guildId));
})
```
These two codes are pretty much the same and should return the same result so, for a JS user there's no difference, but they might choose the second, as there's no need to call a function; For a TS user, the first will show an error like `guildId is possibly null`, while the second will work as expected.

For those who used this method, it can easily be replaced:
```diff
- interaction.inGuild()
+ interaction.guildId // Or `!!interaction.guildId`, or `Boolean(interaction.guildId)`
```

### DIFF
--- a/src/structures/Interaction.js
+++ b/src/structures/Interaction.js
@@ -106,14 +106,6 @@ class Interaction extends Base {
   }
 
   /**
-   * Indicates whether this interaction is received from a guild.
-   * @returns {boolean}
-   */
-  inGuild() {
-    return Boolean(this.guildId && this.member);
-  }
-
-  /**
    * Indicates whether this interaction is a {@link CommandInteraction}.
    * @returns {boolean}
    */
--- a/src/structures/Interaction.js
+++ b/src/structures/Interaction.js
@@ -106,6 +106,14 @@ class Interaction extends Base {
   }
 
   /**
+   * Indicates whether this interaction is received from a guild.
+   * @returns {boolean}
+   */
+  inGuild() {
+    return Boolean(this.guildId && this.member);
+  }
+
+  /**
    * Indicates whether this interaction is a {@link CommandInteraction}.
    * @returns {boolean}
    */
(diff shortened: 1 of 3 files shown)


vladfrangu:
I don't think this should be deleted at all, but rather refactored to actually assert that the `Interaction` instance is received from a guild (consider this: example
```ts
client.on('interactionCreate', async (interaction) => {
	if (interaction.inGuild()) {
		// We know that member is not null now!
		const roles = interaction.member.roles;
		const roleCount = Array.isArray(roles) ? roles.length : roles.cache.size;

		await interaction.reply({ content: `Hello ${interaction.user.tag}, you have ${roleCount} roles 👀`, components: [...] });
	} 
});
```
)

Of course, this is a niche, and you can just use `!` if you know what you're doing, but you shouldn't have to do that. That's my opinion tho, CC @discordjs/the-big-3 for their input

monbrey:
Have discussed with Vlad, this type should achieve the original intention of the method:

```ts
public inGuild(): this is this & { guildId: Snowflake, member: GuildMember | APIInteractionGuildMember };
```

DTrombett:
Ok, that seems like the best solution. I will work on it as soon as possible.

DTrombett:
Changed `inGuild` return type and added tests. LMK if it's ok.

SpaceEEC (approved):

kyranet (approved):

vladfrangu (approved):

PR #6109, opened 2021-07-13
types(Interactions): fix function overloads

thehackerboi69github:
Fixes an issue with `Interaction#defer` and `MessageComponentInteraction#deferUpdate`, due to the overload types the return type would default to `Promise<Message | APIMessage>` when no parameter was passed.

### DIFF
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -1071,7 +1071,7 @@ export class MessageComponentInteraction extends Interaction {
   public webhook: InteractionWebhook;
   public defer(options: InteractionDeferOptions & { fetchReply: true }): Promise<Message | APIMessage>;
   public defer(options?: InteractionDeferOptions): Promise<void>;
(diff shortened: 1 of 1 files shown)


kyranet (approved):
If you could add some type tests in `typings/index.ts` to ensure the returned type is the correct one, this would be perfect. Otherwise LGTM

iCrawl (approved):

typings/index.d.ts
```diff
@@ -419,7 +419,7 @@ export class CommandInteraction extends Interaction {
   public options: Collection<string, CommandInteractionOption>;
   public replied: boolean;
   public webhook: InteractionWebhook;
-  public defer(options?: InteractionDeferOptions & { fetchReply: true }): Promise<Message | APIMessage>;
+  public defer(options: InteractionDeferOptions & { fetchReply: true }): Promise<Message | APIMessage>;
```
DTrombett, line 422: Wait, this means that we can't use `interaction.defer()` without passing any parameter?

vladfrangu (approved):

PR #6110, opened 2021-07-13
feat(Util): add SweptCollection for auto sweeping of caches

ckohen:

⚠️ Updates Collection dep, including breaking changers
⚠️ Updates node requirement to 14.6.0 for Finalizers and WeakRef

Eslint config has been updated to target es2021 as well.

 Adds SweptCollection (mainly for use with threads but has other uses). This can replace Client#sweepMessages, however with the loss of debug events, which is why it was not removed.

Other breaking change is the change in ThreadManager, to automatically sweep archived threads older than 4 hours every hour (default sweep settings)

I'm open to changes for these defaults, kinda just picked them at random.


src/util/Options.js
```diff
@@ -149,6 +153,53 @@ class Options extends null {
...
+   * @returns {CacheFactory}
+   */
+  static cacheWithSweep(settings = {}) {
+    const Collection = require('./Collection');
+    const SweptCollection = require('./SweptCollection');
+
+    return manager => {
+      const setting = settings[manager.name];
+      if (
+        setting === null || setting === undefined || typeof setting !== 'object' || setting.sweepInterval
```
DTrombett, line 168: `typeof setting !== 'object'` should already exclude undefined.
```suggestion
        setting === null || typeof setting !== 'object' || setting.sweepInterval
```

src/util/Options.js
```diff
@@ -149,6 +153,53 @@ class Options extends null {
...
+    const LimitedCollection = require('./LimitedCollection');
+    const SweptCollection = require('./SweptCollection');
+
+    return manager => {
+      const setting = settings[manager.name];
+      if (typeof setting === 'number' && setting !== Infinity) return LimitedCollection(setting);
+      if (
+        setting === null ||
+        setting === undefined ||
+        typeof setting !== 'object' ||
```
DTrombett, line 194: Same as above.

```suggestion
        typeof setting !== 'object' ||
```

src/util/SweptCollection.js
```diff
@@ -0,0 +1,129 @@
...
+
+    /**
+     * A predicate for determining which item to remove when the Collection reaches max size.
+     * Using this can allow the size to grow larger than max size if none of the entries match the predicate.
+     * @type {?Function}
+     */
+    this.maxSizePredicate = maxSizePredicate;
+
+    const intervalFunction = sweepFunction
+      ? () => this.sweep(sweepFunction)
```
DTrombett, line 90: This should be converted to a plain `function` since arrow functions don't have `this` and cannot be binded.

src/util/SweptCollection.js
```diff
@@ -0,0 +1,129 @@
...
+      throw new TypeError('INVALID_TYPE', 'sweepLifetime', 'number');
+    }
+    if (typeof sweepFunction !== 'function' && typeof sweepLifetimeProperty !== 'string') {
+      throw new TypeError('INVALID_TYPE', 'sweepLifetimeProperty', 'number');
+    }
+    if (typeof sweepInterval !== 'number') throw new TypeError('INVALID_TYPE', 'sweepInterval', 'number');
+    if (typeof sweepFunction !== 'function' && typeof sweepArchivedOnly !== 'boolean') {
+      throw new TypeError('INVALID_TYPE', 'sweepArchivedOnly', 'boolean');
+    }
+    if (typeof client !== 'object') throw new TypeError('INVALID_TYPE', 'client', 'Client');
```
DTrombett, line 65: Wouldn't be good here to use `!(client instanceof Client)` instead?

src/util/SweptCollection.js
```diff
@@ -0,0 +1,129 @@
...
+      : () => {
+          if (sweepLifetime <= 0) return;
+          const lifetimeMs = sweepLifetime * 1000;
+          const now = Date.now();
+          this.sweep(entry => {
+            if (!entry[sweepLifetimeProperty]) return false;
+            if (sweepArchivedOnly && !entry.archived) return false;
+            return now - entry[sweepLifetimeProperty] > lifetimeMs;
+          });
+        };
```
DTrombett, line 100: Same as [above](https://github.com/discordjs/discord.js/pull/6110#discussion_r668681218).

src/util/SweptCollection.js
ckohen: I actually bound this out of habit, it's unnecessary. Arrow functions use the parent scope, so I'll just remove the bind

src/util/SweptCollection.js
```diff
@@ -0,0 +1,129 @@
...
+ * @property {number} [sweepInterval=3600] How frequently, in seconds, to sweep the collection.
+ */
+
+/**
+ * A Collection which holds a max amount of entries and sweeps periodically.
+ * The first key is deleted if the Collection has reached max size.
+ * @extends {Collection}
+ * @param {SweptCollectionOptions|Iterable} [options=null] Options for constructing the swept collection.
+ */
+class SweptCollection extends Collection {
```
vladfrangu, line 33: Shouldn't this be called something like.. SweepableCollection? English is beating me, so I'll summon the @discordjs/the-big-3

src/util/SweptCollection.js
ckohen: Technically speaking the jsdoc and typings say that you pass a djs client, but that's not necessarily a requirement,  its just an easy way to describe what it wants. For typings I can make it better, I have no idea what to do for the docs though.

The idea is you could pass your own client with a setInterval method since this doesn't *have* to be used within djs, allowing it to be destroyed with your client (that's the whole reason for it anyways, when done internally, we want to destroy the intervals when the client is destroyed)

src/util/SweptCollection.js
ckohen: I followed the convention of LimitedCollection, open to change though

kyranet: Collections have a sweep method, which makes them... sweepable.

SweptCollection is fine.

src/util/SweptCollection.js
```diff
@@ -0,0 +1,129 @@
...
+ * @property {Function|null} [sweepFunction=null] A custom function passed directly to `sweep()` every `sweepInterval`.
+ * See {@link [Collection#sweep](https://discord.js.org/#/docs/collection/master/class/Collection?scrollTo=sweep)}
+ * for the definition of this function. This overrides the build in function generated using the other properties.
+ * @property {number} [sweepLifetime=14400] How long an entry should stay in the collection
+ * before it is considered sweepable
+ * @property {string} [sweepLifetimeProperty='createdTimestamp'] The property of the entry to check when
+ * determining the entries lifetime.
+ * @property {boolean} [sweepArchivedOnly=false] Whether to only sweep entries with a property `archived: true`.
+ * E.g. for threads.
+ * @property {number} [sweepInterval=3600] How frequently, in seconds, to sweep the collection.
```
kyranet, line 24: So... per hour?

Also, why is this in seconds?

kyranet (changes requested):
Also, just one tiny question... we need to access to the `Client` instance, in an object that is used to build the instance itself, do you have an example on how to use this?

1Computer1:
From the discord, my suggestions on the interface:

- Remove client related stuff after #6113.

- More general interface like so:
```ts
export class SweptCollection<K, V> extends Collection<K, V> {
  public constructor(options: SweptCollectionOptions<K, V>, iterable?: Iterable<readonly [K, V]>);
  public maxSize: number;
  public maxSizePredicate: ((value: V) => boolean) | null;
  public interval: number | null;
}

export interface SweptCollectionOptions<K, V> {
  maxSize?: number;
  shouldKeep?: (value: V) => boolean;
  sweepInterval?: number;
  sweepFilter?: () => (value: V) => boolean;
}
```

Usage example:

```ts
new SweptCollection({
  sweepFilter: () => {
    const now = Date.now();
    return (t) => {
      if (!t.createdTimestamp) return false;
      if (t.archived) return false;
      return now - t.createdTimestamp > lifetimeMs;
    };
  }
});
```

- Possibly remove letting users keep items at max cache due to the fact that `set` is now O(n).

- Too many `cacheWith`s now, I think its best to do this perhaps:
```ts
public static cacheByManager(fs: Record<string, () => Collection<any, any>>): CacheFactory { ... }
```

src/util/SweptCollection.js
ckohen: Yeah, sweeps hourly by default, I have no idea if thats a good interval, it definitely could be pretty long.

Do you mean in seconds vs minutes or seconds vs ms

src/util/SweptCollection.js
```diff
@@ -0,0 +1,129 @@
...
+
+/**
+ * A Collection which holds a max amount of entries and sweeps periodically.
+ * The first key is deleted if the Collection has reached max size.
+ * @extends {Collection}
+ * @param {SweptCollectionOptions|Iterable} [options=null] Options for constructing the swept collection.
+ */
+class SweptCollection extends Collection {
+  constructor(options = null) {
+    if (typeof options !== 'object') throw new TypeError('INVALID_TYPE', 'options', 'object or iterable', true);
```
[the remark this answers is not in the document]
ckohen: I swear I handled that 🤦. I want to not be able to pass no options though, so default is just wrong, will be {} , while still needing the check for null just in caes.

src/util/SweptCollection.js
ckohen: client is now removed from this :)

ckohen:
Did most of what comp suggested, couple of name changes to make it clearer what these things do.

The only thing I didn't touch:
> * Too many `cacheWith`s now, I think it's best to do this perhaps:
> 
> ```ts
> public static cacheByManager(fs: Record<string, () => Collection<any, any>>): CacheFactory { ... }
> ```
What's the point in this, this seems harder than just implementing your own makeCache function. In fact, the only difference  is that you're not building the default case.

src/util/SweptCollection.js
kyranet: Seconds vs ms.

src/util/Options.js
```diff
@@ -98,7 +98,15 @@ class Options extends null {
   static createDefault() {
     return {
       shardCount: 1,
-      makeCache: this.cacheWithLimits({ MessageManager: 200 }),
+      makeCache: this.cacheWithLimitsOrSweep({
+        MessageManager: 200,
+        ThreadManager: {
+          sweepInterval: require('./SweptCollection').filterByLiftetme({
```
kyranet, line 104:
```suggestion
          sweepInterval: require('./SweptCollection').filterByLifetime({
```

src/util/SweptCollection.js
```diff
@@ -0,0 +1,138 @@
...
+   * @property {Function} [excludeFromSweep=null] A function that takes an entry and returns a boolean,
+   * `true` when the entry should not be checked for sweepability.
+   */
+
+  /**
+   * Create a sweepFilter function that uses a lifetime to determine sweepability.
+   * @param {LifetimeFilterOptions} [options={}] The options used to generate the filter function
+   * @returns {Function}
+   */
+  static filterByLiftetme({
```
kyranet, line 110:
```suggestion
  static filterByLifetime({
```

src/util/SweptCollection.js
```diff
@@ -0,0 +1,138 @@
...
+     * @type {?Function}
+     */
+    this.sweepFilter = sweepFilter;
+
+    /**
+     * The id of the interval being used to sweep.
+     * @type {?Number}
+     */
+    this.interval =
+      sweepInterval > 0 && sweepFilter ? setInterval(() => this.sweep(this.sweepFilter()), sweepInterval * 1000) : null;
```
kyranet, line 70: The interval should be unref'd, so it doesn't hang the process.
```suggestion
      sweepInterval > 0 && sweepFilter ? setInterval(() => this.sweep(this.sweepFilter()), sweepInterval * 1000).unref() : null;
```

src/util/SweptCollection.js
```diff
@@ -0,0 +1,138 @@
...
+    }
+    if (excludeFromSweep !== null && typeof excludeFromSweep !== 'function') {
+      throw new TypeError('INVALID_TYPE', 'excludeFromSweep', 'function');
+    }
+    return () => {
+      if (lifetime <= 0) return () => false;
+      const lifetimeMs = lifetime * 1000;
+      const now = Date.now();
+      return entry => {
+        if (excludeFromSweep?.(entry)) {
```
kyranet, line 127: Maybe make a special callback that always returns `false` when `excludeFromSweep` is null?

```javascript
if (excludeFromSweep === null) return () => false;
```

That way we can optimize this kind of operation, and we skip a check in this callback.

kyranet: You can also add this to line 123 with an `||` operator.

kyranet (changes requested):

src/util/SweptCollection.js
ckohen: This is the exact opposite of what its supposed to do, when null it should sweep everything. I've just changed the default to be `() => false` since it'll make it more clear

src/util/SweptCollection.js
ckohen: That is to match the current message sweep options and because sweeping on ms intervals seems impractical, also looks cleaner in end code.

vladfrangu (approved):

kyranet (approved):

typings/index.d.ts
```diff
@@ -1533,6 +1533,14 @@ export class StoreChannel extends GuildChannel {
   public type: 'GUILD_STORE';
 }
 
+export class SweptCollection<K, V> extends Collection<K, V> {
+  public constructor(options?: SweptCollectionOptions<K, V>, iterable?: Iterable<readonly [K, V]>);
+  public interval: number | null;
+  public sweepFilter: ((collection: this) => ((value: V) => boolean) | false) | null;
+
+  public static filterByLifetime(options?: LifetimeFilterOptions): () => (value: any) => boolean;
```
1Computer1, line 1541: Add K, V generics.

typings/index.d.ts
```diff
@@ -3450,6 +3458,12 @@ export type InviteScope =
   | 'gdm.join'
   | 'webhook.incoming';
 
+export interface LifetimeFilterOptions {
```
1Computer1, line 3461: Add K, V generics.

1Computer1: Maybe also consider adding the key to the predicates.

typings/index.d.ts
```diff
@@ -3972,6 +3986,11 @@ export interface StageInstanceEditOptions {
   privacyLevel?: PrivacyLevel | number;
 }
 
+export interface SweptCollectionOptions<K, V> {
+  sweepFilter?: (collection: SweptCollection<K, V>) => ((value: V) => boolean) | false;
```
1Computer1, line 3990: Maybe also consider adding the key to the filter.

src/util/SweptCollection.js
```diff
@@ -0,0 +1,110 @@
...
+
+    /**
+     * The id of the interval being used to sweep.
+     * @type {?Number}
+     */
+    this.interval =
+      sweepInterval > 0 && sweepFilter
+        ? setInterval(() => {
+            const sweepFn = this.sweepFilter(this);
+            if (sweepFn === false) return;
```
1Computer1, line 55: What's this for? If you want to allow not sweeping, maybe just check if the return is `null`/`undefined` instead.

src/util/SweptCollection.js
```diff
@@ -0,0 +1,110 @@
...
+   */
+
+  /**
+   * Create a sweepFilter function that uses a lifetime to determine sweepability.
+   * @param {LifetimeFilterOptions} [options={}] The options used to generate the filter function
+   * @returns {SweepFilter}
+   */
+  static filterByLifetime({
+    lifetime = 14400,
+    getComparisonTimestamp = e => e?.createdTimestamp,
```
1Computer1, line 80: I really don't think this should be the default.

src/util/Options.js
```diff
@@ -132,20 +140,37 @@ class Options extends null {
   }
 
   /**
-   * Create a cache factory using predefined limits.
-   * @param {Record<string, number>} [limits={}] Limits for structures.
+   * Create a cache factory using predefined settings to sweep or limit.
+   * @param {Record<string, SweptCollectionOptions|number>} [settings={}] Settings passed to the relevant constructor.
```
1Computer1, line 144: Should clarify what just a number does.

src/util/Options.js
```diff
@@ -132,20 +140,37 @@ class Options extends null {
...
 
     return manager => {
-      const limit = limits[manager.name];
-      if (limit === null || limit === undefined || limit === Infinity) {
+      const setting = settings[manager.name];
+      if (typeof setting === 'number' && setting !== Infinity) return LimitedCollection(setting);
+      if (
+        setting === null ||
+        typeof setting !== 'object' ||
+        (setting.sweepInterval ? setting.sweepInterval <= 0 || setting.sweepInterval === Infinity : true)
```
1Computer1, line 169: This can just be `setting?.sweepInterval == null || setting.sweepInterval <= 0 || setting.sweepInterval === Infinity`.

1Computer1 (changes requested):

src/util/SweptCollection.js
ckohen: Open to suggestions for a better default, don't know what it would be atm

src/util/SweptCollection.js
ckohen: I want it to be an intentional decision to not sweep, not an oh I forgot to add a return before the functoin, that's why my memory is going out of  control

src/util/Options.js
NotSugden, line 144:
```suggestion
   * @param {Object<string, SweptCollectionOptions|number>} [settings={}] Settings passed to the relevant constructor.
```
`Record` type doesn't exist in jsdoc

src/util/SweptCollection.js
```diff
@@ -0,0 +1,110 @@
...
+ * @param {SweptCollection} collection The collection being swept
+ * @returns {Function|false} Return `false` to skip sweeping, otherwise a function passed to `sweep()`,
+ * See {@link [Collection#sweep](https://discord.js.org/#/docs/collection/master/class/Collection?scrollTo=sweep)}
+ * for the definition of this function.
+ */
+
+/**
+ * Options for defining the behavior of a Swept Collection
+ * @typedef {Object} SweptCollectionOptions
+ * @property {Function|null} [sweepFilter=null] A function run every `sweepInterval` to determine how to sweep
```
NotSugden, line 17:
```suggestion
 * @property {?SweepFilter} [sweepFilter=null] A function run every `sweepInterval` to determine how to sweep
```

src/util/SweptCollection.js
```diff
@@ -0,0 +1,110 @@
...
+ * @property {number} [sweepInterval=3600] How frequently, in seconds, to sweep the collection.
+ */
+
+/**
+ * A Collection which holds a max amount of entries and sweeps periodically.
+ * @extends {Collection}
+ * @param {SweptCollectionOptions} [options={}] Options for constructing the swept collection.
+ * @param {Iterable} [iterable=null] Optional entries passed to the Map constructor.
```
NotSugden, line 25: `Iterable` needs a typedef

src/util/SweptCollection.js
```diff
@@ -0,0 +1,110 @@
...
+    if (sweepFilter !== null && typeof sweepFilter !== 'function') {
+      throw new TypeError('INVALID_TYPE', 'sweepFunction', 'function');
+    }
+    if (typeof sweepInterval !== 'number') throw new TypeError('INVALID_TYPE', 'sweepInterval', 'number');
+
+    super(iterable);
+
+    /**
+     * A function called every sweep interval that returns a function passed to `sweep`
+     * @type {?Function}
```
NotSugden, line 43:
```suggestion
     * @type {?SweepFilter}
```

src/util/SweptCollection.js
```diff
@@ -0,0 +1,110 @@
...
+     */
+    this.sweepFilter = sweepFilter;
+
+    /**
+     * The id of the interval being used to sweep.
+     * @type {?Number}
```
NotSugden, line 49:
```suggestion
     * @type {?number}
```
why just the id? why not the interval object?

src/util/SweptCollection.js
```diff
@@ -0,0 +1,110 @@
...
+          }, sweepInterval * 1000).unref()
+        : null;
+  }
+
+  /**
+   * Options for generating a filter function based on lifetime
+   * @typedef {Object} LifetimeFilterOptions
+   * @property {number} [lifetime=14400] How long an entry should stay in the collection
+   * before it is considered sweepable
+   * @property {Function} [getComparisonTimestamp=`e => e.createdTimestamp`] A function that takes an entry
```
NotSugden, line 67:
```suggestion
   * @property {Function} [getComparisonTimestamp=`e => e?.createdTimestamp`] A function that takes an entry
```
whats this like on the docs website?

src/util/SweptCollection.js
```diff
@@ -0,0 +1,110 @@
...
+  } = {}) {
+    if (typeof lifetime !== 'number') throw new TypeError('INVALID_TYPE', 'lifetime', 'number');
+    if (typeof getComparisonTimestamp !== 'function') {
+      throw new TypeError('INVALID_TYPE', 'getComparisonTimestamp', 'function');
+    }
+    if (typeof excludeFromSweep !== 'function') {
+      throw new TypeError('INVALID_TYPE', 'excludeFromSweep', 'function');
+    }
+    return () => {
+      if (lifetime <= 0) return () => false;
```
NotSugden, line 91: can `false` just be returned here instead of being wrapped in a function? or am i mis understanding how the filter works

src/util/SweptCollection.js
ckohen: it *is* the object, I just still associate the return value with the id 😆

src/util/SweptCollection.js
ckohen: Old site, but this
image

NotSugden: looks a bit gross, perhaps no backticks?

iCrawl: No backticks would apply the users font 👎

src/util/SweptCollection.js
ckohen: I think I wrote this before I allowed a non function return (which is now null instead of false)...yes I should use it.

src/util/SweptCollection.js
1Computer1: Would prefer no default at all, but whatever works.

typings/index.d.ts
```diff
@@ -1527,6 +1527,16 @@ export class StoreChannel extends GuildChannel {
   public type: 'GUILD_STORE';
 }
 
+export class SweptCollection<K, V> extends Collection<K, V> {
+  public constructor(options?: SweptCollectionOptions<K, V>, iterable?: Iterable<readonly [K, V]>);
+  public interval: number | null;
+  public sweepFilter: ((collection: this) => ((value: V, key: K, collection: this) => boolean) | null) | null;
```
1Computer1, line 1533: Should make this into a type, `SweepFilter` or something like that.

typings/index.d.ts
```diff
@@ -1527,6 +1527,16 @@ export class StoreChannel extends GuildChannel {
...
+
+  public static filterByLifetime<K, V>(
+    options?: LifetimeFilterOptions<K, V>,
+  ): () => (value: V, collection: SweptCollection<K, V>) => boolean;
```
1Computer1, line 1537: Wrong type, missing the key.

typings/index.d.ts
```diff
@@ -3999,6 +4017,13 @@ export interface StageInstanceEditOptions {
   privacyLevel?: PrivacyLevel | number;
 }
 
+export interface SweptCollectionOptions<K, V> {
+  sweepFilter?: (
+    collection: SweptCollection<K, V>,
+  ) => ((value: V, key: K, collection: SweptCollection<K, V>) => boolean) | false;
```
1Computer1, line 4023: This is still false instead of null.

src/util/Options.js
```diff
@@ -134,20 +143,51 @@ class Options extends null {
...
     const LimitedCollection = require('./LimitedCollection');
+    const SweptCollection = require('./SweptCollection');
 
     return manager => {
-      const limit = limits[manager.name];
-      if (limit === null || limit === undefined || limit === Infinity) {
+      const setting = settings[manager.name];
+      if (typeof setting === 'number' && setting !== Infinity) return new LimitedCollection(setting);
+      if (
+        setting?.sweepInterval === (undefined || null) ||
```
1Computer1, line 184: This doesn't work.

1Computer1 (changes requested):

src/util/Options.js
ckohen: It doesn't? In my testing it did

ckohen: lol nope, I'm dumb, got falses and trues mixed up, i'll just disable the eslint rule and == check null

src/util/SweptCollection.js
ckohen: Can this be done from mdn docs maybe?

src/managers/GuildEmojiRoleManager.js
```diff
@@ -97,7 +97,7 @@ class GuildEmojiRoleManager extends DataManager {
 
   clone() {
     const clone = new this.constructor(this.emoji);
-    clone._patch(this.cache.keyArray().slice());
+    clone._patch([...this.cache.keys()].slice());
```
shinotheshino, line 100: This `.slice()` is redundant

shinotheshino (changes requested):

src/util/SweptCollection.js
```diff
@@ -0,0 +1,110 @@
...
+     * A function called every sweep interval that returns a function passed to `sweep`
+     * @type {?SweepFilter}
+     */
+    this.sweepFilter = sweepFilter;
+
+    /**
+     * The id of the interval being used to sweep.
+     * @type {?Timeout}
+     */
+    this.interval =
```
SpaceEEC, line 51: These do not seem to be cleared anywhere to me.

Woulnd't that cause leaks with managers that are being discarded at some point (e.g. MessageManager after a channel was deleted)?

SpaceEEC (changes requested):

src/util/SweptCollection.js
ckohen: They are cleared in literally every way possible now, coming at the cost of a breaking change

SpaceEEC (approved):

iCrawl (approved):

1Computer1 (approved):

src/managers/CachedManager.js
```diff
@@ -13,6 +14,18 @@ class CachedManager extends DataManager {
 
     Object.defineProperty(this, '_cache', { value: this.client.options.makeCache(this.constructor, this.holds) });
 
+    let cleanup = this._cache[_cleanupSymbol];
+    const cacheType = this._cache.constructor.name;
```
vladfrangu, line 18: This doesn't seem right... `this._cache` is a form of Collection. Did you mean this.Holds instead?

src/managers/GuildEmojiRoleManager.js
```diff
@@ -72,7 +72,10 @@ class GuildEmojiRoleManager extends DataManager {
       resolvedRoleIds.push(roleId);
     }
 
-    const newRoles = this.cache.keyArray().filter(id => !resolvedRoleIds.includes(id));
+    const newRoles = this.cache
+      .keys()
```
vladfrangu, line 76: Wait... isn't this.cache a collection? keys is an iterator, not a method that returns all keys in an array form

src/managers/GuildEmojiRoleManager.js
```diff
@@ -97,7 +100,7 @@ class GuildEmojiRoleManager extends DataManager {
 
   clone() {
     const clone = new this.constructor(this.emoji);
-    clone._patch(this.cache.keyArray().slice());
+    clone._patch(this.cache.keys().slice());
```
vladfrangu, line 103: Ditto

src/managers/GuildMemberRoleManager.js
```diff
@@ -172,7 +172,7 @@ class GuildMemberRoleManager extends DataManager {
 
   clone() {
     const clone = new this.constructor(this.member);
-    clone.member._roles = [...this.cache.keyArray()];
+    clone.member._roles = this.cache.keys().slice();
```
vladfrangu, line 175: Ditto

src/util/Options.js
```diff
@@ -134,20 +143,52 @@ class Options extends null {
...
+   *     sweepInterval: 300,
+   *     sweepFilter: SweptCollection.filterByLifetime({
+   *       lifetime: 1800,
+   *       getComparisonTimestamp: e => e.editedTimestamp ?? e.createdTimestamp,
+   *     })
+   *   }
+   * });
    */
-  static cacheWithLimits(limits = {}) {
+  static cacheSome(settings = {}) {
```
vladfrangu, line 175: Why the rename? 👀

vladfrangu (changes requested):

src/managers/CachedManager.js
ckohen: The name is probably misleading, when running it I liked `Garbage Collection completed on ThreadManager, which held a SweptCollection`, but open to changes in what this message is.

I suppose it would maybe make sense to say `a SweptCollection of ThreadChannels`?

src/managers/GuildEmojiRoleManager.js
ckohen: I'm kinda suprised I didn't crash while testing this tbh

src/util/Options.js
ckohen: Sweeping isn't really a limit, and it will make more sense if there's any future built in cache types.

src/managers/CachedManager.js
vladfrangu: Yeah, personally I'd rather it say what it holds, not what type of collection it is (plus it's always a SweptCollection currently)

`GC completed on ${this.constructor.name}, which had a ${this._cache.constructor.name} of ${this.Holds.name}` or similar

src/util/Options.js
ckohen: nvm, reverting this for now

PR #6112, opened 2021-07-13
feat(REST): append additional information to the required User Agent

QSmally:

This PR allows you to append information about the bot (like a website, version) to the end of the (already required) user agent. As specified in the [API documentations](https://discord.com/developers/docs/reference#user-agent), users should be able to do so and are encouraged to do it.

Discord.js users could alternatively modify the constants file, where the user agent string is defined, but it can lead to them not following the API specifications.

After having a talk with a few people, I decided to implement a `userAgentSuffix` option to `ClientOptions` for people to do this. Types are already updated. An example of the usage is down below.

```js
const { Client } = require("discord.js");
const botClient = new Client({
    userAgentSuffix: [`MyBot/${/* botversion */}`]
});
```

The implementation of this PR would append the array, joined by `, `, to the end of the already-default user agent.

* Default: `DiscordBot ($djsurl, $djsversion) Node.js/$nodeversion`
* With suffix: `DiscordBot ($djsurl, $djsversion) Node.js/$nodeversion, MyBot/$botversion`

### DIFF
--- a/src/rest/APIRequest.js
+++ b/src/rest/APIRequest.js
@@ -18,7 +18,7 @@ class APIRequest {
     this.retries = 0;
 
     const userAgentSuffix = this.client.options.userAgentSuffix;
-    this.fullUserAgent = `${UserAgent}${userAgentSuffix.length && `, ${userAgentSuffix.join(', ')}`}`;
+    this.fullUserAgent = `${UserAgent}${userAgentSuffix.length ? `, ${userAgentSuffix.join(', ')}` : ''}`;
 
     let queryString = '';
     if (options.query) {
--- a/src/rest/APIRequest.js
+++ b/src/rest/APIRequest.js
@@ -17,6 +17,9 @@ class APIRequest {
     this.options = options;
     this.retries = 0;
 
+    const userAgentSuffix = this.client.options.userAgentSuffix;
+    this.fullUserAgent = `${UserAgent}${userAgentSuffix.length && `, ${userAgentSuffix.join(", ")}`}`;
+
     let queryString = '';
     if (options.query) {
       const query = Object.entries(options.query)
@@ -33,11 +36,14 @@ class APIRequest {
         ? this.client.options.http.api
         : `${this.client.options.http.api}/v${this.client.options.http.version}`;
     const url = API + this.path;
-    let headers = { ...this.client.options.http.headers };
+
+    let headers = {
+      ...this.client.options.http.headers,
+      'User-Agent': this.fullUserAgent
+    };
 
     if (this.options.auth !== false) headers.Authorization = this.rest.getAuth();
     if (this.options.reason) headers['X-Audit-Log-Reason'] = encodeURIComponent(this.options.reason);
-    headers['User-Agent'] = UserAgent;
     if (this.options.headers) headers = Object.assign(headers, this.options.headers);
 
     let body;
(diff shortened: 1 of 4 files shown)


DTrombett:
Why not making it a string instead of an array?

QSmally:
> Why not making it a string instead of an array?

It takes the approach of [RFC7231](https://tools.ietf.org/html/rfc7231#section-5.5.3). Each segment is divided, and it would be useful if those are reflected in the client option as well.

DTrombett:
I think the validateOptions method should be updated as well
https://github.com/discordjs/discord.js/blob/72473e0bf73187a0b3f5bac3ebe4ab273be68d90/src/client/Client.js#L468

QSmally:
> I think the validateOptions method should be updated as well
> 
> https://github.com/discordjs/discord.js/blob/72473e0bf73187a0b3f5bac3ebe4ab273be68d90/src/client/Client.js#L468

Ahh, my bad! I'll fix it in a moment

src/rest/APIRequest.js
```diff
@@ -17,6 +17,9 @@ class APIRequest {
     this.options = options;
     this.retries = 0;
 
+    const userAgentSuffix = this.client.options.userAgentSuffix;
+    this.fullUserAgent = `${UserAgent}${userAgentSuffix.length && `, ${userAgentSuffix.join(", ")}`}`;
```
NotSugden:
```suggestion
    this.fullUserAgent = `${UserAgent}${userAgentSuffix.length ? `, ${userAgentSuffix.join(', ')}` : ''}`;
```

QSmally: I considered doing this. However, further down the line, I saw the same thing with `this.path`. Thought it would make the most sense if it were to be consistent.

almeidx: > I considered doing this. However, further down the line, I saw the same thing with `this.path`. Thought it would make the most sense if it were to be consistent.

It's a different implementation. For `queryString` to be falsy, it has to be an empty string. As for `userAgentSuffix.length`, it has to be 0, resulting in `${UserAgent}0`

NotSugden: guess i forgot to add an actual comment, but yes pretty much what @almeidx said, in this case it would add a 0 after the user agent

QSmally: Ah, my bad! I didn't notice. Fix will be pushed soon™

SpaceEEC (approved):

src/rest/APIRequest.js
```diff
@@ -17,6 +17,9 @@ class APIRequest {
     this.options = options;
     this.retries = 0;
 
+    const userAgentSuffix = this.client.options.userAgentSuffix;
```
vladfrangu, line 20:
```suggestion
    const { userAgentSuffix } = this.client.options;
```

vladfrangu (approved):
Optional destructuring, but otherwise LGTM

iCrawl (approved):

kyranet (approved):
I'm not a fan of this code, since it's the checks and user agent are done in every request rather than doing it only once (maybe in the RESTManager?).

I won't block, we can seek an alternative with a smaller performance impact later in another PR.

QSmally:
@kyranet - I did not notice that the RESTManager was created on every request. Of course, we can look into quite possibly creating the final user agent at startup of the bot, rather than on every request.

PR #6113, opened 2021-07-13
refactor: remove timer utilities from Client

kyranet:
**Please describe the changes this PR makes and why it should be merged:**

:warning: This PR removes the following methods:

- `BaseClient#setTimeout`
- `BaseClient#setInterval`
- `BaseClient#setImmediate`
- `BaseClient#clearTimeout`
- `BaseClient#clearInterval`
- `BaseClient#clearImmediate`

**Why?**

The aforementioned methods existed to make timers not hang the process when it's destroyed and nothing else kept the process on (from the user's side). Since all timers have an `unref` method ([`Immediate#unref`](https://nodejs.org/dist/latest-v16.x/docs/api/timers.html#timers_timeout_unref), [`Immediate#unref`](https://nodejs.org/dist/latest-v16.x/docs/api/timers.html#timers_immediate_unref)), there is not longer a need to keep and track those timers, which can potentially lead to memory leaks.

**Status and versioning classification:**

- Code changes have been tested against the Discord API, or there are no code changes
- I know how to update typings and have done so, or typings don't need updating
- This PR changes the library's interface (methods or parameters added)
- This PR includes breaking changes (methods removed or renamed, parameters moved or removed)

PR #6114, opened 2021-07-14
refactor: remove typing caching

kyranet:

- Added `Typing` structure.
- Renamed `TextBasedChannel#startTyping` to `sendTyping`.
  Additionally, this will not longer run a 9s interval sending additional requests, but instead, it will do so only once without any internal tracking.
- Removed `TextBasedChannel#stopTyping` as it relied on internal tracking.
- Removed `TextBasedChannel#typing` as it relied on internal tracking.
- Removed `TextBasedChannel#typingCount` as it relied on internal tracking.
- Removed `TextBasedChannel#_typing`.
- Changed `typingStart` event to take a `Typing` structure.
- Removed many timers, making typing events more scalable than ever.
- Removed `User#typingIn` as it relied on internal tracking.
- Removed `User#typingSinceIn` as it relied on internal tracking.
- Removed `User#typingDurationIn` as it relied on internal tracking.
- Removed `TypingData` type.

This PR complements #6113 by removing sparse timers, which will allow `Client#destroy` to work as intended.

As a side note, this PR makes channel instances a lot lighter, reducing memory usage by 140 MB per 1 million channels while also reducing the main thread's load by not registering potentially thousands of timers, which results on large delays in when user timers start running, promises start resolving/rejecting, etc.

### DIFF
--- a/src/structures/interfaces/TextBasedChannel.js
+++ b/src/structures/interfaces/TextBasedChannel.js
@@ -6,7 +6,7 @@ const MessagePayload = require('../MessagePayload');
 const SnowflakeUtil = require('../../util/SnowflakeUtil');
 const Collection = require('../../util/Collection');
 const { InteractionTypes } = require('../../util/Constants');
-const { RangeError, TypeError, Error } = require('../../errors');
+const { TypeError, Error } = require('../../errors');
 const InteractionCollector = require('../InteractionCollector');
 
 /**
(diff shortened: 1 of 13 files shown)


src/structures/interfaces/TextBasedChannel.js
```diff
@@ -172,88 +172,14 @@ class TextBasedChannel {
   }
 
   /**
-   * Starts a typing indicator in the channel.
-   * @param {number} [count=1] The number of times startTyping should be considered to have been called
-   * @returns {Promise} Resolves once the bot stops typing gracefully, or rejects when an error occurs
+   * Sends a typing indicator in the channel.
```
advaith1, line 175: should probably mention how long it lasts

typings/index.d.ts
```diff
@@ -1646,6 +1646,25 @@ export class ThreadMemberFlags extends BitField<ThreadMemberFlagsString> {
...
+  public user: PartialUser;
+  public startedTimestamp: number;
+  public readonly startedAt: Date;
+  public readonly guild: Guild | null;
+  public readonly member: GuildMember | null;
+  public inGuild(): this is this & {
+    channel: TextChannel | NewsChannel | ThreadChannel;
+    readonly guild: Guild;
+    readonly member: GuildMember;
+  };
```
DTrombett, line 1665: The `member` getter can returns `null` when the channel is in a guild if it's not cached.
```suggestion
  public inGuild(): this is this & {
    channel: TextChannel | NewsChannel | ThreadChannel;
    readonly guild: Guild;
  };
```

src/structures/interfaces/TextBasedChannel.js
kyranet: How long it lasts is not documented in the [docs](https://discord.com/developers/docs/resources/channel#trigger-typing-indicator).

src/structures/interfaces/TextBasedChannel.js
```diff
@@ -172,88 +172,14 @@ class TextBasedChannel {
...
-   * @type {number}
-   * @readonly
+   * // Start typing in a channel
+   * channel.sendTyping();
    */
-  get typingCount() {
-    return this.client.user._typing.get(this.id)?.count ?? 0;
+  async sendTyping() {
+    await this.client.api.channels(this.id).typing.post();
   }
```
DTrombett, line 183: This is missing the return. Also why is there async / await if the promise is returned?
```suggestion
  sendTyping() {
    return this.client.api.channels(this.id).typing.post();
  }
```

typings/index.d.ts
```diff
@@ -2436,8 +2449,7 @@ export interface TextBasedChannelFields extends PartialTextBasedChannelFields {
     options?: InteractionCollectorOptions<T>,
   ): InteractionCollector<T>;
   createMessageCollector(options?: MessageCollectorOptions): MessageCollector;
-  startTyping(count?: number): Promise<void>;
-  stopTyping(force?: boolean): void;
+  sendTyping(count?: number): Promise<void>;
```
DTrombett, line 2452: The `count` parameter doesn't exist anymore.

```suggestion
  sendTyping(): Promise<void>;
```

src/structures/interfaces/TextBasedChannel.js
kyranet: It returns a `Buffer` instance, which may not be elegant.
image

DTrombett: Ohh, so the async / await without returning is to returns the Promise without the result of the post call? Nice!

almeidx:
This error message can be removed

https://github.com/discordjs/discord.js/blob/98c60789a2bc38a9939b2493c3e9346eb0a0446e/src/errors/Messages.js#L81

iCrawl (approved):

vladfrangu (approved):

SpaceEEC (approved):
🎉

PR #6122, opened 2021-07-16
Fixed deleted flag not being updated

da-the-dev:
So I had noticed that the "deleted" flag wasn't updating when a message was actually deleted, so here I've added a small change. Hope I've fixed it correctly, after all this is my first ever change proposal.

### DIFF
--- a/src/structures/Message.js
+++ b/src/structures/Message.js
@@ -577,6 +577,7 @@ class Message extends Base {
     if (typeof options !== 'object') return Promise.reject(new TypeError('INVALID_TYPE', 'options', 'object', true));
     const { timeout = 0, reason } = options;
     if (timeout <= 0) {
+      this.deleted = true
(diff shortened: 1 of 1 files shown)


DTrombett:
1) This is the wrong place where the assignment should be added. It should be added in [`MessageManager#delete()`](https://github.com/discordjs/discord.js/blob/5ca97c93515d4dfaa2b4951a020abc000115ed4d/src/managers/MessageManager.js#L205).
2) This is automatically done when the bot receives the gateway event [`MESSAGE_DELETE`](https://github.com/discordjs/discord.js/blob/5ca97c93515d4dfaa2b4951a020abc000115ed4d/src/client/actions/MessageDelete.js#L14) so when this event (`messageDelete`) is received the message is automatically flagged as deleted.

iCrawl:
We don't accept PRs to v12.

PR #6127, opened 2021-07-16
added User#asMention

tiagodinis33:
I added User#asMention that returns a mention that can be displayed on a message, because its more intuitive
example:
```js
msg.channel.send(`Hi ${msg.author.asMention}!`)
```
will send:
```Hi @tiagodinis33!```

### DIFF
--- a/src/structures/User.js
+++ b/src/structures/User.js
@@ -5,6 +5,7 @@ const TextBasedChannel = require('./interfaces/TextBasedChannel');
 const { Error } = require('../errors');
 const SnowflakeUtil = require('../util/SnowflakeUtil');
 const UserFlags = require('../util/UserFlags');
+const Message = require('./Message');
 
 /**
  * Represents a user on Discord.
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: src/structures/User.js [hunk 1]
```diff
@@ -91,9 +92,14 @@ class User extends Base {
        * @type {?UserFlags}
        */
       this.flags = new UserFlags(data.public_flags);
+      
     }
   }
 
+  get asMention(){
+    return `<@${this.id}>`;
+  }
```
shinotheshino, line 101: Isn't this the exact same as `User#toString()`?

GoldenAngel2:
`User.toString()` does this.

src/structures/User.js
tiagodinis33: wait, toString already do that?

tiagodinis33: i thougth it would return the object as string

GoldenAngel2: Yes, it does.

shinotheshino: https://github.com/discordjs/discord.js/blob/5ca97c93515d4dfaa2b4951a020abc000115ed4d/src/structures/User.js#L248

tiagodinis33: i mean, return a json

tiagodinis33: but anyway

tiagodinis33: thx for your attention

PR #6154, opened 2021-07-20
refactor(OptionResolver): accept single type instead of an array

vaporox:

The reason why it accepted an array was because sub-commands and sub-command groups used to share the same method, however now that those are split up it is no longer needed.


vladfrangu (approved):

kyranet (approved):

iCrawl (approved):

Debabelkacem:
i think so

PR #6156, opened 2021-07-20
refactor(CommandInteractionOptionResolver): add readonly data property

### LINKED ISSUE #6143: Have stored values available for access for `CommandInteractionOptionResolver`
Hi hi! Thank you for the `CommandInteractionOptionResolver`!

I have a use case for some of my Slash Commands where no options are required but sending across no options results in behaviour still (and not an error). Previously, I would be able to do `if (<CommandInteraction>.options.size === 0) {...}`, but it seems now that I must check _each and every parameter_ to ensure that they are all not present to ensure this behaviour works.

Another use case of I have of my Slash Commands is "repeated" options. For example:
`/ignore evidence add evidence-1: ... evidence-2: ... evidence-3: ...`

I would have parsed this in the past by running `.filter(({ name }) => name.startsWith("evidence")).map(({ value }) => value)` but now I must check each and every option individually!

I have more use cases that the implementation of `CommandInteractionOptionResolver` seems to make harder but...

What I'm trying to get at is that whilst `CommandInteractionOptionResolver` beautifully solves the issue of easily getting your arguments, this implementation alone doesn't allow you to manually parse the data should you have an interesting edge case or use case that doesn't rely on option names alone, for which I described 2 above. I do not wish to check if all my parameters exist to then enact out on some behaviour, but would rather just have the array or `Collection` back so I can simply check the length or size respectively or manipulate to my will.

I'd love an array or `Collection` of options so I can easily run `<Collection>.size` again or `<Array>.length` on to check how many options are present or to be able to iterate over my similarly-named options. I actually see there's a private property `<CommandInteractionOptionResolver>._options` which looks like it solves this, but alas, it is private.

The only way right now is to work with what I have, which means longer and _meh_ code which could easily be solved by supplying the options in an array or `Collection` form again:

For my first described use case, I would have to check if all the parameters do not exist. This means running something like:
```JavaScript
const option1 = interaction.options.getInteger("option1");
const option2 = interaction.options.getInteger("option2");
const option3 = interaction.options.getInteger("option3");
const option4 = interaction.options.getInteger("option4");

if ([option1, option2, option3, option4].every(option => option === null)) {
  // All options have not been sent. Proceed with no-option behaviour here!
}
```
The more options, the better!

For my second described use case, I would again have to check all parameters instead of being able to loop through them which results in this:
```JavaScript
const evidence1 = interaction.options.getString("evidence-1");
const evidence2 = interaction.options.getString("evidence-2");
const evidence3 = interaction.options.getString("evidence-3");
const evidence4 = interaction.options.getString("evidence-4");
const evidence5 = interaction.options.getString("evidence-5");
const evidence6 = interaction.options.getString("evidence-6");
const evidence7 = interaction.options.getString("evidence-7");
const evidence8 = interaction.options.getString("evidence-8");
const evidence9 = interaction.options.getString("evidence-9");
const evidence10 = interaction.options.getString("evidence-10");

const evidenceToAdd = [evidence1, evidence2, evidence3, evidence4, evidence5, evidence6, evidence7, evidence8, evidence9, evidence10].filter(evidence => evidence !== null);
```

As a reminder, this is how I did it in the past on the options: `.filter(({ name }) => name.startsWith("evidence")).map(({ value }) => value)`

**Additional context**
N/A

shinotheshino:
This moves the `_options` property to the public `data` property as a read-only array, which allows custom usage of options.

This closes #6143.

### DIFF
--- a/src/structures/CommandInteractionOptionResolver.js
+++ b/src/structures/CommandInteractionOptionResolver.js
@@ -16,13 +16,6 @@ class CommandInteractionOptionResolver {
     Object.defineProperty(this, 'client', { value: client });
 
     /**
-     * The interaction options array.
-     * @type {CommandInteractionOption[]}
-     * @private
-     */
-    this._options = options ?? [];
-
-    /**
      * The name of the sub-command group.
      * @type {?string}
      * @private
@@ -35,14 +28,24 @@ class CommandInteractionOptionResolver {
      * @private
      */
     this._subCommand = null;
-    if (this._options[0]?.type === 'SUB_COMMAND_GROUP') {
-      this._group = this._options[0].name;
-      this._options = this._options[0].options ?? [];
+
+    let data = options;
+    if (data[0]?.type === 'SUB_COMMAND_GROUP') {
+      this._group = data[0].name;
+      data = data[0].options ?? [];
     }
-    if (this._options[0]?.type === 'SUB_COMMAND') {
-      this._subCommand = this._options[0].name;
-      this._options = this._options[0].options ?? [];
+    if (data[0]?.type === 'SUB_COMMAND') {
+      this._subCommand = data[0].name;
+      data = data[0].options ?? [];
     }
+
+    /**
+     * The interaction options array.
+     * @name CommandInteractionOptionResolver#data
+     * @type {ReadonlyArray<CommandInteractionOption>}
+     * @readonly
+     */
+    Object.defineProperty(this, 'data', { value: Object.freeze(data) });
   }
 
   /**
@@ -52,7 +55,7 @@ class CommandInteractionOptionResolver {
    * @returns {?CommandInteractionOption} The option, if found.
    */
   get(name, required = false) {
-    const option = this._options.find(opt => opt.name === name);
+    const option = this.data.find(opt => opt.name === name);
     if (!option) {
       if (required) {
         throw new TypeError('COMMAND_INTERACTION_OPTION_NOT_FOUND', name);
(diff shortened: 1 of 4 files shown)


iCrawl (approved):

vladfrangu (approved):

kyranet (approved):

src/structures/CommandInteractionOptionResolver.js
```diff
@@ -35,14 +28,24 @@ class CommandInteractionOptionResolver {
...
+      data = data[0].options ?? [];
     }
+
+    /**
+     * The interaction options array.
+     * @name CommandInteractionOptionResolver#data
+     * @type {ReadonlyArray<CommandInteractionOption>}
+     * @readonly
+     */
+    Object.defineProperty(this, 'data', { value: Object.freeze(data) });
```
shinotheshino, line 48: Should this be modified to have `data` be the top-level options array? For cases where people want to access the original interaction option array with the sub-command included.

shinotheshino: @iCrawl @monbrey

monbrey: 🤔 Excellent question. I personally lean to yes, I'd like to see discord.js providing the completely original, unparsed data.

This might mean you keep a private `_data` / `_options` property too if you want to do keep the hoisted sub-command options somewhere.

shinotheshino: Per [discussion](https://discord.com/channels/222078108977594368/682166281826598932/867182323316424774), this will stay as-is in this PR.

SpaceEEC (approved):

PR #6162, opened 2021-07-21
fix(Typings): type attachments to InteractionUpdateOptions

Milo123459:

Currently, if you want to remove attachments from a message you have to use `message.removeAttachments()`, and if you need to edit the content as well that is 2 API calls. Assigning `attachments: []` does work, but sparks errors in TypeScript that require `// @ts-ignore` - which is generally considered a bad practice. This PR fixes the typings allowing you to assign `attachments: []` without having to use ts-ignore.

### DIFF
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -3810,7 +3810,7 @@ export type MessageComponentType = keyof typeof MessageComponentTypes;
 export type MessageComponentTypeResolvable = MessageComponentType | MessageComponentTypes;
 
 export interface MessageEditOptions {
-  attachments?: Array<MessageAttachment | never>;
+  attachments?: (MessageAttachment | never)[];
   content?: string | null;
   embeds?: (MessageEmbed | MessageEmbedOptions)[] | null;
   files?: (FileOptions | BufferResolvable | Stream | MessageAttachment)[];
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -3694,7 +3694,7 @@ export type InteractionResponseType = keyof typeof InteractionResponseTypes;
 
 export type InteractionType = keyof typeof InteractionTypes;
 
-export type InteractionUpdateOptions = Omit<InteractionReplyOptions, 'ephemeral'>;
+export type InteractionUpdateOptions = MessageEditOptions;
 
 export type IntentsString =
   | 'GUILDS'
@@ -3810,7 +3810,7 @@ export type MessageComponentType = keyof typeof MessageComponentTypes;
 export type MessageComponentTypeResolvable = MessageComponentType | MessageComponentTypes;
 
 export interface MessageEditOptions {
-  attachments?: (MessageAttachment | never)[];
+  attachments?: MessageAttachment[];
   content?: string | null;
   embeds?: (MessageEmbed | MessageEmbedOptions)[] | null;
   files?: (FileOptions | BufferResolvable | Stream | MessageAttachment)[];
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -3810,7 +3810,7 @@ export type MessageComponentType = keyof typeof MessageComponentTypes;
 export type MessageComponentTypeResolvable = MessageComponentType | MessageComponentTypes;
 
 export interface MessageEditOptions {
-  attachments?: MessageAttachment[];
+  attachments?: Array<MessageAttachment | never>;
   content?: string | null;
   embeds?: (MessageEmbed | MessageEmbedOptions)[] | null;
   files?: (FileOptions | BufferResolvable | Stream | MessageAttachment)[];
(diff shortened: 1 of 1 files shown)


DTrombett:
```ts
const attachments: MessageAttachment[] = [];
msg.edit({ attachments, content: 'abc' });
```
This works without using TS ignore...

Milo123459:
If that works then logically speaking I would expect `attachments: [] as MessageAttachment[]` to work but I still get the same error if I just gave it a blank array

typings/index.d.ts
```diff
@@ -3810,7 +3810,7 @@ export type MessageComponentType = keyof typeof MessageComponentTypes;
 export type MessageComponentTypeResolvable = MessageComponentType | MessageComponentTypes;
 
 export interface MessageEditOptions {
-  attachments?: MessageAttachment[];
+  attachments?: Array<MessageAttachment | never>;
```
DTrombett, line 3813:
```suggestion
  attachments?: (MessageAttachment | never)[];
```
Djs doesn't use the `Array<Type>` form.

Milo123459:
Also, this error also pops up when using `<MessageComponentInteraction>.update({attachments: []})` - I can't seem to find an interface that corrosponds to the `attachments` field. It's all just chains of Omit's

rmunro-uts:
I can't replicate any error with this.
`message.edit({ attachments: [] });` isn't throwing any TypeScript errors for me.

monbrey:
I can't replicate any issue with `message.edit({ attachments: [] })`, TypeScript isnt complaining.

Milo123459:
Odd. I also get this error when trying to update a MesaageComponentInteraction with no attachments.

monbrey:
`attachments` is not a valid property for `InteractionUpdateOptions`. We briefly discussed this in dev chats but never applied a fix. Ignoring the `Omit` typings, it's basically a chain.
 
`InteractionUpdateOptions > InteractionReplyOptions > WebhookMessageOptions > MessageOptions`

This isn't quite what it should be; `InteractionUpdateOptions` should probably be something more like:
```ts
export interface InteractionUpdateOptions extends Omit<WebhookEditMessageOptions, 'username' | 'avatarURL'> {
  fetchReply?: boolean;
}```
and WebhookEditMessageOptions probably needs to support editing attachments, so:
```ts
export interface WebhookEditMessageOptions extends Pick<
  WebhookMessageOptions,
  'content' | 'embeds' | 'files' | 'allowedMentions' | 'components'
> {
  attachments: MessageAttachment[];
}```

**Comment by Milo123459** [comment]
Picking from MessageOptions makes more sense. I'll try to implement a fix.

**Comment by Milo123459** [comment]
```
export interface WebhookEditMessageOptions extends Pick<
  WebhookMessageOptions,
  'content' | 'embeds' | 'files' | 'allowedMentions' | 'components'
> {
  attachments: MessageAttachment[];
}
```
From what you said, this implies that when you are assigning `attachments` it'll cause the same error this is PR trying to fix. I'll go with `attachments: (MessageAttachment | never)[]` as that's what this PR is fixing.

**Comment by monbrey** [comment]
You're referring to an error I was unable to replicate on `message.edit`. I don't believe typing it as `never` is the correct fix for the issue you described, which does exist on `interaction.update` as no typings for `attachments` exist as all.

**Comment by Milo123459** [comment]
So all this has to do is type `attachments` for `InteractionUpdateOptions`?

**Comment by monbrey** [comment]
> 
> 
> So all this has to do is type `attachments` for `InteractionUpdateOptions`?

I think so, yes. But as its a long chain of Omits as you pointed out, finding the right place/editing that chain is the tricky part.

Otherwise if you are running into the issue on `message.edit({ attachments: [] })`, where those typings do already exist, please provide a reproducible code sample because I wasn't able to get any errors.

**Comment by Milo123459** [comment]
Yea, I think it was something to do with Yarn's PnP and my workspace typescript version, that one seems to of been fixed, this PR should now focus on fixing `InteractionUpdateOptions`

File: typings/index.d.ts [hunk 2]
```diff
@@ -3694,7 +3694,7 @@ export type InteractionResponseType = keyof typeof InteractionResponseTypes;
 
 export type InteractionType = keyof typeof InteractionTypes;
 
-export type InteractionUpdateOptions = Omit<InteractionReplyOptions, 'ephemeral'>;
+export type InteractionUpdateOptions = MessageEditOptions;
```
monbrey, line 3697: This is now missing `fetchReply`

monbrey (changes requested):

typings/index.d.ts
Milo123459: Oops

Milo123459:
Anything else need to be changed?

SpaceEEC (approved):

vladfrangu (approved):

iCrawl (approved):

PR #6163, opened 2021-07-21
fix(Typings): Missing user definition

iAmZiNc:

On guildBanAdd + guildBanRemove events, the "user" property is not included in the typings.


iAmZiNc:
Oh goodness, why did it include all those merge commits?

DTrombett:
There's no `user` parameter passed to those events..

iAmZiNc:
> 
> 
> There's no `user` parameter passed to those events...
> 
> https://github.com/discordjs/discord.js/blob/07017a9699eecc4af7824ace39dd91e8b689f3c6/src/client/actions/GuildBanAdd.js#L16
> 
> https://github.com/discordjs/discord.js/blob/07017a9699eecc4af7824ace39dd91e8b689f3c6/src/client/actions/GuildBanRemove.js#L20

Hm... are you sure?

https://discord.js.org/#/docs/main/master/class/Client?scrollTo=e-guildBanAdd
https://discord.js.org/#/docs/main/master/class/Client?scrollTo=e-guildBanRemove

Either the docs are wrong, or the typings are wrong. I think you're correct, from what it shows there, but the docs should be revised in this case.

iAmZiNc:
Oh, I believe I must've had a cached version of the page. A hard refresh and... poof, they're gone. When were they removed?

iAmZiNc:
Unnecessary, the parameters actually no longer exist.

PR #6165, opened 2021-07-21
feat(CommandInteractionOptionResolver): add sub-command required option

shinotheshino:
This adds a `required` parameter to the `getSubCommand()` and `getSubCommandGroup()` methods, defaulted to `true` (as opposed to `false` in option getters), to allow for optional checking of sub-commands if needed (e.g. logging) without changing the current default behavior that enforces strict command definitions.

### DIFF
--- a/src/structures/CommandInteractionOptionResolver.js
+++ b/src/structures/CommandInteractionOptionResolver.js
@@ -98,7 +98,6 @@ class CommandInteractionOptionResolver {
   /**
    * Gets the selected sub-command.
    * @param {boolean} [required=true] Whether to throw an error if there is no sub-command.
-   * Note that this defaults to true, unlike the other `getX()` methods.
    * @returns {?string} The name of the selected sub-command, or null if not set and not required.
    */
   getSubCommand(required = true) {
@@ -111,7 +110,6 @@ class CommandInteractionOptionResolver {
   /**
    * Gets the selected sub-command group.
    * @param {boolean} [required=true] Whether to throw an error if there is no sub-command group.
-   * Note that this defaults to true, unlike the other `getX()` methods.
    * @returns {?string} The name of the selected sub-command group, or null if not set and not required.
    */
   getSubCommandGroup(required = true) {
--- a/src/structures/CommandInteractionOptionResolver.js
+++ b/src/structures/CommandInteractionOptionResolver.js
@@ -99,7 +99,7 @@ class CommandInteractionOptionResolver {
    * Gets the selected sub-command.
    * @param {boolean} [required=true] Whether to throw an error if there is no sub-command.
    * Note that this defaults to true, unlike the other `getX()` methods.
-   * @returns {string} The name of the selected sub-command.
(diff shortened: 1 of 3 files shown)


src/structures/CommandInteractionOptionResolver.js
```diff
@@ -97,21 +97,25 @@ class CommandInteractionOptionResolver {
 
   /**
    * Gets the selected sub-command.
-   * @returns {string} The name of the selected sub-command.
+   * @param {boolean} [required=true] Whether to throw an error if there is no sub-command.
+   * Note that this defaults to true, unlike the other `getX()` methods.
```
vladfrangu, line 101: Can probably look better as an info block

src/structures/CommandInteractionOptionResolver.js
```diff
@@ -97,21 +97,25 @@ class CommandInteractionOptionResolver {
...
       throw new TypeError('COMMAND_INTERACTION_OPTION_NO_SUB_COMMAND');
     }
     return this._subCommand;
   }
 
   /**
    * Gets the selected sub-command group.
-   * @returns {string} The name of the selected sub-command group.
+   * @param {boolean} [required=true] Whether to throw an error if there is no sub-command group.
+   * Note that this defaults to true, unlike the other `getX()` methods.
```
vladfrangu, line 114: ^

typings/index.d.ts
```diff
@@ -446,8 +446,12 @@ export class CommandInteractionOptionResolver {
 
   public get(name: string, required: true): CommandInteractionOption;
   public get(name: string, required?: boolean): CommandInteractionOption | null;
-  public getSubCommand(): string;
-  public getSubCommandGroup(): string;
+
+  public getSubCommand(required?: true): string;
+  public getSubCommand(required: boolean): string | null;
```
vladfrangu, line 451: This overload will never be matched iirc because its below the one with possible undefined. You should make an overload where required is false, then one for `required?: boolean`

typings/index.d.ts
```diff
@@ -446,8 +446,12 @@ export class CommandInteractionOptionResolver {
...
+  public getSubCommandGroup(required?: true): string;
+  public getSubCommandGroup(required: boolean): string | null;
```
vladfrangu, line 453: Ditto

vladfrangu (changes requested):

typings/index.d.ts
shinotheshino: It matches in `index.ts` though

typings/index.d.ts
```diff
@@ -446,8 +446,12 @@ export class CommandInteractionOptionResolver {
...
+
```
vladfrangu, line 454:
```suggestion
```

vladfrangu (changes requested):
🧹

SpaceEEC (approved):

vladfrangu (approved):

kyranet (approved):

iCrawl (approved):

Issue #6151, opened 2021-07-20
Bug with the (threadUpdate) event. 


GoldenAngel2:
**Please describe the problem you are having in as much detail as possible:**
When a thread's `rateLimitPerUser` is updated to "Off" (0) it doesn't update the property, providing the event with the wrong data.

**Include a reproducible code sample here, if possible:**

```js
const { Client, Intents: { FLAGS } } = require("discord.js");
const client = new Client({ 
    intents: [
        FLAGS.GUILD_MESSAGES,
        FLAGS.GUILD_MEMBERS,
        FLAGS.GUILD_INVITES,
        FLAGS.GUILD_EMOJIS_AND_STICKERS,
        FLAGS.GUILD_BANS,
        FLAGS.GUILDS,
        FLAGS.GUILD_VOICE_STATES,
        FLAGS.GUILD_MESSAGE_REACTIONS,
        FLAGS.GUILD_PRESENCES
    ] 
});

client.on("threadUpdate", (oldThread, thread) => {
    console.log(`Before: ${oldThread.rateLimitPerUser}\nAfter: ${thread.rateLimitPerUser}`);
});

client.login("--TOKEN--");
```

**Further details:**

- discord.js version: 13.0.0-dev (latest master branch commit)
- Node.js version: v16.4.2
- Operating system: Windows 10
- Priority this issue should have – please be realistic and elaborate if possible: Medium

**Relevant client options:**

- partials: none
- gateway intents: `GUILD_MESSAGES, GUILD_MEMBERS, GUILD_INVITES, GUILD_EMOJIS_AND_STICKERS, GUILD_BANS, GUILDS, GUILD_VOICE_STATES, GUILD_MESSAGE_REACTIONS, GUILD_PRESENCES`
- other: none

SpaceEEC:
Looks like Discord is saving 23 bytes by not sending `rate_limit_per_user` when it's `0`. (But not the 30 bytes that could be saved by omitting `locked` and `archived` when they are `false`)

This could be solved by checking for another property in the payload (maybe `thread_metadata`?) to not erroneously overwrite `rateLimitPerUser` when the channel is patched with partial data.

GoldenAngel2:
I've opened an issue on the discord-api-docs repo about the field not being included for `THREAD_UPDATE`
https://github.com/discord/discord-api-docs/issues/3486

GoldenAngel2:
[...]

PR #6172, opened 2021-07-22
fix(ThreadChannel): better property handling

ckohen:

Sets thread auto archive to 0 when not sent in a paylod (exception for interactions)

closes: #6151 

Also Fixes docs and typings for ThreadChannels

### DIFF
--- a/src/managers/ChannelManager.js
+++ b/src/managers/ChannelManager.js
@@ -19,7 +19,7 @@ class ChannelManager extends CachedManager {
    * @name ChannelManager#cache
    */
 
-  _add(data, guild, cache = true, allowUnknownGuild = false, fromInteraction = false) {
+  _add(data, guild, { cache = true, allowUnknownGuild = false, fromInteraction = false } = {}) {
     const existing = this.cache.get(data.id);
     if (existing) {
       if (cache) existing._patch(data, fromInteraction);
@@ -30,7 +30,7 @@ class ChannelManager extends CachedManager {
       return existing;
     }
 
-    const channel = Channel.create(this.client, data, guild, allowUnknownGuild, fromInteraction);
+    const channel = Channel.create(this.client, data, guild, { allowUnknownGuild, fromInteraction });
 
     if (!channel) {
       this.client.emit(Events.DEBUG, `Failed to find guild, or unknown type for channel ${data.id} ${data.type}`);
@@ -99,7 +99,7 @@ class ChannelManager extends CachedManager {
     }
 
     const data = await this.client.api.channels(id).get();
-    return this._add(data, null, cache, allowUnknownGuild);
+    return this._add(data, null, { cache, allowUnknownGuild });
   }
 }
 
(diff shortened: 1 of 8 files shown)


src/managers/ChannelManager.js
```diff
@@ -19,18 +19,18 @@ class ChannelManager extends CachedManager {
    * @name ChannelManager#cache
    */
 
-  _add(data, guild, cache = true, allowUnknownGuild = false) {
+  _add(data, guild, cache = true, allowUnknownGuild = false, fromInteraction = false) {
```
vladfrangu, line 22: At what point should we just make these options an object? @iCrawl

src/structures/ThreadChannel.js
```diff
@@ -62,35 +63,35 @@ class ThreadChannel extends Channel {
...
+       * @type {?Snowflake}
        */
       this.parentId = data.parent_id;
     }
 
     if ('thread_metadata' in data) {
       /**
        * Whether the thread is locked
-       * @type {boolean}
+       * @type {?boolean}
```
vladfrangu, line 74: This can't be undefined

vladfrangu (changes requested):

src/managers/ChannelManager.js
iCrawl: Haha, probably.

src/structures/ThreadChannel.js
ckohen: It only is when received from an interaction because thread_metadata is not present

src/managers/ChannelManager.js
ckohen: Yeah, I was so close to just doing that.

SpaceEEC (approved):

iCrawl (approved):

src/structures/ThreadChannel.js
vladfrangu: You `??` it tho

ckohen: correct, because it's possible it doesn't exist within thread_metadata.

ImRodry: If thread_metadata is not present, wouldn’t that mean that accessing `locked` in an undefined property would result in an error? (Same for all the other properties that come from thread_metadata)

ckohen: Literally 3 lines above where this suggestion it checks this.

ImRodry: Oops my bad. Then where are these set to null?

ckohen: It isn't, the types are correct. For docs the ? means nullish, either null or undefined.

ImRodry: Alright but usually the library sets missing properties to null at the top of the file so idk why that isn’t being done here but both work tbh

vladfrangu: Usually we set them to null if this.\<prop> isn't defined in the payload we received and has never been defined on the class already

kyranet (approved):
Other than @vladfrangu's comments, this looks good to me.

vladfrangu (approved):
If you're sure this change to _add is handled everywhere, LGTM.

PR #6176, opened 2021-07-22
feat: right-clickybois (context menu support for ApplicationCommand and CommandInteraction)

monbrey:
This adds support for creating Application Commands in right-click context menus, and receiving the interactions.

The received data is parsed into the CommandOptionResolver as seamlessly as possible:
- A command acting on a user creates a `USER` type option: `getUser('user)` / `getMember('user')`
- A command acting on a message creates a fake `_MESSAGE` type option for the new `getMessage('message')`

The type of application command is yet to be included in the interaction data, but when it is this will be updated to use that rather than relying on the existence of `target_id`.

### DIFF
--- a/src/structures/BaseCommandInteraction.js
+++ b/src/structures/BaseCommandInteraction.js
@@ -82,7 +82,7 @@ class BaseCommandInteraction extends Interaction {
    * @property {string|number|boolean} [value] The value of the option
    * @property {CommandInteractionOption[]} [options] Additional options if this option is a
    * subcommand (group)
-   * @property {User|APIUser} [user] The resolved user
+   * @property {User} [user] The resolved user
    * @property {GuildMember|APIGuildMember} [member] The resolved member
    * @property {GuildChannel|APIChannel} [channel] The resolved channel
    * @property {Role|APIRole} [role] The resolved role
--- a/src/structures/BaseCommandInteraction.js
+++ b/src/structures/BaseCommandInteraction.js
@@ -17,7 +17,7 @@ class BaseCommandInteraction extends Interaction {
 
     /**
      * The channel this interaction was sent in
-     * @type {?(TextChannel|NewsChannel|DMChannel)}
+     * @type {?TextBasedChannels}
      * @name BaseCommandInteraction#channel
      * @readonly
      */
--- a/src/structures/BaseCommandInteraction.js
+++ b/src/structures/BaseCommandInteraction.js
@@ -80,9 +80,9 @@ class BaseCommandInteraction extends Interaction {
    * @property {string} name The name of the option
    * @property {ApplicationCommandOptionType} type The type of the option
    * @property {string|number|boolean} [value] The value of the option
-   * @property {Collection<string, CommandInteractionOption>} [options] Additional options if this option is a
+   * @property {CommandInteractionOption[]} [options] Additional options if this option is a
    * subcommand (group)
-   * @property {User} [user] The resolved user
+   * @property {User|APIUser} [user] The resolved user
    * @property {GuildMember|APIGuildMember} [member] The resolved member
    * @property {GuildChannel|APIChannel} [channel] The resolved channel
    * @property {Role|APIRole} [role] The resolved role
@@ -91,7 +91,7 @@ class BaseCommandInteraction extends Interaction {
   /**
    * Transforms an option received from the API.
    * @param {APIApplicationCommandOption} option The received option
-   * @param {APIApplicationCommandInteractionData.resolved} resolved The resolved interaction data
+   * @param {APIInteractionDataResolved} resolved The resolved interaction data
    * @returns {CommandInteractionOption}
    * @private
    */
@@ -137,11 +137,6 @@ module.exports = BaseCommandInteraction;
 
 /* eslint-disable max-len */
 /**
- * @external APIApplicationCommandOptionResolved
- * @see {@link https://discord.com/developers/docs/interactions/slash-commands#interaction-applicationcommandinteractiondataresolved}
- */
-
-/**
- * @external APIInteractionDataResolvedOption
- * @see {@link https://discord.com/developers/docs/interactions/slash-commands#sample-application-command-interaction-application-command-interaction-data-resolved-structure}
+ * @external APIInteractionDataResolved
+ * @see {@link https://discord.com/developers/docs/interactions/receiving-and-responding#interaction-object-resolved-data-structure}
  */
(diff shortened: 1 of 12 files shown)


src/structures/CommandInteraction.js
```diff
@@ -130,6 +132,33 @@ class CommandInteraction extends Interaction {
...
+  /**
+   * Resolves and transforms options received from the API for a context menu interaction.
+   * @param {APIApplicationCommandInteractionData} data The interaction data
+   * @returns {CommandInteractionOption[]}
+   * @private
+   */
+  resolveContextMenuOptions({ target_id, resolved }) {
+    const result = [];
+
+    if (resolved.users?.[target_id]) {
```
vladfrangu, line 144: This has a member too

src/structures/CommandInteraction.js
```diff
@@ -130,6 +132,33 @@ class CommandInteraction extends Interaction {
...
+        this.transformOption({ name: 'user', type: ApplicationCommandOptionTypes.USER, value: target_id }, resolved),
+      );
+    }
+
+    if (resolved.messages?.[target_id]) {
+      result.push({
+        name: 'message',
+        type: '_MESSAGE',
+        value: target_id,
+        message: this.channel?.messages._add(resolved.messages[target_id]),
```
vladfrangu, line 155: Return raw data too?

vladfrangu (dismissed):

ThaTiemsz:
`ApplicationCommand` is missing a `type` property.

advaith1:
imo it's a bad idea to stick the target information into `options`; currently context cmds can't have options but they might add support in the future, and also it's not really good ux to just make fake options. it would be better to have dedicated target properties; maybe interaction.target.id, interaction.target.message, interaction.target.user, and interaction.target.member?

monbrey:
Friendly reminder that draft PR means not ready for review

src/structures/ContextMenuInteraction.js
```diff
@@ -0,0 +1,55 @@
...
+        name: 'message',
+        type: '_MESSAGE',
+        value: target_id,
+        message: this.channel?.messages._add(resolved.messages[target_id]),
+      });
+    }
+
+    return result;
+  }
+}
```
NotSugden, line 53: would it not make more sense just to just add properties `user` and `message` to the class?

NotSugden (changes requested):

src/structures/ContextMenuInteraction.js
monbrey: This has been *extensively* debated and this is the approach we've decided on.

iCrawl:
This needs a rebase.

src/structures/ContextMenuInteraction.js
```diff
@@ -0,0 +1,61 @@
...
+  constructor(client, data) {
+    super(client, data);
+    /**
+     * The target of the interaction, parsed into options
+     * @type {CommandInteractionOptionResolver}
+     */
+    this.options = new CommandInteractionOptionResolver(this.client, this.resolveContextMenuOptions(data.data));
+
+    /**
+     * The target of the interaction
```
SpaceEEC, line 21:
```suggestion
     * The id of target of the interaction
```

SpaceEEC (changes requested):
One small doc nit, otherwise LGTM

src/structures/ContextMenuInteraction.js
monbrey: Can I suggest "The id of the target"?

SpaceEEC: Oh, yes, that's better.

SpaceEEC (approved):

kyranet (approved):

ThaTiemsz (changes requested):
`ApplicationCommand#type` should be added so that it's possible to check whether the command is a slash command or context menu.

vladfrangu (dismissed):
LGTM after ThaTiemens comment is resolved

HarmoGlace:
> `ApplicationCommand#type` should be added so that it's possible to check whether the command is a slash command or context menu.

`ApplicationCommandInteractionData` you mean

monbrey:
> `ApplicationCommand#type` should be added so that it's possible to check whether the command is a slash command or context menu.

`isCommand()` and `isContextMenu()` already do this

ThaTiemsz:
> > `ApplicationCommand#type` should be added so that it's possible to check whether the command is a slash command or context menu.
> 
> `isCommand()` and `isContextMenu()` already do this

That's only on `Interaction` though. You can't do that with commands that are cached that you've fetched from the API.
```js
// grab context menu commands
guild.commands.cache.filter(command => command.type === ApplicationCommandTypes.USER || command.type === ApplicationCommandTypes.MESSAGE)
```

monbrey:
> > > `ApplicationCommand#type` should be added so that it's possible to check whether the command is a slash command or context menu.
> > 
> > 
> > `isCommand()` and `isContextMenu()` already do this
> 
> That's only on `Interaction` though. You can't do that with commands that are cached that you've fetched from the API.
> 
> ```js
> // grab context menu commands
> guild.commands.cache.filter(command => command.type === ApplicationCommandTypes.USER || command.type === ApplicationCommandTypes.MESSAGE)
> ```

Right, I misread. Will do.

iCrawl (approved):

vladfrangu (approved):

ThaTiemsz (approved):
You forgot to update the ApplicationCommand types, otherwise LGTM!

src/structures/BaseCommandInteraction.js
```diff
@@ -0,0 +1,142 @@
...
+ * @implements {InteractionResponses}
+ * @abstract
+ */
+class BaseCommandInteraction extends Interaction {
+  constructor(client, data) {
+    super(client, data);
+
+    /**
+     * The channel this interaction was sent in
+     * @type {?(TextChannel|NewsChannel|DMChannel)}
```
NotSugden, line 20:
```suggestion
     * @type {?TextBasedChannels}
```
https://github.com/discordjs/discord.js/blob/47d2ef3e40165d846b5a901709b26dfffea46ef2/src/util/Constants.js#L464-L471

typings/index.d.ts
```diff
@@ -240,6 +242,30 @@ export class BaseClient extends EventEmitter {
   public toJSON(...props: Record<string, boolean | string>[]): unknown;
 }
 
+export abstract class BaseCommandInteraction extends Interaction {
+  public readonly command: ApplicationCommand | ApplicationCommand<{ guild: GuildResolvable }> | null;
+  public readonly channel: TextChannel | DMChannel | NewsChannel | PartialDMChannel | ThreadChannel | null;
```
NotSugden, line 247:
```suggestion
  public readonly channel: TextBasedChannels | null;
```

iCrawl:
This needs a rebase.

PR #6178, opened 2021-07-23
docs(InteractionResponses): Document missing properties

### LINKED ISSUE #6157: Undocumented `fetchReply` option across the `Interaction` methods
Hi hi!

For one of my applications, I need to be able to fetch the message I replied with (from an interaction). `<CommandInteraction>.defer()` returns `Promise<<Message> | void>` which I thought was strange since I never really got a `Message` from that?

Upon looking into the source code of [`InteractionResponses`](https://github.com/discordjs/discord.js/blob/master/src/structures/interfaces/InteractionResponses.js), the options there allow for a `fetchReply` property which returns the `Message` sent. This is really nifty! Just asking if this was undocumented for a reason or is it yet to be added in?

Thank you!

Jiralite:
Hey!

`InteractionResponses` did not document a few `fetchReply` properties to the methods, so I've gone ahead and added those in.

The documentation for the type definition `InteractionDeferUpdateOptions` was completely missing, so I've added that in too!

I also noticed that the `.update()` method documented `WebhookEditMessageOptions` to be sent across. This didn't include the `fetchReply` property, but I did see an interesting `InteractionUpdateOptions` type that wasn't being used _anywhere_ other than itself being defined. It looks a bit wrong though, because it extended `WebhookMessageOptions` which allowed properties like `tts` to be sent through when it wouldn't do anything, ~~so I adjusted this to extend `WebhookEditMessageOptions` instead~~ so I applied the fix from #6162 after reading through it. My attempted fix of this changed `content` to `options` here, so I might as well leave that in for some consistency with the other methods.

This should resolve #6157!

Please check through the wording and typings and stuff and let me know if anything seems amiss!

### DIFF
--- a/src/structures/interfaces/InteractionResponses.js
+++ b/src/structures/interfaces/InteractionResponses.js
@@ -31,6 +31,12 @@ class InteractionResponses {
    */
 
   /**
+   * Options for updating the message received from a {@link ButtonInteraction}.
+   * @typedef {MessageEditOptions} InteractionUpdateOptions
+   * @property {boolean} [fetchReply] Whether to fetch the reply
+   */
+
+  /**
    * Defers the reply to this interaction.
    * @param {InteractionDeferOptions} [options] Options for deferring the reply to this interaction
    * @returns {Promise<Message|void>}
--- a/src/structures/interfaces/InteractionResponses.js
+++ b/src/structures/interfaces/InteractionResponses.js
@@ -14,12 +14,20 @@ class InteractionResponses {
    * Options for deferring the reply to an {@link Interaction}.
    * @typedef {Object} InteractionDeferOptions
    * @property {boolean} [ephemeral] Whether the reply should be ephemeral
+   * @property {boolean} [fetchReply] Whether to return the replied interaction
+   */
+
+  /**
+   * Options for deferring and updating the reply to a {@link ButtonInteraction}.
+   * @typedef {Object} InteractionDeferUpdateOptions
+   * @property {boolean} [fetchReply] Whether to return the replied interaction
    */
 
   /**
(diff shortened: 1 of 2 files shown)


src/structures/interfaces/InteractionResponses.js
```diff
@@ -14,12 +14,20 @@ class InteractionResponses {
    * Options for deferring the reply to an {@link Interaction}.
    * @typedef {Object} InteractionDeferOptions
    * @property {boolean} [ephemeral] Whether the reply should be ephemeral
+   * @property {boolean} [fetchReply] Whether to return the replied interaction
```
vladfrangu, line 17: Shouldn't this specify that it fetches the reply if possible? @discordjs/the-big-3

Jiralite: I personally wasn't sure what to put here, so any suggestions are welcome! Although now that you say that, "Whether to fetch the reply" could be an option!

iCrawl (approved):

vladfrangu (approved):

kyranet (approved):

src/structures/interfaces/InteractionResponses.js
```diff
@@ -178,7 +186,7 @@ class InteractionResponses {
 
   /**
    * Updates the original message of the component on which the interaction was received on.
-   * @param {string|MessagePayload|WebhookEditMessageOptions} options The options for the updated message
+   * @param {string|MessagePayload|InteractionUpdateOptions} options The options for the updated message
```
SpaceEEC, line 189: You also need to create this type in jsdoc.

SpaceEEC (changes requested):

Issue #5672, opened 2021-05-25
Discordjs logic in fetchRecommendedShards doesn't account for large bot sharding when not using the default 1000 servers per shard


scottbucher:
**Please describe the problem you are having in as much detail as possible:**
Essentially, Util.fetchRecommendedShards doesn't take into account large bot sharding. 

![image](https://user-images.githubusercontent.com/47844014/119564851-3ac9e580-bd77-11eb-90f5-74adcbd8982d.png)
[Link to code](https://github.com/discordjs/discord.js/blob/3ac6d550caf1a3490f648899f112488ee3a7095b/src/util/Util.js#L250)

For large bots (over 150,000 servers) discord requires the number of shards to be a multiple of 16, so, when fetching the recommended shard count, discord.js doesn't account for this. It *happens* to work if you leave the guildsPerShard at the default 1000 because discord is auto-rounding to the nearest 16, but if you choose something other than 1000 per shard, discord.js's logic doesn't account for it.

For instance, with my bot, Birthday Bot, I have roughly 167,000 servers, and when doing 2,500 servers per shard I fetch the recommended shard count and receive back 71, however, when I use this, discord sends back that it is invalid sharding settings since 71 isn't a multiple of 16. If I default to the 1000 servers per shard it will return 176, which is a multiple of 16, and thus, works.

**Include a reproducible code sample here, if possible:**

```js
let recommendedShards = await Util.fetchRecommendedShards(MY_BOT_TOKEN, 2500);
```

**Further details:**

- discord.js version: 12.5.1
- Node.js version: 15.5.1
- Operating system: Debian 9
- Priority this issue should have – please be realistic and elaborate if possible: Low

meister03:
@scottbucher | Encountered the same issue. Why dont you divide the recommended Shards with 2?
Then you have the Shard Count for 2000 guilds

KevinNovak:
Recommended Shards would be rounded to a multiple of 16 for large bots, so dividing by 2 could potentially NOT be a multiple of 16.

Ex: discord recommends 48 (3 x 16) shards. 48 / 2 = 24. 24 is not a multiple of 16 and then the bot would not be able to start.

vladfrangu:
Sadly, we cannot detect the bot's guild cound beforehand, to account for the multiple of 16 limit for shards. We can *guesstimate* it by checking if it's >= 150, and then have other logic around it, which is probably what we'll do 👍

scottbucher:
[...]

PR #6184, opened 2021-07-26
feat(FetchRecommendedShardsOptions): account for large bot sharding

scottbucher:
Implements the changes described in #5672. This changes Util.fetchRecommendedShards to take in a FetchRecommendedShardsOption which includes guildsPerShard which is the number of guilds per shard the user wants and a boolean largeBotSharding option which decides if discord.js will round their shards up to a multiple of 16 which is [required for large bot sharding.](https://discord.com/developers/docs/topics/gateway#sharding-for-very-large-bots)

### DIFF
--- a/src/util/Util.js
+++ b/src/util/Util.js
@@ -255,14 +255,21 @@ class Util extends null {
   }
 
   /**
+   * @typedef {Object} FetchRecommendedShardsOptions
+   * @property {number} [guildsPerShard=1000] Number of guilds per shard
+   * @property {boolean} [largeBotSharding=false] Whether or not to round for large bot sharding
+   */
+
+  /**
    * Gets the recommended shard count from Discord.
    * @param {string} token Discord auth token
-   * @param {number} [guildsPerShard=1000] Number of guilds per shard
+   * @param {FetchRecommendedShardsOptions} [options] Options for fetching recommended shard count
    * @returns {Promise<number>} The recommended number of shards
    */
-  static fetchRecommendedShards(token, guildsPerShard = 1000) {
+  static fetchRecommendedShards(token, { guildsPerShard = 1000, largeBotSharding = false } = {}) {
     if (!token) throw new DiscordError('TOKEN_MISSING');
     const defaults = Options.createDefault();
+    const multiple = largeBotSharding ? 16 : 1;
     return fetch(`${defaults.http.api}/v${defaults.http.version}${Endpoints.botGateway}`, {
       method: 'GET',
       headers: { Authorization: `Bot ${token.replace(/^Bot\s*/i, '')}` },
@@ -272,7 +279,7 @@ class Util extends null {
         if (res.status === 401) throw new DiscordError('TOKEN_INVALID');
         throw res;
       })
-      .then(data => data.shards * (1000 / guildsPerShard));
+      .then(data => Math.ceil(data.shards * (1000 / guildsPerShard), multiple) * multiple);
   }
 
   /**
--- a/src/util/Util.js
+++ b/src/util/Util.js
@@ -256,7 +256,7 @@ class Util extends null {
 
   /**
    * @typedef {Object} FetchRecommendedShardsOptions
-   * @property {number} [guildsPerShard=1000] Number of guilds per shard
+   * @property {number} [guildsPerShard=1000] Number of guilds assigned per shard
    * @property {boolean} [largeBotSharding=false] Whether or not to round for large bot sharding
    */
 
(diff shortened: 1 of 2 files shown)


src/util/Util.js
```diff
@@ -254,15 +254,22 @@ class Util extends null {
     return text.replace(/\|\|/g, '\\|\\|');
   }
 
+  /**
+   * @typedef {Object} FetchRecommendedShardsOptions
+   * @property {number} [guildsPerShard=1000] Number of guilds per shard
```
J-Human, line 259:
```suggestion
   * @property {number} [guildsPerShard=1000] Number of guilds assigned per shard
```

src/util/Util.js
```diff
@@ -254,15 +254,22 @@ class Util extends null {
...
+   * @property {boolean} [largeBotSharding=false] Whether or not to round for large bot sharding
+   */
+
   /**
    * Gets the recommended shard count from Discord.
    * @param {string} token Discord auth token
-   * @param {number} [guildsPerShard=1000] Number of guilds per shard
+   * @param {FetchRecommendedShardsOptions} [options] Options for fetching recommended shard count
```
J-Human, line 266:
```suggestion
   * @param {FetchRecommendedShardsOptions} [options] Options for fetching the recommended shard count
```

J-Human (changes requested):

src/util/Util.js
```diff
@@ -272,7 +279,7 @@ class Util extends null {
         if (res.status === 401) throw new DiscordError('TOKEN_INVALID');
         throw res;
       })
-      .then(data => data.shards * (1000 / guildsPerShard));
+      .then(data => Math.ceil(data.shards * (1000 / guildsPerShard), multiple) * multiple);
```
vladfrangu, line 282: Math.ceil doesn't have two inputs

src/util/Util.js
```diff
@@ -254,15 +254,22 @@ class Util extends null {
...
    * @param {string} token Discord auth token
-   * @param {number} [guildsPerShard=1000] Number of guilds per shard
+   * @param {FetchRecommendedShardsOptions} [options] Options for fetching the recommended shard count
    * @returns {Promise<number>} The recommended number of shards
    */
-  static fetchRecommendedShards(token, guildsPerShard = 1000) {
+  static fetchRecommendedShards(token, { guildsPerShard = 1000, largeBotSharding = false } = {}) {
     if (!token) throw new DiscordError('TOKEN_MISSING');
     const defaults = Options.createDefault();
+    const multiple = largeBotSharding ? 16 : 1;
```
vladfrangu, line 272: Personally this shouldn't take it as an option, but rather guesstimate it from what Discord says the recommended shards are (if > 150, then we are in large territory)

vladfrangu (changes requested):

src/util/Util.js
scottbucher: Does discord tell us on the API side what the requirement is for large bot sharding? If not I'm not sure guesstimating here would be good, large bot sharding requirements were originally 250k, then 200k, and now 150k iirc. I feel like giving the user an option is better than just guesstimating since the requirement could change in the future.

vladfrangu: That's also a double edged sword, especially if you use the sharding manager which doesn't (iirc) let you pass in this option. Personally I'd rather have to decrement a >= check but guarantee users that they'll get the shards they should have! @discordjs/the-big-3 whats your take on thios?

iCrawl: We can default to 16 but should allow passing any arbitrary number in case this requirement ever changes.

scottbucher: > We can default to 16 but should allow passing any arbitrary number in case this requirement ever changes.

Okay, so assuming the FetchRecommendedShardsOption stays we can either simply add a largeBotShardMultiple like:
```
guildsPerShard = 1000
largeBotSharding = false
largeBotShardingMultiple = 16
```

Or just ditch the largeBotSharding option entirely by replacing it with a multiple
```
guildPerShard = 1000
roundToMultiple = 1 (and state in the documentation it needs to be set to 16 for large bot sharding)
```

iCrawl: I'd rather go with the 2nd option 👍

kyranet: I'd name it `multipleOf`, since it's not rounding, but ceiling to the next multiple.

scottbucher: Sounds good, I'll make these changes later today.

kyranet (approved):

iCrawl (approved):

SpaceEEC (approved):

PR #6192, opened 2021-07-28
feat(WebhookClient): allow creation of clients via URLs

### LINKED ISSUE #6164: WebhookClient creation with url
Having to provide the webhook id and token instead of just an url can result in more and less clean code which can be quite annoying.

A method overload that allows users to pass the whole webhook url:
```js
  const webhookClient = new Discord.WebhookClient("https://discord.com/api/webhooks/12334567890/asddgsjhjhshfsjhefhshjfhsjfjehsjhf_ashdjhadjhwhja");
```

Using regex to filter out the id and token

**Additional context**
grafik

J-Human:

This allows `WebhookClient`'s to be made with URLs, without the need of removing it or doing any type of parsing. This is useful for pass-through services like Webhook senders, etc., allowing a more convenient time for developers who don't want to use IDs and tokens. Example:
```ts
import { WebhookClient } from 'discord.js';

const webhook = new WebhookClient({ url: 'https://discord.com/api/webhooks/1234567890/abc' });

webhook.send({ content: 'Hi.' });
```

This PR also updates:
* Typings.
* Errors (URLs are passed to a RegEx and validated, although 100% perfection is not possible).
* Tests.

It now mandates initializing Webhook clients should have *one* object parameter, that can contain either ID + token or just a URL (this isn't validated so all 3 are optional).

If I do anything wrong, please forgive me, or if this isn't needed, feel free to give the right feedback/review, I appreciate it.

Fixes #6164

### DIFF
--- a/src/client/WebhookClient.js
+++ b/src/client/WebhookClient.js
@@ -1,6 +1,7 @@
 'use strict';
 
 const BaseClient = require('./BaseClient');
+const { Error } = require('../errors');
 const Webhook = require('../structures/Webhook');
 
 /**
@@ -10,19 +11,40 @@ const Webhook = require('../structures/Webhook');
  */
 class WebhookClient extends BaseClient {
   /**
-   * @param {Snowflake} id The webhook's id
-   * @param {string} token Token of the webhook
+   * The data for the webhook client
+   * @typedef {Object} WebhookClientData
+   * @property {Snowflake} [id] The webhook's id
+   * @property {string} [token] Token of the webhook
+   * @property {string} [url] The full URL of the webhook
+   */
+
+  /**
+   * @param {WebhookClientData} data The data of the webhook
    * @param {ClientOptions} [options] Options for the client
    * @example
    * // Create a new webhook and send a message
-   * const hook = new Discord.WebhookClient('1234', 'abcdef');
+   * const hook = new Discord.WebhookClient({ id: '1234', token: 'abcdef' });
    * hook.send('This will send a message').catch(console.error);
    */
-  constructor(id, token, options) {
+  constructor(data, options) {
     super(options);
     Object.defineProperty(this, 'client', { value: this });
-    this.id = id;
-    Object.defineProperty(this, 'token', { value: token, writable: true, configurable: true });
+    let id, token;
+
+    if ('url' in data) {
+      const url = data.url.match(
+        // eslint-disable-next-line no-useless-escape
+        /^https?:\/\/(?:discord|discordapp)\.com\/api\/webhooks(?:\/v[0-9]\d*)?\/([^\/]+)\/([^\/]+)/i,
+      );
+
+      if (!url || url.length <= 1) throw new Error('WEBHOOK_URL_INVALID');
+
+      id = url[1];
+      token = url[2];
+    }
+
+    this.id = id ?? data.id;
+    Object.defineProperty(this, 'token', { value: token ?? data.id, writable: true, configurable: true });
   }
 
   // These are here only for documentation purposes - they are implemented by Webhook
--- a/src/client/WebhookClient.js
+++ b/src/client/WebhookClient.js
@@ -11,11 +11,11 @@ const Webhook = require('../structures/Webhook');
  */
 class WebhookClient extends BaseClient {
   /**
-   * The data for the webhook client
+   * The data for the webhook client containing either an id and token or just a URL
    * @typedef {Object} WebhookClientData
-   * @property {Snowflake} [id] The webhook's id
-   * @property {string} [token] Token of the webhook
-   * @property {string} [url] The full URL of the webhook
+   * @property {Snowflake} [id] The id of the webhook
+   * @property {string} [token] The token of the webhook
+   * @property {string} [url] The full url for the webhook client
    */
 
   /**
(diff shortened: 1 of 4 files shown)


src/client/WebhookClient.js
```diff
@@ -10,19 +11,40 @@ const Webhook = require('../structures/Webhook');
...
     super(options);
     Object.defineProperty(this, 'client', { value: this });
-    this.id = id;
-    Object.defineProperty(this, 'token', { value: token, writable: true, configurable: true });
+    let id, token;
+
+    if ('url' in data) {
+      const url = data.url.match(
+        // eslint-disable-next-line no-useless-escape
+        /^https?:\/\/(?:discord|discordapp)\.com\/api\/webhooks(?:\/v[0-9]\d*)?\/([^\/]+)\/([^\/]+)/i,
```
vladfrangu, line 37: This regexp doesn't account for subdomains. You could also split by slashes

src/client/WebhookClient.js
```diff
@@ -10,19 +11,40 @@ const Webhook = require('../structures/Webhook');
...
+      );
+
+      if (!url || url.length <= 1) throw new Error('WEBHOOK_URL_INVALID');
+
+      id = url[1];
+      token = url[2];
+    }
+
+    this.id = id ?? data.id;
+    Object.defineProperty(this, 'token', { value: token ?? data.id, writable: true, configurable: true });
```
vladfrangu, line 47: Did you mean token?
```suggestion
    Object.defineProperty(this, 'token', { value: token ?? data.token, writable: true, configurable: true });
```

src/errors/Messages.js
```diff
@@ -101,6 +101,7 @@ const Messages = {
 
   WEBHOOK_MESSAGE: 'The message was not sent by a webhook.',
   WEBHOOK_TOKEN_UNAVAILABLE: 'This action requires a webhook token, but none is available.',
+  WEBHOOK_URL_INVALID: 'The provided webhook URL is not well formed.',
```
vladfrangu, line 104: Up to debate if this is better

```suggestion
  WEBHOOK_URL_INVALID: 'The provided webhook URL is not valid.',
```

typings/index.d.ts
```diff
@@ -4340,6 +4340,12 @@ export type VerificationLevel = keyof typeof VerificationLevels;
 
 export type VoiceBasedChannelTypes = 'GUILD_VOICE' | 'GUILD_STAGE_VOICE';
 
+export interface WebhookClientData {
```
vladfrangu, line 4343: This should be a union of two interfaces that define what props they should have (WebhookClientData = WebhookClientDataURL | WebhookClientDataIdWithToken)

vladfrangu (changes requested):

J-Human:
@vladfrangu fixed! Thanks for reminding. 😅

vladfrangu (approved):

src/client/WebhookClient.js
```diff
@@ -10,19 +11,51 @@ const Webhook = require('../structures/Webhook');
...
     super(options);
     Object.defineProperty(this, 'client', { value: this });
-    this.id = id;
-    Object.defineProperty(this, 'token', { value: token, writable: true, configurable: true });
+    let id, token;
+
+    if ('url' in data) {
+      const url = data.url.match(
+        // eslint-disable-next-line no-useless-escape
+        /^https?:\/\/(?:canary|ptb)?\.?(?:discord|discordapp)\.com\/api\/webhooks(?:\/v[0-9]\d*)?\/([^\/]+)\/([^\/]+)/i,
```
Jiralite, line 48: I'm not sure if we should be allowing domains with discordapp.com still? Webhooks links generated now across the release channels won't be using "discordapp", and I feel if people are still using that, they should update their webhook url instead?

iCrawl: Yeah we can probably remove the "app" part.

SpaceEEC (approved):

src/client/WebhookClient.js
```diff
@@ -10,19 +11,51 @@ const Webhook = require('../structures/Webhook');
...
+      );
+
+      if (!url || url.length <= 1) throw new Error('WEBHOOK_URL_INVALID');
+
+      id = url[1];
```
NotSugden, line 53: i think theres a thing for this
`[, id, token] = url;`? something like that

src/client/WebhookClient.js
```diff
@@ -10,19 +11,51 @@ const Webhook = require('../structures/Webhook');
...
+   * const hook = new Discord.WebhookClient({ id: '1234', token: 'abcdef' });
    * hook.send('This will send a message').catch(console.error);
    */
-  constructor(id, token, options) {
+  constructor(data, options) {
     super(options);
     Object.defineProperty(this, 'client', { value: this });
-    this.id = id;
-    Object.defineProperty(this, 'token', { value: token, writable: true, configurable: true });
+    let id, token;
```
NotSugden, line 43:
```suggestion
    let { id, token } = data;
```
and remove the `id ?? data.id` below (same for token)

NotSugden (changes requested):
the constructor typings need updating still

J-Human:
@NotSugden @Jiralite done! Your suggestions have been committed. Thanks a lot.
* According to Jiralite's comment (and Crawl's decision) - the `discordapp` domain has vanished.
* As per Sugden's comment here, I updated the constructor typings for `WebhookClient` (yes, I forgot to do so), and retrieving the data from either the URL or ID, token has been refactored.

src/client/WebhookClient.js
```diff
@@ -10,17 +11,48 @@ const Webhook = require('../structures/Webhook');
  */
 class WebhookClient extends BaseClient {
   /**
-   * @param {Snowflake} id The webhook's id
-   * @param {string} token Token of the webhook
+   * The data for the webhook client containing either an id and token or just a URL
+   * @typedef {Object} WebhookClientData
+   * @property {(WebhookClientDataIdWithToken|WebhookClientDataURL)} data The data for the webhook client
+   */
```
NotSugden, line 17: this typedef is
```ts
{
  data: {
    id: string;
    token: string;
  } | {
    url: string;
  }
}
```
convert the typedefs below into a single typedef, and use that instead
(the jsdoc typedefs aren't usually handled this way, the typings are left for the more specific unions)

src/client/WebhookClient.js
```diff
@@ -10,17 +11,48 @@ const Webhook = require('../structures/Webhook');
...
     Object.defineProperty(this, 'client', { value: this });
+    let { id, token } = data;
+
+    if ('url' in data) {
+      const url = data.url.match(
+        // eslint-disable-next-line no-useless-escape
+        /^https?:\/\/(?:canary|ptb)?\.?discord\.com\/api\/webhooks(?:\/v[0-9]\d*)?\/([^\/]+)\/([^\/]+)/i,
+      );
+
+      if (!url || url.length <= 1) throw new Error('WEBHOOK_URL_INVALID');
```
NotSugden, line 51:
```suggestion
      if (url?.length !== 2) throw new Error('WEBHOOK_URL_INVALID');
```
```suggestion
      if (url?.length < 2) throw new Error('WEBHOOK_URL_INVALID');
```
either or
can also stay the same prob doesn't matter much

J-Human:
Done.

iCrawl (approved):
Love it, I am actually in need of this too :D

SpaceEEC (approved):

AlexanderGipp:
Thanks for the addition @J-Human

PR #6193, opened 2021-07-28
refactor: use guild-wide route for fetching active threads

ckohen:

Guild wide route for fetching active threads! And since channel wide route its being removed in v10, already based the per channel fetch on guild wide.

### DIFF
--- a/src/managers/GuildChannelManager.js
+++ b/src/managers/GuildChannelManager.js
@@ -1,6 +1,7 @@
 'use strict';
 
 const CachedManager = require('./CachedManager');
+const ThreadManager = require('./ThreadManager');
 const { Error } = require('../errors');
 const GuildChannel = require('../structures/GuildChannel');
 const PermissionOverwrites = require('../structures/PermissionOverwrites');
@@ -149,7 +150,7 @@ class GuildChannelManager extends CachedManager {
    * @param {BaseFetchOptions} [options] Additional options for this fetch
    * @returns {Promise<?GuildChannel|Collection<Snowflake, GuildChannel>>}
    * @example
-   * // Fetch all channels from the guild
+   * // Fetch all channels from the guild (excluding threads)
    * message.guild.channels.fetch()
    *   .then(channels => console.log(`There are ${channels.size} channels.`))
    *   .catch(console.error);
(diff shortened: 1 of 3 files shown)


vladfrangu (approved):

iCrawl (approved):

kyranet (approved):

src/managers/GuildChannelManager.js
```diff
@@ -177,6 +178,27 @@ class GuildChannelManager extends CachedManager {
...
+   * @returns {Promise<FetchedThreads>|Promise<APIActiveThreadsList>}
+   * @example
+   * // Fetch all threads from the guild
+   * message.guild.channels.fetchActiveThreads()
+   *   .then(fetched => console.log(`There are ${fetched.threads.size} threads.`))
+   *   .catch(console.error);
+   */
+  async fetchActiveThreads(cache = true, returnRaw = false) {
+    const raw = await this.client.api.guilds(this.guild.id).threads.active.get();
+    return returnRaw ? raw : ThreadManager._mapThreads(raw, this.client, { guild: this.guild, cache });
```
SpaceEEC, line 195: This (`returnRaw`) feels a bit out of place, we don't do this anywhere else, do we?

typings/index.d.ts
```diff
@@ -2292,6 +2293,8 @@ export class GuildChannelManager extends CachedManager<
   ): Promise<
     Collection<Snowflake, TextChannel | VoiceChannel | CategoryChannel | NewsChannel | StoreChannel | StageChannel>
   >;
+  public fetchActiveThreads(cache?: boolean): Promise<FetchedThreads>;
+  public fetchActiveThreads(cache: boolean, returnRaw: true): Promise<APIThreadList>;
```
NotSugden, line 2297: `cache` has no effect if `returnRaw` is true - perhaps change to a single object parameter and omit cache on this overload?

NotSugden (changes requested):

src/managers/GuildChannelManager.js
NotSugden: what would be the use case for returning the raw threads anyway?

ckohen: its for internals, see
https://github.com/discordjs/discord.js/blob/150a295224ea0de5ea541d624addc8489e00b44a/src/managers/ThreadManager.js#L232

SpaceEEC: I don't think that should be part of the public API then. :eyes:
Maybe move that to a private method or duplicate that api call into the ThreadManager?

ckohen: I was thinking about duplicating the call, especially since its a one-liner. Is that better?

SpaceEEC: I think so.

SpaceEEC (approved):

PR #6206, opened 2021-07-28
docs(MessageMentions): fix documentation for members property

a9ex:

`MessageMentions#members` works also in a ThreadChannel

### DIFF
--- a/src/structures/MessageMentions.js
+++ b/src/structures/MessageMentions.js
@@ -134,7 +134,7 @@ class MessageMentions {
   }
 
   /**
-   * Any members that were mentioned (only in {@link TextChannel}s)
+   * Any members that were mentioned (only in {@link TextChannel}s and {@link ThreadChannel}s)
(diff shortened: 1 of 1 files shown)


src/structures/MessageMentions.js
```diff
@@ -134,7 +134,7 @@ class MessageMentions {
   }
 
   /**
-   * Any members that were mentioned (only in {@link TextChannel}s)
+   * Any members that were mentioned (only in {@link TextChannel}s and {@link ThreadChannel}s)
```
Jiralite, line 137: It looks like this really only works in guilds. Perhaps it would be better to just say that it only works in guilds instead of adding every channel type in a guild that is text-based?

a9ex: Yes it could be better 👍

SpaceEEC (approved):

iCrawl (approved):

PR #6207, opened 2021-07-28
feat(ThreadChannel): add fetchOwner() method

ImRodry:
This PR adds a fetchOwner() method that returns the ThreadMember object corresponding to the Owner of the thread

### DIFF
--- a/src/structures/ThreadChannel.js
+++ b/src/structures/ThreadChannel.js
@@ -189,15 +189,6 @@ class ThreadChannel extends Channel {
   }
 
   /**
-   * The {@link ThreadMember} object of the owner of this thread
-   * @type {?ThreadMember}
-   * @readonly
-   */
-  get owner() {
-    return this.members.cache.get(this.ownerId);
-  }
-
-  /**
    * The time at which this thread's archive status was last changed
    * <info>If the thread was never archived or unarchived, this is the time at which the thread was created</info>
    * @type {?Date}
@@ -246,10 +237,18 @@ class ThreadChannel extends Channel {
   /**
    * Fetches the owner of this thread
    * @param {FetchOwnerOptions} [options] The options for fetching the member
-   * @returns {Promise<GuildMember>}
+   * @returns {Promise<ThreadMember>}
    */
-  fetchOwner(options) {
-    return this.guild.members.fetch({ ...options, user: this.ownerId });
+  async fetchOwner({ cache = true, force = false } = {}) {
+    if (!force) {
+      const existing = this.cache.get(this.ownerId);
+      if (existing) return existing;
+    }
+
+    // We cannot fetch a single thread member, as of this commit's date, Discord API throws with 405
+    const data = await this.client.api.channels(this.thread.id, 'thread-members').get();
+    const owner = data.find(member => member.user_id === this.ownerId);
+    return this._add(owner, cache);
   }
 
   /**
--- a/src/structures/ThreadChannel.js
+++ b/src/structures/ThreadChannel.js
@@ -247,7 +247,7 @@ class ThreadChannel extends Channel {
 
     // We cannot fetch a single thread member, as of this commit's date, Discord API responds with 405
     const members = await this.members.fetch(cache);
-    return members.get(this.ownerId);
+    return members.get(this.ownerId) ?? null;
   }
 
   /**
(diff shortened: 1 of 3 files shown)


src/structures/ThreadChannel.js
```diff
@@ -234,6 +234,23 @@ class ThreadChannel extends Channel {
...
+  async fetchOwner({ cache = true, force = false } = {}) {
+    if (!force) {
+      const existing = this.cache.get(this.ownerId);
+      if (existing) return existing;
+    }
+
+    // We cannot fetch a single thread member, as of this commit's date, Discord API throws with 405
+    const data = await this.client.api.channels(this.thread.id, 'thread-members').get();
+    const owner = data.find(member => member.user_id === this.ownerId);
+    return this._add(owner, cache);
```
SpaceEEC, line 251:
```suggestion
    const members = await this.members.fetch(cache);
    return members.get(this.ownerId);
```
Although I'm not sure if this method is not a bit misleading without being able to actually fetch a single member.

SpaceEEC (changes requested):

src/structures/ThreadChannel.js
ImRodry: I used a method similar to what was used with single role fetching but yeah I know it can be a bit misleading. I'll add your suggestions

src/structures/ThreadChannel.js
```diff
@@ -234,6 +234,23 @@ class ThreadChannel extends Channel {
...
+   * @param {FetchOwnerOptions} [options] The options for fetching the member
+   * @returns {Promise<ThreadMember>}
+   */
+  async fetchOwner({ cache = true, force = false } = {}) {
+    if (!force) {
+      const existing = this.cache.get(this.ownerId);
+      if (existing) return existing;
+    }
+
+    // We cannot fetch a single thread member, as of this commit's date, Discord API throws with 405
```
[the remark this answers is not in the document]
ImRodry: That message is also present on RoleManager#fetch btw but I'll change it

src/structures/ThreadChannel.js
```diff
@@ -234,6 +234,22 @@ class ThreadChannel extends Channel {
...
   }
 
+  /**
+   * Fetches the owner of this thread
+   * @param {FetchOwnerOptions} [options] The options for fetching the member
+   * @returns {Promise<ThreadMember>}
+   */
+  async fetchOwner({ cache = true, force = false } = {}) {
+    if (!force) {
+      const existing = this.cache.get(this.ownerId);
```
kyranet, line 244: What's `this.cache` inside a `ThreadChannel`? Did you mean `this.members.cache.get`?

kyranet (changes requested):

src/structures/ThreadChannel.js
ImRodry: Oopsie yeah forgot that

src/structures/ThreadChannel.js
```diff
@@ -234,6 +234,22 @@ class ThreadChannel extends Channel {
     return this.parent?.permissionsFor(memberOrRole) ?? null;
   }
 
+  /**
+   * Fetches the owner of this thread
+   * @param {FetchOwnerOptions} [options] The options for fetching the member
```
Jiralite, line 239: I'm not sure about this. If a user clicks on this, they'll see the description:
> Options used to fetch the owner of guild

This description isn't accurate. I believe the wording should change to something like:
> Options used to fetch the owner of guild or the owner of a thread

Jiralite (changes requested):

src/structures/ThreadChannel.js
ImRodry: I don't even know why this typedef exists since it's just BaseFetchOptions but sure, I'll add that

Jiralite:
I believe the current implementation of this always expects to return the owner, but this is not true. The owner can leave! This would make the return type possibly `undefined` - in this case, it should be `null`.

src/structures/ThreadChannel.js
```diff
@@ -234,6 +234,22 @@ class ThreadChannel extends Channel {
...
+   * @returns {Promise<ThreadMember>}
```
Jiralite, line 240:
```suggestion
   * @returns {Promise<?ThreadMember>}
```

Jiralite (changes requested):

Jiralite (approved):

vladfrangu (approved):

kyranet (approved):

iCrawl (approved):

SpaceEEC (approved):

PR #6208, opened 2021-07-28
docs(Constants): document the Constants object for enum-like usage

monbrey:
This documents the exported and separately documented typedef values in `util/Constants.js` as one parent object, covering all of the ones that can be used in an enum-like way, e.g. `Constants.MessageButtonStyles.PRIMARY`. As discord.js doesn't export the TS enums in its packaged typings, this alternative allows us to support the usage of Constants instead of relying on magic strings.

I'm not sure if I've done it correctly, since this definition is referencing the enum/object/whatever of all the available types rather than just one singular selected value from the typedef. Open to feedback on this.

Constants that are just an array of string values such as `Constants.MessageTypes` are not documented here to avoid confusion, as they can't be property-accessed like the other. Not sure how to go about including those if necessary. `MessageType[]`?

The definitions of the types are largely copy-pasted from their respective typedefs. I don't consider changing the description of the typing to be in scope of this PR, so any issues should be a new PR changing both locations.

### DIFF
--- a/src/util/Constants.js
+++ b/src/util/Constants.js
@@ -1016,5 +1016,5 @@ function createEnum(keys) {
  * @property {StickerType} StickerTypes The value set for a sticker's type.
  * @property {VerificationLevel} VerificationLevels The value set for the verification levels for a guild.
  * @property {WebhookType} WebhookTypes The value set for a webhook's type.
- * @property {WSEvents} WSEventType The type of a websocket message event.
+ * @property {WSEventType} WSEvents The type of a websocket message event.
  */
--- a/src/util/Constants.js
+++ b/src/util/Constants.js
@@ -986,3 +986,35 @@ function createEnum(keys) {
   }
   return obj;
 }
+
+/**
+ * @typedef {Object} Constants Constants that can be used in an enum or object-like way.
+ * @property {ActivityType} ActivityTypes The type of an activity of a users presence.
+ * @property {APIError} APIErrors An error encountered while performing an API request.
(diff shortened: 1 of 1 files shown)


src/util/Constants.js
```diff
@@ -986,3 +986,35 @@ function createEnum(keys) {
...
+ * @property {OverwriteType} OverwriteTypes An overwrite type.
+ * @property {PartialType} PartialTypes The type of Structure allowed to be a partial.
+ * @property {PremiumTier} PremiumTiers The premium tier (Server Boost level) of a guild.
+ * @property {PrivacyLevel} PrivacyLevels Privacy level of a {@link StageInstance} object.
+ * @property {Status} Status The available statuses of the client.
+ * @property {StickerFormatType} StickerFormatTypes The value set for a sticker's format type.
+ * @property {StickerType} StickerTypes The value set for a sticker's type.
+ * @property {VerificationLevel} VerificationLevels The value set for the verification levels for a guild.
+ * @property {WebhookType} WebhookTypes The value set for a webhook's type.
+ * @property {WSEvents} WSEventType The type of a websocket message event.
```
SpaceEEC, line 1019:
```suggestion
 * @property {WSEventType} WSEvents The type of a websocket message event.
```

SpaceEEC (changes requested):
Otherwise LGTM

vladfrangu (approved):

kyranet (approved):

PR #6210, opened 2021-07-29
refactor(*): use async functions

NotSugden:
This PR refactors all functions to be async instead of chaining a `.then(...)` - also fixes a couple issues where errors are thrown instead of returning a `Promise.reject(...);`

untested

### DIFF
--- a/src/managers/GuildManager.js
+++ b/src/managers/GuildManager.js
@@ -211,10 +211,9 @@ class GuildManager extends CachedManager {
       },
     });
 
-    // eslint-disable-next-line consistent-return
-    return new Promise(resolve => {
-      if (this.client.guilds.cache.has(data.id)) return resolve(this.client.guilds.cache.get(data.id));
+    if (this.client.guilds.cache.has(data.id)) return this.client.guilds.cache.get(data.id);
 
+    return new Promise(resolve => {
       const handleGuild = guild => {
         if (guild.id === data.id) {
           clearTimeout(timeout);
--- a/src/managers/GuildManager.js
+++ b/src/managers/GuildManager.js
@@ -195,45 +195,43 @@ class GuildManager extends CachedManager {
     }
     if (systemChannelFlags) systemChannelFlags = SystemChannelFlags.resolve(systemChannelFlags);
 
-    return new Promise((resolve, reject) =>
-      this.client.api.guilds
-        .post({
-          data: {
-            name,
-            icon,
-            verification_level: verificationLevel,
-            default_message_notifications: defaultMessageNotifications,
-            explicit_content_filter: explicitContentFilter,
-            roles,
-            channels,
-            afk_channel_id: afkChannelId,
-            afk_timeout: afkTimeout,
-            system_channel_id: systemChannelId,
-            system_channel_flags: systemChannelFlags,
-          },
-        })
-        .then(data => {
-          if (this.client.guilds.cache.has(data.id)) return resolve(this.client.guilds.cache.get(data.id));
+    const data = await this.client.api.guilds.post({
+      data: {
+        name,
+        icon,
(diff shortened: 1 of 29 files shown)


src/structures/Integration.js
```diff
@@ -125,17 +125,12 @@ class Integration extends Base {
...
-      .guilds(this.guild.id)
-      .integrations(this.id)
-      .post()
-      .then(() => {
-        this.syncing = false;
-        this.syncedAt = Date.now();
-        return this;
-      });
+    await this.client.api.guilds(this.guild.id).integrations(this.id).post();
+    this.syncing = false;
```
SpaceEEC, line 131: This might be better in a finally, if something goes wrong. ~~Not that anyone uses this method.~~
I also think this endpoint is no longer available to bots. 🤔 (For another PR though)

SpaceEEC (approved):

iCrawl:
This needs a rebase.

src/structures/GuildAuditLogs.js
```diff
@@ -500,20 +501,19 @@ class GuildAuditLogsEntry {
...
+          const invites = await guild.invites.fetch();
+          this.target = invites.find(i => i.code === change) ?? null;
         } else {
           this.target = this.changes.reduce((o, c) => {
             o[c.key] = c.new ?? c.old;
             return o;
           }, {});
-          return this.target;
         }
+        return this.target;
```
vladfrangu, line 516: Are you sure this works still?

vladfrangu (changes requested):

src/structures/GuildAuditLogs.js
NotSugden: i suppose there isn't really any point returning it or assigning it, ill change this

NotSugden: i think this section may need some refactoring, but thats probably something for another PR

src/managers/GuildManager.js
```diff
@@ -195,45 +195,43 @@ class GuildManager extends CachedManager {
...
-              this.client.removeListener(Events.GUILD_CREATE, handleGuild);
-              this.client.decrementMaxListeners();
-              resolve(guild);
-            }
-          };
-          this.client.incrementMaxListeners();
-          this.client.on(Events.GUILD_CREATE, handleGuild);
+    // eslint-disable-next-line consistent-return
+    return new Promise(resolve => {
+      if (this.client.guilds.cache.has(data.id)) return resolve(this.client.guilds.cache.get(data.id));
```
kyranet, line 216: This can be moved outside to prevent constructing a new promise object.

src/managers/GuildMemberManager.js
```diff
@@ -246,17 +246,16 @@ class GuildMemberManager extends CachedManager {
...
-      .post({
+      ({ pruned } = await endpoint.get({ query, reason }));
+    } else {
+      ({ pruned } = await endpoint.post({
         data: { ...query, compute_prune_count },
         reason,
-      })
-      .then(data => data.pruned);
+      }));
+    }
```
kyranet, line 257: This could probably be written as:
```javascript
const { pruned } = await (dry
  ? endpoint.get({ query, reason })
  : endpoint.post({
      data: { ...query, compute_prune_count },
      reason
    }));
```

kyranet (changes requested):

src/managers/GuildManager.js
NotSugden: oh yeah, i meant to ask why it was even done in the promise in the first place, very odd

iCrawl:
This needs a rebase.

NotSugden:
rebased

kyranet (approved):

PR #6212, opened 2021-07-29
types(ButtonMessageOptions): make button/link button typings more specific

suneettipirneni:

With discord buttons and link button components, there are certain property combinations that aren't allowed. Basically, You can't use a `customID` property for buttons with the `LINK` style, for non-link buttons you can't use the `url` property. The issue is that it is only presented at runtime:

```
components[0].components[0].custom_id: A custom id and url cannot both be specified
```

With this PR it adjusts the `ButtonMessageOptions` type to provide a compile-time error, if an invalid `ButtonMessageOptions` combination is provided.

### DIFF
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -3799,11 +3799,11 @@ export interface BaseButtonOptions extends BaseMessageComponentOptions {
 export type MessageButtonOptions = BaseButtonOptions &
   (
     | {
-        style: Exclude<MessageButtonStyle, 'LINK'>;
+        style: Exclude<MessageButtonStyleResolvable, 'LINK' | MessageButtonStyles.LINK>;
         customId: string;
       }
     | {
-        style: 'LINK';
+        style: 'LINK' | MessageButtonStyles.LINK;
         url: string;
       }
   );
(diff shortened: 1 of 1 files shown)


vladfrangu (approved):

kyranet (approved):

typings/index.d.ts
```diff
@@ -3790,15 +3790,24 @@ export interface MessageActivity {
...
 
+export type MessageButtonOptions = BaseButtonOptions &
+  (
+    | {
+        style: Exclude<MessageButtonStyle, 'LINK'>;
+        customId: string;
+      }
+    | {
+        style: 'LINK';
+        url: string;
```
tguichaoua, line 3807:
```suggestion
        style: Exclude<MessageButtonStyleResolvable, 'LINK' | MessageButtonStyles.LINK>;
        customId: string;
      }
    | {
        style: 'LINK' | MessageButtonStyles.LINK;
        url: string;
```

suneettipirneni: Ah good catch!

suneettipirneni: Actually on second thought this may not be needed, `MessageButtonStyleResolvable` isn't an exported type. I'm guessing the only area this would be useful would be for internal implementations.

iCrawl (approved):

SpaceEEC (approved):

PR #6215, opened 2021-07-29
refactor(Message): make thread a getter for accuracy

ckohen:

Fixes a bug where #thread is null after a message update because it does not contain the thread data, relying on thread create instead.

### DIFF
--- a/src/structures/Message.js
+++ b/src/structures/Message.js
@@ -390,8 +390,8 @@ class Message extends Base {
 
   /**
    * The thread started by this message
-   * <info>This can be null when the message has a thread if the cache no longer has the thread.
-   * Use {@link Message#hasThread} to check whether the message has a thread.</info>
+   * <info>This property is not suitable for checking whether a message has a thread,
+   * use {@link Message#hasThread} instead.</info>
    * @type {?ThreadChannel}
    * @readonly
    */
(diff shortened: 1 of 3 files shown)


iCrawl (approved):

src/structures/Message.js
```diff
@@ -386,6 +379,26 @@ class Message extends Base {
...
+   * @type {boolean}
+   * @readonly
+   */
+  get hasThread() {
+    return this.flags.has(MessageFlags.FLAGS.HAS_THREAD);
+  }
+
+  /**
+   * The thread started by this message
+   * <info>This can be null when the message has a thread if the cache no longer has the thread.
```
ImRodry, line 393: Isn’t this already implicit for all getters of this nature? I don’t see a reason to add this info. Otherwise LGTM

ckohen: I added it because threads are for more likely to be out of cache (once they're archived at least). Although by the point they're archived for 4 hours (assuming default options here) the message has probably dropped out of cache too.

ImRodry: Yeah that sounds better

SpaceEEC (approved):

ImRodry (approved):

vladfrangu (approved):

kyranet (approved):

PR #6230, opened 2021-07-30
types: use api-types for stricter typings

ckohen:

-types is now a real dep so lets use it to the fullest

(sorry about the diff...its kinda massive, tried to make it easier by separating the raw tpes)

### DIFF
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -17,17 +17,23 @@ import { Collection } from '@discordjs/collection';
 import { ChildProcess } from 'child_process';
 import {
   APIActionRowComponent,
+  APIApplicationCommand,
+  APIApplicationCommandInteractionData,
+  APIApplicationCommandInteractionDataOption,
+  APIApplicationCommandPermission,
+  APIAuditLogChange,
+  APIEmoji,
   APIInteractionDataResolvedChannel,
   APIInteractionDataResolvedGuildMember,
   APIInteractionGuildMember,
   APIMessage,
   APIMessageComponent,
   APIOverwrite,
+  APIPartialChannel,
   APIPartialEmoji,
+  APIPartialGuild,
   APIRole,
-  APISticker,
-  APIStickerItem,
-  APIStickerPack,
+  APITemplateSerializedSourceGuild,
   APIUser,
   GatewayVoiceServerUpdateDispatchData,
   GatewayVoiceStateUpdateDispatchData,
@@ -35,6 +41,7 @@ import {
 } from 'discord-api-types/v9';
 import { EventEmitter } from 'events';
 import { Stream } from 'stream';
+import { MessagePort, Worker } from 'worker_threads';
 import * as WebSocket from 'ws';
 import {
   ActivityTypes,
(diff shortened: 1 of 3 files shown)


ckohen:
And the instant rebase has been done

typings/index.d.ts
```diff
@@ -238,11 +303,12 @@ export class BitField<S extends string, N extends number | bigint = number> {
   public toJSON(): N extends number ? number : string;
   public valueOf(): N;
   public [Symbol.iterator](): IterableIterator<S>;
-  public static FLAGS: unknown;
+  public static FLAGS: Record<string, number | BigInt>;
```
kyranet, line 306:
```suggestion
  public static FLAGS: Record<string, number | bigint>;
```

typings/index.d.ts
```diff
@@ -973,9 +1045,15 @@ export class LimitedCollection<K, V> extends Collection<K, V> {
...
-  private _update(data: unknown, partial?: boolean): Message;
+  public constructor(
+    client: Client,
+    data: RawMessageData,
+    channel: TextChannel | DMChannel | NewsChannel | ThreadChannel,
+  );
+  private _patch(data: RawPartialMessageData, partial: true): Message;
+  private _patch(data: RawMessageData): Message;
+  private _update(data: RawPartialMessageData, partial: true): Message;
+  private _update(data: RawMessageData): Message;
```
kyranet, line 1056:
```suggestion
  private _patch(data: RawPartialMessageData, partial: true): Message;
  private _patch(data: RawMessageData, partial?: false): Message;
  private _update(data: RawPartialMessageData, partial: true): Message;
  private _update(data: RawMessageData, partial?: false): Message;
```

typings/index.d.ts
```diff
@@ -2573,7 +2662,7 @@ export interface WebhookFields extends PartialWebhookFields {
   readonly createdTimestamp: number;
   delete(reason?: string): Promise<void>;
   edit(options: WebhookEditData, reason?: string): Promise<Webhook>;
-  sendSlackMessage(body: object): Promise<boolean>;
+  sendSlackMessage(body: unknown): Promise<boolean>;
```
kyranet, line 2665: Is it not possible to type this?

kyranet (changes requested):
I love this, just a few changes:

typings/index.d.ts
vladfrangu: We'd need to type it off of Slack's documentation..

typings/index.d.ts
```diff
@@ -2800,9 +2889,9 @@ export type ApplicationFlagsString =
   | 'EMBEDDED';
 
 export interface AuditLogChange {
-  key: string;
-  old?: unknown;
-  new?: unknown;
+  key: APIAuditLogChange['key'];
+  old?: APIAuditLogChange['old_value'];
+  new?: APIAuditLogChange['new_value'];
 }
```
vladfrangu, line 2895: Does this... pick up the unions? If it does, that's awesome!

typings/index.d.ts
```diff
@@ -1292,19 +1371,20 @@ export class NewsChannel extends TextBasedChannel(GuildChannel) {
...
 
 export class OAuth2Guild extends BaseGuild {
+  constructor(client: Client, data: RawOAuth2GuildData);
```
vladfrangu, line 1374:
```suggestion
  public constructor(client: Client, data: RawOAuth2GuildData);
```

typings/index.d.ts
vladfrangu: Shouldn't this be `Record<S, bigint>` instead?

vladfrangu (dismissed):

typings/index.d.ts
kyranet: This is BitField, which can be either.

vladfrangu: Disregarding that bigint part, the key is still relevant

NotSugden: `FLAGS` is static so you can't use the type parameter 😭

typings/index.d.ts
iCrawl: Yeah let's not.

typings/index.d.ts
ckohen: It does! It's quite awesome
image
image

advaith1:
why are some types in rawDataTypes just re-exports from d-api-types? cant those be used directly?

ckohen:
Because I wanted one place for all raw data types, in the event that any of them change it changes in one place and not multiple. Technically speaking I de-duplicated a lot of that, most of the ones that simply seem like reexports are deduplicated versions of a gateway event and 5 different rest endpoint results.

SpaceEEC (approved):

kyranet (approved):

iCrawl (approved):

vladfrangu (approved):

PR #6231, opened 2021-07-30
refactor: move member adding to manager

ckohen:

The long awaited move is here.
Also comes with a bug fix to the method (throws properly when a role is invalid), along with no more param mutation.
Further adds a force param for those of you that don't get member events (if a member has left while the bot is running, they still are cached 😢 )

### DIFF
--- a/src/managers/GuildMemberManager.js
+++ b/src/managers/GuildMemberManager.js
@@ -77,19 +77,21 @@ class GuildMemberManager extends CachedManager {
    * @property {boolean} [mute] Whether the member should be muted (requires `MUTE_MEMBERS`)
    * @property {boolean} [deaf] Whether the member should be deafened (requires `DEAFEN_MEMBERS`)
    * @property {boolean} [force] Whehter to skip the cache check and call the API directly
+   * @property {boolean} [fetchWhenExisting=true] Whether to fetch the user if not cached and already a member
    */
 
   /**
    * Adds a user to the guild using OAuth2. Requires the `CREATE_INSTANT_INVITE` permission.
    * @param {UserResolvable} user The user to add to the guild
    * @param {AddGuildMemberOptions} options Options for adding the user to the guild
-   * @returns {Promise<GuildMember>}
+   * @returns {Promise<GuildMember|null>}
    */
   async add(user, options) {
     const userId = this.client.users.resolveId(user);
     if (!userId) throw new TypeError('INVALID_TYPE', 'user', 'UserResolvable');
-    if (!options?.force) {
-      if (this.cache.has(userId)) return this.cache.get(userId);
+    if (!options.force) {
+      const cachedUser = this.cache.get(userId);
+      if (cachedUser) return cachedUser;
     }
     options.access_token = options.accessToken;
     if (options.roles) {
@@ -106,7 +108,11 @@ class GuildMemberManager extends CachedManager {
     }
     const data = await this.client.api.guilds(this.guild.id).members(userId).put({ data: options });
     // Data is an empty buffer if the member is already part of the guild.
-    return data instanceof Buffer ? this.members.fetch(userId) : this._add(data);
+    let newUser = data instanceof Buffer ? null : this._add(data);
+    if (!newUser && (options.fetchWhenExisting ?? true)) {
+      newUser = this.fetch(userId);
+    }
+    return newUser;
   }
 
   /**
--- a/src/managers/GuildMemberManager.js
+++ b/src/managers/GuildMemberManager.js
@@ -67,6 +67,49 @@ class GuildMemberManager extends CachedManager {
   }
 
   /**
+   * Options used to add a user to a guild using OAuth2.
+   * @typedef {Object} AddGuildMemberOptions
+   * @property {string} accessToken An OAuth2 access token for the user with the `guilds.join` scope granted to the
+   * bot's application
+   * @property {string} [nick] The nickname to give to the member (requires `MANAGE_NICKNAMES`)
+   * @property {Collection<Snowflake, Role>|RoleResolvable[]} [roles] The roles to add to the member
+   * (requires `MANAGE_ROLES`)
+   * @property {boolean} [mute] Whether the member should be muted (requires `MUTE_MEMBERS`)
+   * @property {boolean} [deaf] Whether the member should be deafened (requires `DEAFEN_MEMBERS`)
+   * @property {boolean} [force] Whehter to skip the cache check and call the API directly
+   */
+
+  /**
+   * Adds a user to the guild using OAuth2. Requires the `CREATE_INSTANT_INVITE` permission.
+   * @param {UserResolvable} user The user to add to the guild
+   * @param {AddGuildMemberOptions} options Options for adding the user to the guild
+   * @returns {Promise<GuildMember>}
+   */
+  async add(user, options) {
+    const userId = this.client.users.resolveId(user);
+    if (!userId) throw new TypeError('INVALID_TYPE', 'user', 'UserResolvable');
+    if (!options?.force) {
+      if (this.cache.has(userId)) return this.cache.get(userId);
+    }
+    options.access_token = options.accessToken;
+    if (options.roles) {
+      if (!Array.isArray(options.roles) && !(options.roles instanceof Collection)) {
+        throw new TypeError('INVALID_TYPE', 'options.roles', 'Array or Collection of Roles or Snowflakes', true);
+      }
+      const resolvedRoles = [];
+      for (const role of options.roles.values()) {
+        const resolvedRole = this.guild.roles.resolveId(role);
+        if (!resolvedRole) throw new TypeError('INVALID_ELEMENT', 'Array or Collection', 'options.roles', role);
+        resolvedRoles.push(resolvedRole);
+      }
+      options.roles = resolvedRoles;
+    }
+    const data = await this.client.api.guilds(this.guild.id).members(userId).put({ data: options });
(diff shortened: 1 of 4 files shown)


src/managers/GuildMemberManager.js
```diff
@@ -66,6 +66,49 @@ class GuildMemberManager extends CachedManager {
...
+   * Adds a user to the guild using OAuth2. Requires the `CREATE_INSTANT_INVITE` permission.
+   * @param {UserResolvable} user The user to add to the guild
+   * @param {AddGuildMemberOptions} options Options for adding the user to the guild
+   * @returns {Promise<GuildMember>}
+   */
+  async add(user, options) {
+    const userId = this.client.users.resolveId(user);
+    if (!userId) throw new TypeError('INVALID_TYPE', 'user', 'UserResolvable');
+    if (!options?.force) {
+      if (this.cache.has(userId)) return this.cache.get(userId);
```
kyranet, line 92: Avoid has-set:
```suggestion
      const cachedUser = this.cache.get(userId);
      if (cachedUser) return cachedUser;
```

src/managers/GuildMemberManager.js
```diff
@@ -66,6 +66,49 @@ class GuildMemberManager extends CachedManager {
...
+      for (const role of options.roles.values()) {
+        const resolvedRole = this.guild.roles.resolveId(role);
+        if (!resolvedRole) throw new TypeError('INVALID_ELEMENT', 'Array or Collection', 'options.roles', role);
+        resolvedRoles.push(resolvedRole);
+      }
+      options.roles = resolvedRoles;
+    }
+    const data = await this.client.api.guilds(this.guild.id).members(userId).put({ data: options });
+    // Data is an empty buffer if the member is already part of the guild.
+    return data instanceof Buffer ? this.members.fetch(userId) : this._add(data);
```
kyranet, line 109: I know this is the same behaviour as before, but we should probably make the auto-fetch opt-in in the future.

kyranet (changes requested):
Just one change:

src/managers/GuildMemberManager.js
```diff
@@ -66,6 +66,49 @@ class GuildMemberManager extends CachedManager {
...
+  /**
+   * Adds a user to the guild using OAuth2. Requires the `CREATE_INSTANT_INVITE` permission.
+   * @param {UserResolvable} user The user to add to the guild
+   * @param {AddGuildMemberOptions} options Options for adding the user to the guild
+   * @returns {Promise<GuildMember>}
+   */
+  async add(user, options) {
+    const userId = this.client.users.resolveId(user);
+    if (!userId) throw new TypeError('INVALID_TYPE', 'user', 'UserResolvable');
+    if (!options?.force) {
```
NotSugden, line 91:
```suggestion
    if (!options.force) {
```
options is documented as required

src/managers/GuildMemberManager.js
vladfrangu:
```suggestion
    return data instanceof Buffer ? this.fetch(userId) : this._add(data);
```

vladfrangu (dismissed):

src/managers/GuildMemberManager.js
```diff
@@ -66,6 +66,55 @@ class GuildMemberManager extends CachedManager {
...
+        const resolvedRole = this.guild.roles.resolveId(role);
+        if (!resolvedRole) throw new TypeError('INVALID_ELEMENT', 'Array or Collection', 'options.roles', role);
+        resolvedRoles.push(resolvedRole);
+      }
+      options.roles = resolvedRoles;
+    }
+    const data = await this.client.api.guilds(this.guild.id).members(userId).put({ data: options });
+    // Data is an empty buffer if the member is already part of the guild.
+    let newUser = data instanceof Buffer ? null : this._add(data);
+    if (!newUser && (options.fetchWhenExisting ?? true)) {
```
NotSugden, line 112:
```suggestion
    if (!newUser && options.fetchWhenExisting !== false) {
```
i'd personally do this but either works really

NotSugden (commented):
its probably not important, but `fetchWhenExisting` and `accessToken` are being sent to the API when neither fields are valid (i don't believe it causes an error), just thought it worth noting

src/managers/GuildMemberManager.js
```diff
@@ -66,6 +66,60 @@ class GuildMemberManager extends CachedManager {
...
+    if (!options.force) {
+      const cachedUser = this.cache.get(userId);
+      if (cachedUser) return cachedUser;
+    }
+    const resolvedOptions = {
+      nick: options.nick,
+      mute: options.mute,
+      deaf: options.deaf,
+    };
+    resolvedOptions.access_token = options.accessToken;
```
kyranet, line 101:
```suggestion
      deaf: options.deaf,
      access_token: options.accessToken,
    };
```

src/managers/GuildMemberManager.js
```diff
@@ -66,6 +66,60 @@ class GuildMemberManager extends CachedManager {
...
+      }
+      resolvedOptions.roles = resolvedRoles;
+    }
+    const data = await this.client.api.guilds(this.guild.id).members(userId).put({ data: resolvedOptions });
+    // Data is an empty buffer if the member is already part of the guild.
+    let newUser = data instanceof Buffer ? null : this._add(data);
+    if (!newUser && options.fetchWhenExisting !== false) {
+      newUser = this.fetch(userId);
+    }
+    return newUser;
```
kyranet, line 120: You could probably do something like this, I don't think `_add` returns a falsy value.
```suggestion
    return data instanceof Buffer
    	? options.fetchWhenExisting === false
    		? null
    		: this.fetch(userId)
    	: this._add(data);
```

Prettier will most likely complain about formatting, so please do this in VSCode 🙏🏼

typings/index.d.ts
```diff
@@ -2333,6 +2332,8 @@ export class GuildManager extends CachedManager<Snowflake, Guild, GuildResolvabl
 export class GuildMemberManager extends CachedManager<Snowflake, GuildMember, GuildMemberResolvable> {
   public constructor(guild: Guild, iterable?: Iterable<unknown>);
   public guild: Guild;
+  public add(user: UserResolvable, options: AddGuildMemberOptions & { fetchWhenExisting: true }): Promise<GuildMember>;
+  public add(user: UserResolvable, options: AddGuildMemberOptions): Promise<GuildMember | null>;
```
kyranet, line 2336: Probably needs to be done this way? `fetchWhenExisting` defaults to `true`, so we need to check for a required `false` value, then the rest of overloads.
```suggestion
  public add(user: UserResolvable, options: AddGuildMemberOptions & { fetchWhenExisting: false }): Promise<GuildMember | null>;
  public add(user: UserResolvable, options: AddGuildMemberOptions): Promise<GuildMember>;
```

You can also add a type test.

kyranet (changes requested):

iCrawl (approved):

typings/index.ts
```diff
@@ -398,13 +398,38 @@ client.on('ready', async () => {
...
   });
+
+  // @ts-expect-error
+  assertIsMember(g.members.add(testUserId));
+
+  // @ts-expect-error
+  assertIsMember(g.members.add(testUserId, {}));
+
+  // @ts-expect-error
+  assertIsMember(g.members.add(testUserId, { accessToken: 'totallyRealAccessToken', roles: [g.roles.cache] }));
```
kyranet, line 419: What error does this emit? `fetchWhenExisting` defaults to `true`, so this should not yield an error.

typings/index.ts
```diff
@@ -398,13 +398,38 @@ client.on('ready', async () => {
...
+  // @ts-expect-error
+  assertIsMember(g.members.add(testUserId, { accessToken: 'totallyRealAccessToken', fetchWhenExisting: false }));
+
+  assertIsMember(
+    g.members.add(testUserId, {
+      accessToken: 'totallyRealAccessToken',
+      mute: true,
+      deaf: false,
+      roles: [g.roles.cache.first()!],
+      force: true,
```
kyranet, line 430: Probably want to test when it's set to `true`, so we test all three cases (`undefined`, `false`, and `true`):
```suggestion
      roles: [g.roles.cache.first()!],
      fetchWhenExisting: true,
      force: true,
```

typings/index.ts
```diff
@@ -398,13 +398,38 @@ client.on('ready', async () => {
   });
 });
 
+// This is to check that stuff is the right type
+declare const assertIsMember: (m: Promise<GuildMember>) => void;
```
kyranet, line 402: Perhaps name this `assertIsPromiseMember`? Although the one below (`assertIsMessage`) should probably be updated.

kyranet (changes requested):

typings/index.ts
ckohen: This is an invalid `RoleResolvable`

typings/index.ts
ckohen: Do we need to differentiate? I could union type `Promise<GuildMember> | GuildMember`

kyranet: We want type assertions to be as strict as possible, we probably don't want a method that should be returning a `Promise<T>` to accidentally return `T` 😄

typings/index.ts
kyranet: You can write a description next to `// @ts-expect-error`, please do so.

kyranet (approved):

vladfrangu (approved):

SpaceEEC (approved):

PR #6242, opened 2021-07-31
refactor: merge collections with keeping entries at max

1Computer1:

We want to able to keep things indefinitely in limited/swept caches, so might as well just merge the two since the only difference then is the sweeping.
Also "fixes" some typings (TS sucky wucky).
Probably didn't rename everything, let me know.


iCrawl (approved):

src/util/Options.js
```diff
@@ -103,7 +103,7 @@ class Options extends null {
       makeCache: this.cacheWithLimits({
         MessageManager: 200,
         ThreadManager: {
-          sweepFilter: require('./SweptCollection').filterByLifetime({
+          sweepFilter: require('./LimitedCollection').filterByLifetime({
             getComparisonTimestamp: e => e.archiveTimestamp,
             excludeFromSweep: e => !e.archived,
```
ckohen, line 108: This needs a `sweepInterval` time now

src/util/Options.js
```diff
@@ -144,17 +144,17 @@ class Options extends null {
...
    * @returns {CacheFactory}
    * @example
    * // Store up to 200 messages per channel and discard archived threads if they were archived more than 4 hours ago.
    * Options.cacheWithLimits({
    *    MessageManager: 200,
    *    ThreadManager: {
-   *      sweepFilter: SweptCollection.filterByLifetime({
+   *      sweepFilter: LimitedCollection.filterByLifetime({
    *        getComparisonTimestamp: e => e.archiveTimestamp,
    *        excludeFromSweep: e => !e.archived,
```
ckohen, line 159: ditto

ckohen (changes requested):

ckohen (approved):

src/util/LimitedCollection.js
```diff
@@ -1,30 +1,151 @@
...
+        }
+        const comparisonTimestamp = getComparisonTimestamp(entry, key, coll);
+        if (!comparisonTimestamp || typeof comparisonTimestamp !== 'number') return false;
+        return now - comparisonTimestamp > lifetimeMs;
+      };
+    };
+  }
+
+  [_cleanupSymbol]() {
+    return this.interval ? () => clearInterval(this.interval) : null;
```
NotSugden, line 146:
```suggestion
    return this.interval && () => clearInterval(this.interval);
```
since `interval` is either null or an interval you could just do this

src/util/Options.js
```diff
@@ -175,20 +177,31 @@ class Options extends null {
...
+      }
+      if (typeof setting === 'number') {
+        if (setting === Infinity) {
+          return new Collection();
+        }
+        return new LimitedCollection({ maxSize: setting });
+      }
+      /* eslint-disable eqeqeq */
+      const noSweeping =
+        setting.sweepFilter == null ||
```
NotSugden, line 195: `sweepFilter` is a function if i understand correctly?
```suggestion
        !setting.sweepFilter ||
```
```suggestion
        typeof setting.sweepFilter !== 'function' ||
```

src/util/Options.js
```diff
@@ -175,20 +177,31 @@ class Options extends null {
...
+        setting.sweepInterval == null ||
```
NotSugden, line 196: similarly here could do
```suggestion
        typeof setting.sweepInterval !== 'number' ||
```

typings/index.ts
```diff
@@ -59,9 +63,12 @@ const client: Client = new Client({
     // @ts-expect-error
     Message: 100,
     ThreadManager: {
-      sweepInterval: require('./SweptCollection').filterByLifetime({
-        getComparisonTimestamp: (e: any) => e.archiveTimestamp,
-        excludeFromSweep: (e: any) => !e.archived,
+      maxSize: 1000,
+      keepOverLimit: (x: ThreadChannel) => x.id === '123',
```
NotSugden, line 67:
```suggestion
      keepOverLimit: (x: ThreadChannel) => x.id === '123456789012345678',
```

kyranet (approved):
Other than Sugden's requested changes, LGTM.

src/util/LimitedCollection.js
iCrawl: Doesn't look like it improves the code with those changes.

test/tester2000.js
```diff
@@ -12,6 +12,14 @@ const client = new Client({
   makeCache: Options.cacheWithLimits({
     MessageManager: 10,
     PresenceManager: 10,
+    UserManager: {
+      maxSize: 1,
+      keepOverLimit: v => v.id === client.user.id,
```
vladfrangu, line 17: We should probably make this the default if possible.. @discordjs/the-big-3 thoughts?

vladfrangu (approved):
LGTM, just one thought

test/tester2000.js
iCrawl: No, we don't even have a default on capping users/members.

ckohen (changes requested):

test/tester2000.js
SpaceEEC: I think all structures in a cached manager have a reference to the client, making this additional parameter (and, to some extend, complexity redundant).
I am not sure I really follow your comment however. 😅

SpaceEEC (approved):

test/tester2000.js
ckohen: Ok ignore me, i'm dumb

iCrawl:
This needs a rebase.

src/util/Options.js
1Computer1: Better to have LimitedCollection constructor error for the other cases.

PR #6243, opened 2021-07-31
Fix Collections import

### LINKED ISSUE #6236: error from Collection util
data:    index.js:18319 - /home/debian/tembot/node_modules/discord.js/src/util/Collection.js:6
data:    index.js:18319 - class Collection extends BaseCollection {
data:    index.js:18319 -                          ^
data:    index.js:18319 - TypeError: Class extends value undefined is not a constructor or null
data:    index.js:18319 -     at Object.<anonymous> (/home/debian/tembot/node_modules/discord.js/src/util/Collection.js:6:26)

I believe this maybe because of https://github.com/discordjs/discord.js/commit/dbb59ba1b29b2f75dd8faab5c3004ade51598abc

BannerBomb:
This fixes #6236 which was caused by the @discordjs/collections being deconstructured when importing.

### DIFF
--- a/src/util/Collection.js
+++ b/src/util/Collection.js
@@ -1,6 +1,6 @@
 'use strict';
 
-const { Collection: BaseCollection } = require('@discordjs/collection');
+const BaseCollection = require('@discordjs/collection');
 const Util = require('./Util');
 
 class Collection extends BaseCollection {


ckohen:
The version of Collection used no longer has an export that can be used as such, destructuring is required. The fix to #6236 is updating dependencies. This would actually break the lib.

BannerBomb:
Oh, yea you're right. I also had an older @discordjs/collections installed when I checked the export.

kyranet:
Closing this PR as it's superseded/fixed by #6246

PR #6244, opened 2021-07-31
refactor(Message): accept a single object instead of 3 arguments

ImRodry:
This changes the Message#startThread parameters to be consistent with ThreadManager#create to also prevent the need for any breaking changes in the future if extra parameters need to be added.

### DIFF
--- a/src/structures/Message.js
+++ b/src/structures/Message.js
@@ -742,7 +742,7 @@ class Message extends Base {
    * @param {StartThreadOptions} [options] Options for starting a thread on this message
    * @returns {Promise<ThreadChannel>}
    */
-  startThread({ name, autoArchiveDuration, reason } = {}) {
+  startThread(options = {}) {
     if (!['GUILD_TEXT', 'GUILD_NEWS'].includes(this.channel.type)) {
       return Promise.reject(new Error('MESSAGE_THREAD_PARENT'));
     }
--- a/src/structures/Message.js
+++ b/src/structures/Message.js
@@ -747,7 +747,7 @@ class Message extends Base {
       return Promise.reject(new Error('MESSAGE_THREAD_PARENT'));
     }
     if (this.hasThread) return Promise.reject(new Error('MESSAGE_EXISTING_THREAD'));
-    return this.channel.threads.create({ name, autoArchiveDuration, startMessage: this, reason });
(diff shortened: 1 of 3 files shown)


src/structures/Message.js
```diff
@@ -714,15 +714,22 @@ class Message extends Base {
...
    * Create a new public thread from this message
    * @see ThreadManager#create
-   * @param {string} name The name of the new Thread
-   * @param {ThreadAutoArchiveDuration} autoArchiveDuration How long before the thread is automatically archived
-   * @param {string} [reason] Reason for creating the thread
+   * @param {StartThreadOptions} [options] Options for starting a thread on this message
    * @returns {Promise<ThreadChannel>}
    */
-  startThread(name, autoArchiveDuration, reason) {
+  startThread({ name, autoArchiveDuration, reason } = {}) {
```
NotSugden, line 732: what about 
```suggestion
  startThread(options = {}) {
```
and then below
`create({ ...options, startMessage: this });`

ImRodry: Thought of this but didn't know the best way to implement it, thanks

typings/index.d.ts
```diff
@@ -4316,12 +4318,9 @@ export type ThreadChannelResolvable = ThreadChannel | Snowflake;
 
 export type ThreadChannelTypes = 'GUILD_NEWS_THREAD' | 'GUILD_PUBLIC_THREAD' | 'GUILD_PRIVATE_THREAD';
 
-export interface ThreadCreateOptions<AllowedThreadType> {
-  name: string;
-  autoArchiveDuration: ThreadAutoArchiveDuration;
+export interface ThreadCreateOptions<AllowedThreadType> extends StartThreadOptions {
```
shinotheshino, line 4321: Shouldn't this be:
```suggestion
export interface ThreadCreateOptions<AllowedThreadType extends ThreadChannelTypes> extends StartThreadOptions {
```
?

kyranet (approved):

typings/index.d.ts
ImRodry: No becauseThreadChannelTypes contains types that AllowedThreadTypes doesn’t

vladfrangu (approved):

SpaceEEC (approved):

PR #6245, opened 2021-07-31
fix: array/keyArray removed

1Computer1:

fix


kyranet (approved):
We probably want to add Collection#indexOf at some point so we can get rid of the arrays, but yeah.

iCrawl (approved):

PR #6247, opened 2021-07-31
typings(ApplicationCommandOptionData): Align command data options types more with discord API

suneettipirneni:

Discord outlines that some combinations aren't valid in application command options.

> `choices?`: choices for STRING, INTEGER, and NUMBER types for the user to pick from
> `options?`: If the option is a subcommand or subcommand group type, this nested options will be the parameters

This PR makes any options not adhering to the api spec cause a compile-time error. I didn't put too much thought into naming some of the types, so alternatives are welcome.


advaith1:
i think this is implemented in discord-api-types, have you looked at that?

suneettipirneni:
> i think this is implemented in discord-api-types, have you looked at that?

So I do see that it's implemented there, however I've never used `discord-api-types`, and I'm unsure how to use it's typings with discord.js's typings, because the types conflict. I'm probably missing something here, but can I use `discord-api-types` with discord.js?

kyranet (approved):
It'd be good to have some tests in `typings/tests.ts`, but looks good to me :+1:

typings/index.d.ts
```diff
@@ -2889,19 +2889,56 @@ export interface ApplicationCommandData {
...
+export interface ApplicationCommandChoicesData extends BaseApplicationCommandOptionsData {
+  type: CommandOptionChoiceResolvableType;
   choices?: ApplicationCommandOptionChoice[];
-  options?: ApplicationCommandOptionData[];
 }
 
-export interface ApplicationCommandOption extends ApplicationCommandOptionData {
+export interface ApplicationCommandSubOptionsData extends BaseApplicationCommandOptionsData {
+  type: CommandOptionSubOptionResolvableType;
+  options?: this[];
```
vladfrangu, line 2926: This is inaccurate. In the case of a subcommand group, options is just an "array" of subcommands (usually always has just one value anyways)
In the case of a subcommand, options is any valid option type, excluding subcommand/subcommand group

vladfrangu (changes requested):

typings/index.d.ts
suneettipirneni: Good catch, this has been fixed.

suneettipirneni:
> It'd be good to have some tests in `typings/tests.ts`, but looks good to me

Will do

vladfrangu (approved):

SpaceEEC (approved):

PR #6253, opened 2021-07-31
refactor(GuildChannel): use filter method for #members

Keke27210:

This PR modifies `GuildChannel` to use `Collection#filter` for its `members` getter. `BaseGuildVoiceChannel#members` remains unchanged for optimization purposes (one iteration with a for-of loop instead of two using .filter and .map).

Also removes the administrator permission ignore from `GuildChannel#members`, not sure why this was set? If someone has got Administrator permission, even if they don't have View Channel, they are still supposed to see that channel so they should be included in their members.

### DIFF
--- a/src/structures/BaseGuildVoiceChannel.js
+++ b/src/structures/BaseGuildVoiceChannel.js
@@ -1,6 +1,5 @@
 'use strict';
 
-const { Collection } = require('@discordjs/collection');
 const GuildChannel = require('./GuildChannel');
 const Permissions = require('../util/Permissions');
 
--- a/src/structures/BaseGuildVoiceChannel.js
+++ b/src/structures/BaseGuildVoiceChannel.js
@@ -1,5 +1,6 @@
 'use strict';
 
+const { Collection } = require('@discordjs/collection');
 const GuildChannel = require('./GuildChannel');
 const Permissions = require('../util/Permissions');
 
@@ -36,7 +37,13 @@ class BaseGuildVoiceChannel extends GuildChannel {
    * @readonly
    */
   get members() {
-    return this.guild.voiceStates.cache.filter(s => s.channelId === this.id && s.member);
+    const coll = new Collection();
+    for (const state of this.guild.voiceStates.cache.values()) {
+      if (state.channelId === this.id && state.member) {
+        coll.set(state.id, state.member);
+      }
+    }
+    return coll;
   }
 
   /**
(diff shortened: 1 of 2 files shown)


src/structures/BaseGuildVoiceChannel.js
```diff
@@ -37,13 +36,7 @@ class BaseGuildVoiceChannel extends GuildChannel {
...
    */
   get members() {
-    const coll = new Collection();
-    for (const state of this.guild.voiceStates.cache.values()) {
-      if (state.channelId === this.id && state.member) {
-        coll.set(state.id, state.member);
-      }
-    }
-    return coll;
+    return this.guild.voiceStates.cache.filter(s => s.channelId === this.id && s.member);
```
kyranet, line 39: This is changing the type from `Collection<Snowflake, GuildMember>` to `Collection<Snowflake, VoiceState>`.

kyranet (changes requested):

src/structures/BaseGuildVoiceChannel.js
Keke27210: You are right, my bad - reverting back to the old code as .filter and .map would end up iterating twice over the collection while this is only one iteration.

kyranet (approved):

src/structures/GuildChannel.js
```diff
@@ -258,13 +257,7 @@ class GuildChannel extends Channel {
...
    */
   get members() {
-    const members = new Collection();
-    for (const member of this.guild.members.cache.values()) {
-      if (this.permissionsFor(member).has(Permissions.FLAGS.VIEW_CHANNEL, false)) {
-        members.set(member.id, member);
-      }
-    }
-    return members;
+    return this.guild.members.cache.filter(m => this.permissionsFor(m).has(Permissions.FLAGS.VIEW_CHANNEL));
```
ckohen, line 260:
```suggestion
    return this.guild.members.cache.filter(m => this.permissionsFor(m).has(Permissions.FLAGS.VIEW_CHANNEL, false));
```

This doesn't do what it sounds like, it's *very* confusing. The reason this exists is because the api lets you overwrite the administrator permission on channel, however, it does not do anything, meaning it needs to be ignored. If you have admin from a role, it automatically sets all permissions to true so the administrator check is a noop for that too.

ckohen (changes requested):

vladfrangu (changes requested):
See ckohen's comment

src/structures/GuildChannel.js
Keke27210: Oh right, wasn't aware of that point

iCrawl (approved):

vladfrangu (approved):

SpaceEEC (approved):

PR #6261, opened 2021-08-01
types(ApplicationCommand): applicationCommandData type over raw type

ColdtQ:
This PR Changes the typings for the data variable inside ApplicationCommand. As of now, it requires RawApplicationCommandData, which is the incorrect type. This fixes that by reverting to the ApplicationCommandData type.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5,6 +5,7 @@
   "requires": true,
   "packages": {
     "": {
+      "name": "discord.js",
       "version": "13.0.0-dev",
       "license": "Apache-2.0",
       "dependencies": {
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -190,7 +190,7 @@ export abstract class Application extends Base {
 }
 
 export class ApplicationCommand<PermissionsFetchType = {}> extends Base {
-  public constructor(client: Client, data: RawApplicationCommandData, guild?: Guild, guildId?: Snowflake);
+  public constructor(client: Client, data: ApplicationCommandData, guild?: Guild, guildId?: Snowflake);
   public applicationId: Snowflake;
   public readonly createdAt: Date;
   public readonly createdTimestamp: number;


vladfrangu:
Out of curiosity, 
1. how is it the wrong type? Pretty sure the class takes in raw data from Discord
2. How are you affected by this? 👀

ColdtQ:
> Out of curiosity,
> 
> 1. how is it the wrong type? Pretty sure the class takes in raw data from Discord
> 2. How are you affected by this? 👀

The commit before you PR'd breaking changes, worked fine with the type as ApplicationCommandData

1: Changing it to a raw datatype requires the user to input `application_id` and `id` which I don't think is intended

2: What I said above, when doing 

```

new ApplicationCommand(// params, {
// application_id and id are required here
})

```

vladfrangu:
1. It's very much intended, as it's the data Discord gives us anyways, and is required by the class itself:

https://github.com/discordjs/discord.js/blob/626ff85ae7616a59c95d7338c0df9baead412ce3/src/structures/ApplicationCommand.js#L16-L26

2. You're...not supposed to construct an ApplicationCommand instance yourself, we have fetch methods for that!

ColdtQ:
this is what we are doing right now,

```
      this.slashCommand = new ApplicationCommand<Record<string, unknown>>(
        client,
        {
          ...options.slashCommand,
          name: this.name || options.name || '',
          description: this.description || options.description || '',
          options: options.slashCommand?.options || undefined
        }
      );```
      
      This has worked for a while now, it only works when adding application_id and id now. What is the intended way then?

ImRodry:
If you wanna create a slash command just create an object with the type ApplicationCommandData and then create it like you would normally. There are guides that explain how to do all of that at https://v13-prep.discordjs.guide/interactions/registering-slash-commands.html

ColdtQ:
These are supposed to be Guild only...

ImRodry:
So what? You use the same objects for guild and global slash commands and the guide covers both of those

ColdtQ:
> So what? You use the same objects for guild and global slash commands and the guide covers both of those

I'm using the ApplicationCommand class to represent the Command.slashCommand. If I'm not making an instance of a new applicationCommand for each command I can't do Command.slashCommand to get the application command.

ckohen:
I still don't understand why you need to construct an ApplicationCommand yourself, its not meant to be constructed from user data, it only comes from data discord returns, and reverting this makes this type incorrect.

ImRodry:
Like I said you need to create an object of ApplicationCommandData, not construct the class. Most classes in d.js are not meant to be constructed by the user, this PR should just be closed because it is incorrect and you should use the supported methods that are described in the guide.

PR #6262, opened 2021-08-01
refactor(Channels): fix incorrectly shared properties

monbrey:
This addresses a number of issues with various channel properties being defined too generically, resulting in them existing on channel subclasses where they aren't supported by the API.

### Changes

- New class BaseGuildTextChannel extends GuildChannel, implements TextBasedChannel
- TextChannel and NewsChannel both extend BaseGuildTextChannel
- rateLimitPerUser now only exists on TextChannel
- GuildChannel#setTopic moved to BaseGuildTextChannel and StageChannel
- GuildChannel#createInvite moved to BaseGuildTextChannel, BaseGuildVoiceChannel and StoreChannel
- GuildChannel#fetchInvites moved to BaseGuildTextChannel, BaseGuildVoiceChannel and StoreChannel
- BaseGuildTextChannel#setType now has override types

### DIFF
--- a/src/structures/BaseGuildTextChannel.js
+++ b/src/structures/BaseGuildTextChannel.js
@@ -124,12 +124,11 @@ class BaseGuildTextChannel extends GuildChannel {
    *   .then(hooks => console.log(`This channel has ${hooks.size} hooks`))
    *   .catch(console.error);
    */
-  fetchWebhooks() {
-    return this.client.api.channels[this.id].webhooks.get().then(data => {
-      const hooks = new Collection();
-      for (const hook of data) hooks.set(hook.id, new Webhook(this.client, hook));
-      return hooks;
-    });
+  async fetchWebhooks() {
+    const data = await this.client.api.channels[this.id].webhooks.get();
+    const hooks = new Collection();
+    for (const hook of data) hooks.set(hook.id, new Webhook(this.client, hook));
+    return hooks;
   }
 
   /**
@@ -157,15 +156,14 @@ class BaseGuildTextChannel extends GuildChannel {
     if (typeof avatar === 'string' && !avatar.startsWith('data:')) {
       avatar = await DataResolver.resolveImage(avatar);
     }
-    return this.client.api.channels[this.id].webhooks
-      .post({
-        data: {
-          name,
-          avatar,
-        },
-        reason,
-      })
-      .then(data => new Webhook(this.client, data));
+    const data = await this.client.api.channels[this.id].webhooks.post({
+      data: {
+        name,
+        avatar,
+      },
+      reason,
+    });
+    return new Webhook(this.client, data);
   }
 
   /**
--- a/src/structures/BaseGuildTextChannel.js
+++ b/src/structures/BaseGuildTextChannel.js
@@ -1,0 +1,241 @@
+'use strict';
+
+const { Collection } = require('@discordjs/collection');
+const GuildChannel = require('./GuildChannel');
+const Webhook = require('./Webhook');
+const TextBasedChannel = require('./interfaces/TextBasedChannel');
+const MessageManager = require('../managers/MessageManager');
+const ThreadManager = require('../managers/ThreadManager');
+const DataResolver = require('../util/DataResolver');
+
+/**
+ * Represents a voice-based guild channel on Discord.
+ * @extends {GuildChannel}
+ * @implements {TextBasedChannel}
+ */
+class BaseGuildTextChannel extends GuildChannel {
+  /**
+   * @param {Guild} guild The guild the text channel is part of
+   * @param {APIChannel} data The data for the text channel
+   * @param {Client} [client] A safety parameter for the client that instantiated this
+   */
+  constructor(guild, data, client) {
+    super(guild, data, client);
+
+    /**
+     * A manager of the messages sent to this channel
+     * @type {MessageManager}
+     */
+    this.messages = new MessageManager(this);
+
+    /**
+     * A manager of the threads belonging to this channel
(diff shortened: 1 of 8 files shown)


src/structures/BaseGuildTextChannel.js
```diff
@@ -0,0 +1,241 @@
...
+const { Collection } = require('@discordjs/collection');
+const GuildChannel = require('./GuildChannel');
+const Webhook = require('./Webhook');
+const TextBasedChannel = require('./interfaces/TextBasedChannel');
+const MessageManager = require('../managers/MessageManager');
+const ThreadManager = require('../managers/ThreadManager');
+const DataResolver = require('../util/DataResolver');
+
+/**
+ * Represents a voice-based guild channel on Discord.
```
Jiralite, line 12:
```suggestion
 * Represents a text-based guild channel on Discord.
```

src/structures/BaseGuildTextChannel.js
```diff
@@ -0,0 +1,241 @@
...
+ * @implements {TextBasedChannel}
+ */
+class BaseGuildTextChannel extends GuildChannel {
+  /**
+   * @param {Guild} guild The guild the text channel is part of
+   * @param {APIChannel} data The data for the text channel
+   * @param {Client} [client] A safety parameter for the client that instantiated this
+   */
+  constructor(guild, data, client) {
+    super(guild, data, client);
```
ckohen, line 23: Similar to `GuildChannel` and `Channel` this probably needs to override the default behavior of patching immediately. Surprisingly it hasn't been an issue up to this point, but there is the potential for trying to access `this.messages` in _patch before it gets initialized

Jiralite (changes requested):

ImRodry:
Should probably link #4831

src/structures/TextChannel.js
```diff
@@ -71,36 +18,6 @@ class TextChannel extends GuildChannel {
        */
       this.rateLimitPerUser = data.rate_limit_per_user;
```
SpaceEEC, line 19: The warn now on line 16 can now be removed.

src/structures/TextChannel.js
```diff
@@ -113,93 +30,6 @@ class TextChannel extends GuildChannel {
   setRateLimitPerUser(rateLimitPerUser, reason) {
```
SpaceEEC, line 30: The warn now on line 25 can now be removed.

SpaceEEC (changes requested):

Jiralite (changes requested):
> GuildChannel#setTopic moved to BaseGuildTextChannel

Stage channels would receive this method from `BaseGuildVoiceChannel` which extends `GuildChannel`, which means this would be lost and needs implementing back in somewhere...

Perhaps just stick it in the stage channel itself? That way, we could customise the description of it as it's a unique case for a stage channel?

ckohen:
Stage channel topics seem to have no purpose at this point. The api accepts and sends topic on stage channels, but it doesn't exist in the UI. I suspect this is a remnant after moving to stage instances. But, it probably should be added back directly.

Jiralite:
Is it still documented that it can be used in a stage channel? I see only text channels and announcement channels [here](https://discord.com/developers/docs/resources/channel#modify-channel) upon editing a channel, which is how it is currently called!

Despite the API accepting it, if it does nothing and doesn't seem to be documented... probably best to not add it back in, since stage instances are the way to go right now?

ckohen:
If its undocumented, lets yeet it.

src/structures/GuildChannel.js
```diff
@@ -48,7 +49,7 @@ class GuildChannel extends Channel {
      */
     this.permissionOverwrites = new PermissionOverwriteManager(this);
 
-    this._patch(data);
+    if (data && immediatePatch) this._patch(data);
```
NotSugden, line 52: you can just pass `immediatePatch` into the super and simply remove this patch call because _patch is called in `channel`

src/structures/BaseGuildTextChannel.js
```diff
@@ -0,0 +1,241 @@
...
+   *   .then(hooks => console.log(`This channel has ${hooks.size} hooks`))
+   *   .catch(console.error);
+   */
+  fetchWebhooks() {
+    return this.client.api.channels[this.id].webhooks.get().then(data => {
+      const hooks = new Collection();
+      for (const hook of data) hooks.set(hook.id, new Webhook(this.client, hook));
+      return hooks;
+    });
+  }
```
NotSugden, line 133: make this an async function (#6210)

src/structures/BaseGuildTextChannel.js
```diff
@@ -0,0 +1,241 @@
...
+    }
+    return this.client.api.channels[this.id].webhooks
+      .post({
+        data: {
+          name,
+          avatar,
+        },
+        reason,
+      })
+      .then(data => new Webhook(this.client, data));
```
NotSugden, line 168: #6210

src/structures/GuildChannel.js
monbrey: Not sure I can. The patch on this was hard-coded to false, I don't want to override that by setting a default here.

NotSugden: isn't it the same behaviour both ways? im not crazy?

iCrawl:
This needs a rebase.

iCrawl (approved):

kyranet (approved):

vladfrangu (approved):

SpaceEEC (approved):

Issue #6168, opened 2021-07-22
Bug: message#channel still returns old TextChannel instance after change to news type channel


almostSouji:
**Please describe the problem you are having in as much detail as possible:**

`message#channel` still returns old TextChannel instance after change to news type channel

**Include a reproducible code sample here, if possible:**

1) create a text channel
2) make it a news channel (can be manual through GUI)
3) get `<Message>.channel` inside this (now news) channel

observe: you get a TextChannel of type GUILD_TEXT
reason: https://github.com/discordjs/discord.js/blob/cd47a524afa9cd5af0dfe05d98fa7ab473192bcc/src/client/actions/ChannelUpdate.js#L15-L19

 we do transfer the messages to the new channel, however, each message still has the old channel (due to Message#channel not being a getter but a value)

**Further details:**

- discord.js version: 328501b
- Node.js version: 16.5
- Operating system: n/a
- Priority this issue should have – please be realistic and elaborate if possible: medium

**Relevant client options:**

- partials: n/a
- gateway intents: `GUILDS`, `GUILD_MESSAGES`
- other: n/a

almostSouji:
The most straight-forward and least hacky solution I can think of would be to make Message#channel a getter which retrieves this from cache instead. I can't find any reason so far why this would benefit from being a fix value vs. a getter

Alternatively this could be addressed by iterating over the message cache and setting the new channel to each message (though that seems to be quite the elaborate and hacky work-around)

PR #6271, opened 2021-08-02
fix(Message): make #channel and #guild getters

almostSouji:

While the direct issue outlined in #6168 has been fixed in #6263, the same issue (`Message#channel` returning a text channel) still persists for old messages sent in this same channel before it was converted. As outlined in the issue the least intrusive way to fix this I could think of is making `Message#channel` a getter to retrieve the channel from cache on demand. I use the Client cache (versus the guild cache) to be robust against DMs, if that has any caveats I did not think of, please let me know

- Make `Message#channel` a getter
- Change`Message#guild` getter (relied on `Message#channel` before)
- remove the extra channel from the Message constructor and `MessageManager#_add` call
- use ids within Message structure when possible
- use `Message#channelId` whenever possible in other structures
- shortcut `Message#channel#guild` to `Message#guild`

| :warning: technically this has the caveat of evicting the channel from cache and trying to do operations with the message that rely on its channel. to fix this a plethora of operations on the Message class would need to be changed to use optional chaining or even throw if the channel is not available. however for now we decided to handle it like this: <br><br> DO NOT UNCACHE CHANNELS IF YOU NEED THEM TO DO THINGS |
| --- |

### DIFF
--- a/src/structures/Message.js
+++ b/src/structures/Message.js
@@ -27,16 +27,21 @@ class Message extends Base {
   /**
    * @param {Client} client The instantiating client
    * @param {APIMessage} data The data for the message
-   * @param {TextBasedChannels} channel The channel the message was sent in
    */
-  constructor(client, data, channel) {
+  constructor(client, data) {
     super(client);
 
     /**
-     * The channel that the message was sent in
-     * @type {TextBasedChannels}
+     * The id of the channel the message was sent in
+     * @type {Snowflake}
      */
-    this.channel = channel;
+    this.channelId = data.channel_id;
+
+    /**
+     * The id of the guild the message was sent in, if any
+     * @type {?Snowflake}
+     */
+    this.guildId = data.guild_id ?? null;
 
     /**
      * Whether this message has been deleted
@@ -334,6 +339,15 @@ class Message extends Base {
   }
 
   /**
+   * The channel that the message was sent in
+   * @type {TextChannel|DMChannel|NewsChannel|ThreadChannel}
+   * @readonly
+   */
+  get channel() {
+    return this.client.channels.resolve(this.channelId);
+  }
+
+  /**
    * Whether or not this message is a partial
    * @type {boolean}
    * @readonly
@@ -376,7 +390,7 @@ class Message extends Base {
    * @readonly
    */
   get guild() {
-    return this.channel.guild ?? null;
+    return this.client.guilds.resolve(this.guildId);
   }
 
   /**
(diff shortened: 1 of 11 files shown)


src/structures/Message.js
```diff
@@ -333,6 +338,15 @@ class Message extends Base {
     return clone;
   }
 
+  /**
+   * The channel that the message was sent in
+   * @type {TextChannel|DMChannel|NewsChannel|ThreadChannel}
```
kyranet, line 343:
```suggestion
   * @type {?(TextChannel|DMChannel|NewsChannel|ThreadChannel)}
```

src/structures/MessageMentions.js
```diff
@@ -70,7 +70,7 @@ class MessageMentions {
       } else {
         this.roles = new Collection();
         for (const mention of roles) {
-          const role = message.channel.guild.roles.cache.get(mention);
+          const role = message.guild.roles.cache.get(mention);
```
kyranet, line 73: We probably want to store `guild` outside of the loop, as well as guard it, so we don't execute the getter's logic multiple times.

kyranet (changes requested):

src/structures/Message.js
almostSouji: we get channels with the initial guild payload or channel creates
unless people severely mess things up (sweep channels during runtime) this can not return null

src/structures/Message.js
```diff
@@ -396,7 +410,7 @@ class Message extends Base {
    * @readonly
    */
   get thread() {
-    return this.channel.threads.resolve(this.id);
+    return this.channel?.threads.resolve(this.id) ?? null;
```
NotSugden, line 413: https://github.com/discordjs/discord.js/pull/6271/files#r681339770

if it can never be null then ...?
```suggestion
    return this.channel.threads.resolve(this.id);
```

src/structures/MessagePayload.js
```diff
@@ -167,9 +167,8 @@ class MessagePayload {
 
     let message_reference;
     if (typeof this.options.reply === 'object') {
-      const message_id = this.isMessage
-        ? this.target.channel.messages.resolveId(this.options.reply.messageReference)
-        : this.target.messages.resolveId(this.options.reply.messageReference);
+      const reference = this.options.reply.messageReference;
+      const message_id = this.isMessage ? reference.id ?? reference : this.target.messages.resolveId(reference);
```
NotSugden, line 171: the same logic goes for here

src/structures/Message.js
almostSouji: fully documenting and typing it as nullable is not feasibly possible with how the library is designed at the moment, yet decided to handle it when easily possible, which is the case here, since the return value was already nullable.

iCrawl:
This needs a rebase.

kyranet (approved):

vladfrangu (approved):

SpaceEEC (approved):

PR #6274, opened 2021-08-02
docs: Correct wrong interaction links

Jiralite:
Some links related to interactions were wrong (you can tell straight away because there was no `-` separating the words!) - I've gone ahead and pointed them to the correct area.

`APIInteractionDataResolvedChannel` & `APIInteractionDataResolvedGuildMember` were missing documentation links, so I've gone ahead and pointed them to https://discord.com/developers/docs/interactions/slash-commands#interaction-object-application-command-interaction-data-resolved-structure instead of the usual structure's object. This is because in this link, it tells you _exactly_ what to expect from the data received rather than looking at the whole object and taking guesses:
> Partial Member objects are missing user, deaf and mute fields
> Partial Channel objects only have id, name, type and permissions fields

I've also changed `CommandInteractionOption`'s properties to use `APIInteractionDataResolvedChannel` and `APIInteractionDataResolvedGuildMember` as these are more specific and correct.

I would also correct `APIApplicationCommandOptionResolved`, but I am not sure what this is supposed to be. I also cannot find any typings of this. It looks like https://discord.com/developers/docs/interactions/slash-commands#interaction-object-application-command-interaction-data-resolved-structure but I'm not comfortable with this as it has no typings for me to verify against - is anybody able to confirm this? Maybe @vladfrangu since I can't find it even in [discordjs/discord-api-types](https://github.com/discordjs/discord-api-types)?

Additionally, whilst I corrected the link of `APIApplicationCommandPermissions`, I cannot find any typings of this either...

### DIFF
--- a/src/structures/CommandInteraction.js
+++ b/src/structures/CommandInteraction.js
@@ -92,8 +92,8 @@ class CommandInteraction extends Interaction {
    * @property {CommandInteractionOption[]} [options] Additional options if this option is a
    * subcommand (group)
    * @property {User} [user] The resolved user
-   * @property {GuildMember|APIInteractionDataResolvedGuildMember} [member] The resolved member
-   * @property {GuildChannel|APIInteractionDataResolvedChannel} [channel] The resolved channel
+   * @property {GuildMember|APIInteractionDataResolvedOption} [member] The resolved member
+   * @property {GuildChannel|APIInteractionDataResolvedOption} [channel] The resolved channel
    * @property {Role|APIRole} [role] The resolved role
    */
 
@@ -153,11 +153,6 @@ module.exports = CommandInteraction;
  */
 
 /**
- * @external APIInteractionDataResolvedChannel
- * @see {@link https://discord.com/developers/docs/interactions/slash-commands#interaction-object-application-command-interaction-data-resolved-structure}
- */
-
-/**
- * @external APIInteractionDataResolvedGuildMember
+ * @external APIInteractionDataResolvedOption
  * @see {@link https://discord.com/developers/docs/interactions/slash-commands#interaction-object-application-command-interaction-data-resolved-structure}
  */
--- a/src/structures/CommandInteraction.js
+++ b/src/structures/CommandInteraction.js
@@ -92,8 +92,8 @@ class CommandInteraction extends Interaction {
    * @property {CommandInteractionOption[]} [options] Additional options if this option is a
    * subcommand (group)
    * @property {User} [user] The resolved user
-   * @property {GuildMember|APIGuildMember} [member] The resolved member
-   * @property {GuildChannel|APIChannel} [channel] The resolved channel
+   * @property {GuildMember|APIInteractionDataResolvedGuildMember} [member] The resolved member
+   * @property {GuildChannel|APIInteractionDataResolvedChannel} [channel] The resolved channel
    * @property {Role|APIRole} [role] The resolved role
    */
 
@@ -150,4 +150,14 @@ module.exports = CommandInteraction;
 /**
  * @external APIApplicationCommandOptionResolved
  * @see {@link https://discord.com/developers/docs/interactions/slash-commands#interaction-applicationcommandinteractiondataresolved}
+ */
+
+/**
+ * @external APIInteractionDataResolvedChannel
+ * @see {@link https://discord.com/developers/docs/interactions/slash-commands#interaction-object-application-command-interaction-data-resolved-structure}
+ */
+
+/**
+ * @external APIInteractionDataResolvedGuildMember
+ * @see {@link https://discord.com/developers/docs/interactions/slash-commands#interaction-object-application-command-interaction-data-resolved-structure}
  */
--- a/src/structures/CommandInteraction.js
+++ b/src/structures/CommandInteraction.js
@@ -154,5 +154,5 @@ module.exports = CommandInteraction;
 
 /**
  * @external APIInteractionDataResolvedOption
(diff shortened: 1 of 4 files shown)


src/structures/CommandInteraction.js
```diff
@@ -151,3 +151,13 @@ module.exports = CommandInteraction;
...
+
+/**
+ * @external APIInteractionDataResolvedChannel
+ * @see {@link https://discord.com/developers/docs/interactions/slash-commands#interaction-object-application-command-interaction-data-resolved-structure}
+ */
+
+/**
+ * @external APIInteractionDataResolvedGuildMember
+ * @see {@link https://discord.com/developers/docs/interactions/slash-commands#interaction-object-application-command-interaction-data-resolved-structure}
+ */
```
ImRodry, line 163: Since both lead to the same link you should only add 1 and call it APIInteractionDataResolvedOption, and change it where these 2 appear

ImRodry (changes requested):
As for the errors you will need to disable those eslint rules on those lines. If you install the eslint extension it will help you do that

src/structures/CommandInteraction.js
Jiralite: I do not think this is the right approach. These types are imported from discord-api-types, and if they don't exist there, I don't think we should be creating them over here. Also doing this would imply both the return type of a channel and member would be the same, but this is not correct.

ImRodry: Remember that these are not types and only serve for documentation purposes. Once someone clicks either one of those they will be sent to the same page and will need to figure out what they should be looking at. I argue some may even be confused and call these a bug but there's not much we can do. You could also try a weirder syntax used in typescript which would be something like `APIInteractionDataResolvedOption['member']` so that they know what property to look at when they open the page. No clue how the docs would interpret this however

Jiralite: > Remember that these are not types and only serve for documentation purposes

These _are_ types, I'm not sure what you mean?

```TypeScript
export interface APIInteractionDataResolvedChannel extends Required<APIPartialChannel> {
  permissions: Permissions;
}

export interface APIInteractionDataResolvedGuildMember extends Omit<APIGuildMember, 'user' | 'deaf' | 'mute'> {
  permissions: Permissions;
}
```

Changing them away from these types means we wouldn't be using any types at all...

ImRodry: No these are not types, these are only links in the documentation website. I didn't link the index.d.ts file

Jiralite: Okay so, what you're getting at is we should combine `APIInteractionDataResolvedChannel` and `APIInteractionDataResolvedGuildMember` into `APIInteractionDataResolvedOption` and just make one `@external` reference, right?

The other option I was thinking of was just leading to the usual channel object/guild member object page since most other API-types do that.

ImRodry: Yep that’s it

There are many different types of API external links, there’s no set rule for them so that’d be fine

src/structures/CommandInteraction.js
```diff
@@ -92,8 +92,8 @@ class CommandInteraction extends Interaction {
    * @property {CommandInteractionOption[]} [options] Additional options if this option is a
    * subcommand (group)
    * @property {User} [user] The resolved user
-   * @property {GuildMember|APIGuildMember} [member] The resolved member
-   * @property {GuildChannel|APIChannel} [channel] The resolved channel
+   * @property {GuildMember|APIInteractionDataResolvedOption} [member] The resolved member
```
vladfrangu, line 95: This isn't really... accurate, or *strict*. it's... fine probably, but idk. @discordjs/the-big-3 thoughts?

vladfrangu (changes requested):

src/structures/CommandInteraction.js
ImRodry: There’s no better way to indicate that that is the data that you can receive though

SpaceEEC (approved):

src/structures/CommandInteraction.js
vladfrangu: Yea there is.. there's discord-api-types for it too 👀

kyranet (approved):
Other than Vlad's requests, LGTM.

src/structures/CommandInteraction.js
ImRodry: But how do you do that with documentation?

vladfrangu: @iCrawl might be better to ask here, but you could just have links to them like how it was before. Or point to the raw structures in Discord's documentation 🤷‍♂️

src/structures/CommandInteraction.js
```diff
@@ -151,3 +151,8 @@ module.exports = CommandInteraction;
  * @external APIApplicationCommandOptionResolved
  * @see {@link https://discord.com/developers/docs/interactions/slash-commands#interaction-applicationcommandinteractiondataresolved}
  */
+
+/**
+ * @external APIInteractionDataResolvedOption
+ * @see {@link https://discord.com/developers/docs/interactions/slash-commands#interaction-object-application-command-interaction-data-resolved-structure}
```
iCrawl, line 157:
```suggestion
 * @see {@link https://discord.com/developers/docs/interactions/slash-commands#sample-application-command-interaction-application-command-interaction-data-resolved-structure}
```

src/structures/CommandInteraction.js
iCrawl: Nah, I think its fine.

PR #6278, opened 2021-08-02
types(StartThreadOptions): mark autoArchiveDuration as optional

ImRodry:
Since the addition of default auto archive durations to Discord, the autoArchiveDuration parameter is now optional on the Discord API (tested). This PR marks that parameter as optional on the docs and types.

🌊 Upstream PR: https://github.com/discord/discord-api-docs/pull/3567


kyranet (approved):
LGTM, waiting for upstream PR to be landed first.

iCrawl (approved):

vladfrangu (approved):

ImRodry:
Made `this.channel.defaultArchiveDuration` the default for autoArchiveDuration due to the conversation on the upstream PR where we found that Discord always defaults to 24h no matter what this is set to and passing null is ignored so it will replicate the default behavior of the API

iCrawl:
I'm honestly not sure what to do with this PR in its current state and what seems to me like a Discord API bug.

ImRodry:
> I'm honestly not sure what to do with this PR in its current state and what seems to me like a Discord API bug.

I think we should wait to see what the dapi maintainers say about the other PR and if the bug gets fixed. If it does I can revert the last commit

PR #6289, opened 2021-08-03
feat(Collector): return a boolean on checkEnd

ImRodry:
This PR makes a small change to Collector#checkEnd to make it return a boolean representing whether or not the collector ended.

### DIFF
--- a/src/structures/interfaces/Collector.js
+++ b/src/structures/interfaces/Collector.js
@@ -207,7 +207,6 @@ class Collector extends EventEmitter {
   /**
    * Resets the collector's timeout and idle timer.
    * @param {CollectorResetTimerOptions} [options] Options for reseting
-
    */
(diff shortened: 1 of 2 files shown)


SpaceEEC:
While there probably isn't anything inherently wrong with this change, this method is mostly used internally after handling an event to maybe stop the collector.

What's the use case here? Do you have an example?

ImRodry:
Gonna be honest I've never used this method before but I made this change because it made sense in my head that a function called "checkEnd" would return whether the collector ended or not after running that. Also adding this may make it so the function is used more? Not sure but if it's not meant to be used it can always be marked as private

iCrawl (approved):

kyranet (approved):

vladfrangu (approved):

PR #6306, opened 2021-08-05
refactor(Interaction): rename defer to deferReply

vaporox:
*Just one more PR to defer the v13 release*

**Please describe the changes this PR makes and why it should be merged:**

This renames `.defer()` to `.deferReply()` and with it also `InteractionDeferOptions` for more correctness (you do not defer the interaction itself) and better consistency with `.deferUpdate()` and possible upcoming methods.

I wasn't sure if `.deferred` should be renamed to `.replyDeferred` or something like that, so I kept it at the current name for now.

**Status and versioning classification:**

- I know how to update typings and have done so
- This PR includes breaking changes (methods removed or renamed, parameters moved or removed)

PR #6307, opened 2021-08-05
refactor(Interaction): rename `deferred` to `deferredReply`

muhimur9049:

This PR renames `.deferred` to `.deferredReply` in order to be consistent with the recent change from `.defer()` to `.deferReply()` in PR #6306. 

I wasn't too sure whether to do deferredReply or replyDeferred, but the only reason I went with deferredReply instead is that it sounds more consistent with `deferReply()`. If replyDeferred is decided instead I can change it to that. I guess you can see which is better by looking below and deciding based upon that.

```js
if (interaction.deferredReply) { // do stuff }
```
OR
```js
if (interaction.replyDeferred) { // do stuff }
```

**Update**: Just realized this applies for `deferUpdate` too. What about having both a `deferredReply` and `deferredUpdate`? Would that work?

### DIFF
--- a/src/managers/GuildStickerManager.js
+++ b/src/managers/GuildStickerManager.js
@@ -65,13 +65,13 @@ class GuildStickerManager extends CachedManager {
 
   /**
    * Data that resolves to give a Sticker object. This can be:
-   * * An Sticker object
+   * * A Sticker object
    * * A Snowflake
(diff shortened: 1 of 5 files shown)


monbrey:
I think this was fine as it was. The property applied to both `deferReply` and `deferUpdate`, and could apply to any future type of defer. `deferred` is accurated.

PR #6312, opened 2021-08-06
fix(Caching): sweep archived threads in all channel caches

ckohen:

Fixes a bug in the current sweeping of caches where ThreadChannels would remain in the client and guild channel caches, only being removed from the thread channel caches.

Allows a bit more control over how thread channels are swept from these caches without triggering the process warning for overriding the channel caches.

Also fixes a minor documentation error in LimitedCollection.

### DIFF
--- a/src/managers/ChannelManager.js
+++ b/src/managers/ChannelManager.js
@@ -13,7 +13,11 @@ let cacheWarningEmitted = false;
 class ChannelManager extends CachedManager {
   constructor(client, iterable) {
     super(client, Channel, iterable);
-    if (!cacheWarningEmitted && this._cache.constructor.name !== 'Collection') {
+    const defaultCaching =
+      this._cache.constructor.name === 'Collection' ||
+      ((this._cache.maxSize === undefined || this._cache.maxSize === Infinity) &&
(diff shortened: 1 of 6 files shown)


src/managers/ChannelManager.js
```diff
@@ -13,7 +13,11 @@ let cacheWarningEmitted = false;
 class ChannelManager extends CachedManager {
   constructor(client, iterable) {
     super(client, Channel, iterable);
-    if (!cacheWarningEmitted && this._cache.constructor.name !== 'Collection') {
+    const defaultCaching =
+      this._cache.constructor.name === 'Collection' ||
```
NotSugden, line 17: what about `this._cache.constructor === Collection` ?

ckohen: This is an intentional hardcode on the name

It also technically allows for an extension of Collection the way that discord.js used to due it (e.g `Collection extends BaseCollection`) that won't output the error.

iCrawl (approved):
Really liking the utility function 👍

vladfrangu (approved):

kyranet (approved):

SpaceEEC (approved):

PR #6313, opened 2021-08-06
fix(MessageMentions): check guild exists before adding roles

ckohen:

Fixes a hard crash when resolving mentioned roles in previously uncached messages being replied to.
Also passes the guild id that is received with said message reference so that it is more likely the mentions will be resolved.

### DIFF
--- a/src/structures/Message.js
+++ b/src/structures/Message.js
@@ -304,7 +304,7 @@ class Message extends Base {
     }
 
     if (data.referenced_message) {
-      this.channel?.messages._add(data.referenced_message);
+      this.channel?.messages._add({ guild_id: data.message_reference?.guild_id, ...data.referenced_message });
     }
 
     /**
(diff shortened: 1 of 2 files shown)


iCrawl (approved):

src/structures/Message.js
```diff
@@ -304,7 +304,7 @@ class Message extends Base {
     }
 
     if (data.referenced_message) {
-      this.channel?.messages._add(data.referenced_message);
```
vladfrangu, line 307: Isn't this redundant?

vladfrangu (changes requested):

src/structures/MessageMentions.js
```diff
@@ -71,6 +71,7 @@ class MessageMentions {
         this.roles = new Collection();
         const guild = message.guild;
         for (const mention of roles) {
+          if (!guild) break;
```
kyranet, line 74: Can't we move this outside the loop to avoid a check for each role?

kyranet (changes requested):

vladfrangu (approved):

SpaceEEC (approved):

kyranet (approved):

PR #6325, opened 2021-08-06
fix(Message): handle fetched messages

almostSouji:

PR #6271 has introduced a regression, because fetched messages do not include a `guild_id` property in the rest response. This causes `Message#guild` to become `null` for fetched messages, even if they were fetched from a guild channel.

- try to assign the guild id from the channel, if available
- adapt the `Message#guild` getter to also return the guild from the `Message#channel` getter, if necessary and available

### DIFF
--- a/src/structures/Message.js
+++ b/src/structures/Message.js
@@ -41,7 +41,7 @@ class Message extends Base {
      * The id of the guild the message was sent in, if any
      * @type {?Snowflake}
      */
-    this.guildId = data.guild_id ?? null;
+    this.guildId = data.guild_id ?? this.channel?.guild?.id ?? null;
 
     /**
      * Whether this message has been deleted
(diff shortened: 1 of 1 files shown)


vladfrangu (approved):

src/structures/Message.js
```diff
@@ -390,7 +390,7 @@ class Message extends Base {
    * @readonly
    */
   get guild() {
-    return this.client.guilds.resolve(this.guildId);
+    return this.client.guilds.resolve(this.guildId) ?? this.channel?.guild ?? null;
```
kyranet, line 393: Is this change necessary with the constructor's assignment?

ckohen: Likelihood? No
Theres an extremely small niche use where you fetch a message, the channel is not cached, now you have no guiildId, channel gets cached, now you have guild

SpaceEEC (approved):

iCrawl (approved):

PR #6337, opened 2021-08-07
Update Messages.js

aasim-syed:
I have changed some of the  messages in a more appealing manner


kyranet:
This increases verbosity and is inconsistent with many other errors as well as it introduces typos.

I don't see what's the improvement here.

muhimur9049:
> I have changed some of the messages in a more appealing manner

It only seems like one change to me, and not “some”? 🤔 
Also, there’s no space or capital p after the exclamation mark, making it improper grammar.

PR #6342, opened 2021-08-07
types(ImageURLOptions): allow 'gif' format

almeidx:

`<User>.avatarURL({ format: 'gif' });` throws the following:
Type '"gif"' is not assignable to type 'AllowedImageFormat | undefined'.

### DIFF
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -3786,6 +3786,7 @@ export interface HTTPOptions {
 
 export interface ImageURLOptions extends StaticImageURLOptions {
   dynamic?: boolean;
+  format?: DynamicImageFormat;
 }
 
 export interface IntegrationAccount {
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -3784,10 +3784,10 @@ export interface HTTPOptions {
   headers?: Record<string, string>;
 }
 
-export interface ImageURLOptions extends StaticImageURLOptions {
+export type ImageURLOptions = StaticImageURLOptions & {
   dynamic?: boolean;
   format?: DynamicImageFormat;
-}
+};
 
 export interface IntegrationAccount {
   id: string | Snowflake;
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -3784,10 +3784,10 @@ export interface HTTPOptions {
   headers?: Record<string, string>;
 }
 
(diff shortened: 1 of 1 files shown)


NMW03:
1. You can get gif with `<User>.avatarURL({ dynamic: true })`
2. Shouldn't you add `gif` to [AllowedImageFormat](https://github.com/discordjs/discord.js/blob/394d48649fe4faef8c6b30d5c04962588f86261e/typings/index.d.ts#L2708)?

advaith1:
> Shouldn't you add gif to AllowedImageFormat?

no, that's what `DynamicImageFormat` is for:
https://github.com/discordjs/discord.js/blob/394d48649fe4faef8c6b30d5c04962588f86261e/typings/index.d.ts#L3398

typings/index.d.ts
```diff
@@ -3784,9 +3784,10 @@ export interface HTTPOptions {
   headers?: Record<string, string>;
 }
 
-export interface ImageURLOptions extends StaticImageURLOptions {
+export type ImageURLOptions = StaticImageURLOptions & {
```
kyranet, line 3787:
```suggestion
export interface ImageURLOptions extends StaticImageURLOptions {
```

typings/index.d.ts
```diff
@@ -3784,9 +3784,10 @@ export interface HTTPOptions {
...
   dynamic?: boolean;
-}
+  format?: DynamicImageFormat;
+};
```
kyranet, line 3790:
```suggestion
}
```

typings/index.d.ts
kyranet: Or if it complains (although not suggested, as it makes them non-assignable to each other):

```suggestion
export interface ImageURLOptions extends Omit<StaticImageURLOptions, 'format'> {
```

kyranet (changes requested):

typings/index.d.ts
almeidx: I used the former initially, but I then switched to a `type` on the second commit because it was complaining

vladfrangu (approved):

kyranet (approved):

iCrawl (approved):

SpaceEEC (approved):

PR #6345, opened 2021-08-08
docs: grammar fixes

muhimur9049:

This PR fixes grammar regarding plural possessives. 

In some places, it said something like `The value set for a team members's membership state:`. Not only is this incorrect (since it would be `The value set for a team members' membership state:`), but it does make it too consistent with documentation in other places. 

For example, for stickers, nowhere does it say `The value set for a stickers's type:`; instead it says `The value set for a sticker's type:` which uses genitive case, so the same thing should apply with others.

- [x] This PR **only** includes non-code changes, like changes to documentation, README, etc.

### DIFF
--- a/src/structures/GuildTemplate.js
+++ b/src/structures/GuildTemplate.js
@@ -205,7 +205,7 @@ class GuildTemplate extends Base {
   }
 
   /**
-   * When concatenated with a string, this automatically returns the templates's code instead of the template object.
+   * When concatenated with a string, this automatically returns the templates' code instead of the template object.
    * @returns {string}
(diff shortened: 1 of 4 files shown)


DaStormer:
Prefix should be `docs`; `fix` appears in the change-log under bug fixes

muhimur9049:
> Prefix should be `docs`; `fix` appears in the change-log under bug fixes

Oh right, I did `docs:` in the other PRs oops lol

kyranet (approved):

iCrawl (approved):

PR #6367, opened 2021-08-09
chore(exports): allow imports of specific paths

Mesteery:
This allows to import from specific paths.

**Why?**
In my case:
Jest does not currently support `exports`. So when ESM is enabled, and discord.js is imported, Jest resolves its CJS version (`main`). This commit will  allow to import from `discord.js/src/index.mjs` into the source code to avoid problems with Jest.

In other cases it would allow to import a specific path which is rather interesting.

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,8 +10,11 @@
     "typings"
   ],
   "exports": {
-    "require": "./src/index.js",
-    "import": "./src/index.mjs"
+    "./*": "./*",
+    ".": {
+      "require": "./src/index.js",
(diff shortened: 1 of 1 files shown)


sprayx (approved):

iCrawl (approved):

iCrawl:
Does this work? Is this tested?

vladfrangu (approved):
Sure

SpaceEEC (approved):

PR #6377, opened 2021-08-10
fix(VoiceState): Ensure `suppress` is not `undefined`

Jiralite:
`VoiceState#suppress` can be `undefined`. This is achieved by not knowing the state of a `GuildMember`'s `VoiceState` upon them joining a voice channel or a stage channel.

This can be reproduced by starting the bot up using the following code:
```JavaScript
const Client = new (require("discord.js")).Client({
  intents: [
    "GUILD_VOICE_STATES"
  ]
});

Client.on("voiceStateUpdate", oldState => console.log(oldState));

Client.login("token");
```
<details>
<summary><code>console.log(oldState)</code></summary>

```JavaScript
VoiceState {
  guild: <ref *1> Guild {
    id: '534940241002233896',
    name: undefined,
    icon: undefined,
    features: undefined,
    commands: GuildApplicationCommandManager {
      permissions: [ApplicationCommandPermissionsManager],
      guild: [Circular *1]
    },
    members: GuildMemberManager { guild: [Circular *1] },
    channels: GuildChannelManager { guild: [Circular *1] },
    bans: GuildBanManager { guild: [Circular *1] },
    roles: RoleManager { guild: [Circular *1] },
    presences: PresenceManager {},
    voiceStates: VoiceStateManager { guild: [Circular *1] },
    stageInstances: StageInstanceManager { guild: [Circular *1] },
    invites: GuildInviteManager { guild: [Circular *1] },
    deleted: false,
    available: false,
    shardId: 0
  },
  id: '618976181026422814',
  serverDeaf: null,
  serverMute: null,
  selfDeaf: null,
  selfMute: null,
  selfVideo: null,
  sessionId: null,
  streaming: false,
  channelId: null,
  suppress: undefined,
  requestToSpeakTimestamp: null
}
```
</details>

It seems this happens because of this:
https://github.com/discordjs/discord.js/blob/622f398f4b36b00d0a347b548f63524703c2d395/src/client/actions/VoiceStateUpdate.js#L14

Since the guild member's voice state was not cached, a new `VoiceState` was being made. But there is no default value for `VoiceState#suppress`:
https://github.com/discordjs/discord.js/blob/006edca410f5df53f3272f7a402f2305f73ad14b/src/structures/VoiceState.js#L74
This is the source of it being `undefined`.

This pull request ensures the default value of `VoiceState#suppress` is `false`, as it can be assumed that upon joining a stage channel that they are not being suppressed from speaking.


Jiralite:
Actually, it looks like the received packet includes `suppress`, so I just went ahead and supplied that when a new `VoiceState` was being created.

iCrawl (approved):

src/client/actions/VoiceStateUpdate.js
```diff
@@ -11,7 +11,8 @@ class VoiceStateUpdate extends Action {
     if (guild) {
       // Update the state
       const oldState =
-        guild.voiceStates.cache.get(data.user_id)?._clone() ?? new VoiceState(guild, { user_id: data.user_id });
+        guild.voiceStates.cache.get(data.user_id)?._clone() ??
+        new VoiceState(guild, { user_id: data.user_id, suppress: data.suppress });
```
vladfrangu, line 15: why don't we just spread the object as a whole? will that cause issues?

vladfrangu (changes requested):

src/client/actions/VoiceStateUpdate.js
SpaceEEC: That will cause both voice states to be identical, that doesn't sound ideal.
I think it'd be better to default `suppress` to `true` as the PR initially did.

Jiralite: I see your point! The other option I was thinking of was making it `null`, just in case `VoiceState` is swept by the `makeCache` `ClientOption`, since we can't tell what that was? I'm not familiar with the new cache stuff, but if it's no longer in cache and removed whilst they're a speaker in a stage channel, defaulting to `true` if they leave wouldn't be correct here? If I'm totally wrong, let me know and I'll change the default back to `true` (:

Jiralite: I have just tested this out with this:
```JavaScript
const { Client, Options } = require("discord.js");

const client = new Client({
  intents: [
    "GUILD_VOICE_STATES"
  ],
    VoiceStateManager: 3
  })
});

client.on("voiceStateUpdate", oldState => console.log(oldState)); // Let >= 3 people do whatever first, then leave as a speaker

client.login("token");
```
With a couple of accounts (3), their voice states were created accordingly. However, as a speaker of the stage channel, leaving still rendered `suppress` as `undefined`, so I believe defaulting to `null` would be more appropriate than `true` here. Thoughts?

vladfrangu: Personally I think `null` might be more adequate than defaulting to true... @SpaceEEC @discordjs/the-big-3?

SpaceEEC: I'd like us emitting `null` as voice state if we have none in cache, but that's v14 area.

I think making `supress` nullable falls into that area too, but it wouldn't be as drastic.
So it's probably fine? 🤷 

I definitely prefer `null` over `true`, if that is possible.

Jiralite: I think `null` is fine for the next minor version of discord.js (I see this pull request is aiming for version 13.1 release). If so, I'll go ahead and amend this! (:

Jiralite:
This issue also applies to `VoiceState#streaming`. By not knowing the state of a guild member's `VoiceState` upon them stopping streaming or leaving voice (whilst streaming) etc., `VoiceState#streaming` will be `false`, but in actual fact, this should have been `true`.

This wasn't much of an issue before, but since we have adopted this awesome caching behaviour which means sweeping the cache or setting a limit, there now exists a possibility that this is improperly set.

Should a new pull request be made for this, or should I modify this one to expand it also to the other property?

iCrawl:
Since it's the same scope: VoiceState in this case, I think it should be fine putting it into this PR.

Jiralite:
I have now done this!

vladfrangu (approved):
Well this turned major but LGTM

iCrawl:
Now that I see it, maybe we should do a different PR if you want at least the `suppress` fix inside of v13.

If you don't care too much, thats fine.

Jiralite:
I think it's fine to wait for version 14 for this.

I don't think `suppress` will be that much of an issue (`undefined` vs. `null`; both falsey) compared to a possibly wrong default value of `streaming` (`false` when it might actually be `true`), so I think it's fine to wait it out.

SpaceEEC (approved):

kyranet (approved):

iCrawl:
This needs a rebase.

PR #6383, opened 2021-08-11
feat(Message): add isPartial() method

fasterplayer:
This is to providde an easier method to identify whether or not a message is Partial for Typescript Users

### DIFF
--- a/src/structures/Message.js
+++ b/src/structures/Message.js
@@ -810,7 +810,7 @@ class Message extends Base {
   }
 
   /**
-   * Indicates whether this message is a partial.
+   * Indicates whether this message is partial.
    * @returns {boolean}
    */
   isPartial() {
--- a/src/structures/Message.js
+++ b/src/structures/Message.js
@@ -810,7 +810,7 @@ class Message extends Base {
   }
 
   /**
-   * Indicates whether this Message is Partial.
+   * Indicates whether this message is a partial.
    * @returns {boolean}
    */
   isPartial() {
--- a/src/structures/Message.js
+++ b/src/structures/Message.js
@@ -814,7 +814,7 @@ class Message extends Base {
    * @returns {boolean}
    */
    isPartial() {
-    return this.partial
+    return this.partial;
   }
 
   /**
--- a/src/structures/Message.js
+++ b/src/structures/Message.js
@@ -813,7 +813,7 @@ class Message extends Base {
    * Indicates whether this Message is Partial.
    * @returns {boolean}
    */
-   isPartial() {
+  isPartial() {
     return this.partial;
   }
 
--- a/src/structures/Message.js
+++ b/src/structures/Message.js
@@ -810,6 +810,14 @@ class Message extends Base {
   }
 
   /**
+   * Indicates whether this Message is Partial.
+   * @returns {boolean}
+   */
+   isPartial() {
+    return this.partial
+  }
+
+  /**
    * Used mainly internally. Whether two messages are identical in properties. If you want to compare messages
    * without checking all the properties, use `message.id === message2.id`, which is much more efficient. This
    * method allows you to see if there are differences in content, embeds, attachments, nonce and tts properties.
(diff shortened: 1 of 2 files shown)


PuneetGopinath:
I think title should be `feat(Message): add isPartial() method`

fasterplayer:
> I think title should be `feat(Message): add isPartial() method`

Thanks, changed it.

PuneetGopinath:
Also you added
```js
public isPartial(): this is PartialMessage;
```
Shouldn't that be
```js
public isPartial(): boolen;
```

src/structures/Message.js
```diff
@@ -809,6 +809,14 @@ class Message extends Base {
     return this.edit({ attachments: [] });
   }
 
+  /**
+   * Indicates whether this Message is Partial.
```
DaStormer, line 813:
```suggestion
   * Indicates whether this message is a partial.
```

fasterplayer:
> Also you added
> 
> ```js
> public isPartial(): this is PartialMessage;
> ```
> 
> Shouldn't that be
> 
> ```js
> public isPartial(): boolen;
> ```

No, the point of adding this Method is that using this method is that Typescript will know if the message is a PartialMessage.
But the method will still return a boolean.

src/structures/Message.js
```diff
@@ -809,6 +809,14 @@ class Message extends Base {
     return this.edit({ attachments: [] });
   }
 
+  /**
+   * Indicates whether this message is a partial.
```
vladfrangu, line 813:
```suggestion
   * Indicates whether this message is partial.
```

vladfrangu (changes requested):

vladfrangu (approved):

src/structures/Message.js
DaStormer: There are a lot of occurrences that say "a partial" instead of just "partial." Should those be changed? 🤔

wasdennnoch:
I don't see why this is needed. The typings for the full `message` already say that `partial` is always `false`:
https://github.com/discordjs/discord.js/blob/75b48d8d0f70338c75721df3dbd8fc37866383bc/typings/index.d.ts#L1121
And `Partialize` makes sure that for `PartialMessage` it's always `true`:
https://github.com/discordjs/discord.js/blob/75b48d8d0f70338c75721df3dbd8fc37866383bc/typings/index.d.ts#L4296

TS can infer the type of the message based on that:

fasterplayer:
> I don't see why this is needed. The typings for the full `message` already say that `partial` is always `false`:
> https://github.com/discordjs/discord.js/blob/75b48d8d0f70338c75721df3dbd8fc37866383bc/typings/index.d.ts#L1121
> 
> 
> And `Partialize` makes sure that for `PartialMessage` it's always `true`:
> https://github.com/discordjs/discord.js/blob/75b48d8d0f70338c75721df3dbd8fc37866383bc/typings/index.d.ts#L4296
> 
> TS can infer the type of the message based on that:
> 



> I don't see why this is needed. The typings for the full `message` already say that `partial` is always `false`:
> https://github.com/discordjs/discord.js/blob/75b48d8d0f70338c75721df3dbd8fc37866383bc/typings/index.d.ts#L1121
> 
> 
> And `Partialize` makes sure that for `PartialMessage` it's always `true`:
> https://github.com/discordjs/discord.js/blob/75b48d8d0f70338c75721df3dbd8fc37866383bc/typings/index.d.ts#L4296
> 
> TS can infer the type of the message based on that:
> 
> 
> The only situation in which I can imagine `isPartial()` to be useful is when you type something to be of type `Message`, and then can branch if `isPartial()` says that isn't actually a `PartialMessage`. But then the issue isn't that the Discord.JS typings are wrong - it's that your own app's typings are wrong, which say that an object is always of type `Message` when in fact it isn't.

After making a new test, it seems like you are write. I really don't know how i went through this issue. I think that my issue was only because i was trying to check if message was an instance of Message and it was returning true when it was partial. Really sorry for the time wasted. This PR is actually useless. I'm closing it.

PR #6384, opened 2021-08-11
feat(CommandInteractionResolvedData): access to "raw" resolved data

monbrey:
In the absence of a good way to expose individual resolved data as "mentions", this exposed the complete, parsed `interaction.data.resolved` object as a number of Collections, similar to how MessageMentions are provided.

Aim is to do this in a non-breaking way for 13.1.0, hence using the "resolved" property name, leaving "mentions" available if we want to do a better filtering of it somewhere else down the track.

### DIFF
--- a/src/structures/BaseCommandInteraction.js
+++ b/src/structures/BaseCommandInteraction.js
@@ -80,7 +80,7 @@ class BaseCommandInteraction extends Interaction {
    * @typedef {Object} CommandInteractionResolvedData
    * @property {Collection<string, User>} [users] The resolved users
    * @property {Collection<string, GuildMember|APIGuildMember>} [members] The resolved guild members
-   * @property {Collection<string, Role|APIRole>} [roles] The resolved messages
+   * @property {Collection<string, Role|APIRole>} [roles] The resolved roles
    * @property {Collection<string, Channel|APIChannel>} [channels] The resolved channels
    * @property {Collection<string, Message|APIMessage>} [messages] The resolved messages
    */
--- a/src/structures/BaseCommandInteraction.js
+++ b/src/structures/BaseCommandInteraction.js
@@ -79,10 +79,10 @@ class BaseCommandInteraction extends Interaction {
    * Represents the resolved data of a received command interaction.
    * @typedef {Object} CommandInteractionResolvedData
    * @property {Collection<string, User>} users The resolved users
-   * @property {Collection<string, GuildMember|APIGuildMember} members The resolved guild members
-   * @property {Collection<string, Role|APIRole} roles The resolved messages
-   * @property {Collection<string, Channel|APIChannel} channels The resolved channels
-   * @property {Collection<string, Message|APIMessage} messages The resolved messages
+   * @property {Collection<string, GuildMember|APIGuildMember>} members The resolved guild members
+   * @property {Collection<string, Role|APIRole>} roles The resolved messages
+   * @property {Collection<string, Channel|APIChannel>} channels The resolved channels
+   * @property {Collection<string, Message|APIMessage>} messages The resolved messages
    */
 
   /**
--- a/src/structures/BaseCommandInteraction.js
+++ b/src/structures/BaseCommandInteraction.js
@@ -96,9 +96,9 @@ class BaseCommandInteraction extends Interaction {
 
     if (members) {
       result.members = new Collection();
-      for (const member of Object.values(members)) {
-        const user = users[member.id];
-        result.members.set(member.id, this.guild?.members._add({ user, ...member }) ?? member);
+      for (const [id, member] of Object.entries(members)) {
+        const user = users[id];
+        result.members.set(id, this.guild?.members._add({ user, ...member }) ?? member);
       }
     }
 
--- a/src/structures/BaseCommandInteraction.js
+++ b/src/structures/BaseCommandInteraction.js
@@ -1,5 +1,6 @@
 'use strict';
 
+const { Collection } = require('@discordjs/collection');
 const Interaction = require('./Interaction');
 const InteractionWebhook = require('./InteractionWebhook');
 const InteractionResponses = require('./interfaces/InteractionResponses');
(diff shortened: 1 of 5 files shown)


src/structures/CommandInteraction.js
```diff
@@ -18,6 +18,7 @@ class CommandInteraction extends BaseCommandInteraction {
     this.options = new CommandInteractionOptionResolver(
       this.client,
       data.data.options?.map(option => this.transformOption(option, data.data.resolved)) ?? [],
+      this.transformResolved(data.data.resolved),
```
Jiralite, line 21: On a regular `CommandInteraction`, `data.data.resolved` is `undefined`, so when this called `this.transformResolved`, it will throw a `TypeError`!

Jiralite (changes requested):

typings/index.d.ts
```diff
@@ -3255,6 +3257,14 @@ export interface CommandInteractionOption {
   message?: Message | APIMessage;
 }
 
+export interface CommandInteractionResolvedData {
+  users?: Collection<string, User>;
+  members?: Collection<string, GuildMember | APIInteractionDataResolvedGuildMember>;
+  roles?: Collection<string, Role | APIRole>;
+  channels?: Collection<string, GuildChannel | APIInteractionDataResolvedChannel>;
```
Jiralite, line 3264: Shouldn't `channels` here also allow `DMChannel`s?

src/structures/CommandInteraction.js
monbrey, line 21:
```suggestion
      this.transformResolved(data.data.resolved ?? {}),
```

typings/index.d.ts
monbrey: I don't think so - `interaction.options.getChannel()` is typed as GuildChannels. What channel can you select in a DM?

ckohen: Well, cough, sorry but that should also include Thread Channels at least, I'll test if you can select external channels in DMs

ckohen: Channels in DMs will error unless you select the current DM channel, which is currently possible

vladfrangu (changes requested):

src/structures/BaseCommandInteraction.js
```diff
@@ -74,6 +75,64 @@ class BaseCommandInteraction extends Interaction {
...
+   * @private
+   */
+  transformResolved({ members, users, channels, roles, messages }) {
+    const result = {};
+
+    if (members) {
+      result.members = new Collection();
+      for (const [id, member] of Object.entries(members)) {
+        const user = users[id];
+        result.members.set(id, this.guild?.members._add({ user, ...member }) ?? member);
```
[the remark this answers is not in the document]
iCrawl: What?

We already do this for options. It will only be cached if you are in the guild anyway.

typings/index.d.ts
ckohen: TLDR: Channels can contain any channel type except Partial DM Channel from my testing. This should be updated to include those.

src/structures/BaseCommandInteraction.js
```diff
@@ -74,6 +75,64 @@ class BaseCommandInteraction extends Interaction {
     return this.guild?.commands.cache.get(id) ?? this.client.application.commands.cache.get(id) ?? null;
   }
 
+  /**
+   * Represents the resolved data of a received command interaction.
+   * @typedef {Object} CommandInteractionResolvedData
+   * @property {Collection<string, User>} users The resolved users
+   * @property {Collection<string, GuildMember|APIGuildMember>} members The resolved guild members
+   * @property {Collection<string, Role|APIRole>} roles The resolved messages
```
Jiralite, line 83:
```suggestion
   * @property {Collection<string, Role|APIRole>} roles The resolved roles
```

kyranet (approved):

src/structures/BaseCommandInteraction.js
NotSugden, line 101: i think there should be a check to see if `guild` is not-null before adding to the collection, seems a bit silly to add `undefined` to the collection (even if its rare)

src/structures/BaseCommandInteraction.js
```diff
@@ -74,6 +75,64 @@ class BaseCommandInteraction extends Interaction {
...
+      result.users = new Collection();
+      for (const user of Object.values(users)) {
+        result.users.set(user.id, this.client.users._add(user));
+      }
+    }
+
+    if (roles) {
+      result.roles = new Collection();
+      for (const role of Object.values(roles)) {
+        result.roles.set(role.id, this.guild?.roles._add(role) ?? role);
```
NotSugden, line 115: same here

src/structures/BaseCommandInteraction.js
```diff
@@ -74,6 +75,64 @@ class BaseCommandInteraction extends Interaction {
...
+      result.channels = new Collection();
+      for (const channel of Object.values(channels)) {
+        result.channels.set(channel.id, this.client.channels._add(channel, this.guild) ?? channel);
+      }
+    }
+
+    if (messages) {
+      result.messages = new Collection();
+      for (const message of Object.values(messages)) {
+        result.messages.set(message.id, this.channel?.messages?._add(message) ?? message);
```
NotSugden, line 129: same thing here

NotSugden (changes requested):

typings/index.d.ts
monbrey, line 3264:
```suggestion
  channels?: Collection<string, Channel | APIInteractionDataResolvedChannel>;
```

src/structures/BaseCommandInteraction.js
monbrey:
```suggestion
   * @property {Collection<string, Role|APIRole>} roles The resolved roles
```

src/structures/BaseCommandInteraction.js
```diff
@@ -74,6 +75,64 @@ class BaseCommandInteraction extends Interaction {
     return this.guild?.commands.cache.get(id) ?? this.client.application.commands.cache.get(id) ?? null;
   }
 
+  /**
+   * Represents the resolved data of a received command interaction.
+   * @typedef {Object} CommandInteractionResolvedData
+   * @property {Collection<string, User>} [users] The resolved users
+   * @property {Collection<string, GuildMember|APIGuildMember>} [members] The resolved guild members
+   * @property {Collection<string, Role|APIRole>} [roles] The resolved messages
```
monbrey, line 83:
```suggestion
   * @property {Collection<string, Role|APIRole>} [roles] The resolved roles
```

SpaceEEC (approved):

PR #6386, opened 2021-08-11
workflows: add format code

PuneetGopinath:

I just added a format Workflow, which runs prettier in PRs automatically, so it is easy to format code.

### DIFF
--- a/.github/workflows/format.yml
+++ b/.github/workflows/format.yml
@@ -1,0 +1,24 @@
+name: Format Code
+
+on: [pull_request]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2.3.4
+      - name: Use Node.js 16.x
+        uses: actions/setup-node@v2.4.0
+        with:
+          node-version: "16.x"
+      - name: Install dependencies
+        run: npm install
+      - name: Run format
+        run: npm run prettier
+      - name: Commit and push changes
+        uses: EndBug/add-and-commit@v7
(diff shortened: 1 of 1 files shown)


Fyko:
This is already done in the Test workflow https://github.com/discordjs/discord.js/blob/main/.github/workflows/test.yml#L19

PuneetGopinath:
> This is already done in the Test workflow https://github.com/discordjs/discord.js/blob/main/.github/workflows/test.yml#L19

This is eslint.🤔

Fyko:
Yes, but we use the eslint prettier plugin so when eslint runs, prettier also runs.

PuneetGopinath:
> Yes, but we use the eslint prettier plugin so when eslint runs, prettier also runs.

Only if `eslint src --fix` is used, but I that workflow only `lint` script is used which is `eslint src`
So it only checks for prettier style. But the workflow I added automatically formats code and makes a commit

monbrey:
Generally speaking, not a fan of automated commits. People can just see the eslint tests failing and fix their own code, or run the `lint:fix` script themselves before committing.

PuneetGopinath:
> Generally speaking, not a fan of automated commits. People can just see the eslint tests failing and fix their own code, or run the `lint:fix` script themselves before committing.

Yeah, would work. But automated is easy. I don't know if you don't like automated commits, you can close the pr if you don't want, but I would recommend this. It is easy for new contributors too.

iCrawl:
Yeah we don't do automated code commits 👍

PR #6394, opened 2021-08-11
typings: mark enums as const enums

danny-may:
The enum instances arent actually exported, so shouldnt be declared in the typings as being exported.
Declaring them as const enums instead means they will be replaced by their value at compile time, removing any runtime errors when trying to access the enum value.

This means importing `discord.js/typings/enums` and using the enum will no longer result in a runtime error, without impacting existing code or definitions.


DTrombett:
This will probably create the same problem as https://github.com/discordjs/discord.js/pull/6078

danny-may:
I dont see how as you cant currently use indexers on these types because they arent exported, you get a runtime error if you try.

I do agree it would be best of the enums members were exported and these enums werent converted to const enums, but failing that this is the next best thing as it removes the discrepancy between compile time and runtime

suneettipirneni:
> I dont see how as you cant currently use indexers on these types because they arent exported, you get a runtime error if you try.

#6396 *wink*

suneettipirneni:
> @suneettipirneni +1 for that, but youll still have the same error at runtime because the index.js file doesnt export the enums. If youre allowed to add them to the index.js file then ill close this pr in favour of yours

ohhh, lemme look into that

iCrawl:
Neither of those 2 PRs will work if we don't export the enums in our code. We tried this before, there have been 3-4 PRs on this topic already.

danny-may:
Ive tested this pr in my bot and it works fine without the enum being exported in the js because const enums are a compile time only structure. They get replaced by actual values at runtime

iCrawl:
This still holds true:

https://github.com/discordjs/discord.js/pull/6065#discussion_r664458288

I don't see how this PR handles this.

danny-may:
The issue with [#6065 (comment)](https://github.com/discordjs/discord.js/pull/6065#discussion_r664458288) is that the typescript declaration was declaring there to be an instance available at runtime when there wasnt, hence the undefined you get at runtime. Const enums dont declare anything as existing at runtime. As a toy example:

```ts
enum ThisEnumExistsAtRuntime {
    OPTION_1 = 1,
    OPTION_2 = 2,
    OPTION_3 = 3
}

const enum ThisEnumDoesntExistAtRuntime {
    OPTION_A = 'A',
    OPTION_B = 'B',
    OPTION_C = 'C'
}

const x = ThisEnumExistsAtRuntime.OPTION_1 + ThisEnumExistsAtRuntime.OPTION_2;
const y = ThisEnumDoesntExistAtRuntime.OPTION_A + ThisEnumDoesntExistAtRuntime.OPTION_B;
```
compiles to
```js
"use strict";
var ThisEnumExistsAtRuntime;
(function (ThisEnumExistsAtRuntime) {
    ThisEnumExistsAtRuntime[ThisEnumExistsAtRuntime["OPTION_1"] = 1] = "OPTION_1";
    ThisEnumExistsAtRuntime[ThisEnumExistsAtRuntime["OPTION_2"] = 2] = "OPTION_2";
    ThisEnumExistsAtRuntime[ThisEnumExistsAtRuntime["OPTION_3"] = 3] = "OPTION_3";
})(ThisEnumExistsAtRuntime || (ThisEnumExistsAtRuntime = {}));
const x = ThisEnumExistsAtRuntime.OPTION_1 + ThisEnumExistsAtRuntime.OPTION_2;
const y = "A" /* OPTION_A */ + "B" /* OPTION_B */;
```

If you declare `export enum ...` then typescript expects there to be something there at runtime. The current declarations do this when nothing is exported, so something should be done to fix this. Const enums are designed for exactly this situation where you have a logical enum that isnt exported.

I can change this PR to expose them as interfaces if that would be preferable. It would compile to the exact same code, but would be more annoying to use as in this comment: [#6065 (comment)](https://github.com/discordjs/discord.js/pull/6078#pullrequestreview-706216917)

Debabelkacem (approved):

SpaceEEC (dismissed):
Making those `const enum` breaks:
- non-literal string access, e.g. `ChannelTypes[channel.type]` (errors, but in runtime returns a number)
- Reverse mapping, e.g. `ChannelTypes[ChannelTypes.GUILD_TEXT]` (errors, but in runtime returns a string)

---

We actually export those enums under `Constants` (end of declaration):
https://github.com/discordjs/discord.js/blob/cd4029218fc8b67c395b3cd49a121b1bfa38f021/typings/index.d.ts#L2144

If there is a way to have them, but not be importable from outside of the project, that would be a solution here, but not making them `const enum`s.

iCrawl:
This needs a rebase.

danny-may:
> 
> 
> This needs a rebase.

Done!

kyranet (approved):

suneettipirneni:
If this is merged how will this affect `Constants`? Will it not be needed anymore?

danny-may:
You will still need to use constants if you want to do something like `Constants.ChannelTypes[channel.type]`, otherwise you could use either the enum directly or Constants and have the same result

SpaceEEC (approved):
I don't really see the point, but if everything works. 🤷

iCrawl (approved):

PR #6291, opened 2021-08-03
types(MessageButtonOptions): Clean up and export some button option type definitions

suneettipirneni:
**Please describe the changes this PR makes and why it should be merged:**

This is an addendum to #6212, it simply exports the different button option variants, and cleans up definitions.

**Status and versioning classification:**

Please move lines that apply to you out of the comment:
- Code changes have been tested against the Discord API, or there are no code changes
- I know how to update typings and have done so, or typings don't need updating

PR #6396, opened 2021-08-11
types(ApplicationCommandOptionData): Export enums

suneettipirneni:

A lot of our typings use a `*Types` enum for different type definitions, with the recent changes in #6291 and #6212 this has become more prevalent. We decide type properties based on the `type` field. The issue is that a consumer of the library cannot properly type narrow our types because our types are using unexposed enums.

For example:
```typescript
  if (command.type === 'CHAT_INPUT') {
   // Command is Chat input command type.
  } else {
   // Command is a context menu command type.
  }
```

However this ^ doesn't actually work as expected. If you want to type narrow based on `CHAT_INPUT` you have to include:

` && command.type === ApplicationCommandTypes.CHAT_INPUT`
 in the if clause.

This will compile, and produce expected results, however, `ApplicationCommandTypes` isn't exported, you'll get a runtime error for trying to use something that isn't exported:

```
Cannot find module 'discord.js/typings/enums' (...)
```

### DIFF
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -128,30 +128,7 @@ import {
   RawWidgetMemberData,
 } from './rawDataTypes';
 
-export {
-  ActivityTypes,
-  ApplicationCommandOptionTypes,
-  ApplicationCommandPermissionTypes,
-  ApplicationCommandTypes,
-  ChannelTypes,
-  DefaultMessageNotificationLevels,
-  ExplicitContentFilterLevels,
-  InteractionResponseTypes,
-  InteractionTypes,
-  InviteTargetType,
-  MembershipStates,
-  MessageButtonStyles,
-  MessageComponentTypes,
-  MFALevels,
-  NSFWLevels,
-  OverwriteTypes,
-  PremiumTiers,
-  PrivacyLevels,
-  StickerFormatTypes,
-  StickerTypes,
-  VerificationLevels,
-  WebhookTypes,
-} from './enums';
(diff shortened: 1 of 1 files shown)


typings/index.d.ts
```diff
@@ -128,6 +128,31 @@ import {
...
+  MFALevels,
+  NSFWLevels,
+  OverwriteTypes,
+  PremiumTiers,
+  PrivacyLevels,
+  StickerFormatTypes,
+  StickerTypes,
+  VerificationLevels,
+  WebhookTypes,
+} from './enums';
```
DTrombett, line 154: Not really sure but shouldn't this work too?
```suggestion
export * from './enums';
```

suneettipirneni: Yeah that works, probably a better solution as it requires less maintenance/upkeep.

suneettipirneni:
Since this probably won't work, I'm thinking the `*Types` should be excluded from the type definitions and to stick with string instead. I don't know what others think about this?

vaporox:
The reason why they aren't exported is because they exist as actual values in `Constants` (https://github.com/discordjs/discord.js/blob/main/typings/index.d.ts#L2205-L2243), although it looks like many of them are missing a `typeof`

suneettipirneni:
> The reason why they aren't exported is because they exist as actual values in `Constants` here: https://github.com/discordjs/discord.js/blob/main/typings/index.d.ts#L2215-L2253

Gotcha closing this then

PR #6410, opened 2021-08-12
feat(TypeGuards): add `TypeGuards` utility for commonly used data types

### LINKED ISSUE #6409: Add utility type guards for easier type narrowing

With how types are currently setup it can be quite confusing figuring out how to narrow them to the desired type, for example to check if a command is strictly a context menu type you'd have to write this helper out:

```ts
export function isContextMenuCommand(commandData: ApplicationCommandData): commandData is (MessageApplicationCommandData | UserApplicationCommandData) {
  return commandData.type === 'MESSAGE' || 
         commandData.type === 'USER' || 
         commandData.type === Constants.ApplicationCommandTypes.MESSAGE || 
         commandData.type === Constants.ApplicationCommandTypes.USER;
}
```
It's much worse if you want to check if a command option supports sub-options:
```ts
export function isChoiceBasedOption(option: ApplicationCommandOptionData): option is ApplicationCommandChoicesData {
  return option.type === 'BOOLEAN' || 
         option.type === Constants.ApplicationCommandOptionTypes.BOOLEAN ||
         option.type === 'INTEGER' ||
         option.type === Constants.ApplicationCommandOptionTypes.INTEGER ||
         option.type === 'STRING' ||
         option.type === Constants.ApplicationCommandOptionTypes.STRING ||
         option.type === 'NUMBER' ||
         option.type === Constants.ApplicationCommandOptionTypes.NUMBER;
}
```

This can be extremely confusing for developers, and hard to even come up with in the first place.

Highly used classes don't suffer from these issues, because they have type guards attached to them. Types/interfaces on the other hand are missing this.

I think the best solution to this would to include type guards for different variants of commonly used types. (I'm aware that `discord-api-types` does something like this already, however it's types aren't compatible with `discord.js`'s). These would simply be free functions that return booleans in the source code, and will be defined as a type guard in the `index.d.ts`.

**Additional context**
I'm willing to code this up, but wanted to verify if this is something the maintainers would want.

suneettipirneni:

Closes #6409

### DIFF
--- a/src/util/TypeGuards.js
+++ b/src/util/TypeGuards.js
@@ -1,0 +1,78 @@
+'use strict';
+
+/**
+ * Helpers for checking different variants of common types.
+ */
+class TypeGuards extends null {
+  /**
+   * Verifies if the given command data is chat input command or not.
+   * @param {ApplicationCommandData} commandData The command data to check.
+   * @returns {boolean} True if it conforms to chat input command data, false otherwise.
+   */
+  static isChatInputCommandData(commandData) {
+    return commandData.type === 'CHAT_INPUT' || commandData.type === 1;
+  }
+
+  /**
+   * Verifies if the given command data is a context menu command or not.
+   * @param {ApplicationCommandData} commandData The command data to check.
+   * @returns {boolean} True if it conforms to chat input command data, false otherwise.
+   */
+  static isContextMenuCommandData(commandData) {
+    return (
+      commandData.type === 'MESSAGE' || commandData.type === 'USER' || commandData.type === 2 || commandData.type === 3
+    );
+  }
+
+  /**
+   * Verifies if the given command option data supports choices or not.
+   * @param {ApplicationCommandOptionData} commandOptionData The command option data to check.
+   * @returns {boolean} True if the option supports choices, false otherwise.
+   */
+  static optionDataSupportsChoices(commandOptionData) {
+    return (
+      commandOptionData.type === 'BOOLEAN' ||
+      commandOptionData.type === 5 ||
+      commandOptionData.type === 'INTEGER' ||
+      commandOptionData.type === 4 ||
+      commandOptionData.type === 'STRING' ||
+      commandOptionData.type === 3 ||
+      commandOptionData.type === 'NUMBER' ||
+      commandOptionData.type === 10
+    );
+  }
+
+  /**
+   * Verifies if the given command option data supports options or not.
+   * @param {ApplicationCommandOptionData} commandOptionData The command option data to check.
+   * @returns {boolean} True if the option supports options, false otherwise.
+   */
+  static optionDataSupportsSubOptions(commandOptionData) {
+    return (
+      commandOptionData.type === 'SUB_COMMAND' ||
+      commandOptionData.type === 'SUB_COMMAND_GROUP' ||
+      commandOptionData.type === 1 ||
+      commandOptionData.type === 2
+    );
+  }
+
+  /**
+   * Verifies if the given message button options support URL's or not.
+   * @param {MessageButtonOptions} messageButtonOptions The message button options to check.
+   * @returns {boolean} True if the option supports URL's, false otherwise.
+   */
+  static isLinkButtonOptions(messageButtonOptions) {
+    return messageButtonOptions.style === 'LINK' || messageButtonOptions.style === 5;
+  }
+
+  /**
+   * Verifies if the given message button options support doesn't support URL's or not.
+   * @param {MessageButtonOptions} messageButtonOptions The message button options to check.
+   * @returns {boolean} True if the option doesn't support URL's, false otherwise.
+   */
+  static isInteractionButtonOptions(messageButtonOptions) {
+    return !TypeGuards.isLinkButtonOptions(messageButtonOptions);
+  }
+}
+
+module.exports = TypeGuards;
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -1938,6 +1938,26 @@ export class Formatters extends null {
   public static userMention: typeof userMention;
 }
 
+export class TypeGuards extends null {
+  private constructor();
+  public static isChatInputCommandData(
+    commandData: ApplicationCommandData,
+  ): commandData is ChatInputApplicationCommandData;
+  public static isContextMenuCommandData(
+    commandData: ApplicationCommandData,
+  ): commandData is MessageApplicationCommandData | UserApplicationCommandData;
+  public static optionDataSupportsChoices(
+    commandOptionData: ApplicationCommandOptionData,
+  ): commandOptionData is ApplicationCommandChoicesData;
+  public static optionDataSupportsSubOptions(
+    commandOptionData: ApplicationCommandOptionData,
+  ): commandOptionData is ApplicationCommandSubGroupData | ApplicationCommandSubGroupData;
+  public static isLinkButtonOptions(buttonOptions: MessageButtonOptions): buttonOptions is LinkButtonOptions;
+  public static isInteractionButtonOptions(
+    buttonOptions: MessageButtonOptions,
+  ): buttonOptions is InteractionButtonOptions;
+}
+
 export class VoiceChannel extends BaseGuildVoiceChannel {
   public readonly editable: boolean;
   public readonly speakable: boolean;
--- a/src/index.js
+++ b/src/index.js
@@ -31,6 +31,7 @@ module.exports = {
   ThreadMemberFlags: require('./util/ThreadMemberFlags'),
   UserFlags: require('./util/UserFlags'),
   Util: require('./util/Util'),
+  TypeGuards: require('./util/TypeGuards'),
   version: require('../package.json').version,
 
   // Managers


sprayx (approved):

src/util/TypeGuards.js
```diff
@@ -0,0 +1,78 @@
...
+    );
+  }
+
+  /**
+   * Verifies if the given command option data supports choices or not.
+   * @param {ApplicationCommandOptionData} commandOptionData The command option data to check.
+   * @returns {boolean} True if the option supports choices, false otherwise.
+   */
+  static optionDataSupportsChoices(commandOptionData) {
+    return (
```
NotSugden, line 33: possibly something like
```js
const type = typeof x.type === 'number' ? x.type : Enum[x.type];
return [Enum.X_TYPE, Enum.Y_TYPE].includes(type);
```
would be better (for each function)

NotSugden (changes requested):

src/util/TypeGuards.js
suneettipirneni: Thank you for this, I went ahead an implemented it.

iShibi (changes requested):

src/util/TypeGuards.js
```diff
@@ -0,0 +1,87 @@
...
+  /**
+   * Resolves a given type to an enum equivalent value, and
+   * checks if it's part of the given enum type.
+   * @private
+   * @param {string|number} type The type to resolve
+   * @param {any} object The enum to resolve to
+   * @param {any[]} fields The enum fields to check
+   * @returns {boolean} Whether the type is part of the enum
+   */
+  static isPartofEnum(type, object, fields) {
```
vladfrangu, line 18: Here and everywhere else

```suggestion
  static isPartOfEnum(type, object, fields) {
```

vladfrangu (changes requested):

src/util/TypeGuards.js
vladfrangu: You didn't resolve this?

suneettipirneni: oh my bad, I didn't see my push failed, one sec

vladfrangu (approved):
I don't know if this is needed, since TS can infer from the properties if we have done unions right, but LGTM

src/util/TypeGuards.js
```diff
@@ -0,0 +1,87 @@
...
+   * @private
+   * @param {string|number} type The type to resolve
+   * @param {any} object The enum to resolve to
+   * @param {any[]} fields The enum fields to check
+   * @returns {boolean} Whether the type is part of the enum
+   */
+  static isPartOfEnum(type, object, fields) {
+    const resolvedType = typeof type === 'number' ? type : object[type];
+    fields = fields.map(field => object[field]);
+    return fields.includes(resolvedType);
```
NotSugden, line 21:
```suggestion
    return fields.some(field => object[field] === resolvedType);
```
*(and remove the above line)*

suneettipirneni:
> I don't know if this is needed, since TS can infer from the properties if we have done unions right, but LGTM

You're correct, but I find doing having to do `command.type === 'CHAT_INPUT' && command.type === Constants.ApplicationCommandTypes.CHAT_INPUT`, is too verbose, and somewhat confusing. The typeguards in this context are more like aliases for these checks.

An alternative would be to split up the interfaces even more (add another interface for each variant, one for strings one for enums). But I dunno how I feel about that either.

suneettipirneni:
On reevaluation it seems that `command.type === 'TYPE'` does work as intended without checking enum values, so I'll remove the command checks but keep the option/choice checking as those are still useful.

typings/index.d.ts
```diff
@@ -1938,6 +1938,16 @@ export class Formatters extends null {
...
 }
 
+export class TypeGuards extends null {
+  private constructor();
+  public static optionDataSupportsChoices(
+    commandOptionData: ApplicationCommandOptionData,
+  ): commandOptionData is ApplicationCommandChoicesData;
+  public static optionDataSupportsSubOptions(
+    commandOptionData: ApplicationCommandOptionData,
+  ): commandOptionData is ApplicationCommandSubGroupData | ApplicationCommandSubGroupData;
```
ImRodry, line 1948: Why do you have the same type twice?

typings/index.d.ts
```diff
@@ -1938,6 +1938,16 @@ export class Formatters extends null {
   public static userMention: typeof userMention;
 }
 
+export class TypeGuards extends null {
```
ImRodry, line 1941: Missing declaration for isPartOfEnum

src/util/TypeGuards.js
```diff
@@ -0,0 +1,50 @@
...
+   * @private
+   * @param {string|number} type The type to resolve
+   * @param {any} object The enum to resolve to
+   * @param {string[]} fields The enum fields to check
+   * @returns {boolean} Whether the type is part of the enum
+   */
+  static isPartOfEnum(type, object, fields) {
+    const resolvedType = typeof type === 'number' ? type : object[type];
+    return fields.some(field => object[field] === resolvedType);
+  }
```
ImRodry, line 21: Should probably have some more checks to make sure all parameters are of the desired types and throw errors if not

src/util/TypeGuards.js
```diff
@@ -0,0 +1,50 @@
...
+/**
+ * Helpers for checking different variants of common types.
+ */
+class TypeGuards extends null {
+  /**
+   * Resolves a given type to an enum equivalent value, and
+   * checks if it's part of the given enum type.
+   * @private
+   * @param {string|number} type The type to resolve
+   * @param {any} object The enum to resolve to
```
ImRodry, line 14:
```suggestion
   * @param {Object} object The enum to resolve to
```

src/util/TypeGuards.js
suneettipirneni: This is private, it's only meant for use within the `TypeGuards` class.

ImRodry: So? This is used by the other methods which can still receive invalid types so it’s best to have checks for them here

suneettipirneni: JS already throws an error on its own if you mess a parameter up. I don't see the value of adding another custom one. This isn't typescript, it's javascript where errors like that are normal.

ImRodry: But it’s a less insightful error and the library throws TypeErrors all the time so I don’t see why this shouldn’t either

suneettipirneni: Because as I said earlier it's a private function, any error that is thrown by that function can only be caused by the d.js devs themselves, not the public.

ImRodry: That is not true because you’re passing the parameters sent by the user to the function so it can error out

suneettipirneni: Then I can throw errors in the public functions themselves? Throwing them in the private function makes them less insightful? By the time it reaches the private function we don't know what specific type was trying to be checked.

ImRodry: Sure, I just think they'd be the same so they could all be put in the same place but you do you

src/util/TypeGuards.js
```diff
@@ -0,0 +1,63 @@
...
+    return TypeGuards.isPartOfEnum(commandOptionData.type, ApplicationCommandOptionTypes, [
+      'SUB_COMMAND',
+      'SUB_COMMAND_GROUP',
+    ]);
+  }
+
+  /**
+   * @private
+   * Throws a type error if the object has no `type field`.
+   * @param {Object} object The object to type check.
```
SpaceEEC, line 54: Same here:
```suggestion
   * Throws a type error if the object has no `type field`.
   * @param {Object} object The object to type check.
   * @private
```

src/util/TypeGuards.js
```diff
@@ -0,0 +1,63 @@
...
+ */
+class TypeGuards extends null {
+  /**
+   * Resolves a given type to an enum equivalent value, and
+   * checks if it's part of the given enum type.
+   * @private
+   * @param {string|number} type The type to resolve
+   * @param {Object} object The enum to resolve to
+   * @param {string[]} fields The enum fields to check
+   * @returns {boolean} Whether the type is part of the enum
```
SpaceEEC, line 16: We aren't super consistent with this, but I think putting private below returns is how we do it in most cases:
```suggestion
   * @param {string|number} type The type to resolve
   * @param {Object} object The enum to resolve to
   * @param {string[]} fields The enum fields to check
   * @returns {boolean} Whether the type is part of the enum
   * @private
```

src/util/TypeGuards.js
```diff
@@ -0,0 +1,63 @@
...
+  }
+
+  /**
+   * @private
+   * Throws a type error if the object has no `type field`.
+   * @param {Object} object The object to type check.
+   */
+  static validateType(object) {
+    if (!('type' in object)) {
+      throw new TypeError('Please use a valid type for the type guard.');
```
SpaceEEC, line 58: This should throw a discord.js type error:
https://github.com/discordjs/discord.js/blob/c232baa7154a4f07904ad5b0b8557ebb452bfb66/src/errors/DJSError.js#L59

You can probably use `INVALID_TYPE` here:
https://github.com/discordjs/discord.js/blob/c232baa7154a4f07904ad5b0b8557ebb452bfb66/src/errors/Messages.js#L97

SpaceEEC (changes requested):

typings/index.d.ts
```diff
@@ -1938,6 +1938,18 @@ export class Formatters extends null {
...
+  private constructor();
+  private static isPartOfEnum(type: string | number, object: object, fields: string[]): boolean;
```
kyranet, line 1943: `private static` must be should after `public static`, same for the JS methods.

kyranet (changes requested):

kyranet (approved):

PR #6414, opened 2021-08-13
feat(ApplicationCommand): add #equals

ckohen:

Adds `ApplicationCommand#equals` since checking the equivalence of application commands is a non-trivial task. Especially when handling the instantiated class and raw data.

Oh, and also because @IanMitchell asked for it

### DIFF
--- a/src/structures/ApplicationCommand.js
+++ b/src/structures/ApplicationCommand.js
@@ -157,7 +157,7 @@ class ApplicationCommand extends Base {
   }
 
   /**
-   * Whether this copmmand equals another command. It compares all properties, so for most operations
+   * Whether this command equals another command. It compares all properties, so for most operations
    * it is advisable to just compare `command.id === command2.id` as it is much faster and is often
    * what most users need.
    * @param {ApplicationCommand|ApplicationCommandData|APIApplicationCommand} command The command to compare with
--- a/src/structures/ApplicationCommand.js
+++ b/src/structures/ApplicationCommand.js
@@ -157,6 +157,76 @@ class ApplicationCommand extends Base {
   }
 
   /**
+   * Whether this copmmand equals another command. It compares all properties, so for most operations
+   * it is advisable to just compare `command.id === command2.id` as it is much faster and is often
+   * what most users need.
+   * @param {ApplicationCommand|ApplicationCommandData|APIApplicationCommand} command The command to compare with
+   * @returns {boolean}
+   */
+  equals(command) {
+    // If given an id, early return
+    if (command.id && this.id !== command.id) return false;
+
+    // Check top level parameters
+    const commandType = typeof command.type === 'string' ? command.type : ApplicationCommandTypes[command.type];
+    if (
+      command.name !== this.name ||
+      ('description' in command && command.description !== this.description) ||
+      (commandType && commandType !== this.type) ||
+      // Future proof for options being nullable
+      // TODO: remove ?? 0 on each when nullable
+      (command.options?.length ?? 0) !== (this.options?.length ?? 0) ||
+      (command.defaultPermission ?? command.default_permission ?? true) !== this.defaultPermission
+    ) {
+      return false;
+    }
+
+    // Check a singular option
+    function optionEquals(existing, option) {
+      const optionType = typeof option.type === 'string' ? option.type : ApplicationCommandOptionTypes[option.type];
+      // We don't check name here because name is guaranteed to match via the find operation in optionsEqual
+      if (
+        optionType !== existing.type ||
+        option.description !== existing.description ||
+        (option.required ??
+          (optionType === 'SUB_COMMAND' || optionType === 'SUB_COMMAND_GROUP' ? undefined : false) !==
+            existing.required) ||
+        option.choices?.length !== existing.choices?.length ||
+        option.options?.length !== existing.options?.length
+      ) {
+        return false;
+      }
+
+      if (existing.choices) {
+        for (const choice of existing.choices) {
+          const foundChoice = option.choices.find(c => c.name === choice.name);
+          if (!foundChoice) return false;
+          if (foundChoice.value !== choice.value) return false;
+        }
+      }
+
+      if (existing.options) {
+        return optionsEqual(existing.options, option.options);
+      }
+      return true;
+    }
+
+    // Check an array of options
+    function optionsEqual(existing, options) {
+      for (const option of existing) {
+        const foundOption = options.find(o => o.name === option.name);
+        if (!foundOption) return false;
+        if (!optionEquals(option, foundOption)) return false;
+      }
+      return true;
+    }
+    if (command.options) {
+      return optionsEqual(this.options, command.options);
+    }
+    return true;
+  }
+
+  /**
    * An option for an application command or subcommand.
    * @typedef {Object} ApplicationCommandOption
    * @property {ApplicationCommandOptionType} type The type of the option
--- a/src/structures/ApplicationCommand.js
+++ b/src/structures/ApplicationCommand.js
@@ -219,7 +219,7 @@ class ApplicationCommand extends Base {
       option.name !== existing.name ||
       optionType !== existing.type ||
       option.description !== existing.description ||
-      (option.required ?? (optionType === 'SUB_COMMAND' || optionType === 'SUB_COMMAND_GROUP' ? undefined : false)) !==
+      (option.required ?? (['SUB_COMMAND', 'SUB_COMMAND_GROUP'].includes(optionType) ? undefined : false)) !==
         existing.required ||
       option.choices?.length !== existing.choices?.length ||
       option.options?.length !== existing.options?.length
--- a/src/structures/ApplicationCommand.js
+++ b/src/structures/ApplicationCommand.js
@@ -219,9 +219,8 @@ class ApplicationCommand extends Base {
       option.name !== existing.name ||
       optionType !== existing.type ||
       option.description !== existing.description ||
-      (option.required ??
-        (optionType === 'SUB_COMMAND' || optionType === 'SUB_COMMAND_GROUP' ? undefined : false) !==
-          existing.required) ||
+      (option.required ?? (optionType === 'SUB_COMMAND' || optionType === 'SUB_COMMAND_GROUP' ? undefined : false)) !==
+        existing.required ||
       option.choices?.length !== existing.choices?.length ||
       option.options?.length !== existing.options?.length
     ) {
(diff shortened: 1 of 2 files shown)


src/structures/ApplicationCommand.js
```diff
@@ -156,6 +156,76 @@ class ApplicationCommand extends Base {
     return this.manager.delete(this, this.guildId);
   }
 
+  /**
+   * Whether this copmmand equals another command. It compares all properties, so for most operations
```
JKLorenzo, line 160:
```suggestion
   * Whether this command equals another command. It compares all properties, so for most operations
```

typings/index.d.ts
```diff
@@ -219,6 +219,7 @@ export class ApplicationCommand<PermissionsFetchType = {}> extends Base {
   public type: ApplicationCommandType;
   public delete(): Promise<ApplicationCommand<PermissionsFetchType>>;
   public edit(data: ApplicationCommandData): Promise<ApplicationCommand<PermissionsFetchType>>;
+  public isEqual(command: ApplicationCommand | ApplicationCommandData | RawApplicationCommandData): boolean;
```
JKLorenzo, line 222:
```suggestion
  public equals(command: ApplicationCommand | ApplicationCommandData | RawApplicationCommandData): boolean;
```

JKLorenzo (changes requested):

src/structures/ApplicationCommand.js
```diff
@@ -156,6 +156,76 @@ class ApplicationCommand extends Base {
...
 
+  /**
+   * Whether this command equals another command. It compares all properties, so for most operations
+   * it is advisable to just compare `command.id === command2.id` as it is much faster and is often
+   * what most users need.
+   * @param {ApplicationCommand|ApplicationCommandData|APIApplicationCommand} command The command to compare with
+   * @returns {boolean}
+   */
+  equals(command) {
+    // If given an id, early return
```
vladfrangu, line 167: Need to describe why it could early return

```suggestion
    // If given an id, check if the id matches
```

vladfrangu (changes requested):

src/structures/ApplicationCommand.js
```diff
@@ -156,6 +156,76 @@ class ApplicationCommand extends Base {
...
+        option.options?.length !== existing.options?.length
+      ) {
+        return false;
+      }
+
+      if (existing.choices) {
+        for (const choice of existing.choices) {
+          const foundChoice = option.choices.find(c => c.name === choice.name);
+          if (!foundChoice) return false;
+          if (foundChoice.value !== choice.value) return false;
```
NotSugden, line 204:
```suggestion
          if (!foundChoice || foundChoice.value !== choice.value) return false;
```

src/structures/ApplicationCommand.js
```diff
@@ -156,6 +156,76 @@ class ApplicationCommand extends Base {
...
+      // Future proof for options being nullable
+      // TODO: remove ?? 0 on each when nullable
+      (command.options?.length ?? 0) !== (this.options?.length ?? 0) ||
+      (command.defaultPermission ?? command.default_permission ?? true) !== this.defaultPermission
+    ) {
+      return false;
+    }
+
+    // Check a singular option
+    function optionEquals(existing, option) {
```
NotSugden, line 185:
```suggestion
    const optionEquals = (existing, option) => {
```

src/structures/ApplicationCommand.js
```diff
@@ -156,6 +156,76 @@ class ApplicationCommand extends Base {
...
+      }
+
+      if (existing.options) {
+        return optionsEqual(existing.options, option.options);
+      }
+      return true;
+    }
+
+    // Check an array of options
+    function optionsEqual(existing, options) {
```
NotSugden, line 215:
```suggestion
    const optionsEqual = (existing, options) => {
```
and move this up (maybe even make them private functions on the class?)

ckohen: I originally considered making them private functions but didn't want to because the types are gonna be a mess...however, I think  I will do this though and they can be static functions so even better.

src/structures/ApplicationCommand.js
```diff
@@ -156,6 +156,91 @@ class ApplicationCommand extends Base {
...
+
+    // Check top level parameters
+    const commandType = typeof command.type === 'string' ? command.type : ApplicationCommandTypes[command.type];
+    if (
+      command.name !== this.name ||
+      ('description' in command && command.description !== this.description) ||
+      (commandType && commandType !== this.type) ||
+      // Future proof for options being nullable
+      // TODO: remove ?? 0 on each when nullable
+      (command.options?.length ?? 0) !== (this.options?.length ?? 0) ||
```
ImRodry, line 178: You don't actually need the `?? 0` because you'd be comparing undefined to undefined if neither one of the commands have options, which is still valid and produces the same output

ckohen: `this.options` always exists currently, the `?.` on the RHS is misleading. I could remove the one on the RHS only but this is accounting for the (yes unlikely) case that someone changed that property in runtime

ImRodry: They can also provide an object that they construct in which case it won't have options so it's good to keep both ?. there, just saying that because of your TODO comment, since you can easily remove those

vladfrangu (approved):

SpaceEEC (approved):

src/structures/ApplicationCommand.js
```diff
@@ -156,6 +156,90 @@ class ApplicationCommand extends Base {
...
+   * In most cases it is better to compare using {@link ApplicationCommand#equals}
+   * @param {ApplicationCommandOptionData[]} existing The options on the existing command,
+   * should be {@link ApplicationCommand#options}
+   * @param {ApplicationCommandOptionData[]|APIApplicationCommandOption[]} options The options to compare against
+   * @returns {boolean}
+   */
+  static optionsEqual(existing, options) {
+    if (existing.length !== options.length) return false;
+    for (const option of existing) {
+      const foundOption = options.find(o => o.name === option.name);
```
NotSugden, line 201: would the indexes be the same for both arrays?

could do
```js
return existing.every((option, idx) => this._optionEquals(option, options[idx]));
```
in any case an `every` call would probably be what i'd use here

src/structures/ApplicationCommand.js
```diff
@@ -156,6 +156,90 @@ class ApplicationCommand extends Base {
...
+      (option.required ?? (['SUB_COMMAND', 'SUB_COMMAND_GROUP'].includes(optionType) ? undefined : false)) !==
+        existing.required ||
+      option.choices?.length !== existing.choices?.length ||
+      option.options?.length !== existing.options?.length
+    ) {
+      return false;
+    }
+
+    if (existing.choices) {
+      for (const choice of existing.choices) {
```
NotSugden, line 231: the same here as the above

src/structures/ApplicationCommand.js
ImRodry: The order in which you send options does actually matter so two commands should not be considered equal if they have the same options in a different order

ckohen: IIRC, discord sometimes reorders things, and reordered options / choices do not actually get taken into account.
I used for...of for two reasons
1. return directly from the loop (its one less op too!)
2. make it clear that the loop stops as soon as it hits a non-equal option
Bonus 3. If for some reason any of the functions need to be async in the future, we don't need to rewrite the loop

ImRodry: Discord does take order into account and deploys options in the exact order you provide them, so this change is important

ckohen: Hmm, it does take it into account, but not everything is deployed in the order provided (client side that is, API still has the correct order) E.g. subcommand groups are **always** displayed before subcommands on the same level. 

Its hard to say which behavior would be more confusing given that. I imagine most people store the command definition somewhere in their code, but if you were to reconstruct a command from the way the client displayed it (again, you could just fetch it to be sure its identical), it could be different but actually appear identical.

Not to mention that the way subcommands / options are displayed looks to be being changed soon.

ImRodry: If the API still has the same order then you can probably compare the indexes just fine, since it would actually change something

monbrey: Why restrict this to relying on the index order being the same between both API and user definitions if we have a method that is more robust and reliable?

ImRodry: The idea is to check index order and then use the method tho?

NotSugden: > If for some reason any of the functions need to be async in the future, we don't need to rewrite the loop

i don't see a function like this becoming async any time soon 😄 

on another note, the inital idea behind suggesting using the index was simply just not to iterate through the second options for every single option

ImRodry: But you can’t just check for the index which I think is what you’re suggesting right?

ckohen: > i don't see a function like this becoming async any time soon 😄

Absolutely not heh

Okay, so here's my thoughts
1. Provide an optional param, enforceOrdering that will enforce strict equality in array orders
2. To alleviate concern with iterations, iterate through the second set of options once, storing them in a map by name to be able to perform the very quick get based on the name.

ImRodry: Why would you need a parameter for that? I don’t see a reason to leave that disabled tbf
Point 2 is fine by me tho

iCrawl (approved):

kyranet (approved):
We really need unit tests for this, but looks good to me.

ImRodry:
I strongly disagree with the need for an option here and specially it being set to false by default. Option order is something that will change how Discord handles your commands so if you really want that option it should default to true. The best in my opinion would be to default to true and remove that option.
Also why do you still have TODO comments?

ckohen:
I have TODO comments because they're TODO for v14, not for now.

As for your disagreement. No, order doesn't change how discord handles your commands, it changes how it is stored on the API. As I've already said, the client does not strictly respect this order and the info blocks I added for those parameters clearly indicates the thought process for that.

ImRodry:
> I have TODO comments because they're TODO for v14, not for now.
> 
> As for your disagreement. No, order doesn't change how discord handles your commands, it changes how it is stored on the API. As I've already said, the client does not strictly respect this order and the info blocks I added for those parameters clearly indicates the thought process for that.

The client may not strictly respect that order, but it does to some extent. So if you change the order of an option or a subcommand even, the client will display it differently, which is why I think it should at least default to true. 
As for the TODO comments, why are they for v14?

PR #6419, opened 2021-08-13
feat(GuildChannel): Add `publiclyViewable` getter

suneettipirneni:

Adds a getter which indicates if the channel can be viewed by the public or not.

### DIFF
--- a/src/structures/GuildChannel.js
+++ b/src/structures/GuildChannel.js
@@ -157,6 +157,15 @@ class GuildChannel extends Channel {
   }
 
   /**
+   * Checks if the channel is viewable by the public.
+   * @returns {boolean} true if public, false otherwise.
+   */
+  get publiclyViewable() {
+    const { everyone } = this.guild.roles;
+    return this.permissionsFor(everyone)?.has(['VIEW_CHANNEL']) ?? false;
+  }
+
+  /**
    * Gets the overall set of permissions for a member or role in this channel, taking into account channel overwrites.
    * @param {GuildMemberResolvable|RoleResolvable} memberOrRole The member or role to obtain the overall permissions for
    * @returns {?Readonly<Permissions>}
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -808,6 +808,7 @@ export class GuildChannel extends Channel {
   public permissionOverwrites: PermissionOverwriteManager;
   public readonly permissionsLocked: boolean | null;
   public readonly position: number;
+  public readonly publiclyViewable: boolean;
   public rawPosition: number;
   public type: Exclude<keyof typeof ChannelTypes, 'DM' | 'GROUP_DM' | 'UNKNOWN'>;
   public readonly viewable: boolean;


src/structures/GuildChannel.js
```diff
@@ -156,6 +156,15 @@ class GuildChannel extends Channel {
     return [...sorted.values()].indexOf(sorted.get(this.id));
   }
 
+  /**
+   * Checks if the channel is viewable by the public.
+   * @type {boolean} true if public, false otherwise.
+   */
+  get publiclyViewable() {
+    const { everyone } = this.guild.roles;
+    return this.permissionsFor(everyone)?.has(['VIEW_CHANNEL']) ?? false;
```
NotSugden, line 165:
```suggestion
    return this.permissionsFor(everyone)?.has(Permissions.FLAGS.VIEW_CHANNEL) ?? false;
```

advaith1:
imo a `private` getter would be better, and closer to Discord's UI

suneettipirneni:
> imo a `private` getter would be better, and closer to Discord's UI

Good point, I'll refactor the method to that.

src/structures/GuildChannel.js
```diff
@@ -156,6 +156,15 @@ class GuildChannel extends Channel {
     return [...sorted.values()].indexOf(sorted.get(this.id));
   }
 
+  /**
+   * Checks if the channel is not viewable by users with the @everyone role.
```
advaith1, line 160:
```suggestion
   * Checks if the channel is not viewable by users with only the @everyone role.
```

src/structures/GuildChannel.js
```diff
@@ -156,6 +156,15 @@ class GuildChannel extends Channel {
     return [...sorted.values()].indexOf(sorted.get(this.id));
   }
 
+  /**
+   * Checks if the channel is not viewable by users with only the @everyone role.
+   * @type {boolean} true if private, false otherwise.
+   */
+  get private() {
+    const { everyone } = this.guild.roles;
+    return !(this.permissionsFor(everyone)?.has(Permissions.FLAGS.VIEW_CHANNEL) ?? false);
```
kyranet, line 165: Unless roles are being swept, which we do not support, `permissionsFor` will never return null.
```suggestion
    return this.permissionsFor(everyone).has(Permissions.FLAGS.VIEW_CHANNEL);
```

kyranet (changes requested):

src/structures/GuildChannel.js
almeidx: @kyranet probably wanted to suggest this instead:
```suggestion
    return !this.permissionsFor(everyone).has(Permissions.FLAGS.VIEW_CHANNEL);
```

Otherwise, the method would return `true` if the channel was public

kyranet: Fixed, thanks @almeidx

src/structures/GuildChannel.js
```diff
@@ -156,6 +156,15 @@ class GuildChannel extends Channel {
     return [...sorted.values()].indexOf(sorted.get(this.id));
   }
 
+  /**
+   * Checks if the channel is not viewable by users with only the @everyone role.
+   * @type {boolean} true if private, false otherwise.
+   */
+  get private() {
+    const { everyone } = this.guild.roles;
+    return !this.permissionsFor(everyone).has(Permissions.FLAGS.VIEW_CHANNEL);
```
NotSugden, line 165: may aswell just use `this.guild.roles.everyone` here instead of making a variable for it

src/structures/GuildChannel.js
```diff
@@ -156,6 +156,14 @@ class GuildChannel extends Channel {
     return [...sorted.values()].indexOf(sorted.get(this.id));
   }
 
+  /**
+   * Checks if the channel is not viewable by users with only the @everyone role.
+   * @type {boolean} true if private, false otherwise.
+   */
+  get private() {
+    return !this.permissionsFor(this.guild.roles.everyone).has(Permissions.FLAGS.VIEW_CHANNEL);
```
advaith1, line 164:
```suggestion
    return !this.permissionsFor(this.guild.id).has(Permissions.FLAGS.VIEW_CHANNEL);
```

src/structures/GuildChannel.js
```diff
@@ -156,6 +156,14 @@ class GuildChannel extends Channel {
     return [...sorted.values()].indexOf(sorted.get(this.id));
   }
 
+  /**
+   * Checks if the channel is not viewable by users with only the @everyone role.
+   * @type {boolean} true if private, false otherwise.
```
SpaceEEC, line 161: This seems quite redundant to me.
```suggestion
   * @type {boolean}
```

src/structures/GuildChannel.js
kyranet: I'd argue that this suggestion's code feels hacky, and does the same steps.

- If `Snowflake` -> Get `?Role` -> `permissionsFor(Role?)` 
- Get `Role` -> If `Role` -> `permissionsFor(Role)`

Only difference is that `permissionsFor(Snowflake)`'s return type is nullable (because the case of invalid role IDs) whereas `permissionsFor(Role)` is not, so in the event when we rewrite to TypeScript, this would be one nullish assertion operator less.

Also, in the event Discord ever changes the `@everyone`'s role ID to be something else, `roles.everyone` would be updated and this wouldn't need further changes, but `this.guildId` would.

kyranet (approved):

PR #6425, opened 2021-08-14
feat(MessageTypes): add CONTEXT_MENU_COMMAND (23)

advaith1:
- https://github.com/discord/discord-api-docs/pull/3651

this might not actually ship

### DIFF
--- a/src/util/Constants.js
+++ b/src/util/Constants.js
@@ -374,6 +374,7 @@ exports.InviteScopes = [
  * * APPLICATION_COMMAND
  * * THREAD_STARTER_MESSAGE
  * * GUILD_INVITE_REMINDER
+ * * CONTEXT_MENU_COMMAND
  * @typedef {string} MessageType
  * @see {@link https://discord.com/developers/docs/resources/channel#message-object-message-types}
  */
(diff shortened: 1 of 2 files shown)


kyranet (approved):

iCrawl (approved):

SpaceEEC (approved):

src/util/Constants.js
```diff
@@ -374,6 +374,7 @@ exports.InviteScopes = [
  * * APPLICATION_COMMAND
```
ImRodry, line 374: Shouldn't this be renamed to SLASH_COMMAND like it was on the upstream PR? Now that context menus are a thing, having an APPLICATION_COMMAND message type can be quite vague

ckohen: That would be a breaking change, it'll have to wait for v14

ImRodry: pretty sure there was a PR before v13 to change that, why wasn't that merged 🤔

vladfrangu (approved):

src/util/Constants.js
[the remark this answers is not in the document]
iCrawl: Separate PR

PR #6429, opened 2021-08-14
fix(BaseClient): Remove selfbot ability

xHyroM:

Removing this will prevent people from raiding easier. Yes they can do fork, but I don't think a well known library like this should support it.

How can you make a self bot?
`new Client({_tokenType:""})`

### DIFF
--- a/src/client/BaseClient.js
+++ b/src/client/BaseClient.js
@@ -24,7 +24,7 @@ class BaseClient extends EventEmitter {
      * @type {RESTManager}
      * @private
      */
-    this.rest = new RESTManager(this, "Bot");
+    this.rest = new RESTManager(this, 'Bot');
   }
 
   /**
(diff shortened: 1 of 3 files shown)


vladfrangu (approved):

xHyroM:
Alternatively, I can make it so that there will be support for `bot` type and `bearer` but I don't think it's necessary to do that and just do it this way.

SpaceEEC (approved):
This is a, now not needed anymore, remnant of #1883, which I think was RPC related.

kyranet (approved):

vaporox:
So why not remove that parameter now that it's hard coded?

xHyroM:
> 
> 
> So why not remove that parameter now that it's hard coded?

Good point. I've already deleted it.

iCrawl (approved):

PR #6444, opened 2021-08-15
feat(`BaseGuildTextChannel`): Add `BaseGuildTextChannel#sendable` getter.

WiseDevHelper:

This PR adds a getter which indicates if the client can send messages in it.

### DIFF
--- a/src/structures/BaseGuildTextChannel.js
+++ b/src/structures/BaseGuildTextChannel.js
@@ -88,6 +88,18 @@ class BaseGuildTextChannel extends GuildChannel {
   }
 
   /**
+   * Whether the client user can send messages in this channel.
+   * @type {boolean}
+   * @readonly
+   */
+  get sendable() {
+    if (this.client.user.id === this.guild.ownerId) return true;
+    const permissions = this.permissionsFor(this.client.user);
+    if (!permissions) return false;
+    return permissions.has(Permissions.FLAGS.SEND_MESSAGES, false);
+  }
+
+  /**
    * Sets the default auto archive duration for all newly created threads in this channel.
    * @param {ThreadAutoArchiveDuration} defaultAutoArchiveDuration The new default auto archive duration
    * @param {string} [reason] Reason for changing the channel's default auto archive duration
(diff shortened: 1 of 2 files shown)


monbrey:
Not a fan of this or #6419. Permission helpers don't need to bloat the library imo

ImRodry:
> Not a fan of this or #6419. Permission helpers don't need to bloat the library imo

the library is literally full of permission helpers @monbrey, some classes have more than 5, some even close to 10

wasdennnoch:
If this is actually implemented, this also has to check for `VIEW_CHANNEL` permissions. Without those, `SEND_MESSAGES` is implicitly denied.

monbrey:
> > Not a fan of this or #6419. Permission helpers don't need to bloat the library imo
> 
> the library is literally full of permission helpers @monbrey, some classes have more than 5, some even close to 10

And we have on multiple occassions suggested we don't want to go further in that direction and increase the problem. Unless that's changed and I haven't realised.

ImRodry:
> > > Not a fan of this or #6419. Permission helpers don't need to bloat the library imo
> > 
> > 
> > the library is literally full of permission helpers @monbrey, some classes have more than 5, some even close to 10
> 
> And we have on multiple occassions suggested we don't want to go further in that direction and increase the problem. Unless that's changed and I haven't realised.

I haven’t seen any signs of this slowing down because they keep getting introduced as new classes come. However I’m still not a fan of this PR, but I really wanna see the other one merged because it does represent a feature within Discord and it’s something I’ve tried to suggest before

iCrawl:
> I haven’t seen any signs of this slowing down because they keep getting introduced as new classes come.

I don't remember merging any significant permission helpers in a long time.

WiseDevHelper:
> Not a fan of this or #6419. Permission helpers don't need to bloat the library imo

Added this to maintain consistency with `ThreadChannel`. Alternatively, the getter from that class could be removed.


> If this is actually implemented, this also has to check for `VIEW_CHANNEL` permissions. Without those, `SEND_MESSAGES` is implicitly denied.

Fixed 👍

PR #6447, opened 2021-08-15
fix(GuildChannel): Fix manageable method for voice-channels

### LINKED ISSUE #6446: Missing Permissions on channel update after all needed tests

I catch a **Missing Permissions** error on a channel edit when all needed checks are done.
 - `<GuildChannel>.manageable` is tested
 - `<Role>.editable` is not tested but we know that we have the `MANAGE_ROLES` permission because we create a role. The target role has just be created so under the position of the bot.
 
Here is a sentry report error: https://sentry.draftbot.fr/share/issue/1309451c008c4269bd2e4a3a85294a31/

```js
const role = await member.guild.roles.create({
  name: 'Not verified',
  reason: 'Role required for members who have not completed the captcha.',
});

for (const channel of member.guild.channels.cache.array()) {
    if(!channel.manageable) continue;
    await channel.updateOverwrite(role, { VIEW_CHANNEL: false }, {
        type: 0,
        reason: 'Update of the channel for the implementation of the captcha.',
    });
}
```

**Further details:**

- discord.js version: **v13.0.0-dev** [this commit](https://github.com/discordjs/discord.js/commit/e980948de55e91e59c9e3293ac76bc645a058a53)
- Node.js version: **v14**
- Operating system: **Debian GNU/Linux 10 (buster)**
- Priority this issue should have – please be realistic and elaborate if possible: **Not priority**

**Relevant client options:**

- partials: MESSAGE, REACTION, GUILD_MEMBER
- gateway intents: GUILDS, GUILD_MEMBERS, GUILD_BANS, GUILD_EMOJIS, GUILD_VOICE_STATES, GUILD_MESSAGES, GUILD_MESSAGE_REACTIONS, DIRECT_MESSAGES, DIRECT_MESSAGE_REACTIONS
- other: none

- Its a production error have only tested the issue on v13 version; Commit hash: `e980948de55e91e59c9e3293ac76bc645a058a53`

DraftProducts:

Resolves #6446

`<GuildChannel>.manageable` is not efficient with voice channels because `<GuildChannel>.viewable` is not checked on voice channels.

```js
if (VoiceBasedChannelTypes.includes(this.type)) { 
   // connect permission check
 } else if (!this.viewable) { 
   return false;
 }
```

### DIFF
--- a/src/structures/GuildChannel.js
+++ b/src/structures/GuildChannel.js
@@ -514,7 +514,7 @@ class GuildChannel extends Channel {
     const permissions = this.permissionsFor(this.client.user);
     if (!permissions) return false;
     const bitfield = VoiceBasedChannelTypes.includes(this.type)
-      ? Permissions.FLAGS.VIEW_CHANNEL | Permissions.FLAGS.MANAGE_CHANNELS | Permissions.FLAGS.CONNECT
+      ? Permissions.FLAGS.MANAGE_CHANNELS | Permissions.FLAGS.CONNECT
       : Permissions.FLAGS.VIEW_CHANNEL | Permissions.FLAGS.MANAGE_CHANNELS;
     return permissions.has(bitfield, false);
   }
--- a/src/structures/GuildChannel.js
+++ b/src/structures/GuildChannel.js
@@ -514,10 +514,11 @@ class GuildChannel extends Channel {
     if (!this.viewable) {
       return false;
     }
-    if (VoiceBasedChannelTypes.includes(this.type)) {
-      if (!this.permissionsFor(this.client.user).has(Permissions.FLAGS.CONNECT, false)) {
-        return false;
-      }
+    if (
+      VoiceBasedChannelTypes.includes(this.type) &&
+      !this.permissionsFor(this.client.user).has(Permissions.FLAGS.CONNECT, false)
+    ) {
+      return false;
     }
     return this.permissionsFor(this.client.user).has(Permissions.FLAGS.MANAGE_CHANNELS, false);
   }
--- a/src/structures/GuildChannel.js
+++ b/src/structures/GuildChannel.js
@@ -511,16 +511,13 @@ class GuildChannel extends Channel {
    */
   get manageable() {
     if (this.client.user.id === this.guild.ownerId) return true;
-    if (!this.viewable) {
+    if (!this.viewable) return false;
+    const permissions = this.permissionsFor(this.client.user);
+    if (!permissions) return false;
+    if (VoiceBasedChannelTypes.includes(this.type) && !permissions.has(Permissions.FLAGS.CONNECT, false)) {
       return false;
     }
-    if (
-      VoiceBasedChannelTypes.includes(this.type) &&
-      !this.permissionsFor(this.client.user).has(Permissions.FLAGS.CONNECT, false)
-    ) {
-      return false;
-    }
-    return this.permissionsFor(this.client.user).has(Permissions.FLAGS.MANAGE_CHANNELS, false);
+    return permissions.has(Permissions.FLAGS.MANAGE_CHANNELS, false);
   }
 
   /**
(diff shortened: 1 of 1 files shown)


src/structures/GuildChannel.js
```diff
@@ -511,11 +511,13 @@ class GuildChannel extends Channel {
...
-      if (!this.permissionsFor(this.client.user).has(Permissions.FLAGS.CONNECT, false)) {
-        return false;
-      }
-    } else if (!this.viewable) {
+    if (!this.viewable) {
+      return false;
+    }
+    if (
+      VoiceBasedChannelTypes.includes(this.type) &&
+      !this.permissionsFor(this.client.user).has(Permissions.FLAGS.CONNECT, false)
```
NotSugden, line 519: put `this.permissionsFor(...)` into a variable so its not calculated twice unnessercerily

src/structures/GuildChannel.js
```diff
@@ -511,11 +511,13 @@ class GuildChannel extends Channel {
...
     if (this.client.user.id === this.guild.ownerId) return true;
-    if (VoiceBasedChannelTypes.includes(this.type)) {
-      if (!this.permissionsFor(this.client.user).has(Permissions.FLAGS.CONNECT, false)) {
-        return false;
-      }
-    } else if (!this.viewable) {
+    if (!this.viewable) {
+      return false;
+    }
+    if (
```
NotSugden, line 517: this can be combined with the above if statement

ckohen:
This is incorrect, you can manage channels without permission to view them, the error in the original issue is caused because the bot did not have permission to view the channel, and thus could not change other roles permission on it.

DraftProducts:
So you can manage voice-channel without view permission but for text channels, it is needed?

ckohen:
Correct, I should have specified voice, but the permission bit to manage the channel is actually `CONNECT` on voice channels, regardless of the `VIEW_CHANNEL` bit. The error actually explains it better, you get Missing Access when you can't change overwrites on the channel, and Missing Permissions when you can't change one of the permissions you are trying to change

src/structures/GuildChannel.js
```diff
@@ -511,14 +511,13 @@ class GuildChannel extends Channel {
...
     if (this.client.user.id === this.guild.ownerId) return true;
-    if (VoiceBasedChannelTypes.includes(this.type)) {
-      if (!this.permissionsFor(this.client.user).has(Permissions.FLAGS.CONNECT, false)) {
-        return false;
-      }
-    } else if (!this.viewable) {
+    if (!this.viewable) return false;
+    const permissions = this.permissionsFor(this.client.user);
+    if (!permissions) return false;
+    if (VoiceBasedChannelTypes.includes(this.type) && !permissions.has(Permissions.FLAGS.CONNECT, false)) {
```
NotSugden, line 517:
```suggestion
    if (!permissions || (VoiceBasedChannelTypes.includes(this.type) && !permissions.has(Permissions.FLAGS.CONNECT, false))) {
```

kyranet (commented):
Can't we just do this?

```javascript
// Owner always can manage:
if (this.client.user.id === this.guild.ownerId) return true;

// Get the permissions, if we couldn't resolve to a member, return false:
const permissions = this.permissionsFor(this.client.user);
if (!permissions) return false;

// Get the bitfield and see if we fulfill it:
const bitfield = VoiceBasedChannelTypes.includes(this.type)
  ? Permissions.FLAGS.VIEW_CHANNEL | Permissions.FLAGS.MANAGE_CHANNELS | Permissions.FLAGS.CONNECT
  : Permissions.FLAGS.VIEW_CHANNEL | Permissions.FLAGS.MANAGE_CHANNELS;
return permissions.has(Permissions.FLAGS.MANAGE_CHANNELS, false);
```

It's far more efficient, resolves `MemberResolvable` and creates a `Permissions` instance only once (as opposed to twice).

SpaceEEC (approved):

PR #6452, opened 2021-08-16
typings: require component type in raw object formats

monbrey:
Opening this one up for discussion as to pros and cons - this change enforces the `type` property when using JSON object format for components, rather than builders.

This works and enforces the param, but does mean that objects typed as the base typedefs:
 - `MessageActionRowOptions`
 - `MessageButtonOptions`
 - `MessageSelectMenuOptions`

are not compatible with `MessageOptions` anymore, and instead must be typed as:
 - `Required<BaseMessageComponentOptions> & MessageActionRowOptions` for action rows
 - `MessageActionRowComponentOptions` for buttons and selects.

Open to feedback on ways to address that without change the fact that the type is not required in the base typedefs, such as creating separately named types e.g. \
`type RawMessageActionRowData = Required<BaseMessageComponentOptions> & MessageActionRowOptions`

### DIFF
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -4011,7 +4011,9 @@ export type MembershipState = keyof typeof MembershipStates;
 
 export type MessageActionRowComponent = MessageButton | MessageSelectMenu;
 
-export type MessageActionRowComponentOptions = MessageButtonOptions | MessageSelectMenuOptions;
+export type MessageActionRowComponentOptions =
+  | (Required<BaseMessageComponentOptions> & MessageButtonOptions)
+  | (Required<BaseMessageComponentOptions> & MessageSelectMenuOptions);
 
 export type MessageActionRowComponentResolvable = MessageActionRowComponent | MessageActionRowComponentOptions;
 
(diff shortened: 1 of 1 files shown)


suneettipirneni:
Having the separately named type would be helpful, otherwise I'm a fan of enforcing `type` to be provided.

JKLorenzo:
Adding the `type` property to `MessageActionRowOptions`, `BaseButtonOptions`, and `MessageSelectMenuOptions` interfaces actually enforces you to supply a type.

```js
export interface MessageActionRowOptions extends BaseMessageComponentOptions {
  type: 'ACTION_ROW' | MessageComponentTypes.ACTION_ROW;
  components: MessageActionRowComponentResolvable[];
}
```

```js
export interface BaseButtonOptions extends BaseMessageComponentOptions {
  type: 'BUTTON' | MessageComponentTypes.BUTTON;
  disabled?: boolean;
  emoji?: EmojiIdentifierResolvable;
  label?: string;
}
```

```js
export interface MessageSelectMenuOptions extends BaseMessageComponentOptions {
  type: 'SELECT_MENU' | MessageComponentTypes.SELECT_MENU;
  customId?: string;
  disabled?: boolean;
  maxValues?: number;
  minValues?: number;
  options?: MessageSelectOptionData[];
  placeholder?: string;
}
```

#### This also shows that you can't have an action row type on an action row component:

image


#### And lastly, it lets you do type checking:

image


I don't know if this breaks anything though... 😃

monbrey:
> Adding the `type` property to `MessageActionRowOptions`, `BaseButtonOptions`, and `MessageSelectMenuOptions` interfaces actually enforces you to supply a type.

This is exactly what I don't want to do, as it would then enforce specifying type if providing parameters the constructors of these objects, as I said in the issue.
> without change the fact that the type is not required in the base typedefs

iCrawl (approved):

kyranet (approved):

vladfrangu (approved):

SpaceEEC (approved):

PR #6453, opened 2021-08-16
fix(BaseClient): produce an error for non object options

J-Human:

This throws an error to the user if the `options` provided is not an object. It prevents more confusing and misleading errors given by the process. For example, one may have something like this:
```ts
new WebhookClient('id', 'token');
```
Which is, in fact, an outdated way of creating Webhooks. I wouldn't say we *shouldn't* place it because it's the user's fault, though. Since the options parameter is a string, it cannot append the default arguments, hence producing a misleading error:
`cannot append property shardCount on string "..."`
I've seen lots of users in the server that have errors like the above. 

Furthermore, this prevents using the wrong data in other parts of client-related things. Some users don't have a proper IDE, hence they won't receive any feedback from the typings either, this does it for the interpreter.


Keke27210:
Documentation is pretty clear on how to build each class, if people don't want to be bothered to check it, it's their problem.
discord.js is a beginner-friendly library (in spite of itself? e.g. the stricter Snowflake type) but its source code should not be containing extra-(useless)-code.

JavaScript being weakly typed is one of its forces, but also one of its drawbacks (which is why TypeScript was born), but imo it's not up to the library to check for those inputs. If people want to code "safely", they can switch to TypeScript, get a good linter, or even better (but maybe a little bit too much) read documentation.

All of this to say I'm not really a big fan of that, as it would probably start up a trend of implementing type checks everywhere in the source code that'd end up taking a little bit of performance at runtime.

src/client/BaseClient.js
```diff
@@ -13,11 +14,15 @@ class BaseClient extends EventEmitter {
   constructor(options = {}) {
     super();
 
-    /**
-     * The options the client was instantiated with
-     * @type {ClientOptions}
-     */
-    this.options = Util.mergeDefault(Options.createDefault(), options);
+    if (typeof options === 'object') {
```
NotSugden, line 17: i think it would be cleaner to
```js
if (typeof options !== 'object' || options === null) {
  throw [...];
}

this.options = [...];
```

kyranet (approved):

src/client/BaseClient.js
```diff
@@ -13,6 +14,10 @@ class BaseClient extends EventEmitter {
   constructor(options = {}) {
     super();
 
+    if (typeof options !== 'object' || options === null) {
+      throw new Error('INVALID_TYPE', 'options', 'object', true);
```
kyranet, line 18:
```suggestion
      throw new TypeError('INVALID_TYPE', 'options', 'object', true);
```

src/client/BaseClient.js
```diff
@@ -1,6 +1,7 @@
 'use strict';
 
 const EventEmitter = require('events');
+const { Error } = require('../errors');
```
kyranet, line 4:
```suggestion
const { TypeError } = require('../errors');
```

kyranet (changes requested):
Just noticed something, the `'INVALID_TYPE'` error is used exclusively with `TypeError` across the library.

vladfrangu (approved):
LGTM, although we should decide if we validate all inputs, just some, or none.. personally a fan of runtime validation (you can see that in builders)

kyranet (approved):

SpaceEEC (approved):

iCrawl:
This needs a rebase.

J-Human:
Done

iCrawl (approved):

PR #6455, opened 2021-08-16
feat(Guild): Add `Guild#maximumBitrate` getter.

fasterplayer:
This PR adds a getter to Guild which indicates the maximum allowed bitrate in this guild.
This is similar to the `ThreadAutoArchiveDuration['MAX']` option.

### DIFF
--- a/src/structures/Guild.js
+++ b/src/structures/Guild.js
@@ -317,9 +317,9 @@ class Guild extends AnonymousGuild {
     }
 
     /**
-   * The maximum bitrate based on the guild premium_tiers level and the features
-   * @type {?number}
-   */
+     * The maximum bitrate based on the guild premium_tiers level and the features
+     * @type {?number}
+     */
     this.maximumBitrate = 96000;
 
     if (this.features.includes('VIP_REGIONS')) {
--- a/src/structures/Guild.js
+++ b/src/structures/Guild.js
@@ -317,6 +317,33 @@ class Guild extends AnonymousGuild {
     }
 
     /**
+   * The maximum bitrate based on the guild premium_tiers level and the features
+   * @type {?number}
+   */
+    this.maximumBitrate = 96000;
+
+    if (this.features.includes('VIP_REGIONS')) {
+      this.maximumBitrate = 384000;
+    } else {
+      switch (data.premium_tier) {
+        case 0:
+          this.maximumBitrate = 96000;
+          break;
+        case 1:
+          this.maximumBitrate = 128000;
+          break;
+        case 2:
+          this.maximumBitrate = 256000;
+          break;
+        case 3:
+          this.maximumBitrate = 384000;
+          break;
+        default:
+          this.maximumBitrate = 96000;
+      }
+    }
+
+    /**
      * The use count of the vanity URL code of the guild, if any
      * <info>You will need to fetch this parameter using {@link Guild#fetchVanityData} if you want to receive it</info>
      * @type {?number}
--- a/src/structures/Guild.js
+++ b/src/structures/Guild.js
@@ -542,19 +542,19 @@ class Guild extends AnonymousGuild {
   get maximumBitrate() {
     if (this.features.includes('VIP_REGIONS')) {
       return 384000;
-    } else {
-      switch (this.premiumTier) {
-        case PremiumTiers[this.premiumTier] === 0:
-          return 96000;
-        case PremiumTiers[this.premiumTier] === 1:
-          return 128000;
-        case PremiumTiers[this.premiumTier] === 2:
-          return 256000;
-        case PremiumTiers[this.premiumTier] === 3:
-          return 384000;
-        default:
-          return 96000;
-      }
+    } 
+    
+    switch (PremiumTiers[this.premiumTier]) {
+      case 0:
+        return 96000;
+      case 1:
+        return 128000;
+      case 2:
+        return 256000;
+      case 3:
+        return 384000;
+      default:
+        return 96000;
     }
   }
 
--- a/src/structures/Guild.js
+++ b/src/structures/Guild.js
@@ -543,6 +543,7 @@ class Guild extends AnonymousGuild {
     if (this.features.includes('VIP_REGIONS')) {
       return 384000;
     }
+
     switch (PremiumTiers[this.premiumTier]) {
       case PremiumTiers.TIER_1:
         return 128000;
--- a/src/structures/Guild.js
+++ b/src/structures/Guild.js
@@ -317,33 +317,6 @@ class Guild extends AnonymousGuild {
     }
 
     /**
-     * The maximum bitrate based on the guild premium_tiers level and the features
-     * @type {?number}
-     */
-    this.maximumBitrate = 96000;
-
-    if (this.features.includes('VIP_REGIONS')) {
-      this.maximumBitrate = 384000;
-    } else {
-      switch (data.premium_tier) {
-        case 0:
-          this.maximumBitrate = 96000;
-          break;
-        case 1:
-          this.maximumBitrate = 128000;
-          break;
-        case 2:
-          this.maximumBitrate = 256000;
-          break;
-        case 3:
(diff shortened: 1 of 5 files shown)


Jiralite:
Shouldn't this be a getter?

fasterplayer:
I'll update it later today.

fasterplayer:
> Shouldn't this be a getter?

Updated into a getter! Thanks for the advice. I'm new to this and trying to help the best i can!

src/structures/Guild.js
```diff
@@ -534,6 +534,30 @@ class Guild extends AnonymousGuild {
...
+        case PremiumTiers.indexOf(this.premiumTier) === 0:
+          return 96000;
+        case PremiumTiers.indexOf(this.premiumTier) === 1:
+          return 128000;
+        case PremiumTiers.indexOf(this.premiumTier) === 2:
+          return 256000;
+        case PremiumTiers.indexOf(this.premiumTier) === 3:
+          return 384000;
+        default:
+          return 96000;
```
Jiralite, line 556: `PremiumTiers` is actually this:
```JavaScript
{
  '0': 'NONE',
  '1': 'TIER_1',
  '2': 'TIER_2',
  '3': 'TIER_3',
  NONE: 0,
  TIER_1: 1,
  TIER_2: 2,
  TIER_3: 3
}
```
The `.indexOf()` on this will cause a `TypeError`!

Jiralite (changes requested):

src/structures/Guild.js
```diff
@@ -534,6 +534,30 @@ class Guild extends AnonymousGuild {
     );
   }
 
+  /**
+   * Return maximum bitrate based on the guild premium_tiers level and the features
```
Jiralite, line 538: Personally, I'd change this line to this:
```suggestion
   * The maximum bitrate available for this guild
```
I think it's more concise? Just throwing this out there!

src/structures/Guild.js
fasterplayer: Fixed. Damn, it would have been a big mistake.

src/structures/Guild.js
```diff
@@ -534,6 +534,30 @@ class Guild extends AnonymousGuild {
...
+        case PremiumTiers.indexOf(this.premiumTier) === 1:
+          return 128000;
+        case PremiumTiers.indexOf(this.premiumTier) === 2:
+          return 256000;
+        case PremiumTiers.indexOf(this.premiumTier) === 3:
+          return 384000;
+        default:
+          return 96000;
+      }
+    }
```
kyranet, line 558:
```suggestion
    if (this.features.includes('VIP_REGIONS')) {
      return 384000;
    }

    switch (this.premiumTier) {
      case PremiumTiers.NONE:
        return 96000;
      case PremiumTiers.TIER_1:
        return 128000;
      case PremiumTiers.TIER_2:
        return 256000;
      case PremiumTiers.TIER_3:
        return 384000;
      default:
        return 96000;
    }
```

Alternatively, we can also define an object somewhere and do the following:

```javascript
return this.features.includes('VIP_REGIONS') ? 384_000 : PremiumBitrates[this.premiumTier];
```

kyranet (changes requested):

src/structures/Guild.js
```diff
@@ -534,6 +534,30 @@ class Guild extends AnonymousGuild {
...
+          return 96000;
+        case PremiumTiers[this.premiumTier] === 1:
+          return 128000;
+        case PremiumTiers[this.premiumTier] === 2:
+          return 256000;
+        case PremiumTiers[this.premiumTier] === 3:
+          return 384000;
+        default:
+          return 96000;
+      }
```
NotSugden, line 557: this isn't how switch statements work...

NotSugden (changes requested):

src/structures/Guild.js
```diff
@@ -534,6 +534,30 @@ class Guild extends AnonymousGuild {
...
+   * @readonly
+   */
+  get maximumBitrate() {
+    if (this.features.includes('VIP_REGIONS')) {
+      return 384000;
+    } 
+    
+    switch (PremiumTiers[this.premiumTier]) {
+      case 0:
+        return 96000;
```
Keke27210, line 549: Wouldn't it be better to have a proper enum for that instead of leaving magic numbers in the middle of the code?

src/structures/Guild.js
fasterplayer: > this isn't how switch statements work...

You are right. I did it all wrong. I think that the last commit would be the best.


>Alternatively, we can also define an object somewhere and do the following:

The reason i went with a switch statement is that if tomorrow, discord releases a Tier 4, we won't have an issue as it would at least return the default bitrate.  But other options would also work.

src/structures/Guild.js
fasterplayer: > Wouldn't it be better to have a proper enum for that instead of leaving magic numbers in the middle of the code?

It could be a better idea. If you all prefer to go with this option, i can do it.

ImRodry: I don't think an enum is the way to go because these are just maximum values and bitrates can have all sorts of values.
Either way did you test this? Are you 100% sure this works?

src/structures/Guild.js
```diff
@@ -534,6 +534,30 @@ class Guild extends AnonymousGuild {
...
+   * @type {number}
+   * @readonly
+   */
+  get maximumBitrate() {
+    if (this.features.includes('VIP_REGIONS')) {
+      return 384000;
+    } 
+    
+    switch (PremiumTiers[this.premiumTier]) {
+      case 0:
```
NotSugden, line 548:
```suggestion
      default:
      case 0:
```
seeing as they are the same this is valid

iCrawl: Not even sure why you would suggest invalid code.

fasterplayer: True. This is not working. Switch back to the previous default.

NotSugden: bit of a derp the logic behind it yes, did edit though to suggest a much more logical approach

NotSugden: ah i see, the code was invalid because i thought it would be implied to remove the already existing default case

but of course as i said removing the 0 case is a smarter approach

fasterplayer: Yea, that's was i was going to commit.

src/structures/Guild.js
fasterplayer: Yes, it is! Tested!

src/structures/Guild.js
```diff
@@ -534,6 +534,27 @@ class Guild extends AnonymousGuild {
...
+  get maximumBitrate() {
+    if (this.features.includes('VIP_REGIONS')) {
+      return 384000;
+    }
+    switch (PremiumTiers[this.premiumTier]) {
+      case 1:
+        return 128000;
+      case 2:
+        return 256000;
+      case 3:
```
kyranet, line 551:
```suggestion
      case PremiumTiers.TIER_1:
        return 128000;
      case PremiumTiers.TIER_2:
        return 256000;
      case PremiumTiers.TIER_3:
```

kyranet (changes requested):

kyranet (approved):

iCrawl (approved):

src/structures/Guild.js
```diff
@@ -534,6 +534,27 @@ class Guild extends AnonymousGuild {
...
+  /**
+   * The maximum bitrate available for this guild
+   * @type {number}
+   * @readonly
+   */
+  get maximumBitrate() {
+    if (this.features.includes('VIP_REGIONS')) {
+      return 384000;
+    }
+    switch (PremiumTiers[this.premiumTier]) {
```
vladfrangu, line 546:
```suggestion
    }

    switch (PremiumTiers[this.premiumTier]) {
```

vladfrangu (changes requested):

vladfrangu (approved):

PR #6462, opened 2021-08-17
refactor(RequestHandler): Do not return await

summetdev:

- There's no reason `return`'ing `await ...` because function ([RequestHandler#push](https://github.com/discordjs/discord.js/blob/main/src/rest/RequestHandler.js#L50)) itself returns a `Promise`.
- [`no-return-await`](https://github.com/discordjs/discord.js/blob/main/.eslintrc.json#L79) is also enabled.

### DIFF
--- a/src/rest/RequestHandler.js
+++ b/src/rest/RequestHandler.js
@@ -47,7 +47,7 @@ class RequestHandler {
   async push(request) {
     await this.queue.wait();
     try {
-      return await this.execute(request);
+      return this.execute(request);
     } finally {
       this.queue.shift();
     }


monbrey:
I believe this will change the behaviour of exceptions inside this try/catch block if it's not awaited.

ckohen:
As mon said, this is awaited due to the finally call, which is intended. eslint doesn't error on this because the behavior is different.

E.g.
Not awaited:
Request is queued, waits for previous requests, as the request goes out, return active request, finally block is called and lets the next request pass even though this hasn't finished

Awaited:
Request is queued, waits for previous requests, wait for request to complete, return result, finally block is called and lets the next request pass

summetdev:
OK, thank you for clarification!

PR #6466, opened 2021-08-17
feat(Threads): update permissions and fix getters

advaith1:
- Docs PR: https://github.com/discord/discord-api-docs/pull/3672
- Update USE_*\_THREADS description (rename to CREATE_ in v14)
- Adds SEND_MESSAGES_IN_THREADS
- Fixes permission getters checking outdated `private_thread` type
- Fixes sendable checking archived but not locked or manageable
- Updates sendable to use the new permission
- Fixes sendable sometimes returning undefined (https://github.com/discordjs/discord.js/pull/6466#discussion_r690652609)


src/util/Permissions.js
```diff
@@ -91,9 +91,10 @@ class Permissions extends BitField {
  * * `USE_APPLICATION_COMMANDS`
  * * `REQUEST_TO_SPEAK`
  * * `MANAGE_THREADS`
- * * `USE_PUBLIC_THREADS`
- * * `USE_PRIVATE_THREADS`
+ * * `USE_PUBLIC_THREADS` (create threads)
+ * * `USE_PRIVATE_THREADS` (create private threads)
```
ckohen, line 95: You can add a TODO comment here and in the object so we don't forget

ckohen (changes requested):

src/structures/ThreadChannel.js
```diff
@@ -423,15 +425,10 @@ class ThreadChannel extends Channel {
...
-      this.permissionsFor(this.client.user)?.any(
-        [
-          Permissions.FLAGS.SEND_MESSAGES,
-          this.type === 'GUILD_PRIVATE_THREAD'
-            ? Permissions.FLAGS.USE_PRIVATE_THREADS
-            : Permissions.FLAGS.USE_PUBLIC_THREADS,
-        ],
+      !(this.archived && this.locked && !this.manageable) &&
+      (this.type !== 'GUILD_PRIVATE_THREAD' || this.joined || this.manageable) &&
+      this.permissionsFor(this.client.user)?.has(
```
NotSugden, line 430: you need to account for `this.permissionsFor(...)` returning null, if its null the getter will return undefined

src/structures/ThreadChannel.js
```diff
@@ -423,15 +425,10 @@ class ThreadChannel extends Channel {
...
+        this.isThread() ? Permissions.FLAGS.SEND_MESSAGES_IN_THREADS : Permissions.FLAGS.SEND_MESSAGES,
```
ckohen, line 431: `this` is always a thread, no need to check it here

src/structures/ThreadChannel.js
ckohen: The getter would return false because its in a giant boolean logic chain and null or undefined == false right?

advaith1: actually i think sugden is correct here; it's the last one in the chain so if the first 2 are falsy, the value of the last one will be returned

ckohen: oh right there's no Boolean(), yeah

src/util/Permissions.js
iCrawl, line 95:
```suggestion
 * * `USE_PUBLIC_THREADS` (deprecated)
 * * `USE_PRIVATE_THREADS` (deprecated)
```

And add the new ones in the appropriate place.

src/util/Permissions.js
```diff
@@ -132,9 +133,11 @@ Permissions.FLAGS = {
   USE_APPLICATION_COMMANDS: 1n << 31n,
   REQUEST_TO_SPEAK: 1n << 32n,
   MANAGE_THREADS: 1n << 34n,
+  // TODO: Rename USE_*_THREADS to CREATE_*_THREADS in v14
   USE_PUBLIC_THREADS: 1n << 35n,
   USE_PRIVATE_THREADS: 1n << 36n,
```
iCrawl, line 138: Same here, keep those, and add the new ones with the new name.

typings/index.d.ts
```diff
@@ -4331,7 +4331,8 @@ export type PermissionString =
   | 'MANAGE_THREADS'
   | 'USE_PUBLIC_THREADS'
   | 'USE_PRIVATE_THREADS'
```
iCrawl, line 4333: Same as above.

src/util/Permissions.js
```diff
@@ -132,9 +135,13 @@ Permissions.FLAGS = {
   USE_APPLICATION_COMMANDS: 1n << 31n,
   REQUEST_TO_SPEAK: 1n << 32n,
   MANAGE_THREADS: 1n << 34n,
+  // TODO: Remove deprecated USE_*_THREADS flags in v14
   USE_PUBLIC_THREADS: 1n << 35n,
+  CREATE_PUBLIC_THREADS: 1n << 35n,
   USE_PRIVATE_THREADS: 1n << 36n,
+  CREATE_PRIVATE_THREADS: 1n << 36n,
```
kyranet, line 142: Just one question:

With this patch, what does `new Permissions(3n << 35n).toArray()` return? `['USE_PUBLIC_THREADS', 'USE_PRIVATE_THREADS']` or `['CREATE_PUBLIC_THREADS', 'CREATE_PRIVATE_THREADS']`?

almeidx: @kyranet neither. It actually returns:
```js
[
  'USE_PUBLIC_THREADS',
  'CREATE_PUBLIC_THREADS',
  'USE_PRIVATE_THREADS',
  'CREATE_PRIVATE_THREADS'
]
```

kyranet: Isn't this actually breaking? @discordjs/the-big-3

vladfrangu: I don't understand how flag `1n << 35n` can return both public and private... Other than that..I think it should be fine?

kyranet: Not really, at least in my bot (which relies on `Permissions#toArray` for a few things), this would cause duplicates to happen, which can be troublesome, that's why I think this change is `semver: major`.

The rest of the changes in this PR are fine and fit in either `semver: patch` or `semver: minor`.

SpaceEEC: This scenario already happened once.
We used to have aliases (due to name changes) for `USE_EXTERNAL_EMOJIS` and `MANAGE_ROLES` until the old names were removed with v12.
(Introduced with https://github.com/discordjs/discord.js/commit/16fe48d405f4dd9c40a68737b05aedc3f4f0d201)
I do not remember any trouble around these. Therefore I suppose it's fine to add these, not that they are strictly needed.

SpaceEEC (approved):

iCrawl (approved):
This needs a rebase.

kyranet (approved):

vladfrangu (approved):

PR #6469, opened 2021-08-18
feat(CommandInteractionOptionResolver): make `subcommand` and `group` public, and add raw `hoistedOptions`

QSmally:

While in the process of engineering my bot application, I found the `CommandInteractionOptionResolver`'s way of managing subcommand(group)s rather unnecessary.

As `_group` and `_subcommand` were already processed, there was no need to implement two separate methods fetching it (`getSubcommandGroup` and `getSubcommand`), except for consistency and the `required` parameter. I ended up using the `_group` property instead, and I'm proposing that it should be public because of that.

Besides that, `_hoistedOptions` is also rather useful to navigate yourself but wasn't public either. I implemented a change similar to the `data` property (which is the same, just not hoisted).

People can now navigate the options with `group`, `subcommand` and `hoistedOptions` to manage the interaction arguments themselves (like I would do, personally, with my own implementation).

### DIFF
--- a/src/structures/CommandInteractionOptionResolver.js
+++ b/src/structures/CommandInteractionOptionResolver.js
@@ -18,16 +18,14 @@ class CommandInteractionOptionResolver {
     /**
      * The name of the subcommand group.
      * @type {?string}
-     * @private
      */
-    this._group = null;
+    this.group = null;
 
     /**
      * The name of the subcommand.
      * @type {?string}
-     * @private
      */
-    this._subcommand = null;
+    this.subcommand = null;
 
     /**
      * The bottom-level options for the interaction.
@@ -39,12 +37,12 @@ class CommandInteractionOptionResolver {
 
     // Hoist subcommand group if present
     if (this._hoistedOptions[0]?.type === 'SUB_COMMAND_GROUP') {
-      this._group = this._hoistedOptions[0].name;
+      this.group = this._hoistedOptions[0].name;
       this._hoistedOptions = this._hoistedOptions[0].options ?? [];
     }
     // Hoist subcommand if present
     if (this._hoistedOptions[0]?.type === 'SUB_COMMAND') {
-      this._subcommand = this._hoistedOptions[0].name;
+      this.subcommand = this._hoistedOptions[0].name;
       this._hoistedOptions = this._hoistedOptions[0].options ?? [];
     }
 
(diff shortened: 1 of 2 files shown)


monbrey:
We intentionally made these design decisions already to provide a consistent interface. The resolver provides additional error handling and type checking of arguments and this encourages it's usage.

I don't think we want to start providing every middle-ground of half-parsed stages of raw data.

QSmally:
@monbrey - I might agree to that, although the `group` and `subcommand` properties were quite useful to me personally.

monbrey:
Why are they more or less useful than `getSubcommandGroup` and `getSubcommand` which provide optional error throwing?

QSmally:
For those who would want to process the command arguments, including subcommands, on their own - `getSubcommand` exists, but it requires the programmer to do `getSubcommand(false)` each time while they ultimately can just use the `group` property. It might not be as big of a deal as the latter change I proposed, but at the same time, these properties are already computed at initialisation of the resolver instance.

src/structures/CommandInteractionOptionResolver.js
```diff
@@ -56,6 +54,14 @@ class CommandInteractionOptionResolver {
      */
     Object.defineProperty(this, 'data', { value: Object.freeze([...options]) });
 
+    /**
+     * The hoisted interaction options array, so without any subcommand(groups).
+     * @name CommandInteractionOptionResolver#hoistedOptions
+     * @type {ReadonlyArray<CommandInteractionOption>}
+     * @readonly
```
ImRodry, line 61: I don't fully agree with this PR but this readonly tag is only for getters iirc

QSmally: @ImRodry - I pretty much copied the `data` property which is right above it. I wasn't sure how to document it otherwise

kyranet: `@readonly` is also used for `Object.defineProperty` too.

shinotheshino:
I don't think this is necessary. Users can use the `.data` option if they want to do their own handling. This PR just adds redundancy and goes against the original design choices for adding `.getCommand()` and `getSubcommand()`.

QSmally:
I'd argue that the methods are unconventional. Both properties, `subcommand` and `group` are already computed. And except for the line that throws an error if it's not there, it's practically useless in my opinion.

ImRodry:
I said it was useless on the PR that added them too but they kept justifying that they threw errors so that was a good thing even though you should only really use these in commands that you know have subcommands or groups. If it didn't change then, I doubt it's gonna change now that v13 is out tbh

PR #6476, opened 2021-08-20
typings: Allow message component interaction collectors to infer collected interaction types

suneettipirneni:

As it stands now if you use this code:

```ts
const collector = message.createMessageComponentCollector({ componentType: 'SELECT_MENU' });

collector.on('collect', interaction => {
  // Interaction is type of `MessageComponentInteraction`
});
```

You'll end up with a `MessageComponentInteraction` in the `collect` handler.

Currently there are two trivial ways to narrow this type.

1) Type Guard it:
```ts
collector.on('collect', interaction => {
  if (!interaction.isSelectMenu()) {
    return;
  }
  // Interaction is type of `SelectMenuInteraction`
});
```

2) Use the generic parameter:
```ts
.createMessageComponentCollector<SelectMenuInteraction>(...);
```

However, even though these methods are straightforward, these types really shouldn't have to be narrowed at all for collectors, where `componentType` is specified.

We already stated the `componentType` in the `options`, which means we have proven which component types we should receive.  In addition the d.js collector will *always* return a `ButtonInteraction` if `componentType` is `'BUTTON'` and a `SelectMenuInteraction` if `componentType` is a `'SELECT_MENU'`. If all else fails then we know at the very least it's a `MessageComponentInteraction`.  The function should infer these parameters and give the correct `InteractionCollector<T>` as a return value.

This PR addresses this by inferring interaction collector types via the options given, this allows our first example to properly deduce that, in fact, `SelectMenuInteraction` is being collected rather than `MessageComponentInteraction`:

```ts
const collector = message.createMessageComponentCollector({ componentType: 'SELECT_MENU' });

collector.on('collect', interaction => {
  // Interaction is type of `SelectMenuInteraction`, 
  // function properly inferred types from given parameters.
});
```

### DIFF
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -1171,7 +1171,9 @@ export class Message extends Base {
   public createMessageComponentCollector<
     T extends MessageComponentType | MessageComponentTypes = MessageComponentTypes.ACTION_ROW,
   >(
-    options?: { componentType?: T } & InteractionCollectorOptionsResolvable,
+    options?:
+      | ({ componentType?: T } & InteractionCollectorOptionsResolvable)
+      | InteractionCollectorOptions<MessageComponentInteraction>,
   ): ConditionalType<MappedInteractionCollectorOptions[T]>;
   public delete(): Promise<Message>;
   public edit(content: string | MessageEditOptions | MessagePayload): Promise<Message>;
@@ -2714,7 +2716,9 @@ export interface TextBasedChannelFields extends PartialTextBasedChannelFields {
   createMessageComponentCollector<
     T extends MessageComponentType | MessageComponentTypes = MessageComponentTypes.ACTION_ROW,
   >(
-    options?: { componentType?: T } & InteractionCollectorOptionsResolvable,
+    options?:
+      | ({ componentType?: T } & InteractionCollectorOptionsResolvable)
+      | InteractionCollectorOptions<MessageComponentInteraction>,
   ): ConditionalType<MappedInteractionCollectorOptions[T]>;
   sendTyping(): Promise<void>;
 }
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -1091,7 +1091,7 @@ export class LimitedCollection<K, V> extends Collection<K, V> {
   public static filterByLifetime<K, V>(options?: LifetimeFilterOptions<K, V>): SweepFilter<K, V>;
 }
 
-type ConditionalInteractionCollector<T extends InteractionCollectorOptionsResolvable> =
+type ConditionalInteractionCollector<T extends InteractionCollectorOptionsResolvable | undefined> =
   T extends MessageInteractionCollectorOptions
     ? InteractionCollector<MessageComponentInteraction>
     : T extends ButtonInteractionCollectorOptions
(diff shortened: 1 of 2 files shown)


AustrianDev:
First, I can't see any code (only typings)
And for me, this isn't required because it can already be filtered with the filter option :

```js
const filter = (interaction) => interaction.isButton();

const collector = message.createMessageComponentCollector({
  filter
});
```

DTrombett:
Wdym? The filter parameter has nothing to do with the type of `interaction` in the collector.

Also, wouldn't be good to add tests for this in [`typings/tests.ts`](https://github.com/suneettipirneni/discord.js/blob/typings/collector-inference/typings/tests.ts)

suneettipirneni:
> Also, wouldn't be good to add tests for this in [`typings/tests.ts`](https://github.com/suneettipirneni/discord.js/blob/typings/collector-inference/typings/tests.ts)?

Tests have been added

typings/index.d.ts
```diff
@@ -3911,7 +3917,7 @@ export interface IntegrationAccount {
   name: string;
 }
 
-export interface InteractionCollectorOptions<T extends Interaction> extends CollectorOptions<[T]> {
+export interface InteractionCollectorOptions<T extends Interaction> extends CollectorOptions<T[]> {
```
vladfrangu, line 3920: This will make it loose strictness and accuracy... It should stay as a tuple, or otherwise always pass in a tuple to T

vladfrangu (changes requested):

typings/index.d.ts
suneettipirneni: Good point, that has been fixed.

typings/tests.ts
```diff
@@ -484,6 +489,31 @@ client.on('messageCreate', ({ channel }) => {
...
+  );
+  assertType<InteractionCollector<SelectMenuInteraction>>(selectMenuCollector);
+
+  // Verify that message component interactions are default collected types.
+  const defaultCollector = message.createMessageComponentCollector();
+  assertType<Promise<InteractionCollector<MessageComponentInteraction>>>(message.awaitMessageComponent());
+  assertType<InteractionCollector<MessageComponentInteraction>>(defaultCollector);
+
+  // Verify that additional options don't affect default collector types.
+  const semiDefaultCollector = message.createMessageComponentCollector({ interactionType: 'APPLICATION_COMMAND' });
```
vladfrangu, line 515: Doesn't this throw an error / shouldn't it throw a type error?

vladfrangu (approved):
LGTM, just one question

typings/tests.ts
suneettipirneni: So `interactionType` is an allowed option currently for `InteractionCollectorOptions` which it seems that `createMesaageComponent()` uses, IIRC thats how it was setup before this PR.

But I do see your point, this may be a mistake, since interaction type doesn't have anything to do with component interactions.

kyranet (approved):
Those are some very complex types, but as long as tests pass, LGTM

iCrawl (approved):

SpaceEEC (approved):

PR #6478, opened 2021-08-20
feat: export BaseCommandInteraction

wakfi:
Please describe the changes this PR makes and why it should be merged:
This PR adds BaseCommandInteraction to the index.js file.

This PR changes the library's interface (methods or parameters added)

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -71,6 +71,7 @@ module.exports = {
   Application: require('./structures/interfaces/Application'),
   ApplicationCommand: require('./structures/ApplicationCommand'),
   Base: require('./structures/Base'),
+  BaseCommandInteraction: require('./structures/BaseCommandInteraction'),
   BaseGuild: require('./structures/BaseGuild'),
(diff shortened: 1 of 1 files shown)


monbrey:
What's the reason for exporting this one? It shouldn't ever be constructed, everything should be a subclass of it.

wakfi:
I'm wanting to use it for polymorphic behavior. I have an Interaction, which could be any of the subclasses, but I'm only interested in accessing the properties defined in the abstract BaseCommandInteraction. I'd like to be able to run `instanceof` checks with BaseCommandInteraction to accomplish this, rather than needing to check each subclass separately. 

I'm working in TypeScript, and this class is exported in the type definitions which makes it appear to be usable in this way, but since it isn't exported in index.js the import is always undefined.

vladfrangu (approved):

iCrawl (approved):

SpaceEEC (approved):

PR #6484, opened 2021-08-21
feat(Message): Add method to resolve a component by id

Jiralite:
Now that the API enforces custom ids to be unique in components in a message, I want to go ahead and propose this new method!

This will add the `<Message>.resolveComponent()` method. By supplying a custom id, the desired component will be returned (or `null` if not found). The use case for this is when the desired behaviour is altering the state of some other specific component upon the interaction of a component on the same message. For me, this is altering the `.disabled` property of a button or a select menu.

I understand I can simply index the components array like so: `<ButtonInteraction>.message.components[<number>].components[<number>]`. However, I don't think this is quite right. I am assuming that the component I want to edit will always be on a specific row on a specific column. Why should I do this when I have a custom id? Finding a component by a custom id seems like a more intuitive approach, hence why I have this method!

Feel free to suggest/request changes!

### DIFF
--- a/src/structures/Message.js
+++ b/src/structures/Message.js
@@ -817,6 +817,15 @@ class Message extends Base {
   }
 
   /**
+   * Resolves a component by a custom id.
+   * @param {string} customId The custom id to resolve against
+   * @returns {?MessageActionRowComponent}
+   */
+  resolveComponent(customId) {
+    return this.components.flatMap(row => row.components).find(component => component.customId === customId) ?? null;
+  }
+
+  /**
    * Used mainly internally. Whether two messages are identical in properties. If you want to compare messages
    * without checking all the properties, use `message.id === message2.id`, which is much more efficient. This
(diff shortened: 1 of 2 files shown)


vladfrangu (approved):

ImRodry:
This doesn’t allow you to resolve link components, why not check for that too?

Jiralite:
Link buttons don't have a custom id, so there is nothing unique about them. You can have the same structure of a link button for 25 buttons in a message. It's not ideal to have a  resolver for something that is not unique (just yet)!

ImRodry:
I was suggesting using the same logic but for the button's URL. I know you can have the same URL in different buttons but I don't see why anyone would do that. You could also add a warning saying it will return the first button with the given URL if a URL is passed.
Alternatively I think you should add some sort of note saying LINK buttons will not work as they do not have customIds

SpaceEEC (approved):

iCrawl (approved):

kyranet (approved):

PR #6485, opened 2021-08-21
docs(ThreadManager): Fix typo

Jiralite:
Fixes a typo. Also tweaked the wording to refer to the guild's features as this is more accurate.

### DIFF
--- a/src/managers/ThreadManager.js
+++ b/src/managers/ThreadManager.js
@@ -66,7 +66,7 @@ class ThreadManager extends CachedManager {
    * * `1440` (1 day)
    * * `4320` (3 days) <warn>This is only available when the guild has the `THREE_DAY_THREAD_ARCHIVE` feature.</warn>
    * * `10080` (7 days) <warn>This is only available when the guild has the `SEVEN_DAY_THREAD_ARCHIVE` feature.</warn>
-   * * `'MAX'` Based on the guilds boost count
+   * * `'MAX'` Based on the guild's features
    * @typedef {number|string} ThreadAutoArchiveDuration
    */
 


vladfrangu (approved):

src/managers/ThreadManager.js
```diff
@@ -66,7 +66,7 @@ class ThreadManager extends CachedManager {
    * * `1440` (1 day)
    * * `4320` (3 days) <warn>This is only available when the guild has the `THREE_DAY_THREAD_ARCHIVE` feature.</warn>
    * * `10080` (7 days) <warn>This is only available when the guild has the `SEVEN_DAY_THREAD_ARCHIVE` feature.</warn>
-   * * `'MAX'` Based on the guilds boost count
+   * * `'MAX'` Based on the guild's features
```
ImRodry, line 69: Why remove this
```suggestion
   * * `'MAX'` Based on the guild's boost level
```

ckohen: It's not *necessarily* the boost count that grants features. They are usually tied, but some guilds may have this feature with even 0 boosts.

ImRodry: Yeah only if they’re partnered or verified, in which case boost level 3 is implicit even with no boosts

ajpalkovic: I wouldn't say that is necessarily accurate

the guild feature will exist for private threads if they have private threads, same for 3/7 day times

SpaceEEC (approved):

iCrawl (approved):

kyranet (approved):

PR #6488, opened 2021-08-21
feat(ThreadChannel): add fetchStarterMessage

ImRodry:
This PR adds a method to ThreadChannels called fetchStarterMessage that fetches the message that started the thread when this exists. I didn't find any better way to tell if a thread channel was started by a message or not so this just fetches the message with the same ID as the channel from the parent channel.

### DIFF
--- a/src/structures/ThreadChannel.js
+++ b/src/structures/ThreadChannel.js
@@ -254,6 +254,17 @@ class ThreadChannel extends Channel {
   }
 
   /**
+   * Fetches the message that started this thread, if any.
+   * <info>This in only available when the thread started from a message in the parent channel.
+   * If you just need the id of that message, use {@link ThreadChannel#id this channel's id} instead.</info>
+   * @param {BaseFetchOptions} [options] Additional options for this fetch
+   * @returns {Promise<?Message>}
+   */
+  fetchStarterMessage(options) {
+    return this.channel.messages.fetch(this.id, options);
+  }
+
+  /**
    * The options used to edit a thread channel
    * @typedef {Object} ThreadEditData
    * @property {string} [name] The new name for the thread
--- a/src/structures/ThreadChannel.js
+++ b/src/structures/ThreadChannel.js
@@ -255,10 +255,10 @@ class ThreadChannel extends Channel {
 
   /**
    * Fetches the message that started this thread, if any.
-   * <info>This in only available when the thread started from a message in the parent channel.
-   * If you just need the id of that message, use {@link ThreadChannel#id this channel's id} instead.</info>
+   * <info>This only works when the thread started from a message in the parent channel, otherwise the promise will
+   * reject. If you just need the id of that message, use {@link ThreadChannel#id this channel's id} instead.</info>
    * @param {BaseFetchOptions} [options] Additional options for this fetch
-   * @returns {Promise<?Message>}
+   * @returns {Promise<Message>}
    */
   fetchStarterMessage(options) {
     return this.channel.messages.fetch(this.id, options);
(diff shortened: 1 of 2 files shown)


ImRodry:
After thinking about this for a while I opted by adding a method instead to be more in line with the direction the library has been going in recently. The getter was unreliable as it got the message from cache so this just simply fetches it.

monbrey:
When can this method return `null`? Wouldn't it throw if it can't fetch the message?

ImRodry:
> When can this method return `null`? Wouldn't it throw if it can't fetch the message?

Good point, will update

src/structures/ThreadChannel.js
```diff
@@ -253,6 +253,17 @@ class ThreadChannel extends Channel {
     return members.get(this.ownerId) ?? null;
   }
 
+  /**
+   * Fetches the message that started this thread, if any.
+   * <info>This only works when the thread started from a message in the parent channel, otherwise the promise will
+   * reject. If you just need the id of that message, use {@link ThreadChannel#id this channel's id} instead.</info>
```
vladfrangu, line 259: I don't see the point of that documentation for ThreadChannel#id link - especially if threads ever stop sharing IDs with their message

```suggestion
   * reject. If you just need the id of that message, use {@link ThreadChannel#id} instead.</info>
```

vladfrangu (dismissed):

src/structures/ThreadChannel.js
ImRodry: If they intentionally made it like this, I doubt this will ever change, but I'll apply that suggestion

src/structures/ThreadChannel.js
```diff
@@ -253,6 +253,17 @@ class ThreadChannel extends Channel {
...
 
+  /**
+   * Fetches the message that started this thread, if any.
+   * <info>This only works when the thread started from a message in the parent channel, otherwise the promise will
+   * reject. If you just need the id of that message, use {@link ThreadChannel#id} instead.</info>
+   * @param {BaseFetchOptions} [options] Additional options for this fetch
+   * @returns {Promise<Message>}
+   */
+  fetchStarterMessage(options) {
+    return this.channel.messages.fetch(this.id, options);
```
PuneetGopinath, line 264:
```suggestion
    return this.channel.messages.fetch(this.id, options) ?? null;
```

PuneetGopinath (commented):
I guess it has to be nullable as I tried deleting the starter message in a thread

ImRodry:
@PuneetGopinath refer to monbrey’s comment: https://github.com/discordjs/discord.js/pull/6488#issuecomment-903110014

src/structures/ThreadChannel.js
ImRodry: Even if fetch didn’t throw an error this is still invalid

NotSugden: this is not valid, `MessageManager#fetch` (`MessageManager#_fetchId`) does not return null

PuneetGopinath:
> @PuneetGopinath refer to monbrey’s comment: [#6488 (comment)](https://github.com/discordjs/discord.js/pull/6488#issuecomment-903110014)

Ohk. Didn't notice, but the `?? null` wasn't there.

PuneetGopinath:
> @PuneetGopinath refer to monbrey’s comment: [#6488 (comment)](https://github.com/discordjs/discord.js/pull/6488#issuecomment-903110014)

Ohk.

monbrey:
> Ohk. Didn't notice, but the `?? null` wasn't there.

Nor should it be there, because fetch will either resolve to a Message, or throw. Nullish-coallescing is not possible or correct in any sense.

vladfrangu (approved):

SpaceEEC (approved):

kyranet (approved):

iCrawl (approved):

PR #6493, opened 2021-08-21
feat: add support for GuildScheduledEvent

iShibi:

Adds support for `GuildScheduledEvent`:
- discord/discord-api-docs#3586

Todo:
- [x] Add `GuildScheduledEvent` class
- [x] Define all properties of the `GuildScheduledEvent` class
- [x] Add Complete Documentation
- [x] Add Complete Typings
- [x] `Get` List of guild events in a guild
- [x] `Post` Create a guild event
- [x] `Get` A guild event by id
- [x] `Delete` Delete a guild event
- [x] `Patch` Modify a guild event
- [x] Add `GuildEventManager`
- [x] Add `MANAGE_EVENTS` permission
- [x] Add Gateway Events:
   - [x] GUILD_SCHEDULED_EVENT_CREATE
   - [x] GUILD_SCHEDULED_EVENT_UPDATE
   - [x] GUILD_SCHEDULED_EVENT_DELETE
   - [x] GUILD_SCHEDULED_EVENT_USER_ADD
   - [x] GUILD_SCHEDULED_EVENT_USER_REMOVE
- [x] Add `GuildScheduledEvent` related audit logs

- Code changes have been tested against the Discord API (⚠️ Lots of changes, I def missed some)


src/util/Constants.js
```diff
@@ -1031,6 +1031,24 @@ exports.PrivacyLevels = createEnum([null, 'PUBLIC', 'GUILD_ONLY']);
  */
 exports.PremiumTiers = createEnum(['NONE', 'TIER_1', 'TIER_2', 'TIER_3']);
 
+/**
+ * The entity type of a {@link GuildEvent}:
+ * * SCHEDULED
+ * * ACTIVE
+ * * COMPLETED
+ * * CANCELED
+ * @typedef {string} GuildEventEntityType
```
ImRodry, line 1040: Once the upstream PR lands, please add the `@see` link (maybe add a TODO comment so you don't forget)

src/util/Constants.js
```diff
@@ -1031,6 +1031,24 @@ exports.PrivacyLevels = createEnum([null, 'PUBLIC', 'GUILD_ONLY']);
...
+ */
+exports.GuildEventEntityTypes = createEnum([null, 'SCHEDULED', 'ACTIVE', 'COMPLETED', 'CANCELED']);
+
+/**
+ * The status of a {@link GuildEvent}:
+ * * NONE
+ * * STAGE_INSTANCE
+ * @typedef {string} GuildEventStatus
```
ImRodry, line 1048: Same as above

src/structures/GuildEvent.js
```diff
@@ -0,0 +1,125 @@
...
+
+/**
+ * Represents an event in a {@link Guild}.
+ * @extends {Base}
+ */
+class GuildEvent extends Base {
+  /**
+   *
+   * @param {Client} client The instantiating client
+   * @param {*} data The data for the guild event
```
ImRodry, line 14: Once the upstream PR lands, please add the `@external` API docs link (maybe add a TODO comment so you don't forget)

src/structures/GuildEvent.js
```diff
@@ -0,0 +1,125 @@
...
+    /**
+     * The entity id
+     * @type {?Snowflake}
+     */
+    this.entityId = data.entity_id;
+
+    /**
+     * The metadata for the guild event
+     */
+    this.entityMetadata = data.entity_metadata;
```
ImRodry, line 103: You can probably make the typedef for this already

ImRodry (commented):
I know this is still a draft, these are just some notes for the future mostly

src/util/Constants.js
iShibi: sure, I'll resolve these convos only when it's done so that I don't forget

src/structures/GuildEvent.js
```diff
@@ -0,0 +1,128 @@
...
+     */
+    this.guildId = data.guild_id;
+
+    this._patch(data);
+  }
+
+  _patch(data) {
+    if ('channel_id' in data) {
+      /**
+       * The id of the stage channel this event belongs to
```
ImRodry, line 38: Oh forgot this but this is why the CI is failing
```suggestion
       * The id of the stage channel this event belongs to
       * @type {Snowflake}
```

src/managers/GuildEventManager.js
```diff
@@ -0,0 +1,171 @@
...
+   */
+
+  /**
+   * Obtains one or multiple guild events from Discord, or the guild cache if it's aready available.
+   * @param {GuildEventResolvable|FetchGuildEventOptions|FetchGuildEventsOptions} [options] The id of the guild event or
+   * options
+   * @returns {Promise<GuildEvent|Collection<Snowflake, GuildEvent>>}
+   */
+  async fetch(options = {}) {
+    const id = this.resolveId(options) ?? this.resolveId(options.guildEvent);
```
NotSugden, line 53:
```suggestion
    const id = this.resolveId(options.guildEvent ?? options);
```
this would probably make more sense

src/managers/GuildEventManager.js
```diff
@@ -0,0 +1,171 @@
...
+  /**
+   * Creates a new guild event.
+   * @param {GuildEventCreateOptions} options Options for creating the guild event
+   * @returns {Promise<GuildEvent>}
+   */
+  async create(options) {
+    if (typeof options !== 'object') throw new TypeError('INVALID_TYPE', 'options', 'object', true);
+    let { privacyLevel, entityType, channel } = options;
+
+    if (!privacyLevel) throw new TypeError('INVALID_TYPE', 'privacyLevel', 'string or number', false);
```
NotSugden, line 94: this would throw on `0`, granted 0 is not a valid privacy level but the error would seem to indicate that neither a string or a number was passed

src/managers/GuildEventManager.js
ImRodry: You’d need to add nullish coalescing then because options might be a snowflake

NotSugden: a snowflake isn't null/undefined

ImRodry: Ah nvm I thought that would throw an error but I just confused it for sth else

src/managers/GuildEventManager.js
iShibi: Not sure why I chose to validate `privacyLevel` here. The lib leaves the validation of these types of required values for the api. Changed it in f1e6d8e

src/managers/GuildEventManager.js
```diff
@@ -0,0 +1,167 @@
...
+
+  /**
+   * Options used to create a guild event.
+   * @typedef {Object} GuildEventCreateOptions
+   * @property {string} name The name of the event
+   * @property {Date} scheduledStartTime The time to schedule the event at
+   * @property {PrivacyLevel|number} privacyLevel The privacy level of the event
+   * @property {GuildEventEntityType|number} entityType The scheduled entity type of the event
+   * @property {string} [description] The description of the event
+   * @property {StageChannel|Snowflake} [channel] The stage channel of the event
```
Jiralite, line 41: You could create a `StageChannelResolvable` `@typedef` for this, akin to others such as [`VoiceChannelResolvable`](https://discord.js.org/#/docs/main/main/typedef/VoiceChannelResolvable) and [TextChannelResolvable](https://discord.js.org/#/docs/main/main/typedef/TextChannelResolvable) etc.

iShibi: upstream changed these to just `channel id` now, so it won't be required here. This combination is referenced in a lot of places, so a new PR adding the typedef replacing all of them will be a better idea.

Jiralite: > This combination is referenced in a lot of places

I see that now! I have opened #6568 then.

src/structures/GuildEvent.js
```diff
@@ -0,0 +1,136 @@
...
+     * The time at which the guild event will start
+     * @type {Date}
+     */
+    this.scheduledStartTime = data.scheduled_start_time;
+
+    /**
+     * The time at which the guild event will end, or `null` if the event does not have a scheduled time to end
+     * @type {?Date}
+     */
+    this.scheduledEndTime = data.scheduled_end_time;
```
ImRodry, line 81: Same reason as above, should also create a timestamp getter imo
```suggestion
    this.scheduledEndAt = new Date(data.scheduled_end_time);
```

src/structures/GuildEvent.js
```diff
@@ -0,0 +1,136 @@
...
+     * The entity id
+     * @type {?Snowflake}
+     */
+    this.entityId = data.entity_id;
+
+    /**
+     * The metadata for the guild event
+     * @type {GuildEventEntityMetadata}
+     */
+    this.entityMetadata = new GuildEventEntityMetadata(data.entity_metadata);
```
ImRodry, line 111: Just like threads, we probably don't need a class for this and the properties from the metadata can be added directly on the GuildEvent structure

src/util/Constants.js
```diff
@@ -1031,6 +1046,26 @@ exports.PrivacyLevels = createEnum([null, 'PUBLIC', 'GUILD_ONLY']);
  */
 exports.PremiumTiers = createEnum(['NONE', 'TIER_1', 'TIER_2', 'TIER_3']);
 
+/**
+ * The status of a {@link GuildEvent}:
+ * * SCHEDULED
+ * * ACTIVE
+ * * COMPLETED
+ * * CANCELED
+ * @typedef {string} GuildEventStatus
```
ImRodry, line 1055: Should add a `@see` link (note for the future once upstream PR is merged)

src/util/Constants.js
```diff
@@ -1031,6 +1046,26 @@ exports.PrivacyLevels = createEnum([null, 'PUBLIC', 'GUILD_ONLY']);
...
+ */
+exports.GuildEventStatuses = createEnum([null, 'SCHEDULED', 'ACTIVE', 'COMPLETED', 'CANCELED']);
+
+/**
+ * The entity type of a {@link GuildEvent}:
+ * * NONE
+ * * STAGE_INSTANCE
+ * * VOICE
+ * * LOCATION
+ * @typedef {string} GuildEventEntityType
```
ImRodry, line 1065: Should add a `@see` link (note for the future once upstream PR is merged)

ImRodry (changes requested):
Some things I noticed after going through this PR again
I also think you should change Invite#INVITES_PATTERN to `/discord(?:(?:app)?\.com\/invite|\.gg(?:\/invite)?)\/([\w-]{2,255})$/gi` (added the $ at the end to match the end of the string) and add a new property under GuildEvent called EVENTS_PATTERN with the following regex: `/discord(?:(?:app)?\.com\/invite|\.gg(?:\/invite)?)\/[\w-]{2,255}\?event=([0-9]{17,19})/gi`

src/structures/GuildEvent.js
```diff
@@ -0,0 +1,136 @@
...
+       * @type {string}
+       */
+      this.image = data.image;
+    }
+
+    /**
+     * The time at which the guild event will start
+     * @type {Date}
+     */
+    this.scheduledStartTime = data.scheduled_start_time;
```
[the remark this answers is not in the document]
iShibi: The property you are referring to is creation time constructed using id which I *definitely didn't forget* to implement 🤞. About these start and end fields, I believe they should be kept as it is for keeping consistency with API naming. I could add a `scheduledStartTimeAt` getter to return the `Date` object and do the same for the end time.

iShibi:
This is ready for review

src/client/actions/GuildScheduledEventUserRemove.js
```diff
@@ -0,0 +1,30 @@
...
+      const guildScheduledEvent = guild.scheduledEvents.cache.get(data.guild_scheduled_event_id);
+      const user = client.users.cache.get(data.user_id);
+
+      /**
+       * Emitted whenever a user unsubscribes from a guild scheduled event
+       * @event Client#guildScheduledEventUserRemove
+       * @param {GuildScheduledEvent} guildScheduledEvent The guild scheduled event
+       * @param {User} user user who unsubscribed
+       */
+      client.emit(Events.GUILD_SCHEDULED_EVENT_USER_ADD, guildScheduledEvent, user);
```
ImRodry, line 21: Typo in the event name ?

iShibi: can't find it, what is the typo?

ImRodry: ADD instead of REMOVE

iShibi: ah, I see. Thanks for noticing.

src/structures/GuildScheduledEvent.js
```diff
@@ -0,0 +1,300 @@
...
+      this.userCount = data.user_count;
+    }
+
+    if ('creator' in data) {
+      /**
+       * The user that created this guild scheduled event
+       * @type {User}
+       */
+      this.creator = this.client.users._add(data.creator);
+    }
```
Jiralite, line 124: This is documented & typed as always available but is not sent when fetching an invite with an event id rendering this property `undefined`. It's also [documented at Discord](https://discord.com/developers/docs/resources/guild-scheduled-event#guild-scheduled-event-object-guild-scheduled-event-structure) as such. This should be nullable.

Jiralite (changes requested):

src/client/actions/GuildScheduledEventUserRemove.js
```diff
@@ -0,0 +1,30 @@
...
+
+    if (guild) {
+      const guildScheduledEvent = guild.scheduledEvents.cache.get(data.guild_scheduled_event_id);
+      const user = client.users.cache.get(data.user_id);
+
+      /**
+       * Emitted whenever a user unsubscribes from a guild scheduled event
+       * @event Client#guildScheduledEventUserRemove
+       * @param {GuildScheduledEvent} guildScheduledEvent The guild scheduled event
+       * @param {User} user user who unsubscribed
```
Jiralite, line 19:
```suggestion
       * @param {User} user The user who unsubscribed
```

src/managers/GuildScheduledEventManager.js
```diff
@@ -0,0 +1,226 @@
...
+    // TODO: find a way to return collection of members. (Edge case??): some users can be non-members
+    // Can't test it rn because setting privacy level to 'PUBLIC' returns error
+    // if (options?.withMember) {
+    //
+    // }
+
+    return data.users.reduce(
+      (coll, rawUserData) => coll.set(rawUserData.id, this.client.users._add(rawUserData)),
+      new Collection(),
+    );
```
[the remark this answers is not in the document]
iShibi: Used to be the same thing with stage instances, I assume bots can't still make them public(?).
Also, there isn't any way to set `PUBLIC` events in the client yet is there?

src/structures/GuildScheduledEvent.js
iShibi: I believe abaa35ad3af6a4e2ce1c2e95fa8765783b89930f fixed this and all the others like it

ImRodry:
In what case can those properties be null?

src/managers/GuildScheduledEventManager.js
```diff
@@ -0,0 +1,254 @@
...
+   * * A GuildScheduledEvent object
+   * @typedef {Snowflake|GuildScheduledEvent} GuildScheduledEventResolvable
+   */
+
+  /**
+   * Options used to create a guild scheduled event.
+   * @typedef {Object} GuildScheduledEventCreateOptions
+   * @property {string} name The name of the guild scheduled event
+   * @property {DateResolvable} scheduledStartTime The time to schedule the event at
+   * @property {DateResolvable} [scheduledEndTime] The time to end the event at
```
Jiralite, line 42: Perhaps note that this property is only required for external events?

src/managers/GuildScheduledEventManager.js
```diff
@@ -0,0 +1,254 @@
...
+   * @typedef {Object} FetchGuildScheduledEventSubscribersOptions
+   * @property {number} [limit] The maximum numbers of users to fetch
+   * @property {boolean} [withMember] Whether to fetch guild member data of the users
+   */
+
+  /**
+   * Fetches subscribers of a guild scheduled event.
+   * @param {GuildScheduledEventResolvable} guildScheduledEvent The guild scheduled event to fetch subscribers of
+   * @param {FetchGuildScheduledEventSubscribersOptions} [options] Options for fetching the subscribers
+   * @returns {Promise<Collection<Snowflake, User> | Collection<Snowflkae, GuildMember>>}
```
Jiralite, line 231:
```suggestion
   * @returns {Promise<Collection<Snowflake, User> | Collection<Snowflake, GuildMember>>}
```

src/structures/GuildScheduledEvent.js
```diff
@@ -0,0 +1,310 @@
...
+/**
+ * Represents a scheduled event in a {@link Guild}.
+ * @extends {Base}
+ */
+class GuildScheduledEvent extends Base {
+  /**
+   * @param {Client} client The instantiating client
+   * @param {APIGuildScheduledEvent} data The data for the guild scheduled event
+   */
+  constructor(client, data) {
```
Jiralite, line 20: I believe this should be undocumented per https://github.com/discordjs/discord.js/pull/6732.

src/managers/GuildScheduledEventManager.js
```diff
@@ -0,0 +1,254 @@
...
+      description,
+      scheduledEndTime,
+    } = options;
+
+    if (typeof privacyLevel === 'string') privacyLevel = PrivacyLevels[privacyLevel];
+    if (typeof entityType === 'string') entityType = GuildScheduledEventEntityTypes[entityType];
+    if (typeof status === 'string') status = GuildScheduledEventStatuses[status];
+    const channelId = this.guild.channels.resolveId(channel);
+
+    speakers = speakers?.map(user => this.client.users.resolveId(user));
```
almeidx, line 186:
```suggestion
    speakers &&= speakers.map(user => this.client.users.resolveId(user));
```

src/managers/GuildScheduledEventManager.js
```diff
@@ -0,0 +1,254 @@
...
+      scheduledStartTime,
+      description,
+      scheduledEndTime,
+    } = options;
+
+    if (typeof privacyLevel === 'string') privacyLevel = PrivacyLevels[privacyLevel];
+    if (typeof entityType === 'string') entityType = GuildScheduledEventEntityTypes[entityType];
+    const channelId = this.guild.channels.resolveId(channel);
+
+    speakers = speakers?.map(user => this.client.users.resolveId(user));
```
almeidx, line 74:
```suggestion
    speakers &&= speakers.map(user => this.client.users.resolveId(user));
```

almeidx (changes requested):

Jiralite:
We have another pull request: https://github.com/discord/discord-api-docs/pull/4113

src/structures/GuildScheduledEvent.js
```diff
@@ -0,0 +1,310 @@
...
+
+    if ('creator_id' in data) {
+      /**
+       * The id of the user that created this guild scheduled event
+       * @type {?Snowflake}
+       */
+      this.creatorId = data.creator_id;
+    } else {
+      this.creatorId ??= null;
+    }
```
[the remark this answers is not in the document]
iShibi: 👍

src/managers/GuildScheduledEventManager.js
iShibi: upstream removed it

src/managers/GuildScheduledEventManager.js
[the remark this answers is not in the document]
iShibi: This has resolved itself

iShibi:
I guess this is it, *whew*

waiting on https://github.com/discordjs/discord-api-types/pull/186 to merge. In the meantime:
1) Let me know if I forgot to add a shortcut method that can be useful.
2) What to do if `scheduledEvent` and/or `user` is not cached for `Guild_Scheduled_Event_User_*` events. **These are experimental gateway events, I hope they replace the ids in the payload with their respective objects.**
3) TEST THIS PR☠️

typings/index.d.ts
```diff
@@ -495,6 +499,10 @@ export abstract class Channel extends Base {
 
 export type If<T extends boolean, A, B = null> = T extends true ? A : T extends false ? B : A | B;
 
+export interface ClientFetchInviteOptions {
```
ImRodry, line 502: Shouldn’t this be at the very bottom in the mixins region?

iShibi: yes, but lazy me just left it there 😴

ImRodry:
The new classes should probably also be exported in index right

Jiralite:
New event-related error codes added:

- https://github.com/discord/discord-api-docs/pull/4117

src/client/actions/GuildScheduledEventUpdate.js
```diff
@@ -0,0 +1,30 @@
...
+const Action = require('./Action');
+const { Events } = require('../../util/Constants');
+
+class GuildScheduledEventUpdateAction extends Action {
+  handle(data) {
+    const client = this.client;
+    const guild = client.guilds.cache.get(data.guild_id);
+
+    if (guild) {
+      const oldGuildScheduledEvent = guild.scheduledEvents.cache.get(data.id)._clone() ?? null;
```
GoldenAngel2, line 12:
```suggestion
      const oldGuildScheduledEvent = guild.scheduledEvents.cache.get(data.id)?._clone() ?? null;
```
If the event isn't cached it will cause an error on this line.

iShibi: can't find that end part in the documentation

GoldenAngel2: > can't find that end part in the documentation

@iShibi It seems like it's only sent for the `GUILD_CREATE` event
from startup

iShibi: If there isn't already a PR for this can you open one at https://github.com/discord/discord-api-docs

marceaupatu123:
That looks amazing thanks a lot !

iCrawl:
This needs a rebase.

src/managers/GuildScheduledEventManager.js
```diff
@@ -0,0 +1,247 @@
...
+        channel_id: channelId,
+        name,
+        privacy_level: privacyLevel,
+        scheduled_start_time: new Date(scheduledStartTime).toISOString(),
+        scheduled_end_time: scheduledEndTime ? new Date(scheduledEndTime).toISOString() : undefined,
+        description,
+        entity_type: entityType,
+        entity_metadata: {
+          location,
+        },
```
vladfrangu, line 76: Do we want to abstract it like this? CC @discordjs/the-big-4

src/managers/GuildScheduledEventManager.js
```diff
@@ -0,0 +1,247 @@
...
+    if (typeof status === 'string') status = GuildScheduledEventStatuses[status];
+    const channelId = this.guild.channels.resolveId(channel);
+
+    const data = await this.client.api.guilds(this.guild.id, 'scheduled-events', guildScheduledEventId).patch({
+      data: {
+        channel_id: channelId ?? undefined,
+        name,
+        privacy_level: privacyLevel,
+        scheduled_start_time: scheduledStartTime ? new Date(scheduledStartTime).toISOString() : undefined,
+        scheduled_end_time: scheduledEndTime ? new Date(scheduledEndTime).toISOString() : undefined,
```
vladfrangu, line 181: Can this be set to null?

src/structures/GuildScheduledEvent.js
```diff
@@ -0,0 +1,311 @@
...
+      /**
+       * The user that created this guild scheduled event
+       * @type {?User}
+       */
+      this.creator = this.client.users._add(data.creator);
+    } else {
+      this.creator ??= this.client.users.resolve(this.creatorId);
+    }
+
+    if (data.entity_metadata) {
```
vladfrangu, line 126: Ditto

src/structures/GuildScheduledEvent.js
```diff
@@ -0,0 +1,311 @@
...
+   * Sets a new time to schedule the event at.
+   * @param {DateResolvable} scheduledStartTime The time to schedule the event at
+   * @returns {Promise<GuildScheduledEvent>}
+   * @example
+   * // Set start time of a guild scheduled event
+   * guildScheduledEvent.setScheduledStartTime('September 24, 2022'))
+   *  .then(guildScheduledEvent => console.log(`Set the start time to: ${guildScheduledEvent.scheduledStartTime}`))
+   *  .catch(console.error);
+   */
+  setScheduledStartTime(scheduledStartTime) {
```
vladfrangu, line 250: Where's endTime, privacyLevel, etc?

vladfrangu (changes requested):

src/structures/GuildScheduledEvent.js
iShibi: It can have only one value which is `GUILD_ONLY`. So, I omitted the shortcut method for it.

src/client/actions/GuildScheduledEventUserAdd.js
```diff
@@ -0,0 +1,30 @@
...
+const { Events } = require('../../util/Constants');
+
+class GuildScheduledEventUserAddAction extends Action {
+  handle(data) {
+    const client = this.client;
+    const guild = client.guilds.cache.get(data.guild_id);
+
+    if (guild) {
+      const guildScheduledEvent = guild.scheduledEvents.cache.get(data.guild_scheduled_event_id);
+      const user = client.users.cache.get(data.user_id);
```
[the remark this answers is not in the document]
iShibi: yeah, was thinking about the same thing. These two are experimental events, so we can either not support them right now or only emit them when both `GuildScheduledEvent` and `User` are cached.

This could be fixed if Discord returns full objects instead of just ids for them.

src/managers/GuildScheduledEventManager.js
kyranet: Probably not? I think we should stick to being close with Discord.

We could also make a builder for this, and since builders need to be very close to Discord's types, it can be very odd if we mix the two, as well as making both interfaces incompatible.

src/client/actions/GuildScheduledEventUserRemove.js
```diff
@@ -0,0 +1,30 @@
...
+const { Events } = require('../../util/Constants');
+
+class GuildScheduledEventUserRemoveAction extends Action {
+  handle(data) {
+    const client = this.client;
+    const guild = client.guilds.cache.get(data.guild_id);
+
+    if (guild) {
+      const guildScheduledEvent = guild.scheduledEvents.cache.get(data.guild_scheduled_event_id);
+      const user = client.users.cache.get(data.user_id);
```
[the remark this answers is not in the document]
kyranet: Aren't we able to use `Action#getUser` here? At worst we'll have a partial User, at best we'll have a full object.

src/client/actions/GuildScheduledEventUpdate.js
GoldenAngel2: Update: https://github.com/discord/discord-api-docs/pull/4139 got merged, so the field can be added now.

src/client/actions/GuildScheduledEventUserRemove.js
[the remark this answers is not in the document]
iShibi: added `GUILD_SCHEDULED_EVENT` partial to deal with the issue of stuff being uncached for these two events

src/managers/GuildScheduledEventManager.js
```diff
@@ -68,7 +68,11 @@ class GuildScheduledEventManager extends CachedManager {
         name,
         privacy_level: privacyLevel,
         scheduled_start_time: new Date(scheduledStartTime).toISOString(),
-        scheduled_end_time: scheduledEndTime ? new Date(scheduledEndTime).toISOString() : undefined,
+        scheduled_end_time: scheduledEndTime
+          ? new Date(scheduledEndTime).toISOString()
+          : scheduledEndTime === null
+          ? scheduledEndTime
+          : undefined,
```
vladfrangu, line 75: any reason you don't just.. set it to scheduledEndTime instead of doing these checks? Everything truthy will go down the new Date path, might as well simplify it

src/managers/GuildScheduledEventManager.js
```diff
@@ -178,7 +182,11 @@ class GuildScheduledEventManager extends CachedManager {
         name,
         privacy_level: privacyLevel,
         scheduled_start_time: scheduledStartTime ? new Date(scheduledStartTime).toISOString() : undefined,
-        scheduled_end_time: scheduledEndTime ? new Date(scheduledEndTime).toISOString() : undefined,
+        scheduled_end_time: scheduledEndTime
+          ? new Date(scheduledEndTime).toISOString()
+          : scheduledEndTime === null
+          ? scheduledEndTime
+          : undefined,
```
vladfrangu, line 189: Ditto

vladfrangu (dismissed):

src/managers/GuildScheduledEventManager.js
iShibi: I don't know why I did it like that ^._.^

genius

src/managers/GuildScheduledEventManager.js
vladfrangu: @iCrawl thoughts?

iCrawl: I don't think we need to.

ImRodry:
Just out of curiosity: why are we keeping the entityMetadata object in here if it was removed in threads and all of its properties placed on the class itself?

iShibi:
The metadata was merged into the scheduled event class after you pointed it out. The latest commit just makes the structure of create/edit options objects similar to how Discord takes it. See: `MessageOptions#allowedMentions`.

ImRodry:
It was vlad who pointed it out and kyra said this https://github.com/discordjs/discord.js/pull/6493#discussion_r756086609 but I don’t understand this being brought up here when it wasn’t brought up in threads. Seems kinda inconsistent to me and tbh I’d prefer removing the metadata object

src/managers/GuildScheduledEventManager.js
```diff
@@ -72,7 +78,7 @@ class GuildScheduledEventManager extends CachedManager {
         description,
         entity_type: entityType,
         entity_metadata: {
```
vladfrangu, line 80: Just pass in options.entityMetadata instead

vladfrangu (dismissed):

src/managers/GuildScheduledEventManager.js
ImRodry: This will have to be changed in the future in case other properties in snake case are added inside the object, I think we should just put them all in one object instead of having this metadata object which seems to just be an api thing with no real use

vladfrangu: Sure, let's keep the object as is and manually provide variables but we don't know what may come to the entityMetadata field in the future, and we try to stay as close to the API as possible

ImRodry: I know we try to stay close to the API but this wasn’t done with threads simply because it causes unnecessary burden and doesn’t bring any benefits to the users (at least that’s how I see it, if there was a better reason there please let me know) so I don’t see why we can’t do the same here

src/structures/GuildScheduledEvent.js
```diff
@@ -0,0 +1,325 @@
...
+     * @type {number}
+     */
+    this.scheduledStartTime = data.scheduled_start_time;
+
+    /**
+     * The timestamp the guild scheduled event will end at,
+     * or `null` if the event does not have a scheduled time to end
+     * @type {?number}
+     */
+    this.scheduledEndTime = data.scheduled_end_time;
```
ImRodry, line 80: These are ISO strings, not numbers. Can you convert these to timestamps in ms?

src/structures/GuildScheduledEvent.js
```diff
@@ -0,0 +1,325 @@
...
+     * The channel id in which the scheduled event will be hosted, or `null` if entity type is `EXTERNAL`
+     * @type {?Snowflake}
+     */
+    this.channelId = data.channel_id;
+
+    /**
+     * The id of the user that created this guild scheduled event
+     * @type {?Snowflake}
+     */
+    this.creatorId = data.creator_id;
```
ImRodry, line 51: Shouldn't these have an `in` check?

src/structures/GuildScheduledEvent.js
```diff
@@ -0,0 +1,325 @@
...
+     * The type of hosting entity associated with the scheduled event
+     * @type {GuildScheduledEventEntityType}
+     */
+    this.entityType = GuildScheduledEventEntityTypes[data.entity_type];
+
+    /**
+     * The id of the hosting entity associated with the scheduled event
+     * @type {?Snowflake}
+     */
+    this.entityId = data.entity_id;
```
ImRodry, line 104: ditto for the `in` check

src/structures/GuildScheduledEvent.js
```diff
@@ -0,0 +1,325 @@
...
+  }
+
+  /**
+   * Fetches subscribers of this guild scheduled event.
+   * @param {FetchGuildScheduledEventSubscribersOptions} [options] Options for fetching the subscribers
+   * @returns {Promise<Collection<Snowflake, User> | Collection<Snowflake, GuildMember>>}
+   */
+  fetchSubscribers(options) {
+    return this.guild.scheduledEvents.fetchSubscribers(this.id, options);
+  }
```
ImRodry, line 317: Seeing as there is a property called userCount, shouldn't this be fetchUsers, or possibly even moved into a GuildScheduledEventUserManager? There are events for these so we could cache them pretty easily

ImRodry (changes requested):

src/structures/GuildScheduledEvent.js
```diff
@@ -0,0 +1,325 @@
...
+   * @param {GuildScheduledEventStatus|number} status The status of the guild scheduled event
+   * @returns {Promise<GuildScheduledEvent>}
+   * @example
+   * // Set status of a guild scheduled event
+   * guildScheduledEvent.setStatus('ACTIVE')
+   *  .then(guildScheduledEvent => console.log(`Set the status to: ${guildScheduledEvent.status}`))
+   *  .catch(console.error);
+   */
+  setStatus(status) {
+    return this.edit({ status });
```
ImRodry, line 293: Is there any way we could check for this in the typings at least?
image
(taken from the Discord docs, cc @suneettipirneni)

src/structures/GuildScheduledEvent.js
iShibi: good catch

src/structures/GuildScheduledEvent.js
suneettipirneni: If there's a way to know the status of an event compile-time, then it is possible.

ImRodry (commented):
Can you add a GuildScheduledEvent#url getter? It should look like this:
`https://discord.com/events/${this.guildId}/${this.id}` (perhaps even add it to the CDN endpoints)

src/structures/GuildScheduledEvent.js
ImRodry: Well the status property is of type GuildScheduledEventStatus which is one of the following:
image
so I guess it would be possible? Just no clue how

suneettipirneni: I'm saying is there a way to know *which* event status the event has before runtime?

ImRodry: Probably with a series of `extends` checks on the status property? Should be something similar to what was done with component collector types right?

suneettipirneni:
But you provide the type in the collector components, here you’re not explicitly providing the status type

Thanks, Suneet Tipirneni

src/structures/GuildScheduledEvent.js
ImRodry: Moving the convo back here, if you provide a type parameter in the class (`GuildScheduledEvent<Status extends GuildScheduledEventStatus = GuildScheduledEventStatus>`) you can then use that parameter and check if it extends each one of the status types. I think if the type parameter is used as a property it can be inferred, but I'm not sure

src/structures/GuildScheduledEvent.js
```diff
@@ -0,0 +1,358 @@
...
+  get createdTimestamp() {
+    return SnowflakeUtil.deconstruct(this.id).timestamp;
+  }
+
+  /**
+   * The time the guild scheduled event was created at
+   * @type {Date}
+   * @readonly
+   */
+  get cretedAt() {
```
Underpants-dev, line 168: There is a typo in this line. It should have been `createdAt()` instead of `cretedAt()`.

Underpants-dev (changes requested):

src/structures/GuildScheduledEvent.js
```diff
@@ -327,6 +337,17 @@ class GuildScheduledEvent extends Base {
...
+
+  /**
+   * When concatenated with a string, this automatically concatenates the event's URL instead of the object.
+   * @returns {string}
+   * @example
+   * // Logs: Event: https://discord.com/events/412345678901234567/499876543211234567
+   * console.log(`Event: ${guildScheduledEvent}`);
+   */
+  toString() {
+    return this.url;
```
ImRodry, line 349: These urls don’t get embed when sent so it would probably be better to try and find an invite from the guild’s invite cache to the channel the event is in (if any) and add the event id. If no invites are found then you can default to this URL

James-Firth: imho there should be a function to grab both URLs and I would say using `this.url` as the default `toString` allows someone using this library to minimize the permissions of their bot.

If it uses the invite link then the bot _must_ have permissions to create invites.

I'm also optimistically hoping discord will add an embed to this style URL, as the user experience isn't great as-is.

ImRodry: Discord cannot add an embed to this url as this doesn’t grant you access to the server the event is on. The toString method also cannot create invites as it cannot be asynchronous so the only possibility would be for it to get from cache or default to the regular url

kyranet: That hybrid behaviour would be a mess anyways. If anything, it should be `fetchEmbed()` if we even want to abstract it.

ImRodry: That’d be misleading since it gives you an invite and not an embed object. I think a method called share or something similar would work better and this would create an invite or fetch an existing one

iShibi: tell me why the current behaviour isn't enough. The method does exactly what it should, not getting embedded is on Discord.

src/structures/GuildScheduledEvent.js
iShibi: 1) The method name is self-descriptive currently. A subscriber can be a user as well as a member, so it makes sense to keep it as it is.
2) Those two events are experimental currently, so maybe in some other PR??

ImRodry: How are the events experimental if they’re included in this PR

suneettipirneni:
Just a heads up dapi-types have been bumped, so that should fix your TS build issues

src/structures/GuildScheduledEvent.js
iShibi: Only added the basic support for them, adding more features around experimental events doesn't seem right.

src/structures/GuildScheduledEvent.js
```diff
@@ -0,0 +1,390 @@
...
+   * guildScheduledEvent.setDescription('A virtual birthday party')
+   *  .then(guildScheduledEvent => console.log(`Set the description to: ${guildScheduledEvent.description}`))
+   *  .catch(console.error);
+   */
+  setDescription(description) {
+    return this.edit({ description, entityMetadata: { location: this.location } });
+  }
+
+  /**
+   * Sets the new status of the guild scheduled event.
```
ImRodry, line 305: I would recommend adding an info tag of sorts saying that if users are on TypeScript they should use the typeguards in order to make sure they don't get errors when setting the status. That info tag could be here and on #edit, or only on the typeguards, your call

src/structures/GuildAuditLogs.js
```diff
@@ -575,6 +588,19 @@ class GuildAuditLogsEntry {
...
+      this.target =
+        guild.scheduledEvents.cache.get(data.target_id) ??
+        new GuildScheduledEvent(
+          guild.client,
+          this.changes.reduce(
+            (o, c) => {
+              o[c.key] = c.new ?? c.old;
+              return o;
+            },
+            { id: data.target_id },
```
Jiralite, line 601: A removed guild scheduled event doesn't have the guild id in `this.changes`. This will result in the guild id being `undefined`.

```suggestion
            {
              id: data.target_id,
              guild_id: guild.id
            },
```

src/structures/GuildScheduledEvent.js
```diff
@@ -0,0 +1,390 @@
...
+      this.description = data.description;
+    } else {
+      this.description ??= null;
+    }
+
+    /**
+     * The timestamp the guild scheduled event will start at
+     * @type {number}
+     */
+    this.scheduledStartTime = new Date(data.scheduled_start_time).getTime();
```
Jiralite, line 82: This is not sent when fetching a removed guild scheduled event from the audit logs and will return `NaN`, so I would suggest making this nullable just like `scheduledEndTime`.

I noticed this is documented as not nullable in Discord's API documentation, so it's probably not documented correctly there too?

Jiralite (commented):
I've done some more messing around on this branch!

src/structures/GuildScheduledEvent.js
```diff
@@ -0,0 +1,390 @@
...
+     * The id of the guild this guild scheduled event belongs to
+     * @type {Snowflake}
+     */
+    this.guildId = data.guild_id;
+
+    /**
+     * Whether the guild scheduled event has been deleted
+     * @type {boolean}
+     */
+    this.deleted = false;
```
Jiralite, line 36: Well, discord.js constructing a deleted guild scheduled event doesn't show it's deleted here because this boolean is never actually modified. Interesting!

iShibi: It gets modified in the action handler:
https://github.com/discordjs/discord.js/blob/611fd63845d8901f6a0db2bf789088057463e3f0/src/client/actions/GuildScheduledEventDelete.js#L15

src/structures/GuildScheduledEvent.js
iShibi: What's the status on this? I started with a separate structure, so should I revert back to that or keep it just like threads?

src/managers/GuildScheduledEventManager.js
iShibi: Resolving this because this has been dealt with. A similar conversation is open above for further discussion.

src/structures/GuildScheduledEvent.js
```diff
@@ -303,6 +303,7 @@ class GuildScheduledEvent extends Base {
 
   /**
    * Sets the new status of the guild scheduled event.
+   * <info>Use this method in conjunction with status type-guards like {@link GuildScheduledEvent.isScheduled}</info>
```
ImRodry, line 306: Make sure this says "if you're using TypeScript", otherwise we'll get people confused on why they need to use this

src/structures/GuildScheduledEvent.js
Jiralite: Not when fetched from the audit logs.

src/structures/GuildScheduledEvent.js
```diff
@@ -0,0 +1,386 @@
...
+   *  .catch(console.error);
+   */
+  setLocation(location) {
+    return this.edit({ entityMetadata: { location } });
+  }
+
+  /**
+   * Fetches subscribers of this guild scheduled event.
+   * @param {FetchGuildScheduledEventSubscribersOptions} [options] Options for fetching the subscribers
+   * @returns {Promise<Collection<Snowflake, User> | Collection<Snowflake, GuildMember>>}
```
Jiralite, line 336:
```suggestion
   * @returns {Promise<Collection<Snowflake, GuildScheduledEventUser>>}
```

Jiralite (changes requested):

src/structures/GuildScheduledEvent.js
Jiralite: `{@link GuildScheduledEvent.isScheduled}` should be `{@link GuildScheduledEvent#isScheduled}`.

src/structures/GuildScheduledEvent.js
iShibi: fixed

Jiralite:
Might need to remove .vscode/settings.json?

src/structures/GuildScheduledEvent.js
```diff
@@ -0,0 +1,387 @@
...
+    return this.edit({ name, entityMetadata: { location: this.location } });
+  }
+
+  /**
+   * Sets a new time to schedule the event at.
+   * @param {DateResolvable} scheduledStartTime The time to schedule the event at
+   * @returns {Promise<GuildScheduledEvent>}
+   * @example
+   * // Set start time of a guild scheduled event
+   * guildScheduledEvent.setScheduledStartTime('September 24, 2022'))
```
Jiralite, line 268: Syntax error here!

src/structures/GuildScheduledEvent.js
```diff
@@ -0,0 +1,387 @@
...
+    return this.edit({ scheduledStartTime, entityMetadata: { location: this.location } });
+  }
+
+  /**
+   * Sets a new time to end the event at.
+   * @param {?DateResolvable} scheduledEndTime The time to end the event at
+   * @returns {Promise<GuildScheduledEvent>}
+   * @example
+   * // Set end time of a guild scheduled event
+   * guildScheduledEvent.setScheduledEndTime('September 25, 2022'))
```
Jiralite, line 282: Syntax error here!

Jiralite (changes requested):

iShibi:
> Might need to remove .vscode/settings.json?

:blameCrawl:

src/structures/GuildScheduledEvent.js
```diff
@@ -216,6 +217,17 @@ class GuildScheduledEvent extends Base {
...
   }
 
+  /**
+   * Creates an invite url to this guild scheduled event.
+   * @param {InviteResolvable} invite The invite to the guild channel of this guild scheduled event
+   * <info>For an `EXTERNAL` guild scheduled event, this can be an invite to any guild channel</info>
+   * @returns {string}
+   */
+  createInviteUrl(invite) {
+    const code = DataResolver.resolveInviteCode(invite);
```
ImRodry, line 227: This should be generating a new invite and take inviteoptions as an argument imo, or maybe have an option to try to get from cache and only create if no invite to the channel is found

src/structures/GuildScheduledEvent.js
```diff
@@ -0,0 +1,400 @@
...
+   * The URL to the guild scheduled event
+   * @type {string}
+   * @readonly
+   */
+  get url() {
+    return Endpoints.scheduledEvent(this.client.options.http.scheduledEvent, this.guildId, this.id);
+  }
+
+  /**
+   * Creates an invite url to this guild scheduled event.
```
Jiralite, line 221:
```suggestion
   * Creates an invite URL to this guild scheduled event.
```

src/structures/GuildScheduledEvent.js
```diff
@@ -218,14 +218,26 @@ class GuildScheduledEvent extends Base {
...
+   */
+  async createInviteUrl(options) {
+    let channelId = this.channelId;
+    if (this.entityType === 'EXTERNAL') {
+      if (!options?.channel) throw new Error('INVITE_OPTIONS_MISSING_CHANNEL');
+      channelId = this.guild.channels.resolveId(options.channel);
+      if (!channelId) throw new Error('GUILD_CHANNEL_RESOLVE');
+    }
+    const invite = await this.guild.invites.create(channelId, options);
+    return `${this.client.options.http.invite}/${invite.code}?event=${this.id}`;
```
ImRodry, line 240: I recommend using Endpoints.invite here and adding an option to that method which is the eventId (in an object)

iShibi: but why

ImRodry: consistency and maintainability?

src/structures/GuildScheduledEvent.js
iShibi: @vladfrangu this is the only thing that's left (i guess)

henriqueleite42 (approved):

kyranet:
You have merge conflicts

typings/index.d.ts
```diff
@@ -1103,6 +1109,50 @@ export class GuildPreview extends Base {
...
+  public userCount: number | null;
+  public creator: User | null;
+  public readonly createdTimestamp: number;
+  public readonly createdAt: Date;
+  public readonly scheduledStartAt: Date;
+  public readonly scheduledEndAt: Date | null;
+  public readonly channel: VoiceChannel | StageChannel | null;
+  public readonly guild: Guild | null;
+  public readonly url: string;
+  public createInviteUrl(options?: CreateGuildScheduledEventInviteUrlOptions): Promise<string>;
```
Jiralite, line 1137:
```suggestion
  public createInviteURL(options?: CreateGuildScheduledEventInviteURLOptions): Promise<string>;
```

typings/index.d.ts
```diff
@@ -4015,6 +4110,10 @@ export interface ConstantsStatus {
   DISCONNECTED: 5;
 }
 
+export interface CreateGuildScheduledEventInviteUrlOptions extends CreateInviteOptions {
```
Jiralite, line 4113:
```suggestion
export interface CreateGuildScheduledEventInviteURLOptions extends CreateInviteOptions {
```

src/structures/GuildScheduledEvent.js
```diff
@@ -0,0 +1,412 @@
...
+   * @type {string}
+   * @readonly
+   */
+  get url() {
+    return Endpoints.scheduledEvent(this.client.options.http.scheduledEvent, this.guildId, this.id);
+  }
+
+  /**
+   * Options used to create an invite URL to a {@link GuildScheduledEvent}
+   * @typedef {CreateInviteOptions} CreateGuildScheduledEventInviteUrlOptions
```
Jiralite, line 222:
```suggestion
   * @typedef {CreateInviteOptions} CreateGuildScheduledEventInviteURLOptions
```

src/structures/GuildScheduledEvent.js
```diff
@@ -0,0 +1,412 @@
...
+   * @property {GuildInvitableChannelResolvable} [channel] The channel to create the invite in.
+   * <warn>This is required when the `entityType` of `GuildScheduledEvent` is `EXTERNAL`, gets ignored otherwise</warn>
+   */
+
+  /**
+   * Creates an invite URL to this guild scheduled event.
+   * @param {CreateGuildScheduledEventInviteUrlOptions} [options] The options to create the invite
```
Jiralite, line 229:
```suggestion
   * @param {CreateGuildScheduledEventInviteURLOptions} [options] The options to create the invite
```

src/structures/GuildScheduledEvent.js
```diff
@@ -0,0 +1,412 @@
...
+   * @returns {Promise<string>}
+   */
+  async createInviteUrl(options) {
```
Jiralite, line 232:
```suggestion
  async createInviteURL(options) {
```

Jiralite (commented):
These should be stylised as *URL (see other methods and such) so otherwise, this introduces inconsistency:

src/structures/GuildScheduledEvent.js
kyranet: I still think that it should be `gse.metadata.location`.

src/structures/GuildScheduledEvent.js
```diff
@@ -0,0 +1,412 @@
...
+    }
+  }
+
+  /**
+   * The timestamp the guild scheduled event was created at
+   * @type {number}
+   * @readonly
+   */
+  get createdTimestamp() {
+    return SnowflakeUtil.deconstruct(this.id).timestamp;
```
kyranet, line 162: If you have rebased, you should be able to use [`SnowflakeUtil.timestampFrom`](https://discord.js.org/#/docs/main/main/class/SnowflakeUtil?scrollTo=s-timestampFrom) now:
```suggestion
    return SnowflakeUtil.timestampFrom(this.id);
```

kyranet (changes requested):

Jiralite:
Is it worth throwing when attempting to use a method that uses an API call if you've received a guild scheduled event object from a guild you're not in?

For example, if the client fetches an invite alongside a guild scheduled event, all the methods won't work (`delete()`, `fetchSubscribers()`, `edit()` etc.).

Jiralite:
Note: More audit log types for guild scheduled events need to be added per this pull request:

- https://github.com/discordjs/discord.js/pull/6647

iShibi:
> Note: More audit log types for guild scheduled events need to be added per this pull request:
> 
> * [types: improve audit logs' typings #6647](https://github.com/discordjs/discord.js/pull/6647)

this PR gonna outlive me 😩

kyranet:
Please rebase (again), there are merge conflicts :eyes:

Jiralite:
Seems as of this pull request, the once experimental events are no longer experimental:

- https://github.com/discord/discord-api-docs/pull/4216

Jiralite:
Another note: time to use the deleted getters & setters from b0937502d35a6519948b7320a0cc361acf1a1755!

iCrawl:
This needs a rebase 👍

src/managers/GuildScheduledEventManager.js
```diff
@@ -0,0 +1,261 @@
...
+      data: {
+        channel_id: typeof channel === 'undefined' ? channel : this.guild.channels.resolveId(channel),
+        name,
+        privacy_level: privacyLevel,
+        scheduled_start_time: scheduledStartTime ? new Date(scheduledStartTime).toISOString() : undefined,
+        scheduled_end_time: scheduledEndTime ? new Date(scheduledEndTime).toISOString() : scheduledEndTime,
+        description,
+        entity_type: entityType,
+        status,
+        entity_metadata: {
```
ImRodry, line 194: Entity metadata can only be sent if the event is external, otherwise it causes an error

src/client/actions/GuildScheduledEventDelete.js
```diff
@@ -0,0 +1,33 @@
...
+
+class GuildScheduledEventDeleteAction extends Action {
+  handle(data) {
+    const client = this.client;
+    const guild = client.guilds.cache.get(data.guild_id);
+
+    if (guild) {
+      const guildScheduledEvent = guild.scheduledEvents._add(data);
+      if (guildScheduledEvent) {
+        guild.scheduledEvents.cache.delete(guildScheduledEvent.id);
```
vladfrangu, line 15: Just wanted to say how ironic it is... we're adding what we're removing.. Might wanna not cache the result of _add if possible?

src/managers/GuildScheduledEventManager.js
vladfrangu: CC @iShibi, if that's so, we should probably only append entity_metadata if the user provides it

src/structures/GuildScheduledEvent.js
```diff
@@ -0,0 +1,440 @@
...
+  Endpoints,
+} = require('../util/Constants');
+const SnowflakeUtil = require('../util/SnowflakeUtil');
+
+/**
+ * @type {WeakSet<GuildScheduledEvent>}
+ * @private
+ * @internal
+ */
+const deletedGuildScheduledEvents = new WeakSet();
```
vladfrangu, line 18: Hmm, since we're planning on removing this and `Structure#deleted`, should we just not implement it here? @discordjs/the-big-4

src/structures/GuildScheduledEvent.js
```diff
@@ -0,0 +1,440 @@
...
+    } else {
+      this.description ??= null;
+    }
+
+    /**
+     * The timestamp the guild scheduled event will start at
+     * <info>This can be potentially `null` only when it's an {@link AuditLogEntryTarget}</info>
+     * @type {?number}
+     */
+    this.scheduledStartTime = data.scheduled_start_time ? new Date(data.scheduled_start_time).getTime() : null;
```
vladfrangu, line 85: Should we be consistent and suffix them in `Timestamp`?

vladfrangu (changes requested):

src/managers/GuildScheduledEventManager.js
ImRodry: Oh if you want the error:
```
entity_metadata: This type of event should not have entity metadata.
```

.vscode/settings.json
```diff
@@ -0,0 +1,28 @@
+{
```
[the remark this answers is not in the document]
iShibi: I'm sure I already removed this, it's not showing up in changes to me.

src/client/actions/GuildScheduledEventDelete.js
iShibi: I really can't recall why I used `_add`  here and not just get it from cache 😾. Replaced it with `getScheduledEvent` in b2ec54e instead. It can emit a partial too now.

src/structures/GuildScheduledEvent.js
iShibi: Makes sense but what about keeping consistency with API naming?

src/managers/GuildScheduledEventManager.js
iShibi: This seems to be happening with `VOICE` only, for `STAGE_INSTANCE` it just ignores it, which is how it should be I guess.

ImRodry: Yeah it did happen on a VOICE event for me, should still be fixed regardless

src/structures/GuildScheduledEvent.js
kyranet: I'm leaning towards not adding this at all, since it's going to be instantly deprecated. I'll be working on a PR in a bit to deprecate all `.deleted` later to prepare for the semver-major PR.

src/structures/GuildScheduledEvent.js
kyranet: `scheduledStartTimeTimestamp` storing a timestamp and `scheduledStartTime` being a getter that returns a `Date` from it makes sense to me.

iShibi: date getters have `At` at the end, we would be breaking one consistency to keep another.
<sub>ℹ️ Current names: `scheduledStartAt` & `scheduledEndAt`</sub>

kyranet: Or neither: `startsTimestamp` and `startsAt`? More consistent with what we have (similar to `{joined,created}{At,Timestamp}`), but differs a lot from Discord's naming?

src/managers/GuildScheduledEventManager.js
iShibi: This issue has been fixed

src/structures/GuildScheduledEvent.js
iShibi: If we are diverging from Discord's naming this time then I would suggest: `scheduledStartTimestamp` & keep the `scheduledStartAt`.

src/structures/GuildScheduledEvent.js
iShibi: aight, I'll remove these

Jiralite:
- https://github.com/discord/discord-api-docs/pull/4237

src/managers/GuildScheduledEventManager.js
```diff
@@ -0,0 +1,270 @@
...
+    if (typeof options !== 'object') throw new TypeError('INVALID_TYPE', 'options', 'object', true);
+    let { privacyLevel, entityType, channel, name, scheduledStartTime, description, scheduledEndTime, entityMetadata } =
+      options;
+
+    if (typeof privacyLevel === 'string') privacyLevel = PrivacyLevels[privacyLevel];
+    if (typeof entityType === 'string') entityType = GuildScheduledEventEntityTypes[entityType];
+
+    let entity_metadata, channel_id;
+    if (entityType === 3) {
+      channel_id = null;
```
vladfrangu, line 76: For safety, set it to undefined instead, unless you intentionally meant to set it to null

src/managers/GuildScheduledEventManager.js
```diff
@@ -0,0 +1,270 @@
...
+      entity_metadata = { location: entityMetadata?.location };
+    } else {
+      channel_id = this.guild.channels.resolveId(channel);
+      if (!channel_id) throw new Error('GUILD_VOICE_CHANNEL_RESOLVE');
+      entity_metadata = null;
```
vladfrangu, line 81: Ditto

src/managers/GuildScheduledEventManager.js
vladfrangu: Fixed how?

src/structures/GuildScheduledEvent.js
vladfrangu: I personally agree with @iShibi here

src/structures/GuildScheduledEvent.js
```diff
@@ -0,0 +1,416 @@
...
+   * @example
+   * // Delete a guild scheduled event
+   * guildScheduledEvent.delete()
+   *  .then(guildScheduledEvent => console.log(guildScheduledEvent))
+   *  .catch(console.error);
+   */
+  async delete() {
+    await this.guild.scheduledEvents.delete(this.id);
+    const clone = this._clone();
+    return clone;
```
vladfrangu, line 274: Can't you just return `this`?

src/structures/GuildScheduledEvent.js
```diff
@@ -0,0 +1,416 @@
...
+   * @param {string} name The new name of the guild scheduled event
+   * @returns {Promise<GuildScheduledEvent>}
+   * @example
+   * // Set name of a guild scheduled event
+   * guildScheduledEvent.setName('Birthday Party')
+   *  .then(guildScheduledEvent => console.log(`Set the name to: ${guildScheduledEvent.name}`))
+   *  .catch(console.error);
+   */
+  setName(name) {
+    return this.edit({ name, entityMetadata: { location: this.location } });
```
vladfrangu, line 288: this.location doesn't exist. Did you mean `entityMetadata: this.entityMetadata`?

src/structures/GuildScheduledEvent.js
```diff
@@ -0,0 +1,416 @@
...
+  }
+
+  /**
+   * Sets a new time to schedule the event at.
+   * @param {DateResolvable} scheduledStartTime The time to schedule the event at
+   * @returns {Promise<GuildScheduledEvent>}
+   * @example
+   * // Set start time of a guild scheduled event
+   * guildScheduledEvent.setScheduledStartTime('September 24, 2022')
```
vladfrangu, line 297: Maybe show an ISO string instead? Could confuse users? Not sure whats the ideal solution here...

src/structures/GuildScheduledEvent.js
```diff
@@ -0,0 +1,416 @@
...
+  }
+
+  /**
+   * Sets the new location of the guild scheduled event.
+   * @param {string} location The location of the guild scheduled event
+   * @returns {Promise<GuildScheduledEvent>}
+   * @example
+   * // Set location of a guild scheduled event
+   * guildScheduledEvent.setLocation('Earth')
+   *  .then(guildScheduledEvent => console.log(`Set the location to: ${guildScheduledEvent.location}`))
```
vladfrangu, line 356: .location doesn't exist

typings/index.d.ts
```diff
@@ -1119,6 +1124,49 @@ export class GuildPreview extends Base {
...
+  public readonly url: string;
+  public createInviteURL(options?: CreateGuildScheduledEventInviteURLOptions): Promise<string>;
+  public edit(options: GuildScheduledEventEditOptions): Promise<GuildScheduledEvent>;
+  public delete(): Promise<GuildScheduledEvent>;
+  public setName(name: string): Promise<GuildScheduledEvent>;
+  public setScheduledStartTime(scheduledStartTime: DateResolvable): Promise<GuildScheduledEvent>;
+  public setScheduledEndTime(scheduledEndTime: DateResolvable | null): Promise<GuildScheduledEvent>;
+  public setDescription(description: string): Promise<GuildScheduledEvent>;
+  public setStatus(status: GuildScheduledEventSetStatusArg<S>): Promise<GuildScheduledEvent>;
+  public setLocation(location: string): Promise<GuildScheduledEvent>;
```
vladfrangu, line 1159: You'll want to get the new status from edit/setStatus somehow and pass it down
```suggestion
  public edit(options: GuildScheduledEventEditOptions): Promise<GuildScheduledEvent>;
  public delete(): Promise<GuildScheduledEvent<S>>;
  public setName(name: string): Promise<GuildScheduledEvent<S>>;
  public setScheduledStartTime(scheduledStartTime: DateResolvable): Promise<GuildScheduledEvent<S>>;
  public setScheduledEndTime(scheduledEndTime: DateResolvable | null): Promise<GuildScheduledEvent<S>>;
  public setDescription(description: string): Promise<GuildScheduledEvent<S>>;
  public setStatus(status: GuildScheduledEventSetStatusArg<S>): Promise<GuildScheduledEvent>;
  public setLocation(location: string): Promise<GuildScheduledEvent<S>>;
```

vladfrangu (changes requested):

src/managers/GuildScheduledEventManager.js
iShibi: oopsie, added the fix only to `create`, should be there in the next commit for `edit` too

src/structures/GuildScheduledEvent.js
ImRodry: Why not a unix timestamp?

src/structures/GuildScheduledEvent.js
iShibi: this was a dirtttty hack that I forgot to update, fortunately, it's not required anymore. So, I'll just remove it.

src/structures/GuildScheduledEvent.js
iShibi: any valid Date parameter would work but discord sends ISO strings so it makes sense to show the example using it

src/structures/GuildScheduledEvent.js
iShibi: I did this with StageInstance too and I honestly have no idea why :pepeHands:

typings/index.d.ts
iShibi: ✔️

iCrawl (approved):

typings/index.d.ts
```diff
@@ -2923,6 +2981,29 @@ export class GuildInviteManager extends DataManager<string, Invite, InviteResolv
   public delete(invite: InviteResolvable, reason?: string): Promise<Invite>;
 }
 
+export class GuildScheduledEventManager extends CachedManager<
```
SpaceEEC, line 2984: I think this manager should be added to `Caches` (line 3671 as of the commit I am reviewing)

SpaceEEC (changes requested):

src/structures/GuildScheduledEvent.js
```diff
@@ -0,0 +1,422 @@
...
+    } else {
+      this.description ??= null;
+    }
+
+    /**
+     * The timestamp the guild scheduled event will start at
+     * <info>This can be potentially `null` only when it's an {@link AuditLogEntryTarget}</info>
+     * @type {?number}
+     */
+    this.scheduledStartTimestamp = data.scheduled_start_time ? new Date(data.scheduled_start_time).getTime() : null;
```
kyranet, line 78:
```suggestion
    this.scheduledStartTimestamp = data.scheduled_start_time ? Date.parse(data.scheduled_start_time) : null;
```

src/structures/GuildScheduledEvent.js
```diff
@@ -0,0 +1,422 @@
...
+
+    /**
+     * The timestamp the guild scheduled event will end at,
+     * or `null` if the event does not have a scheduled time to end
+     * @type {?number}
+     */
+    this.scheduledEndTimestamp = data.scheduled_end_time ? new Date(data.scheduled_end_time).getTime() : null;
```
kyranet, line 85:
```suggestion
    this.scheduledEndTimestamp = data.scheduled_end_time ? Date.parse(data.scheduled_end_time) : null;
```

src/structures/GuildScheduledEvent.js
```diff
@@ -0,0 +1,422 @@
...
+  }
+
+  /**
+   * The time the guild scheduled event will end at,
+   * or `null` if the event does not have a scheduled time to end
+   * @type {?Date}
+   * @readonly
+   */
+  get scheduledEndAt() {
+    return this.scheduledEndTimestamp ? new Date(this.scheduledEndTimestamp) : null;
```
kyranet, line 195: Following the new semantics we're using in latest PRs:
```suggestion
    return this.scheduledEndTimestamp && new Date(this.scheduledEndTimestamp);
```

kyranet (changes requested):

kyranet (approved):

SpaceEEC (approved):

iCrawl:
This needs a rebase

src/structures/GuildScheduledEvent.js
```diff
@@ -0,0 +1,422 @@
...
+   * // Set start time of a guild scheduled event
+   * guildScheduledEvent.setScheduledStartTime('2022-09-24T00:00:00+05:30')
+   *  .then(guildScheduledEvent => console.log(`Set the start time to: ${guildScheduledEvent.scheduledStartTime}`))
+   *  .catch(console.error);
+   */
+  setScheduledStartTime(scheduledStartTime, reason) {
+    return this.edit({ scheduledStartTime, reason });
+  }
+
+  // TODO: scheduledEndTime gets reset on passing null but it hasn't been documented
```
vladfrangu, line 306: So what are we doing with this?

vladfrangu (approved):

src/structures/GuildScheduledEvent.js
iShibi: Would it be a breaking change if we make the method take `null` later on once it gets documented *somehow* (no I'm definitely not being lazy here😹)

iCrawl: Might happen before v14 anyway so we should be fine.

PR #6496, opened 2021-08-21
feat(Invite): support friend invites

advaith1:
Adds support for friend invites:
* https://github.com/discord/discord-api-docs/pull/3659

To be non-breaking, this PR adds a `supportFriendInvites` ClientOption: by default, it is `false`, and fetching a friend invite will continue to throw an error. If it is set to `true`, fetching a friend invite will return an Invite with channel as null.

The guide should recommend setting `supportFriendInvites` to true.

### DIFF
--- a/src/util/Options.js
+++ b/src/util/Options.js
@@ -67,6 +67,9 @@
  * @property {boolean} [failIfNotExists=true] Default value for {@link ReplyMessageOptions#failIfNotExists}
  * @property {string[]} [userAgentSuffix] An array of additional bot info to be appended to the end of the required
  * [User Agent](https://discord.com/developers/docs/reference#user-agent) header
+ * @property {boolean} [supportFriendInvites=false] Whether to support friend invites (where channel is null),
+ * if false, an error will be thrown when fetching a friend invite.
+ * This option exists for backwards compatibility, and will always be true in a future version.
  * @property {PresenceData} [presence={}] Presence data to use upon login
  * @property {IntentsResolvable} intents Intents to enable for this connection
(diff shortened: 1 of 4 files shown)


almeidx (commented):
The `supportFriendInvites` client option is missing in the typings

src/util/Options.js
```diff
@@ -123,6 +126,7 @@ class Options extends null {
       restSweepInterval: 60,
       failIfNotExists: true,
       userAgentSuffix: [],
+      supportFriendInvites: false,
```
vladfrangu, line 129: If we want to move forward with feature flags, we should probably group them under one object. Thoughts @discordjs/the-big-3?

vladfrangu (changes requested):

src/util/Options.js
SpaceEEC: If we are going that way, grouping them would in my opinion be the best way to go about this.

advaith1: what should it be called? `features`?

iCrawl:
This needs a rebase.

almeidx:
The upstream pr was closed but `channel` is still nullable as per https://github.com/discord/discord-api-docs/commit/b9663021282673f07d08728147a7bdc6cc45a82a

iCrawl:
is this still a feature discord is exploring / wanting to go forward with @advaith1 ?

ImRodry:
They said it isn’t: https://github.com/discord/discord-api-docs/pull/3659#issuecomment-988178670 but the PR could be changed to just make channel nullable

iCrawl:
Separate PR then 👍

PR #6500, opened 2021-08-22
feat(CommandInteractionOptionResolver): add default value for get[option]

Wondermarin:

Adds the `defaultValue` parameter for `getBoolean`, `getString`, `getInteger`, and `getNumber` in cases when the option is optional.

I'm not sure whether it's worth adding this parameter for the rest of the options, so for now I've stopped only on the "primitive" ones.


advaith1:
```js
getString('field', false, 'default')
```
is longer than
```js
getString('field') ?? 'default'
```
so this doesn't seem very useful

Wondermarin:
> ```js
> getString('field', false, 'default')
> ```
> 
> is longer than
> 
> ```js
> getString('field') ?? 'default'
> ```
> 
> so this doesn't seem very useful

Yes, it's true.

However, in TypeScript, the solution you are talking about is unsafe, because with this operator you can assign a string to `getInteger` and it will not cause an error.

And the current implementation looks more expensive:
```ts
// Current implementation.
const x = getInteger("x") ?? "5"; // no error

// "Correct" implementation.
const x: number = getInteger("x") ?? 5;

// PR implementation.
const x = getInteger("x", false, 5);
```

There may also be problems using ECMAScript in the version when there was no 'nullish coalescing operator' yet. Then the code will look like this:
```js
// Current implementation.
let x = getInteger("x");

if (!x) x = 5;

// PR implementation.
const x = getInteger("x", false, 5);
```

In any case, I don't think that the implementation in this PR has no right to life. I agree that it may be **more** expensive in the example that you specified, but it may also be **less** expensive in other cases.
And in conclusion, it would be nice, as for me, if this feature was natively implemented in the library.

suneettipirneni:
> > ```js
> 
> > getString('field', false, 'default')
> 
> > ```
> 
> > 
> 
> > is longer than
> 
> > 
> 
> > ```js
> 
> > getString('field') ?? 'default'
> 
> > ```
> 
> > 
> 
> > so this doesn't seem very useful
> 
> 
> 
> Yes, it's true.
> 
> 
> 
> However, in TypeScript, the solution you are talking about is unsafe, because with this operator you can assign a string to `getInteger` and it will not cause an error.
> 
> 
> 
> And the current implementation looks more expensive:
> 
> ```ts
> 
> // Current implementation.
> 
> const x = getInteger("x") ?? "5"; // no error
> 
> 
> 
> // "Correct" implementation.
> 
> const x: number = getInteger("x") ?? 5;
> 
> 
> 
> // PR implementation.
> 
> const x = getInteger("x", false, 5);
> 
> ```
> 
> 
> 
> There may also be problems using ECMAScript in the version when there was no 'nullish coalescing operator' yet. Then the code will look like this:
> 
> ```js
> 
> // Current implementation.
> 
> let x = getInteger("x");
> 
> 
> 
> if (!x) x = 5;
> 
> 
> 
> // PR implementation.
> 
> const x = getInteger("x", false, 5);
> 
> ```
> 
> 
> 
> In any case, I don't think that the implementation in this PR has no right to life. I agree that it may be **more** expensive in the example that you specified, but it may also be **less** expensive in other cases.
> 
> 
> 
> And in conclusion, it would be nice, as for me, if this feature was natively implemented in the library.

Typescript will evaluate the type to `number | string` in the example you gave. Even if someone does this in their code, you won't be able to use the type `number | string` as a number or string without casting at some point to a `number` or `string`. If you get a union type composed of two orthogonal types it's pretty clear IMO that you're using orthogonal/mismatched types.

Wondermarin:
> > However, in TypeScript, the solution you are talking about is unsafe, because with this operator you can assign a string to `getInteger` and it will not cause an error.
> 
> Typescript will evaluate the type to `number | string` in the example you gave. Even if someone does this in their code, you won't be able to use the type `number | string` as a number or string without casting at some point to a `number` or `string`. If you get a union type composed of two orthogonal types it's pretty clear IMO that you're using orthogonal/mismatched types. Try to pass it in as a `string` parameter, that's a compiler error. Try to assign it to a `number` type that's also a compiler error.

It is clear that if you use this in the code:
```ts
const x = getInteger("x") ?? "5"; // x: number | string;
```

Then TypeScript **SHOULD NOT** output an error, but here we are talking about basic type protection, using `getInteger` you expect to get a number, using native support for "default values" your IDE will report an error at the stage of writing code, because you are trying to assign a string to a number.

kyranet:
I'd consider that as user-error, and it's their responsibility to do such things. Besides, the default is always evaluated and cannot be done lazily, and you can't stop them from using `getString('field') ?? 'default'` either way.

Issue #6379, opened 2021-08-10
Add support for invitable field for private threads


ckohen:
**Is your feature request related to a problem? Please describe.**

Upstream PR: 
- https://github.com/discord/discord-api-docs/pull/3600

**Describe the ideal solution**

- [x] Add invitable to the ThreadChannel structure
- [x] Add support for editing invitable (private threads only!)
- [x] Add support for invitable field when creating private threads

**Describe alternatives you've considered**

N/A

**Additional context**

Waiting to implement this until it is deployed to the API so testing of behavior can be done.

PR #6501, opened 2021-08-22
feat(Threads): add support for invitable in private threads

ckohen:

Closes: #6379 

Fixes a bug in typings that didn't pass the allowable thread types properly for news channels since this relies on it.

⚠️ The `invitable` field in the thread creation endpoint does not seem to be respected at the moment, it is supported here, but the API does not utilize it, always setting true.
- https://github.com/discord/discord-api-docs/issues/3700

### DIFF
--- a/src/managers/ThreadManager.js
+++ b/src/managers/ThreadManager.js
@@ -78,6 +78,8 @@ class ThreadManager extends CachedManager {
    * @property {ThreadChannelTypes|number} [type] The type of thread to create. Defaults to `GUILD_PUBLIC_THREAD` if
    * created in a {@link TextChannel} <warn>When creating threads in a {@link NewsChannel} this is ignored and is always
    * `GUILD_NEWS_THREAD`</warn>
+   * @property {boolean} [invitable] Whether non-moderators can add other non-moderators to the thread
+   * <info>Can only be set when type will be `GUILD_PRIVATE_THREAD` (12)</info>
    */
 
   /**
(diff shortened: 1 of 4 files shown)


src/structures/ThreadChannel.js
```diff
@@ -109,6 +117,9 @@ class ThreadChannel extends Channel {
       if (!this.archiveTimestamp) {
         this.archiveTimestamp = null;
       }
+      if (!this.invitable) {
+        this.invitable = null;
+      }
```
ImRodry, line 122: Since we're using node v16.6 now, why not change this and all the other ones above to the following
```suggestion
        this.invitable ??= null;
```

typings/index.d.ts
```diff
@@ -4571,6 +4575,7 @@ export type ThreadChannelTypes = 'GUILD_NEWS_THREAD' | 'GUILD_PUBLIC_THREAD' | '
 export interface ThreadCreateOptions<AllowedThreadType> extends StartThreadOptions {
   startMessage?: MessageResolvable;
   type?: AllowedThreadType;
+  invitable?: AllowedThreadType extends 'GUILD_PRIVATE_THREAD' | 12 ? boolean : undefined;
```
SpaceEEC, line 4578: I didn't test this, but wouldn't `never` be more appropriate here?
```suggestion
  invitable?: AllowedThreadType extends 'GUILD_PRIVATE_THREAD' | 12 ? boolean : never;
```

src/managers/ThreadManager.js
```diff
@@ -78,6 +78,8 @@ class ThreadManager extends CachedManager {
    * @property {ThreadChannelTypes|number} [type] The type of thread to create. Defaults to `GUILD_PUBLIC_THREAD` if
    * created in a {@link TextChannel} <warn>When creating threads in a {@link NewsChannel} this is ignored and is always
    * `GUILD_NEWS_THREAD`</warn>
+   * @property {boolean} [invitable] Whether non-moderators can add other non-moderators to the thread
+   * <info>Can only be set when type will be `GUILD_PRIVATE_THREAD` (12)</info>
```
iCrawl, line 82:
```suggestion
   * <info>Can only be set when type will be `GUILD_PRIVATE_THREAD`</info>
```

SpaceEEC (approved):

PR #6509, opened 2021-08-23
feat(MessageAttachment): allow files to be marked as spoilers

### LINKED ISSUE #6473: Set MessageAttachment as spoiler for remote file
I would like to send an attachment message based on an url as a spoiler.

Create a setter for the spoiler property of the MessageAttachment class

```js
const post = getPostFromReddit('surprise')
const attachment = new MessageAttachment(post.url)
attachment.spoiler = true
discordChannel.send(post.title, attachment)
```

Download the image and send it as a file with the spoiler tag in the filename.

**Additional context**
I like the fact that you can send an attachment message from a remote image.

J-Human:

This allows users to mark media as spoilers before sending them to Discord's API. The motivation behind this is for convenience, as in, users not needing to place the `SPOILER_` string.

You'll see a validation if the given argument is false, as it's possible to have the spoiler string inside the *string itself*, hence removing the title users would expect. 

I have adapted this PR from `#6473` but I didn't make it an equivalent setter as the function is more easy to use.

Resolves #6473

### DIFF
--- a/src/managers/MessageManager.js
+++ b/src/managers/MessageManager.js
@@ -178,7 +178,7 @@ class MessageManager extends CachedManager {
 
   /**
    * Adds a reaction to a message, even if it's not cached.
-   * @param {MessageResolvable} message The messag to react to
+   * @param {MessageResolvable} message The message to react to
    * @param {EmojiIdentifierResolvable} emoji The emoji to react with
    * @returns {Promise<void>}
    */
--- a/src/structures/MessageAttachment.js
+++ b/src/structures/MessageAttachment.js
@@ -50,8 +50,13 @@ class MessageAttachment {
    */
   setSpoiler(spoiler = true) {
     if (spoiler === this.spoiler) return this;
-
-    this.name = spoiler ? `SPOILER_${this.name}` : this.name.slice('SPOILER_'.length);
+    
+    if (!spoiler) {
+      while (this.spoiler) {
+        this.name.slice('SPOILER_'.length);
+      }
+    }
+    this.name = `SPOILER_${this.name}`;
     return this;
   }
 
--- a/src/structures/MessageAttachment.js
+++ b/src/structures/MessageAttachment.js
@@ -51,8 +51,7 @@ class MessageAttachment {
   setSpoiler(spoiler = true) {
     if (spoiler === true && this.spoiler === false) {
       this.name = `SPOILER_${this.name}`;
-    }
-    if (spoiler === false && this.spoiler === true) {
+    } else if (spoiler === false && this.spoiler === true) {
       this.name = this.name.replace('SPOILER_', '');
     }
     return this;
--- a/src/structures/MessageAttachment.js
+++ b/src/structures/MessageAttachment.js
@@ -49,10 +49,12 @@ class MessageAttachment {
    * @returns {MessageAttachment} This attachment
    */
   setSpoiler(spoiler = true) {
-    if (spoiler === true && this.spoiler === false) {
-      this.name = `SPOILER_${this.name}`;
-    } else if (spoiler === false && this.spoiler === true) {
-      this.name = this.name.replace('SPOILER_', '');
+    if (spoiler !== this.spoiler) {
+      if (spoiler) {
+        this.name = `SPOILER_${this.name}`;
+      } else {
+        this.name = this.name.replace(/^SPOILER_/, '');
+      }
     }
     return this;
   }
--- a/src/structures/MessageAttachment.js
+++ b/src/structures/MessageAttachment.js
@@ -43,6 +43,21 @@ class MessageAttachment {
     return this;
   }
 
+  /**
+   * Sets whether this attachment is a spoiler
+   * @param {boolean} [spoiler=true] Whether the attachment should be marked as a spoiler
+   * @returns {MessageAttachment} This attachment
+   */
+  setSpoiler(spoiler = true) {
+    if (spoiler === true && this.spoiler === false) {
+      this.name = `SPOILER_${this.name}`;
+    }
+    if (spoiler === false && this.spoiler === true) {
+      this.name = this.name.replace('SPOILER_', '');
+    }
+    return this;
+  }
+
   _patch(data) {
     /**
      * The attachment's id
@@ -93,7 +108,7 @@ class MessageAttachment {
    * @readonly
    */
   get spoiler() {
-    return Util.basename(this.url).startsWith('SPOILER_');
+    return Util.basename(this.url ?? this.name).startsWith('SPOILER_');
   }
 
   toJSON() {
--- a/src/structures/MessageAttachment.js
+++ b/src/structures/MessageAttachment.js
@@ -50,11 +50,12 @@ class MessageAttachment {
    */
   setSpoiler(spoiler = true) {
     if (spoiler === this.spoiler) return this;
-    
+
     if (!spoiler) {
       while (this.spoiler) {
         this.name.slice('SPOILER_'.length);
       }
+      return this;
     }
     this.name = `SPOILER_${this.name}`;
     return this;
--- a/src/structures/MessageAttachment.js
+++ b/src/structures/MessageAttachment.js
@@ -49,13 +49,9 @@ class MessageAttachment {
    * @returns {MessageAttachment} This attachment
    */
   setSpoiler(spoiler = true) {
-    if (spoiler !== this.spoiler) {
-      if (spoiler) {
-        this.name = `SPOILER_${this.name}`;
-      } else {
-        this.name = this.name.replace(/^SPOILER_/, '');
-      }
-    }
+    if (spoiler === this.spoiler) return this;
+
+    this.name = spoiler ? `SPOILER_${this.name}` : this.name.slice('SPOILER_'.length);
     return this;
   }
 
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -1205,6 +1205,7 @@ export class MessageAttachment {
   public width: number | null;
   public setFile(attachment: BufferResolvable | Stream, name?: string): this;
   public setName(name: string): this;
+  public setSpoiler(spoiler?: boolean): this;
   public toJSON(): unknown;
 }
 


src/structures/MessageAttachment.js
```diff
@@ -43,6 +43,22 @@ class MessageAttachment {
     return this;
   }
 
+  /**
+   * Sets or unsets the attachment as a spoiler
+   * @param {boolean} spoiler Whether the attachment should be in a spoiler
```
vBread, line 48:
```suggestion
   * Sets whether this attachment is a spoiler
   * @param {boolean} spoiler Whether the attachment should be marked as a spoiler
```
Wording here could be improved to be consistent with other descriptions (e.g., [`.setNSFW`](https://discord.js.org/#/docs/main/stable/class/BaseGuildTextChannel?scrollTo=setNSFW))

src/structures/MessageAttachment.js
```diff
@@ -43,6 +43,22 @@ class MessageAttachment {
     return this;
   }
 
+  /**
+   * Sets whether this attachment is a spoiler
+   * @param {boolean} spoiler Whether the attachment should be marked as a spoiler
+   * @returns {MessageAttachment} This attachment
+   */
+  setSpoiler(spoiler) {
+    if (spoiler === true) {
```
ckohen, line 52: You should do the reverse validation here to avoid ending up with `SPOILER_SPOILER_name.png` or something

src/structures/MessageAttachment.js
```diff
@@ -43,6 +43,22 @@ class MessageAttachment {
...
+  /**
+   * Sets whether this attachment is a spoiler
+   * @param {boolean} spoiler Whether the attachment should be marked as a spoiler
+   * @returns {MessageAttachment} This attachment
+   */
+  setSpoiler(spoiler) {
+    if (spoiler === true && this.spoiler === false) {
+      this.name = `SPOILER_${this.name}`;
+      return this;
+    } else if (spoiler === false && this.spoiler === true) {
```
ckohen, line 55: No need for else if here, just use if since the first statement always returns

src/structures/MessageAttachment.js
```diff
@@ -43,6 +43,21 @@ class MessageAttachment {
...
+   * @returns {MessageAttachment} This attachment
+   */
+  setSpoiler(spoiler) {
+    if (spoiler === true && this.spoiler === false) {
+      this.name = `SPOILER_${this.name}`;
+      return this;
+    } else {
+      if (this.spoiler === true) this.name.replace('SPOILER_', '');
+      return this;
+    }
```
ckohen, line 58:
```suggestion
    }
    if (this.spoiler === true) this.name.replace('SPOILER_', '');
    return this;
```

That's not quite what I meant 😆 The else is redundant since the first  if block always returns

Jiralite (commented):
Has this been tested? This method is calling the spoiler getter, which returns the basename utility method and passes the URL across. But that won't exist via instantiating this class and will throw an error?

src/structures/MessageAttachment.js
```diff
@@ -43,6 +43,20 @@ class MessageAttachment {
...
+   * Sets whether this attachment is a spoiler
+   * @param {boolean} spoiler Whether the attachment should be marked as a spoiler
+   * @returns {MessageAttachment} This attachment
+   */
+  setSpoiler(spoiler) {
+    if (spoiler === true && !this.name.startsWith('SPOILER_')) {
+      this.name = `SPOILER_${this.name}`;
+      return this;
+    }
+    if (this.name.startsWith('SPOILER_')) this.name.replace('SPOILER_', '');
```
ImRodry, line 56: What if spoiler is true and the name already starts with SPOILER_ ? I believe this should also be checked for if spoiler is false
Also spoiler should probably be true by default like other setX methods

src/managers/MessageManager.js
```diff
@@ -178,7 +178,7 @@ class MessageManager extends CachedManager {
 
   /**
    * Adds a reaction to a message, even if it's not cached.
-   * @param {MessageResolvable} message The messag to react to
+   * @param {MessageResolvable} message The message to react to
```
ImRodry, line 181: Oh and why is this being fixed again? You should probably rebase xD

src/structures/MessageAttachment.js
```diff
@@ -43,6 +43,20 @@ class MessageAttachment {
...
+   * Sets whether this attachment is a spoiler
+   * @param {boolean} [spoiler=true] Whether the attachment should be marked as a spoiler
+   * @returns {MessageAttachment} This attachment
+   */
+  setSpoiler(spoiler = true) {
+    if (spoiler === true && !this.name.startsWith('SPOILER_')) {
+      this.name = `SPOILER_${this.name}`;
+      return this;
+    }
+    if (spoiler === false && this.name.startsWith('SPOILER_')) this.name.replace('SPOILER_', '');
```
ImRodry, line 56: String.prototype.replace does not change the original string, so this is needed
```suggestion
    if (spoiler === false && this.name.startsWith('SPOILER_')) {
        this.name = this.name.replace('SPOILER_', '');
    }
```

J-Human: Yep, seems like I forgot about that, since commits ago.

J-Human: Wait, couldn't it just be `if (spoiler === false && this.name.startsWith('SPOILER_')) this.name = this.name.replace('SPOILER_', '');`? I recall @ ckohen suggesting to disregard the if/else chain, so...

ImRodry: What are you suggesting? I just moved it to a new like because otherwise ESLint would complain

ImRodry:
I suggested the brackets exactly because of that. You should install eslint locally so that it tells you about these errors before pushing.

src/structures/MessageAttachment.js
J-Human: Woops

ImRodry (approved):
Imo you only need 1 return at the end, the others are redundant. Otherwise LGTM

SpaceEEC (approved):

Jiralite:
This is a bit strange now. If a user runs `.setSpoiler()`, the attachment will be marked as a spoiler that is yet to be sent. If they decide to run the `.spoiler` getter to see if it's been marked as a spoiler, it'll throw an error because it's expecting a URL. Right now it's valid only for received messages... maybe it should be adjusted for to-be messages also?

src/structures/MessageAttachment.js
```diff
@@ -43,6 +43,21 @@ class MessageAttachment {
     return this;
   }
 
+  /**
+   * Sets whether this attachment is a spoiler
+   * @param {boolean} [spoiler=true] Whether the attachment should be marked as a spoiler
+   * @returns {MessageAttachment} This attachment
+   */
+  setSpoiler(spoiler = true) {
+    if (spoiler === true && !this.name.startsWith('SPOILER_')) {
```
NotSugden, line 52: i would personally go for this approach
```js
const isSpoiler = this.name.startsWith("SPOILER_");
if (spoiler != isSpoiler) {
  if (spoiler) {
    this.name = `SPOILER_${this.name}`;
  } else {
    this.name = this.name.slice(8);
    // or
    this.name = this.name.replace(/^SPOILER_/, '');
  }
}
return this;
```

or at the least putting the `this.name.startsWith(...);` in a variable instead of calling it twice, and using an else if statement below

ImRodry: I agree and slice may actually be better

ImRodry:
> This is a bit strange now. If a user runs `.setSpoiler()`, the attachment will be marked as a spoiler that is yet to be sent. If they decide to run the `.spoiler` getter to see if it's been marked as a spoiler, it'll throw an error because it's expecting a URL. Right now it's valid only for received messages... maybe it should be adjusted for to-be messages also?

Tbh it should’ve been like that from the beginning, and it’s as simple as adding `?? this.name` inside the Util.basename call

src/structures/MessageAttachment.js
kyranet: We could also maake an `isSpoiler` getter, and flatten indentation a bit:

```js
if (spoiler === this.isSpoiler) return this;

this.name = spoiler
  ? `SPOILER_${this.name}`
  : this.name.slice('SPOILER_'.length);

return this;
```

ImRodry: But this.spoiler already exists ?

NotSugden: yeah, i put the simple `isSpoiler` in the suggested code because of the problem mentioned before of it requiring a URL, if thats fixed then the getter could be used instead

ImRodry: Yeah fixing the getter would be the best approach like I suggested below this

J-Human:
> > This is a bit strange now. If a user runs `.setSpoiler()`, the attachment will be marked as a spoiler that is yet to be sent. If they decide to run the `.spoiler` getter to see if it's been marked as a spoiler, it'll throw an error because it's expecting a URL. Right now it's valid only for received messages... maybe it should be adjusted for to-be messages also?
> 
> Tbh it should’ve been like that from the beginning, and it’s as simple as adding `?? this.name` inside the Util.basename call

Yeah, I actually agree with this idea since it's the shortest approach to do and it simplifies the job and any future tasks that needs name validation. I'll commit this, along with the if/else-if thing, as it was suggested to be removed by ckohen earlier on.

NotSugden (changes requested):
remove `x === boolean` in favour of `x` and `!x`

J-Human:
Done, suggestions have been committed

src/structures/MessageAttachment.js
```diff
@@ -43,6 +43,22 @@ class MessageAttachment {
...
+   */
+  setSpoiler(spoiler = true) {
+    if (spoiler !== this.spoiler) {
+      if (spoiler) {
+        this.name = `SPOILER_${this.name}`;
+      } else {
+        this.name = this.name.replace(/^SPOILER_/, '');
+      }
+    }
+    return this;
```
kyranet, line 59: Just to remove indentation and using ternaries for simplicity:
```suggestion
    if (spoiler === this.spoiler) return this;

    this.name = spoiler ? `SPOILER_${this.name}` : this.name.replace(/^SPOILER_/, '');
    return this;
```

Also, I still prefer this variation to avoid using RegExp due to its performance implications:
```suggestion
    if (spoiler === this.spoiler) return this;

    this.name = spoiler ? `SPOILER_${this.name}` : this.name.slice('SPOILER_'.length);
    return this;
```

In runtime, doing `'SPOILER_'.length` over and over doesn't matter because since it's a constant string, it'll be optimised to just the length, so it's equally as fast as just `8` while being a lot more readable.

kyranet (changes requested):

src/structures/MessageAttachment.js
vladfrangu: Doing .slice would also mean sequential calls to setSpoiler(false) will make the name go \*poof\* tho. Now, if that's something we want to prevent is a different story

kyranet: No, because `this.spoiler` returns `true` when it's prefixed with `SPOILER_`.

If you set `spoiler` to false, and `this.spoiler` is also false (not prefixed), no operation is done.

vladfrangu: Brain.OS is still loading... 😅

J-Human: But the goal of removing the spoiler is only at the start, isn't it? We wouldn't want messing up the user's image if it was intentionally placed with some spoiler string, it would make more sense if it only occurred once in the *start* of the string, right? So I agree with kyranet's suggestion, except I'm a bit hesitated with regards the number 8 since Rodry's comment makes sense, regarding randomness and knowing where it comes from.

kyranet: I'm suggesting the usage of `'SPOILER_'.length` over magic numbers (`8`), not the other way around.

J-Human: Ah, got it. All good then

J-Human:
This is unusual. It seems all 4 checks failed to request npm. Kindly rerun them again later. 😄

kyranet:
~~It seems like npm hates you.~~ I'll be re-running until it goes through.

kyranet (approved):

src/structures/MessageAttachment.js
ImRodry: You can also try a while loop that checks if the string still starts with SPOILER_ since I know some soul out there will run into this problem at some point. That way you don’t delete other words further into the name and only get rid of the spoiler itself @J-Human

src/structures/MessageAttachment.js
```diff
@@ -43,6 +43,23 @@ class MessageAttachment {
...
+   */
+  setSpoiler(spoiler = true) {
+    if (spoiler === this.spoiler) return this;
+    
+    if (!spoiler) {
+      while (this.spoiler) {
+        this.name.slice('SPOILER_'.length);
+      }
+    }
+    this.name = `SPOILER_${this.name}`;
```
ImRodry, line 59: You either need an else here or a return statement above, otherwise this removes and then adds back the spoiler

J-Human: Alright

J-Human:
Done

src/structures/MessageAttachment.js
```diff
@@ -43,6 +43,24 @@ class MessageAttachment {
...
+   * Sets whether this attachment is a spoiler
+   * @param {boolean} [spoiler=true] Whether the attachment should be marked as a spoiler
+   * @returns {MessageAttachment} This attachment
+   */
+  setSpoiler(spoiler = true) {
+    if (spoiler === this.spoiler) return this;
+
+    if (!spoiler) {
+      while (this.spoiler) {
+        this.name.slice('SPOILER_'.length);
```
NotSugden, line 56: is it worth making a constant for this?

eg just
```js
const SPOILER_PREFIX = 'SPOILER_';
```
at the top or exported in constants

NotSugden (changes requested):

iCrawl (approved):

src/structures/MessageAttachment.js
iCrawl: We should be fine without.

Issue #6510, opened 2021-08-23
Bot crash while it updates itself


Positive2022:
**Please describe the problem you are having in as much detail as possible:**
I was updating my bot's pfp and I had my bots console opened cuz I was updating it. Just when i updated the bot's pfp, the bot crashed giving out an error.

Error:
```js
/home/container/node_modules/discord.js/src/client/actions/UserUpdate.js:11
    const oldUser = newUser._update(data);
                            ^

TypeError: Cannot read property '_update' of undefined
    at UserUpdateAction.handle (/home/container/node_modules/discord.js/src/client/actions/UserUpdate.js:11:29)
    at Object.module.exports [as USER_UPDATE] (/home/container/node_modules/discord.js/src/client/websocket/handlers/USER_UPDATE.js:4:29)
    at WebSocketManager.handlePacket (/home/container/node_modules/discord.js/src/client/websocket/WebSocketManager.js:345:31)
    at WebSocketShard.onPacket (/home/container/node_modules/discord.js/src/client/websocket/WebSocketShard.js:443:22)
    at WebSocketShard.onMessage (/home/container/node_modules/discord.js/src/client/websocket/WebSocketShard.js:300:10)
    at WebSocket.onMessage (/home/container/node_modules/ws/lib/event-target.js:132:16)
    at WebSocket.emit (node:events:394:28)
    at Receiver.receiverOnMessage (/home/container/node_modules/ws/lib/websocket.js:970:20)
    at Receiver.emit (node:events:394:28)
    at Receiver.dataMessage (/home/container/node_modules/ws/lib/receiver.js:517:14)
```

**Further details:**

- discord.js version: 13.1.0
- Node.js version: 16.1.0
- Operating system: LInux
- Priority this issue should have – please be realistic and elaborate if possible: Medium

**Relevant client options:**

- partials: none
- gateway intents: none
- other: All non privilleged intents.

Positive2022:
Note: Also happened when I changed the bots username.
Note 2: The pfp error that occured when I changed application's pfp which ofcourse changes bot user's pfp.

PR #6511, opened 2021-08-23
fix(UserUpdateAction): rely on client.user when ids match

Keke27210:

When the `UserManager` cache is disabled, updating the client user will result in a crash. The user update action gets the old state of the user via the `UserManager` cache, even if it's the client user. However, when it's disabled, nothing is being returned and so `newUser._update` throws a type error.

This fix makes the code get `client.user` if `client.user.id` and `data.id` are equal.

### DIFF
--- a/src/client/actions/UserUpdate.js
+++ b/src/client/actions/UserUpdate.js
@@ -7,7 +7,7 @@ class UserUpdateAction extends Action {
   handle(data) {
     const client = this.client;
 
-    const newUser = client.users.cache.get(data.id);
+    const newUser = data.id === client.user.id ? client.user : client.users.cache.get(data.id);
     const oldUser = newUser._update(data);
 
     if (!oldUser.equals(newUser)) {


SpaceEEC (approved):

iShibi:
What if the `UserUpdate` gets triggered by a non-client user who isn't cached? The `newUser` will be undefined in that case too and the same error will occur. I think you should use the [`getUser`](https://github.com/discordjs/discord.js/blob/70cc0295f833cd988ea627b37d20536f73e21630/src/client/actions/Action.js#L87) method here.

iShibi:
There is a chance that `getUser` will too return `undefined` if the client didn't opt for `USER` partial. The [`GuildMemberUpdate`](https://github.com/discordjs/discord.js/blob/70cc0295f833cd988ea627b37d20536f73e21630/src/client/actions/GuildMemberUpdate.js#L20) action handler does it the right way.

Keke27210:
> There is a chance that `getUser` will too return `undefined` if the client didn't opt for `USER` partial. The [`GuildMemberUpdate`](https://github.com/discordjs/discord.js/blob/70cc0295f833cd988ea627b37d20536f73e21630/src/client/actions/GuildMemberUpdate.js#L20) action handler does it the right way.

Discord's `USER_UPDATE` is only emitted for client user's changes. Changes for other users have to go through the `GUILD_MEMBER_UPDATE` event, where the `UserUpdate` action is not being called if the user wasn't in cache already.

That means if this action is being called by anything else than the actual `USER_UPDATE` event, the user is in the cache. Which also means the only user that may not be at this point, is the client user.

Now, if people have been messing enough with the caches, the library shouldn't be handling that for them. Enabling the `GUILD_MEMBERS` intent and disabling the `GuildManager` or the `UserManager` cache(s) is definitely going to lead to undefined behaviors but would be up to the end user responsibility imo.

The client user would be the only exception to that rule as the action is being called no matter what intents are being enabled. Which would mean the crash would occur as soon as the `UserManager` is disabled (#6510).

That's mainly why I haven't gotten further in the patch, thought that wouldn't be "that useful". That's also why I'd like to hear some maintainers point-of-views because I must admit I'm not very confident on that particular point.

kyranet (approved):

iCrawl (approved):

PR #6520, opened 2021-08-24
ci: setup npm auto deprecate workflow

Favna:

TL;DR: Script to automatically deprecate older npm versions, runs on a cron job of 01:00 UTC

---

Quite a while ago now, around the time we introduced auto published dev builds, we had a discussion in prof chat about how to handle old "dev" versions. From that spawned changes to the dev publisher to first deprecate and then publish a new version. A near perfect solution, but as seen at the time of writing, not a flawless one. This what NPM versions tab currently looks like:

image

Back then I mentioned how I would be working on a utility that could automatically deprecate versions as based on what I could get as info from `registry.npmjs.com`. Well I finally had time to make it now that I have vacation, so here it is.

This new workflow will run every night at 01:00 UTC. It scans all the versions that are on NPM and filters out those that:
- are not the "latest" tag
- do not have a dist tag associated with them
- are not already deprecated
- match the glob provided by the `--name` option (uses the micromatch lib)

So given the earlier screenshot that means that only the boxed version would get matched and deprecated with this script:

image

---

As as aside, this CLI tool _can_ also be used to manually bulk deprecate versions if needed in the future (i.e. deprecating whole v13 when we go v14 with `--name "13*"`) but just keep in mind that it is not capable of prompting for NPM OTP so it _must_ be used with an NPM Automation Token in your Env vars as `NODE_AUTH_TOKEN`

I have tested the CLI  tool against _all_ of Sapphire's current packages, successfully deprecating many old versions there.

---

Lastly, for now I pass the options as CLI flags. The tool _can_ also read a JSON or YAML config file in the project root, so if that is preferable let me know and I'll make the change.

---

### DIFF
--- a/.github/workflows/auto-deprecate.yml
+++ b/.github/workflows/auto-deprecate.yml
@@ -1,0 +1,31 @@
+name: NPM Auto Deprecate
+
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: '0 1 * * *'
+
+jobs:
+  auto-deprecate:
+    name: NPM Auto Deprecate
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout Project
+        uses: actions/checkout@v2
+      - name: Use Node.js v16
+        uses: actions/setup-node@v2
+        with:
+          node-version: 16
+      - name: Restore CI Cache
+        uses: actions/cache@v2.1.6
+        id: cache-restore
+        with:
+          path: node_modules
+          key: ${{ runner.os }}-16-${{ hashFiles('**/package-lock.json') }}
+      - name: Install Dependencies if Cache Miss
+        if: ${{ !steps.cache-restore.outputs.cache-hit }}
+        run: npm ci
+      - name: Deprecate versions
+        run: npx npm-deprecate --name "*dev*" --package "discord.js"
+        env:
+          NODE_AUTH_TOKEN: ${{ secrets.NPM_PUBLISH_TOKEN }}
--- a/package-lock.json
+++ b/package-lock.json
@@ -21,6 +21,7 @@
         "@commitlint/cli": "^13.1.0",
         "@commitlint/config-angular": "^13.1.0",
         "@discordjs/docgen": "^0.10.0",
+        "@favware/npm-deprecate": "^1.0.2",
         "@types/node": "^16.4.12",
         "conventional-changelog-cli": "^2.1.1",
         "cross-env": "^7.0.3",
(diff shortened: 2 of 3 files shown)


.github/workflows/auto-deprecate.yml
```diff
@@ -0,0 +1,31 @@
...
+        uses: actions/cache@v2.1.6
+        id: cache-restore
+        with:
+          path: node_modules
+          key: ${{ runner.os }}-16-${{ hashFiles('**/package-lock.json') }}
+      - name: Install Dependencies if Cache Miss
+        if: ${{ !steps.cache-restore.outputs.cache-hit }}
+        run: npm ci
+      - name: Deprecate versions
+        run: npx npm-deprecate --name "*dev*" --package "discord.js"
```
ckohen, line 29: Would a --no-install be desirable here?

Favna: It shouldn't matter because I added it as a devdep. 

As as aside, `--no-install` is deprecated, see: https://docs.npmjs.com/cli/v7/commands/npx#compatibility-with-older-npx-versions

ckohen: Wasn't aware --no was already implemented, nice.

> It shouldn't matter because I added it as a devdep.

Even more reason to do it, and with npm's replacement for npx, it does prompt if the command can't be found, meaning a stalled workflow if the dep fails to install the command properly, technically that's on your package, but its a thought.

vladfrangu: > it does prompt if the command can't be found

I've actually never seen it prompt in CI actually... maybe they check for that too?

ckohen: I haven't gotten a chance to run npm exec / npx in CI yet, so cannot confirm or deny with that, just remember that the old npx did not prompt at all. 

It does seem like CI detection is performed though in order to prevent this behavior, just not sure what the result is
> Interactive mode is not supported in CI environments when standard input is a TTY, to prevent hangs.

vladfrangu: The result is a warning + installing + executing from what I saw 👀

ImRodry:
Shouldn’t you remove the auto deprecate step in publish-dev then? Or is this just to make sure no versions are left over?

Favna:
@ImRodry That's a possibility. For now it's an extra step to clean up remainders. Whether to remove that step is something I first want to hear @iCrawl's thoughts on.

.github/workflows/auto-deprecate.yml
```diff
@@ -0,0 +1,31 @@
...
+        env:
+          NODE_AUTH_TOKEN: ${{ secrets.NPM_PUBLISH_TOKEN }}
```
iCrawl, line 31:
```suggestion
name: NPM auto deprecate
on:
  workflow_dispatch:
  schedule:
    - cron: '0 1 * * *'
jobs:
  auto-deprecate:
    name: NPM auto deprecate
    runs-on: ubuntu-latest
    steps:
      - name: Checkout repository
        uses: actions/checkout@v2

      - name: Install Node v16
        uses: actions/setup-node@v2
        with:
          node-version: 16

      - name: Install dependencies
        run: npm ci --ignore-scripts

      - name: Deprecate versions
        run: npx npm-deprecate --name "*dev*" --package "discord.js"
        env:
          NODE_AUTH_TOKEN: ${{ secrets.NPM_PUBLISH_TOKEN }}
```

iCrawl: Just to bring things in-line with the other actions we have

iCrawl (changes requested):

.github/workflows/auto-deprecate.yml
Favna: Added the flag and also changed it to `npm exec` since that's the new way. Tested the command how it's written now locally and it passes args properly (I made it so adding `--verbose` / `-v` will print the options it receives)

PR #6530, opened 2021-08-27
refactor: Remove `Util.removeMentions()`

Jiralite:
This method removes the user & role mentioning ability by inserting `\u200b` after the `@`. This essentially sanitises input to prevent mention injections when running `Message#cleanContent`. However, `Message#cleanContent` should only return the content with mentions replaced with names. Otherwise, this ruins the original message. 

`allowedMentions` has been implemented into discord.js for quite some time now and supersedes this "workaround". Therefore, I believe it's time we remove this method. Users should sanitise their own sending of messages via the `allowedMentions` feature if they don't want mentions to occur.

### DIFF
--- a/packages/discord.js/src/util/Util.js
+++ b/packages/discord.js/src/util/Util.js
@@ -1,7 +1,6 @@
 'use strict';
 
 const { parse } = require('node:path');
-const process = require('node:process');
 const { Collection } = require('@discordjs/collection');
 const fetch = require('node-fetch');
 const { Colors, Endpoints } = require('./Constants');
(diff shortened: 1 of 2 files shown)


iCrawl (approved):

SpaceEEC (approved):
🧹

vladfrangu (approved):
While I agree with the rest that this 🧹 moment is good, I'm a bit worried that `@${member.displayName}` COULD cause potential abuse vectors till people update their code to use allowedMentions (if they haven't already). Still, can be documented well in a changelog so LGTM

PR #6540, opened 2021-08-28
refactor: ES2021 features

ImRodry:
This PR makes use of ES2021's features: nullish assignment, underscore numeric operator, and String.prototype.replaceAll! Here are the changes:
- Replaced (hopefully) all `if (prop) prop = X` and `if (typeof prop === 'undefined') prop = X` with `prop ??= X`. There may be situations where ||= might be more appropriate but I didn't find any when going through these, however please let me know if you do
- Added _ to separate big numbers (bigger than 999) but only on numbers with 000 at some point in their structure. If there is for example a 58395766 out there, I did not catch that. I also purposely left out Constants as these would become harder to compare to Discord's API docs
- Replaced some String.prototype.replace with String.prototype.replaceAll in Util methods
- Applied prettier on one of the test files because I messed up the formatting and it did that afterward, so might as well

- This PR **only** includes non-code changes, like changes to documentation, README, etc. (the changes are in the code but should not have any impact on anything on the library)


src/client/websocket/WebSocketManager.js
```diff
@@ -19,7 +19,7 @@ const BeforeReadyWhitelist = [
 ];
 
 const UNRECOVERABLE_CLOSE_CODES = Object.keys(WSCodes).slice(1).map(Number);
-const UNRESUMABLE_CLOSE_CODES = [1000, 4006, 4007];
+const UNRESUMABLE_CLOSE_CODES = [1_000, 4_006, 4_007];
```
kyranet, line 22: Why not use `discord-api-types`? And same for the other codes in this and other files.

src/util/Util.js
```diff
@@ -181,7 +181,7 @@ class Util extends null {
    * @returns {string}
    */
   static escapeCodeBlock(text) {
-    return text.replace(/```/g, '\\`\\`\\`');
+    return text.replaceAll(/```/g, '\\`\\`\\`');
```
kyranet, line 184:
```suggestion
    return text.replaceAll('```', '\\`\\`\\`');
```

src/util/Util.js
```diff
@@ -190,7 +190,7 @@ class Util extends null {
    * @returns {string}
    */
   static escapeInlineCode(text) {
-    return text.replace(/(?<=^|[^`])`(?=[^`]|$)/g, '\\`');
+    return text.replaceAll(/(?<=^|[^`])`(?=[^`]|$)/g, '\\`');
```
kyranet, line 193: Unnecessary, please revert to `replace` as RegExp with global flag will go thru all instances.
```suggestion
    return text.replace(/(?<=^|[^`])`(?=[^`]|$)/g, '\\`');
```

src/util/Util.js
```diff
@@ -200,12 +200,12 @@ class Util extends null {
    */
   static escapeItalic(text) {
     let i = 0;
-    text = text.replace(/(?<=^|[^*])\*([^*]|\*\*|$)/g, (_, match) => {
+    text = text.replaceAll(/(?<=^|[^*])\*([^*]|\*\*|$)/g, (_, match) => {
```
kyranet, line 203: Unnecessary, please revert to `replace` as RegExp with global flag will go thru all instances.
```suggestion
    text = text.replace(/(?<=^|[^*])\*([^*]|\*\*|$)/g, (_, match) => {
```

src/util/Util.js
```diff
@@ -200,12 +200,12 @@ class Util extends null {
...
       if (match === '**') return ++i % 2 ? `\\*${match}` : `${match}\\*`;
       return `\\*${match}`;
     });
     i = 0;
-    return text.replace(/(?<=^|[^_])_([^_]|__|$)/g, (_, match) => {
+    return text.replaceAll(/(?<=^|[^_])_([^_]|__|$)/g, (_, match) => {
```
kyranet, line 208: Unnecessary, please revert to `replace` as RegExp with global flag will go thru all instances.
```suggestion
    return text.replace(/(?<=^|[^_])_([^_]|__|$)/g, (_, match) => {
```

src/util/Util.js
```diff
@@ -218,7 +218,7 @@ class Util extends null {
    */
   static escapeBold(text) {
     let i = 0;
-    return text.replace(/\*\*(\*)?/g, (_, match) => {
+    return text.replaceAll(/\*\*(\*)?/g, (_, match) => {
```
kyranet, line 221: Unnecessary, please revert to `replace` as RegExp with global flag will go thru all instances.
```suggestion
    return text.replace(/\*\*(\*)?/g, (_, match) => {
```

src/util/Util.js
```diff
@@ -231,7 +231,7 @@ class Util extends null {
    */
   static escapeUnderline(text) {
     let i = 0;
-    return text.replace(/__(_)?/g, (_, match) => {
+    return text.replaceAll(/__(_)?/g, (_, match) => {
```
kyranet, line 234: Unnecessary, please revert to `replace` as RegExp with global flag will go thru all instances.
```suggestion
    return text.replace(/__(_)?/g, (_, match) => {
```

src/util/Util.js
```diff
@@ -243,7 +243,7 @@ class Util extends null {
    * @returns {string}
    */
   static escapeStrikethrough(text) {
-    return text.replace(/~~/g, '\\~\\~');
+    return text.replaceAll(/~~/g, '\\~\\~');
```
kyranet, line 246:
```suggestion
    return text.replaceAll('~~', '\\~\\~');
```

src/util/Util.js
```diff
@@ -252,7 +252,7 @@ class Util extends null {
    * @returns {string}
    */
   static escapeSpoiler(text) {
-    return text.replace(/\|\|/g, '\\|\\|');
+    return text.replaceAll(/\|\|/g, '\\|\\|');
```
kyranet, line 255:
```suggestion
    return text.replaceAll('||', '\\|\\|');
```

src/util/Util.js
```diff
@@ -577,7 +577,7 @@ class Util extends null {
    * @returns {string}
    */
   static removeMentions(str) {
-    return str.replace(/@/g, '@\u200b');
+    return str.replaceAll(/@/g, '@\u200b');
```
kyranet, line 580:
```suggestion
    return str.replaceAll('@', '@\u200b');
```

src/util/Util.js
```diff
@@ -621,7 +621,7 @@ class Util extends null {
    * @returns {string}
    */
   static cleanCodeBlockContent(text) {
-    return text.replace(/```/g, '`\u200b``');
+    return text.replaceAll(/```/g, '`\u200b``');
```
kyranet, line 624:
```suggestion
    return text.replaceAll('```', '`\u200b``');
```

kyranet (dismissed):

src/client/websocket/WebSocketManager.js
ImRodry: Seems like that’s out of the scope of this PR and I’m also not very familiar with that package

src/client/Client.js
```diff
@@ -181,7 +181,7 @@ class Client extends BaseClient {
       );
       this.sweepMessageInterval = setInterval(
         this.sweepMessages.bind(this),
-        this.options.messageSweepInterval * 1000,
+        this.options.messageSweepInterval * 1_000,
```
vladfrangu, line 184: Am I the only one not a fan of using separators for such small numbers? I personally start using them from 10k+... @discordjs/the-big-3 thoughts?

src/structures/interfaces/TextBasedChannel.js
```diff
@@ -297,7 +297,7 @@ class TextBasedChannel {
     if (Array.isArray(messages) || messages instanceof Collection) {
       let messageIds = messages instanceof Collection ? [...messages.keys()] : messages.map(m => m.id ?? m);
       if (filterOld) {
-        messageIds = messageIds.filter(id => Date.now() - SnowflakeUtil.deconstruct(id).timestamp < 1209600000);
+        messageIds = messageIds.filter(id => Date.now() - SnowflakeUtil.deconstruct(id).timestamp < 1_209_600_000);
```
vladfrangu, line 300: Not for this PR but we should probably move this magic number to a const someday 👀

src/util/Options.js
```diff
@@ -135,7 +135,6 @@ class Options extends null {
         version: 9,
       },
       http: {
-        agent: {},
```
vladfrangu, line 138: Why was this removed?

vladfrangu (changes requested):

src/client/Client.js
ImRodry: Personally I don’t think it looks very good on 1000 but other 4-digit numbers bigger than that are fine imo. I just ended up changing all because some places had 1000 and other numbers so it would seem odd for one to have it and not the others

kyranet: My thoughts are the same.

src/structures/interfaces/TextBasedChannel.js
ImRodry: What even is this number?

src/util/Options.js
ImRodry: Probably some conflicts I came across when doing this. Thought I undid them all bug I’ll add this back

src/structures/interfaces/TextBasedChannel.js
ckohen: 2 weeks IIRC

ImRodry: Could probably make this a constant inside this function then, since it’s not used anywhere and wouldn’t be out of the scope of the PR iirc

src/util/Options.js
ImRodry: Amended the commit

src/structures/interfaces/TextBasedChannel.js
iCrawl: Nah. We good.

src/structures/ClientUser.js
```diff
@@ -25,8 +25,8 @@ class ClientUser extends User {
        * @type {?boolean}
        */
       this.mfaEnabled = typeof data.mfa_enabled === 'boolean' ? data.mfa_enabled : null;
-    } else if (typeof this.mfaEnabled === 'undefined') {
-      this.mfaEnabled = null;
+    } else {
+      this.mfaEnabled ??= null;
```
xiJulian, line 29: i think that this should be left unchanged

ImRodry: Why

xiJulian: nvm seems like i didn't understand what does the nullish assignment actually do

ImRodry: This only assigns the value if the property on the left-hand side is nullable

SpaceEEC (approved):

kyranet (approved):

ImRodry:
Cherry-picked the commit from #6544 since this PR was gonna touch that file anyway and it doesn't hurt to change this

src/client/websocket/WebSocketManager.js
iCrawl: Then it's time to get familiar with it, I don't think this is out of scope at all, considering you just cherry-picked something completely out of scope yourself

ImRodry: Alright I’ll look into it in a bit, would also appreciate if you could suggest the changes and I’ll add the necessary imports above.

ImRodry: Addressed in the latest commit and also removed the out of scope commit, let me know if that works

iCrawl (approved):

PR #6551, opened 2021-08-29
fix(xxxable): follow more properly with discord behavior

tignear:
In rare cases, the result returned by the current XXXable property may not return the exact result, but it will return the correct result.
However, it seems difficult to maintain this, so consider removing this property. 

- Cannot delete channels that are not viewable/connectable(have CONNECT permission).
- ~~Cannot edit messages on channels that cannot send messages.~~
  - Cannot edit messages on channels that cannot view messages.
- Cannot pin messages on channels that cannot view messages.
- Cannot delete messages on channels that cannot view messages.
- ~~Messages that are not system messages can be pinned.~~
  - #6585 

### DIFF
--- a/src/structures/Message.js
+++ b/src/structures/Message.js
@@ -548,7 +548,11 @@ class Message extends Base {
    * @readonly
    */
   get editable() {
-    return this.author.id === this.client.user.id;
+    return (
+      this.author.id === this.client.user.id &&
+      (!this.guild ||
+        (this.channel.viewable && this.channel.permissionsFor(this.client.user).has(Permissions.FLAGS.SEND_MESSAGES)))
+    );
   }
 
   /**
@@ -557,10 +561,18 @@ class Message extends Base {
    * @readonly
    */
   get deletable() {
-    return Boolean(
-      !this.deleted &&
-        (this.author.id === this.client.user.id ||
-          this.channel.permissionsFor?.(this.client.user)?.has(Permissions.FLAGS.MANAGE_MESSAGES)),
+    if (this.deleted) {
+      return false;
+    }
+    if (!this.guild) {
+      return this.author.id === this.client.user.id;
+    }
+    if (!this.channel.viewable) {
+      return false;
+    }
+    return (
+      this.author.id === this.client.user.id ||
+      this.channel.permissionsFor(this.client.user).has(Permissions.FLAGS.MANAGE_MESSAGES, false)
     );
   }
 
@@ -570,9 +582,11 @@ class Message extends Base {
    * @readonly
    */
   get pinnable() {
-    return Boolean(
+    return (
       !this.system &&
-        (!this.guild || this.channel.permissionsFor(this.client.user)?.has(Permissions.FLAGS.MANAGE_MESSAGES, false)),
+      (!this.guild ||
+        (this.channel.viewable &&
+          this.channel.permissionsFor(this.client.user).has(Permissions.FLAGS.MANAGE_MESSAGES, false)))
     );
   }
 
--- a/src/structures/Message.js
+++ b/src/structures/Message.js
@@ -550,6 +550,7 @@ class Message extends Base {
   get editable() {
     return (
       this.author.id === this.client.user.id &&
+      !this.deleted &&
       (!this.guild ||
         (this.channel.viewable && this.channel.permissionsFor(this.client.user).has(Permissions.FLAGS.SEND_MESSAGES)))
     );
@@ -584,6 +585,7 @@ class Message extends Base {
   get pinnable() {
     return (
       !this.system &&
+      !this.deleted &&
       (!this.guild ||
         (this.channel.viewable &&
           this.channel.permissionsFor(this.client.user).has(Permissions.FLAGS.MANAGE_MESSAGES, false)))
@@ -616,7 +618,8 @@ class Message extends Base {
       this.channel.viewable &&
       this.channel.permissionsFor(this.client.user).has(Permissions.FLAGS.SEND_MESSAGES) &&
       (this.author.id === this.client.user.id ||
-        this.channel.permissionsFor(this.client.user).has(Permissions.FLAGS.MANAGE_MESSAGES))
+        this.channel.permissionsFor(this.client.user).has(Permissions.FLAGS.MANAGE_MESSAGES)) &&
+      !this.deleted
     );
   }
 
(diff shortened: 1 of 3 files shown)


tignear:
Which should I use `permissionsFor().has(X)` or `permissionsFor().has(X,false)`?
This is work as same(It seems that I was confused when I created f28178c).

src/structures/Message.js
```diff
@@ -548,7 +548,11 @@ class Message extends Base {
    * @readonly
    */
   get editable() {
-    return this.author.id === this.client.user.id;
+    return (
+      this.author.id === this.client.user.id &&
```
DTrombett, line 552: Shouldn't this (and `pinnable`) check if the message has been deleted too (like on `deletable`)?
```suggestion
      this.author.id === this.client.user.id && !this.deleted &&
```

NotSugden (changes requested):
`GuildChannel#permissionsFor` can return null if the member for the client user is uncached

src/structures/Message.js
```diff
@@ -557,10 +561,18 @@ class Message extends Base {
...
-          this.channel.permissionsFor?.(this.client.user)?.has(Permissions.FLAGS.MANAGE_MESSAGES)),
+    if (this.deleted) {
+      return false;
+    }
+    if (!this.guild) {
+      return this.author.id === this.client.user.id;
+    }
+    if (!this.channel.viewable) {
+      return false;
+    }
```
ckohen, line 572:
```suggestion
    if (this.deleted || !this.channel?.viewable) {
      return false;
    }
    if (!this.guild) {
      return this.author.id === this.client.user.id;
    }
```
These can be combined, and channel is nullable

ckohen (changes requested):

src/structures/Message.js
```diff
@@ -572,7 +572,7 @@ class Message extends Base {
     }
     return (
       this.author.id === this.client.user.id ||
-      this.channel.permissionsFor(this.client.user).has(Permissions.FLAGS.MANAGE_MESSAGES, false)
+      this.channel.permissionsFor(this.client.user).has(Permissions.FLAGS.MANAGE_MESSAGES)
```
ckohen: Please revert this, this override is required due to something consider not a bug by discord (see https://github.com/discordjs/discord.js/pull/2756)

src/structures/Message.js
```diff
@@ -585,8 +585,7 @@ class Message extends Base {
     return (
       !this.system &&
       (!this.guild ||
-        (this.channel.viewable &&
-          this.channel.permissionsFor(this.client.user).has(Permissions.FLAGS.MANAGE_MESSAGES, false)))
+        (this.channel.viewable && this.channel.permissionsFor(this.client.user).has(Permissions.FLAGS.MANAGE_MESSAGES)))
```
ckohen: ditto

ckohen (changes requested):

src/structures/Message.js
tignear: DMChannel does not have a "viewable" property.
This is not combined.

ckohen: You'd have the same problem with the existing code then

tignear: If "existing" is meaning b75073c, fixed at 0b6a55d.
If "existing" is meaning that code below, why?
https://github.com/discordjs/discord.js/blob/main/src/structures/Message.js#L560-L564

ckohen: Yeah that's fine. You should add some comments explaining why the ordering is such, its not intuitive that's whats happening, someone else may accidentally combine them in the future.

tignear: Added. Thank you.

DraftProducts (approved):

iCrawl (approved):

kyranet (approved):

src/structures/Message.js
```diff
@@ -548,7 +548,13 @@ class Message extends Base {
    * @readonly
    */
   get editable() {
-    return this.author.id === this.client.user.id;
+    return Boolean(
+      this.author.id === this.client.user.id &&
+        !this.deleted &&
+        (!this.guild ||
+          (this.channel?.viewable &&
+            this.channel.permissionsFor(this.client.user)?.has(Permissions.FLAGS.SEND_MESSAGES, false))),
```
SpaceEEC, line 556: > Cannot edit messages on channels that cannot send messages.

You can do this just fine.

src/structures/Message.js
```diff
@@ -595,14 +613,16 @@ class Message extends Base {
...
-      this.channel.type === 'GUILD_NEWS' &&
-      !this.flags.has(MessageFlags.FLAGS.CROSSPOSTED) &&
-      this.type === 'DEFAULT' &&
-      this.channel.viewable &&
-      this.channel.permissionsFor(this.client.user).has(Permissions.FLAGS.SEND_MESSAGES) &&
-      (this.author.id === this.client.user.id ||
-        this.channel.permissionsFor(this.client.user).has(Permissions.FLAGS.MANAGE_MESSAGES))
+    const bitfield =
+      Permissions.FLAGS.SEND_MESSAGES |
+      (this.author.id === this.client.user.id ? 0 : Permissions.FLAGS.MANAGE_MESSAGES);
```
SpaceEEC, line 618: This throws:
`TypeError: Cannot mix BigInt and other types, use explicit conversions`

```suggestion
      (this.author.id === this.client.user.id ? Permissions.defaultBit : Permissions.FLAGS.MANAGE_MESSAGES);
```

src/structures/VoiceChannel.js
```diff
@@ -8,22 +8,13 @@ const Permissions = require('../util/Permissions');
...
-  get deletable() {
-    return super.deletable && this.permissionsFor(this.client.user).has(Permissions.FLAGS.CONNECT, false);
-  }
-
   /**
    * Whether the channel is editable by the client user
    * @type {boolean}
    * @readonly
    */
   get editable() {
```
SpaceEEC, line 16: Probably out of scope for this PR, but we should deprecate this getter.
Having both `VoiceChannel#editable` and `VoiceChannel#manageable` is plain confusing.

SpaceEEC (changes requested):

tignear:
Partial testing is here: https://github.com/tignear/discord.js-pr-6551-test/blob/master/main.test.js

SpaceEEC (approved):

PR #6553, opened 2021-08-29
chore: clean up tsconfig.json

thehackerboi69github:
Cleans up `tsconfig.json`:
- `declaration`, `removeComments` and `sourceMap` are false by default and were explicitly set to false here too, so I have removed them.
- `alwaysStrict` is true by default if `strict` is set, and it was explicitly set to true here too, so I have removed it.
- `skipDefaultLibCheck` is now deprecated as it has been renamed to `skipLibCheck`, so I have updated it to `skipLibCheck`.
- Specifying `esnext` in the `lib` option includes all `esnext` typings, all other `esnext` options are unnecessary, so I have removed them.

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -2,14 +2,10 @@
   "compilerOptions": {
     "strict": true,
     "moduleResolution": "node",
-    "declaration": false,
-    "removeComments": false,
-    "alwaysStrict": true,
     "pretty": false,
     "module": "commonjs",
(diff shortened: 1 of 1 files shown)


iCrawl:
The tsconfig isn't actually really used for anything nor really relevant right now, I'd like to hold off on changing or touching it until we actually need it, in which case well use the one from the builders repo.

PR #6560, opened 2021-08-29
workflow: issue forms

almostSouji:

Users oftentimes ignore the issue template, no longer! [Issue Forms](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/syntax-for-issue-forms) to the rescue! (note: labels are not present on my fork, hence missing in below screenshots)

<details>
<summary>Bug Report example</summary>

image

image

</details>

<details>
<summary>Feature Request example</summary>

image

</details>


vladfrangu (approved):
Me likey

iCrawl (approved):

ImRodry:
For the issue template, are you able to automatically put something in the text box so that users already get the code block?
Also idk if this is possible but could you have a list with all the intents and partials and let users select multiple items from that list?

almostSouji:
> For the issue template, are you able to automatically put something in the text box so that users already get the code block?

`render: javascript` renders it as javascript codeblock, as the hint says, no need to put backticks.

> Also idk if this is possible but could you have a list with all the intents and partials and let users select multiple items from that list?

Only with checkboxes, which i'm not a fan of, as they are used for task lists on GitHub.

ImRodry:
> render: javascript renders it as javascript codeblock, as the hint says, no need to put backticks.

Ah mb didn't see that. You can probably move that text to the text box hint, so that it's easier for people to see
> Only with checkboxes, which i'm not a fan of, as they are used for task lists on GitHub.

So you can't have a list like the priority list in which you can pick multiple items?

Jiralite:
Can users specify to produce TypeScript code instead (in the case of a TypeScript-related issue), or is this not a problem? Just making sure!

almostSouji:
> Can users specify to produce TypeScript code instead (in the case of a TypeScript-related issue), or is this not a problem? Just making sure!

This would require a second codeblock element, The feature is in a early beta, there is no way to design flows at this point in time (select js -> get js, select ts -> get ts cb). So if people use ts it will be rendered as js codeblock at the moment.

> So you can't have a list like the priority list in which you can pick multiple items?

I suppose I overlooked this [feature](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/syntax-for-githubs-form-schema#dropdown), but i'm not too sure how beneficial it is in this case

ImRodry:
> I suppose I overlooked this feature, but i'm not too sure how beneficial it is in this case

Only drawback I can see is having to update the list every time a partial/intent is added, but as long as reviewers don't forget that I don't see why not

almostSouji:
Multi selects are nicer than anticipated

image

I made the code sample clear by... well, placeholder'ing a code sample

image

ImRodry:
Looks fancy! Just to be sure, that placeholder will disappear once the user starts typing something right?

almostSouji:
That is in fact how placeholders operate.

ImRodry (approved):
Would also be nice to see this in PRs once GitHub adds the feature!

vBread:
> This would require a second code block element, The feature is in an early beta, there is no way to design flows at this point in time (select js -> get js, select ts -> get ts CB). So if people use ts it will be rendered as js code block at the moment. 

Couldn't you use TS since TS highlighting uses JS?

PR #3822, opened 2020-02-24
typings(MessageEmbed): mark properties as able to be undefined

NotSugden:
**Please describe the changes this PR makes and why it should be merged:**
This PR marks the color, description, provider, title, and url properties as able to be undefined, since they are able to be undefined

**Status**
- [x] Code changes have been tested against the Discord API, or there are no code changes
- [x] I know how to update typings and have done so, or typings don't need updating

**Semantic versioning classification:**  
- [ ] This PR changes the library's interface (methods or parameters added)
  - [ ] This PR includes breaking changes (methods removed or renamed, parameters moved or removed)
- [] This PR **only** includes non-code changes, like changes to documentation, README, etc.

PR #6563, opened 2021-08-29
typings(MessageReference): correctly add undefined type

ImRodry:
This PR fixes the types for MessageReference as guildId can be undefined and messageId can be undefined, not null.

### DIFF
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -4289,8 +4289,8 @@ export type MessageReactionResolvable =
 
 export interface MessageReference {
   channelId: Snowflake;
-  guildId: Snowflake;
-  messageId: Snowflake | null;
+  guildId: Snowflake | undefined;
+  messageId: Snowflake | undefined;
 }
 
 export type MessageResolvable = Message | Snowflake;


typings/index.d.ts
```diff
@@ -4289,8 +4289,8 @@ export type MessageReactionResolvable =
 
 export interface MessageReference {
   channelId: Snowflake;
-  guildId: Snowflake;
-  messageId: Snowflake | null;
+  guildId: Snowflake | undefined;
+  messageId: Snowflake | undefined;
```
NotSugden, line 4293:
```suggestion
  guildId?: Snowflake;
  messageId?: Snowflake;
```

NotSugden (changes requested):

typings/index.d.ts
iCrawl: I'd rather go with the latter because: https://devblogs.microsoft.com/typescript/announcing-typescript-4-4/#exact-optional-property-types

ImRodry: I believe these should stay as undefined because all 3 properties will always be present, but be undefined if we don’t get that data from Discord. It’s something worth looking into for v14 but for now I think this would be more accurate.

kyranet (approved):

typings/index.d.ts
NotSugden: there may be other places this change needs applying then, i seem to remember a discussion a while ago about just prefering `?: ...` over `| undefined`

ImRodry: The two types mean different things though. `?:` is for missing parameters and undefined is for, well, undefined parameters. If you want I can go through the file to ensure these are correctly set, I believe that would still be within the scope of the PR

NotSugden: yeah i know they are different, i just seem to remember a discussion a while back about preferring `?:` for whatever reason

almeidx: Are you sure you're not confusing the jsdocs and the types?

NotSugden: [here] this is the discussion i'm talking about

if you view at [this commit](https://github.com/discordjs/discord.js/blob/967b533e9d5f08e2be50698fb98d38820d29d88b/src/structures/MessageEmbed.js) said properties are always present, but either a string or undefined

almeidx: > [here](https://github.com/discordjs/discord.js/pull/3822#pullrequestreview-363274200) this is the discussion i'm talking about

What @SpaceEEC is no longer valid as stated in the link iCrawl sent

iCrawl: Yes, but we will enable this option, so we def should go with what it is now or even do both. Preferably both.

```suggestion
  guildId?: Snowflake | undefined;
  messageId?: Snowflake | undefined;
```

ImRodry: > Yes, but we will enable this option, so we def should go with what it is now or even do both. Preferably both.

Why both though? They will never be actually missing

iCrawl: Ah, I read that wrong, we good then 👍

iCrawl (approved):

vladfrangu (approved):

SpaceEEC (approved):

PR #6571, opened 2021-08-31
types(MessageOptions): add missing typings

OliverHenrichs:
typings added for ephemeral and flags options

Signed-off-by: Oliver Henrichs <Oliver Henrichs>

When using MessageOptions in MessagePayload's constructor, the typings hide the fact that[ ephemeral-bool is available](https://github.com/discordjs/discord.js/blob/main/src/structures/MessagePayload.js#L150) and used.
Therefore typescript-programmers currently cannot create ephemeral MessagePayloads.

Regarding the other fix to flags I am unsure about the typings. This was my best guess after browsing the library for a while.
I thought it's better to try fix all the issues with typings for this interface.

### DIFF
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -4300,6 +4300,8 @@ export interface MessageOptions {
   reply?: ReplyOptions;
   stickers?: StickerResolvable[];
   attachments?: MessageAttachment[];
+  ephemeral?: boolean;
+  flags?: BitFieldResolvable<string, number | bigint>;
 }
 
 export type MessageReactionResolvable =


ImRodry:
Indeed you cannot make messages ephemeral and never have been. You also should not be editing your message flags either

OliverHenrichs:
> Indeed you cannot make messages ephemeral and never have been. You also should not be editing your message flags either

Then it's weird that MessagePayload is used in Interaction replies, which can be ephemeral. My usage of payloads in that context made me create these changes.

ImRodry:
> > Indeed you cannot make messages ephemeral and never have been. You also should not be editing your message flags either
> 
> Then it's weird that MessagePayload is used in Interaction replies, which can be ephemeral. My usage of payloads in that context made me create these changes.

The interface you're editing is used for sending regular text messages, so these changes are already incorrect. If you wanna be able to create `MessagePayload`s (no idea why in the world you would, but sure) for interactions then I guess the InteractionReplyOptions interface should be added to the MessagePayload constructor and these changes should be reverted.
And, like I said previously, editing flags has never been something the library supports as these are added in the code depending on what you're doing

OliverHenrichs:
> > > Indeed you cannot make messages ephemeral and never have been. You also should not be editing your message flags either
> > 
> > 
> > Then it's weird that MessagePayload is used in Interaction replies, which can be ephemeral. My usage of payloads in that context made me create these changes.
> 
> The interface you're editing is used for sending regular text messages, so these changes are already incorrect. If you wanna be able to create `MessagePayload`s (no idea why in the world you would, but sure) for interactions then I guess the InteractionReplyOptions interface should be added to the MessagePayload constructor and these changes should be reverted.
> And, like I said previously, editing flags has never been something the library supports as these are added in the code depending on what you're doing

You are right.
And I just realized that your suggested solution is already implemented in InteractionReplyOptions.
Sorry for the inconvenience.

PR #6576, opened 2021-09-01
feat(ClientOptions): waitGuildTimeout amount client option

kylerchin:

Sometimes bots hang while connecting for 15 seconds because 1 single guild is messed up. For bots with Continuous integration, this can create significant impacts to uptime, user experience, and lag.

This adds the optional `fetchGuildTimeout` option under ClientOptions, where the bot developer can optionally specify how long the client should hang for. If not specified, this will remain the default 15 seconds.

I'm happy to discuss how this can be improved, including the names of the options and variables, and I'd also be happy to write documentation!

👉🏻👈🏻 my first pr to this repo, I'm honoured I can be a part of this community.

EDIT: I'm wondering if the "in X sec" should be changed to "in X ms".  


kylerchin:
P.S. I've added this to my own bot, cutting the startup time in half in 1 line of code
Before
Before
image
After
After

src/client/websocket/WebSocketShard.js
```diff
@@ -475,12 +475,22 @@ class WebSocketShard extends EventEmitter {
...
+    // Step 2. Create a timeout that will mark the shard as ready if there are still unavailable guilds
+    // * the timeout should be 15 seconds by default, but this can be optionally be changed in the client options
+
+    var guildTimeoutAmount = 15000;
+
+    if (this.manager.client.options.fetchGuildTimeout) {
+      guildTimeoutAmount = this.manager.client.options.fetchGuildTimeout;
+    }
+
+    var guildTimeoutAmountInSeconds = guildTimeoutAmount / 1000;
```
DTrombett, line 487: 🧹
```suggestion
    const guildTimeoutAmount = this.manager.client.options.fetchGuildTimeout || 1500;
```

src/client/websocket/WebSocketShard.js
```diff
@@ -475,12 +475,22 @@ class WebSocketShard extends EventEmitter {
...
+
     this.readyTimeout = setTimeout(
       () => {
         this.debug(
           `Shard ${hasGuildsIntent ? 'did' : 'will'} not receive any more guild packets` +
-            `${hasGuildsIntent ? ' in 15 seconds' : ''}.\n   Unavailable guild count: ${this.expectedGuilds.size}`,
+            `${hasGuildsIntent ? `in ${guildTimeoutAmountInSeconds} seconds` : ''}.\n   Unavailable guild count: ${this.expectedGuilds.size}`,
```
DTrombett, line 493:
```suggestion
            `${hasGuildsIntent ? `in ${Math.round(guildTimeoutAmount / 1000)} seconds` : ''}.\n   Unavailable guild count: ${this.expectedGuilds.size}`,
```

DTrombett (commented):
This is also missing documentation and validation in the Client class.

thehackerboi69github:
Wouldn't `shardReadyTimeout` be a more appropriate name for this?

src/client/websocket/WebSocketShard.js
ckohen: You sure you want 0 to be treated as default? If not use ??

NotSugden: what about setting the default here
https://github.com/discordjs/discord.js/blob/d16ada97087e0324134484c9c0724b52a2839bdf/src/util/Options.js#L110-L146

ckohen: I mean yes that would be preferred once validation is added 😎

DTrombett: > You sure you want 0 to be treated as default? If not use ??

If `15000` is added as a default option this isn't needed as `this.manager.client.options.fetchGuildTimeout` can be used... However, why you should use 0? Does this mean no timeout at all or no guilds at all?

src/client/websocket/WebSocketShard.js
kylerchin: What if we just used ms?

kylerchin:
I've tried my best to make the recommended changes, let me know if we need to change anything else or we have any other thoughts about this before merging.

ImRodry:
Yeah I’m pretty sure that text you added goes over the 120 char limit, you should install ESLint to be aware of these things since the CI requires approval to run

src/client/websocket/WebSocketShard.js
```diff
@@ -475,12 +475,17 @@ class WebSocketShard extends EventEmitter {
...
+    // * a timeout time of zero will skip this timeout
+
+    const guildTimeoutAmount = this.manager.client.options.fetchGuildTimeout || 1500;
+
     this.readyTimeout = setTimeout(
       () => {
         this.debug(
           `Shard ${hasGuildsIntent ? 'did' : 'will'} not receive any more guild packets` +
-            `${hasGuildsIntent ? ' in 15 seconds' : ''}.\n   Unavailable guild count: ${this.expectedGuilds.size}`,
+            `${hasGuildsIntent ? `in ${guildTimeoutAmount} milliseconds` : ''}.\n   Unavailable guild count: ${this.expectedGuilds.size}`,
```
thehackerboi69github, line 488:
```suggestion
            `${hasGuildsIntent ? ` in ${guildTimeoutAmount} milliseconds` : ''}.\n   Unavailable guild count: ${this.expectedGuilds.size}`,
```

ImRodry: Should probably also say ms instead of milliseconds

kylerchin: > Yeah I’m pretty sure that text you added goes over the 120 char limit, you should install ESLint to be aware of these things since the CI requires approval to run

thank you! I've corrected this issue now and it passed the lint :) @ImRodry 
I've learned a lot and I hope this PR is helpful to all discord js bot devs!

src/client/websocket/WebSocketShard.js
```diff
@@ -475,12 +475,19 @@ class WebSocketShard extends EventEmitter {
...
+
+    const guildTimeoutAmount = this.manager.client.options.fetchGuildTimeout || 1500;
+
     this.readyTimeout = setTimeout(
       () => {
         this.debug(
           `Shard ${hasGuildsIntent ? 'did' : 'will'} not receive any more guild packets` +
-            `${hasGuildsIntent ? ' in 15 seconds' : ''}.\n   Unavailable guild count: ${this.expectedGuilds.size}`,
+            `${hasGuildsIntent ? ` in ${guildTimeoutAmount} ms` : ''}.\n` +
+            `Unavailable guild count: ${this.expectedGuilds.size}`,
```
ImRodry, line 490: I think this isn't allowed either. Just format with prettier and it will adjust it for you

kylerchin: Thank you, i will try this in a little bit.

src/client/websocket/WebSocketShard.js
DTrombett: Well, it's the same. I personally think that if the option is in ms maybe the message logged should be in ms too but it's better that official djs collaborators decide lol

ImRodry: The option should probably be in seconds instead since you're not gonna want nor need a timeout of less than 1 second

DTrombett:
You can add the default value for this option in the `Options.createDefault()` static method:
https://github.com/discordjs/discord.js/blob/stable/src/util/Options.js#L102
And add validation in the `Client#_validateOptions()` method:
https://github.com/discordjs/discord.js/blob/stable/src/client/Client.js#L542

kylerchin:
I'm debating if this should be titled "fetchGuildTimeout" or "fetchGuildsTimeout" or something else. I'm happy to discuss this and improve the experience for developers.

src/client/websocket/WebSocketShard.js
kylerchin: I've used the prettier version! it looks so much better :) thank you

src/client/websocket/WebSocketShard.js
kylerchin: I think it's kinda weird to have decimals in the options but I'm not entirely sure. I think the console log and options should be the same formatting (in milliseconds) to make the experience smoother for devs.

ImRodry: But you shouldn’t need decimals for the same reason you shouldn’t need milliseconds. Your guilds will never arrive that fast so it’s unreasonable to set a lower timeout. If the dev REALLY wants that then they can use decimals, but I don’t see why they would

src/client/websocket/WebSocketShard.js
```diff
@@ -475,12 +475,20 @@ class WebSocketShard extends EventEmitter {
       return;
     }
     const hasGuildsIntent = new Intents(this.manager.client.options.intents).has(Intents.FLAGS.GUILDS);
-    // Step 2. Create a 15s timeout that will mark the shard as ready if there are still unavailable guilds
+    // Step 2. Create a timeout that will mark the shard as ready if there are still unavailable guilds
+    // * the timeout should be 15 seconds by default
+    // * this can be optionally be changed in the client options via fetchGuildTimeout
+    // * a timeout time of zero will skip this timeout
+
+    const guildTimeoutAmount = this.manager.client.options.fetchGuildTimeout || 1500;
```
monbrey, line 483:
```suggestion
    const guildTimeoutAmount = this.manager.client.options.fetchGuildTimeout || 15000;
```

src/client/websocket/WebSocketShard.js
kylerchin: Thank you for pointing that out. I was looking at the client options and the request-based ones all use milliseconds. Devs are already quite familiar with setTimeout in JS and other similar stuff being based on milliseconds.

Does this change have to be fulfilled for the option to be merged into the repo?

src/util/Options.js
```diff
@@ -69,6 +69,9 @@
  * [User Agent](https://discord.com/developers/docs/reference#user-agent) header
  * @property {PresenceData} [presence={}] Presence data to use upon login
  * @property {IntentsResolvable} intents Intents to enable for this connection
+ * @property {number} [fetchGuildTimeout=15000] Time in milliseconds that Clients with the GUILDS intent should hang
+ * and wait for missing guilds to be recieved before starting the bot.
+ * If not specified, the default is 15 seconds.
```
thehackerboi69github, line 74:
```suggestion
 * @property {number} [fetchGuildTimeout=15000] Time in milliseconds to wait for missing guilds to be received before marking as ready (0 to skip this timeout)
```

src/client/websocket/WebSocketShard.js
ImRodry: Yeah that’s true. I’m not a maintainer I’m just giving my personal opinion so let’s wait for one of the @discordjs/the-big-3 to say something about this

src/util/Options.js
ImRodry: The only good thing about this change is that it mentions that 0 skips the timeout. Other than that, why??
You removed "Clients with the GUILDS intent" and "If not specified, the default is 15 seconds" which imo are quite important mentions, what's the purpose of this?

src/client/websocket/WebSocketShard.js
```diff
@@ -475,12 +475,20 @@ class WebSocketShard extends EventEmitter {
       return;
     }
     const hasGuildsIntent = new Intents(this.manager.client.options.intents).has(Intents.FLAGS.GUILDS);
-    // Step 2. Create a 15s timeout that will mark the shard as ready if there are still unavailable guilds
+    // Step 2. Create a timeout that will mark the shard as ready if there are still unavailable guilds
+    // * the timeout should be 15 seconds by default
```
vladfrangu, line 479:
```suggestion
    // * The timeout is 15 seconds by default
```

src/client/websocket/WebSocketShard.js
```diff
@@ -475,12 +475,20 @@ class WebSocketShard extends EventEmitter {
...
+    // * this can be optionally be changed in the client options via fetchGuildTimeout
```
vladfrangu, line 480:
```suggestion
    // * This can be optionally be changed in the client options via the `fetchGuildTimeout` option
```

src/client/websocket/WebSocketShard.js
```diff
@@ -475,12 +475,20 @@ class WebSocketShard extends EventEmitter {
...
+    // * a timeout time of zero will skip this timeout
```
vladfrangu, line 481: While this is documentation in internal code, users shouldn't use a timeout of 0 without acknowledging the difference it can make

src/client/websocket/WebSocketShard.js
vladfrangu: Also this default should be in the client option, so no need for `?? 15_000`

src/util/Options.js
```diff
@@ -124,6 +128,7 @@ class Options extends null {
       failIfNotExists: true,
       userAgentSuffix: [],
       presence: {},
+      fetchGuildTimeout: 15000,
```
vladfrangu, line 131:
```suggestion
      fetchGuildTimeout: 15_000,
```

src/util/Options.js
```diff
@@ -69,6 +69,10 @@
  * [User Agent](https://discord.com/developers/docs/reference#user-agent) header
  * @property {PresenceData} [presence={}] Presence data to use upon login
  * @property {IntentsResolvable} intents Intents to enable for this connection
+ * @property {number} [fetchGuildTimeout=15000] Time in milliseconds that Clients with the GUILDS intent should hang
```
vladfrangu, line 72:
```suggestion
 * @property {number} [fetchGuildTimeout=15000] Time in milliseconds that clients with the `GUILDS` intent should hang
```

vladfrangu (dismissed):

src/util/Options.js
```diff
@@ -69,6 +69,10 @@
...
+ * and wait for missing guilds to be recieved before starting the bot
```
thehackerboi69github, line 73:
```suggestion
 * and wait for missing guilds to be received before starting the bot
```

src/client/websocket/WebSocketShard.js
```diff
@@ -475,12 +475,20 @@ class WebSocketShard extends EventEmitter {
       return;
     }
     const hasGuildsIntent = new Intents(this.manager.client.options.intents).has(Intents.FLAGS.GUILDS);
-    // Step 2. Create a 15s timeout that will mark the shard as ready if there are still unavailable guilds
+    // Step 2. Create a timeout that will mark the shard as ready if there are still unavailable guilds
+    // * The timeout is 15 seconds by default
+    // * This can be optionally be changed in the client options via the `fetchGuildTimeout` option
+    // * a timeout time of zero will skip this timeout, which potentially could cause the Client to miss guilds.
+
+    const guildTimeoutAmount = this.manager.client.options.fetchGuildTimeout ?? 15000;
```
thehackerboi69github, line 483:
```suggestion
    const guildTimeoutAmount = this.manager.client.options.fetchGuildTimeout;
```

thehackerboi69github: Defaults will be assigned in the client constructor when not specified, so there's no need to specify it here as well

thehackerboi69github (changes requested):

src/util/Options.js
```diff
@@ -69,6 +69,10 @@
  * [User Agent](https://discord.com/developers/docs/reference#user-agent) header
  * @property {PresenceData} [presence={}] Presence data to use upon login
  * @property {IntentsResolvable} intents Intents to enable for this connection
+ * @property {number} [fetchGuildTimeout=15000] Time in milliseconds that clients with the `GUILDS` intent should hang
+ * and wait for missing guilds to be received before starting the bot
+ * If not specified, the default is 15 seconds.
```
thehackerboi69github, line 74: Is this line really needed? The `[fetchGuildTimeout=15000]` in L72 will already show the default in the docs.

src/util/Options.js
```diff
@@ -69,6 +69,10 @@
  * [User Agent](https://discord.com/developers/docs/reference#user-agent) header
  * @property {PresenceData} [presence={}] Presence data to use upon login
  * @property {IntentsResolvable} intents Intents to enable for this connection
+ * @property {number} [fetchGuildTimeout=15000] Time in milliseconds that clients with the `GUILDS` intent should hang
```
1chiSensei, line 72:
```suggestion
 * @property {number} [fetchGuildTimeout=15000] Time in milliseconds that bots with the `GUILDS` intent should hang
```

Shouldn't be this called bots instead of clients?

1chiSensei (changes requested):

src/util/Options.js
ImRodry: Not at all

src/util/Options.js
```diff
@@ -69,6 +69,10 @@
  * [User Agent](https://discord.com/developers/docs/reference#user-agent) header
  * @property {PresenceData} [presence={}] Presence data to use upon login
  * @property {IntentsResolvable} intents Intents to enable for this connection
+ * @property {number} [fetchGuildTimeout=15000] Time in milliseconds that bots with the `GUILDS` intent should hang
+ * and wait for missing guilds to be received before starting the bot
+ * If not specified, the default is 15 seconds.
+ * (0 to skip this timeout)
```
SpaceEEC, line 75: A few non-binding rephrasing suggestions:
- `hang and wait`: seems redundant
- `missing`: feels a bit misleading (might just be me)
- `starting the bot`: `Client#login` starts it. It's just not ready yet, a small nit.
- `skip this timeout`: `not wait` seems to be a bit more concise

```suggestion
 * @property {number} [waitGuildTimeout=15000] Time in milliseconds that bots with the `GUILDS` intent should wait for
 * initial guild's data to be received before emitting ready. (0 to not wait)
```

SpaceEEC (commented):
Maybe rename the option to `waitGuildTimeout` or something the like that?
I think that `fetch` might be a bit misleading as this isn't something we directly "requested" in that sense, that's just part of the login and ready flow. (Probably not a good explanation, but I can't come up with anything better at the moment)

src/client/websocket/WebSocketShard.js
```diff
@@ -475,12 +475,20 @@ class WebSocketShard extends EventEmitter {
       return;
     }
     const hasGuildsIntent = new Intents(this.manager.client.options.intents).has(Intents.FLAGS.GUILDS);
-    // Step 2. Create a 15s timeout that will mark the shard as ready if there are still unavailable guilds
+    // Step 2. Create a timeout that will mark the shard as ready if there are still unavailable guilds
+    // * The timeout is 15 seconds by default
+    // * This can be optionally be changed in the client options via the `fetchGuildTimeout` option
+    // * a timeout time of zero will skip this timeout, which potentially could cause the Client to miss guilds.
+
+    const guildTimeoutAmount = this.manager.client.options.fetchGuildTimeout;
```
thehackerboi69github:
```suggestion
    const { waitGuildTimeout } = this.manager.client.options;
```

src/client/websocket/WebSocketShard.js
```diff
@@ -475,12 +475,20 @@ class WebSocketShard extends EventEmitter {
...
+
     this.readyTimeout = setTimeout(
       () => {
         this.debug(
           `Shard ${hasGuildsIntent ? 'did' : 'will'} not receive any more guild packets` +
-            `${hasGuildsIntent ? ' in 15 seconds' : ''}.\n   Unavailable guild count: ${this.expectedGuilds.size}`,
+            `${hasGuildsIntent ? ` in ${guildTimeoutAmount} ms` : ''}.\nUnavailable guild count: ${
```
thehackerboi69github, line 489:
```suggestion
            `${hasGuildsIntent ? ` in ${waitGuildTimeout} ms` : ''}.\nUnavailable guild count: ${
```

src/client/websocket/WebSocketShard.js
```diff
@@ -489,7 +497,7 @@ class WebSocketShard extends EventEmitter {
 
         this.emit(ShardEvents.ALL_READY, this.expectedGuilds);
       },
-      hasGuildsIntent ? 15_000 : 0,
+      hasGuildsIntent ? guildTimeoutAmount : 0,
```
thehackerboi69github, line 500:
```suggestion
      hasGuildsIntent ? waitGuildTimeout : 0,
```

typings/index.d.ts
```diff
@@ -3353,6 +3353,7 @@ export interface ClientOptions {
   userAgentSuffix?: string[];
   presence?: PresenceData;
   intents: BitFieldResolvable<IntentsString, number>;
+  fetchGuildTimeout?: number;
```
thehackerboi69github:
```suggestion
  waitGuildTimeout?: number;
```

kylerchin:
@SpaceEEC the changes have been applied

thehackerboi69github (changes requested):

src/util/Options.js
```diff
@@ -69,6 +69,7 @@
  * [User Agent](https://discord.com/developers/docs/reference#user-agent) header
  * @property {PresenceData} [presence={}] Presence data to use upon login
  * @property {IntentsResolvable} intents Intents to enable for this connection
+ * @property {number} [waitGuildTimeout=15000] Time in milliseconds that Clients with the GUILDS intent should wait for missing guilds to be recieved before starting the bot. If not specified, the default is 15 seconds.
```
SpaceEEC, line 72: To comply with the line length limit:
```suggestion
 * @property {number} [waitGuildTimeout=15000] Time in milliseconds that Clients with the GUILDS intent should wait for
 * missing guilds to be recieved before starting the bot. If not specified, the default is 15 seconds.
```

src/util/Options.js
```diff
@@ -115,8 +116,8 @@ class Options extends null {
       messageSweepInterval: 0,
       invalidRequestWarningInterval: 0,
       partials: [],
-      restWsBridgeTimeout: 5_000,
-      restRequestTimeout: 15_000,
+      restWsBridgeTimeout: 5000,
+      restRequestTimeout: 15000,
```
SpaceEEC, line 120: I don't think we want to revert those:
```suggestion
      restWsBridgeTimeout: 5_000,
      restRequestTimeout: 15_000,
```

SpaceEEC (changes requested):
You also need to add the documented default for `waitGuildTimeout` in `Options.createDefault`, otherwise an error will be thrown if omitted.

SpaceEEC (approved):

src/client/websocket/WebSocketShard.js
```diff
@@ -475,12 +475,20 @@ class WebSocketShard extends EventEmitter {
       return;
     }
     const hasGuildsIntent = new Intents(this.manager.client.options.intents).has(Intents.FLAGS.GUILDS);
-    // Step 2. Create a 15s timeout that will mark the shard as ready if there are still unavailable guilds
+    // Step 2. Create a timeout that will mark the shard as ready if there are still unavailable guilds
+    // * The timeout is 15 seconds by default
+    // * This can be optionally be changed in the client options via the `waitGuildTimeout` option
+    // * a timeout time of zero will skip this timeout, which potentially could cause the Client to miss guilds.
```
vladfrangu, line 481: Setting this to 0 *might* break many things. Are we sure we should allow it?

vladfrangu (changes requested):

kyranet (approved):

kylerchin:
everything in here looks done! 

let me know if anything else needs to be changed, otherwise, this will be it! happy that this will get merged soon!

thehackerboi69github:
The default value is still not in  `Options.createDefault()`

Jiralite:
Check the reviews I submitted also.

src/util/Options.js
```diff
@@ -69,6 +69,8 @@
  * [User Agent](https://discord.com/developers/docs/reference#user-agent) header
  * @property {PresenceData} [presence={}] Presence data to use upon login
  * @property {IntentsResolvable} intents Intents to enable for this connection
+ * @property {number} [waitGuildTimeout=15_000] Time in milliseconds that Clients with the GUILDS intent should wait for
+ * missing guilds to be recieved before starting the bot. If not specified, the default is 15 seconds.
```
almeidx, line 73:
```suggestion
 * @property {number} [waitGuildTimeout=15_000] Time in milliseconds that Clients with the GUILDS intent should wait for
 * missing guilds to be received before starting the bot.
```

almeidx (changes requested):

src/util/Options.js
```diff
@@ -69,6 +69,8 @@
  * [User Agent](https://discord.com/developers/docs/reference#user-agent) header
  * @property {PresenceData} [presence={}] Presence data to use upon login
  * @property {IntentsResolvable} intents Intents to enable for this connection
+ * @property {number} [waitGuildTimeout=15000] Time in milliseconds that Clients with the GUILDS intent should wait for
+ * missing guilds to be recieved before starting the bot. If not specified, the default is 15 seconds.
```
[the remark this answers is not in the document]
almeidx: And the
> If not specified, the default is 15 seconds.

bit should be removed

thehackerboi69github:
What does.that last commit have to do with this PR?

kylerchin:
It fixes a bug in the discord name acronym system. Apologies.

src/structures/BaseGuild.js
```diff
@@ -61,10 +61,14 @@ class BaseGuild extends Base {
...
   get nameAcronym() {
-    return this.name
+    if (this.name) {
+      return this.name
       .replace(/'s /g, ' ')
       .replace(/\w+/g, e => e[0])
       .replace(/\s/g, '');
+    }  else {
+    return null;
+    }
```
kyranet, line 71: Please revert this, do not introduce unrelated changes in a PR.

kyranet (changes requested):

kylerchin:
I reverted the changes. Really sorry.

kylerchin:
Those changes have been moved to https://github.com/discordjs/discord.js/pull/7050/

kyranet (approved):

thehackerboi69github:
> The default value is still not in  `Options.createDefault()`

thebigwolfy (approved):

vladfrangu (changes requested):
The default is still not added in `Options.createDefault()`

kylerchin:
Apologies for not updating this, I will change it.

Sorry, i've been busy with finals.

kylerchin:
Again, i'm really sorry about the delay. I added it to options now

vladfrangu:
All I want for Christmas..issss... a rebaseeeeee 🎵

iCrawl (approved):

vladfrangu (approved):

PR #6585, opened 2021-09-03
fix(Message): pinnable returning false in non-DEFAULT messages

ImRodry:
This PR fixes a bug with Message#pinnable where it would return false if the type of message was not DEFAULT in some types that can be pinned. This PR fixes that so that it will now work on non system messages.

### DIFF
--- a/src/structures/Message.js
+++ b/src/structures/Message.js
@@ -570,10 +570,9 @@ class Message extends Base {
    * @readonly
    */
   get pinnable() {
-    return (
-      ['DEFAULT', 'REPLY', 'APPLICATION_COMMAND', 'THREAD_STARTER_MESSAGE'].includes(this.type) &&
-      (!this.guild || this.channel.permissionsFor(this.client.user).has(Permissions.FLAGS.MANAGE_MESSAGES, false)) &&
-      !this.system
+    return Boolean(
+      !this.system &&
+        (!this.guild || this.channel.permissionsFor(this.client.user)?.has(Permissions.FLAGS.MANAGE_MESSAGES, false))
     );
   }
 
--- a/src/structures/Message.js
+++ b/src/structures/Message.js
@@ -573,6 +573,7 @@ class Message extends Base {
     return (
       ['DEFAULT', 'REPLY', 'APPLICATION_COMMAND', 'THREAD_STARTER_MESSAGE'].includes(this.type) &&
       (!this.guild || this.channel.permissionsFor(this.client.user).has(Permissions.FLAGS.MANAGE_MESSAGES, false))
+      !this.system
     );
   }
 
--- a/src/structures/Message.js
+++ b/src/structures/Message.js
@@ -571,7 +571,7 @@ class Message extends Base {
    */
   get pinnable() {
     return (
-      this.type === 'DEFAULT' &&
+      ['DEFAULT', 'REPLY', 'APPLICATION_COMMAND', 'THREAD_STARTER_MESSAGE'].includes(this.type) &&
       (!this.guild || this.channel.permissionsFor(this.client.user).has(Permissions.FLAGS.MANAGE_MESSAGES, false))
     );
   }
--- a/src/structures/Message.js
+++ b/src/structures/Message.js
@@ -572,7 +572,7 @@ class Message extends Base {
   get pinnable() {
     return Boolean(
       !this.system &&
-        (!this.guild || this.channel.permissionsFor(this.client.user)?.has(Permissions.FLAGS.MANAGE_MESSAGES, false))
+        (!this.guild || this.channel.permissionsFor(this.client.user)?.has(Permissions.FLAGS.MANAGE_MESSAGES, false)),
     );
   }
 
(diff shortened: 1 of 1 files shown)


SpaceEEC (changes requested):
> If you have information that shows otherwise please review this PR mentioning that.

I just tested this and received an error:
`DiscordAPIError: Cannot execute action on a system message` (Code `50021`)

ImRodry:
> > If you have information that shows otherwise please review this PR mentioning that.
> 
> I just tested this and received an error:
> `DiscordAPIError: Cannot execute action on a system message` (Code `50021`)

Totally forgot about system messages, fixed in https://github.com/discordjs/discord.js/commits/95c0575e9bffb98e1e1b8a284b8d736f633b07ba @SpaceEEC

src/structures/Message.js
```diff
@@ -571,8 +571,9 @@ class Message extends Base {
    */
   get pinnable() {
     return (
-      this.type === 'DEFAULT' &&
-      (!this.guild || this.channel.permissionsFor(this.client.user).has(Permissions.FLAGS.MANAGE_MESSAGES, false))
+      ['DEFAULT', 'REPLY', 'APPLICATION_COMMAND', 'THREAD_STARTER_MESSAGE'].includes(this.type) &&
```
kyranet, line 574: You can probably make this a constant.

src/structures/Message.js
```diff
@@ -571,8 +571,9 @@ class Message extends Base {
...
+      (!this.guild || this.channel.permissionsFor(this.client.user).has(Permissions.FLAGS.MANAGE_MESSAGES, false)) &&
+      !this.system
```
kyranet, line 576: I'm confused as of why this is needed, when all it does is check whether or not `this.type` is one of a few:

https://github.com/discordjs/discord.js/blob/f57791c5feff1f9808369f8a7f95dab0900c6f0c/src/structures/Message.js#L69-L73

And we're already checking for that type in line 574 from this PR.

kyranet (changes requested):

src/structures/Message.js
```diff
@@ -571,8 +571,9 @@ class Message extends Base {
    */
   get pinnable() {
     return (
-      this.type === 'DEFAULT' &&
-      (!this.guild || this.channel.permissionsFor(this.client.user).has(Permissions.FLAGS.MANAGE_MESSAGES, false))
+      ['DEFAULT', 'REPLY', 'APPLICATION_COMMAND', 'THREAD_STARTER_MESSAGE'].includes(this.type) &&
+      (!this.guild || this.channel.permissionsFor(this.client.user).has(Permissions.FLAGS.MANAGE_MESSAGES, false)) &&
```
NotSugden, line 575: `permissionsFor(...)` can return null

NotSugden (changes requested):

src/structures/Message.js
ImRodry: Sorry I got confused by @SpaceEEC’s comment then. Space can you tell me how you got the error you mentioned above e.g. what message type it was?

src/structures/Message.js
ImRodry: How so?

NotSugden: if the client user's guildmember (`Guild#me`) is uncached then it will return null

ImRodry: I believe that can never be uncached unless the user disables the cache which is expected. I’m not sure why this isn’t in the warning in ClientOptions#makeCache though. Is it really worth adding optional chaining?

src/structures/Message.js
SpaceEEC: This is the code in question:
`message.channel.messages.fetch().then(m => m.find(m => m.type == 'THREAD_STARTER_MESSAGE')).then(m => m.pin())`

ImRodry: Ooohh okay my bad
Can you show me what a THREAD_STARTER_MESSAGE is or what it looks like btw?

src/structures/Message.js
ImRodry: Not needed anymore since this is just the reverse of Message#system

ImRodry:
_coding on mobile is hard_

src/structures/Message.js
SpaceEEC: It's just an empty message with a reference to the message the thread was created on:
<details>
<summary>THREAD_STARTER_MESSAGE  message</summary>

```js
{
  id: '884893836730572802',
  type: 21,
  content: '',
  channel_id: '884893816438542396',
  author: [Object],
  attachments: [],
  embeds: [],
  mentions: [],
  mention_roles: [],
  pinned: false,
  mention_everyone: false,
  tts: false,
  timestamp: '2021-09-07T20:12:11.209000+00:00',
  edited_timestamp: null,
  flags: 0,
  components: [],
  message_reference: {
    channel_id: 'id',
    guild_id: 'id',
    message_id: '884893816438542396'
  },
  referenced_message: {
    id: '884893816438542396',
    type: 0,
    content: 'test',
    channel_id: 'id',
    author: [Object],
    attachments: [],
    embeds: [],
    mentions: [],
    mention_roles: [],
    pinned: false,
    mention_everyone: false,
    tts: false,
    timestamp: '2021-09-07T20:12:06.371000+00:00',
    edited_timestamp: null,
    flags: 32,
    components: [],
    thread: [Object]
  }
}
```
</details>

SpaceEEC (approved):

kyranet (approved):

iCrawl (approved):

vladfrangu (approved):

PR #6586, opened 2021-09-03
refactor: node-fetch & @discordjs/formdata -> undici

KhafraDev:

Removed the dependencies of `node-fetch` and `@discordjs/formdata` in favor of `undici`.

Undici contains spec-compliant `fetch` and `FormData` implementations, is maintained by
NodeJS contributors, and is just basically all around better. [Repo](https://github.com/nodejs/undici)

Complete list of changes:
- Removed `@discordjs/form-data` package and its 4 dependencies.
- Removed `node-fetch` (still installed as a dev package)
- [MessageAttachment](https://discord.js.org/#/docs/main/stable/class/MessageAttachment) and [MessageAttachment.setFile](https://discord.js.org/#/docs/main/stable/class/MessageAttachment?scrollTo=setFile) now accept a `Blob` parameter. 
- Likewise, [DataResolver.resolveFile](https://discord.js.org/#/docs/main/stable/class/DataResolver?scrollTo=s-resolveFile) also accepts a Blob. It also returns a `Blob` now, whereas it used to return a `Buffer` or `ReadableStream`.
- [DataResolver.resolveBase64](https://discord.js.org/#/docs/main/stable/class/DataResolver?scrollTo=s-resolveBase64) now returns `Promise<string>`.
- [DataResolver.resolveFileAsBuffer](https://discord.js.org/#/docs/main/stable/class/DataResolver?scrollTo=s-resolveFileAsBuffer) is no longer used anywhere in the code, so like previously unused utility functions, will likely get removed in v14 (if this pr lands).
- The library now requires Node >=16.7.0 (bumped from Node >=16.6.0)

edit: I've been personally using this on my bot (roughly 21,000 members) that has made thousands of http requests and there has been no issue. This PR is likely ready for production use if anyone reading this wants to try it out. ;)

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,7 +24,7 @@ discord.js is a powerful [Node.js](https://nodejs.org) module that allows you to
 
 ## Installation
 
-**Node.js 16.6.0 or newer is required.**
+**Node.js 16.7.0 or newer is required.**  
 
 ```sh-session
 npm install discord.js
--- a/packages/discord.js/package.json
+++ b/packages/discord.js/package.json
@@ -52,11 +52,9 @@
     "@discordjs/builders": "^0.11.0",
     "@discordjs/collection": "^0.4.0",
     "@sapphire/async-queue": "^1.1.9",
-    "@types/node-fetch": "^2.5.12",
     "@types/ws": "^8.2.2",
     "discord-api-types": "^0.26.0",
-    "form-data": "^4.0.0",
-    "node-fetch": "^2.6.1",
+    "undici": "^4.12.1",
     "ws": "^8.4.0"
   },
   "devDependencies": {
@@ -77,6 +75,6 @@
     "typescript": "^4.5.4"
   },
   "engines": {
-    "node": ">=16.6.0"
+    "node": ">=16.7.0"
   }
 }
--- a/packages/discord.js/src/managers/GuildMemberManager.js
+++ b/packages/discord.js/src/managers/GuildMemberManager.js
@@ -1,7 +1,7 @@
 'use strict';
 
-const { Buffer } = require('node:buffer');
 const { setTimeout } = require('node:timers');
+const { types } = require('node:util');
 const { Collection } = require('@discordjs/collection');
 const CachedManager = require('./CachedManager');
 const { Error, TypeError, RangeError } = require('../errors');
@@ -115,7 +115,11 @@ class GuildMemberManager extends CachedManager {
     }
     const data = await this.client.api.guilds(this.guild.id).members(userId).put({ data: resolvedOptions });
     // Data is an empty buffer if the member is already part of the guild.
-    return data instanceof Buffer ? (options.fetchWhenExisting === false ? null : this.fetch(userId)) : this._add(data);
+    return types.isArrayBuffer(data)
+      ? options.fetchWhenExisting === false
+        ? null
+        : this.fetch(userId)
+      : this._add(data);
   }
 
   /**
--- a/packages/discord.js/src/rest/APIRequest.js
+++ b/packages/discord.js/src/rest/APIRequest.js
@@ -2,8 +2,7 @@
 
 const https = require('node:https');
 const { setTimeout } = require('node:timers');
-const FormData = require('form-data');
-const fetch = require('node-fetch');
+const { fetch, FormData } = require('undici');
 const { UserAgent } = require('../util/Constants');
 
 let agent = null;
@@ -61,7 +60,6 @@ class APIRequest {
           body.append('payload_json', JSON.stringify(this.options.data));
         }
       }
-      headers = Object.assign(headers, body.getHeaders());
       // eslint-disable-next-line eqeqeq
     } else if (this.options.data != null) {
       body = JSON.stringify(this.options.data);
--- a/packages/discord.js/src/rest/APIRequest.js
+++ b/packages/discord.js/src/rest/APIRequest.js
@@ -8,14 +8,6 @@ const { UserAgent } = require('../util/Constants');
 
 let agent = null;
 
-function toBlob(value) {
-  if (Buffer.isBuffer(value)) {
-    return new Blob([value]);
-  } else {
-    return value;
-  }
-}
-
 class APIRequest {
   constructor(rest, method, path, options) {
     this.rest = rest;
--- a/packages/discord.js/src/rest/APIRequest.js
+++ b/packages/discord.js/src/rest/APIRequest.js
@@ -1,6 +1,5 @@
 'use strict';
 
-const { Blob } = require('node:buffer');
 const https = require('node:https');
 const { setTimeout } = require('node:timers');
 const { fetch, FormData } = require('undici');
--- a/packages/discord.js/src/rest/APIRequest.js
+++ b/packages/discord.js/src/rest/APIRequest.js
@@ -1,11 +1,20 @@
 'use strict';
 
+const { Blob } = require('node:buffer');
 const https = require('node:https');
 const { setTimeout } = require('node:timers');
 const { fetch, FormData } = require('undici');
 const { UserAgent } = require('../util/Constants');
 
 let agent = null;
+
+function toBlob(value) {
+  if (Buffer.isBuffer(value)) {
+    return new Blob([value]);
+  } else {
+    return value;
+  }
+}
 
 class APIRequest {
   constructor(rest, method, path, options) {
@@ -51,7 +60,7 @@ class APIRequest {
     if (this.options.files?.length) {
       body = new FormData();
       for (const [index, file] of this.options.files.entries()) {
-        if (file?.file) body.append(file.key ?? `files[${index}]`, file.file, file.name);
+        if (file?.file) body.append(file.key ?? file.name ?? `files[${index}]`, file.file, file.name);
       }
       if (typeof this.options.data !== 'undefined') {
         if (this.options.dontUsePayloadJSON) {
--- a/packages/discord.js/src/rest/APIRequest.js
+++ b/packages/discord.js/src/rest/APIRequest.js
@@ -51,7 +51,7 @@ class APIRequest {
     if (this.options.files?.length) {
       body = new FormData();
       for (const [index, file] of this.options.files.entries()) {
-        if (file?.file) body.append(file.key ?? file.name ?? `files[${index}]`, file.file, file.name);
+        if (file?.file) body.append(file.key ?? `files[${index}]`, file.file, file.name);
       }
       if (typeof this.options.data !== 'undefined') {
         if (this.options.dontUsePayloadJSON) {
--- a/packages/discord.js/src/rest/APIRequest.js
+++ b/packages/discord.js/src/rest/APIRequest.js
@@ -8,6 +8,14 @@ const { UserAgent } = require('../util/Constants');
 
 let agent = null;
 
+function toBlob(value) {
+  if (Buffer.isBuffer(value)) {
+    return new Blob([value]);
+  } else {
+    return value;
+  }
+}
+
 class APIRequest {
   constructor(rest, method, path, options) {
     this.rest = rest;
--- a/packages/discord.js/src/rest/RequestHandler.js
+++ b/packages/discord.js/src/rest/RequestHandler.js
@@ -12,7 +12,7 @@ const {
 
 function parseResponse(res) {
   if (res.headers.get('content-type').startsWith('application/json')) return res.json();
-  return res.buffer();
+  return res.arrayBuffer();
 }
 
 function getAPIOffset(serverDate) {
--- a/packages/discord.js/src/rest/RequestHandler.js
+++ b/packages/discord.js/src/rest/RequestHandler.js
@@ -12,7 +12,7 @@ const {
 
 function parseResponse(res) {
   if (res.headers.get('content-type').startsWith('application/json')) return res.json();
-  return res.arrayBuffer().then(buffer => Buffer.from(buffer));
+  return res.arrayBuffer();
 }
 
 async function consumeBody(res) {
--- a/packages/discord.js/src/rest/RequestHandler.js
+++ b/packages/discord.js/src/rest/RequestHandler.js
@@ -12,7 +12,7 @@ const {
 
 function parseResponse(res) {
   if (res.headers.get('content-type').startsWith('application/json')) return res.json();
-  return res.arrayBuffer();
+  return res.arrayBuffer().then(buffer => Buffer.from(buffer));
 }
 
 function getAPIOffset(serverDate) {
--- a/packages/discord.js/src/rest/RequestHandler.js
+++ b/packages/discord.js/src/rest/RequestHandler.js
@@ -12,7 +12,7 @@ const {
 
 function parseResponse(res) {
   if (res.headers.get('content-type').startsWith('application/json')) return res.json();
-  return res.arrayBuffer();
+  return res.arrayBuffer().then(buffer => Buffer.from(buffer));
 }
 
 async function consumeBody(res) {
--- a/packages/discord.js/src/rest/RequestHandler.js
+++ b/packages/discord.js/src/rest/RequestHandler.js
@@ -12,7 +12,7 @@ const {
 
 function parseResponse(res) {
   if (res.headers.get('content-type').startsWith('application/json')) return res.json();
-  return res.arrayBuffer().then(buffer => Buffer.from(buffer));
+  return res.arrayBuffer();
 }
 
 function getAPIOffset(serverDate) {
--- a/packages/discord.js/src/rest/RequestHandler.js
+++ b/packages/discord.js/src/rest/RequestHandler.js
@@ -15,6 +15,13 @@ function parseResponse(res) {
   return res.arrayBuffer();
 }
 
+async function consumeBody(res) {
+  if (res.body === null) return;
+  // eslint-disable-next-line no-unused-vars, no-empty
+  for await (const _chunk of res.body) {
+  }
+}
+
 function getAPIOffset(serverDate) {
   return new Date(serverDate).getTime() - Date.now();
 }
@@ -306,6 +313,8 @@ class RequestHandler {
     if (res.status >= 400 && res.status < 500) {
       // Handle ratelimited requests
       if (res.status === 429) {
+        consumeBody(res);
+
         const isGlobal = this.globalLimited;
         let limit, timeout;
         if (isGlobal) {
@@ -352,6 +361,8 @@ class RequestHandler {
 
     // Handle 5xx responses
     if (res.status >= 500 && res.status < 600) {
+      consumeBody(res);
+
       // Retry the specified number of times for possible serverside issues
       if (request.retries === this.manager.client.options.retryLimit) {
         throw new HTTPError(res.statusText, res.constructor.name, res.status, request);
@@ -361,6 +372,7 @@ class RequestHandler {
       return this.execute(request);
     }
 
+    consumeBody(res);
     // Fallback in the rare case a status code outside the range 200..=599 is returned
     return null;
   }
--- a/packages/discord.js/src/structures/MessageAttachment.js
+++ b/packages/discord.js/src/structures/MessageAttachment.js
@@ -7,7 +7,7 @@ const Util = require('../util/Util');
  */
 class MessageAttachment {
   /**
-   * @param {BufferResolvable|Stream} attachment The file
+   * @param {BufferResolvable|Stream|Blob} attachment The file
    * @param {string} [name=null] The name of the file, if any
    * @param {APIAttachment} [data] Extra data
    */
@@ -33,7 +33,7 @@ class MessageAttachment {
 
   /**
    * Sets the file of this attachment.
-   * @param {BufferResolvable|Stream} attachment The file
+   * @param {BufferResolvable|Stream|Blob} attachment The file
    * @param {string} [name=null] The name of the file, if any
    * @returns {MessageAttachment} This attachment
    */
--- a/packages/discord.js/src/structures/MessageAttachment.js
+++ b/packages/discord.js/src/structures/MessageAttachment.js
@@ -7,7 +7,7 @@ const Util = require('../util/Util');
  */
 class MessageAttachment {
   /**
-   * @param {BufferResolvable|Stream|Blob} attachment The file
+   * @param {BufferResolvable|Stream|Blob|File} attachment The file
    * @param {string} [name=null] The name of the file, if any
    * @param {APIAttachment} [data] Extra data
    */
@@ -17,7 +17,7 @@ class MessageAttachment {
      * The name of this attachment
      * @type {?string}
      */
-    this.name = name;
+    this.name = attachment.name ?? name;
     if (data) this._patch(data);
   }
 
--- a/packages/discord.js/src/structures/MessagePayload.js
+++ b/packages/discord.js/src/structures/MessagePayload.js
@@ -219,7 +219,7 @@ class MessagePayload {
 
   /**
    * Resolves a single file into an object sendable to the API.
-   * @param {BufferResolvable|Stream|FileOptions|MessageAttachment|Blob} fileLike Something that could be resolved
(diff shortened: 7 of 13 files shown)


iCrawl:
https://github.com/nodejs/undici#undicifetchinput-init-promise

> Only supported on Node 16+.

>This is experimental and is not yet fully compliant with the Fetch Standard. We plan to ship breaking changes to this feature until it is out of experimental.

https://nodejs.org/api/documentation.html#documentation_stability_index

> Stability: 1 - Experimental. The feature is not subject to Semantic Versioning rules. Non-backward compatible changes or removal may occur in any future release. Use of the feature is not recommended in production environments.

KhafraDev:
> 
> 
> https://github.com/nodejs/undici#undicifetchinput-init-promise
> 
> > Only supported on Node 16+.
> 
> > This is experimental and is not yet fully compliant with the Fetch Standard. We plan to ship breaking changes to this feature until it is out of experimental.
> 
> https://nodejs.org/api/documentation.html#documentation_stability_index
> 
> > Stability: 1 - Experimental. The feature is not subject to Semantic Versioning rules. Non-backward compatible changes or removal may occur in any future release. Use of the feature is not recommended in production environments.

I completely agree, just posting it here for some discussion. I assume most of the animosity in the previous discussion linked above was that *someone* had to do the work, so I decided to just see what happens.

Although I believe v16 is about to be LTS which should mean that both Blob and web streams should be removed from experimental status (?).

ImRodry:
Why not use axios though?

KhafraDev:
> Why not use axios though?

I've never personally used `axios`, so I don't know what - if any - benefits it provides over a standardized fetch. However, just looking it over I can point out a few things:

- For Node, Axios wraps around http(s) which means it's much slower than undici (see: https://github.com/nodejs/undici/#benchmarks)
- By using a fetch implementation, it means that the API is standardized and, once spec-compliant, unlikely to change in the future.
- I have no idea if `axios` can handle the weird, non-spec compliant @discordjs/formdata package (for that matter, node-fetch v3 removed this behavior as well). Undici provides an implementation for node without external dependencies.
- Undici is maintained by nodejs maintainers under the nodejs organization. It's possible that this might land in core, and there have been talks of doing so. If it lands in core, we can simply remove the `const { fetch } = require('undici');` and it will work without changes.

iCrawl (approved):
Despite the comment above, I'm okay with the overall change once we know the stability will not be `1` in Node 16 LTS

src/rest/RequestHandler.js
```diff
@@ -11,7 +11,7 @@ const Util = require('../util/Util');
 
 function parseResponse(res) {
   if (res.headers.get('content-type').startsWith('application/json')) return res.json();
-  return res.buffer();
+  return res.arrayBuffer();
```
kyranet, line 14: This returns an [`ArrayBuffer`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/ArrayBuffer), if it's following the fetch API.

Fortunately, we do not have a method that returns `Promise<Buffer>` from a REST call, so this is not `semver: major` 😅

kyranet (approved):

src/rest/RequestHandler.js
KhafraDev: https://github.com/nodejs/undici/blob/main/types/fetch.d.ts#L38 

types say that this returns a Buffer, but after testing this seems to be incorrect 😄

will submit a pr/issue over there to get this fixed

KhafraDev: If needed, I can convert to buffers with just
```js
Buffer.from(ArrayBuffer)
```

iCrawl: I'm sure they were just referring to the fact that this is indeed different now, but since we only use it internally (and even there it can be either a buffer or array buffer) it doesn't matter.

kyranet: > but since we only use it internally (and even there it can be either a buffer or array buffer) it doesn't matter.

Actually, now that you mention it, it *does* matter:

https://github.com/discordjs/discord.js/blob/8cc3885739127ebc7023b6358ed0259ba7527250/src/managers/GuildMemberManager.js#L114-L116

This is broken after this PR given the current code, we need to either convert it to a buffer, or to use [`util.types.isArrayBuffer()`](https://nodejs.org/dist/latest-v16.x/docs/api/util.html#util_util_types_isarraybuffer_value).

KhafraDev: converted to Buffer in 6ac0b13

iCrawl: I'm not sure whats more expensive here.

Just naively glancing over it, checking if its an array buffer instead of creating a new buffer should be lighter?

src/rest/RequestHandler.js
```diff
@@ -11,7 +11,7 @@ const Util = require('../util/Util');
 
 function parseResponse(res) {
   if (res.headers.get('content-type').startsWith('application/json')) return res.json();
-  return res.buffer();
+  return res.arrayBuffer().then(buffer => Buffer.from(buffer));
```
NotSugden, line 14: can we use async/await here

src/rest/RequestHandler.js
kyranet: We probably can catch all instances of those checks with just `instanceof Buffer` (or `Buffer.isBuffer`), but if I had to choose, I'd go for using `isArrayBuffer`.

~~It's also ironic how this PR is made with the premise of increasing the library's performance, while at the same time, also introducing overhead.~~

Bottom line is that it doesn't matter, because `Buffer` extends `Uint8Array`, which can take an `ArrayBuffer` as parameter, which is also used to efficiently share memory between typed arrays, so it's just a tiny wrapper on it, the memory isn't duplicated.

Of course, we don't read those buffers, so it's a bit useless to do the conversion.

KhafraDev: Okay, I removed the conversion from Buffer -> ArrayBuffer and checked if the data is an ArrayBuffer instead (other than the tests which were previously buffers so I kept those changes).

jimmywarting: I suggest using ArrayBuffer/Uint8Array instead of Buffer whenever possible, including Buffer to Web & Deno is not good

advaith1: discord.js and unidici are node-only

jimmywarting: I just mean that: working with Buffer is considered as a bad practices for cross compatibility support

kyranet: First, we intentionally removed browser support a while ago, see #5113.

Second, we have too many dependencies with Node.js to even be able to support Deno (or browser, for that matter).

And working with Buffer isn't a bad practice when the project is Node.js-only, it has a bunch of useful methods Uint8Array lacks of, and is a "Stability: 2" feature, which suggests it's both a stable and a recommended API.

Sure, we could change some stuff so we check for `Uint8Array` instances, but as far as Node.js goes, a lot of things take and return `Buffer` instances, and it can also confuse users when they have a `Buffer` but see a method that takes `Uint8Array`, as the names are very different (even though Buffer extends Uint8Array).

If anything, we can always look into cross-platform support with the TypeScript rewrite, but not for v13.

Favna (approved):
> Why not use axios though?

Adding to the arguments given above
1. Axios also has a known bug in that you cannot send a `body` when sending a DELETE request. 
2. ~~Axios is currently still on `0.x` which per strict semver means "initial development. Anything MAY change at any time. The public API SHOULD NOT be considered stable" ([source](https://semver.org/#spec-item-4)). Therefore Axios should not be considered stable to begin with. Hardly a library we can afford to have in a project as large and with as many users as DiscordJS.~~ I have been corrected, apparently Axios doesn't use semver but [0ver](https://0ver.org). If you treasure your sanity, do not open that URL.
3. Axios has a backlog of issues and open pull requests from here to Saturn. If, aside from point 1, we'd ever run into an issue then the likelihood of having to just fork it and publish @discordjs/axios would be extremely large because likewise the likelihood of them merging a PR and releasing that in a timely fashion is extremely small.

J-Human (approved):

advaith1:
From unidici's readme:

> #### Specification Compliance
>
> This section documents parts of the [Fetch Standard](https://fetch.spec.whatwg.org) which Undici does
not support or does not fully implement.
>
> ##### Garbage Collection
>
> * https://fetch.spec.whatwg.org/#garbage-collection
>
> The [Fetch Standard](https://fetch.spec.whatwg.org) allows users to skip consuming the response body by relying on
[garbage collection](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Memory_Management#garbage_collection) to release connection resources. Undici does the same. However,
garbage collection in Node is less aggressive and deterministic (due to the lack
of clear idle periods that browser have through the rendering refresh rate)
which means that leaving the release of connection resources to the garbage collector
can lead to excessive connection usage, reduced performance (due to less connection re-use),
and even stalls or deadlocks when running out of connections. Therefore, it is highly 
recommended to always either consume or cancel the response body.
>
> ```js
> // Do
> const headers = await fetch(URL)
>   .then(async res => {
>     for await (const chunk of res) {
>       // force consumption of body
>     }
>     return res.headers
>   })
>
> // Do not
> const headers = await fetch(url)
>   .then(res => res.headers)
> ```

will this be an issue?

iShibi:
`FormData#append` takes a `Blob` as value. `undici` throws an error here, because `file.file` is not an instance of `Blob`.
https://github.com/discordjs/discord.js/blob/72e3958bc4b47b562296a64d683f95ec6995520f/src/rest/APIRequest.js#L53

KhafraDev:
> 
> 
> `FormData#append` takes a `Blob` as value. `undici` throws an error here, because `file.file` is not an instance of `Blob`.
> 
> https://github.com/discordjs/discord.js/blob/72e3958bc4b47b562296a64d683f95ec6995520f/src/rest/APIRequest.js#L53

Can `file.file` be anything other than a Buffer?

jimmywarting:
> Can file.file be anything other than a Buffer? From testing I haven't seen any other value. ty :)

Using undici is really living on the edge of experimental apis...
undici is using latest stuff from NodeJS and requires the newest and experimental api's there is... which is still unstable

- Now undici is using [require('node:buffer').Blob](https://github.com/nodejs/undici/blob/9f53d8216cc70f35de14a5f43d73f8608ab98a21/lib/fetch/body.js#L7) and not something else throughout their entire codebase (in fetch & FormData)
- undici created a own [File class](https://github.com/nodejs/undici/blob/main/lib/fetch/file.js) that extends the builtin NodeJS Blob
- And theirs FormData instance requires the input to be nothing but a [instance](https://github.com/nodejs/undici/blob/9f53d8216cc70f35de14a5f43d73f8608ab98a21/lib/fetch/formdata.js#L28-L32) of a NodeJS Blob coming from `node:buffer` 

So no... you can't append a buffer into their FormData instance, you have to construct a File using there version of the File implementation.

## About Blobs...

One of the reason why fetch-blob haven't switched to using the built in version of NodeJS Blob impl is b/c they are pretty darn useless unless you can't get them backed up by the filesystem somehow, otherwise they are just useless containers that stores a buffers in memory where you could as well use a uint8array or ArrayBuffer directly instead.

With `fetch-blob` you can get a File from the filesystem without holding anything in the memory. 
something of which node's is looking into implementing https://github.com/nodejs/node/issues/37340 But they are blocked by support for async blob sources https://github.com/nodejs/node/issues/37338 and also implementing a File class https://github.com/nodejs/node/issues/39015

Only then when File and blobs backed up by the file system would we over at node-fetch start to consider using the built in impl. and marking fetch-blob as legacy. cuz right now it serves a purpose not even NodeJS have fixed in their version.

## About FormData

Now to the version of undici FormData impl... knowing what i have written earlier about node's wortless Blob container and how strict undici FormData only accept a version of NodeJS own Blob implementation...
Do you think undici dose a grate job at sending large payloads with files embedded to a few GB of files? the answer is No.
You are required to read all files, construct a new Blob (or a File that extends Blob) with buffers and it's all held in memory.

Unlike one other spec compatible FormData implementations such as [formdata-polyfill](https://www.npmjs.com/package/formdata-polyfill) it dose not use instanceof checks. It can work with other Blob/File implementations as long as they look like a Blob, so it can accept both a fetch-blob and a NodeJS blob

and it can generate a FormData boundary/payload without ever reading the files content.

undici is more focused on writing something they wish to see being implemented into core one day. not something that is best suitable for userland as they require the latest experimental features 
undici is very new and not as much battle tested as node-fetch

if i would choose between undici and node-fetch then i would _maybe_ go with node-fetch 2.x or 3.x b/c it's 
- supported on more node's versions
- have been tested a lot by many users
- is less strict about what Blob, File, AbortSignal, URLSearchParams and FormData you are using. So it can work with files backed up by filesystem or other implementations cuz it dose not use `instanceof` checks instead uses duck typing to check if it looks like a duck and quack like a duck. This have been a feature of node-fetch since a very long beginning b/c nodejs lacks all those things.

KhafraDev:
The issues you brought up are completely out of scope for this repo and mostly don't apply (either to the platform, usage, or users).

iShibi:
The [`toBlob`](https://github.com/discordjs/discord.js/blob/0207b4a1204d019256c2c73d4754dba4d996884e/src/rest/APIRequest.js#L10) function doesn't work currently. The reason is that [`DataResolver#resolveFile`](https://github.com/discordjs/discord.js/blob/fe9500538e76423e49e16a1e1756eb04b5f40531/src/util/DataResolver.js#L108) returns a [`ReadableStream`](https://nodejs.org/dist/latest-v16.x/docs/api/webstreams.html) (Web Stream) for a **URL**, as that's what [`Response.body`](https://github.com/nodejs/undici#responsebody) is in undici and a [`ReadStream`](https://nodejs.org/api/stream.html#stream_class_stream_readable) (Nodejs Stream) for a **local file path**, as that's what `fs#createReadStream` returns.

Neither of them returns `true` for `Buffer#isBuffer`, hence they can't be converted into a Blob, which results in undici throwing as it expects a Blob as a value in `FormDate#append`.

I did some tinkering and got the `FormData#append` working which results in the bot correctly uploading the attachments. However, <ins>I don't know how **efficient** of an approach it is or what effects it has on the memory</ins>. Still, I'll write down what I did:

1) For a URL, we can directly return `Response.blob()` in [`DataResolver#resolveFile`](https://github.com/discordjs/discord.js/blob/fe9500538e76423e49e16a1e1756eb04b5f40531/src/util/DataResolver.js#L108) :
```js
static async resolveFile(resource) {
  if (typeof resource === 'string') {
    if (/^https?:\/\//.test(resource)) {
      const res = await fetch(resource);
      // return res.body;
      return res.blob();
    }
    // ...
  }
}
```
`Buffer#isBuffer` will return `false` for Blob, hence `toBlob` will pass it as it is to `FormData#append`.

2) For a local file path, we can create a `Buffer` using `fs.readFile`:
```js
static async resolveFile(resource) {
  if (typeof resource === 'string') {
    // ...
      const file = path.resolve(resource);
      fs.stat(file, (err, stats) => {
        if (err) return reject(err);
        if (!stats.isFile()) return reject(new DiscordError('FILE_NOT_FOUND', file));
        return fs.readFile(file, (error, buf) => {
          if (error) return reject(error);
          return resolve(buf);
        });
      });
    });
  }
}
``` 
`toBlob` will then convert this `Buffer` into a Blob

KhafraDev:
I got everything fixed & tested with both `Buffer` and node `Readable streams` - will push the changes in a few hours.

In `resolveFile` I ended up following your suggestions but I converted everything to blobs there and remove the `toBlob` function.

src/util/DataResolver.js
```diff
@@ -103,22 +104,32 @@ class DataResolver extends null {
   /**
    * Resolves a BufferResolvable to a Buffer or a Stream.
```
iShibi, line 105:
```suggestion
   * Resolves a BufferResolvable or a Stream to a Blob
```

iShibi (changes requested):
Works fine. Make sure to update the typings too:
https://github.com/discordjs/discord.js/blob/fe9500538e76423e49e16a1e1756eb04b5f40531/typings/index.d.ts#L628

KhafraDev:
I think this PR should be classified as `semver-major` due to 0618e7f9c22a7f3901669071d79fe2b6521c5a91
- DataResolver.resolveFileAsBuffer is no longer used (not removed however)
- DataResolver.resolveBase64 is now asynchronous (doesn't change behavior of the lib, but will affect anyone using it).
- DataResolver.resolveFile now accepts a `Blob` parameter (ie. MessageAttachment should support Blobs I believe).

iCrawl:
This needs a rebase.

src/util/DataResolver.js
```diff
@@ -78,11 +79,16 @@ class DataResolver extends null {
...
    * Resolves a Base64Resolvable to a Base 64 image.
-   * @param {Base64Resolvable} data The base 64 resolvable you want to resolve
-   * @returns {?string}
+   * @param {Base64Resolvable|Blob} data The base 64 resolvable you want to resolve
+   * @returns {Promise<string>}
    */
-  static resolveBase64(data) {
+  static async resolveBase64(data) {
     if (Buffer.isBuffer(data)) return `data:image/jpg;base64,${data.toString('base64')}`;
+    if (data instanceof Blob) {
```
vladfrangu, line 87: Question: don't blobs have a type too? If they do, can't we use that for the `data:content-type`?

vladfrangu (changes requested):
Please fix lint

src/util/DataResolver.js
KhafraDev: Yes, fixed in fd18ecf64d2ae364657159eb6128ce235c9c13cc. If the type isn't specified, it's an empty string so I still default it to `image/jpg` though.

KhafraDev:
> 
> 
> Please fix lint

guess the precommit hook wasn't working, thought it would have caught that for me 😕

kyranet:
> I think this PR should be classified as `semver-major` due to [0618e7f](https://github.com/discordjs/discord.js/commit/0618e7f9c22a7f3901669071d79fe2b6521c5a91)

Done!

iCrawl:
This needs a rebase 👍

KhafraDev:
How would you guys feel if I used [`stream/consumers`](https://nodejs.org/api/webstreams.html#webstreams_utility_consumers)? They were added in v16.7.0 (so the min node version would have to get bumped from v16.6.0) but they would simplify some logic and allow the lib to support both node streams and web streams without extra logic.

iShibi:
This PR is `semver:major` now, so a bump in node version wouldn't be an issue 👍

src/util/DataResolver.js
```diff
@@ -101,24 +108,31 @@ class DataResolver extends null {
...
-   * @returns {Promise<Buffer|Stream>}
+   * Resolves a BufferResolvable, Blob, or Stream to a Blob.
+   * @param {BufferResolvable|Stream|Blob|File} resource The buffer or stream resolvable to resolve
+   * @returns {Promise<Blob>}
    */
   static async resolveFile(resource) {
-    if (Buffer.isBuffer(resource) || resource instanceof stream.Readable) return resource;
+    if (resource instanceof Blob) return resource;
+    if (Buffer.isBuffer(resource)) return new Blob([resource]);
+    if (resource !== null && resource !== undefined && resource[Symbol.asyncIterator]) {
```
iShibi, line 118:
```suggestion
    if (resource?.[Symbol.asyncIterator]) {
```

src/util/DataResolver.js
```diff
@@ -101,24 +108,31 @@ class DataResolver extends null {
...
+      return stream.consumers.blob(resource);
```
iShibi, line 119:
```suggestion
      return blob(resource);
```

KhafraDev:
> 
> 
> `stream.consumers.blob()` throws error, because `consumers` is undefined

image
vscode lied to me lol

(fixed in a9dd2fe5ecd6c52bdc6e14e6b4920a8ffe842e75)

iCrawl:
This needs a rebase.

KhafraDev:
> I really don't like that we have to manually consume the body if we won't use it (maybe one day they'll handle that for us 🤞), but otherwise LGTM once it gets stable
```
The Fetch Standard allows users to skip consuming the response body by relying on garbage collection
to release connection resources. Undici does the same. However, garbage collection in Node is less aggressive
and deterministic (due to the lack of clear idle periods that browser have through the rendering refresh rate)
which means that leaving the release of connection resources to the garbage collector can lead to excessive
connection usage, reduced performance (due to less connection re-use), and even stalls or deadlocks when
running out of connections. Therefore, it is highly recommended to always either consume or cancel the 
response body.
```
It seems like an issue with NodeJS itself and not the package. The difference between `node-fetch` and `undici` is that undici attempts to follow the spec whereas node-fetch just doesn't really care. It might also be an underlying implementation because nf uses http(s) and undici uses sockets.

package.json
```diff
@@ -52,12 +52,10 @@
   "dependencies": {
     "@discordjs/builders": "^0.9.0",
     "@discordjs/collection": "^0.3.2",
-    "@discordjs/form-data": "^3.0.1",
     "@sapphire/async-queue": "^1.1.9",
-    "@types/node-fetch": "^2.5.12",
     "@types/ws": "^8.2.2",
     "discord-api-types": "^0.25.2",
-    "node-fetch": "^2.6.1",
+    "undici": "^4.9.2",
```
xHyroM, line 58: A new version of umidici, 4.12.0, was released 12h ago

KhafraDev:
https://github.com/nodejs/node/pull/40971 👀

kyranet:
Just to clarify, are WHATWG streams the only thing that was in experimental? @KhafraDev

KhafraDev:
@kyranet buffer.Blob still remains

KhafraDev:
> Although I believe v16 is about to be LTS which should mean that both Blob and web streams should be removed from experimental status (?).

I was slightly off with this one, by about 1 LTS version, but v18 will have a global Blob and experimental status removed as well :)
https://github.com/nodejs/node/pull/41270

KhafraDev:
hey 😏

KhafraDev:
I have no interest in keeping this PR open as it likely won't land.

ImRodry:
I know this was closed but undici's fetch just landed on node as an experimental flag, so now is an even better time to remake this PR IMO
https://github.com/nodejs/node/pull/41749

suneettipirneni:
> I know this was closed but undici's fetch just landed on node as an experimental flag, so now is an even better time to reopen this PR IMO
> 
> 
> 
> - https://github.com/nodejs/node/pull/41749

This was closed because it's going to be added to /rest now. Also there's a roadblock in mocking some of the request data.

ImRodry:
> > I know this was closed but undici's fetch just landed on node as an experimental flag, so now is an even better time to reopen this PR IMO
> > 
> > * [lib: add fetch nodejs/node#41749](https://github.com/nodejs/node/pull/41749)
> 
> This was closed because it's going to be added to /rest now. Also there's a roadblock in mocking some of the request data.

This was closed because the author didn't feel like keeping it up. There's no open issue about that to my knowledge

suneettipirneni:
> > > I know this was closed but undici's fetch just landed on node as an experimental flag, so now is an even better time to reopen this PR IMO
> 
> > > 
> 
> > > * [lib: add fetch nodejs/node#41749](https://github.com/nodejs/node/pull/41749)
> 
> > 
> 
> > This was closed because it's going to be added to /rest now. Also there's a roadblock in mocking some of the request data.
> 
> 
> 
> This was closed because the author didn't feel like keeping it up. There's no open issue about that to my knowledge

I've spoken to the author and my reasons for why it's closed are from what they told me.

ckohen:
> I know this was closed but undici's fetch just landed on node as an experimental flag, so now is an even better time to reopen this PR IMO

How could you even reopen this PR? The majority of the point of the PR has been made obsolete because /rest has already replaced it.

As you've already been  told, the author was told the switch to /rest was happening soon and thus this probably couldn't land. Between me and the author we are looking to make this PR a reality in /rest but switching the tests over is a major rewrite and since there are still things that are experimental in undici fetch we aren't rushing to do that.

PR #6603, opened 2021-09-07
feat(ApplicationCommand): add version property

iShibi:

Adds the new `version` field of `ApplicationCommand`:
- discord/discord-api-docs#3524

- This PR changes the library's interface

### DIFF
--- a/src/structures/ApplicationCommand.js
+++ b/src/structures/ApplicationCommand.js
@@ -77,6 +77,12 @@ class ApplicationCommand extends Base {
      * @type {boolean}
      */
     this.defaultPermission = data.default_permission;
+
+    /**
+     * Autoincrementing version identifier updated during substantial record changes
+     * @type {Snowflake}
+     */
+    this.version = data.version;
   }
 
   /**
--- a/src/structures/ApplicationCommand.js
+++ b/src/structures/ApplicationCommand.js
@@ -185,7 +185,7 @@ class ApplicationCommand extends Base {
       // TODO: remove ?? 0 on each when nullable
       (command.options?.length ?? 0) !== (this.options?.length ?? 0) ||
       (command.defaultPermission ?? command.default_permission ?? true) !== this.defaultPermission ||
-      command.version !== this.version
+      ('version' in command && command.version !== this.version)
     ) {
       return false;
     }
--- a/src/structures/ApplicationCommand.js
+++ b/src/structures/ApplicationCommand.js
@@ -180,12 +180,12 @@ class ApplicationCommand extends Base {
     if (
       command.name !== this.name ||
       ('description' in command && command.description !== this.description) ||
+      ('version' in command && command.version !== this.version) ||
       (commandType && commandType !== this.type) ||
       // Future proof for options being nullable
       // TODO: remove ?? 0 on each when nullable
       (command.options?.length ?? 0) !== (this.options?.length ?? 0) ||
-      (command.defaultPermission ?? command.default_permission ?? true) !== this.defaultPermission ||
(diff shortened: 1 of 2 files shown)


src/structures/ApplicationCommand.js
```diff
@@ -77,6 +77,12 @@ class ApplicationCommand extends Base {
      * @type {boolean}
      */
     this.defaultPermission = data.default_permission;
+
+    /**
+     * Autoincrementing version identifier updated during substantial record changes
+     * @type {Snowflake}
+     */
+    this.version = data.version;
```
vladfrangu, line 85: This should be checked in equals too

vladfrangu (dismissed):

src/structures/ApplicationCommand.js
iShibi: sure

src/structures/ApplicationCommand.js
```diff
@@ -178,7 +184,8 @@ class ApplicationCommand extends Base {
       // Future proof for options being nullable
       // TODO: remove ?? 0 on each when nullable
       (command.options?.length ?? 0) !== (this.options?.length ?? 0) ||
-      (command.defaultPermission ?? command.default_permission ?? true) !== this.defaultPermission
+      (command.defaultPermission ?? command.default_permission ?? true) !== this.defaultPermission ||
+      command.version !== this.version
```
kyranet, line 188: This should be the same as the first checks:
```suggestion
      ('version' in command && command.version !== this.version)
```

Also, shouldn't we move this check with description and others?

kyranet (changes requested):

src/structures/ApplicationCommand.js
ckohen: I think this check should be moved right after id, in a standalone form. If versions are equal there's no reason to check the rest

iShibi: that sounds right. btw why are we using `in` keyword even for fields that should always be present, why not use optional chaining instead 👀

ckohen: because purposefully setting a value to undefined or null (though technically invalid) causes the equal check to happen properly

src/structures/ApplicationCommand.js
ImRodry: Why should it? You can use equals with ApplicationCommandData and the version is something only Discord provides, you cannot have a version in your command data

ckohen: Equals takes an ApplicationCommand too

ImRodry: Yeah I know that, that’s exactly why I made that comment because just comparing them like it was initially done was pretty stupid, but doing a check like the id check is better

src/structures/ApplicationCommand.js
```diff
@@ -77,6 +77,12 @@ class ApplicationCommand extends Base {
      * @type {boolean}
      */
     this.defaultPermission = data.default_permission;
+
+    /**
+     * Autoincrementing version identifier updated during substantial record changes
```
ImRodry, line 82: Isn’t this unnecessarily long?
```suggestion
     * Version identifier updated every time the command is edited
```

iShibi: That was added by night after some disagreements around the few initial descriptions, so I think we should keep it.

SpaceEEC (approved):

iCrawl (approved):

kyranet (approved):

vladfrangu (approved):

PR #6614, opened 2021-09-08
feat(CategoryChannel): add createChannel shortcut method

ImRodry:
This PR adds a method to CategoryChannel called createChannel that creates a channel inside that category. It also fixes a bug in the options interface to disallow GUILD_NEWS_THREAD types on both the GuildChannelCreateOptions and the CategoryCreateChannelOptions interfaces

### DIFF
--- a/src/structures/CategoryChannel.js
+++ b/src/structures/CategoryChannel.js
@@ -26,6 +26,35 @@ class CategoryChannel extends GuildChannel {
    * @param {SetParentOptions} [options={}] The options for setting the parent
    * @returns {Promise<GuildChannel>}
    */
+
+  /**
+   * Options for creating a channel using {@link CategoryChannel#createChannel}.
+   * @typedef {Object} CategoryCreateChannelOptions
+   * @property {ChannelType|number} [type='GUILD_TEXT'] The type of the new channel.
+   * @property {string} [topic] The topic for the new channel
+   * @property {boolean} [nsfw] Whether the new channel is nsfw
+   * @property {number} [bitrate] Bitrate of the new channel in bits (only voice)
+   * @property {number} [userLimit] Maximum amount of users allowed in the new channel (only voice)
+   * @property {OverwriteResolvable[]|Collection<Snowflake, OverwriteResolvable>} [permissionOverwrites]
+   * Permission overwrites of the new channel
+   * @property {number} [position] Position of the new channel
+   * @property {number} [rateLimitPerUser] The ratelimit per user for the new channel
+   * @property {string} [reason] Reason for creating the new channel
+   */
+
+  /**
+   * Creates a new channel on this category.
+   * <info>You cannot create a channel of type `GUILD_CATEGORY` inside a CategoryChannel.</info>
+   * @param {string} name The name of the new channel
+   * @param {CategoryCreateChannelOptions} options Options for creating the new channel
+   * @returns {Promise<GuildChannel>}
+   */
+  createChannel(name, options = {}) {
+    return this.guild.channels.create(name, {
+      parent: this.id,
+      ...options,
+    });
+  }
 }
 
 module.exports = CategoryChannel;
--- a/src/structures/CategoryChannel.js
+++ b/src/structures/CategoryChannel.js
@@ -49,7 +49,7 @@ class CategoryChannel extends GuildChannel {
    * @param {CategoryCreateChannelOptions} options Options for creating the new channel
    * @returns {Promise<GuildChannel>}
    */
-  createChannel(name, options = {}) {
+  createChannel(name, options) {
     return this.guild.channels.create(name, {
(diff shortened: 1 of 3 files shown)


src/structures/CategoryChannel.js
```diff
@@ -26,6 +26,35 @@ class CategoryChannel extends GuildChannel {
...
+   * Creates a new channel on this category.
+   * <info>You cannot create a channel of type `GUILD_CATEGORY` inside a CategoryChannel.</info>
+   * @param {string} name The name of the new channel
+   * @param {CategoryCreateChannelOptions} options Options for creating the new channel
+   * @returns {Promise<GuildChannel>}
+   */
+  createChannel(name, options = {}) {
+    return this.guild.channels.create(name, {
+      parent: this.id,
+      ...options,
```
NotSugden, line 55: i think `...options` should be done first, just incase a parent is passed in the options by accident or however

typings/index.d.ts
```diff
@@ -3199,6 +3223,34 @@ export type CacheWithLimitsOptions = {
     : never;
 };
 
+export interface CategoryCreateChannelOptions {
+  permissionOverwrites?: OverwriteResolvable[] | Collection<Snowflake, OverwriteResolvable>;
+  topic?: string;
+  type?: Exclude<
```
NotSugden, line 3229: maybe out of acope for this PR, but i think this should have a better type

```ta
type ExcludeEnum<Enum, Keys extends keyof Enum> = Exclude<keyof typeof Enum | Enum, Keys | Enum[Keys]>
```

because as-is this type is unnessercerily big

src/structures/CategoryChannel.js
ImRodry: Good point, GitHub copilot suggested this and I didn't even think twice lol

typings/index.d.ts
ImRodry: I'm not experienced with types on this level so if you could provide an example of how this would look like I could add it, and since the PR already touches this type I don't see why not

NotSugden:
```ts
ExcludeEnum<ChannelType,
  'DM'
>
```
instead of
```ts
Exclude<keyof typeof ChannelType | ChannelType,
  'DM',
  ChannelType.DM
>
```

ImRodry: Will try this and commit if it goes well

NotSugden: the type might need a bit of fine tuning but yea

src/structures/CategoryChannel.js
```diff
@@ -26,6 +26,35 @@ class CategoryChannel extends GuildChannel {
...
+   */
+
+  /**
+   * Creates a new channel on this category.
+   * <info>You cannot create a channel of type `GUILD_CATEGORY` inside a CategoryChannel.</info>
+   * @param {string} name The name of the new channel
+   * @param {CategoryCreateChannelOptions} options Options for creating the new channel
+   * @returns {Promise<GuildChannel>}
+   */
+  createChannel(name, options = {}) {
```
vladfrangu, line 52: Maybe don't default options

And before you say, yes, it'll be fine
image

vladfrangu (approved):
Smallest nit, but can be merged without it so LGTM

src/structures/CategoryChannel.js
NotSugden: i think there are a few places where
`{ ...(x ?? {}) }` is used perhaps that could be changed if undefined is valid

ImRodry: Never seen that notation Sugden but I’ll change it since the main method already defaults these to an empty object

anandre:
I don't think that this method brings any needed utility to the library.  Adding a method that AFAIK is only on managers to separate class seems like it's cluttering up the library for no good reason.  All this does is add more to the library for the convenience of setting one value for the user.  Recently, the library has been removing extraneous helper/utility methods, and this is going in exactly the opposite direction.

typings/index.d.ts
ImRodry: Decided I'll add this in a different PR after this one is merged since I also wanna use this ExcludeEnum type in other places, so I'll mark this as resolved for now

ImRodry:
@anandre the methods the library has been recently removing are only ones that were unreliable and caused more hassles than they solved (e.g. typing-related properties and methods, User#presence, etc). This method is indeed a convenience and I recognize that it would be better to have a children manager, however, that cannot be done until v14. If the reviewers prefer we can hold this off and transform CategoryChannel#children into a manager later on, however I don't see why this couldn't be added personally as it is quite convenient and seems to be getting some positive feedback too.

SpaceEEC (approved):

iCrawl (approved):

PR #6621, opened 2021-09-10
feat: Add support for API command types in `ApplicationCommandManager`

suneettipirneni:
Allows api types to be accepted in `ApplicationCommandManager` methods. This means typescript will no longer complain about type mismatches when using something like builders for example.

### DIFF
--- a/src/managers/ApplicationCommandManager.js
+++ b/src/managers/ApplicationCommandManager.js
@@ -100,7 +100,7 @@ class ApplicationCommandManager extends CachedManager {
 
   /**
    * Creates an application command.
-   * @param {ApplicationCommandData|RESTPostAPIApplicationCommandsJSONBody} command The command
+   * @param {ApplicationCommandData|APIApplicationCommand} command The command
    * @param {Snowflake} [guildId] The guild's id to create this command in,
    * ignored when using a {@link GuildApplicationCommandManager}
    * @returns {Promise<ApplicationCommand>}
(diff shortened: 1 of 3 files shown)


src/managers/ApplicationCommandManager.js
```diff
@@ -100,7 +100,7 @@ class ApplicationCommandManager extends CachedManager {
 
   /**
    * Creates an application command.
-   * @param {ApplicationCommandData} command The command
+   * @param {ApplicationCommandData|RESTPostAPIApplicationCommandsJSONBody} command The command
```
ImRodry, line 103: You can’t just use names of interfaces from the typings here. These have to be declared as `@external` links at the end of a file and this isn’t the naming scheme we use for these external links either. I believe there is already an APIApplicationCommand though, but please double check that

src/structures/ApplicationCommand.js
```diff
@@ -298,6 +298,31 @@ class ApplicationCommand extends Base {
...
+   */
+  static transformCommand(command) {
+    return {
+      name: command.name,
+      description: command.description,
+      type: typeof command.type === 'number' ? command.type : ApplicationCommandTypes[command.type],
+      options: command.options?.map(o => ApplicationCommand.transformOption(o)),
+      default_permission: command.defaultPermission,
+    };
+  }
```
ckohen, line 316: Why did this get duplicated

src/structures/ApplicationCommand.js
```diff
@@ -298,6 +298,31 @@ class ApplicationCommand extends Base {
...
+
+  /**
+   * Whether or not the given object is api command data or not.
+   * @param {Object} command The command object to check.
+   * @returns {boolean} True if the object is api command data, false otherwise.
+   */
+  static isAPICommandData(command) {
+    return 'default_permission' in command || 'guild_id' in command;
+  }
```
ckohen, line 325: This isn't a definitive check, and as mentioned either below or above, I believe its unnecessary

src/managers/ApplicationCommandManager.js
```diff
@@ -115,14 +115,14 @@ class ApplicationCommandManager extends CachedManager {
    */
   async create(command, guildId) {
     const data = await this.commandPath({ guildId }).post({
-      data: this.constructor.transformCommand(command),
+      data: ApplicationCommand.isAPICommandData(command) ? command : this.constructor.transformCommand(command),
```
ckohen, line 118: Instead of changing this everywhere, you should use the transformCommand function and just do
```js
default_permission: command.defaultPermission ?? comand.default_permission
```

ckohen (changes requested):

kyranet (approved):

iCrawl (approved):

SpaceEEC (approved):

PR #6629, opened 2021-09-12
types(MessageEmbed): allow APIEmbed type

### LINKED ISSUE #6627: Bug: MessageEmbed not accepting a date string even though it would be possible to do so
### Issue description

1. Create an object that has a type of `APIEmbed` from `discord-api-types`
2. Supply it to the `MessageEmbed` constructor
3. Observe TypeScript error

This issue could easily be fixed by making the setup function accept strings too, but throw an error if the input is not a valid date string. I would gladly write it in.

### Codesample

```typescript
import {MessageEmbed} from "discord.js"
import {APIEmbed} from "discord-api-types"

const API = <APIEmbed>{} // this is wrong, but it still throws the same TS error
const Embed = new MessageEmbed(API)
```

### discord.js version

13.1.0

### Node.js version

16.6.2

### Operating system

MacOS

Low (slightly annoying)

_No response_

None, this doesn't relate to gateway intents

_No response_

ImRodry:
This PR adds support for the APIEmbed type in the MessageEmbed constructor. The code didn't need to be updated as it already supported this type. Fixes #6627

### DIFF
--- a/src/structures/MessageEmbed.js
+++ b/src/structures/MessageEmbed.js
@@ -30,9 +30,9 @@ class MessageEmbed {
    * @property {Partial<MessageEmbedFooter>} [footer] The footer of this embed
    */
 
+  // eslint-disable-next-line valid-jsdoc
   /**
    * @param {MessageEmbed|MessageEmbedOptions|APIEmbed} [data={}] MessageEmbed to clone or raw embed data
-   * @param {boolean} [skipValidation=false] Whether to skip validation of embed fields
    */
   constructor(data = {}, skipValidation = false) {
     this.setup(data, skipValidation);
--- a/src/structures/MessageEmbed.js
+++ b/src/structures/MessageEmbed.js
@@ -8,13 +8,6 @@ const Util = require('../util/Util');
  */
 class MessageEmbed {
   /**
-   * @name MessageEmbed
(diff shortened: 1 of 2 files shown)


xMooddyy (approved):

src/structures/MessageEmbed.js
```diff
@@ -37,6 +30,10 @@ class MessageEmbed {
    * @property {Partial<MessageEmbedFooter>} [footer] The footer of this embed
    */
 
+  /**
```
SpaceEEC, line 33: Please revert documenting the second parameter of the constructor, it's intentionally undocumented because it's only to be used internally (since Discord doesn't hold itself to the limitation it's imposing).

SpaceEEC (changes requested):

src/structures/MessageEmbed.js
ImRodry: Addressed in https://github.com/discordjs/discord.js/pull/6629/commits/870d811245c0aa8a95ebd0543fbdca9412c0b735

SpaceEEC (approved):

kyranet (approved):

iCrawl (approved):

PR #6633, opened 2021-09-13
feat: add support for role icons

iShibi:

Starts work on role icons:
- discord/discord-api-docs#3815
- discord/discord-api-docs#3847

### DIFF
--- a/src/managers/RoleManager.js
+++ b/src/managers/RoleManager.js
@@ -106,7 +106,7 @@ class RoleManager extends CachedManager {
    * @property {number} [position] The position of the new role
    * @property {boolean} [mentionable] Whether or not the new role should be mentionable
    * @property {?(BufferResolvable|Base64Resolvable)} [icon] The icon for the role
-   * @property {?string} [unicodeEmoji] The unicode emoji for the role
+   * @property {?string} [unicodeEmoji] The name of the unicode emoji for the role
    * @property {string} [reason] The reason for creating this role
    */
 
--- a/src/managers/RoleManager.js
+++ b/src/managers/RoleManager.js
@@ -105,7 +105,7 @@ class RoleManager extends CachedManager {
    * @property {PermissionResolvable} [permissions] The permissions for the new role
    * @property {number} [position] The position of the new role
    * @property {boolean} [mentionable] Whether or not the new role should be mentionable
-   * @property {BufferResolvable|Base64Resolvable} [icon] The icon for the role
+   * @property {?(BufferResolvable|Base64Resolvable)} [icon] The icon for the role
    * @property {string} [reason] The reason for creating this role
    */
 
--- a/src/managers/RoleManager.js
+++ b/src/managers/RoleManager.js
@@ -106,6 +106,8 @@ class RoleManager extends CachedManager {
    * @property {number} [position] The position of the new role
    * @property {boolean} [mentionable] Whether or not the new role should be mentionable
    * @property {?(BufferResolvable|Base64Resolvable|EmojiResolvable)} [icon] The icon for the role
+   * <warn>The `EmojiResolvable` should belong to the same guild as the role.
+   * If not, pass the emoji's URL directly</warn>
    * @property {?string} [unicodeEmoji] The unicode emoji for the role
    * @property {string} [reason] The reason for creating this role
    */
@@ -137,6 +139,7 @@ class RoleManager extends CachedManager {
     if (icon) {
       const guildEmojiURL = this.guild.emojis.resolve(icon)?.url;
       icon = guildEmojiURL ? await DataResolver.resolveImage(guildEmojiURL) : await DataResolver.resolveImage(icon);
+      if (typeof icon !== 'string') icon = undefined;
     }
 
     const data = await this.client.api.guilds(this.guild.id).roles.post({
@@ -195,6 +198,7 @@ class RoleManager extends CachedManager {
     if (icon) {
       const guildEmojiURL = this.guild.emojis.resolve(icon)?.url;
       icon = guildEmojiURL ? await DataResolver.resolveImage(guildEmojiURL) : await DataResolver.resolveImage(icon);
+      if (typeof icon !== 'string') icon = undefined;
     }
 
     const _data = {
--- a/src/managers/RoleManager.js
+++ b/src/managers/RoleManager.js
@@ -106,7 +106,7 @@ class RoleManager extends CachedManager {
    * @property {number} [position] The position of the new role
    * @property {boolean} [mentionable] Whether or not the new role should be mentionable
    * @property {?(BufferResolvable|Base64Resolvable)} [icon] The icon for the role
-   * @property {string} [unicodeEmoji] The unicode emoji for the role
+   * @property {?string} [unicodeEmoji] The unicode emoji for the role
    * @property {string} [reason] The reason for creating this role
    */
 
--- a/src/managers/RoleManager.js
+++ b/src/managers/RoleManager.js
@@ -106,7 +106,7 @@ class RoleManager extends CachedManager {
    * @property {number} [position] The position of the new role
    * @property {boolean} [mentionable] Whether or not the new role should be mentionable
    * @property {?(BufferResolvable|Base64Resolvable|EmojiResolvable)} [icon] The icon for the role
-   * @property {?string} [unicodeEmoji] The name of the unicode emoji for the role
+   * @property {?string} [unicodeEmoji] The unicode emoji for the role
    * @property {string} [reason] The reason for creating this role
    */
 
--- a/src/managers/RoleManager.js
+++ b/src/managers/RoleManager.js
@@ -105,8 +105,8 @@ class RoleManager extends CachedManager {
    * @property {PermissionResolvable} [permissions] The permissions for the new role
    * @property {number} [position] The position of the new role
    * @property {boolean} [mentionable] Whether or not the new role should be mentionable
-   * @property {string} [reason] The reason for creating this role
    * @property {BufferResolvable|Base64Resolvable} [icon] The icon for the role
+   * @property {string} [reason] The reason for creating this role
    */
 
   /**
--- a/src/managers/RoleManager.js
+++ b/src/managers/RoleManager.js
@@ -4,6 +4,7 @@ const { Collection } = require('@discordjs/collection');
 const CachedManager = require('./CachedManager');
 const { TypeError } = require('../errors');
 const Role = require('../structures/Role');
+const DataResolver = require('../util/DataResolver');
 const Permissions = require('../util/Permissions');
 const { resolveColor, setPosition } = require('../util/Util');
 
@@ -105,6 +106,7 @@ class RoleManager extends CachedManager {
    * @property {number} [position] The position of the new role
    * @property {boolean} [mentionable] Whether or not the new role should be mentionable
    * @property {string} [reason] The reason for creating this role
+   * @property {BufferResolvable|Base64Resolvable} [icon] The icon for the role
    */
 
   /**
@@ -128,9 +130,10 @@ class RoleManager extends CachedManager {
    *   .catch(console.error);
    */
   async create(options = {}) {
-    let { name, color, hoist, permissions, position, mentionable, reason } = options;
+    let { name, color, hoist, permissions, position, mentionable, reason, icon } = options;
     color &&= resolveColor(color);
     if (typeof permissions !== 'undefined') permissions = new Permissions(permissions);
+    if (icon) icon = await DataResolver.resolveImage(icon);
 
     const data = await this.client.api.guilds(this.guild.id).roles.post({
       data: {
@@ -139,6 +142,7 @@ class RoleManager extends CachedManager {
         hoist,
         permissions,
         mentionable,
+        icon,
       },
       reason,
     });
@@ -188,6 +192,7 @@ class RoleManager extends CachedManager {
       hoist: data.hoist,
       permissions: typeof data.permissions === 'undefined' ? undefined : new Permissions(data.permissions),
       mentionable: data.mentionable,
+      icon: typeof data.icon === 'undefined' ? undefined : await DataResolver.resolveImage(data.icon),
     };
 
     const d = await this.client.api.guilds(this.guild.id).roles(role.id).patch({ data: _data, reason });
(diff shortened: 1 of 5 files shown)


src/managers/RoleManager.js
```diff
@@ -105,6 +106,7 @@ class RoleManager extends CachedManager {
    * @property {number} [position] The position of the new role
    * @property {boolean} [mentionable] Whether or not the new role should be mentionable
    * @property {string} [reason] The reason for creating this role
+   * @property {BufferResolvable|Base64Resolvable} [icon] The icon for the role
```
ImRodry, line 109: This should at least appear before reason, same for the code right below

src/managers/RoleManager.js
```diff
@@ -188,6 +192,7 @@ class RoleManager extends CachedManager {
       hoist: data.hoist,
       permissions: typeof data.permissions === 'undefined' ? undefined : new Permissions(data.permissions),
       mentionable: data.mentionable,
+      icon: typeof data.icon === 'undefined' ? undefined : await DataResolver.resolveImage(data.icon),
```
ImRodry, line 195: Is passing null any different from undefined here? If not you can just call the resolveImage function

src/util/Constants.js
```diff
@@ -73,6 +73,8 @@ exports.Endpoints = {
         makeImageUrl(`${root}/team-icons/${teamId}/${hash}`, { size, format }),
       Sticker: (stickerId, stickerFormat) =>
         `${root}/stickers/${stickerId}.${stickerFormat === 'LOTTIE' ? 'json' : 'png'}`,
+      // TODO: Need more info
+      RoleIcon: (roleId, hash, format) => `${root}/role-icons/${roleId}/${hash}.${format === 'png' ? 'png' : 'jpeg'}`,
```
ImRodry, line 77: The upstream PR mentioned png/jpeg only but it seems like you can set other formats too so this should be updated to use makeImageUrl instead
https://cdn.discordapp.com/role-icons/860078619648851978/cf3ced8600b777c9486c6d8d84fb4327.webp?size=64
Also did you miss the Role#iconURL method or is this coming later?

ImRodry (commented):
Small nitpicks as I know this is still a WIP

src/managers/RoleManager.js
vladfrangu: null would reset the icon most likely

ImRodry: Then we can probably do this
```suggestion
      icon: (await DataResolver.resolveImage(data.icon)) ?? data.icon,
```

vladfrangu: Isn't that inconsistent then with how every other icon/image-taking methods do it?

ImRodry: I’m not sure but this seems more optimal don’t you think?

src/util/Constants.js
iShibi: yup, I wasn't sure about the info that was available at the moment so I just kept it like that. I did forget about the method tho, so thanks for that.

ImRodry: This can be resolved then I guess

src/util/Constants.js
```diff
@@ -73,6 +73,8 @@ exports.Endpoints = {
         makeImageUrl(`${root}/team-icons/${teamId}/${hash}`, { size, format }),
       Sticker: (stickerId, stickerFormat) =>
         `${root}/stickers/${stickerId}.${stickerFormat === 'LOTTIE' ? 'json' : 'png'}`,
+      RoleIcon: (roleId, hash, format = 'webp', size) =>
```
ImRodry, line 76: Probably out of the scope of this PR but why does every method here default the format to webp? Why not just do that on the makeImageUrl function?

src/managers/RoleManager.js
```diff
@@ -128,9 +130,10 @@ class RoleManager extends CachedManager {
    *   .catch(console.error);
    */
   async create(options = {}) {
-    let { name, color, hoist, permissions, position, mentionable, reason } = options;
+    let { name, color, hoist, permissions, position, mentionable, reason, icon } = options;
     if (color) color = resolveColor(color);
     if (typeof permissions !== 'undefined') permissions = new Permissions(permissions);
+    if (icon) icon = await DataResolver.resolveImage(icon);
```
almeidx, line 136:
```suggestion
    icon &&= await DataResolver.resolveImage(icon);
```

almeidx (changes requested):

src/managers/RoleManager.js
monbrey: Shouldn't this also be done on the color line too?

monbrey: Shouldn't this also be done on the color line too?

NotSugden (commented):
can't you also pass null for the role icon? if so docs should be updated

src/managers/RoleManager.js
NotSugden: could so
```js
data.icon == undefined ? data.icon : ...
```
but honestly thats just a bit weird what is there now works fine

ImRodry: I'm just trying to simplify the code and I think my suggestion works fine since resolveImage will return null if data.icon is falsy, and the nullish coalescing makes sure that the passed value is the value of data.icon in case this is falsy

Lulalaby:
> 
> 
> can't you also pass null for the role icon? if so docs should be updated

Yes u can

src/managers/RoleManager.js
iShibi: Adding it here will bring inconsistency. I think you should open another PR that adds this change throughout the library, so that it can be discussed and tracked at a single place.

src/managers/RoleManager.js
iShibi: same as my above comment: This can be added for the whole lib in a dedicated PR for the sake of discussion and consistency

src/util/Constants.js
iShibi: `makeImageUrl` does default `format` to `webp`. Since, some of the methods there don't call `makeImageUrl`, they have to default it themselves, which seems to be done even for other methods. There is one more inconsistency, the methods are taking format and size sometimes together in an object and sometimes not.

src/managers/RoleManager.js
```diff
@@ -104,6 +105,7 @@ class RoleManager extends CachedManager {
    * @property {PermissionResolvable} [permissions] The permissions for the new role
    * @property {number} [position] The position of the new role
    * @property {boolean} [mentionable] Whether or not the new role should be mentionable
+   * @property {?(BufferResolvable|Base64Resolvable)} [icon] The icon for the role
```
ImRodry, line 108: Why doesn't this support an EmojiResolvable?

ImRodry:
New upstream PR: https://github.com/discord/discord-api-docs/pull/3847

iShibi:
This is ready for review ✨

src/managers/RoleManager.js
iShibi: I believe `unicode_emoji`(128b9c9) resolves this

ImRodry: No it doesn’t, I said guild emojis not unicode emojis. There should be a method that allows you to pass a guild emoji and uploads that as the role icon

Lulalaby: ^ this

iShibi: Added in 4f7c58d

kyranet (approved):

ledlamp:
hi so i guess it's worth mentioning that, so i'm using this and if I .setIcon() to an animated emoji, absolutely nothing happens, not even an error. But in discord ui you can pick an animated emoji and it'll use the first frame.

Lulalaby:
> 
> 
> hi so i guess it's worth mentioning that, so i'm using this and if I .setIcon() to an animated emoji, absolutely nothing happens, not even an error. But in discord ui you can pick an animated emoji and it'll use the first frame.

yeah, that's a client thing.
But yeah, discord seems like not to throw errors

ImRodry:
> hi so i guess it's worth mentioning that, so i'm using this and if I .setIcon() to an animated emoji, absolutely nothing happens, not even an error. But in discord ui you can pick an animated emoji and it'll use the first frame.

Well the fact that the url property on an emoji is a getter and not a function like it is everywhere else makes it hard to add a fix for this. We could just get the png image by calling the CDN endpoint directly because that way we’d always make sure to get the first frame of the emoji

ledlamp:
Nvm it's working now. I found that setting to a gif url works and turns out `member.roles.color.setIcon(guild.emojis.cache.find(x=>x.animated))` works too so I have no idea what happened that time 😐

src/managers/RoleManager.js
```diff
@@ -104,6 +105,8 @@ class RoleManager extends CachedManager {
    * @property {PermissionResolvable} [permissions] The permissions for the new role
    * @property {number} [position] The position of the new role
    * @property {boolean} [mentionable] Whether or not the new role should be mentionable
+   * @property {?(BufferResolvable|Base64Resolvable|EmojiResolvable)} [icon] The icon for the role
+   * @property {?string} [unicodeEmoji] The name of the unicode emoji for the role
```
advaith1, line 109:
```suggestion
   * @property {?string} [unicodeEmoji] The unicode emoji for the role
```

src/structures/Role.js
```diff
@@ -83,6 +83,18 @@ class Role extends Base {
...
     this.deleted = false;
 
+    /**
+     * The icon hash of the role
+     * @type {?string}
+     */
+    this.icon = data.icon;
+
+    /**
+     * The name of the unicode emoji for the role
```
advaith1, line 93:
```suggestion
     * The unicode emoji for the role
```

src/structures/Role.js
```diff
@@ -183,6 +195,8 @@ class Role extends Base {
    * @property {number} [position] The position of the role
    * @property {PermissionResolvable} [permissions] The permissions of the role
    * @property {boolean} [mentionable] Whether or not the role should be mentionable
+   * @property {?(BufferResolvable|Base64Resolvable|EmojiResolvable)} [icon] The icon for the role
+   * @property {?string} [unicodeEmoji] The name of the unicode emoji for the role
```
advaith1, line 199:
```suggestion
   * @property {?string} [unicodeEmoji] The unicode emoji for the role
```

src/structures/Role.js
```diff
@@ -292,6 +306,31 @@ class Role extends Base {
...
+  }
+
+  /**
+   * Sets a new unicode emoji for the role.
+   * @param {?string} unicodeEmoji The name of the new unicode emoji for the role
+   * @param {string} [reason] Reason for changing the role's unicode emoji
+   * @returns {Promise<Role>}
+   * @example
+   * // Set a new unicode emoji for the role
+   * role.setUnicodeEmoji('robot')
```
advaith1, line 326:
```suggestion
   * @param {?string} unicodeEmoji The new unicode emoji for the role
   * @param {string} [reason] Reason for changing the role's unicode emoji
   * @returns {Promise<Role>}
   * @example
   * // Set a new unicode emoji for the role
   * role.setUnicodeEmoji('🤖')
```

advaith1 (changes requested):
unicode_emoji is now the actual emoji, not a text name

src/structures/Role.js
iShibi: heyy, thanks for the suggestions. I committed these but forgot to pull the changes before rebasing. It's lost now 🙇🏼‍♀️

ImRodry: lost? 🤔

src/managers/RoleManager.js
```diff
@@ -128,9 +131,13 @@ class RoleManager extends CachedManager {
    *   .catch(console.error);
    */
   async create(options = {}) {
-    let { name, color, hoist, permissions, position, mentionable, reason } = options;
+    let { name, color, hoist, permissions, position, mentionable, reason, icon, unicodeEmoji } = options;
     if (color) color = resolveColor(color);
     if (typeof permissions !== 'undefined') permissions = new Permissions(permissions);
+    if (icon) {
+      const guildEmojiURL = this.guild.emojis.resolve(icon)?.url;
+      icon = guildEmojiURL ? await DataResolver.resolveImage(guildEmojiURL) : await DataResolver.resolveImage(icon);
```
ImRodry, line 139: If the author shares a server with the client and uses an emoji from that server, this should be able to use that as well
```suggestion
      const emojiURL = this.client.emojis.resolve(icon)?.url;
      icon = emojiURL ? await DataResolver.resolveImage(emojiURL) : await DataResolver.resolveImage(icon);
```

src/managers/RoleManager.js
```diff
@@ -182,12 +191,20 @@ class RoleManager extends CachedManager {
       });
     }
 
+    let icon = data.icon;
+    if (icon) {
+      const guildEmojiURL = this.guild.emojis.resolve(icon)?.url;
+      icon = guildEmojiURL ? await DataResolver.resolveImage(guildEmojiURL) : await DataResolver.resolveImage(icon);
```
ImRodry, line 197: ditto
```suggestion
      const emojiURL = this.guild.emojis.resolve(icon)?.url;
      icon = emojiURL ? await DataResolver.resolveImage(emojiURL) : await DataResolver.resolveImage(icon);
```

ImRodry (changes requested):

src/managers/RoleManager.js
iShibi: change `this.guild.emojis` to `this.client.emojis` like above

iShibi: On second thought, `Client#emojis` seems really expensive and it's gonna be called every time there is an icon param irrespective of whether it's an EmojiResolvable or not. Do we really want to add this when (I assume) Discord client too only shows emojis in the guild?

ImRodry: The discord client does not show any guild emojis all, this is just a convenience. And how is it expensive?

advaith1: the client does show guild emojis, and client.emojis iterates through all guilds to get all of their emojis on every access

ImRodry: The difference would be barely noticeable if you’re just using get though

iCrawl:
This needs a rebase.

src/managers/RoleManager.js
iShibi: Made the behaviour similar to Discord client by only allowing the same guild emojis. Added warnings about this behaviour along with how to use emojis from other guilds as an icon

ImRodry: Sounds good to me

SpaceEEC (approved):

vladfrangu (approved):

iCrawl:
This needs a rebase.

iShibi:
Rebased 😓

PR #6636, opened 2021-09-14
refactor(src/structures): speedy optimizations

vierofernando:

changes in this PR:
- Removed `Boolean()` calls and replace it with the `!!` operator.
- Reduced `VoiceChannel#joinable` to just one line
- Repeated `EventEmitter#on` calls are chained because `EventEmitter#on` returns itself

### DIFF
--- a/src/structures/Interaction.js
+++ b/src/structures/Interaction.js
@@ -110,7 +110,7 @@ class Interaction extends Base {
    * @returns {boolean}
    */
   inGuild() {
-    return Boolean(this.guildId && this.member);
+    return this.guildId && this.member;
   }
 
   /**
--- a/src/structures/Interaction.js
+++ b/src/structures/Interaction.js
@@ -110,7 +110,7 @@ class Interaction extends Base {
    * @returns {boolean}
    */
   inGuild() {
-    return this.guildId && this.member;
+    return !!(this.guildId && this.member);
   }
 
   /**
--- a/src/structures/Message.js
+++ b/src/structures/Message.js
@@ -557,7 +557,7 @@ class Message extends Base {
    * @readonly
    */
   get deletable() {
-    return (
+    return !!(
       !this.deleted &&
         (this.author.id === this.client.user.id ||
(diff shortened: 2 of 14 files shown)


src/structures/Message.js
```diff
@@ -557,10 +557,10 @@ class Message extends Base {
    * @readonly
    */
   get deletable() {
-    return Boolean(
+    return (
```
ImRodry, line 560: Once again this cannot be removed otherwise it might return undefined due to the last check being possibly undefined

ImRodry (changes requested):

src/structures/Interaction.js
```diff
@@ -110,7 +110,7 @@ class Interaction extends Base {
    * @returns {boolean}
    */
   inGuild() {
-    return Boolean(this.guildId && this.member);
+    return this.guildId && this.member;
```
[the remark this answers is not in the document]
vierofernando: already resolved in the third commit

src/structures/Message.js
VoltrexMaster: Can be replaced with `!!` to keep consistency, same for the review below.

ImRodry: Yes that's what I meant

vladfrangu:
Was the change from `Boolean` to `!!` done for perf. reasons? Because they're practically the same, only one is more verbose than the other (and I did try it with other inputs, and the perf change is negligible)
image

For the `VoiceChannel#joinable` change... personally I'm a fan of the previous way since it was more verbose (to me it reads better) and didn't ~~ab~~use ternaries, but I'm certain the sentiment is not shared, so as long as it works the same, 🤷‍♂️ from me.

As for the chained .on calls.. Eh, 🦐, personally I prefer them separate in places like those, but I'm neutral about it as long as it's consistent everywhere else :D

monbrey:
I'm kinda with Vlad on this - though I don't necessarily see any problems with the PR I'm really struggling to see what it actually achieves. This seems more like just refactoring for code style preferences than any actual optimizations.

PR #6637, opened 2021-09-14
refactor(Constants): better type error in cdn endpoints

ImRodry:
This PR mainly improves the error given when a user passes an invalid type to the size parameter in CDN endpoints, so it will now throw a TypeError instead, to prevent confusing errors like this one:
image
Along with that, there were also other minor changes:
- Removed the redundant webp default formats from all functions as the main makeImageUrl function already does this.
- Changed the dynamic if statement to only trigger if the hash starts with a_ to make it more readable
- Set the default value of dynamic in ImageURLOptions in the docs to false, instead of having that information written down.

### DIFF
--- a/src/util/Constants.js
+++ b/src/util/Constants.js
@@ -46,15 +46,15 @@ exports.Endpoints = {
       Asset: name => `${root}/assets/${name}`,
       DefaultAvatar: discriminator => `${root}/embed/avatars/${discriminator}.png`,
       Avatar: (userId, hash, format, size, dynamic = false) => {
-        if (dynamic) format = hash.startsWith('a_') ? 'gif' : format;
+        if (dynamic && hash.startsWith('a_')) format = 'gif';
         return makeImageUrl(`${root}/avatars/${userId}/${hash}`, { format, size });
       },
       Banner: (id, hash, format, size, dynamic = false) => {
-        if (dynamic) format = hash.startsWith('a_') ? 'gif' : format;
+        if (dynamic && hash.startsWith('a_')) format = 'gif';
         return makeImageUrl(`${root}/banners/${id}/${hash}`, { format, size });
       },
       Icon: (guildId, hash, format, size, dynamic = false) => {
-        if (dynamic) format = hash.startsWith('a_') ? 'gif' : format;
+        if (dynamic && hash.startsWith('a_')) format = 'gif';
         return makeImageUrl(`${root}/icons/${guildId}/${hash}`, { format, size });
       },
       AppIcon: (appId, hash, { format, size } = {}) =>
--- a/src/util/Constants.js
+++ b/src/util/Constants.js
@@ -57,10 +57,8 @@ exports.Endpoints = {
         if (dynamic && hash.startsWith('a_')) format = 'gif';
         return makeImageUrl(`${root}/icons/${guildId}/${hash}`, { format, size });
       },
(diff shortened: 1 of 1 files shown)


src/util/Constants.js
```diff
@@ -42,34 +42,33 @@ function makeImageUrl(root, { format = 'webp', size } = {}) {
 exports.Endpoints = {
   CDN(root) {
     return {
-      Emoji: (emojiId, format = 'webp') => `${root}/emojis/${emojiId}.${format}`,
```
vladfrangu, line 45: I think these should stay in the event anyone manually uses the CDN export..at least for those that actually don't use the function in question (so from the looks of things..this one)

vladfrangu (changes requested):

src/util/Constants.js
ImRodry: Oops didn't realize this one didn't use it, will revert

ImRodry: Actually idk why this endpoint doesn't use the makeImageUrl because it should support different sizes, but that would be a semver major so I'll simply revert this change and leave that for v14

src/util/Constants.js
```diff
@@ -45,31 +45,30 @@ exports.Endpoints = {
...
         return makeImageUrl(`${root}/banners/${id}/${hash}`, { format, size });
       },
-      Icon: (guildId, hash, format = 'webp', size, dynamic = false) => {
-        if (dynamic) format = hash.startsWith('a_') ? 'gif' : format;
+      Icon: (guildId, hash, format, size, dynamic = false) => {
+        if (dynamic && hash.startsWith('a_')) format = 'gif';
         return makeImageUrl(`${root}/icons/${guildId}/${hash}`, { format, size });
       },
-      AppIcon: (appId, hash, { format = 'webp', size } = {}) =>
+      AppIcon: (appId, hash, { format, size } = {}) =>
```
kyranet, line 60: If you're going to do this, you can skip the destructure and just pass the object.

kyranet (changes requested):

src/util/Constants.js
ImRodry: Changed all instances where this happened to have an options object instead with no default, since makeImageUrl already has said default

vladfrangu (approved):

kyranet (approved):

iCrawl (approved):

SpaceEEC (approved):

src/util/Constants.js
```diff
@@ -19,6 +19,7 @@ const AllowedImageFormats = ['webp', 'png', 'jpg', 'jpeg', 'gif'];
 const AllowedImageSizes = Array.from({ length: 9 }, (e, i) => 2 ** (i + 4));
 
 function makeImageUrl(root, { format = 'webp', size } = {}) {
+  if (typeof size !== 'number') throw new TypeError('INVALID_TYPE', 'size', 'number');
```
GoldenAngel2, line 22: Did anyone actually test this? 
Because just doing anything without the size field being there errors out. 
Ex: `.iconURL()`, `.avatarURL()`, `.displayAvatarURL()`

cc: @ImRodry @vladfrangu

ImRodry: oh shit

ImRodry: Addressed in #6686, guess I forgot to test not passing a size, sorry!

PR #6640, opened 2021-09-15
feat(ApplicationCommand): add support for channel_types

ckohen:

🚀 

- https://github.com/discord/discord-api-docs/pull/3825

### DIFF
--- a/src/structures/ApplicationCommand.js
+++ b/src/structures/ApplicationCommand.js
@@ -2,7 +2,7 @@
 
 const Base = require('./Base');
 const ApplicationCommandPermissionsManager = require('../managers/ApplicationCommandPermissionsManager');
-const { ApplicationCommandOptionTypes, ApplicationCommandTypes } = require('../util/Constants');
+const { ApplicationCommandOptionTypes, ApplicationCommandTypes, ChannelTypes } = require('../util/Constants');
 const SnowflakeUtil = require('../util/SnowflakeUtil');
 
 /**
@@ -131,6 +131,12 @@ class ApplicationCommand extends Base {
    * @property {boolean} [required] Whether the option is required
    * @property {ApplicationCommandOptionChoice[]} [choices] The choices of the option for the user to pick from
    * @property {ApplicationCommandOptionData[]} [options] Additional options if this option is a subcommand (group)
+   * @property {ChannelType[]|number[]} [channelTypes] When the option type is channel,
+   * the allowed types of channels that can be selected
+   * @property {number[]} [channel_types] When the option type is channel,
+   * the API data for allowed types of channels that can be selected
+   * <warn>This is provided for compatability with something like `@discordjs/builders`
+   * and will be discarded when `channelTypes` is present</warn>
    */
 
   /**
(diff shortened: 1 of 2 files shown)


src/structures/ApplicationCommand.js
```diff
@@ -131,6 +131,12 @@ class ApplicationCommand extends Base {
    * @property {boolean} [required] Whether the option is required
    * @property {ApplicationCommandOptionChoice[]} [choices] The choices of the option for the user to pick from
    * @property {ApplicationCommandOptionData[]} [options] Additional options if this option is a subcommand (group)
+   * @property {ChannelType[]|number[]} [channelTypes] When the option type is channel,
+   * the allowed types of channels that can be selected
+   * @property {number[]} [channel_types] When the option type is channel,
+   * the API data for allowed types of channels that can be selected
+   * <warn>This is provided for compatability with something like `@discordjs/builders`
```
ThaTiemsz, line 138:
```suggestion
   * <warn>This is provided for compatibility with something like `@discordjs/builders`
```

vladfrangu (approved):
The warn block could use some rephrasing, but otherwise LGTM

typings/index.d.ts
```diff
@@ -3077,10 +3083,11 @@ export interface ApplicationCommandNonOptionsData extends BaseApplicationCommand
 export type ApplicationCommandOptionData =
   | ApplicationCommandSubGroupData
   | ApplicationCommandNonOptionsData
+  | ApplicationCommandChannelOptionData
   | ApplicationCommandChoicesData
   | ApplicationCommandSubCommandData;
 
-export type ApplicationCommandOption = ApplicationCommandOptionData & {
+export type ApplicationCommandOption = Omit<ApplicationCommandOptionData, 'ApplicationCommandChannelOptionRawData'> & {
```
almeidx, line 3090: What is `ApplicationCommandChannelOptionRawData` supposed to be?

ckohen: Oh, that's residual, do need to fix what that was trying to though (the ApplicationCommandOption type shouldn't have channel_types ever)

almeidx: `Omit` is used to remove certain keys from a type. You were likely looking for `Exclude`, which is used for removing a type from a union.

Although, you could just omit the `channel_types` key🤷

ckohen: When it existed it was Exclude, forget to change the key and just put quotes around it 💀. But omitting doesn't work as intended, will figure something out.

ckohen: On second thought, just gonna rewrite this whole union. It causes options to show up on all types when it shouldn't anyways

src/structures/ApplicationCommand.js
```diff
@@ -239,7 +245,8 @@ class ApplicationCommand extends Base {
       (option.required ?? (['SUB_COMMAND', 'SUB_COMMAND_GROUP'].includes(optionType) ? undefined : false)) !==
         existing.required ||
       option.choices?.length !== existing.choices?.length ||
-      option.options?.length !== existing.options?.length
+      option.options?.length !== existing.options?.length ||
+      (option.channelTypes?.length ?? option.channel_types.length) !== existing.channelTypes?.length
```
NotSugden, line 249:
```suggestion
      ((option.channelTypes ?? option.channel_types).length) !== existing.channelTypes?.length
```

ckohen: I was never really a fan of this pattern, but in this case it works.

vladfrangu (approved):

iCrawl (approved):

SpaceEEC (approved):

PR #6641, opened 2021-09-15
docs(Shard): Point to correct events

Jiralite:
The events in the `@link`s didn't actually exist, but now they do!

One cannot do `{@link Client#shardReady}` as this would redirect to the method which doesn't exist. Prefixing `event:` to this will redirect to the correct area (the event). This will require https://github.com/discordjs/website/pull/104 to be merged as this is new parsing.

### DIFF
--- a/src/sharding/Shard.js
+++ b/src/sharding/Shard.js
@@ -311,7 +311,7 @@ class Shard extends EventEmitter {
       if (message._ready) {
         this.ready = true;
         /**
-         * Emitted upon the shard's {@link Client#ready} event.
+         * Emitted upon the shard's {@link Client#event:shardReady} event.
          * @event Shard#ready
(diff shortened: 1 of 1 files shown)


ImRodry:
Don’t you need to change `@emits` tags too or will those be automatically fixed with the other PR?

Jiralite:
> Don’t you need to change `@emits` tags too or will those be automatically fixed with the other PR?

`@emits` tags are for event emitters, so by default, I made them reference only events. This means you don't have to do `@emits Client#event:shardReady`, you can simply do `@emits Client#shardready`. The redirection of these will be fixed with the new parsing!

iCrawl (approved):

SpaceEEC (approved):

PR #6646, opened 2021-09-15
fix: allow null to be passed in order to reset icon/avatar

iShibi:

This PR updates documentation and typings to allow `null` as a valid value for icon/avatar of a `Guild`, `ClientUser`, and `Webhook`.

It also fixes a bug in `ClientUser#edit` and `Guild#edit` where the `icon` option wasn't being resolved, making it impossible to set a new icon using a path to the local file.

⚠️ From the docs for [`Modify Guild Docs`](https://discord.com/developers/docs/resources/guild#modify-guild), it seems that `splash`, `discovery_splash`, and `banner` can also be set to `null`. My test guild doesn't have those features, so I didn't add changes related to them  here. It would be great if someone can confirm it.

### DIFF
--- a/src/managers/GuildManager.js
+++ b/src/managers/GuildManager.js
@@ -146,7 +146,7 @@ class GuildManager extends CachedManager {
    * @property {DefaultMessageNotificationLevel|number} [defaultMessageNotifications] The default message notifications
    * for the guild
    * @property {ExplicitContentFilterLevel} [explicitContentFilter] The explicit content filter level for the guild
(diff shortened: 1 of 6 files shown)


vladfrangu:
> ⚠️ From the docs for `Modify Guild Docs`, it seems that `splash`, `discovery_splash`, and `banner` can also be set to `null`. My test guild doesn't have those features, so I didn't add changes related to them here. It would be great if someone can confirm it.

You're correct on that assumption

iCrawl (approved):

SpaceEEC (approved):

kyranet (approved):

PR #6647, opened 2021-09-15
types: improve audit logs' typings

noftaly:

This PR adds strong typings to Audit Logs. For example, you could fetch them with a type like this:
```ts
const logs1 = await guild.fetchAuditLogs({ type: 'MEMBER_UPDATE' });
// > logs1.target => User
// > logs1.extra => null

const logs2 = await guild.fetchAuditLogs({ type: 'STAGE_INSTANCE_CREATE' });
// > logs2.target => StageInstance
// > logs2.extra => GuildChannel | { id: Snowflake }

// etc.
```

**To think about before merging**

1. ~~Currently, we don't allow `number`s anymore in `GuildAuditLogsFetchOptions.type` because it's not compatible with `GuildAuditLogsAction`.
We could fix that by creating a mapping from numbers to `GuildAuditLogsAction`.
This could be done easily but is not really clean. Tell me what you would prefer, this is why I've made this PR a draft (Crawl told me on discord we should keep numbers, because it would be a too big regression otherwise)~~ 
Fixed by afaed98

2. ~~I chose to write the current code with generics rather than many interfaces out of simplicity. However, this means that the typing of `logs1` will be `GuildAuditLogs<"MEMBER_UPDATE">`, and `logs1.entries.first()` will be `GuildAuditLogsEntry<"MEMBER_UPDATE", "MEMBER_UPDATE", "UPDATE", "USER">` (yeah...)~~
It doesn't seem to bother any of you so I'll consider this as solved :)

As this is my first contribution to the library, don't refrain from suggesting how it could be improved or what I should maybe do differently. I'm happy to discuss! :)

This is "semver:minor"


typings/index.d.ts
```diff
@@ -779,33 +781,148 @@ export class GuildAuditLogs {
...
+      : User
+    : TTargetType extends 'INTEGRATION'
+    ? Integration
+    : TTargetType extends 'CHANNEL' | 'THREAD'
+    ? GuildChannel | ThreadChannel | { id: Snowflake; [x: string]: unknown }
+    : TTargetType extends 'STAGE_INSTANCE'
+    ? StageInstance
+    : TTargetType extends 'STICKER'
+    ? Sticker
+    : Role | GuildEmoji | { id: Snowflake } | null;
```
suneettipirneni, line 924: Instead of using long conditional type chains, you should consider using a mapped type instead where `TYPE` is mapped to `ObjectType` https://www.typescriptlang.org/docs/handbook/2/mapped-types.html

Then easily reference the varient type later on:
`MyMappedType['USER'] // Returns User Type`

This reduces the work that the typechecker needs to do, and makes it less verbose.

suneettipirneni (changes requested):

typings/index.d.ts
suneettipirneni: An example can be found in this diff: https://github.com/discordjs/discord.js/pull/6476

noftaly: Done in https://github.com/noftaly/discord.js/commit/70a4b72b2f30a4eb037639b21c0f8e958d69ba2c

typings/index.d.ts
```diff
@@ -779,33 +781,114 @@ export class GuildAuditLogs {
...
+        | 'MESSAGE_UNPIN'
+        | 'INTEGRATION_DELETE'
+        | 'STAGE_INSTANCE_DELETE'
+        | 'STICKER_DELETE'
+        | 'THREAD_DELETE'
+    ? 'DELETE'
+    : TAction extends
+        | 'GUILD_UPDATE'
+        | 'CHANNEL_UPDATE'
+        | 'CHANNEL_OVERWRITE_UPDATE'
```
suneettipirneni, line 823: These can use mapped types as well, although not as trivial, but can be accomplished like this:

```ts
export type AuditUpdateActions = {
  [key in keyof typeof AuditUpdateTypes]: 'UPDATE';
};

export type AuditDeleteActions = {
  [key in keyof typeof AuditDeleteTypes]: 'DELETE';
};

...

export type AllActions = AuditUpdateActions & AuditDeleteActions & ...;
```

Where `AuditDeleteTypes` and `AuditUpdateTypes` are enums with the subset of values that correspond to `'DELETE'` or `'UPDATE'`

suneettipirneni (changes requested):

typings/index.d.ts
noftaly: Great idea! I've done it in https://github.com/noftaly/discord.js/commit/65f08d25ed00717b458ce2fe15fea2ea17442cbd and expended it to TTargetType, even though I didn't exactly do it as you proposed to prevent too much duplication

suneettipirneni (approved):

iCrawl:
This needs a rebase.

typings/index.d.ts
```diff
@@ -3823,63 +3821,105 @@ export interface GuildApplicationCommandPermissionData {
...
+  STAGE_INSTANCE_CREATE: GuildChannel | { id: Snowflake };
+  STAGE_INSTANCE_DELETE: GuildChannel | { id: Snowflake };
+  STAGE_INSTANCE_UPDATE: GuildChannel | { id: Snowflake };
+}
+
+interface GuildAuditLogsEntryTargetField<TActionType extends GuildAuditLogsActionType> {
+  USER: User | null;
+  GUILD: Guild;
+  WEBHOOK: Webhook;
+  INVITE: Promise<Invite> | { [x: string]: unknown };
```
vladfrangu, line 3909: Isn't it gonna be a resolved Invite, and not a promise when you fetch it?

vladfrangu (changes requested):

typings/index.d.ts
```diff
@@ -3823,63 +3821,105 @@ export interface GuildApplicationCommandPermissionData {
...
+  STAGE_INSTANCE: StageInstance;
+  STICKER: Sticker;
+}
+
+export interface GuildAuditLogsFetchOptions<T extends GuildAuditLogsAction> {
   before?: Snowflake | GuildAuditLogsEntry;
   limit?: number;
   user?: UserResolvable;
-  type?: GuildAuditLogsAction | number;
+  type?: T;
```
SpaceEEC, line 3922: This breaks the ability to use a number, e.g. `guild.fetchAuditLogs({ type: GuildAuditLogs.Actions.BOT_ADD });` (or even a generic `number`)

SpaceEEC (approved):

SpaceEEC (changes requested):
This breaks the ability to use a number, e.g. `guild.fetchAuditLogs({ type: GuildAuditLogs.Actions.BOT_ADD });` (or even a generic `number`)

typings/index.d.ts
noftaly: yes, that's written in my PR message and that's why I need your advice before merging 😄

noftaly:
I added some tests, don't hesitate to tell me if you want more.

I also added support for IDs, I didn't find a better than this 😕
Please tell me what you think, and if you have better ideas! I'll happy to discuss :]

typings/index.d.ts
noftaly: Fixed by afaed98 !

typings/index.d.ts
```diff
@@ -872,7 +872,9 @@ export class Guild extends AnonymousGuild {
   public edit(data: GuildEditData, reason?: string): Promise<Guild>;
   public editWelcomeScreen(data: WelcomeScreenEditData): Promise<WelcomeScreen>;
   public equals(guild: Guild): boolean;
-  public fetchAuditLogs(options?: GuildAuditLogsFetchOptions): Promise<GuildAuditLogs>;
+  public fetchAuditLogs<T extends keyof GuildAuditLogsIds | GuildAuditLogsAction = 'ALL'>(
```
suneettipirneni, line 875: Consider making a type alias for `keyof GuildAuditLogsIds | GuildAuditLogsAction`:

```ts
export type AuditLogsResolvable = keyof GuildAuditLogsIds | GuildAuditLogsAction;
```

suneettipirneni (changes requested):

typings/index.d.ts
noftaly: Done in 626aab1!

typings/tests.ts
```diff
@@ -1117,3 +1120,48 @@ client.on('interactionCreate', async interaction => {
 declare const shard: Shard;
 
 assertType<Promise<number | null>>(shard.eval(c => c.readyTimestamp));
+
+// Test audit logs
+declare const guild: Guild;
+assertType<Promise<GuildAuditLogs<'MEMBER_KICK'>>>(guild.fetchAuditLogs({ type: 'MEMBER_KICK' }));
+assertType<Promise<any>>(guild.fetchAuditLogs({ type: 20 }));
```
vladfrangu, line 1127: Promise any? That doesn't seem right..

vladfrangu (changes requested):

typings/tests.ts
noftaly: Oops it isn't indeed, some leftovers from testings and debugging. Fixed in 2227d1f

kyranet:
There are merge conflicts in this PR.

iCrawl:

typings/index.test-d.ts
```diff
@@ -1062,3 +1065,71 @@ collector.on('end', (collection, reason) => {
 });
 
 expectType<Promise<number | null>>(shard.eval(c => c.readyTimestamp));
+
+// Test audit logs
+declare const guild: Guild;
+expectType<Promise<GuildAuditLogs<'MEMBER_KICK'>>>(guild.fetchAuditLogs({ type: 'MEMBER_KICK' }));
+expectType<Promise<GuildAuditLogs<20>>>(guild.fetchAuditLogs({ type: GuildAuditLogs.Actions.MEMBER_KICK }));
```
kyranet, line 1072: Are we not able to use the enums from discord-api-types? At least to avoid using magic numbers.

kyranet (changes requested):

typings/index.test-d.ts
noftaly: Fixed in a250ec7 !

noftaly: Uh-oh, forgot to run the tests, but they don't pass anymore...

```
Parameter type Promise<GuildAuditLogs<AuditLogEvent.MemberKick>> is not identical to argument type Promise<GuildAuditLogs<20>>
```
But `AuditLogEvent.MemberKick` *is* 20

kyranet (approved):

vladfrangu (approved):

SpaceEEC (approved):

PR #6649, opened 2021-09-16
fix(Shard): eval promise never resolves

### LINKED ISSUE #6626: Bug: <Shard>.eval(() => someFunction) returns infinite promise
### Issue description
When the function is used in [\<Shard\>](https://github.com/discordjs/discord.js/blob/main/src/sharding/Shard.js).eval(), it returns infinite promise.

Steps to reproduce:
1. Create sharding manager
2. Create some shards
3. Get any shard and try to .eval(() => something) using function

You will get approximately this result:
image

But if you pass a string instead of a function to eval, it will work:
image

Notes: 
* The 'mode' of the sharding manager does not affect whether this bug exists or not.
* \<ShardingManager\>.broadcastEval(() => someFunction, { shard: 0 }) works fine.
* \<Shard\>.eval('string') works fine.
* The 'Discordoo' package used in the screenshots does not affect whether this bug exists or not.
* The number of shards does not affect whether this bug exists or not.

### Codesample

```typescript
// manager.js
const Discord = require('discord.js')

const manager = new Discord.ShardingManager('./shard.js', {
  totalShards: 1,
  token: process.env.TOKEN
})

manager.spawn()
  .then(shards => {
    console.log('getting guilds...')
    shards.get(0).eval(client => client.guilds.cache.size)
      .then(guilds => console.log('guilds:', guilds))
  })

// shard.js
const Discord = require('discord.js')

  intents: [ 'GUILDS' ]
})

client.login(process.env.TOKEN)
  .then(() => {
    console.log('ready!')
  })
```

### discord.js version

13.1.0

### Node.js version

16.6.0

### Operating system

Windows 10 Dev 21390.2025

Low (slightly annoying)

_No response_

GUILDS

discord.js@13.2.0-dev.1631361856.be8912a

DTrombett:
fix #6626 

Passing a function as a parameter in `Shard#eval()` would return a promise that never resolves due to stringifying the function as described in the linked issue.
This PR fixes that problem.

### DIFF
--- a/src/sharding/Shard.js
+++ b/src/sharding/Shard.js
@@ -270,33 +270,35 @@ class Shard extends EventEmitter {
    * @returns {Promise<*>} Result of the script execution
    */
   eval(script) {
+    // Stringify the script if it's a Function
+    const _eval = typeof script === 'function' ? `(${script})(this)` : script;
+
     // Shard is dead (maybe respawning), don't cache anything and error immediately
     if (!this.process && !this.worker) return Promise.reject(new Error('SHARDING_NO_CHILD_EXISTS', this.id));
 
     // Cached promise from previous call
-    if (this._evals.has(script)) return this._evals.get(script);
+    if (this._evals.has(_eval)) return this._evals.get(_eval);
 
     const promise = new Promise((resolve, reject) => {
       const child = this.process ?? this.worker;
 
       const listener = message => {
-        if (message?._eval !== script) return;
+        if (message?._eval !== _eval) return;
         child.removeListener('message', listener);
-        this._evals.delete(script);
+        this._evals.delete(_eval);
         if (!message._error) resolve(message._result);
         else reject(Util.makeError(message._error));
       };
       child.on('message', listener);
 
-      const _eval = typeof script === 'function' ? `(${script})(this)` : script;
       this.send({ _eval }).catch(err => {
         child.removeListener('message', listener);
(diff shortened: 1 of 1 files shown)


src/sharding/Shard.js
```diff
@@ -270,33 +270,35 @@ class Shard extends EventEmitter {
...
   eval(script) {
+    // Stringify the script if it's a Function
+    const _eval = typeof script === 'function' ? `(${script})(this)` : script;
+
     // Shard is dead (maybe respawning), don't cache anything and error immediately
     if (!this.process && !this.worker) return Promise.reject(new Error('SHARDING_NO_CHILD_EXISTS', this.id));
 
     // Cached promise from previous call
-    if (this._evals.has(script)) return this._evals.get(script);
+    if (this._evals.has(_eval)) return this._evals.get(_eval);
```
NotSugden, line 280: completely unrelated but this return should be wrapped in `Promise.resolve(...)` !

DTrombett: Yeah, `Shard#_evals` is a Map of promises so `this._evals.get(_eval)` will return a Promise (maybe already resolved) of the last execution with the same code.

vladfrangu (approved):
Looks fine to me, if you tested it and it works, LGTM

iCrawl (approved):

SpaceEEC (approved):

PR #6659, opened 2021-09-18
chore(IssueForms): remove default titles

ImRodry:
This PR brings 2 changes to issue forms:
- Removed the default titles from both. This is so that users don't accidentally submit an issue without a proper title, so GitHub will alert them if they try to. The type of issue (bug/request) can already be inferred through the labels
- Changed the Node.js version field in the bug report form to mention that you can also provide your typescript version there, if your issue is strictly related to TS, as discussed [on the discord server](https://canary.discord.com/channels/222078108977594368/884983786545238047/888774406128689222)
And also fixes a small typo in "Codesample"

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -60,7 +60,7 @@ body:
       label: Node.js version
       description: |
         Which version of Node.js are you using? Run `node --version` in your project directory and paste the output.
-        If your issue is directly related to TypeScript, please include the version of TypeScript you are using as well. (`npm list typescript`)
+        If you are using TypeScript, please include its version (`npm list typescript`) as well. 
       placeholder: Node.js version 16.6+ is required for version 13.0.0+
     validations:
       required: true
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -60,7 +60,7 @@ body:
       label: Node.js version
       description: |
         Which version of Node.js are you using? Run `node --version` in your project directory and paste the output.
-        If your issue is directly related to Typescript, please include the version of Typescript you are using as well. (`npm list typescript`)
+        If your issue is directly related to TypeScript, please include the version of TypeScript you are using as well. (`npm list typescript`)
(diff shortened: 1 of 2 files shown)


.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -59,7 +58,9 @@ body:
     id: node-version
     attributes:
       label: Node.js version
-      description: Which version of Node.js are you using? Run `node --version` in your project directory and paste the output.
+      description: |
+        Which version of Node.js are you using? Run `node --version` in your project directory and paste the output.
+        If your issue is directly related to TypeScript, please include the version of TypeScript you are using as well. (`npm list typescript`)
```
almostSouji, line 63:
```suggestion
        If you are using TypeScript include its version as well: `npm list typescript`
```
or similar, the proposed version is unnecessarily wordy

almostSouji (changes requested):

.github/ISSUE_TEMPLATE/bug_report.yml
almostSouji: doesn't hurt to provide it and people are often not even sure what the issue is related to. the longer text gets the less likely people are to actually read the instruction and even consider it

ImRodry: Alright then how does this sound
```suggestion
        If you are using TypeScript, please include its version as well. (`npm list typescript`)
```
(Or maybe with npm ls, why don’t we suggest that one instead)

almostSouji: > (Or maybe with npm ls, why don’t we suggest that one instead)

the general paradigm of verbosity, knowing what things do is generally preferable, hence why we generally use `--version` and `install`, rather than the short flags.

parenthesis after the sentence generally refer to a reference so

```suggestion
        If you are using TypeScript, please include its version (`npm list typescript`) as well. 
```
or even
```suggestion
        If you are using TypeScript, please include its version (`npm list typescript`). 
```
dropping the "as well"

ImRodry: went with the first one, thanks

almostSouji (approved):

PR #6668, opened 2021-09-20
types(Interaction): allow `Interaction` cached properties to be type narrowed

suneettipirneni:

Allows you to narrow out a `GuildMember` or a `APIGuildMember` from the `GuildMember | APIGuildMember` union.

```ts
if (!interaction.inCachedGuild()) {
  return;
}

// Member is narrowed to `GuildMember`
const { member } = interaction;
```

Alternatively...
```ts

if (!interaction.inRawGuild() {
  return;
}

// Member is narrowed to `APIGuildMember`
const { member } = interaction;

```

**Also look at changes here: https://github.com/discordjs/discord.js/pull/6668#issuecomment-928168884**

Some have brought up "Why not use `instanceof interaction.member`"

The answer is because it's not as elegant and it doesn't mutate any other types on the interaction such as `guild` and the removal and/or assertion of properties being null. In some circumstances it can be misleading, the methods are more explicit in their purpose.


typings/index.d.ts
```diff
@@ -1022,7 +1021,8 @@ export class Interaction extends Base {
   public type: InteractionType;
   public user: User;
   public version: number;
-  public inGuild(): this is this & { guildId: Snowflake; member: GuildMember | APIInteractionGuildMember };
+  public inGuild(bot?: boolean): this is this & { guildId: Snowflake; member: GuildMember | APIInteractionGuildMember };
+  public inGuild(bot: true): this is this & { guildId: Snowflake; member: GuildMember };
```
vladfrangu, line 1025: These need to be inversed

vladfrangu (changes requested):

typings/index.d.ts
suneettipirneni: I'm confused, if we are checking for the cached guild with the bot and it finds the guild isn't this correct?

NotSugden, line 1025: you can also note that `guild` will be present too, because if `bot` is passed it explictly checks if `guild` is not null

src/structures/Interaction.js
```diff
@@ -107,9 +107,13 @@ class Interaction extends Base {
...
    * Indicates whether this interaction is received from a guild.
+   * @param {boolean} [bot=false] Whether to check if the interaction is from a bot in a guild.
    * @returns {boolean}
    */
-  inGuild() {
+  inGuild(bot = false) {
+    if (bot) {
+      return Boolean(this.guild && this.member);
+    }
     return Boolean(this.guildId && this.member);
```
NotSugden, line 117:
```suggestion
    return Boolean((bot ? this.guild : this.guildId) && this.member);
```

typings/index.d.ts
ImRodry: he's just saying the bot?: boolean overload should always be last

src/structures/Interaction.js
```diff
@@ -107,10 +107,11 @@ class Interaction extends Base {
 
   /**
    * Indicates whether this interaction is received from a guild.
+   * @param {boolean} [bot=false] Whether to check if the interaction is from a bot in a guild.
```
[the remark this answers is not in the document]
Jiralite: Make the parameter called `checkMe` (similar to `Guild#me`), maybe. Also, I would word it as "in the guild" rather than "on the guild".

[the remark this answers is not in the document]
ImRodry: The in/on is always a tricky one for me xD but both sound good and I agree with the checkMe

monbrey:
Although I like the concept of what this checks I can't help but feel the interface / parameter is really awkward to use in practices. Combining the truthy conditions to achieve the assertions, although confusing, would work from what I can see. However this method returning false doesn't seem to have been considered.

`CommandInteraction#inGuild(true); // false` - does this assert that its *not* a `GuildMember`, but it could still be `APIInteractionGuildMember | null`? Used in this way it doesn't actually help you determine if this is a DM or not? I don't even know if you can have falsy assertions.

I think combining these into one method with a boolean parameter is where I'm finding it awkward. I feel like `inGuild()` should clearly assert the difference between a DM and a Guild, then a separate method should assert the existence of a bot user.

ImRodry:
When this was brought up on the discord server I also thought of it as a separate method, the only issue is I wouldn’t know what to call it but if someone does then I think it would be a preferred scenario

suneettipirneni:
Substantial changes have been made since this PR originally was opened, rather than just being able to narrow regular `Interactions` it now allows interaction sub types to be narrowed as well using conditional typings. This allows a much more robust type checking system for interactions. For example let's say I want to know if a command interaction is in a cached guild, given a base `Interaction`:


```ts
// Narrow two levels
if (interaction.isCommand() && interaction.isInCachedGuild()) {
  // interaction is now of type CommandInteraction<'cached'>;
  
  // Because we are in a cached guild, we can intuitively narrow `Message | APIMessage` to `Message`
  const message: Message = interaction.fetchReply();
}
```

It now can not only infer guild specific properties are of a certain type, but can also infer typings for methods specific to `CommandInteraction`.

Alternatively you can use the types to give expectations as to what your function recieves:

```ts
{
  ...
  // The command handler run method only accepted command interactions where the bot is in the same guild.
  run(interaction: CommandInteraction<'cached'>) {
    // No need to verify cached or not at this point, the parameter type guarantees it's in a cached guild.
  }
}
```
This eliminates redundant type checking/casts in areas where the developer knows the guild will be cached.

SpaceEEC (approved):

typings/index.d.ts
```diff
@@ -1014,6 +1033,29 @@ export class Intents extends BitField<IntentsString> {
...
+> = State extends 'cached'
+  ? CachedType
+  : State extends 'raw'
+  ? RawType
+  : State extends 'present'
+  ? PresentType
+  : Fallback;
+
+export interface GuildInteraction<Cached extends GuildCacheState = GuildCacheState> extends Interaction {
+  guildId: CacheTypeReducer<Cached, Snowflake>;
```
vladfrangu, line 1053: Is there any reason you don't just mark this as a Snowflake anyways? If it's a *Guild*Interaction, guildId is always present

typings/tests.ts
```diff
@@ -825,7 +825,29 @@ declare const booleanValue: boolean;
...
+    assertType<GuildMember>(interaction.member);
+  } else if (interaction.inRawGuild()) {
+    assertType<APIInteractionGuildMember>(interaction.member);
+  } else {
+    assertType<APIGuildMember | GuildMember | null>(interaction.member);
+  }
+
+  if (interaction.isContextMenu()) {
+    if (interaction.inCachedGuild()) {
+      assertType<Guild>(interaction.guild);
```
vladfrangu, line 838: Can you also assert that the Interaction is still a contextMenu one? Or better yet, assert that all of these checks work? (aka if you first check if it's a ContextMenu one, and you check if it's cached or not, you still have a ContextMenu one, same for raw, etc)

typings/tests.ts
```diff
@@ -1,4 +1,4 @@
-import { APIInteractionGuildMember } from 'discord-api-types';
+import { APIGuildMember, APIInteractionGuildMember, APIMessage } from 'discord-api-types';
```
vladfrangu, line 1: Why is this not versioned 😢 
I know it falls outside the scope of this PR but please version this

vladfrangu (changes requested):

src/structures/Interaction.js
```diff
@@ -120,6 +120,22 @@ class Interaction extends Base {
     return Boolean(this.guildId && this.member);
   }
 
+  /**
+   * Indicates whether this interaction is both received from a guild and is cached.
```
kyranet, line 124:
```suggestion
   * Indicates whether or not this interaction is both cached and received from a guild.
```

src/structures/Interaction.js
```diff
@@ -120,6 +120,22 @@ class Interaction extends Base {
...
+   * @returns {boolean}
+   */
+  inCachedGuild() {
+    return Boolean(this.guild && this.member);
+  }
+
+  /**
+   * Indicates whether this interaction is both received from a guild and is not cached.
```
kyranet, line 132:
```suggestion
   * Indicates whether or not this interaction is received from an uncached guild.
```

kyranet (changes requested):

vladfrangu (approved):

iCrawl (approved):

kyranet (approved):

PR #6672, opened 2021-09-22
feat: add support for autocomplete interactions

OfficialSirH:
- [Document Application Commands Autocompletion](https://github.com/discord/discord-api-docs/pull/3849)
 

### DIFF
--- a/src/structures/AutocompleteInteraction.js
+++ b/src/structures/AutocompleteInteraction.js
@@ -1,0 +1,26 @@
+'use strict';
+
+const CommandInteractionOptionResolver = require('./CommandInteractionOptionResolver');
+const Interaction = require('./Interaction');
+
+/**
+ * Represents a autocomplete interaction.
+ * @extends {Interaction}
+ */
+class AutocompleteInteraction extends Interaction {
+  constructor(client, data) {
+    super(client, data);
+
+    /**
+     * The options passed to the command.
+     * @type {CommandInteractionOptionResolver}
+     */
+    this.options = new CommandInteractionOptionResolver(
+      this.client,
+      data.data.options?.map(option => this.transformOption(option, data.data.resolved)) ?? [],
+      this.transformResolved(data.data.resolved ?? {}),
+    );
+  }
+}
+
+module.exports = AutocompleteInteraction;
--- a/src/structures/CommandInteractionOptionResolver.js
+++ b/src/structures/CommandInteractionOptionResolver.js
@@ -239,6 +239,18 @@ class CommandInteractionOptionResolver {
     const option = this._getTypedOption(name, '_MESSAGE', ['message'], required);
     return option?.message ?? null;
   }
+
+  /**
+   * Gets a focused option.
+   * @param {name} name The name of the option.
+   * @param {boolean} [required=false] Whether to throw an error if the option is not found.
+   * @returns {?(string)}
+   * The value of the option, or null if not set and not required.
+   */
+  getFocused(name, required = false) {
+    const option = this._getTypedOption(name, 'STRING', ['value'], required);
+    return option?.value ?? null;
+  }
 }
 
 module.exports = CommandInteractionOptionResolver;
--- a/src/structures/Interaction.js
+++ b/src/structures/Interaction.js
@@ -130,6 +130,17 @@ class Interaction extends Base {
   }
 
   /**
+   * Indicates whether this interaction is a {@link AutocompleteInteraction}
+   * @returns {boolean}
+   */
+  isAutocomplete() {
+    return (
+      InteractionTypes[this.type] === InteractionTypes.APPLICATION_COMMAND_AUTOCOMPLETE &&
+      typeof this.targetId === 'undefined'
+    );
+  }
+
+  /**
    * Indicates whether this interaction is a {@link MessageComponentInteraction}.
    * @returns {boolean}
    */
--- a/src/structures/interfaces/InteractionResponses.js
+++ b/src/structures/interfaces/InteractionResponses.js
@@ -218,6 +218,35 @@ class InteractionResponses {
     return options.fetchReply ? this.fetchReply() : undefined;
   }
 
+  /**
+   * Sends results for the autocomplete of this interaction.
+   * @param {InteractionResultOptions} options The options for the autocomplete
+   * @returns {Promise<void>}
+   * @example
+   * // respond to autocomplete interaction
+   * interaction.autocomplete({
+   *  options: [
+   *   {
+   *    name: 'Option 1',
+   *    value: 'option1',
+   *   },
+   *  ],
+   * })
+   *  .then(console.log)
+   * .catch(console.error);
+   */
+  async result(options) {
+    if (this.replied) throw new Error('INTERACTION_ALREADY_REPLIED');
+
+    await this.client.api.interactions(this.id, this.token).callback.post({
+      data: {
+        type: InteractionResponseTypes.APPLICATION_COMMAND_AUTOCOMPLETE_RESULT,
+        data: options,
+      },
+    });
+    this.replied = true;
+  }
+
   static applyToClass(structure, ignore = []) {
     const props = [
       'deferReply',
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -557,6 +557,11 @@ export class CommandInteraction extends BaseCommandInteraction {
   public options: CommandInteractionOptionResolver;
 }
 
+export class AutocompleteInteraction extends Interaction {
+  public options: CommandInteractionOptionResolver;
+  public result: (name: string, required: boolean) => Promise<string | null>;
+}
+
 export class CommandInteractionOptionResolver {
   public constructor(client: Client, options: CommandInteractionOption[], resolved: CommandInteractionResolvedData);
   public readonly client: Client;
@@ -611,6 +616,7 @@ export class CommandInteractionOptionResolver {
   ): NonNullable<CommandInteractionOption['member' | 'role' | 'user']> | null;
   public getMessage(name: string, required: true): NonNullable<CommandInteractionOption['message']>;
   public getMessage(name: string, required?: boolean): NonNullable<CommandInteractionOption['message']> | null;
+  public getFocused(name: string, require?: boolean): string | null;
 }
 
 export class ContextMenuInteraction extends BaseCommandInteraction {
@@ -1025,6 +1031,7 @@ export class Interaction extends Base {
   public inGuild(): this is this & { guildId: Snowflake; member: GuildMember | APIInteractionGuildMember };
   public isButton(): this is ButtonInteraction;
   public isCommand(): this is CommandInteraction;
+  public isAutocomplete(): this is AutocompleteInteraction;
   public isContextMenu(): this is ContextMenuInteraction;
   public isMessageComponent(): this is MessageComponentInteraction;
   public isSelectMenu(): this is SelectMenuInteraction;
@@ -3432,6 +3439,8 @@ export interface CommandInteractionOption {
   name: string;
   type: ApplicationCommandOptionType;
   value?: string | number | boolean;
+  autocomplete?: boolean;
+  focused?: boolean;
   options?: CommandInteractionOption[];
   user?: User;
   member?: GuildMember | APIInteractionDataResolvedGuildMember;
@@ -4071,6 +4080,10 @@ export type InteractionType = keyof typeof InteractionTypes;
 
 export interface InteractionUpdateOptions extends MessageEditOptions {
   fetchReply?: boolean;
+}
+
+export interface InteractionResultOptions {
+  options: CommandInteractionOption[];
 }
 
 export type IntentsString =
--- a/src/structures/BaseCommandInteraction.js
+++ b/src/structures/BaseCommandInteraction.js
@@ -162,6 +162,7 @@ class BaseCommandInteraction extends Interaction {
 
     if ('value' in option) result.value = option.value;
     if ('options' in option) result.options = option.options.map(opt => this.transformOption(opt, resolved));
+    if ('focused' in option) result.focused = option.focused;
 
     if (resolved) {
       const user = resolved.users?.[option.value];
--- a/src/util/Constants.js
+++ b/src/util/Constants.js
@@ -954,7 +954,13 @@ exports.ApplicationCommandPermissionTypes = createEnum([null, 'ROLE', 'USER']);
  * @typedef {string} InteractionType
  * @see {@link https://discord.com/developers/docs/interactions/receiving-and-responding#interaction-object-interaction-type}
  */
-exports.InteractionTypes = createEnum([null, 'PING', 'APPLICATION_COMMAND', 'MESSAGE_COMPONENT']);
+exports.InteractionTypes = createEnum([
+  null,
+  'PING',
+  'APPLICATION_COMMAND',
+  'MESSAGE_COMPONENT',
+  'APPLICATION_COMMAND_AUTOCOMPLETE',
+]);
 
 /**
  * The type of an interaction response:
@@ -975,6 +981,7 @@ exports.InteractionResponseTypes = createEnum([
   'DEFERRED_CHANNEL_MESSAGE_WITH_SOURCE',
   'DEFERRED_MESSAGE_UPDATE',
   'UPDATE_MESSAGE',
+  'APPLICATION_COMMAND_AUTOCOMPLETE_RESULT',
 ]);
 /* eslint-enable max-len */
 
--- a/typings/enums.d.ts
+++ b/typings/enums.d.ts
@@ -66,12 +66,14 @@ export const enum InteractionResponseTypes {
   DEFERRED_CHANNEL_MESSAGE_WITH_SOURCE = 5,
   DEFERRED_MESSAGE_UPDATE = 6,
   UPDATE_MESSAGE = 7,
+  APPLICATION_COMMAND_AUTOCOMPLETE_RESULT = 8,
 }
 
 export const enum InteractionTypes {
   PING = 1,
   APPLICATION_COMMAND = 2,
   MESSAGE_COMPONENT = 3,
+  APPLICATION_COMMAND_AUTOCOMPLETE = 4,
 }
 
 export const enum InviteTargetType {


src/structures/AutocompleteInteraction.js
```diff
@@ -0,0 +1,26 @@
+'use strict';
+
+const CommandInteractionOptionResolver = require('./CommandInteractionOptionResolver');
+const Interaction = require('./Interaction');
+
+/**
+ * Represents a autocomplete interaction.
+ * @extends {Interaction}
+ */
+class AutocompleteInteraction extends Interaction {
```
ckohen, line 10: Is this tested? `transformOption` and `transformResolved` don't exist on Interaction.

src/structures/CommandInteractionOptionResolver.js
```diff
@@ -239,6 +239,18 @@ class CommandInteractionOptionResolver {
...
+
+  /**
+   * Gets a focused option.
+   * @param {name} name The name of the option.
+   * @param {boolean} [required=false] Whether to throw an error if the option is not found.
+   * @returns {?(string)}
+   * The value of the option, or null if not set and not required.
+   */
+  getFocused(name, required = false) {
+    const option = this._getTypedOption(name, 'STRING', ['value'], required);
```
ckohen, line 251: The option is not always going to be a string type. It can be any type that accepts the choices / autocomplete parameter

src/structures/Interaction.js
```diff
@@ -129,6 +129,17 @@ class Interaction extends Base {
...
   }
 
+  /**
+   * Indicates whether this interaction is a {@link AutocompleteInteraction}
+   * @returns {boolean}
+   */
+  isAutocomplete() {
+    return (
+      InteractionTypes[this.type] === InteractionTypes.APPLICATION_COMMAND_AUTOCOMPLETE &&
+      typeof this.targetId === 'undefined'
```
ckohen, line 139: No need for the targetId check on this one for now.

src/structures/interfaces/InteractionResponses.js
```diff
@@ -218,6 +218,35 @@ class InteractionResponses {
     return options.fetchReply ? this.fetchReply() : undefined;
   }
 
+  /**
+   * Sends results for the autocomplete of this interaction.
+   * @param {InteractionResultOptions} options The options for the autocomplete
+   * @returns {Promise<void>}
+   * @example
+   * // respond to autocomplete interaction
+   * interaction.autocomplete({
```
ckohen, line 227: This example doesn't match the function name, which I don't love right now either. Maybe `sendResult` or something. Not the best at naming things so I'd wait for more input on that though.

src/structures/interfaces/InteractionResponses.js
```diff
@@ -218,6 +218,35 @@ class InteractionResponses {
...
+   *   {
+   *    name: 'Option 1',
+   *    value: 'option1',
+   *   },
+   *  ],
+   * })
+   *  .then(console.log)
+   * .catch(console.error);
+   */
+  async result(options) {
```
ckohen, line 238: Need to add this to the `applyToClass` function and then go exclude it from other interaction types

ckohen (changes requested):

src/structures/CommandInteractionOptionResolver.js
monbrey: I don't see any reason to make this an option you have to get by name either. There can only ever be one focused option, so I'd design it more like `getSubcommand()` which can only be singular.

src/structures/AutocompleteInteraction.js
monbrey: It might be able to extend `CommandInteraction` based on the params

ckohen: Oh, I don't think the best option is to extend `BaseCommandInteraction` though, there's too many properties that don't apply to this interaction type. Need to find some way to have the relevant properties and methods without too much code duplication.

ckohen: Update, it appears resolved doesn't exist for these interactions, so the code duplication could be minimal. I think that would be the best approach.

src/structures/interfaces/InteractionResponses.js
ckohen: On second thought, this should probably just go directly in the `AutocompleteInteraction` class. That class cannot use any of the response types and this type cannot be used by anything else, so no need to overcomplicate it.

src/structures/AutocompleteInteraction.js
```diff
@@ -0,0 +1,56 @@
+'use strict';
+
+const BaseCommandInteraction = require('./BaseCommandInteraction');
+const CommandInteractionOptionResolver = require('./CommandInteractionOptionResolver');
+const { InteractionResponseTypes } = require('../util/Constants');
+
+/**
+ * Represents a autocomplete interaction.
+ * @extends {Interaction}
```
iShibi, line 9:
```suggestion
 * @extends {BaseCommandInteraction}
```

typings/index.d.ts
```diff
@@ -557,6 +557,11 @@ export class CommandInteraction extends BaseCommandInteraction {
   public options: CommandInteractionOptionResolver;
 }
 
+export class AutocompleteInteraction extends Interaction {
```
iShibi, line 560: This in code extends `BaseCommandInteraction`

iShibi (changes requested):

typings/index.d.ts
```diff
@@ -557,6 +557,11 @@ export class CommandInteraction extends BaseCommandInteraction {
   public options: CommandInteractionOptionResolver;
 }
 
+export class AutocompleteInteraction extends BaseCommandInteraction {
+  public options: CommandInteractionOptionResolver;
+  public result: (options: InteractionResultOptions) => Promise<void>;
```
vladfrangu, line 562:
```suggestion
  public result(options: InteractionResultOptions): Promise<void>;
```

vladfrangu (changes requested):

src/structures/Interaction.js
```diff
@@ -129,6 +129,14 @@ class Interaction extends Base {
     return InteractionTypes[this.type] === InteractionTypes.APPLICATION_COMMAND && typeof this.targetId !== 'undefined';
   }
 
+  /**
+   * Indicates whether this interaction is a {@link AutocompleteInteraction}
```
iShibi, line 133:
```suggestion
   * Indicates whether this interaction is an {@link AutocompleteInteraction}
```

src/structures/CommandInteractionOptionResolver.js
```diff
@@ -239,6 +239,20 @@ class CommandInteractionOptionResolver {
     const option = this._getTypedOption(name, '_MESSAGE', ['message'], required);
     return option?.message ?? null;
   }
+
+  /**
+   * Gets the focused option.
+   * @param {boolean} [required=false] Whether to throw an error if the option is not found.
+   * @returns {?(string|number)}
+   * The value of the option, or null if not set and not required.
```
iShibi, line 247:
```suggestion
   * @param {boolean} [required=false] Whether to throw an error if the option is not found
   * @returns {?(string|number)} The value of the option, or null if not set and not required
```

iShibi (changes requested):

src/structures/AutocompleteInteraction.js
```diff
@@ -0,0 +1,54 @@
...
+/**
+ * Represents an autocomplete interaction.
+ * @extends {BaseCommandInteraction}
+ */
+class AutocompleteInteraction extends BaseCommandInteraction {
+  constructor(client, data) {
+    super(client, data);
+
+    /**
+     * The options passed to the command.
```
iShibi, line 16:
```suggestion
     * The options passed to the command
```

src/structures/CommandInteractionOptionResolver.js
```diff
@@ -239,6 +239,19 @@ class CommandInteractionOptionResolver {
     const option = this._getTypedOption(name, '_MESSAGE', ['message'], required);
     return option?.message ?? null;
   }
+
+  /**
+   * Gets the focused option.
+   * @param {boolean} [required=false] Whether to throw an error if the option is not found.
```
iShibi, line 245:
```suggestion
   * @param {boolean} [required=false] Whether to throw an error if the option is not found
```

greysdawn:
Do you think it would be better for `getFocused()` to return the whole option instead of just the value of it? Seeing as multiple options can have autocomplete on (afaik) and devs will probably want to know which out of all the options is currently focused using that

I know you can also just check if an individual option is `.focused` but I feel like making `getFocused()` shortcut to the full option, or at least including its name, might be better 🤔

src/structures/AutocompleteInteraction.js
```diff
@@ -0,0 +1,56 @@
...
+    this.options = new CommandInteractionOptionResolver(
+      this.client,
+      data.data.options?.map(option => this.transformOption(option, data.data.resolved)) ?? [],
+      this.transformResolved(data.data.resolved ?? {}),
+    );
+  }
+
+  /**
+   * Sends results for the autocomplete of this interaction.
+   * @param {InteractionResultOptions} options The options for the autocomplete
```
Jiralite, line 28: `InteractionResultOptions` needs defining!

Jiralite (changes requested):

src/structures/AutocompleteInteraction.js
OfficialSirH: Could you clarify what you mean? Cause the type can be found on line 4087 in index.d.ts so I'm unsure of what you mean to define it.

Jiralite: This is JSDoc. You've used it without creating a `@typedef` for it, so it won't show up in the documentation. If you search around the repository, you'll see every custom type has an associated `@typedef` for it (and those that don't won't show up on the documentation). All you have to do is create one and you're good to go.

This will actually be the same as [`ApplicationCommandOptionChoice`](https://discord.js.org/#/docs/main/main/typedef/ApplicationCommandOptionChoice), so lucky for you it's just a copy-paste.

OfficialSirH: nevermind, my bad.

src/structures/interfaces/InteractionResponses.js
greysdawn: i think `sendResult` works personally, it's pretty clear and concise  
worst case if something similar to `reply` is preferred, `respond` might work? i still like the first better though

src/structures/AutocompleteInteraction.js
```diff
@@ -0,0 +1,56 @@
...
+
+const BaseCommandInteraction = require('./BaseCommandInteraction');
+const CommandInteractionOptionResolver = require('./CommandInteractionOptionResolver');
+const { InteractionResponseTypes } = require('../util/Constants');
+
+/**
+ * Represents an autocomplete interaction.
+ * @extends {BaseCommandInteraction}
+ */
+class AutocompleteInteraction extends BaseCommandInteraction {
```
ckohen, line 11: I was suggesting to continue extending Interaction and duplicate the few necessary parameters and methods, since webhook, ephemeral, deferred (from BaseCommandInteraction) don't apply here

src/structures/AutocompleteInteraction.js
```diff
@@ -0,0 +1,56 @@
...
+    super(client, data);
+
+    /**
+     * The options passed to the command
+     * @type {CommandInteractionOptionResolver}
+     */
+    this.options = new CommandInteractionOptionResolver(
+      this.client,
+      data.data.options?.map(option => this.transformOption(option, data.data.resolved)) ?? [],
+      this.transformResolved(data.data.resolved ?? {}),
```
ckohen, line 22: We don't get resolved data on this ever right now. It's possible that's a bug, but I doubt it

src/structures/CommandInteractionOptionResolver.js
```diff
@@ -239,6 +239,19 @@ class CommandInteractionOptionResolver {
     const option = this._getTypedOption(name, '_MESSAGE', ['message'], required);
     return option?.message ?? null;
   }
+
+  /**
+   * Gets the focused option.
+   * @param {boolean} [required=false] Whether to throw an error if the option is not found
+   * @returns {?(string|number)} The value of the option, or null if not set and not required
+   */
+  getFocused(required = false) {
```
ckohen, line 248: This is kinda annoying to deal with in typings right now. This function exists on normal command interactions, where focussed options will never exist, but when the interaction type is autocomplete then its always present, so passing true to this function feels wrong (you can't get an autocomplete interaction without a focused option). Maybe this also needs to be part of the `AutocompleteInteraction`? Thoughts?

ckohen (changes requested):
This PR should also be adding support for the autocomplete parameter in ApplicationCommand.options[n]!

src/structures/interfaces/InteractionResponses.js
ImRodry: Why not just follow the reply naming scheme from every other interaction?

src/structures/CommandInteractionOptionResolver.js
ckohen: What I'm saying is the required option is a hack for typings purposes here. There's no real reason to actually use it. And yes that applies to context menu interactions too! We should probably be using type generics instead

ImRodry: I guess since this is the only one that would always be required in an Autocomplete interaction it could be removed without necessarily creating an inconsistency but it should definitely be kept on the option resolver and throw an error when used on other interaction types

src/structures/interfaces/InteractionResponses.js
greysdawn: if you mean naming the function `reply`, tbh i don't think that's very accurate to what the function does  
mostly because `reply` everywhere else implies sending some kind of message that the user will see, not just a list of options they can choose from in an interface that's not very message-like

ImRodry: reply and respond are synonyms

greysdawn: and i said i personally preferred `sendResult` over respond for that exact reason 😅  
i'm easy either way on it, just explaining my line of thought with not naming it `reply`

ImRodry: Yeah sure then, sendResult sounds good

ImRodry:
> Do you think it would be better for `getFocused()` to return the whole option instead of just the value of it? Seeing as multiple options can have autocomplete on (afaik) and devs will probably want to know which out of all the options is currently focused using that
> 
> I know you can also just check if an individual option is `.focused` but I feel like making `getFocused()` shortcut to the full option, or at least including its name, might be better 🤔

As for this it could probably be toggleable through a parameter like sendFull and, like one of the discussions above mentioned, the required parameter should be removed since it doesn't make much sense to be there, and should just be replaced by this one IMO

src/structures/AutocompleteInteraction.js
```diff
@@ -0,0 +1,89 @@
...
+   * interaction.sendResult([
+   *  {
+   *    name: 'Option 1',
+   *    value: 'option1',
+   *  },
+   * ])
+   *  .then(console.log)
+   *  .catch(console.error);
+   */
+  async result(options) {
```
vladfrangu, line 74: Example isn't identical with example. Also.. the naming is a bit weird, I'd personally name it respond. Thoughts @discordjs/the-big-3?

src/structures/CommandInteractionOptionResolver.js
```diff
@@ -239,6 +239,18 @@ class CommandInteractionOptionResolver {
     const option = this._getTypedOption(name, '_MESSAGE', ['message'], required);
     return option?.message ?? null;
   }
+
+  /**
+   * Gets the focused option.
+   * @param {boolean} [getFull=false] Whether to throw an error if the option is not found
```
vladfrangu, line 245: Not at all what this option represents, please update the comment

src/structures/CommandInteractionOptionResolver.js
```diff
@@ -239,6 +239,18 @@ class CommandInteractionOptionResolver {
...
+   * @returns {string|number|ApplicationCommandOptionChoice}
+   * The value of the option, or the whole option if getFull is true
+   */
+  getFocused(getFull = false) {
+    const focusedOption = this._hoistedOptions.find(option => option.focused);
+    if (!focusedOption) throw new TypeError('AUTOCOMPLETE_INTERACTION_OPTION_NO_FOCUSED_OPTION');
```
vladfrangu, line 251: You need to define this message in Errors.js

typings/index.d.ts
```diff
@@ -557,6 +557,12 @@ export class CommandInteraction extends BaseCommandInteraction {
   public options: CommandInteractionOptionResolver;
 }
 
+export class AutocompleteInteraction extends Interaction {
+  public options: CommandInteractionOptionResolver;
+  private transformOption(option: APIApplicationCommandOption, resolved: null): CommandInteractionOption;
+  public sendResult(options: ApplicationCommandOptionChoice[]): Promise<void>;
```
vladfrangu, line 563: More inconsistency

vladfrangu (changes requested):

typings/index.d.ts
```diff
@@ -611,10 +625,20 @@ export class CommandInteractionOptionResolver {
   ): NonNullable<CommandInteractionOption['member' | 'role' | 'user']> | null;
   public getMessage(name: string, required: true): NonNullable<CommandInteractionOption['message']>;
   public getMessage(name: string, required?: boolean): NonNullable<CommandInteractionOption['message']> | null;
+  public getFocused<
+    T extends InteractionOptionResolverReturn<Type, AutocompleteInteraction, ApplicationCommandOptionChoice>,
+  >(getFull: true): T;
+  public getFocused<T extends InteractionOptionResolverReturn<Type, AutocompleteInteraction, string | number>>(
+    getFull?: boolean,
+  ): T;
```
OfficialSirH, line 633: > What I'm saying is the required option is a hack for typings purposes here. There's no real reason to actually use it. And yes that applies to context menu interactions too! We should probably be using type generics instead

@ckohen Would this type generic work out for what you were suggesting or do you have a better suggestion for this?

SpaceEEC (changes requested):
I think `autocomplete` needs to be compared in `ApplicationCommand#_optionEquals` as well.

typings/index.d.ts
ckohen: Yes! this works great The functions themselves don't need generics though, just move them like so (not sure why the diff looks so weird here).

```suggestion
 public getFocused(
   getFull: true
 ): InteractionOptionResolverReturn<Type, AutocompleteInteraction, ApplicationCommandOptionChoice>;
 public getFocused(
    getFull?: boolean,
  ): InteractionOptionResolverReturn<Type, AutocompleteInteraction, string | number>;
```

kyranet (approved):

Rodentman87 (approved):

src/structures/AutocompleteInteraction.js
```diff
@@ -0,0 +1,114 @@
...
+     * @type {?TextBasedChannels}
+     * @name AutocompleteInteraction#channel
+     * @readonly
+     */
+
+    /**
+     * The id of the channel this interaction was sent in
+     * @type {Snowflake}
+     * @name AutocompleteInteraction#channelId
+     */
```
vladfrangu, line 26: Is this actually needed?

src/structures/AutocompleteInteraction.js
```diff
@@ -0,0 +1,114 @@
...
+    return this.guild?.commands.cache.get(id) ?? this.client.application.commands.cache.get(id) ?? null;
+  }
+
+  /**
+   * Transforms an option received from the API.
+   * @param {APIApplicationCommandOption} option The received option
+   * @returns {CommandInteractionOption}
+   * @private
+   */
+  transformOption(option) {
```
vladfrangu, line 71: Shouldn't this be static?

vladfrangu (approved):

vladfrangu (changes requested):
Pressed wrong button, comments above still are valid

src/structures/AutocompleteInteraction.js
SpaceEEC: I think so, `AutocompleteInteraction` always has a channelId, which is not guaranteed in `Interaction`.

src/structures/AutocompleteInteraction.js
SpaceEEC: It's not static in `BaseCommandInteraction` either, so probably not.

src/structures/AutocompleteInteraction.js
```diff
@@ -0,0 +1,114 @@
...
+class AutocompleteInteraction extends Interaction {
+  constructor(client, data) {
+    super(client, data);
+
+    /**
+     * The channel this interaction was sent in
+     * @type {?TextBasedChannels}
+     * @name AutocompleteInteraction#channel
+     * @readonly
+     */
```
SpaceEEC, line 20: This is redundant isn't it? It's identical to the type in the base class:
https://github.com/discordjs/discord.js/blob/34b2ad0d8ee39eb9cc964a4334a803b2ab48cae5/src/structures/Interaction.js#L97-L101

src/structures/AutocompleteInteraction.js
```diff
@@ -0,0 +1,114 @@
...
+   * interaction.respond([
+   *  {
+   *    name: 'Option 1',
+   *    value: 'option1',
+   *  },
+   * ])
+   *  .then(console.log)
+   *  .catch(console.error);
+   */
+  async respond(options) {
```
SpaceEEC, line 99: What's the reason to name this method `respond` instead of `reply`, which is what `InteractionResponses` is using?

vladfrangu (approved):
LGTM after a rebase and upstream lands

SpaceEEC (changes requested):

src/structures/AutocompleteInteraction.js
almeidx: https://github.com/discordjs/discord.js/pull/6672#discussion_r714726081

greysdawn: there's a [conversation](https://github.com/discordjs/discord.js/pull/6672#discussion_r713621675) about it towards the top of the PR, consensus was pretty much "reply doesn't seem quite right here, respond is close so i suppose it works"

it's mostly because it isn't sending content the command's users will see as a message (or, well, reply), so using the same naming scheme doesn't seem right. it also doesn't accept the same data as the other reply functions, so naming it something else should be less confusing for lib users

(just my 2¢ at least)

src/structures/AutocompleteInteraction.js
OfficialSirH: If it's redundant, wouldn't BaseCommandInteraction also be doing something redundant since it also has it despite it already being in the base class?

SpaceEEC: Probably

ImRodry:
Hey @OfficialSirH I think there was a problem in your rebase, could you try again?

OfficialSirH:
> Hey @OfficialSirH I think there was a problem in your rebase, could you try again?

What's the problem with the rebase?

almeidx:
> > Hey @OfficialSirH I think there was a problem in your rebase, could you try again?
> 
> What's the problem with the rebase?

Look at the Files Changed tab

iCrawl:
Thats normal, because it was a merge not a rebase.

PR #6675, opened 2021-09-22
fix(GuildEmoji): Cache restricted roles and author data

Jiralite:
Guild emojis that were restricted to roles were not being stored in discord.js over the gateway - an empty array would always be set. As such, methods like `GuildEmojiRoleManager.add()` would completely wipe all other roles from the emoji (since it calls `.set()` from the cache which is empty), and then only add what you specified, which is bad behaviour and not intended.

I believe I managed to fix this by setting the default value to the provided array. I have tested this and it works as expected.

### DIFF
--- a/src/structures/GuildEmoji.js
+++ b/src/structures/GuildEmoji.js
@@ -30,7 +30,7 @@ class GuildEmoji extends BaseGuildEmoji {
      * @type {Snowflake[]}
      * @private
      */
-    Object.defineProperty(this, '_roles', { value: [], writable: true });
+    Object.defineProperty(this, '_roles', { value: data.roles, writable: true });
   }
 
   /**
--- a/src/structures/GuildEmoji.js
+++ b/src/structures/GuildEmoji.js
@@ -30,7 +30,9 @@ class GuildEmoji extends BaseGuildEmoji {
      * @type {Snowflake[]}
      * @private
      */
-    Object.defineProperty(this, '_roles', { value: data.roles, writable: true });
+    Object.defineProperty(this, '_roles', { value: [], writable: true });
+
+    this._patch(data);
   }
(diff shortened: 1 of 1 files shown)


src/structures/GuildEmoji.js
```diff
@@ -30,7 +30,7 @@ class GuildEmoji extends BaseGuildEmoji {
      * @type {Snowflake[]}
      * @private
      */
-    Object.defineProperty(this, '_roles', { value: [], writable: true });
```
vladfrangu, line 33: Oh, this happens because _patch is called before defining the field? Is that the case? If _patch isn't called at all, thats the bug that should be fixed instead 👀

Jiralite: Ah yeah, always calling `._patch()` will also correct the author data if not known beforehand - thank you! Should I leave `data.roles` as the default value here, as it seems to be always present (and an empty array if no roles are provided)?

vladfrangu: I think you can leave it as an empty array by default and call _patch, do test this tho

Jiralite: Empty array or default array whilst calling `._patch()` works fine in the event that:
1. Emoji is simply not cached but then fetched from the manager (role and author data is populated as expected)
2. Emoji is received from the gateway (author data not present and is `null` as expected, roles populated as expected)
3. Emoji is cached without author data, but fetching with `{ force: true }` or fetching the author populates the `.author` property

I'll amend `data.roles` back to `[]` and call `this._patch()`!

vladfrangu (approved):

SpaceEEC (approved):

iCrawl (approved):

PR #6679, opened 2021-09-23
fix(types): make transformCommand public

wutv:

Since the method is static, and can be used for more than internal code, it should be classified as public. 


monbrey:
In theory, but it's not quite that simple. By making it public, we're no longer able to change or replace it in non-major version since it becomes a public-facing part of the library's interface.

suneettipirneni:
This was something that I wanted to utilize and expose in the proposed `type-utils` [module](https://github.com/discordjs/discord.js-modules/pull/57), however it's still not confirmed whether this will be a thing or not.

vladfrangu:
I agree with @monbrey here, might be better as a separate util package instead

Issue #6681, opened 2021-09-23
Bug: Resolving src/client/actions via filesystems breaks bundlers


alula:
### Issue description

Some "smart" person decided that resolving modules statically is too spammy and replaced it with code that scans filesystem at one point: https://github.com/discordjs/discord.js/pull/6102

Therefore using discord.js with bundlers like webpack breaks d.js completely and causes very unexpected behavior, since everything is actually bundled in a single .js file, `__dirname` points to current directory and modules the code is looking for do not actually exist on the filesystem:

```
  18:18:32.925  worker    ERROR    Whoops: Error: Cannot find module './.env.example'
    at webpackContextResolve (webpack://kyoko/./.yarn/cache/discord.js-npm-13.1.0-8440b31213-90f629be62.zip/node_modules/discord.js/src/client/actions/_sync_^\.\/.*$?:109:11)
    at webpackContext (webpack://kyoko/./.yarn/cache/discord.js-npm-13.1.0-8440b31213-90f629be62.zip/node_modules/discord.js/src/client/actions/_sync_^\.\/.*$?:104:11)
    at new ActionsManager (webpack://kyoko/./.yarn/cache/discord.js-npm-13.1.0-8440b31213-90f629be62.zip/node_modules/discord.js/src/client/actions/ActionsManager.js?:13:174)
    at new Client (webpack://kyoko/./.yarn/cache/discord.js-npm-13.1.0-8440b31213-90f629be62.zip/node_modules/discord.js/src/client/Client.js?:100:20)
    at MockMonitorClient.eval (webpack://kyoko/./src/worker/index.ts?:114:24)
    at MockMonitorClient.start (webpack://kyoko/./src/worker/index.ts?:61:85)
    at eval (webpack://kyoko/./src/worker/index.ts?:147:19)
    at eval (webpack://kyoko/./src/worker/index.ts?:148:3)
    at Object../src/worker/index.ts (/home/alula/workspace/kyoko/kyoko-v4/dist/worker.js:118:1)
    at __webpack_require__ (/home/alula/workspace/kyoko/kyoko-v4/dist/worker.js:6662:42) {
  code: 'MODULE_NOT_FOUND'
}
```

The bug can be very easily resolved by reverting #6102.

### Codesample

```typescript
Not applicable
```

### discord.js version

13.1.0

### Node.js version

16.10

### Operating system

Linux

### Priority this issue should have

Medium (should be fixed soon)

### Which partials do you have configured?

No Partials

[...]

PR #6102, opened 2021-07-12
refactor(ActionsManager): less complex code

xHyroM:
I've looked at the `ActionsManager.js` file, and saw that it was pretty impractical. So, here's my easier and way less complex version using `fs`.

**Status and versioning classification:**
- Code changes have been tested against the Discord API, or there are no code changes
- I know how to update typings and have done so, or typings don't need updating
- This PR changes the library's interface (methods or parameters added)

PR #6682, opened 2021-09-23
Revert #6102 (see #6681)

alula:

Revert changes made in pull request #6102, which causes bundlers like webpack to break (#6681).

### DIFF
--- a/src/client/actions/ActionsManager.js
+++ b/src/client/actions/ActionsManager.js
@@ -6,12 +6,49 @@ class ActionsManager {
   constructor(client) {
     this.client = client;
 
-    const files = fs.readdirSync(__dirname);
-
-    for (const file of files) {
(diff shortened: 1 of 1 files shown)


vladfrangu:
Why? We do not support webpack and have stopped supporting it ever since we removed browser support.

alula:
This is a pretty trivial change and Webpack isn't only used in browser environments.

alula:
actually for some context read from here: https://discord.com/channels/222078108977594368/682166281826598932/890684079186530364

alula:
The raised concern seems to be caused by "less maintainable code" and "often forgetting to export things". 
I proposed a slightly different approach - add tests that check if all modules in the directory are referenced properly - which also might seem help with similar issues in other parts of the library.

Issue #6667, opened 2021-09-20
Bug: cache fetch option not working as intended


KokoNeot:
### Issue description

1. Fetch all invites for specific guild
2. Use some of the invites from the guild
3. get the old invites from the cache, fetch the guild invites again after joining with invites.fetch({ cache: false }) 
4. see that the new fetch returns what is already in the cache but not the newly fetched data/invites

### Codesample

```typescript
const { Client, Collection } = require('Discord.js')
const bot = new Client({ intents: ["GUILD_INVITES", "GUILD_MEMBERS"] })

const inviteCache = new Collection()

bot.on('inviteCreate', async inv => {
    inviteCache.set(inv.guild.id, await inv.guild.invites.fetch())
})

bot.on('guildMemberAdd', async member => {
    const cachedInvites = bot.invites.get(member.guild.id)

    console.log(cachedInvites.get('invCode').uses) //If new invite then expected = 0, recieved = 0

    member.guild.invites.fetch({ cache: false }).then(newInvites => {
        console.log(cachedInvites.get('invCode').uses) //expected = 0, recieved = 1
        console.log(newInvites.get('invCode').uses)  //expected = 1, recieved = 1
    })
})
```

### discord.js version

13.1.0

### Node.js version

16.6.1

### Operating system

Windows 10 pro

### Priority this issue should have

Medium (should be fixed soon)

### Which partials do you have configured?

USER, CHANNEL, MESSAGE, REACTION

### Which gateway intents are you subscribing to?

GUILDS, GUILD_MEMBERS, GUILD_INVITES, GUILD_VOICE_STATES, GUILD_MESSAGES, GUILD_MESSAGE_REACTIONS, DIRECT_MESSAGES, DIRECT_MESSAGE_REACTIONS

### I have tested this issue on a development release

I did not test as I dont know how :(

KokoNeot:
I'm not familiar with this new bug form so I'm gonna add here additional information

on this line it returns whatever is in the cache and not the new data. Because if cache is set to false the new data wont be stored in the cache so its not returned and it returns the old data which are already in the cache.
https://github.com/discordjs/discord.js/blob/d0bc4d7ff037035d4f315a0286b06f0a1ade29c1/src/managers/CachedManager.js#L49

and if this is really a bug then it might be with other fetch methods which accept cache option

vladfrangu:
[...]

PR #6685, opened 2021-09-24
fix(CachedManager): return updated data when cache is false

iShibi:

fixes: #6667

Under the circumstances where a structure is present in the cache, `<Manager>_add()` returns stale data if `cache` is set to false instead of returning the updated data without caching it.


SpaceEEC (changes requested):
`_patch` does not necessarily return the updated structure, you need to explicitly return the appropriate instance. 😅

KokoNeot:
Thanks for the fix!

vladfrangu (approved):

iCrawl (approved):

SpaceEEC (approved):

PR #6686, opened 2021-09-24
fix(Constants): allow undefined size

ImRodry:
Fixes a bug accidentally introduced by #6637 where an error would be thrown when passing an undefined size (this is what happens when you only test the behavior you're looking for and not other scenarios 😅)

### DIFF
--- a/src/util/Constants.js
+++ b/src/util/Constants.js
@@ -19,7 +19,7 @@ const AllowedImageFormats = ['webp', 'png', 'jpg', 'jpeg', 'gif'];
 const AllowedImageSizes = Array.from({ length: 9 }, (e, i) => 2 ** (i + 4));
 
 function makeImageUrl(root, { format = 'webp', size } = {}) {
-  if (typeof size !== 'number') throw new TypeError('INVALID_TYPE', 'size', 'number');
+  if (size && typeof size !== 'number') throw new TypeError('INVALID_TYPE', 'size', 'number');
   if (format && !AllowedImageFormats.includes(format)) throw new Error('IMAGE_FORMAT', format);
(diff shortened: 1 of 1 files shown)


GoldenAngel2 (approved):
👍

src/util/Constants.js
```diff
@@ -19,7 +19,7 @@ const AllowedImageFormats = ['webp', 'png', 'jpg', 'jpeg', 'gif'];
 const AllowedImageSizes = Array.from({ length: 9 }, (e, i) => 2 ** (i + 4));
 
 function makeImageUrl(root, { format = 'webp', size } = {}) {
-  if (typeof size !== 'number') throw new TypeError('INVALID_TYPE', 'size', 'number');
+  if (size && typeof size !== 'number') throw new TypeError('INVALID_TYPE', 'size', 'number');
```
vladfrangu, line 22:
```suggestion
  if (typeof size !== 'undefined' && typeof size !== 'number') throw new TypeError('INVALID_TYPE', 'size', 'number');
```

vladfrangu (changes requested):

src/util/Constants.js
ImRodry: Does this work too? Looks cleaner imo
```suggestion
  if (!['undefined', 'number'].includes(typeof size)) throw new TypeError('INVALID_TYPE', 'size', 'number');
```

vladfrangu: Sure, that works too

vladfrangu (approved):

SpaceEEC (approved):

iCrawl (approved):

PR #6689, opened 2021-09-25
fix

Cod-USER:
fix interface PartialChannelData


Jiralite:
#6683 addresses issues with this interface. You can suggest changes over there!

PR #6694, opened 2021-09-25
fix: don't patch set data with undefined

### LINKED ISSUE #6688: Bug: Voice channel loses properties after being received as interaction option
### Issue description

1. Fetch a voice channel.
2. Execute an interaction that has channel parameters and do a getChannel of the same channel.
3. Observe that some properties of the voice channel were lost (they existed before and ceased to exist).

### Codesample

```typescript
// In the ready event.

const channel = await client.channels.fetch("699644022804840468");
console.log(channel.bitrate); // In my case it returns 128000

// In the interactionCreate event.

const { id } = interaction.options.getChannel("channel");
const channel = await client.channels.fetch(id);
console.log(channel.bitrate); // Returns undefined.
```

### discord.js version

13.2.0-dev

### Node.js version

16.6.2

### Operating system

Ubuntu 18 Server

Medium (should be fixed soon)

No Partials

GUILDS, GUILD_MEMBERS, GUILD_EMOJIS_AND_STICKERS, GUILD_INVITES, GUILD_MESSAGES, DIRECT_MESSAGES

839974c

ImRodry:
This PR fixes a bug across the entire library where some properties could be lost by calling ._patch. The diff is quite large because this fix was applied to all _patch methods on the library on all properties except ids. If you do not agree with some of the changes, feel free to leave a review and we can discuss them as I understand some of these may be redundant, but it's always better to be sure. Fixes #6688 

### DIFF
--- a/src/structures/AnonymousGuild.js
+++ b/src/structures/AnonymousGuild.js
@@ -15,36 +15,47 @@ class AnonymousGuild extends BaseGuild {
   }
 
   _patch(data) {
-    this.features = data.features;
-    /**
-     * The hash of the guild invite splash image
-     * @type {?string}
-     */
-    this.splash = data.splash;
+    if ('features' in data) this.features = data.features;
 
-    /**
-     * The hash of the guild banner
-     * @type {?string}
-     */
-    this.banner = data.banner;
+    if ('splash' in data) {
+      /**
+       * The hash of the guild invite splash image
+       * @type {?string}
+       */
+      this.splash = data.splash;
+    }
 
-    /**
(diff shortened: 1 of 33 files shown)


kyranet (approved):

iCrawl (approved):

ImRodry:
Rebased and decided not to apply the `'ephemeral' in data` check for MessageAttachment#ephemeral since that is intentionally not sent for the library to infer that the property is false, so it should always be set

SpaceEEC (approved):

vladfrangu (approved):

PR #6695, opened 2021-09-26
feat: Allow webhooks to fetch, edit and delete messages in threads

### LINKED ISSUE #6592: Bug: `Webhook.editMessage()` into a thread doesn't work
### Issue description

With a webhook (created by the client) I send a message in a thread -> Everything is fine
```javascript
webhook.send({content: firstContent, threadId: id});
```

But later I come to edit this message -> Everything is bad :sweat_smile: 

```javascript
let msg = channel.messages.fetch(msgId)
webhook.editMessage(msg, {content: newContent});
```

But it works if the message is not in a thread and I did not find anything in the discord.js doc for this "problem"
image

### Codesample

```typescript
function sendMsg(message, threadId)
{
	let hook = message.channel.fetchWebhooks().first()
	hook.send({content: message.content, threadId: threadId})
}

async function editMsg(message, messageOldId)
{
	let msgOld = await message.channel.messages.fetch(messageOldId);
	let hook = await msgOld.fetchWebhook()
	hook.editMessage(msgOld, {content: message.content});
}
```

### discord.js version

13.0.1

### Node.js version

16.6+

### Operating system

Windows

Medium (should be fixed soon)

_No response_

GUILDS, GUILD_MESSAGES

_No response_

Jiralite:
This will allow webhooks to fetch, edit and delete messages in threads.

Webhook#fetchMessage() unfortunately didn't already have an object for its second parameter. I would have liked to change it to what I commented [here](https://github.com/discordjs/discord.js/issues/6592#issuecomment-927191151) but to keep it not breaking, I instead added a new client option for webhooks that as a sort of opt-in feature flag that makes the library parse the second parameter as an object, thus allowing access to fetching a message in a thread.

Here's an example:
```JavaScript
const { WebhookClient } = require("discord.js");

const webhook1 = new WebhookClient({
  url: "webhook_url1"
});

const webhook2 = new WebhookClient({
  url: "webhook_url2"
}, {
  allowWebhookThreadFetching: true // With this enabled, I can now fetch messages in threads!
});

webhook1.fetchMessage("message_id1", false).then(console.log); // Success!
// No error or change in code as we did not opt-in to fetching in threads. Parameter is unchanged.

webhook2.fetchMessage("message_id2", {
  cache: false,
  threadId: "thread_id"
}).then(console.log); // Success!
```

This should resolve #6592, additional. I am open to adjustments and feedback!

### DIFF
--- a/src/structures/Webhook.js
+++ b/src/structures/Webhook.js
@@ -257,7 +257,7 @@ class Webhook {
   async fetchMessage(message, cacheOrOptions) {
     const options = { cache: true, threadId: undefined };
 
-    if (this.client.options.allowedWebhookThreadFetching) {
+    if (this.client.options.allowWebhookThreadFetching) {
       options.cache = cacheOrOptions.cache ?? true;
       options.threadId = cacheOrOptions.threadId;
     } else {
--- a/src/structures/Webhook.js
+++ b/src/structures/Webhook.js
@@ -105,6 +105,8 @@ class Webhook {
    * @property {MessageAttachment[]} [attachments] Attachments to send with the message
    * @property {MessageActionRow[]|MessageActionRowOptions[]} [components]
    * Action rows containing interactive components for the message (buttons, select menus)
+   * @property {Snowflake} [threadId] The id of the thread this message belongs to
+   * <info>For interaction webhooks, this property is ignored</info>
    */
 
   /**
@@ -236,17 +238,43 @@ class Webhook {
   }
 
   /**
+   * Options that can be passed to fetch a message.
+   * @typedef {options} WebhookFetchMessageOptions
+   * @property {boolean} [cache=true] Whether to cache the message.
+   * @property {Snowflake} [threadId] The id of the thread this message belongs to.
+   * <info>For interaction webhooks, this property is ignored</info>
+   */
+
+  /**
    * Gets a message that was sent by this webhook.
    * @param {Snowflake|'@original'} message The id of the message to fetch
-   * @param {boolean} [cache=true] Whether to cache the message
+   * @param {boolean|WebhookFetchMessageOptions} [cacheOrOptions=true|cacheOrOptions={}] If
+   * {@link ClientOptions#allowWebhookThreadFetching} is `true`, this will be {@link WebhookFetchMessageOptions}
+   * to enable fetching messages in threads. Otherwise, this will be the `cache` parameter.
    * @returns {Promise<Message|APIMessage>} Returns the raw message data if the webhook was instantiated as a
    * {@link WebhookClient} or if the channel is uncached, otherwise a {@link Message} will be returned
    */
-  async fetchMessage(message, cache = true) {
+  async fetchMessage(message, cacheOrOptions) {
+    const options = { cache: true, threadId: undefined };
+
+    if (this.client.options.allowedWebhookThreadFetching) {
+      options.cache = cacheOrOptions.cache ?? true;
+      options.threadId = cacheOrOptions.threadId;
+    } else {
+      options.cache = cacheOrOptions ?? true;
+    }
+
     if (!this.token) throw new Error('WEBHOOK_TOKEN_UNAVAILABLE');
 
-    const data = await this.client.api.webhooks(this.id, this.token).messages(message).get();
-    return this.client.channels?.cache.get(data.channel_id)?.messages._add(data, cache) ?? data;
+    const data = await this.client.api
+      .webhooks(this.id, this.token)
+      .messages(message)
+      .get({
+        query: {
+          thread_id: options.threadId,
+        },
+      });
+    return this.client.channels?.cache.get(data.channel_id)?.messages._add(data, options.cache) ?? data;
   }
 
   /**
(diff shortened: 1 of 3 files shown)


src/structures/Webhook.js
```diff
@@ -235,18 +237,44 @@ class Webhook {
...
    * @param {Snowflake|'@original'} message The id of the message to fetch
-   * @param {boolean} [cache=true] Whether to cache the message
+   * @param {boolean|WebhookFetchMessageOptions} [cacheOrOptions=true|cacheOrOptions={}] If
+   * {@link ClientOptions#allowWebhookThreadFetching} is `true`, this will be {@link WebhookFetchMessageOptions}
+   * to enable fetching messages in threads. Otherwise, this will be the `cache` parameter.
    * @returns {Promise<Message|APIMessage>} Returns the raw message data if the webhook was instantiated as a
    * {@link WebhookClient} or if the channel is uncached, otherwise a {@link Message} will be returned
    */
-  async fetchMessage(message, cache = true) {
+  async fetchMessage(message, cacheOrOptions) {
```
ImRodry, line 257: I really do not like this approach mainly because it’s hard to accurately display in typings and also unnecessarily complex. You should just add a third parameter and then change this in v14, maybe add a warning saying it will change

monbrey:
What's the point of the client option? Why would we ever NOT want to allow this?

src/structures/Webhook.js
monbrey: You can't add a third option nicely here in a non-breaking way - it would have to come after `cache` which currently has a default, forcing them to always pass true/false in order to push to that third option. 

I think this is just best served by being a v14 PR either way.

ckohen: How about...keep this options, default it to `{ cache: true }`. Then instead of a client option, just accept the legacy option type by utilizing a `typeof options === 'object'`, and if its not let the old code run

ImRodry: Or you could add the third parameter and pass undefined to the second one if you wanna use the default, I’m pretty sure that would work anyway

monbrey: That a horrible interface for the user.

Jiralite:
I have gone with @ckohen's desired approach. Feel free to re-review!

src/structures/Webhook.js
```diff
@@ -235,18 +237,40 @@ class Webhook {
...
+   * @property {Snowflake} [threadId] The id of the thread this message belongs to.
+   * <info>For interaction webhooks, this property is ignored</info>
+   */
+
   /**
    * Gets a message that was sent by this webhook.
    * @param {Snowflake|'@original'} message The id of the message to fetch
-   * @param {boolean} [cache=true] Whether to cache the message
+   * @param {WebhookFetchMessageOptions|boolean} [cacheOrOptions={}] The options to provide to fetch the message.
+   * A boolean may be passed instead to specify whether to cache the message.
```
ckohen, line 252: Something to indicate that a boolean option is deprecated would be nice

typings/index.d.ts
```diff
@@ -2825,12 +2825,15 @@ export function WebhookMixin<T>(Base?: Constructable<T>): Constructable<T & Webh
...
+  deleteMessage(message: MessageResolvable | APIMessage | '@original', threadId?: Snowflake): Promise<void>;
   editMessage(
     message: MessageResolvable | '@original',
     options: string | MessagePayload | WebhookEditMessageOptions,
   ): Promise<Message | APIMessage>;
-  fetchMessage(message: Snowflake | '@original', cache?: boolean): Promise<Message | APIMessage>;
+  fetchMessage(
+    message: Snowflake | '@original',
+    cacheOrOptions?: WebhookFetchMessageOptions | boolean,
+  ): Promise<Message | APIMessage>;
```
ckohen, line 2836: If tslint lets you, splitting this overload out would be nice so that you can `@deprecated` the boolean overload

ckohen (changes requested):

typings/index.d.ts
Jiralite: I was only able to do this by disabling the `unified-signatures` rule.

SpaceEEC (approved):

src/structures/Webhook.js
```diff
@@ -235,18 +237,40 @@ class Webhook {
...
    * @returns {Promise<Message|APIMessage>} Returns the raw message data if the webhook was instantiated as a
    * {@link WebhookClient} or if the channel is uncached, otherwise a {@link Message} will be returned
    */
-  async fetchMessage(message, cache = true) {
+  async fetchMessage(message, cacheOrOptions = { cache: true }) {
+    if (typeof cacheOrOptions === 'object') {
+      cacheOrOptions.cache ??= true;
+    } else {
+      cacheOrOptions = { cache: cacheOrOptions };
+    }
```
kyranet, line 261: Not a fan of mutating objects, we can do this instead:
```suggestion
    if (typeof cacheOrOptions === 'boolean') {
      cacheOrOptions = { cache: cacheOrOptions };
    }
```

We don't need to default `options.cache` to `true`, because it's defaulted to that same value in the manager:

https://github.com/discordjs/discord.js/blob/3b14883e349c23a15aec225bb133ce49ed440817/src/managers/CachedManager.js#L46

kyranet (changes requested):

kyranet (approved):

ckohen (approved):

PR #6701, opened 2021-09-27
types(ChannelTypes): remove unknown from CHANNEL options

ImRodry:
This PR simply removes the ability to put UNKNOWN in the channelTypes array in a CHANNEL option on a chat input application command.

### DIFF
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -3080,8 +3080,8 @@ export type ApplicationCommandData =
 
 export interface ApplicationCommandChannelOptionData extends BaseApplicationCommandOptionsData {
   type: CommandOptionChannelResolvableType;
-  channelTypes?: (keyof typeof ChannelTypes | ChannelTypes)[];
-  channel_types?: ChannelTypes[];
+  channelTypes?: Exclude<keyof typeof ChannelTypes | ChannelTypes, 'UNKNOWN' | ChannelTypes.UNKNOWN>[];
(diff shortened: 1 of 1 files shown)


ImRodry:
Gonna bring this here from Discord so that the discussion isn't lost.
I removed the UNKNOWN type because passing this actually throws an error on Discord's end but I believe DM and GROUP_DM should be removed too because passing these with no other type causes the option to be unusable because you will be left with no channels to choose from, however, the API accepts both of these types, even by themselves, so I'd like to know from @discordjs/the-big-3 if I should remove those too or let them stay

kyranet (approved):

iCrawl (approved):

SpaceEEC (approved):

ImRodry:
> Gonna bring this here from [Discord](https://canary.discord.com/channels/222078108977594368/682166281826598932/892072942735859824) so that the discussion isn't lost. I removed the UNKNOWN type because passing this actually throws an error on Discord's end but I believe DM and GROUP_DM should be removed too because passing these with no other type causes the option to be unusable because you will be left with no channels to choose from, however, the API accepts both of these types, even by themselves, so I'd like to know from @discordjs/the-big-3 if I should remove those too or let them stay
> 
> EDIT: having a CHANNEL option with the only DM in the channelTypes array doesn't accept DM channel ids either, causing the following error to be thrown: image

So I guess we’re allowing dm channel types to be sent here?

SpaceEEC:
I'm in favor of not validating those and let Discord do that.

vladfrangu (approved):

PR #6710, opened 2021-09-28
refactor: use the `node:` protocol

almeidx:
- https://github.com/nodejs/node/issues/38343
- https://nodejs.org/api/modules.html#modules_core_modules

### DIFF
--- a/src/client/BaseClient.js
+++ b/src/client/BaseClient.js
@@ -1,6 +1,6 @@
 'use strict';
 
-const EventEmitter = require('events');
+const EventEmitter = require('node:events');
 const RESTManager = require('../rest/RESTManager');
 const Options = require('../util/Options');
 const Util = require('../util/Util');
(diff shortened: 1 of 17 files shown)


kyranet (approved):

SpaceEEC (approved):

ckohen:
If we're going with this, could we add the [eslint plugin](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/prefer-node-protocol.md) as well? The plugin has a lot of rules and we'd only be using one though, so more thoughts needed.

almeidx:
> If we're going with this, could we add the [eslint plugin](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/prefer-node-protocol.md) as well? The plugin has a lot of rules and we'd only be using one though, so more thoughts needed.

Would I enable it on this pull request?

ckohen:
Yes, if that is the decision, it should be implemented here

vladfrangu (approved):

iCrawl (approved):

PR #6715, opened 2021-09-29
types(Vanity): Make `uses` non-nullable

Jiralite:
Guild#fetchVanityData() seems to always return a number for `uses`, even in the event that:

- A guild has no vanity URL access
- A guild has vanity URL access but has not set it
- A guild has a vanity URL with no uses

In the above cases, Discord returns 0 for `uses`, so this can never be `null`.


SpaceEEC (approved):

ImRodry (approved):
Title should say something along the lines of `typings(Vanity): make uses non-nullable`

kyranet (approved):

vladfrangu (approved):

iCrawl (approved):

PR #6721, opened 2021-09-30
refactor(User): set accentColor and banner to undefined when not yet received

ImRodry:
This PR changes the accentColor and banner properties in User to default to undefined when they're not present in the payload since they need to be force-fetched for the client to receive them. I'm doing this because these properties can actually be null after being fetched (meaning they're not present) and that would make it hard to tell those two scenarios apart. To go along with these changes I also made hexAccentColor return accentColor when this is falsy (so either null or undefined, to avoid duplicating code) and made User#bannerURL throw an error when being called before the user is fetched. Despite the new error, this should not be semver major as these properties are not live on stable yet.

### DIFF
--- a/src/structures/User.js
+++ b/src/structures/User.js
@@ -89,7 +89,7 @@ class User extends Base {
        * @type {?number}
        */
       this.accentColor = data.accent_color;
-    } else if (this.banner !== null) {
+    } else if (this.accentColor !== null) {
       this.accentColor ??= undefined;
     }
 
--- a/src/structures/User.js
+++ b/src/structures/User.js
@@ -175,7 +175,7 @@ class User extends Base {
    * @readonly
    */
   get hexAccentColor() {
-    if (!this.accentColor && typeof this.accentColor !== 'number') return this.accentColor;
+    if (typeof this.accentColor !== 'number') return this.accentColor;
     return `#${this.accentColor.toString(16).padStart(6, '0')}`;
   }
 
--- a/src/structures/User.js
+++ b/src/structures/User.js
@@ -74,23 +74,23 @@ class User extends Base {
     if ('banner' in data) {
       /**
        * The user banner's hash
-       * <info>The user must be force fetched for this property to be present</info>
+       * <info>The user must be force fetched for this property to be present or be updated</info>
        * @type {?string}
        */
       this.banner = data.banner;
-    } else {
-      this.banner ??= null;
+    } else if (this.banner !== null) {
+      this.banner ??= undefined;
     }
 
     if ('accent_color' in data) {
       /**
        * The base 10 accent color of the user's banner
-       * <info>The user must be force fetched for this property to be present</info>
+       * <info>The user must be force fetched for this property to be present or be updated</info>
        * @type {?number}
        */
       this.accentColor = data.accent_color;
-    } else {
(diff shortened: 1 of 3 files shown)


ImRodry:
Sorry for the two force-pushes, had some issues in the first versions but it should be all good now!

src/structures/User.js
```diff
@@ -179,17 +179,19 @@ class User extends Base {
    * @readonly
    */
   get hexAccentColor() {
-    if (!this.accentColor) return null;
+    if (!this.accentColor) return this.accentColor;
     return `#${this.accentColor.toString(16).padStart(6, '0')}`;
```
NotSugden, line 183:
```suggestion
    return this.accentColor && `#${this.accentColor.toString(16).padStart(6, '0')}`;
```
i believe this would work too (removing the above line also)

src/structures/User.js
```diff
@@ -78,23 +78,23 @@ class User extends Base {
...
       /**
        * The base 10 accent color of the user's banner
-       * <info>The user must be force fetched for this property to be present</info>
+       * <info>The user must be force fetched for this property to be present or be updated</info>
        * @type {?number}
        */
       this.accentColor = data.accent_color;
-    } else {
-      this.accentColor ??= null;
+    } else if (this.banner !== null) {
```
NotSugden, line 96:
```suggestion
    } else if (this.accentColor !== null) {
```
C&P go brrrrr

NotSugden (changes requested):

src/structures/User.js
NotSugden: actually on second thoughts

if accentColor is 0 then 0 would be returned (in both instances) so perhaps a more robust check is required

ImRodry: I can probably just add a check if accentColor isn’t a number on the first one

src/structures/User.js
ImRodry: Oopsie, you saw nothing

src/structures/User.js
```diff
@@ -179,17 +179,19 @@ class User extends Base {
    * @readonly
    */
   get hexAccentColor() {
-    if (!this.accentColor) return null;
+    if (!this.accentColor && typeof this.accentColor !== 'number') return this.accentColor;
```
vladfrangu, line 182: Couldn't you simplify this to just the typeof check?

vladfrangu (changes requested):

src/structures/User.js
ImRodry: I’d rather not because people could set the accentColor on their own and thus break it a bit, I think this is more secure

SpaceEEC: I don't think we do this anywhere else, do we?
Generally, if you mess with properties you aren't supposed to mess with, expect weird behavior. 🤷

ImRodry: Sure then I guess

vladfrangu (approved):

SpaceEEC (approved):

iCrawl:
This needs a rebase.

ImRodry:
Rebased!

kyranet (approved):

iCrawl:
This needs a rebase.

PR #6734, opened 2021-10-01
fix(ReactionCollector): only call the filter function once

### LINKED ISSUE #6461: ReactionCollector filter is called twice
Upon adding a new reaction to a message with a `ReactionCollector` on it, the collector's filter function is called twice. This issue began with `v13.0.0`. I've checked the source code and found that the filter is called once to determine whether the `create` event should be emitted...
https://github.com/discordjs/discord.js/blob/cfde127072ae06b63f17ba19b3e0d28fddef97d6/src/structures/ReactionCollector.js#L104-L106
...and again in a parent method, before emitting the `collect` event
https://github.com/discordjs/discord.js/blob/cfde127072ae06b63f17ba19b3e0d28fddef97d6/src/structures/interfaces/Collector.js#L100-L108

As a user of the library, it would be ideal if the filter function is only called once per reaction creation. In my situation, I call a third-party API in my collector's filter. The call is only intended to be placed once per reaction collection, not twice.

```js
const { Client, Intents } = require('discord.js');

const client = new Client({ intents: [Intents.FLAGS.GUILDS, Intents.FLAGS.GUILD_MESSAGES, Intents.FLAGS.GUILD_MESSAGE_REACTIONS] });

client.on('messageCreate', (msg) => {
    msg.createReactionCollector({
        filter: function(r, u) {
            console.log('in filter');   //should appear twice
            return false;
        },
    });
});

client.login("DISCORD_TOKEN");
```

**Further details:**

- discord.js version: `13.1.0`
- Node.js version: `16.6.1`
- Operating system: Windows 10 Version 21H1
- Priority this issue should have – please be realistic and elaborate if possible: Low, is not a critical problem

**Relevant client options:**

- partials: none
- gateway intents: Same as in code snippet (`GUILDS`, `GUILD_MESSAGES`, `GUILD_MESSAGE_REACTIONS`)
- other: none

SpaceEEC:

This PR resolved #6461 by only calling the filter function once.

### DIFF
--- a/src/structures/ReactionCollector.js
+++ b/src/structures/ReactionCollector.js
@@ -67,6 +67,17 @@ class ReactionCollector extends Collector {
     });
 
     this.on('collect', (reaction, user) => {
+      /**
+       * Emitted whenever a reaction is newly created on a message. Will emit only when a new reaction is
+       * added to the message, as opposed to {@link Collector#collect} which which will
+       * be emitted even when a reaction has already been added to the message.
+       * @event ReactionCollector#create
+       * @param {MessageReaction} reaction The reaction that was added
+       * @param {User} user The user that added the reaction
+       */
+      if (reaction.count === 1) {
+        this.emit('create', reaction, user);
+      }
       this.total++;
       this.users.set(user.id, user);
     });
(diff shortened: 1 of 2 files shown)


src/structures/ReactionCollector.js
```diff
@@ -81,10 +92,10 @@ class ReactionCollector extends Collector {
    * Handles an incoming reaction for possible collection.
    * @param {MessageReaction} reaction The reaction to possibly collect
    * @param {User} user The user that added the reaction
```
vaporox, line 94: This line should be removed (and the method signature needs to be changed in the typings too)

SpaceEEC: I'd argue that it should just stay there: The parameter is still there, since the event emits them both.

vaporox: It's still there for the `collect` event, but not for the `collect` method :eyes:

iCrawl (approved):

vladfrangu (approved):

Issue #6707, opened 2021-09-28
Add an event for getting 


ThatSkyz:
### Feature

I'm always frustrated when I cannot easily debug issues regarding http requests and rate limits in Discord.js when requests and responses are send and received.

### Ideal solution or implementation

I would like to see something such as the following:
```js
client.on('httpRequest', (request, response) => {
    if (!request.path.startsWith('/api/v9/channels/1234/messages') {
        // or whatever, I'm not 100% sure on how it works internally in Discord.js
        return undefined;
    }
    console.log(request.path, response.statusCode, response.headers, response.body); // Just an example
});
```

### Alternative solutions or implementations

I manually patch the library to log information required when I am trying to debug issues regarding my bot.

Megatank58:
You should be able to use `client.ws.on(...)`

PR #6739, opened 2021-10-02
feat(Client): add apiResponse event for debugging

ckohen:

Adds an event emitted on every response received from the API, closes: #6707 
This is quite useful for debugging purposes, especially when trying to figure out if something may be a Discord bug and not on the library side.

Some things to consider:

- Name of the event, I feel like this is accurate now, but naming things is hard
- In order to accomplish this, I've added docs / types to APIRequest, which makes changing it potentially breaking (e.g. if we wanted to fully replace rest with `@discordjs/rest`). To avoid this I could pull out specific properties from the request and only emit those, bonus to this being less chance of mutating (only matters when the request needs to be retried).
- In order to type the response, I had to add `@types/node-fetch` (since we don't target DOM), which does not have a version matching the version in the library since v3 has built in types. This could easily be solved by #6586 or updating node-fetch


typings/index.d.ts
```diff
@@ -182,6 +183,18 @@ export abstract class AnonymousGuild extends BaseGuild {
   public splashURL(options?: StaticImageURLOptions): string | null;
 }
 
+export class APIRequest {
+  private constructor(rest: unknown, method: string, path: string, options: unknown);
+  public client: Client;
+  public method: string;
```
iShibi, line 189: This could be a union of these: https://github.com/discordjs/discord.js/blob/994abf509e6b10b536c23f928f100307f2dc0109/src/rest/APIRouter.js#L4

ckohen: That's not *strictly* true though. Someone could do the unthinkable and call `client.rest.request('options', ...` for example. I don't necessarily disagree, though because that's an unsupported use case I suppose.

src/rest/APIRequest.js
```diff
@@ -80,3 +107,8 @@ class APIRequest {
 }
 
 module.exports = APIRequest;
+
+/**
+ * @external HTTPMethod
```
ImRodry, line 112: Shouldn't this be declared on the website instead, just like other things like Function and string are?

src/rest/RequestHandler.js
```diff
@@ -176,6 +176,19 @@ class RequestHandler {
       return this.execute(request);
     }
 
+    if (this.manager.client.listenerCount(API_RESPONSE)) {
+      /**
+       * Emitted after every api request has received a response.
```
ImRodry, line 181:
```suggestion
       * Emitted after every API request has received a response.
```

src/rest/RequestHandler.js
```diff
@@ -315,3 +328,8 @@ class RequestHandler {
 }
 
 module.exports = RequestHandler;
+
+/**
+ * @external Response
```
ImRodry, line 333: Same as the mozilla link above

typings/index.d.ts
ImRodry: I remember also making a PR to node-fetch to specify the types of methods and they said they supported all methods, case insensitive, so this should still remain as string imo

typings/index.d.ts
```diff
@@ -182,6 +183,18 @@ export abstract class AnonymousGuild extends BaseGuild {
...
+  public options: unknown;
+  public path: string;
+  private rest: unknown;
+  public retries: number;
+  public route: string;
+  private make(): Promise<Response>;
```
ImRodry, line 195: Could you place private methods and properties at the top and separate them with a new line like in other classes?

src/rest/APIRequest.js
ckohen: I didn't think so because they are not under the javascript scope. However, even if we end up adding it via the website, I would like to keep them here until said PR is merged.

ImRodry: Let's hear what @iCrawl thinks then

typings/index.d.ts
ckohen: ngl, I'm fully anticipating making this an interface per point 2, just wanted to gather thoughts first, so I didn't bother formatting it yet.

ImRodry: Personally I think this is fine as it is atm but sure, let's wait for more feedback on that

SpaceEEC (changes requested):
I think the event name is fine.

I'd be in favor of extracting relevant properties of `APIRequest` into an interface instead of documenting and emitting it.

This also does not seem to address the request for an `apiRequest` (when a request is being made) event, which was part of the linked issue. :eyes:

src/rest/RequestHandler.js
```diff
@@ -176,6 +176,19 @@ class RequestHandler {
...
+      /**
+       * Emitted after every api request has received a response.
+       * This event does not necessarily correlate to completion of the request, e.g. when hitting a rate limit.
+       * <info>This is an informational event that is emitted quite frequently,
+       * it is highly recommended to check `request.path` to filter the data.</info>
+       * @event Client#apiResponse
+       * @param {APIRequest} request The request that triggered this response
+       * @param {Response} response The response received from the discord API
+       */
+      this.manager.client.emit(API_RESPONSE, request, res);
```
vladfrangu, line 189: I think we shouldn't emit the response we get but a clone of it (res.clone()), to allow users to tap into the raw data without breaking our code due to the data being read already

vladfrangu (changes requested):
Main issue I see is below. Also, I agree with @SpaceEEC, we shouldn't document the full internal class we have but just an "object" of it (honestly you might be able to just do `{ ...request }` as well), because users shouldn't call `make()` themselves

ThatSkyz:
> This also does not seem to address the request for an `apiRequest` (when a request is being made) event, which was part of the linked issue. 👀

The `apiResponse` event is what I was looking for, though it would be ideal to also have an `apiRequest` event too since some use cases might not require the response and I'm not sure what would happen if the request actually fails (if the event would get emitted or not).

KhafraDev:
> which does not have a version matching the version in the library since v3 has built in types. This could easily be solved by refactor: node-fetch & @discordjs/formdata -> undici #6586 or updating node-fetch

Discord.js cannot upgrade to node-fetch v3 due to it being ESM-only whilst this project is CJS.

typings/index.d.ts
vBread: `string` can still be added to a union while preserving autocompletion through a type.

```ts
type LiteralUnion<T> = T | (string & { _?: never });

declare const method: LiteralUnion<'get' | 'post' | 'delete' | 'patch' | 'put'>;
```

src/rest/RequestHandler.js
ckohen: Agh, yes yes, I totally forgot that was an option, and its exactly what I was looking for

ckohen:
> Discord.js cannot upgrade to node-fetch v3 due to it being ESM-only whilst this project is CJS.

I think we'd be able to async import it. Regardless, that's the harder of the options at this point

ckohen:
At what point would you expect the apiRequest event to be fired? I'm going to put it right before the request for now, but there could be an argument for emitting before waiting out ratelimits.

typings/index.d.ts
ckohen: I mostly think it would be okay to call out the methods directly here without the string type since you'd have to be using a private property / method already so you'd already be ignoring ts.

src/rest/APIRequest.js
iCrawl: I think this should be fine honestly.

iCrawl:
This needs a rebase.

PR #6740, opened 2021-10-02
ci(Test): add ci caching

Milo123459:

This adds caching to the CI, which will decrease the amount of time it takes for the CI to execute.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -19,6 +19,13 @@ jobs:
         with:
           node-version: 16
 
+      - uses: actions/cache@v2
+        with:
+          path: ~/.npm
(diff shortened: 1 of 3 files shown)


ImRodry:
Not entirely sure how this works, but shouldn't the cache action come after the `npm ci` command?

Milo123459:
> Not entirely sure how this works, but shouldn't the cache action come after the `npm ci` command?

No, it shouldn't. Also, now that I think about it: It really should be `npm i`, not `npm ci`. CI removes the node_modules making the caching essentially useless.

iCrawl:
Yeah this is intended and will stay like this.

npm i has the liberty of updating packages in the semver minor and patch range if you use a `^` (carrot).

Also the benefit of a cache here is very minor to not needed for us 👍

PR #6741, opened 2021-10-02
feat(Interaction): add memberPermissions

ObscuritySRL:
Implements `<Interaction>.member.permissions`, if exists, sent with an interaction. `null` if the interaction was sent via DM.

Attempting to fetch an _uncached_ channel without `VIEW_CHANNEL` will result in a `Missing Access` error. However, interactions can still be received and responded to in channels which a bot does not have `VIEW_CHANNEL` for. Unless we use `<Interaction>.member.permissions` [sent with the interaction](https://canary.discord.com/developers/docs/interactions/slash-commands#receiving-an-interaction) , it's impossible to get the permissions of a member in an uncached channel.

Large bots which cannot afford to cache tens of millions of channels, roles, and permissions, could rely on the permissions sent with the interaction.

### DIFF
--- a/src/structures/Interaction.js
+++ b/src/structures/Interaction.js
@@ -73,7 +73,7 @@ class Interaction extends Base {
      * The permissions of the member, if one exists, in the channel this interaction was executed in
      * @type {?Readonly<Permissions>}
      */
-    this.memberPermissions = data.member?.permissions ? new Permissions(data.member.permissions) : null;
+    this.memberPermissions = data.member?.permissions ? new Permissions(data.member.permissions).freeze() : null;
   }
 
   /**
--- a/src/structures/Interaction.js
+++ b/src/structures/Interaction.js
@@ -70,7 +70,7 @@ class Interaction extends Base {
     this.version = data.version;
 
     /**
-     * This member permissions, if exists, in the channel this interaction was sent in
+     * The permissions of the member, if one exists, in the channel this interaction was executed in
      * @type {?Permissions}
      */
     this.memberPermissions = data.member?.permissions != null ? new Permissions(data.member.permissions) : null;
--- a/src/structures/Interaction.js
+++ b/src/structures/Interaction.js
@@ -3,6 +3,7 @@
 const Base = require('./Base');
 const { InteractionTypes, MessageComponentTypes } = require('../util/Constants');
 const SnowflakeUtil = require('../util/SnowflakeUtil');
+const Permissions = require('../util/Permissions');
 
 /**
  * Represents an interaction.
@@ -67,6 +68,12 @@ class Interaction extends Base {
      * @type {number}
      */
     this.version = data.version;
+
+    /**
+     * This member permissions, if exists, in the channel this interaction was sent in
+     * @type {?Permissions}
+     */
+     this.memberPermissions = data.member?.permissions ? new Permissions( data.member.permissions ) : null;
   }
 
   /**
--- a/src/structures/Interaction.js
+++ b/src/structures/Interaction.js
@@ -71,7 +71,7 @@ class Interaction extends Base {
 
     /**
      * The permissions of the member, if one exists, in the channel this interaction was executed in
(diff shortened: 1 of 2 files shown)


src/structures/Interaction.js
```diff
@@ -67,6 +68,12 @@ class Interaction extends Base {
      * @type {number}
      */
     this.version = data.version;
+
+    /**
+     * This member permissions, if exists, in the channel this interaction was sent in
+     * @type {?Permissions}
+     */
+     this.memberPermissions = data.member?.permissions ? new Permissions( data.member.permissions ) : null;
```
ImRodry, line 76: Prettier please, also update typings

ImRodry: Actually, I believe data.member.permissions can be 0, so this should be a typeof or an in check instead

ObscuritySRL: Good catch. The possible 0 permission was also accounted for. 👍

src/structures/Interaction.js
```diff
@@ -73,7 +73,7 @@ class Interaction extends Base {
      * This member permissions, if exists, in the channel this interaction was sent in
      * @type {?Permissions}
      */
-     this.memberPermissions = data.member?.permissions ? new Permissions( data.member.permissions ) : null;
+    this.memberPermissions = data.member?.permissions != null ? new Permissions(data.member.permissions) : null;
```
ImRodry, line 76: Once again, please install prettier and ESLint. Also I deleted the comment because I realized that permissions is sent as a string and not an integer, which this added check wouldn't even fix, so you can keep it to `data.member?.permissions`

src/structures/Interaction.js
```diff
@@ -67,6 +68,12 @@ class Interaction extends Base {
      * @type {number}
      */
     this.version = data.version;
+
+    /**
+     * This member permissions, if exists, in the channel this interaction was sent in
```
ImRodry, line 73: Current description can be improved on gramatically
```suggestion
     * The permissions of the member, if one exists, in the channel this interaction was executed in```

**Review by ImRodry — CHANGES_REQUESTED** [review]

File: src/structures/Interaction.js (hunk 2 above)
Comment (ObscuritySRL) [reply src/structures/Interaction.js]: Prettier unfortunately stopped functioning for me a few weeks ago — resolved the lint order issue.

File: src/structures/Interaction.js [hunk 4]
```diff
@@ -67,6 +68,12 @@ class Interaction extends Base {
      * @type {number}
      */
     this.version = data.version;
+
+    /**
+     * The permissions of the member, if one exists, in the channel this interaction was executed in
+     * @type {?Permissions}
```
Jiralite, line 74:
```suggestion
     * @type {?Readonly<Permissions>}
```

typings/index.d.ts
```diff
@@ -1032,6 +1032,7 @@ export class Interaction extends Base {
   public type: InteractionType;
   public user: User;
   public version: number;
+  public memberPermissions: Permissions | null;
```
Jiralite, line 1035:
```suggestion
  public memberPermissions: Readonly<Permissions> | null;
```

src/structures/Interaction.js
```diff
@@ -67,6 +68,12 @@ class Interaction extends Base {
...
+     */
+    this.memberPermissions = data.member?.permissions ? new Permissions(data.member.permissions) : null;
```
Jiralite, line 76:
```suggestion
    this.memberPermissions = data.member?.permissions ? new Permissions(data.member.permissions).freeze() : null;
```

Jiralite (commented):
I suppose these should be immutable!

SpaceEEC (approved):

vladfrangu (approved):

ImRodry (approved):
Title should say `feat(Interaction): add memberPermissions` or something similar, otherwise LGTM

iCrawl (approved):

PR #6742, opened 2021-10-02
fix(Invite): #guild never resolving into a Guild

ImRodry:
This PR fixed a bug with Invite#guild where it was documented as possibly being InviteGuild or Guild but it could never be the latter since `data.guild` can never be an instanceof Guild, since that's the raw data received from Discord. This shouldn't be breaking as the docs and types already had this behavior predicted, it was just poorly implemented

### DIFF
--- a/src/structures/Invite.js
+++ b/src/structures/Invite.js
@@ -26,7 +26,7 @@ class Invite extends Base {
      */
     this.guild = null;
     if (data.guild) {
-      this.guild = this.client.guilds.resolve(data.guild) ?? new InviteGuild(this.client, data.guild);
+      this.guild = this.client.guilds.resolve(data.guild.id) ?? new InviteGuild(this.client, data.guild);
     }
 
     /**
--- a/src/structures/Invite.js
+++ b/src/structures/Invite.js
@@ -20,14 +20,13 @@ class Invite extends Base {
 
   _patch(data) {
     const InviteGuild = require('./InviteGuild');
-    const Guild = require('./Guild');
     /**
      * The guild the invite is for including welcome screen data if present
      * @type {?(Guild|InviteGuild)}
      */
     this.guild = null;
     if (data.guild) {
-      this.guild = data.guild instanceof Guild ? data.guild : new InviteGuild(this.client, data.guild);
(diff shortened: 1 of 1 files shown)


src/structures/Invite.js
```diff
@@ -20,14 +20,13 @@ class Invite extends Base {
...
     const InviteGuild = require('./InviteGuild');
-    const Guild = require('./Guild');
     /**
      * The guild the invite is for including welcome screen data if present
      * @type {?(Guild|InviteGuild)}
      */
     this.guild = null;
     if (data.guild) {
-      this.guild = data.guild instanceof Guild ? data.guild : new InviteGuild(this.client, data.guild);
+      this.guild = this.client.guilds.resolve(data.guild) ?? new InviteGuild(this.client, data.guild);
```
SpaceEEC, line 29: Since `GuildManager#resolve` doesn't accept an api guild object:
```suggestion
      this.guild = this.client.guilds.resolve(data.guild.id) ?? new InviteGuild(this.client, data.guild);
```

SpaceEEC (changes requested):

src/structures/Invite.js
ImRodry: It does and I tested it 🤔

ImRodry: I thought I tested this but I was executing the wrong method, I am the entire circus rn

SpaceEEC: `GuildManager#resolve` only checks for specific class instances:
https://github.com/discordjs/discord.js/blob/d70bdf2ec4c8dbf1f840ec9e4cfde206f4d61745/src/managers/GuildManager.js#L107-L118

`DataManager#resolve` then checks for whether the object is an instance of `Guild` or whether it's a string being a key of a cached value:
https://github.com/discordjs/discord.js/blob/d70bdf2ec4c8dbf1f840ec9e4cfde206f4d61745/src/managers/DataManager.js#L39-L43

I don't see how it would have worked in your test, it doesn't work on my end.

ImRodry: Yeah sorry for the confusion, my bad

SpaceEEC (approved):

iCrawl (approved):

PR #6746, opened 2021-10-03
types: add ExcludeEnum utility for easier typings

ImRodry:
This PR adds a new type utility called `ExcludeEnum` to make it easier to exclude properties from an enum in scenarios where we want both the keys and the values to be excluded, to avoid creating huge Exclude types. This shouldn't change any types and just makes it easier to add more of these types in the future. All new types were tested and they work as they did before (which is still not optimal due to a TypeScript bug that doesn't let us exclude numbers).
While working on this PR I also noticed the MessageTypes enum was missing and was instead only declared as a string type union, so I added that and changed the MessageType type to be `keyof MessageTypes`. Although that change isn't strictly related to the new type utility, I felt like it would make sense to keep it in this PR, but I can revert and put it in another one if you wish.


typings/index.d.ts
```diff
@@ -4499,30 +4484,7 @@ export type MessageTarget =
...
-  | 'GUILD_DISCOVERY_REQUALIFIED'
-  | 'GUILD_DISCOVERY_GRACE_PERIOD_INITIAL_WARNING'
-  | 'GUILD_DISCOVERY_GRACE_PERIOD_FINAL_WARNING'
-  | 'THREAD_CREATED'
-  | 'REPLY'
-  | 'APPLICATION_COMMAND'
-  | 'THREAD_STARTER_MESSAGE'
-  | 'GUILD_INVITE_REMINDER'
-  | 'CONTEXT_MENU_COMMAND';
+export type MessageType = keyof MessageTypes;
```
vaporox, line 4487:
```suggestion
export type MessageType = keyof typeof MessageTypes;
```

ImRodry: oh bruh

vladfrangu (approved):
If it works, LGTM

SpaceEEC (approved):

kyranet (approved):

iCrawl (approved):

PR #6748, opened 2021-10-03
types(GuildBanManager): remove can return null

ImRodry:
This PR fixes the types and docs for GuildBanManager#remove which can return null


SpaceEEC (approved):

vladfrangu (approved):
Honestly I'd rather we returned the id instead of null but LGTM

ImRodry:
Same as that would be consistent with #create, but that's something for v14 if we do that

kyranet (approved):

iCrawl (approved):

PR #6753, opened 2021-10-03
docs: fetchReply info + example

almostSouji:

Users often get confused when trying to get a message from `CommandInteraction#reply`, as per default the returned Promise resolved with a void value. 

This PR clarifies this with an infobox as well as example code. To not overload the first example, the embed usage moved into the second example, together with the `ephemeral` option, providing an alternative non-empty message to `content`. 

### DIFF
--- a/src/structures/interfaces/InteractionResponses.js
+++ b/src/structures/interfaces/InteractionResponses.js
@@ -69,18 +69,19 @@ class InteractionResponses {
 
   /**
    * Creates a reply to this interaction.
+   * <info>By default `void`, provide the `fetchReply` option to fetch the bot's response.</info>
    * @param {string|MessagePayload|InteractionReplyOptions} options The options for the reply
    * @returns {Promise<Message|APIMessage|void>}
    * @example
-   * // Reply to the interaction with an embed
+   * // Reply to the interaction and fetch the response
    * const embed = new MessageEmbed().setDescription('Pong!');
    *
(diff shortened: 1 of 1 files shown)


iCrawl (approved):

SpaceEEC (approved):

src/structures/interfaces/InteractionResponses.js
```diff
@@ -69,18 +69,19 @@ class InteractionResponses {
 
   /**
    * Creates a reply to this interaction.
+   * <info>By default `void`, provide the `fetchReply` option to fetch the bot's response.</info>
```
Milo123459, line 72:
```suggest
   * <info>By default, the return type is `void`. Provide the `fetchReply` option to fetch the bot's response with the return type being `Message`</info>
```

vladfrangu: ^

almostSouji: don't remove the period and we can talk about it

almostSouji: On second thought, listing all the response types doesn't add any benefit to the info block. that's what the function's return type is meant to provide. Listing `void` and `Message` would be wrong, as it does return `Promise<*>` for either case. (+ the possibility to return an `APIMessage` instead. Covering all this takes up a lot of space and probably results in less people reading the info.

PR #6754, opened 2021-10-04
fix(AllowedImageSizes): add new image sizes

ObscuritySRL:
The user banner update added two new image sizes — 300 and 600. These new sizes work for all images, including guild icons, user avatars, and user banners.

### DIFF
--- a/src/util/Constants.js
+++ b/src/util/Constants.js
@@ -16,7 +16,7 @@ exports.WSCodes = {
 
 const AllowedImageFormats = ['webp', 'png', 'jpg', 'jpeg', 'gif'];
 
-const AllowedImageSizes = [...Array.from({ length: 9 }, (e, i) => 2 ** (i + 4)), 300, 600].sort((a, b) => a - b);
+const AllowedImageSizes = [16, 32, 64, 128, 256, 300, 512, 600, 1024, 2048, 4096];
 
 function makeImageUrl(root, { format = 'webp', size } = {}) {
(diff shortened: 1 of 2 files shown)


src/util/Constants.js
```diff
@@ -16,7 +16,7 @@ exports.WSCodes = {
 
 const AllowedImageFormats = ['webp', 'png', 'jpg', 'jpeg', 'gif'];
 
-const AllowedImageSizes = Array.from({ length: 9 }, (e, i) => 2 ** (i + 4));
+const AllowedImageSizes = [...Array.from({ length: 9 }, (e, i) => 2 ** (i + 4)), 300, 600].sort((a, b) => a - b);
```
kyranet, line 19: Since the code is already complicated and hard to read, wouldn't it be worth to simply specify the numbers?

ImRodry: This is internal code for private methods tho and it works perfectly. What I don’t think is needed is the sort function since we don’t send this array anywhere

ObscuritySRL: > 
> 
> Since the code is already complicated and hard to read, wouldn't it be worth to simply specify the numbers?

I would have rather just supplied the actual numbers as well — especially if Discord is going to continue adding sizes that don't fit the power of two rule.

On the other hand, as mentioned above this is only used internally.

iCrawl (approved):

vladfrangu (approved):

SpaceEEC (approved):

PR #6757, opened 2021-10-04
refactor: use optional chaining

almeidx:
There were still a few places where optional chaining could be used but wasn't

### DIFF
--- a/src/rest/RequestHandler.js
+++ b/src/rest/RequestHandler.js
@@ -177,7 +177,7 @@ class RequestHandler {
     }
 
     let sublimitTimeout;
-    if (res && res.headers) {
+    if (res?.headers) {
       const serverDate = res.headers.get('date');
       const limit = res.headers.get('x-ratelimit-limit');
       const remaining = res.headers.get('x-ratelimit-remaining');
--- a/src/rest/RequestHandler.js
+++ b/src/rest/RequestHandler.js
@@ -177,7 +177,7 @@ class RequestHandler {
     }
 
     let sublimitTimeout;
-    if (res?.headers) {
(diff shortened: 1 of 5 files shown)


kyranet (approved):

src/rest/RequestHandler.js
```diff
@@ -177,7 +177,7 @@ class RequestHandler {
     }
 
     let sublimitTimeout;
-    if (res && res.headers) {
+    if (res?.headers) {
```
vladfrangu, line 180: Here's a genuine question, when will res not be defined 👀

vladfrangu (approved):

src/rest/RequestHandler.js
ImRodry: In #6758 maybe?

vladfrangu: It's not even related and we clearly get a res because the DiscordAPIError emitted has the HTTP status code

ImRodry: I’m not sure where that status code is coming from though because cloudflare’s error is 1020 and you get no json object back at all so it sort of seems like it’s made up somewhere along the way

almeidx: I don't see how `res` could be anything other than a `Response` at this point, I'll remove the optional chain

Typings:
```ts
declare function fetch(
    url: RequestInfo,
    init?: RequestInit
): Promise<Response>;
```

ckohen: yeah, the next few if statements all use res. so if it was not defined we'd have gotten errors long ago

iCrawl (approved):

PR #6767, opened 2021-10-05
fix: correct the error checking bounds

hannesfant:
Prevents 0xffffff from being classified as an invalid value. Probably just an old typo

### DIFF
--- a/src/util/Util.js
+++ b/src/util/Util.js
@@ -469,7 +469,7 @@ class Util extends null {
       color = (color[0] << 16) + (color[1] << 8) + color[2];
     }
 
-    if (color < 0 || color > 0xffffff) throw new RangeError('COLOR_RANGE');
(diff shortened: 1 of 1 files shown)


src/util/Util.js
```diff
@@ -469,7 +469,7 @@ class Util extends null {
       color = (color[0] << 16) + (color[1] << 8) + color[2];
     }
 
-    if (color < 0 || color > 0xffffff) throw new RangeError('COLOR_RANGE');
+    if (color < 0 || color >= 0xffffff) throw new RangeError('COLOR_RANGE');
```
almeidx, line 472: The current code throws if `color` is greater than `0xffffff`. You made it throw if it's greater than or equal to `0xffffff`

hannesfant: That's totally my bad, no idea how I didn't notice that

PR #6768, opened 2021-10-05
GuildMemberRoleManager: add role that sets the role icon/emoji

Markus-Rost:

Adding `icon` property to GuildMemberRoleManager for easy access to the role used to display the role icon for the member. This matches existing properties like hoist, highest and color.

### DIFF
--- a/src/managers/GuildMemberRoleManager.js
+++ b/src/managers/GuildMemberRoleManager.js
@@ -48,6 +48,17 @@ class GuildMemberRoleManager extends DataManager {
   }
 
   /**
+   * The role of the member used to set their role icon
+   * @type {?Role}
+   * @readonly
+   */
+  get icon() {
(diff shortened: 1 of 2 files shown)


src/managers/GuildMemberRoleManager.js
```diff
@@ -47,6 +47,17 @@ class GuildMemberRoleManager extends DataManager {
...
 
+  /**
+   * The role of the member used to set their role icon
+   * @type {?Role}
+   * @readonly
+   */
+  get icon() {
+    const iconRoles = this.cache.filter(role => role.icon || role.unicodeEmoji);
+    if (!iconRoles.size) return null;
+    return iconRoles.reduce((prev, role) => (!prev || role.comparePositionTo(prev) > 0 ? role : prev));
```
almeidx, line 58: This code is duplicated a few times in this file, might be worth making a function for it instead

Issue #6595, opened 2021-09-05
Bug: guild.name is undefined


PuneetGopinath:
### Issue description

Guild#name is undefined and Guild#id is always `880150602204713060` no matter what the guild Is. I got the guild object from the guildDelete event
Steps to reproduce:
- Copy the sample code
- Kick the bot from a guild

### Codesample

```typescript
const { Client, Intents } = require("discord.js");
const client = new Client({ intents: [Intents.FLAGS.GUILDS] });
client.on("guildDelete", (guild) => {
    console.log("guildDelete event triggered", guild.name, guild.id);
});
```

### discord.js version

13.2.0-dev.1630800304.08edc0b

### Node.js version

v16.8.0 

### Operating system

Linux

### Priority this issue should have

High (immediate attention needed)

### Which partials do you have configured?

CHANNEL

### Which gateway intents are you subscribing to?

GUILDS, GUILD_MEMBERS, GUILD_VOICE_STATES, GUILD_PRESENCES, GUILD_MESSAGES, GUILD_MESSAGE_REACTIONS, DIRECT_MESSAGES, DIRECT_MESSAGE_REACTIONS

SpaceEEC:
I'm unable to reproduce this, I am seeing the expected name and ID just fine.

PuneetGopinath:
> I'm unable to reproduce this, I am seeing the expected name and ID just fine.

Sometimes it is working for me too. Sometimes it doesn't work 
![Screenshot_20210905-114827_Discord-Canary~2](https://user-images.githubusercontent.com/76863199/132117540-a526fc45-3a09-4008-a8ed-23ef4c46186c.jpg)

PuneetGopinath:
Idk what to do. I want it to work always

Palm2570Playz:
Try fetching the guild with `force` option true

Palm2570Playz:
Try fetching the guild with `force` option `true` to use force fetch and receive data directly from Discord API, don't know if this will help though.
 
```js
client.on("guildDelete", async (g) => {
   const guild = await client.guilds.fetch({ guild: g.id, force: true })
   // Code
})```

kyranet:
The only way for a guild to have empty data, is to have an unavailable guild.

Please check [`Guild#available`](https://discord.js.org/#/docs/main/stable/class/Guild?scrollTo=available).

CodeSpent:
[...]

PR #6769, opened 2021-10-05
fix(Guild): guild.available is never set on new joins

cycloptux:

This fixes a bug that was introduced in https://github.com/discordjs/discord.js/pull/6694 where, if the `data` passed to the constructor of a Guild doesn't explicitly have the `unavailable` field, then the corresponding `available` field of the cached Guild will never be set, resulting in `this.available` returning `undefined` (false-y) despite the guild being available.

This scenario specifically happens when inviting a bot to a new guild since the `unavailable` field is missing, whereas the `GUILD_CREATE` event the client receives on startup has the field and is processed correctly.

### DIFF
--- a/src/structures/Guild.js
+++ b/src/structures/Guild.js
@@ -137,7 +137,7 @@ class Guild extends AnonymousGuild {
     this.id = data.id;
     if ('name' in data) this.name = data.name;
     if ('icon' in data) this.icon = data.icon;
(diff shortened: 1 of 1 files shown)


PuneetGopinath:
I guess this bug caused #6595

almeidx:
> I guess this bug caused #6595

That issue was made long before the pull request that introduced this bug was

PR #6782, opened 2021-10-07
feat(GuildMemberUpdate): emit on user updates

Bas950:
This PR changes the guildMemberUpdate event in a (hopefully) `semver: minor` way in order to emit it on user updates like Discord intends us to. The `USER_UPDATE` event will still be emitted on both the `GUILD_MEMBER_UPDATE` and the `PRESENCE_UPDATE` gateway events, however we will not fire presenceUpdate events as presences do not exist on users like it was discussed [here](https://discord.com/channels/222078108977594368/682166281826598932/895689920667062323). There is only 1 possible issue with this which is that the events will fire when we don't have a user's flags cached and they have flags in reality. This is due to the fact that Discord does not send the `public_flags` property on startup and thus it doesn't get cached. The only alternative would be not to emit the event when the old flags are not present, however this would become an issue in scenarios where the first userUpdate received on a session is a flags update.

### DIFF
--- a/packages/discord.js/src/client/actions/PresenceUpdate.js
+++ b/packages/discord.js/src/client/actions/PresenceUpdate.js
@@ -16,8 +16,8 @@ class PresenceUpdateAction extends Action {
     const guild = this.client.guilds.cache.get(data.guild_id);
     if (!guild) return;
 
-    const oldPresence = guild.presences.cache.get(user.id)?._clone() ?? null;
-    let member = guild.members.cache.get(user.id);
+    const oldPresence = guild.presences.cache.get(data.user.id)?._clone() ?? null;
+    let member = guild.members.cache.get(data.user.id);
     if (!member && data.status !== 'offline') {
       member = guild.members._add({
         user,
--- a/packages/discord.js/src/client/Client.js
+++ b/packages/discord.js/src/client/Client.js
@@ -506,6 +506,9 @@ class Client extends BaseClient {
     if (typeof options.failIfNotExists !== 'boolean') {
       throw new TypeError('CLIENT_INVALID_OPTION', 'failIfNotExists', 'a boolean');
     }
+    if (typeof options.emitUserUpdatesPerGuild !== 'boolean') {
+      throw new TypeError('CLIENT_INVALID_OPTION', 'emitUserUpdatesPerGuild', 'a boolean');
+    }
   }
 }
 
--- a/packages/discord.js/src/client/actions/GuildMemberUpdate.js
+++ b/packages/discord.js/src/client/actions/GuildMemberUpdate.js
@@ -29,7 +29,7 @@ class GuildMemberUpdateAction extends Action {
          * @param {GuildMember} newMember The member after the update
          */
         if (shard.status === Status.READY && (!member.equals(old) || !old.user._equals(data.user))) {
-          client.emit(Events.GUILD_MEMBER_UPDATE, old, member);
+          client.emit(Events.GuildMemberUpdate, old, member);
         }
       } else {
         const newMember = guild.members._add(data);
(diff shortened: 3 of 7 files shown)


Bas950:
@SpaceEEC Could you please give some details as to why you downvoted?

Bas950:
I agree with y'alls concerns for memory usage going up, hence why I think it would be a good idea to make this a ClientOption.
image
image
I have currently named it `emitUserUpdatesPerGuild` (uncommitted as [per request of Crawl](https://discord.com/channels/222078108977594368/682166281826598932/904444673261727755)), if you guys like this idea then I will commit it and add it to this PR.
Let me know if you guys have a better naming for the added option, then I can also improve that :)

typings/index.d.ts
```diff
@@ -18,8 +18,8 @@ import {
   TimestampStylesString,
   underscore,
   userMention,
-} from '@discordjs/builders';
-import { Collection } from '@discordjs/collection';
+} from "@discordjs/builders";
```
imranbarbhuiya, line 21: Why double quotes are used here?

typings/index.d.ts
```diff
@@ -48,14 +47,15 @@ import {
...
-import { MessagePort, Worker } from 'node:worker_threads';
-import * as WebSocket from 'ws';
+} from "discord-api-types/v9";
+import { Response } from "node-fetch";
+import { ChildProcess } from "node:child_process";
+import { EventEmitter } from "node:events";
+import { AgentOptions } from "node:https";
+import { Stream } from "node:stream";
+import { MessagePort, Worker } from "node:worker_threads";
+import * as WebSocket from "ws";
```
imranbarbhuiya, line 57: same

typings/index.d.ts
```diff
@@ -137,7 +137,7 @@ import {
   RawWelcomeScreenData,
   RawWidgetData,
   RawWidgetMemberData,
-} from './rawDataTypes';
+} from "./rawDataTypes";
```
imranbarbhuiya, line 140: ^^

imranbarbhuiya (changes requested):

Bas950:
@imranbarbhuiya thanks for notifying, fixed the issues, also rebased the pr to fix all merge conflicts.

src/structures/GuildMember.js
```diff
@@ -62,7 +62,7 @@ class GuildMember extends Base {
        * The user that this guild member instance represents
        * @type {?User}
        */
-      this.user = this.client.users._add(data.user, true);
+      this.user = this.client.users._add(data.user, true)._clone();
```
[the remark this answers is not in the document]
Bas950: @vladfrangu I added a setting for it, hopefully you could rereview your standpoint as it's better as a setting since most peeps indeed don't need this.

packages/discord.js/src/client/actions/PresenceUpdate.js
```diff
@@ -6,18 +6,18 @@ const { Events } = require('../../util/Constants');
 class PresenceUpdateAction extends Action {
   handle(data) {
     let user = this.client.users.cache.get(data.user.id);
-    if (!user && data.user?.username) user = this.client.users._add(data.user);
+    if (!user && data.user.username) user = this.client.users._add(data.user);
     if (!user) return;
 
-    if (data.user?.username) {
+    if (data.user.username) {
```
imranbarbhuiya, line 12: #7248 fixed it. So ig you can leave it as it is.

Bas950: Rebased pr to fix it :)

ImRodry:
This needs a rebase

thebigwolfy (approved):

iCrawl:
We decided to not move forward with this and rather have a different take on this.

PR #6797, opened 2021-10-08
fix(CommandInteractionOption): don't set undefined props

ImRodry:
This PR fixes certain props in ApplicationCommand#options being set to undefined where they weren't documented/typed as such. This means that no property should now be set to undefined and will only be present if the user has passed it when creating the command.
Also changed the typings for the ApplicationCommand#transformOption method to include overloads and API types.

### DIFF
--- a/src/structures/ApplicationCommand.js
+++ b/src/structures/ApplicationCommand.js
@@ -327,10 +327,11 @@ class ApplicationCommand extends Base {
    */
 
   /**
-   * Transforms an {@link ApplicationCommandOptionData} object into something that can be used with the API.
+   * Transforms an {@link ApplicationCommandOptionData} object into something that can be used by the API or vice-versa.
    * @param {ApplicationCommandOptionData} option The option to transform
    * @param {boolean} [received] Whether this option has been received from Discord
-   * @returns {APIApplicationCommandOption}
+   * @returns {APIApplicationCommandOption|ApplicationCommandOption} API-compatible option if resolved is true,
+   * otherwise an ApplicationCommandOption
    * @private
    */
   static transformOption(option, received) {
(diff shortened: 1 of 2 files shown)


kyranet:
What's the issue with having `undefined`s? It's intentional that we do this because mutating an object's properties affects the over-all performance (due to object shapes shifting) of the method and everything that interacts with its properties, and keys which value is `undefined` aren't JSON-serialized.

Also, this kinda duplicates #5620.

> This PR fixes certain props in ApplicationCommand#options being set to undefined where they weren't documented/typed as such.

To fix this, simply change the types to match.

ImRodry:
> That makes no sense btw, undefined values when stringified simply "disappear", they don't have a value, they are not "undefined" they do not exist.

That still doesn't change the fact that this behavior isn't documented anywhere and isn't expected to some extent

ImRodry:
> Or you just document them as being undefined and reduce all the overhead it procudes.

I can document them as undefined but I do not agree with properties that cannot be set being undefined, so something still needs to be changed about that. The only reason I removed **all** undefined props was because of what vlad said on Discord

ImRodry:
> All your if checks? That's overhead. Setting them to undefined really doesn't matter since, as I've actually told you before on Discord, once you JSON.stringify they are gone

If your concern is performance that shouldn't really be an issue since this function isn't called that often (only when commands are fetched). I'm also not talking about using JSON.stringify here, I'm talking about what the actual object holds. Having these set to undefined also doesn't allow us to do `if ('channelTypes' in option)` for example, which we should be able to do.

iCrawl:
For 13.x we decide against doing that, may be worth revisiting in 14.

PR #6801, opened 2021-10-09
feat(Interaction): add `isApplicationCommand` check

manuelvleeuwen:
Introduces `Interaction#isApplicationCommand`, which indicates whether or not the interaction is a `BaseCommandInteraction` (as of now the only subclasses are `CommandInteraction` or `ContextMenuInteraction`, but there might come more in the future).

### DIFF
--- a/src/structures/Interaction.js
+++ b/src/structures/Interaction.js
@@ -137,7 +137,7 @@ class Interaction extends Base {
   }
 
   /**
-   * Indicates whether this interaction is n {@link BaseCommandInteraction}.
+   * Indicates whether this interaction is a {@link BaseCommandInteraction}.
    * @returns {boolean}
    */
   isApplicationCommand() {


iCrawl:
This is a semver-major and seems too big of a change, considering we don't even know when we'll release the next major and what has changed on Discords side to maybe justify this.

thehackerboi69github:
Yeah, I agree that renaming might be too big of a change, but still a check like `Interaction#isApplicationCommand` would be useful, for example: I defer all application command interactions by default (I'm making API calls that delay for quite long) and I can run them without checking whether they are slash commands or context menus.
```js
interaction.isCommand() || interaction.isContextMenu()
```
also works but imo a single check for this is cleaner.

ImRodry:
You can add the isApplicationCommand method just fine without needing to rename the class

PR #6808, opened 2021-10-10
fix(Types): export more types

favna:

Bit annoying that without this we have to copy over types to our own code so lets just export them. Personally (and some other Sapphire users) ran into `ExcludeEnum`, Souji asked me to also export the collector ones for Yuu. 

Then I figured, might as well do a check with `/^type /` to find any not yet exported types, after this PR everything is exported

### DIFF
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -455,7 +455,7 @@ export abstract class Channel extends Base {
   public toString(): ChannelMention;
 }
 
-type If<T extends boolean, A, B = null> = T extends true ? A : T extends false ? B : A | B;
+export type If<T extends boolean, A, B = null> = T extends true ? A : T extends false ? B : A | B;
 
 export class Client<Ready extends boolean = boolean> extends BaseClient {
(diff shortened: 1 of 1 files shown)


ImRodry:
@Favna just curious, in what scenarios do you need to use these types yourself?

Favna:
> For your example you can easily do `as ActivitiesOptions["type"]` and I'm not sure what the issue is on yuu but the same probably applies as well

Doesn't matter. That's not an argument to not export the types. Besides, hate to be the one to give you a reality check, but you're not a maintainer so it's up to people like Souji, Crawl, Kyra and co to make the call whether they want to merge this and considering Souji specifically asked me to make the collector changes... well you do the math.

ImRodry:
I know, I'm just voicing my opinion as I don't think that a discord API library should be exporting generic type utilities because it seems quite out of the scope of the package. There are other alternatives that work just as fine like the one I just mentioned but I know it's up to them in the end to say whether to merge this or not, I just think it's a bad idea

Issue #6539, opened 2021-08-27
Roadmap: cache and sweeping improvements


kyranet:
Starting with Discord.js v13.0.0, we removed a lot of add-ons, such as `Channel#_typing` (https://github.com/discordjs/discord.js/pull/6114), `User#lastMessageId`, `User#lastMessageChannelId` (https://github.com/discordjs/discord.js/pull/6046), and so on. Those changes have helped reduce Discord.js' memory footprint greatly.

With Discord adding archivable thread channels, we saw a need to add a second sweeper, alongside message's. In an attempt to generalize sweeping caches, we added custom caching (https://github.com/discordjs/discord.js/pull/6013) and later sweeping caches (https://github.com/discordjs/discord.js/pull/6110). The latter (`SweptCollection`, now `LimitedCollection`) is the new alternative for sweeping, and noticed it's biggest improvement was to allow us to use them in even very nested collections (such as `client` -> `channel` -> `message` -> `reactions`), but as much as it helped, it has a bounce effect and does actually increase memory usage when used.

A `LimitedCollection` differs from `Collection` by several properties:

- `maxSize`, which is usually a small integer.
- `keepOverLimit`, which is a function, often a reference.
- `sweepFilter`, which is also a function, often a reference.
- `interval`, which is either an interval (with a lot of tracking properties) or null.

If you were to use `LimitedCollection` on a very duplicated manager, such as `ReactionManager`, which also stays empty most of the time, this would use far more memory than it should, despite the collection's very purpose being to reduce it.

Additionally, if a `LimitedCollection` is used, a lot of extra memory would be used for the [`FinalizationRegistry`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/FinalizationRegistry), which is used to hook into the runtime's garbage collector and clear the timers in said collections, so that they don't leak memory. However, this brings a serious issue: the cleanup is bound in every single instance of them, that's a full function (which has large memory overhead) being created, as well as it being stored in both a `Set` and finally the finalizer itself, alongside a message and a name:

https://github.com/discordjs/discord.js/blob/fc51f61f0cdc66087e918165646aee4be74e6e56/src/managers/CachedManager.js#L17-L28

As a side note, sweeping collections are cancelled a bit late, if you set the interval to 2 minutes, chances are that once its manager is not longer accessible, it'll run at least twice, as the GC usually runs every 5 or 10 minutes, and of course, we don't want to be sweeping unaccessible memory.

[...]

PR #6825, opened 2021-10-13
feat(Client): add global sweepers

ckohen:

<div align="center">

| ⚠️ This is tested, but not all combinations of sweepers have been validated. |
| - |

</div>

- This aims to tackle point 2 in  #6539: Global Sweepers. This does not completely satisfy what was outlined there however, it still relies on `LimitedCollection` to provide max size functionality. After doing the work to get to this point, that seemed out of scope, and can easily be tackled in another PR. This does mean that you are still storing extra properties on LimitedCollection for now, however hopefully not tracking a timeout.

In order to keep this semver: minor, default options have unfortunately <strong>not</strong> been switched to use the sweepers in this PR.

### DIFF
--- a/src/util/Options.js
+++ b/src/util/Options.js
@@ -37,9 +37,9 @@
  * You can use your own function, or the {@link Options} class to customize the Collection used for the cache.
  * <warn>Overriding the cache used in `GuildManager`, `ChannelManager`, `GuildChannelManager`, `RoleManager`,
  * and `PermissionOverwriteManager` is unsupported and **will** break functionality</warn>
- * @property {number} [messageCacheLifetime=0] DEPRECATED: Use `makeCache` with a `LimitedCollection` instead.
+ * @property {number} [messageCacheLifetime=0] DEPRECATED: Pass `lifetime` to `sweepers.messages` instead.
  * How long a message should stay in the cache until it is considered sweepable (in seconds, 0 for forever)
- * @property {number} [messageSweepInterval=0] DEPRECATED: Use `makeCache` with a `LimitedCollection` instead.
+ * @property {number} [messageSweepInterval=0] DEPRECATED: Pass `interval` to `sweepers.messages` instead.
  * How frequently to remove messages from the cache that are older than the message cache lifetime
  * (in seconds, 0 for never)
  * @property {MessageMentionOptions} [allowedMentions] Default value for {@link MessageOptions#allowedMentions}
--- a/src/util/Options.js
+++ b/src/util/Options.js
@@ -269,4 +269,19 @@ class Options extends null {
   }
 }
 
+/**
+ * The default settings passed to {@link Options.sweepers} (for v14).
+ * The sweepers that this changes are:
+ * * `threads` - Sweep archived threads every hour, removing those archived more than 4 hours ago
+ * <info>If you want to keep default behavior and add on top of it you can use this object and add on to it, e.g.
+ * `sweepers: { ...Options.defaultSweeperSettings, messages: { interval: 300, lifetime: 600 } })`</info>
+ * @type {SweeperOptions}
+ */
+Options.defaultSweeperSettings = {
+  threads: {
+    interval: 3600,
+    lifetime: 14400,
+  },
+};
+
 module.exports = Options;
--- a/src/util/Options.js
+++ b/src/util/Options.js
@@ -70,8 +70,25 @@
  * [User Agent](https://discord.com/developers/docs/reference#user-agent) header
  * @property {PresenceData} [presence={}] Presence data to use upon login
  * @property {IntentsResolvable} intents Intents to enable for this connection
+ * @property {SweeperOptions} [sweepers={}] Options for cache sweeping
  * @property {WebsocketOptions} [ws] Options for the WebSocket
  * @property {HTTPOptions} [http] HTTP options
+ */
+
+/**
+ * Options for {@link Sweepers} defining the behavior of cache sweeping
+ * @typedef {Object<SweeperKey, SweepOptions>} SweeperOptions
+ */
+
+/**
+ * Options for sweeping a single type of item from cache
+ * @typedef {Object} SweepOptions
+ * @property {number} interval The interval (in seconds) at which to perform sweeping of the item
+ * @property {number} [lifetime] How long an item should stay in cache until it is considered sweepable.
+ * <warn>This property is only valid for the `invites`, `messages`, and `threads` keys. The `filter` property
+ * is mutually exclusive to this property and takes priority</warn>
+ * @property {GlobalSweepFilter} filter The function used to determine the function passed to the sweep method
+ * <info>This property is optional when the key is `invites`, `messages`, or `threads` and `lifetime` is set</info>
  */
 
 /**
@@ -125,6 +142,7 @@ class Options extends null {
       failIfNotExists: true,
       userAgentSuffix: [],
       presence: {},
+      sweepers: {},
       ws: {
         large_threshold: 50,
         compress: false,
--- a/src/util/Sweepers.js
+++ b/src/util/Sweepers.js
@@ -364,10 +364,11 @@ class Sweepers {
     let items = 0;
 
     for (const guild of this.client.guilds.cache.values()) {
-      if (guild[key].cache.size === 0) continue;
+      const cache = guild[key].cache;
+      if (cache.size === 0) continue;
       guilds++;
 
-      items += guild[key].cache.sweep(filter);
+      items += cache.sweep(filter);
     }
 
     if (emit) {
--- a/src/util/Sweepers.js
+++ b/src/util/Sweepers.js
@@ -135,8 +135,9 @@ class Sweepers {
     let messages = 0;
 
     for (const channel of this.client.channels.cache.values()) {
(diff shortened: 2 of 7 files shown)


src/util/Sweepers.js
```diff
@@ -0,0 +1,450 @@
...
+
+  /**
+   * Initialize an interval for sweeping
+   * @param {string} intervalKey The name of the property that stores the interval for this sweeper
+   * @param {string} sweepKey The name of the function that sweeps the desired caches
+   * @param {Object} opts Validated options for a sweep
+   * @private
+   */
+  _initInterval(intervalKey, sweepKey, opts) {
+    if (opts.interval > 0 && opts.interval !== Infinity) {
```
kyranet, line 439: You can invert the condition to reduce indentation.

src/util/Sweepers.js
```diff
@@ -0,0 +1,450 @@
...
+      throw new TypeError('INVALID_TYPE', `sweepers.${key}`, 'object', true);
+    }
+    if (typeof props.interval !== 'number') {
+      throw new TypeError('INVALID_TYPE', `sweepers.${key}.interval`, 'number');
+    }
+    if (['invites', 'messages', 'threads'].includes(key) && !('filter' in props)) {
+      if (typeof props.lifetime !== 'number') {
+        throw new TypeError('INVALID_TYPE', `sweepers.${key}.lifetime`, 'number');
+      }
+      return;
```
kyranet, line 424: Admitedly, this was a bit difficult to understand, perhaps in the future Zod will make this easier to understand 😅.

In the meantime, we can probably put a comment here, explaining that we can define a lifetime without a filter.

src/util/Sweepers.js
```diff
@@ -0,0 +1,450 @@
...
+     * The options the sweepers were instantiated with
+     * @type {SweeperOptions}
+     */
+    this.options = options;
+
+    /**
+     * A record of interval timeout that is used to sweep the indicated items, or null if not being swept
+     * @type {Object<SweeperKey, ?Timeout>}
+     */
+    this.intervals = {};
```
kyranet, line 35: We can probably define all the possible keys here, so we can see the potential list of possible sweepers in this place. Also to avoid [shape changing](https://mathiasbynens.be/notes/shapes-ics).

kyranet (changes requested):
Just a few things, nice work overall ^^

src/util/Sweepers.js
ckohen: I can use Object.fromEntries to avoid shape changing I believe. The list already exists in Constants and is just a ctrl-click away (will be right in that call now. If you still want to manually call it out, that can be done though.

src/util/Sweepers.js
```diff
@@ -0,0 +1,449 @@
...
+   * @param {string} sweepKey The name of the function that sweeps the desired caches
+   * @param {Object} opts Validated options for a sweep
+   * @private
+   */
+  _initInterval(intervalKey, sweepKey, opts) {
+    if (opts.interval <= 0 || opts.interval === Infinity) return;
+    this.intervals[intervalKey] = setInterval(() => {
+      const sweepFn = opts.filter();
+      if (sweepFn === null) return;
+      if (typeof sweepFn !== 'function') throw new TypeError('SWEEP_FILTER_RETURN');
```
kyranet, line 443: Should those be inside the interval and not outside? It gets cloned options, so those values are never changing.

src/util/Sweepers.js
```diff
@@ -0,0 +1,449 @@
...
+  sweepVoiceStates(filter) {
+    return this._sweepGuildDirectProp('voiceStates', filter, { outputName: 'voice states' }).items;
+  }
+
+  /**
+   * Cancels all sweeping intervals
+   * @returns {void}
+   */
+  destroy() {
+    for (const key of Object.keys(this.intervals)) {
```
kyranet, line 280: I'm pretty sure this can be done too:
```suggestion
    for (const key of SweeperKeys) {
```

kyranet (dismissed):
Just the last 2 requests, everything else looks fine to me.

src/util/Sweepers.js
ckohen: opts.filter doesn't change, the return value from calling opts.filter can.

src/util/Sweepers.js
ckohen: whoopsies...

kyranet (approved):

src/util/Options.js
```diff
@@ -251,4 +269,19 @@ class Options extends null {
...
 
+/**
+ * The default settings passed to {@link Options.sweepers} (for v14).
+ * The sweepers that this changes are:
+ * * `threads` - Sweep archived threads every hour, removing those archived more than 4 hours ago
+ * <info>If you want to keep default behavior and add on top of it you can use this object and add on to it, e.g.
+ * `sweepers: { ...Options.defaultSweeperSettings, messages: { interval: 300, lifetime: 600 } })`</info>
+ * @type {SweeperOptions}
+ */
+Options.defaultSweeperSettings = {
```
vladfrangu, line 280: Shouldn't you also place these defaults in `createDefault`?

vladfrangu (changes requested):

src/util/Options.js
ckohen: No, I cannot remove the existing make cache sweeper options in semver minor and adding sweepers on top of it would cause a degree of performance degradation with no actual changes.

vladfrangu (approved):

src/util/Options.js
kyranet: You can add a TODO for later, unless you plan on making the semver major PR quickly after this is merged.

ckohen: That PR is the very next PR I am going to write, there's a lot more TODO, and this is high on the priority list in that TODO list.

iCrawl (approved):

src/util/Sweepers.js
```diff
@@ -0,0 +1,449 @@
...
+   */
+  sweepMessages(filter) {
+    if (typeof filter !== 'function') {
+      throw new TypeError('INVALID_TYPE', 'filter', 'function');
+    }
+    let channels = 0;
+    let messages = 0;
+
+    for (const channel of this.client.channels.cache.values()) {
+      if (!channel.messages || channel.messages.cache.size === 0) continue;
```
SpaceEEC, line 138: Instead of `channel.messages` you could use `channel.isText()`.

SpaceEEC: It feels a bit inconsistent to not count channels that do not have messages cached but count channels that have messages cached even if not sweeping anything in there.

src/util/Sweepers.js
```diff
@@ -0,0 +1,449 @@
...
+  sweepReactions(filter) {
+    if (typeof filter !== 'function') {
+      throw new TypeError('INVALID_TYPE', 'filter', 'function');
+    }
+    let channels = 0;
+    let messages = 0;
+    let reactions = 0;
+
+    for (const channel of this.client.channels.cache.values()) {
+      if (!channel.messages || channel.messages.cache.size === 0) continue;
```
SpaceEEC, line 170: The same about the "inconsistent" counting in regards to empty channels vs channels without sweeping anything here.

src/util/Sweepers.js
```diff
@@ -0,0 +1,449 @@
...
+      channels++;
+
+      for (const message of channel.messages.cache.values()) {
+        if (message.reactions.cache.size === 0) continue;
```
SpaceEEC, line 174: Same as above

src/util/Sweepers.js
```diff
@@ -0,0 +1,449 @@
...
+   */
+  sweepThreadMembers(filter) {
+    if (typeof filter !== 'function') {
+      throw new TypeError('INVALID_TYPE', 'filter', 'function');
+    }
+
+    let threads = 0;
+    let members = 0;
+    for (const channel of this.client.channels.cache.values()) {
+      if (!ThreadChannelTypes.includes(channel.type) || channel.members.cache.size === 0) continue;
```
SpaceEEC, line 210: Same as above

src/util/Sweepers.js
```diff
@@ -0,0 +1,449 @@
...
+    if (typeof filter !== 'function') {
+      throw new TypeError('INVALID_TYPE', 'filter', 'function');
+    }
+
+    let guilds = 0;
+    let items = 0;
+
+    for (const guild of this.client.guilds.cache.values()) {
+      const { cache } = guild[key];
+      if (cache.size === 0) continue;
```
SpaceEEC, line 393: Same as above

src/util/Sweepers.js
SpaceEEC: Instead of `channel.messages` you could use `channel.isText()`.

SpaceEEC (changes requested):
I'm not sure about the importance of the count bits, but here they are. 😅

src/util/Sweepers.js
ckohen: Yeah, it felt weird when I was looking at the output of the event, but it made sense at the same time. I think I'll end up going with this change though since it'll match the old message sweeping style.

SpaceEEC (approved):

PR #6841, opened 2021-10-15
fix(ApplicationCommandManager): update guild command cache on c/u/d

advaith1:
### Please describe the changes this PR makes and why it should be merged:
Currently, the cache is not updated when guild commands are created, updated, or deleted; this PR fixes that.

[discussion](https://canary.discord.com/channels/222078108977594368/682166281826598932/890029573008425011) (about a month ago)

### Status and versioning classification:


vladfrangu:
> This fix was made because the events no longer emit - is it worth adding deprecation warnings to these events?

If there isn't one already, yeah, probably should let users know the event won't fire

advaith1:
wouldn't that be out of scope for this pr?

PR #6872, opened 2021-10-21
Update publish-dev.yml

vijayymmeena:

image

As you can see in the above image, there were four NPM releases for the same head in past 2 days. So, I would like to propose this PR to prevent this from happening and only release an update if it is necessary. 

### DIFF
--- a/.github/workflows/publish-dev.yml
+++ b/.github/workflows/publish-dev.yml
@@ -29,6 +29,6 @@ jobs:
       - name: Publish
         run: |
           npm version --git-tag-version=false $(jq --raw-output '.version' package.json).$(date +%s).$(git rev-parse --short HEAD)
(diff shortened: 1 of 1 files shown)


ImRodry:
Please don’t remove the date. We did this so that we could update more easily and dependabot could also pick up on new releases. I would really like to see this PR go forward but not by changing the versioning scheme

oceanroleplay:
@ImRodry I wish the same, but I am not sure, how to make the if condition works with date. Perhaps someone can help about this.

vladfrangu:
Well, the last part is the short commit hash, you can probably use that when checking if the version should be published (check latest from npm with HEAD or w/e in git)

oceanroleplay:
@ImRodry this will work just perfect, what do you think?

iCrawl:
You running a dev version in production is not our or my cup of tea.

Additionally, you could just let your dependabot update your deps once a week instead of looking for new ones every day.

oceanroleplay:
@iCrawl just to confirm, you added this PR in milestone after you closed it. If that is correct, ignore.

PR #6887, opened 2021-10-25
feat(Guild): add premiumProgressbarEnabled

ImRodry:

- https://github.com/discord/discord-api-docs/pull/4001

This PR adds Discord's new feature: premium_progress_bar_enabled, which is only available on certain Guilds at the moment. Here's what it looks like:
image
And in the settings:
image
Upstream PR:
- None yet!

### DIFF
--- a/src/structures/Guild.js
+++ b/src/structures/Guild.js
@@ -875,7 +875,7 @@ class Guild extends AnonymousGuild {
       _data.description = data.description;
     }
     if (data.preferredLocale) _data.preferred_locale = data.preferredLocale;
-    if (data.premiumProgressBarEnabled) _data.premium_progress_bar_enabled = data.premiumProgressBarEnabled;
+    if ('premiumProgressBarEnabled' in data) _data.premium_progress_bar_enabled = data.premiumProgressBarEnabled;
     const newData = await this.client.api.guilds(this.id).patch({ data: _data, reason });
     return this.client.actions.GuildUpdate.handle(newData).updated;
   }
--- a/src/structures/Guild.js
+++ b/src/structures/Guild.js
@@ -167,6 +167,14 @@ class Guild extends AnonymousGuild {
       this.large = Boolean(data.large);
     }
 
+    if ('premium_progress_bar_enabled' in data) {
+      /**
+       * Whether this guild has its premium progress bar enabled (GOAL bar below the guild name on the client)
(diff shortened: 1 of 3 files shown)


vladfrangu (dismissed):

ImRodry:
This feature has now rolled out to every guild so I'm marking this as ready for review. As for the review above, I believe the description is fine as it is but I'll be glad to change it if better descriptions come up. I believe we should still mention the client name as this is something that is done in other properties, maybe we could remove the API name as that can be deduced from the property name

kyranet (approved):

ImRodry:
Upstream PR was merged! 🎉

src/structures/Guild.js
```diff
@@ -866,6 +875,7 @@ class Guild extends AnonymousGuild {
       _data.description = data.description;
     }
     if (data.preferredLocale) _data.preferred_locale = data.preferredLocale;
+    if (data.premiumProgressBarEnabled) _data.premium_progress_bar_enabled = data.premiumProgressBarEnabled;
```
vladfrangu, line 878: This won't set it to false if i call it with false...

vladfrangu (changes requested):

src/structures/Guild.js
```diff
@@ -167,6 +167,14 @@ class Guild extends AnonymousGuild {
       this.large = Boolean(data.large);
     }
 
+    if ('premium_progress_bar_enabled' in data) {
```
vladfrangu, line 170: I still think this `in` check is useless since it's always provided according to the docs.. Thoughts @discordjs/the-big-4

vladfrangu (changes requested):

src/structures/Guild.js
ImRodry: right, my bad
```suggestion
    if ('premiumProgressBarEnabled' in data) _data.premium_progress_bar_enabled = data.premiumProgressBarEnabled;
```

src/structures/Guild.js
ImRodry: I followed what was done on the lines above. There are properties that are never missing according to the docs and still have that `in` check. I think that should be done in a separate PR if anything don't you think?

vladfrangu (approved):

SpaceEEC (approved):

PR #6919, opened 2021-10-29
fix: assert channel types in message actions

SpaceEEC:

This PR ensures that message related actions take place in a channel that is supported to avoid throwing an unhandled exceptions. (Read: text in voice, also future-proofing)

Changed actions:
- MessageCreate
- MessageDelete
- MessageReactionAdd
- MessageReactionRemove
- MessageReactionRemoveAll
- MessageReactionRemoveEmoji
- MessageUpdate
- TypingStart

The reaction actions previously were ensuring that the action was not taking place in a voice channel (as per type) and the typing start action in a text channel (as per type).

I went with `Channel#isText` since it seems to be the less error-prone method to check whether messages can be accessed in said channel, since it's just `return 'messages' in this`.

### DIFF
--- a/src/client/actions/MessageCreate.js
+++ b/src/client/actions/MessageCreate.js
@@ -10,6 +10,8 @@ class MessageCreateAction extends Action {
     const client = this.client;
     const channel = this.getChannel(data);
     if (channel) {
+      if (!channel.isText()) return {};
+
       const existing = channel.messages.cache.get(data.id);
       if (existing) return { message: existing };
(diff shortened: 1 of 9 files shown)


ImRodry:
Voice channels will soon have messages though, are you sure you wanna do this?

SpaceEEC:
Yes, I am.

iCrawl (approved):

iCrawl (approved):

PR #6920, opened 2021-10-29
chore: next dev version

oceanroleplay:

prepare for next dev release

note: please take action before next 4 hour because dev workflow will be executed after that.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "discord.js",
-  "version": "13.3.0",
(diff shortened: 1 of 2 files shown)


oceanroleplay:
@iCrawl would you recommend to release minor dev version after each stable version? because if you like that, then I can update dev workflow to automate this task and never prepare for this after each stable release? if yes, then let me know, I will create another PR later for that.

iCrawl:
I'd like to keep things as they are right now. This was just an oversight.

PR #6933, opened 2021-11-01
types(Message): `guild` should be non-null when message is in guild

suneettipirneni:

Makes it so `guild` and `member` are non-null when message is in guild.


wasdennnoch:
Note that messages sent by a webhook don't have a member object. Additionally, when fetching a message from a user who is no longer a member of the guild, `member` will also be null.

typings/index.d.ts
```diff
@@ -1357,11 +1357,11 @@ export class Message<Cached extends boolean = boolean> extends Base {
   public embeds: MessageEmbed[];
   public groupActivityApplication: ClientApplication | null;
   public guildId: Snowflake | null;
-  public readonly guild: Guild | null;
+  public readonly guild: If<Cached, Guild, null>;
```
asportnoy, line 1360:
```suggestion
  public readonly guild: If<Cached, Guild>;
```
Since the default value is `null`, the parameter is not needed. This is also consistent with other uses of that type.

typings/index.d.ts
```diff
@@ -1357,11 +1357,11 @@ export class Message<Cached extends boolean = boolean> extends Base {
...
   public readonly hasThread: boolean;
   public id: Snowflake;
   public interaction: MessageInteraction | null;
-  public readonly member: GuildMember | null;
+  public readonly member: If<Cached, GuildMember, null>;
```
asportnoy, line 1364:
```suggestion
  public readonly member: If<Cached, GuildMember | null>;
```
See @wasdennnoch's comment

vladfrangu: At that point we might as well keep it as `GuildMember | null`

SpaceEEC: I think this should stay as-is, a message in a guild does not guarantee you a member:
e.g. webhook messages, old messages from someone that left the guild since

SpaceEEC (changes requested):

iCrawl:

Jiralite:
I suppose the guild id can also be non-null then?

iCrawl (approved):

Issue #6935, opened 2021-11-02
ChannelManager has wrong type parameter


isaiahtaylor:
### Issue description

The `holds` parameter (`Channel`) in the following snippet is wrong:

```ts
export class ChannelManager extends CachedManager<Snowflake, Channel, ChannelResolvable> {
  private constructor(client: Client, iterable: Iterable<RawChannelData>);
  public fetch(id: Snowflake, options?: FetchChannelOptions): Promise<Channel | null>;
}
```

I am aware of the explanations given in #3622, #4708, and #4357. However, these explanations do not seem to be aware of unions.

https://github.com/discordjs/discord.js/issues/3622#issuecomment-565526740
>And that is completely correct. Here are the docs.
Since CategoryChannel, TextChannel, VoiceChannel and other types extend GuildChannel, which itself extends Channel, there is no way there will be send for all of them. You can't send messages to voice channels nor categories.

The issue is that the Cache Manager doesn't hold / yield instances of the super class `Channel`, it holds instances of child classes. We need to interact with those children as their own unique signatures, like `send` or on the other hand voice-specific properties.

https://github.com/discordjs/discord.js/issues/3622#issuecomment-565526742
>The Channel class does not have a send method, because categories and voice channels are also a child classes of the Channel class. When you're certain you have a text channel you can cast it to a TextChannel

The statement "because categories and voice channels are also child classes of the Channel class" is not a good argument for why there shouldn't just be a correct type, like a union. If you need to specify the _ingress_ side of the cache manager to be able to support `Channel` at a minimum, you can use a Generic for that:

```ts
export type ChannelManager = MetaChannelManager<TextChannel | VoiceChannel | CategoryChannel | NewsChannel | StoreChannel | StageChannel | WhateverOtherTypes>;

export class MetaChannelManager<T extends Channel> extends CachedManager<Snowflake, T, ChannelResolvable> {
  private constructor(client: Client, iterable: Iterable<RawChannelData>);
  public fetch(id: Snowflake, options?: FetchChannelOptions): Promise<Channel | null>;
}
```

Also, why be "certain" when you can just check?

```ts
if (channel.type === "GUILD_TEXT") {
  // channel becomes a TextChannel in this block
  channel.send("cats are cool");
} else {
  Debug("Hey, you've got the wrong channel ID! This isn't a text channel!");
}
```

https://github.com/discordjs/discord.js/issues/3622#issuecomment-565547568
[...]

PR #6939, opened 2021-11-02
fix: change channel to union type

isaiahtaylor:

See #6935

### DIFF
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -444,7 +444,17 @@ export class CategoryChannel extends GuildChannel {
 
 export type CategoryChannelResolvable = Snowflake | CategoryChannel;
 
-export abstract class Channel extends Base {
+export type Channel =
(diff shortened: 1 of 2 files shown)


suneettipirneni:
Firstly, you'll need to modify all of the respective implementation javascript files for these types to be accurate. There's no such thing as a `BaseChannel` in the js code.

Secondly, you'll still need to explain the practical differences as to how this is an improvement. 

It's fine stating that this [pattern] violates [another pattern]. But if you don't provide any practical examples of how it effects the codebase moving forward then it doesn't seem to be worth the overhead of maintenance in addition to the already existing overhead when adding new channel types.

iCrawl:

isaiahtaylor:
Just to be clear I opened this PR as a starting point to go along with #6935. I don't intend to do anything further with it. But I do think it would be helpful if someone picked it up!

PR #6945, opened 2021-11-04
docs(README): Use discord.js to register slash command

waki285:

The previous README shows how to create a slash command using a package that is essentially unnecessary.
This will be changed so that it can be registered using only discord.js.

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,33 +42,32 @@ pnpm add discord.js
 
 ## Example usage
 
-Install all required dependencies:
-```sh-session
-npm install discord.js @discordjs/rest discord-api-types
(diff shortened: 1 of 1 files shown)


almostSouji:
Using API types and the REST module here was a deliberate decision. Caching behavior and websocket events (the prime use cases for the discord.js Client) are not required to push commands. We suggest people use the tool best fit for the job. For pushing command changes on demand (not on ready or similar, of course) the REST module is sufficient.

This is also consistent with our guide page on slash commands, for that same reason: <https://discordjs.guide/interactions/registering-slash-commands.html>

PR #6948, opened 2021-11-05
+ Information

voidx999:
Added some information on the README.md for users to understand what the links send you to

The changes that are in this PR is where on the README.MD it includes information on the links for people to make it easier to understand. The PR should be merged because it could help people understand the meanings of the links, and the redirect it brings you to.

### DIFF
--- a/README.md
+++ b/README.md
@@ -99,19 +99,19 @@ client.login('token');
 
 ## Links
 
-- [Website](https://discord.js.org/) ([source](https://github.com/discordjs/website))
-- [Documentation](https://discord.js.org/#/docs)
+- [Website](https://discord.js.org/) ([source](https://github.com/discordjs/website)), Official Website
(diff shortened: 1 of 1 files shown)


el-durango:
The hyperlinks already explain what they link to, which makes this unnecessary.

Jiralite (commented):
I'm not sure how I feel about all of this. It was neatly presented, but the readability is reduced with this addition. I'm also not sure if this is even needed as they seem self-explanatory to me (the support server is exactly that, the documentation is exactly that... the documentation). If someone doesn't know what npm is at this point, well...

At the very least, I'd recommend changing everything to `[markdown](link "hover")` - add your extra information in the `hover` parameter so when users hover over the link, they can see the extra information. Otherwise, this seems unneeded.

iCrawl:
I think the links are fine as is and not confusing at all, and certainly not to a degree those descriptions are needed.

PR #6949, opened 2021-11-06
feat(GuildMember): Added communicationDisabledUntil property

ZeldaFan0225:

This PR includes the already present communication_disabled_until property returned by the API.
This should be merged so as soon Discord releases that feature people can use it.

Relevant PR in discord-api-docs:
- https://github.com/discord/discord-api-docs/pull/4075

### DIFF
--- a/src/errors/Messages.js
+++ b/src/errors/Messages.js
@@ -150,6 +150,9 @@ const Messages = {
   NOT_IMPLEMENTED: (what, name) => `Method ${what} not implemented on ${name}.`,
 
   SWEEP_FILTER_RETURN: 'The return value of the sweepFilter function was not false or a Function',
+
+  COMMUNICATION_DISABLED_PAST_DATE: 'The given date can\'t be in the past',
+  COMMUNICATION_DISABLED_NEGATIVE_DURATION: 'The given duration can\' be negative',
 };
 
 for (const [name, message] of Object.entries(Messages)) register(name, message);
--- a/src/managers/GuildMemberManager.js
+++ b/src/managers/GuildMemberManager.js
@@ -237,8 +237,16 @@ class GuildMemberManager extends CachedManager {
 
     // Clone the data object for immutability
     const _data = { ...data };
-    _data.communication_disabled_until = new Date(Date.now() + _data.communicationDisabledUntil * 1000).toISOString();
-    if (_data.communicationDisabledUntil < 0) throw new Error('The seconds have to be positive');
+    if (_data.communicationDisabledUntil) {
+      if (_data.communicationDisabledUntil instanceof Date) {
+        _data.communication_disabled_until = _data.communicationDisabledUntil;
+        if(_data.communication_disabled_until.getTime() < Date.now()) throw new Error("The given date can't be in the past");
+      } else {
+        _data.communication_disabled_until = new Date(Date.now() + _data.communicationDisabledUntil * 1000);
+        if(_data.communication_disabled_until.getTime() < Date.now()) throw new Error("The given amount of seconds can't be negative");
+      };
+      _data.communication_disabled_until = _data.communication_disabled_until.toISOString();
+    };
     if (_data.channel) {
       _data.channel = this.guild.channels.resolve(_data.channel);
       if (!(_data.channel instanceof BaseGuildVoiceChannel)) {
--- a/src/managers/GuildMemberManager.js
+++ b/src/managers/GuildMemberManager.js
@@ -237,8 +237,8 @@ class GuildMemberManager extends CachedManager {
 
     // Clone the data object for immutability
     const _data = { ...data };
-    _data.communication_disabled_until = new Date(Date.now()+(_data.communicationDisabledUntil*1000)).toISOString();
-    if(_data.communicationDisabledUntil < 0) throw new Error("The seconds have to be positive");
+    _data.communication_disabled_until = new Date(Date.now() + _data.communicationDisabledUntil * 1000).toISOString();
+    if (_data.communicationDisabledUntil < 0) throw new Error('The seconds have to be positive');
     if (_data.channel) {
       _data.channel = this.guild.channels.resolve(_data.channel);
       if (!(_data.channel instanceof BaseGuildVoiceChannel)) {
--- a/src/managers/GuildMemberManager.js
+++ b/src/managers/GuildMemberManager.js
@@ -240,10 +240,10 @@ class GuildMemberManager extends CachedManager {
     if (_data.communicationDisabledUntil) {
       if (_data.communicationDisabledUntil instanceof Date) {
         _data.communication_disabled_until = _data.communicationDisabledUntil;
-        if(_data.communication_disabled_until.getTime() < Date.now()) throw new Error("The given date can't be in the past");
+        if(_data.communication_disabled_until.getTime() < Date.now()) throw new Error('COMMUNICATION_DISABLED_PAST_DATE');
       } else {
         _data.communication_disabled_until = new Date(Date.now() + _data.communicationDisabledUntil * 1000);
-        if(_data.communication_disabled_until.getTime() < Date.now()) throw new Error("The given amount of seconds can't be negative");
+        if(_data.communication_disabled_until.getTime() < Date.now()) throw new Error('COMMUNICATION_DISABLED_PAST_DATE');
       };
       _data.communication_disabled_until = _data.communication_disabled_until.toISOString();
       delete _data.communicationDisabledUntil;
--- a/src/managers/GuildMemberManager.js
+++ b/src/managers/GuildMemberManager.js
@@ -246,6 +246,7 @@ class GuildMemberManager extends CachedManager {
         if(_data.communication_disabled_until.getTime() < Date.now()) throw new Error("The given amount of seconds can't be negative");
       };
       _data.communication_disabled_until = _data.communication_disabled_until.toISOString();
+      delete _data.communicationDisabledUntil;
     };
     if (_data.channel) {
       _data.channel = this.guild.channels.resolve(_data.channel);
--- a/src/managers/GuildMemberManager.js
+++ b/src/managers/GuildMemberManager.js
@@ -237,6 +237,8 @@ class GuildMemberManager extends CachedManager {
 
     // Clone the data object for immutability
     const _data = { ...data };
+    _data.communication_disabled_until = new Date(Date.now()+(_data.communicationDisabledUntil*1000)).toISOString();
+    if(_data.communicationDisabledUntil < 0) throw new Error("The seconds have to be positive");
     if (_data.channel) {
       _data.channel = this.guild.channels.resolve(_data.channel);
       if (!(_data.channel instanceof BaseGuildVoiceChannel)) {
--- a/src/structures/GuildMember.js
+++ b/src/structures/GuildMember.js
@@ -86,8 +86,10 @@ class GuildMember extends Base {
       this.premiumSinceTimestamp = data.premium_since ? new Date(data.premium_since).getTime() : null;
     }
     if ('roles' in data) this._roles = data.roles;
-    if('communication_disabled_until' in data) {
-      this.communicationDisabledUntilTimestamp = data.communication_disabled_until ? new Date(data.communication_disabled_until).getTime() : null;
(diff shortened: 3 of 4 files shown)


src/structures/GuildMember.js
```diff
@@ -80,6 +86,7 @@ class GuildMember extends Base {
       this.premiumSinceTimestamp = data.premium_since ? new Date(data.premium_since).getTime() : null;
     }
     if ('roles' in data) this._roles = data.roles;
+    this.communicationDisabledUntil = data.communication_disabled_until ? new Date(data.communication_disabled_until) : null;
```
el-durango, line 89: Shouldn't you use the `'prop' in obj` syntax here?

ZeldaFan0225: I don't think so since the communication_disabled_until property passed by discord is either `null` or a ISO8601 timestamp
image

ImRodry: But it could be missing from partial GuildMember objects, like the ones sent in guild member update events

el-durango: I mean something like:
```js
if ('communication_disabled_until' in data) this.communicationDisabledUntil = new Date (..)
```

ZeldaFan0225: i added it (see below)

src/structures/GuildMember.js
```diff
@@ -46,6 +46,12 @@ class GuildMember extends Base {
      */
     this.nickname = null;
 
+    /**
+     * When this member will be unmuted, if they are muted
+     * @type {?Date}
+     */
+    this.communicationDisabledUntilTimestamp = null;
```
ImRodry, line 53: So... is it a date or a timestamp?
Also this could probably be called communicationDisabledTimestamp, thoughts @discordjs/the-big-4 ?

src/structures/GuildMember.js
```diff
@@ -80,6 +86,9 @@ class GuildMember extends Base {
       this.premiumSinceTimestamp = data.premium_since ? new Date(data.premium_since).getTime() : null;
     }
     if ('roles' in data) this._roles = data.roles;
+    if('communication_disabled_until' in data) {
+      this.communicationDisabledUntilTimestamp = data.communication_disabled_until ? new Date(data.communication_disabled_until).getTime() : null;
```
ImRodry, line 90: you can assume that data.communication_disabled_until is always truthy here

src/structures/GuildMember.js
```diff
@@ -284,6 +302,18 @@ class GuildMember extends Base {
...
+  
+  /**
+   * Disables or enables communication for this member.
+   * @param {?Date} The duration in milliseconds the guild member should have communication disabled for, or `null` to enable communication for them
+   * @param {string} [reason] Reason for setting the nickname
+   * @returns {Promise<GuildMember>}
+   */
+  disableCommunication(duration, reason) {
+    const timestamp = Date.now()+duration;
+    return this.edit({communication_disabled_until: new Date(timestamp).toString()}, reason);
```
ImRodry, line 314: I highly doubt discord would require you to send an ISO string to modify a guild member, it's likely to be a value in seconds, but this still needs to be added to the #edit method (also formatting)

typings/index.d.ts
```diff
@@ -1032,6 +1032,7 @@ export class GuildMember extends PartialTextBasedChannel(Base) {
   public readonly displayHexColor: HexColorString;
   public readonly displayName: string;
   public guild: Guild;
+  public readonly communicationDisabledUntil: Date | null;
```
ImRodry, line 1035: Missing communicationDisabledUntilTimestamp, method and #edit parameter

ImRodry (changes requested):
This PR should be marked as a draft until the upstream PR is created/merged

src/structures/GuildMember.js
ZeldaFan0225: changed it to be a number now

ImRodry: Please resolve the conversation then

src/structures/GuildMember.js
ZeldaFan0225: no because it can also be null which would be falsy (when referring to the ternary)
also i dont know if it exists on a partial member so i put it in just to be sure

typings/index.d.ts
ZeldaFan0225: added

src/structures/GuildMember.js
```diff
@@ -284,6 +302,18 @@ class GuildMember extends Base {
...
+  
+  /**
+   * Disables or enables communication for this member.
+   * @param {?Date} The duration in milliseconds the guild member should have communication disabled for, or `null` to enable communication for them
+   * @param {string} [reason] Reason for setting the nickname
+   * @returns {Promise<GuildMember>}
+   */
+  disableCommunication(duration, reason) {
+    const timestamp = Date.now()+duration;
+    return this.edit({communication_disabled_until: new Date(timestamp).toISOString()}, reason);
```
ImRodry, line 314: The edit method accepts properties in camelCase. All of these checks should be done there and the duration should be a value in seconds with how long to mute the member for.

src/structures/GuildMember.js
```diff
@@ -284,6 +302,18 @@ class GuildMember extends Base {
     return this.edit({ nick }, reason);
   }
 
+  
+  /**
+   * Disables or enables communication for this member.
+   * @param {?Date} The duration in milliseconds the guild member should have communication disabled for, or `null` to enable communication for them
```
ImRodry, line 308: This probably crosses the length limit, you should consider splitting into multiple lines
```suggestion
   * @param {number|null} seconds The duration in seconds the guild member should have communication disabled for, or `null` to enable communication for them
```

typings/index.d.ts
```diff
@@ -1053,6 +1055,7 @@ export class GuildMember extends PartialTextBasedChannel(Base) {
   public createDM(force?: boolean): Promise<DMChannel>;
   public deleteDM(): Promise<DMChannel>;
   public displayAvatarURL(options?: ImageURLOptions): string;
+  public disableCommunication(duration: number, reason?: string): Promise<GuildMember>;
```
ImRodry, line 1058: The GuildMemberEditData interface still needs to be updated
```suggestion
  public disableCommunication(seconds: number, reason?: string): Promise<GuildMember>;
```

ImRodry (changes requested):

src/structures/GuildMember.js
ZeldaFan0225: what checks exactly?

ImRodry: Everything this function is doing. It should only be a call to this.edit with the provided value.

ZeldaFan0225: i moved the parsing to GuildMemberManager.edit

src/managers/GuildMemberManager.js
```diff
@@ -237,6 +237,8 @@ class GuildMemberManager extends CachedManager {
 
     // Clone the data object for immutability
     const _data = { ...data };
+    _data.communication_disabled_until = new Date(Date.now() + _data.communicationDisabledUntil * 1000).toISOString();
+    if (_data.communicationDisabledUntil < 0) throw new Error('The seconds have to be positive');
```
ImRodry, line 241: Seeing as the API supports an ISO string, you could make this accept a Date or a number which could be a UNIX timestamp in ms or the amount of seconds from now. You'd have to do more elaborate checks though. Also, you're currently passing a camelCased value to the API, which you should delete instead. Lastly, the error should be created in Messages.js and the message can be improved a lot.

typings/index.d.ts
ImRodry, line 1058:
```suggestion
  public disableCommunication(seconds: number | null, reason?: string): Promise<GuildMember>;
```

ImRodry (changes requested):
Still missing the new parameter in GuildMemberEditData

src/managers/GuildMemberManager.js
```diff
@@ -237,6 +237,17 @@ class GuildMemberManager extends CachedManager {
 
     // Clone the data object for immutability
     const _data = { ...data };
+    if (_data.communicationDisabledUntil) {
+      if (_data.communicationDisabledUntil instanceof Date) {
+        _data.communication_disabled_until = _data.communicationDisabledUntil;
+        if(_data.communication_disabled_until.getTime() < Date.now()) throw new Error('COMMUNICATION_DISABLED_PAST_DATE');
```
ckohen, line 243: Discord accepts dates in the past currently (and doesn't reset on sending null), so the lib shouldn't throw

src/managers/GuildMemberManager.js
```diff
@@ -237,6 +237,17 @@ class GuildMemberManager extends CachedManager {
...
+      } else {
+        _data.communication_disabled_until = new Date(Date.now() + _data.communicationDisabledUntil * 1000);
+        if(_data.communication_disabled_until.getTime() < Date.now()) throw new Error('COMMUNICATION_DISABLED_PAST_DATE');
```
ckohen, line 246: Ditto

src/errors/Messages.js
```diff
@@ -150,6 +150,9 @@ const Messages = {
   NOT_IMPLEMENTED: (what, name) => `Method ${what} not implemented on ${name}.`,
 
   SWEEP_FILTER_RETURN: 'The return value of the sweepFilter function was not false or a Function',
+
+  COMMUNICATION_DISABLED_PAST_DATE: 'The given date can\'t be in the past',
+  COMMUNICATION_DISABLED_NEGATIVE_DURATION: 'The given duration can\' be negative',
```
ckohen, line 155: This error is unused

ckohen (changes requested):
Based on the some of things seen by testing this (yes you can actually test it already, client doesn't do anything with it though), the API seems highly likely to change, just as an FYI

GodderE2D:
I feel like the terms "muted" and "unmuted" shouldn't be used here, but we'll wait for UI descriptions to see what we can replace them with, if at all.

ImRodry:
Please add the PR to the description: https://github.com/discord/discord-api-docs/pull/4075

milanmdev:
Discord released a "MODERATE_MEMBERS" permission - this should be added to this PR (#7020)

https://github.com/discord/discord-api-docs/pull/4075#issuecomment-975721690

src/structures/GuildMember.js
```diff
@@ -284,6 +304,18 @@ class GuildMember extends Base {
     return this.edit({ nick }, reason);
   }
 
+  
+  /**
+   * Disables or enables communication for this member.
+   * @param {?number | Date} seconds The duration in seconds or a date in the future until the member should have
+   * their communication disabled. Passing `null` will enable communication for them. 
+   * @param {string} [reason] Reason for setting the nickname
```
elderlabs, line 312: This feels like a typo as it does not appear a nickname is set here. 😉

elderlabs, line 312: Typo as a nickname is not set here, but instead the reason for the mute.

src/structures/GuildMember.js
```diff
@@ -284,6 +304,18 @@ class GuildMember extends Base {
     return this.edit({ nick }, reason);
   }
 
+  
+  /**
+   * Disables or enables communication for this member.
+   * @param {?number | Date} seconds The duration in seconds or a date in the future until the member should have
```
almeidx, line 310: The documentation doesn't render the `?` properly otherwise
```suggestion
   * @param {?(number|Date)} seconds The duration in seconds or a date in the future until the member should have
```

src/structures/GuildMember.js
```diff
@@ -284,6 +304,18 @@ class GuildMember extends Base {
...
+   * their communication disabled. Passing `null` will enable communication for them. 
+   * @param {string} [reason] Reason for setting the nickname
+   * @returns {Promise<GuildMember>}
+   */
+  disableCommunication(seconds, reason) {
```
almeidx, line 315: The name of the first argument doesn't match the one in the typings
Not to mention that `seconds` sounds misleading since it accepts dates

almeidx (changes requested):

ImRodry:
@ZeldaFan0225 are you no longer going to maintain this PR?

ZeldaFan0225:
@ImRodry no thats why i closed

PR #6956, opened 2021-11-07
fix(MessageReaction): Cache myself when `MessageReaction#me` is `true`.

GrapeColor:
It is inconsistent that I am not cached in `ReactionUserManager` when `MessageReaction#me` is `true`.
So when I create a `ReactionUserManager`, I cache myself.


vladfrangu (approved):

SpaceEEC (approved):
Probably falls under the partial-undefined-state-ness, but I suppose this doesn't hurt. 🤷

kyranet (approved):

PR #6957, opened 2021-11-07
feat(Channel): improve typeguards

ImRodry:
This PR creates typeguards for every channel type and renames the existing ones to isTextBased and isVoiceBased, as this was already what they did, allowing for accurate isText and isVoice typeguards.

### DIFF
--- a/packages/discord.js/src/structures/Channel.js
+++ b/packages/discord.js/src/structures/Channel.js
@@ -127,7 +127,7 @@ class Channel extends Base {
   }
 
   /**
-   * Indicates whether this channel is a {@link PartialGroupDMChannel}
+   * Indicates whether this channel is a {@link PartialGroupDMChannel}.
    * @returns {boolean}
    */
   isGroupDM() {
@@ -143,7 +143,7 @@ class Channel extends Base {
   }
 
   /**
-   * Indicates whether this channel is a {@link NewsChannel}
+   * Indicates whether this channel is a {@link NewsChannel}.
    * @returns {boolean}
    */
   isNews() {
@@ -151,7 +151,7 @@ class Channel extends Base {
   }
 
   /**
-   * Indicates whether this channel is a {@link StoreChannel}
+   * Indicates whether this channel is a {@link StoreChannel}.
    * @returns {boolean}
    */
   isStore() {
--- a/packages/discord.js/src/structures/Channel.js
+++ b/packages/discord.js/src/structures/Channel.js
@@ -3,7 +3,7 @@
 const { DiscordSnowflake } = require('@sapphire/snowflake');
 const { ChannelType } = require('discord-api-types/v9');
 const Base = require('./Base');
-const { ThreadChannelType } = require('../util/Constants');
+const { ThreadChannelTypes } = require('../util/Constants');
 let CategoryChannel;
 let DMChannel;
 let NewsChannel;
@@ -163,7 +163,7 @@ class Channel extends Base {
    * @returns {boolean}
    */
   isThread() {
-    return ThreadChannelType.includes(this.type);
+    return ThreadChannelTypes.includes(this.type);
   }
 
   /**
--- a/packages/discord.js/src/structures/Channel.js
+++ b/packages/discord.js/src/structures/Channel.js
@@ -3,6 +3,7 @@
 const { DiscordSnowflake } = require('@sapphire/snowflake');
 const { ChannelType } = require('discord-api-types/v9');
 const Base = require('./Base');
+const { ThreadChannelType } = require('../util/Constants');
 let CategoryChannel;
 let DMChannel;
 let NewsChannel;
@@ -11,7 +12,6 @@ let StoreChannel;
 let TextChannel;
 let ThreadChannel;
 let VoiceChannel;
-const { ThreadChannelTypes, VoiceBasedChannelTypes } = require('../util/Constants');
 
 /**
  * Represents any channel on Discord.
@@ -103,19 +103,59 @@ class Channel extends Base {
   }
 
   /**
-   * Indicates whether this channel is {@link TextBasedChannels text-based}.
+   * Indicates whether this channel is a {@link TextChannel}.
    * @returns {boolean}
    */
   isText() {
-    return 'messages' in this;
+    return this.type === ChannelType[ChannelType.GuildText];
   }
 
   /**
-   * Indicates whether this channel is {@link BaseGuildVoiceChannel voice-based}.
+   * Indicates whether this channel is a {@link DMChannel}.
+   * @returns {boolean}
+   */
+  isDM() {
+    return this.type === ChannelType[ChannelType.DM];
+  }
+
+  /**
+   * Indicates whether this channel is a {@link VoiceChannel}.
    * @returns {boolean}
    */
   isVoice() {
-    return VoiceBasedChannelTypes.includes(this.type);
+    return this.type === ChannelType[ChannelType.GuildVoice];
+  }
+
+  /**
+   * Indicates whether this channel is a {@link PartialGroupDMChannel}
+   * @returns {boolean}
+   */
+  isGroupDM() {
+    return this.type === ChannelType[ChannelType.GroupDM];
+  }
+
+  /**
+   * Indicates whether this channel is a {@link CategoryChannel}.
+   * @returns {boolean}
+   */
+  isCategory() {
+    return this.type === ChannelType[ChannelType.GuildCategory];
+  }
+
+  /**
+   * Indicates whether this channel is a {@link NewsChannel}
+   * @returns {boolean}
+   */
+  isNews() {
+    return this.type === ChannelType[ChannelType.GuildNews];
+  }
+
+  /**
+   * Indicates whether this channel is a {@link StoreChannel}
+   * @returns {boolean}
+   */
+  isStore() {
+    return this.type === ChannelType[ChannelType.GuildStore];
   }
 
   /**
@@ -123,7 +163,31 @@ class Channel extends Base {
    * @returns {boolean}
    */
   isThread() {
-    return ThreadChannelTypes.includes(this.type);
+    return ThreadChannelType.includes(this.type);
+  }
+
+  /**
+   * Indicates whether this channel is a {@link StageChannel}.
+   * @returns {boolean}
+   */
+  isStage() {
+    return this.type === ChannelType[ChannelType.GuildStageVoice];
+  }
+
+  /**
+   * Indicates whether this channel is {@link TextBasedChannels text-based}.
+   * @returns {boolean}
+   */
+  isTextBased() {
+    return 'messages' in this;
+  }
+
+  /**
+   * Indicates whether this channel is {@link BaseGuildVoiceChannel voice-based}.
+   * @returns {boolean}
+   */
+  isVoiceBased() {
+    return 'bitrate' in this;
   }
 
   static create(client, data, guild, { allowUnknownGuild, fromInteraction } = {}) {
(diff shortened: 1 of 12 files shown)


src/structures/Channel.js
```diff
@@ -108,27 +108,91 @@ class Channel extends Base {
...
+    return this.type === 'GUILD_STORE';
   }
 
   /**
    * Indicates whether this channel is a {@link ThreadChannel}.
    * @returns {boolean}
    */
   isThread() {
-    return ThreadChannelTypes.includes(this.type);
+    return 'autoArchiveDuration' in this;
```
papsavas, line 171: Wouldn't `this.type.endsWith('_THREAD')` or atleast `includes('THREAD')` be more suitable according to their [naming](https://github.com/discord/discord-api-docs/blob/master/docs/resources/Channel.md?plain=1#L54-L56)?

ImRodry: What if they add more thread types that don’t end with _THREAD? Unlikely, but still possible and I don’t think a thread would ever not have autoArchiveDuration, however I’m happy to revert if the maintainers think I should, just did that for consistency with the isTextBased method

papsavas: A defined `parentId` can also indicate a channel's category. Perhaps a combination of checks is safer

ImRodry: I can always revert to the thread channel types check, but I think it’d be best to use the same check across all methods

vladfrangu: I agree with @papsavas here, checking `autoArchiveDuration` makes this type guard feel out of place... Either use an array of thread types and check if this.type is present (ideal), or check if the type includes THREAD (less ideal but still better)

ImRodry: Alright I’ll change these ones to check for the type and keep the isXBased ones checking for a property in the channel

src/structures/Channel.js
```diff
@@ -108,19 +108,59 @@ class Channel extends Base {
   }
 
   /**
-   * Indicates whether this channel is {@link TextBasedChannels text-based}.
+   * Indicates whether this channel is a {@link TextChannel}.
    * @returns {boolean}
    */
   isText() {
-    return 'messages' in this;
+    return this.type === 'GUILD_TEXT';
```
papsavas, line 115:
```suggestion
    return this.type === ChannelTypes.GUILD_TEXT;
```

src/structures/Channel.js
```diff
@@ -108,19 +108,59 @@ class Channel extends Base {
...
   }
 
   /**
-   * Indicates whether this channel is {@link BaseGuildVoiceChannel voice-based}.
+   * Indicates whether this channel is a {@link DMChannel}.
+   * @returns {boolean}
+   */
+  isDM() {
+    return this.type === 'DM';
```
papsavas, line 123:
```suggestion
    return this.type === ChannelTypes.DM;
```

src/structures/Channel.js
```diff
@@ -108,19 +108,59 @@ class Channel extends Base {
...
+  }
+
+  /**
+   * Indicates whether this channel is a {@link VoiceChannel}.
    * @returns {boolean}
    */
   isVoice() {
-    return VoiceBasedChannelTypes.includes(this.type);
+    return this.type === 'GUILD_VOICE';
```
papsavas, line 131:
```suggestion
    return this.type === ChannelTypes.GUILD_VOICE;
```

src/structures/Channel.js
```diff
@@ -108,19 +108,59 @@ class Channel extends Base {
...
+  }
+
+  /**
+   * Indicates whether this channel is a {@link PartialGroupDMChannel}
+   * @returns {boolean}
+   */
+  isGroupDM() {
+    return this.type === 'GROUP_DM';
```
papsavas, line 139:
```suggestion
    return this.type === ChannelTypes.GROUP_DM;
```

src/structures/Channel.js
```diff
@@ -108,19 +108,59 @@ class Channel extends Base {
...
+  }
+
+  /**
+   * Indicates whether this channel is a {@link CategoryChannel}.
+   * @returns {boolean}
+   */
+  isCategory() {
+    return this.type === 'GUILD_CATEGORY';
```
papsavas, line 147:
```suggestion
    return this.type === ChannelTypes.GUILD_CATEGORY;
```

src/structures/Channel.js
```diff
@@ -108,19 +108,59 @@ class Channel extends Base {
...
+  }
+
+  /**
+   * Indicates whether this channel is a {@link NewsChannel}
+   * @returns {boolean}
+   */
+  isNews() {
+    return this.type === 'GUILD_NEWS';
```
papsavas, line 155:
```suggestion
    return this.type === ChannelTypes.GUILD_NEWS;
```

src/structures/Channel.js
```diff
@@ -108,19 +108,59 @@ class Channel extends Base {
...
+  }
+
+  /**
+   * Indicates whether this channel is a {@link StoreChannel}
+   * @returns {boolean}
+   */
+  isStore() {
+    return this.type === 'GUILD_STORE';
```
papsavas, line 163:
```suggestion
    return this.type === ChannelTypes.GUILD_STORE;
```

src/structures/Channel.js
```diff
@@ -131,6 +171,30 @@ class Channel extends Base {
     return ThreadChannelTypes.includes(this.type);
   }
 
+  /**
+   * Indicates whether this channel is a {@link StageChannel}.
+   * @returns {boolean}
+   */
+  isStage() {
+    return this.type === 'GUILD_STAGE_VOICE';
```
papsavas, line 179:
```suggestion
    return this.type === ChannelTypes.GUILD_STAGE_VOICE;
```

papsavas (changes requested):
Feels like this would be preferred rather than raw strings

src/structures/Channel.js
vladfrangu: I'm all for enum usages but this would never return true... pretty sure that ChannelTypes.GUILD_TEXT is a number, not a string

papsavas: I tested this and you're right. I will change it to ```ChannelTypes[ChannelTypes.X]``` for now but perhaps a function for this would be a little more subtle

suneettipirneni: I mean, to be fair using an indexed access here seems like a lot more verbose, and it (maybe...?) has performance implications rather than using a string comparison.

papsavas: All (afaik) true, but i still believe that raw strings are not the way to go. If discord decides to rename channel types for some reason (especially since they work with int numbers so all good), it gets messy. We should try and minimize the exposure of raw strings all around the project and gather them in the least possible files needed. This way, something changes, you change it once.

ImRodry: If the type name changed you’d be having to change the name of the method most likely anyway. Isn’t this solution much worse performance wise? And I mean you still have the channel type string there so it doesn’t improve anything in that regard

vladfrangu (approved):

iCrawl (approved):

SpaceEEC (changes requested):
You also need to update usages of `isText` to `isTextBased`.

ImRodry:
CI will fail until #7215 is merged

kyranet (approved):

SpaceEEC (approved):

packages/discord.js/typings/index.d.ts
```diff
@@ -1054,7 +1062,7 @@ export abstract class GuildChannel extends Channel {
   public setName(name: string, reason?: string): Promise<this>;
   public setParent(channel: CategoryChannelResolvable | null, options?: SetParentOptions): Promise<this>;
   public setPosition(position: number, options?: SetChannelPositionOptions): Promise<this>;
-  public isText(): this is TextChannel | NewsChannel;
+  public isTextBased(): this is TextChannel | NewsChannel | ThreadChannel;
```
Jiralite, line 1065: I'm wondering if it's better to have this as the intersection `GuildBasedChannel & TextBasedChannel` instead? It makes it a lil' more dynamic then, right?

packages/discord.js/src/structures/Channel.js
```diff
@@ -102,19 +102,59 @@ class Channel extends Base {
...
+   * Indicates whether this channel is a {@link VoiceChannel}.
    * @returns {boolean}
    */
   isVoice() {
-    return VoiceBasedChannelTypes.includes(this.type);
+    return this.type === ChannelTypes[ChannelTypes.GUILD_VOICE];
+  }
+
+  /**
+   * Indicates whether this channel is a {@link PartialGroupDMChannel}
```
Jiralite, line 129:
```suggestion
   * Indicates whether this channel is a {@link PartialGroupDMChannel}.
```

packages/discord.js/src/structures/Channel.js
```diff
@@ -102,19 +102,59 @@ class Channel extends Base {
...
+  /**
+   * Indicates whether this channel is a {@link CategoryChannel}.
+   * @returns {boolean}
+   */
+  isCategory() {
+    return this.type === ChannelTypes[ChannelTypes.GUILD_CATEGORY];
+  }
+
+  /**
+   * Indicates whether this channel is a {@link NewsChannel}
```
Jiralite, line 145:
```suggestion
   * Indicates whether this channel is a {@link NewsChannel}.
```

packages/discord.js/src/structures/Channel.js
```diff
@@ -102,19 +102,59 @@ class Channel extends Base {
...
+   * @returns {boolean}
+   */
+  isNews() {
+    return this.type === ChannelTypes[ChannelTypes.GUILD_NEWS];
+  }
+
+  /**
+   * Indicates whether this channel is a {@link StoreChannel}
```
Jiralite, line 153:
```suggestion
   * Indicates whether this channel is a {@link StoreChannel}.
```

Jiralite (commented):
Some consistency suggestions and a TypeScript comment!

packages/discord.js/src/structures/Channel.js
ImRodry: y tho

packages/discord.js/typings/index.d.ts
ImRodry: Absolutely, will do

packages/discord.js/src/structures/Channel.js
Jiralite: Because you've done it for some others, and that's just really... unsettling?

iCrawl:
With #7077 merged you'll have to adapt your PR to it :+1:

packages/discord.js/src/structures/Channel.js
```diff
@@ -3,6 +3,7 @@
 const { DiscordSnowflake } = require('@sapphire/snowflake');
 const { ChannelType } = require('discord-api-types/v9');
 const Base = require('./Base');
+const { ThreadChannelType } = require('../util/Constants');
```
almeidx, line 6:
```suggestion
const { ThreadChannelTypes } = require('../util/Constants');
```

packages/discord.js/src/structures/Channel.js
```diff
@@ -103,27 +103,91 @@ class Channel extends Base {
...
+    return this.type === ChannelType[ChannelType.GuildStore];
   }
 
   /**
    * Indicates whether this channel is a {@link ThreadChannel}.
    * @returns {boolean}
    */
   isThread() {
-    return ThreadChannelTypes.includes(this.type);
+    return ThreadChannelType.includes(this.type);
```
almeidx, line 166:
```suggestion
    return ThreadChannelTypes.includes(this.type);
```

Issue #6958, opened 2021-11-07
Remove `*Data` types from Discord.js types


suneettipirneni:
### Feature

A big pain point I've noticed recently is that there's a good deal of duplicated work when it comes to both discord-api-types and the discord.js-equivalent types. 95% of the time changes made in one need to be reflected in the other. And it's not always as simple as `snake_case` to `camelCase`. As has been evident by the latest updates, our types have become more complex involving tagged unions, intersections and custom utility types. A couple of examples of are command option types, and the autocomplete option types.

When slash commands (or insert any thing with a `type` prop) were added, their type in d.js was significantly weaker than it's discord-api-types counterpart. It wasn't less strict on purpose, it just wasn't based on the dapi types. And this mistake is completely understandable, because not everyone is going to look at the DAPI types implementation when making a PR for discord.js. However this represents a problem with how we use types: there's no single source of truth, therefore it invites disparity.

Because of this redundancy I'm proposing that we remove `*Data` types from discord.js and use `discord-api-types` as the replacement. As for the concerns of doing this, I'll try my best to address them below:

### We can't use dapi types because they're `snake_cased` while djs types are `camelCase`:
As far as I can tell this is the biggest roadblock as to not using DAPI types directly in discord.js. Luckily with the power of mapped types in typescript, this issue can be solved. You can combine template string literals with key remapping to convert any snake cased object into a camel-cased one:

```ts
// Types that transform snake_case to camelCase
type CamelCase<S extends string> = S extends `${infer prefix}_${infer suffix}${infer tail}`
  ? `${Lowercase<prefix>}${Uppercase<suffix>}${CamelCase<tail>}`
  : Lowercase<S>

type Camelize<T> = {
    [K in keyof T as CamelCase<string &K>]: T[K] extends {} ? Camelize<T[K]> : T[K]
}
```

Then in the type definition we simply use this utility type:
```ts
// Psuedo code
addCommandOption(data: Camelize<APIApplicationCommandOption>);
```
All keys from `APIApplicationCommandOption` are now camel cased even it's sub-objects **(yes, `Camelize<S>` is recursive)**.

Personally, I like the aesthetics of `Camelize<APIApplicationCommandOption>` because it signifies to the caller that this is simply a camel-cased version of the DAPI types, nothing more, nothing less.

### How would enums be handled:
[...]

PR #6959, opened 2021-11-09
types(ApplicationCommandManager): Deprecate old `*Data` type usages and allow camel cased dapi types to be used

suneettipirneni:

Ref: #6958 


ImRodry:
Do we really need to add a dependency for types only? Most people don't use typescript and the package offers much more than the type that is being imported, I believe we could simply create the type ourselves instead

suneettipirneni:
> Do we really need to add a dependency for types only? Most people don't use typescript and the package offers much more than the type that is being imported, I believe we could simply create the type ourselves instead

It's a dev dep, so why does it matter. And the camel case util type here is more sophisticated. It's not that simple to implement one our own that's as robust as this one.

ImRodry:
If it’s a dev dev then these types won’t even work at all because the package wouldn’t get installed. We can simply copy the type from them and add it ourselves

typings/index.d.ts
```diff
@@ -137,6 +138,7 @@ import {
   RawWidgetData,
   RawWidgetMemberData,
 } from './rawDataTypes';
+import type { CamelCasedPropertiesDeep } from 'type-fest';
```
almeidx, line 141:
```suggestion
```

almeidx (changes requested):

suneettipirneni:
We won't be using our own version of Camelizer. We are going to stick with the package that was originally brought up in this PR.

typings/index.d.ts
```diff
@@ -2627,7 +2635,11 @@ export abstract class CachedManager<K, Holds, R> extends DataManager<K, Holds, R
   private _add(data: unknown, cache?: boolean, { id, extras }?: { id: K; extras: unknown[] }): Holds;
 }
 
-export type ApplicationCommandDataResolvable = ApplicationCommandData | RESTPostAPIApplicationCommandsJSONBody;
+export type ApplicationCommandDataResolvable =
+  | Camelize<RESTPostAPIApplicationCommandsJSONBody>
+  | RESTPostAPIApplicationCommandsJSONBody;
+
+export type foo = Camelize<RESTPostAPIApplicationCommandsJSONBody>['options'];
```
almeidx, line 2642: 😕

tslint.json
```diff
@@ -24,6 +24,7 @@
     "array-type": [true, "array"],
     "one-line": false,
     "no-any-union": false,
-    "void-return": false
+    "void-return": false,
+    "unified-signatures": false
```
Jiralite, line 28: With the introduction of `unified-signatures: false`, I believe these can be safely removed now:

https://github.com/discordjs/discord.js/blob/9d7152f0fdcd879cad654fe4fdd658fac57ceb02/typings/index.d.ts#L2341-L2345

https://github.com/discordjs/discord.js/blob/9d7152f0fdcd879cad654fe4fdd658fac57ceb02/typings/index.d.ts#L3095-L3099

vladfrangu:
Well what do ya know! This needs a rebase

typings/tests.ts
```diff
@@ -880,13 +898,18 @@ declare const applicationNonChoiceOptionData: ApplicationCommandOptionData & {
   // @ts-expect-error
   applicationNonChoiceOptionData.choices;
 }
+declare const applicationChoiceOptionData: APIApplicationCommandOption & { type: ApplicationCommandOptionType.String };
+{
+  // Choices should be available.
+  applicationChoiceOptionData.choices;
+}
```
vladfrangu, line 905: What's the point of this test? You're just checking a discord-api-type 👀

vladfrangu (approved):
🎉

iCrawl:

SpaceEEC (approved):

PR #6966, opened 2021-11-10
refactor(MessageEmbed): Utilise an object approach for `.setAuthor()`

### LINKED ISSUE #6963: Consider an object approach for `MessageEmbed#setAuthor()`
### Feature

I'm always frustrated when I have to pass `undefined` through `MessageEmbed#setAuthor()` in order to reach the last parameter:

```JavaScript
const { MessageEmbed } = require("discord.js");
const embed = new MessageEmbed();
// data is an object from an API
embed.setAuthor(data.author, undefined, data.link);
```

I don't want an icon appearing in the embed, but discord.js enforces that the `iconURL` parameter must be provided before supplying the `url` parameter (of which I only want). Supplying `undefined` here seems... off; `null` would be a better choice, no?

Pass an object as the sole parameter:

```JavaScript
const { MessageEmbed } = require("discord.js");
const embed = new MessageEmbed();
// data is an object from an API
embed.setAuthor({ name: data.author, url: data.link });
```

This will allow one to not specify an `undefined` property to reach one's desired parameters.

An alternative solution could be to simply replace `undefined` with `null` (allow a `null` value). TypeScript insists that this cannot be done. However, with the object approach, this wouldn't even be needed.

Another alternative solution would be to swap the priorities with the last two arguments. Whilst that does solve _this_ issue, the same problem will be encountered for wanting the `iconURL` parameter without specifying `undefined`.

### Other context

_No response_

Jiralite:
As described in #6963, accessing the last parameter when one did not need the parameter before led to an ungraceful situation. This pull request resolves #6963.

### DIFF
--- a/src/structures/MessageEmbed.js
+++ b/src/structures/MessageEmbed.js
@@ -346,14 +346,28 @@ class MessageEmbed {
   }
 
   /**
+   * The options to provide for setting an author for a {@link MessageEmbed}.
+   * @typedef {Object} MessageEmbedAuthor
+   * @property {string} name The name of this author.
+   * @property {string} [url] The URL of this author.
+   * @property {string} [iconURL] The icon URL of this author.
+   */
+
+  // TODO: Remove the deprecated parameters.
+  /* eslint-disable-next-line valid-jsdoc */
+  /**
    * Sets the author of this embed.
-   * @param {string} name The name of the author
-   * @param {string} [iconURL] The icon URL of the author
-   * @param {string} [url] The URL of the author
+   * @param {string | MessageEmbedAuthor} options The options to provide for the author.
+   * A string may simply be provided if only the author name is desirable.
    * @returns {MessageEmbed}
    */
-  setAuthor(name, iconURL, url) {
-    this.author = { name: Util.verifyString(name, RangeError, 'EMBED_AUTHOR_NAME'), iconURL, url };
+  setAuthor(options = {}, deprecatedIconURL, deprecatedAuthorURL) {
+    if (typeof nameOrOptions === 'string') {
+      options = { name: options, url: deprecatedAuthorURL, iconURL: deprecatedIconURL };
+    }
+
+    const { name, url, iconURL } = options;
+    this.author = { name: Util.verifyString(name, RangeError, 'EMBED_AUTHOR_NAME'), url, iconURL };
     return this;
   }
 
--- a/src/structures/MessageEmbed.js
+++ b/src/structures/MessageEmbed.js
@@ -356,7 +356,7 @@ class MessageEmbed {
   // TODO: Remove the deprecated code in the following method and typings.
   /**
    * Sets the author of this embed.
-   * @param {string | EmbedAuthorData} options The options to provide for the author.
+   * @param {string|EmbedAuthorData} options The options to provide for the author.
    * A string may simply be provided if only the author name is desirable.
    * @param {string} [deprecatedIconURL] The icon URL of this author.
    * <warn>This parameter is **deprecated**. Use the `options` parameter instead.</warn>
--- a/src/structures/MessageEmbed.js
+++ b/src/structures/MessageEmbed.js
@@ -347,7 +347,7 @@ class MessageEmbed {
 
   /**
    * The options to provide for setting an author for a {@link MessageEmbed}.
-   * @typedef {Object} MessageEmbedAuthor
+   * @typedef {Object} EmbedAuthorData
    * @property {string} name The name of this author.
    * @property {string} [url] The URL of this author.
    * @property {string} [iconURL] The icon URL of this author.
@@ -356,7 +356,7 @@ class MessageEmbed {
   // TODO: Remove the deprecated code in the following method and typings.
   /**
    * Sets the author of this embed.
-   * @param {string | MessageEmbedAuthor} options The options to provide for the author.
+   * @param {string | EmbedAuthorData} options The options to provide for the author.
    * A string may simply be provided if only the author name is desirable.
    * @param {string} [deprecatedIconURL] The icon URL of this author.
    * <warn>This parameter is **deprecated**. Use the `options` parameter instead.</warn>
--- a/src/structures/MessageEmbed.js
+++ b/src/structures/MessageEmbed.js
@@ -359,12 +359,12 @@ class MessageEmbed {
    * @param {string | MessageEmbedAuthor} options The options to provide for the author.
    * A string may simply be provided if only the author name is desirable.
    * @param {string} [deprecatedIconURL] The icon URL of this author. This parameter is **deprecated**.
-   * @param {string} [deprecatedAuthorURL] The URL of this author. This parameter is **deprecated**.
+   * @param {string} [deprecatedURL] The URL of this author. This parameter is **deprecated**.
    * @returns {MessageEmbed}
    */
-  setAuthor(options = {}, deprecatedIconURL, deprecatedAuthorURL) {
+  setAuthor(options = {}, deprecatedIconURL, deprecatedURL) {
     if (typeof nameOrOptions === 'string') {
-      options = { name: options, url: deprecatedAuthorURL, iconURL: deprecatedIconURL };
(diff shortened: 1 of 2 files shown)


src/structures/MessageEmbed.js
```diff
@@ -345,15 +345,30 @@ class MessageEmbed {
...
+   * @param {string | MessageEmbedAuthor} options The options to provide for the author.
+   * A string may simply be provided if only the author name is desirable.
+   * @param {string} [deprecatedIconURL] The icon URL of this author. This parameter is **deprecated**.
+   * @param {string} [deprecatedURL] The URL of this author. This parameter is **deprecated**.
    * @returns {MessageEmbed}
    */
-  setAuthor(name, iconURL, url) {
-    this.author = { name: Util.verifyString(name, RangeError, 'EMBED_AUTHOR_NAME'), iconURL, url };
+  setAuthor(options = {}, deprecatedIconURL, deprecatedURL) {
+    if (typeof nameOrOptions === 'string') {
```
suneettipirneni, line 366: I might be missing something here, but where is `nameOrOptions` defined?

Jiralite: Oh no!

src/structures/MessageEmbed.js
```diff
@@ -345,15 +345,30 @@ class MessageEmbed {
     return this;
   }
 
+  /**
+   * The options to provide for setting an author for a {@link MessageEmbed}.
+   * @typedef {Object} MessageEmbedAuthor
+   * @property {string} name The name of this author.
+   * @property {string} [url] The URL of this author.
+   * @property {string} [iconURL] The icon URL of this author.
```
suneettipirneni, line 353:
```suggestion
   * @property {string} name The name of this author
   * @property {?string} [url] The URL of this author
   * @property {?string} [iconURL] The icon URL of this author
```

src/structures/MessageEmbed.js
```diff
@@ -345,15 +345,32 @@ class MessageEmbed {
...
    * Sets the author of this embed.
-   * @param {string} name The name of the author
-   * @param {string} [iconURL] The icon URL of the author
-   * @param {string} [url] The URL of the author
+   * @param {string | MessageEmbedAuthor} options The options to provide for the author.
+   * A string may simply be provided if only the author name is desirable.
+   * @param {string} [deprecatedIconURL] The icon URL of this author.
+   * <warn>This parameter is **deprecated**. Use the `options` parameter instead.</warn>
+   * @param {string} [deprecatedURL] The URL of this author.
+   * <warn>This parameter is **deprecated**. Use the `options` parameter instead.</warn>
```
suneettipirneni, line 364: Just for convention purposes

```suggestion
   * @param {string | MessageEmbedAuthor} options The options to provide for the author
   * A string may simply be provided if only the author name is desirable
   * @param {string} [deprecatedIconURL] The icon URL of this author
   * <warn>This parameter is **deprecated**. Use the `options` parameter instead</warn>
   * @param {string} [deprecatedURL] The URL of this author
   * <warn>This parameter is **deprecated**. Use the `options` parameter instead</warn>
```

suneettipirneni (changes requested):

src/structures/MessageEmbed.js
ImRodry: Why?

src/structures/MessageEmbed.js
```diff
@@ -345,15 +345,32 @@ class MessageEmbed {
...
+   * @property {string} [iconURL] The icon URL of this author.
+   */
+
+  // TODO: Remove the deprecated code in the following method and typings.
   /**
    * Sets the author of this embed.
-   * @param {string} name The name of the author
-   * @param {string} [iconURL] The icon URL of the author
-   * @param {string} [url] The URL of the author
+   * @param {string | EmbedAuthorData} options The options to provide for the author.
```
almeidx, line 359:
```suggestion
   * @param {string|EmbedAuthorData} options The options to provide for the author.
```

almeidx (changes requested):

src/structures/MessageEmbed.js
```diff
@@ -345,15 +345,32 @@ class MessageEmbed {
...
+   * @property {string} [iconURL] The icon URL of this author.
+   */
+
+  // TODO: Remove the deprecated code in the following method and typings.
   /**
    * Sets the author of this embed.
-   * @param {string} name The name of the author
-   * @param {string} [iconURL] The icon URL of the author
-   * @param {string} [url] The URL of the author
+   * @param {string|EmbedAuthorData} options The options to provide for the author.
```
vladfrangu, line 359: Technically options is optional now
```suggestion
   * @param {string|EmbedAuthorData} [options={}] The options to provide for the author.
```

src/structures/MessageEmbed.js
vladfrangu: Why what?

vladfrangu (changes requested):

src/structures/MessageEmbed.js
Jiralite: This has given me a thought: how does one remove an author with this method? Should we allow `null` to empty the author object just like`MessageEmbed#setTimestamp()`?

vladfrangu: I mean calling .setAuthor will already make it empty, no?

Jiralite: That's true, but calling a set method with no arguments feels like it should error or be set with a default value. `null` is being explicit to remove. Thoughts?

src/structures/MessageEmbed.js
ImRodry: I… can’t remember? Just ignore this bruh

src/structures/MessageEmbed.js
vladfrangu: We don't do it for any other method, like setDescription or such, they all require a string, be it empty... @discordjs/the-big-4 what do?

kyranet: Currently, it won't set to `null`, but rather will cause an error to be thrown on line 372:

```js
    const { name, url, iconURL } = options;
```

Perhaps we can do a manual check and set it to the default value when `null` is passed?

SpaceEEC: I'd be fine with clearing the underlying properties (or setting to null) when null is passed to a `set__` method.

src/structures/MessageEmbed.js
```diff
@@ -345,15 +345,38 @@ class MessageEmbed {
...
+   * <warn>This parameter is **deprecated**. Use the `options` parameter instead.</warn>
    * @returns {MessageEmbed}
    */
-  setAuthor(name, iconURL, url) {
-    this.author = { name: Util.verifyString(name, RangeError, 'EMBED_AUTHOR_NAME'), iconURL, url };
+  setAuthor(options, deprecatedIconURL, deprecatedURL) {
+    if (options === null) {
+      this.author = {};
+      return this;
+    }
```
kyranet, line 372: Those lines should be moved after the typeof check to keep the behaviour of the overload.

Otherwise `.setAuthor(null, url, iconURL)` will make the embed not throw.

kyranet (changes requested):

kyranet (approved):

src/structures/MessageEmbed.js
```diff
@@ -345,15 +345,38 @@ class MessageEmbed {
...
+
+    if (typeof options === 'string') {
+      options = { name: options, url: deprecatedURL, iconURL: deprecatedIconURL };
```
vladfrangu, line 375: Nit: should we emit a deprecation warning about this? @discordjs/the-big-4

vladfrangu (approved):
Outside the nit, this LGTM

SpaceEEC (approved):

src/structures/MessageEmbed.js
SpaceEEC: Probably a reasonable thing to do. (Maybe also on other places where we currently don't emit a warning.)

iCrawl (approved):

PR #6975, opened 2021-11-12
feat(ThreadChannel): add `ThreadChannel#viewable`

### LINKED ISSUE #6969: Message.editable doesn't work in threads
### Issue description

Steps to reproduce:
1. register event for the `messageCreate` event.
2. Send a message in a thread
3. See how it fails.

Possible cause (I'm almost certain that this is the cause):
It tries to run `this.channel?.viewable` on the thread channel but thread channels do not have a `viewable` property resulting in null/false

```javascript
  get editable() {
    return Boolean(this.author.id === this.client.user.id && !this.deleted && (!this.guild || this.channel?.viewable));
  }
```
[Source](https://github.com/discordjs/discord.js/blob/3c857a63632ef11a1f7aae00c98e1f1b9909bc87/src/structures/Message.js#L554)

### Code sample

```typescript
client.on('messageCreate', (message) => {
    if (message.channel.type == ChannelType.GuildPublicThread) {
        if (message.editable) {
            message.edit("Edited message");
        }
    }
})
```

### discord.js version

13.3.1

### Node.js version

16.6.1

### Operating system

Windows

Low (slightly annoying)

No Partials

GUILDS, GUILD_MEMBERS, GUILD_BANS, GUILD_MESSAGES, GUILD_MESSAGE_REACTIONS, GUILD_MESSAGE_TYPING

_No response_

suneettipirneni:

Closes #6969 

### DIFF
--- a/src/structures/Message.js
+++ b/src/structures/Message.js
@@ -551,7 +551,15 @@ class Message extends Base {
    * @readonly
    */
   get editable() {
-    return Boolean(this.author.id === this.client.user.id && !this.deleted && (!this.guild || this.channel?.viewable));
+    const precheck = Boolean(
+      this.author.id === this.client.user.id && !this.deleted && (!this.guild || this.channel?.viewable),
+    );
+    // Regardless of permissions thread messages cannot be edited if
+    // the thread is locked.
+    if (this.channel?.isThread()) {
+      return precheck && !this.channel.locked;
+    }
+    return precheck;
   }
 
   /**
--- a/src/structures/ThreadChannel.js
+++ b/src/structures/ThreadChannel.js
@@ -443,6 +443,18 @@ class ThreadChannel extends Channel {
   }
 
   /**
+   * Whether the thread is viewable by the client user
+   * @type {boolean}
+   * @readonly
+   */
+  get viewable() {
+    if (this.client.user.id === this.guild.ownerId) return true;
+    const permissions = this.permissionsFor(this.client.user);
+    if (!permissions) return false;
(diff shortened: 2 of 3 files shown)


Jiralite (changes requested):
Whilst this pull request _does_ fix the issue linked, it doesn't fix all instances of it.

Consider the scenario where a thread becomes locked or the client fetches a thread that happens to be locked without the manage threads permission. Then, consider the client fetching a message and running `Message#editable`. Through that, the result will be `true` but it should be `false` because the thread is locked and messages cannot be edited.

suneettipirneni:
> Whilst this pull request _does_ fix the issue linked, it doesn't fix all instances of it.
> 
> 
> 
> Consider the scenario where a thread becomes locked or the client fetches a thread that happens to be locked without the manage threads permission. Then, consider the client fetching a message and running `Message#editable`. Through that, the result will be `true` but it should be `false` because the thread is locked and messages cannot be edited.

Good catch! I'm thinking even if it's locked and you have manage_threads perms `editable` should probably still be `false`, simply because the current state of the thread is still archived, and I believe editing a message doesn't implicitly unarchive it.

Jiralite:
> I'm thinking even if it's locked and you have manage_threads perms `editable` should probably still be `false`

Right.

iCrawl (approved):

Jiralite (approved):

SpaceEEC (approved):

kyranet (approved):

PR #6982, opened 2021-11-13
fix(Util): ID sorting in `discordSort`

That-Guy977:
The Discord client sorts channels with the same position by ID, ascending. Previously `discordSort` would sort descending

- Swapped sort order of `a` and `b` when sorting by ID (which is required when `rawPosition` is duplicate between the 2 channels

From testing, the client sorts ascending, but `discordSort` would sort descending

- Typings don't need updating (Only values have changed)

### DIFF
--- a/src/util/Util.js
+++ b/src/util/Util.js
@@ -484,8 +484,8 @@ class Util extends null {
     return collection.sorted(
       (a, b) =>
         a.rawPosition - b.rawPosition ||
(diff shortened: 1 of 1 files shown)


That-Guy977:
This needs further discussion, as this change will break positions for roles, as channels are sorted ascending, but roles are sorted descending.

PR #6983, opened 2021-11-13
CI: Use `tsd` for Testing Types

suneettipirneni:

Uses the `tsd` package to run tests on types. This allows more nuanced type checking than `tsc`, and provides clear output when types don't match. ~~Not all of the `assertType`'s have been replaced, I've only done a few to demo how it works.~~


typings/index.test-d.ts
```diff
@@ -0,0 +1,1056 @@
...
+assertType<null>(loggedOutClient.readyAt);
+assertType<null>(loggedOutClient.readyTimestamp);
+assertType<string | null>(loggedOutClient.token);
+assertType<null>(loggedOutClient.uptime);
+assertType<null>(loggedOutClient.user);
+
+// Test type transformation:
+declare const assertType: <T>(value: T) => asserts value is T;
+declare const serialize: <T>(value: T) => Serialized<T>;
+declare const notPropertyOf: <T, P extends PropertyKey>(value: T, property: P & Exclude<P, keyof T>) => void;
```
vladfrangu, line 707: These should be moved at the top, since they're used above their definition too

.eslintignore
```diff
@@ -0,0 +1 @@
+/typings/**/*
```
GodderE2D, line 1:
```suggestion
/typings/**/*

```

package.json
```diff
@@ -4,7 +4,7 @@
   "description": "A powerful library for interacting with the Discord API",
   "scripts": {
     "test": "npm run lint && npm run docs:test && npm run lint:typings && npm run test:typescript",
-    "test:typescript": "tsc --noEmit",
+    "test:typescript": "npx tsd",
```
iCrawl, line 7:
```suggestion
    "test:typescript": "tsc --noEmit && tsd",
```

At this point we might as well just do this and don't change the workflow at all.

kyranet:
CI is failing.

Also, isn't the new file very similar to our `index.ts` file? Can't we re-use it somehow?

suneettipirneni:
The index.ts file is the same file, it's just showing as a new file in the diff.

iCrawl (approved):

iCrawl:
I'll just move forward to merge this, since it is a better alternative to maintaining some util functions ourselves and have proper testing this way.

PR #6984, opened 2021-11-14
Update README.md

IceXPR:
Renamed "token" to be BOT_TOKEN more clear to new developers.

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,7 +59,7 @@ const commands = [{
   description: 'Replies with Pong!'
 }]; 
 
-const rest = new REST({ version: '9' }).setToken('token');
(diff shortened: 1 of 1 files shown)


GodderE2D (changes requested):
`BOT_TOKEN` is not defined in that code sample, so if one copied & pasted the except, it will throw a ReferenceError. I suggest renaming it to `'bot token'`, if at all. I don't think this is a necessary change.

DTrombett:
If it's really necessary, I suggest to replace `'token'` with something like `'put your token here'`.

IceXPR:
I used caps in the update as this is consistent with the other user required values.  It is also good practice to have them in caps as this should be an ENV_VARIABLE.

PR #6994, opened 2021-11-16
chore: Emit deprecation messages

Jiralite:
This pull request emits deprecation warnings to deprecated parts of the library.

This was not done for `MessageEmbed#type` as that was just a property (though it could technically be made into a getter?) and the application events that don't emit at all.

### DIFF
--- a/src/structures/Guild.js
+++ b/src/structures/Guild.js
@@ -34,6 +34,9 @@ const DataResolver = require('../util/DataResolver');
 const SystemChannelFlags = require('../util/SystemChannelFlags');
 const Util = require('../util/Util');
 
+let deprecationEmittedForSetChannelPositions = false;
+let deprecationEmittedForSetRolePositions = false;
+
 /**
  * Represents a guild (or a server) on Discord.
  * <info>It's recommended to see if a guild is available before performing operations or reading data from it. You can
@@ -1195,6 +1198,15 @@ class Guild extends AnonymousGuild {
    *   .catch(console.error);
    */
   setChannelPositions(channelPositions) {
+    if (!deprecationEmittedForSetChannelPositions) {
+      process.emitWarning(
+        'This method is deprecated. Use GuildChannelManager#setPositions instead.',
+        'DeprecationWarning',
+      );
+
+      deprecationEmittedForSetChannelPositions = true;
+    }
+
     return this.channels.setPositions(channelPositions);
   }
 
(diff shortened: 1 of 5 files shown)


vladfrangu (approved):

src/util/Util.js
```diff
@@ -578,6 +580,15 @@ class Util extends null {
    * @deprecated Use {@link BaseMessageOptions#allowedMentions} instead.
    */
   static removeMentions(str) {
+    if (!deprecationEmittedForRemoveMentions) {
+      process.emitWarning(
+        'This method is deprecated. Utilise allowed mentions when sending a message instead.',
```
ImRodry, line 585: Why doesn't this follow the same pattern as the others?
```suggestion
        'This method is deprecated. Use MessageOptions#allowedMentions when sending a message instead.',
```

ImRodry (changes requested):

kyranet (approved):
I'm guessing that the typings already have `@deprecated`, LGTM.

src/structures/Guild.js
```diff
@@ -1195,6 +1198,15 @@ class Guild extends AnonymousGuild {
    *   .catch(console.error);
    */
   setChannelPositions(channelPositions) {
+    if (!deprecationEmittedForSetChannelPositions) {
+      process.emitWarning(
+        'This method (Guild#setChannelPositions) is deprecated. Use GuildChannelManager#setPositions instead.',
```
ImRodry, line 1203: Not gonna suggest this for all as that'd be too spammy but what do you think of this format?
```suggestion
        'This Guild#setChannelPositions method is deprecated. Use GuildChannelManager#setPositions instead.',
```

almeidx: > The Guild#setChannelPositions method is deprecated

ImRodry: Fixed

suneettipirneni:
Another message can be added as a result of #6889

src/structures/Guild.js
Jiralite: Sure, I can do that!

advaith1:
can permission functions warn on usage of the USE_*_THREADS permissions?

Jiralite:
> can permission functions warn on usage of the USE_*_THREADS permissions?

I can for sure check them in `Permissions#any`, `Permissions#has` & `Permissions#missing` but I'm not sure if I can for the rest of the methods. They're implemented from the parent class of which conflicting bits may be a possibility from other extended classes... but it's a start?

PR #6997, opened 2021-11-17
chore(ThreadMemberManager): make `#fetch` only take one object parameter

suneettipirneni:

@Jiralite pointed out this one out

### DIFF
--- a/packages/discord.js/src/managers/ThreadManager.js
+++ b/packages/discord.js/src/managers/ThreadManager.js
@@ -69,6 +69,7 @@ class ThreadManager extends CachedManager {
    * `GUILD_NEWS_THREAD`</warn>
    * @property {boolean} [invitable] Whether non-moderators can add other non-moderators to the thread
    * <info>Can only be set when type will be `GUILD_PRIVATE_THREAD`</info>
+   * @property {number} [rateLimitPerUser] The rate limit per user (slowmode) for the new channel in seconds
    */
 
   /**
@@ -104,6 +105,7 @@ class ThreadManager extends CachedManager {
     type,
     invitable,
     reason,
+    rateLimitPerUser,
   } = {}) {
     let path = this.client.api.channels(this.channel.id);
     if (type && typeof type !== 'string' && typeof type !== 'number') {
@@ -133,6 +135,7 @@ class ThreadManager extends CachedManager {
         auto_archive_duration: autoArchiveDuration,
         type: resolvedType,
         invitable: resolvedType === ChannelTypes.GUILD_PRIVATE_THREAD ? invitable : undefined,
+        rate_limit_per_user: rateLimitPerUser,
       },
       reason,
     });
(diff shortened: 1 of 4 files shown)


typings/index.d.ts
```diff
@@ -5091,6 +5095,7 @@ export interface ThreadCreateOptions<AllowedThreadType> extends StartThreadOptio
   startMessage?: MessageResolvable;
   type?: AllowedThreadType;
   invitable?: AllowedThreadType extends 'GUILD_PRIVATE_THREAD' | 12 ? boolean : never;
+  rateLimitPerUser?: number;
```
ImRodry, line 5098: Once again xD

ImRodry (commented):
Personally, I don’t agree with this. Despite the member parameter being optional, the options object is only ever used if a member is provided, so it wouldn’t make sense to do fetch({force:true}) or even fetch({cache:true}) as that can simply be replaced by fetch(true). Either way, just my two cents

suneettipirneni:
> Personally, I don’t agree with this. Despite the member parameter being optional, the options object is only ever used if a member is provided, so it wouldn’t make sense to do fetch({force:true}) or even fetch({cache:true}) as that can simply be replaced by fetch(true). Either way, just my two cents

Yeah I'm starting to agree with this, I think I'll just remove the deprecation on the boolean overload.

iCrawl:

SpaceEEC:
Isn't this breaking? The second parameter exists in stable (13.3.1)
https://github.com/discordjs/discord.js/blob/13.3.1/src/managers/ThreadMemberManager.js#L110-L121

suneettipirneni:
> Isn't this breaking? The second parameter exists in stable (13.3.1)
> 
> https://github.com/discordjs/discord.js/blob/13.3.1/src/managers/ThreadMemberManager.js#L110-L121

Welp, semver: major it is

vladfrangu (changes requested):
Code doesn't align with implementation

packages/discord.js/src/managers/ThreadManager.js
```diff
@@ -105,7 +104,6 @@ class ThreadManager extends CachedManager {
     type,
     invitable,
     reason,
-    rateLimitPerUser,
```
vladfrangu, line 108: Why was this removed?

vladfrangu (changes requested):

packages/discord.js/src/managers/ThreadManager.js
```diff
@@ -69,7 +69,6 @@ class ThreadManager extends CachedManager {
    * `GUILD_NEWS_THREAD`</warn>
    * @property {boolean} [invitable] Whether non-moderators can add other non-moderators to the thread
    * <info>Can only be set when type will be `GUILD_PRIVATE_THREAD`</info>
-   * @property {number} [rateLimitPerUser] The rate limit per user (slowmode) for the new channel in seconds
```
almeidx, line 72: This should probably be added back too

vladfrangu (approved):

iCrawl (approved):

SpaceEEC (approved):

PR #6998, opened 2021-11-17
fix: Interaction channel type should be `GuildTextBasedChannels` when in guild

suneettipirneni:


iCrawl (approved):

typings/index.d.ts
```diff
@@ -1190,7 +1191,7 @@ export class Interaction<Cached extends CacheType = CacheType> extends Base {
   private readonly _cacheType: Cached;
   protected constructor(client: Client, data: RawInteractionData);
   public applicationId: Snowflake;
-  public readonly channel: CacheTypeReducer<Cached, TextBasedChannels | null>;
+  public readonly channel: CacheTypeReducer<Cached, GuildTextBasedChannel, APIInteractionDataResolvedChannel>;
```
vladfrangu, line 1194: Are you sure the raw type is correct? I also see you import APIChannel but not use it

Also won't this technically break actual DM interactions? Or is `cached` supposed to only affect GuildChannels?

vladfrangu (changes requested):

typings/index.d.ts
suneettipirneni: I just pushed some changes, I think they look ok but you can double check.

iCrawl:

kyranet (approved):

SpaceEEC (approved):

PR #7002, opened 2021-11-18
fix(Util): fix sorting for GuildChannels

That-Guy977:

`Util#discordSort` sorts Ids descending when the `rawPosition` is the same, which is consistent with `Role` sorting, but is not for `GuildChannel` sorting, which sorts Ids ascending

- Typings don't need updating

~~Note: I've been unable to test the changes nor some similar versions, due to issues as I mentioned [here](https://discord.com/channels/222078108977594368/682166281826598932/909663555312291920); I'm not sure what I'm doing wrong, as I've only modified the required files and this method, but it seems to break the entire `Util` exports
I've just decided to create this PR in the hopes that someone else with more experience can fix these mistakes~~
Issues resolved, changes tested

### DIFF
--- a/src/util/Util.js
+++ b/src/util/Util.js
@@ -483,9 +483,10 @@ class Util extends null {
    */
   static discordSort(collection) {
     return collection.sorted((a, b) => {
+      if (a.rawPosition !== b.rawPosition) return a.rawPosition - b.rawPosition;
       const aId = BigInt(a.id);
       const bId = BigInt(b.id);
-      return a.rawPosition - b.rawPosition || Number(a instanceof GuildChannel ? aId - bId : bId - aId);
+      return Number(a instanceof GuildChannel ? aId - bId : bId - aId);
     });
   }
 
--- a/src/util/Util.js
+++ b/src/util/Util.js
@@ -482,12 +482,12 @@ class Util extends null {
    * @returns {Collection}
    */
   static discordSort(collection) {
-    return collection.sorted((a, b) => {
-      if (a.rawPosition !== b.rawPosition) return a.rawPosition - b.rawPosition;
-      const aId = BigInt(a.id);
-      const bId = BigInt(b.id);
-      return Number(a instanceof GuildChannel ? aId - bId : bId - aId);
-    });
+    const isGuildChannel = collection.first() instanceof GuildChannel;
+    return collection.sorted(
+      isGuildChannel
+        ? (a, b) => a.rawPosition - b.rawPosition || Number(BigInt(a.id) - BigInt(b.id))
+        : (a, b) => a.rawPosition - b.rawPosition || Number(BigInt(b.id) - BigInt(a.id)),
+    );
   }
 
   /**
--- a/src/util/Util.js
+++ b/src/util/Util.js
@@ -6,6 +6,7 @@ const fetch = require('node-fetch');
 const { Colors, Endpoints } = require('./Constants');
 const Options = require('./Options');
 const { Error: DiscordError, RangeError, TypeError } = require('../errors');
+const GuildChannel = require('../structures/GuildChannel');
 const has = (o, k) => Object.prototype.hasOwnProperty.call(o, k);
 const isObject = d => typeof d === 'object' && d !== null;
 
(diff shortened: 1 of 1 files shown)


src/util/Util.js
```diff
@@ -481,12 +482,11 @@ class Util extends null {
...
-    return collection.sorted(
-      (a, b) =>
-        a.rawPosition - b.rawPosition ||
-        parseInt(b.id.slice(0, -10)) - parseInt(a.id.slice(0, -10)) ||
-        parseInt(b.id.slice(10)) - parseInt(a.id.slice(10)),
-    );
+    return collection.sorted((a, b) => {
+      const aId = BigInt(a.id);
+      const bId = BigInt(b.id);
+      return a.rawPosition - b.rawPosition || Number(a instanceof GuildChannel ? aId - bId : bId - aId);
```
kyranet, line 488: We should avoid converting to BigInt unless both `rawPosition`s are the same, otherwise it'll cause some unneeded overhead.

kyranet (changes requested):

kyranet:
Come think of it, can't we read the first value in the collection and determine whether or not it's a collection of GuildChannels? That way we do 1 instanceof check rather than `N * log(N)` times (being `N` the collection's size), which can be quite expensive as `instanceof` is an `O(N)` operation (being `N` the amount of prototypes it takes to get to null), resulting on an `O(N^2)` cost?

I'm thinking more along the lines of:
```javascript
// Nothing to sort if it has 0 or 1 entry:
if (collection.size < 2) return collection.clone();

// Check whether or not the first value is a GuildChannel, if it is, assume the rest
// of entries also are:
const isGuildChannel = collection.first() instanceof GuildChannel;
return collection.sorted(isGuildChannel ?
	(a, b) => a.rawPosition - b.rawPosition || Number(BigInt(a.id) - BigInt(b.id)) :
	(a, b) => a.rawPosition - b.rawPosition || Number(BigInt(b.id) - BigInt(a.id)));
```

This way it keeps being `O(N * log N)` while also fixes the bug.

That-Guy977:
Would it really be necessary to add a case specifically for 0 or 1 elements in the collection? Wouldn't the internal use of Array#sort just not do much at all?

kyranet:
That's because with this trick, we're reading the first element, so we need to make sure the collection is not empty (`if (collection.size === 0) return collection.clone();`), however, since we cannot sort a collection with one entry (it's always sorted), then we change the condition from `=== 0` to `< 2`, that's why the suggested code is like that.

That-Guy977:
But if there aren't elements, the sort function isn't run anyways, so I don't understand *why* we would need to make sure that it isn't empty
It'd just be `undefined instanceof GuildChannel`, no? and that would be false but it wouldn't really matter

kyranet:
Fair enough, feel free to remove that case.

That-Guy977:
This is what i got to after a few style iterations
```js
static discordSort(collection) {
  const isGuildChannel = collection.first() instanceof GuildChannel;
  return collection.sorted(
    (a, b) =>
      a.rawPosition - b.rawPosition ||
      Number(isGuildChannel ? BigInt(a.id) - BigInt(b.id) : BigInt(b.id) - BigInt(a.id)),
  );
}
```
Does this look good to you?

kyranet:
You're back to making a check `N * log(N)` times.

That-Guy977:
Ah, i missed the purpose of the seperate functions, I'll just go with your method minus the size check then

kyranet (approved):
As long as this is tested, looks good to me.

That-Guy977:
The issues i mentioned in the initial comment seem to have fixed themselves, so I was able to test it, and the position is consistent with the UI for both channels and roles with this version
However, I'm now getting these warnings/errors from sending the message output, not sure if this is related/significant? I tested this in the node repl
```sh-session
(node:24339) Warning: Accessing non-existent property 'idToBinary' of module exports inside circular dependency
Uncaught TypeError: Util.idToBinary is not a function
    at Function.deconstruct (/Users/that_guy977/Desktop/Stuff/Github/test/node_modules/discord.js/src/util/SnowflakeUtil.js:63:25)
    at Message._patch (/Users/that_guy977/Desktop/Stuff/Github/test/node_modules/discord.js/src/structures/Message.js:62:43)
    at new Message (/Users/that_guy977/Desktop/Stuff/Github/test/node_modules/discord.js/src/structures/Message.js:48:10)
    at MessageManager._add (/Users/that_guy977/Desktop/Stuff/Github/test/node_modules/discord.js/src/managers/CachedManager.js:58:32)
    at MessageManager._add (/Users/that_guy977/Desktop/Stuff/Github/test/node_modules/discord.js/src/managers/MessageManager.js:31:18)
    at TextChannel.send (/Users/that_guy977/Desktop/Stuff/Github/test/node_modules/discord.js/src/structures/interfaces/TextBasedChannel.js:177:59)
```
Seems to occur with the other Util methods as well

SpaceEEC (approved):

iCrawl (approved):

PR #7013, opened 2021-11-20
types: allow `MessageActionRow` to specify a generic component type

suneettipirneni:

This strengthen types for `MessageActionRow` by enforcing that elements within its component array must be homogeneous. I.e. You can't add a button to a row with a select menu. In addition `#components` will give a more specific type rather than just `MessageComponentTypeResolvable`.


typings/index.d.ts
```diff
@@ -1418,21 +1418,15 @@ export class Message<Cached extends boolean = boolean> extends Base {
   public inGuild(): this is Message<true> & this;
 }
 
-export class MessageActionRow extends BaseMessageComponent {
-  public constructor(data?: MessageActionRow | MessageActionRowOptions | APIActionRowComponent);
+export type MessageActionRowComponentType<T> = T | MessageActionRowComponentResolvable[];
+
+export class MessageActionRow<T = MessageActionRowComponent> extends BaseMessageComponent {
```
ImRodry, line 1423: Sorry I didn't catch this earlier, but shouldn't this T be of type MessageComponentTypeResolvable just like in collectors? For consistency purposes that is

suneettipirneni: No, mainly because the collectors accept the number or string representation of a component in its `options.componentType`. This only needs the `MessageActionRowComponent`, since it's the only type used in the class.

ImRodry: Alright then this works just fine! Thanks

ImRodry (approved):

monbrey:
While technically correct, isn't this not very future proof?

ImRodry:
If your concern is multiple component types being allowed in the same row in the future, that's not gonna be an issue because you can simply pass both of those types and it will work, which is what the default type is doing atm. Other than that I don't see anything that would make this less future proof ?

suneettipirneni:
> While technically correct, isn't this not very future proof? I don't think typings need to enforce something that is just coincidentally true, it's not a hard rule.

So my initial PR description is somewhat misleading. 

This doesn't actually enforce that every element in the component row needs to be homogeneous. It instead allows users to narrow whatever the array type of components down to something more specific as long as it's covariant to `MessageComponentResolable`.

For example let's assume that discord allows buttons and text components to be on the same row.

This code would be used by the developer:
```ts
const row = new MessageActionRow<MessageButton | MessageTextInput>();
```

Without this parameter there isn't a way to narrow this down without a type guard. If you're the one who created the row you should also have information as to what type(s) this row contains. As a result, you should be able to provide a more specific type.

This generic parameter includes being able to specify type combinations that may not be available today but will maybe be in the future.

If you don't care for this kind of narrowing, the type behavior is exactly the same as before if no generic parameter is utilized.

Korabi-dev (approved):

Korabi-dev:
I think this commit should be merged, great idea.

Korabi-dev:
My bad, I didnt mean commit. Rather meant PR.

iCrawl:
This needs a rebase.

vladfrangu (approved):

suneettipirneni:
> This needs a rebase.

done

iCrawl:

kyranet (approved):

suneettipirneni:
Superseded by #7040

ImRodry:
Since #7040 is semver major, could this be reopened please?

PR #7019, opened 2021-11-22
types: make event emitters use mapped event types

suneettipirneni:

Uses mapped types instead of manually declaring them twice within each class that extends `EventEmitter`. Similar to what `Client` already does for its events. 

### DIFF
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -609,6 +609,12 @@ export class ClientVoiceManager {
 
 export { Collection } from '@discordjs/collection';
 
+export interface CollectorEventTypes<K, V, F extends unknown[] = []> {
+  collect: [V, ...F];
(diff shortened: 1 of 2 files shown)


typings/index.d.ts
```diff
@@ -609,6 +609,12 @@ export class ClientVoiceManager {
 
 export { Collection } from '@discordjs/collection';
 
+export interface CollectorEventTypes<K, V, F extends unknown[] = []> {
+  collect: [V, ...F];
+  dispose: [V, ...F];
+  end: [collected: Collection<K, V>];
```
ImRodry, line 615: Where’d the reason go?

SpaceEEC (approved):

iCrawl (approved):

PR #7026, opened 2021-11-24
Handle Memory Leak

Nigelrex:
Handle Memory Leak that closes the node process if the process has hit the max memory limit

Added a `.catch()` to handle and close the process before the process starts to eat up a ton of memory in a few minutes UP to 2GB

This PR nomore includes breaking changes (methods removed or renamed, parameters moved or removed)

typings don't need updating

https://user-images.githubusercontent.com/75080457/143230670-41746d78-1fa0-4499-b601-e110169cbadc.mp4

### DIFF
--- a/src/rest/APIRequest.js
+++ b/src/rest/APIRequest.js
@@ -22,8 +22,8 @@ class APIRequest {
     let queryString = '';
     if (options.query) {
       const query = Object.entries(options.query)
-        .filter(([, value]) => value !== null && typeof value !== 'undefined')
-        .flatMap(([key, value]) => (Array.isArray(value) ? value.map(v => [key, v]) : [[key, value]]));
+        .filter(([ , value ]) => value !== null && typeof value !== 'undefined')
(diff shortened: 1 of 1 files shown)


Nigelrex:
I didn't wnat to do the full testing my here you go, I also made those changes!

Nigelrex:
I didn't want to do the full testing my here you go, I also made those changes!

vladfrangu:
This Pull request:

- reverts changes that were made to be compatible with the newer API features
- stylistic changes that our linter will scream at
- kills the process at any error encountered by node fetch, which is highly undesirable

If you're encountering ram issues while sending files, maybe try using read streams instead, I think we handle those. If we don't, that's what should be implemented instead

iCrawl:
Additionally, to the mentioned things from Vlad, this is a horrible way to measure your RAM usage, the underlying issue is a completely different.

To fix this all you must do is simply make sure you have an active DNS resolvable record

Nigelrex:
B r u h, I sent a video and showed that if there was a connection error your RAM usage by the nodejs process goes up untill it exceeds the limit and then it crashes

PR #6709, opened 2021-09-28
feat(ModuleImporter): implement import module compatible with bundlers

nrauschcom:
**Please describe the changes this PR makes and why it should be merged:**

In this PR, I implemented a ModuleImporter helper class, which currently supports Webpack and File System Imports.

The Util can import all files from a given directory without listing them manually and thus circumvents missed imports.

In an earlier approach [#6102] this issue was solved by doing a file system search, which currently is the only change preventing the use of code bundlers (Webpack, Rollup, etc) in d.js@>13. Read [this comment](https://github.com/discordjs/discord.js/issues/6681#issuecomment-929087260) on why bundling is important for some use cases other than Web Development.

My approach will actually just work for File Sytsem import or Webpack, because I'm not very familiar with other bundlers and the implementation seemed not so straightforward. Potentially Rollup works out-of-the-box with [rollup-plugin-require-context](https://www.npmjs.com/package/rollup-plugin-require-context).

I tested both file system and WebPack in my projects.

I'm open for discussion and feedback.

Close #6681

**Status and versioning classification:**

- Code changes have been tested against the Discord API, or there are no code changes
- I know how to update typings and have done so, or typings don't need updating

PR #7034, opened 2021-11-25
fix(ActionsManager): revert to manual requires

### LINKED ISSUE #7032: Bug: Cannot find module when running with esbuild
### Issue description

I'm using esbuild to bundle my code, along with a dependency on discord.js, before deploying it to AWS Lambda.

https://esbuild.github.io/

Read [this comment](https://github.com/discordjs/discord.js/issues/6681#issuecomment-929087260) on why bundling is important for some use cases other than Web Development. (thanks @nrauschcom)

It fails due to the imports in `src/client/websocket/handlers/index.js` which are generated from `Constants.WSEvents`. If that is fixed or commented out, then it fails due to the imports in `ActionsManager.js` which are generated from a directory listing.

## How to reproduce

1. Create package.json file:
```
{
  "dependencies": {
    "discord.js": "13.3.1"
  },
  "devDependencies": {
    "esbuild": "0.13.14"
  }
}
```

2. Create index.js file:
```
const { Client, Intents } = require("discord.js");
new Client({ intents: [Intents.FLAGS.GUILDS] });
```

3. Bundle index.js with esbuild and then run with it node:
```
$ npm install
$ ./node_modules/.bin/esbuild index.js --bundle --platform=node --target=node16 | node
```

4. node fails with "Error: Cannot find module" thrown from `src/client/websocket/handlers/index.js`:
```
node:internal/modules/cjs/loader:936
  throw err;
  ^

Error: Cannot find module './READY.js'
Require stack:
- /Users/stevengbrown/Documents/discordjstest/[stdin]
    at Function.Module._resolveFilename (node:internal/modules/cjs/loader:933:15)
    at Function.Module._load (node:internal/modules/cjs/loader:778:27)
    at Module.require (node:internal/modules/cjs/loader:1005:19)
    at require (node:internal/modules/cjs/helpers:102:18)
    at node_modules/discord.js/src/client/websocket/handlers/index.js ([stdin]:19177:24)
    at __require ([stdin]:21:44)
    at node_modules/discord.js/src/client/websocket/WebSocketManager.js ([stdin]:19191:26)
    at __require ([stdin]:21:44)
    at node_modules/discord.js/src/client/Client.js ([stdin]:28474:28)
    at __require ([stdin]:21:44) {
  code: 'MODULE_NOT_FOUND',
  requireStack: [ '/Users/stevengbrown/Documents/discordjstest/[stdin]' ]
}
```

## Workaround

discord.js can be marked as external when using esbuild which means it isn't included in the bundle and discord.js along with all of its direct and transitive dependencies must be copied across from the node_modules directory instead. It is possible to have a script do this, but of course it would nicer if discord.js worked out of the box.

https://esbuild.github.io/api/#non-analyzable-imports

## History

Related to issues #6680 and #6681.

PR #6102 changed `ActionsManager.js` to generate the imports from a directory listing, unintentionally breaking bundlers (it was intended as a refactor).

PR #6682 proposed to revert `ActionsManager.js` to use manual imports, but was closed without merging due to maintainability concerns.

PR #6709 fixed this issue for Webpack by making use of `require.context`, but not for other bundlers such as esbuild.

## Possible solutions

esbuild requires the import paths to be string literals. This would make discord.js compatible with both esbuild and Webpack.

Either:

1. Update `src/client/websocket/handlers/index.js` and `ActionsManager.js` to use manual imports.

OR

2. Have a script automatically generate those imports from the directory listings during commit, npm install, npm test and npm publish.

With either solution, a check would be added to make sure that every entry in `Constants.WSEvents` has an associated event handler.

I'm happy to prepare a PR.

### Code sample

```typescript
const { Client, Intents } = require("discord.js");
new Client({ intents: [Intents.FLAGS.GUILDS] });
```

### discord.js version

13.3.1

### Node.js version

16.13.0

### Operating system

macOS Big Sur

Medium (should be fixed soon)

No Partials

GUILDS

802a5ac

kyranet:

Reverts #6102 and #6709, the list has been generated with the ~~`ls` command~~ source generator script included in this Pull Request to ensure it's up to date.

The first Pull Request proposed this change to reduce the complexity, however, what once was simple turned out to be a complete nightmare as issues of all kinds started to show up. Just to clarify, it's a bad practice for a library to be doing FS operations on its own.

~~Following up with the concerns in #6682, if we're worried about that, I'm more than willing to write an ESM script that scans for the files and generates a valid JS file, it wouldn't be the first time I generate code with a script.~~ Following #5542 and taking bundler support further, I decided to make a script to generate the requires with a single command.

Fixes #7032

**Note**: The `.eslintrc.json` in `scripts` was written to solve an issue with different source types, since the scripts folder uses `module`.


scripts/generateRequires.mjs
```diff
@@ -0,0 +1,41 @@
...
+  const outputFile = new URL('../src/client/websocket/handlers/index.js', import.meta.url);
+
+  await writeFile(outputFile, lines.join('\n'));
+}
+
+async function writeClientActionImports() {
+  const lines = ["'use strict';\n", 'class ActionsManager {', '  constructor(client) {', '    this.client = client;\n'];
+
+  const actionsDirectory = new URL('../src/client/actions', import.meta.url);
+  for (const file of await readdir(actionsDirectory)) {
```
StevenGBrown, line 22: I don't think `readdir` gives any guarantees about the order, I suggest sticking a `.sort()` in there to make sure that everyone gets the same result

StevenGBrown (approved):
Thanks @kyranet, have confirmed that it works

scripts/generateRequires.mjs
kyranet: You're right, addressed in d6c13168.

SpaceEEC (approved):

iCrawl (approved):

alula:
:^)

PR #7023, opened 2021-11-22
feat: Add Modals and Text Inputs

suneettipirneni:
**Please describe the changes this PR makes and why it should be merged:**

Adds support for modal submit interactions as well as text input components.

Prerequisites:
- https://github.com/discordjs/discord-api-types/pull/243
- https://github.com/discordjs/discord.js/pull/7422
- https://github.com/discordjs/discord.js/pull/7424
- https://github.com/discordjs/discord.js/pull/7464

Ref:
- https://github.com/discord/discord-api-docs/pull/4459

#### Sending
```ts
import {
  ActionRow,
  TextInputComponent,
  TextInputStyle,
  Modal,
  type ModalActionRowComponent,
} from 'discord.js';

// Create the modal
const modal = new Modal()
  .setTitle('My Awesome Form')
  .setCustomId('AwesomeForm');

// Create text input fields
const tvShowInputComponent = new TextInputComponent()
  .setCustomId('tvField')
  .setLabel('Favorite TV show')
  .setStyle(TextInputStyle.Short);

const haikuInputComponent = new TextInputComponent()
  .setCustomId('haikuField')
  .setLabel('Write down your favorite haiku')
  .setStyle(TextInputStyle.Paragraph);

const rows = [tvShowInputComponent, haikuInputComponent].map(
  (component) =>
    new ActionRow<ModalActionRowComponent>().addComponents(component)
);

// Add action rows to form
modal.addComponents(...rows);

// --- snip ---

// Present the modal to the user
await interaction.presentModal(modal);
```

#### Receiving
```ts
client.on('interactionCreate', async (interaction) => {
  if (!interaction.isModalSubmit()) {
    return;
  }
  
  // Verify we have the right modal
  if (interaction.customId !== 'AwesomeForm') {
    return;
  }

  // Extract text input components from submitted modal
  const [favTVShow, favHaiku] = ['tvField', 'haikuField']
    .map(id => interaction.fields.getTextInputValue(id));
  
  // ACK the interaction
  await interaction.reply(`Favorite TV Show: ${favTVShow}, Favorite haiku: ${favHaiku}`);
});
```

### TODO:

- [x] Add the rest of the typings
- [x] Implement the proper type guards
- [x] Refine and finish doc types and descriptions
- [x] Fix any discrepancies with data deserialization of components.
- [x] Throw API errors in unsupported configurations
- [x] Test

**Status and versioning classification:**
- Code changes have been tested against the Discord API, or there are no code changes
[...]

PR #7040, opened 2021-11-26
Refactor: Make message components agnostic components

suneettipirneni:

~~This deprecates all other `Message*` types, so this shouldn't be breaking.~~

It's breaking
 
This makes all components regardless of being a message component, or a modal component all inherit from the same base, and have no difference in variance between message and modal implementations. This is key refactor needed for the modal PR (#7023), and prevents future class proliferation of other component type variants.


src/ComponentFactory.js
```diff
@@ -0,0 +1,49 @@
+'use strict';
+
+const { ComponentTypes, Events } = require('./util/Constants');
+
+/**
+ * A factory for creating and resolving components.
+ */
+class Components {
```
ImRodry, line 8: Class and file names should probably match

src/structures/ActionRow.js
```diff
@@ -0,0 +1,104 @@
...
+   * * ActionRowComponent
+   * * ActionRowComponentOptions
+   * @typedef {ActionRowComponent|ActionRowComponentOptions} ActionRowComponentResolvable
+   */
+
+  /**
+   * @typedef {BaseComponentOptions} ActionRowOptions
+   * @property {ActionRowComponentResolvable[]} [components]
+   * The components to place in this action row
+   */
```
ImRodry, line 39: Please put these in places where they’re relevant so it’s easier to find and edit them later on (please do the same for all files)

src/structures/ActionRow.js
```diff
@@ -0,0 +1,104 @@
...
+    this.components = data.components?.map(c => Components.create(c, client)) ?? [];
+  }
+
+  /**
+   * Adds components to the action row.
+   * @param {...ActionRowComponentResolvable[]} components The components to add
+   * @returns {ActionRow}
+   */
+  addComponents(...components) {
+    this.components.push(...components.flat(Infinity).map(c => Components.create(c)));
```
ImRodry, line 61:
```suggestion
    this.components.push(...components.flat(Infinity).map(c => Components.create(c, client)));
```

src/structures/ActionRow.js
```diff
@@ -0,0 +1,104 @@
...
+
+  /**
+   * Removes, replaces, and inserts components in the action row.
+   * @param {number} index The index to start at
+   * @param {number} deleteCount The number of components to remove
+   * @param {...ActionRowComponentResolvable[]} [components] The replacing components
+   * @returns {ActionRow}
+   */
+  spliceComponents(index, deleteCount, ...components) {
+    this.components.splice(index, deleteCount, ...components.flat(Infinity).map(c => Components.create(c)));
```
ImRodry, line 83:
```suggestion
    this.components.splice(index, deleteCount, ...components.flat(Infinity).map(c => Components.create(c, create)));
```

src/structures/BaseComponent.js
```diff
@@ -0,0 +1,57 @@
...
+   * * string
+   * * number
+   * @typedef {string|number|ComponentType} ComponentTypeResolvable
+   */
+
+  /**
+   * @param {BaseComponent|BaseComponentOptions} data The options for this component
+   */
+  constructor(data = {}) {
+    this.type = 'type' in data ? BaseComponent.resolveType(data.type) : null;
```
ImRodry, line 43: Why can this be null and why isn’t it documented?

src/structures/BaseMessageComponent.js
```diff
@@ -43,6 +46,15 @@ class BaseMessageComponent {
    * @param {BaseMessageComponent|BaseMessageComponentOptions} [data={}] The options for this component
    */
   constructor(data) {
+    if (!emittedDeprecationNotice) {
+      process.emitWarning(
+        'BaseMessageComponent is deprecated and will be removed in a future release, please use BaseComponent instead',
```
ImRodry, line 51: Matches the ones from #6994 
```suggestion
        'The BaseMessageComponent class is deprecated. Use BaseComponent instead',
```

src/structures/ButtonComponent.js
```diff
@@ -0,0 +1,160 @@
...
+     * A unique string to be sent in the interaction when clicked
+     * @type {?string}
+     */
+    this.customId = data.custom_id ?? data.customId ?? null;
+
+    /**
+     * The style of this button
+     * @type {?ButtonStyle}
+     */
+    this.style = data.style ? ButtonComponent.resolveStyle(data.style) : null;
```
ImRodry, line 44: How can this be missing?

src/structures/MessageActionRow.js
```diff
@@ -42,6 +45,15 @@ class MessageActionRow extends BaseMessageComponent {
   constructor(data = {}, client = null) {
     super({ type: 'ACTION_ROW' });
 
+    if (!emittedDeprecationNotice) {
+      process.emitWarning(
+        'MessageActionRow is deprecated and will be removed in a future release, please use ActionRow instead',
```
ImRodry, line 50:
```suggestion
        'The MessageActionRow class is deprecated. Use ActionRow instead',
```

src/structures/MessageButton.js
```diff
@@ -26,6 +29,15 @@ class MessageButton extends BaseMessageComponent {
   constructor(data = {}) {
     super({ type: 'BUTTON' });
 
+    if (!emittedDeprecationNotice) {
+      process.emitWarning(
+        'MessageButton is deprecated and will be removed in a future release, please use ButtonComponent instead',
```
ImRodry, line 34:
```suggestion
        'The MessageButton class is deprecated. Use ButtonComponent instead',
```

src/structures/MessageSelectMenu.js
```diff
@@ -43,6 +46,16 @@ class MessageSelectMenu extends BaseMessageComponent {
   constructor(data = {}) {
     super({ type: 'SELECT_MENU' });
 
+    if (!emittedDeprecationNotice) {
+      process.emitWarning(
+        `MessageSelectMenu is deprecated and will be removed in a future release,
+         please use SelectMenuComponent instead`,
```
ImRodry, line 52: Pretty sure this would have a huge gap in the middle, so please update this one to match the others

src/util/Constants.js
```diff
@@ -1019,10 +1019,22 @@ exports.InteractionResponseTypes = createEnum([
  * * BUTTON
  * * SELECT_MENU
  * @typedef {string} MessageComponentType
+ * @deprecated Use @{@link ComponentTypes} instead
```
ImRodry, line 1022: That’s not how you link
```suggestion
 * @deprecated Use {@link ComponentTypes} instead
```

src/util/Constants.js
```diff
@@ -1019,10 +1019,22 @@ exports.InteractionResponseTypes = createEnum([
...
+/**
+ * The type of a message component
+ * * ACTION_ROW
+ * * BUTTON
+ * * SELECT_MENU
+ * * INPUT_TEXT
+ * @typedef {string} ComponentType
+ * @see {@link https://discord.com/developers/docs/interactions/message-components#component-object-component-types}
+ */
+exports.ComponentTypes = createEnum([null, 'ACTION_ROW', 'BUTTON', 'SELECT_MENU', 'INPUT_TEXT']);
```
ImRodry, line 1036: Should INPUT_TEXT really be included in this PR?

typings/index.test-d.ts
```diff
@@ -665,6 +668,7 @@ client.on('interaction', async interaction => {
 
   const actionRow = new MessageActionRow({ components: [button] });
 
+  // @ts-expect-error
```
ImRodry, line 671: What error are you expecting here?

typings/index.test-d.ts
```diff
@@ -1061,4 +1065,12 @@ collector.on('end', (collection, reason) => {
   expectType<string>(reason);
 });
 
+const row = new ActionRow();
+row.addComponents(new ButtonComponent());
+row.addComponents(new SelectMenuComponent());
```
ImRodry, line 1070: This can be moved into one call, not that it matters a lot tho

ImRodry (changes requested):

src/structures/ActionRow.js
kyranet: Where does client come from, Rodry?

src/structures/ActionRow.js
monbrey: How is it not relevant here in this file?

src/structures/ActionRow.js
vladfrangu: The constructor parameter, kyra.

vladfrangu: Actually, no, client shouldn't even be passed down here. It will make the error be emitted as a debug message instead of thrown

src/structures/ButtonComponent.js
vladfrangu: And everything else too? `new ButtonComponent().setStyle()` thats how.

src/structures/SelectMenuComponent.js
```diff
@@ -0,0 +1,205 @@
...
+   * Transforms the select menu into a plain object
+   * @returns {APIMessageSelectMenu} The raw data of this select menu
+   */
+  toJSON() {
+    return {
+      custom_id: this.customId,
+      disabled: this.disabled,
+      placeholder: this.placeholder,
+      min_values: this.minValues,
+      max_values: this.maxValues ?? (this.minValues ? this.options.length : undefined),
```
vladfrangu, line 172: Is the defaulting to this.options.length needed? I don't recall.. 👀

typings/index.d.ts
```diff
@@ -4582,14 +4677,26 @@ export type MessageChannelComponentCollectorOptions<T extends MessageComponentIn
...
   | MessageButtonOptions
   | MessageSelectMenuOptions;
 
+export type ComponentOptions =
+  | BaseComponentOptions
+  | ActionRowOptions
+  | ButtonComponentOptions
+  | SelectMenuComponentOptions;
+
+// export type ComponentOptions = BaseComponentOptions;
```
vladfrangu, line 4693: 👀

vladfrangu (changes requested):

src/structures/ActionRow.js
ImRodry: To be honest this file is good but some of the others have all of the typedefs at the very top where they could be scattered through the file as they are used

src/structures/ButtonComponent.js
ImRodry: Yeah forgot about the constructor, I thought this was the only property that would be always set

src/util/Constants.js
```diff
@@ -1110,7 +1135,9 @@ function createEnum(keys) {
  * @property {InteractionType} InteractionTypes The type of an {@link Interaction} object.
  * @property {MembershipState} MembershipStates The value set for a team member's membership state.
  * @property {MessageButtonStyle} MessageButtonStyles The style of a message button.
+ * @property {ButtonStyle} MButtonStyles The style of a button.
```
ImRodry, line 1138:
```suggestion
 * @property {ButtonStyle} ButtonStyles The style of a button.
```

src/structures/ActionRow.js
suneettipirneni: This is the same format being used beforehand

ImRodry: Alright then

src/structures/SelectMenuComponent.js
suneettipirneni: I'm assuming so, since it was being done in `MessageSelectMenu`

typings/index.test-d.ts
suneettipirneni: The error is you can't use the old `Message*` classes as components.

ImRodry: I thought that was only deprecated though. Wouldn’t that make this semver major?

suneettipirneni: No because this error only affects typescript, in this specific method it's hard to mark the method type as deprecated. But using the old classes here still work as expected.

ImRodry: Can’t you simply deprecate the class/constructor, emit the deprecation warning and keep it working without the error? I know typescript breaking changes don’t really make things semver major but it would be nice if the error was avoided imo

src/index.js
```diff
@@ -65,18 +65,21 @@ exports.WebSocketManager = require('./client/websocket/WebSocketManager');
...
 // Structures
+exports.ActionRow = require('./structures/ActionRow');
 exports.Activity = require('./structures/Presence').Activity;
 exports.AnonymousGuild = require('./structures/AnonymousGuild');
 exports.Application = require('./structures/interfaces/Application');
 exports.ApplicationCommand = require('./structures/ApplicationCommand');
 exports.AutocompleteInteraction = require('./structures/AutocompleteInteraction');
 exports.Base = require('./structures/Base');
+exports.BaseComponent = require('./structures/BaseComponent');
 exports.BaseCommandInteraction = require('./structures/BaseCommandInteraction');
```
ImRodry, line 76: P comes after M 😅 
```suggestion
exports.BaseCommandInteraction = require('./structures/BaseCommandInteraction');
exports.BaseComponent = require('./structures/BaseComponent');
```

src/structures/ActionRow.js
```diff
@@ -0,0 +1,104 @@
...
+      components: this.components.map(c => c.toJSON()),
+      type: ComponentTypes[this.type],
+    };
+  }
+}
+
+module.exports = ActionRow;
+
+/**
+ * @external APIMessageComponent
```
ImRodry, line 102: Should probably remove this from MessageActionRow.js

suneettipirneni:
Just realized that this is going to need to be semver: major, because it changes the type of `Message#components`.

ImRodry:
I don’t think there’s a use case where that would break though is there?

suneettipirneni:
> I don’t think there’s a use case where that would break though is there?

The only place I can think of is using `instanceof` checks. Although doing these types of checks is pointless anyways. But I'll defer this to @discordjs/the-big-4

ImRodry:
Yeah there’s always that but I don’t think anyone would be doing this here. You can also make the old classes extend the new ones and _I think_ the instanceof checks wouldn’t break here, right? This would work since I believe all properties and methods are the same right

suneettipirneni:
If I made the new classes extend the old ones it would defeat the purpose of this PR.

ImRodry:
My suggestion is to make the old ones extend the new ones. The properties are the same, but documented differently (at least how I see it) so it wouldn’t defeat the purpose would it?

suneettipirneni:
> My suggestion is to make the old ones extend the new ones. The properties are the same, but documented differently (at least how I see it) so it wouldn’t defeat the purpose would it?

Perhaps, but then you might run into the issue where `instanceof BaseMessageComponent` won't work. 🤔

ImRodry:
I don't think anyone in their right mind would do such a thing but it you're right :/
Can we ignore that small nit?

src/structures/ActionRow.js
```diff
@@ -0,0 +1,104 @@
...
+  }
+
+  /**
+   * Sets the components of the action row.
+   * @param {...ActionRowComponentResolvable[]} components The components to set
+   * @returns {ActionRow}
+   */
+  setComponents(...components) {
+    this.spliceComponents(0, this.components.length, components);
+    return this;
```
vladfrangu, line 72:
```suggestion
    return this.spliceComponents(0, this.components.length, components);
```

vladfrangu (approved):
Small nit otherwise LGTM to me

typings/index.d.ts
```diff
@@ -340,10 +343,18 @@ export abstract class BaseCommandInteraction<Cached extends CacheType = CacheTyp
   public deferReply(options: InteractionDeferReplyOptions & { fetchReply: true }): Promise<GuildCacheMessage<Cached>>;
   public deferReply(options?: InteractionDeferReplyOptions): Promise<void>;
   public deleteReply(): Promise<void>;
+  /** @deprecated use `editReply(string | MessagePayload | WebhookEditMessageOptions)` instead */
+  public editReply(options: string | MessagePayload | _WebhookEditMessageOptions): Promise<GuildCacheMessage<Cached>>;
   public editReply(options: string | MessagePayload | WebhookEditMessageOptions): Promise<GuildCacheMessage<Cached>>;
```
kyranet, line 348: Question, does this not deprecate `editReply('Hello World')`? Or am I missing something?

suneettipirneni: Good catch, this has been fixed in the latest commit.

kyranet (approved):

iCrawl (approved):

kyranet:
Please rebase, there are merging conflicts.

iCrawl:
This needs a rebease.

suneettipirneni:
So we I'm probably going to close this PR and open another one later that uses the message components from builders instead.

suneettipirneni:
Superceded by #7195

PR #7041, opened 2021-11-26
docs(MessageReference): Fix static link

Jiralite:

First of all, `{@link MessageFlags#FLAGS message flag})` is an invalid redirect as it's a static link. Instead of correcting it, I removed the extra description as it seemed to fit the others. Feel free to request changes to only correct the redirect instead.

### DIFF
--- a/src/structures/Message.js
+++ b/src/structures/Message.js
@@ -293,7 +293,7 @@ class Message extends Base {
     /**
      * Reference data sent in a message that contains ids identifying the referenced message.
      * This can be present in the following types of message:
-     * * IS_CROSSPOST
+     * * Crossposted messages (IS_CROSSPOST {@link MessageFlags.FLAGS message flag})
      * * CHANNEL_FOLLOW_ADD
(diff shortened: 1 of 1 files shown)


ImRodry:
IS_CROSSPOST is not a message type therefore we cannot say this like this. I think the static issue is something that needs to be fixed on the website instead but the link still redirects to the correct page iirc so it’s no big deal and I don’t think this PR is needed

Jiralite:
> IS_CROSSPOST is not a message type therefore we cannot say this like this.

Understood!


>  I think the static issue is something that needs to be fixed on the website instead but the link still redirects to the correct page, just not the correct method

No this is not correct. To reference something that's static, it needs to be joined with `.` To reference something that's from the instance, it needs to be joined with `#`. To reference something that's an event, it needs to be referenced with `event:`.

ImRodry:
Was this tested?

SpaceEEC (approved):

iCrawl (approved):

PR #7049, opened 2021-11-29
feat(Utils): add `camelize()` function

suneettipirneni:

Adds a camel-casing function using [camelcase-keys](https://www.npmjs.com/package/camelcase-keys). This provides a way for users to create a type-compliant camel-cased object.

### DIFF
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -2241,6 +2241,7 @@ export class Util extends null {
     reason?: string,
   ): Promise<{ id: Snowflake; position: number }[]>;
   public static splitMessage(text: string, options?: SplitOptions): string[];
+  public static camelize<T extends Record<string, any> | readonly any[]>(input: T): Camelize<T>;
 }
 
 export class Formatters extends null {
--- a/package-lock.json
+++ b/package-lock.json
@@ -15,6 +15,7 @@
         "@sapphire/async-queue": "^1.1.8",
         "@types/node-fetch": "^2.5.12",
         "@types/ws": "^8.2.0",
+        "camelcase-keys": "^7.0.1",
         "discord-api-types": "^0.25.0",
         "node-fetch": "^2.6.1",
         "type-fest": "^2.5.3",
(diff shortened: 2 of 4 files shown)


iCrawl (approved):

iCrawl:
This needs a rebase.

typings/index.d.ts
```diff
@@ -2241,6 +2241,7 @@ export class Util extends null {
     reason?: string,
   ): Promise<{ id: Snowflake; position: number }[]>;
   public static splitMessage(text: string, options?: SplitOptions): string[];
+  public static camelize<T extends Record<string, any> | readonly any[]>(input: T): Camelize<T>;
```
kyranet, line 2244: JFYI, this won't match structures such as the ones defined by builders. Instead, do the following:

```suggestion
  public static camelize<T extends {}>(input: T): Camelize<T>;
```

Unlike `object`, `{}` does **not** match `null`, and since arrays are objects, they're already supported by it.

It is possible that a TSLint rule will forbid that, add a ignore-next-line if that happens. You can also do `interface NonNullObject {}` with the ignore above, and use it.

kyranet (changes requested):

typings/index.d.ts
suneettipirneni: This seems to match the builders currently:
```ts
const commands = new SlashCommandBuilder().setName('test').setDescription('test').toJSON();
expectType<Camelize<RESTPostAPIApplicationCommandsJSONBody>>(Util.camelize(commands));
```

kyranet: I have no idea why it does, but it's completely invalid as it doesn't define a dynamic property accessor, which non-Record interfaces lack of.

suneettipirneni: Doesn't it match `Record<string, any>`? Since it's returning a json object from `#toJSON()`?

vladfrangu: Can you create a typescript playground with the issue you encountered @kyranet? Could be old TS issues?

kyranet: Just checked, apparently it complains when it's `Record<string, unknown>`, but `Record<string, any>` works fine.

I personally still prefer using `{}` for three reasons:

1. Doesn't rely on `any` (which might unintentionally change to `unknown` in the future, since... you know, `unknown` is preferred over `any` as it's stricter and disallows uncasted property access).
2. In case of a returned type, you cannot read properties from it unless it's casted, similar to `unknown` but restricted to non-null objects. On the contrary, `Record<string, any>` allows you to access to properties that might not even exist, and you won't even know (unless `noPropertyAccessFromIndexSignature` is enabled, but it's been a while since I saw somebody use that).
3. Better semantics, `Record<string, any>` literally means the following:
    ```ts
    interface Foo {
      [key: string]: any;
    }
    ```

    Which is wrong in many cases, `RESTPostAPIApplicationCommandsJSONBody` (and most interfaces) define a set of properties, but do not define a dynamic property access (`[key: string]: any`) member.

    `{}`, however, matches any object, from those that have no keys, to those that do, just like `{ 
foo: 'bar', hello: 'world' }` matches `{ foo: string }` as well as `{ hello: string }`.

iCrawl:
This needs a rebase.

suneettipirneni:
Gonna close this, there's not much use in this anymore now that #7052 is now merged.

PR #7054, opened 2021-12-01
types(Interaction): Narrow `memberPermissions`

Jiralite:
Narrowing this type as it is present from guilds.

Note: Prettier modified the file elsewhere.

### DIFF
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -1227,7 +1227,7 @@ export class Interaction<Cached extends CacheType = CacheType> extends Base {
   public type: InteractionType;
   public user: User;
   public version: number;
-  public memberPermissions: Readonly<Permissions> | null;
+  public memberPermissions: CacheTypeReducer<Cached, Readonly<Permissions>>;
   public inGuild(): this is Interaction<'present'>;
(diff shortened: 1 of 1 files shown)


vladfrangu:
Won't this make it return null for non cached guilds?

Jiralite:
> Won't this make it return null for non cached guilds?

The `RawType` (from a guild, specifically raw) and `PresentType` (from a guild) defaults to the `CachedType` (from a guild, specifically cached) in this case, so no:

```TypeScript
export type CacheTypeReducer<
  State extends CacheType,
  CachedType,
  RawType = CachedType,
  PresentType = CachedType | RawType,
  Fallback = PresentType | null,
> = [State] extends ['cached']
  ? CachedType
  : [State] extends ['raw']
  ? RawType
  : [State] extends ['present']
  ? PresentType
  : Fallback;
```

The only time it will be `null` is in the fallback where no guild checks are done as it is possible to receive this from a direct message.

vladfrangu (approved):

SpaceEEC (approved):

kyranet (approved):

iCrawl (approved):

PR #7060, opened 2021-12-03
docs(GuildStickerManager): fix create() example

SethCohen:
The docs have improper examples and can result in confusion. The proper input for the `tags` parameter in `.create()` shouldnt take in a unicode emoji; as per Discord's docs and as indicated in the Discord.js docs itself in the tags description, the proper input should be the **string name** of a unicode emoji.
e.g. 🍌 should be "banana"

I was confused as to why when I would create a sticker, there would be no unicode emoji attached to it - until I read the actual discord docs and saw the error. With this fix in the docs, hopefully users - like me - won't become confused as to why there is no emoji representation of their sticker(s).

**What the discord.js docs promote:**

**What the actual input should be:**

### DIFF
--- a/src/managers/GuildStickerManager.js
+++ b/src/managers/GuildStickerManager.js
@@ -47,12 +47,12 @@ class GuildStickerManager extends CachedManager {
    * @returns {Promise<Sticker>} The created sticker
    * @example
    * // Create a new sticker from a URL
-   * guild.stickers.create('https://i.imgur.com/w3duR07.png', 'rip', '🪦')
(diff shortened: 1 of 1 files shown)


SethCohen:
An alternative solution would be to implement a Map that maps a unicode emoji to discord's name representation so if the user does input a unicode emoji, before discord.js sends it to discord, it converts it to the string name representation. But that ones more of a Feature Request kind of deal, tbh..

mafineeek (approved):

SpaceEEC (approved):

vladfrangu (approved):
Thanks Dis cord

iCrawl (approved):

PR #7063, opened 2021-12-05
feat(ApplicationCommand): add setX methods for easier editing

ImRodry:
This PR adds 4 new methods to ApplicationCommand to make it easier to edit them. These are setName, setDescription, setDefaultPermission and setOptions. Although the edit method says that type conversions are supported, this does not seem to be the case when testing with the API so I decided not to add a setType method.

### DIFF
--- a/src/structures/ApplicationCommand.js
+++ b/src/structures/ApplicationCommand.js
@@ -189,7 +189,7 @@ class ApplicationCommand extends Base {
   }
 
   /**
-   * Edits the default permission of this ApplicationCommand.
+   * Edits the default permission of this ApplicationCommand
    * @param {boolean} [defaultPermission=true] The default permission for this command
    * @returns {Promise<ApplicationCommand>}
    */
--- a/src/structures/ApplicationCommand.js
+++ b/src/structures/ApplicationCommand.js
@@ -171,6 +171,42 @@ class ApplicationCommand extends Base {
   }
 
   /**
+   * Edits the name of this ApplicationCommand
+   * @param {string} name The new name of the command
+   * @returns {Promise<ApplicationCommand>}
+   */
+  setName(name) {
+    return this.edit({ name });
+  }
+
+  /**
+   * Edits the description of this ApplicationCommand
+   * @param {string} description The new description of the command
+   * @returns {Promise<ApplicationCommand>}
+   */
+  setDescription(description) {
+    return this.edit({ description });
+  }
+
+  /**
+   * Edits the default permission of this ApplicationCommand.
+   * @param {boolean} [defaultPermission=true] The default permission for this command
+   * @returns {Promise<ApplicationCommand>}
+   */
+  setDefaultPermission(defaultPermission = true) {
+    return this.edit({ defaultPermission });
+  }
+
+  /**
+   * Edits the options of this ApplicationCommand
+   * @param {ApplicationCommandOptionData[]} options The options to set for this command
+   * @returns {Promise<ApplicationCommand>}
+   */
+  setOptions(options) {
+    return this.edit({ options });
+  }
+
+  /**
    * Deletes this command.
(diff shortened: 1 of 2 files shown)


src/structures/ApplicationCommand.js
```diff
@@ -170,6 +170,42 @@ class ApplicationCommand extends Base {
...
+   * Edits the description of this ApplicationCommand
+   * @param {string} description The new description of the command
+   * @returns {Promise<ApplicationCommand>}
+   */
+  setDescription(description) {
+    return this.edit({ description });
+  }
+
+  /**
+   * Edits the default permission of this ApplicationCommand.
```
ceilingfans, line 192:
```suggestion
   * Edits the default permission of this ApplicationCommand
```

ceilingfans (changes requested):
Looks good to me

src/structures/ApplicationCommand.js
ImRodry: Sure ig

typings/index.d.ts
```diff
@@ -226,6 +226,10 @@ export class ApplicationCommand<PermissionsFetchType = {}> extends Base {
   public version: Snowflake;
   public delete(): Promise<ApplicationCommand<PermissionsFetchType>>;
   public edit(data: ApplicationCommandData): Promise<ApplicationCommand<PermissionsFetchType>>;
+  public setName(name: string): Promise<ApplicationCommand<PermissionsFetchType>>;
+  public setDescription(description: string): Promise<ApplicationCommand<PermissionsFetchType>>;
+  public setDefaultPermission(defaultPermission?: boolean): Promise<ApplicationCommand<PermissionsFetchType>>;
+  public setOptions(options: ApplicationCommandOptionData[]): Promise<ApplicationCommand<PermissionsFetchType>>;
```
thehackerboi69github, line 232:
```suggestion
  public setName(name: string): Promise<this>;
  public setDescription(description: string): Promise<this>;
  public setDefaultPermission(defaultPermission?: boolean): Promise<this>;
  public setOptions(options: ApplicationCommandOptionData[]): Promise<this>;
```

thehackerboi69github (changes requested):

typings/index.d.ts
ImRodry: It should return the same thing as the edit method

suneettipirneni: This isn't a builder it's a manager.

thehackerboi69github: What exactly is the difference though?

suneettipirneni: > What exactly is the difference though?

Managers are backed by the api, whereas builders only output json that _may_ be used with the api later on

thehackerboi69github: No, like what is the difference between `this` and `ApplicationCommand` in this case?

suneettipirneni: `this` doesn't take generic parameters whilst `ApplicationCommand` does.

thehackerboi69github: But `this` is already an `ApplicationCommand<PermissionsFetchType>` right?

joseywoermann: > But `this` is already an `ApplicationCommand<PermissionsFetchType>` right?

If so, shouldn't it be `typeof this`? Because `this` is an object, not a type.

suneettipirneni: `this` is always a type, it just represents the type of the enclosing class.

vladfrangu (approved):

Mysterious-Dev (approved):

thebigwolfy (approved):

kyranet (approved):

SpaceEEC (approved):

mafineeek (approved):
good

PR #7064, opened 2021-12-06
Make options parameter to Client implementation class constructor optional, or possibly a blank object

StoneCypher:
Hi.  New user.  Could be wrong.

Most of the tutorials I find show code like `const client = new Discord.client();`.

[The actual constructor implementation](https://github.com/discordjs/discord.js/blob/main/src/client/Client.js#L38) defers to a class inherited constructor superclass, [which offers a blank object as a default value](https://github.com/discordjs/discord.js/blob/main/src/client/BaseClient.js#L13).

The subordinate class never actually refers to the constructor argument again after calling the superclass; instead it refers to the class members, which were set by the parent method.

On these grounds, I believe that you intend to allow a user to be allowed to pass 

1. Nothing
2. A blank object

Currently, when running in TypeScript, you are required to pass a valid instance of `ClientOptions`, which means that as a new user you have to figure out what the hell an `Intent` is.

I think this isn't intentional.

I have made the parameter optional, and I have allowed the substitution of a blank object for the argument.

Please consider.  TY.

<br/><br/>

Because the following seems unlikely to be intentional:

image

<br/><br/>

- I know how to update typings and have done so ~~, or typings don't need updating~~
    - In fact this is 100% a typing change
- ***this is arguable*** This PR includes breaking changes (methods removed or renamed, parameters moved or removed)
    - The change is to be more permissive, in line with your existing JS usage and your own documentation

### DIFF
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -500,7 +500,7 @@ export abstract class Channel extends Base {
 export type If<T extends boolean, A, B = null> = T extends true ? A : T extends false ? B : A | B;
 
 export class Client<Ready extends boolean = boolean> extends BaseClient {
-  public constructor(options: ClientOptions);
+  public constructor(options? : ClientOptions | {});
   private actions: unknown;
   private presence: ClientPresence;
   private _eval(script: string): unknown;


ckohen:
This is completely intentional, you cannot connect to the gateway without specifying intents. If you want no intents, you need to specifically pass intents: 0. Forcing users to learn about intents allows them to be more knowledgable about what they actually need and to only pick those that they will actually use.

StoneCypher:
> This is completely intentional, you cannot connect to the gateway without specifying intents. If you want no intents, you need to specifically pass intents: 0. Forcing users to learn about intents allows them to be more knowledgable about what they actually need and to only pick those that they will actually use.

Has it always been this way?

I was introduced to your library by youtube tutorials.  On watching three of them in a row ([1](https://www.youtube.com/watch?v=_o8lwjVNJsg), [2](https://www.youtube.com/watch?v=BmKXBVdEV0g), [3](https://www.youtube.com/watch?v=HMmIm1RZiBM)) , they all pass nothing there, and appear to be working.

I could change this PR to throw an informative message instead, if you prefer?

&nbsp;

> Forcing users to learn about intents allows them to be more knowledgable about what they actually need and to only pick those that they will actually use.

I haven't found your onboarding yet.  It's not clear where to learn this.  It's not clear if these are analogous to the intents I used when signing up the bot.  If they are, it's not clear why I have to write them a second time, if they match, if they should be a subset, or what.

It's difficult if the tutorials are out of step with the library, and the library doesn't have a one page "getting started" in its docs, y'know?

StoneCypher:
Also, are you certain?

The code appears to be explicitly written to permit this.

ckohen:
This is new as of v13, which uses gateway version 9 instead of 6  / 7, in which intents were optional.

The official guide can be found at https://discordjs.guide and its completely open source so if you feel its lacking something, feel free to make a PR.

> Also, are you certain?

Yep, the code doesn't validate options but it does happen
https://github.com/discordjs/discord.js/blob/01f8d1bed564a07d40b184dc7ff686a895ddda31/src/client/Client.js#L73
Which is this function here, which the first thing it does is validate intents, otherwise throwing a very informative error
https://github.com/discordjs/discord.js/blob/01f8d1bed564a07d40b184dc7ff686a895ddda31/src/client/Client.js#L542-L547

StoneCypher:
Ok

PR #7069, opened 2021-12-06
fix(ReactionCollector): check for channel.threads

kyranet:
**Please describe the changes this PR makes and why it should be merged:**

`Message#channel` becomes `null` when the channel is deleted, so we cannot rely on it for handling channel deletions of a thread's parent. However, we can do the reverse and check whether the message's channel ID is a thread contained by the deleted channel.

Fixes #7066

**Status and versioning classification:**

- Code changes have been tested against the Discord API, or there are no code changes
- I know how to update typings and have done so, or typings don't need updating

PR #7070, opened 2021-12-06
fix(InteractionCollector): handle thread (and parent) deletion

### LINKED ISSUE #6905: Possible event listener leak via interaction collectors
### Issue description

There is a possibility with interaction collectors that attached listeners are never removed. This happens in threads when the interaction collector's options have no `time` specified.

Consider this example where we are listening to component interactions:

```JavaScript
const collector = <Message>.channel.createMessageComponentCollector({
  max: 5
});

collector.on("collect", interaction => {
  console.log("Collected something!");
  if (interaction.customId === "test") collector.stop();
});

collector.on("end", console.log);
```

No `time` is specified, but that's alright because we (expect to) manually end our interaction collector.

Now, let's suppose the parent channel is deleted before we receive the interaction. In this case, the interaction collector detects the deletion of a channel and runs the following code:

https://github.com/discordjs/discord.js/blob/fae4abf2f7071425639e8d1ea1a8598403fa2a07/src/structures/InteractionCollector.js#L199-L209

However, this condition is _always_ `false`. `this.channelId` is the id of a `ThreadChannel`, but the deleted channel is a `GuildChannel`. Therefore, the interaction collector is not stopped.

So, what now? Well, the interaction collector will keep going, despite the channel being deleted (and thus the thread). It will never receive components ever again since the interaction collector is bound to a deleted thread and therefore, there are listeners here that have not been removed.

### Code sample

```typescript
const { Client, Intents } = require('discord.js');

const client = new Client({
  intents: [
    Intents.FLAGS.GUILDS,
    Intents.FLAGS.GUILD_MESSAGES
  ]
});

client.on("messageCreate", async message => {
  if (message.content === "start") {
    await message.channel.send({
      content: "help",
      components: [
        {
          type: "ACTION_ROW",
          components: [
            {
              type: "BUTTON",
              label: "hi",
              style: "PRIMARY",
              customId: "test",
              disabled: false
            }
          ]
        }
      ]
    });

    const collector = message.channel.createMessageComponentCollector({
      max: 5
    });

    // Now delete the parent channel and log `client.eventNames()`

    collector.on("collect", () => console.log("Collected something!"));
    collector.on("end", console.log);
  }
});

client.login("token");
```

### discord.js version

13.3.0-dev

### Node.js version

16.6.1

### Operating system

macOS Big Sur 11.6

Medium (should be fixed soon)

No Partials

GUILDS, GUILD_MESSAGES

fae4abf

kyranet:

Backports the thread (and thread parent) channel deletion from #6902 to `InteractionCollector` with the fixed checks from #7069.

Fixes #6905

### DIFF
--- a/src/structures/InteractionCollector.js
+++ b/src/structures/InteractionCollector.js
@@ -105,7 +105,9 @@ class InteractionCollector extends Collector {
 
     if (this.channelId) {
       this._handleChannelDeletion = this._handleChannelDeletion.bind(this);
+      this._handleThreadDeletion = this._handleThreadDeletion.bind(this);
       this.client.on(Events.CHANNEL_DELETE, this._handleChannelDeletion);
+      this.client.on(Events.THREAD_DELETE, this._handleThreadDeletion);
     }
 
     if (this.guildId) {
@@ -120,6 +122,7 @@ class InteractionCollector extends Collector {
       this.client.removeListener(Events.MESSAGE_DELETE, this._handleMessageDeletion);
       this.client.removeListener(Events.MESSAGE_BULK_DELETE, bulkDeleteListener);
       this.client.removeListener(Events.CHANNEL_DELETE, this._handleChannelDeletion);
+      this.client.removeListener(Events.THREAD_DELETE, this._handleThreadDeletion);
       this.client.removeListener(Events.GUILD_DELETE, this._handleGuildDeletion);
       this.client.decrementMaxListeners();
     });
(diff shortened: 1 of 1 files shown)


vladfrangu (approved):

kyranet:
For the code sample in the linked issue, you're using `message.channel.createMessageComponentCollector`, which will always set a guild the client is in.

If you're somehow passing an invalid guild ID (or one that the bot isn't in), we can't determine in any way, and I believe that falls under [user error](https://en.wikipedia.org/wiki/User_error).

iCrawl (approved):

SpaceEEC (approved):

PR #7071, opened 2021-12-07
Cleaned up some code and comments in Util.js

Enadel710:
I fixed some of the grammar in the comments of the flatten function of Util.js and inverted a condition to make the code more understandable.

### DIFF
--- a/src/util/Util.js
+++ b/src/util/Util.js
@@ -16,7 +16,7 @@ let deprecationEmittedForRemoveMentions = false;
  */
 class Util extends null {
   /**
-   * Flatten an object. Any properties that are collections will get converted to an array of keys.
+   * Static method to flatten an object. Any properties that are collections will get converted to an array of keys.
    * @param {Object} obj The object to flatten.
    * @param {...Object<string, boolean|string>} [props] Specific properties to include/exclude.
    * @returns {Object}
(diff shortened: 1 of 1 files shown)


src/util/Util.js
```diff
@@ -16,7 +16,7 @@ let deprecationEmittedForRemoveMentions = false;
  */
 class Util extends null {
   /**
-   * Flatten an object. Any properties that are collections will get converted to an array of keys.
+   * Static method to flatten an object. Any properties that are collections will get converted to an array of keys.
```
kyranet, line 19: We don't use the words "static" and "method" to describe a... static method.

Instead, the following is more grammatically correct:
```suggestion
   * Flattens an object out. Any properties that are collections will get converted to an array of keys.
```

src/util/Util.js
```diff
@@ -33,23 +33,24 @@ class Util extends null {
...
+        if (element instanceof Collection) out[newProp] = Array.from(element.keys());
+        // If valueOf is a Collection, use its array of keys
+        else if (valueOf instanceof Collection) out[newProp] = Array.from(valueOf.keys());
+        // If it's an array, flatten each element
+        else if (Array.isArray(element)) out[newProp] = element.map(e => Util.flatten(e));
+        // If it's an object with a primitive `valueOf`, use that value
+        else if (typeof valueOf !== 'object') out[newProp] = valueOf;
+        // If it's a primitive, set out to be that primitive value
+        else if (!elemIsObj) out[newProp] = element;
+      }
```
kyranet, line 53: Please revert this, it was fine as it was, there's no real readability improvement here.

While you may think that sticking to basic syntax would increase the readability, the increase of curly braces can lead to the opposite effect, which is why that alongside to reduce the indentation in our code, we chose to use `continue` here.

kyranet (changes requested):

PR #7073, opened 2021-12-07
feat(Collector): yield all collected values

### LINKED ISSUE #4139: feat: Change return type of for-await-of Collectors to return any param collected

Currently, a Collector will only ever return the first item that gets emitted from a `collect` event when you use the `for-await-of` syntax.

https://github.com/discordjs/discord.js/blob/026691702da7d1fcf9a959f00082f07f1e14b13d/src/structures/interfaces/Collector.js#L222-L223

This is *fine* for message collectors, but MessageReaction Collectors return two elements on `collect`: MessageReaction and User.

That means you're limited to either:

1. Extend the ReactionCollector class and overwrite the async symbol to something like this (which is a simple copy paste)

```diff
async *[Symbol.asyncIterator]() {
	const queue = [];
- const onCollect = item => queue.push(item);
+ const onCollect = (...items) => queue.push(items);
	this.on('collect', onCollect);
```

(doable, but for TS users, thats a `@ts-ignore` needed, not really nice)

2. Using the event callbacks, which is fine, but not as clean as `for-await-of`

A simple (but sadly breaking) solution would be changing the return type of

https://github.com/discordjs/discord.js/blob/026691702da7d1fcf9a959f00082f07f1e14b13d/typings/index.d.ts#L259

to default to `unknown[]` in the base Collector class (citation needed), `[Message]` in the MessageCollector, and `[MessageReaction, User]` for ReactionCollector. No need to add `this.collected` to the typing of the result as you can access it in the loop.

This would mean that for ReactionCollectors, you now do

```js
for await (const [reaction, user] of collector) {}
```

and for MessageCollector you do

```js
for await (const [message] of collector) {}
```

which sadly doesn't make full sense but for consistency sake it's what would have to be done

Could you solve this yourself by extending the class? Absolutely! But the collectors are very powerful and having this out of the box would be amazing!

EDIT: Since Collector takes a `K, V`, you COULD make V be `[MessageReaction, User]`; however the code also needs to be updated to handle that, soo...

**Additional context**

None

kyranet:

Fixes #4139 by yielding an array of all the parameters received from the `collect` event.

### DIFF
--- a/packages/discord.js/src/structures/interfaces/Collector.js
+++ b/packages/discord.js/src/structures/interfaces/Collector.js
@@ -236,7 +236,7 @@ class Collector extends EventEmitter {
    */
   async *[Symbol.asyncIterator]() {
     const queue = [];
-    const onCollect = item => queue.push(item);
+    const onCollect = (...item) => queue.push(item);
     this.on('collect', onCollect);
 
     try {
--- a/packages/discord.js/typings/index.d.ts
+++ b/packages/discord.js/typings/index.d.ts
@@ -670,7 +670,7 @@ export abstract class Collector<K, V, F extends unknown[] = []> extends EventEmi
   public handleDispose(...args: unknown[]): Promise<void>;
   public stop(reason?: string): void;
   public resetTimer(options?: CollectorResetTimerOptions): void;
-  public [Symbol.asyncIterator](): AsyncIterableIterator<V>;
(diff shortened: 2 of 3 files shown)


iCrawl (approved):

SpaceEEC (approved):

vladfrangu (approved):
Thank. You.

PR #7074, opened 2021-12-07
refactor: make `Structure#deleted` a getter to a WeakSet

kyranet:

`Structure#deleted` is a value that's almost always `false`, except typically during a small set of time (when structures are deleted). By moving the information to WeakSets, we'll be able to only store those that are deleted (and automatically clean themselves once the entries are freed from memory by the GC), resulting on large memory reductions over the millions of entries with a `deleted` field.

|      Entries | Before (Total) | After (Total) |             Diff | Before (Used) | After (Used) |            Diff |
| :----------: | :------------: | :-----------: | :--------------: | :-----------: | :----------: | :-------------: |
| `10_000_000` |      1.1547 GB |     1.0946 GB |   60.1 MB (5.2%) |     1.1115 GB |    1.0368 GB | 74.7 MB (6.72%) |
|  `1_000_000` |    133.9063 MB |   126.1563 MB |  7.75 MB (5.79%) |   100.0864 MB |   92.4568 MB | 7.63 MB (7.62%) |
|    `100_000` |     30.9063 MB |    21.9063 MB |    9 MB (29.12%) |    13.7924 MB |   13.0284 MB |  764 KB (5.54%) |

> ℹ️ The tests are done measuring the memory usage before and after populating a map with `{ id: string; deleted: false }` in the case of "Before", and `{ id: string }` in the case of "After", being `id = (i + 100_000_000).toString()`. As such, only the differences matter as it shows the overhead of the `deleted` field.

> ℹ️ `@discordjs/docgen` was updated to fix an error in the `test` script.

Also adds `Sticker#deleted` in the typings and the docs, it was assigned in the event, but the structure did not set the field.


Jiralite:
Looks like the CI is failing because wherever `const X = new WeakSet()` is defined, it's expecting that to be a constructor, property, method or event on the class but it's none of them... maybe it's best to just remove the JSDoc on them since they're private?

Also, it's throwing because we're using 0.10.0 of the generator which doesn't include https://github.com/discordjs/docgen/pull/15 which will resolve this, but doing the above will fix the issue (but this should probably be done also).

vladfrangu (approved):

iCrawl:
This needs a rebase.

kyranet:
Rebased and tested, the results are consistent with those from v13.3.1.

SpaceEEC (commented):
Why wasn't the `deleted` getter/setter moved into `Base`?

kyranet:
Because it's not set by all structures, but only a few.

SpaceEEC (approved):

iCrawl (approved):

Mirdukkk:
Let's test your implementation.
-

I have created two classes, one uses just a true/false property to determine whether an entity has been deleted, the second uses WeakMap for this.

Both classes have been created 10 million times.
50% of these classes marked as deleted.

For testing, I used a bot and the eval command. The bot does not have any cache. It also does not handle any events other than MESSAGE_CREATE.

**Let's measure RAM usage in the first case (property):**
[click](https://user-images.githubusercontent.com/44965055/145408865-f230e1e0-c029-4581-ac9b-247148041f0e.png)
Result: 422.6640625 MB.

**Let's measure RAM usage in the second case (WeakMap):**
[click](https://user-images.githubusercontent.com/44965055/145409095-0879c40a-7a21-490c-b187-2dbb96ac4426.png)
Result: 520.4375 MB.

You can say that this is an error due to adding a large number of objects to the WeakMap in a short time. So let's look at the RAM consumption 80 seconds after creating the array:
[click](https://user-images.githubusercontent.com/44965055/145409410-2b871a29-a023-4296-a2c0-1e1b4315fe72.png)
Result: 475.5390625 MB.

**At best, your implementation consumes 50 megabytes more than a normal property.**

You probably haven't noticed, but the eval command also counts how long the code was executed.
The code using just a property for implementation was executed in 2238510500 nanoseconds or 2238.510ms or 2.2s.
The code using WeakMap for implementation was executed in 139577886800 nanoseconds or 139577.887ms or 139s.
At the same time, the code using the WeapMap used 100% of the CPU and the bot went offline because of this.

**Let's see how long it takes us to check how many entities are marked as deleted.**

First let's create a globally accessible array with classes that use WeakMap:
[click](https://user-images.githubusercontent.com/44965055/145410724-3654c946-1923-4d6e-b12e-55c8f15e0b62.png)
Then we will do the same, but for classes that use just a property:
[click](https://user-images.githubusercontent.com/44965055/145410894-87cd435e-92b1-4427-86b6-348524dd909b.png)

And now let's see how long it will take us to count the number of entities marked as deleted in both cases:
[click](https://user-images.githubusercontent.com/44965055/145411092-3a3649a8-b9c8-4f5c-b59d-988d211dc9d9.png)
One hundred milliseconds vs sixty seconds.
And also look at the CPU usage during the count: [click](https://cdn.discordapp.com/attachments/677558850026733586/918504896377024532/2021-12-09_17-08-30_1639058910.mp4)

**Based on everything that is written above, I believe that implementation needs to be returned to its original state.**

Answers to possible questions:
1. The time spent on code execution is measured like this:
image
2. The bot does not use djs/eris. There could be no random spikes in consumption.
3. You got less consumption because you just used objects with and without a property, but did not take into account that the reference values inside WeakMap also take up RAM.
4. The code of a class that uses just a property: 
```js
class Entity {

  get deleted() {
    return !!this._deleted
  }

  set deleted(value) {
    const condition = !!value

    if (condition) {
      this._deleted = true
    } else {
      this._deleted = false
    }
  }
  
}

let deleted = false
const isDeletedFn = () => {
  deleted = !deleted
  return deleted
}

const mapFn = () => {
  const ent = new Entity()
  ent.deleted = isDeletedFn()
  return ent
}

const arr = Array.from({ length: 10_000_000 }, mapFn)
```
5. The code of a class that uses WeakMap:
```js
const deletedEntities = new WeakSet()

class Entity {
  
  get deleted() {
    return deletedEntities.has(this)
  }

  set deleted(value) {
    const condition = !!value

    if (condition) {
      deletedEntities.add(this)
    }
  }
  
}

let deleted = false
const isDeletedFn = () => {
  deleted = !deleted
  return deleted
}

const mapFn = () => {
  const ent = new Entity()
  ent.deleted = isDeletedFn()
  return ent
}

const arr = Array.from({ length: 10_000_000 }, mapFn)
```
6. Node.js v17.1.0

kyranet:
Not going to bother reading after that "50%", sorry, but 99.99% of the times, the instance is just gone quickly after `deleted` is set to `true`.

If you're storing deleted instances, that's your fault, because discord.js does not.

Mirdukkk:
Are you kidding? And the CPU usage when checking .deleted doesn't bother you? Ok. Discord.js has never been a good-performance library, and now it has only gotten worse. Good luck.

kyranet:
The CPU when checking `.deleted` should be overall pretty low because discord.js barely checks it, and a lot of people also don't check for that. In fact, people usually have far more CPU than RAM these days, although yes, that CPU usage is a concern to me. Although I'm at blame for not reading all of it thoroughly (I'm sometimes sent benchmarks that are usually invalid or meaningless), it'd have been easier to follow if you split the text in two parts, one for RAM and another for CPU, as it stands it was a bit hidden between RAM metrics and screenshots.

Honestly, I'm leaning towards just removing `.deleted` for v14, it was never reliable to start with, and it won't be at all in the soon future with v15's lower caching dependency that we aim to achieve to reduce the gotchas when using the library.

I'll make an issue regarding that field's removal in a bit, then discuss with the rest of the team whether we want to revert this or continue using WeakSet (since it brings benefits to the majority), and if we want to remove the field due to how troublesome it can be.

Mirdukkk:
I tested everything again, but marking only 1% of the entities as deleted. Indeed, you are right, so the consumption is less by 100 megabytes than it was with classes that use just a property. But still, when checking .deleted CPU usage is 50% more than with just property. 

Therefore, in general, you are right: if the WeakMap will be almost empty, then the RAM consumption will be less, as well, due to the fact that the property .deleted is used infrequently, the perceived CPU consumption will be only a couple of percent more. 

But still, if someone will store deleted entities, the consumption of both RAM and CPU will grow very much. I think it's worth writing somewhere in the docs next to the instructions for overriding the cache that storing deleted entities is strictly not recommended. For example, [here](https://discord.js.org/#/docs/main/stable/typedef/ClientOptions):
image
I'm sorry for the last comment, I was very angry when you didn't even read what I spent time on.

kyranet:
Sorry for the delay, I created #7091 pointing all the issues with `deleted` and proposing the removal of the field in favor of tracking them on the user's side however they want.

I thank you a lot for bringing me the performance impact this change caused despite how beneficial it was for RAM usage, and I'm glad you could see the positive memory impact this PR has, I just assumed that CPU-wise it'd be just twice or thrice slower, which would have been reasonable (think that `#private` properties are WeakMap-based, and they're just a small fraction slower than regular properties).

Mirdukkk:
You did a good job! You have well justified your position on the removal of this property, and indeed, I agree with almost everything. That's great! I am glad that I helped you see the problems with CPU usage, and that an almost unused property can be easily removed. Thank you for working on this library!

PR #7079, opened 2021-12-08
refactor(SnowflakeUtil): switch to `@sapphire/snowflake`

kyranet:

Replaces `SnowflakeUtil` with `DiscordSnowflake` from `@sapphire/snowflake`, following the decision taken with `@discordjs/rest`.


That-Guy977:
What about possible customization of those fields through an optional options parameter? As well as possibly the increment
I think the more customizable behavior from that could be nice to have, although I don't really see much practical use for this (although I don't for not hardcoding them either)

kyranet:
If we go for that path, we might as well just nuke all snowflake stuff from this package and use [`@sapphire/snowflake`](https://www.npmjs.com/package/@sapphire/snowflake). [`@discordjs/rest`](https://github.com/discordjs/discord.js-modules/blob/main/packages/rest/package.json#L49) already does.

That-Guy977:
I don't see any reason not to, that seems like a pretty good option.

wasdennnoch:
The process ID is only allocated 5 bits in the Snowflake format. Generally, PIDs go above 32, and this would cause the PID's bits to extend into the worker ID and even the timestamp if >1024 (10 bits), falsifying them since all Snowflake components are ORed.

kyranet:
Nice catch @wasdennnoch! I'm currently working on https://github.com/sapphiredev/utilities/pull/231 since we'll remove `SnowflakeUtil` in favour of the Sapphire package. I have applied your observation in the linked PR, thanks 😅

src/index.js
```diff
@@ -25,7 +25,6 @@ exports.MessageFlags = require('./util/MessageFlags');
 exports.Options = require('./util/Options');
 exports.Permissions = require('./util/Permissions');
 exports.RateLimitError = require('./rest/RateLimitError');
-exports.SnowflakeUtil = require('./util/SnowflakeUtil');
```
iCrawl, line 28: We don't intend to expose it?

kyranet: What name should we expose it as? `SnowflakeUtil` (same as now)? `DiscordSnowflake` (same as dep)?

iCrawl: I don't really care honestly, I was just wondering if this was a conscious decision

kyranet: I was thinking among the lines of letting users install the dependency if they need it, however, we'd naturally expose a part of the library since `TwitterSnowflake` is not really related to discord.js.

However, `@discordjs/collection`'s `Collection` is re-exported, although it might have been for retro-compatibility since we exported it for many versions, the same can't really be said with the new library, which has similar methods but different inputs and outputs.

I don't really care either, I can see the use of re-exporting `DiscordSnowflake` as `SnowflakeUtil` for ease of transition and exposing something that might be useful to users, were they to need it, but at the same time, this is mostly an utility class we use internally, like `AsyncQueue` from `@sapphire/async-queue`, which is not exported.

iCrawl: Yeah the problem with them installing deps themselves is that version numbers may diverge and if it's useful for the end user, a re-export solves that problem.

kyranet: Re-exported in b0d1c8baebe147476ae324729bba12383cbaf079 as `SnowflakeUtil`, I figured we didn't need to use `Discord` in the name as this is inside a Discord-focused library.

iCrawl:
This needs a rebase.

src/structures/interfaces/TextBasedChannel.js
```diff
@@ -358,3 +357,4 @@ module.exports = TextBasedChannel;
 
 // Fixes Circular
 const MessageManager = require('../../managers/MessageManager');
+const { DiscordSnowflake } = require('@sapphire/snowflake');
```
vladfrangu, line 360: Is this really needed down here?

vladfrangu (changes requested):

src/structures/interfaces/TextBasedChannel.js
kyranet: No, auto-import did that, it's fixed.

vladfrangu (approved):

iCrawl (approved):

SpaceEEC (approved):

PR #7087, opened 2021-12-09
refactor(UserManager): Move methods to the manager

Jiralite:
This moves the following methods to the manager:

- `createDM()`
- `deleteDM()`
- `fetchFlags()`

The `send()` method is also on the manager.

The `User` class isn't needed for these requests which in turn only made these possible via doing:

```JavaScript
<Client>.users.fetch("user_id").then(async user => {
  const dmChannel = await user.createDM();
  // ...
});
```

But now, only one API call will be made:

```JavaScript
<Client>.users.createDM("user_id").then(dmChannel => {
  // ...
});
```

A nice change from this is the `send()` method on the manager. To send a message to an uncached user, you would first have to fetch them (`UserManager#fetch()`) and then run `User#send()` which internally calls `User#createDM()`. That's 3 API requests, but on the manager, a DM is created first (`User#createDM()`) and then sent (`DMChannel#send()`). 1 less!

### DIFF
--- a/src/managers/UserManager.js
+++ b/src/managers/UserManager.js
@@ -105,6 +105,17 @@ class UserManager extends CachedManager {
   }
 
   /**
+   * Sends a message to a user.
+   * @param {UserResolvable} user The UserResolvable to identify
+   * @param {string|MessagePayload|MessageOptions} options The options to provide
+   * @returns {Promise<Message>}
+   */
+
+  async send(user, options) {
+    return (await this.createDM(user)).send(options);
+  }
+
+  /**
    * Resolves a {@link UserResolvable} to a {@link User} object.
    * @param {UserResolvable} user The UserResolvable to identify
    * @returns {?User}
--- a/src/managers/UserManager.js
+++ b/src/managers/UserManager.js
@@ -32,7 +32,7 @@ class UserManager extends CachedManager {
    */
 
   /**
-   * The DM between the client's user and this user
+   * The DM between the client's user and a user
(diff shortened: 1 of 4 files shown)


kyranet (approved):

vladfrangu (approved):

iCrawl (approved):

src/managers/UserManager.js
```diff
@@ -32,26 +32,49 @@ class UserManager extends CachedManager {
    */
 
   /**
-   * Resolves a {@link UserResolvable} to a {@link User} object.
+   * The DM between the client's user and this user
```
SpaceEEC, line 35:
```suggestion
   * The DM between the client's user and a user
```

SpaceEEC (changes requested):
Just a small nit

src/managers/UserManager.js
```diff
@@ -70,6 +93,50 @@ class UserManager extends CachedManager {
...
+    return (await this.fetch(user, options)).flags;
+  }
+
+  /**
+   * Sends a message to a user.
+   * @param {UserResolvable} user The UserResolvable to identify
+   * @param {string|MessagePayload|MessageOptions} options The options to provide
+   * @returns {Promise<Message>}
+   */
+
```
SpaceEEC, line 113:
```suggestion
```

SpaceEEC (dismissed):
Otherwise lgtm

SpaceEEC (approved):

PR #7089, opened 2021-12-09
fix(GuildAuditLog): Remove `Promise`s in constructor

Jiralite:
The constructor shouldn't be utilising `Promise`s, so I've removed them by checking the cache and defaulting to creating an `Invite`.

`changes` also included `channel_id` & `inviter_id` which weren't being stored in the `Invite` class. I think this would be a loss if we don't expose that data as at least the developer can fetch the user/resolve the channel themselves as full objects are not sent across.

### DIFF
--- a/src/structures/GuildAuditLogs.js
+++ b/src/structures/GuildAuditLogs.js
@@ -2,11 +2,11 @@
 
 const { Collection } = require('@discordjs/collection');
 const Integration = require('./Integration');
+const Invite = require('./Invite');
 const { StageInstance } = require('./StageInstance');
 const { Sticker } = require('./Sticker');
 const Webhook = require('./Webhook');
 const { OverwriteTypes, PartialTypes } = require('../util/Constants');
-const Permissions = require('../util/Permissions');
 const SnowflakeUtil = require('../util/SnowflakeUtil');
 const Util = require('../util/Util');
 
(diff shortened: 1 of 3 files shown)


kyranet (approved):

src/structures/GuildAuditLogs.js
```diff
@@ -502,19 +502,21 @@ class GuildAuditLogsEntry {
...
-            o[c.key] = c.new ?? c.old;
-            return o;
-          }, {});
-        }
-      });
+      let change = this.changes.find(c => c.key === 'code');
+      change = change.new ?? change.old;
+
+      this.target =
+        guild.invites.cache.get(change) ??
```
suneettipirneni, line 509: Is this fetch correct? type `AuditLogChange` doesn't seem to be the same as `Invite` which this cache stores.

suneettipirneni (changes requested):

src/structures/GuildAuditLogs.js
Jiralite: I noticed on create, delete & update, `data.target_id` is always `null`.

Here's an example output of fetching an entry of an invite that was created:

```
{
  id: 'id',
  user_id: 'user_id',
  target_id: null,
  action_type: 40,
  changes: [
    { key: 'code', new_value: 'code' },
    { key: 'channel_id', new_value: 'channel_id' },
    { key: 'inviter_id', new_value: 'inviter_id' },
    { key: 'uses', new_value: 0 },
    { key: 'max_uses', new_value: 0 },
    { key: 'max_age', new_value: 604800 },
    { key: 'temporary', new_value: false }
  ]
}
```

`code` there is the invite code. This seems to be the only way to grab the invite code, so:

```JavaScript
let change = this.changes.find(c => c.key === 'code');
change = change.new ?? change.old;
```
This should make `change` be the invite code. Then `<Guild>..invites.cache` is keyed by the invite code, so this should be fine. Unless I'm missing something here (which I hope not).

src/structures/Invite.js
```diff
@@ -106,12 +106,24 @@ class Invite extends Base {
       this.maxUses ??= null;
     }
 
+    if ('inviter_id' in data) {
+      /**
+       * The user's id who created this invite
+       * @type {?Snowflake}
+       */
+      this.inviterId = data.inviter_id;
+      this.inviter = this.client.users.resolve(data.inviter_id);
```
vladfrangu, line 115: This should really be turned to a getter in the next major version, maybe slap a todo on this? CC @discordjs/the-big-4

vladfrangu (approved):

src/structures/Invite.js
kyranet: Sure 👍🏼

Jiralite: Will that also be the case for `this.channel`?

vladfrangu: Yes

Jiralite: Added with 78477b2fb28e5754e475ed25547b96acdac47d49!

iCrawl (approved):

thebigwolfy (approved):

SpaceEEC (approved):

PR #7096, opened 2021-12-10
refactor(RoleManager): Move some methods over from Role

Jiralite:
This moves the following methods to the manager:

- `comparePositions()`
- `delete()`

The `RoleManager` can manage the `delete()` API method.

As for `Role.comparePositions()`... not too sure why it's a static method on `Role` to be honest. It would make more sense for this to be moved to the manager. The manager stores the cache of roles, so I believe this method may reside here nicely. It's also no longer a static method and accepts `RoleResolvable`s.

```JavaScript
// client is a Client
const guild = client.guilds.resolve("guild_id");
const role1 = guild.roles.resolve("role_id1");
const role2 = guild.roles.resolve("role_id2");

// Old way of using the method
const { Role } = require("discord.js");
Role.comparePositions(role1, role2);

// New way of using the method
guild.roles.comparePositions(role1, role2);
```

### DIFF
--- a/src/managers/RoleManager.js
+++ b/src/managers/RoleManager.js
@@ -219,9 +219,9 @@ class RoleManager extends CachedManager {
   }
 
   /**
-   * Deletes the role.
+   * Deletes a role.
    * @param {RoleResolvable} role The role to delete
-   * @param {string} [reason] Reason for deleting this role
+   * @param {string} [reason] Reason for deleting the role
    * @returns {Promise<Role>}
    * @example
    * // Delete a role
--- a/src/managers/RoleManager.js
+++ b/src/managers/RoleManager.js
@@ -219,6 +219,24 @@ class RoleManager extends CachedManager {
   }
 
   /**
+   * Deletes the role.
+   * @param {RoleResolvable} role The role to delete
+   * @param {string} [reason] Reason for deleting this role
+   * @returns {Promise<Role>}
+   * @example
+   * // Delete a role
+   * guild.roles.delete('222079219327434752', 'The role needed to go')
+   *   .then(deleted => console.log(`Deleted role ${deleted.name}`))
+   *   .catch(console.error);
+   */
+  async delete(role, reason) {
+    const id = this.resolveId(role);
+    await this.client.api.guilds[this.guild.id].roles[id].delete({ reason });
+    this.client.actions.GuildRoleDelete.handle({ guild_id: this.guild.id, role_id: id });
+    return this;
+  }
+
+  /**
    * Batch-updates the guild's role positions
    * @param {GuildRolePosition[]} rolePositions Role positions to update
    * @returns {Promise<Guild>}
@@ -242,6 +260,21 @@ class RoleManager extends CachedManager {
       guild_id: this.guild.id,
       roles: rolePositions,
     }).guild;
+  }
+
+  /**
+   * Compares the positions of two roles.
+   * @param {RoleResolvable} role1 First role to compare
+   * @param {RoleResolvable} role2 Second role to compare
+   * @returns {number} Negative number if the first role's position is lower (second role's is higher),
+   * positive number if the first's is higher (second's is lower), 0 if equal
+   */
+  comparePositions(role1, role2) {
+    const resolvedRole1 = this.resolve(role1);
+    const resolvedRole2 = this.resolve(role2);
+    if (!resolvedRole1 || !resolvedRole2) throw new TypeError('INVALID_TYPE', 'role', 'Role nor a Snowflake');
+    if (resolvedRole1.position === resolvedRole2.position) return resolvedRole2.id - resolvedRole1.id;
+    return resolvedRole1.position - resolvedRole2.position;
   }
 
   /**
--- a/src/managers/RoleManager.js
+++ b/src/managers/RoleManager.js
@@ -273,7 +273,11 @@ class RoleManager extends CachedManager {
     const resolvedRole1 = this.resolve(role1);
     const resolvedRole2 = this.resolve(role2);
     if (!resolvedRole1 || !resolvedRole2) throw new TypeError('INVALID_TYPE', 'role', 'Role nor a Snowflake');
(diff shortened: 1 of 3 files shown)


kyranet (approved):

src/managers/RoleManager.js
```diff
@@ -244,6 +262,21 @@ class RoleManager extends CachedManager {
...
+   * @param {RoleResolvable} role1 First role to compare
+   * @param {RoleResolvable} role2 Second role to compare
+   * @returns {number} Negative number if the first role's position is lower (second role's is higher),
+   * positive number if the first's is higher (second's is lower), 0 if equal
+   */
+  comparePositions(role1, role2) {
+    const resolvedRole1 = this.resolve(role1);
+    const resolvedRole2 = this.resolve(role2);
+    if (!resolvedRole1 || !resolvedRole2) throw new TypeError('INVALID_TYPE', 'role', 'Role nor a Snowflake');
+    if (resolvedRole1.position === resolvedRole2.position) return resolvedRole2.id - resolvedRole1.id;
```
vladfrangu, line 276: Should we maybe BigInt the ids when doing a `-`? CC @kyranet

vladfrangu (changes requested):

src/managers/RoleManager.js
kyranet: I don't think it matters here, the strings are casted to numbers (`'4' - '3'` returns `1`) before the operation. Since only the 2 digits will be imprecise, and they're always defined by the last field (increment, 12 bits) by a small set of them, it'd require the same worker and process to create two roles in the exact same millisecond and within a few hundred ops, which seem extremely unlikely.

However, if you want correctness and handling that extreme edge-case, be my guest

src/managers/RoleManager.js
```diff
@@ -218,6 +218,24 @@ class RoleManager extends CachedManager {
...
+   * // Delete a role
+   * guild.roles.delete('222079219327434752', 'The role needed to go')
+   *   .then(deleted => console.log(`Deleted role ${deleted.name}`))
+   *   .catch(console.error);
+   */
+  async delete(role, reason) {
+    const id = this.resolveId(role);
+    await this.client.api.guilds[this.guild.id].roles[id].delete({ reason });
+    this.client.actions.GuildRoleDelete.handle({ guild_id: this.guild.id, role_id: id });
+    return this;
```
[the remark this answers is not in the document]
Jiralite: Thank you for pointing it out! Fixed with 08d31a4d204064a998cf0d27feb73de26036b345.

src/managers/RoleManager.js
Jiralite: Handled nonetheless!

src/managers/RoleManager.js
```diff
@@ -273,7 +273,7 @@ class RoleManager extends CachedManager {
     const resolvedRole1 = this.resolve(role1);
     const resolvedRole2 = this.resolve(role2);
     if (!resolvedRole1 || !resolvedRole2) throw new TypeError('INVALID_TYPE', 'role', 'Role nor a Snowflake');
-    if (resolvedRole1.position === resolvedRole2.position) return resolvedRole2.id - resolvedRole1.id;
+    if (resolvedRole1.position === resolvedRole2.position) return BigInt(resolvedRole2.id) - BigInt(resolvedRole1.id);
```
muchnameless, line 276: wouldn't this change the return type to `number|bigint`?

Jiralite: Yep, thank you for pointing that out. Fixed with 35922f5a6ba4286e93948daa9b1c580827156553.

src/managers/RoleManager.js
```diff
@@ -218,6 +218,24 @@ class RoleManager extends CachedManager {
...
+   * // Delete a role
+   * guild.roles.delete('222079219327434752', 'The role needed to go')
+   *   .then(deleted => console.log(`Deleted role ${deleted.name}`))
+   *   .catch(console.error);
+   */
+  async delete(role, reason) {
+    const resolvedRole = this.resolve(role);
+    await this.client.api.guilds[this.guild.id].roles[resolvedRole.id].delete({ reason });
+    this.client.actions.GuildRoleDelete.handle({ guild_id: this.guild.id, role_id: resolvedRole.id });
+    return resolvedRole;
```
kyranet, line 236: We can just do what `MessageManager#delete` (and possibly other places as well) do: return void so it doesn't rely on caching.

https://github.com/discordjs/discord.js/blob/ee91f5a19cdcf15f17a80fc9a25a382b7cbcde9f/src/managers/MessageManager.js#L206-L211

Please revert to `resolveId`.

```suggestion
    const resolvedId = this.resolveId(role);
    await this.client.api.guilds[this.guild.id].roles[resolvedId].delete({ reason });
    this.client.actions.GuildRoleDelete.handle({ guild_id: this.guild.id, role_id: resolvedId });
```

Also update JSDocs to match those changes.

src/structures/Role.js
```diff
@@ -406,10 +406,8 @@ class Role extends Base {
    *   .then(deleted => console.log(`Deleted role ${deleted.name}`))
    *   .catch(console.error);
    */
-  async delete(reason) {
-    await this.client.api.guilds[this.guild.id].roles[this.id].delete({ reason });
-    this.client.actions.GuildRoleDelete.handle({ guild_id: this.guild.id, role_id: this.id });
-    return this;
+  delete(reason) {
+    return this.guild.roles.delete(this.id, reason);
```
kyranet, line 410: Alongside the other suggestion:
```suggestion
  async delete(reason) {
    await this.guild.roles.delete(this.id, reason);
    return this;
```

kyranet (changes requested):

src/managers/RoleManager.js
Jiralite: This looks better! I have done this and updated the relevant typings.

iCrawl (approved):

kyranet (approved):

thebigwolfy (approved):

SpaceEEC (approved):

PR #6835, opened 2021-10-14
fix(MessageComponentInteraction): component getter is not null anymore

chrdek:
__Please describe the changes this PR makes and why it should be merged:__

This refers issue #6481 - The component getter of a MessageComponentInteraction cannot be null.
Removed null coal. from both TS typings/main getter of the component.

(following  init commit of d6e6244)

&nbsp;
__Status and versioning classification:__
- This includes updates on typings index.d.ts and  MessageComponentInteraction core structures.
- Testing layer remains -as-is.

PR #7099, opened 2021-12-11
types(MessageComponentInteraction): Ensure `component` is not `null`

Jiralite:
`<MessageComponentInteraction>.component` cannot be `null` and was addressed in #6835. However, #6867 reverted this.

I have also narrowed the received component structure if one already narrowed the interaction type to a `ButtonInteraction` or a `SelectMenuInteraction`. Previously, either of the two would yield either structures but now, they will yield their own structures.

### DIFF
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -448,6 +448,13 @@ export class BitField<S extends string, N extends number | bigint = number> {
 
 export class ButtonInteraction<Cached extends CacheType = CacheType> extends MessageComponentInteraction<Cached> {
   private constructor(client: Client, data: RawMessageButtonInteractionData);
+  public readonly component: CacheTypeReducer<
+    Cached,
+    MessageButton,
+    APIButtonComponent,
+    MessageButton | APIButtonComponent,
+    MessageButton | APIButtonComponent
+  >;
   public componentType: 'BUTTON';
   public inGuild(): this is ButtonInteraction<'present'>;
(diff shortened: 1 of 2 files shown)


kyranet (approved):

typings/index.d.ts
```diff
@@ -448,6 +448,13 @@ export class BitField<S extends string, N extends number | bigint = number> {
 
 export class ButtonInteraction<Cached extends CacheType = CacheType> extends MessageComponentInteraction<Cached> {
   private constructor(client: Client, data: RawMessageButtonInteractionData);
+  public readonly component: CacheTypeReducer<
+    Cached,
+    MessageButton,
+    APIButtonComponent,
+    MessageButton | APIButtonComponent,
```
vladfrangu, line 455: When can it be a raw component? We always build it, right?

Jiralite: If an interaction is received from a guild the client is not in, it will be a "raw" interaction. The `message` property will be as-is without any parsing by discord.js, which means the `components` on it will also be raw:

https://github.com/discordjs/discord.js/blob/ee91f5a19cdcf15f17a80fc9a25a382b7cbcde9f/src/structures/MessageComponentInteraction.js#L23-L27

vladfrangu (approved):

SpaceEEC (approved):

iCrawl (approved):

PR #7104, opened 2021-12-12
feat(GuildMember): Add timeouts

Jiralite:
## Description

- https://github.com/discord/discord-api-docs/pull/4075

## Checklist

- [x] Add `communication_disabled_until` property in the `GuildMember` class
- [x] Support modifying `communication_disabled_until` in `GuildMemberManager#edit()`
  - [x] Add helper method(s) to `GuildMember`
- [x] Add `MODERATE_MEMBERS` permission
- [x] Add necessary checks to permission helpers

## Classification

- Typings have been updated
- This pull request changes the library's interface

### DIFF
--- a/src/managers/GuildMemberManager.js
+++ b/src/managers/GuildMemberManager.js
@@ -226,6 +226,7 @@ class GuildMemberManager extends CachedManager {
   /**
    * Data that can be resolved to a Date object for timeouts. This can be:
    * * A number in seconds
+   * * A DateResolvable
    * @typedef {DateResolvable} TimeoutDateResolvable
    */
 
--- a/src/managers/GuildMemberManager.js
+++ b/src/managers/GuildMemberManager.js
@@ -224,6 +224,12 @@ class GuildMemberManager extends CachedManager {
   }
 
   /**
+   * Data that can be resolved to a Date object for timeouts. This can be:
+   * * A number in seconds
+   * @typedef {DateResolvable} TimeoutDateResolvable
+   */
+
+  /**
    * The data for editing a guild member.
    * @typedef {Object} GuildMemberEditData
    * @property {?string} [nick] The nickname to set for the member
@@ -232,7 +238,7 @@ class GuildMemberManager extends CachedManager {
    * @property {boolean} [deaf] Whether or not the member should be deafened
    * @property {GuildVoiceChannelResolvable|null} [channel] Channel to move the member to
    * (if they are connected to voice), or `null` if you want to disconnect them from voice
-   * @property {Date|number|null} [communicationDisabledUntil] The date, timestamp, or time in seconds
+   * @property {TimeoutDateResolvable|null} [communicationDisabledUntil] The date, timestamp, or time in seconds
    * for the member's communication to be disabled until. Provide `null` to remove the timeout.
    */
 
--- a/src/managers/GuildMemberManager.js
+++ b/src/managers/GuildMemberManager.js
@@ -239,7 +239,7 @@ class GuildMemberManager extends CachedManager {
    * @property {boolean} [deaf] Whether or not the member should be deafened
    * @property {GuildVoiceChannelResolvable|null} [channel] Channel to move the member to
    * (if they are connected to voice), or `null` if you want to disconnect them from voice
-   * @property {TimeoutDateResolvable|null} [communicationDisabledUntil] The date, timestamp, or time in seconds
+   * @property {TimeoutDateResolvable|null} [communicationDisabledUntil] The date, timestamp, or time in milliseconds
    * for the member's communication to be disabled until. Provide `null` to remove the timeout.
    */
 
@@ -274,13 +274,14 @@ class GuildMemberManager extends CachedManager {
       if (_data.communicationDisabledUntil === null) {
         _data.communication_disabled_until = null;
       } else {
-        let date = new Date(_data.communicationDisabledUntil);
-        if (date.getUTCFullYear() <= 2015) {
-          // Assume seconds
-          date = new Date(Date.now() + _data.communicationDisabledUntil * 1000);
+        if (
+          typeof _data.communicationDisabledUntil === 'number' &&
+          _data.communicationDisabledUntil < SnowflakeUtil.EPOCH
+        ) {
+          _data.communicationDisabledUntil += Date.now();
         }
 
-        _data.communication_disabled_until = date.toISOString();
+        _data.communication_disabled_until = new Date(_data.communicationDisabledUntil).toISOString();
       }
     }
 
--- a/src/managers/GuildMemberManager.js
+++ b/src/managers/GuildMemberManager.js
@@ -232,6 +232,8 @@ class GuildMemberManager extends CachedManager {
    * @property {boolean} [deaf] Whether or not the member should be deafened
    * @property {GuildVoiceChannelResolvable|null} [channel] Channel to move the member to
    * (if they are connected to voice), or `null` if you want to disconnect them from voice
+   * @property {number|null} [communicationDisabledUntil] The timestamp
+   * for the member's communication to be disabled until. Provide `null` to remove the timeout.
    */
 
   /**
@@ -260,6 +262,7 @@ class GuildMemberManager extends CachedManager {
       _data.channel = undefined;
     }
     _data.roles &&= _data.roles.map(role => (role instanceof Role ? role.id : role));
+    _data.communication_disabled_until = _data.communicationDisabledUntil;
     let endpoint = this.client.api.guilds(this.guild.id);
     if (id === this.client.user.id) {
       const keys = Object.keys(_data);
--- a/src/managers/GuildMemberManager.js
+++ b/src/managers/GuildMemberManager.js
@@ -262,7 +262,7 @@ class GuildMemberManager extends CachedManager {
       _data.channel = undefined;
     }
     _data.roles &&= _data.roles.map(role => (role instanceof Role ? role.id : role));
-    _data.communication_disabled_until = _data.communicationDisabledUntil;
+    _data.communication_disabled_until = new Date(_data.communicationDisabledUntil).toISOString();
     let endpoint = this.client.api.guilds(this.guild.id);
     if (id === this.client.user.id) {
       const keys = Object.keys(_data);
--- a/src/managers/GuildMemberManager.js
+++ b/src/managers/GuildMemberManager.js
@@ -232,7 +232,7 @@ class GuildMemberManager extends CachedManager {
    * @property {boolean} [deaf] Whether or not the member should be deafened
    * @property {GuildVoiceChannelResolvable|null} [channel] Channel to move the member to
    * (if they are connected to voice), or `null` if you want to disconnect them from voice
-   * @property {number|null} [communicationDisabledUntil] The timestamp or time in seconds
+   * @property {Date|number|null} [communicationDisabledUntil] The date, timestamp or time in seconds
    * for the member's communication to be disabled until. Provide `null` to remove the timeout.
    */
 
--- a/src/managers/GuildMemberManager.js
+++ b/src/managers/GuildMemberManager.js
@@ -232,7 +232,7 @@ class GuildMemberManager extends CachedManager {
    * @property {boolean} [deaf] Whether or not the member should be deafened
    * @property {GuildVoiceChannelResolvable|null} [channel] Channel to move the member to
    * (if they are connected to voice), or `null` if you want to disconnect them from voice
-   * @property {Date|number|null} [communicationDisabledUntil] The date, timestamp or time in seconds
+   * @property {Date|number|null} [communicationDisabledUntil] The date, timestamp, or time in seconds
    * for the member's communication to be disabled until. Provide `null` to remove the timeout.
    */
 
--- a/src/managers/GuildMemberManager.js
+++ b/src/managers/GuildMemberManager.js
@@ -263,10 +263,8 @@ class GuildMemberManager extends CachedManager {
     }
     _data.roles &&= _data.roles.map(role => (role instanceof Role ? role.id : role));
 
-    if (typeof _data.communicationDisabledUntil !== 'undefined') {
-      _data.communication_disabled_until =
-        _data.communicationDisabledUntil && new Date(_data.communicationDisabledUntil).toISOString();
-    }
+    _data.communication_disabled_until =
+      _data.communicationDisabledUntil && new Date(_data.communicationDisabledUntil).toISOString();
 
     let endpoint = this.client.api.guilds(this.guild.id);
     if (id === this.client.user.id) {
--- a/src/managers/GuildMemberManager.js
+++ b/src/managers/GuildMemberManager.js
@@ -262,7 +262,8 @@ class GuildMemberManager extends CachedManager {
       _data.channel = undefined;
     }
     _data.roles &&= _data.roles.map(role => (role instanceof Role ? role.id : role));
-    _data.communication_disabled_until = new Date(_data.communicationDisabledUntil).toISOString();
+    _data.communication_disabled_until =
+      _data.communicationDisabledUntil === null ? null : new Date(_data.communicationDisabledUntil).toISOString();
     let endpoint = this.client.api.guilds(this.guild.id);
     if (id === this.client.user.id) {
       const keys = Object.keys(_data);
--- a/src/managers/GuildMemberManager.js
+++ b/src/managers/GuildMemberManager.js
@@ -277,7 +277,7 @@ class GuildMemberManager extends CachedManager {
         let date = new Date(_data.communicationDisabledUntil);
         if (date.getUTCFullYear() <= 2015) {
           // Assume seconds
-          date = new Date(date.setTime(Date.now() + date.getTime() * 1000));
+          date = new Date(Date.now() + _data.communicationDisabledUntil * 1000);
         }
 
         _data.communication_disabled_until = date.toISOString();
--- a/src/managers/GuildMemberManager.js
+++ b/src/managers/GuildMemberManager.js
@@ -270,10 +270,7 @@ class GuildMemberManager extends CachedManager {
         let date = new Date(_data.communicationDisabledUntil);
         if (date.getUTCFullYear() <= 2015) {
           // Assume seconds
-          const dateFromSeconds = new Date();
-          date = new Date(
-            dateFromSeconds.setUTCSeconds(dateFromSeconds.getUTCSeconds() + _data.communicationDisabledUntil),
-          );
+          date = new Date(date.setTime(Date.now() + date.getTime() * 1000));
         }
 
         _data.communication_disabled_until = date.toISOString();
--- a/src/managers/GuildMemberManager.js
+++ b/src/managers/GuildMemberManager.js
@@ -232,7 +232,7 @@ class GuildMemberManager extends CachedManager {
    * @property {boolean} [deaf] Whether or not the member should be deafened
    * @property {GuildVoiceChannelResolvable|null} [channel] Channel to move the member to
    * (if they are connected to voice), or `null` if you want to disconnect them from voice
-   * @property {number|null} [communicationDisabledUntil] The timestamp
+   * @property {number|null} [communicationDisabledUntil] The timestamp or time in seconds
    * for the member's communication to be disabled until. Provide `null` to remove the timeout.
    */
 
@@ -262,8 +262,24 @@ class GuildMemberManager extends CachedManager {
       _data.channel = undefined;
     }
     _data.roles &&= _data.roles.map(role => (role instanceof Role ? role.id : role));
-    _data.communication_disabled_until =
-      _data.communicationDisabledUntil === null ? null : new Date(_data.communicationDisabledUntil).toISOString();
+
+    if (typeof _data.communicationDisabledUntil !== 'undefined') {
+      if (_data.communicationDisabledUntil === null) {
+        _data.communication_disabled_until = null;
+      } else {
+        let date = new Date(_data.communicationDisabledUntil);
+        if (date.getUTCFullYear() <= 2015) {
+          // Assume seconds
+          const dateFromSeconds = new Date();
+          date = new Date(
+            dateFromSeconds.setUTCSeconds(dateFromSeconds.getUTCSeconds() + _data.communicationDisabledUntil),
+          );
+        }
+
+        _data.communication_disabled_until = date.toISOString();
+      }
+    }
+
     let endpoint = this.client.api.guilds(this.guild.id);
     if (id === this.client.user.id) {
       const keys = Object.keys(_data);
--- a/src/managers/GuildMemberManager.js
+++ b/src/managers/GuildMemberManager.js
@@ -224,6 +224,17 @@ class GuildMemberManager extends CachedManager {
   }
 
   /**
+   * The data for editing a guild member.
+   * @typedef {Object} GuildMemberEditData
+   * @property {?string} [nick] The nickname to set for the member
+   * @property {Collection<Snowflake, Role>|RoleResolvable[]} [roles] The roles or role ids to apply
+   * @property {boolean} [mute] Whether or not the member should be muted
+   * @property {boolean} [deaf] Whether or not the member should be deafened
+   * @property {GuildVoiceChannelResolvable|null} [channel] Channel to move the member to
+   * (if they are connected to voice), or `null` if you want to disconnect them from voice
+   */
+
+  /**
    * Edits a member of the guild.
    * <info>The user must be a member of the guild</info>
    * @param {UserResolvable} user The member to edit
--- a/src/managers/GuildMemberManager.js
+++ b/src/managers/GuildMemberManager.js
@@ -224,13 +224,6 @@ class GuildMemberManager extends CachedManager {
   }
 
   /**
-   * Data that can be resolved to a Date object for timeouts. This can be:
-   * * A number in seconds
-   * * A DateResolvable
-   * @typedef {DateResolvable} TimeoutDateResolvable
-   */
-
-  /**
    * The data for editing a guild member.
    * @typedef {Object} GuildMemberEditData
    * @property {?string} [nick] The nickname to set for the member
@@ -239,8 +232,8 @@ class GuildMemberManager extends CachedManager {
    * @property {boolean} [deaf] Whether or not the member should be deafened
    * @property {GuildVoiceChannelResolvable|null} [channel] Channel to move the member to
    * (if they are connected to voice), or `null` if you want to disconnect them from voice
-   * @property {TimeoutDateResolvable|null} [communicationDisabledUntil] The date, timestamp, or time in milliseconds
-   * for the member's communication to be disabled until. Provide `null` to remove the timeout.
+   * @property {DateResolvable|null} [communicationDisabledUntil] The date or timestamp
+   * for the member's communication to be disabled until. Provide `null` to enable communication again.
    */
 
   /**
@@ -271,18 +264,8 @@ class GuildMemberManager extends CachedManager {
     _data.roles &&= _data.roles.map(role => (role instanceof Role ? role.id : role));
 
     if (typeof _data.communicationDisabledUntil !== 'undefined') {
-      if (_data.communicationDisabledUntil === null) {
-        _data.communication_disabled_until = null;
-      } else {
-        if (
-          typeof _data.communicationDisabledUntil === 'number' &&
-          _data.communicationDisabledUntil < SnowflakeUtil.EPOCH
-        ) {
-          _data.communicationDisabledUntil += Date.now();
-        }
-
-        _data.communication_disabled_until = new Date(_data.communicationDisabledUntil).toISOString();
-      }
+      _data.communication_disabled_until =
+        _data.communicationDisabledUntil && new Date(_data.communicationDisabledUntil).toISOString();
     }
 
     let endpoint = this.client.api.guilds(this.guild.id);
--- a/src/structures/BaseGuildVoiceChannel.js
+++ b/src/structures/BaseGuildVoiceChannel.js
@@ -69,6 +69,8 @@ class BaseGuildVoiceChannel extends GuildChannel {
   get joinable() {
     if (!this.viewable) return false;
     const permissions = this.permissionsFor(this.client.user);
+    if (!permissions) return false;
+
     // This flag allows joining even if timed out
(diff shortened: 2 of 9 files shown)


src/managers/GuildMemberManager.js
```diff
@@ -223,6 +223,19 @@ class GuildMemberManager extends CachedManager {
...
+  /**
+   * The data for editing a guild member.
+   * @typedef {Object} GuildMemberEditData
+   * @property {?string} [nick] The nickname to set for the member
+   * @property {Collection<Snowflake, Role>|RoleResolvable[]} [roles] The roles or role ids to apply
+   * @property {boolean} [mute] Whether or not the member should be muted
+   * @property {boolean} [deaf] Whether or not the member should be deafened
+   * @property {GuildVoiceChannelResolvable|null} [channel] Channel to move the member to
+   * (if they are connected to voice), or `null` if you want to disconnect them from voice
+   * @property {Date|number|null} [communicationDisabledUntil] The date, timestamp or time in seconds
```
suneettipirneni, line 235: This should be applied to the `GuildMemberEditData` within the typings.

suneettipirneni (changes requested):

src/managers/GuildMemberManager.js
Jiralite: Added with b13faaedf58dfec6f28751077cd1b53a0ec0ec63!

almeidx: Oxford comma?
```suggestion
   * @property {Date|number|null} [communicationDisabledUntil] The date, timestamp, or time in seconds
```

suneettipirneni: I was referring to the interface `GuildMemberEditData`

almeidx: yeah, I don't know why I commented on your review instead of making my own 😅

almeidx, line 235: Oxford comma
```suggestion
   * @property {Date|number|null} [communicationDisabledUntil] The date, timestamp, or time in seconds
```

Jiralite: Added with f48b948718056de871495a2de6473a186468ed86!

typings/index.d.ts
```diff
@@ -4971,7 +4975,8 @@ export type PermissionString =
   | 'CREATE_PRIVATE_THREADS'
   | 'USE_EXTERNAL_STICKERS'
   | 'SEND_MESSAGES_IN_THREADS'
-  | 'START_EMBEDDED_ACTIVITIES';
+  | 'START_EMBEDDED_ACTIVITIES'
+  | 'MODERATE MEMBERS';
```
ToasterSticks, line 4979: Shouldn't this be `MODERATE_MEMBERS`? Clicked the wrong button last time 😄

Jiralite: Oh no!

ToasterSticks: Anyway

thebigwolfy (approved):

src/managers/GuildMemberManager.js
```diff
@@ -223,6 +223,25 @@ class GuildMemberManager extends CachedManager {
     return data.reduce((col, member) => col.set(member.user.id, this._add(member, cache)), new Collection());
   }
 
+  /**
+   * Data that can be resolved to a Date object for timeouts. This can be:
+   * * A number in seconds
+   * @typedef {DateResolvable} TimeoutDateResolvable
```
ImRodry, line 229: The way the code is set atm this accepts a date object, as well as an ISO8601 string (see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/Date)
```suggestion
   * Data that can be resolved into a Date object for timeouts. This can be:
   * * A number in seconds
   * * An {@link https://en.wikipedia.org/wiki/ISO_8601 ISO 8601} string
   * * A {@link Date} object
   * @typedef {DateResolvable} TimeoutDateResolvable
```

src/managers/GuildMemberManager.js
```diff
@@ -249,6 +268,21 @@ class GuildMemberManager extends CachedManager {
...
     _data.roles &&= _data.roles.map(role => (role instanceof Role ? role.id : role));
+
+    if (typeof _data.communicationDisabledUntil !== 'undefined') {
+      if (_data.communicationDisabledUntil === null) {
+        _data.communication_disabled_until = null;
+      } else {
+        let date = new Date(_data.communicationDisabledUntil);
+        if (date.getUTCFullYear() <= 2015) {
+          // Assume seconds
+          date = new Date(date.setTime(Date.now() + date.getTime() * 1000));
```
ImRodry, line 279: What exactly is this doing?

src/structures/GuildMember.js
```diff
@@ -337,6 +346,22 @@ class GuildMember extends Base {
...
+   * for the member's communication to be disabled until. Provide `null` to remove the timeout.
+   * @param {string} [reason] The reason for this timeout.
+   * @returns {Promise<GuildMember>}
+   * @example
+   * // Timeout a guild member for 5 minutes
+   * guildMember.timeout(300, 'They deserved it')
+   *   .then(console.log)
+   *   .catch(console.error);
+   */
+  timeout(timeout, reason) {
```
ImRodry, line 361: This is the client name, don't we usually follow the API name?
```suggestion
  disableCommunication(timeout, reason) {
```

typings/index.d.ts
```diff
@@ -4475,12 +4478,15 @@ export type GuildFeatures =
   | 'PRIVATE_THREADS'
   | 'ROLE_ICONS';
 
+export type TimeoutDateResolvable = DateResolvable;
```
ImRodry, line 4481: What's the point of this type?

ImRodry (changes requested):

src/managers/GuildMemberManager.js
Jiralite: That's exactly what this type definition explains. What you have done is repeat ISO 8601 and the date object twice.

src/managers/GuildMemberManager.js
Jiralite: This adds seconds. Is there an issue?

ImRodry: Seems a bit messy the way you check if the year is before 2015 and add the seconds if it is, couldn’t this be very prone to errors? I think you’d be better off checking if the number is smaller than 28 days in seconds (the max according to the pr) and add that if so

src/managers/GuildMemberManager.js
ImRodry: Well then this should still be fixed, but disregard the previous suggestion (yes this is done elsewhere, see https://discord.js.org/#/docs/main/stable/typedef/EmojiIdentifierResolvable for example)
```suggestion
   * Data that can be resolved to a Date object for timeouts. This can be:
   * * A number in seconds
   * * A DateResolvable
   * @typedef {DateResolvable} TimeoutDateResolvable
```

src/managers/GuildMemberManager.js
Jiralite: What if the limit is changed? Maybe the limit is dropped to the client's: 1 week. Maybe it's increased 30 days, or 50, or 60... Hardcoding a limit isn't an option I would want to take. Also, what's very prone to errors here?

typings/index.d.ts
Jiralite: The same reason as other types that do this, I suppose. Future-proofing if the types need to be amended and also to match with the relevant JSDoc.

ImRodry: I don’t recall seeing many other types that do this and consistency with JSDoc has never been a rule to my knowledge. I think this type could be easily removed

src/managers/GuildMemberManager.js
ImRodry: The limit will not increase because the current one is the 32-bit integer limit afaik (and as documented in the PR). If it decreases then the value will still work and you’d like Discord throw that error instead. This code seems unnecessarily complex and doesn’t feel accurate at all to me

src/structures/GuildMember.js
```diff
@@ -337,6 +347,30 @@ class GuildMember extends Base {
...
 
+  /**
+   * Data that can be resolved to a Date object for timeouts. This can be:
+   * * A number in milliseconds
+   * * A DateResolvable
+   * @typedef {DateResolvable} TimeoutDateResolvable
+   */
+
+  /**
+   * Timeouts this guild member.
```
ThaTiemsz, line 358:
```suggestion
   * Times this guild member out.
```

src/structures/GuildMember.js
```diff
@@ -337,6 +347,30 @@ class GuildMember extends Base {
...
+   * @param {TimeoutDateResolvable|null} timeout The date, timestamp, or time in milliseconds
+   * for the member's communication to be disabled until. Provide `null` to remove the timeout.
+   * @param {string} [reason] The reason for this timeout.
+   * @returns {Promise<GuildMember>}
+   * @example
+   * // Timeout a guild member for 5 minutes
```
ThaTiemsz, line 364:
```suggestion
   * // Time a guild member out for 5 minutes
```

ThaTiemsz (changes requested):

src/managers/GuildMemberManager.js
```diff
@@ -223,6 +223,19 @@ class GuildMemberManager extends CachedManager {
...
+  /**
+   * The data for editing a guild member.
+   * @typedef {Object} GuildMemberEditData
+   * @property {?string} [nick] The nickname to set for the member
+   * @property {Collection<Snowflake, Role>|RoleResolvable[]} [roles] The roles or role ids to apply
+   * @property {boolean} [mute] Whether or not the member should be muted
+   * @property {boolean} [deaf] Whether or not the member should be deafened
+   * @property {GuildVoiceChannelResolvable|null} [channel] Channel to move the member to
+   * (if they are connected to voice), or `null` if you want to disconnect them from voice
+   * @property {DateResolvable|null} [communicationDisabledUntil] The date or timestamp
```
ImRodry, line 235: Why does this method not support TimeoutDateResolvable?

ckohen: Aren't DateResolvable and TimeoutDateResolvable the same thing now anyways?

ckohen: Documentation needs updating, but that makes total sense then, you can't disable communcation until a duration, but you can timeout for a duration

ckohen: IMO, the shortcut methods are a bridge between the client implementation and the API implementation, but then again, this isn't up to me or you. Its true, this is a little different from a large majority of the library, but I personally really like it.

ImRodry: It's such an easy implementation that I don't see why such an inconsistency would be introduced, it makes a lot of sense imo

ImRodry: Well then how about naming the property timeout in the edit method too? If it was done in the GuildMember method I don't see why not, right?

ckohen: Again, like I said earlier, IMO it is the *helpers* responsbility to bridge the gap, not the manager, the manager should remain fairly low level.

ImRodry: Usually it's in the manager where all the processing of the data happens, not the helpers

src/managers/GuildMemberManager.js
```diff
@@ -249,6 +262,12 @@ class GuildMemberManager extends CachedManager {
       _data.channel = undefined;
     }
     _data.roles &&= _data.roles.map(role => (role instanceof Role ? role.id : role));
+
+    if (typeof _data.communicationDisabledUntil !== 'undefined') {
+      _data.communication_disabled_until =
+        _data.communicationDisabledUntil && new Date(_data.communicationDisabledUntil).toISOString();
```
vladfrangu, line 268: Does this set it to null if we provide null? Asking to be sure

src/structures/GuildMember.js
```diff
@@ -159,6 +171,15 @@ class GuildMember extends Base {
     return this.joinedTimestamp ? new Date(this.joinedTimestamp) : null;
   }
 
+  /**
+   * The time this member's timeout will be removed
+   * @type {?Date}
+   * @readonly
+   */
+  get communicationDisabledUntil() {
+    return this.communicationDisabledUntilTimestamp && new Date(this.communicationDisabledUntilTimestamp);
```
vladfrangu, line 180: Do we usually do it like this?

src/util/Permissions.js
```diff
@@ -144,6 +145,7 @@ Permissions.FLAGS = {
   USE_EXTERNAL_STICKERS: 1n << 37n,
   SEND_MESSAGES_IN_THREADS: 1n << 38n,
   START_EMBEDDED_ACTIVITIES: 1n << 39n,
+  MODERATE_MEMBERS: 1n << 40n,
```
vladfrangu, line 148: If its not added to -types yet, can you please add it? 🙏

vladfrangu (changes requested):

src/managers/GuildMemberManager.js
Jiralite: Yep! `null && true` for example will return `null`. In this case, providing `null && new Date(null)` will return `null`.

src/structures/GuildMember.js
Jiralite: Currently, we do a ternary: `property ? new Date(property) : null`. However, as of #7108, it will change to what is here.

src/util/Permissions.js
Jiralite: See https://github.com/discordjs/discord-api-types/pull/235!

src/structures/GuildMember.js
meister03: Won't it be better to use a object instead of multiply parameters, would be more consistence to `.ban` ...

Jiralite: The API calls are usually done by the manager by resolving the respective id, collecting all the JSON query data in an object followed by additional headers (`reason` typically). Sometimes, there won't be a `reason` and helper methods do not need an `id` - they are called from the structure itself.

This results in the cases of helper methods _usually_ having 2 parameters: the data to send across and the reason. If there's no reason, then it's not included. If there's no need for any data, then there won't be any parameters. The [`Guild`](https://discord.js.org/#/docs/main/main/class/Guild) class is a prime example of this.

`<GuildMember>.ban()` is the inconsistent one: providing a `reason` was part of the JSON query data which is likely why it was included in the type definition, but this is deprecated as of a while ago and has since been moved out of the JSON query data sent to Discord (but still in the type definition). It is simply still reused across the library.

I would also argue that anything that doesn't follow this pattern is inconsistent, such as [`GuildStickerManager#create()`](https://discord.js.org/#/docs/main/main/class/GuildStickerManager?scrollTo=create) which seems off to be honest... 4 possible parameters instead of 1 object with a `reason` parameter?

Jiralite: > This is the client name, don't we usually follow the API name?

I'd like to think that helper methods help bridge the gap between the client and the API. That's why I went for "timeout" here. It's known to be the timeout feature so I'm sure that's what people expect to see! Welcome to hear feedback (especially from maintainers) about whether to change it back.

ImRodry: There are other methods such as `setRateLimitPerUser` that have a different name in the API and the client and yet Discord.js went with the API name for all methods and properties so I think it would make sense to do the same here, as I don't recall this being done anywhere else other than in JSDocs and this would introduce a weird inconsistency in the library

xHyroM:
Timeouts already released, merge when?

suneettipirneni:
> Timeouts already released, merge when?

docs need to be merged first

src/structures/GuildMember.js
kyranet: We should probably name this either `timeoutUntil`, `setTimeoutUntil`, or `disableCommunicationUntil` since it takes a date or an UNIX timestamp, and not a duration.

But I'm more inclined for `disableCommunicationUntil` because timeouts remind us of the `setTimeout` global, which takes a callback and a duration, which neither of which this method takes.

Jiralite: Just to be clear, you're moving towards "disableCommunicationUntil" with no duration?

kyranet: If we take a duration, `timeout` and `disableCommunicationFor` are fine.

Also, why not both? cc @discordjs/the-big-4

ImRodry:
Upstream PR landed 🎉

src/structures/GuildMember.js
Jiralite: Just to be clear, this helper method already accepts a date and a time in milliseconds (duration).

donzee529:
Can it be merged now? Timeout docs were released..

ImRodry:
Patience is a virtue

xHyroM:
Why is there a `disableCommunicationUntil` method and a `timeout` method? Wouldn't it be better if they were named similarly? There will be unnecessary chaos in this. Better would be `disableCommunicationUntil` and `disableCommunication` or `timeoutUntil` `timeout`.

xHyroM (commented):
Why is there a `communicationDisabledUntil` property and a `timeout` method? Wouldn't it be better if they were named similarly? There will be unnecessary chaos in this. Better would be `communicationDisabledUntil` and `disableCommunication` or `timeoutUntil` `timeout`.

src/structures/GuildMember.js
vladfrangu: Honestly, I'm also much more of a fan of the `disableCommunicationUntil` name for a method that acts just as a shortcut, and something like `timeout` or w/e that gives you the support for passing in a duration

xcheez1 (commented):
It doesnt work, i get this error:
/app/node_modules/discord.js/src/managers/GuildManager.js:111
guild instanceof GuildMember ||
TypeError: Right-hand side of 'instanceof' is not callable

Jiralite:
That error isn't relevant to this pull request. File a new issue if you are able to reproduce it with relevant details.

DTrombett:
> app/node_modules/discord.js/src/managers/GuildManager.js:111

I don't see this file was touched in this PR...

src/structures/BaseGuildVoiceChannel.js
```diff
@@ -68,8 +68,11 @@ class BaseGuildVoiceChannel extends GuildChannel {
    */
   get joinable() {
     if (!this.viewable) return false;
-    if (!this.permissionsFor(this.client.user).has(Permissions.FLAGS.CONNECT, false)) return false;
-    return true;
+    const permissions = this.permissionsFor(this.client.user);
```
kyranet, line 71: This needs a check for nullability, `permissionsFor(User)`'s return is nullable.

As for why the old code didn't account for that... it's a bug and would have thrown an error if `guild.me` was not cached.

src/structures/GuildChannel.js
```diff
@@ -510,6 +510,10 @@ class GuildChannel extends Channel {
     if (this.client.user.id === this.guild.ownerId) return true;
     const permissions = this.permissionsFor(this.client.user);
     if (!permissions) return false;
+    // This flag allows managing even if timed out
```
kyranet, line 513: Nit:
```suggestion

    // This flag allows managing even if timed out
```

kyranet (changes requested):

xcheez1:
i did not touch the file, i changed what you said to change and this happen...

xcheez1:
> > app/node_modules/discord.js/src/managers/GuildManager.js:111
> 
> I don't see that file was touched in this PR... I'm testing it right now, however.
> 
> Edit: it works as expected and the member is successfully timed out!

can you send me what you made?

kyranet:
@xcheez1 please do not scope creep unrelated PRs, go to the [official Discord server](https://discord.gg/djs) instead.

Further off-topic messages will be hidden.

src/structures/GuildChannel.js
ImRodry: Can admins even be timed out?

vladfrangu: Yes, yes they can

kyranet: Yes, as a matter of a fact, @vladfrangu timed me out in Sapphire, which has ADMINISTRATOR.

src/structures/GuildMember.js
Jiralite: See bab6adc64bbe57ba6cbb74ff3ba69d5de96a82c4!

ethhaqn:
`(node:8124) DeprecationWarning: Passing strings for the URL or the icon's URL for MessageEmbed#setAuthor is deprecated. Pass a sole object instead.`
Were some other changes made that weren't discussed?

ImRodry:
@ethhaqn please stick to the topic of the PR, questions like that should be taken to the Discord server

ethhaqn:
@ImRodry As far as I'm aware, this is part of the PR. Since I installed this version of the code, I've been getting this warning.

Glazelf:
> @ImRodry As far as I'm aware, this is part of the PR. Since I installed this version of the code, I've been getting a warning.

Check the "Files Changed" tab here and point to where this was changed? :p

vladfrangu (approved):

src/structures/GuildMember.js
```diff
@@ -337,6 +346,39 @@ class GuildMember extends Base {
...
+   * @returns {Promise<GuildMember>}
+   * @example
+   * // Time a guild member out for 5 minutes
+   * guildMember.timeout(5 * 60 * 1000, 'They deserved it')
+   *   .then(console.log)
+   *   .catch(console.error);
+   */
+  timeout(timeout, reason) {
+    if (timeout !== null) timeout += Date.now();
+    return this.edit({ communicationDisabledUntil: timeout }, reason);
```
kyranet, line 379: We can do this and simplify a lot:
```suggestion
    return this.disableCommunicationUntil(Date.now() + timeout, reason);
```

kyranet (changes requested):

src/structures/BaseGuildVoiceChannel.js
```diff
@@ -68,8 +68,13 @@ class BaseGuildVoiceChannel extends GuildChannel {
...
+    if (!permissions) return false;
+
+    // This flag allows joining even if timed out
+    if (permissions.has(Permissions.FLAGS.ADMINISTRATOR, false)) return true;
+    if (!permissions.has(Permissions.FLAGS.CONNECT, false)) return false;
+    return this.guild.me.communicationDisabledUntilTimestamp < Date.now();
```
Syjalo, line 77:
```suggestion
    const { communicationDisabledUntilTimestamp } = this.guild.me;
    return communicationDisabledUntilTimestamp ? communicationDisabledUntilTimestamp < Date.now() : true;
```
If `communicationDisabledUntilTimestamp` will be `null` it will return `false`, but if `communicationDisabledUntilTimestamp` is `null` you aren't timeouted and can join the channel.

Syjalo (changes requested):

src/structures/BaseGuildVoiceChannel.js
ImRodry: Or
```suggestion
    return this.guild.me.communicationDisabledUntilTimestamp ?? 0 < Date.now();
```

iCrawl (approved):

src/structures/BaseGuildVoiceChannel.js
Jiralite: I'm confused by these suggestions... can you explain why this change is needed?

kyranet: They're not needed at all, `null < Date.now()` is always executed as `0 < Date.now()` because the `<` operator converts both expressions to a number primitive, and since `null` has neither `@@toPrimitive` nor `toValue`... it falls back to default behavior: `Number(null)`, which is effectively `0`.

So doing `null < Date.now()` is equivalent to doing `0 < Date.now()`.

src/structures/GuildMember.js
Jiralite: ESLint is fine here, but Prettier wanted to remove the brackets: eb57f31030ea83e17cd69460b0a79f531fb6aa05

SpaceEEC (approved):

Jiralite:
I believe that's all permission checks done. Feel free to re-review!

src/structures/ThreadChannel.js
```diff
@@ -423,13 +423,19 @@ class ThreadChannel extends Channel {
...
+    const permissions = this.permissionsFor(this.client.user);
+    if (!permissions) return false;
+    // This flag allows joining even if timed out
+    if (permissions.has(Permissions.FLAGS.ADMINISTRATOR, false)) return true;
+
     return (
       !this.archived &&
       !this.joined &&
-      this.permissionsFor(this.client.user)?.has(
+      this.permissionsFor(this.client.user).has(
```
kyranet, line 434:
```suggestion
      permissions.has(
```

kyranet (changes requested):

kyranet (approved):

Mysterious-Dev (approved):

src/structures/GuildChannel.js
TheCreativeGod: You can still send messages with the API but not with the client if you have ADMINISTRATOR and you are timed out, and for I can see even if you are timed out in this scenario you can still connect to voice channels.

src/structures/VoiceChannel.js
```diff
@@ -46,7 +46,12 @@ class VoiceChannel extends BaseGuildVoiceChannel {
    * @readonly
    */
   get speakable() {
-    return this.permissionsFor(this.client.user).has(Permissions.FLAGS.SPEAK, false);
+    const permissions = this.permissionsFor(this.client.user);
+    if (!permissions) return false;
+    // This flag allows speaking even if timed out
+    if (permissions.has(Permissions.FLAGS.ADMINISTRATOR, false)) return true;
+    if (!permissions.has(Permissions.FLAGS.SPEAK, false)) return false;
+    return this.guild.me.communicationDisabledUntilTimestamp < Date.now();
```
ImRodry, line 54: How are timeouts related to speaking in a voice channel if you can’t join it to begin with?

Sasiko: Administrator can receive timeout  however they are still able to join VC

ImRodry: In that case they can also speak in the vc

src/structures/BaseGuildVoiceChannel.js
ImRodry: Sorry, my suggestion was wrong, but you should still consider @Syjalo's suggestion because otherwise, this will return false when a user is not timed out, which is misleading

src/structures/ThreadChannel.js
```diff
@@ -423,13 +423,19 @@ class ThreadChannel extends Channel {
...
     return (
       !this.archived &&
       !this.joined &&
-      this.permissionsFor(this.client.user)?.has(
+      permissions.has(
         this.type === 'GUILD_PRIVATE_THREAD' ? Permissions.FLAGS.MANAGE_THREADS : Permissions.FLAGS.VIEW_CHANNEL,
         false,
-      )
+      ) &&
+      this.guild.me.communicationDisabledUntilTimestamp < Date.now()
```
ImRodry, line 438: You can always join threads, regardless of whether you're timed out or not

ImRodry (changes requested):
Could GuildMember#moderatable be added?

src/structures/BaseGuildVoiceChannel.js
TheCreativeGod, line 77: You could do an `ADMINISTRATOR` check and then `return` if the you (don't) have `timeout` and have `CONNECT`
```suggestion
    return this.guild.me.communicationDisabledUntilTimestamp < Date.now() && permissions.has(Permissions.FLAGS.CONNECT, false);
```

src/structures/ThreadChannel.js
```diff
@@ -439,7 +445,12 @@ class ThreadChannel extends Channel {
    * @readonly
    */
   get manageable() {
-    return this.permissionsFor(this.client.user)?.has(Permissions.FLAGS.MANAGE_THREADS, false);
+    const permissions = this.permissionsFor(this.client.user);
+    if (!permissions) return false;
+    // This flag allows managing even if timed out
+    if (permissions.has(Permissions.FLAGS.ADMINISTRATOR, false)) return true;
+    if (!permissions.has(Permissions.FLAGS.MANAGE_THREADS, false)) return false;
+    return this.guild.me.communicationDisabledUntilTimestamp < Date.now();
```
TheCreativeGod, line 453: Can't we use a single return for `MANAGE_THREADS` and `timeouts` here?
```suggestion
    return this.guild.me.communicationDisabledUntilTimestamp < Date.now() && permissions.has(Permissions.FLAGS.MANAGE_THREADS, false);
```

src/structures/VoiceChannel.js
TheCreativeGod, line 54: We can use a single return here for `SPEAK` and the `timeout`
```suggestion
    return this.guild.me.communicationDisabledUntilTimestamp < Date.now() && permissions.has(Permissions.FLAGS.SPEAK, false);
```

TheCreativeGod (changes requested):
See if we can use a single return adding the `timeout` with specific permissions

src/structures/ThreadChannel.js
Jiralite: Thank you for pointing this out! I have verified this and removed this check.

Jiralite:
> See if we can use a single return adding the `timeout` with specific permissions

Sure, but Prettier would want these to be on multiple lines anyway. I'll do it nonetheless.

mafineeek (approved):

src/structures/GuildMember.js
```diff
@@ -255,6 +275,15 @@ class GuildMember extends Base {
     return this.manageable && this.guild.me.permissions.has(Permissions.FLAGS.BAN_MEMBERS);
   }
 
+  /**
+   * Whether this member is moderatable by the client user
+   * @type {boolean}
+   * @readonly
+   */
+  get moderatable() {
+    return this.manageable && (this.guild.me?.permissions.has(Permissions.FLAGS.MODERATE_MEMBERS) ?? false);
```
ImRodry, line 284:
```suggestion
    if (!this.guild.me) return new Error('GUILD_UNCACHED_ME')
    return this.manageable && this.guild.me.permissions.has(Permissions.FLAGS.MODERATE_MEMBERS);
```

PR #7106, opened 2021-12-13
Change default

thebigwolfy:
Change default remove all roles to the member.
*Sorry i am french ^^*

### DIFF
--- a/src/managers/GuildMemberRoleManager.js
+++ b/src/managers/GuildMemberRoleManager.js
@@ -162,7 +162,7 @@ class GuildMemberRoleManager extends DataManager {
   }
 
   /**
-   * Sets the roles applied to the member.
+   * Sets the roles applied to the member
    * @param {Collection<Snowflake, Role>|RoleResolvable[]} roles The roles or role ids to apply
    * @param {string} [reason] Reason for applying the roles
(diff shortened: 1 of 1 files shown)


Jiralite (commented):
Calling `GuildMember#roles#set()` with no arguments seems ambiguous to me. For example:

```JavaScript
<GuildMember>.roles.set([]);
```

By passing an empty array, this clearly indicates to me that we're setting no roles. But doing this:

```JavaScript
<GuildMember>.roles.set();
```

This indicates to me that a value is being set about roles and that value is not _nothing_. It's _something_, but it's not obvious what it is.

I'm not in agreement with this change, personally.

kyranet:
Between bad linting and what @Jiralite said, there's also a breaking change introduced by returning the instance instead of a clone, as seen below:

https://github.com/discordjs/discord.js/blob/2ce244b502293c07ff1c3e929b2afb861e341d88/src/structures/GuildMember.js#L287-L289

https://github.com/discordjs/discord.js/blob/2ce244b502293c07ff1c3e929b2afb861e341d88/src/managers/GuildMemberManager.js#L262-L264

Due to those reasons, I'm inclined to close this PR, it does not offer a significant change that improves discord.js's API.

PR #7114, opened 2021-12-15
chore(GuildMemberManager): Update nick path

Jiralite:
To edit the client's nickname in a guild, it was using [this path](https://discord.com/developers/docs/resources/guild#modify-current-user-nick) which has been marked as deprecated. The correct path to use is right above it.


vladfrangu (approved):

iCrawl (approved):

SpaceEEC (approved):
And here I thought this edge-case finally is no longer. :(

PR #7115, opened 2021-12-15
fix(Message): xable check ephemeral

muchnameless:
 - added checks for ephemeral in `Message#editable`, `Message#deletable` and `Message#pinnable`
 - since `Message#guild` is now a getter it should be more performant to check `!this.guildId` instead of `!this.guild` (I don't know if this change is considered out of scope for this PR)

### DIFF
--- a/src/structures/Message.js
+++ b/src/structures/Message.js
@@ -584,7 +584,10 @@ class Message extends Base {
    */
   get editable() {
     const precheck = Boolean(
-      this.author.id === this.client.user.id && !deletedMessages.has(this) && (!this.guild || this.channel?.viewable),
+      this.author.id === this.client.user.id &&
+        !deletedMessages.has(this) &&
+        !this.flags.has(MessageFlags.FLAGS.EPHEMERAL) &&
+        (!this.guildId || this.channel?.viewable),
     );
     // Regardless of permissions thread messages cannot be edited if
     // the thread is locked.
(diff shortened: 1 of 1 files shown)


src/structures/Message.js
```diff
@@ -584,7 +584,10 @@ class Message extends Base {
    */
   get editable() {
     const precheck = Boolean(
-      this.author.id === this.client.user.id && !deletedMessages.has(this) && (!this.guild || this.channel?.viewable),
+      this.author.id === this.client.user.id &&
+        !deletedMessages.has(this) &&
+        !this.flags.has(MessageFlags.FLAGS.EPHEMERAL) &&
```
vladfrangu, line 589: Ephemeral messages can be edited

vladfrangu (changes requested):

src/structures/Message.js
muchnameless: not via `Message#edit` though

muchnameless: one can argue either way, imo if you edit ephemeral messages via `Interaction#editReply` you wouldn't use this getter since you don't need any permissions for it but it's up to you maintainers whether to include this check

kyranet: > not via `Message#edit` though

What if we make an `InteractionMessage` that takes the interaction and makes the message editable? 🤔

Would also remove the need to use not-so-intuitive methods.

iCrawl:
Based on feedback given above we decided to go against this for the 13.x iteration, should definitely be revisited in 14 though.

muchnameless:
should I reopen a reduced version of this which only adds the fix to `deletable` and `pinnable` since those will always error for ephemeral messages?

PR #7120, opened 2021-12-16
types: make channel types a lot stricter

kyranet:

Brings a new set of strict channel types based on @ImRodry's typeguards from #6957. This change also makes sure that the set of possible channel types are strict, explicit, and matching the actual classes obtained during runtime.

### DIFF
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -386,7 +386,7 @@ export class BaseGuildEmoji extends Emoji {
   public requiresColons: boolean | null;
 }
 
-export class BaseGuildTextChannel extends TextBasedChannel(GuildChannel) {
+export class BaseGuildTextChannel extends TextBasedChannelMixin(GuildChannel) {
   protected constructor(guild: Guild, data?: RawGuildChannelData, client?: Client, immediatePatch?: boolean);
   public defaultAutoArchiveDuration?: ThreadAutoArchiveDuration;
   public messages: MessageManager;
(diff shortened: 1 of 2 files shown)


vladfrangu (approved):

typings/index.d.ts
```diff
@@ -521,9 +521,9 @@ export abstract class Channel extends Base {
   public id: Snowflake;
   public readonly partial: false;
   public type: keyof typeof ChannelTypes;
-  public delete(): Promise<Channel>;
-  public fetch(force?: boolean): Promise<Channel>;
-  public isText(): this is TextBasedChannels;
+  public delete(): Promise<this>;
+  public fetch(force?: boolean): Promise<this>;
```
ImRodry, line 525: Didn’t @suneettipirneni say something about returning `this` and how it didn’t make sense in this situation?

suneettipirneni: It makes sense here, because if you narrow a type the promise on this class has the narrowed return type as well.

ImRodry: Oh alright, when does it not make sense then?

suneettipirneni: The other PR was more of an aesthetic thing, the type that was discussed before wasn't narrowed. But even in the case before it's fine.

ImRodry: Alright, thanks

iCrawl (approved):

SpaceEEC (approved):

PR #7130, opened 2021-12-20
This should solve the API hiccup instance

Itz-Hyperz:
This should solve the API hiccup instance, and allow the library to recover from it without throwing fatal errors.

https://cdn.hyperz.net/main/4OB9Ap.png

It's a bad practice to not have common instances like this checked for when the API has a hiccup like it did a few moments ago and crashed bots that ran this function. This change will implement a fix for the above error. I am not the only person who got this error either.

Other: https://discord.com/channels/222078108977594368/824411059443204127/922550844195496047

### DIFF
--- a/src/client/actions/MessageCreate.js
+++ b/src/client/actions/MessageCreate.js
@@ -11,7 +11,7 @@ class MessageCreateAction extends Action {
     const channel = this.getChannel(data);
     if (channel) {
       if (!channel.isText()) return {};
-
+      if(channel.messages == undefined) return;
       const existing = channel.messages.cache.get(data.id);
       if (existing) return { message: existing };
       const message = channel.messages._add(data);


kyranet:
That's what the check above is for, `isText()` checks for whether or not it has messages:

https://github.com/discordjs/discord.js/blob/b9b60a37b33aed2f85ef1e6cb6ae54481e946b92/src/structures/Channel.js#L146-L148

Itz-Hyperz:
That clearly isn't checking for when it returns this error:

https://cdn.hyperz.net/main/4OB9Ap.png

Itz-Hyperz:
or something else somewhere is breaking this, as it exits the process entirely on that error over a hiccup that wasn't in your control.

kyranet:
In your screenshot, line 13 is the following:

```js
const existing = channel.messages.cache.get(data.id);
```

However, in current main, it's this:

```js
if (!channel.isText()) return {};
```

The `isText` check was added in #6919, which should be released in [v13.3.1](https://github.com/discordjs/discord.js/releases/tag/13.3.1).

Itz-Hyperz:
ahhh, thank you.

Itz-Hyperz:
sorry to bother.

PR #7132, opened 2021-12-20
types(CategoryChannel): `createChannel` should default to a text channel

suneettipirneni:

Fixes issue where if `type` isn't explicitly set in the options the return type wouldn't be a text channel.

### DIFF
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -503,7 +503,7 @@ export class CategoryChannel extends GuildChannel {
   /** @deprecated See [Self-serve Game Selling Deprecation](https://support-dev.discord.com/hc/en-us/articles/4414590563479) for more information */
   public createChannel(
     name: string,
-    options: CategoryCreateChannelOptions & { type: 'GUILD_STORE' },
+    options: CategoryCreateChannelOptions & { type: 'GUILD_STORE' | ChannelTypes.GUILD_STORE },
   ): Promise<StoreChannel>;
 
   public createChannel(
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -494,19 +494,21 @@ export type CategoryChannelTypes = ExcludeEnum<
 export class CategoryChannel extends GuildChannel {
   public readonly children: Collection<Snowflake, Exclude<NonThreadGuildBasedChannel, CategoryChannel>>;
   public type: 'GUILD_CATEGORY';
+
+  public createChannel<T extends Exclude<CategoryChannelTypes, 'GUILD_STORE'>>(
+    name: string,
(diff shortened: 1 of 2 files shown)


typings/index.test-d.ts
```diff
@@ -854,9 +854,7 @@ declare const categoryChannel: CategoryChannel;
   expectType<Promise<NewsChannel>>(categoryChannel.createChannel('name', { type: 'GUILD_NEWS' }));
   expectDeprecated(categoryChannel.createChannel('name', { type: 'GUILD_STORE' }));
   expectType<Promise<StageChannel>>(categoryChannel.createChannel('name', { type: 'GUILD_STAGE_VOICE' }));
-  expectType<Promise<TextChannel | VoiceChannel | NewsChannel | StoreChannel | StageChannel>>(
-    categoryChannel.createChannel('name', {}),
-  );
+  expectType<Promise<TextChannel>>(categoryChannel.createChannel('name', {}));
```
ImRodry, line 857: Why not just this?
```suggestion
  expectType<Promise<TextChannel>>(categoryChannel.createChannel('name'));
```

ImRodry (approved):
Tested and this works fine, just a small nit

typings/index.test-d.ts
suneettipirneni: That option has always been required hence the empty object, but I presume it could be made optional

ImRodry: it has always been optional iirc, better to test both actually, given the TS error

ImRodry: @suneettipirneni what do you think? ^^

iCrawl (approved):

kyranet (approved):

typings/index.d.ts
```diff
@@ -494,20 +494,16 @@ export type CategoryChannelTypes = ExcludeEnum<
...
     name: string,
     options: CategoryCreateChannelOptions & { type: T },
   ): Promise<MappedChannelCategoryTypes[T]>;
-
+  /** @deprecated See [Self-serve Game Selling Deprecation](https://support-dev.discord.com/hc/en-us/articles/4414590563479) for more information */
   public createChannel(
     name: string,
-    options: CategoryCreateChannelOptions,
-  ): Promise<TextChannel | VoiceChannel | NewsChannel | StoreChannel | StageChannel>;
+    options: CategoryCreateChannelOptions & { type: 'GUILD_STORE' },
```
vladfrangu, line 504: Small nit but this won't match the number

vladfrangu (approved):

SpaceEEC (approved):

iCrawl:
This needs a rebase.

PR #7143, opened 2021-12-24
fix: Remove deprecation warning on `Message#cleanContent`

### LINKED ISSUE #7142: Message.cleanContent sends a deprecation warning
### Issue description

The `Util.removeMentions` method is deprecated, but it is still used by `Util.cleanContent` which is used by [`Message.cleanContent`](https://github.com/discordjs/discord.js/blob/stable/src/structures/Message.js#L473)

So even though `Message.cleanContent` is not deprecated, it sends a deprecation warning on any call on a message that contains a mention.

Steps to reproduce:
1. Run the included code (event that logs the `cleanContent` of all recieved messages)
2. Send a message with a user mention

### Code sample

```typescript
const Discord = require('discord.js')
const client = new Discord.Client({ intents: ['GUILDS', 'GUILD_MESSAGES'] })

client
  .on('ready', () => {
    console.log(`Logged in as ${client.user.tag}!`)
  })
  .on('warn', console.log)
  .on('messageCreate', msg => {
    console.log(msg.cleanContent)
  })

client.login('token')
```

### discord.js version

13.4.0

### Node.js version

16.13.1

### Operating system

Windows

Medium (should be fixed soon)

No Partials

GUILDS, GUILD_MESSAGES

_No response_

Jiralite:
Resolves #7142.

### DIFF
--- a/src/util/Util.js
+++ b/src/util/Util.js
@@ -529,11 +529,7 @@ class Util extends null {
    * @deprecated Use {@link BaseMessageOptions#allowedMentions} instead.
    */
   static removeMentions(str) {
-    return Util._removeMentions(str, false);
-  }
-
-  static _removeMentions(str, fromCleanContent) {
-    if (!deprecationEmittedForRemoveMentions && !fromCleanContent) {
+    if (!deprecationEmittedForRemoveMentions) {
       process.emitWarning(
         'The Util.removeMentions method is deprecated. Use MessageOptions#allowedMentions instead.',
         'DeprecationWarning',
@@ -542,6 +538,10 @@ class Util extends null {
       deprecationEmittedForRemoveMentions = true;
     }
 
+    return Util._removeMentions(str);
+  }
+
+  static _removeMentions(str) {
     return str.replaceAll('@', '@\u200b');
   }
 
@@ -559,15 +559,15 @@ class Util extends null {
         const id = input.replace(/<|!|>|@/g, '');
         if (channel.type === 'DM') {
           const user = channel.client.users.cache.get(id);
-          return user ? Util._removeMentions(`@${user.username}`, true) : input;
+          return user ? Util._removeMentions(`@${user.username}`) : input;
         }
 
         const member = channel.guild.members.cache.get(id);
         if (member) {
-          return Util._removeMentions(`@${member.displayName}`, true);
+          return Util._removeMentions(`@${member.displayName}`);
         } else {
           const user = channel.client.users.cache.get(id);
-          return user ? Util._removeMentions(`@${user.username}`, true) : input;
+          return user ? Util._removeMentions(`@${user.username}`) : input;
         }
       })
       .replace(/<#[0-9]+>/g, input => {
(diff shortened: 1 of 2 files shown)


iCrawl (approved):

vladfrangu (approved):

Seblor:
I'm not familiar with the inner workings of DJS, but from an outsider's point of view, this looks like a stopgap solution.

The `Util.removeMentions` method is being deprecated because of the introduction of `allowedMentions` (https://github.com/discordjs/discord.js/pull/6530). So maybe the code from `Util.removeMentions` should be moved to the `Message.cleanContent` getter, or the `removeMentions` should not be deprecated.

Jiralite:
`Message#cleanContent` calls `Util.cleanContent()` which calls `Util.removeMentions()` which sanitises input to prevent mentionability by a simple `str.replaceAll('@', '@\u200b')`. Once `Util.removeMentions()` is removed, this replacement will no longer exist. `Message#cleanContent` will still call `Util.cleanContent()` and will still act as it always has - the only difference being that mentions are not sanitised.

The reason for this is in case people send clean content as a new message which may cause `@everyone` or `@here` injections. Since `allowedMentions` exist, that should be the preferred method to prevent mentions in your message.

Seblor:
That makes sense, thank you for the clarification.

ImRodry:
Shouldn't cleanContent still return @Rodry#4020 instead of <@240875059953139714>? Maybe it shouldn't only remove mentions but also make them readable outside of Discord

vladfrangu:
The only thing this PR does is prevent the deprecation warning from showing in .cleanContent, not remove the replace to username behavior

ImRodry:
> Once `Util.removeMentions()` is removed, this replacement will no longer exist. `Message#cleanContent` will still call `Util.cleanContent()` and will still act as it always has - the only difference being that mentions are not sanitised.

I was commenting on this

vladfrangu:
It means that if you cleanContent a member mention, whose display name is `everyone`, you will get an `@everyone` in your chat instead of `@\u200beveryone`.

src/util/Util.js
```diff
@@ -529,7 +529,11 @@ class Util extends null {
    * @deprecated Use {@link BaseMessageOptions#allowedMentions} instead.
    */
   static removeMentions(str) {
-    if (!deprecationEmittedForRemoveMentions) {
+    return Util._removeMentions(str, false);
+  }
+
+  static _removeMentions(str, fromCleanContent) {
+    if (!deprecationEmittedForRemoveMentions && !fromCleanContent) {
       process.emitWarning(
```
kyranet, line 537: This should be something like this:
```js
  static removeMentions(str) {
    if (!deprecationEmittedForRemoveMentions) {
      deprecationEmittedForRemoveMentions = true;
      process.emitWarning(
        // ...
      );
    }

    return Util._removeMentions(str);
  }

  static _removeMentions(str) {
    return str.replaceAll('@', '@\u200b');
  }
```

Likewise in `cleanContent`, remove the second parameter.

kyranet (changes requested):

src/util/Util.js
Jiralite: > Also, does the lack of JSDoc automatically make this private/hidden from the docs?

Yep! By having no JSDoc, it's simply not parsed into the .json.

kyranet (approved):

PR #7144, opened 2021-12-24
fix(Permissions): toArray shouldn't check admin

### LINKED ISSUE #6267: Permissions#missing returns all permissions in v13

In v13, Permissions#missing returns all permissions, not only the provided ones.

```js
guild.me.permissions.missing(Permissions.FLAGS.ADMINISTRATOR)
```
This should only return the `ADMINISTRATOR` permission, since it's the only one passed as argument, in v12 works as intended but v13 seems to return all permissions, i'm not sure if those are all of them but it returns permissions that the bot actually has.

v12 returns:
image

v13 returns:
image

This also happens if you provide an array as the argument.

**Further details:**

- discord.js version: `13.0.0-dev.t1627831263.774f5b7`
- Node.js version: `14.17.3`
- Operating system: `Windows 10`
- Priority this issue should have – please be realistic and elaborate if possible: `Low`

**Relevant client options:**

- partials: none
- gateway intents: `GUILD_MEMBERS, GUILD_MESSAGES, GUILDS`
- other: none

DTrombett:

When a member is an administrator, he does have all permissions but it's still possible to add/remove other permissions. Although it doesn't have any effect, the bitfield is still different and the `toArray()` method should return the permissions of the bitfield.
This can be done by passing `false` to the `toArray()` call.
This change also resolves #6267, which seems a regression added in #5911.

I've also considered defaulting the `checkAdmin` parameter to false in `Permissions#toArray()` but it would be inconsistent with other methods and useless as I don't see any use case where an array with all the flags is needed for an administrator and could be confused with a person with really every permission.

### DIFF
--- a/src/util/Permissions.js
+++ b/src/util/Permissions.js
@@ -31,7 +31,7 @@ class Permissions extends BitField {
    * @returns {string[]}
    */
   missing(bits, checkAdmin = true) {
-    return checkAdmin && this.has(this.constructor.FLAGS.ADMINISTRATOR) ? [] : super.missing(bits, checkAdmin);
+    return checkAdmin && this.has(this.constructor.FLAGS.ADMINISTRATOR) ? [] : super.missing(bits);
   }
 
   /**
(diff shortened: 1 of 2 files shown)


vladfrangu (approved):

kyranet (approved):
We seriously need to make unit tests for this, it's a really troublesome class.

lucalves (approved):

iCrawl (approved):

Issue #7145, opened 2021-12-25
isTimeouted or Timeouted


rodycouto:
### Feature

whether the guild member is under timeout or not
it can be return a Boolean result

### Ideal solution or implementation

```js
if (<GuildMember>.timeouted) {
   return message.channel.send('This member is already muted.')
}
```

ImRodry:
What’s the use case for this? Can’t you just check if communicationDisabledUntilTimestamp isn’t null? cc @iShibi

iCrawl:
No because discord doesn't reset the value once the timeout is run out

PR #7146, opened 2021-12-25
feat(GuildMember): add method to check timeout

iShibi:

closes: #7145

### DIFF
--- a/src/structures/GuildMember.js
+++ b/src/structures/GuildMember.js
@@ -306,11 +306,8 @@ class GuildMember extends Base {
    * Whether this member is currently timed out
    * @returns {boolean}
    */
-  isTimedOut() {
-    return (
-      typeof this.communicationDisabledUntilTimestamp === 'number' &&
-      this.communicationDisabledUntilTimestamp > Date.now()
-    );
+  isCommunicationDisabled() {
+    return this.communicationDisabledUntilTimestamp > Date.now();
   }
 
   /**
--- a/src/structures/GuildMember.js
+++ b/src/structures/GuildMember.js
@@ -303,7 +303,7 @@ class GuildMember extends Base {
   }
 
   /**
-   * Whether this member is already under time out
+   * Whether this member is currently timed out
    * @returns {boolean}
    */
   isTimedOut() {
--- a/src/structures/GuildMember.js
+++ b/src/structures/GuildMember.js
@@ -303,6 +303,17 @@ class GuildMember extends Base {
   }
 
   /**
+   * Whether this member is already under time out
+   * @returns {boolean}
+   */
+  isTimedOut() {
+    return (
+      typeof this.communicationDisabledUntilTimestamp === 'number' &&
+      this.communicationDisabledUntilTimestamp > Date.now()
+    );
+  }
+
+  /**
    * Returns `channel.permissionsFor(guildMember)`. Returns permissions for a member in a guild channel,
    * taking into account roles and permission overwrites.
    * @param {GuildChannelResolvable} channel The guild channel to use as context
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -1123,7 +1123,7 @@ export class GuildMember extends PartialTextBasedChannel(Base) {
   public deleteDM(): Promise<DMChannel>;
   public displayAvatarURL(options?: ImageURLOptions): string;
   public edit(data: GuildMemberEditData, reason?: string): Promise<GuildMember>;
-  public isTimedOut(): this is GuildMember & {
+  public isCommunicationDisabled(): this is GuildMember & {
     communicationDisabledUntilTimestamp: number;
     communicationDisabledUntil: Date;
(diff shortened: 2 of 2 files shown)


src/structures/GuildMember.js
```diff
@@ -302,6 +302,17 @@ class GuildMember extends Base {
     return this.manageable && (this.guild.me?.permissions.has(Permissions.FLAGS.MODERATE_MEMBERS) ?? false);
   }
 
+  /**
+   * Whether this member is already under time out
```
Jiralite, line 306:
```suggestion
   * Whether this member is currently timed out
```

src/structures/GuildMember.js
```diff
@@ -302,6 +302,17 @@ class GuildMember extends Base {
...
+   * @returns {boolean}
+   */
+  isTimedOut() {
+    return (
+      typeof this.communicationDisabledUntilTimestamp === 'number' &&
+      this.communicationDisabledUntilTimestamp > Date.now()
+    );
+  }
```
Jiralite, line 314: The `typeof` check is not needed here, simply `this.communicationDisabledUntilTimestamp > Date.now()` is enough. `null` will convert to `0`, so this is fine.

Such a simple check also makes me wonder if this should really be implemented, to be honest.

Jiralite (changes requested):

advaith1:
why is it not a getter?

Jiralite:
Getters cannot help narrow types:

https://github.com/discordjs/discord.js/blob/4e7601244d33dfa3a9d9d5eac600b187d37afbda/typings/index.d.ts#L1126-L1129

src/structures/GuildMember.js
iShibi: > The `typeof` check is not needed here, simply `this.communicationDisabledUntilTimestamp > Date.now()` is enough. `null` will convert to `0`, so this is fine.

true, but it's always better to be on the safe side

> Such a simple check also makes me wonder if this should really be implemented, to be honest.

concise, succinct, self-descriptive, and of course narrows down `communicationDisabledUntilTimestamp` and `communicationDisabledUntil` as non-null properties like you pointed out below.
<sub>(I may or may not have used a dictionary while writing this down 🐱)</sub>

src/structures/GuildMember.js
```diff
@@ -302,6 +302,17 @@ class GuildMember extends Base {
     return this.manageable && (this.guild.me?.permissions.has(Permissions.FLAGS.MODERATE_MEMBERS) ?? false);
   }
 
+  /**
+   * Whether this member is currently timed out
+   * @returns {boolean}
+   */
+  isTimedOut() {
+    return (
+      typeof this.communicationDisabledUntilTimestamp === 'number' &&
```
ImRodry, line 311: You could probably remove this check since the next line will return false if communicationDisabledUntilTimestamp is null

almeidx: https://github.com/discordjs/discord.js/pull/7146#discussion_r775127198

src/structures/GuildMember.js
```diff
@@ -302,6 +302,17 @@ class GuildMember extends Base {
     return this.manageable && (this.guild.me?.permissions.has(Permissions.FLAGS.MODERATE_MEMBERS) ?? false);
   }
 
+  /**
+   * Whether this member is currently timed out
+   * @returns {boolean}
+   */
+  isTimedOut() {
```
S222em, line 309: Shouldnt this be named `isCommunicationDisabled`? There was a discussion about the naming in #7104

src/structures/GuildMember.js
ImRodry: oops

src/structures/GuildMember.js
kyranet: > but it's always better to be on the safe side

It's literally ECMAScript's spec and it cannot change (without breaking the web, which they actively avoid).

src/structures/GuildMember.js
iShibi: thanks 👍

kyranet (approved):

iCrawl (approved):

typings/index.d.ts
```diff
@@ -1122,6 +1123,10 @@ export class GuildMember extends PartialTextBasedChannel(Base) {
   public deleteDM(): Promise<DMChannel>;
   public displayAvatarURL(options?: ImageURLOptions): string;
   public edit(data: GuildMemberEditData, reason?: string): Promise<GuildMember>;
+  public isCommunicationDisabled(): this is GuildMember & {
+    communicationDisabledUntilTimestamp: number;
+    communicationDisabledUntil: Date;
```
vladfrangu, line 1128:
```suggestion
    readonly communicationDisabledUntil: Date;
```

vladfrangu (dismissed):

SpaceEEC (approved):

PR #7149, opened 2021-12-25
refactor: Default *URL methods to animated hash

Jiralite:
If a user's avatar icon is animated (for example), Discord lets us know about that by prefixing the avatar hash with `a_`. If it is not animated, the aforementioned prefix is not present. discord.js enforces everything to be static, but I believe the proper approach would be to default to the avatar hash provided.

This pull request will yield results as such:

```JavaScript
// Let "animatedUser" be a User object with an animated avatar hash
// Let "staticUser" be a User object with a regular avatar hash

animatedUser.displayAvatarURL();
// https://cdn.discordapp.com/avatars/user_id/a_hash.gif

animatedUser.displayAvatarURL({ forceStatic: true });
// https://cdn.discordapp.com/avatars/user_id/a_hash.webp

animatedUser.displayAvatarURL({ format: "png" });
// https://cdn.discordapp.com/avatars/user_id/a_hash.gif

animatedUser.displayAvatarURL({ format: "png", forceStatic: true });
// https://cdn.discordapp.com/avatars/user_id/a_hash.png

// ...

staticUser.displayAvatarURL();
// https://cdn.discordapp.com/avatars/user_id/hash.webp

staticUser.displayAvatarURL({ forceStatic: true });
// https://cdn.discordapp.com/avatars/user_id/hash.webp

staticUser.displayAvatarURL({ format: "png" });
// https://cdn.discordapp.com/avatars/user_id/hash.png

staticUser.displayAvatarURL({ format: "png", forceStatic: true });
// https://cdn.discordapp.com/avatars/user_id/hash.png
```

You'll see in the third animated example that `animatedUser` with the desired format of "png" was in fact still animated. This is because we default to if the avatar hash is animated. If the avatar hash is static, it will change appropriately.

In addition to this, I removed the "Asset" CDN as I could not find a reference where this was being used and `StaticImageURLOptions` has been removed as I believe this to be no longer relevant.


almeidx (commented):
Some methods use an if statement to check if the hash is not null (`if (!<hash>) return`) and  others that simply return with a condition `return <hash> && getCdnURL(<hash>)`

Should probably be more consistent with this

src/util/Constants.js
```diff
@@ -14,68 +14,50 @@ exports.WSCodes = {
...
- * Options for static Image URLs.
- * @typedef {Object} StaticImageURLOptions
- * @property {string} [format='webp'] One of `webp`, `png`, `jpg`, `jpeg`.
+ * Options for image URLs.
+ * @typedef {Object} ImageURLOptions
+ * @property {string} [format] One of `webp`, `png`, `jpg`, `jpeg`.
+ * @property {boolean} [forceStatic=false] If `true`, the format will be as specified or `webp`.
  * @property {number} [size] One of `16`, `32`, `56`, `64`, `96`, `128`, `256`, `300`, `512`, `600`, `1024`, `2048`,
- * `4096`
+ * `4096`.
```
ImRodry, line 36: You could probably make a typedef with the sizes that would make it easier to update in the future

src/util/Constants.js
```diff
@@ -14,68 +14,50 @@ exports.WSCodes = {
...
- * @property {boolean} [dynamic=false] If true, the format will dynamically change to `gif` for animated avatars.
- */
-
-/**
- * Options for static Image URLs.
- * @typedef {Object} StaticImageURLOptions
- * @property {string} [format='webp'] One of `webp`, `png`, `jpg`, `jpeg`.
+ * Options for image URLs.
+ * @typedef {Object} ImageURLOptions
+ * @property {string} [format] One of `webp`, `png`, `jpg`, `jpeg`.
```
ImRodry, line 33: The default is still webp so that shouldn’t be removed
On that topic though, why don’t we change that to png?

src/util/Constants.js
```diff
@@ -14,68 +14,50 @@ exports.WSCodes = {
...
+ * @property {boolean} [forceStatic=false] If `true`, the format will be as specified or `webp`.
```
ImRodry, line 34: Why can’t this stay as dynamic and default to true? People could set it to false if they want the static image

ImRodry (changes requested):

rmunro-uts:
Looks like this is in line with the conversation we had in library-discussion back in November 👍

src/util/Constants.js
mdashlw: imo a default-false boolean is better than default true. though i think `force` is unnecessary and it can be called just `static`

Jiralite: `static` is a reserved keyword.

src/util/Constants.js
```diff
@@ -14,68 +14,50 @@ exports.WSCodes = {
...
 
-function makeImageUrl(root, { format = 'webp', size } = {}) {
+function makeImageUrl(root, { hash, format = 'webp', forceStatic = false, size } = {}) {
   if (!['undefined', 'number'].includes(typeof size)) throw new TypeError('INVALID_TYPE', 'size', 'number');
-  if (format && !AllowedImageFormats.includes(format)) throw new Error('IMAGE_FORMAT', format);
+  if (!AllowedImageFormats.includes(format)) throw new Error('IMAGE_FORMAT', format);
   if (size && !AllowedImageSizes.includes(size)) throw new RangeError('IMAGE_SIZE', size);
-  return `${root}.${format}${size ? `?size=${size}` : ''}`;
+  if (forceStatic) format ??= 'webp';
+  else if (hash?.startsWith('a_')) format = 'gif';
```
mdashlw, line 26: `format` cannot be `null` or `undefined` at this point, also not to repeat the default twice
```suggestion
  if (!forceStatic && hash?.startsWith('a_')) format = 'gif';
```

src/util/Constants.js
imranbarbhuiya: Why not using `steady`? It's a smaller word with the same meaning.

src/util/Constants.js
imranbarbhuiya: As discord api don't provide something like static or dynamic so why don't use format only and remove the `forceStatic` if anyone wants it in static form they can pass `png` or `webp` in format.

iCrawl (approved):

iCrawl:
This needs a rebase.

src/util/Constants.js
Jiralite: Have done this!

src/util/Constants.js
Jiralite: I don't believe .png is going to be the default option.

I removed the default option here because the output can be .gif. I'm not really sure how to document that as a default?

packages/discord.js/src/util/Constants.js
```diff
@@ -15,68 +15,67 @@ exports.WSCodes = {
   4014: 'DISALLOWED_INTENTS',
 };
 
-const AllowedImageFormats = ['webp', 'png', 'jpg', 'jpeg', 'gif'];
```
vladfrangu, line 18: This removal feels like a regression

vladfrangu (changes requested):

packages/discord.js/src/structures/User.js
```diff
@@ -183,9 +182,9 @@ class User extends Base {
    * @param {ImageURLOptions} [options={}] Options for the Image URL
    * @returns {?string}
    */
-  bannerURL({ format, size, dynamic } = {}) {
+  bannerURL(options = {}) {
     if (!this.banner) return this.banner;
-    return this.client.rest.cdn.Banner(this.id, this.banner, format, size, dynamic);
+    return this.client.rest.cdn.Banner(this.id, this.banner, options);
```
ImRodry, line 187: You can do the same thing you did on the other ones with `this.banner && …`

src/util/Constants.js
ImRodry: The advantages of using png as the default are obvious but I guess that’s not a topic for here
I think you should keep the default as webp and, on the forceStatic documentation, mention that, if omitted, that will change the format to gif if possible

packages/discord.js/src/structures/User.js
Jiralite: Done this!

vladfrangu (approved):

src/util/Constants.js
Jiralite: Should be done as of 911f7ad92d5922f12ccd35415aae901fef9aa106!

src/util/Constants.js
Jiralite: Which is why it's not resolved. (:

iCrawl (approved):

SpaceEEC (approved):

iCrawl:
This needs a rebase.

packages/discord.js/src/util/Constants.js
```diff
@@ -15,68 +15,76 @@ exports.WSCodes = {
...
+ * * `png`
+ * * `jpg`
+ * * `jpeg`
+ * @typedef {string} ImageFormat
+ */
+
+/**
+ * Options for image URLs.
+ * @typedef {Object} ImageURLOptions
+ * @property {string} [ImageFormat='webp'] An image format.
```
vaporox, line 60:
```suggestion
 * @property {ImageFormat} [format='webp'] An image format.
```

kyranet (approved):

PR #7153, opened 2021-12-26
refactor(MessageEmbed): Deprecate strings for `setAuthor()` (completely) and `setFooter()`

Jiralite:
Due to #7067, `setAuthor()` and `setFooter()` should only passing objects and not strings as per the builder's implementation. This pull request deprecates the passing of strings.

### DIFF
--- a/src/structures/MessageEmbed.js
+++ b/src/structures/MessageEmbed.js
@@ -4,6 +4,9 @@ const { RangeError } = require('../errors');
 const Util = require('../util/Util');
 
 let deprecationEmittedForSetAuthor = false;
+let deprecationEmittedForSetFooter = false;
+
+// TODO: Remove the deprecated code for `setAuthor()` and `setMethod()`.
 
 /**
  * Represents an embed in a message (image/video preview, rich embed, etc.)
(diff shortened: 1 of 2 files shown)


kyranet (approved):

ImRodry:
Why can't builders be changed to support the current implementation? I think it makes much more sense to do that instead

vladfrangu:
I'm not a fan of this required parameters provided 1 after the other with an optional object for optional parameters at all, and I don't know if the others are or aren't on the same page. What is each of those parameters?? You don't know by name, unless you read the docs or hover over the function definition in your editor, whereas with objects you know from the getgo what's what

vladfrangu (approved):

iCrawl (approved):

ImRodry:
VSCode actually introduced a feature recently with TS 4.5 that adds the name of the parameter to the function call so unless you have a variable called text and set that in the text parameter of the footer, it will display as `text: “whatever you want”` so you can easily know what parameter you’re passing. I think that it makes perfect sense to be able to just pass one string for setAuthor and setFooter because that’s probably what you want most of the time anyway

kyranet:
Yeah, try doing that on a PR review in GitHub or anywhere else that isn't powered by TypeScript.

ImRodry:
Well if you’re reviewing a PR you should probably know how the library works to begin with so that shouldn’t be an issue

vladfrangu:
That's missing the point that not everyone is using TypeScript, not everyone is using VSCode/IntelliJ, and that in the end an object doesn't cause any harm at all.

SpaceEEC (approved):

PR #7157, opened 2021-12-27
refactor: import timers, `process`, and `Buffer` from `node:`

kyranet:

Following [this Tweet](https://twitter.com/kt3k/status/1475391353375031301), this PR imports `process`, `Buffer`, and timers (`setTimeout`, `setInterval`, and `setImmediate`) from their respective modules.

**But why?**

tl;dr: *if Node.js recommends a way to do something over another, we abide to that, also known as following best practices.*

This has been asked several times in the server, while I do not know the very exact reason (Node.js developers aren't sharing many details in their PRs), I have read the same thing:

The imports tell bundlers (and other tools) where certain globals come from, and what kind they are, for example, Node.js's `setTimeout` isn't a compatible implementation of the DOM one, one returns a `NodeJS.Timer` and the other returns a `number`, however, bundlers can't tell if a package needs one or another. Importing timers from `node:timers` is a sensible approach to this, since it tells the bundler we're using Node.js's.

And now my theories...

It's possible that Node.js might look into deprecating some of the globals (specially those that aren't used in other platforms) in favour of importing from a package, this remains to be confirmed, but it's likely.

Also, there's a possibility they also recommend this change to protect libraries from any kind of global mutation, it only takes an inexperienced developer writing `process = something` to trigger a quite funny error to debug. Since the imports are const-assigned, we keep a reference of those global objects during the entire runtime, and as such, this is far less likely to happen, also with ESM, developers won't be able to change the exports of modules, so that's one problem that'll be fixed in the (far) future.

For the performance kids, because the globals are now `const` assigned, it's faster than pulling from the global, since the value cannot change.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -171,6 +171,30 @@
     "prefer-template": "error",
     "rest-spread-spacing": "error",
     "template-curly-spacing": "error",
-    "yield-star-spacing": "error"
+    "yield-star-spacing": "error",
+
+    "no-restricted-globals": [
+      "error",
+      {
+        "name": "Buffer",
+        "message": "Import Buffer from `node:buffer` instead of using the global"
+      },
+      {
+        "name": "process",
+        "message": "Import process from `node:process` instead of using the global"
+      },
+      {
+        "name": "setTimeout",
+        "message": "Import setTimeout from `node:timers` instead of using the global"
+      },
+      {
+        "name": "setInterval",
+        "message": "Import setInterval from `node:timers` instead of using the global"
+      },
+      {
+        "name": "setImmediate",
+        "message": "Import setImmediate from `node:timers` instead of using the global"
+      }
+    ]
   }
 }
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -177,23 +177,23 @@
       "error",
       {
         "name": "Buffer",
-        "message": "Import Buffer from `node:buffer` instead of using the global"
+        "message": "Import Buffer from `node:buffer` instead"
       },
       {
         "name": "process",
-        "message": "Import process from `node:process` instead of using the global"
+        "message": "Import process from `node:process` instead"
       },
       {
         "name": "setTimeout",
-        "message": "Import setTimeout from `node:timers` instead of using the global"
+        "message": "Import setTimeout from `node:timers` instead"
       },
       {
         "name": "setInterval",
-        "message": "Import setInterval from `node:timers` instead of using the global"
+        "message": "Import setInterval from `node:timers` instead"
       },
       {
         "name": "setImmediate",
-        "message": "Import setImmediate from `node:timers` instead of using the global"
+        "message": "Import setImmediate from `node:timers` instead"
       }
     ]
   }
--- a/test/webhooktest.js
+++ b/test/webhooktest.js
@@ -2,6 +2,7 @@
 
 const fs = require('node:fs');
 const path = require('node:path');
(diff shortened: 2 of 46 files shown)


iCrawl (approved):

test/webhooktest.js
```diff
@@ -2,6 +2,7 @@
 
 const fs = require('node:fs');
 const path = require('node:path');
+const { setTimeout: sleep } = require('node:timers');
```
muchnameless, line 5:
```suggestion
const { setTimeout: sleep } = require('node:timers/promises');
```

almeidx:
Might be worth enabling the [no-restricted-globals eslint rule](https://eslint.org/docs/rules/no-restricted-globals) and disable `process`, `Buffer`, etc.

.eslintrc.json
```diff
@@ -171,6 +171,30 @@
...
     "rest-spread-spacing": "error",
     "template-curly-spacing": "error",
-    "yield-star-spacing": "error"
+    "yield-star-spacing": "error",
+
+    "no-restricted-globals": [
+      "error",
+      {
+        "name": "Buffer",
+        "message": "Import Buffer from `node:buffer` instead of using the global"
```
almeidx, line 180:
```suggestion
        "message": "Import Buffer from `node:buffer` instead"
```

.eslintrc.json
```diff
@@ -171,6 +171,30 @@
...
+      },
+      {
+        "name": "process",
+        "message": "Import process from `node:process` instead of using the global"
```
almeidx, line 184:
```suggestion
        "message": "Import process from `node:process` instead"
```

.eslintrc.json
```diff
@@ -171,6 +171,30 @@
...
+      },
+      {
+        "name": "setTimeout",
+        "message": "Import setTimeout from `node:timers` instead of using the global"
```
almeidx, line 188:
```suggestion
        "message": "Import setTimeout from `node:timers` instead"
```

.eslintrc.json
```diff
@@ -171,6 +171,30 @@
...
+      },
+      {
+        "name": "setInterval",
+        "message": "Import setInterval from `node:timers` instead of using the global"
```
almeidx, line 192:
```suggestion
        "message": "Import setInterval from `node:timers` instead"
```

.eslintrc.json
```diff
@@ -171,6 +171,30 @@
...
+      },
+      {
+        "name": "setImmediate",
+        "message": "Import setImmediate from `node:timers` instead of using the global"
```
almeidx, line 196:
```suggestion
        "message": "Import setImmediate from `node:timers` instead"
```

almeidx (changes requested):

SpaceEEC (approved):

vladfrangu (approved):

blattersturm:
I'm still missing `clearTimeout` being imported here in files such as `GuildMemberManager.js`. This may work well in pure Node runtimes, but elsewhere this may lead to a mismatch.

kyranet:
Oh. Good catch, I'll make a PR in a bit.

blattersturm:
@kyranet have you managed to get this fixed yet? People are still [hitting issues](https://forum.cfx.re/t/discord-js-13-0-0/4778143/14?u=d-bubble) in our embedded runtime.

kyranet:
Yes, I just opened https://github.com/discordjs/discord.js/pull/7269 to fix that, sorry for the delay, I have been busy with exams 😅

PR #7174, opened 2021-12-31
docs(RoleManager): fix incorrect example

ceilingfans:
This PR fixes the example provided in `RoleManger#delete()` where `deleted` is used as a `Role` when it is `undefined`

### DIFF
--- a/packages/discord.js/src/managers/RoleManager.js
+++ b/packages/discord.js/src/managers/RoleManager.js
@@ -227,7 +227,7 @@ class RoleManager extends CachedManager {
    * @example
    * // Delete a role
    * guild.roles.delete('222079219327434752', 'The role needed to go')
-   *   .then(deleted => console.log(`Deleted role ${deleted.name}`))
+   *   .then(() => console.log('Deleted the role'))
    *   .catch(console.error);
    */
   async delete(role, reason) {


src/managers/RoleManager.js
```diff
@@ -227,7 +227,7 @@ class RoleManager extends CachedManager {
    * @example
    * // Delete a role
    * guild.roles.delete('222079219327434752', 'The role needed to go')
-   *   .then(deleted => console.log(`Deleted role ${deleted.name}`))
+   *   .then(() => console.log('Deleted the role'))
```
imranbarbhuiya, line 230: The role isn't undefined here. It's the role which got deleted.

vladfrangu: The manager doesn't return anything, the role class does.

vladfrangu (approved):

src/managers/RoleManager.js
ImRodry: The method returns a Promise<void>, u sure?

imranbarbhuiya: > The manager doesn't return anything, the role class does.

Oh my bad.

iCrawl (approved):

PR #7177, opened 2022-01-02
chore: bump LICENSE year

Syjalo:
Bump LICENSE year

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -175,8 +175,8 @@
 
    END OF TERMS AND CONDITIONS
 
-   Copyright 2015 - 2021 Noel Buechler
-   Copyright 2015 - 2021 Amish Shah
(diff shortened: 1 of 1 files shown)


favna (changes requested):
TL;DR: Very likely this won't be merged and you should stop this insane PR farming.

---

This is not how open source licenses work. You don't need to update the year at all. You're supposed to list the year of first publication.

https://opensource.stackexchange.com/a/5779

https://softwareengineering.stackexchange.com/a/130479

https://opensource.stackexchange.com/a/1523

In fact the way I understand it is that updating the license year is subsequently also damaging for your project because it implies that any new files added after the license change are licensed as of the current year, instead of as of the original publication date of the work.

Look at it this way. Is Mickey Mouse licensed as of 2022, or as of 1923. That's right, it's the latter. Open source software works the exact same.

PR #7188, opened 2022-01-06
refactor: remove required from getMember

imranbarbhuiya:
Removed `required` from `CommandInteractionOptionResolver#getMember`

In all other types, we can set an option to require. That's why we know that this value will be present or not when we are writing the code. But `getMember` is totally different. In this case, discord neither provide a `MEMBER` type nor a way to limit `USER` type to member only. That's why if we set a `USER` option which is required then also `getMember` can be null as in `USER` type we can provide a user Id of any user who aren't present in that guild.

Here's a example

```ts
    {
      name: "user",
      description: "The user to play against",
      required: true,
      type: "USER",
    }
```
Here the option is required. But if I use. 

```ts
const member = options.getMember("user", true)
```

Then, as this input can be any user who aren't even a member of that guild, so this line of code will through error.


src/structures/CommandInteractionOptionResolver.js
```diff
@@ -85,7 +85,7 @@ class CommandInteractionOptionResolver {
    * Gets an option by name and property and checks its type.
    * @param {string} name The name of the option.
    * @param {ApplicationCommandOptionType} type The type of the option.
-   * @param {string[]} properties The properties to check for for `required`.
+   * @param {string[]} properties The properties to check for `required`.
```
ImRodry, line 88: This is correct as-is, doesn’t need to be changed and is also out of the scope of the PR

src/structures/CommandInteractionOptionResolver.js
```diff
@@ -196,12 +196,11 @@ class CommandInteractionOptionResolver {
   /**
    * Gets a member option.
    * @param {string} name The name of the option.
-   * @param {boolean} [required=false] Whether to throw an error if the option is not found.
    * @returns {?(GuildMember|APIGuildMember)}
-   * The value of the option, or null if not set and not required.
+   * The value of the option, or null if the user is not present in the guild or not set and not required.
```
ImRodry, line 200:
```suggestion
   * The value of the option, or null if the user is not present in the guild or the option is not set.
```

typings/index.d.ts
```diff
@@ -49,10 +48,10 @@ import {
   RESTPostAPIApplicationCommandsJSONBody,
   Snowflake,
 } from 'discord-api-types/v9';
+import { Response } from 'node-fetch';
 import { ChildProcess } from 'node:child_process';
 import { EventEmitter } from 'node:events';
 import { AgentOptions } from 'node:https';
-import { Response } from 'node-fetch';
```
ImRodry, line 55: Sure…?

typings/index.d.ts
```diff
@@ -5463,8 +5460,6 @@ export interface ShardingManagerOptions {
   execArgv?: string[];
 }
 
-export { Snowflake };
-
```
ImRodry, line 5467: Please don’t add miscellaneous changes to your PR

ImRodry (changes requested):

typings/index.d.ts
imranbarbhuiya: It was done by ts organize import which organized it alphabetically. You can check it's preset in line 51. Idk which convention do you use. Should I turn that off?

typings/index.d.ts
imranbarbhuiya: This line is also moved to line 648

ImRodry: No clue what that is but it doesn’t sound relevant to this PR

imranbarbhuiya: reverted

vladfrangu (approved):

kyranet (approved):

iCrawl (approved):

imranbarbhuiya:
I need to remove some more types from this `test-d.ts` which were removed from main branch. Waiting for #7215 to merge then i can commit the changes.

SpaceEEC (approved):

Issue #5581, opened 2021-05-02
Duplicate shard


TGTGamer:
Heya guys, I've got an issue where the bot is duplicating the shard and I'm not sure why or what could be causing it. Can I get some advice?

manager.ts
```ts
const config: Config = require('../config.json')
const manager = new Discord.ShardingManager(`${__dirname}/bot.js`, {
  token: process.env.TOKEN || process.env.NPM_PACKAGE_TOKEN || config.token
})
manager.on('shardCreate', (shard: any) => {
  L.log(new loggingData('300', `Launching shard ${shard.id}`))
})
manager.spawn().catch((reason: any) => {
  if (reason && reason.size > 0)
    L.log(
      new loggingData('500', 'manager.events', {
        errors: reason,
        translate: true,
        T: {
          replace: {
            this: reason.toString()
          }
        }
      })
    )
})
```

**Further details:**

- discord.js version: 12.5.3
- Node.js version: 15.8.0
- Operating system: Tested on Linux, Win10
- Priority this issue should have – please be realistic and elaborate if possible: 

<details>
    <summary><b>Log</b></summary>

```log
[NOTICE]  Launching shard 0
[DEBUG]  [CLIENT] Preparing to connect to the gateway...
[NOTICE]  Launching shard 0
[DEBUG]  [CLIENT] 429 hit on route &#x2F;gateway&#x2F;bot
[NOTICE]  Attempting to load events 'ready.js' of admin module
...
[NOTICE]  Attempting to load commands 'refresh.js' of public module
[DEBUG]  [CLIENT] Preparing to connect to the gateway...
[DEBUG]  [CLIENT] 429 hit on route &#x2F;gateway&#x2F;bot
[NOTICE]  Attempting to load events 'ready.js' of admin module
...
[NOTICE]  Attempting to load commands 'refresh.js' of public module
[DEBUG]  [CLIENT] [WS =&gt; Manager] Fetched Gateway Information
    URL: wss:&#x2F;&#x2F;gateway.discord.gg
    Recommended Shards: 1
[DEBUG]  [CLIENT] [WS =&gt; Manager] Session Limit Information
    Total: 1000
    Remaining: 965
[DEBUG]  [CLIENT] [WS =&gt; Manager] Spawning shards: 0
[DEBUG]  [CLIENT] [WS =&gt; Shard 0] [CONNECT]
    Gateway    : wss:&#x2F;&#x2F;gateway.discord.gg&#x2F;
    Version    : 6
[...]

Issue #5364, opened 2021-03-01
Shard 0's process exited before its Client became ready.


MaximKing1:
**Please describe the problem you are having in as much detail as possible:**
I start the bot, when it starts using the ShardingManaer i get this error.

```
StriderM   | Error [SHARDING_IN_PROCESS]: Shards are still being spawned.
StriderM   | (node:1) UnhandledPromiseRejectionWarning: Error [SHARDING_READY_DIED]: Shard 0's process exited before its Client became ready.
```
then after it crashes with no other data, somethimes i get this

```
StriderM   | (node:1) UnhandledPromiseRejectionWarning: Error [SHARDING_READY_DIED]: Shard 0's process exited before its Client became ready.
StriderM   |     at Shard.onDeath (/usr/src/Strider/node_modules/discord.js/src/sharding/Shard.js:158:16)
StriderM   |     at Object.onceWrapper (events.js:422:26)
StriderM   |     at Shard.emit (events.js:315:20)
StriderM   |     at Shard._handleExit (/usr/src/Strider/node_modules/discord.js/src/sharding/Shard.js:384:10)
StriderM   |     at ChildProcess.emit (events.js:315:20)
StriderM   |     at Process.ChildProcess._handle.onexit (internal/child_process.js:277:12)
```

**Include a reproducible code sample here, if possible:**

Sharding Manager:
```js
// Import All The Packages
require("dotenv").config();
const { ShardingManager } = require("discord.js");

// Create Your ShardingManger Instance
const manager = new ShardingManager("./index.js", {
  execArgv: ['--trace-warnings'],
  token: process.env.TOKEN,  
  shardArgs: ["--color"]
});

// Spawn The Shards
manager.spawn();
```

index file:
```js
const { Client } = require("discord.js");
const client = new Client({
  ws: { intents: 32511 },
  restTimeOffset: 0,
  rateLimitAsError: true,
  disableMentions: "everyone",
}); // Require The Bot Client

// Get Our Packages
const glob = require("glob");

// Command & Event Files Using Glob
const commandFiles = glob.sync("./Helpers/commands/**/*.js");
for (const file of commandFiles) {
  const command = require(file);
  client.commands.set(command.name, command);
}
const eventFiles = glob.sync("./Helpers/events/**/*.js");
for (const file of eventFiles) {
  const event = require(file);
  const eventName = /\/events.(.*).js/.exec(file)[1];
  client.on(eventName, event.bind(null, client));
}

// Login To The Discord API
[...]

Issue #4095, opened 2020-04-18
Spam reconnection 


JusteLoneWolf:
**Please describe the problem you are having in as much detail as possible:**

Since several weeks (about 3 weeks) my spam bot reconnection to Discord, it happens a few hours or a few minutes after the launch of the shards (Less shards the more it will take time to spam reconnection) I have not made any modification for 4 weeks on my bot so I wonder if it comes from discord directly or from my code.

I have already run it locally and same problem and several of my friends have the same problem on a different library of discord (eris) but also on discord.js 

Namely I use Discord.js 11.6.4 

I currently have 25 servers so I launched my bot plus 25 shards I had no problem for 2-3 minutes then the shards were restarted one after the other. 

Namely that I'm not testing another bot on the same VPS or PC

https://imgur.com/c2Xbf0l.png

**Further details:**

- discord.js version: 11.6.5
- Node.js version: 12.16.1
- Operating system: Linux Debian (VPS)
Windows 10 (Local)
- Priority this issue should have - please be realistic and elaborate if possible: On a scale of 1 to 10 I would say 6/10

Translated with www.DeepL.com/Translator (free version)

zechaos031:
https://imgur.com/tManVca.png another image to illustrate (5 to 10 messages per minute)

adrifcastr:
that is expected behaviour.
Discord psuhed a gateway change that will regularly reconnect you.

zechaos031:
As far as I'm concerned, it's a reconnection every 1 or 2 minutes, I don't think that's really normal.

Monbrey:
You can read more about it from the announcement in the [Discord API Server](https://discordapp.com/channels/81384788765712384/381871767846780928/695380999063863406)

vladfrangu:
Firstly: if you have 25 guilds, you do NOT need 25 shards.. you need 1. You can start sharding when you get to 2.5k servers

Secondly; Discord has now made a change in the gateway API where they will ask connections to reconnect at random times. We implement that accordingly in both v11 and v12 of discord.js, the latter logging when it happens. 

Besides, this isn't really a problem as the bot RESUMED, which doesn't drain the Identify Limit. 

TL;DR: for 25 guilds you don't need to shard to 25 shards, what you've presented is normal, no need to worry about it!

zechaos031:
I know a shards is 2500 guild if I just wanted to see if it was a waiter or something else that was a problem. 

[...]

Issue #3890, opened 2020-03-03
client.login has no ratelimit backoff


MattIPv4:
**Please describe the problem you are having in as much detail as possible:**

Well, I got one of my bots ratelimited on /gateway/bot it seems. Watching the debug event, it seems to be continually returning 429s on /gateway/bot whilst trying to log in.

It seems to be getting these at a very consistent rate, which would suggest to me the login method has no logic to consider a ratelimited response and back-off etc.?

**Include a reproducible code sample here, if possible:**

```js
// This requires you be ratelimited
const client = new Client();
client.on('debug', console.log);
await client.login(token);
// Observe cotinual ratelimit logs
```

Logs from a slightly more fleshed out bot script:

```
 MNGR | 20/3/3 22:9:33  INFO  Launched shard 0
 S0 | 20/3/3 22:9:34  INFO  Logging in...
 S0 | 20/3/3 22:9:34  DEBUG  Provided token: [SNIP]
 S0 | 20/3/3 22:9:34  DEBUG  Preparing to connect to the gateway...
 S0 | 20/3/3 22:9:34  DEBUG  429 hit on route /gateway/bot
 S0 | 20/3/3 22:9:34  DEBUG  429 hit on route /gateway/bot
 S0 | 20/3/3 22:9:35  DEBUG  429 hit on route /gateway/bot
 S0 | 20/3/3 22:9:35  DEBUG  429 hit on route /gateway/bot
 S0 | 20/3/3 22:9:35  DEBUG  429 hit on route /gateway/bot
 S0 | 20/3/3 22:9:35  DEBUG  429 hit on route /gateway/bot
 S0 | 20/3/3 22:9:36  DEBUG  429 hit on route /gateway/bot
 S0 | 20/3/3 22:9:36  DEBUG  429 hit on route /gateway/bot
 S0 | 20/3/3 22:9:36  DEBUG  429 hit on route /gateway/bot
 S0 | 20/3/3 22:9:37  DEBUG  429 hit on route /gateway/bot
 S0 | 20/3/3 22:9:37  DEBUG  429 hit on route /gateway/bot
 S0 | 20/3/3 22:9:37  DEBUG  429 hit on route /gateway/bot
 S0 | 20/3/3 22:9:37  DEBUG  429 hit on route /gateway/bot
^C
```

**Further details:**

- discord.js version: 12.0.1
- Node.js version: 12.15.0
- Operating system: Windows server 2016
[...]

PR #7204, opened 2022-01-08
feat: added `@discordjs/sharder` package

kyranet:

I heard you wanted a nice and complete sharder. :eyes:

This sharder draws inspiration from the strategy and plug-in system Sapphire has largely adopted over the months, the way Kurasuta handles messages and shards separately, and the structure of Discord.js's built-in sharding manager to make the update easy, alongside a few sprinkles of my own based on my experience when building a cross-process intercommunication system.

:warning: **This PR is under heavy WIP**: docs and tests are missing, they'll be done as I progress. The API may also change as new needs arise when building more code.

Let's talk about this sharder's structure! It is divided in three components:

- **The Sharding Manager**: it's the central class that handles the shards, from starting them up, to closing.
- **The Shards**: they're strategies that define how the "channel" should be handled from the sharding manager, they can be of any kind, from workers (`WorkerShardHandler`) and forks (`ForkProcessShardHandler`) to cross-server clusters going thru local clusters (`ClusterProcessShardHandler`). Yes, you heard that. Although we'll most likely ship only the 3 local ones. You can make your own custom shard strategies.
- **The Messengers**: they're also strategies, but they define how data must be serialized and/or deserialized as well as handling message queues, we'll provide `RawMessageHandler` (data as-is, without converting to a Buffer or a string), `JsonMessageHandler` (JSON), and `BinaryMessageHandler` (V8 serder)... You can plug in your own ones, such as the Erlang External Term Format (ETF), YAML, TOML, etc.

This clear separation between responsibilities allow for greater customization of the core functionality of the library, reducing the need to use external libraries.

And coming in the future, we'll probably release **The Message Mappers**, which give a final touch-up to a message, such as appending a header/footer or even, using OpenSSL to encrypt messages over the network.

**Issues to handle:**

- ✔ **DONE** #2882: I'll change `respawn: boolean` to `respawn: number`, accepting `-1` and `Infinity` for unlimited respawns, but a number that is decreased by one on each shard until it's ready, in which case it'll reset the counter.
- #3560: I had an eureka moment here, it'll require a small refactor in Discord.js's WebSocket system but... turns out we can query `/gateway/bot` in the sharder, only once, and pass the data to the shards during spawn. We could also implement a query system so we can get that data from the ShardingManager without querying Discord more than once a day by caching and updating the result accordingly.
- #3890: See above.
- #4095: Status unknown.
- #5364: The new ready handler should be a lot cleaner and far more efficient, but we'll see.
- #5581: Can it happen with this sharder? Let's find out.

**Todo:**

- [x] Consolidate the sharder's API, make everything more consistent.
- [x] Make `ProcessShardHandler` abstract, for the two classes below.
- [x] Implement `ForkProcessShardHandler`.
- [x] Implement `CusterProcessShardHandler`.
- [ ] Implement `WorkerShardHandler`.
- [x] Implement `RawMessageHandler`.
- [x] Implement `BinaryMessageHandler`.
- [x] Implement built-in cluster sharding (e.g. 2 processes with 5 shards each to have 10 shards) support.
- [x] Implement `fetchRecommendedShards`.
- [ ] Implement shard client to interact with the shard manager, `ShardTunnel`? `ShardClient`?
- [ ] Type the events.
- [ ] Add documentation.
- [ ] Add tests.

**Note:** This is a port of https://github.com/discordjs/discord.js-modules/pull/87


xHyroM:
🎉

meister03:
Is there a reason, why @discordjs/sharder module is using the cluster module instead of child-processes. The Cluster module has some layers built on top of child_processes, which can make it slower.
Or has it been chosen, bc it allows using the same port?

kyranet:
I fail to see where you see cluster being the only/forced option to use, @meister03, it's simply a strategy.

If somebody wants to use cluster over child processes, they're free to do so.

Deivu:
This looks promising as a replacement for old d.js sharder. One thing though, can we implement our op system in this sharder by tapping on ipc messages, and do the same send() reply() fashion like how the old veza does? eg: if we want to send messages our selves from main process to sub process? One reason why I would like a system like this, is because I don't want to use broadcastEval on getting data across my sub process.

meister03:
> This looks promising as a replacement for old d.js sharder. One thing though, can we implement our op system in this sharder by tapping on ipc messages, and do the same send() reply() fashion like how the old veza does? eg: if we want to send messages our selves from main process to sub process? One reason why I would like a system like this, is because I don't want to use broadcastEval on getting data across my sub process.

You can reply to messages, thats been done with the `track` and `waitforId` functions(MessageHandler), when I recognized right.

DraftProducts:
I think there are a few features from Discordeno that we could learn from and would be a big improvement for discord.js.

https://github.com/discordeno/discordeno#gateway
image

vladfrangu:
A lot of those would go in the @discordjs/ws package, not in sharder, and are actually planned to a degree (for instance you can host the ws in a different process or machine, and in the future, you can make d.js pull from that process using NATS or whatever floats your boat)!
*I'd also like to clarify that technically there is "downtime" when resharding, as the connections have to close and reopen with an identify payload being sent with the new shard configuration but 🤫*

iCrawl:
I would love to take inspiration of discordeno, if it were any good that is.

Most of that text is just random fluff with no actual improvement or impact, or a complete lack and misunderstanding of how sharding works to make it look more appealing.

DraftProducts:
Sure @iCrawl, I agree with you.
Sadly, like any promotional text, it can happen that the reality is embellished (#linkdin), but behind these beautiful words, there is necessarily something concrete and/or a concept that could improve Discord.js.
I'm sure that downtime can be optimized.
If we use the current `ShardingManager` spawn system with a large amount of shards, an update of is painful.
I am convinced that we can find a way to launch the shards in a more optimized way, if only with the max_concurrency implemented by Discord.

Pitofui:
Honestly your all just reading the text and have 0 idea how the package runs. Honestly do you even know how it works from the core. Discord deno uses a proxy system that allows the gateway to stay running which allows you to restart only your bot. The rest hNdler and gateway handler are separate entities. Look at the source code before you start randomly judging and also look at the advanced bot setup to see how its arranged.

Drxckzyz:
> A lot of those would go in the @discordjs/ws package, not in sharder, and are actually planned to a degree (for instance you can host the ws in a different process or machine, and in the future, you can make d.js pull from that process using NATS or whatever floats your boat)!
> _I'd also like to clarify that technically there is "downtime" when resharding, as the connections have to close and reopen with an identify payload being sent with the new shard configuration but 🤫_

Actually it is close to zero downtime as you only lose a few events, you have the new shard and the old shards stays connected until the new one is ready

phamleduy04:
any updates on this?

kyranet:
Superseded by #8859, the new PR is up-to-date and implements the RFC. Also has a much simpler API.

PR #7211, opened 2022-01-08
refactor: make `intersect` perform a true intersection

suneettipirneni:

See: 
- https://github.com/discordjs/collection/issues/58

### DIFF
--- a/packages/collection/src/index.ts
+++ b/packages/collection/src/index.ts
@@ -629,10 +629,13 @@ export class Collection<K, V> extends Map<K, V> {
 	 *
 	 * @param other The other Collection to filter against
 	 */
-	public intersect<T>(other: Collection<K, T>): Collection<K, T> {
-		const coll = new this.constructor[Symbol.species]<K, T>();
+	public intersect<T>(other: Collection<K, T>): Collection<K, T | V> {
+		const coll = new this.constructor[Symbol.species]<K, T | V>();
 		for (const [k, v] of other) {
-			if (this.has(k)) coll.set(k, v);
+			const v2 = this.get(k);
+			if (v2 !== undefined && Object.is(v, v2)) {
+				coll.set(k, v);
+			}
 		}
 		return coll;
 	}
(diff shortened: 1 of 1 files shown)


vladfrangu (approved):

packages/collection/src/index.ts
```diff
@@ -629,10 +629,13 @@ export class Collection<K, V> extends Map<K, V> {
...
 	 * @param other The other Collection to filter against
 	 */
-	public intersect<T>(other: Collection<K, T>): Collection<K, T> {
-		const coll = new this.constructor[Symbol.species]<K, T>();
+	public intersect<T>(other: Collection<K, T>): Collection<K, T | V> {
+		const coll = new this.constructor[Symbol.species]<K, T | V>();
 		for (const [k, v] of other) {
-			if (this.has(k)) coll.set(k, v);
+			const v2 = this.get(k);
+			if (v2 !== undefined && Object.is(v, v2)) {
```
almeidx, line 636: What if `V` extends `undefined`? Probably want to use `this.has()` instead

packages/collection/src/index.ts
```diff
@@ -625,14 +625,16 @@ export class Collection<K, V> extends Map<K, V> {
...
 	 *
 	 * @param other The other Collection to filter against
 	 */
-	public intersect<T>(other: Collection<K, T>): Collection<K, T> {
-		const coll = new this.constructor[Symbol.species]<K, T>();
+	public intersect<T>(other: Collection<K, T>): Collection<K, T | V> {
+		const coll = new this.constructor[Symbol.species]<K, T | V>();
 		for (const [k, v] of other) {
-			if (this.has(k)) coll.set(k, v);
+			if (this.has(k) && Object.is(v, other.get(k))) {
```
kyranet, line 635: Did you mean to do this instead?
```suggestion
			if (this.has(k) && Object.is(v, this.get(k))) {
```

`other.get(k)` (while iterating `other`) equals to `v`, so this statement will never be false.

kyranet (changes requested):
Are you sure you meant to type the values of the resulting collection as `T | V`? I don't see any code that would set `V` into the collection, only `T`.

kyranet (approved):

iCrawl (approved):

SpaceEEC (approved):

PR #7216, opened 2022-01-08
typings: include base class in `GuildVoiceChannelResolvable`

AndyClausen:

Currently, something like this will throw type errors:

```ts
const channel: GuildChannel;
const member: GuildMember
if (channel.isVoice()) {
  member.voice.setChannel(channel);
}
```

because `isVoice` type guards channel to `BaseGuildVoiceChannel`, and `setChannel`'s takes `VoiceBasedChannel | Snowflake` which resolves to `VoiceChannel | StageChannel | string` - none of which are compatible with `BaseGuildVoiceChannel`.

I'm not sure this is the right approach in the long run, since I think there's something fundamentally wrong with how voice channels are typed compared to the typing of text channels, but this should work fine as a hotfix.

### DIFF
--- a/packages/discord.js/typings/index.d.ts
+++ b/packages/discord.js/typings/index.d.ts
@@ -4790,7 +4790,7 @@ export interface GuildScheduledEventUser<T> {
 
 export type GuildTemplateResolvable = string;
 
-export type GuildVoiceChannelResolvable = VoiceBasedChannel | Snowflake;
+export type GuildVoiceChannelResolvable = VoiceBasedChannel | BaseGuildVoiceChannel | Snowflake;
 
 export type HexColorString = `#${string}`;
 


ImRodry:
This reminded me to change #6957 to make isVoiceBased return VoiceBasedChannel instead of the base class. This might fix the issue you're running into but let the maintainers decide whether they want this or not

kyranet:
I personally don't want to, because that's a class you'll never get, and the union types are always preferred, but yeah... we need to fix the casts in the library.

AndyClausen:
> make isVoiceBased return VoiceBasedChannel instead of the base class. This might fix the issue you're running into

It definitely would. I don't think the solution in this PR is the best, I like your PR much more, but it'd fix some issues for now.

vladfrangu (approved):

AndyClausen:
Closed in favour of #6957

PR #7217, opened 2022-01-08
refactor(subcommandgroup): required default to false

imranbarbhuiya:

All other getX methods required is default to false. But for `subCommand` and `subCommandGroup` it's true.
`getSubCommand` have a valid reason that it's default to true because if there is no subCommand then why you are trying to get the value. But `getSubCommandGroup` is different.

Here's an example of how it's different,
```js
{
name: "name",
description: "description",
type: 1,
options: []
},
{
name: "name-group",
description: "description",
type: 2,
options: [
      {
          name: "name",
          description: "description",
          type: 1,
          options: []
      },
]
},
```
Here when a user uses `name` subCommand (1st one) and I try to check which `name` command was used, so I'll check `interaction.options.getSubCommandGroup()` first,  but it'll throw error because group is empty, but I've a subCommandGroup for this command.

### DIFF
--- a/packages/discord.js/src/structures/CommandInteractionOptionResolver.js
+++ b/packages/discord.js/src/structures/CommandInteractionOptionResolver.js
@@ -116,10 +116,10 @@ class CommandInteractionOptionResolver {
 
   /**
    * Gets the selected subcommand group.
-   * @param {boolean} [required=true] Whether to throw an error if there is no subcommand group.
+   * @param {boolean} [required=false] Whether to throw an error if there is no subcommand group.
    * @returns {?string} The name of the selected subcommand group, or null if not set and not required.
    */
-  getSubcommandGroup(required = true) {
+  getSubcommandGroup(required = false) {
     if (required && !this._group) {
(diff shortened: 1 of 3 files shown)


ImRodry:
You can’t have a subcommand group without a subcommand so why do you need to get the name of the group even? And I feel like that’s such a niche use case it wouldn’t hurt to pass the argument as false

imranbarbhuiya:
> You can’t have a subcommand group without a subcommand so why do you need to get the name of the group even? And I feel like that’s such a niche use case it wouldn’t hurt to pass the argument as false

I just didn't added subcommand inside of the subCommandGroup but i can have a subCommand with name `name` and a subcommand inside of the group with the same name so i need to check the group name.

imranbarbhuiya:
> that’s such a niche use case it wouldn’t hurt to pass the argument as false

yeah we can use false ofc and this is what I'm using rn but I think as there is a valid reason to make it false so I created the pr.

SpaceEEC (approved):

kyranet (approved):

PR #7108, opened 2021-12-13
refactor(Dates): save timestamps everywhere and use Date.parse

vaporox:
**Please describe the changes this PR makes and why it should be merged:**

- `Client#readyAt`, `GuildTemplate#createdAt` and `GuildTemplate#updatedAt` are now getters as opposed to their timestamp counterparts which are now properties
- Change from `new Date(x).getTime()` to `Date.parse(x)` where applicable for improved performance
- Fixed the timestamp property on the result of `MessageEmbed#toJSON` which was a `Date` instance and not a string
- Consistent implementation of date getters

**Status and versioning classification:**

- Code changes have been (partially) tested against the Discord API
- I know how to update typings and have done so

PR #7140, opened 2021-12-22
fix(BaseMessageComponent): don't create new class instances

ImRodry:
**Please describe the changes this PR makes and why it should be merged:**
Feel free to give the PR a better title as I didn't know what to call this but this basically resolves #7139 by returning the component instance in BaseMessageComponent#create when the received data is already a discord.js class

**Status and versioning classification:**
- Code changes have been tested against the Discord API, or there are no code changes
- I know how to update typings and have done so, or typings don't need updating

PR #7218, opened 2022-01-08
fix(MessageEmbed): createdAt field can be zero

Raiden-Ei:
* #7108

After above PR, `createdAt` field can be a number 0. It will result in an error since `Number` class doesn't have `toISOString` method.

This PR changes the value of `timestamp` field, to deal with number zero.

### DIFF
--- a/packages/discord.js/src/structures/MessageEmbed.js
+++ b/packages/discord.js/src/structures/MessageEmbed.js
@@ -88,7 +88,7 @@ class MessageEmbed {
      * @type {?number}
      */
     // Date.parse() cannot be used here because data.timestamp might be a number
-    this.timestamp = 'timestamp' in data ? new Date(data.timestamp).getTime() : null;
+    this.timestamp = new Date(data.timestamp).getTime() || null;
 
     /**
      * Represents a field of a MessageEmbed
@@ -515,7 +515,7 @@ class MessageEmbed {
       type: 'rich',
       description: this.description,
       url: this.url,
-      timestamp: this.createdAt?.toISOString?.(),
+      timestamp: this.createdAt?.toISOString(),
       color: this.color,
       fields: this.fields,
       thumbnail: this.thumbnail,
--- a/packages/discord.js/src/structures/MessageEmbed.js
+++ b/packages/discord.js/src/structures/MessageEmbed.js
@@ -515,7 +515,7 @@ class MessageEmbed {
       type: 'rich',
       description: this.description,
       url: this.url,
-      timestamp: this.createdAt?.toISOString(),
+      timestamp: this.createdAt?.toISOString?.(),
       color: this.color,
       fields: this.fields,
       thumbnail: this.thumbnail,
--- a/packages/discord.js/src/structures/MessageEmbed.js
+++ b/packages/discord.js/src/structures/MessageEmbed.js
@@ -88,6 +88,7 @@ class MessageEmbed {
      * @type {?number}
      */
     // Date.parse() cannot be used here because data.timestamp might be a number
+    // Additionally, the nullish coalescing operator cannot be used here as we're checking for 0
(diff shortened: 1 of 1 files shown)


ImRodry:
You should fix how timestamp is set instead, by changing it to
```js
this.timestamp = new Date(data.timestamp).getTime() || null
```

packages/discord.js/src/structures/MessageEmbed.js
```diff
@@ -88,7 +88,7 @@ class MessageEmbed {
      * @type {?number}
      */
     // Date.parse() cannot be used here because data.timestamp might be a number
-    this.timestamp = 'timestamp' in data ? new Date(data.timestamp).getTime() : null;
+    this.timestamp = new Date(data.timestamp).getTime() || null;
```
ImRodry, line 91: Great! Should probably just leave this comment here to prevent confusion in the future
```suggestion
    // Additionally, the nullish coalescing operator cannot be used here as we're checking for 0
    this.timestamp = new Date(data.timestamp).getTime() || null;
```

Raiden-Ei: Done 😃

ImRodry (approved):

kyranet:
The fix is actually to change this line:

https://github.com/discordjs/discord.js/blob/ae876d962453ccf843f8e6f70666a98a3173bb75/packages/discord.js/src/structures/MessageEmbed.js#L244

Defaulting a perfectly valid value (`0`) to `null` is not the right approach.

Koyamie:
I feel like there is another issue going on there, my embeds are "setup" two times for an unknown reason.
When I `console.log` the setup function while doing a `new MessageEmbed()` I have two logs, one with empty object from `data`, one with my full `MessageEmbed` which cause timestamp to be 0 because `'timestamp' in data` when `timestamp = null` is `true`.
The second log doesn't appear when I do not send the embed.

kyranet:
MessagePayload builds MessageEmbed instances to transform embeds data even if they're already MessageEmbed instances.

Koyamie:
> MessagePayload builds MessageEmbed instances to transform embeds data even if they're already MessageEmbed instances.

Alright so we have to make sure data.timestamp is not null to build the date on setup and not fixing line **244** like you said.

packages/discord.js/src/structures/MessageEmbed.js
```diff
@@ -88,7 +88,8 @@ class MessageEmbed {
      * @type {?number}
      */
     // Date.parse() cannot be used here because data.timestamp might be a number
-    this.timestamp = 'timestamp' in data ? new Date(data.timestamp).getTime() : null;
+    // Additionally, the nullish coalescing operator cannot be used here as we're checking for 0
```
almeidx, line 91: IMO this comment is very much unnecessary. It's pretty obvious we're *only* checking for 0 since `<Date>.getTime()` will always return a number

vladfrangu (approved):

vladfrangu (approved):

ImRodry:
> MessagePayload builds MessageEmbed instances to transform embeds data even if they're already MessageEmbed instances.

This was fixed for message components in #7140 so it might be worth it to fix it here too, will PR

iCrawl (approved):

kyranet:
> > MessagePayload builds MessageEmbed instances to transform embeds data even if they're already MessageEmbed instances.
> 
> Alright so we have to make sure data.timestamp is not null to build the date on setup and not fixing line **244** like you said. Because line **244** works perfectly fine when we do not mess up on setup.

No, you just change line 244 to account for nullish instead of falsy values.

SpaceEEC (approved):

ImRodry:
What happened to @kyranet’s review?

PR #7226, opened 2022-01-09
fix(timestamps): account for timestamps of 0 when creating Dates

ImRodry:
For some reason, @kyranet's suggestion in #7218 was ignored so this PR fixes all instances where timestamps could be 0 on the library. I had to disable the eqeqeq rule for this in order to check for null and undefined timestamps but if you have a better idea please let me know.

Note about the change on the GuildMemberManager file: Discord throws an error when passing a timestamp that is more than 28 days in the past or the future, however, without this fix you'd be able to pass 0 and get no error, as the API would treat that as null which ends up doing the same thing but is not exactly the expected behavior. With this fix, we transform every timestamp into an ISO string and let Discord error when it should. I'll gladly undo this if you think it's necessary however

### DIFF
--- a/packages/discord.js/src/managers/GuildMemberManager.js
+++ b/packages/discord.js/src/managers/GuildMemberManager.js
@@ -266,7 +266,10 @@ class GuildMemberManager extends CachedManager {
     _data.roles &&= _data.roles.map(role => (role instanceof Role ? role.id : role));
 
     _data.communication_disabled_until =
-      _data.communicationDisabledUntil && new Date(_data.communicationDisabledUntil).toISOString();
+      // eslint-disable-next-line eqeqeq
+      _data.communicationDisabledUntil != null
+        ? new Date(_data.communicationDisabledUntil).toISOString()
+        : _data.communicationDisabledUntil;
 
     let endpoint = this.client.api.guilds(this.guild.id);
     if (id === this.client.user.id) {
--- a/packages/discord.js/src/structures/MessageEmbed.js
+++ b/packages/discord.js/src/structures/MessageEmbed.js
@@ -88,8 +88,8 @@ class MessageEmbed {
      * @type {?number}
      */
     // Date.parse() cannot be used here because data.timestamp might be a number
-    // Additionally, the nullish coalescing operator cannot be used here as we're checking for 0
-    this.timestamp = new Date(data.timestamp).getTime() || null;
+    // eslint-disable-next-line eqeqeq
+    this.timestamp = data.timestamp != null ? new Date(data.timestamp).getTime() : null;
 
     /**
      * Represents a field of a MessageEmbed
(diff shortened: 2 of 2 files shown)


packages/discord.js/src/structures/MessageEmbed.js
```diff
@@ -88,8 +88,8 @@ class MessageEmbed {
      * @type {?number}
      */
     // Date.parse() cannot be used here because data.timestamp might be a number
-    // Additionally, the nullish coalescing operator cannot be used here as we're checking for 0
-    this.timestamp = new Date(data.timestamp).getTime() || null;
+    // eslint-disable-next-line eqeqeq
+    this.timestamp = data.timestamp != null ? new Date(data.timestamp).getTime() : null;
```
imranbarbhuiya, line 92: I'm just curious to know that why we can't use `!==` here?

ImRodry: Read the description

imranbarbhuiya: Oh sorry, got it's because of 0. right?

ImRodry: No... it's because of undefined

ImRodry:
Accidentally forgot to commit some changes, all good now

vladfrangu (approved):

packages/discord.js/src/managers/GuildMemberManager.js
```diff
@@ -266,7 +266,10 @@ class GuildMemberManager extends CachedManager {
     _data.roles &&= _data.roles.map(role => (role instanceof Role ? role.id : role));
 
     _data.communication_disabled_until =
-      _data.communicationDisabledUntil && new Date(_data.communicationDisabledUntil).toISOString();
+      // eslint-disable-next-line eqeqeq
+      _data.communicationDisabledUntil != null
+        ? new Date(_data.communicationDisabledUntil).toISOString()
+        : _data.communicationDisabledUntil;
```
Jiralite, line 272: Why is this change needed? 0 is not a value that is accepted here?

thebigwolfy (approved):

packages/discord.js/src/managers/GuildMemberManager.js
ImRodry: Please read the description of the PR

iCrawl (approved):

packages/discord.js/src/managers/GuildMemberManager.js
NotSugden: would `_data.communicationDisabledUntil ??= ...` work here ?

ImRodry: That would only work if the value was nullish, and &&= wouldn't either because of 0

NotSugden: oh yes, im reading the logic in reverse MB

kyranet (approved):

PR #7238, opened 2022-01-10
refactor(files): file data can be much more than buffer

ckohen:

The data for a file is directly appended to formdata, which accepts a multitude of types, not just a buffer, this corrects the types and changes the name of the property accordingly (not every single stream type is covered, but it is the natively supported set)

### DIFF
--- a/packages/rest/src/lib/RequestManager.ts
+++ b/packages/rest/src/lib/RequestManager.ts
@@ -8,8 +8,6 @@ import type { IHandler } from './handlers/IHandler';
 import { SequentialHandler } from './handlers/SequentialHandler';
 import type { RESTOptions, RestEvents } from './REST';
 import { DefaultRestOptions, DefaultUserAgent } from './utils/constants';
-import type { ReadStream } from 'fs';
-import type { ServerResponse } from 'http';
 
 let agent: Agent | null = null;
 
@@ -30,7 +28,7 @@ export interface RawFile {
 	/**
 	 * The actual data for the file
 	 */
-	fileData: string | number | boolean | Buffer | ReadStream | ServerResponse;
+	fileData: string | number | boolean | Buffer;
 }
 
 /**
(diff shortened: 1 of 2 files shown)


packages/rest/src/lib/RequestManager.ts
```diff
@@ -28,7 +30,7 @@ export interface RawFile {
 	/**
 	 * The actual data for the file
 	 */
-	rawBuffer: Buffer;
+	fileData: string | number | boolean | Buffer | ReadStream | ServerResponse;
```
vladfrangu, line 33: Is it readstream or readable from the stream module?

ckohen: The packages specifies a file stream

KhafraDev, line 33: All of these values other than string aren't [spec compliant](https://developer.mozilla.org/en-US/docs/Web/API/FormData/append) and will make using a different library harder.

ckohen: That's not actually true, the only non-spec compliant are the streams, which based on vlads comment above may be unnecessary anyways. The spec calls for .toString() to be callled on anything that is not a Blob or USVstring

KhafraDev: The spec doesn't have value converted to a USV string though? https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#create-an-entry

ckohen: Interesting that it is not in the bare bones html spec, however both the [web api](https://developer.mozilla.org/en-US/docs/Web/API/FormData/append) and [undici](https://github.com/nodejs/undici/blob/main/lib/fetch/formdata.js#L37) convert any non string / blob into a string

Regardless, when we do switch to undici, the second argument can *only* be of type `Blob` when providing a third argument, so we are going to need to translate anyways since Discord now essentially requires the third parameter

KhafraDev: From experience the spec will often leave basic things out of the spec and have the reader infer what their implementation should do. In this case, the param should theoretically be `any` other than Symbols and objects with a Symbol.toStringTag that throws an error I guess.

However, since it's written in ts, the values will need to be converted to string/Blob before being able to append them anyways, once a switch is made.

vladfrangu (approved):

kyranet (approved):

SpaceEEC (approved):

PR #7239, opened 2022-01-10
fix: `ref` delay for rate limited requests

suneettipirneni:

The request handler relied on awaiting `delay` here:
https://github.com/discordjs/discord.js/blob/b2dfb2ec70c33d30cf9b4a0b6757fbc6c750826c/packages/rest/src/lib/handlers/SequentialHandler.ts#L278

This works if this isn't the only code left in the event loop. However if there is no other code left and the program always ends up on awaiting the delay, it will never actually `await` because it's been unref'd. 


packages/rest/src/lib/handlers/SequentialHandler.ts
```diff
@@ -252,7 +252,7 @@ export class SequentialHandler {
 				// Set RateLimitData based on the route-specific limit
 				limit = this.limit;
 				timeout = this.timeToReset;
-				delay = sleep(timeout, undefined, { ref: false });
+				delay = sleep(timeout, undefined);
```
almeidx, line 255:
```suggestion
				delay = sleep(timeout);
```

almeidx (changes requested):

vladfrangu (approved):

kyranet (approved):

iCrawl (approved):

SpaceEEC (approved):

PR #7244, opened 2022-01-10
chore: improve linting

iCrawl:


packages/rest/src/lib/handlers/SequentialHandler.ts
```diff
@@ -304,11 +304,12 @@ export class SequentialHandler {
 		try {
 			// node-fetch typings are a bit weird, so we have to cast to any to get the correct signature
 			// Type 'AbortSignal' is not assignable to type 'import("discord.js-modules/node_modules/@types/node-fetch/externals").AbortSignal'
+			// eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
 			res = await fetch(url, { ...options, signal: controller.signal as any });
 		} catch (error: unknown) {
 			// Retry the specified number of times for possible timed out requests
 			if (error instanceof Error && error.name === 'AbortError' && retries !== this.manager.options.retries) {
-				return this.runRequest(routeId, url, options, bodyData, ++retries);
+				return void this.runRequest(routeId, url, options, bodyData, ++retries);
```
vladfrangu, line 312: I don't think this is correct

vladfrangu (changes requested):

packages/rest/src/lib/handlers/SequentialHandler.ts
iCrawl: Yeah I wasn't sure whether to `return await` or `return void` here to make it shut up.

iCrawl: Tests saved us! `return await` is the way to go here.

packages/rest/src/lib/REST.ts
```diff
@@ -171,20 +171,20 @@ export interface RestEvents {
 }
 
 export interface REST {
-	on<K extends keyof RestEvents>(event: K, listener: (...args: RestEvents[K]) => void): this;
-	on<S extends string | symbol>(event: Exclude<S, keyof RestEvents>, listener: (...args: any[]) => void): this;
+	on: (<K extends keyof RestEvents>(event: K, listener: (...args: RestEvents[K]) => void) => this) &
```
vladfrangu, line 174: I really dislike this but 🤷

vladfrangu (approved):

packages/rest/src/lib/REST.ts
iCrawl: 🤷

packages/voice/src/VoiceConnection.ts
```diff
@@ -261,10 +261,14 @@ export class VoiceConnection extends TypedEmitter<VoiceConnectionEvents> {
 	 */
 	public set state(newState: VoiceConnectionState) {
 		const oldState = this._state;
+		// eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
 		const oldNetworking: Networking | undefined = Reflect.get(oldState, 'networking');
+		// eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
 		const newNetworking: Networking | undefined = Reflect.get(newState, 'networking');
```
kyranet, line 267: Can't we change the casts to this? Or does ESLint still complain?
```suggestion
		const oldNetworking = Reflect.get(oldState, 'networking') as Networking | undefined;
		const newNetworking = Reflect.get(newState, 'networking') as Networking | undefined;
```

packages/rest/src/lib/handlers/SequentialHandler.ts
```diff
@@ -304,11 +304,12 @@ export class SequentialHandler {
 		try {
 			// node-fetch typings are a bit weird, so we have to cast to any to get the correct signature
 			// Type 'AbortSignal' is not assignable to type 'import("discord.js-modules/node_modules/@types/node-fetch/externals").AbortSignal'
+			// eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
 			res = await fetch(url, { ...options, signal: controller.signal as any });
 		} catch (error: unknown) {
 			// Retry the specified number of times for possible timed out requests
 			if (error instanceof Error && error.name === 'AbortError' && retries !== this.manager.options.retries) {
-				return this.runRequest(routeId, url, options, bodyData, ++retries);
+				return await this.runRequest(routeId, url, options, bodyData, ++retries);
```
kyranet, line 312: By adding this await, the finally block (line 317 in this PR) will be called later, causing intervals to accumulate without clearing between recursive calls.

What exactly is the ESLint error that required this to be added?

packages/voice/src/audio/AudioPlayer.ts
```diff
@@ -163,6 +164,19 @@ export type AudioPlayerEvents = {
...
+ * Stringifies an AudioPlayerState instance.
+ *
+ * @param state - The state to stringify
+ */
+function stringifyState(state: AudioPlayerState) {
+	return JSON.stringify({
+		...state,
+		resource: Reflect.has(state, 'resource'),
+		stepTimeout: Reflect.has(state, 'stepTimeout'),
+	});
```
kyranet, line 177: I just realised... aren't those Reflect.has useless? o.o

packages/voice/src/VoiceConnection.ts
kyranet: If so, then this applies to everywhere disabling this rule.

kyranet (changes requested):
Voice code is extremely painful to read with all those comments, but either way, requested changes here!

packages/voice/src/audio/AudioPlayer.ts
iCrawl: Maybe, maybe not, but I'm not going to plan on doing big code changes in a linting PR.

packages/rest/src/lib/handlers/SequentialHandler.ts
iCrawl: https://typescript-eslint.io/rules/return-await/

ckohen: Not actually sure this matters in this particular instance because to reach that code I'm fairly certain the timeout has already finished, causing the AbortError

kyranet (approved):

packages/discord.js/test/sendtest.js
```diff
@@ -1,11 +1,11 @@
 'use strict';
 
+const fetch = require('node-fetch');
```
ImRodry, line 3: Shouldn’t node imports stay before everything else and not be sorted alphabetically like this was?

iCrawl: Don't care much about those test files regardless.

ImRodry: Ah true, I didn’t realize it was a test file

SpaceEEC (approved):

PR #7245, opened 2022-01-11
types: add `ReadonlyCollection`

suneettipirneni:

Provides a utility type for ensuring a given collection doesn't get mutated.

### DIFF
--- a/packages/collection/src/index.ts
+++ b/packages/collection/src/index.ts
@@ -10,6 +10,12 @@ export interface CollectionConstructor {
 }
 
 /**
+ * Represents an immutable version of a collection
+ */
+export type ReadonlyCollection<K, V> = ReadonlyMap<K, V> &
+	Omit<Collection<K, V>, 'forEach' | 'ensure' | 'sweep' | 'sort'>;
+
+/**
  * Separate interface for the constructor so that emitted js does not have a constructor that overwrites itself
  *
  * @internal
@@ -557,7 +563,7 @@ export class Collection<K, V> extends Map<K, V> {
 	 * @example
 	 * const newColl = someColl.clone();
 	 */
-	public clone() {
+	public clone(): Collection<K, V> {
 		return new this.constructor[Symbol.species](this);
 	}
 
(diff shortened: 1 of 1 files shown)


packages/collection/src/index.ts
```diff
@@ -9,6 +9,12 @@ export interface CollectionConstructor {
 	readonly [Symbol.species]: CollectionConstructor;
 }
 
+/**
+ * Represents an immutable version of a collection
+ */
+export type ReadonlyCollection<K, V> = ReadonlyMap<K, V> &
+	Omit<Collection<K, V>, 'forEach' | 'ensure' | 'sweep' | 'sort'>;
```
vladfrangu, line 16: Won't this pick set, delete, etc? Also, why are you omitting forEach?

vladfrangu (changes requested):

packages/collection/src/index.ts
suneettipirneni: I just fixed the get/set/delete. However `forEach` is omitted because ReadonlyMap uses its own signature that only accepts readonly maps. Since collection simply inherits from map I omitted from the collection and used `ReadonlyMap`s foreach instead.

packages/collection/src/index.ts
```diff
@@ -9,6 +9,12 @@ export interface CollectionConstructor {
 	readonly [Symbol.species]: CollectionConstructor;
 }
 
+/**
+ * Represents an immutable version of a collection
+ */
+export type ReadonlyCollection<K, V> = ReadonlyMap<K, V> &
+	Omit<Collection<K, V>, 'forEach' | 'ensure' | 'sweep' | 'sort' | 'get' | 'set' | 'delete'>;
```
vladfrangu, line 16:
```suggestion
	Omit<Collection<K, V>, 'forEach' | 'ensure' | 'reverse' | 'sweep' | 'sort' | 'get' | 'set' | 'delete'>;
```

vladfrangu (changes requested):

kyranet (approved):

vladfrangu (approved):

SpaceEEC (approved):

PR #7248, opened 2022-01-11
refactor: PresenceUpdate and demuxProbe

legendhimslef:
In the `PresenceUpdate#handle` there are **null** checks at lines 9 and 12. It's redundant since at line 8 the `id` is accessed from `data.user` without any null check.

In the `demuxProbe` function, Execution falls through after this Promise 'reject' call. So adding a return statement after the reject to stop unintended code execution.


Jiralite:
The first half of the pull request is addressed by #6782.

SpaceEEC (approved):

iCrawl (approved):

PR #7255, opened 2022-01-12
feat: rest hash and handler sweeping

suneettipirneni:

See https://github.com/discordjs/discord.js-modules/issues/40

Sweeps hashes based on the lifetime of a given hash. Lifetimes are based on the last time a particular hash was used relative to the current data/time. In addition this sweeps inactive handlers.

This introduces new options for `REST`:
- `hashSweepInterval`: The delay before each hash sweep interval.
- `hashLifetime`: The maximum amount of time a hash can exist before being swept.
- `handlerSweepInterval`: The delay before each handler sweep interval

### DIFF
--- a/packages/rest/src/lib/REST.ts
+++ b/packages/rest/src/lib/REST.ts
@@ -6,6 +6,7 @@ import type { AgentOptions } from 'node:https';
 import type { RequestInit, Response } from 'node-fetch';
 import type { HashData } from '../index';
 import type Collection from '@discordjs/collection';
+import type { IHandler } from './handlers/IHandler';
 
 /**
  * Options to be passed when creating the REST instance
@@ -86,6 +87,11 @@ export interface RESTOptions {
 	 * @default 86_400_000
 	 */
 	hashLifetime: number;
+	/**
+	 * The amount of time in milliseconds that passes between each hash sweep. (defaults to 1h)
+	 * @default 3_600_000
+	 */
+	handlerSweepInterval: number;
 }
 
 /**
@@ -181,6 +187,7 @@ export interface RestEvents {
 	newListener: [name: string, listener: (...args: any) => void];
 	removeListener: [name: string, listener: (...args: any) => void];
 	hashSweep: [sweptHashes: Collection<string, HashData>];
+	handlerSweep: [sweptHandlers: Collection<string, IHandler>];
 }
 
 export interface REST {
--- a/packages/rest/src/lib/REST.ts
+++ b/packages/rest/src/lib/REST.ts
@@ -77,13 +77,13 @@ export interface RESTOptions {
 	 */
 	version: string;
 	/**
-	 * The amount of time in milliseconds that passes between each hash sweep
-	 * @default 3_600_000
+	 * The amount of time in milliseconds that passes between each hash sweep. (defaults to 4h)
+	 * @default 14_400_000
 	 */
 	hashSweepInterval: number;
 	/**
-	 * The maximum amount of time a hash can exist in milliseconds without being hit with a request
-	 * @default 21_600_000
+	 * The maximum amount of time a hash can exist in milliseconds without being hit with a request (defaults to 24h)
+	 * @default 86_400_000
 	 */
 	hashLifetime: number;
 }
--- a/packages/rest/src/lib/REST.ts
+++ b/packages/rest/src/lib/REST.ts
@@ -4,6 +4,8 @@ import { InternalRequest, RequestData, RequestManager, RequestMethod, RouteLike 
 import { DefaultRestOptions, RESTEvents } from './utils/constants';
 import type { AgentOptions } from 'node:https';
 import type { RequestInit, Response } from 'node-fetch';
+import type { HashData } from '../index';
+import type Collection from '@discordjs/collection';
 
 /**
  * Options to be passed when creating the REST instance
@@ -74,6 +76,16 @@ export interface RESTOptions {
 	 * @default '9'
 	 */
 	version: string;
+	/**
+	 * The amount of time in milliseconds that passes between each hash sweep
+	 * @default 3_600_000
+	 */
+	hashSweepInterval: number;
+	/**
+	 * The maximum amount of time a hash can exist in milliseconds without being hit with a request
+	 * @default 21_600_000
+	 */
+	hashLifetime: number;
 }
 
 /**
@@ -168,6 +180,7 @@ export interface RestEvents {
 	response: [request: APIRequest, response: Response];
 	newListener: [name: string, listener: (...args: any) => void];
 	removeListener: [name: string, listener: (...args: any) => void];
+	hashSweep: [sweptHashes: Collection<string, HashData>];
 }
 
 export interface REST {
@@ -197,7 +210,8 @@ export class REST extends EventEmitter {
 		this.requestManager = new RequestManager(options)
 			.on(RESTEvents.Debug, this.emit.bind(this, RESTEvents.Debug))
 			.on(RESTEvents.RateLimited, this.emit.bind(this, RESTEvents.RateLimited))
-			.on(RESTEvents.InvalidRequestWarning, this.emit.bind(this, RESTEvents.InvalidRequestWarning));
+			.on(RESTEvents.InvalidRequestWarning, this.emit.bind(this, RESTEvents.InvalidRequestWarning))
+			.on(RESTEvents.HashSweep, this.emit.bind(this, RESTEvents.HashSweep));
 
 		this.on('newListener', (name, listener) => {
 			if (name === RESTEvents.Request || name === RESTEvents.Response) this.requestManager.on(name, listener);
--- a/packages/rest/src/lib/REST.ts
+++ b/packages/rest/src/lib/REST.ts
@@ -4,7 +4,7 @@ import { InternalRequest, RequestData, RequestManager, RequestMethod, RouteLike 
 import { DefaultRestOptions, RESTEvents } from './utils/constants';
 import type { AgentOptions } from 'node:https';
 import type { RequestInit, Response } from 'node-fetch';
-import type { HashData } from '../index';
+import type { HashData } from './RequestManager';
 import type Collection from '@discordjs/collection';
 import type { IHandler } from './handlers/IHandler';
 
--- a/packages/rest/src/lib/RequestManager.ts
+++ b/packages/rest/src/lib/RequestManager.ts
@@ -210,7 +210,7 @@ export class RequestManager extends EventEmitter {
 			validateMaxInterval(this.options.hashSweepInterval);
 			this.hashTimer = setInterval(() => {
 				// Only allocate a swept collection for event.
-				const sweptHashes: Collection<string, HashData> = new Collection<string, HashData>();
+				const sweptHashes = new Collection<string, HashData>();
 
 				const currentDate = Date.now();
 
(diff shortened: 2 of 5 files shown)


packages/rest/src/lib/RequestManager.ts
```diff
@@ -181,6 +191,43 @@ export class RequestManager extends EventEmitter {
...
+			// Begin sweep hash based on lifetimes
+			this.hashes.sweep((v, k) => {
+				// `-1` indicates a global hash
+				if (v.lastAccess === -1) return false;
+
+				// Check if lifetime has been exceeded
+				const shouldSweep = Math.floor(curDate - v.lastAccess) > this.options.hashLifetime;
+
+				// Add hash to collection of swept hashes
+				if (shouldSweep && isListeningToSweeps) {
```
vladfrangu, line 214: You should always emit the debug event if needed, even if not listening to sweeps

packages/rest/src/lib/RequestManager.ts
```diff
@@ -181,6 +191,43 @@ export class RequestManager extends EventEmitter {
...
+					if (this.listenerCount(RESTEvents.Debug)) {
```
vladfrangu, line 215: Cache the return of listenerCount at the top

packages/rest/src/lib/handlers/SequentialHandler.ts
```diff
@@ -355,7 +355,16 @@ export class SequentialHandler {
 			// Let library users know when rate limit buckets have been updated
 			this.debug(['Received bucket hash update', `  Old Hash  : ${this.hash}`, `  New Hash  : ${hash}`].join('\n'));
 			// This queue will eventually be eliminated via attrition
-			this.manager.hashes.set(`${method}:${routeId.bucketRoute}`, hash);
+			this.manager.hashes.set(`${method}:${routeId.bucketRoute}`, { value: hash, lastAccess: Date.now() });
+		} else if (hash) {
```
vladfrangu, line 359: Do you need to check if hash is present?

vladfrangu (changes requested):

packages/rest/src/lib/handlers/SequentialHandler.ts
suneettipirneni: Wouldn't this need to be checked if the route has a global hash?

vladfrangu: Oh, probably

SpaceEEC (dismissed):

packages/rest/src/lib/REST.ts
```diff
@@ -74,6 +76,16 @@ export interface RESTOptions {
...
 	 */
 	version: string;
+	/**
+	 * The amount of time in milliseconds that passes between each hash sweep
+	 * @default 3_600_000
+	 */
+	hashSweepInterval: number;
+	/**
+	 * The maximum amount of time a hash can exist in milliseconds without being hit with a request
+	 * @default 21_600_000
```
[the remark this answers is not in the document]
suneettipirneni: I actually wanted some input on this from @discordjs/the-big-4, what are some reasonable defaults for the hash lifetime and the default sweep interval

packages/rest/src/lib/RequestManager.ts
```diff
@@ -181,6 +191,41 @@ export class RequestManager extends EventEmitter {
 		this.options = { ...DefaultRestOptions, ...options };
 		this.options.offset = Math.max(0, this.options.offset);
 		this.globalRemaining = this.options.globalRequestsPerSecond;
+
+		setInterval(() => {
+			// Only allocate a swept collection if there are listeners
+			const sweptHashes: Collection<string, HashData> | null =
+				this.listenerCount(RESTEvents.HashSweep) > 0 ? new Collection<string, HashData>() : null;
```
ckohen, line 198: Listener Count will *always* be 1 here since you're proxying the event through the rest event emitter

packages/rest/src/lib/RequestManager.ts
```diff
@@ -181,6 +191,41 @@ export class RequestManager extends EventEmitter {
...
+
+			const listeningToDebug = this.listenerCount(RESTEvents.Debug) > 0;
```
ckohen, line 200: Ditto

packages/rest/src/lib/RequestManager.ts
```diff
@@ -181,6 +191,41 @@ export class RequestManager extends EventEmitter {
...
+
+				// Add hash to collection of swept hashes
+				if (shouldSweep && sweptHashes) {
+					// Add to swept hashes
+					sweptHashes.set(k, v);
+				}
+
+				if (listeningToDebug) {
+					this.emit(RESTEvents.Debug, `Hash ${v.value} for ${k} swept due to lifetime being exceeded`);
+				}
```
ckohen, line 219: Emitting this event is not computationally intensive, simply emitting is more efficient than checking

packages/rest/src/lib/RequestManager.ts
```diff
@@ -181,6 +191,41 @@ export class RequestManager extends EventEmitter {
 		this.options = { ...DefaultRestOptions, ...options };
 		this.options.offset = Math.max(0, this.options.offset);
 		this.globalRemaining = this.options.globalRequestsPerSecond;
+
+		setInterval(() => {
```
ckohen, line 195: If someone sets the hash sweep interval to infinity or 0 you should not set this interval

packages/rest/src/lib/REST.ts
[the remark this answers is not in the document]
ckohen: I vote 24 hours, the build up isn't too bad

ckohen (changes requested):
I was going to do sweeping of handlers in djs but since we're sweeping hashes here I guess we should sweep handlers too. I do think that would be within the scope of this PR (most of the groundwork was layed long ago), but if you don't think so, I'll make that PR once this is merged

packages/rest/src/lib/RequestManager.ts
ckohen: Looks like this was resolved without changes? (diff shows changes but I can't actually see any)

suneettipirneni: I think the line numbers changed a lot, but it's in the PR diff

ckohen: You're still checking the listener count that will always be 1 in the PR diff

packages/rest/src/lib/REST.ts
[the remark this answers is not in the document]
kyranet: Problem with 24h intervals is that you're assuming Node.js's timers are wall clock timers, but it's not, since they're very slow.

24h intervals can run as short as a couple of hours given extreme cases. 30 minutes is generally the highest value where the timing is somewhat accurate with wall clock's time.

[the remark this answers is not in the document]
ckohen: I didn't realize you were talking about the interval, I was talking about the lifetime

packages/rest/src/lib/RequestManager.ts
```diff
@@ -181,6 +191,73 @@ export class RequestManager extends EventEmitter {
...
+			}
+		};
+
+		if (this.options.hashSweepInterval !== 0 && this.options.hashSweepInterval !== Infinity) {
+			validateMaxInterval(this.options.hashSweepInterval);
+			setInterval(() => {
+				// Only allocate a swept collection if there are listeners
+				const sweptHashes: Collection<string, HashData> = new Collection<string, HashData>();
+
+				const curDate = Date.now();
```
kyranet, line 212: I'm not a fan of this name, please use full names.

packages/rest/src/lib/RequestManager.ts
```diff
@@ -181,6 +191,73 @@ export class RequestManager extends EventEmitter {
...
+				});
+
+				// Fire event
+				this.emit(RESTEvents.HashSweep, sweptHashes);
+			}, this.options.hashSweepInterval).unref();
+		}
+
+		if (this.options.handlerSweepInterval !== 0 && this.options.handlerSweepInterval !== Infinity) {
+			validateMaxInterval(this.options.handlerSweepInterval);
+			setInterval(() => {
```
kyranet, line 241: Same here.

kyranet (changes requested):

packages/rest/src/lib/RequestManager.ts
```diff
@@ -174,13 +184,83 @@ export class RequestManager extends EventEmitter {
...
+
+				// Fire event
+				this.emit(RESTEvents.HashSweep, sweptHashes);
+			}, this.options.hashSweepInterval).unref();
+		}
+
+		if (this.options.handlerSweepInterval !== 0 && this.options.handlerSweepInterval !== Infinity) {
+			validateMaxInterval(this.options.handlerSweepInterval);
+			this.handlerTimer = setInterval(() => {
+				const sweptHandlers: Collection<string, IHandler> = new Collection<string, IHandler>();
```
vladfrangu, line 245:
```suggestion
				const sweptHandlers = new Collection<string, IHandler>();
```

packages/rest/src/lib/RequestManager.ts
```diff
@@ -174,13 +184,83 @@ export class RequestManager extends EventEmitter {
...
+			if (interval > 14_400_000) {
+				throw new Error('Cannot set an interval greater than 4 hours');
+			}
+		};
+
+		if (this.options.hashSweepInterval !== 0 && this.options.hashSweepInterval !== Infinity) {
+			validateMaxInterval(this.options.hashSweepInterval);
+			this.hashTimer = setInterval(() => {
+				// Only allocate a swept collection for event.
+				const sweptHashes: Collection<string, HashData> = new Collection<string, HashData>();
```
vladfrangu, line 213:
```suggestion
				const sweptHashes = new Collection<string, HashData>();
```

packages/rest/src/lib/RequestManager.ts
```diff
@@ -174,13 +184,83 @@ export class RequestManager extends EventEmitter {
...
+		const validateMaxInterval = (interval: number) => {
+			if (interval > 14_400_000) {
+				throw new Error('Cannot set an interval greater than 4 hours');
+			}
+		};
+
+		if (this.options.hashSweepInterval !== 0 && this.options.hashSweepInterval !== Infinity) {
+			validateMaxInterval(this.options.hashSweepInterval);
+			this.hashTimer = setInterval(() => {
+				// Only allocate a swept collection for event.
```
vladfrangu, line 212: Comment isn't valid anymore

packages/rest/src/lib/RequestManager.ts
```diff
@@ -323,6 +406,20 @@ export class RequestManager extends EventEmitter {
 		return { url, fetchOptions };
 	}
 
+	/**
+	 * Stops the hash sweeping interval
+	 */
+	public clearHashSweeper() {
+		return clearInterval(this.hashTimer);
```
vladfrangu, line 413: Don't return this imo, just make this return void

packages/rest/src/lib/RequestManager.ts
```diff
@@ -323,6 +406,20 @@ export class RequestManager extends EventEmitter {
...
+	}
+
+	/**
+	 * Stops the request handler sweeping interval
+	 */
+	public clearHandlerSweeper() {
+		return clearInterval(this.handlerTimer);
```
vladfrangu, line 420: Ditto

vladfrangu (changes requested):

iCrawl (approved):

kyranet (approved):

vladfrangu (approved):

SpaceEEC (approved):

PR #7265, opened 2022-01-13
chore(cdn): default to animated images

suneettipirneni:

Ports #7149 to /rest

### DIFF
--- a/packages/rest/src/lib/CDN.ts
+++ b/packages/rest/src/lib/CDN.ts
@@ -24,7 +24,7 @@ export interface BaseImageURLOptions {
 }
 
 /**
- * The options used for image URLs with dynamic content
+ * The options used for image URLs with animated content
  */
 export interface ImageURLOptions extends BaseImageURLOptions {
 	/**
--- a/packages/rest/src/lib/CDN.ts
+++ b/packages/rest/src/lib/CDN.ts
@@ -201,8 +201,13 @@ export class CDN {
 	 */
 	private makeURL(
 		route: string,
-		{ allowedExtensions = ALLOWED_EXTENSIONS, extension = 'png', size }: Readonly<MakeURLOptions> = {},
+		{ allowedExtensions = ALLOWED_EXTENSIONS, extension = 'webp', size }: Readonly<MakeURLOptions> = {},
 	): string {
+		// Stickers cannot have a `.webp` extension
+		if (route.startsWith('/stickers/')) {
+			extension = 'png';
+		}
+
 		extension = String(extension).toLowerCase();
 
 		if (!allowedExtensions.includes(extension)) {
--- a/packages/rest/src/lib/CDN.ts
+++ b/packages/rest/src/lib/CDN.ts
@@ -158,7 +158,11 @@ export class CDN {
 	 * @param extension The extension of the sticker
 	 */
 	public sticker(stickerId: string, extension?: StickerExtension): string {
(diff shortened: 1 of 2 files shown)


vaporox:
The PR you mentioned removes `StaticImageURLOptions` in favor of `ImageURLOptions`, should this be done here too with `BaseImageURLOptions`?

suneettipirneni:
> The PR you mentioned removes `StaticImageURLOptions` in favor of `ImageURLOptions`, should this be done here too with `BaseImageURLOptions`?

I don't think so, mainly because djs didn't have a dedicated method for static and dynamic image url functions, while /rest does. So having `BaseImageURLOptions` still has use here.

packages/rest/src/lib/CDN.ts
```diff
@@ -201,8 +201,13 @@ export class CDN {
 	 */
 	private makeURL(
 		route: string,
-		{ allowedExtensions = ALLOWED_EXTENSIONS, extension = 'png', size }: Readonly<MakeURLOptions> = {},
+		{ allowedExtensions = ALLOWED_EXTENSIONS, extension = 'webp', size }: Readonly<MakeURLOptions> = {},
 	): string {
+		// Stickers cannot have a `.webp` extension
+		if (route.startsWith('/stickers/')) {
+			extension = 'png';
+		}
```
kyranet, line 209: We cannot set `extension: 'json'` with this, and this code is rather troublesome since it may throw false positives or replace the format in unexpected ways, plus this is running for every single route as opposed to stickers only.

kyranet (changes requested):

kyranet (approved):

iCrawl (approved):

vladfrangu (approved):

packages/rest/src/lib/CDN.ts
```diff
@@ -8,19 +8,48 @@ import {
...
+	 */
 	extension?: ImageExtension;
+	/**
+	 * The size specified in the image URL
+	 */
 	size?: ImageSize;
 }
 
+/**
+ * The options used for image URLs with dynamic content
```
Jiralite, line 27: This might be better wording now that `dynamic` is no more:
```suggestion
 * The options used for image URLs with animated content
```

Jiralite (approved):

packages/rest/src/lib/CDN.ts
```diff
@@ -158,7 +187,11 @@ export class CDN {
 	 * @param extension The extension of the sticker
 	 */
 	public sticker(stickerId: string, extension?: StickerExtension): string {
-		return this.makeURL(`/stickers/${stickerId}`, { allowedExtensions: ALLOWED_STICKER_EXTENSIONS, extension });
+		// Stickers cannot have a `.webp` extension
+		return this.makeURL(`/stickers/${stickerId}`, {
+			allowedExtensions: ALLOWED_STICKER_EXTENSIONS,
+			extension: extension ?? 'png', // Stickers cannot have a `.webp` extension, so we default to a `.png`
+		});
```
almeidx, line 194: No need to mention it twice
```suggestion
		return this.makeURL(`/stickers/${stickerId}`, {
			allowedExtensions: ALLOWED_STICKER_EXTENSIONS,
			extension: extension ?? 'png', // Stickers cannot have a `.webp` extension, so we default to a `.png`
		});
```

SpaceEEC (approved):

PR #7077, opened 2021-12-08
refactor: remove discord.js enums and use discord-api-types enums instead

suneettipirneni:
**Please describe the changes this PR makes and why it should be merged:**

TODO:

- [x] Make sure spots weren't missed
- [x] Test for any regressions

Note: I won't be updating the enums for any components. I'll wait for #7040 to be merged before I modify anything with components.

**Status and versioning classification:**
Please move lines that apply to you out of the comment:
- Code changes have been tested against the Discord API, or there are no code changes
- I know how to update typings and have done so, or typings don't need updating
- This PR includes breaking changes (methods removed or renamed, parameters moved or removed)

PR #7267, opened 2022-01-14
fix: follow type of `discord-api-types` package

Raiden-Ei:
* #7077 

After this, `CommandInteractionOptionResolver#getX` only accepts `discord-api-types`'s type. It throws an error because old thing is ALL UPPERCASE, while `discord-api-types` using PascalCase.


ImRodry:
This change was already proposed on another PR and rejected by @iCrawl so this will probably have the same faith

Raiden-Ei:
> This change was already proposed on another PR and rejected by @iCrawl so this will probably have the same faith

But it's fact that those methods are broken. how to fix the problem without following dapi-types package's way?

iCrawl:
Not use -dev is a way.

If you need interaction locales theres also 13.6.0

PR #7286, opened 2022-01-17
feat(guildemojimanager): add `delete` and `edit` methods

ceilingfans:
This PR adds the `delete` and `edit` methods to the `GuildEmojiManager` as to use these methods, you would have to fetch the emoji from the manager before calling the methods on the `GuildEmoji` itself.

- _Code changes have been tested against the Discord API,_ or there are no code changes

### DIFF
--- a/packages/discord.js/src/errors/Messages.js
+++ b/packages/discord.js/src/errors/Messages.js
@@ -113,7 +113,6 @@ const Messages = {
 
   EMOJI_TYPE: 'Emoji must be a string or GuildEmoji/ReactionEmoji',
   EMOJI_MANAGED: 'Emoji is managed and has no Author.',
-  EMOJI_RESOLVE: 'Could not resolve the emoji.',
   MISSING_MANAGE_EMOJIS_AND_STICKERS_PERMISSION: guild =>
     `Client must have Manage Emojis and Stickers permission in guild ${guild} to see emoji authors.`,
   NOT_GUILD_STICKER: 'Sticker is a standard (non-guild) sticker and has no author.',
--- a/packages/discord.js/src/errors/Messages.js
+++ b/packages/discord.js/src/errors/Messages.js
@@ -113,6 +113,7 @@ const Messages = {
 
   EMOJI_TYPE: 'Emoji must be a string or GuildEmoji/ReactionEmoji',
   EMOJI_MANAGED: 'Emoji is managed and has no Author.',
+  EMOJI_RESOLVE: 'Could not resolve the emoji',
   MISSING_MANAGE_EMOJIS_AND_STICKERS_PERMISSION: guild =>
     `Client must have Manage Emojis and Stickers permission in guild ${guild} to see emoji authors.`,
   NOT_GUILD_STICKER: 'Sticker is a standard (non-guild) sticker and has no author.',
--- a/packages/discord.js/src/errors/Messages.js
+++ b/packages/discord.js/src/errors/Messages.js
@@ -113,7 +113,7 @@ const Messages = {
 
   EMOJI_TYPE: 'Emoji must be a string or GuildEmoji/ReactionEmoji',
   EMOJI_MANAGED: 'Emoji is managed and has no Author.',
-  EMOJI_RESOLVE: 'Could not resolve the emoji',
+  EMOJI_RESOLVE: 'Could not resolve the emoji.',
   MISSING_MANAGE_EMOJIS_AND_STICKERS_PERMISSION: guild =>
     `Client must have Manage Emojis and Stickers permission in guild ${guild} to see emoji authors.`,
   NOT_GUILD_STICKER: 'Sticker is a standard (non-guild) sticker and has no author.',
--- a/packages/discord.js/src/managers/GuildEmojiManager.js
+++ b/packages/discord.js/src/managers/GuildEmojiManager.js
@@ -105,7 +105,7 @@ class GuildEmojiManager extends BaseGuildEmojiManager {
    * Deletes an emoji.
    * @param {EmojiResolvable} emoji The Emoji resolvable to delete
    * @param {string} [reason] Reason for deleting the emoji
-   * @returns {?Promise<void>}
+   * @returns {Promise<void>}
    */
   async delete(emoji, reason) {
     const id = this.resolveId(emoji);
--- a/packages/discord.js/src/managers/GuildEmojiManager.js
+++ b/packages/discord.js/src/managers/GuildEmojiManager.js
@@ -100,6 +100,45 @@ class GuildEmojiManager extends BaseGuildEmojiManager {
     for (const emoji of data) emojis.set(emoji.id, this._add(emoji, cache));
     return emojis;
   }
+
+  /**
+   * Deletes an emoji.
+   * @param {EmojiResolvable} emoji The Emoji resolvable to delete
+   * @param {string} reason Reason for deleting the emoji
+   * @returns {?Promise<GuildEmoji>}
+   */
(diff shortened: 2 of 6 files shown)


packages/discord.js/src/managers/GuildEmojiManager.js
```diff
@@ -100,6 +100,45 @@ class GuildEmojiManager extends BaseGuildEmojiManager {
...
+
+  /**
+   * Edits an emoji.
+   * @param {EmojiResolvable} emoji The Emoji resolvable to edit
+   * @param {GuildEmojiEditData} data The new data for the emoji
+   * @param {string} [reason] Reason for editing this emoji
+   * @returns {Promise<GuildEmoji>}
+   */
+  async edit(emoji, data, reason) {
+    const resolvedEmoji = this.resolve(emoji);
```
vladfrangu, line 125: Ditto

vladfrangu (changes requested):

packages/discord.js/src/managers/GuildEmojiManager.js
```diff
@@ -100,6 +100,45 @@ class GuildEmojiManager extends BaseGuildEmojiManager {
...
+    if (!resolvedEmoji) throw new TypeError('EMOJI_RESOLVE');
+    const roles = data.roles?.map(r => r.id ?? r);
```
kyranet, line 127: This should use `guild.roles.resolveId`.

kyranet (dismissed):
Please make [`GuildEmoji#edit`](https://github.com/discordjs/discord.js/blob/9054f2f7ad7f246431e5f53403535bf301c27a80/packages/discord.js/src/structures/GuildEmoji.js#L107) and [`GuildEmoji#delete`](https://github.com/discordjs/discord.js/blob/9054f2f7ad7f246431e5f53403535bf301c27a80/packages/discord.js/src/structures/GuildEmoji.js#L139) use the manager methods to reduce code duplication.

packages/discord.js/src/errors/Messages.js
```diff
@@ -113,6 +113,7 @@ const Messages = {
 
   EMOJI_TYPE: 'Emoji must be a string or GuildEmoji/ReactionEmoji',
   EMOJI_MANAGED: 'Emoji is managed and has no Author.',
+  EMOJI_RESOLVE: 'Could not resolve the emoji.',
```
ImRodry, line 116: This isn’t needed, you can use INVALID_TYPE instead

ImRodry (commented):
Funny how I was just about to make a PR that included this lol

packages/discord.js/src/managers/GuildEmojiManager.js
```diff
@@ -100,6 +100,45 @@ class GuildEmojiManager extends BaseGuildEmojiManager {
...
   }
+
+  /**
+   * Deletes an emoji.
+   * @param {EmojiResolvable} emoji The Emoji resolvable to delete
+   * @param {string} reason Reason for deleting the emoji
+   * @returns {?Promise<GuildEmoji>}
+   */
+  async delete(emoji, reason) {
+    const resolvedEmoji = this.resolve(emoji);
```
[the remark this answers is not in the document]
ceilingfans: Alright, I changed it to use that instead

packages/discord.js/src/managers/GuildEmojiManager.js
ceilingfans: Changed this too, thanks for pointing it out

packages/discord.js/src/errors/Messages.js
ceilingfans: You're right, I changed it, thanks for pointing it out

packages/discord.js/src/managers/GuildEmojiManager.js
```diff
@@ -100,6 +100,44 @@ class GuildEmojiManager extends BaseGuildEmojiManager {
     for (const emoji of data) emojis.set(emoji.id, this._add(emoji, cache));
     return emojis;
   }
+
+  /**
+   * Deletes an emoji.
+   * @param {EmojiResolvable} emoji The Emoji resolvable to delete
+   * @param {string} [reason] Reason for deleting the emoji
+   * @returns {?Promise<void>}
```
ImRodry, line 108: There’s no such thing as a nullable promise

packages/discord.js/src/managers/GuildEmojiManager.js
```diff
@@ -100,6 +100,44 @@ class GuildEmojiManager extends BaseGuildEmojiManager {
...
+   * Edits an emoji.
+   * @param {EmojiResolvable} emoji The Emoji resolvable to edit
+   * @param {GuildEmojiEditData} data The new data for the emoji
+   * @param {string} [reason] Reason for editing this emoji
+   * @returns {Promise<GuildEmoji>}
+   */
+  async edit(emoji, data, reason) {
+    emoji = this.resolve(emoji);
+    if (!emoji) throw new TypeError('INVALID_TYPE', 'emoji', 'EmojiResolvable', true);
+    const roles = data.roles?.map(r => this.guild.roles.resolveId(r) ?? r);
```
ImRodry, line 126: This should work
```suggestion
    const roles = data.roles?.map(this.guild.roles.resolveId);
```

kyranet: No, it won't due to unbound `this`.

ImRodry: Well then at least the `?? r` can be removed, right?

kyranet: Yes, because `resolveId` already resolves an ID, so if it returns null... `r` is **definitely** not valid data

ImRodry: Well then could you unresolve this?
```suggestion
    const roles = data.roles?.map(r => this.guild.roles.resolveId(r));
```

SpaceEEC (approved):

vladfrangu (changes requested):

packages/discord.js/src/managers/GuildEmojiManager.js
```diff
@@ -100,6 +100,44 @@ class GuildEmojiManager extends BaseGuildEmojiManager {
...
+
+  /**
+   * Edits an emoji.
+   * @param {EmojiResolvable} emoji The Emoji resolvable to edit
+   * @param {GuildEmojiEditData} data The new data for the emoji
+   * @param {string} [reason] Reason for editing this emoji
+   * @returns {Promise<GuildEmoji>}
+   */
+  async edit(emoji, data, reason) {
+    emoji = this.resolve(emoji);
```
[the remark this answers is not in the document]
ceilingfans: Okay, I changed it to use resolveId

vladfrangu (approved):

packages/discord.js/src/managers/GuildEmojiManager.js
```diff
@@ -100,6 +100,48 @@ class GuildEmojiManager extends BaseGuildEmojiManager {
...
+      .patch({
+        data: {
+          name: data.name,
+          roles,
+        },
+        reason,
+      });
+    const existing = this.cache.get(id);
+    if (existing) {
+      const clone = emoji._clone();
```
kyranet, line 139: `emoji` might not be an object, so this method may throw, instead, use the existing value from the cache, which we get at line 137.
```suggestion
      const clone = existing._clone();
```

kyranet (changes requested):
Just one thing, and some testing would be nice.

kyranet (approved):

PR #7288, opened 2022-01-17
BaseClient: pass `captureRejections` option

KhafraDev:
- Currently, the Client class cannot properly handle async errors being 
  thrown.
- With this change, when an error is thrown within an event, Client will
  automatically emit an error event if a listener exists, or otherwise
  exits with an unhandledRejection as it currently does..

```js
client.on('ready', () => console.log('ready'));
client.on('error', (e) => console.error(e.message)); // "whoops"

client.on('messageCreate', async (m) => {
    // do some async task here that throws an error
    // this will no longer exit the program with an
    // unhandled rejection!
    throw new Error('whoops');
});
```
https://nodejs.org/api/events.html#capture-rejections-of-promises

This is experimental, but a [PR](https://github.com/nodejs/node/commit/9a85efaa7f5f22c6905bf47d672dd73738787437) has already been committed that removes the experimental status.

### DIFF
--- a/packages/discord.js/src/client/BaseClient.js
+++ b/packages/discord.js/src/client/BaseClient.js
@@ -13,7 +13,7 @@ const Util = require('../util/Util');
  */
 class BaseClient extends EventEmitter {
   constructor(options = {}) {
-    super();
+    super({ captureRejections: true });
 
     if (typeof options !== 'object' || options === null) {
       throw new TypeError('INVALID_TYPE', 'options', 'object', true);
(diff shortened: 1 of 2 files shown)


almeidx:
Maybe add a note on the error event telling people to avoid using async

From the link you included in the description:
> The 'error' events that are generated by the captureRejections behavior do not have a catch handler to avoid infinite error loops: the recommendation is to **not use async functions as 'error' event handlers.**

SpaceEEC (approved):

iCrawl (approved):

kyranet (approved):

PR #7290, opened 2022-01-18
feat: Enum resolvers & internal enum string removal

### LINKED ISSUE #6136: option to not transform data received from the api

With [discord-api-types](https://github.com/discordjs/discord-api-types) there isn't really a reason to modify values received from the api anymore. Obviously object keys should still be modified (guild_id -> guildId, ...) but channel types (for example) should not be transformed from numbers -> strings.

For example:
```ts
import { ChannelType } from 'discord-api-types';
declare const channel: import('discord.js').GuildChannel;

// current method
if (channel.type === 'TEXT') { ... }

// proposed
if (channel.type === ChannelType.GUILD_TEXT) { ... }
```

N/A

**Additional context**

It could also be an option passed to the client? For instance:
```ts
const client = new Client({ ..., transform: false }); // transform defaults to true
```
this way it wouldn't be a breaking change

The benefits of doing this would be less maintenance burden (and the bugs that come with it), for example:
https://github.com/discordjs/discord.js/pull/5251/files

(there are others, just an example)

### LINKED ISSUE #7276: The type of structures are different

discord.js

### Issue description

The type of structures declaration in Constants are different with the type of structures declaration

### Code sample

```typescript
// Constants
exports.ThreadChannelTypes = ['GUILD_NEWS_THREAD', 'GUILD_PUBLIC_THREAD', 'GUILD_PRIVATE_THREAD'];
exports.TextBasedChannelTypes = [
  'DM',
  'GUILD_TEXT',
  'GUILD_NEWS',
  'GUILD_NEWS_THREAD',
  'GUILD_PUBLIC_THREAD',
  'GUILD_PRIVATE_THREAD',
];

// index.d.ts
ThreadChannelTypes: ThreadChannelTypeKey[]; // ['GuildNewsThread', 'GuildPublicThread', 'GuildPrivateThread']
TextBasedChannelTypes: TextBasedChannelTypes[]; // ['DM', 'GuildText', 'GuildNews', 'GuildNewsThread', 'GuildPublicThread', 'GuildPrivateThread'];

// And more...
```

### Package version

14.0.0-dev.1642205165.7c07976

### Node.js version

16.13.1

### Operating system

_No response_

No Partials

No Intents

7c07976

suneettipirneni:

TODO:
- [x] Preliminary Testing

This makes discord.js no longer use string representations of enums internally. As a result if strings are still desired to be used, you must use the an enum resolver:

```ts
import { EnumResolvers } from 'discord.js';
const { resolveChannelType } = EnumResolvers
...

category.createChannel({ type: resolveChannelType('GUILD_TEXT') });
```

Otherwise you can use the enums directly in your own code:

```ts
import { ChannelType } from 'discord.js';
...

category.createChannel({ type: ChannelType.GuildText });
```

Closes #6136
Closes #7276


KhafraDev:
closes #6136?

suneettipirneni:
> closes #6136?

yeah

Syjalo:
Also may be closes #7276

packages/discord.js/src/structures/Channel.js
```diff
@@ -27,7 +27,8 @@ class Channel extends Base {
      * The type of the channel
```
Syjalo, line 27: The Channel type has not been changed (I can't reach it)

Syjalo:
Why do we need resolvers, can't we just export the enums from `discord-api-type` in index.js?

packages/discord.js/src/util/EnumResolvers.js
```diff
@@ -0,0 +1,223 @@
...
+    }
+
+    return func(key);
+  };
+}
+
+// Injects wrapper into class static methods.
+function applyPreconditioner(obj) {
+  Object.getOwnPropertyNames(obj).forEach(name => {
+    console.log(name);
```
Syjalo, line 211:
```suggestion
```
Should be removed

Syjalo (changes requested):
Debug `console.log` that got into PR

packages/discord.js/src/structures/ContextMenuCommandInteraction.js
```diff
@@ -29,9 +29,9 @@ class ContextMenuCommandInteraction extends CommandInteraction {
 
     /**
      * The type of the target of the interaction; either USER or MESSAGE
-     * @type {ApplicationCommandType}
+     * @type {ApplicationCommandType.User|ApplicationCommandType.Message}
```
ImRodry, line 32: Will this render properly in the docs?

packages/discord.js/src/util/EnumResolvers.js
```diff
@@ -0,0 +1,223 @@
...
+        return ApplicationCommandOptionType.Integer;
+      case 'BOOLEAN':
+        return ApplicationCommandOptionType.Boolean;
+      case 'USER':
+        return ApplicationCommandOptionType.User;
+      case 'CHANNEL':
+        return ApplicationCommandOptionType.Channel;
+      case 'ROLE':
+        return ApplicationCommandOptionType.Role;
+      default:
```
ImRodry, line 93: Missed number and mentionable

packages/discord.js/src/util/EnumResolvers.js
```diff
@@ -0,0 +1,223 @@
...
+      return key;
+    }
+
+    return func(key);
+  };
+}
+
+// Injects wrapper into class static methods.
+function applyPreconditioner(obj) {
+  Object.getOwnPropertyNames(obj).forEach(name => {
```
ImRodry, line 210: I don’t recall seeing many uses of forEach in the library, wouldn’t it be better to use for of?

packages/discord.js/typings/index.d.ts
```diff
@@ -837,6 +831,18 @@ export class DataResolver extends null {
...
+  private constructor();
+  public static resolveChannelType(key: string | number): ChannelType;
+  public static resolveInteractionType(key: string | number): InteractionType;
+  public static resolveApplicationCommandType(key: string | number): ApplicationCommandType;
+  public static resolveApplicationCommandOptionType(key: string | number): ApplicationCommandOptionType;
+  public static resolveApplicationCommandPermissionType(key: string | number): ApplicationCommandPermissionType;
+  public static resolveComponentType(key: string | number): ComponentType;
+  public static resolveButtonStyle(key: string | number): ButtonStyle;
+  public static resolveMessageType(key: string | number): MessageType;
+}
```
ImRodry, line 844: How about the privates?

packages/discord.js/typings/index.d.ts
```diff
@@ -2837,23 +2832,20 @@ export class GuildApplicationCommandManager extends ApplicationCommandManager<Ap
...
 
 export class GuildChannelManager extends CachedManager<Snowflake, GuildBasedChannel, GuildChannelResolvable> {
   private constructor(guild: Guild, iterable?: Iterable<RawGuildChannelData>);
   public readonly channelCountWithoutThreads: number;
   public guild: Guild;
   /** @deprecated See [Self-serve Game Selling Deprecation](https://support-dev.discord.com/hc/en-us/articles/4414590563479) for more information */
   public create(name: string, options: GuildChannelCreateOptions & { type: 'GuildStore' }): Promise<StoreChannel>;
-  public create<T extends GuildChannelTypes>(
+  /** @deprecated See [Self-serve Game Selling Deprecation](https://support-dev.discord.com/hc/en-us/articles/4414590563479) for more information */
+  public create<T extends CategoryChannelType | ChannelType.GuildCategory>(
```
ImRodry, line 2848: This isn’t correct, also above type wasn’t changed

packages/discord.js/typings/index.d.ts
```diff
@@ -1507,17 +1504,17 @@ export class Message<Cached extends boolean = boolean> extends Base {
   public system: boolean;
   public readonly thread: ThreadChannel | null;
   public tts: boolean;
-  public type: MessageTypeKey;
+  public type: MessageType;
```
ImRodry, line 1507: What’s `MessageType` (same for other types) and why can we use it without `keyof typeof`?

ImRodry (changes requested):

suneettipirneni:
> Why do we need resolvers, can't we just export the enums from `discord-api-type` in index.js?

They are reexported, but they're const enums. That means you can't use reverse key mapping with them.

packages/discord.js/src/structures/ApplicationCommand.js
```diff
@@ -325,11 +325,8 @@ class ApplicationCommand extends Base {
     }
 
     if (existing.channelTypes) {
-      const newTypes = (option.channelTypes ?? option.channel_types).map(type =>
-        typeof type === 'number' ? ChannelType[type] : type,
-      );
       for (const type of existing.channelTypes) {
-        if (!newTypes.includes(type)) return false;
+        if (!option.channelTypes.includes(type)) return false;
```
vladfrangu, line 329: This doesn't handle builder data anymore because of the `option.channelTypes ?? option.channel_types` condition missing

packages/discord.js/src/structures/Channel.js
```diff
@@ -27,7 +27,8 @@ class Channel extends Base {
...
      * @type {ChannelType}
      */
-    this.type = type ?? 'UNKNOWN';
+    // TODO: Better way of handling unknown?
```
vladfrangu, line 30: We won't have Channels without types anyways, let alone unknown types.

packages/discord.js/src/structures/MessageMentions.js
```diff
@@ -112,13 +111,11 @@ class MessageMentions {
...
       } else {
         this.crosspostedChannels = new Collection();
-        const channelTypes = Object.keys(ChannelType);
         for (const d of crosspostedChannels) {
-          const type = channelTypes[d.type];
           this.crosspostedChannels.set(d.id, {
             channelId: d.id,
             guildId: d.guild_id,
-            type: type ?? 'UNKNOWN',
+            type: d.type ?? -1,
```
vladfrangu, line 118: When will type be undefined?

packages/discord.js/src/util/EnumResolvers.js
```diff
@@ -0,0 +1,223 @@
...
+  ComponentType,
+  ButtonStyle,
+  ApplicationCommandOptionType,
+  ChannelType,
+  ApplicationCommandPermissionType,
+  MessageType,
+} = require('discord-api-types/v9');
+
+function unknownKeyStrategy(val) {
+  console.log(`Could not resolve enum value for ${val}`);
```
vladfrangu, line 15: Shouldn't we throw instead? `-1` can yield undefined behavior

packages/discord.js/src/util/EnumResolvers.js
```diff
@@ -0,0 +1,223 @@
...
+  return -1;
+}
+
+class EnumResolvers extends null {
+  static resolveChannelType(key) {
```
vladfrangu, line 20: You should define typedefs for all of these key parameters

packages/discord.js/typings/index.d.ts
suneettipirneni: `MessageType` is an enum from -types. I don't understand the second question, nothing should be using `keyof` anymore since djs doesn't accept strings directly.

packages/discord.js/src/util/EnumResolvers.js
```diff
@@ -0,0 +1,223 @@
...
+    if (typeof key === 'number') {
+      return key;
+    }
+
+    return func(key);
+  };
+}
+
+// Injects wrapper into class static methods.
+function applyPreconditioner(obj) {
```
vladfrangu, line 209: Not a blocker but I can't wait for TS for this 😂

packages/discord.js/typings/index.d.ts
```diff
@@ -473,43 +473,37 @@ export class ButtonInteraction<Cached extends CacheType = CacheType> extends Mes
...
+  | ChannelType.GuildPublicThread
+  | ChannelType.GuildNewsThread
+  | ChannelType.GuildPrivateThread
+  | ChannelType.GuildCategory
 >;
 
 export class CategoryChannel extends GuildChannel {
   public readonly children: Collection<Snowflake, Exclude<NonThreadGuildBasedChannel, CategoryChannel>>;
-  public type: 'GuildCategory';
+  public type: ChannelType.GuildCategory;
```
vladfrangu, line 496: Does this actually enforce it to the value? I know in TS we had to cast as const

packages/discord.js/typings/index.d.ts
vladfrangu: What about them? 👀

packages/discord.js/typings/index.d.ts
```diff
@@ -837,6 +831,18 @@ export class DataResolver extends null {
...
+export class EnumResolvers extends null {
+  private constructor();
+  public static resolveChannelType(key: string | number): ChannelType;
+  public static resolveInteractionType(key: string | number): InteractionType;
+  public static resolveApplicationCommandType(key: string | number): ApplicationCommandType;
+  public static resolveApplicationCommandOptionType(key: string | number): ApplicationCommandOptionType;
+  public static resolveApplicationCommandPermissionType(key: string | number): ApplicationCommandPermissionType;
+  public static resolveComponentType(key: string | number): ComponentType;
+  public static resolveButtonStyle(key: string | number): ButtonStyle;
+  public static resolveMessageType(key: string | number): MessageType;
```
vladfrangu, line 843: All of these should have `number` be the actual enum instead.

packages/discord.js/typings/index.d.ts
```diff
@@ -2837,23 +2832,20 @@ export class GuildApplicationCommandManager extends ApplicationCommandManager<Ap
...
-    GuildCategory: CategoryChannel;
-  }
-> &
-  MappedChannelCategoryTypes;
+export type MappedGuildChannelTypes = {
+  [ChannelType.GuildCategory]: CategoryChannel;
+} & MappedChannelCategoryTypes;
 
-export type GuildChannelTypes = CategoryChannelTypes | ChannelType.GuildCategory | 'GuildCategory';
+export type GuildChannelTypes = CategoryChannelType | ChannelType.GuildCategory | 'GuildCategory';
```
vladfrangu, line 2839: Whats `GuildCategory` doing here?
```suggestion
export type GuildChannelTypes = CategoryChannelType | ChannelType.GuildCategory;
```

packages/discord.js/typings/index.d.ts
```diff
@@ -2837,23 +2832,20 @@ export class GuildApplicationCommandManager extends ApplicationCommandManager<Ap
...
 
 export class GuildChannelManager extends CachedManager<Snowflake, GuildBasedChannel, GuildChannelResolvable> {
   private constructor(guild: Guild, iterable?: Iterable<RawGuildChannelData>);
   public readonly channelCountWithoutThreads: number;
   public guild: Guild;
   /** @deprecated See [Self-serve Game Selling Deprecation](https://support-dev.discord.com/hc/en-us/articles/4414590563479) for more information */
   public create(name: string, options: GuildChannelCreateOptions & { type: 'GuildStore' }): Promise<StoreChannel>;
```
vladfrangu, line 2846: & type is wrong

vladfrangu (changes requested):

packages/discord.js/typings/index.d.ts
ImRodry: Why does it not?

suneettipirneni: > This makes discord.js no longer use string representations of enums internally. As a result if strings are still desired to be used, you must use the an enum resolver

ImRodry: I hadn’t read that, that’s very annoying
Couldn’t you use these resolvers whenever there’s user input to allow us to pass strings?

packages/discord.js/src/structures/Channel.js
ImRodry: If a new channel type is introduced it would be unknown wouldn’t it…

packages/discord.js/typings/index.d.ts
suneettipirneni: No, we won't be using resolvers internally. You either will need to use the enum or the resolver.

packages/discord.js/typings/index.d.ts
suneettipirneni: Yeah not entirely sure what you're asking here.

packages/discord.js/src/structures/Channel.js
suneettipirneni: Not sure what you mean here, do you mind elaborating 😀

almeidx: image

packages/discord.js/typings/index.d.ts
ImRodry: They’re not typed here when they should be, right?

packages/discord.js/src/structures/Channel.js
suneettipirneni: ah ok I was looking below not above

packages/discord.js/typings/index.d.ts
suneettipirneni: Yeah it complains if it isn't an enum value.

packages/discord.js/typings/index.d.ts
suneettipirneni: which private fields? There are none

packages/discord.js/src/util/Constants.js
```diff
@@ -1,6 +1,7 @@
 'use strict';
 
 const process = require('node:process');
+const { ChannelType } = require('discord-api-types');
```
almeidx, line 4:
```suggestion
const { ChannelType } = require('discord-api-types/v9');
```

almeidx (changes requested):

packages/discord.js/typings/index.d.ts
ImRodry: Ahhh mb the functions are outside of the class, didn’t notice

packages/discord.js/src/index.js
```diff
@@ -16,6 +16,7 @@ exports.BitField = require('./util/BitField');
 exports.Collection = require('@discordjs/collection').Collection;
 exports.Constants = require('./util/Constants');
 exports.DataResolver = require('./util/DataResolver');
+exports.EnumResolvers = require('./util/EnumResolvers');
 exports.DiscordAPIError = require('./rest/DiscordAPIError');
```
vaporox, line 20:
```suggestion
exports.DiscordAPIError = require('./rest/DiscordAPIError');
exports.EnumResolvers = require('./util/EnumResolvers');
```
The rest of this list is sorted alphabetically

packages/discord.js/src/structures/AutocompleteInteraction.js
```diff
@@ -64,7 +64,7 @@ class AutocompleteInteraction extends Interaction {
   transformOption(option) {
     const result = {
       name: option.name,
-      type: ApplicationCommandOptionType[option.type],
+      type: option.type,
```
vaporox, line 67: This change makes the method useless, you could probably remove it completely

vaporox (changes requested):
Next to this, there's also another use inside `CommandInteraction#transformOption` that needs to be updated

packages/discord.js/src/structures/Webhook.js
```diff
@@ -59,7 +58,7 @@ class Webhook {
        * The type of the webhook
        * @type {WebhookType}
        */
-      this.type = WebhookType[data.type];
+      this.type = data.type;
```
muchnameless, line 61: the typeguards in this class (`isChannelFollower` and `isIncoming`) need to be updated (again)

almeidx:
Missed this: https://github.com/suneettipirneni/discord.js/blob/1dd64f9324c8274f86f8562e7554774ab86d41f0/packages/discord.js/src/structures/GuildChannel.js#L314

almeidx:
Here are a few more:

https://github.com/suneettipirneni/discord.js/blob/1dd64f9324c8274f86f8562e7554774ab86d41f0/packages/discord.js/src/structures/GuildChannel.js#L329

https://github.com/suneettipirneni/discord.js/blob/1dd64f9324c8274f86f8562e7554774ab86d41f0/packages/discord.js/src/structures/ApplicationCommand.js#L51

https://github.com/suneettipirneni/discord.js/blob/53b644964888f01a670d182ff8fb99b370af243e/packages/discord.js/src/managers/ApplicationCommandManager.js#L210

https://github.com/suneettipirneni/discord.js/blob/53b644964888f01a670d182ff8fb99b370af243e/packages/discord.js/src/managers/ApplicationCommandPermissionsManager.js#L403

This if and the others one below as well: https://github.com/suneettipirneni/discord.js/blob/53b644964888f01a670d182ff8fb99b370af243e/packages/discord.js/src/managers/GuildManager.js#L184

https://github.com/suneettipirneni/discord.js/blob/53b644964888f01a670d182ff8fb99b370af243e/packages/discord.js/src/managers/GuildManager.js#L194

https://github.com/suneettipirneni/discord.js/blob/53b644964888f01a670d182ff8fb99b370af243e/packages/discord.js/src/managers/GuildManager.js#L207

All of the type guards here: https://github.com/suneettipirneni/discord.js/blob/53b644964888f01a670d182ff8fb99b370af243e/packages/discord.js/src/structures/Channel.js#L108

https://github.com/suneettipirneni/discord.js/blob/53b644964888f01a670d182ff8fb99b370af243e/packages/discord.js/src/struc tures/ClientPresence.js#L56 - also on the else if below this one - should probably use the enum on line 51 instead of a magic number too

https://github.com/suneettipirneni/discord.js/blob/53b644964888f01a670d182ff8fb99b370af243e/packages/discord.js/src/structures/CommandInteraction.js#L154

https://github.com/suneettipirneni/discord.js/blob/53b644964888f01a670d182ff8fb99b370af243e/packages/discord.js/src/structures/Guild.js#L744

https://github.com/suneettipirneni/discord.js/blob/53b644964888f01a670d182ff8fb99b370af243e/packages/discord.js/src/structures/GuildAuditLogs.js#L346 - and the one on the case under this one - also, is `Number()` needed on the switch condition?

https://github.com/suneettipirneni/discord.js/blob/53b644964888f01a670d182ff8fb99b370af243e/packages/discord.js/src/structures/Interaction.js#L157 - and on the isSelectMenu() guard

https://github.com/suneettipirneni/discord.js/blob/53b644964888f01a670d182ff8fb99b370af243e/packages/discord.js/src/structures/PermissionOverwrites.js#L74

packages/discord.js/src/structures/ContextMenuCommandInteraction.js
```diff
@@ -28,10 +28,10 @@ class ContextMenuCommandInteraction extends CommandInteraction {
     this.targetId = data.data.target_id;
 
     /**
-     * The type of the target of the interaction; either USER or MESSAGE
+     * The type of the target of the interaction; either `.User` or `.Message`
      * @type {ApplicationCommandType}
```
almeidx, line 32: Shouldn't the type be `ApplicationCommandType.User|ApplicationCommandType.Message`?

packages/discord.js/src/structures/Guild.js
```diff
@@ -1306,8 +1290,8 @@ class Guild extends AnonymousGuild {
     return Util.discordSort(
       this.channels.cache.filter(
         c =>
-          (['GUILD_TEXT', 'GUILD_NEWS', 'GUILD_STORE'].includes(channel.type)
-            ? ['GUILD_TEXT', 'GUILD_NEWS', 'GUILD_STORE'].includes(c.type)
+          ([ChannelType.GuildText, ChannelType.GuildNews, ChannelType.GuildStore].includes(channel.type)
+            ? [ChannelType.GuildText, ChannelType.GuildNews, ChannelType.GuildStore].includes(c.type)
```
almeidx, line 1294: Use a variable instead of repeating twice?

vladfrangu (approved):

kyranet (approved):

packages/discord.js/src/structures/Interaction.js
```diff
@@ -154,85 +154,79 @@ class Interaction extends Base {
...
   /**
    * Indicates whether this interaction is a {@link ButtonInteraction}.
    * @returns {boolean}
    */
   isButton() {
-    return (
-      InteractionType[this.type] === InteractionType.MessageComponent &&
-      ComponentType[this.componentType] === ComponentType.Button
-    );
+    return this.type === InteractionType.MessageComponent && this.componentType === ComponentType.Button;
```
almeidx, line 213:
```suggestion
    return this.isMessageComponent() && this.componentType === ComponentType.Button;
```

packages/discord.js/src/structures/Interaction.js
```diff
@@ -154,85 +154,79 @@ class Interaction extends Base {
...
   /**
    * Indicates whether this interaction is a {@link SelectMenuInteraction}.
    * @returns {boolean}
    */
   isSelectMenu() {
-    return (
-      InteractionType[this.type] === InteractionType.MessageComponent &&
-      ComponentType[this.componentType] === ComponentType.SelectMenu
-    );
+    return this.type === InteractionType.MessageComponent && this.componentType === ComponentType.SelectMenu;
```
almeidx, line 221:
```suggestion
    return this.isMessageComponent() && this.componentType === ComponentType.SelectMenu;
```

packages/discord.js/src/structures/Interaction.js
```diff
@@ -154,85 +154,79 @@ class Interaction extends Base {
...
+    return this.type === InteractionType.ApplicationCommand;
   }
 
   /**
    * Indicates whether this interaction is a {@link ChatInputCommandInteraction}.
    * @returns {boolean}
    */
   isChatInputCommand() {
-    return InteractionType[this.type] === InteractionType.ApplicationCommand && typeof this.targetId === 'undefined';
+    return this.type === InteractionType.ApplicationCommand && typeof this.targetId === 'undefined';
```
almeidx, line 165:
```suggestion
    return this.isCommand() && typeof this.targetId === 'undefined';
```

packages/discord.js/src/structures/Interaction.js
```diff
@@ -154,85 +154,79 @@ class Interaction extends Base {
...
   }
 
   /**
    * Indicates whether this interaction is a {@link ContextMenuCommandInteraction}
    * @returns {boolean}
    */
   isContextMenuCommand() {
-    return InteractionType[this.type] === InteractionType.ApplicationCommand && typeof this.targetId !== 'undefined';
+    return this.type === InteractionType.ApplicationCommand && typeof this.targetId !== 'undefined';
```
almeidx, line 173:
```suggestion
    return this.isCommand() && typeof this.targetId !== 'undefined';
```

almeidx (changes requested):

iCrawl (approved):

almeidx:
This fixes #7276 too

SpaceEEC (changes requested):
I found one more thing: https://github.com/suneettipirneni/discord.js/blob/2bfc94b8e92d1aa56d9ccaf167692cfa02eeb643/packages/discord.js/src/util/Constants.js#L87

Otherwise this is probably good to go

SpaceEEC (approved):
🎉

PR #7299, opened 2022-01-18
feat(Collection): add merging functions

1Computer1:

#7211 removes a use-case from `intersection`, so might as well just cover them all with `merge`.
Then `combineEntries` is basically `merge` but for a bunch of entries instead of two collections.

Prior art [here](https://hackage.haskell.org/package/containers-0.6.4.1/docs/Data-Map-Lazy.html#v:intersection) and [here](https://hackage.haskell.org/package/containers-0.6.4.1/docs/Data-Map-Merge-Lazy.html#v:merge) and [here](https://hackage.haskell.org/package/containers-0.6.4.1/docs/Data-Map-Lazy.html#v:fromListWith).

@suneettipirneni 

### DIFF
--- a/packages/collection/src/index.ts
+++ b/packages/collection/src/index.ts
@@ -695,7 +695,7 @@ export class Collection<K, V> extends Map<K, V> {
 		other: ReadonlyCollection<K, T>,
 		whenInSelf: (value: V, key: K) => Keep<R>,
 		whenInOther: (valueOther: T, key: K) => Keep<R>,
-		whenInBoth: (value: V, valueOther: T, key: K) => Keep<R>
+		whenInBoth: (value: V, valueOther: T, key: K) => Keep<R>,
 	): Collection<K, R> {
 		const coll = new this.constructor[Symbol.species]<K, R>();
 		const keys = new Set([...this.keys(), ...other.keys()]);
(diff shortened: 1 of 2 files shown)


iCrawl (approved):

SpaceEEC (approved):

packages/collection/src/index.ts
```diff
@@ -666,6 +666,54 @@ export class Collection<K, V> extends Map<K, V> {
...
+			if (this.has(k) && other.has(k)) {
+				const r = whenInBoth(this.get(k)!, other.get(k)!, k);
+				if (r.keep) coll.set(k, r.value);
+			} else if (this.has(k)) {
+				const r = whenInSelf(this.get(k)!, k);
+				if (r.keep) coll.set(k, r.value);
+			} else if (other.has(k)) {
+				const r = whenInOther(other.get(k)!, k);
+				if (r.keep) coll.set(k, r.value);
+			}
```
vladfrangu, line 712: Can you cache the results of this.has/other.has, and re-use them? 
```suggestion
			const hasInSelf = this.has(k);
			const hasInOther = other.has(k);

			if (hasInSelf && hasInOther) {
				const r = whenInBoth(this.get(k)!, other.get(k)!, k);
				if (r.keep) coll.set(k, r.value);
			} else if (hasInSelf) {
				const r = whenInSelf(this.get(k)!, k);
				if (r.keep) coll.set(k, r.value);
			} else if (hasInOther) {
				const r = whenInOther(other.get(k)!, k);
				if (r.keep) coll.set(k, r.value);
			}
```

vladfrangu (changes requested):

kyranet (approved):

PR #7306, opened 2022-01-19
feat(threadchannel): add `createdTimestamp` field

suneettipirneni:

This also adds an `isPrivate` type guard for threads.

### DIFF
--- a/packages/discord.js/src/structures/ThreadChannel.js
+++ b/packages/discord.js/src/structures/ThreadChannel.js
@@ -214,7 +214,7 @@ class ThreadChannel extends Channel {
    * @readonly
    */
   get createdAt() {
-    return this.createdTimestamp === null ? null : new Date(this.createdTimestamp);
+    return this.createdTimestamp && new Date(this.createdTimestamp);
   }
 
   /**
--- a/packages/discord.js/src/structures/ThreadChannel.js
+++ b/packages/discord.js/src/structures/ThreadChannel.js
@@ -110,7 +110,7 @@ class ThreadChannel extends Channel {
          */
         this.createTimestamp = Date.parse(data.thread_metadata.create_timestamp);
       } else {
-        this.createTimestamp = null;
+        this.createTimestamp ??= null;
       }
     } else {
       this.locked ??= null;
--- a/packages/discord.js/src/structures/ThreadChannel.js
+++ b/packages/discord.js/src/structures/ThreadChannel.js
@@ -214,8 +214,16 @@ class ThreadChannel extends Channel {
    * @readonly
    */
   get createdAt() {
+    if (this.createdTimestamp !== null) {
+      return new Date(this.createdTimestamp);
+    }
+
     // Private threads can rely on their ID for creation date
-    return (this.createdTimestamp ??= this.type === ChannelType.PrivateThread ? super.createdTimestamp : null);
+    if (this.type === ChannelType.PrivateThread) {
+      return super.createdAt;
+    }
+
+    return null;
   }
 
   /**
--- a/packages/discord.js/src/structures/ThreadChannel.js
+++ b/packages/discord.js/src/structures/ThreadChannel.js
@@ -214,7 +214,7 @@ class ThreadChannel extends Channel {
    * @readonly
    */
   get createdAt() {
-    return this.createdTimestamp === null ? null : super.createdAt;
+    return this.createdTimestamp === null ? null : new Date(this.createdTimestamp);
   }
 
   /**
--- a/packages/discord.js/src/structures/ThreadChannel.js
+++ b/packages/discord.js/src/structures/ThreadChannel.js
@@ -526,7 +526,7 @@ class ThreadChannel extends Channel {
   }
 
   /**
-   * Whether or not this thread is a private thread or not
+   * Whether this thread is a private thread
    * @returns {boolean}
    */
   isPrivate() {
--- a/packages/discord.js/src/structures/ThreadChannel.js
+++ b/packages/discord.js/src/structures/ThreadChannel.js
@@ -214,7 +214,8 @@ class ThreadChannel extends Channel {
    * @readonly
    */
   get createdAt() {
-    return this.createdTimestamp && new Date(this.createdTimestamp);
+    // Private threads can rely on their ID for creation date
+    return (this.createdTimestamp ??= this.type === ChannelType.PrivateThread ? super.createdTimestamp : null);
   }
 
   /**
--- a/packages/discord.js/src/structures/ThreadChannel.js
+++ b/packages/discord.js/src/structures/ThreadChannel.js
@@ -118,7 +118,7 @@ class ThreadChannel extends Channel {
       this.invitable ??= null;
     }
 
-    this.createdTimestamp ??= this.type === ChannelType.PrivateThread ? super.createdTimestamp : null;
+    this.createdTimestamp ??= this.type === ChannelType.GuildPrivateThread ? super.createdTimestamp : null;
 
     if ('owner_id' in data) {
       /**
--- a/packages/discord.js/src/structures/ThreadChannel.js
+++ b/packages/discord.js/src/structures/ThreadChannel.js
@@ -101,6 +101,17 @@ class ThreadChannel extends Channel {
        * @type {?number}
        */
       this.archiveTimestamp = Date.parse(data.thread_metadata.archive_timestamp);
+
+      if ('create_timestamp' in data.thread_metadata) {
+        /**
+         * The timestamp when this thread was created. This isn't available for threads
+         * created before 2022-01-09
+         * @type {?number}
+         */
+        this.createTimestamp = Date.parse(data.thread_metadata.create_timestamp);
+      } else {
+        this.createTimestamp = null;
+      }
     } else {
       this.locked ??= null;
       this.archived ??= null;
--- a/packages/discord.js/src/structures/ThreadChannel.js
+++ b/packages/discord.js/src/structures/ThreadChannel.js
@@ -213,16 +213,7 @@ class ThreadChannel extends Channel {
    * @readonly
    */
   get createdAt() {
-    if (this.createdTimestamp !== null) {
-      return new Date(this.createdTimestamp);
-    }
-
-    // Private threads can rely on their ID for creation date
-    if (this.type === ChannelType.PrivateThread) {
-      return super.createdAt;
-    }
-
-    return null;
+    return this.createdTimestamp && new Date(this.createdTimestamp);
   }
 
   /**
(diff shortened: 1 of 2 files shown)


packages/discord.js/src/structures/ThreadChannel.js
```diff
@@ -101,6 +101,17 @@ class ThreadChannel extends Channel {
...
+
+      if ('create_timestamp' in data.thread_metadata) {
+        /**
+         * The timestamp when this thread was created. This isn't available for threads
+         * created before 2022-01-09
+         * @type {?number}
+         */
+        this.createTimestamp = Date.parse(data.thread_metadata.create_timestamp);
+      } else {
+        this.createTimestamp = null;
```
ImRodry, line 113:
```suggestion
        this.createTimestamp ??= null;
```

Jiralite, line 113: Is `data.thread_metadata` always defined? If not, could be possibly `undefined` still (see lines 116-120)?

Jiralite (commented):
Thoughts on adding `ThreadChannel#createdAt`?

packages/discord.js/src/structures/ThreadChannel.js
ImRodry: Yes it is, see line 71

packages/discord.js/src/structures/ThreadChannel.js
```diff
@@ -101,6 +101,17 @@ class ThreadChannel extends Channel {
...
+
+      if ('create_timestamp' in data.thread_metadata) {
+        /**
+         * The timestamp when this thread was created. This isn't available for threads
+         * created before 2022-01-09
+         * @type {?number}
+         */
+        this.createTimestamp = Date.parse(data.thread_metadata.create_timestamp);
+      } else {
+        this.createTimestamp ??= null;
```
almeidx, line 113: This should probably be set on the else statement right below this one as well

suneettipirneni:
> Thoughts on adding `ThreadChannel#createdAt`?

Added

packages/discord.js/typings/index.d.ts
```diff
@@ -513,7 +513,7 @@ export type CategoryChannelResolvable = Snowflake | CategoryChannel;
 export abstract class Channel extends Base {
   public constructor(client: Client, data?: RawChannelData, immediatePatch?: boolean);
   public readonly createdAt: Date;
-  public readonly createdTimestamp: number;
+  public readonly createdTimestamp: number | null;
```
almeidx, line 516: Was this intentional? If yes, then `createdAt` would be nullable too

suneettipirneni: true

packages/discord.js/src/structures/ThreadChannel.js
```diff
@@ -101,7 +101,19 @@ class ThreadChannel extends Channel {
...
+
+      if ('create_timestamp' in data.thread_metadata) {
+        /**
+         * The timestamp when this thread was created. This isn't available for threads
+         * created before 2022-01-09
+         * @type {?number}
+         */
+        this.createdTimestamp = Date.parse(data.thread_metadata.create_timestamp);
+      } else {
+        this.createTimestamp ??= null;
```
almeidx, line 113:
```suggestion
        this.createdTimestamp ??= null;
```

packages/discord.js/src/structures/ThreadChannel.js
```diff
@@ -196,6 +208,15 @@ class ThreadChannel extends Channel {
     return this.archiveTimestamp && new Date(this.archiveTimestamp);
   }
 
+  /**
+   * The time the thread was created at
+   * @type {?Date}
+   * @readonly
+   */
+  get createdAt() {
+    return this.createdTimestamp === null ? null : new Date(this.createdTimestamp);
```
almeidx, line 217:
```suggestion
    return this.createdTimestamp && new Date(this.createdTimestamp);
```

packages/discord.js/src/structures/ThreadChannel.js
```diff
@@ -101,7 +101,19 @@ class ThreadChannel extends Channel {
...
+
+      if ('create_timestamp' in data.thread_metadata) {
+        /**
+         * The timestamp when this thread was created. This isn't available for threads
+         * created before 2022-01-09
+         * @type {?number}
+         */
+        this.createdTimestamp = Date.parse(data.thread_metadata.create_timestamp);
+      } else {
+        this.createdTimestamp ??= null;
```
ckohen, line 113: Okay this is gonna be slightly painful, but on private threads you are guaranteed the id of the thread is the created timestamp (can't create a private thread from an existing message), so if createdTimestamp is nullish and not present in the payload I think we can get away doing  `this.createdTimestamp ??= this.type === ChannelTypes.PrivateThread ? super.createdTimestamp :  null`

packages/discord.js/src/structures/ThreadChannel.js
```diff
@@ -196,6 +208,16 @@ class ThreadChannel extends Channel {
...
   }
 
+  /**
+   * The time the thread was created at
+   * @type {?Date}
+   * @readonly
+   */
+  get createdAt() {
+    // Private threads can rely on their ID for creation date
+    return (this.createdTimestamp ??= this.type === ChannelType.PrivateThread ? super.createdTimestamp : null);
```
kyranet, line 218: This is returning a number, not a date, also it's... assigning? I don't think we have a getter that does side-effects, and it feels wrong to be honest.

kyranet (dismissed):

packages/discord.js/typings/index.d.ts
```diff
@@ -2235,6 +2238,11 @@ export class ThreadChannel extends TextBasedChannelMixin(Channel) {
   public rateLimitPerUser: number | null;
   public type: ThreadChannelType;
   public readonly unarchivable: boolean;
+  public isPrivate(): this is this & {
+    createdTimestamp: number;
+    createdAt: Date;
```
Jiralite, line 2243: Should `createdAt: Date` here be prefixed with `readonly` or does that not matter?

packages/discord.js/src/structures/ThreadChannel.js
```diff
@@ -196,6 +208,24 @@ class ThreadChannel extends Channel {
...
+   */
+  get createdAt() {
+    if (this.createdTimestamp !== null) {
+      return new Date(this.createdTimestamp);
+    }
+
+    // Private threads can rely on their ID for creation date
+    if (this.type === ChannelType.PrivateThread) {
+      return super.createdAt;
+    }
```
almeidx, line 224: You're not doing this for `createdTimestamp` so it will still be nullable regardless of the thread type

You could just assign `this.createdTimestamp` to `super.createdTimestamp` in the constructor in case the thread is private

packages/discord.js/src/structures/ThreadChannel.js
```diff
@@ -495,6 +525,14 @@ class ThreadChannel extends Channel {
     return this.archived && (this.locked ? this.manageable : this.sendable);
   }
 
+  /**
+   * Whether or not this thread is a private thread or not
```
almeidx, line 529:
```suggestion
   * Whether this thread is a private thread
```

almeidx: Or maybe just
```suggestion
   * Whether this thread is private
```

packages/discord.js/src/structures/ThreadChannel.js
```diff
@@ -196,6 +207,24 @@ class ThreadChannel extends Channel {
...
+    if (this.createdTimestamp !== null) {
+      return new Date(this.createdTimestamp);
+    }
+
+    // Private threads can rely on their ID for creation date
+    if (this.type === ChannelType.PrivateThread) {
+      return super.createdAt;
+    }
+
+    return null;
```
almeidx, line 225:
```suggestion
    return this.createdTimestamp && new Date(this.createdTimestamp);
```

packages/discord.js/src/structures/ThreadChannel.js
```diff
@@ -109,6 +118,8 @@ class ThreadChannel extends Channel {
       this.invitable ??= null;
     }
 
+    this.createdTimestamp ??= this.type === ChannelType.PrivateThread ? super.createdTimestamp : null;
```
Syjalo, line 121:
```suggestion
    this.createdTimestamp ??= this.type === ChannelType.GuildPrivateThread ? super.createdTimestamp : null;
```

Syjalo (changes requested):

packages/discord.js/src/structures/ThreadChannel.js
```diff
@@ -196,6 +207,24 @@ class ThreadChannel extends Channel {
...
+   * The time the thread was created at
+   * @type {?Date}
+   * @readonly
+   */
+  get createdAt() {
+    if (this.createdTimestamp !== null) {
+      return new Date(this.createdTimestamp);
+    }
+
+    // Private threads can rely on their ID for creation date
```
Jiralite, line 220: Small nit for consistency!
```suggestion
    // Private threads can rely on their id for creation date
```

packages/discord.js/src/structures/ThreadChannel.js
```diff
@@ -196,6 +207,24 @@ class ThreadChannel extends Channel {
...
+    if (this.type === ChannelType.PrivateThread) {
```
Jiralite, line 221:
```suggestion
    if (this.type === ChannelType.GuildPrivateThread) {
```

Syjalo: Same as https://github.com/discordjs/discord.js/pull/7306#discussion_r789728451

packages/discord.js/src/structures/ThreadChannel.js
```diff
@@ -109,6 +118,8 @@ class ThreadChannel extends Channel {
       this.invitable ??= null;
     }
 
+    this.createdTimestamp ??= this.type === ChannelType.GuildPrivateThread ? super.createdTimestamp : null;
```
ImRodry, line 121: How is this true and how would super.createdTimestamp be defined if the one here isn’t?

vladfrangu: super would take from the base class 👀

ImRodry: Yeah I know but how would that work here?

Syjalo: > Yeah I know but how would that work here?

If a thread has a starter message the tread id will be the starter message id, else it will be unique id. We can be sure that the id is unique only when the channel type is GuildPrivateThread. So we can be sure that createdTimestamp is correct only when the channel type is GuildPrivateThread.

iCrawl (approved):

SpaceEEC (approved):

vladfrangu (approved):

kyranet (approved):

PR #7316, opened 2022-01-21
types: fix MessageMentions channel types

oof2win2:

Fixed typings to allow for other channel types other than TextBasedChannel in MessageMentions.channels, such as voice. Also removed DMChannels from those types, as you cannot mention a DMChannel - the ID of a DMChannel is the user's ID, which is only valid under a user mention.

**Status and versioning classification:** Patch

### DIFF
--- a/packages/discord.js/src/structures/MessageMentions.js
+++ b/packages/discord.js/src/structures/MessageMentions.js
@@ -161,20 +161,6 @@ class MessageMentions {
     let matches;
     while ((matches = this.constructor.CHANNELS_PATTERN.exec(this._content)) !== null) {
       const chan = this.client.channels.cache.get(matches[1]);
-      if (
-        ![
-          ChannelType.GuildText,
-          ChannelType.GuildVoice,
-          ChannelType.GuildNews,
-          ChannelType.GuildNewsThread,
-          ChannelType.GuildPublicThread,
-          ChannelType.GuildPrivateThread,
-          ChannelType.GuildStageVoice,
-        ].includes(chan.type)
-      ) {
-        // Go onto the next channel, as this one is not a text channel
-        continue;
-      }
       if (chan) this._channels.set(chan.id, chan);
     }
     return this._channels;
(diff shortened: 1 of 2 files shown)


imranbarbhuiya:
You can mention a dm channel. Also dm channels I'd isn't same as user ID.

oof2win2:
Taking the ID from the URL and pasting it as a usual channel mention doesn't work, it only sends the name of the DM recipient.
Screenshot 2022-01-21 at 17 37 33
Screenshot 2022-01-21 at 17 39 21

Jiralite:
It doesn't matter what happens in your client - `DMChannel`s can be parsed if you use the channel's id. If you look at our implementation:

https://github.com/discordjs/discord.js/blob/c10afeadc702ab98bec5e077b3b92494a9596f9c/packages/discord.js/src/structures/MessageMentions.js#L157-L166

You can see that we match via a regular expression here. So yes, a `DMChannel` can be matched here but also any other kind of channel - `StageChannel`... `StoreChannel`... so the typings here still aren't correct, however.

Jiralite:
The opposite actually: the JavaScript isn't the issue, the typings are.

oof2win2:
Should DMChannels still be excluded? IMO they should since you shouldn't be able to mention them, but it's a channel type.

Syjalo:
> Should DMChannels still be excluded? IMO they should since you shouldn't be able to mention them, but it's a channel type.

It's shouldn't be excluded. You able to mention it, but the desktop version of Discord don't resolves it. In mobile version of Discord the DMChannel mentions works fine. Also discord.js includes the mention to the channels collection.

oof2win2:
I see. The current code in this PR allows for any channel to be mentioned.

SpaceEEC (approved):

iCrawl (approved):

vladfrangu (approved):

kyranet (approved):

PR #7317, opened 2022-01-21
feat(minor): add application_id to Webhook

oof2win2:
Add the `application_id` property to the Webhook class. It is available with the Discord API itself, however, it is not exposed in discord.js

**Status and versioning classification:** Minor

### DIFF
--- a/packages/discord.js/src/structures/Webhook.js
+++ b/packages/discord.js/src/structures/Webhook.js
@@ -85,6 +85,16 @@ class Webhook {
       this.owner ??= null;
     }
 
+    if ('application_id' in data) {
+      /**
+       * The application that created this webhook
+       * @type {?Snowflake}
+       */
+      this.applicationId = data.application_id;
+    } else {
+      this.applicationId ??= null;
+    }
+
     if ('source_guild' in data) {
       /**
        * The source guild of the webhook
--- a/packages/discord.js/src/structures/Webhook.js
+++ b/packages/discord.js/src/structures/Webhook.js
@@ -408,15 +408,18 @@ class Webhook {
    * @returns {boolean}
    */
   isUserCreated() {
-    return Boolean(this.owner);
+    if (this.type !== WebhookType.Incoming) return false;
+    if (!this.owner) return false;
+    if (this.owner.bot) return false;
+    return true;
   }
 
   /**
-   * Whether this webhook is created by an application and not a user.
+   * Whether this webhook is created by an application.
    * @returns {boolean}
    */
   isApplicationCreated() {
-    return Boolean(this.applicationId);
+    return this.type === WebhookType.Application;
   }
 
   /**
--- a/packages/discord.js/src/structures/Webhook.js
+++ b/packages/discord.js/src/structures/Webhook.js
@@ -404,6 +404,22 @@ class Webhook {
   }
 
   /**
+   * Whether or not this webhook is created by a user
+   * @returns {boolean}
+   */
+  isUserCreated() {
+    return Boolean(this.owner);
+  }
+
+  /**
+   * Whether or not this webhook is created by an application and not a user
+   * @returns {boolean}
+   */
+  isApplicationCreated() {
+    return Boolean(this.applicationId);
+  }
+
+  /**
    * Whether or not this webhook is a channel follower webhook.
    * @returns {boolean}
    */
--- a/packages/discord.js/src/structures/Webhook.js
+++ b/packages/discord.js/src/structures/Webhook.js
@@ -404,7 +404,7 @@ class Webhook {
   }
 
   /**
-   * Whether or not this webhook is created by a user
+   * Whether or not this webhook is created by a user.
    * @returns {boolean}
    */
   isUserCreated() {
@@ -412,7 +412,7 @@ class Webhook {
   }
 
   /**
-   * Whether or not this webhook is created by an application and not a user
+   * Whether or not this webhook is created by an application and not a user.
    * @returns {boolean}
    */
   isApplicationCreated() {
--- a/packages/discord.js/src/structures/Webhook.js
+++ b/packages/discord.js/src/structures/Webhook.js
@@ -404,7 +404,7 @@ class Webhook {
   }
 
   /**
-   * Whether or not this webhook is created by a user.
+   * Whether this webhook is created by a user.
    * @returns {boolean}
    */
   isUserCreated() {
@@ -412,7 +412,7 @@ class Webhook {
   }
 
   /**
-   * Whether or not this webhook is created by an application and not a user.
+   * Whether this webhook is created by an application and not a user.
    * @returns {boolean}
    */
   isApplicationCreated() {
--- a/packages/discord.js/typings/index.d.ts
+++ b/packages/discord.js/typings/index.d.ts
@@ -2461,7 +2461,7 @@ export class Webhook extends WebhookMixin() {
     token: string | null;
   };
   public isApplicationCreated(): this is this & {
-    type: WebhookType.Incoming;
+    type: WebhookType.Application;
     applicationId: Snowflake;
     owner: User | APIUser;
   };
--- a/packages/discord.js/typings/index.d.ts
+++ b/packages/discord.js/typings/index.d.ts
@@ -2453,7 +2453,7 @@ export class Webhook extends WebhookMixin() {
   public sourceChannel: NewsChannel | APIPartialChannel | null;
   public token: string | null;
   public type: WebhookType;
-  public application_id?: Snowflake;
+  public applicationId: Snowflake | null;
   public isIncoming(): this is this & { token: string };
   public isChannelFollower(): this is this & {
     sourceGuild: Guild | APIPartialGuild;
--- a/packages/discord.js/typings/index.d.ts
+++ b/packages/discord.js/typings/index.d.ts
@@ -2455,18 +2455,25 @@ export class Webhook extends WebhookMixin() {
   public type: WebhookType;
   public applicationId: Snowflake | null;
   public isUserCreated(): this is this & {
+    type: WebhookType.Incoming;
     applicationId: null;
     owner: User | APIUser;
     token: null;
   };
   public isApplicationCreated(): this is this & {
+    type: WebhookType.Incoming;
     applicationId: Snowflake;
     owner: null;
   };
-  public isIncoming(): this is this & { token: string };
+  public isIncoming(): this is this & {
+    type: WebhookType.Incoming;
+    token: string;
+  };
   public isChannelFollower(): this is this & {
+    type: WebhookType.ChannelFollower;
     sourceGuild: Guild | APIPartialGuild;
     sourceChannel: NewsChannel | APIPartialChannel;
+    token: null;
   };
 }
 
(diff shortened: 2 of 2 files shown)


packages/discord.js/typings/index.d.ts
```diff
@@ -2453,6 +2453,7 @@ export class Webhook extends WebhookMixin() {
   public sourceChannel: NewsChannel | APIPartialChannel | null;
   public token: string | null;
   public type: WebhookType;
+  public application_id?: Snowflake;
```
Dev-CasperTheGhost, line 2456: Shouldn't it be this?

```suggestion
  public applicationId: Snowflake | null;
```

Jiralite:
From https://discord.com/developers/docs/resources/webhook#webhook-resource, it seems there is something other than the `application_id` that is missing. There is also a `user` property that is exposed. Might add that too?

Additionally, a type guard like the other webhook types could be nice to narrow down the application's id.

oof2win2:
User is already present on the class as `owner`, which I find odd, but I guess it's in line with other types of objects. Not sure what you mean with a type guard like other webhook types @Jiralite?

Jiralite:
> User is already present on the class as `owner`

Got you.

As for the type guard, look at `isIncoming()` and `isChannelFollower()` for example which narrow types based on the webhook's type. If a webhook's type is of type 3, it's owned by an application and thus, the id for sure won't be `null` (and the `owner` too likely).

oof2win2:
Added type restrictions to check for `owner`, `token`, and `applicationId`.

packages/discord.js/src/structures/Webhook.js
```diff
@@ -393,6 +403,22 @@ class Webhook {
     return this.avatar && this.client.rest.cdn.Avatar(this.id, this.avatar, options);
   }
 
+  /**
+   * Whether or not this webhook is created by a user
```
vladfrangu, line 407:
```suggestion
   * Whether or not this webhook is created by a user.
```

packages/discord.js/src/structures/Webhook.js
```diff
@@ -393,6 +403,22 @@ class Webhook {
...
+   * @returns {boolean}
+   */
+  isUserCreated() {
+    return Boolean(this.owner);
+  }
+
+  /**
+   * Whether or not this webhook is created by an application and not a user
```
vladfrangu, line 415:
```suggestion
   * Whether or not this webhook is created by an application and not a user.
```

vladfrangu (changes requested):

packages/discord.js/src/structures/Webhook.js
```diff
@@ -393,6 +403,22 @@ class Webhook {
     return this.avatar && this.client.rest.cdn.Avatar(this.id, this.avatar, options);
   }
 
+  /**
+   * Whether or not this webhook is created by a user.
```
almeidx, line 407:
```suggestion
   * Whether this webhook is created by a user.
```

packages/discord.js/src/structures/Webhook.js
```diff
@@ -393,6 +403,22 @@ class Webhook {
...
+   * @returns {boolean}
+   */
+  isUserCreated() {
+    return Boolean(this.owner);
+  }
+
+  /**
+   * Whether or not this webhook is created by an application and not a user.
```
almeidx, line 415:
```suggestion
   * Whether this webhook is created by an application and not a user.
```

almeidx: Alternatively:
```suggestion
   * Whether this webhook is created by an application.
```

almeidx (approved):

packages/discord.js/typings/index.d.ts
```diff
@@ -2453,6 +2453,16 @@ export class Webhook extends WebhookMixin() {
...
+    applicationId: null;
+    owner: User | APIUser;
+    token: null;
+  };
+  public isApplicationCreated(): this is this & {
+    applicationId: Snowflake;
+    owner: null;
+  };
   public isIncoming(): this is this & { token: string };
   public isChannelFollower(): this is this & {
```
Jiralite, line 2467: Can't the type of webhook also be narrowed here?

oof2win2: Narrowed the types for all of the checks.

packages/discord.js/typings/index.d.ts
```diff
@@ -2453,10 +2453,27 @@ export class Webhook extends WebhookMixin() {
...
+  public isUserCreated(): this is this & {
+    type: WebhookType.Incoming;
+    applicationId: null;
+    owner: User | APIUser;
+    token: null;
+  };
+  public isApplicationCreated(): this is this & {
+    type: WebhookType.Incoming;
+    applicationId: Snowflake;
+    owner: null;
```
Jiralite, line 2466: This isn't correct, if there is an application id, then the `owner` is obviously defined!

packages/discord.js/src/structures/Webhook.js
```diff
@@ -393,6 +403,22 @@ class Webhook {
     return this.avatar && this.client.rest.cdn.Avatar(this.id, this.avatar, options);
   }
 
+  /**
+   * Whether this webhook is created by a user.
+   * @returns {boolean}
+   */
+  isUserCreated() {
+    return Boolean(this.owner);
+  }
```
Jiralite, line 412: This isn't accurate - if you truly want to know if it was created by a user and not a bot, you would have to check if the `owner` is a bot too.

packages/discord.js/typings/index.d.ts
```diff
@@ -2453,10 +2453,27 @@ export class Webhook extends WebhookMixin() {
   public sourceChannel: NewsChannel | APIPartialChannel | null;
   public token: string | null;
   public type: WebhookType;
-  public isIncoming(): this is this & { token: string };
+  public applicationId: Snowflake | null;
+  public isUserCreated(): this is this & {
+    type: WebhookType.Incoming;
+    applicationId: null;
+    owner: User | APIUser;
+    token: null;
```
Jiralite, line 2461: I wouldn't document this as `null`. Simply fetching a webhook in a channel will return the token.

packages/discord.js/typings/index.d.ts
```diff
@@ -2453,10 +2453,27 @@ export class Webhook extends WebhookMixin() {
...
+  public isIncoming(): this is this & {
+    type: WebhookType.Incoming;
+    token: string;
+  };
   public isChannelFollower(): this is this & {
+    type: WebhookType.ChannelFollower;
     sourceGuild: Guild | APIPartialGuild;
     sourceChannel: NewsChannel | APIPartialChannel;
+    token: null;
   };
```
Jiralite, line 2477: The owner will be defined here! Also, `applicationId` will be `null` here, even if added by a bot.

packages/discord.js/src/structures/Webhook.js
```diff
@@ -393,6 +403,22 @@ class Webhook {
...
+
+  /**
+   * Whether this webhook is created by an application and not a user.
```
Jiralite, line 415: I would remove the "and not a user" portion of this string, it's not needed information.
```suggestion
   * Whether this webhook is created by an application.
```

packages/discord.js/src/structures/Webhook.js
```diff
@@ -393,6 +403,22 @@ class Webhook {
...
+   * @returns {boolean}
+   */
+  isApplicationCreated() {
+    return Boolean(this.applicationId);
```
Jiralite, line 419: A more accurate check is to check the type, right?

Jiralite (changes requested):

packages/discord.js/typings/index.d.ts
advaith1: no, bot-created webhooks (as well as oauth2-created webhooks) have an application id

packages/discord.js/src/structures/Webhook.js
oof2win2: Not sure why I added this. Is it fine if I remove it? If it's incoming, the user/developer can check themselves, no?

packages/discord.js/typings/index.d.ts
oof2win2: So `string | null` I assume?

packages/discord.js/typings/index.d.ts
oof2win2: The example on the docs shows that application webhooks can have an application id
https://discord.com/developers/docs/resources/webhook#webhook-object-example-application-webhook

oof2win2: Seems that Jiralite was correct. `application_id` is null on following webhooks

packages/discord.js/typings/index.d.ts
oof2win2: Hmm. The issue here is that if it's fetched with the owner, the token is not sent back, and I assume it's the same the other way round.

packages/discord.js/src/structures/Webhook.js
```diff
@@ -393,6 +403,25 @@ class Webhook {
...
 
+  /**
+   * Whether this webhook is created by a user.
+   * @returns {boolean}
+   */
+  isUserCreated() {
+    if (this.type !== WebhookType.Incoming) return false;
+    if (!this.owner) return false;
+    if (this.owner.bot) return false;
+    return true;
```
almeidx, line 414:
```suggestion
    return Boolean(this.type === WebhookType.Incoming && this.owner && !this.owner.bot)
```

packages/discord.js/typings/index.d.ts
```diff
@@ -2453,10 +2453,29 @@ export class Webhook extends WebhookMixin() {
...
   public token: string | null;
   public type: WebhookType;
-  public isIncoming(): this is this & { token: string };
+  public applicationId: Snowflake | null;
+  public isUserCreated(): this is this & {
+    type: WebhookType.Incoming;
+    applicationId: null;
+    owner: User | APIUser;
+    token: string | null;
+  };
```
almeidx, line 2462: `token` is already typed as `string | null`
```suggestion
  public isUserCreated(): this is this & {
    type: WebhookType.Incoming;
    applicationId: null;
    owner: User | APIUser;
  };
```

almeidx (changes requested):

SpaceEEC (approved):

vladfrangu (approved):

packages/discord.js/src/structures/Webhook.js
```diff
@@ -393,6 +403,22 @@ class Webhook {
...
+    return Boolean(this.type === WebhookType.Incoming && this.owner && !this.owner.bot);
+  }
+
+  /**
+   * Whether this webhook is created by an application.
+   * @returns {boolean}
+   */
+  isApplicationCreated() {
+    return this.type === WebhookType.Application;
+  }
```
Jiralite, line 420: The type it's checking doesn't match with the types.

Jiralite (changes requested):

packages/discord.js/src/structures/Webhook.js
oof2win2: Why not?

Jiralite: You fixed this in https://github.com/discordjs/discord.js/pull/7317/commits/79fed2fe1304175c2ac922dd50c7f44a24588429.

oof2win2: Oh okay, forgot about making that commit.

kyranet (approved):

iCrawl (approved):

PR #7298, opened 2022-01-18
refactor: use `@discordjs/rest`

vaporox:
**Please describe the changes this PR makes and why it should be merged:**

This PR replaces the current REST implementation in `discord.js` with the `@discordjs/rest` package.

<details>

<summary>Breaking Changes</summary>

- `DiscordAPIError`, `HTTPError` and `RateLimitError` are now re-exported and thus have a different interface.
- `MessagePayload#data` was renamed to `MessagePayload#body`.
- The type of `MessagePayload#files` was changed to `?RawFile[]`.
- `MessagePayload#resolveData` was renamed to `MessagePayload#resolveBody`.
- The return type of `MessagePayload#resolveFiles` was changed to `Promise<RawFile>`.
- `Constants.Endpoints` was removed.
- `RATE_LIMIT`, `INVALID_REQUEST_WARNING`, `API_RESPONSE` and `API_REQUEST` were removed from `Constants.Events`.
- `DataResolver#resolveFileAsBuffer` was removed and implemented into `DataResolver#resolveFile` instead.
- `invalidRequestWarningInterval`, `restTimeOffset`, `restRequestTimeout`, `restSweepInterval`, `restGlobalRateLimit`, `rejectOnRateLimit`, `retryLimit`, `userAgentSuffix` and `http` were removed from `ClientOptions` and replaced by `rest`.

</details>

**Status and versioning classification:**

- Code changes have been tested against the Discord API
- I know how to update typings and have done so
- This PR includes breaking changes (methods removed or renamed, parameters moved or removed)

PR #7318, opened 2022-01-22
rest: prefer arrayBuffer over buffer

KhafraDev:

- `Response.buffer` is not a spec compliant method.
- Buffer only exists in node and it's just a wrapper for Uin8Arrays anyways.
- It's getting [removed](https://github.com/node-fetch/node-fetch/issues/1452) in v4, if its decided to stay with node-fetch and move towards ESM.

I have no idea if types need updating, sorry 😕 

### DIFF
--- a/packages/rest/__tests__/REST.test.ts
+++ b/packages/rest/__tests__/REST.test.ts
@@ -205,12 +205,12 @@ test('urlEncoded', async () => {
 		['code', 'very-invalid-code'],
 	]);
 	expect(
-		await api.post('/urlEncoded', {
+		new Uint8Array(await api.post('/urlEncoded', {
 			body,
 			passThroughBody: true,
 			auth: false,
-		}),
-	).toStrictEqual(Buffer.from(body.toString()));
+		}) as ArrayBuffer),
+	).toStrictEqual(new Uint8Array(Buffer.from(body.toString())));
 });
 
 test('postEcho', async () => {
--- a/packages/rest/__tests__/REST.test.ts
+++ b/packages/rest/__tests__/REST.test.ts
@@ -205,11 +205,13 @@ test('urlEncoded', async () => {
 		['code', 'very-invalid-code'],
 	]);
 	expect(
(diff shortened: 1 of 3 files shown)


KhafraDev:
Also pinging @ckohen since he is pretty involved in rest

ckohen:
At first glance, I'm inclined to not make this switch until we utilize undici because of the way node-fetch handles consuming the body. Internally, node-fetch seems to much prefer using buffer() since that is what their consume function converts to natively (Their consume body function probably is why buffer() exists at all). While small, there is a slight performance hit from converting that buffer, not sure if that's going to end up mattering much though.

I guess its a matter of if we want to do part of the work now rather than later (also makes undici a semver:minor / patch PR I think?) which is up to the maintainers.

To note: should be obvious, but this will need changes in the main lib in #7298

KhafraDev:
I doubt that this change will have any meaningful performance impact, as node-fetch's arrayBuffer method is simply just using `buffer.slice` after calling `consumeBody`.

The only difference is that users will have to deal with ArrayBuffers instead of Buffers on api routes that don't have an `application/json` Content-Type header. Even then, converting from ArrayBuffer to Buffers (Uint8Arrays) will not be slowing down user applications 😄.

ckohen:
Yeah, slight could definitely be considered an exaggeration for the performance implications lol, especially considering the slim amount of endpoints that don't return JSON to begin with

jimmywarting:
converting it to a node buffer have not so much cost to it i think,
quite simple to do:
```js
const buffer = await res.arrayBuffer().then(Buffer.from)
```
But ultimately i think buffer should be discouraged for better cross platform functionality
https://github.com/cross-js/cross-js#dont-use-buffer

vladfrangu (approved):

iCrawl:
Can you rebase this PR? @KhafraDev

iCrawl (approved):

kyranet (approved):

KhafraDev:
@iCrawl rebased 👍

PR #7320, opened 2022-01-22
feat!: add CategoryChannelChildManager

ImRodry:

This PR changes CategoryChannel#children to be a manager and moves CategoryChannel#createChannel to CategoryChannelChildManager#create. This allows for better organization and consistency with other methods in the library and also allows for easier expandability in the future

### DIFF
--- a/packages/discord.js/src/managers/CategoryChannelChildManager.js
+++ b/packages/discord.js/src/managers/CategoryChannelChildManager.js
@@ -1,13 +1,13 @@
 'use strict';
 
-const CachedManager = require('./CachedManager');
+const DataManager = require('./DataManager');
 const GuildChannel = require('../structures/GuildChannel');
 
 /**
  * Manages API methods for CategoryChannels' children and stores their cache.
- * @extends {CachedManager}
+ * @extends {DataManager}
  */
-class CategoryChannelChildManager extends CachedManager {
+class CategoryChannelChildManager extends DataManager {
   constructor(channel, iterable) {
     super(channel.client, GuildChannel, iterable);
     /**
@@ -15,12 +15,24 @@ class CategoryChannelChildManager extends CachedManager {
      * @type {CategoryChannel}
      */
     this.channel = channel;
+  }
 
-    /**
-     * The guild this manager belongs to
-     * @type {Guild}
-     */
-    this.guild = channel.guild;
+  /**
+   * The channels that are a part of this category
+   * @type {Collection<Snowflake, GuildChannel>}
+   * @readonly
+   */
+  get cache() {
+    return this.guild.channels.cache.filter(c => c.parentId === this.channel.id);
+  }
+
+  /**
+   * The guild this manager belongs to
+   * @type {Guild}
+   * @readonly
+   */
+  get guild() {
+    return this.channel.guild;
   }
 
   /**
@@ -51,15 +63,6 @@ class CategoryChannelChildManager extends CachedManager {
       ...options,
       parent: this.channel.id,
     });
-  }
-
-  /**
-   * The channels that are a part of this category
-   * @type {Collection<Snowflake, GuildChannel>}
-   * @readonly
-   */
-  get cache() {
-    return this.guild.channels.cache.filter(c => c.parentId === this.channel.id);
   }
 }
 
--- a/packages/discord.js/src/managers/CategoryChannelChildManager.js
+++ b/packages/discord.js/src/managers/CategoryChannelChildManager.js
@@ -1,0 +1,66 @@
+'use strict';
+
+const CachedManager = require('./CachedManager');
+const GuildChannel = require('../structures/GuildChannel');
+
+/**
+ * Manages API methods for CategoryChannels' children and stores their cache.
+ * @extends {CachedManager}
+ */
+class CategoryChannelChildManager extends CachedManager {
+  constructor(channel, iterable) {
+    super(channel.client, GuildChannel, iterable);
+    /**
+     * The category channel this manager belongs to
+     * @type {CategoryChannel}
+     */
+    this.channel = channel;
+
+    /**
+     * The guild this manager belongs to
+     * @type {Guild}
+     */
+    this.guild = channel.guild;
+  }
+
+  /**
+   * Options for creating a channel using {@link CategoryChannel#createChannel}.
+   * @typedef {Object} CategoryCreateChannelOptions
+   * @property {ChannelType} [type=ChannelType.GuildText] The type of the new channel.
+   * @property {string} [topic] The topic for the new channel
+   * @property {boolean} [nsfw] Whether the new channel is NSFW
+   * @property {number} [bitrate] Bitrate of the new channel in bits (only voice)
+   * @property {number} [userLimit] Maximum amount of users allowed in the new channel (only voice)
+   * @property {OverwriteResolvable[]|Collection<Snowflake, OverwriteResolvable>} [permissionOverwrites]
+   * Permission overwrites of the new channel
+   * @property {number} [position] Position of the new channel
+   * @property {number} [rateLimitPerUser] The rate limit per user (slowmode) for the new channel in seconds
+   * @property {string} [rtcRegion] The specific region of the new channel.
+   * @property {string} [reason] Reason for creating the new channel
+   */
+
+  /**
+   * Creates a new channel within this category.
+   * <info>You cannot create a channel of type {@link ChannelType.GuildCategory} inside a CategoryChannel.</info>
+   * @param {string} name The name of the new channel
+   * @param {CategoryCreateChannelOptions} options Options for creating the new channel
+   * @returns {Promise<GuildChannel>}
+   */
+  create(name, options) {
+    return this.guild.channels.create(name, {
+      ...options,
+      parent: this.channel.id,
+    });
+  }
+
+  /**
+   * The channels that are a part of this category
+   * @type {Collection<Snowflake, GuildChannel>}
+   * @readonly
+   */
+  get cache() {
+    return this.guild.channels.cache.filter(c => c.parentId === this.channel.id);
+  }
+}
+
+module.exports = CategoryChannelChildManager;
--- a/packages/discord.js/src/managers/CategoryChannelChildManager.js
+++ b/packages/discord.js/src/managers/CategoryChannelChildManager.js
@@ -8,8 +8,8 @@ const GuildChannel = require('../structures/GuildChannel');
  * @extends {DataManager}
  */
 class CategoryChannelChildManager extends DataManager {
-  constructor(channel, iterable) {
-    super(channel.client, GuildChannel, iterable);
+  constructor(channel) {
+    super(channel.client, GuildChannel);
     /**
      * The category channel this manager belongs to
      * @type {CategoryChannel}
--- a/packages/discord.js/src/managers/CategoryChannelChildManager.js
+++ b/packages/discord.js/src/managers/CategoryChannelChildManager.js
@@ -4,7 +4,7 @@ const DataManager = require('./DataManager');
 const GuildChannel = require('../structures/GuildChannel');
 
 /**
- * Manages API methods for CategoryChannels' children and stores their cache.
+ * Manages API methods for CategoryChannels' children.
  * @extends {DataManager}
  */
 class CategoryChannelChildManager extends DataManager {
--- a/packages/discord.js/src/structures/CategoryChannel.js
+++ b/packages/discord.js/src/structures/CategoryChannel.js
@@ -1,6 +1,7 @@
 'use strict';
 
 const GuildChannel = require('./GuildChannel');
+const CategoryChannelChildManager = require('../managers/CategoryChannelChildManager');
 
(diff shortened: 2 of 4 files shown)


Syjalo (changes requested):
I think the cache should be updated on `channelCreate`, `channelDelete` and `channelUpdate`.

ImRodry:
That's why it's a getter, it just gets from the guild channels cache every time it's called

packages/discord.js/src/managers/CategoryChannelChildManager.js
```diff
@@ -0,0 +1,62 @@
+'use strict';
+
+const CachedManager = require('./CachedManager');
+const GuildChannel = require('../structures/GuildChannel');
+
```
Syjalo, line 5:
```suggestion

/**
 * Manages API methods for GuildChannels and stores their cache.
 * @extends {CachedManager}
 */
```

packages/discord.js/src/structures/CategoryChannel.js
```diff
@@ -1,19 +1,22 @@
...
-  get children() {
-    return this.guild.channels.cache.filter(c => c.parentId === this.id);
+  constructor(guild, data, client) {
+    super(guild, data, client);
+
+    /**
+     * Channels that are a part of this category
+     * @type {Collection<Snowflake, GuildChannel>}
+     * @readonly
+     */
```
Syjalo, line 18:
```suggestion
    /**
     * A manager of the channels belonging to this category channel
     * @type {CategoryChannelChildManager}
     */
```

Syjalo (changes requested):

packages/discord.js/src/managers/CategoryChannelChildManager.js
```diff
@@ -0,0 +1,66 @@
...
+   * @param {CategoryCreateChannelOptions} options Options for creating the new channel
+   * @returns {Promise<GuildChannel>}
+   */
+  create(name, options) {
+    return this.guild.channels.create(name, {
+      ...options,
+      parent: this.channel.id,
+    });
+  }
+
```
Syjalo, line 55:
```suggestion

  /**
     * Batch-updates the category's channels' positions.
     * @param {ChannelPosition[]} channelPositions Channel positions to update
     * @returns {Promise<Guild>}
     * @example
     * channel.children.setPositions([{ channel: channelId, position: newChannelIndex }])
     *   .then(guild => console.log(`Updated channel positions for ${guild}`))
     *   .catch(console.error);
     */
  async setPositions(channelPositions) {
    channelPositions = channelPositions.map(r => ({
      id: this.client.channels.resolveId(r.channel),
      position: r.position,
      lock_permissions: r.lockPermissions,
      parent_id: typeof r.parent === 'undefined' ? this.channel.id : this.guild.channels.resolveId(r.parent),
    }));

    return this.guild.channels.setPositions(channelPositions);
  }

```

Syjalo (changes requested):

packages/discord.js/src/managers/CategoryChannelChildManager.js
ImRodry: What would be the benefit of having this here?

Syjalo: > this.channel.id

This will set parent of a channel to the category channel when you will set positions.

ImRodry: I mean I can see the use case but shouldn't this only work for the channels within the category? There's no way to prevent that from happening

Syjalo: > I mean I can see the use case but shouldn't this only work for the channels within the category? There's no way to prevent that from happening

Do you mean it shouldn't move a channel from other categories?

ImRodry: Yes. I'll wait for @discordjs/the-big-4 and commit if they think it's worth it

packages/discord.js/src/structures/CategoryChannel.js
```diff
@@ -1,19 +1,21 @@
...
-  get children() {
-    return this.guild.channels.cache.filter(c => c.parentId === this.id);
+  constructor(guild, data, client) {
+    super(guild, data, client);
+
+    /**
+     * A manager of the channels belonging to this category
+     * @type {CategoryChannelChildManager}
+     */
+    this.children = new CategoryChannelChildManager(this);
```
kyranet, line 18: This could be a getter, no need for the extra memory overhead.

kyranet (changes requested):

packages/discord.js/src/structures/CategoryChannel.js
ImRodry: A getter for a manager? Do we do that anywhere else?

Jiralite: https://github.com/discordjs/discord.js/blob/11e5e5ac5b70138f56332eb3e61a42443670b0de/packages/discord.js/src/client/Client.js#L172-L183

ImRodry: Not sure what difference this makes since the cache for this manager is a getter but sure, I'll change that

ImRodry:
I believe the CI fail is unrelated to this PR as it is thrown on /rest but if it isn't, please let me know and I'll fix it

suneettipirneni:
> I believe the CI fail is unrelated to this PR as it is thrown on /rest but if it isn't, please let me know and I'll fix it

run this to restart the CI
```
git commit --amend --no-edit
git push -f
```

ImRodry:
Thanks! That fixed it and I'll keep that in mind for the future

packages/discord.js/typings/index.d.ts
```diff
@@ -2718,6 +2707,23 @@ export class BaseGuildEmojiManager extends CachedManager<Snowflake, GuildEmoji,
   public resolveIdentifier(emoji: EmojiIdentifierResolvable): string | null;
 }
 
+export class CategoryChannelChildManager extends CachedManager<Snowflake, GuildChannel, GuildChannelResolvable> {
```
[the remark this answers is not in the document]
kyranet: As a matter of a fact, abstract classes (including `Channel` and `GuildChannel`) should never be used. Instead, we must use any of those:

https://github.com/discordjs/discord.js/blob/11e5e5ac5b70138f56332eb3e61a42443670b0de/packages/discord.js/typings/index.d.ts#L5018-L5039

Perhaps make a new type for non-category and non-thread guild channels?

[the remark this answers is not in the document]
ImRodry: Do I need to change the GuildBasedChannel type too or is that fine to keep?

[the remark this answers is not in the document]
kyranet: Fine to keep, please do not change that.

packages/discord.js/src/managers/CategoryChannelChildManager.js
```diff
@@ -0,0 +1,66 @@
+'use strict';
+
+const CachedManager = require('./CachedManager');
+const GuildChannel = require('../structures/GuildChannel');
+
+/**
+ * Manages API methods for CategoryChannels' children and stores their cache.
```
vitorlops, line 7:
```suggestion
 * Manages API methods for CategoryChannel's children and stores their cache.
```

ImRodry: That’s not correct english

monbrey: Agree, its either `the CategoryChannel's` singular, or `CategoryChannels'` plural

vitorlops: Got it, thanks explaining.

packages/discord.js/src/structures/CategoryChannel.js
```diff
@@ -1,19 +1,20 @@
...
-   * Channels that are a part of this category
-   * @type {Collection<Snowflake, GuildChannel>}
+   * A manager of the channels belonging to this category
+   * @type {CategoryChannelChildManager}
    * @readonly
    */
   get children() {
-    return this.guild.channels.cache.filter(c => c.parentId === this.id);
+    return new CategoryChannelChildManager(this);
   }
```
vladfrangu, line 18: This doesn't need to be a getter anymore

vladfrangu (changes requested):

packages/discord.js/src/structures/CategoryChannel.js
ImRodry: I was told to make it a getter by @kyranet though ?

kyranet: The problem with making it a property is that it makes the memory usage higher, @vladfrangu. This is a perfectly computed class, it can be a getter just fine.

vladfrangu: How does it make memory usage higher? Does it even matter? Creating an instance of this class every time you access it doesn't seem like a smart idea either, and it doesn't store any data itself 🤷

kyranet: Well, this class stores `guild` and `channel`.

Then inherits `CachedManager`, which creates a Collection (`_cache`) no matter what, with `Object.defineProperty`:

https://github.com/discordjs/discord.js/blob/cd2f5660521ae4a668727cd246c6e5b0abe53219/packages/discord.js/src/managers/CachedManager.js#L14

Then inherits `DataManager` (which this class should extend anyways), and this stores `holds` using `Object.defineProperty`

https://github.com/discordjs/discord.js/blob/cd2f5660521ae4a668727cd246c6e5b0abe53219/packages/discord.js/src/managers/DataManager.js#L22

And it finally extends `BaseManager`, which creates a property of `client`, also with `Object.defineProperty`:

https://github.com/discordjs/discord.js/blob/cd2f5660521ae4a668727cd246c6e5b0abe53219/packages/discord.js/src/managers/BaseManager.js#L15

Now that's a total of 5 properties, one of them being a brand new `Map` (which are very heavy), and 3 of those properties are done so with `Object.defineProperty` (and you know the performance implications of this).

Making it a property will force the code to always run, resulting on a large performance and memory impact as categories are on average the 17% of all channels (at least on my bot, out of 384,183 channels, 65,587 are category channels).

Furthermore, `CategoryChannel#children` isn't that used, nor it's that useful, it's a utility that provides nothing but convenience. *As a matter of a fact, I haven't even used this property in my [bot's code](https://github.com/skyra-project/skyra/search?l=TypeScript&q=.children), which proves that even kitchensinks may not use it.* I have yet to see real world usage that uses this property efficiently.

If we are to lower the CPU and RAM usage of discord.js, we should start by not forcing the storage of data that not everyone needs.

vladfrangu (approved):

packages/discord.js/src/managers/CategoryChannelChildManager.js
```diff
@@ -0,0 +1,66 @@
...
+ * @extends {CachedManager}
+ */
+class CategoryChannelChildManager extends CachedManager {
```
kyranet, line 10: Shouldn't this extend `DataManager`? All `CachedManager` brings is a private readonly `_cache` property (being a map) and the private `_add` method, none of which is used here.

Using `DataManager` would allow us to skip allocating a Collection every time an `CategoryChannelChildManager` is created.

packages/discord.js/src/managers/CategoryChannelChildManager.js
```diff
@@ -0,0 +1,66 @@
...
+     * The category channel this manager belongs to
+     * @type {CategoryChannel}
+     */
+    this.channel = channel;
+
+    /**
+     * The guild this manager belongs to
+     * @type {Guild}
+     */
+    this.guild = channel.guild;
```
kyranet, line 23: Is it necessary to always store this? We can make it a getter.

kyranet (changes requested):

packages/discord.js/src/managers/CategoryChannelChildManager.js
ImRodry: I just extended CachedManager because this has a cache, although it isn't used directly so I guess I could change that yeah

packages/discord.js/src/managers/CategoryChannelChildManager.js
ImRodry: Yeah it's unnecessary, I don't know if the getter is even needed, what do you think?

kyranet: I think it's more a convenience, just like `message.guild` before the `guildId` days (hint: it was just `this.channel.guild`).

kyranet (changes requested):
Just one last thing before approving.

kyranet (approved):

Mysterious-Dev (approved):

iCrawl (approved):

SpaceEEC (approved):

PR #7324, opened 2022-01-22
fix: update node-fetch to 2.6.7

YodaLightsabr:
Fix a major security vulnerability with node-fetch versions below 2.6.7 and 3.1.1.

### DIFF
--- a/packages/discord.js/package.json
+++ b/packages/discord.js/package.json
@@ -55,7 +55,7 @@
     "@types/ws": "^8.2.2",
     "discord-api-types": "^0.26.1",
     "form-data": "^4.0.0",
-    "node-fetch": "^2.6.1",
+    "node-fetch": "^2.6.7",
     "ws": "^8.4.0"
   },
   "devDependencies": {
--- a/yarn.lock
+++ b/yarn.lock
@@ -5853,15 +5853,10 @@ nock@^13.2.1:
     lodash.set "^4.3.2"
     propagate "^2.0.0"
 
-node-fetch@2.6.1:
-  version "2.6.1"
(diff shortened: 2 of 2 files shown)


almeidx:
Why was the package file not updated too?

wildercode:
oh yes, that is very important

KhafraDev:
This doesn't even get rid of the warning because npm audit sucks and reports the vulnerability in "node-fetch  <3.1.1"

YodaLightsabr:
That's an NPM issue, and regardless since there aren't any breaking changes node-fetch should still be updated to 2.6.7.

yourfriendoss:
Need this out ASAP

kyranet:
This change is unneeded, the vulnerability is related to the cookie header leaking, we do not use those in discord.js. Besides, we're using `^2.6.1`, as you can see below:

https://github.com/discordjs/discord.js/blob/988a51b7641f8b33cc9387664605ddc02134859d/package.json#L60

Package managers resolve to the latest minor and patch within the v2 range, so they'll resolve that version as `2.6.7`, as it is the highest compatible version within semver ([see here](https://jubianchi.github.io/semver-check/#/%5E2.6.1/2.6.7)).

And even if we bump it, `npm audit` will still complain because it uses `<3.1.1`, which includes *any* version, including `2.6.7` ([see here](https://jubianchi.github.io/semver-check/#/%3C3.1.1/2.6.1)).

Save yourself the trouble and disable `npm audit`, this is one of the false positives it reports, and there are many more included in https://overreacted.io/npm-audit-broken-by-design/.

YodaLightsabr:
Ok, cool.

PR #7343, opened 2022-01-25
fix: don't create new instances of builders classes

ImRodry:

This PR makes it so that if a builders class is passed onto createComponent it is returned as-is, without creating a new instance of it so we can keep referencing it. Also changed types of ActionRow slightly to make it easier to type it

### DIFF
--- a/packages/builders/src/components/ActionRow.ts
+++ b/packages/builders/src/components/ActionRow.ts
@@ -3,6 +3,8 @@ import type { ButtonComponent, SelectMenuComponent } from '..';
 import type { Component } from './Component';
 import { createComponent } from './Components';
 
+export type MessageComponent = ActionRowComponent | ActionRow;
+
 export type ActionRowComponent = ButtonComponent | SelectMenuComponent;
 
 // TODO: Add valid form component types
@@ -10,7 +12,7 @@ export type ActionRowComponent = ButtonComponent | SelectMenuComponent;
 /**
  * Represents an action row component
  */
-export class ActionRow<T extends ActionRowComponent> implements Component {
+export class ActionRow<T extends ActionRowComponent = ActionRowComponent> implements Component {
 	public readonly components: T[] = [];
 	public readonly type = ComponentType.ActionRow;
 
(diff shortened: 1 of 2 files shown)


packages/builders/src/components/ActionRow.ts
```diff
@@ -3,14 +3,16 @@ import type { ButtonComponent, SelectMenuComponent } from '..';
...
+
 export type ActionRowComponent = ButtonComponent | SelectMenuComponent;
 
 // TODO: Add valid form component types
 
 /**
  * Represents an action row component
  */
-export class ActionRow<T extends ActionRowComponent> implements Component {
+export class ActionRow<T extends ActionRowComponent = ActionRowComponent> implements Component {
```
suneettipirneni, line 15: Out of scope?

ImRodry: Eh maybe? I explained in the description it was just a small thing to make it easier to type but I can revert if you want

packages/builders/src/components/Components.ts
```diff
@@ -15,14 +15,15 @@ export interface MappedComponentTypes {
 export function createComponent<T extends keyof MappedComponentTypes>(
 	data: APIMessageComponent & { type: T },
 ): MappedComponentTypes[T];
-export function createComponent(data: APIMessageComponent): Component {
+export function createComponent<C extends MessageComponent>(data: C): C;
+export function createComponent(data: APIMessageComponent | MessageComponent): Component {
```
vladfrangu, line 19: Shouldn't you just.. type it as `data: APIMessageComponent | Component`?

vladfrangu (dismissed):

packages/builders/src/components/Components.ts
ImRodry: can't
image

iCrawl (approved):

vladfrangu (approved):

SpaceEEC (approved):

kyranet (approved):

PR #7350, opened 2022-01-26
chore: make `@types/node-fetch` a regular dependency

suneettipirneni:

We use `node-fetch` types within the public rest API definitions so it should be a non-dev dependency.

**Status and versioning classification:**:


vladfrangu (approved):

kyranet (approved):

iCrawl (approved):

ImRodry:
@discordjs/rest does use node-fetch types but none of these are exported by discord.js nor used as types in the typings file, are they?

vladfrangu:
Consider generated definition files, which cause issues in your project if you try to compile a TS project without them 🙂 
It's also why we have @types/ws in discord.js

ImRodry:
> Consider generated definition files, which cause issues in your project if you try to compile a TS project without them 🙂 It's also why we have @types/ws in discord.js

Well `@types/ws` is imported in the typings file whereas `node-fetch` isn't. Furthermore none of the imported or exported types rely on types coming from `node-fetch`, which is why I was asking
https://github.com/discordjs/discord.js/blob/355f579771771a28a293c327a38574c8918d18f8/packages/discord.js/typings/index.d.ts#L87

almeidx:
The change was done in /rest and not discord.js

SpaceEEC (approved):

oceanroleplay:
@iCrawl A patch release is needed for this, as builds fail for the development version.
image

however, confirm this at your end.

ImRodry:
/rest was added in v14 so all you need is the latest dev release

oceanroleplay:
> /rest was added in v14 so all you need is the latest dev release

Thank you but I am very well aware about that, I am using `^14.0.0-dev.1643674011.803665`

PR #7352, opened 2022-01-26
feat(Interaction): add userId

vitorlops:
Get the user ID will be easier, as it already have the properties `channel` and `channelId`, I think is good adding the `userId`.

### DIFF
--- a/packages/discord.js/src/structures/Interaction.js
+++ b/packages/discord.js/src/structures/Interaction.js
@@ -58,6 +58,12 @@ class Interaction extends Base {
     this.user = this.client.users._add(data.user ?? data.member.user);
 
     /**
+     * The id of the user which sent this interaction
+     * @type {?Snowflake}
+     */
+    this.userId = data.user?.id ?? null;
+
+    /**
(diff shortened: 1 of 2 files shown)


vladfrangu:
What's the point of this, you're saving exactly one character (a `.`) with this PR (without even handling the guild member user) 👀

almeidx:
The reasoning behind `channelId` is that discord doesn't send the channel in the interaction payload, only the channel_id. 
https://discord.com/developers/docs/interactions/receiving-and-responding#interaction-object-interaction-structure

That isn't the case for `user`

vitorlops:
> What's the point of this, you're saving exactly one character (a `.`) with this PR (without even handling the guild member user) 👀

Oh, I did something wrong then. 

> The reasoning behind `channelId` is that discord doesn't send the channel object in the interaction payload, only the channel_id. https://discord.com/developers/docs/interactions/receiving-and-responding#interaction-object-interaction-structure
> 
> That isn't the case for `user`

Oh, I didn't know that (thanks for the explanation), do you think it's unecessary?

PR #7354, opened 2022-01-27
chore: use `performance.now` for time-based unit tests

suneettipirneni:

This could fix the flaky unit test issues for /rest.

> Also unlike Date.now(), the values returned by performance.now() always increase at a constant rate, independent of the system clock (which might be adjusted manually or skewed by software like NTP).

[Source](https://developer.mozilla.org/en-US/docs/Web/API/Performance/now)


iCrawl (approved):

kyranet (approved):

SpaceEEC (approved):
🙏

PR #7358, opened 2022-01-28
fix(MessageEmbed): set footer to undefined

imranbarbhuiya:
If you set `embed.setFooter(null)` it set's footer to `{}`. When checking `MessageEmbed`.length, it checks `this.footer?.text.length`. In this case, footer is defined but footer.text isn't, so it'll throw an error `can't read property of undefined (reading length)`. 

In builder, it's correct, that's why after discussing in `library-discussion` I'm sending the pr in stable branch.

### DIFF
--- a/src/structures/MessageEmbed.js
+++ b/src/structures/MessageEmbed.js
@@ -430,7 +430,7 @@ class MessageEmbed {
    */
   setFooter(options, deprecatedIconURL) {
     if (options === null) {
-      this.footer = {};
+      this.footer = undefined;
       return this;
     }
 


kyranet:
We do not accept PRs to stable anymore, it causes main to need a rebase and with that, all PRs with it.

imranbarbhuiya:
> We do not accept PRs to stable anymore, it causes main to need a rebase and with that, all PRs with it.

This bug isn't present is builder, that's why it's not present in main branch too. 

So where should I change it?  I think this change is needed because after v14 release some people will still use v13.

imranbarbhuiya:
After #7431 ig this can be added in v13

iCrawl (approved):

PR #7364, opened 2022-01-28
feat(richpresence): add support for url image assets

suneettipirneni:

No upstream PR yet

### DIFF
--- a/packages/discord.js/src/structures/Presence.js
+++ b/packages/discord.js/src/structures/Presence.js
@@ -333,13 +333,13 @@ class RichPresenceAssets {
 
     /**
      * The large image asset's id
-     * @type {?Snowflake|string}
+     * @type {?(Snowflake|string)}
      */
     this.largeImage = assets.large_image ?? null;
 
     /**
      * The small image asset's id
-     * @type {?Snowflake|string}
+     * @type {?(Snowflake|string)}
      */
     this.smallImage = assets.small_image ?? null;
   }
@@ -399,10 +399,15 @@ class RichPresenceAssets {
    * @private
    */
   static _isURLImage(imageKey) {
-    const URLRegex = /(https?|ftp|file):\/\//gm;
-    return URLRegex.test(imageKey);
+    return RichPresenceAssets.URL_PATTERN.test(imageKey);
   }
 }
+
+/**
+ * Regular expression that matches images URLs from image keys
+ * @type {RegExp}
+ */
+RichPresenceAssets.URL_PATTERN = /(https?|ftp|file):\/\//gm;
 
 exports.Presence = Presence;
 exports.Activity = Activity;
--- a/packages/discord.js/src/structures/Presence.js
+++ b/packages/discord.js/src/structures/Presence.js
@@ -333,13 +333,13 @@ class RichPresenceAssets {
 
     /**
      * The large image asset's id
-     * @type {?Snowflake}
+     * @type {?string}
      */
     this.largeImage = assets.large_image ?? null;
 
     /**
      * The small image asset's id
-     * @type {?Snowflake}
+     * @type {?string}
      */
     this.smallImage = assets.small_image ?? null;
   }
@@ -351,6 +351,7 @@ class RichPresenceAssets {
    */
   smallImageURL(options = {}) {
     if (!this.smallImage) return null;
+    if (RichPresenceAssets._isURLImage(this.smallImage)) return this.smallImage;
     if (this.smallImage.includes(':')) {
       const [platform, id] = this.smallImage.split(':');
       switch (platform) {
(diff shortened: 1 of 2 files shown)


packages/discord.js/src/structures/Presence.js
```diff
@@ -333,13 +333,13 @@ class RichPresenceAssets {
 
     /**
      * The large image asset's id
-     * @type {?Snowflake}
+     * @type {?string}
```
ImRodry, line 336:
```suggestion
     * @type {?(Snowflake|string)}
```

packages/discord.js/src/structures/Presence.js
```diff
@@ -333,13 +333,13 @@ class RichPresenceAssets {
...
      */
     this.largeImage = assets.large_image ?? null;
 
     /**
      * The small image asset's id
-     * @type {?Snowflake}
+     * @type {?string}
```
ImRodry, line 342:
```suggestion
     * @type {?(Snowflake|string)}
```

packages/discord.js/src/structures/Presence.js
```diff
@@ -389,6 +391,17 @@ class RichPresenceAssets {
 
     return this.activity.presence.client.rest.cdn.appAsset(this.activity.applicationId, this.largeImage, options);
   }
+
+  /**
+   * Tests whether an image key is a URL or not
+   * @private
+   * @param {string} imageKey The image key to test
+   * @returns {boolean}
```
ImRodry, line 399: Usually `@private` is the last tag, also looks nicer
```suggestion
   * @param {string} imageKey The image key to test
   * @returns {boolean}
   * @private
```

packages/discord.js/src/structures/Presence.js
```diff
@@ -389,6 +391,17 @@ class RichPresenceAssets {
...
+   */
+  static _isURLImage(imageKey) {
+    const URLRegex = /(https?|ftp|file):\/\//gm;
```
ImRodry, line 402: maybe export this regex?

packages/discord.js/typings/index.d.ts
```diff
@@ -1776,12 +1776,13 @@ export class ReactionEmoji extends Emoji {
 
 export class RichPresenceAssets {
   private constructor(activity: Activity, assets: RawRichPresenceAssets);
-  public largeImage: Snowflake | null;
+  public largeImage: string | null;
   public largeText: string | null;
-  public smallImage: Snowflake | null;
+  public smallImage: string | null;
```
ImRodry, line 1781: kinda redundant, I know, but this is done elsewhere and if the Snowflake type is ever changed it can be changed here too
```suggestion
  public smallImage: Snowflake | string | null;
```

packages/discord.js/typings/index.d.ts
```diff
@@ -1776,12 +1776,13 @@ export class ReactionEmoji extends Emoji {
 
 export class RichPresenceAssets {
   private constructor(activity: Activity, assets: RawRichPresenceAssets);
-  public largeImage: Snowflake | null;
+  public largeImage: string | null;
```
ImRodry, line 1779:
```suggestion
  public largeImage: Snowflake | string | null;
```

ImRodry (changes requested):

packages/discord.js/typings/index.d.ts
suneettipirneni: I mean Snowflake | string resolves to just `string`

ImRodry: Yeah I know but

> this is done elsewhere and if the Snowflake type is ever changed it can be changed here too

packages/discord.js/src/structures/Presence.js
suneettipirneni: I don't see the use

ImRodry: Other patterns are exported in other classes and the function isn't public so I think the regex should be so users can use this if they need to

suneettipirneni:
This seems to be deployed and live on the API

packages/discord.js/src/structures/Presence.js
```diff
@@ -389,8 +391,24 @@ class RichPresenceAssets {
...
+  static _isURLImage(imageKey) {
+    return RichPresenceAssets.URL_PATTERN.test(imageKey);
+  }
 }
 
+/**
+ * Regular expression that matches images URLs from image keys
+ * @type {RegExp}
+ */
+RichPresenceAssets.URL_PATTERN = /(https?|ftp|file):\/\//gm;
```
kyranet, line 410:
```suggestion
RichPresenceAssets.URL_PATTERN = /^(?:https?|ftp|file):\/\//gm;
```

kyranet (changes requested):

packages/discord.js/src/structures/Presence.js
almeidx: `gm` flags sound redundant in this scenario

vladfrangu: The URLs also don't start with https/ftp/file but rather with mp://some_thing/url according to the shown videos in DDevs

suneettipirneni:
After looking at the demo again, it seems we don't need to check for any special cases. Discord transforms the url for us, when returning it back.

PR #7365, opened 2022-01-28
revert: refactor(invite): make `channel` a getter

ImRodry:

This PR makes Invite.channel a regular property again as PartialGroupDMChannels do not get cached, so it was impossible to obtain them. This also removes Invite#channelId as this is not documented on Discord's side and thus is not sent

### DIFF
--- a/packages/discord.js/src/structures/Invite.js
+++ b/packages/discord.js/src/structures/Invite.js
@@ -151,16 +151,12 @@ class Invite extends Base {
       this.targetType ??= null;
     }
 
-    if ('channel_id' in data) {
+    if ('channel' in data) {
       /**
-       * The channel's id this invite is for
-       * @type {?Snowflake}
+       * The channel this invite is for
+       * @type {?Channel}
        */
-      this.channelId = data.channel_id;
-    }
-
-    if (data.channel) {
-      this.channelId ??= data.channel.id;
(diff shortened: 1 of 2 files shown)


kyranet (approved):

vladfrangu (approved):

iCrawl (approved):

Jiralite:
> This also removes Invite#channelId as this is not documented on Discord's side and thus is not sent

It's sent an documented via audit logs?

ImRodry:
> > This also removes Invite#channelId as this is not documented on Discord's side and thus is not sent
> 
> It's sent and documented via audit logs?

Audit logs don’t send invite objects, all they do is send the data through AuditLogEntry#changes so unless djs uses that to make an invite that shouldn’t be an issue, right?

Jiralite:
I guess so. The other change that comes would be `Invite#inviterId`, so that would have to be looked upon too.

ImRodry:
It turns out djs does transform the changes data into an Invite object as seen in
https://github.com/discordjs/discord.js/blob/388f53550cca7ded7350a050fda03c36e4c1fdf7/packages/discord.js/src/structures/GuildAuditLogs.js#L416-L420
Will revert this later today

SpaceEEC (approved):

PR #7259, opened 2022-01-12
feat(interaction): add `isRepliable` type guard

suneettipirneni:
**Please describe the changes this PR makes and why it should be merged:**

@almostSouji will be happy about this

**Status and versioning classification:**
Please move lines that apply to you out of the comment:
- Code changes have been tested against the Discord API, or there are no code changes
- I know how to update typings and have done so, or typings don't need updating
- This PR changes the library's interface (methods or parameters added)

PR #7371, opened 2022-01-31
feat: add `isRepliable` type guard to `InteractionResponses`

Syjalo:
Addition to #7259. 

### DIFF
--- a/packages/discord.js/src/structures/interfaces/InteractionResponses.js
+++ b/packages/discord.js/src/structures/interfaces/InteractionResponses.js
@@ -225,6 +225,14 @@ class InteractionResponses {
     return options.fetchReply ? this.fetchReply() : undefined;
   }
 
+  /**
+   * Indicates whether this interaction can be replied to.
+   * @returns {boolean}
+   */
+  isRepliable() {
+    return !(this.deferred || this.replied);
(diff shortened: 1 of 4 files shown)


packages/discord.js/src/structures/interfaces/InteractionResponses.js
```diff
@@ -225,6 +225,14 @@ class InteractionResponses {
     return options.fetchReply ? this.fetchReply() : undefined;
   }
 
+  /**
+   * Indicates whether this interaction can be replied to.
+   * @returns {boolean}
+   */
+  isRepliable() {
```
imranbarbhuiya, line 232: Interaction#isRepliable() is a type guard.

It narrows Interaction to all the Interactions which can be replied.  Like 6ou can't reply to autocomplete Interaction. So it removes it.

imranbarbhuiya, line 232: Interaction#isRepliable() is a type guard.

It narrows Interaction to all the Interactions which can be replied.  Like 6ou can't reply to autocomplete Interaction. So it removes it.

monbrey:
This seems worse than the existing implementation

PR #7375, opened 2022-01-31
docs: correct intent docs

suneettipirneni:


vladfrangu (approved):

iCrawl (approved):

kyranet (approved):

SpaceEEC (changes requested):
image
Very informative.

---

image
How is a non-clickable type "correct"?

Much less since it's of the same type as `IntentsBitField.Flags`, we surely don't accept that in the constructor.

---

And lastly: I see nothing wrong with the type before the change, it seems to be accurate to me.

ImRodry:
> How is a non-clickable type "correct"?

Isn’t the goal to make a -types documentation website and have these links point there? There’s a bunch of types in the docs that come from the package already

PR #7384, opened 2022-02-01
fix(guildmember): check if member has administrator permission

fowlerro:
Discord doesn't allow to timeout a member, which has administrator permission, even if he is manageable.
I added check to `guildMember.moderatable` which returns false if member has `Administrator` permission

### DIFF
--- a/packages/discord.js/src/structures/GuildMember.js
+++ b/packages/discord.js/src/structures/GuildMember.js
@@ -264,8 +264,8 @@ class GuildMember extends Base {
    */
   get moderatable() {
     return (
-      this.manageable &&
       !this.permissions.has(PermissionFlagsBits.Administrator) &&
+      this.manageable &&
       (this.guild.me?.permissions.has(PermissionFlagsBits.ModerateMembers) ?? false)
     );
   }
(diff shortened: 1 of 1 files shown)


ImRodry:
Discord *does* allow you to timeout a member with administrator permission, this is not needed IMO

almeidx:
- https://github.com/discord/discord-api-docs/pull/4390

thebigwolfy (approved):

packages/discord.js/src/structures/GuildMember.js
```diff
@@ -263,7 +263,11 @@ class GuildMember extends Base {
    * @readonly
    */
   get moderatable() {
-    return this.manageable && (this.guild.me?.permissions.has(PermissionFlagsBits.ModerateMembers) ?? false);
+    return (
+      this.manageable &&
+      !this.permissions.has(PermissionFlagsBits.Administrator) &&
```
vladfrangu, line 268: Shouldn't this check be done before the rest instead? (would make it a faster path for members that have Admin perm)

fowlerro: Oh I see, I changed that

SpaceEEC (approved):

DraftProducts (approved):

thebigwolfy (approved):

iCrawl (approved):

kyranet (approved):
Interesting find, I didn't notice this behaviour until now.

PR #7385, opened 2022-02-01
chore: add keepNames tsup option

DTrombett:

Add the `keepNames` option to tsup configs to avoid errors or such to have a minified name like `D`.

### DIFF
--- a/packages/builders/tsup.config.ts
+++ b/packages/builders/tsup.config.ts
@@ -6,6 +6,7 @@ export const tsup: Options = {
 	entryPoints: ['src/index.ts'],
 	format: ['esm', 'cjs'],
 	minify: true,
+	keepNames: true,
 	skipNodeModulesBundle: true,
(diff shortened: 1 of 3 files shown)


kyranet (approved):

ImRodry:
Can we please just remove minify? Looking at source code to trace back an error is extremely painful with this option and the size reduction is very small (although relatively significant, it’s only 15kb in rest for example)

vladfrangu:
I don't see how seeing where in our code the issue started from helps you at all

vladfrangu (approved):

ImRodry:
> I don't see how seeing where in our code the issue started from helps you at all

There can always be bugs in the package and even if there aren’t it’s good to know what methods are being called and how the code reached that point. That should not be taken away from us

SpaceEEC (approved):

iCrawl (approved):

PR #7389, opened 2022-02-02
docs: ApplicationCommandData typedef

KonkenBonken:
Updated ApplicationCommandData documentation to add default value of `type`, that `description` only is used on CHAT_INPUT and that the name must be in all lowercase when used in a CHAT_INPUT

### DIFF
--- a/packages/discord.js/src/structures/ApplicationCommand.js
+++ b/packages/discord.js/src/structures/ApplicationCommand.js
@@ -127,8 +127,9 @@ class ApplicationCommand extends Base {
   /**
    * Data for creating or editing an application command.
    * @typedef {Object} ApplicationCommandData
-   * @property {string} name The name of the command, must be in all lowercase if type is CHAT_INPUT
-   * @property {string} description The description of the command, if type is CHAT_INPUT
+   * @property {string} name The name of the command, must be in all lowercase if type is
+   * {@link ApplicationCommandType.ChatInput}
+   * @property {string} description The description of the command, if type is {@link ApplicationCommandType.ChatInput}
    * @property {ApplicationCommandType} [type=ApplicationCommandType.ChatInput] The type of the command
    * @property {ApplicationCommandOptionData[]} [options] Options for the command
    * @property {boolean} [defaultPermission=true] Whether the command is enabled by default when the app is added to a guild
--- a/packages/discord.js/src/structures/ApplicationCommand.js
+++ b/packages/discord.js/src/structures/ApplicationCommand.js
@@ -132,7 +132,8 @@ class ApplicationCommand extends Base {
    * @property {string} description The description of the command, if type is {@link ApplicationCommandType.ChatInput}
    * @property {ApplicationCommandType} [type=ApplicationCommandType.ChatInput] The type of the command
    * @property {ApplicationCommandOptionData[]} [options] Options for the command
-   * @property {boolean} [defaultPermission=true] Whether the command is enabled by default when the app is added to a guild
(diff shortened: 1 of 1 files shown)


vladfrangu (approved):

packages/discord.js/src/structures/ApplicationCommand.js
```diff
@@ -127,11 +127,11 @@ class ApplicationCommand extends Base {
   /**
    * Data for creating or editing an application command.
    * @typedef {Object} ApplicationCommandData
-   * @property {string} name The name of the command
-   * @property {string} description The description of the command
-   * @property {ApplicationCommandType} [type] The type of the command
+   * @property {string} name The name of the command, must be in all lowercase if type is CHAT_INPUT
+   * @property {string} description The description of the command, if type is CHAT_INPUT
+   * @property {ApplicationCommandType} [type="CHAT_INPUT"] The type of the command
```
ImRodry, line 132: I think this default was here for backwards compatibility when v13 released, we can probably get rid of it now or is that not needed?

almeidx (commented):
Should probably use `ApplicationCommandType.ChatInput` instead of `CHAT_INPUT`

packages/discord.js/src/structures/ApplicationCommand.js
KonkenBonken: I don't get what you mean sorry

KonkenBonken:
>

packages/discord.js/src/structures/ApplicationCommand.js
ImRodry: The type default can be removed now, as it was there for backwards compatibility but the maintainers should probably say if they want that

packages/discord.js/src/structures/ApplicationCommand.js
```diff
@@ -127,11 +127,11 @@ class ApplicationCommand extends Base {
   /**
    * Data for creating or editing an application command.
    * @typedef {Object} ApplicationCommandData
-   * @property {string} name The name of the command
-   * @property {string} description The description of the command
-   * @property {ApplicationCommandType} [type] The type of the command
+   * @property {string} name The name of the command, must be in all lowercase if type is CHAT_INPUT
+   * @property {string} description The description of the command, if type is CHAT_INPUT
```
kyranet, line 131:
```suggestion
   * @property {string} name The name of the command, must be in all lowercase if type is
   * {@link ApplicationCommandType.ChatInput}
   * @property {string} description The description of the command, if type is {@link ApplicationCommandType.ChatInput}
```

packages/discord.js/src/structures/ApplicationCommand.js
```diff
@@ -127,11 +127,11 @@ class ApplicationCommand extends Base {
...
+   * @property {ApplicationCommandType} [type=ApplicationCommandType.ChatInput] The type of the command
    * @property {ApplicationCommandOptionData[]} [options] Options for the command
-   * @property {boolean} [defaultPermission] Whether the command is enabled by default when the app is added to a guild
+   * @property {boolean} [defaultPermission=true] Whether the command is enabled by default when the app is added to a guild
```
kyranet, line 134: Lint is failing here because the line is too long.
```suggestion
   * @property {boolean} [defaultPermission=true] Whether the command is enabled by default when the app is added to a
   * guild
```

kyranet (changes requested):

kyranet (approved):

iCrawl (approved):

ImRodry:
commitlint!

SpaceEEC (approved):

PR #7398, opened 2022-02-04
docs: fix example about filter options

Dirain1700:
In docs(example: [message.awaitReactions()](https://discord.js.org/#/docs/discord.js/stable/class/Message?scrollTo=awaitReactions)), filter is first argument and no used cologne. However, this is undesirable because it causes malfunction.
ref: https://github.com/InkoHX/vm2-discordjs/pull/50/commits/32056dbdc3c2f87fc9f413dc122794dc74d0d35a

### DIFF
--- a/packages/discord.js/src/structures/Message.js
+++ b/packages/discord.js/src/structures/Message.js
@@ -449,8 +449,8 @@ class Message extends Base {
    * @returns {ReactionCollector}
    * @example
    * // Create a reaction collector
-   * const filter = (reaction, user) => reaction.emoji.name === '👌' && user.id === 'someId';
-   * const collector = message.createReactionCollector({ filter, time: 15_000 });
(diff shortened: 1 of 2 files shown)


imranbarbhuiya:
https://github.com/InkoHX/vm2-discordjs/commit/32056dbdc3c2f87fc9f413dc122794dc74d0d35a

This is basic javascript. You should know about it before using djs.

Dirain1700:
> [InkoHX/vm2-discordjs@32056db](https://github.com/InkoHX/vm2-discordjs/commit/32056dbdc3c2f87fc9f413dc122794dc74d0d35a)
> 
> This is basic javascript. You should know about it before using djs.

reactionFilter and messageFilter is defiend with basic JavaScript, but filter is exist on djs
but this seems invalod, I will close this.

ImRodry:
I’ve seen many people call their filters the most random names and then pass them with short-hand object notation and then wonder why it doesn’t work so this may not be the worst idea

PR #7402, opened 2022-02-04
feat(channel): add .url getter

Milo123459:

This simply adds a `.url` getter on channels. This can be specifically useful when working with moving threads between 2 channels.

### DIFF
--- a/packages/discord.js/src/structures/Channel.js
+++ b/packages/discord.js/src/structures/Channel.js
@@ -59,11 +59,11 @@ class Channel extends Base {
 
   /**
    * The URL to the channel
-   * @type {String}
+   * @type {string}
    * @readonly
    */
   get url() {
-    return `https://discord.com/channels/${this.isDM() ? "@me" : this.guild.id}/${this.id}`;
+    return `https://discord.com/channels/${this.isDM() ? '@me' : this.guild.id}/${this.id}`;
   }
 
   /**
--- a/packages/discord.js/src/structures/Channel.js
+++ b/packages/discord.js/src/structures/Channel.js
@@ -63,7 +63,7 @@ class Channel extends Base {
    * @readonly
    */
   get url() {
-    return `https://discord.com/channels/${this.isDMBased() ? '@me' : this.guild.id}/${this.id}`;
+    return `https://discord.com/channels/${this.isDMBased() ? '@me' : this.guildId}/${this.id}`;
   }
 
   /**
--- a/packages/discord.js/src/structures/Channel.js
+++ b/packages/discord.js/src/structures/Channel.js
@@ -58,6 +58,15 @@ class Channel extends Base {
   }
 
   /**
+   * The URL to the channel
+   * @type {String}
+   * @readonly
+   */
+  get url() {
+    return `https://discord.com/channels/${this.isDM() ? "@me" : this.guild.id}/${this.id}`;
+  }
+
+  /**
    * Whether this Channel is a partial
    * <info>This is always false outside of DM channels.</info>
    * @type {boolean}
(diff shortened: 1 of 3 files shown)


Milo123459:
Tests are failing from places I haven't even touched..?

packages/discord.js/src/structures/Channel.js
```diff
@@ -57,6 +57,15 @@ class Channel extends Base {
     return new Date(this.createdTimestamp);
   }
 
+  /**
+   * The URL to the channel
+   * @type {string}
+   * @readonly
+   */
+  get url() {
+    return `https://discord.com/channels/${this.isDM() ? '@me' : this.guild.id}/${this.id}`;
```
ImRodry, line 66:
```suggestion
    return `https://discord.com/channels/${this.isDMBased() ? '@me' : this.guild.id}/${this.id}`;
```

packages/discord.js/src/structures/ThreadChannel.js
```diff
@@ -230,6 +230,15 @@ class ThreadChannel extends Channel {
     return this.guild.channels.resolve(this.parentId);
   }
 
+  /**
+   * The URL to the channel
+   * @type {string}
+   * @readonly
+   */
+  get url() {
+    return `https://discord.com/channels/${this.guild.id}/${this.parent.id}/${this.id}`;
```
ImRodry, line 239: This is not correct, just remove it and keep the one from Channel

Milo123459: Mind explaining why? The URL for threads is channels/guild_id/channel_id/thread_id.

ImRodry: Simply because it isn’t. The format is the exact same one as the Channel one

Milo123459: Ah, I see that without referencing the parent channel it still works. Is it still worth removing this implementation? I feel this one is more explicit and better.

ImRodry: That URL links to the thread’s starter message and not the thread channel, so it’s not even correct

Jiralite: It is wrong. This does not take you to the channel, but the message of the thread. If the thread was orphaned, this would go nowhere.

packages/discord.js/typings/index.d.ts
```diff
@@ -491,6 +491,7 @@ export abstract class Channel extends Base {
   public id: Snowflake;
   public readonly partial: false;
   public type: ChannelType;
+  public url: string;
```
almeidx, line 494:
```suggestion
  public readonly url: string;
```

almeidx (changes requested):

packages/discord.js/src/structures/Channel.js
```diff
@@ -57,6 +57,15 @@ class Channel extends Base {
     return new Date(this.createdTimestamp);
   }
 
+  /**
+   * The URL to the channel
+   * @type {string}
+   * @readonly
+   */
+  get url() {
+    return `https://discord.com/channels/${this.isDMBased() ? '@me' : this.guild.id}/${this.id}`;
```
kyranet, line 66: This should use `guildId` since it's always guaranteed to be set:
```suggestion
    return `https://discord.com/channels/${this.isDMBased() ? '@me' : this.guildId}/${this.id}`;
```

kyranet (changes requested):

packages/discord.js/src/structures/Channel.js
Milo123459: Good catch!

kyranet (approved):

vladfrangu (approved):

iCrawl (approved):

thebigwolfy (approved):

PR #7403, opened 2022-02-04
chore(CONTRIBUTING.md): update to yarn instructions

GodderE2D:

Apparently, we're using Yarn now, so I think this should be updated. I also provided a link for how to install Yarn for people who don't know.

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -11,7 +11,7 @@ is a great boon to your development process.
 To get ready to work on the codebase, please do the following:
 
 1. Fork & clone the repository, and make sure you're on the **main** branch
-2. Run `yarn` ([install](https://yarnpkg.com/getting-started/install))
+2. Run `yarn --immutable` ([install](https://yarnpkg.com/getting-started/install))
 3. Code your heart out!
 4. Run `yarn test` to run ESLint and ensure any JSDoc changes are valid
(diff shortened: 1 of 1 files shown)


.github/CONTRIBUTING.md
```diff
@@ -11,7 +11,7 @@ is a great boon to your development process.
 To get ready to work on the codebase, please do the following:
 
 1. Fork & clone the repository, and make sure you're on the **main** branch
-2. Run `npm ci`
+2. Run `yarn` ([install](https://yarnpkg.com/getting-started/install))
```
kyranet, line 14: The flag keeps the lockfile unchanged and installs everything with the locked versions, similar to `npm ci`, and this also makes the install step *a lot faster*:
```suggestion
2. Run `yarn --immutable` ([install](https://yarnpkg.com/getting-started/install))
```

This is a Yarn v3 flag, which we use since 37c1cb4.

kyranet (changes requested):

kyranet (approved):

iCrawl (approved):

Milo123459:
Hmm, what about setting [this option](https://yarnpkg.com/configuration/yarnrc#enableImmutableInstalls) to true in the YarnRC? This is immutable when running a command like `yarn`, but should be mutable when running something like `yarn add`.

PR #7404, opened 2022-02-04
refactor: Make public builder props getters

suneettipirneni:

Stores raw api data in a `data` object and allows access to builder properties via getters.

### Rationale

There's quite a few benefits in using getters for builders:

- Currently builders relies on `Reflect#set` which is much slower than direct prop assignment, in one benchmark [I did it was waaay slower than direct assignment](https://www.measurethat.net/Benchmarks/ShowResult/263962). Beyond performance `Reflect#set` is less typesafe since property keys are strings and it allows values to be set as `any`. This PR removes the usage of these methods, and preserves the `readonly` nature of the fields.

- The classes are now more flexible. For example, right now if we decided to add a field to any builder class, it would be outputted in `toJSON` simply because `toJSON` just spreads `this`: `{ ...this }`. In addition, since we now have getters we can do any computation *before* a value is returned unlike a regular field.

- Constructor assignment is easy, in some cases it's just assigning `this.data = data`. `toJSON` is straightforward as well, as it's essentially just returning a copy of `this.data`.

- Allows for idiomatic js/ts class field naming. The props are simply retrieved from `this.data` so they're not constrained to any naming conventions used by the API.

### DIFF
--- a/packages/builders/src/components/ActionRow.ts
+++ b/packages/builders/src/components/ActionRow.ts
@@ -1,4 +1,4 @@
-import { type APIActionRowComponent, ComponentType } from 'discord-api-types/v9';
+import { type APIActionRowComponent, ComponentType, APIMessageComponent } from 'discord-api-types/v9';
 import type { ButtonComponent, SelectMenuComponent } from '..';
 import { Component } from './Component';
 import { createComponent } from './Components';
@@ -9,17 +9,31 @@ export type ActionRowComponent = ButtonComponent | SelectMenuComponent;
 
 // TODO: Add valid form component types
 
+export interface ActionRowData extends Omit<APIActionRowComponent, 'type' | 'components'> {
+	type?: ComponentType.ActionRow;
+	components?: Exclude<APIMessageComponent, APIActionRowComponent>[];
+}
+
 /**
  * Represents an action row component
  */
 export class ActionRow<T extends ActionRowComponent = ActionRowComponent> extends Component<
 	Omit<APIActionRowComponent, 'components'>
 > {
-	public readonly components: T[] = [];
+	public readonly components: T[];
 
-	public constructor({ components, ...data }: Omit<APIActionRowComponent, 'type'>) {
-		super({ type: ComponentType.ActionRow, ...data });
-		this.components = components.map(createComponent) as T[];
+	public constructor(data?: ActionRowData) {
+		// We don't destructure directly in the constructor because it can't properly
+		// handle possibly-undefined data, which causes invalid destructure runtime errors.
+		if (data?.components) {
+			const { components: initComponents, ...initData } = data;
+			super({ type: ComponentType.ActionRow, ...initData });
+		} else {
+			super({ type: ComponentType.ActionRow, ...data });
+		}
+
+		// eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
+		this.components = (data?.components?.map(createComponent) ?? []) as T[];
 	}
 
 	public get type(): ComponentType.ActionRow {
--- a/packages/builders/src/components/ActionRow.ts
+++ b/packages/builders/src/components/ActionRow.ts
@@ -12,12 +12,14 @@ export type ActionRowComponent = ButtonComponent | SelectMenuComponent;
 /**
  * Represents an action row component
  */
-export class ActionRow<T extends ActionRowComponent = ActionRowComponent> extends Component<APIActionRowComponent> {
+export class ActionRow<T extends ActionRowComponent = ActionRowComponent> extends Component<
+	Omit<APIActionRowComponent, 'components'>
+> {
 	public readonly components: T[] = [];
 
-	public constructor(data?: APIActionRowComponent & { type?: ComponentType.ActionRow }) {
+	public constructor({ components, ...data }: Omit<APIActionRowComponent, 'type'>) {
 		super({ type: ComponentType.ActionRow, ...data });
-		this.components = (data?.components.map(createComponent) ?? []) as T[];
+		this.components = components.map(createComponent) as T[];
 	}
 
 	public get type(): ComponentType.ActionRow {
--- a/packages/builders/src/components/ActionRow.ts
+++ b/packages/builders/src/components/ActionRow.ts
@@ -12,8 +12,7 @@ export type ActionRowComponent = ButtonComponent | SelectMenuComponent;
 /**
  * Represents an action row component
  */
-export class ActionRow<T extends ActionRowComponent = ActionRowComponent> extends Component {
-	protected declare data: APIActionRowComponent;
+export class ActionRow<T extends ActionRowComponent = ActionRowComponent> extends Component<APIActionRowComponent> {
 	public readonly components: T[] = [];
 
 	public constructor(data?: APIActionRowComponent & { type?: ComponentType.ActionRow }) {
--- a/packages/builders/src/components/ActionRow.ts
+++ b/packages/builders/src/components/ActionRow.ts
@@ -17,8 +17,7 @@ export class ActionRow<T extends ActionRowComponent = ActionRowComponent> extend
 	public readonly components: T[] = [];
 
 	public constructor(data?: APIActionRowComponent & { type?: ComponentType.ActionRow }) {
-		super(data);
-		this.data.type ??= ComponentType.ActionRow;
+		super({ type: ComponentType.ActionRow, ...data });
 		this.components = (data?.components.map(createComponent) ?? []) as T[];
 	}
 
@@ -41,8 +40,7 @@ export class ActionRow<T extends ActionRowComponent = ActionRowComponent> extend
 	 * @param components The components to set this row to
 	 */
 	public setComponents(components: T[]) {
-		this.components.slice(0, this.components.length);
-		this.components.push(...components);
+		this.components.splice(0, this.components.length, ...components);
 		return this;
 	}
 
--- a/packages/builders/src/components/ActionRow.ts
+++ b/packages/builders/src/components/ActionRow.ts
@@ -36,10 +36,6 @@ export class ActionRow<T extends ActionRowComponent = ActionRowComponent> extend
 		this.components = (data?.components?.map(createComponent) ?? []) as T[];
 	}
 
-	public get type(): ComponentType.ActionRow {
-		return this.data.type;
-	}
-
 	/**
 	 * Adds components to this action row.
 	 * @param components The components to add to this action row.
--- a/packages/builders/src/components/ActionRow.ts
+++ b/packages/builders/src/components/ActionRow.ts
@@ -1,6 +1,6 @@
-import { APIActionRowComponent, ComponentType } from 'discord-api-types/v9';
+import { type APIActionRowComponent, ComponentType } from 'discord-api-types/v9';
 import type { ButtonComponent, SelectMenuComponent } from '..';
-import type { Component } from './Component';
+import { Component } from './Component';
 import { createComponent } from './Components';
 
 export type MessageComponent = ActionRowComponent | ActionRow;
@@ -12,12 +12,18 @@ export type ActionRowComponent = ButtonComponent | SelectMenuComponent;
 /**
  * Represents an action row component
  */
-export class ActionRow<T extends ActionRowComponent = ActionRowComponent> implements Component {
+export class ActionRow<T extends ActionRowComponent = ActionRowComponent> extends Component {
+	protected declare data: APIActionRowComponent;
 	public readonly components: T[] = [];
-	public readonly type = ComponentType.ActionRow;
 
 	public constructor(data?: APIActionRowComponent & { type?: ComponentType.ActionRow }) {
+		super(data);
+		this.data.type ??= ComponentType.ActionRow;
 		this.components = (data?.components.map(createComponent) ?? []) as T[];
+	}
+
+	public get type(): ComponentType.ActionRow {
+		return this.data.type;
 	}
 
 	/**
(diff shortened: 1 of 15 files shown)


packages/builders/src/components/selectMenu/UnsafeSelectMenu.ts
```diff
@@ -1,45 +1,60 @@
...
-		this.placeholder = data?.placeholder;
-		this.min_values = data?.min_values;
-		this.max_values = data?.max_values;
-		/* eslint-disable @typescript-eslint/non-nullable-type-assertion-style */
-		this.custom_id = data?.custom_id as string;
-		/* eslint-enable @typescript-eslint/non-nullable-type-assertion-style */
-		this.disabled = data?.disabled;
+		super(data);
+		this.data.type ??= ComponentType.SelectMenu;
+		this.data.options ??= [];
```
kyranet, line 15: Please do not mutate input, and since we're mutating it... let's clone it.
```suggestion
		super({ type: ComponentType.SelectMenu, optioms: [], ...data });
```

packages/builders/src/components/Component.ts
```diff
@@ -3,13 +3,23 @@ import type { APIMessageComponent, ComponentType } from 'discord-api-types/v9';
 /**
  * Represents a discord component
  */
-export interface Component {
+export abstract class Component {
+	/**
+	 * The api data associated with this component
+	 */
+	protected data!: APIMessageComponent;
```
kyranet, line 10: Is the assertion needed? It seems set in the constructor.

packages/builders/src/components/ActionRow.ts
```diff
@@ -35,13 +41,14 @@ export class ActionRow<T extends ActionRowComponent = ActionRowComponent> implem
 	 * @param components The components to set this row to
 	 */
 	public setComponents(components: T[]) {
-		Reflect.set(this, 'components', [...components]);
+		this.components.slice(0, this.components.length);
+		this.components.push(...components);
```
kyranet, line 45:
```suggestion
		this.components.splice(0, this.components.length, ...components);
```

packages/builders/src/components/ActionRow.ts
```diff
@@ -12,14 +12,20 @@ export type ActionRowComponent = ButtonComponent | SelectMenuComponent;
...
  */
-export class ActionRow<T extends ActionRowComponent = ActionRowComponent> implements Component {
+export class ActionRow<T extends ActionRowComponent = ActionRowComponent> extends Component {
+	protected declare data: APIActionRowComponent;
 	public readonly components: T[] = [];
-	public readonly type = ComponentType.ActionRow;
 
 	public constructor(data?: APIActionRowComponent & { type?: ComponentType.ActionRow }) {
+		super(data);
+		this.data.type ??= ComponentType.ActionRow;
```
kyranet, line 21:
```suggestion
		super({ type: ComponentType.ActionRow, ...data });
```

packages/builders/src/components/selectMenu/UnsafeSelectMenu.ts
```diff
@@ -84,15 +99,16 @@ export class UnsafeSelectMenuComponent implements Component {
 	 * Sets the options on this select menu
 	 * @param options The options to set on this select menu
 	 */
-	public setOptions(options: SelectMenuOption[]) {
-		Reflect.set(this, 'options', [...options]);
+	public setOptions(options: UnsafeSelectMenuOption[]) {
+		this.options.splice(0, this.options.length);
+		this.options.push(...options);
```
kyranet, line 104:
```suggestion
		this.options.splice(0, this.options.length, ...options);
```

packages/builders/src/components/ActionRow.ts
kyranet: Also, just saying... this needs an unit test if nothing failed with this change, specially since `.slice` does not mutate the source array.

kyranet: As a side note, shouldn't we clone the array and do a direct assignment just like the old code did? This way we don't mutate input.

packages/builders/src/messages/embed/UnsafeEmbed.ts
```diff
@@ -22,80 +13,97 @@ export interface FooterOptions {
 }
 
 export class UnsafeEmbed implements APIEmbed {
+	protected data!: APIEmbed;
+
+	public constructor(data: APIEmbed = {}) {
+		this.data = data;
+		this.data.fields = data.fields ?? [];
```
kyranet, line 20:
```suggestion
		this.data = { fields: [], ...data };
```

packages/builders/src/messages/embed/UnsafeEmbed.ts
```diff
@@ -22,80 +13,97 @@ export interface FooterOptions {
...
+	public get video() {
+		return this.data.video;
+	}
 
 	/**
 	 * The embed author data
 	 */
-	public readonly author?: APIEmbedAuthor;
+	public get author() {
+		return this.data.author;
```
kyranet, line 92: JFYI, `embed.author.icon_url` is still a thing. Should we not map it?

packages/builders/src/messages/embed/UnsafeEmbed.ts
```diff
@@ -22,80 +13,97 @@ export interface FooterOptions {
...
-		this.image = data.image;
-		this.video = data.video;
-		this.author = data.author;
-		this.provider = data.provider;
-		this.footer = data.footer;
-		this.fields = data.fields ?? [];
-
-		if (data.timestamp) this.timestamp = new Date(data.timestamp).toISOString();
+	public get footer() {
+		return this.data.footer;
```
kyranet, line 106: Same here (`embed.footer.icon_url`).

kyranet (changes requested):

packages/builders/src/components/ActionRow.ts
suneettipirneni: > Also, just saying... this needs an unit test if nothing failed with this change, specially since `.slice` does not mutate the source array.

The source array gets overwritten in toJSON()

kyranet: I'm saying that `this.components.slice(0, this.components.length);` without return or assignment is basically a NOP.

packages/builders/src/components/Component.ts
```diff
@@ -1,15 +1,27 @@
...
+	public abstract readonly type: ComponentType;
+
 	/**
 	 * Converts this component to an API-compatible JSON object
 	 */
-	toJSON: () => APIMessageComponent;
+	public abstract toJSON(): APIMessageComponent;
+
+	public constructor(data: APIBaseMessageComponent<ComponentType>) {
+		this.data = data as DataType;
```
kyranet, line 25:
```suggestion
	public constructor(data: DataType) {
		this.data = data;
```

packages/builders/src/components/selectMenu/UnsafeSelectMenu.ts
```diff
@@ -1,45 +1,60 @@
...
-		this.disabled = data?.disabled;
+export class UnsafeSelectMenuComponent extends Component<Omit<APISelectMenuComponent, 'options'>> {
+	public readonly options: UnsafeSelectMenuOption[] = [];
+
+	public constructor(
+		data?: APISelectMenuComponent & { type?: ComponentType.SelectMenu; options?: APISelectMenuOption[] },
+	) {
+		super({ type: ComponentType.SelectMenu, ...data });
+		// eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
+		this.options = data?.options?.map((o) => new UnsafeSelectMenuOption(o)) ?? [];
```
kyranet, line 16: Same as `ActionRow`, `options` is being passed to super, which makes `usmc.data.options` and `usmc.options` both exist with mismatching values after using `usmc.addOptions()` or `usmc.setOptions()`.

The way to fix this is by changing the `data` parameter to `{ options, ...data }`.

packages/builders/src/components/ActionRow.ts
```diff
@@ -12,14 +12,18 @@ export type ActionRowComponent = ButtonComponent | SelectMenuComponent;
...
  * Represents an action row component
  */
-export class ActionRow<T extends ActionRowComponent = ActionRowComponent> implements Component {
+export class ActionRow<T extends ActionRowComponent = ActionRowComponent> extends Component<APIActionRowComponent> {
 	public readonly components: T[] = [];
-	public readonly type = ComponentType.ActionRow;
 
 	public constructor(data?: APIActionRowComponent & { type?: ComponentType.ActionRow }) {
+		super({ type: ComponentType.ActionRow, ...data });
 		this.components = (data?.components.map(createComponent) ?? []) as T[];
```
kyranet, line 20: Do we want to have `ar.data.components` and `ar.components`, even mismatching in values after using `ar.setComponents()` or `ar.addComponents()`?

The way to fix this is by changing the `data` parameter to `{ components, ...data }`.

Also make sure to edit the extends so it omits `components` just like `UnsafeSelectMenuComponent` does.

packages/builders/src/messages/embed/UnsafeEmbed.ts
```diff
@@ -1,101 +1,144 @@
...
+		this.data = { fields: [], ...data };
+		if (data.timestamp) this.data.timestamp = new Date(data.timestamp).toISOString();
+	}
+
 	/**
 	 * An array of fields of this embed
 	 */
-	public readonly fields: APIEmbedField[];
+	public get fields() {
+		return this.data.fields ?? [];
```
kyranet, line 40: Is the `?? []` needed? We're always ensuring an array in the constructor.

kyranet: Alternatively, we could just not initialise fields (but keep `this.data = { ...data }` to clone the object) and then create the array if it doesn't exist in the field methods, this way we save some memory when initialising embeds since we won't be instantiating an array all the time, even for objects that don't have them.

kyranet (changes requested):

kyranet (approved):
I still don't like that UnsafeEmbed always instantiates fields, but w/e, everything else LGTM.

suneettipirneni:
> I still don't like that UnsafeEmbed always instantiates fields, but w/e, everything else LGTM.

I've made the `fields` prop on embeds lazily initialized.

packages/builders/src/messages/embed/UnsafeEmbed.ts
```diff
@@ -126,7 +169,8 @@ export class UnsafeEmbed implements APIEmbed {
 	 * @param fields The fields to add
 	 */
 	public addFields(...fields: APIEmbedField[]): this {
-		this.fields.push(...Embed.normalizeFields(...fields));
+		if (!this.data.fields) this.data.fields = [];
+		this.fields?.push(...Embed.normalizeFields(...fields));
```
almeidx, line 173: Why is this using the `fields` getter? Also, should use the UnsafeEmbed class as to not perform validation
```suggestion
		this.data.fields ??= [];
		this.data.fields.push(...UnsafeEmbed.normalizeFields(...fields));
```

packages/builders/src/messages/embed/UnsafeEmbed.ts
```diff
@@ -138,7 +182,7 @@ export class UnsafeEmbed implements APIEmbed {
 	 * @param fields The replacing field objects
 	 */
 	public spliceFields(index: number, deleteCount: number, ...fields: APIEmbedField[]): this {
-		this.fields.splice(index, deleteCount, ...Embed.normalizeFields(...fields));
+		this.fields?.splice(index, deleteCount, ...Embed.normalizeFields(...fields));
```
almeidx, line 185: `data` property + `UnsafeEmbed` instead of `Embed`
```suggestion
		this.data.fields?.splice(index, deleteCount, ...UnsafeEmbed.normalizeFields(...fields));
```

packages/builders/src/messages/embed/UnsafeEmbed.ts
```diff
@@ -147,7 +191,7 @@ export class UnsafeEmbed implements APIEmbed {
 	 * @param fields The fields to set
 	 */
 	public setFields(...fields: APIEmbedField[]) {
-		this.spliceFields(0, this.fields.length, ...fields);
+		this.spliceFields(0, this.fields?.length ?? 0, ...fields);
```
almeidx, line 194:
```suggestion
		this.spliceFields(0, this.data.fields?.length ?? 0, ...fields);
```

packages/builders/src/messages/embed/UnsafeEmbed.ts
```diff
@@ -105,7 +148,7 @@ export class UnsafeEmbed implements APIEmbed {
 		return (
 			(this.title?.length ?? 0) +
 			(this.description?.length ?? 0) +
-			this.fields.reduce((prev, curr) => prev + curr.name.length + curr.value.length, 0) +
+			(this.fields ?? []).reduce((prev, curr) => prev + curr.name.length + curr.value.length, 0) +
 			(this.footer?.text.length ?? 0) +
 			(this.author?.name.length ?? 0)
 		);
```
almeidx, line 154: Shouldn't all of these be using the `data` property instead of the getters?
```suggestion
		return (
			(this.data.title?.length ?? 0) +
			(this.data.description?.length ?? 0) +
			(this.data.fields ?? []).reduce((prev, curr) => prev + curr.name.length + curr.value.length, 0) +
			(this.data.footer?.text.length ?? 0) +
			(this.data.author?.name.length ?? 0)
		);
```

packages/builders/src/messages/embed/UnsafeEmbed.ts
```diff
@@ -1,101 +1,144 @@
...
+		};
+	}
 
 	/**
 	 * Received video data
 	 */
-	public readonly video?: APIEmbedVideo;
+	public get video(): APIEmbedVideo | undefined {
+		if (!this.data.video) return undefined;
+		return this.data.video;
```
almeidx, line 109:
```suggestion
		return this.data.video;
```

almeidx (changes requested):

PR #7409, opened 2022-02-05
refactor: replace `WSCodes`, `WSEvents`, and `InviteScopes` with `discord-api-types` equivalent

almeidx:
Removed the `WSCodes`, `WSEvents`, and `InviteScopes` constants in favour of the `discord-api-types` equivalent.

- In the process of replacing the usages of `WSCodes` I noticed that the WebSocketManager file had a constant that used the `RPCErrorCodes` enum from `discord-api-types`, which I found odd. https://github.com/discordjs/discord.js/blob/d8184f94dd08daab37195f52828e06af5ed1c1e0/packages/discord.js/src/client/websocket/WebSocketManager.js#L27-L31
This was done in https://github.com/discordjs/discord.js/pull/6540, which seems to have been an oversight unless I'm missing something (let me know if this is the case!).
I replaced those with the `GatewayCloseCodes` equivalent error codes. However, previously, the array had the error code `1000`, which is [not documented](https://discord.com/developers/docs/topics/opcodes-and-status-codes#gateway-gateway-close-event-codes) (and not present in the -types enum), so I assumed its usage was a mistake and replaced it with `4000`, which is the "Unknown Error" equivalent in the gateway event close codes.


vladfrangu:
> However, previously, the array had the error code 1000, which is [not documented](https://discord.com/developers/docs/topics/opcodes-and-status-codes#gateway-gateway-close-event-codes) (and not present in the -types enum), so I assumed its usage was a mistake
It's usage was definitely not a mistake.

iCrawl:
1000 doesn't really need to be documented because its part of the WebSocket spec itself, it just so happened we used it and had it in the enum because of that.

vladfrangu:
Also, 4000 is a resumable close code, whereas 1000 isn't, so please fix that 🙏

vladfrangu (approved):

iCrawl (approved):

kyranet (approved):

PR #7410, opened 2022-02-05
fix(components): addMultiple/setMultiple accepting array and objects

thehairy:
This PR allows arrays in the following methods again:
```
SelectMenuComponent#addOptions
SelectMenuComponent#setOptions
ActionRow#addComponents
ActionRow#setComponents
Embed#addComponents
Embed#setComponents
```

### DIFF
--- a/packages/builders/src/messages/embed/UnsafeEmbed.ts
+++ b/packages/builders/src/messages/embed/UnsafeEmbed.ts
@@ -125,8 +125,8 @@ export class UnsafeEmbed implements APIEmbed {
 	 *
 	 * @param fields The fields to add
 	 */
-	public addFields(...fields: APIEmbedField[]): this {
-		this.fields.push(...Embed.normalizeFields(...fields));
+	public addFields(...fields: APIEmbedField[] | APIEmbedField[][]): this {
+		this.fields.push(...Embed.normalizeFields(...fields.flat(2)));
 		return this;
 	}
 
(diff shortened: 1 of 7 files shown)


kyranet:
You can just spread the array, `.flat()` is a very slow function.

packages/builders/src/messages/embed/UnsafeEmbed.ts
```diff
@@ -125,8 +125,8 @@ export class UnsafeEmbed implements APIEmbed {
 	 *
 	 * @param fields The fields to add
 	 */
-	public addFields(...fields: APIEmbedField[]): this {
-		this.fields.push(...Embed.normalizeFields(...fields));
+	public addFields(...fields: APIEmbedField[] | APIEmbedField[][]): this {
+		this.fields.push(...Embed.normalizeFields(...fields.flat(2)));
```
muchnameless, line 129: isn't the `flat`tening already done in [`normalizeFields`](https://github.com/thehairy/discord.js/blob/0de772769646055a72c7f44ae7a2c54fa591b18f/packages/builders/src/messages/embed/UnsafeEmbed.ts#L266)? That said, is `normalizeFields` even needed for `UnsafeEmbed`?

thebigwolfy (approved):

SpaceEEC (commented):
To add on to what kyra wrote:
I think it would be better to improve the documentation (with examples probably) to make it more obvious that explicit arrays are not supported and that you need to spread them (if you aren't just specifying it literally, then just drop the brackets).

thehairy:
> To add on to what kyra wrote: I think it would be better to improve the documentation (with examples probably) to make it more obvious that explicit arrays are not supported and that you need to spread them (if you aren't just specifying it literally, then just drop the brackets).

So we don't want to support arrays at all now?
Sounded different in internals when I brought that up.

thehairy:
Since we are going to stick with rest parameters only, I'm gonna close this.

PR #7414, opened 2022-02-06
fix: use png as extension for defaultAvatarURL

HerrTxbias:
This PR adds the options object to the defaultAvatarURL method, so no unexpected image formats are getting returned, when a user doesn't have an Avatar.

### DIFF
--- a/packages/discord.js/src/structures/User.js
+++ b/packages/discord.js/src/structures/User.js
@@ -149,11 +149,11 @@ class User extends Base {
 
   /**
    * A link to the user's default avatar
-   * @param {ImageURLOptions} [options={}] Options for the Image URL
-   * @returns {string}
+   * @type {string}
+   * @readonly
    */
-  defaultAvatarURL(options = {}) {
-    return this.client.rest.cdn.defaultAvatar(this.discriminator % 5, options);
+  get defaultAvatarURL() {
+    return this.client.rest.cdn.defaultAvatar(this.discriminator % 5);
   }
 
   /**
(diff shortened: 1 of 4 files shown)


Jiralite:
> so no unexpected image formats are getting returned

In what case is this happening? What are you encountering to warrant this change?

HerrTxbias:
> > so no unexpected image formats are getting returned
> 
> In what case is this happening? What are you encountering to warrant this change?

The default extension the rest package uses for CDN links is WEBP. See [here](https://github.com/discordjs/discord.js/blob/main/packages/rest/src/lib/CDN.ts#L248-L250).

But the default Avatars are not available as WEBP.

HerrTxbias:
I just checked the documentation again -> https://discord.com/developers/docs/reference#image-formatting-cdn-endpoints. It looks like they are only available as PNG, so i'm gonna change it to always use PNG as the extension.

Jiralite:
> I just checked the documentation again -> https://discord.com/developers/docs/reference#image-formatting-cdn-endpoints. It looks like they are only available as PNG, so i'm gonna change it to always use PNG as the extension.

Yea, this is correct. I would instead change this to always pass a .png:

https://github.com/discordjs/discord.js/blob/d8184f94dd08daab37195f52828e06af5ed1c1e0/packages/rest/src/lib/CDN.ts#L111-L117

Jiralite (approved):

kyranet (approved):

vladfrangu (approved):

SpaceEEC (approved):

iCrawl (approved):

PR #7424, opened 2022-02-07
refactor: allow discord.js builders to accept camelCase

### LINKED ISSUE #7376: [discord.js@dev] Using JSON structures do not accept camelCase.

discord.js

### Issue description

Steps to reproduce bug:
1. use the JSON way to create embeds, components or application commands
2. use `camelCase` for properties, such as `customId` or `minValues`
3. use the `.send()` method and place a `.catch()` 
4. execute the code

This may be because builders use `snake_case` but just reporting it to be sure

### Code sample

```javascript
message.reply({

  embeds: [{ 
    title: "Some title",
    description: "Some description",
    footer: { text: "some footer text", iconURL: message.member.displayAvatarURL() }    
  }],

  components: [
    { type: 1, components: [{ type: 2, label: "Button Label", style: 1, customId: "somecode" }] },
    { type: 1, components: [
      { type: 3, customId: "selectMenu", minValues: 2, options: [
        { label: "Label1", description: "Desc1", value: "l1" },
        { label: "Label2", description: "Desc2", value: "l2" }
      ] }
    ] }
  ]

})
```

### Package version

14.0.0-dev.1643544396.388f535

### Node.js version

16.9

### Operating system

Windows 10

Medium (should be fixed soon)

388f535

suneettipirneni:

Closes #7376 

Extends builder components in discord.js to allow both `camelCase` and `snake_case` inputs.


packages/discord.js/src/util/Components.js
```diff
@@ -0,0 +1,70 @@
...
+
+/**
+ * @typedef {BaseComponentData} ActionRowData
+ * @property {ComponentData[]} components
+ */
+
+/**
+ * @typedef {BaseComponentData} ButtonComponentData
+ * @property {ButtonStyle} style
+ * @property {boolean?} disabled
```
vitorlops, line 16:
```suggestion
 * @property {?boolean} disabled
```
The `?` shouldn't you go first?

imranbarbhuiya:
Why not embed too? also can we get that `#equal` in embed?

or can we get `embedEqual`, `xComponentEqual` in new component file?

KhafraDev:
I still don't see why this can't be left to the user 🤷

suneettipirneni:
> I still don't see why this can't be left to the user 🤷

We had discussions about this internally, and we decided that this library shouldn't be that low level. We'd prefer to keep idiomatic js field names and have the library take care of the rest. The regular builders without json transformations are still available directly from the builders packages.

KhafraDev:
However it doesn't make much sense in this case, and leads to a bunch of other discussions about what the library should do. In this sense, snake_case is the only case accepted by Discord, so the data the user provides is the data that Discord receives.

suneettipirneni:
> However it doesn't make much sense in this case, and leads to a bunch of other discussions about what the library should do. In this sense, snake_case is the only case accepted by Discord, so the data the user provides is the data that Discord receives.

I mean almost all of props in the library structures use camelcased naming schemes. Beyond classes a lot of pure json option objects are intentionally camelcased. So it would be inconsistent if message payload data didn't follow suit.

KhafraDev:
For the library it makes sense, but this is data that is being sent to Discord, which only accepts snake_case.

suneettipirneni:
> For the library it makes sense, but this is data that is being sent to Discord, which only accepts snake_case.

It's sent on behalf of the user by the library. The library the developer never sees any of the actual endpoints like they would if the used a lower level library like /rest for example.

monbrey:
> For the library it makes sense, but this is data that is being sent to Discord, which only accepts snake_case.

The data being sent to Discord is still snake_case, and isn't relevant to the changes being made here.

The standalone builders which exist under the /builders structures also haven't changed. They still only accept snake_case because they were designed to be used with /rest, to send data to Discord.

This PR extends those builder classes within the discord.js package to accept camelCase, which is consistent with the rest of discord.js and idiomatic JavaScript conventions.

KhafraDev:
> and isn't relevant to the changes being made here.

It's entirely relevant... this change adds another layer of abstraction between the data a dev provides and the data sent to Discord. Not only this, but the behavior of discord.js is no longer uniform with builders with this change, which will definitely lead to confusion.

> Up to the developer to decide. In this case, the developers of discord.js are free to decide on a convention. It's camelCase

and yet, both snake_case & camelCase are accepted here 😉. Theoretically, discord.js should only accept camelCase, since it's *the* convention throughout the rest of the lib, while builders uses snake_case.

monbrey:
> the behavior of discord.js is no longer uniform with builders with this change

It never should have been. discord.js should be consistent with discord.js

> discord.js should only accept camelCase

I agree, at least as far as the end user interface goes. But these classes are also used internally, when we receive snake_case data from Discord.

They always accepted both in the first place.

iCrawl (approved):

SpaceEEC (approved):

thebigwolfy (approved):

kyranet (approved):

ImRodry:
Your commit was meant to say embeds not modals I imagine but why not this one? It’d be within the scope since we’re extending builders classes

suneettipirneni:
> Your commit was meant to say embeds not modals I imagine but why not this one? It’d be within the scope since we’re extending builders classes

Mainly because the getters PR makes comparisons between components much easier.

PR #7433, opened 2022-02-09
feat(GuildChannelManager) Incluses data 'name' in object

thebigwolfy:
Hello, i am french. Sorry for language 😅 Thanks you, so, i inclused in function "create" and "createWebhook" the data "name". I am me logic 🤔

### DIFF
--- a/packages/discord.js/src/managers/GuildChannelManager.js
+++ b/packages/discord.js/src/managers/GuildChannelManager.js
@@ -115,7 +115,8 @@ class GuildChannelManager extends CachedManager {
    *   .catch(console.error);
    * @example
    * // Create a new channel with permission overwrites
-   * guild.channels.create('new-voice', {
+   * guild.channels.create({
+   *   name: 'new-voice',
    *   type: ChannelType.GuildVoice,
    *   permissionOverwrites: [
    *      {
@@ -126,8 +127,8 @@ class GuildChannelManager extends CachedManager {
    * })
    */
   async create(
-    name,
     {
+      name,
       type,
       topic,
       nsfw,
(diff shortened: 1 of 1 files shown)


imranbarbhuiya:
You need to update the typings too

thebigwolfy:
That is to say ?

thebigwolfy:
> You need to update the typings too

That is to say ?

almeidx (changes requested):
You need to update:
- the jsdocs (you updated the examples on them, but not the parameters and the options typedefs)
- the [typings](https://github.com/discordjs/discord.js/blob/main/packages/discord.js/typings/index.d.ts#L2769)

packages/discord.js/src/managers/GuildChannelManager.js
```diff
@@ -180,14 +181,15 @@ class GuildChannelManager extends CachedManager {
...
+   * guild.channels.createWebhook('222197033908436994', {
+   *   name: 'Hello!',
    *   avatar: 'https://i.imgur.com/mI8XcpG.jpg',
    *   reason: 'Needed a cool new Webhook'
    * })
    *   .then(console.log)
    *   .catch(console.error)
    */
-  async createWebhook(channel, name, { avatar, reason } = {}) {
+  async createWebhook(channel, { name, avatar, reason } = {}) {
```
thehairy, line 192:
```suggestion
  async createWebhook({ channel, name, avatar, reason } = {}) {
```

thehairy (changes requested):
Why do you put name in the object, but not the channel?
If you move that, I would just make it a whole object.

packages/discord.js/src/managers/GuildChannelManager.js
imranbarbhuiya: I agree. Btw in your change channel is required but adding `= {}` the whole object will become optional. No?

almeidx: The channel name is not optional either, so moving the name to the options object it would already make it required.

So the `= {}` should be removed

thebigwolfy:
I have editing code ( docs... ), so, i don't push merging, why ?

PR #7439, opened 2022-02-09
refactor: allow typescript files in discord.js

suneettipirneni:

This modifies the build behavior of discord.js. It compiles all js and ts files into `./dist`. This is beneficial as it allows us granularly transition the codebase to typescript, rather than try to lump a massive amount of changes in all at once.

As a proof of concept I converted some classes to typescript. I've also testing running the package with after it being built with the typescript compiler and everything seems to be working as expected.

TODO:
- [x] Use eslint overrides to parse both javascript and typescript with different parsers.


kyranet (changes requested):
Before going much further, please do not use `export default`. Prefer named exports.

packages/discord.js/src/index.js
```diff
@@ -97,6 +97,9 @@ exports.UnsafeEmbed = require('@discordjs/builders').UnsafeEmbed;
 exports.Emoji = require('./structures/Emoji').Emoji;
 exports.Guild = require('./structures/Guild').Guild;
 exports.GuildAuditLogs = require('./structures/GuildAuditLogs');
+/**
+ * @type {GuildAuditLogsEntry}
+ */
```
ImRodry, line 102: Why?

packages/discord.js/package.json
```diff
@@ -12,7 +12,7 @@
     "prepublishOnly": "yarn lint && yarn test",
     "changelog": "git cliff --prepend ./CHANGELOG.md -l -c ./cliff.toml -r ../../ --include-path 'packages/discord.js/*'"
   },
-  "main": "./src/index.js",
+  "main": "./dist/src/index.js",
```
ImRodry, line 15: Usually src is the source and dist is the compiled folder, shouldn’t it be just dist/index.js?

packages/discord.js/src/util/Util.ts
```diff
@@ -18,34 +16,41 @@ class Util extends null {
    * @param {...Object<string, boolean|string>} [props] Specific properties to include/exclude.
    * @returns {Object}
    */
-  static flatten(obj, ...props) {
+  static flatten(obj: object, ...props: Iterable<[string, boolean | string]>[]): unknown {
```
ImRodry, line 19: It’s good practice to use unknown in this specific situation 
```suggestion
  static flatten(obj: unknown, ...props: Iterable<[string, boolean | string]>[]): unknown {
```

packages/discord.js/src/util/Util.ts
```diff
@@ -130,7 +135,7 @@ class Util extends null {
       codeBlockContent = true,
       inlineCodeContent = true,
     } = {},
-  ) {
+  ): string {
```
ImRodry, line 138: Can’t this return type be inferred?

ImRodry (changes requested):

packages/discord.js/src/index.js
suneettipirneni: Because it can't compile without the explicit type annotation here.

packages/discord.js/src/util/Util.ts
suneettipirneni: `unknown` is a wider type than `object`

ImRodry: Exactly, and this function accepts any kind of input as seen by the check in the line right below this one, so this should be changed

packages/discord.js/src/index.js
ImRodry: But it’s js ?

packages/discord.js/src/util/Util.ts
```diff
@@ -316,7 +321,7 @@ class Util extends null {
    * @returns {Object}
    * @private
    */
-  static cloneObject(obj) {
+  static cloneObject(obj: object) {
```
imranbarbhuiya, line 324:
```suggestion
  static cloneObject<T>(obj: T)T {
```

why not something like this?

packages/discord.js/src/util/Util.ts
```diff
@@ -327,13 +332,14 @@ class Util extends null {
    * @returns {Object}
    * @private
    */
-  static mergeDefault(def, given) {
+  static mergeDefault(def: Record<string, unknown>, given: Record<string, unknown>) {
```
imranbarbhuiya, line 335:
```suggestion
  static mergeDefault<T>(def: Record<string, T>, given: Record<string, T>): Record<string, T> {
```

packages/discord.js/src/index.js
suneettipirneni: Yea but it's run through the TS compiler.

packages/discord.js/src/util/Util.ts
```diff
@@ -384,7 +390,7 @@ class Util extends null {
    * @returns {number}
    * @private
    */
-  static moveElementInArray(array, element, newIndex, offset = false) {
+  static moveElementInArray(array: unknown[], element: unknown, newIndex: number, offset = false) {
```
imranbarbhuiya, line 393:
```suggestion
  static moveElementInArray<T>(array: T[], element: T, newIndex: number, offset = false): T[] {
```

packages/discord.js/src/util/Util.ts
```diff
@@ -499,10 +505,18 @@ class Util extends null {
    * @returns {Promise<Channel[]|Role[]>} Updated item list, with `id` and `position` properties
    * @private
    */
-  static async setPosition(item, position, relative, sorted, client, route, reason) {
+  static async setPosition(
```
imranbarbhuiya, line 508: Its return type will be mixed of both Channel and Role. No?

imo adding 2 overload types is necessary.

packages/discord.js/src/util/Util.ts
```diff
@@ -542,7 +556,9 @@ class Util extends null {
         }
       })
       .replace(/<#[0-9]+>/g, input => {
-        const mentionedChannel = channel.client.channels.cache.get(input.replace(/<|#|>/g, ''));
+        const mentionedChannel = channel.client.channels.cache.get(
+          input.replace(/<|#|>/g, ''),
+        ) as GuildTextBasedChannel;
```
imranbarbhuiya, line 561:
```suggestion
        )!;
```

iirc it returns this type so assertion isn't needed.

packages/discord.js/src/index.js
ImRodry: Alright ig

packages/discord.js/tsconfig.json
```diff
@@ -3,8 +3,12 @@
...
+    "lib": ["ESNext"],
 
     // Completeness
-    "skipDefaultLibCheck": true
-  }
+    "skipDefaultLibCheck": true,
+    "allowJs": true,
+    "outDir": "./dist"
+  },
+  "include": ["src/**/*", "typings/**/*"]
```
imranbarbhuiya, line 13: This will compile to `dist/src`. So why not move types inside src and auto gen types?

imranbarbhuiya (changes requested):

packages/discord.js/tsconfig.json
suneettipirneni: If we autogen type files we'll end up with definitions that are worse than our manual definitions, since they're pulled from js.

packages/discord.js/package.json
suneettipirneni: We use the index.d.ts types which aren't in the source

packages/discord.js/src/util/Util.ts
suneettipirneni: It doesn't return that type.

packages/discord.js/src/util/Util.ts
suneettipirneni: doesn't seem to like when it's not annotated

imranbarbhuiya:
Can we get a new branch for typescript conversion so that anyone can contribute by converting atleast 1 file.

monbrey:
> Can we get a new branch for typescript conversion so that anyone can contribute by converting atleast 1 file.

Isn't the point of this feature being in main that separate branching wouldn't be required to contribute individual files.

packages/discord.js/package.json
ImRodry: This could be the time to move that folder then since having dist/src doesn’t make much sense imo

imranbarbhuiya: > This could be the time to move that folder then since having dist/src doesn’t make much sense imo

https://github.com/discordjs/discord.js/pull/7439#discussion_r803236250

packages/discord.js/src/util/Colors.ts
```diff
@@ -1,6 +1,4 @@
-'use strict';
-
-module.exports = {
+export const Colors = {
```
kyranet, line 1: Why are many things an enum, but this one isn't?

packages/discord.js/src/util/Util.ts
```diff
@@ -558,12 +593,20 @@ class Util extends null {
...
+import type {
+  AnyChannel,
+  ColorResolvable,
+  EmojiIdentifierResolvable,
+  GuildTextBasedChannel,
+  MakeErrorOptions,
+  TextBasedChannel,
+} from '../../typings';
+import type { Role } from '../structures/Role';
+import type Client from '../client/Client';
```
kyranet, line 612: Those don't need to go at the bottom, since they're type imports.

packages/rest/__tests__/util.ts
```diff
@@ -0,0 +1,14 @@
...
+}
+
+export function jsonHeaders(headers: Record<string, string> = {}) {
+	return {
+		headers: {
+			'content-type': 'application/json',
+			...headers,
+		},
+	};
+}
```
kyranet, line 14: What's... this?

kyranet (changes requested):

packages/discord.js/src/util/Util.ts
vladfrangu: Blame VSC Autoimport :P

packages/discord.js/src/util/Colors.ts
vladfrangu: It does align more with how it was before, and I don't see why we'd need a reverse mapping

packages/discord.js/package.json
```diff
@@ -59,8 +59,11 @@
   "devDependencies": {
```
vladfrangu, line 59: For a future PR, shouldn't we move all dev deps to the root of the project? CC @iCrawl

vladfrangu (changes requested):

packages/discord.js/src/util/Util.ts
```diff
@@ -327,13 +342,16 @@ class Util extends null {
    * @returns {Object}
    * @private
    */
-  static mergeDefault(def, given) {
+  private static mergeDefault<T>(def: Record<string, T>, given?: Record<string, T>): Record<string, T> {
```
imranbarbhuiya, line 345:
```suggestion
  private static mergeDefault<T>(def: Record<string, T>, given?: Record<string, Partial<T>>): Record<string, T> {
```

packages/discord.js/src/util/Util.ts
```diff
@@ -402,8 +420,8 @@ class Util extends null {
    * @param {boolean} [allowEmpty=true] Whether an empty string should be allowed
    * @returns {string}
    */
-  static verifyString(
-    data,
+  public static verifyString(
+    data: string,
```
imranbarbhuiya, line 424: This string can be unknown type and its return type should be `data is string`

ckohen:
> Also import files with .js extension. So that in future if u plan to make the lib ems only then you've to readd the extensions.

Alienating a large portion of the userbase doesn't sound great to me

Anyways, we can use the esm / TS syntax for imports and exports in js files now and typescript will compile to cjs compatible, even if the rest of the file hasn't changed to TS yet

packages/discord.js/package.json
ImRodry: I don’t see how that’s related to what I said. I simply suggested moving the typings folder into src

xHyroM:
> Also import files with `.js` extension. So that in future if u plan to make the lib ems only then you've to readd the extensions. Saying it now just because you are changing all the imports to named import so u can add js along with that.

I think someone already said this here but as with the rest module everything will be converted to commonjs. I think discord.js can't even afford to make a change like "the module will be ESM only, you'll need typescript or things like that". If only for the reason that very many people who use discord.js are beginners and then they would all complain. In my opinion it's probably best to do what `@sapphire/framework` has, that the code will be converted to cjs and then using the https://www.npmjs.com/package/gen-esm-wrapper module an ESM wrapper will be generated.

packages/discord.js/src/util/Colors.ts
suneettipirneni: I'm sticking with `const` here because a color can be any `number`, this structure just defines some predetermined ones.

packages/discord.js/tsconfig.json
```diff
@@ -3,8 +3,12 @@
   "compilerOptions": {
```
IRONM00N, line 3: Could we enable could we enable [declarationMap](https://www.typescriptlang.org/tsconfig#declarationMap) too? It would be nice to be able to see the source code instead of the generated typings.

PR #7457, opened 2022-02-13
fix(dataresolver): ensure fetched file is convert to a buffer

nev-r:
creating an emoji using a web URL is currently nonfunctional in `main` but working in `stable`.

`stable`->`main` removed `resolveFileAsBuffer` but `DataResolver.resolveBase64` still expected a Buffer,
and `resolveFile` provided a ReadableStream.

there's no reason to readd/reimplement that function since it only acted as an intermediary between those 2 other static methods, in this one specific case. so now `resolveFile` will directly return a Buffer, as the typing suggests it's supposed to.

### DIFF
--- a/packages/discord.js/src/util/DataResolver.js
+++ b/packages/discord.js/src/util/DataResolver.js
@@ -1,7 +1,7 @@
 'use strict';
 
 const { Buffer } = require('node:buffer');
-const fs = require('node:fs');
+const fs = require('node:fs/promises');
 const path = require('node:path');
 const stream = require('node:stream');
 const fetch = require('node-fetch');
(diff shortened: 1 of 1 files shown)


packages/discord.js/src/util/DataResolver.js
```diff
@@ -67,7 +67,7 @@ class DataResolver extends null {
       return image;
     }
     const file = await this.resolveFile(image);
-    return DataResolver.resolveBase64(file);
+    return this.resolveBase64(file);
```
almeidx, line 70: The method is static
```suggestion
    return DataResolver.resolveBase64(file);
```

nev-r: hm, just standardizing across the file. every other other static->static reference is using `this`. should the rest of the file be swapped over the other way then? as DataResolver is just a container for static methods, there isn't really a binding problem here.

DTrombett: > The method is static
> ```suggestion
>     return DataResolver.resolveBase64(file);
> ```

`this` is `DataResolver` here so no need of this change...

almeidx: > just standardizing across the file. every other static->static reference is using `this`

Didn't notice the rest of the static methods were using `this`, you can ignore this suggestion 

> should the rest of the file be swapped over the other way then

IMO, yes, but that's not up to me

SpaceEEC (approved):
LGTM

I went ahead and also applied this fix to the part resolving a local file path. 😄

vladfrangu (approved):

PR #7464, opened 2022-02-14
fix: use case converter for json component serialization

suneettipirneni:

Fixes a bug where components that were converted snake case would have their properties overwritten with `undefined` when spreading.

**Note: this library has been tested against strings like `URL` and it serializes it into snake case properly.**


packages/discord.js/src/util/Components.js
```diff
@@ -43,28 +43,49 @@
...
-      options: data?.options,
-      placeholder: data?.placeholder,
-      min_values: data?.minValues ?? data?.min_values,
-      max_values: data?.maxValues ?? data?.max_values,
-      components: data?.components?.map(c => Components.transformJSON(c)),
-    };
-  }
-}
+/**
+ * @typedef {Object} EmbedData
```
vladfrangu, line 47: Embed stuff in Components?

vladfrangu (approved):

packages/discord.js/src/util/Components.js
```diff
@@ -42,29 +42,3 @@
 /**
  * @typedef {ActionRowData|ButtonComponentData|SelectMenuComponentData} ComponentData
  */
-
-class Components extends null {
```
ImRodry, line 46: a file with only typedefs?

packages/discord.js/src/util/Embeds.js
```diff
@@ -46,36 +46,3 @@
  * @property {string} value
  * @property {?boolean} inline
  */
-
-class Embeds extends null {
```
ImRodry, line 50: ditto

packages/discord.js/src/util/Transformers.js
```diff
@@ -0,0 +1,20 @@
...
+class Transformers extends null {
+  /**
+   * Transforms camel-cased keys into snake cased keys
+   * @param {*} obj The object to transform
+   * @returns {*}
+   */
+  static toSnakeCase(obj = {}) {
+    return transform(obj, (acc, value, key, target) => {
+      const camelKey = Array.isArray(target) ? key : snakeCase(key);
+      acc[camelKey] = Boolean(value) && typeof value === 'object' ? Transformers.toSnakeCase(value) : value;
```
ImRodry, line 15: why Boolean(value)?

packages/discord.js/src/util/Components.js
suneettipirneni: Yeah

packages/discord.js/src/util/Transformers.js
suneettipirneni: Because `null`'s `typeof` is `object`

ImRodry: > Because `null`'s `typeof` is `object`

a ternary is an if, an if turns everything into a boolean already

packages/discord.js/src/util/Components.js
ImRodry: You can move them to Util, this feels weird like that

packages/discord.js/typings/index.d.ts
```diff
@@ -2407,17 +2407,11 @@ export class Formatters extends null {
   public static userMention: typeof userMention;
 }
 
-export type ComponentData = ActionRowComponentData | ButtonComponentData | SelectMenuComponentData;
-
-export class Components extends null {
-  private constructor();
-  public static transformJSON(data: ComponentData | APIMessageComponent): APIMessageComponent;
+export class Transformers extends null {
```
kyranet, line 2410: This is not exported, though.

packages/discord.js/src/util/Transformers.js
```diff
@@ -0,0 +1,20 @@
...
+
+class Transformers extends null {
+  /**
+   * Transforms camel-cased keys into snake cased keys
+   * @param {*} obj The object to transform
+   * @returns {*}
+   */
+  static toSnakeCase(obj = {}) {
+    return transform(obj, (acc, value, key, target) => {
+      const camelKey = Array.isArray(target) ? key : snakeCase(key);
```
kyranet, line 14: I don't find this in the docs... what is `target`? And in which cases can it be an array? The [docs](https://lodash.com/docs/4.17.15#transform) are unclear about it.

kyranet (changes requested):

packages/discord.js/src/util/Transformers.js
suneettipirneni: `target` is the original object it can be an array if any props on the object are an array or the object is an array itself.

iCrawl (approved):

packages/discord.js/src/util/Transformers.js
```diff
@@ -0,0 +1,23 @@
...
+
+class Transformers extends null {
+  /**
+   * Transforms camel-cased keys into snake cased keys
+   * @param {*} obj The object to transform
+   * @returns {*}
+   */
+  static toSnakeCase(obj = {}) {
+    if (typeof obj !== 'object' || !obj) return obj;
+    if (Array.isArray(obj)) return obj.map(Transformers.toSnakeCase);
```
vladfrangu, line 13:
```suggestion
    if (Array.isArray(obj)) return obj.map((item) => Transformers.toSnakeCase(item));
```

vladfrangu (changes requested):

vladfrangu (approved):

kyranet (approved):

SpaceEEC (approved):
Looking fine as-is, but maybe we want to use a built-in for that (see comment) instead.

PR #7466, opened 2022-02-15
feat: add missing `v13` component methods

suneettipirneni:

Adds
- `Embed#hexColor`
- `#equals`
- Select menu allows objects to be passed in to `setOptions` and `addOptions` instead of exclusively just classes.

### DIFF
--- a/packages/builders/__tests__/messages/embed.test.ts
+++ b/packages/builders/__tests__/messages/embed.test.ts
@@ -131,12 +131,9 @@ describe('Embed', () => {
 
 			// @ts-expect-error
 			expect(() => embed.setColor('RED')).toThrowError();
-		});
-
-		test('GIVEN an embed with a valid color THEN does not throw error', () => {
-			const embed = new Embed();
-			expect(() => embed.setColor([42, 36, 100])).not.toThrowError();
-			expect(() => embed.setColor(0xffffff)).not.toThrowError();
+			// @ts-expect-error
+			expect(() => embed.setColor([42, 36])).toThrowError();
+			expect(() => embed.setColor([42, 36, 1000])).toThrowError();
 		});
 	});
 
--- a/packages/builders/__tests__/messages/embed.test.ts
+++ b/packages/builders/__tests__/messages/embed.test.ts
@@ -115,10 +115,8 @@ describe('Embed', () => {
 		});
 
 		test('GIVEN an embed using Embed#setColor THEN returns valid toJSON data', () => {
-			const embed = new Embed();
-			embed.setColor(0xff0000);
-
-			expect(embed.toJSON()).toStrictEqual({ color: 0xff0000 });
+			expect(new Embed().setColor(0xff0000).toJSON()).toStrictEqual({ color: 0xff0000 });
+			expect(new Embed().setColor([242, 66, 245]).toJSON()).toStrictEqual({ color: 0xf242f5 });
 		});
 
 		test('GIVEN an embed with a pre-defined color THEN unset color THEN return valid toJSON data', () => {
@@ -133,6 +131,12 @@ describe('Embed', () => {
 
 			// @ts-expect-error
 			expect(() => embed.setColor('RED')).toThrowError();
+		});
+
+		test('GIVEN an embed with a valid color THEN does not throw error', () => {
+			const embed = new Embed();
+			expect(() => embed.setColor([42, 36, 100])).not.toThrowError();
+			expect(() => embed.setColor(0xffffff)).not.toThrowError();
 		});
 	});
 
--- a/packages/builders/src/components/ActionRow.ts
+++ b/packages/builders/src/components/ActionRow.ts
@@ -2,6 +2,7 @@ import { type APIActionRowComponent, ComponentType, APIMessageComponent } from '
 import type { ButtonComponent, SelectMenuComponent } from '..';
 import { Component } from './Component';
 import { createComponent } from './Components';
+import isEqual from 'fast-deep-equal';
 
 export type MessageComponent = ActionRowComponent | ActionRow;
 
@@ -45,5 +46,12 @@ export class ActionRow<T extends ActionRowComponent = ActionRowComponent> extend
 			...this.data,
 			components: this.components.map((component) => component.toJSON()),
 		};
+	}
+
+	public equals(other: APIActionRowComponent<APIMessageComponent> | ActionRow) {
+		if (other instanceof ActionRow) {
+			return isEqual(other.data, this.data);
+		}
+		return isEqual(other, this.data);
 	}
 }
--- a/packages/builders/src/components/ActionRow.ts
+++ b/packages/builders/src/components/ActionRow.ts
@@ -50,8 +50,11 @@ export class ActionRow<T extends ActionRowComponent = ActionRowComponent> extend
 
 	public equals(other: APIActionRowComponent<APIMessageComponent> | ActionRow) {
 		if (other instanceof ActionRow) {
-			return isEqual(other.data, this.data);
+			return isEqual(other.data, this.data) && isEqual(other.components, this.components);
 		}
-		return isEqual(other, this.data);
+		return isEqual(other, {
+			...this.data,
+			components: this.components.map((component) => component.toJSON()),
+		});
 	}
 }
--- a/packages/builders/src/components/Component.ts
+++ b/packages/builders/src/components/Component.ts
@@ -22,6 +22,7 @@ export abstract class Component<
 	protected readonly data: DataType;
 
 	public abstract toJSON(): APIMessageComponent;
+
 	public abstract equals(other: Component | APIActionRowComponentTypes): boolean;
 
 	public constructor(data: DataType) {
--- a/packages/builders/src/components/Component.ts
+++ b/packages/builders/src/components/Component.ts
@@ -1,5 +1,11 @@
 import type { JSONEncodable } from '../util/jsonEncodable';
-import type { APIBaseComponent, APIMessageComponent, ComponentType } from 'discord-api-types/v9';
+import type {
+	APIActionRowComponentTypes,
+	APIBaseComponent,
+	APIMessageComponent,
+	ComponentType,
+} from 'discord-api-types/v9';
+import type { Equatable } from '../util/equatable';
 
 /**
  * Represents a discord component
@@ -8,17 +14,15 @@ export abstract class Component<
 	DataType extends Partial<APIBaseComponent<ComponentType>> & {
 		type: ComponentType;
 	} = APIBaseComponent<ComponentType>,
-> implements JSONEncodable<APIMessageComponent>
+> implements JSONEncodable<APIMessageComponent>, Equatable<Component | APIActionRowComponentTypes>
 {
 	/**
 	 * The API data associated with this component
 	 */
 	protected readonly data: DataType;
 
-	/**
-	 * Converts this component to an API-compatible JSON object
-	 */
 	public abstract toJSON(): APIMessageComponent;
+	public abstract equals(other: Component | APIActionRowComponentTypes): boolean;
 
 	public constructor(data: DataType) {
 		this.data = data;
--- a/packages/builders/src/messages/embed/Assertions.ts
+++ b/packages/builders/src/messages/embed/Assertions.ts
@@ -25,12 +25,14 @@ export const authorNamePredicate = fieldNamePredicate.nullable();
 
 export const urlPredicate = z.string().url().nullish();
 
+export const RGBPredicate = z.number().int().gte(0).lte(255);
 export const colorPredicate = z
 	.number()
+	.int()
 	.gte(0)
 	.lte(0xffffff)
 	.nullable()
-	.or(z.tuple([z.number(), z.number(), z.number()]));
+	.or(z.tuple([RGBPredicate, RGBPredicate, RGBPredicate]));
 
 export const descriptionPredicate = z.string().min(1).max(4096).nullable();
 
--- a/packages/builders/src/messages/embed/Assertions.ts
+++ b/packages/builders/src/messages/embed/Assertions.ts
@@ -25,7 +25,12 @@ export const authorNamePredicate = fieldNamePredicate.nullable();
 
 export const urlPredicate = z.string().url().nullish();
 
-export const colorPredicate = z.number().gte(0).lte(0xffffff).nullable();
+export const colorPredicate = z
+	.number()
+	.gte(0)
+	.lte(0xffffff)
+	.nullable()
+	.or(z.tuple([z.number(), z.number(), z.number()]));
 
 export const descriptionPredicate = z.string().min(1).max(4096).nullable();
 
--- a/packages/builders/src/messages/embed/UnsafeEmbed.ts
+++ b/packages/builders/src/messages/embed/UnsafeEmbed.ts
@@ -333,10 +333,10 @@ export class UnsafeEmbed implements Equatable<APIEmbed | UnsafeEmbed> {
 	}
 
 	public equals(other: UnsafeEmbed | APIEmbed) {
-		if (other instanceof UnsafeEmbed) {
-			return isEqual(other.data, this.data);
-		}
-		return isEqual(other, this.data);
+		const { image: thisImage, thumbnail: thisThumbnail, ...thisData } = this.data;
+		const data = other instanceof UnsafeEmbed ? other.data : other;
+		const { image, thumbnail, ...otherData } = data;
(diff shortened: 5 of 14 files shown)


ImRodry:
Can you also add hexColor back and overwrite setColor to use Util.resolveColor?

packages/builders/src/components/Component.ts
```diff
@@ -8,17 +14,15 @@ export abstract class Component<
...
+> implements JSONEncodable<APIMessageComponent>, Equatable<Component | APIActionRowComponentTypes>
 {
 	/**
 	 * The API data associated with this component
 	 */
 	protected readonly data: DataType;
 
-	/**
-	 * Converts this component to an API-compatible JSON object
-	 */
```
ImRodry, line 20: this can stay, right?

packages/builders/src/components/ActionRow.ts
```diff
@@ -46,4 +47,11 @@ export class ActionRow<T extends ActionRowComponent = ActionRowComponent> extend
 			components: this.components.map((component) => component.toJSON()),
 		};
 	}
+
+	public equals(other: APIActionRowComponent<APIMessageComponent> | ActionRow) {
```
ImRodry, line 51: does this really belong in builders? Thought it'd be something for discord.js so we can compare with Embeds from there 🤔

vladfrangu: I don't see why not? Could be useful outside of d.js

packages/builders/src/util/equatable.ts
```diff
@@ -0,0 +1,3 @@
+export interface Equatable<T> {
```
vladfrangu, line 1: Whats this?

vladfrangu (changes requested):

packages/builders/src/components/ActionRow.ts
ImRodry: Yeah but my question is if you will be able to throw the Embed class from discord.js into this and have it work.
I also just realized halfway through writing this that this method wasn't added to Embed so, please? 🥺

packages/builders/src/util/equatable.ts
suneettipirneni: It's similar to JSONEncodable, it just signals that a builder supports `.equals`

packages/builders/src/components/Component.ts
suneettipirneni: I removed it because I realized the doc is inherited from `JSONEncodable` already

packages/builders/src/messages/embed/UnsafeEmbed.ts
```diff
@@ -315,6 +325,13 @@ export class UnsafeEmbed {
 		return { ...this.data };
 	}
 
+	public equals(other: UnsafeEmbed | APIEmbed) {
+		if (other instanceof UnsafeEmbed) {
+			return isEqual(other.data, this.data);
```
imranbarbhuiya, line 330: This will check for hight, weight and all in thumbnail, image which was removed in v13.3 iirc

packages/builders/src/util/equatable.ts
```diff
@@ -0,0 +1,14 @@
...
+	 * Whether or not this is equal to another structure
+	 */
+	equals: (other: T) => boolean;
+}
+
+/**
+ * Indicates if an object is equatable or not.
+ * @param maybeEquatable The object to check against
+ */
+export function isEquatable(maybeEquatable: unknown): maybeEquatable is Equatable<unknown> {
```
imranbarbhuiya, line 12:
```suggestion
export function isEquatable<T>(maybeEquatable: T): maybeEquatable is Equatable<T> {
```

Or when you'll use isEqualable it'll become unknown.

imranbarbhuiya (changes requested):

packages/builders/src/util/equatable.ts
suneettipirneni: This is inaccurate because it doesn't take `API` types into account.

iCrawl:
This needs a rebase.

packages/builders/src/components/ActionRow.ts
vladfrangu: > Yeah but my question is if you will be able to throw the Embed class from discord.js

Technically speaking yes, since d.js's extends builders class, it'll work just fine.
image

packages/builders/src/messages/embed/UnsafeEmbed.ts
```diff
@@ -6,6 +6,10 @@ import type {
 	APIEmbedImage,
 	APIEmbedVideo,
 } from 'discord-api-types/v9';
+import type { Equatable } from '../../util/equatable';
+import isEqual from 'fast-deep-equal';
+
+export type RGBTuple = [red: number, blue: number, green: number];
```
vladfrangu, line 12:
```suggestion
export type RGBTuple = [red: number, green: number, blue: number];
```

vladfrangu (dismissed):

packages/builders/src/messages/embed/UnsafeEmbed.ts
ImRodry: Embeds aren't people! It's checking for the height and width of the images, along with the proxy URLs

packages/builders/src/messages/embed/UnsafeEmbed.ts
```diff
@@ -164,6 +168,14 @@ export class UnsafeEmbed {
 		);
 	}
 
+	/**
+	 * The hex color of the current color of the embed
+	 */
+	public get hexColor() {
+		if (!this.data.color) return undefined;
+		return `#${this.data.color}` as const;
```
ImRodry, line 176: huh? That's not how you turn a base-10 into hex...
```suggestion
		return typeof this.data.color === 'number' ? `#${this.data.color.toString(16).padStart(6, '0')}` : null;
```

KhafraDev:
I haven't tested much but wouldn't something to this have be equal?

```js
import { deepStrictEqual } from 'assert';

export const equals = (a, b) => {
    try {
        deepStrictEqual(a, b);
        return true;
    } catch {
        return false;
    }
}
```

It works with classes, nested objects (as far as 2-3 levels as I tested) and it's simple enough. 🤷

kyranet:
Feels like using the wrong tool to do the right thing. Also, the fact it throws an error (and thus needing a try/catch) has severe performance implications.

KhafraDev:
I wouldn't necessarily call it the wrong tool, from its description: `Tests for deep equality between the actual and expected parameters. "Deep" equality means that the enumerable "own" properties of child objects are recursively evaluated also by the following rules.`

For performance, I doubt that using `try/catch` is going to be a bottleneck. It might be slower, but *slower* does not mean slow.

suneettipirneni:
> I wouldn't necessarily call it the wrong tool, from its description: `Tests for deep equality between the actual and expected parameters. "Deep" equality means that the enumerable "own" properties of child objects are recursively evaluated also by the following rules.`
> 
> For performance, I doubt that using `try/catch` is going to be a bottleneck. It might be slower, but _slower_ does not mean slow.

If you were going to give a built-in example node's `util.isDeepStrictEqual` is probably better alternative than `assert.isDeepStrictEqual`.

But going back to performance assert.isDeepEquals is one of slowest implementations of deep equality. According [fast-deep-equal's benchmarks](https://www.npmjs.com/package/fast-deep-equal) it's about 6.5x slower than `fast-deep-equals`

KhafraDev:
> If you were going to give a built-in example node's util.isDeepStrictEqual is probably better alternative than assert.isDeepStrictEqual.

I didn't know this existed but this seems like a good alternative.

almeidx:
I never saw the point in the `.equals()` methods, why are they being added back?

They're not being used internally, and now there's going to be a brand new dependency on the package just for them. Why can't the users that will be using the method (if any) install the dependency and compare the objects themselves?

monbrey:
> They're not being used internally, and now there's going to be a brand new dependency on the package just for them. Why can't the users that will be using the method (if any) install the dependency and compare the objects themselves?

I agree with the premise that users could do this themselves, but dependencies in a package are not a bad thing.

ImRodry:
> I never saw the point in the `.equals()` methods, why are they being added back?
> 
> They're not being used internally, and now there's going to be a brand new dependency on the package just for them. Why can't the users that will be using the method (if any) install the dependency and compare the objects themselves?

If you never saw the point that’s your problem but you’re not the only person using the package. There’s plenty of use cases and I initially PR’d it so you can assume at least I do + if there’s no replacement and no good reason to remove it it shouldn’t be removed. I agree that the dependency is extremely necessary snd the old implementation should be brought back but apparently I’m alone in that

almeidx:
> dependencies in a package are not a bad thing.

I wasn't trying to say that dependencies are a bad thing. What I was trying to say is that, in this case, the implementation for the method is so simple (using an external package; as simple as just calling a function) that I don't see the need for having it in this package

suneettipirneni:
> > dependencies in a package are not a bad thing.
> 
> I wasn't trying to say that dependencies are a bad thing. What I was trying to say is that, in this case, the implementation for the method is so simple (using an external package; as simple as just calling a function) that I don't see the need for having it in this package

For one, if you're using typescript `data` isn't publicly exposed so it doesn't get mutated externally. Also in some cases like `Embed`, `ActionRow`, and `SelectMenu` there are special cases you need to handle when doing comparisons. For example `SelectMenu` stores options in a separate `options` array rather than directly on the data itself. `Embed` would do equality based on image height and proxy url by default, which are props that will never be equal when compared to a received Embed from the API.

suneettipirneni:
> For one, if you're using typescript data isn't publicly exposed so it doesn't get mutated externally.

This being said, yes you could extend the class to access `data` if need be.

packages/builders/src/components/Component.ts
```diff
@@ -8,17 +14,15 @@ export abstract class Component<
...
 	public abstract toJSON(): APIMessageComponent;
+	public abstract equals(other: Component | APIActionRowComponentTypes): boolean;
```
IRONM00N, line 25:
```suggestion

	public abstract equals(other: Component | APIActionRowComponentTypes): boolean;
```
everything else gets a line in between

KhafraDev:
> For one, if you're using typescript data isn't publicly exposed so it doesn't get mutated externally.

I think this is more so an issue with the semantics. According to [kyra's post](https://github.com/discordjs/discord.js/issues/6567), the raw data should be available to users ("[t]he raw data must be exposed so users can access it"). By allowing access to the data would also forgo any need for future PRs to add utility functions as well.

```ts
export const embedEquals = (a: WhateverData, b: WhateverData) => {
    return util.deepStrictEqual(a.data, b.data);
}

export const embedHex = (a: Embed) => {
    return `#${a.data.color.toString(16).padStart(6, '0')}`;
}
```

suneettipirneni:
> p.s.: I'm fine with this change if `util.deepStrictEqual` is used in either case. Otherwise, almeidx's arguments still apply and I do agree with them.

`util.deepStrictEqual` is extremely slow compared to the `fast-deep-equals` package, I'm not sure the mantra of "avoid extra dependencies" is enough to warrant not using it. This is especially true when the package is [less than 1 kb](https://bundlephobia.com/package/fast-deep-equal@3.1.3), and looking at the package code, the implementation is anything but trivial.

Once again the argument that it should be handled by the user is a rationale that could be applied to any utility method in the library. That being said, yes there are situations where the library isn't responsible for functionality that doesn't fit in the scope of library. But I don't think adding an `.equals` method is even close to that. If we're responsible for dictating the internal values we should also be responsible for deciding equality between those values. This is especially true considering that other non-builder discord.js entities also have equals() methods.

KhafraDev:
Slower does not mean slow. At 46,000 ops a second (fast deep equal benchmark) I would say that it’s well past fast enough. I would be surprised if equals has been used 46k times since it’s inception (half joking lol).

suneettipirneni:
> Slower does not mean slow. At 46,000 ops a second (fast deep equal benchmark) I would say that it’s well past fast enough. I would be surprised if equals has been used 46k times since it’s inception (half joking lol).

It's still bottom of the barrel when it comes to performance if I can get better performance at the cost of 1kb I'll take that instead. In terms of usage it's hard to tell how many times a builder method is used because it could be invoked only once or be invoked every time a command is run. Furthermore you can have situations where you need to run a find() on an array that would use .equals for every element. It's a very primitive base method that can be applied in a bunch of different areas. Regardless of the use cases it would be nice if the built in solution can scale well if needed.

KhafraDev:
> Regardless of the use cases it would be nice if the built in solution can scale well if needed.

46k ops/s does scale well. As said earlier, maybe not as fast as other dependencies, but it is fast enough for 100% of use cases. If someone using it needs the best performance humanly possible (spoiler: no one does), then they can override the equals method by extending the class and implement it using whatever dependency they want.

> it could be invoked only once or be invoked every time a command is run.

It still wouldn't be the bottleneck. I was also joking that the only person I've ever seen need this method is Rodry, and even with it being gone entirely from v14, he's the only person who has wanted it to be re-implemented.

> It's a very primitive base method that can be applied in a bunch of different areas.

Have any real life use cases?

suneettipirneni:
> 46k ops/s does scale well. As said earlier, maybe not as fast as other dependencies, but it is fast enough for 100% of use cases. If someone using it needs the best performance humanly possible (spoiler: no one does), then they can override the equals method by extending the class and implement it using whatever dependency they want.

Sure, but it'd be nice if the library included a common and highly used external library so performance isn't always a concern for bots that want to scale.

> I was also joking that the only person I've ever seen need this method is Rodry, and even with it being gone entirely from v14, he's the only person who has wanted it to be re-implemented.

This is a not a very good point. I we want to get into semantics I can list at least four people who wanted it restored (rodry,IronM00n imranbarcaria and smidul). But still these numbers prove absolutely nothing, and there's no way they can extrapolated to the already small percentage of users using v14. User request numbers are pretty much meaningless.

> Have any real life use cases?

Here's some real world support messages that needed them:
https://discord.com/channels/222078108977594368/447463323135508480/930914561350049842

https://discord.com/channels/222078108977594368/824411059443204127/914678861755744386

https://discord.com/channels/222078108977594368/824411059443204127/918374538457280512

In addition, one thing that wasn't mentioned is tracking differences between messages like on the message update event. You'd have to do a value comparison for tracking those differences.

SpaceEEC (approved):

packages/builders/src/messages/embed/UnsafeEmbed.ts
```diff
@@ -164,6 +168,14 @@ export class UnsafeEmbed {
 		);
 	}
 
+	/**
+	 * The hex color of the current color of the embed
+	 */
+	public get hexColor() {
+		if (!this.data.color) return undefined;
+		return typeof this.data.color === 'number' ? `#${this.data.color.toString(16).padStart(6, '0')}` : null;
```
kyranet, line 176:
```suggestion
		return typeof this.data.color === 'number' ? `#${this.data.color.toString(16).padStart(6, '0')}` : this.data.color;
```

kyranet (approved):

kyranet (changes requested):
Sorry, wrong button. ^

imranbarbhuiya (changes requested):

packages/builders/src/messages/embed/Assertions.ts
```diff
@@ -25,7 +25,12 @@ export const authorNamePredicate = fieldNamePredicate.nullable();
 
 export const urlPredicate = z.string().url().nullish();
 
-export const colorPredicate = z.number().gte(0).lte(0xffffff).nullable();
+export const colorPredicate = z
+	.number()
+	.gte(0)
+	.lte(0xffffff)
+	.nullable()
+	.or(z.tuple([z.number(), z.number(), z.number()]));
```
almeidx, line 33: Perhaps you should add more validation to the RGB values too? (e.g. int, min, max, etc.)

packages/builders/__tests__/messages/embed.test.ts
```diff
@@ -134,6 +132,12 @@ describe('Embed', () => {
 			// @ts-expect-error
 			expect(() => embed.setColor('RED')).toThrowError();
 		});
+
+		test('GIVEN an embed with a valid color THEN does not throw error', () => {
```
kyranet, line 136: It should test for throwing errors, we already test for not throwing in lines 117-120.

kyranet (changes requested):

vladfrangu (approved):

kyranet (approved):

suneettipirneni:
Don't merge yet, need to adjust one thing in action row

PR #7490, opened 2022-02-17
refactor(`Message`): use raw data storage for messages

suneettipirneni:
**Please describe the changes this PR makes and why it should be merged:**

Caches applicable message data within an internal `data` field. Makes props of the data getters in the class instead of storing them.

**Status and versioning classification:**
- Code changes have been tested against the Discord API, or there are no code changes
- I know how to update typings and have done so, or typings don't need updating
- This PR changes the library's interface (methods or parameters added)

PR #7495, opened 2022-02-18
refactor(`Interaction`): use raw data storage for interactions

imranbarbhuiya:
followup of #7490 for interactions.

### DIFF
--- a/package.json
+++ b/package.json
@@ -44,7 +44,7 @@
 		"conventional-changelog-cli": "^2.2.2",
 		"husky": "^7.0.4",
 		"prettier": "^2.5.1",
-		"turbo": "1.1.2"
+		"turbo": "1.1.4"
 	},
 	"engines": {
 		"node": ">=16.9.0"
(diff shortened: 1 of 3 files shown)


suneettipirneni (changes requested):
`data` fields that aren't a collection or store classful objects shouldn't have each of its fields be assigned directly in the constructor. The object itself should be assigned directly to the internal `data` field. In addition, no new properties should be added to the `data` field such as `command_id` and `command_name`.

imranbarbhuiya:
> `data` fields that aren't a collection or store classful objects shouldn't have each of its fields be assigned directly in the constructor. The object itself should be assigned directly to the internal `data` field.

I've assigned it as `data.data` or should I assign all properties to `data` 
```js
this.data = {...this.data, data.data}
```

something like this?

suneettipirneni:
> > `data` fields that aren't a collection or store classful objects shouldn't have each of its fields be assigned directly in the constructor. The object itself should be assigned directly to the internal `data` field.
> 
> I've assigned it as `data.data` or should I assign all properties to `data`
> 
> ```js
> this.data = {...this.data, data.data}
> ```
> 
> something like this?

yeah

PR #7499, opened 2022-02-18
refactor(message): make constructor protected

SuperchupuDev:
Ideally this should be merged after #7490 gets merged

This PR makes the constructor of `Message` protected, making it able to create extended classes out of it in a way that doesn't break the library. Since classes like `Interaction` have a protected constructor already, I don't see why `Message` should be kept as a private class.

This can be useful for safely extending the message class when the raw data is stored in `Message#data` (when #7490 gets merged), allowing you to do `new ExtendedMessage(client, message.data)` so you can add your own methods and properties.

### DIFF
--- a/packages/discord.js/typings/index.d.ts
+++ b/packages/discord.js/typings/index.d.ts
@@ -1501,7 +1501,7 @@ export interface MappedInteractionTypes<Cached extends boolean = boolean> {
 
 export class Message<Cached extends boolean = boolean> extends Base {
   private readonly _cacheType: Cached;
-  private constructor(client: Client, data: RawMessageData);
+  protected constructor(client: Client, data: RawMessageData);
   private _patch(data: RawPartialMessageData | RawMessageData): void;
 
   public activity: MessageActivity | null;


ImRodry:
> Since classes like Interaction have a protected constructor already, I don't see why Message should be kept as a private class.

The Interaction class is protected because it's extended by other classes within the library, which are private themselves and not protected IIRC. I don't see what other properties you'd add to a message or why you'd need to use the constructor at all when you can use a MessageOptions object or something similar

SuperchupuDev:
> > Since classes like Interaction have a protected constructor already, I don't see why Message should be kept as a private class.
> 
> The Interaction class is protected because it's extended by other classes within the library, which are private themselves and not protected IIRC. I don't see what other properties you'd add to a message or why you'd need to use the constructor at all when you can use a MessageOptions object or something similar

In my usecase it can be useful for adding custom methods like localeReply (`message.localeReply('ERROR')` for example). It would make everything easier and cleaner by making it a method instead of manually calling the translate function each time. I had no idea the other interaction classes were private, so I'd argue it would also make sense for them to be protected too (with the same usecase i explained above)

vladfrangu:
1. Consistency, Message shouldn't be the only class getting this treatment. A huge chunk of classes, that don't have further extensions, have private constructors
2. You really shouldn't extend our classes just to add a method... it's one of the reasons we completely removed Structures.extend (alongside the plethora of pitfalls it came with), and constructing a new class *just* to get that method in is quite more expensive than just doing what I suggested below. 

You should instead build a localeReply function in your code that takes in a context (be it a channel, a guild, a message, an interaction, my hopes and dreams, whatever floats your boat) and the message to translate and replies there 😄

SuperchupuDev:
Okay then, thanks 👍

PR #7500, opened 2022-02-18
docs: correctly type getters

IRONM00N:
- made all getters be typed correctly as getters - this improves typescript error messages and makes it clearer to the end-user that the value is calculated 
- remove `GuildChannel#calculatedPosition` from the typings since it doesn't exist in the source code
- corrected a few properties that should(n't) have been marked as readonly 

### DIFF
--- a/packages/discord.js/typings/index.d.ts
+++ b/packages/discord.js/typings/index.d.ts
@@ -46,7 +46,6 @@ import {
   APIPartialEmoji,
   APIPartialGuild,
   APIRole,
-  APIAttachment,
   APISelectMenuComponent,
   APITemplateSerializedSourceGuild,
   APIUser,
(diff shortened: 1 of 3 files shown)


kyranet (approved):
This was a task that I wanted to get done for a while, thanks! 🙏🏼

vladfrangu (approved):

packages/discord.js/typings/index.d.ts
```diff
@@ -2454,7 +2452,7 @@ export class Webhook extends WebhookMixin() {
   public avatar: string;
   public avatarURL(options?: ImageURLOptions): string | null;
   public channelId: Snowflake;
-  public client: Client;
+  public readonly client: Client;
```
ImRodry, line 2455: This isn't a getter but a property defined through Object.defineProperty, should it be documented this way or with the Readonly type?

IRONM00N: Well, it isn't `writable` but it also isn't a getter so it should be like this. I think Readonly would only be used with `Object.freeze`

kyranet: That's correct, `Object.defineProperty` by default assigns readonly properties.

iCrawl (approved):

PR #6050, opened 2021-07-05
refactor(Intents): remove computed shorts

almostSouji:
**Please describe the changes this PR makes and why it should be merged:**

- ⚠️ remove `Intents.ALL`
- ⚠️ remove `Intents.PRIVILEGED`
- ⚠️ remove `Intents.NON_PRIVILEGED`

We currently provide static shorthands on the Intents bitfield wrapper (inspired by Permissions when implementing). This raises multiple issues:

- End users are lead to just pass `Intents.ALL` no matter what they actually need for the bot to work, as evident from library support channels
- Discord might change the grouping of which intents are privileged and which are not at any time. They don't make any promise that this will stay at member presences and guild members. Removal accordingly improves maintainability, as we do not need to push releases to change groupings going forward.

**Status and versioning classification:**

- Code changes have been tested against the Discord API, or there are no code changes
- I know how to update typings and have done so, or typings don't need updating
- This PR includes breaking changes (methods removed or renamed, parameters moved or removed)

PR #7503, opened 2022-02-19
feat: add Partials class

ImRodry:

This PR transforms Partials into a class that extends BitField so that we can have properties like All, which adds all partials into an array, and other methods from the inherited class

### DIFF
--- a/packages/discord.js/typings/index.d.ts
+++ b/packages/discord.js/typings/index.d.ts
@@ -1912,6 +1912,11 @@ export class OAuth2Guild extends BaseGuild {
   public permissions: Readonly<PermissionsBitField>;
 }
 
+export class Partials extends BitField<keyof typeof PartialsFlags> {
+  public static Flags: typeof PartialsFlags;
+  public static All: PartialsFlags[];
+}
+
 export class PartialGroupDMChannel extends Channel {
   private constructor(client: Client, data: RawPartialGroupDMChannelData);
   public type: ChannelType.GroupDM;
@@ -3826,7 +3831,7 @@ export interface ClientOptions {
   shardCount?: number;
   makeCache?: CacheFactory;
   allowedMentions?: MessageMentionOptions;
-  partials?: Partials[];
+  partials?: PartialsFlags[];
   failIfNotExists?: boolean;
   presence?: PresenceData;
   intents: BitFieldResolvable<GatewayIntentsString, number>;
@@ -4861,7 +4866,7 @@ export interface PartialRoleData extends RoleData {
   id?: Snowflake | number;
 }
 
-export enum Partials {
+declare enum PartialsFlags {
   User,
   Channel,
   GuildMember,
--- a/packages/discord.js/src/client/actions/Action.js
+++ b/packages/discord.js/src/client/actions/Action.js
@@ -43,7 +43,7 @@ class GenericAction {
         },
         this.client.channels,
         id,
-        Partials.Channel,
+        Partials.Flags.Channel,
       )
     );
   }
@@ -60,7 +60,7 @@ class GenericAction {
         },
         channel.messages,
         id,
-        Partials.Message,
+        Partials.Flags.Message,
         cache,
       )
     );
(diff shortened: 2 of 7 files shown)


ItsJackson (commented):
There’s a error running the ESLint check, with code style issues in these file(s)

ItsJackson (approved):

vladfrangu:
And what exactly is the benefit from this? It feels like overcomplicating for no reason 🤔

ImRodry:
Quoting from the pr description:
> so that we can have properties like All, which adds all partials into an array, and other methods from the inherited class

almeidx:
> so that we can have properties like All, which adds all partials into an array

Is `Partials.All` something that is actually wanted though? Perhaps not, for the same reasoning as to why `Intents.ALL` was removed (#6050)

ImRodry:
> > so that we can have properties like All, which adds all partials into an array
> 
> Is `Partials.All` something that is actually wanted though? Perhaps not, for the same reasoning as to why `Intents.ALL` was removed (#6050)

Intents.All wasn't wanted because people are not supposed to have all intents and they should specify the ones they want individually. The same doesn't apply to Partials, where you can have all partials enabled with no issues

vladfrangu:
You can have all intents enabled too and you won't have issues, so I don't get why that matters 👀

ImRodry:
That's not what I meant. You're not supposed to have all intents enabled because you often won't need all of them, but you may need all partials and there's nothing wrong with having them all enabled

xHyroM:
I'm leaning towards deleting Partials.All. It's useless, you rarely need all partials.

almeidx:
> and there's nothing wrong with having them all enabled

I disagree. Let's say a new partial is added and your bot is dependent on events/caches of that structure. It now might receive partial structures and you very likely will need to adapt your code if you were accessing data that can be missing (which could not be missing previously).

And besides that, `Partials.All` doesn't actually tell you *what* partials are enabled, and you'd likely need to go to the documentation to find out what partials exist. As opposed to passing all the partials that you need, which is more verbose and doesn't require an external service (e.g. the documentation) to know what is it you're enabling (as is the case for the intents)

ImRodry:
I assume you’d only enable all partials if you knew what partials there were to begin with. Not only that but new partials are only ever added when new structures are added (for example when scheduled events came out) so you wouldn’t need to adapt your code since you wouldn’t have event listeners for something that just came out

monbrey:
I feel like this change is a little counter-intuitive and doesn't really deliver any benefit over the base enum/constant.

Partials aren't (or weren't) a calculated Bitfield and I don't see the need for them to be. Is this all just to have `Partials.All` added?

ImRodry:
Mainly yes but if we want to move away from strings and partials are now treated as numbers and not strings, it would make sense to treat Partials as a Bitfield IMO

vladfrangu:
So, in a nutshell, this PR...

1. Adds a Partials.All which is not even checked by the rest of the flags (like ADMINISTRATOR is checked for permissions), because adding all the 4 partials we have that you need to an array is..annoying?
2. Changes an enum of Strings -> numbers to a bitfield class with a Flags property (which we moved away from); the enum was consistent with the general move in v14 from strings to numbers as is, and there were no more magic numbers/strings
3. Goes against intents where you're supposed to only use the ones you need

I don't know what the other maintainers think, but to me this sounds like bloat and overcomplicating an already complex concept. Do note that this is *my personal* opinion, and now saying this in the name of the entire team. But I don't see any real benefit from this 😅

ImRodry:
So let me address each one of those points individually
1. Partials.All being an array of numbers is correct (this also made me realize there was a bug which I just fixed) because when you access an enum property it's returned as a number, so all properties of that enum would be an array of all the numbers. Also you have 6, not 4 partial types, and most of the time you do want all of them to make sure you get all the data from events. This would also not be an issue when a new partial type is added because, as mentioned previously, this only happens when a new structure is added, so you'd need to add event listeners for that anyway
2. The flags property is literally the exact same as the current Flags enum, I just copied the code over so it's the exact same
3. I don't understand the comparison with intents because there is an actual downside to using all intents, whereas the same doesn't apply to partials, as mentioned previously

packages/discord.js/typings/index.d.ts
```diff
@@ -4788,7 +4793,7 @@ export interface PartialRoleData extends RoleData {
   id?: Snowflake | number;
 }
 
-export enum Partials {
+export enum PartialsFlags {
```
almeidx, line 4796: Doesn't look like this is exported

ImRodry: I’ll just not export this then

packages/discord.js/typings/index.d.ts
```diff
@@ -1705,6 +1705,11 @@ export class OAuth2Guild extends BaseGuild {
   public permissions: Readonly<PermissionsBitField>;
 }
 
+export class Partials extends BitField<keyof typeof PartialsFlags> {
+  public static Flags: typeof PartialsFlags;
+  public static All: number[];
```
almeidx, line 1710:
```suggestion
  public static All: PartialsFlags[];
```

almeidx (changes requested):

ImRodry:
@iCrawl would appreciate an explanation on why this was closed

iCrawl:
Explanations were given above.

Issue #7492, opened 2022-02-18
rest clears token on 401s


advaith1:
### Which package is this bug report for?

rest

### Issue description

djs/rest clears the stored token whenever it gets a 401. however, this is causing issues in production, since Discord returns 401s for issues other than an invalid bot token (such as an invalid webhook token).

djs/rest seems to even do this when auth is set to false.

https://github.com/discordjs/discord.js/blob/298b22604b6e0ad0f19c2fb59513a60c682f4f17/packages/rest/src/lib/handlers/SequentialHandler.ts#L480-L482

Vlad told me to make a gh issue

### Package version

@discordjs/rest@0.3.0

### Node.js version

16.14.0

### Priority this issue should have

Medium (should be fixed soon)

ckohen:
FYI, I looked into things that 401 with valid tokens before we migrated to /rest because I noticed it could be a problem. I went through a bunch of endpoints and made an [issue](https://github.com/discord/discord-api-docs/issues/3945) with the only one I found, didn't try webhooks though. That's gonna be a problem. We probably want to just check if auth is set on the request before calling that

PR #7508, opened 2022-02-19
fix(RequestHandler): only reset tokens for authenticated 401s

vladfrangu:

Closes https://github.com/discordjs/discord.js/issues/7492 by resetting the token only when an authorized request fails with a 401.

### DIFF
--- a/packages/rest/src/lib/REST.ts
+++ b/packages/rest/src/lib/REST.ts
@@ -160,7 +160,7 @@ export interface APIRequest {
 	/**
 	 * The data that was used to form the body of this request
 	 */
-	data: Pick<InternalRequest, 'files' | 'body'>;
+	data: Pick<InternalRequest, 'files' | 'body' | 'auth'>;
 	/**
 	 * The number of times this request has been attempted
 	 */
--- a/packages/rest/src/lib/REST.ts
+++ b/packages/rest/src/lib/REST.ts
@@ -1,6 +1,13 @@
 import { EventEmitter } from 'node:events';
 import { CDN } from './CDN';
-import { InternalRequest, RequestData, RequestManager, RequestMethod, RouteLike } from './RequestManager';
+import {
+	HandlerRequestData,
+	InternalRequest,
+	RequestData,
+	RequestManager,
+	RequestMethod,
+	RouteLike,
+} from './RequestManager';
 import { DefaultRestOptions, RESTEvents } from './utils/constants';
 import type { AgentOptions } from 'node:https';
 import type { RequestInit, Response } from 'node-fetch';
@@ -160,7 +167,7 @@ export interface APIRequest {
 	/**
 	 * The data that was used to form the body of this request
 	 */
-	data: Pick<InternalRequest, 'files' | 'body' | 'auth'>;
+	data: HandlerRequestData;
 	/**
 	 * The number of times this request has been attempted
 	 */
--- a/packages/rest/src/lib/RequestManager.ts
+++ b/packages/rest/src/lib/RequestManager.ts
@@ -290,10 +290,16 @@ export class RequestManager extends EventEmitter {
 			this.createHandler(hash.value, routeId.majorParameter);
 
 		// Resolve the request into usable fetch/node-fetch options
-		const { url, fetchOptions } = this.resolveRequest(request);
+		const { url, fetchOptions, isAuthenticated } = this.resolveRequest(request);
 
 		// Queue the request
-		return handler.queueRequest(routeId, url, fetchOptions, { body: request.body, files: request.files });
+		return handler.queueRequest(
+			routeId,
+			url,
+			fetchOptions,
+			{ body: request.body, files: request.files },
+			isAuthenticated,
+		);
 	}
 
 	/**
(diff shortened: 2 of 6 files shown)


ckohen:
Instead of passing around an extra property everywhere, can you not just check for the Authorization Header?

I know a user can *technically* override it when passing auth: false, but that's not supported

vladfrangu:
We'd have to normalize the headers to always be lowercased no matter if we or users provided it I suppose... But of a pitfall of how we pass data down from the manager to the handler

ckohen:
What do you mean by that? the header is passed as Authorization, not authorization

packages/rest/src/lib/RequestManager.ts
```diff
@@ -290,10 +290,14 @@ export class RequestManager extends EventEmitter {
...
 		// Resolve the request into usable fetch/node-fetch options
-		const { url, fetchOptions } = this.resolveRequest(request);
+		const { url, fetchOptions, isAuthenticated } = this.resolveRequest(request);
 
 		// Queue the request
-		return handler.queueRequest(routeId, url, fetchOptions, { body: request.body, files: request.files });
+		return handler.queueRequest(routeId, url, fetchOptions, {
+			body: request.body,
+			files: request.files,
+			auth: isAuthenticated,
```
ckohen, line 299: We have direct access to the original request here.

```suggestion
			auth: request.auth !== false,
```

packages/rest/src/lib/handlers/SequentialHandler.ts
```diff
@@ -218,14 +219,14 @@ export class SequentialHandler {
 	 * @param routeId The generalized api route with literal ids for major parameters
 	 * @param url The fully resolved url to make the request to
 	 * @param options The node-fetch options needed to make the request
-	 * @param bodyData The data that was used to form the body, passed to any errors generated
+	 * @param requestData The data that was used to form the body, passed to any errors generated
```
ckohen, line 222:
```suggestion
	 * @param requestData Extra data from the users request needed for errors and additional processing
```

packages/rest/src/lib/RequestManager.ts
```diff
@@ -290,10 +290,14 @@ export class RequestManager extends EventEmitter {
 			this.createHandler(hash.value, routeId.majorParameter);
 
 		// Resolve the request into usable fetch/node-fetch options
-		const { url, fetchOptions } = this.resolveRequest(request);
+		const { url, fetchOptions, isAuthenticated } = this.resolveRequest(request);
```
ckohen, line 293:
```suggestion
		const { url, fetchOptions } = this.resolveRequest(request);
```

packages/rest/src/lib/RequestManager.ts
```diff
@@ -315,7 +319,11 @@ export class RequestManager extends EventEmitter {
 	 * Formats the request data to a usable format for fetch
 	 * @param request The request data
 	 */
-	private resolveRequest(request: InternalRequest): { url: string; fetchOptions: RequestInit } {
+	private resolveRequest(request: InternalRequest): {
+		url: string;
+		fetchOptions: RequestInit;
+		isAuthenticated: boolean;
+	} {
```
ckohen, line 326:
```suggestion
private resolveRequest(request: InternalRequest): { url: string; fetchOptions: RequestInit } {
```

packages/rest/src/lib/RequestManager.ts
```diff
@@ -406,7 +414,7 @@ export class RequestManager extends EventEmitter {
 			method: request.method,
 		};
 
-		return { url, fetchOptions };
+		return { url, fetchOptions, isAuthenticated: request.auth !== false };
```
ckohen, line 417:
```suggestion
		return { url, fetchOptions };
```

ckohen (changes requested):

packages/rest/src/lib/REST.ts
```diff
@@ -160,7 +160,7 @@ export interface APIRequest {
 	/**
 	 * The data that was used to form the body of this request
 	 */
-	data: Pick<InternalRequest, 'files' | 'body'>;
+	data: Pick<InternalRequest, 'files' | 'body' | 'auth'>;
```
kyranet, line 163: We need to make this a type, it's referenced in a few places already.

kyranet (changes requested):

SpaceEEC (approved):

packages/rest/src/lib/RequestManager.ts
```diff
@@ -293,7 +293,11 @@ export class RequestManager extends EventEmitter {
 		const { url, fetchOptions } = this.resolveRequest(request);
 
 		// Queue the request
-		return handler.queueRequest(routeId, url, fetchOptions, { body: request.body, files: request.files });
+		return handler.queueRequest(routeId, url, fetchOptions, {
+			body: request.body,
+			files: request.files,
+			auth: request.auth !== false,
```
nearlySplat, line 299: Couldn't this be `!!request.auth`?

almeidx: If anything, it would be `!request.auth`. I guess there's a reason why vlad didn't make it that from the start

vladfrangu: Nop, and here's why
image
auth should be set to `false` only if it was provided as `false`

iCrawl (approved):

PR #7511, opened 2022-02-19
refactor: use `Readonly` types in builders where applicable

suneettipirneni:

Changes return types and some parameter types to use readonly inputs/outputs.

### DIFF
--- a/packages/builders/src/components/ActionRow.ts
+++ b/packages/builders/src/components/ActionRow.ts
@@ -36,7 +36,7 @@ export class ActionRow<T extends ActionRowComponent = ActionRowComponent> extend
 	 * Sets the components in this action row
 	 * @param components The components to set this row to
 	 */
-	public setComponents(...components: ReadonlyArray<Readonly<T>>) {
+	public setComponents(...components: readonly Readonly<T>[]) {
 		this.components.splice(0, this.components.length, ...components);
 		return this;
 	}
--- a/packages/builders/src/components/ActionRow.ts
+++ b/packages/builders/src/components/ActionRow.ts
@@ -17,7 +17,7 @@ export class ActionRow<T extends ActionRowComponent = ActionRowComponent> extend
 > {
 	public readonly components: T[];
 
-	public constructor({ components, ...data }: Partial<APIActionRowComponent<APIMessageComponent>> = {}) {
+	public constructor({ components, ...data }: Readonly<Partial<APIActionRowComponent<APIMessageComponent>>> = {}) {
 		super({ type: ComponentType.ActionRow, ...data });
 		this.components = (components?.map((c) => createComponent(c)) ?? []) as T[];
 	}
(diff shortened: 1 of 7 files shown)


almeidx:
Would `readonly T[]` syntax be preferred over `ReadonlyArray<T>` given that `T[]` is already being used instead of `Array<T>`?

suneettipirneni:
> Would `readonly T[]` syntax be preferred over `ReadonlyArray<T>` given that `T[]` is already being used instead of `Array<T>`?

Pretty sure `readonly T[]` isn't a valid syntax.

almeidx:
> Pretty sure `readonly T[]` isn't a valid syntax.

It is ([handbook](https://www.typescriptlang.org/docs/handbook/2/objects.html#the-readonlyarray-type))

suneettipirneni:
> > Pretty sure `readonly T[]` isn't a valid syntax.
> 
> It is ([handbook](https://www.typescriptlang.org/docs/handbook/2/objects.html#:~:text=Just%20as%20TypeScript%20provides%20a%20shorthand%20syntax%20for%20Array%3CType%3E%20with%20Type%5B%5D%2C%20it%20also%20provides%20a%20shorthand%20syntax%20for%20ReadonlyArray%3CType%3E%20with%20readonly%20Type%5B%5D.))

Ah ok I'll adjust them then.

iCrawl (approved):

packages/builders/src/components/ActionRow.ts
```diff
@@ -27,7 +27,7 @@ export class ActionRow<T extends ActionRowComponent = ActionRowComponent> extend
 	 * @param components The components to add to this action row.
 	 * @returns
 	 */
-	public addComponents(...components: T[]) {
+	public addComponents(...components: ReadonlyArray<Readonly<T>>) {
```
kyranet, line 30:
```suggestion
	public addComponents(...components: readonly Readonly<T>[]) {
```

packages/builders/src/messages/embed/Assertions.ts
```diff
@@ -17,7 +17,7 @@ export const embedFieldsArrayPredicate = embedFieldPredicate.array();
 
 export const fieldLengthPredicate = z.number().lte(25);
 
-export function validateFieldLength(amountAdding: number, fields?: APIEmbedField[]): void {
+export function validateFieldLength(amountAdding: number, fields?: ReadonlyArray<Readonly<APIEmbedField>>): void {
```
kyranet, line 20:
```suggestion
export function validateFieldLength(amountAdding: number, fields?: readonly Readonly<APIEmbedField>[]): void {
```

packages/builders/src/messages/embed/UnsafeEmbed.ts
```diff
@@ -299,7 +300,7 @@ export class UnsafeEmbed implements Equatable<APIEmbed | UnsafeEmbed> {
 	 *
 	 * @param timestamp The timestamp or date
 	 */
-	public setTimestamp(timestamp: number | Date | null = Date.now()): this {
+	public setTimestamp(timestamp: number | Readonly<Date> | null = Date.now()): this {
```
kyranet, line 303: We really need [ReadonlyDate](https://github.com/microsoft/TypeScript/issues/29274), heh.

`Readonly<Date>` is a bit useless here, it doesn't have public fields, only methods.

kyranet (changes requested):

BenPatersonxx:
@suneettipirneni

PR #7520, opened 2022-02-20
feat(message): add `reason` on pin and unpin

### LINKED ISSUE #7519: Support audit log reasons for pinning and unpinning messages

discord.js

### Feature

Add a `reason: string` parameter for `Message(Manager)#(un)pin`. #4869 removed support for including an audit log reason when pinning and unpinning messages, citing that it isn’t supported, but from my testing it is.

screenshot of pinning message audit log entry with reason

`X-Audit-Log-Reason` is supported for pinning messages and the reason appears in the client. For unpinning messages, the reason doesn’t appear in the client but is returned by the API in `/guilds/{guild.id}/audit-logs`.

Add a `reason: string` parameter for `Message(Manager)#(un)pin`.

_No response_

### Other context

Related: #4359

almeidx:
Closes #7519

### DIFF
--- a/packages/discord.js/src/structures/Message.js
+++ b/packages/discord.js/src/structures/Message.js
@@ -681,6 +681,7 @@ class Message extends Base {
 
   /**
    * Pins this message to the channel's pinned messages.
+   * @param {string} [reason] Reason for pinning
    * @returns {Promise<Message>}
    * @example
    * // Pin a message
@@ -688,14 +689,15 @@ class Message extends Base {
    *   .then(console.log)
    *   .catch(console.error)
    */
-  async pin() {
+  async pin(reason) {
     if (!this.channel) throw new Error('CHANNEL_NOT_CACHED');
-    await this.channel.messages.pin(this.id);
+    await this.channel.messages.pin(this.id, reason);
     return this;
   }
 
   /**
    * Unpins this message from the channel's pinned messages.
+   * @param {string} [reason] Reason for unpinning
    * @returns {Promise<Message>}
    * @example
    * // Unpin a message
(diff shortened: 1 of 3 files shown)


packages/discord.js/src/structures/Message.js
```diff
@@ -681,31 +681,33 @@ class Message extends Base {
 
   /**
    * Pins this message to the channel's pinned messages.
+   * @param {string} [reason] Reason for pinning
    * @returns {Promise<Message>}
    * @example
    * // Pin a message
    * message.pin()
```
ImRodry, line 688: I would add a reason to the example, same below

packages/discord.js/src/structures/Message.js
```diff
@@ -681,31 +681,33 @@ class Message extends Base {
...
     return this;
   }
 
   /**
    * Unpins this message from the channel's pinned messages.
+   * @param {string} [reason] Reason for unpinning
    * @returns {Promise<Message>}
    * @example
    * // Unpin a message
    * message.unpin()
```
FnrDev, line 704:
```suggestion
   * message.unpin("needed to go")
```

packages/discord.js/src/structures/Message.js
FnrDev, line 688:
```suggestion
   * message.pin("Important message for new members")
```

vladfrangu (approved):

packages/discord.js/src/structures/Message.js
kyranet: If that's to be done, it should be a separate example, otherwise people will understand it's required when it's optional.

SpaceEEC (approved):

iCrawl (approved):

PR #7395, opened 2022-02-03
refactor(builders-methods): make methods consistent

BenjammingKirby:
**Please describe the changes this PR makes and why it should be merged:**

This makes builders methods such as `<Embed>addFields` and `<SlashCommand..Option>.addChoice`, `<ActionRow>.setComponents`, `<SlashCommand..Option>.setChoices` and `<Embed>.setFields` consistent 

now `<SlashCommand..Option>`'s `.setChoices`, `.addChoices`, `.addChoice` **all take in Objects** like the embed methods instead of a tuple array

`<SlashCommand..Option>`'s `.setChoices` and `addChoices` as well as `<ActionRow>.setComponents` all now take a *rest parameter* instead of an array of their respective inputs to be more consistent with methods such as `<Embed>.setFields`

**Status and versioning classification:**
- Code changes have been tested against the Discord API, or there are no code changes
- I know how to update typings and have done so, or typings don't need updating
- This PR includes breaking changes (methods removed or renamed, parameters moved or removed)

PR #7522, opened 2022-02-21
refactor(Embed): remove add field

imranbarbhuiya:
This PR removes `Embed#addField` in favour of `Embed#addFields` because you are just saving a `s` when using `addField`.

For those who don't wanna replace all `addField` manually can try this regex replace. You can ofc make a better one but using these 2 also works fine. Idk how can I use condition in replace string else it can be simplified to one regex using `?` after inline.

```js
// 1. use this one first
// regex
.addField\(((.|[\n])*?),((.|[\n])*?),(.*true)\)
//replace with
.addFileds({name: $1, value: $3, inline: $5})
```

```js
// 2. use this one after replacing all inline fields
//regex
.addField\(((.|[\n])*?),((.|[\n])*?)\)
//replace with
.addFileds({name: $1, value: $3})
```

cross verify before replacing.

### DIFF
--- a/packages/builders/__tests__/messages/embed.test.ts
+++ b/packages/builders/__tests__/messages/embed.test.ts
@@ -318,15 +318,6 @@ describe('Embed', () => {
 			});
 		});
 
-		test('GIVEN an embed using Embed#addField THEN returns valid toJSON data', () => {
-			const embed = new Embed();
-			embed.addField({ name: 'foo', value: 'bar' });
-
-			expect(embed.toJSON()).toStrictEqual({
-				fields: [{ name: 'foo', value: 'bar' }],
-			});
-		});
-
 		test('GIVEN an embed using Embed#addFields THEN returns valid toJSON data', () => {
 			const embed = new Embed();
 			embed.addFields({ name: 'foo', value: 'bar' });
--- a/packages/builders/src/messages/embed/UnsafeEmbed.ts
+++ b/packages/builders/src/messages/embed/UnsafeEmbed.ts
@@ -176,15 +176,6 @@ export class UnsafeEmbed implements Equatable<APIEmbed | UnsafeEmbed> {
 	}
 
 	/**
-	 * Adds a field to the embed (max 25)
-	 *
-	 * @param field The field to add.
-	 */
-	public addField(field: APIEmbedField): this {
-		return this.addFields(field);
-	}
-
-	/**
 	 * Adds fields to the embed (max 25)
 	 *
 	 * @param fields The fields to add


kyranet (approved):

vladfrangu:
I think the method should stay as is, although I can see how it's redundant compared to `addFields`... 🤷 I'll let the others give in their thoughts too

nev-r:
i think it's nicer if d.js is ugly so that downstream code can look clean.
allocating a new array and wrapping it in extra syntax to add a single field
```js
addFields([field])
```
just to make the method happy, is kind of a tragedy.

i'd personally rather recommend people spread into addFields than the above mess

vladfrangu:
That's not how the spread arguments work, you don't need to wrap them in brackets

nev-r:
i didn't say they need to be in brackets?
i said i would rather do `addField(...myfields)` or `addFields(...myfields)`
than be forced to to `addFields([myfield])`

BenjammingKirby:
why just remove addField and not other builder methods like addComponent an such?

BenjammingKirby:
> i didn't say they need to be in brackets?
i said i would rather do addField(...myfields) or addFields(...myfields)
than be forced to to addFields([myfield])
(if the goal is to narrow to 1 method)

the methods exclusively take a rest parameter since <https://github.com/discordjs/discord.js/pull/7395>

SpaceEEC (approved):

nev-r:
> the methods exclusively take a rest parameter since #7395

ah, thanks. then yeah, such a little stub passthrough function like addField seems unnecessary

suneettipirneni:
> why just remove addField and not other builder methods like addChoice an such?

If this goes through they should be removed.

BenjammingKirby:
> I think this method should be changed to match v13’s MessageEmbed#addField instead because it is quite useless as it is right now. I don’t think removing it is the way to go because there are many situations where you may only need one field and creating the object just adds unnecessary visual burden to the code

you are forgetting why the change was made
changing to objects was the right call since some of the methods may take multiple optional properties, like setAuthor
what do you want? `setAuthor(something, undefined, something)`? an object is much better
and with this change they are now useless

ImRodry:
That argument makes no sense because fields are composed of 2 required parameters and 1 optional one

BenjammingKirby:
> That argument makes no sense because fields are composed of 2 required parameters and 1 optional one

i never said this particular one doesn't, but i did give you an example didn't i?
so what do you suggest? keep it like this for just this one method but for the others use objects? that to me makes no sense, we should try to stay consistent

ImRodry:
The example you gave is invalid because it comes from a situation with many optional parameters, which this one isn’t and you’d never be passing undefined for the first or second values because that would be an error

advaith1:
If the goal is to remove redundancy then imo addFields should be removed instead of addField, which is probably used a lot more.

almostSouji:
Semantically, `addFields` makes much more sense, you can add multiple, but adding one is fine too. Whereas adding multiple to `addField` is not nearly as intuitive

suneettipirneni:
> If the goal is to remove redundancy then imo addFields should be removed instead of addField, which is probably used a lot more.

Considering what's used more it's kinda hard to tell I wouldn't say for certain that one is used a lot more than the other. However, grammatically it makes more sense for it be plural since it accepts a rest parameter which is 1..n.

iCrawl (approved):

PR #7526, opened 2022-02-22
feat(GuildMemberManager): add `GuildMemberManager#fetchMe()`

Syjalo:
There is property `GuildMemberManager#me` and it can be `null`. And to fetch the client user you should do `guild.members.fetch(guild.client.user.id)`.
This PR makes much easier to fetch client user as GuildMember of the guild.

Example:
```js
if (!guild.members.me) {
  await guild.members.fetchMe();
}
```


ImRodry:
You will always receive your bot’s GuildMember object on every guild and this just sounds like a lazy shortcut, I don’t see a point in this

Syjalo:
> You will always receive your bot’s GuildMember object on every guild

You won't receive it if the `Guilds` intent is not enabled.

> This just sounds like a lazy shortcut

So `Guild#fetchOwner()` is also a lazy shortcut?

almeidx:
> You will always receive your bot’s GuildMember object on every guild 

`Guild#me` is a getter that relies on the cache, and because of sweepers it might not always be present

ImRodry:
> You won't receive it if the Guilds intent is not enabled.

You need the Guilds intent for basically everything else with a bot so I still don’t understand the need for this

> So Guild#fetchOwner() is also a lazy shortcut?

yes but at least that one has a reasoning behind it which is that the ownerId is provided by Discord so you need a way to fetch it because you may not always have it. For Guild#me it will always be there

> Guild#me is a getter that relies on the cache, and because of sweepers it might not always be present

It’s there to begin with. If you choose to remove it then that’s an issue you gotta sort out, that’s why it’s documented as nullable

ObscuritySRL:
> You need the Guilds intent for basically everything

Correct me if I'm wrong, but can bots not receive and respond to interactions without `Guilds`? If your bot doesn't involve any form of guild moderation, interactions will often times contain all the information you need — guild Id, member info, roles, and permissions, etc.

ImRodry:
> > You need the Guilds intent for basically everything
> 
> Correct me if I'm wrong, but can bots not receive and respond to interactions without `Guilds`? If your bot doesn't involve any form of guild moderation, interactions will often times contain all the information you need — guild Id, member info, roles, and permissions, etc. Unique case, sure, but AFAIK one where `Guilds` is not required.

If you don’t have the Guilds intent you won’t have guilds where you can use this method in, because Discord doesn’t send them

ObscuritySRL:
> > > You need the Guilds intent for basically everything
> > 
> > 
> > Correct me if I'm wrong, but can bots not receive and respond to interactions without `Guilds`? If your bot doesn't involve any form of guild moderation, interactions will often times contain all the information you need — guild Id, member info, roles, and permissions, etc. Unique case, sure, but AFAIK one where `Guilds` is not required.
> 
> If you don’t have the Guilds intent you won’t have guilds where you can use this method in, because Discord doesn’t send them

Forgot about this. Would it not still be possible to do `const me = new Guild(client, {id: someGuildId}).fetchMe()`? Still not sure how useful that'd be - I haven't used DJS apart from API types recently, but I recall doing something similar to add channel followers when I wasn't caching them a while back.

vladfrangu (approved):
I'm 🤷 on this

vladfrangu:
> > If you don’t have the Guilds intent you won’t have guilds where you can use this method in, because Discord doesn’t send them
> 
> Forgot about this. Would it not still be possible to do `const me = new Guild(client, {id: someGuildId}).fetchMe()` to get member info about the client user? Still not sure how useful that'd be - I haven't used DJS apart from API types recently, but I recall doing something similar to add channel followers when I wasn't caching them a while back.
> 
> **Edit**: Correction - at the time it was easiest to add a channel follower for a channel on another shard with `new Channel(...).addFollower()` or whatever.

That points a lot to the fact there's still tons of methods not accessible to managers and only to built structures, which is something we need to fix in the future

kyranet:
Come think of it, shouldn't we do `guild.members.fetchMe()` instead of `guild.fetchMe()`?

We could also move `guild.me` to `guild.members.me` in another PR.

Syjalo:
> Come think of it, shouldn't we do `guild.members.fetchMe()` instead of `guild.fetchMe()`?
> 
> We could also move `guild.me` to `guild.members.me` in another PR.

I think so. We could also move `guild.fetchOwner()`.

vladfrangu (approved):

kyranet (approved):

BenPatersonxx:
@iCrawl @SpaceEEC please review

SpaceEEC (approved):

iCrawl (approved):

PR #7529, opened 2022-02-22
types: default to text on channel create

ImRodry:

This fixes a bug in the types where providing an options object and no type property would create a channel that would be a union of all the channel types


ImRodry:
Turns out this was caused by trying to assign a channel with a type that wasn't CategoryChannel to the parent property of the options object. I didn't realize this when making the PR so this isn't needed

PR #7539, opened 2022-02-24
fix(ThreadMembersUpdate): Only emit added & removed thread members

### LINKED ISSUE #6513: Possibility of same old & new structure for `Client#threadMembersUpdate`
`<Client>.threadMembersUpdate` has the possibility of emitting the same `Collection` for both the `oldMembers` & `newMembers` parameters.

This is achieved by the following:
1. Join a thread
2. Restart your bot to ensure reproducibility. **Do not fetch active threads**!
3. Leave the thread

Upon leaving the thread, `oldMembers` and `newMembers` will emit the same `Collection`. Needless to say, this is quite... pointless! This is because the thread channel's members didn't have the member in question cached, ergo the same structures.

```js
const client = new (require("discord.js").Client)({
  intents: [
    "GUILDS",
    "GUILD_MEMBERS"
  ]
});

client.once("ready", () => {
  const thread = client.channels.resolve("thread_id"); // Thread
  console.log(thread.members.cache); // Collection(0) [Map] {}
});

// Emitting from thread_id
client.on("threadMembersUpdate", (oldMembers, newMembers) => {
  console.log(oldMembers.equals(newMembers)); // true
});

client.login("token");
```

**Possible Fixes:**
I guess the simple answer is the make the `oldMembers` parameter nullable after these operations:
https://github.com/discordjs/discord.js/blob/70cc0295f833cd988ea627b37d20536f73e21630/src/client/actions/ThreadMembersUpdate.js#L11-L20
We could just check to see if the size is equal and return `null` if so. By sending out `null`, we are essentially saying that an event occurred that we did not have enough data about (the old cache was unreliable).

~~I guess another way to do it would be to emit a "fake" `ThreadMember` that we create solely for it to be removed here just to ensure that the event fires with an actual `ThreadMember`. I'm unsure if this is possible.~~

**Further details:**

- discord.js version: 13.2.0-dev
- Node.js version: 16.6.1
- Operating system: macOS Big Sur 11.5.2
- Priority this issue should have – please be realistic and elaborate if possible: Low

**Relevant client options:**

- partials: none
- gateway intents: GUILDS & GUILD_MEMBERS
- other: none
- I have also tested the issue on latest main; Commit hash: f39b597e3120701287e86acdb247b4815f693c91

Jiralite:
As explained in #6513, the event may emit the **same** `oldMembers` and `newMembers` which was... not useful. It's not possible to tell what actually changed there - only that _something_ changed. This is due to discord.js always assuming the thread member leaving is cached.

I believe it'd be best to do nothing (like other events) if only uncached thread members leave. By doing this, the event does not fire as it would emit the same `Collection` for both parameters. We can utilise a new partial (`Partials.ThreadMember`) for people who still want that data. This solves that issue, but there is another.

discord.js emits these added & removed thread members alongside the existing thread member cache. Since Discord does not send the current thread members, this is susceptible to a caching issue where there may be some thread members in both payloads... or there may not. To find out what truly changed, you would have to perform operations on the received `Collection`s to find out the thread members that caused this gateway event to occur. Therefore, I removed this caching behaviour and aligned it to the API: only emitting the added & removed thread members.

Now, the event works like this:

```JavaScript
<Client>.on(Events.ThreadMembersUpdate, (thread, addedMembers, removedMembers) => {
  console.log(thread); // This is the thread where members got updated!
  
  if (addedMembers.size > 0) {
    // Only the members that were added are here.
  }

  if (removedMembers.size > 0) {
    // Only the members that were removed are here.
    if (removedMembers.first().partial) {
      // Uncached thread member left the thread.
      // This will be included with the new partial (`Partials.ThreadMember`).
    }
  }
});
```

The thread channel is being emitted too to avoid the user doing `addedMembers.first()?.thread ?? removedMembers.first().thread`. We already have the thread internally so it seemed reasonable to include it. Additionally, users can access the thread's member cache to do what they need as before.

Feedback is welcome and appreciated!

This resolves #6513.

### DIFF
--- a/packages/discord.js/src/client/actions/ThreadMembersUpdate.js
+++ b/packages/discord.js/src/client/actions/ThreadMembersUpdate.js
@@ -19,8 +19,8 @@ class ThreadMembersUpdateAction extends Action {
       );
 
       data.removed_member_ids?.reduce((removedMembersIds, removedMembersId) => {
-        const cachedThreadMember = thread.members.cache.get(removedMembersId);
-        if (cachedThreadMember) removedMembersIds.set(cachedThreadMember.id, cachedThreadMember);
+        const threadMember = this.getThreadMember(removedMembersId, thread.members);
+        if (threadMember) removedMembersIds.set(threadMember.id, threadMember);
         thread.members.cache.delete(removedMembersId);
         return removedMembersIds;
       }, removedMembers);
--- a/packages/discord.js/src/client/actions/ThreadMembersUpdate.js
+++ b/packages/discord.js/src/client/actions/ThreadMembersUpdate.js
@@ -33,10 +33,11 @@ class ThreadMembersUpdateAction extends Action {
       /**
        * Emitted whenever members are added or removed from a thread. Requires `GUILD_MEMBERS` privileged intent
        * @event Client#threadMembersUpdate
+       * @param {ThreadChannel} thread The thread where members got updated
        * @param {Collection<Snowflake, ThreadMember>} addedMembers The members that were added
        * @param {Collection<Snowflake, ThreadMember>} removedMembers The members that were removed
        */
-      client.emit(Events.ThreadMembersUpdate, addedMembers, removedMembers);
+      client.emit(Events.ThreadMembersUpdate, thread, addedMembers, removedMembers);
     }
     return {};
   }
--- a/packages/discord.js/src/client/actions/ThreadMembersUpdate.js
+++ b/packages/discord.js/src/client/actions/ThreadMembersUpdate.js
@@ -1,5 +1,6 @@
 'use strict';
 
+const { Collection } = require('@discordjs/collection');
 const Action = require('./Action');
 const Events = require('../../util/Events');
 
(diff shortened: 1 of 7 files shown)


packages/discord.js/src/client/actions/ThreadMembersUpdate.js
```diff
@@ -8,24 +9,35 @@ class ThreadMembersUpdateAction extends Action {
...
+        // Uncached thread member(s) left.
+        return {};
+      }
 
       /**
        * Emitted whenever members are added or removed from a thread. Requires `GUILD_MEMBERS` privileged intent
        * @event Client#threadMembersUpdate
-       * @param {Collection<Snowflake, ThreadMember>} oldMembers The members before the update
-       * @param {Collection<Snowflake, ThreadMember>} newMembers The members after the update
+       * @param {ThreadChannel} thread The thread where members got updated
```
ImRodry, line 36: This should be the last parameter imo

packages/discord.js/src/client/actions/ThreadMembersUpdate.js
```diff
@@ -8,24 +9,35 @@ class ThreadMembersUpdateAction extends Action {
...
+        (_addedMembers, addedMember) => _addedMembers.set(addedMember.user_id, thread.members._add(addedMember)),
+        addedMembers,
+      );
 
-      data.removed_member_ids?.forEach(memberId => {
-        thread.members.cache.delete(memberId);
-      });
+      data.removed_member_ids?.reduce((removedMembersIds, removedMembersId) => {
+        const cachedThreadMember = thread.members.cache.get(removedMembersId);
+        if (cachedThreadMember) removedMembersIds.set(cachedThreadMember.id, cachedThreadMember);
```
SpaceEEC, line 23: I think something allowing you to know who left regardless of whether the thread member was cached would be nice.
(Partials or just a `null` value maybe?)

SpaceEEC (approved):

packages/discord.js/src/client/actions/ThreadMembersUpdate.js
kyranet: Yes, we should at very least give some data back, probably a PartialUser so we can read the IDs when iterating the maps with `.values()`

Jiralite: I've gone ahead and created partial thread members. This should resolve the uncached behaviour here.

SpaceEEC (approved):

kyranet (approved):

iCrawl:
This needs a rebase.

iCrawl:
This needs a rebase.

iCrawl (approved):

PR #7542, opened 2022-02-24
fix: Handle partial data for `Typing#user`

### LINKED ISSUE #7102: `Client#typingStart#User` has `null` values incorrectly
### Issue description

The `Typing` class has a `user` property which appears to be typed as a `PartialUser`. When I access this object, I seem to always see the full, expected structure.

If I attempt to access `Typing#user#username`, `Typing#user#discriminator` or `Typing#user#tag` from a guild, TypeScript tells me that these properties are `null`, but when I log them, they are in fact not:

<details>
<summary>Image</summary>

image

</details>

Even if the user was uncached in a guild, I receive them.

Now, in direct messages (oh boy), absolutely everything but the id is `null` if only the `CHANNEL` partial is specified. Specifying the `USER` partial makes no difference here:
<details>
<summary>Response</summary>

```JavaScript
User {
  id: '618976181026422814',
  bot: null,
  system: null,
  flags: null,
  username: null,
  discriminator: null,
  avatar: null,
  banner: undefined,
  accentColor: undefined
}
```

</details>

Should discord.js emit this even if only the `id` is present?

To me, it looks like `PartialUser` isn't the right type for this property anyhow.

### Code sample

```typescript
import { Client, Intents, Constants } from "discord.js";

const client = new Client({
  intents: [
    Intents.FLAGS.GUILDS,
    Intents.FLAGS.GUILD_MESSAGE_TYPING,
    Intents.FLAGS.DIRECT_MESSAGE_TYPING
  ],
  partials: [
    Constants.PartialTypes.CHANNEL // Test in direct messages via an uncached user too!
  ]
});

client.on(Constants.Events.TYPING_START, typingStart => {
  console.log(typingStart); // Check the `user` property!
});

client.login();
```

### discord.js version

13.4.0-dev

### Node.js version

Node.js 16.13.0 & TypeScript 4.5.2

### Operating system

macOS Monterery 

Medium (should be fixed soon)

CHANNEL

GUILDS, GUILD_MESSAGE_TYPING, DIRECT_MESSAGE_TYPING

ee91f5a

Jiralite:
As detailed in #7102, `Typing#user` had the potential to be an unwanted partial. This is because `DMChannel#_patch()` always adds the user, even if the data only contained the user id. Additionally, the typings weren't correct for this.

Since there is a possibility of not receiving the full recipient data, only the `recipientId` is now on the `DMChannel` class. `recipient` is now a getter. If the `USER` partial is specified or a full structure is received, we add the recipient to the cache which in turn will respectively emit `Client#typingStart`.

This fully resolves #7102.

### DIFF
--- a/packages/discord.js/src/structures/DMChannel.js
+++ b/packages/discord.js/src/structures/DMChannel.js
@@ -4,6 +4,8 @@ const { ChannelType } = require('discord-api-types/v9');
 const { Channel } = require('./Channel');
 const TextBasedChannel = require('./interfaces/TextBasedChannel');
 const MessageManager = require('../managers/MessageManager');
+const { Formatters } = require('../util/Formatters');
+const Partials = require('../util/Partials');
 
 /**
  * Represents a direct message channel between two users.
@@ -29,10 +31,14 @@ class DMChannel extends Channel {
 
     if (data.recipients) {
       /**
-       * The recipient on the other end of the DM
-       * @type {User}
+       * The recipient's id
+       * @type {Snowflake}
        */
-      this.recipient = this.client.users._add(data.recipients[0]);
+      this.recipientId = data.recipients[0].id;
+
+      if ('username' in data.recipients[0] || this.client.options.partials.includes(Partials.Users)) {
+        this.client.users._add(data.recipients[0]);
+      }
     }
 
     if ('last_message_id' in data) {
@@ -64,12 +70,20 @@ class DMChannel extends Channel {
   }
 
   /**
+   * The recipient on the other end of the DM
+   * @type {?User}
+   */
+  get recipient() {
+    return this.client.users.resolve(this.recipientId);
+  }
+
+  /**
    * Fetch this DMChannel.
    * @param {boolean} [force=true] Whether to skip the cache check and request the API
    * @returns {Promise<DMChannel>}
    */
   fetch(force = true) {
-    return this.recipient.createDM(force);
+    return this.client.users.createDM(this.recipientId, force);
   }
 
   /**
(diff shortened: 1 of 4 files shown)


packages/discord.js/src/structures/DMChannel.js
```diff
@@ -29,10 +31,14 @@ class DMChannel extends Channel {
...
       /**
-       * The recipient on the other end of the DM
-       * @type {User}
+       * The recipient's id
+       * @type {Snowflake}
        */
-      this.recipient = this.client.users._add(data.recipients[0]);
+      this.recipientId = data.recipients[0].id;
+
+      if ('username' in data.recipients[0] || this.client.options.partials.includes(Partials.Users)) {
```
almeidx, line 39: Perhaps create a variable for `data.recipients[0]`?

Jiralite: Sure! Done with c6032a6b80593ad8432556d4c701f9283ada6309.

SpaceEEC (approved):

packages/discord.js/src/structures/DMChannel.js
```diff
@@ -81,7 +98,7 @@ class DMChannel extends Channel {
    * console.log(`Hello from ${channel}!`);
    */
   toString() {
-    return this.recipient.toString();
+    return Formatters.userMention(this.recipientId);
```
kyranet, line 101: Instead of using the Formatters namespace class, you should use builder's function directly.

kyranet (dismissed):

packages/discord.js/src/structures/DMChannel.js
Jiralite: Should be done with 41097f3ae44d1d4ee43015ed75c6d6c125e1c3d1.

vladfrangu (approved):

kyranet (approved):

PR #7546, opened 2022-02-25
fix: only check `instanceof Component` once

suneettipirneni:

Previously a check was done for each component type. This simplifies the logic to one check for a class-based component.

### DIFF
--- a/packages/builders/src/components/Components.ts
+++ b/packages/builders/src/components/Components.ts
@@ -17,13 +17,17 @@ export function createComponent<T extends keyof MappedComponentTypes>(
 ): MappedComponentTypes[T];
 export function createComponent<C extends MessageComponent>(data: C): C;
 export function createComponent(data: APIMessageComponent | MessageComponent): Component {
+	if (data instanceof Component) {
+		return data;
+	}
+
 	switch (data.type) {
 		case ComponentType.ActionRow:
-			return (data instanceof ActionRow ? data : new ActionRow(data)) as Component;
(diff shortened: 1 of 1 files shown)


packages/builders/src/components/Components.ts
```diff
@@ -17,13 +17,17 @@ export function createComponent<T extends keyof MappedComponentTypes>(
 ): MappedComponentTypes[T];
 export function createComponent<C extends MessageComponent>(data: C): C;
 export function createComponent(data: APIMessageComponent | MessageComponent): Component {
+	if (data instanceof Component) {
+		return data;
+	}
```
ImRodry, line 22: Surely you could make this into 1 line right?

kyranet, line 22:
```suggestion
	if (data instanceof Component) return data;
```

kyranet (changes requested):

SpaceEEC (approved):

packages/builders/src/components/Components.ts
vladfrangu: I don't agree with this change 😂

vladfrangu (approved):

kyranet (approved):

packages/builders/src/components/Components.ts
ImRodry: 🤔

iCrawl (approved):

PR #7547, opened 2022-02-25
refactor: replace zod with shapeshift

imranbarbhuiya:
wait for 
- [x] v2 release


imranbarbhuiya:
ci will fail bc union takes rest param which isn't intentional. So I'm not making it rest param now.

kyranet (commented):
Just FYI, you'll need to update `lengthGte` to `lengthGe` and `lengthLte` to `lengthLe` once v2 is released.

Also, we're not releasing v2 just yet, we need to write documentation for everything.

iCrawl:
This needs a rebase.

kyranet:
[@sapphire/shapeshift](https://www.npmjs.com/package/@sapphire/shapeshift) v2 just released, you can now rebase the PR and use it 🎉

imranbarbhuiya:
> [@sapphire/shapeshift](https://www.npmjs.com/package/@sapphire/shapeshift) v2 just released, you can now rebase the PR and use it 🎉

done

packages/builders/src/interactions/slashCommands/mixins/ApplicationCommandOptionChannelTypesMixin.ts
```diff
@@ -16,13 +16,13 @@ const allowedChannelTypes = [
...
-			...ZodLiteral<ChannelType>[]
-		],
+const channelTypesPredicate = s.array(
+	s.union(
+		...(allowedChannelTypes.map((type) => s.literal(type)) as [
+			LiteralValidator<ChannelType>,
+			LiteralValidator<ChannelType>,
+			...LiteralValidator<ChannelType>[]
+		]),
 	),
```
Khasms, line 26:
```suggestion
	s.union(...allowedChannelTypes.map((type) => s.literal(type))),
```
This cast is no longer needed.

packages/builders/src/interactions/slashCommands/mixins/ApplicationCommandOptionChannelTypesMixin.ts
```diff
@@ -1,5 +1,5 @@
 import { ChannelType } from 'discord-api-types/v9';
-import { z, ZodLiteral } from 'zod';
+import { s, LiteralValidator } from '@sapphire/shapeshift';
```
Khasms, line 2:
```suggestion
import { s } from '@sapphire/shapeshift';
```
The cast this is used in is no longer needed.

Khasms (changes requested):

packages/builders/src/interactions/modals/Assertions.ts
```diff
@@ -1,9 +1,9 @@
-import { z } from 'zod';
+import { s } from '@sapphire/shapeshift';
 import { ActionRowBuilder, type ModalActionRowComponentBuilder } from '../..';
 import { customIdValidator } from '../../components/Assertions';
 
-export const titleValidator = z.string().min(1).max(45);
-export const componentsValidator = z.array(z.instanceof(ActionRowBuilder)).min(1);
+export const titleValidator = s.string.lengthGe(1).lengthLe(45);
+export const componentsValidator = s.array(s.instance(ActionRowBuilder)).lengthGe(1);
```
kyranet, line 6:
```suggestion
export const componentsValidator = s.instance(ActionRowBuilder).array.lengthGe(1);
```

packages/builders/src/components/Assertions.ts
```diff
@@ -1,43 +1,42 @@
...
-export const buttonStyleValidator = z.number().int().min(ButtonStyle.Primary).max(ButtonStyle.Link);
+export const buttonStyleValidator = s.nativeEnum(ButtonStyle);
 
-export const placeholderValidator = z.string().max(150);
-export const minMaxValidator = z.number().int().min(0).max(25);
+export const placeholderValidator = s.string.lengthLe(150);
+export const minMaxValidator = s.number.int.ge(0).le(25);
 
-export const optionsValidator = z.object({}).array().nonempty();
+export const optionsValidator = s.object({}).array.lengthGe(1);
```
kyranet, line 22: Why is this an empty object?

By the way, `zod`'s `z.object()` defaults to a passthrough object validator, however, `@sapphire/shapeshift` has stricter defaults, and as such, it defaults to an ignore object validator.

This means, while `[{ foo: 'bar' }]` is valid for this validator, the properties would be stripped and will always result on `[{}]`. If unknown properties are desired to be kept (albeit no validation done for them), you need to use `.passthrough`.

kyranet (changes requested):

packages/builders/src/components/Assertions.ts
imranbarbhuiya: s.object is default to ignore so if u pass any key value it'll ignore them and will return `{}`. In this case 
https://github.com/discordjs/discord.js/blob/87a6b8445bfbf3981cd39813fe961dfa1c7f2bce/packages/builders/src/components/Assertions.ts#L29

we don't need the return value. It's just there to check the input is an array of objects. So imo adding `passthrough` here is redundant.

packages/builders/src/interactions/slashCommands/mixins/ApplicationCommandOptionChannelTypesMixin.ts
```diff
@@ -16,15 +16,7 @@ const allowedChannelTypes = [
...
-const channelTypesPredicate = z.array(
-	z.union(
-		allowedChannelTypes.map((type) => z.literal(type)) as [
-			ZodLiteral<ChannelType>,
-			ZodLiteral<ChannelType>,
-			...ZodLiteral<ChannelType>[]
-		],
-	),
-);
+const channelTypesPredicate = s.array(s.union(...allowedChannelTypes.map((type) => s.literal(type))));
```
vladfrangu, line 19: I wonder in the future if there's any way to use nativeEnum for this too (probably not without remaking the enum here so this is fine)

vladfrangu (approved):

kyranet (approved):

iCrawl (approved):

favna (approved):
tfw was looking at some builders code, realised I never checked if this was already made and turns out it already was.

SpaceEEC (approved):

PR #7548, opened 2022-02-25
Fix(types): Allow Date object in`MessageEmbed.timestamp` 

AuroraTook:
In raw `MessageEmbed` object, Allow `MessageEmbed.timestamp` to use Date object

### DIFF
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -1691,7 +1691,7 @@ export class MessageEmbed {
   public readonly length: number;
   public provider: MessageEmbedProvider | null;
   public thumbnail: MessageEmbedThumbnail | null;
-  public timestamp: number | null;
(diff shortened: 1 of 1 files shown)


ImRodry:
You should probably removed the dozens of unrelated changes in this PR

kyranet (commented):
In which case have you seen `MessageEmbed#timestamp` to be a `Date`? The source code indicates it can only be either a number or null:

https://github.com/discordjs/discord.js/blob/f0960698d2ef51d4586f3c89d8b7c94faf5d921b/src/structures/MessageEmbed.js#L90

And the options also accepts dates:

https://github.com/discordjs/discord.js/blob/f0960698d2ef51d4586f3c89d8b7c94faf5d921b/typings/index.d.ts#L5090-L5094

PR #7575, opened 2022-02-27
fix(guild): throw if ownerId falsey

### LINKED ISSUE #7574: guild.fetchOwner() is returning a collection of guild members

discord.js

### Issue description

Using Discord 13.6.0 I am trying to get the id of the guild, is not working for me.

Trying `guild.ownerId` returns undefined, but I was then directed to try `guild.fetchOwner()`, but this returns a list of members of the guild instead of a `GuildMember`.

Looking at the Discord source code suggests that if `guild.ownerId` is undefined, this will return all the members of the guild:

``js
fetchOwner(options) {
    return this.members.fetch({ ...options, user: this.ownerId });
}
```


### Code sample

```typescript
To get the guild:


  async getGuild (guildId?: string) {
    const client = await this.getClient();
    const guildResolvable = client.guilds.resolveId(guildId || this.guildId);

    const guild = client.guilds.resolve(guildResolvable);
    if (!guild) {
      throw new Error('Unable to access guild');
    }
    return guild;
  }


This returns a guild with `ownerId` being `undefined, but then calling `guild.fetchOwner()` returns a collection of form `Collection(11) [Map] { ... }`.
```

### Package version

13.6.0

### Node.js version

16.6.0

### Operating system

macOS

Medium (should be fixed soon)

No Partials

GuildMembers, GuildMessages

_No response_

monbrey:
Fixes #7574 and changes the method to follow our prefer-async style.

I guess it's technically breaking since the method can now throw an error where it didn't before (although it probably should have)

### DIFF
--- a/packages/discord.js/src/structures/Guild.js
+++ b/packages/discord.js/src/structures/Guild.js
@@ -451,8 +451,12 @@ class Guild extends AnonymousGuild {
    * @param {BaseFetchOptions} [options] The options for fetching the member
    * @returns {Promise<GuildMember>}
    */
-  fetchOwner(options) {
-    return this.members.fetch({ ...options, user: this.ownerId });
+  async fetchOwner(options) {
+    if (!this.ownerId) {
+      throw new Error('FETCH_OWNER_ID');
+    }
+    const member = await this.members.fetch({ ...options, user: this.ownerId });
+    return member;
   }
 
   /**
--- a/packages/discord.js/src/errors/Messages.js
+++ b/packages/discord.js/src/errors/Messages.js
@@ -90,6 +90,7 @@ const Messages = {
   CHANNEL_NOT_CACHED: 'Could not find the channel where this message came from in the cache!',
   STAGE_CHANNEL_RESOLVE: 'Could not resolve channel to a stage channel.',
   GUILD_SCHEDULED_EVENT_RESOLVE: 'Could not resolve the guild scheduled event.',
+  FETCH_OWNER_ID: "Couldn't resolve the guild ownerId to fetch the member.",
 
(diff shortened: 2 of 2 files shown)


monbrey:
Just to pre-empt some expected comments:

- Yes I know ownerId isn't supposed to ever be undefined in a proper Guild object
- Yes I know the only time it probably ever is would be when you don't specify the Guilds intent
- This is still a good check and will probably become even more relevant as we shift towards the raw data storage approach

packages/discord.js/src/structures/Guild.js
```diff
@@ -451,8 +451,12 @@ class Guild extends AnonymousGuild {
...
    * @returns {Promise<GuildMember>}
    */
-  fetchOwner(options) {
-    return this.members.fetch({ ...options, user: this.ownerId });
+  async fetchOwner(options) {
+    if (!this.ownerId) {
+      throw new Error('FETCH_OWNER_ID');
+    }
+    const member = await this.members.fetch({ ...options, user: this.ownerId });
+    return member;
```
ImRodry, line 459: Can’t you just return this without assigning a constant and remove the await? What’s the point of this variable?

monbrey: Making the linter happy that this isn't an async method without await.

ImRodry: Then don’t make this an async method and return a Promise.reject

monbrey: But we prefer async methods - I checked that internally first.

ImRodry: Not really, most methods return a promise rejection when they don’t use the await keyword anywhere else.

kyranet (approved):

iCrawl (approved):

vladfrangu (approved):

SpaceEEC (approved):

PR #7584, opened 2022-03-01
refactor: Don't return builders from API data

suneettipirneni:

Component Builders currently maintain a responsibility it shouldn't. That responsibility is representing received data from the API. The issue with doing this is that it assumes that received data and in-progress build data is *always* the same type. However, in many places this isn't the case:

- Builders select menu component's `customId` is optional when it's always non-null from the API
- Button `style` is always required while it's optional in builders.
- Select Menu `options` are optional in builders while always non-null from the API
- Modal Submit Interactions don't return the full set of fields that a builder uses, which means you have a huge discrepancy between available fields and non-available fields.

Because of this, builders are now more aligned to their original goal: *build data*, don't represent it

### Relieving responsibility for builders:

Most noticeably, builders now have a `Builder` suffix. This makes it extremely clear that the class is purely a builder, nothing more and nothing less.

For example `ButtonComponent` -> `ButtonBuilder`

#### Removal of getters from builders

Keep in mind that builders are meant to be used in places beyond just discord.js. In situations outside of discord.js getters on builders hold little-to-no utility. The main reason as to why is because if you're constructing a structure you already have all the data needed to construct it. If you already have that data in the first place there's no need to retrieve it from a getter. 

The main use case of getters on builders was to get data *when received from discord.js*. However, with this PR removing the responsibility of builders to represent structures, the getters are no longer needed.

Instead, getters have been relocated to their respective finalized structures.

### Finalized Structures

`ButtonComponent`, `SelectMenuComponent`, `Embed` etc. are now structures within discord.js fully responsible for representing data received from the API. This allows us to always have API-accurate representations of the data. In addition, it prevents mutation side-effects by making the structures immutable.

For example, previously a user could call a `setX` method on a returned API structure. While on the surface it seems harmless, it actually modifies the cache. The cache should *never* be modified by anything externally, if it is modified it's no longer a valid cache since the data is out of sync.

*How would I modify a received structure, if I want to make a small change?*
Like this:
```ts
const modified = EmbedBuilder.from(receivedEmbed)
  .setTitle('foo');

interaction.reply({ embeds: [modified] })
```

This preserves immutability while also allowing you to use preexisting component data.

### Is this RichEmbed vs MessageEmbed all over again?

tdlr; no

Firstly, if I were to ask someone: "Which class is a builder? `RichEmbed` or `MessageEmbed`? ", most people probably wouldn't know. This is because the names don't really indicate the differences between these embeds. `Rich` over `Message` doesn't mean anything to me, so figuring out which one to use is difficult.

In this PR structures are named clearly as to what they do. "Embed vs EmbedBuilder" is a much clearer distinction. I can clearly tell one of the structures "builds" an embed while the other structure *is* an embed.

Secondly, both `RichEmbed` and `MessageEmbed` were both capable of representing a received embed structure. Whereas in this PR *only* the finalized structures can represent API responses, not the builders. For example:

```ts
// RichEmbed, MessageEmbed
const { title } = richEmbed; // OK
const { title } = messageEmbed // OK

// Embed, EmbedBuilder
const { title } = embed; // OK
const { title } = embedBuilder; // Error - property title doesn't exist on `EmbedBuilder`
```


packages/discord.js/src/structures/ActionRow.js
```diff
@@ -1,11 +1,21 @@
...
 
-class ActionRow extends BuildersActionRow {
-  constructor(data) {
-    super(Transformers.toSnakeCase(data));
+/**
+ * Represents an action row
+ * @extends {Component}
+ */
+class ActionRow extends Component {
+  constructor({ components, data } = {}) {
```
kyranet, line 11:
```suggestion
  constructor({ components, ...data } = {}) {
```

kyranet (changes requested):

packages/discord.js/src/structures/ActionRow.js
```diff
@@ -1,11 +1,21 @@
...
+ */
+class ActionRow extends Component {
+  constructor({ components, ...data } = {}) {
+    super(data);
+    /**
+     * The components in this action row
+     * @type {Component[]}
+     * @readonly
+     */
+    this.components = components.map(c => Components.createComponent(c));
```
kyranet, line 18:
```suggestion
    this.components = components?.map(c => Components.createComponent(c)) ?? [];
```

kyranet (approved):
LGTM

packages/discord.js/src/structures/ActionRow.js
ImRodry: You don't need the `?? []`, right?

packages/discord.js/typings/index.d.ts
```diff
@@ -501,19 +545,40 @@ export interface EmbedData {
...
+  private constructor(data?: APIEmbed);
+  public get fields(): APIEmbedField[] | undefined;
+  public get title(): string | undefined;
+  public get description(): string | undefined;
+  public get url(): string | undefined;
+  public get color(): number | undefined;
+  public get timestamp(): string | undefined;
+  public get thumbnail(): EmbedImageData | undefined;
+  public get image(): EmbedImageData | undefined;
+}
```
ImRodry, line 580: where's #equals?

packages/discord.js/src/structures/Embed.js
```diff
@@ -1,11 +1,195 @@
...
+    return this.data.provider;
+  }
+
+  /**
+   * The embed footer data
+   * @type {?EmbedFooterData}
+   * @readonly
+   */
+  get footer() {
+    if (!this.data.footer) return undefined;
```
ImRodry, line 143: explicit undefined is never a good idea, I think this would be better
```suggestion
    if (!this.data.footer) return this.data.footer;
```

packages/discord.js/src/structures/ButtonComponent.js
```diff
@@ -1,11 +1,64 @@
...
+    return this.data.custom_id;
+  }
+
+  /**
+   * The URL of this button (only defined on link buttons)
+   * @type {?string}
+   * @readonly
+   */
+  get url() {
+    return this.data.url;
```
imranbarbhuiya, line 61:
```suggestion
  /**
   * The custom id of this button
   * @type {string}
   * @readonly
   */
  get customId() {
    return this.data.custom_id;
```

You can't receive interaction from a link button so url doesn't make sense here.

packages/discord.js/typings/index.d.ts
```diff
@@ -476,16 +493,43 @@ export class ButtonInteraction<Cached extends CacheType = CacheType> extends Mes
...
+export class Component<T extends APIMessageComponent = APIMessageComponent> {
+  public data: T;
+  public get type(): T['type'];
+  public toJSON(): T;
+  public equals(other: this | T): boolean;
+}
+
+export class ButtonComponent extends Component<APIButtonComponent> {
+  private constructor(data?: APIButtonComponent);
+  public get style(): ButtonStyle;
```
imranbarbhuiya, line 505:
```suggestion
  public get style(): Omit<ButtonStyle, "LINK">;
```

packages/discord.js/typings/index.d.ts
```diff
@@ -476,16 +493,43 @@ export class ButtonInteraction<Cached extends CacheType = CacheType> extends Mes
...
+  public get label(): string | undefined;
+  public get emoji(): APIMessageComponentEmoji | undefined;
+  public get disabled(): boolean | undefined;
+  public get customId(): string | undefined;
+  public get url(): string | undefined;
+}
```
imranbarbhuiya, line 511:
```suggestion
  public get customId(): string;
}
```

imranbarbhuiya (changes requested):

packages/discord.js/src/structures/ActionRow.js
suneettipirneni: It would, optional chaining makes it undefined if components is undefined.

packages/discord.js/src/structures/ButtonComponent.js
monbrey: These aren't solely for interactions though? ButtonComponents can be LINK type and exist in Messages. Which your bot absolutely sees/receives/handles.

packages/discord.js/src/structures/ActionRow.js
kyranet: The constructor does `= {}`, so if no data is passed, this will error.

If explicit data is expected, then remove the default in the constructor.

packages/discord.js/src/structures/ButtonComponent.js
imranbarbhuiya: Yeah. I forgot sorry

packages/discord.js/src/structures/Embed.js
suneettipirneni: Can you explain why explicit `undefined` is never a good idea?

ImRodry: It’s just general bad practice and there’s linting rules to disallow it. I think you should even return null unless there’s a problem with that

packages/discord.js/src/structures/Embed.js
```diff
@@ -1,11 +1,195 @@
...
-class Embed extends BuildersEmbed {
-  constructor(data) {
-    super(Transformers.toSnakeCase(data));
+class Embed {
+  /**
+   * Creates a new embed object
+   * @param {APIEmbed} data API embed data
+   * @private
+   */
+  constructor(data = {}) {
```
muchnameless, line 11:
```suggestion
  constructor(data) {
```
`{ ...undefined }` in L17 works fine

packages/builders/src/components/textInput/UnsafeTextInput.ts
```diff
@@ -1,73 +1,17 @@
...
-	 * Whether this text input is required
-	 */
-	public get required() {
-		return this.data.required;
-	}
-
 	/**
 	 * Sets the custom id for this text input
-	 * @param customId The custom id of this text input
+	 * @param customId The custom id of this text inputå
```
SpaceEEC, line 14:
```suggestion
	 * @param customId The custom id of this text input
```

SpaceEEC (changes requested):
Otherwise looking good

iCrawl:
This needs a rebase.

vladfrangu:
In Crawl's words, this needs a rebase

packages/builders/src/components/Components.ts
```diff
@@ -1,37 +1,43 @@
...
-			return new SelectMenuComponent(data);
+			return new SelectMenuBuilder(data);
 		case ComponentType.TextInput:
-			return new TextInputComponent(data);
+			return new TextInputBuilder(data);
 		default:
-			throw new Error(`Cannot serialize component type: ${(data as APIBaseComponent<ComponentType>).type}`);
+			// @ts-expect-error
+			// eslint-disable-next-line @typescript-eslint/restrict-template-expressions
+			console.warn(`Cannot properly serialize component type: ${data.type}`);
```
vladfrangu, line 40: Why? This should throw an error not return anything

packages/discord.js/src/structures/Component.js
```diff
@@ -0,0 +1,52 @@
...
+   * Creates a new component from API data
+   * @param {APIMessageComponent} data The API component data
+   * @private
+   */
+  constructor(data) {
+    /**
+     * The API data associated with this button component
+     * @type {APIMessageComponent}
+     */
+    this.data = { ...data };
```
vladfrangu, line 19: any reason for cloning the data?

packages/discord.js/src/structures/SelectMenuComponent.js
```diff
@@ -1,11 +1,64 @@
...
+   * @type {?boolean}
+   * @readonly
+   */
+  get disabled() {
+    return this.data.disabled ?? null;
+  }
+
+  /**
+   * The options in this select menu
+   * @type {APISelectMenuOption[]}
```
vladfrangu, line 57: This doesn't seem ideal for a d.js component, shouldn't it return camelCase data

packages/discord.js/src/util/Components.js
```diff
@@ -56,3 +56,37 @@
...
+      case ComponentType.ActionRow:
+        return new ActionRow(data);
+      case ComponentType.Button:
+        return new ButtonComponent(data);
+      case ComponentType.SelectMenu:
+        return new SelectMenuComponent(data);
+      default:
+        // This is done for future-proofing purposes to prevent
+        // crashes when unknown component types are found.
+        console.warn(`Found unknown component type: ${data.type}`);
```
vladfrangu, line 81: Ditto as builders, we don't return unknown channels for instance

vladfrangu (dismissed):

packages/discord.js/src/structures/SelectMenuComponent.js
suneettipirneni: `APISelectMenuOption` has camelCased properties already.

iCrawl (approved):

vladfrangu (approved):

packages/discord.js/src/structures/Message.js
```diff
@@ -1,6 +1,6 @@
 'use strict';
 
-const { createComponent, Embed } = require('@discordjs/builders');
+const { Embed } = require('@discordjs/builders');
```
SpaceEEC, line 3: This should probably import `'./embed'` instead.

SpaceEEC (changes requested):
Otherwise probably fine

packages/discord.js/src/structures/Message.js
suneettipirneni: Fixed

SpaceEEC (approved):

Issue #7373, opened 2022-01-31
Util.escapeInlineCode fails to properly escape inline code in some cases


dshepsis:
### Which package is this bug report for?

discord.js

### Issue description

Have the bot send a message with the content:
```js
Util.escapeInlineCode("``a`")
```
The function will return ``` ``a\` ```, indicating only the last backtick was escaped by prepending a backslash. The bot's message will appear as shown below:

![image](https://user-images.githubusercontent.com/16926704/151745944-5af2ef9b-643b-4908-b4b0-d4f16676cfb1.png)

As you can see, it's rendered as inline code with the content ``` `a\ ```, indicating the first and last backtick were interpreted as marking inline code, even though the last backtick is supposed to be escaped.

In order to correctly escape this string, `Util.escapeInlineCode` should've returned ``` \`\`a` ``` or ``` \`\`a\` ```.

Note that the `a` can be replaced with any string without backticks, as far as I can tell, without affecting the issue. Perhaps there are other kinds of strings you could craft that would cause similar failures, but I couldn't find any.

I'm not sure the exact best way to fix this method, as it may be dependent on exactly how Discord's own markdown parser is built.

### Code sample

```typescript
Util.escapeInlineCode("``a`")
```

### Package version

discord.js@13.5.1

### Node.js version

v16.13.1

### Operating system

Windows

### Priority this issue should have

Low (slightly annoying)

### Which partials do you have configured?

Not applicable (subpackage bug)

### Which gateway intents are you subscribing to?

Not applicable (subpackage bug)

PR #7587, opened 2022-03-02
fix(Util): escapeInlineCode properly

imranbarbhuiya:
fixes: #7373

The purpose of this method is to remove all the markdowns but as shown in the issue that it's not able to remove `inlineCode` for some strings where additional backticks are used. The current change works fine for any input.

### DIFF
--- a/packages/discord.js/src/util/Util.js
+++ b/packages/discord.js/src/util/Util.js
@@ -189,7 +189,7 @@ class Util extends null {
    * @returns {string}
    */
   static escapeInlineCode(text) {
-    return text.replace(/(?<=^|[^`])`(?=[^`]|$)/g, '\\`');
+    return text.replaceAll("`", '\\`');
   }
 
   /**
--- a/packages/discord.js/src/util/Util.js
+++ b/packages/discord.js/src/util/Util.js
@@ -189,7 +189,7 @@ class Util extends null {
    * @returns {string}
    */
   static escapeInlineCode(text) {
-    return text.replaceAll("`", '\\`');
+    return text.replaceAll('`', '\\`');
   }
 
   /**


iCrawl (approved):

SpaceEEC (approved):

iCrawl (approved):

dshepsis:
I don't this patch is correct. This simply escapes all backticks. As a result, it does escape inline code, but also escapes code blocks, even though that's a separate function and not the intended functionality of this function by itself. A fix should be implemented, and tests should be added, to ensure that this function does not escape backticks which form a code block in discord. I've tried a few different strings, and I'm not sure exactly what the correct rules are. For example:

This one:
~~~
```a`
~~~
Results in inline code containing "``a".

This one:
~~~
`a```b```
~~~
Results in plaintext "`a" followed by a code block containing "b"


And this one:
~~~
`a\```b```
~~~
Results in plaintext "`a\```b```"

There are probably more weird corner cases like this. Not really sure what the best solution is, other than to randomly generate a bunch of these, and then progressively add more of them as test cases until there are no discrepancies between what Discord and Discord.js identify as inline code or a code block.

imranbarbhuiya:
> I don't this patch is correct. This simply escapes all backticks. As a result, it does escape inline code, but also escapes code blocks, even though that's a separate function and not the intended functionality of this function by itself. A fix should be implemented, and tests should be added, to ensure that this function does not escape backticks which form a code block in discord. I've tried a few different strings, and I'm not sure exactly what the correct rules are. For example:
> 
> This one:
> 
> ```
> ```a`
> ```
> 
> Results in inline code containing "``a".
> 
> This one:
> 
> ```
> `a```b```
> ```
> 
> Results in plaintext "`a" followed by a code block containing "b"
> 
> And this one:
> 
> ```
> `a\```b```
> ```
> 
> Results in plaintext "`a```b```"
> 
> There are probably more weird corner cases like this. Not really sure what the best solution is, other than to randomly generate a bunch of these, and then progressively add more of them as test cases until there are no discrepancies between what Discord and Discord.js identify as inline code or a code block.

I'll probably create a new pr to ignore code blocks.

monbrey:
Why would it be a new PR if it's been identified as a defect of this PR?

PR #7593, opened 2022-03-03
refactor(embed): allow hex strings in `setColor()`

suneettipirneni:

Originally I had assumed hex number literals would suffice over hex strings. However, many external services may return colors as hex strings because it's a popular format. This PR allows hex strings to be passed into `setColor` for discord.js embed builders.

### DIFF
--- a/packages/discord.js/src/structures/Embed.js
+++ b/packages/discord.js/src/structures/Embed.js
@@ -4,11 +4,19 @@ const { Embed: BuildersEmbed } = require('@discordjs/builders');
 const Transformers = require('../util/Transformers');
 const { Util } = require('../util/Util');
 
+/**
+ * Represents an embed object
+ */
 class Embed extends BuildersEmbed {
   constructor(data) {
     super(Transformers.toSnakeCase(data));
   }
 
+  /**
+   * Sets the color of this embed
+   * @param {ColorResolvable} color The color of the embed
+   * @returns {Embed}
+   */
   setColor(color) {
     if (color === null) {
       return super.setColor(null);
--- a/packages/discord.js/src/structures/Embed.js
+++ b/packages/discord.js/src/structures/Embed.js
@@ -2,10 +2,18 @@
 
 const { Embed: BuildersEmbed } = require('@discordjs/builders');
 const Transformers = require('../util/Transformers');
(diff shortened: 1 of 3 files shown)


packages/discord.js/src/structures/Embed.js
```diff
@@ -2,11 +2,19 @@
...
 const { Embed: BuildersEmbed } = require('@discordjs/builders');
 const Transformers = require('../util/Transformers');
+const { Util } = require('../util/Util');
 
 class Embed extends BuildersEmbed {
   constructor(data) {
     super(Transformers.toSnakeCase(data));
   }
+
+  setColor(color) {
```
ImRodry, line 12: this should be documented because the allowed type is different (also the main class should be documented too ig)

ImRodry (commented):
BTW this doesn't only allow hex strings, it also allows Color strings e.g. "Blurple"

packages/discord.js/src/structures/Embed.js
```diff
@@ -2,11 +2,28 @@
...
+ * Represents an embed object
+ */
 class Embed extends BuildersEmbed {
   constructor(data) {
     super(Transformers.toSnakeCase(data));
   }
+
+  /**
+   * Sets the color of this embed
+   *
```
ImRodry, line 17:
```suggestion
```

packages/discord.js/src/structures/Embed.js
```diff
@@ -2,11 +2,28 @@
...
+   * @param {ColorResolvable} color The color of the embed
```
ImRodry, line 18:
```suggestion
   * @param {ColorResolvable} color The new color of the embed
```

suneettipirneni: Sorry, but the docs are currently mirroring the way it's documented in the builders package. Going to keep this the same.

ImRodry: There’s always room for improvement and having it like this makes it seem like it’s the current color but alright 🤔

iCrawl (approved):

vladfrangu (approved):

SpaceEEC (approved):

PR #7616, opened 2022-03-07
refactor: allow builders to accept emoji strings

suneettipirneni:


kyranet (approved):

iCrawl (approved):

packages/discord.js/typings/index.test-d.ts
```diff
@@ -1358,7 +1358,11 @@ new SelectMenuComponent({
 
 new ButtonComponent({
   style: ButtonStyle.Danger,
-});
+})
+  .setEmoji('<a:foo:123>')
+  .setEmoji('<:foo:123>')
+  // @ts-expect-error
+  .setEmoji(':foobar:123');
```
vladfrangu, line 1365: Util.parseEmoji can also match `foobar:123`, `:foobar:123`, etc. Might want to update types if we want to be extra strict, or just keep it as string

vladfrangu (changes requested):

packages/discord.js/typings/index.test-d.ts
imranbarbhuiya: It should allow default emojis too. So imo string will be better option.

ImRodry: It should just allow everything that parseEmoji allows, just like in v13

imranbarbhuiya: > It should just allow everything that parseEmoji allows, just like in v13
It should also allow `APIMessageComponentEmoji`.

packages/discord.js/src/structures/SelectMenuOption.js
```diff
@@ -0,0 +1,23 @@
...
+const { SelectMenuOption: BuildersSelectMenuOption } = require('@discordjs/builders');
+const Util = require('../util/Util');
+
+/**
+ * Represents a select menu option
+ */
+class SelectMenuOption extends BuildersSelectMenuOption {
+  /**
+   * Sets the emoji to display on this option
+   * @param {APIMessageComponentEmoji} emoji The emoji to display on this option
```
muchnameless, line 12:
```suggestion
   * @param {ComponentEmojiResolvable} emoji The emoji to display on this option
```

muchnameless (changes requested):

packages/discord.js/typings/index.test-d.ts
```diff
@@ -1367,7 +1367,16 @@ new SelectMenuBuilder({
 
 new ButtonBuilder({
   style: ButtonStyle.Danger,
-});
+})
+  .setEmoji('<a:foo:123>')
+  .setEmoji('<:foo:123>')
+  .setEmoji('foobar:123')
+  .setEmoji('normal')
```
vladfrangu, line 1374:
```suggestion
  .setEmoji('😏')
```

vladfrangu (approved):
LGTM, just add this

SpaceEEC (approved):

PR #7620, opened 2022-03-07
types: return `ReadonlyCollection`s where applicable

suneettipirneni:

Return readonly collections in areas where we return data based on the API.

### DIFF
--- a/packages/discord.js/typings/index.d.ts
+++ b/packages/discord.js/typings/index.d.ts
@@ -28,7 +28,7 @@ import {
   userMention,
   ModalActionRowComponentBuilder,
 } from '@discordjs/builders';
-import { Collection } from '@discordjs/collection';
+import { Collection, ReadonlyCollection } from '@discordjs/collection';
 import { BaseImageURLOptions, ImageURLOptions, RawFile, REST, RESTOptions } from '@discordjs/rest';
 import {
   APIActionRowComponent,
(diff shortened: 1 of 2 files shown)


kyranet (approved):

vladfrangu (approved):

iCrawl (approved):

SpaceEEC (changes requested):
How do you define "applicable"?

For example `BaseGuildTextChannel#fetchWebhooks` returns a `ReadonlyCollection`, while `BaseGuildTextChannel#fetchInvites` returns a regular `Collection`.

suneettipirneni:
> How do you define "applicable"?
> 
> 
> 
> For example `BaseGuildTextChannel#fetchWebhooks` returns a `ReadonlyCollection`, while `BaseGuildTextChannel#fetchInvites` returns a regular `Collection`.

Looks like blab oversight.

almeidx:
`ReadonlyCollection` should be exported in the typings too

SpaceEEC (changes requested):
There are still quite a few `Collection`s being returned where I'd expect a `ReadonlyCollection` (if I understand this PR correctly).
Most notable all the `fetch*` methods, but also `GuildAuditLogs`' properties as well as `*Collector`'s callbacks.

But it feels like most references of `Collection` in there should be `ReadonlyCollection`s.

PR #7623, opened 2022-03-09
feat: allow `createMessageComponentCollector` without using `fetchReply`

suneettipirneni:

Often times the only reason I use `fetchReply` on `Interaction#reply` is to simply enable component collectors. As a result I never touch any part of the actual fetched message.

This PR enables you to create component collectors without the need to fetch the actual message.

### How does it work?

Previously it was assumed that doing something like this wasn't possible without having the actual message data. However, it actually is possible via comparing interaction ID's.

Every interaction created is given a unique id. Whenever a message component interaction is received it contains a message object with a nested interaction object pointing to the original interaction. This nested interaction object contains the same interaction id. Whenever a reply is sent to an interaction, it attaches the interaction to the message object. We receive this entire structure whenever a message component interaction is fired. **Every replied interaction has an id that is tied to no more than one message**

#### What does this mean?

This means given an interaction id we can tell the message it came from by comparing the message's associated interaction id. Not only that, but we already know the channel the interaction was sent in, so we can also utilize that in our collector.

This removes the need for a network request if all you need to do is create a component collector.

Ergo, this now works:
```ts
client.on('interactionCreate', async interaction => {
  const reply = await interaction.reply('foo'); // No need to `fetchReply`
  const collector = reply.createMessageComponentCollector();
})
```


packages/discord.js/src/structures/InteractionCollector.js
```diff
@@ -6,14 +6,14 @@ const Events = require('../util/Events');
...
- * @property {TextBasedChannelResolvable} [channel] The channel to listen to interactions from
  * @property {ComponentType} [componentType] The type of component to listen for
  * @property {GuildResolvable} [guild] The guild to listen to interactions from
  * @property {InteractionType} [interactionType] The type of interaction to listen for
  * @property {number} [max] The maximum total amount of interactions to collect
  * @property {number} [maxComponents] The maximum number of components to collect
  * @property {number} [maxUsers] The maximum number of users to interact
- * @property {Message|APIMessage} [message] The message to listen to interactions from
+ * @property {Snowflake} [channelId] The channel id of the message interaction response
+ * @property {Snowflake} [messageInteractionId] The message with a given interaction id to listen to interactions from
```
monbrey, line 16: Is this a message id, or an interaction id? The description only confused me further

>The message

So its a message id?

>with a given interaction id 

So I'm giving an interaction id now?

And either way, messageId is still a property, but I don't see where this can be separately provided as an option anymore since you removed the `message` option

packages/discord.js/src/structures/InteractionCollector.js
```diff
@@ -141,6 +147,7 @@ class InteractionCollector extends Collector {
     if (this.interactionType && interaction.type !== this.interactionType) return null;
     if (this.componentType && interaction.componentType !== this.componentType) return null;
     if (this.messageId && interaction.message?.id !== this.messageId) return null;
+    if (this.messageInteractionId && interaction.message?.interaction?.id !== this.messageInteractionId) return null;
```
monbrey, line 150: Alright, so now that I've seen this line it's clear to me that its an interaction id, but my above comments about the description still applies

packages/discord.js/src/structures/PartialInteractionMessage.js
```diff
@@ -0,0 +1,35 @@
+'use strict';
+
+const { InteractionType } = require('discord-api-types/v9');
+
+class PartialInteractionMessage {
```
monbrey, line 5: Totally happy with what this is representing, but `MessageInteraction` isn't even a class, just a typedef for the `message.interaction` property.

This has the naming backwards, `InteractionMessage`, is defined as a "partial" class yet has a method that it's "full" equivalent can't have because it's just an object

monbrey (commented):
Love what this is aiming to achieve and fully support the concept, but I have some concerns with the text and naming conventions.

packages/discord.js/src/structures/InteractionCollector.js
suneettipirneni: Oops I accidentally removed `message` and `channel`

packages/discord.js/src/structures/InteractionCollector.js
```diff
@@ -133,6 +143,7 @@ class InteractionCollector extends Collector {
    * @private
    */
   collect(interaction) {
+    console.log({ got: interaction.message?.interaction?.id, wanted: this.messageInteractionId });
```
muchnameless, line 146:
```suggestion
```

muchnameless (changes requested):

packages/discord.js/src/structures/InteractionResponse.js
```diff
@@ -0,0 +1,61 @@
...
+    /**
+     * The interaction associated with the interaction response
+     * @type {Interaction}
+     */
+    this.interaction = interaction;
+    /**
+     * The id of the original interaction response
+     * @type {Snowflake}
+     */
+    this.responseInteractionId = responseInteractionId ?? interaction.id;
```
kyranet, line 24: `message.interactionResponse.responseInteractionId` sounds... funky, because it contains "interaction response" and then the same but backwards ("response interaction").

Plus the class is also called `InteractionResponse`, so I believe this field should be named just `id`, as it's an interaction response ID.

iCrawl (approved):

packages/discord.js/src/structures/InteractionResponse.js
```diff
@@ -0,0 +1,61 @@
+'use strict';
+
+const { InteractionType } = require('discord-api-types/v9');
```
vladfrangu, line 3:
```suggestion
const { InteractionType } = require('discord-api-types/v10');
```

vladfrangu (changes requested):

kyranet (approved):

packages/discord.js/typings/index.d.ts
```diff
@@ -407,13 +407,13 @@ export abstract class CommandInteraction<Cached extends CacheType = CacheType> e
   public inCachedGuild(): this is CommandInteraction<'cached'>;
   public inRawGuild(): this is CommandInteraction<'raw'>;
   public deferReply(options: InteractionDeferReplyOptions & { fetchReply: true }): Promise<GuildCacheMessage<Cached>>;
-  public deferReply(options?: InteractionDeferReplyOptions): Promise<void>;
+  public deferReply(options?: InteractionDeferReplyOptions): Promise<InteractionResponse>;
```
vladfrangu, line 410: Pass Cached to the response

packages/discord.js/typings/index.d.ts
```diff
@@ -407,13 +407,13 @@ export abstract class CommandInteraction<Cached extends CacheType = CacheType> e
...
   public deleteReply(): Promise<void>;
   public editReply(options: string | MessagePayload | WebhookEditMessageOptions): Promise<GuildCacheMessage<Cached>>;
   public fetchReply(): Promise<GuildCacheMessage<Cached>>;
   public followUp(options: string | MessagePayload | InteractionReplyOptions): Promise<GuildCacheMessage<Cached>>;
   public reply(options: InteractionReplyOptions & { fetchReply: true }): Promise<GuildCacheMessage<Cached>>;
-  public reply(options: string | MessagePayload | InteractionReplyOptions): Promise<void>;
+  public reply(options: string | MessagePayload | InteractionReplyOptions): Promise<InteractionResponse>;
```
vladfrangu, line 416: Ditto

packages/discord.js/typings/index.d.ts
```diff
@@ -1727,17 +1741,17 @@ export class MessageComponentInteraction<Cached extends CacheType = CacheType> e
   public inCachedGuild(): this is MessageComponentInteraction<'cached'>;
   public inRawGuild(): this is MessageComponentInteraction<'raw'>;
   public deferReply(options: InteractionDeferReplyOptions & { fetchReply: true }): Promise<GuildCacheMessage<Cached>>;
-  public deferReply(options?: InteractionDeferReplyOptions): Promise<void>;
+  public deferReply(options?: InteractionDeferReplyOptions): Promise<InteractionResponse>;
```
vladfrangu, line 1744: Ditto

vladfrangu (dismissed):
Basically pass Cached to all InteractionResponses

iCrawl (approved):

PR #7629, opened 2022-03-09
feat: Add `premiumSubscriptionCount` to `InviteGuild`

Jiralite:
Backports the following pull requests to version 13:

- #7451 
- #7641

### DIFF
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -184,6 +184,7 @@ export abstract class AnonymousGuild extends BaseGuild {
   public banner: string | null;
   public description: string | null;
   public nsfwLevel: NSFWLevel;
+  public premiumSubscriptionCount: number | null;
   public splash: string | null;
   public vanityURLCode: string | null;
   public verificationLevel: VerificationLevel;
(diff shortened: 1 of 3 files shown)


typings/index.d.ts
```diff
@@ -184,6 +184,7 @@ export abstract class AnonymousGuild extends BaseGuild {
   public banner: string | null;
   public description: string | null;
   public nsfwLevel: NSFWLevel;
+  public premiumSubscriptionCount: number | null;
```
vladfrangu, line 187: Wait a minute...isn't it optional or number :thonk:

vladfrangu (approved):

typings/index.d.ts
Jiralite: It's [documented as potentially missing](https://discord.com/developers/docs/resources/guild#guild-object) so it looks like there will be a case where this will be `undefined`. Seems like we should handle that.

iCrawl (approved):

kyranet (approved):

SpaceEEC (approved):

PR #7642, opened 2022-03-12
refactor(AutocompleteInteraction): use rest param in respond

imranbarbhuiya:
To make the lib consistent `AutocompleteInteraction#respond` will now take rest param instead of an array.


vladfrangu:
I genuinely don't like this change/move in general, as it makes us have to do a useless spread in our code to send the data over instead of just passing the computed array through, but I'll let the others decide too.

suneettipirneni:
Afaik the only reason we use rest params in builders is because builders are supposed to follow a declarative code style (elements of builders a visually laid out to show its structure). In places that aren't builders, we usually use imperative arrays which works with imperative structures like array literals and/or array variables.

PR #7644, opened 2022-03-12
feat(Collector): add `ignore` event

ImRodry:

This PR adds an option to the collector class with a function to be called when an item is filtered out of the collector, useful when working with interactions and sending error messages for unwanted ones.

### DIFF
--- a/packages/discord.js/src/structures/interfaces/Collector.js
+++ b/packages/discord.js/src/structures/interfaces/Collector.js
@@ -59,10 +59,10 @@ class Collector extends EventEmitter {
     this.collected = new Collection();
 
     /**
-     * The items filtered out by this collector
+     * The items that were not collected by this collector
      * @type {Collection}
      */
-    this.filtered = new Collection();
+    this.ignored = new Collection();
 
     /**
      * Whether this collector has finished collecting
@@ -121,14 +121,14 @@ class Collector extends EventEmitter {
           this._idletimeout = setTimeout(() => this.stop('idle'), this.options.idle).unref();
         }
       } else {
-        this.filtered.set(collectedId, args[0]);
+        this.ignored.set(collectedId, args[0]);
 
         /**
-         * Emitted whenever an element is filtered out of the collector.
-         * @event Collector#filtered
+         * Emitted whenever an element is not collected by the collector.
+         * @event Collector#ignore
          * @param {...*} args The arguments emitted by the listener
          */
-        this.emit('filtered', ...args);
+        this.emit('ignore', ...args);
       }
     }
     this.checkEnd();
--- a/packages/discord.js/src/structures/interfaces/Collector.js
+++ b/packages/discord.js/src/structures/interfaces/Collector.js
@@ -97,9 +97,9 @@ class Collector extends EventEmitter {
    */
   async handleCollect(...args) {
     const collectedId = await this.collect(...args);
-    const filterResult = await this.filter(...args, this.collected);
 
     if (collectedId) {
+      const filterResult = await this.filter(...args, this.collected);
       if (filterResult) {
         this.collected.set(collectedId, args[0]);
 
--- a/packages/discord.js/src/structures/interfaces/Collector.js
+++ b/packages/discord.js/src/structures/interfaces/Collector.js
@@ -59,12 +59,6 @@ class Collector extends EventEmitter {
     this.collected = new Collection();
 
     /**
-     * The items that were not collected by this collector
-     * @type {Collection}
-     */
-    this.ignored = new Collection();
-
-    /**
      * Whether this collector has finished collecting
      * @type {boolean}
      */
@@ -121,8 +115,6 @@ class Collector extends EventEmitter {
           this._idletimeout = setTimeout(() => this.stop('idle'), this.options.idle).unref();
         }
       } else {
-        this.ignored.set(collectedId, args[0]);
-
         /**
          * Emitted whenever an element is not collected by the collector.
          * @event Collector#ignore
--- a/packages/discord.js/src/structures/interfaces/Collector.js
+++ b/packages/discord.js/src/structures/interfaces/Collector.js
@@ -15,17 +15,9 @@ const Util = require('../../util/Util');
  */
 
 /**
- * Function to be called when an item is filtered out of the collector
- * @typedef {Function} CollectorOnFiltered
- * @param {...*} args Any arguments filtered out by the collector
- * @returns {void|Promise<void>}
- */
-
-/**
  * Options to be applied to the collector.
  * @typedef {Object} CollectorOptions
  * @property {CollectorFilter} [filter] The filter applied to this collector
- * @property {CollectorOnFiltered} [onFiltered] The function to call when an item is filtered out of the collector
  * @property {number} [time] How long to run the collector for in milliseconds
  * @property {number} [idle] How long to stop the collector after inactivity in milliseconds
  * @property {boolean} [dispose=false] Whether to dispose data when it's deleted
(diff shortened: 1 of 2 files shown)


almeidx:
Shouldn't this be an event instead?

Jiralite (changes requested):
"Discarded" would be a better word choice here. It would make for a more concise description and wouldn't leave `onFiltered` open to interpretation.

ImRodry:
I disagree. I don't see how `onFiltered` is open for interpretation and it makes sense seeing as the other option is called filter

ImRodry:
> Shouldn't this be an event instead?

Personally I would prefer it to be like this but if the maintainers prefer an event, I can also do that

Jiralite:
> Shouldn't this be an event instead?

I believe this should be an event instead.

I'm also not sure about this idea altogether... the purpose of a collector is to collect items. In here, we're specifically looking at what we're discarding. Seems one should handle this in the filter instead?

ImRodry:
The way I see it is that the filter is only meant to return a boolean and not do anything with the received data. You can technically put all your code in the filter which would then make you question whether a collect event is needed at all (or a filter, depending from the perspective). I feel like having this option would make people’s code more organized as you’re telling the collector what to do when it removes an item from the get-go

monbrey:
+1 for this concept but also +1 for it being an event

ImRodry:
> +1 for this concept but also +1 for it being an event

what should the event be called? `filtered`?

almeidx:
To me at least, "filtered" makes it sound like it would only trigger when things pass the filter (as in are accepted through the filter), as per the definition of the word:
> 1. pass (a liquid, gas, light, or sound) through a device to remove unwanted material.
> "the eye filters out ultraviolet radiation"

I believe @Jiralite's suggestion makes it more clear

BaumianerNiklas:
> what should the event be called? `filtered`?

Regardless of which word gets chosen here, shouldn't it use the infinitive as that's used in every other event name? So `filter` instead of `filtered` (if that's what gets decided on), like how it's `collect` instead of `collected`.

ImRodry:
> > what should the event be called? `filtered`?
> 
> Regardless of which word gets chosen here, shouldn't it use the infinitive as that's used in every other event name? So `filter` instead of `filtered` (if that's what gets decided on), like how it's `collect` instead of `collected`.

I understand the thought process but I don't think we should use that same logic here. We use `collect` for when the collector **collects** something. Here, using the name `filter` for the event could make it seem like it's called when the filter is called, which is not the point. I'll stick to `filtered` for now as this implies the event is called when something is filtered, and not when anything reaches the filter. Will also make it an event in the next commit

BaumianerNiklas:
That's reasonable, though still a huge inconsistency I feel like. I would actually agree with going with something like `discard` as an event name then

ImRodry:
I just don't feel like discard has any correlation with the options on the collector. A filter filters items in and out. Items filtered in are collected and sent through the collect event. Items filtered out are filtered and sent through the filtered event. This is also in line with the meaning of the filtered word when used as an adjective: https://en.wiktionary.org/wiki/filtered#Adjective

monbrey:
The event names are not adjectives. The fact you had to add the additional qualifiers "filtered **in**" and "filtered **out**" demonstrates exactly why "filtered" is not clear enough on its own.

The filter function makes a choice to collect, or discard/ignore/whatever the item. However every single received item must be filtered in order for it to make that decision.

almostSouji:
"filtered" does not fit the general event naming conventions, that would be on "filter".
However, "filter" is also ambiguous in meaning, is the element kept or ignored? Is it filtered for or filtered out?

Discard isn't the right choice either.
It's semantically very hard to discern from "dispose", which also already exists as an event and option.
Discard as well as dispose has a strong connotation to trash, "throwing away" or "getting rid of something" (that was present before). As such it fits the "dispose" event, which removes something from the collected items.

Here, however, we need something to express the opposite of "collect". So what does this actually represent, what happens here.
Well, a collector filters incoming events and either "collect"s the structure or does not collect it. Trying to find the opposite of collecting, the best I can come up with is "ignore". The structure is ignored, not collected.

This clearly sets it apart from the semantics of throwing things away (dispose is no longer confusable with discard) and clearly implies which end of the filter something comes out of. If something is collected it's kept, if something is ignored, it's not collected.

almostSouji:
Separating this part out, because agreement above does not need to reflect this side note as well:
Overall my opinion on an event with this functionality: 👍 
While it can arguably be done with a filter function with side effects right now, I'm personally not a huge fan of side effects, when they can be avoided. Tends to be a pain when debugging and trying to evaluate correctness.

ImRodry:
Applied @almostSouji's suggestion for the name of the event

packages/discord.js/src/structures/interfaces/Collector.js
```diff
@@ -58,6 +58,12 @@ class Collector extends EventEmitter {
      */
     this.collected = new Collection();
 
+    /**
+     * The items that were not collected by this collector
+     * @type {Collection}
+     */
+    this.ignored = new Collection();
```
imranbarbhuiya, line 65: are there any reason why r u storing it in a collection?
bc I don't see any use case here like `collected` collection have which is used in `end` event

packages/discord.js/typings/index.d.ts
```diff
@@ -801,6 +802,7 @@ export abstract class Collector<K, V, F extends unknown[] = []> extends EventEmi
 
   public readonly client: Client;
   public collected: Collection<K, V>;
+  public ignore: Collection<K, V>;
```
imranbarbhuiya, line 805:
```suggestion
  public ignored: Collection<K, V>;
```

imranbarbhuiya (changes requested):

packages/discord.js/typings/index.d.ts
ImRodry: god damn it

packages/discord.js/src/structures/interfaces/Collector.js
ImRodry: For the exact same reason there is a collected collection. It's useless to me, but if someone added it there must be a use, and you may always want to do something with the ignored items at the end, who knows. Either way, the possibility is there

packages/discord.js/src/structures/interfaces/Collector.js
```diff
@@ -96,21 +102,33 @@ class Collector extends EventEmitter {
...
-       */
-      this.emit('collect', ...args);
-
-      if (this._idletimeout) {
-        clearTimeout(this._idletimeout);
-        this._idletimeout = setTimeout(() => this.stop('idle'), this.options.idle).unref();
+    const collectedId = await this.collect(...args);
+    const filterResult = await this.filter(...args, this.collected);
+
+    if (collectedId) {
```
muchnameless, line 108:
```suggestion

    if (collectedId) {
        const filterResult = await this.filter(...args, this.collected);
```
the filter should only be called for items that pass the collect check

muchnameless (changes requested):

packages/discord.js/src/structures/interfaces/Collector.js
imranbarbhuiya: Well, then I'll prefer adding it in the end event then. `collected, ignored, reason`.

kyranet: JFYI, I'm not approving unless this is removed.

This collection is useless, it can and **will** result on extremely large memory usage, for something not a single soul cares about. Entries that are ignored should stay ignored, no mirroring from the collected items, which are something we care about.

ImRodry: You're right, I didn't think of that

Syjalo:
[REQUEST] Add `ignored` collection and add it to the last parameter of `filter` function.
Maybe also provide `collected` and `ignored` collections in `collect` and `ignore` events.

ImRodry:
I had an `ignored` collection in the first commit of this PR but was asked to remove it because it would be taking up memory unnecessarily, especially when receiving a large number of items in a collector. As for the other change I won't comment on that because that would be out of the scope of this PR

kyranet (approved):

BenPatersonxx:
@iCrawl @SpaceEEC please review

SpaceEEC (approved):

iCrawl (approved):

PR #7649, opened 2022-03-13
types: fix regressions

ImRodry:

This PR fixes a bunch of regressions and missed changes from #7584 

### DIFF
--- a/packages/builders/src/components/selectMenu/UnsafeSelectMenu.ts
+++ b/packages/builders/src/components/selectMenu/UnsafeSelectMenu.ts
@@ -11,7 +11,7 @@ export class UnsafeSelectMenuBuilder extends ComponentBuilder<
 	/**
 	 * The options within this select menu
 	 */
-	protected readonly options: UnsafeSelectMenuOptionBuilder[];
+	public readonly options: UnsafeSelectMenuOptionBuilder[];
 
 	public constructor(data?: Partial<APISelectMenuComponent>) {
 		const { options, ...initData } = data ?? {};
--- a/packages/discord.js/src/structures/MessagePayload.js
+++ b/packages/discord.js/src/structures/MessagePayload.js
@@ -3,6 +3,7 @@
 const { Buffer } = require('node:buffer');
 const { isJSONEncodable } = require('@discordjs/builders');
 const { MessageFlags } = require('discord-api-types/v10');
+const ActionRowBuilder = require('./ActionRowBuilder');
 const { RangeError } = require('../errors');
 const DataResolver = require('../util/DataResolver');
 const MessageFlagsBitField = require('../util/MessageFlagsBitField');
@@ -131,9 +132,7 @@ class MessagePayload {
       }
     }
 
-    const components = this.options.components?.map(c =>
-      isJSONEncodable(c) ? c.toJSON() : this.target.client.options.jsonTransformer(c),
-    );
+    const components = this.options.components?.map(c => new ActionRowBuilder(c).toJSON());
 
     let username;
     let avatarURL;
--- a/packages/discord.js/src/structures/MessagePayload.js
+++ b/packages/discord.js/src/structures/MessagePayload.js
@@ -132,7 +132,7 @@ class MessagePayload {
       }
     }
 
-    const components = this.options.components?.map(c => new ActionRowBuilder(c).toJSON());
+    const components = this.options.components?.map(c => (isJSONEncodable(c) ? c : new ActionRowBuilder(c)).toJSON());
 
     let username;
     let avatarURL;
--- a/packages/discord.js/typings/index.d.ts
+++ b/packages/discord.js/typings/index.d.ts
@@ -245,9 +245,6 @@ export class ActionRowBuilder<
 export type MessageActionRowComponent = ButtonComponent | SelectMenuComponent;
 export type ModalActionRowComponent = TextInputComponent;
 
-export type MessageActionRowBuilder = ButtonBuilder | SelectMenuBuilder;
-export type ModalActionRowBuilder = TextInputBuilder;
-
 export class ActionRow<T extends MessageActionRowComponent | ModalActionRowComponent> {
   private constructor(data: APIActionRowComponent<APIMessageActionRowComponent>);
   public readonly components: T[];
--- a/packages/discord.js/typings/index.d.ts
+++ b/packages/discord.js/typings/index.d.ts
@@ -224,7 +224,7 @@ export type ActionRowComponent = MessageActionRowComponent | ModalActionRowCompo
 
 export type ActionRowComponentBuilder = MessageActionRowComponentBuilder | ModalActionRowComponentBuilder;
 
-export interface ActionRowData<T extends ActionRowComponentBuilder | ActionRowComponentData> extends BaseComponentData {
+export interface ActionRowData<T extends ActionRowComponent | ActionRowComponentData> extends BaseComponentData {
   components: T[];
 }
 
@@ -4731,7 +4731,7 @@ export interface MessageEditOptions {
   components?: (
     | JSONEncodable<APIActionRowComponent<APIMessageActionRowComponent>>
     | ActionRowBuilder<MessageActionRowComponentBuilder>
-    | (Required<BaseComponentData> & ActionRowData<MessageActionRowComponentData | ActionRowComponentBuilder>)
+    | (Required<BaseComponentData> & ActionRowData<MessageActionRowComponentData | MessageActionRowComponentBuilder>)
     | APIActionRowComponent<APIMessageActionRowComponent>
   )[];
 }
@@ -4773,7 +4773,7 @@ export interface MessageOptions {
   components?: (
     | JSONEncodable<APIActionRowComponent<APIMessageActionRowComponent>>
     | ActionRowBuilder<MessageActionRowComponentBuilder>
-    | (Required<BaseComponentData> & ActionRowData<MessageActionRowComponentData | ActionRowComponentBuilder>)
+    | (Required<BaseComponentData> & ActionRowData<MessageActionRowComponentData | MessageActionRowComponentBuilder>)
     | APIActionRowComponent<APIMessageActionRowComponent>
   )[];
   allowedMentions?: MessageMentionOptions;
--- a/packages/discord.js/typings/index.d.ts
+++ b/packages/discord.js/typings/index.d.ts
@@ -224,7 +224,7 @@ export type ActionRowComponent = MessageActionRowComponent | ModalActionRowCompo
 
 export type ActionRowComponentBuilder = MessageActionRowComponentBuilder | ModalActionRowComponentBuilder;
 
-export interface ActionRowData<T extends ActionRowComponent | ActionRowComponentData> extends BaseComponentData {
+export interface ActionRowData<T extends ActionRowComponentBuilder | ActionRowComponentData> extends BaseComponentData {
   components: T[];
 }
 
--- a/packages/discord.js/typings/index.d.ts
+++ b/packages/discord.js/typings/index.d.ts
@@ -222,7 +222,9 @@ export type ActionRowComponentData = MessageActionRowComponentData | ModalAction
 
 export type ActionRowComponent = MessageActionRowComponent | ModalActionRowComponent;
 
-export interface ActionRowData<T extends ActionRowBuilder | ActionRowComponentData> extends BaseComponentData {
+export type ActionRowComponentBuilder = MessageActionRowComponentBuilder | ModalActionRowComponentBuilder;
+
+export interface ActionRowData<T extends ActionRowComponentBuilder | ActionRowComponentData> extends BaseComponentData {
   components: T[];
 }
 
@@ -233,7 +235,7 @@ export class ActionRowBuilder<
 > extends BuilderActionRow<T> {
   constructor(
     data?:
-      | ActionRowData<MessageActionRowComponentData | ModalActionRowComponentData | ActionRowBuilder>
+      | ActionRowData<MessageActionRowComponentData | ModalActionRowComponentData | ActionRowComponentBuilder>
       | (Omit<APIActionRowComponent<APIMessageActionRowComponent | APIModalActionRowComponent>, 'type'> & {
           type?: ComponentType.ActionRow;
         }),
@@ -4728,8 +4730,8 @@ export interface MessageEditOptions {
   allowedMentions?: MessageMentionOptions;
   components?: (
     | JSONEncodable<APIActionRowComponent<APIMessageActionRowComponent>>
-    | ActionRowBuilder<MessageActionRowBuilder>
-    | (Required<BaseComponentData> & ActionRowData<MessageActionRowComponentData | ActionRowBuilder>)
+    | ActionRowBuilder<MessageActionRowComponentBuilder>
+    | (Required<BaseComponentData> & ActionRowData<MessageActionRowComponentData | ActionRowComponentBuilder>)
     | APIActionRowComponent<APIMessageActionRowComponent>
   )[];
 }
@@ -4770,8 +4772,8 @@ export interface MessageOptions {
   embeds?: (JSONEncodable<APIEmbed> | APIEmbed)[];
   components?: (
     | JSONEncodable<APIActionRowComponent<APIMessageActionRowComponent>>
-    | ActionRowBuilder<MessageActionRowBuilder>
-    | (Required<BaseComponentData> & ActionRowData<MessageActionRowComponentData | ActionRowBuilder>)
+    | ActionRowBuilder<MessageActionRowComponentBuilder>
+    | (Required<BaseComponentData> & ActionRowData<MessageActionRowComponentData | ActionRowComponentBuilder>)
     | APIActionRowComponent<APIMessageActionRowComponent>
   )[];
   allowedMentions?: MessageMentionOptions;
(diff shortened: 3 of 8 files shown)


almeidx:
Can you fix this too? 
https://github.com/discordjs/discord.js/blob/549716e4fcec89ca81216a6d22aa8e623175e37a/packages/discord.js/src/structures/ModalSubmitInteraction.js#L43

Should be `Components.createComponent(c)` with `const Components = require('../util/Components');`

And the `createComponent()` method itself doesn't support text input components
https://github.com/discordjs/discord.js/blob/549716e4fcec89ca81216a6d22aa8e623175e37a/packages/discord.js/src/util/Components.js#L71-L80

BenjammingKirby:
the `createComponent` function in builders has been renamed to `createComponentBuilder` but in some cases it still uses `createComponent`

packages/discord.js/typings/index.d.ts
```diff
@@ -221,7 +221,7 @@ export type ActionRowComponentData = MessageActionRowComponentData | ModalAction
 
 export type ActionRowComponent = MessageActionRowComponent | ModalActionRowComponent;
 
-export interface ActionRowData<T extends ActionRowComponent | ActionRowComponentData> extends BaseComponentData {
+export interface ActionRowData<T extends ActionRowBuilder | ActionRowComponentData> extends BaseComponentData {
```
vladfrangu, line 224: This is wrong, an action row cannot have action row builders

packages/discord.js/typings/index.d.ts
```diff
@@ -232,7 +232,7 @@ export class ActionRowBuilder<
 > extends BuilderActionRow<T> {
   constructor(
     data?:
-      | ActionRowData<MessageActionRowComponentData | ModalActionRowComponentData | ActionRowComponent>
+      | ActionRowData<MessageActionRowComponentData | ModalActionRowComponentData | ActionRowBuilder>
```
vladfrangu, line 235: Ditto

packages/discord.js/typings/index.d.ts
```diff
@@ -4733,14 +4736,14 @@ export type MessageChannelComponentCollectorOptions<T extends MessageComponentIn
...
-  embeds?: (Embed | APIEmbed)[] | null;
+  embeds?: (JSONEncodable<APIEmbed> | APIEmbed)[] | null;
   files?: (FileOptions | BufferResolvable | Stream | MessageAttachment)[];
   flags?: BitFieldResolvable<MessageFlagsString, number>;
   allowedMentions?: MessageMentionOptions;
   components?: (
     | JSONEncodable<APIActionRowComponent<APIMessageActionRowComponent>>
-    | ActionRow<MessageActionRowComponent>
-    | (Required<BaseComponentData> & ActionRowData<MessageActionRowComponentData | MessageActionRowComponent>)
+    | ActionRowBuilder<MessageActionRowBuilder>
```
vladfrangu, line 4745: do you mean MessageActionRowComponent?

packages/discord.js/typings/index.d.ts
```diff
@@ -4781,8 +4784,8 @@ export interface MessageOptions {
   embeds?: (JSONEncodable<APIEmbed> | APIEmbed)[];
   components?: (
     | JSONEncodable<APIActionRowComponent<APIMessageActionRowComponent>>
-    | ActionRow<MessageActionRowComponent>
-    | (Required<BaseComponentData> & ActionRowData<MessageActionRowComponentData | MessageActionRowComponent>)
+    | ActionRowBuilder<MessageActionRowBuilder>
```
vladfrangu, line 4787: Ditto

vladfrangu (dismissed):

packages/discord.js/typings/index.d.ts
ImRodry: These types are supposed to accept ActionRow builders with MessageActionRowComponentBuilders. Gonna be honest I got a bit confused with the names but I think the next commit should fix all of this. Please re-review if it doesn't

packages/discord.js/typings/index.d.ts
```diff
@@ -4733,14 +4738,14 @@ export type MessageChannelComponentCollectorOptions<T extends MessageComponentIn
...
+  embeds?: (JSONEncodable<APIEmbed> | APIEmbed)[] | null;
   files?: (FileOptions | BufferResolvable | Stream | MessageAttachment)[];
   flags?: BitFieldResolvable<MessageFlagsString, number>;
   allowedMentions?: MessageMentionOptions;
   components?: (
     | JSONEncodable<APIActionRowComponent<APIMessageActionRowComponent>>
-    | ActionRow<MessageActionRowComponent>
-    | (Required<BaseComponentData> & ActionRowData<MessageActionRowComponentData | MessageActionRowComponent>)
+    | ActionRowBuilder<MessageActionRowComponentBuilder>
+    | (Required<BaseComponentData> & ActionRowData<MessageActionRowComponentData | ActionRowComponentBuilder>)
```
vladfrangu, line 4748:
```suggestion
    | (Required<BaseComponentData> & ActionRowData<MessageActionRowComponentData | MessageActionRowComponentBuilder>)
```

packages/discord.js/typings/index.d.ts
```diff
@@ -4781,8 +4786,8 @@ export interface MessageOptions {
   embeds?: (JSONEncodable<APIEmbed> | APIEmbed)[];
   components?: (
     | JSONEncodable<APIActionRowComponent<APIMessageActionRowComponent>>
-    | ActionRow<MessageActionRowComponent>
-    | (Required<BaseComponentData> & ActionRowData<MessageActionRowComponentData | MessageActionRowComponent>)
+    | ActionRowBuilder<MessageActionRowComponentBuilder>
+    | (Required<BaseComponentData> & ActionRowData<MessageActionRowComponentData | ActionRowComponentBuilder>)
```
vladfrangu, line 4790:
```suggestion
    | (Required<BaseComponentData> & ActionRowData<MessageActionRowComponentData | MessageActionRowComponentBuilder>)
```

packages/discord.js/typings/index.d.ts
```diff
@@ -232,7 +234,7 @@ export class ActionRowBuilder<
 > extends BuilderActionRow<T> {
   constructor(
     data?:
-      | ActionRowData<MessageActionRowComponentData | ModalActionRowComponentData | ActionRowComponent>
+      | ActionRowData<MessageActionRowComponentData | ModalActionRowComponentData | ActionRowComponentBuilder>
```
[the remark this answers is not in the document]
ImRodry: This should be taking builders, not components

packages/discord.js/typings/index.d.ts
```diff
@@ -221,7 +221,9 @@ export type ActionRowComponentData = MessageActionRowComponentData | ModalAction
 
 export type ActionRowComponent = MessageActionRowComponent | ModalActionRowComponent;
 
-export interface ActionRowData<T extends ActionRowComponent | ActionRowComponentData> extends BaseComponentData {
+export type ActionRowComponentBuilder = MessageActionRowComponentBuilder | ModalActionRowComponentBuilder;
+
+export interface ActionRowData<T extends ActionRowComponentBuilder | ActionRowComponentData> extends BaseComponentData {
```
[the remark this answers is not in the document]
ImRodry: applied this but this was actually wrong, action row data needs to take a builder, not a received component

packages/discord.js/typings/index.d.ts
```diff
@@ -4781,8 +4786,8 @@ export interface MessageOptions {
   embeds?: (JSONEncodable<APIEmbed> | APIEmbed)[];
   components?: (
     | JSONEncodable<APIActionRowComponent<APIMessageActionRowComponent>>
-    | ActionRow<MessageActionRowComponent>
-    | (Required<BaseComponentData> & ActionRowData<MessageActionRowComponentData | MessageActionRowComponent>)
+    | ActionRowBuilder<MessageActionRowComponentBuilder>
```
suneettipirneni, line 4789:
```suggestion
    | JSONEncodable<APIActionRowComponent<APIMessageActionRowComponent>>
```

packages/discord.js/typings/index.d.ts
```diff
@@ -4781,8 +4786,8 @@ export interface MessageOptions {
...
+    | (Required<BaseComponentData> & ActionRowData<MessageActionRowComponentData | MessageActionRowComponentBuilder>)
```
suneettipirneni, line 4790:
```suggestion
    | (Required<BaseComponentData> & ActionRowData<MessageActionRowComponentData | JSONEncodable<APIActionRowComponent<APIMessageActionRowComponent>>>)
```

suneettipirneni (changes requested):
[code to interfaces, not implementations](https://betterprogramming.pub/code-against-interfaces-not-implementations-37b30e7ab992) moment

packages/discord.js/typings/index.d.ts
```diff
@@ -4733,14 +4738,14 @@ export type MessageChannelComponentCollectorOptions<T extends MessageComponentIn
...
-  embeds?: (Embed | APIEmbed)[] | null;
+  embeds?: (JSONEncodable<APIEmbed> | APIEmbed)[] | null;
   files?: (FileOptions | BufferResolvable | Stream | MessageAttachment)[];
   flags?: BitFieldResolvable<MessageFlagsString, number>;
   allowedMentions?: MessageMentionOptions;
   components?: (
     | JSONEncodable<APIActionRowComponent<APIMessageActionRowComponent>>
-    | ActionRow<MessageActionRowComponent>
-    | (Required<BaseComponentData> & ActionRowData<MessageActionRowComponentData | MessageActionRowComponent>)
+    | ActionRowBuilder<MessageActionRowComponentBuilder>
```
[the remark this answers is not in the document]
ImRodry: this is already an accepted type in the line right above, what

packages/discord.js/typings/index.d.ts
```diff
@@ -4733,14 +4738,14 @@ export type MessageChannelComponentCollectorOptions<T extends MessageComponentIn
...
+    | (Required<BaseComponentData> & ActionRowData<MessageActionRowComponentData | MessageActionRowComponentBuilder>)
```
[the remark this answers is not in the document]
ImRodry: This is incompatible with the type of ActionRowData:
image

packages/discord.js/typings/index.d.ts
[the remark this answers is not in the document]
suneettipirneni: Then `ActionRowBuilder<MessageActionRowComponentBuilder>` can be removed from this union and the other one since it's already assignable to `JSONEncodable<APIActionRowComponent<APIMessageActionRowComponent>>`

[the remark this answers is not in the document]
ImRodry: seems like you're right, will remove that then

packages/discord.js/typings/index.d.ts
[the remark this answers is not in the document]
suneettipirneni: The `ActionRowData` type seems to be incorrect. `ActionRowData` is used in `MessageOptions`, which is currently inaccurate because message options will either snake_case the json, or if it's a class it'll call `toJSON`. So if you pass `ActionRow` data with class components it wouldn't give the correct result. The only time classes are permitted in json is when passing it into an action row builder constructor.

ImRodry:
Rebased and pushed a new commit that fixes a bug in the MessagePayload class that wouldn't allow for objects to be mixed with builder classes

packages/discord.js/src/structures/MessagePayload.js
```diff
@@ -131,9 +132,7 @@ class MessagePayload {
       }
     }
 
-    const components = this.options.components?.map(c =>
-      isJSONEncodable(c) ? c.toJSON() : this.target.client.options.jsonTransformer(c),
-    );
+    const components = this.options.components?.map(c => new ActionRowBuilder(c).toJSON());
```
muchnameless, line 135: if `c` is already an instance of `ActionRowBuilder` this will result in a nested `data` property of `{ type: 1, data: { type: 1 } }` since the constructor isn't designed to take class instances. To fix this you could either do an `instanceof` check or revert to the old behaviour and adjust the typings. It's a question of whether mixing builders and raw action row objects should be supported, if someone wanted to mix them they could always manually call `toJSON` on the component builders. An advantage of `isJSONEncodable` over `instanceof` is that it accepts a wider range of input, e.g. if you `import` a `ComponentBuilder` via esm from the builders package it wouldn't pass the `instanceof`-check since d.js uses the commonjs imports

packages/builders/src/components/selectMenu/UnsafeSelectMenu.ts
```diff
@@ -11,7 +11,7 @@ export class UnsafeSelectMenuBuilder extends ComponentBuilder<
 	/**
 	 * The options within this select menu
 	 */
-	protected readonly options: UnsafeSelectMenuOptionBuilder[];
+	public readonly options: UnsafeSelectMenuOptionBuilder[];
```
muchnameless, line 14: can't comment on other files but there's something similar in [`UnsafeModalBuilder`](https://github.com/discordjs/discord.js/blob/main/packages/builders/src/interactions/modals/UnsafeModal.ts#L9) where you can change `protected readonly data` to `public readonly data`. Also data is being [cloned twice](https://github.com/discordjs/discord.js/blob/main/packages/builders/src/interactions/modals/UnsafeModal.ts#L12-L13) in the other file, the constructor can simply do `this.data = data;`

packages/discord.js/src/structures/MessagePayload.js
```diff
@@ -131,9 +132,7 @@ class MessagePayload {
       }
     }
 
-    const components = this.options.components?.map(c =>
-      isJSONEncodable(c) ? c.toJSON() : this.target.client.options.jsonTransformer(c),
-    );
+    const components = this.options.components?.map(c => (isJSONEncodable(c) ? c : new ActionRowBuilder(c)).toJSON());
```
suneettipirneni, line 135: This is incorrect, `JSONEncodable` means that the object has a `toJSON` method it doesn't mean the object is JSON itself.

```suggestion
    const components = this.options.components?.map(c => (isJSONEncodable(c) ? c.toJSON() : new ActionRowBuilder(c)).toJSON());
```

suneettipirneni (changes requested):

packages/discord.js/src/structures/MessagePayload.js
ImRodry: you should probably read the full thing, this is still calling toJSON on c

suneettipirneni: Ah I see. However I noticed another problem, `ActionRowBuilder` *is* JSONEncodable so only one branch of the condition will ever execute.

ImRodry: what? This is simply creating an action row builder if there isn't one already aka if it isn't a builder instance

suneettipirneni: Yeah, I don't know what I'm on right now. Ignore me

suneettipirneni (approved):

Mysterious-Dev (approved):

packages/discord.js/typings/index.d.ts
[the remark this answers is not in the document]
vladfrangu: If you check, my suggestion is what yours does but with type reuse for the unions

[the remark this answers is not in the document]
ImRodry: ActionRowComponent is a received component, ActionRowComponentBuilder is a builder class, so it's not the same and this suggestion introduces an error. I'll fix it by doing this instead
```suggestion
      | ActionRowData<ActionRowComponentData | ActionRowComponentBuilder>
```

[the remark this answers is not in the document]
ImRodry: did I actually just commit the wrong suggestion omg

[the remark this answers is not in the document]
vladfrangu: Oh I see what you meant now, gotcha

packages/discord.js/typings/index.d.ts
```diff
@@ -4705,14 +4710,13 @@ export type MessageChannelComponentCollectorOptions<T extends MessageComponentIn
 export interface MessageEditOptions {
   attachments?: MessageAttachment[];
   content?: string | null;
-  embeds?: (Embed | APIEmbed)[] | null;
```
vladfrangu, line 4708: Small request, can you add a test that ensures builders are supported? (both the d.js one and /builders one?)

vladfrangu (approved):
Small request, but doesn't block merge

KirmiwDev:
When will this pull be merged? I've been waiting a long time

packages/discord.js/typings/index.d.ts
ImRodry: Will try to do that tomorrow, can't finish it today

Syjalo:
> When will this pull be merged? I've been waiting a long time

When it's ready. Just wait.

iCrawl (approved):

KirmiwDev:
Have you finished rodry?

kyranet (approved):

Mysterious-Dev:
Any news for this ?

skaneprime:
Looks to me That it's ready to go 
Waiting for the last reviewer?

SpaceEEC (approved):

PR #7660, opened 2022-03-14
Update PULL_REQUEST_TEMPLATE.md

Gary50613:

I think checkbox will be better for reviewing PR

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -2,11 +2,9 @@
 
 **Status and versioning classification:**
 
-<!--
-Please move lines that apply to you out of the comment:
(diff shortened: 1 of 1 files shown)


Jiralite:
This is not feasible for the reason listed in #5269:

> GitHub uses checkboxes as todo lists only. We currently use them as checkboxes for versioning classification, which intersects badly with GitHub native features (milestone progress, progress notifications, the progress bar in PR overview), to solve the issue this PR moves classification and optional lines into a comment at the bottom (see next point) to be moved out if applicable.

Gary50613:
Thanks for the response, I thought it would be cool because [microsoft/PowerToys](https://github.com/microsoft/PowerToys/blob/main/.github/pull_request_template.md) also use checkbox for checklists

PR #7684, opened 2022-03-17
feat(CommandInteraction): add support for localized slash commands

suneettipirneni:

Add support for localized names and descriptions in slash command payloads.

### DIFF
--- a/packages/discord.js/src/managers/ApplicationCommandManager.js
+++ b/packages/discord.js/src/managers/ApplicationCommandManager.js
@@ -216,7 +216,9 @@ class ApplicationCommandManager extends CachedManager {
   static transformCommand(command) {
     return {
       name: command.name,
+      name_localizations: command.nameLocalizations,
       description: command.description,
+      description_localizations: command.descriptionLocalizations,
       type: command.type,
       options: command.options?.map(o => ApplicationCommand.transformOption(o)),
       default_permission: command.defaultPermission ?? command.default_permission,
--- a/packages/discord.js/src/managers/ApplicationCommandManager.js
+++ b/packages/discord.js/src/managers/ApplicationCommandManager.js
@@ -216,9 +216,9 @@ class ApplicationCommandManager extends CachedManager {
   static transformCommand(command) {
     return {
       name: command.name,
-      name_localizations: command.nameLocalizations,
+      name_localizations: command.nameLocalizations ?? command.name_localizations,
       description: command.description,
-      description_localizations: command.descriptionLocalizations,
+      description_localizations: command.descriptionLocalizations ?? command.description_localizations,
       type: command.type,
       options: command.options?.map(o => ApplicationCommand.transformOption(o)),
       default_permission: command.defaultPermission ?? command.default_permission,
--- a/packages/discord.js/src/managers/ApplicationCommandManager.js
+++ b/packages/discord.js/src/managers/ApplicationCommandManager.js
@@ -75,7 +75,7 @@ class ApplicationCommandManager extends CachedManager {
    * @typedef {BaseFetchOptions} FetchApplicationCommandOptions
    * @property {Snowflake} [guildId] The guild's id to fetch commands for, for when the guild is not cached
    * @property {LocaleString} [locale] The locale to use when fetching this command
-   * @property {boolean} [withLocalizations] Whether or not to fetch all localization data
+   * @property {boolean} [withLocalizations] Whether to fetch all localization data
    */
 
   /**
--- a/packages/discord.js/src/managers/ApplicationCommandManager.js
+++ b/packages/discord.js/src/managers/ApplicationCommandManager.js
@@ -95,7 +95,18 @@ class ApplicationCommandManager extends CachedManager {
    *   .catch(console.error);
    */
   async fetch(id, { guildId, cache = true, force = false, locale, withLocalizations } = {}) {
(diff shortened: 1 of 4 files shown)


ImRodry:
Discord API docs PR https://github.com/discord/discord-api-docs/pull/4653

ImRodry (changes requested):
Missed the edit methods in ApplicationCommand (also please provide some examples in those)

packages/discord.js/src/managers/ApplicationCommandManager.js
```diff
@@ -216,7 +216,9 @@ class ApplicationCommandManager extends CachedManager {
   static transformCommand(command) {
     return {
       name: command.name,
+      name_localizations: command.nameLocalizations,
       description: command.description,
+      description_localizations: command.descriptionLocalizations,
```
Smidul, line 221: Shouldn't both also check their variants in `snake_case`?

packages/discord.js/src/structures/ApplicationCommand.js
```diff
@@ -190,6 +210,22 @@ class ApplicationCommand extends Base {
...
   }
 
+  /**
+   * Edits the localized names of this ApplicationCommand
+   * @param {Object<string, string>} nameLocalizations The new localized names for the command
+   * @returns {Promise<ApplicationCommand>}
+   * @example
+   * // Edit the name localizations of this command
+   * command.setLocalizedNames({
+   *   'en-US': 'ping',
```
ImRodry, line 220: wouldn't en-US be the one in `name`? This should probably have a different locale

FedorLap2006 (changes requested):
LGTM. However, docs state that behaviour of `GET` endpoints is changed, therefore they won't return `name_localizations` & `description_localizations` by default, instead they will return `name_localized` & `description_localized`, which are not included in the PR. The way this behaviour could be altered is by supplying `?with_localizations=true`. Not sure what's the best solution here though.

suneettipirneni:
> LGTM. However, docs state that behaviour of `GET` endpoints is changed, therefore they won't return `name_localizations` & `description_localizations` by default, instead they will return `name_localized` & `description_localized`, which are not included in the PR. The way this behaviour could be altered is by supplying `?with_localizations=true`. Not sure what's the best solution here though.

I can add it as an option in the `ApplicationCommandManager`

FedorLap2006:
> > LGTM. However, docs state that behaviour of `GET` endpoints is changed, therefore they won't return `name_localizations` & `description_localizations` by default, instead they will return `name_localized` & `description_localized`, which are not included in the PR. The way this behaviour could be altered is by supplying `?with_localizations=true`. Not sure what's the best solution here though.
> 
> I can add it as an option in the `ApplicationCommandManager`

Problem is, when it's set to false (i.e. endpoint provides `localized` fields) you would need to pass an additional header to alter the language. Which I'm not sure where would be the best to put

packages/discord.js/src/structures/ApplicationCommand.js
```diff
@@ -70,6 +80,16 @@ class ApplicationCommand extends Base {
       this.description = data.description;
     }
 
+    if ('description_localizations' in data) {
+      /**
+       * The description localizations for this command
+       * @type {?Object<string, string>}
```
xHyroM, line 86: You can use `LocaleString` from `discord-api-types`

xHyroM (changes requested):

mafineeek (approved):

packages/discord.js/src/managers/ApplicationCommandManager.js
```diff
@@ -92,19 +94,26 @@ class ApplicationCommandManager extends CachedManager {
    *   .then(commands => console.log(`Fetched ${commands.size} commands`))
    *   .catch(console.error);
    */
-  async fetch(id, { guildId, cache = true, force = false } = {}) {
+  async fetch(id, { guildId, cache = true, force = false, locale, withLocalizations } = {}) {
+    const fetchOptions = {
+      headers: {
+        'X-Discord-Locale': locale,
+      },
+      query: withLocalizations ? new URLSearchParams({ with_localizations: withLocalizations }) : undefined,
```
vladfrangu, line 102: Nit
```suggestion
      query: typeof withLocalizations === 'boolean' ? new URLSearchParams({ with_localizations: withLocalizations }) : undefined,
```

vladfrangu (approved):

SpaceEEC (approved):

kyranet (approved):

kyranet:
There are merge conflicts, please update or rebase your branch.

packages/discord.js/src/managers/ApplicationCommandManager.js
```diff
@@ -74,6 +74,8 @@ class ApplicationCommandManager extends CachedManager {
    * Options used to fetch Application Commands from Discord
    * @typedef {BaseFetchOptions} FetchApplicationCommandOptions
    * @property {Snowflake} [guildId] The guild's id to fetch commands for, for when the guild is not cached
+   * @property {LocaleString} [locale] The locale to use when fetching this command
+   * @property {boolean} [withLocalizations] Whether or not to fetch all localization data
```
Jiralite, line 78:
```suggestion
   * @property {boolean} [withLocalizations] Whether to fetch all localization data
```

PR #7686, opened 2022-03-17
feat: add getters to builders

ImRodry:

This PR adds back getters to the builder classes in discord.js that were removed in #7584 as that PR didn't take into account the possibility of creating agnostic functions that accept builder instances as input and modify them according to their props. 
I also fixed the docs in some places related to this (aka component classes)

### DIFF
--- a/packages/discord.js/typings/index.d.ts
+++ b/packages/discord.js/typings/index.d.ts
@@ -17,6 +17,7 @@ import {
   quote,
   roleMention,
   SelectMenuBuilder as BuilderSelectMenuComponent,
+  SelectMenuOptionBuilder as BuilderSelectMenuOption,
   TextInputBuilder as BuilderTextInputComponent,
   UnsafeSelectMenuOptionBuilder as BuildersSelectMenuOption,
   spoiler,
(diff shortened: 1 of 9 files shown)


monbrey:
Why can't the agnostic functions access `<Builder>.data.property`?

ImRodry:
Because those properties are snake_cased

monbrey:
> Because those properties are snake_cased

I might be missing something but that doesn't seem to address why those properties couldn't be used, unless you consider that to no longer be "agnostic"?

ImRodry:
There are no properties in discord.js that are snake_cased, so these shouldn't be any different, it's purely for the sake of consistency

packages/discord.js/typings/index.d.ts
```diff
@@ -541,35 +544,62 @@ export class ButtonComponent extends Component<APIButtonComponent> {
...
 export class TextInputBuilder extends BuilderTextInputComponent {
   public constructor(data?: TextInputComponentData | APITextInputComponent);
   public static from(other: JSONEncodable<APITextInputComponent> | APITextInputComponent): TextInputBuilder;
+  public get label(): string | null;
+  public get style(): TextInputStyle | null;
+  public get maxLength(): number | null;
+  public get minLength(): number | null;
+  public get required(): boolean | null;
+  public get value(): string | null;
+  public get placeholder(): string | null;
```
vitorlops, line 595: It's missing `customId`.

ImRodry:
That was a very personal comment but thank you for that. I understand your points and I believe my code could be adjusted to do what you suggested and it would even look better. I will close this PR now as I understand it is not needed and will open a separate one simply to document the missed typedefs and classes.

ImRodry:
Turns out the comment above addressed one situation while still leaving room for others where this PR is needed. I'm reopening this because other people might face similar issues as I have and this will facilitate their task to upgrade a lot.

suneettipirneni:
> Turns out the comment above addressed one situation while still leaving room for others where this PR is needed.

Which you already were given more-than valid solution(s) too.

ImRodry:
If you could read the full comment I’d appreciate it

suneettipirneni:
Second half of the comment is a moot point considering its basis (the first half) is misleading.

ImRodry:
I'm making this PR to improve the library, not for myself

ImRodry:
@iCrawl would appreciate an explanation on why this was closed seeing as it even had positive reactions

iCrawl:
Because we have no plans to re-introduce them and here's plenty of explanation above.

PR #7698, opened 2022-03-21
feat: support sodium-native lib for voice

ducktrshessami:
I noticed sodium/libsodium-wrappers/tweetnacl aren't being maintained properly, so I figure we should be using an encryption library that *is* being maintained.
~~Though this really isn't my field so it could be entirely unimportant to ensure an up to date encryption library~~

To be more specific about "tested against the Discord API", I ran the code using the [basic voice example](https://github.com/discordjs/discord.js/tree/main/packages/voice/examples/basic), it worked, and `yarn test` also produced no issues.

### DIFF
--- a/packages/voice/src/util/Secretbox.ts
+++ b/packages/voice/src/util/Secretbox.ts
@@ -5,6 +5,31 @@ interface Methods {
 }
 
 const libs = {
+	'sodium-native': (sodium: any): Methods => ({
+		open: (buffer: Buffer, nonce: Buffer, secretKey: Uint8Array) => {
+			if (Boolean(buffer)) {
+				// eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
+				const output = Buffer.allocUnsafe(buffer.length - sodium.crypto_box_MACBYTES);
+				// eslint-disable-next-line @typescript-eslint/no-unsafe-member-access, @typescript-eslint/no-unsafe-call
+				if (sodium.crypto_secretbox_open_easy(output, buffer, nonce, secretKey)) return output;
+			}
+			return null;
+		},
+		close: (opusPacket: Buffer, nonce: Buffer, secretKey: Uint8Array) => {
+			// eslint-disable-next-line @typescript-eslint/no-unsafe-member-access, @typescript-eslint/no-unsafe-argument, @typescript-eslint/restrict-plus-operands
+			const output = Buffer.allocUnsafe(opusPacket.length + sodium.crypto_box_MACBYTES);
+			// eslint-disable-next-line @typescript-eslint/no-unsafe-member-access, @typescript-eslint/no-unsafe-call
+			sodium.crypto_secretbox_easy(output, opusPacket, nonce, secretKey);
+			return output;
+		},
+		random: (n: any, buffer?: Buffer) => {
+			// eslint-disable-next-line @typescript-eslint/no-unsafe-argument
+			if (!buffer) buffer = Buffer.allocUnsafe(n);
+			// eslint-disable-next-line @typescript-eslint/no-unsafe-member-access, @typescript-eslint/no-unsafe-call
+			sodium.randombytes_buf(buffer);
+			return buffer;
+		},
+	}),
 	sodium: (sodium: any): Methods => ({
 		// eslint-disable-next-line @typescript-eslint/no-unsafe-assignment, @typescript-eslint/no-unsafe-member-access
 		open: sodium.api.crypto_secretbox_open_easy,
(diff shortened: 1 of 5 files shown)


packages/voice/src/util/Secretbox.ts
```diff
@@ -5,6 +5,31 @@ interface Methods {
...
+		close: (opusPacket: Buffer, nonce: Buffer, secretKey: Uint8Array) => {
+			// eslint-disable-next-line @typescript-eslint/no-unsafe-member-access, @typescript-eslint/no-unsafe-argument, @typescript-eslint/restrict-plus-operands
+			const output = Buffer.allocUnsafe(opusPacket.length + sodium.crypto_box_MACBYTES);
+			// eslint-disable-next-line @typescript-eslint/no-unsafe-member-access, @typescript-eslint/no-unsafe-call
+			sodium.crypto_secretbox_easy(output, opusPacket, nonce, secretKey);
+			return output;
+		},
+		random: (n: any, buffer?: Buffer) => {
+			// eslint-disable-next-line @typescript-eslint/no-unsafe-argument
+			if (!buffer) buffer = Buffer.allocUnsafe(n);
```
vitorlops, line 27:
```suggestion
		// eslint-disable-next-line @typescript-eslint/no-unsafe-argument
		random: (n: any, buffer: Buffer = Buffer.allocUnsafe(n)) => {
```
I guess you could just do this instead, since if param already has an initializer, it's optional.

ducktrshessami: I just copied it from the sodium implementation down below lol

packages/voice/src/util/Secretbox.ts
```diff
@@ -5,6 +5,30 @@ interface Methods {
 }
 
 const libs = {
+	'sodium-native': (sodium: any): Methods => ({
+		open: (buffer: Buffer, nonce: Buffer, secretKey: Uint8Array) => {
+			if (Boolean(buffer)) {
```
vladfrangu, line 10: Isn't this always going to return true?

packages/voice/src/util/Secretbox.ts
```diff
@@ -5,6 +5,30 @@ interface Methods {
...
+		},
+		close: (opusPacket: Buffer, nonce: Buffer, secretKey: Uint8Array) => {
+			// eslint-disable-next-line @typescript-eslint/no-unsafe-member-access, @typescript-eslint/no-unsafe-argument, @typescript-eslint/restrict-plus-operands
+			const output = Buffer.allocUnsafe(opusPacket.length + sodium.crypto_box_MACBYTES);
+			// eslint-disable-next-line @typescript-eslint/no-unsafe-member-access, @typescript-eslint/no-unsafe-call
+			sodium.crypto_secretbox_easy(output, opusPacket, nonce, secretKey);
+			return output;
+		},
+		// eslint-disable-next-line @typescript-eslint/no-unsafe-argument
+		random: (n: any, buffer: Buffer = Buffer.allocUnsafe(n)) => {
```
vladfrangu, line 26: is `n` really `any` or is it a number?

vladfrangu (dismissed):

packages/voice/src/util/Secretbox.ts
ducktrshessami: there was a test that failed because it doesn't pass a buffer :/

packages/voice/src/util/Secretbox.ts
ducktrshessami: I was definitely just going for keeping the established style of the other implementations, but I'll change it

packages/voice/src/util/Secretbox.ts
ducktrshessami: Yeah there's only the one secret box test, which fails without the null check, so I opted to respect the test. I realize that there probably isn't going to be a situation where a buffer isn't passed during actual run, so I'll change it if you want. At that point I figure an empty buffer should be passed in the test though.

vladfrangu: Out of curiosity, what test are you talking about? 👀

ducktrshessami: [this one](https://github.com/discordjs/discord.js/blob/main/packages/voice/__tests__/Secretbox.test.ts)

vladfrangu: I guess just change this to `if (buffer)`? 🤷 I don't really know what is the expected output when giving wrong inputs..

ducktrshessami:
```
@discordjs/voice:format: \discord.js\packages\voice\src\util\Secretbox.ts
@discordjs/voice:format:   10:8  error  Unnecessary conditional, value is always truthy  @typescript-eslint/no-unnecessary-condition
```
🤷‍♂️
I figured `!!buffer` would look nicer than another 'disable next line' comment, but the linter changed it to bool(buffer)
I'll just do another comment

kyranet (approved):

vladfrangu (approved):

SpaceEEC (approved):

PR #7720, opened 2022-03-27
fix: clarify that packages need to be locally built

MajesticString:


imranbarbhuiya:
We don't need to build for tests. No?

MajesticString:
> We don't need to build for tests. No?

Tests seem to run the build task, but for example, if you just clone the repo and install packages and open packages/discord.js/typings/index.d.ts, vscode will yell at you because a bunch of the dependencies are from the workspace.

vladfrangu (approved):

SpaceEEC (approved):

kyranet (approved):

iCrawl (approved):

PR #7726, opened 2022-03-28
feat(createWebhook): webhook name validation

vitorlops:
This PR adds a webhook name validation as per
- https://github.com/discord/discord-api-docs/pull/4625

### DIFF
--- a/packages/discord.js/src/managers/GuildChannelManager.js
+++ b/packages/discord.js/src/managers/GuildChannelManager.js
@@ -199,7 +199,7 @@ class GuildChannelManager extends CachedManager {
     const invalidNames = ['everyone', 'here'];
     const invalidSubstrings = ['@', '#', ':', '```', 'discord', 'clyde'];
     if (invalidNames.includes(name) || invalidSubstrings.some(c => name.includes(c))) {
-      throw new TypeError('INVALID_WEBHOOK_NAME');
+      throw new TypeError('WEBHOOK_NAME_INVALID');
     }
 
     const data = await this.client.rest.post(Routes.channelWebhooks(id), {
--- a/packages/discord.js/src/managers/GuildChannelManager.js
+++ b/packages/discord.js/src/managers/GuildChannelManager.js
@@ -1,8 +1,8 @@
 'use strict';
 
+const process = require('node:process');
 const { Collection } = require('@discordjs/collection');
 const { ChannelType, Routes } = require('discord-api-types/v10');
-const process = require('node:process');
 const CachedManager = require('./CachedManager');
 const ThreadManager = require('./ThreadManager');
 const { Error, TypeError } = require('../errors');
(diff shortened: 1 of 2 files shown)


packages/discord.js/src/managers/GuildChannelManager.js
```diff
@@ -191,9 +191,17 @@ class GuildChannelManager extends CachedManager {
...
     if (!id) throw new TypeError('INVALID_TYPE', 'channel', 'GuildChannelResolvable');
+
     if (typeof avatar === 'string' && !avatar.startsWith('data:')) {
       avatar = await DataResolver.resolveImage(avatar);
     }
+
+    const invalidNames = ['everyone', 'here'];
+    const invalidSubstrings = ['@', '#', ':', '```', 'discord', 'clyde'];
+    if (invalidNames.includes(name) || invalidSubstrings.some(c => name.includes(c))) {
+      throw new TypeError('INVALID_WEBHOOK_NAME');
```
vladfrangu, line 202: Is this error message registered? If not, it should be

vladfrangu (changes requested):
I don't really know if we want to do this validation (as we don't do..a lot of validation for items that might change at the API level + confusables being usable for bypassing), but this is the only thing I noticed that needs to be fixed if we do go forward with this

packages/discord.js/src/managers/GuildChannelManager.js
almeidx: It is not

vitorlops: Now it is, thanks for pointing this out! :)

vitorlops:
I agree, I was able to create a webhook with one of the substrings that they said I couldn't. So I [commented](https://github.com/discord/discord-api-docs/pull/4625#issuecomment-1084713065) on the upstream PR, let's see if they reply. 👀
The only one that is not working is "clyde", and it's case-insensitive. So I probably need to also make `name` be `name.toLowerCase()` at the `if`.

Markus-Rost:
> The following restrictions are additionally enforced for usernames:
> 
> 1. Usernames cannot contain the following substrings: '@', '#', ':', '```', 'discord'
> 2. Usernames cannot be: 'everyone', 'here'

The validation you added is only enforced on usernames, but webhooks follow only the nickname restriction so besides clyde they are all valid webhook names.

almostSouji:
The error Discord return
```
DiscordAPIError[50035]: Invalid Form Body
name[WEBHOOK_NAME_INVALID_CONTAINS]: Name cannot contain "clyde"
```
is more verbose than the replacement error introduced here and this isn't as common or restrictive enough of an error to warrant library level validation.

vitorlops:
I see, you are right. I am going to close it then. Thanks for the explanation!

PR #7000, opened 2021-11-18
feat(ClientApplication): app authorization links and tags

suneettipirneni:
**Please describe the changes this PR makes and why it should be merged:**

# Number 7000! 🎉

Ref:
- https://github.com/discord/discord-api-docs/pull/4112

**Status and versioning classification:**
- Code changes have been tested against the Discord API, or there are no code changes
- I know how to update typings and have done so, or typings don't need updating
- This PR changes the library's interface (methods or parameters added)

PR #7731, opened 2022-03-29
feat: app authorization links and tags for v13

SuperchupuDev:
Backports #7000 into v13

### DIFF
--- a/src/structures/ClientApplication.js
+++ b/src/structures/ClientApplication.js
@@ -45,6 +45,8 @@ class ClientApplication extends Application {
         scopes: data.install_params.scopes,
         permissions: new Permissions(data.install_params.permissions).freeze(),
       };
+    } else {
+      this.installParams ??= null;
     }
 
     if ('custom_install_url' in data) {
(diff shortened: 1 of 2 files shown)


src/structures/ClientApplication.js
```diff
@@ -23,6 +30,27 @@ class ClientApplication extends Application {
   _patch(data) {
     super._patch(data);
 
+    this.tags = data.tags ?? [];
```
Jiralite, line 33: This isn't documented.

Jiralite (changes requested):

src/structures/ClientApplication.js
SuperchupuDev: What do you mean? It got documented in discord/discord-api-docs#4112 right?

Jiralite: Correct. However, it has no JSDoc here. Without JSDoc, it will not appear in discord.js's documentation. Look at the implementation in #7000 for example:

https://github.com/discordjs/discord.js/blob/0673d38afe63b6f46783f3f7dd7bbe17a0a015cf/packages/discord.js/src/structures/ClientApplication.js#L33-L37

The first 4 lines are the JSDoc that describes the property, but that is not here. This will result in the property not showing in our documentation.

kyranet (approved):

vladfrangu (approved):

SpaceEEC (approved):

PR #7736, opened 2022-04-01
refactor: Remove nickname parsing

Jiralite:

- https://github.com/discord/discord-api-docs/issues/4734#issuecomment-1086156692
- https://github.com/discord/discord-api-docs/commit/a8f7638e79391bb70e15996d1d88488a3b6364b2

### DIFF
--- a/packages/discord.js/src/structures/MessageMentions.js
+++ b/packages/discord.js/src/structures/MessageMentions.js
@@ -19,7 +19,7 @@ class MessageMentions {
    * @type {RegExp}
    * @memberof MessageMentions
    */
-  static UsersPattern = /<@!?(\d{17,19})>/g;
+  static UsersPattern = /<@(\d{17,19})>/g;
 
   /**
    * Regular expression that globally matches role mentions like `<@&297577916114403338>`
--- a/packages/discord.js/src/structures/MessageMentions.js
+++ b/packages/discord.js/src/structures/MessageMentions.js
@@ -19,7 +19,7 @@ class MessageMentions {
    * @type {RegExp}
    * @memberof MessageMentions
    */
-  static UsersPattern = /<@(\d{17,19})>/g;
+  static UsersPattern = /<@!?(\d{17,19})>/g;
 
   /**
    * Regular expression that globally matches role mentions like `<@&297577916114403338>`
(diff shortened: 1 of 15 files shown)


packages/discord.js/src/structures/MessageMentions.js
```diff
@@ -19,7 +19,7 @@ class MessageMentions {
    * @type {RegExp}
    * @memberof MessageMentions
    */
-  static UsersPattern = /<@!?(\d{17,19})>/g;
+  static UsersPattern = /<@(\d{17,19})>/g;
```
imranbarbhuiya, line 22: `<@!id>` is also a valid user mention. The issue is it won't show nickname/username but instead both `<@id>` and `<@!id>` will show the same result.

packages/discord.js/src/util/Util.js
```diff
@@ -526,7 +526,7 @@ class Util extends null {
    */
   static cleanContent(str, channel) {
     str = str
-      .replace(/<@!?[0-9]+>/g, input => {
+      .replace(/<@[0-9]+>/g, input => {
```
imranbarbhuiya, line 529: same

packages/discord.js/src/structures/MessageMentions.js
ImRodry: They made it seem like they were gonna remove it though https://github.com/discord/discord-api-docs/issues/4734#issuecomment-1086156692

splatterxl: No, they're just not going to send it from the client. Making an extremely breaking change like that with such short notice is not really what Discord would do.

ImRodry:
Should also change this
https://github.com/discordjs/discord.js/blob/ac4bc3a6c81f7b86bc5ec2b05069122a601224e7/packages/discord.js/src/structures/GuildMember.js#L431

almostSouji:
Discord is removing it, but does it hurt to just keep it in the parsing section for legacy reasons (parsing messages where this was still the case)? I do agree with removing it from #toString, so d.js doesn't send it anymore, but do not see any benefit in removing it from the parsing regex

vladfrangu:
Agreed with @almostSouji, let's remove it just from toString() (aka make toString() call this.user.toString() :megu:)

kyranet (approved):

vladfrangu (approved):

iCrawl (approved):

SpaceEEC (approved):

PR #7740, opened 2022-04-03
feat(EmbedBuilder): don't unset optional params if not passed

ImRodry:

This PR makes it so if you don't pass a value to Embed setter methods that take in an object that value doesn't get unset and instead keeps its old value.

### DIFF
--- a/packages/builders/src/messages/embed/UnsafeEmbed.ts
+++ b/packages/builders/src/messages/embed/UnsafeEmbed.ts
@@ -83,7 +83,11 @@ export class UnsafeEmbedBuilder {
 			return this;
 		}
 
-		this.data.author = { name: options.name, url: options.url, icon_url: options.iconURL };
+		this.data.author = {
+			name: options.name,
+			url: options.url ?? this.data.author?.url,
+			icon_url: options.iconURL ?? this.data.author?.icon_url,
+		};
 		return this;
 	}
 
@@ -123,7 +127,7 @@ export class UnsafeEmbedBuilder {
 			return this;
 		}
 
-		this.data.footer = { text: options.text, icon_url: options.iconURL };
+		this.data.footer = { text: options.text, icon_url: options.iconURL ?? this.data.footer?.icon_url };
 		return this;
 	}
 


imranbarbhuiya:
What if anyone wants to unset these fields? If they pass `null` or `undefined` when reseting the name, it'll still keep the old properties.

monbrey:
> What if anyone wants to unset these params? If they pass `null` or `undefined` when reseting the name, it'll still keep the old properties.

Yeah, tend to agree here. Passing explicit `null` would/should be the correct way to unset a property

ImRodry:
I'll check for undefined then

vladfrangu:
I don't really approve this PR, these are `set` methods that should set the fields with whatever you pass in, not extend.

ImRodry:
Exactly, they should set the fields with what you pass in, and not change what you don't pass in.

vladfrangu:
Ok, let me clarify. `set*` should overwrite the entire object, not just extend it with new properties you pass in. Besides, whats the point of this? You always have to call setAuthor/setFooter with the name fields present anyways.

suneettipirneni:
This PR also doesn't address combinations where multiple fields are required. For example you should never be able to pass in an author icon url, without an author being defined in EmbedAuthorData. This PR implies all fields are optional.

IMO the correct solution here is to have a nested builder for nested objects that allow constituent parts to be changed Like so:

```ts
<EmbedBuilder>.author.setIconUrl()
```

vladfrangu:
Or...just leave it as is. There's no point in nesting this deeper into miniature builders. Just call `.set*` when you have all the data prepared.

suneettipirneni:
> Or...just leave it as is. There's no point in nesting this deeper into miniature builders. Just call `.set*` when you have all the data prepared.

Yeahhh that's probably the optimal solution tbh

ImRodry:
Actually if #7686 is merged this isn't needed, I just don't know whether that PR will be merged or not

Syjalo:
What about `.patchX()` or `.editX()` methods? Them can edit the X instead of overwrite the X.

vladfrangu:
> What about `.patchX()` or `.editX()` methods? Them can edit the X instead of overwrite the X.

Or just... yknow

```ts
const author = {
	name: '123 look at me',
	icon_url: someDefaultIconUrl(),
};

if (wowMyConditionIsTrue) {
	author.icon_url = someFancierIconUrl();
}

embed.setAuthor(author);
```

There is no point in bloating the builder code for things like these.

ImRodry:
That doesn’t work on agnostic functions… I just want to be able to change the footer text without changing the iconURL

vladfrangu:
1. I have no idea what you're doing that cannot be adapted to our builders. 
2. Embed#data is public unlike other builders (although the fact its a data object is a separate discussion)

ImRodry:
It’s public but it’s snake_cased, so either djs should have a getter on top of it or it should be made private IMO

vladfrangu:
We've discussed about this, and we'll be rejecting this pull request *in its current state*. Changing the behavior of `set*` methods would be a misnomer to what the method is actually supposed to do - set the object to whatever you pass into it.

As a workaround, you can always call `.setFooter({ ...embed.footer, name: 'overwritten' })` instead in your code (or `...embed.data.footer` if its nested) and you'll achieve the exact same behavior!

With that said, what you're requesting is more of a `update*` method (par example `.updateFooter`). Feel free to open a PR for that instead! :D (or alter this one and follow up, and I'll reopen it for you)

ImRodry:
> As a workaround, you can always call .setFooter({ ...embed.footer, name: 'overwritten' }) instead in your code (or ...embed.data.footer if its nested) and you'll achieve the exact same behavior!

I believe the types wouldn't quite work since you could be mixing snake_case with camelCase here but I'd need to check, gonna admit I didn't think of using the spread operator

> With that said, what you're requesting is more of a update* method (par example .updateFooter). Feel free to open a PR for that instead! :D (or alter this one and follow up, and I'll reopen it for you)

good idea, might PR this later on but I'll open a new PR if that does happen, thanks for the feedback!

PR #7744, opened 2022-04-04
feat: add `makeURLSearchParams` utility function

kyranet:

When using `URLSearchParams`'s constructor, we had to be careful not to pass nullish values, and furthermore, TypeScript complains if we use non-string values in the constructor as it's typed as the record types are typed as `string[][]` and `Record<string, string>`.

The `makeURLSearchParams` is an utility that safely serializes the data, stripping out invalid values (undefined, null, symbol, function, objects without toString override, etc).

### DIFF
--- a/packages/discord.js/src/client/Client.js
+++ b/packages/discord.js/src/client/Client.js
@@ -2,6 +2,7 @@
 
 const process = require('node:process');
 const { Collection } = require('@discordjs/collection');
+const { makeURLSearchParams } = require('@discordjs/rest');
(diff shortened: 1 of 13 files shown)


vladfrangu (approved):

ImRodry:
Shouldn’t it be makeURLSearchParams?

kyranet:
... yes, good catch, I'll fix once I'm on PC

SpaceEEC (approved):

iCrawl (approved):

PR #7747, opened 2022-04-05
feat(rest): use undici

KhafraDev:

Transfers from `node-fetch` and `form-data` to `undici`.

**WIP:**
- [x] Finish re-writing tests and remove `nock`.
- [x] Report a bunch of mock bugs I've found with undici.
- [x] Use `undici.request` ~~when a `FormData` body isn't being sent~~ because it's way faster.
- [x] Replace `postFile` tests

- Code changes have been tested against the Discord API, or there are no code changes (WIP)
- This PR includes breaking changes (methods removed or renamed, parameters moved or removed) (just to be safe)


ckohen:
oh good, I was literally gonna start working on this in like 20 minutes, I can't write directly from here so I'll PR tests to your fork

KhafraDev:
@ckohen (or anyone else) if you can figure out why https://github.com/discordjs/discord.js/blob/9c118a9aa55b09ebdd03c446572dcd318787199a/packages/rest/__tests__/RequestHandler.test.ts#L319 test causes the 429 cloudfare test to fail it would be greatly appreciated

packages/rest/__tests__/RequestHandler.test.ts
```diff
@@ -306,57 +300,199 @@ test('Handle sublimits', async () => {
...
-	const unexepectedSublimit = api.get('/unexpected').then((res) => {
-		firstResolvedTime = performance.now();
-		return res;
-	});
-	const queuedSublimit = api.get('/unexpected').then((res) => {
-		secondResolvedTime = performance.now();
-		return res;
-	});
+if (Boolean(false))
+	test('Handle unexpected 429', async () => {
```
vladfrangu, line 320: Use test.skip for now with a todo

packages/rest/__tests__/RequestManager.test.ts
```diff
@@ -1,9 +1,21 @@
...
+setGlobalDispatcher(mockAgent);
+
+const mockPool = mockAgent.get('https://discord.com');
+
+mockPool
+	.intercept({
+		path: genPath('/simpleGet'),
+		method: 'GET',
+	})
+	.reply(200, 'Well this is awkward...');
```
vladfrangu, line 18: Should have the same beforeAll/afterAll or beforeEach/afterEach

packages/rest/__tests__/REST.test.ts
```diff
@@ -1,209 +1,236 @@
...
+const res = await fetch('http://localhost:3000/foo');
+
+console.log(res.status, await res.text()); // 200 foo
+*/
+
+const mockAgent = new MockAgent();
+mockAgent.disableNetConnect(); // don't accidentally make requests to Discord
+setGlobalDispatcher(mockAgent); // this allows the Agent to intercept requests
+
+const mockPool = mockAgent.get('https://discord.com');
```
vladfrangu, line 37: Ditto with before/afterEach/all

packages/rest/src/lib/RequestManager.ts
```diff
@@ -372,15 +372,28 @@ export class RequestManager extends EventEmitter {
...
-				formData.append(file.key ?? `files[${index}]`, file.data, file.name);
+				const fileKey = file.key ?? `files[${index}]`;
+
+				// https://developer.mozilla.org/en-US/docs/Web/API/FormData/append#parameters
+				// FormData.append only accepts a string or Blob.
+				// https://developer.mozilla.org/en-US/docs/Web/API/Blob/Blob#parameters
+				// The Blob constructor accepts TypedArray/ArrayBuffer, strings, and Blobs.
+				if (typeof file.data === 'string') {
+					formData.append(fileKey, file.data, file.name);
+				} else if (Buffer.isBuffer(file.data)) {
```
vladfrangu, line 383: This won't actually match ArrayBuffers or TypedArrays, just nodejs buffers

packages/rest/src/lib/RequestManager.ts
```diff
@@ -406,7 +417,7 @@ export class RequestManager extends EventEmitter {
 
 		const fetchOptions = {
 			agent: this.agent,
-			body: finalBody,
+			body: finalBody ?? null,
```
vladfrangu, line 420: Won't this cause issues?

packages/rest/src/lib/RequestManager.ts
```diff
@@ -389,8 +402,6 @@ export class RequestManager extends EventEmitter {
 
 			// Set the final body to the form data
 			finalBody = formData;
-			// Set the additional headers to the form data ones
-			additionalHeaders = formData.getHeaders();
```
vladfrangu, line 393: Is this not needed anymore? 👀 Does form-data still work?

vladfrangu (changes requested):

packages/rest/src/lib/RequestManager.ts
KhafraDev: Undici exports its own `FormData`, is there a requirement on form-data?

packages/rest/src/lib/RequestManager.ts
KhafraDev: Undici's `RequestInit` type doesn't include `undefined` as a value;

```
Type '{ agent: httpAgent; body: string | ArrayBuffer | AsyncIterable<Uint8Array> | Blob | FormData | Iterable<Uint8Array> | DataView | URLSearchParams | null | undefined; headers: Record<...>; method: RequestMethod; }' is not assignable to type 'RequestInit' with 'exactOptionalPropertyTypes: true'. Consider adding 'undefined' to the types of the target's properties.
  Types of property 'body' are incompatible.
    Type 'string | ArrayBuffer | AsyncIterable<Uint8Array> | Blob | FormData | Iterable<Uint8Array> | DataView | URLSearchParams | null | undefined' is not assignable to type 'BodyInit'.
      Type 'undefined' is not assignable to type 'BodyInit'.ts(2375)
[RequestManager.ts(324, 67): ]()The expected type comes from property 'fetchOptions' which is declared here on type '{ url: string; fetchOptions: RequestInit; }'
```

vladfrangu: It does but doesn't 🥲
Try setting it to the manual undefined property?

packages/rest/src/lib/RequestManager.ts
KhafraDev: https://github.com/discordjs/discord.js/blob/402514ff323ccf1f8c95d295f044cf0bb5547c2e/packages/rest/src/lib/RequestManager.ts#L30

packages/rest/src/lib/RequestManager.ts
vladfrangu: No no as in do the headers get automatically added for content type when form data is sent? Aka does this change work with the api

packages/rest/src/lib/RequestManager.ts
KhafraDev: I would *love* to enable TypedArray/ArrayBuffer/Blob/File, etc in the future though. Just not in this PR Lol

packages/rest/__tests__/REST.test.ts
KhafraDev: 87c91bbe277766860497ec34876caeaabb39d40b

packages/rest/__tests__/RequestManager.test.ts
KhafraDev: 87c91bbe277766860497ec34876caeaabb39d40b

packages/rest/__tests__/RequestHandler.test.ts
KhafraDev: 87c91bbe277766860497ec34876caeaabb39d40b

packages/rest/src/lib/RequestManager.ts
vladfrangu: Thats fair

packages/rest/src/lib/RequestManager.ts
KhafraDev: Yes, undici does all of this for us 😄

packages/rest/src/lib/RequestManager.ts
KhafraDev: fixed in 405b6f050fa0275572b9de8988807966cb1548dd

packages/rest/src/lib/RequestManager.ts
```diff
@@ -406,12 +415,15 @@ export class RequestManager extends EventEmitter {
...
 		const fetchOptions = {
 			agent: this.agent,
-			body: finalBody,
 			// eslint-disable-next-line @typescript-eslint/consistent-type-assertions
 			headers: { ...(request.headers ?? {}), ...additionalHeaders, ...headers } as Record<string, string>,
 			method: request.method,
 		};
 
+		if (finalBody !== undefined) {
+			Reflect.set(fetchOptions, 'body', finalBody);
```
almeidx, line 424: Assuming the reason you're using `Reflect.set()` here is because of the type, can't you just explicitly type `fetchOptions` as `RequestInit`?

KhafraDev: This is my lazy temporary solution because undici's fetch doesn't support an `agent` option.

KhafraDev: fixed in d9dd20e55606643ffab1bf88a7802815458d9bd6

packages/rest/__tests__/RequestHandler.test.ts
```diff
@@ -330,33 +367,132 @@ test('Handle unexpected 429', async () => {
...
-	const promise = api.get('/slow');
-	await expect(promise).rejects.toThrowError('The user aborted a request.');
-}, 10000);
+	mockPool
+		.intercept({
+			path: genPath('/slow'),
+			method: 'GET',
+		})
+		.reply(200, '')
+		.delay(8000)
```
ckohen, line 480: This is....*very* slow, waiting for the old test already felt like ages, lets not wait longer (remember we try twice). If needed you can add another rest handler that has a much shorter timeout

packages/rest/src/lib/RequestManager.ts
```diff
@@ -291,7 +288,7 @@ export class RequestManager extends EventEmitter {
 			this.handlers.get(`${hash.value}:${routeId.majorParameter}`) ??
 			this.createHandler(hash.value, routeId.majorParameter);
 
-		// Resolve the request into usable fetch/node-fetch options
+		// Resolve the request into usable fetch
```
ckohen, line 291:
```suggestion
		// Resolve the request into usable fetch options
```

packages/rest/src/lib/RequestManager.ts
```diff
@@ -404,14 +406,17 @@ export class RequestManager extends EventEmitter {
...
+			body: finalBody!,
 			// eslint-disable-next-line @typescript-eslint/consistent-type-assertions
 			headers: { ...(request.headers ?? {}), ...additionalHeaders, ...headers } as Record<string, string>,
 			method: request.method,
 		};
 
+		if (finalBody !== undefined) {
+			Reflect.set(fetchOptions, 'body', finalBody);
+		}
+
```
ckohen, line 419:
```suggestion
```

Looks like you forgot to remove this

packages/rest/src/lib/RequestManager.ts
```diff
@@ -404,14 +406,17 @@ export class RequestManager extends EventEmitter {
 			}
 		}
 
-		const fetchOptions = {
-			agent: this.agent,
-			body: finalBody,
+		const fetchOptions: RequestInit = {
+			body: finalBody!,
```
ckohen, line 410: This assertion is under false pretense at first glance, you should leave a comment here explaining its actual purpose

packages/rest/src/lib/handlers/SequentialHandler.ts
```diff
@@ -1,6 +1,6 @@
 import { setTimeout as sleep } from 'node:timers/promises';
 import { AsyncQueue } from '@sapphire/async-queue';
-import fetch, { RequestInit, Response } from 'node-fetch';
+import { fetch, RequestInit, Response } from 'undici';
```
ckohen, line 3:
```suggestion
import { fetch, type RequestInit, type Response } from 'undici';
```

packages/rest/src/lib/handlers/SequentialHandler.ts
```diff
@@ -303,10 +303,7 @@ export class SequentialHandler implements IHandler {
 		let res: Response;
 
 		try {
-			// node-fetch typings are a bit weird, so we have to cast to any to get the correct signature
-			// Type 'AbortSignal' is not assignable to type 'import("discord.js-modules/node_modules/@types/node-fetch/externals").AbortSignal'
-			// eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
```
ckohen, line 308: 🧹

ckohen (changes requested):
Looks like you didn't need me to write tests after all 😌

Just a few nits, also looks like you should regen your lockfile, it still has the collection version change in it.

packages/rest/src/lib/RequestManager.ts
KhafraDev: Coincidentally fixed in c6d27d055d82503ecfb4b6282f9cd5166698b98e

packages/rest/src/lib/RequestManager.ts
KhafraDev: Coincidentally fixed in c6d27d055d82503ecfb4b6282f9cd5166698b98e (again lol)

ckohen:
That timing was uncanny, that newest commit resolves some of what I just requested

packages/rest/__tests__/RequestHandler.test.ts
KhafraDev: That is the only timeout that works 100%, otherwise the test was flaky maybe 30-40% of the time.

KhafraDev:
I had those changes done for a few hours and was waiting to commit it all at once with a larger commit 😄

packages/rest/__tests__/RequestHandler.test.ts
ckohen: what i mean is add another api handler with timeout: 250 set in the options so that you can decrease this

packages/rest/src/lib/handlers/SequentialHandler.ts
KhafraDev: fixed in c8fdd310579c6473fd5f0423c04dd636ec2d5fce

packages/rest/src/lib/RequestManager.ts
KhafraDev: fixed in c8fdd310579c6473fd5f0423c04dd636ec2d5fce

packages/rest/__tests__/RequestHandler.test.ts
KhafraDev: that makes sense, will do

KhafraDev: done in fa21eb416ac1f80f349e1be0be3d178b9fb1b3f2

KhafraDev:
Added an option to set an undici Agent, this allows for using [proxies](https://github.com/nodejs/undici/blob/e780b16c1e8e75011e17fab0d0aa38ccedfe4356/docs/api/ProxyAgent.md#class-proxyagent) and more.

Solves https://github.com/discordjs/discord.js/issues/3039 without 'patching' discord.js too. 👀

KhafraDev:
Now the library will have a massive increase in req/second, but is now a pretty major change.

Also, tests are failing due to a Jest bug. This is why no one recommends using jest.

packages/rest/src/lib/REST.ts
```diff
@@ -233,6 +233,10 @@ export class REST extends EventEmitter {
 		this.on('removeListener', (name, listener) => {
 			if (name === RESTEvents.Request || name === RESTEvents.Response) this.requestManager.off(name, listener);
 		});
+
+		if (options.agent) {
+			setGlobalDispatcher(options.agent);
```
vladfrangu, line 238: Won't this cause issues? Should it be global? I'd argue the agent should only affect requests done through the module

packages/rest/src/lib/RequestManager.ts
```diff
@@ -404,14 +407,25 @@ export class RequestManager extends EventEmitter {
...
+			additionalHeaders['content-type'] = resolvedBody[1]; // multipart header
+		} else {
+			finalBody = resolvedBody;
+		}
+
+		const fetchOptions: RequestOptions = {
 			// eslint-disable-next-line @typescript-eslint/consistent-type-assertions
 			headers: { ...(request.headers ?? {}), ...additionalHeaders, ...headers } as Record<string, string>,
-			method: request.method,
+			method: `${request.method}` as Dispatcher.HttpMethod,
```
vladfrangu, line 422: Any reason for `${}`? Can't you just cast it?

packages/rest/src/lib/handlers/SequentialHandler.ts
```diff
@@ -329,17 +326,18 @@ export class SequentialHandler implements IHandler {
 					data: requestData,
 					retries,
 				},
-				res.clone(),
+				{ ...res },
```
vladfrangu, line 329: Does this do the same? Will this allow using the body twice?

packages/rest/src/lib/handlers/SequentialHandler.ts
```diff
@@ -468,24 +466,24 @@ export class SequentialHandler implements IHandler {
...
 			return this.runRequest(routeId, url, options, requestData, retries);
-		} else if (res.status >= 500 && res.status < 600) {
+		} else if (status >= 500 && status < 600) {
 			// Retry the specified number of times for possible server side issues
 			if (retries !== this.manager.options.retries) {
 				return this.runRequest(routeId, url, options, requestData, ++retries);
 			}
 			// We are out of retries, throw an error
-			throw new HTTPError(res.statusText, res.constructor.name, res.status, method, url, requestData);
+			throw new HTTPError(status.toString(), res.constructor.name, status, method, url, requestData);
```
vladfrangu, line 475: statusText is now redundant so maybe we should remove it if undici doesn't return anything similar

packages/rest/src/lib/RequestManager.ts
```diff
@@ -404,14 +407,25 @@ export class RequestManager extends EventEmitter {
 			}
 		}
 
-		const fetchOptions = {
-			agent: this.agent,
-			body: finalBody,
+		const resolvedBody = await resolveBody(finalBody);
+
+		if (Array.isArray(resolvedBody)) {
```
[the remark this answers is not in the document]
KhafraDev: `undici.fetch` works with FormData, `undici.request` doesn't :(

packages/rest/src/lib/handlers/SequentialHandler.ts
KhafraDev: with `undici.request` the body is a ReadableStream so I'm not sure how it could be cloned to the same degree tbh

packages/rest/src/lib/REST.ts
KhafraDev: Setting is per request is possible with `undici.request` but causes issues with the tests because they use an Agent too... I'll get around to fixing this however.

packages/rest/src/lib/RequestManager.ts
KhafraDev: Left overs from before I changed the enum to be uppercase

packages/rest/src/lib/REST.ts
KhafraDev: This is much easier to fix than I previously thought.

packages/rest/src/lib/handlers/SequentialHandler.ts
ckohen: I don't think this works, you're just copying a reference to the readable stream mat this point.

Because .body is readonly you also can't just pipe body to two passthroughs and replace body on the original response, though I suppose you could try reassigning res to { ...res, body: bodyStreamA } and use { ...res, body: bodyStreamB } as the emitted value (slowdowns are okay here, that's why we check for the listener)

packages/rest/src/lib/REST.ts
```diff
@@ -233,10 +233,6 @@ export class REST extends EventEmitter {
 		this.on('removeListener', (name, listener) => {
 			if (name === RESTEvents.Request || name === RESTEvents.Response) this.requestManager.off(name, listener);
 		});
-
-		if (options.agent) {
```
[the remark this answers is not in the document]
KhafraDev: So you want an option to set a "global" agent (global in the sense that every request done for this REST instance will have this default agent) and an option to set per-request agents?

packages/rest/src/lib/handlers/SequentialHandler.ts
KhafraDev: It doesn't work, but working with typescript and cloning streams doesn't work great :/. If you can work it out and send a PR to my fork I'd be more than happy to bring it in 😉

packages/rest/src/lib/RequestManager.ts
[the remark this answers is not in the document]
KhafraDev: Will be fixed by https://github.com/nodejs/undici/pull/1332/

packages/rest/src/lib/REST.ts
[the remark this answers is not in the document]
vladfrangu: Yep! Just like how auth works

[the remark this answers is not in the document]
KhafraDev: Done in 8cac0014d0b221f03af59710641a29043b252b4f!

packages/rest/src/lib/handlers/SequentialHandler.ts
ckohen: uggh right, this isn't a *normal* readable, its undicis bodyreadable which isn't even exported and its not a web stream so can't just tee it. We'd have to reimplement undici's BodyReadable but as a BodyPassthrough in order to do this really

KhafraDev: We can use `stream/consumers` for the BodyMixin stuff, the main problem is getting typescript to like it

ckohen: I'll take a look at that later today. 

This whole thing is making me wonder if it's really worth it to use undicis proprietary request over the spec fetch implementation. Yes it's faster, but at some point the fetch api provides enough niceties that it's worth the difference (also should a discord bot ever be making that many req/s anyways? Unless you're proxying all requests from a sharded bot to a single instance). Just not sure if we're at that point or not

KhafraDev: This seems to be the only troublesome part, and `node-fetch`'s body was a ReadableStream as well (their cloning doesn't work great either). Using `request` also gives us a lower-leveled implementation that is much more customizable; for example, users of the library can now use proxies ( #3039 ), set custom request timeouts on header/body parsing, logging (using diagnostic channel), and much more. With fetch, you're limited to what the spec allows, and it isn't much...

ckohen: It sounds like we lose a good bit of that stuff if we clone it ourselves

Maybe undici would be open to adding a clone function to responsedata so that types make a whole lot more sense

Otherwise I'll look at how node-fetch did it (I think I have a pretty good guess) and see if we can apply it here

KhafraDev: Thinking it over... we can remove these events entirely, as undici provides diagnostic_channel support. Here's an example of getting the body sent to a server:

```js
import diagnosticsChannel from 'diagnostics_channel'
import { request } from 'undici';

diagnosticsChannel.channel('undici:request:create').subscribe(({ request }) => {
  console.log(request.body.toString()); // 'Hello, world!'
});

await request('http://httpbin.org/post', {
  method: 'POST',
  headers: {
    'hello': 'world'
  },
  body: 'Hello, world!'
})

```

KhafraDev: I've gone ahead and removed `request` and `response` events,
request - https://github.com/nodejs/undici/blob/main/docs/api/DiagnosticsChannel.md#undicirequestcreate
response - https://github.com/nodejs/undici/blob/main/docs/api/DiagnosticsChannel.md#undicirequesttrailers

ckohen: trailers does not satisfy the same functionality that our response event did. Undici only provides the trailers, not the returned body (because that would involve cloning it, an inherently slow operation. This event was only really useful for when there could potentially be something wrong with djs and we needed to see the raw api response body that was causing it.

KhafraDev: I think they'd accept a PR, considering that it would only be cloned if the channel is being subscribed to.

KhafraDev: The body is sent in the diagnostic channel, although not cloned & hidden via a symbol. Should theoretically be possible to clone and expose it in this event,

ckohen: yeah, took a look at the code, and the *very* indirect way to get the body for now (and it should be since its not cloned) and I don't mind it. This seems like something that undici should probably have anyways, so we shouldn't stall this PR on that basis.

ckohen:
With the global agent settable, we can simplify tests a little bit, got a PR coming your way with a sample of what that would look like (and a couple other tests for coverage purposes)

.vscode/settings.json
```diff
@@ -5,5 +5,8 @@
 	"editor.codeActionsOnSave": {
 		"source.fixAll": true,
 		"source.organizeImports": false
+	},
+	"files.exclude": {
+		"node_modules": true
```
ckohen, line 10: Up to crawl, but I don't think this should be added

KhafraDev: This was added just to make it easier on my end, was going to remove it (and revert the collection dep thing) once everything was good. 👍

vladfrangu:
This needs a world wide ~web~ rebase

.vscode/settings.json
vladfrangu: Personally I don't think it hurts to have, but I'll let Crawl decide too :D

packages/rest/src/lib/RequestManager.ts
```diff
@@ -404,14 +425,32 @@ export class RequestManager extends EventEmitter {
...
+		if (finalBody !== undefined) {
+			fetchOptions.body = finalBody as Exclude<RequestOptions['body'], undefined>;
+		}
+
+		// Prioritize setting an agent per request, use the agent for this instance otherwise.
+		if (request.dispatcher) {
+			fetchOptions.dispatcher = request.dispatcher;
+		} else if (this.agent) {
+			fetchOptions.dispatcher = this.agent;
+		}
```
vladfrangu, line 452:
```suggestion
		fetchOptions.dispatcher = request.dispatcher ?? this.agent;
```

vladfrangu (changes requested):

packages/rest/src/lib/REST.ts
```diff
@@ -189,8 +153,6 @@ export interface RestEvents {
 	invalidRequestWarning: [invalidRequestInfo: InvalidRequestWarningData];
 	restDebug: [info: string];
 	rateLimited: [rateLimitInfo: RateLimitData];
-	request: [request: APIRequest];
```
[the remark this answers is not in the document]
KhafraDev: request - https://github.com/nodejs/undici/blob/main/docs/api/DiagnosticsChannel.md#undicirequestcreate
response - https://github.com/nodejs/undici/blob/main/docs/api/DiagnosticsChannel.md#undicirequesttrailers

Not the exact same but close enough

packages/rest/src/lib/REST.ts
```diff
@@ -226,13 +191,22 @@ export class REST extends EventEmitter {
...
-			if (name === RESTEvents.Request || name === RESTEvents.Response) this.requestManager.on(name, listener);
-		});
-		this.on('removeListener', (name, listener) => {
-			if (name === RESTEvents.Request || name === RESTEvents.Response) this.requestManager.off(name, listener);
-		});
+	/**
+	 * Gets the agent set for this instance
+	 */
+	public getAgent() {
+		return this.requestManager.agent;
```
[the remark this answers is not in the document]
KhafraDev: No it isn't

packages/rest/src/lib/RequestManager.ts
```diff
@@ -53,6 +52,10 @@ export interface RequestData {
 	 * If providing as BodyInit, set `passThroughBody: true`
 	 */
 	body?: BodyInit | unknown;
+	/**
+	 * The [Agent]({@link https://undici.nodejs.org/#/docs/api/Agent}) to use for the request.
```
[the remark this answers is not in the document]
KhafraDev: I don't know how the jsdoc parser works, I just made sure it worked with intellisense so you're probably correct

packages/rest/src/lib/REST.ts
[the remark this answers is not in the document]
KhafraDev: I'll make an issue on their repo to see if they could support cloning somehow (or at least expose the un-cloned stream via a public property).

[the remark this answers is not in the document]
KhafraDev: I found a potential solution, just waiting now to see.

packages/rest/src/lib/RequestManager.ts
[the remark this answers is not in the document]
KhafraDev: Fixed in 2fe1c6c3598e1c704496ceff1738b6046deed631 (finally!)

KhafraDev:
This PR is ready. Only thing missing is a real replacement to the `response` event but one of the undici maintainers didn't like my proposed solution to sending a consumed body (see: https://github.com/nodejs/undici/pull/1369). The event itself was re-added in 483719a1b3eddb9454cf2e3f2b5b92ce44b040fc.

However, all the bugs I found were fixed and all the tests are now running. Please merge so I don't have to maintain this pr permanently 🙏.

iCrawl (approved):

packages/rest/package.json
```diff
@@ -50,14 +50,12 @@
 	},
 	"homepage": "https://discord.js.org",
 	"dependencies": {
-		"@discordjs/collection": "workspace:^",
```
iCrawl, line 53: Will need to revert this.

KhafraDev: done

packages/rest/src/lib/RequestManager.ts
```diff
@@ -203,6 +203,7 @@ export class RequestManager extends EventEmitter {
 		this.options = { ...DefaultRestOptions, ...options };
 		this.options.offset = Math.max(0, this.options.offset);
 		this.globalRemaining = this.options.globalRequestsPerSecond;
+		this.agent = options.agent ?? null;
```
iCrawl, line 206: The DefaultRestOptions use `getGlobalDispatcher()` for the agent, then we check if the user has provided an agent via `RESTOptions` -> `options`, and if this isn't provided, we set it to `null`. So I guess this will collide here, no?

KhafraDev: `options` is the param passed to the constructor by the dev, it can have `agent` missing. Although the DefaultOptions's agent should be a getter so that it can update if a global dispatcher is set after loading REST...

KhafraDev: You're right that it was set to the global dispatcher, but I'm not entirely sure how to be honest. I think being able to set a dispatcher for the Rest manager instance will solve every use case though.

iCrawl: Yeah I do think it solves every use case, but the expectation of people setting the global dispatcher and it not working seems like we need to fix it somehow down the line 🤔 
Like yes, it would probably route all traffic, even non-discord one through that, but if people expect that(?)

KhafraDev: Fixed this issue - had to update discord.js' undici so that the dispatcher was truly global! I was wondering why `getGlobalDispatcher` wasn't returning the ProxyAgent.

packages/discord.js/package.json
```diff
@@ -56,7 +56,7 @@
     "fast-deep-equal": "^3.1.3",
     "lodash.snakecase": "^4.1.1",
     "tslib": "^2.3.1",
-    "undici": "^4.16.0",
+    "undici": "5.2.0",
```
kyranet, line 59: Is there a reason why this doesn't use `^`?

packages/rest/__tests__/REST.test.ts
```diff
@@ -1,209 +1,245 @@
...
+mockAgent.disableNetConnect();
+setGlobalDispatcher(mockAgent);
+
+const mockPool = mockAgent.get('http://localhost:3000');
+mockPool.intercept({ path: '/foo', method: 'GET' }).reply(200, 'foo');
+
+const res = await fetch('http://localhost:3000/foo');
+
+console.log(res.status, await res.text()); // 200 foo
+*/
```
kyranet, line 31: Is this supposed to be left there?

SpaceEEC (approved):

packages/rest/package.json
```diff
@@ -53,11 +53,9 @@
 		"@discordjs/collection": "workspace:^",
 		"@sapphire/async-queue": "^1.3.1",
 		"@sapphire/snowflake": "^3.2.1",
-		"@types/node-fetch": "^2.6.1",
 		"discord-api-types": "^0.29.0",
-		"form-data": "^4.0.0",
-		"node-fetch": "^2.6.7",
-		"tslib": "^2.3.1"
+		"tslib": "^2.3.1",
+		"undici": "5.2.0"
```
kyranet, line 58: Likewise here.

kyranet (changes requested):

packages/discord.js/package.json
KhafraDev: idk I just did `yarn workspace @discordjs/rest add undici@5.2.0`. I use npm typically if you couldn't tell

packages/rest/__tests__/REST.test.ts
KhafraDev: It's pretty helpful for people who haven't used undici mocks before IMO. I used that as a basic template for everything.

packages/discord.js/package.json
iCrawl: NPM would do the same. The command itself was correct, you could also just cd into the packages folder, but you need to add the carret manually if you want to show it up (same behaviour as npm) like: `yarn workspace @discordjs/rest add undici@^5.2.0`

packages/rest/__tests__/REST.test.ts
kyranet: Then you can document it as such, a template.

I won't doubt that any of us will find that comment and delete it thinking it was just some left-overs.

packages/discord.js/package.json
KhafraDev: npm does it correctly in my experience, even with specificity the caret, yarn doesn't update it properly😕

KhafraDev: okay so without specifying the version, yarn will add the caret automatically

packages/rest/__tests__/REST.test.ts
KhafraDev: I ended up removing it, the tests show how to properly use mocks.

KhafraDev:
Oh

PR #7759, opened 2022-04-08
refactor(builders): use arrays instead of rest parameters

suneettipirneni:

Originally I had the opinion that using rest parameters allowed for a clearer UI DSL. However I've since conceded this opinion.

The main reason for this is that this design decision has resulted in a degraded UX for many users. It seems that every other message in the dev channel has to do with someone using an array instead of rest parameters. Rest paramters seem to be an unexpected input type for most.

"Why not support both rest and array parameters"

To be clear JS doesn't support this. The only way to achieve this is via a workaround, that *emulates* this behavior. This workaround involves accepting a a rest parameter 2D arrays. 

I fear that the types here are confusing since they imply you must pass in multiple 2D arrays as rest parameters, when in reality you can pass a single 1D array. This may introduce an issue in place of a resolved one. In addition, many people seem to default to array parameters anyways, so I think this workaround holds little-to-no utility.

### DIFF
--- a/packages/builders/__tests__/components/actionRow.test.ts
+++ b/packages/builders/__tests__/components/actionRow.test.ts
@@ -44,8 +44,8 @@ const rowWithSelectMenuData: APIActionRowComponent<APIMessageActionRowComponent>
 describe('Action Row Components', () => {
 	describe('Assertion Tests', () => {
 		test('GIVEN valid components THEN do not throw', () => {
-			expect(() => new ActionRowBuilder().addComponents(new ButtonBuilder())).not.toThrowError();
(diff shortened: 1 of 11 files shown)


vladfrangu (approved):

iCrawl (approved):

kyranet (approved):

SpaceEEC (approved):

ImRodry:
Honestly I didn’t think this would be merged…
Why is supporting both forms an issue? The PR description addressed this but also didn’t explain much, I feel like if a user passes both rest params or an array and they both work they wouldn’t have to worry about it and it wouldn’t be “confusing” at all. This is just limiting the options we have for no reason

PR #7734, opened 2022-03-31
feat(GuildBanManager): Support pagination results

Jiralite:
**Please describe the changes this PR makes and why it should be merged:**
See https://github.com/discord/discord-api-docs/commit/808d94b82c04d734b622d23688a3cdbe2e0903ea.

**Status and versioning classification:**
- Code changes have been tested against the Discord API, or there are no code changes
- I know how to update typings and have done so, or typings don't need updating
- This PR changes the library's interface (methods or parameters added)

PR #7763, opened 2022-04-09
fix(MessageManager): Allow caching option of an unspecified limit

### LINKED ISSUE #7093: `MessageManager#fetch()` does not work as documented
### Issue description

There seems to be a bug in the documentation or the code itself for [`MessageManager#fetch()`](https://discord.js.org/#/docs/main/main/class/MessageManager?scrollTo=fetch).

Both parameters are listed as optional. I proceed to only make use of the `options` parameter as I don't care about specifying a message limit, but I _do_ care about caching them. TypeScript yells that the overload is not valid:

```TypeScript
<Client>.once(<Constants>.Events.CLIENT_READY, async () => {
  const channel = <Client>.channels.resolve("channel_id") as TextChannel;

  // error TS2769: No overload matches this call.
  await channel.messages.fetch({
    cache: false
  });

  console.log(channel.messages.cache.size); // 50
});
```

Upon executing the above code sample in JavaScript, 50 messages are fetched as expected as is the default but the returned messages are cached as not expected.

Either the internal code logic and typings are not correct here, or this isn't documented correctly. Personally, I'd lean towards supporting this because, why not?

### Code sample

```typescript
import { Client, Constants, Intents, TextChannel } from "discord.js";

const client = new Client({
  intents: [
    Intents.FLAGS.GUILDS,
    Intents.FLAGS.GUILD_MESSAGES
  ]
});

client.once(Constants.Events.CLIENT_READY, async () => {
  const channel = client.channels.resolve("765676229478711366") as TextChannel;

  await channel.messages.fetch({
    cache: false
  });


  console.log(channel.messages.cache.size);
});

client.login();
```

### discord.js version

13.4.0-dev

### Node.js version

Node.js 16.13.0 & TypeScript 4.4.4

### Operating system

macOS Monterey 12.0.1

Low (slightly annoying)

No Partials

GUILDS, GUILD_MESSAGES

6e57b65

Jiralite:
This pull request resolves #7093. This implementation is actually the same as #7734, so this should feel familiar and consistent.

I have also renamed `ChannelLogsQueryOptions` to `FetchMessagesOptions` to better reflect this type definition and for consistency amongst other type definitions.

### DIFF
--- a/packages/discord.js/src/managers/MessageManager.js
+++ b/packages/discord.js/src/managers/MessageManager.js
@@ -84,7 +84,7 @@ class MessageManager extends CachedManager {
    *   .then(messages => console.log(`${messages.filter(m => m.author.id === '84484653687267328').size} messages`))
    *   .catch(console.error);
    */
-  fetch(options = {}) {
+  fetch(options) {
     if (!options) return this._fetchMany();
     const { message, cache, force } = options;
     const resolvedMessage = this.resolveId(message ?? options);
--- a/packages/discord.js/src/managers/MessageManager.js
+++ b/packages/discord.js/src/managers/MessageManager.js
@@ -35,40 +35,79 @@ class MessageManager extends CachedManager {
   }
 
   /**
-   * The parameters to pass in when requesting previous messages from a channel. `around`, `before` and
-   * `after` are mutually exclusive. All the parameters are optional.
-   * @typedef {Object} ChannelLogsQueryOptions
-   * @property {number} [limit] Number of messages to acquire
-   * @property {Snowflake} [before] The message's id to get the messages that were posted before it
-   * @property {Snowflake} [after] The message's id to get the messages that were posted after it
-   * @property {Snowflake} [around] The message's id to get the messages that were posted around it
+   * Data that can be resolved to a Message object. This can be:
+   * * A Message
+   * * A Snowflake
+   * @typedef {Message|Snowflake} MessageResolvable
    */
 
   /**
-   * Gets a message, or messages, from this channel.
+   * Options used to fetch a message.
+   * @typedef {BaseFetchOptions} FetchMessageOptions
+   * @property {MessageResolvable} [message] The message to fetch
+   */
+
+  /**
+   * Options used to fetch multiple messages.
+   * @typedef {Object} FetchMessagesOptions
+   * @property {number} [limit] The maximum number of messages to return
+   * @property {Snowflake} [before] Consider only messages before this id
+   * @property {Snowflake} [after] Consider only messages after this id
+   * @property {Snowflake} [around] Consider only messages around this id
+   * @property {boolean} [cache] Whether to cache the fetched messages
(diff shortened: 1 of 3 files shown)


packages/discord.js/src/managers/MessageManager.js
```diff
@@ -34,40 +34,79 @@ class MessageManager extends CachedManager {
...
-   * // Get messages and filter by user id
+   * // Fetch messages and filter by a user id
    * channel.messages.fetch()
    *   .then(messages => console.log(`${messages.filter(m => m.author.id === '84484653687267328').size} messages`))
    *   .catch(console.error);
    */
-  fetch(message, { cache = true, force = false } = {}) {
-    return typeof message === 'string' ? this._fetchId(message, cache, force) : this._fetchMany(message, cache);
+  fetch(options = {}) {
+    if (!options) return this._fetchMany();
```
vladfrangu, line 87: This will never trigger

packages/discord.js/src/managers/MessageManager.js
```diff
@@ -34,40 +34,79 @@ class MessageManager extends CachedManager {
...
    * @example
-   * // Get messages and filter by user id
+   * // Fetch messages and filter by a user id
    * channel.messages.fetch()
    *   .then(messages => console.log(`${messages.filter(m => m.author.id === '84484653687267328').size} messages`))
    *   .catch(console.error);
    */
-  fetch(message, { cache = true, force = false } = {}) {
-    return typeof message === 'string' ? this._fetchId(message, cache, force) : this._fetchMany(message, cache);
+  fetch(options = {}) {
```
vladfrangu, line 86:
```suggestion
  fetch(options = { cache: true }) {
```

packages/discord.js/src/managers/MessageManager.js
```diff
@@ -34,40 +34,79 @@ class MessageManager extends CachedManager {
...
+      if (existing && !existing.partial) return existing;
+    }
+
+    const data = await this.client.rest.get(Routes.channelMessage(this.channel.id, message));
+    return this._add(data, cache);
+  }
+
+  async _fetchMany(options = {}) {
+    const data = await this.client.rest.get(Routes.channelMessages(this.channel.id), {
+      query: new URLSearchParams(options),
```
vladfrangu, line 106: If this is merged after https://github.com/discordjs/discord.js/pull/7744, we need to adapt the code

vladfrangu (changes requested):

packages/discord.js/src/managers/MessageManager.js
Jiralite: The default here should be removed as a whole. The `_add()` methods already default this property to true and also resolves your following review.

vladfrangu:
This needs a rebase :3

SpaceEEC (approved):

iCrawl (approved):

PR #7764, opened 2022-04-09
makeContent should accept empty strings

Ortovoxx:

Currently if you send an embed with an empty string as the content you get a `RangeError [MESSAGE_CONTENT_TYPE]: Message content must be a non-empty string.` and are supposed to supply `null` when indicating that the message content is not needed. However the method, namely `makeContent`, that actually causes this error simply converts that null into an empty string.

It appears to be, and do correct me if I am wrong, that the method errors if an empty string is the content but then returns an empty string anyway?

To fix this I have changed the `verifyString` function to accept empty strings stopping the errors

### DIFF
--- a/packages/discord.js/src/errors/Messages.js
+++ b/packages/discord.js/src/errors/Messages.js
@@ -71,7 +71,7 @@ const Messages = {
 
   MESSAGE_BULK_DELETE_TYPE: 'The messages must be an Array, Collection, or number.',
   MESSAGE_NONCE_TYPE: 'Message nonce must be an integer or a string.',
-  MESSAGE_CONTENT_TYPE: 'Message content must be a non-empty string.',
+  MESSAGE_CONTENT_TYPE: 'Message content must be a string.',
 
   SPLIT_MAX_LEN: 'Chunk exceeds the max length and contains no split characters.',
 
--- a/packages/discord.js/src/structures/MessagePayload.js
+++ b/packages/discord.js/src/structures/MessagePayload.js
@@ -105,7 +105,7 @@ class MessagePayload {
     if (this.options.content === null) {
       content = '';
     } else if (typeof this.options.content !== 'undefined') {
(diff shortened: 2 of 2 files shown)


ImRodry:
Shouldn’t you change the verifyString method instead to accept empty strings and change the error message?

Ortovoxx:
Yeah that is what I've done? I've changed the `allowEmpty` argument of `verifyString` to be true instead of false so it will allow through empty strings. The error message is unchanged as it won't be thrown anymore.

ImRodry:
I’m saying it’s probably best to remove allowEmpty altogether and always accept empty strings since this is probably not used anywhere else (correct me if I’m wrong tho)

Ortovoxx:
It is referenced twice once in this PR and another [here](https://github.com/Ortovoxx/discord.js/blob/79a494fba4c18f173a3184d4dcdf410b270d2782/packages/discord.js/src/util/Util.js#L72) but it uses a single argument so will allow through empty strings.

I don't see why we should change the `verifyString` function though? Yes it would solve the problem but seems more complex than it needs to be when simply changing from false to true fixes it equally as well. Perhaps there may be a future use for this function and I don't think we should jeopardise that.

imranbarbhuiya:
https://github.com/discordjs/discord.js/blob/79a494fba4c18f173a3184d4dcdf410b270d2782/packages/discord.js/src/errors/Messages.js#L74

You should update the message also

Ortovoxx:
Good point. The error message isn't technically necessary now but I've updated it anyway.

imranbarbhuiya:
> Good point. The error message isn't technically necessary now but I've updated it anyway.

It's necessary for non-string values. Like if someone enters an array,..

vladfrangu (approved):

iCrawl (approved):

kyranet (approved):

SpaceEEC (approved):

PR #7772, opened 2022-04-11
chore: reexport all builders in discord.js

suneettipirneni:

The reason for this PR is to reduce the import confusion caused by the discord.js using builders as dependency. Currently the guide suggests to manually install builders to make slash commands. This however causes some issues when it comes to auto-imports.

For example this is the auto import vsc will give when builders is manually installed:

Screen Shot 2022-04-11 at 6 52 01 PM

As you can see it prefers the pure builders form of the builder over the discord.js version. Most users are looking for the djs versions and are confused when some methods don't work as intended.

Removing the manual installation of builders results in auto-imports that looks like so:

Screen Shot 2022-04-11 at 6 52 53 PM

Even though builders is in `node_modules` vsc won't reccomend it because it's not in the projects `package.json`

Since we already use builders as regular dependency we can just reexport the command builders as-is from discord.js. This means users *of discord.js* don't need to manually install builders in order to use any of them. The "creating commands" section of the guide will be updated. In addition, the migration guide will recommend uninstalling the builders standalone package if it's already installed.

This also fixes a bug where the emoji validator wouldn't like nullish emoji ids to be accepted.


packages/discord.js/src/index.js
```diff
@@ -193,13 +194,26 @@ exports.GuildVerificationLevel = require('discord-api-types/v10').GuildVerificat
...
+exports.SlashCommandBooleanOption = require('@discordjs/builders').SlashCommandBooleanOption;
+exports.SlashCommandChannelOption = require('@discordjs/builders').SlashCommandChannelOption;
+exports.SlashCommandIntegerOption = require('@discordjs/builders').SlashCommandIntegerOption;
+exports.SlashCommandNumberOption = require('@discordjs/builders').SlashCommandNumberOption;
+exports.SlashCommandMentionableOption = require('@discordjs/builders').SlashCommandMentionableOption;
+exports.SlashCommandRoleOption = require('@discordjs/builders').SlashCommandRoleOption;
+exports.SlashCommandStringOption = require('@discordjs/builders').SlashCommandStringOption;
+exports.SlashCommandSubcommandBuilder = require('@discordjs/builders').SlashCommandSubcommandBuilder;
+exports.SlashCommandSubcommandGroupBuilder = require('@discordjs/builders').SlashCommandSubcommandGroupBuilder;
+exports.SlashCommandUserOption = require('@discordjs/builders').SlashCommandUserOption;
```
ImRodry, line 216: Instead you could just do `exports = { …, …require(‘@discordjs/builders’) }` ?

suneettipirneni: This would overwrite the exports we do for djs custom classes

ImRodry: You didn’t read it all, it doesn’t overwrite those

imranbarbhuiya: It does. What he mean is djs exports `ActionRowBuilder` so it'll be present in exports. Then when adding djs/builder that `ActionRowBuilder` will override the djs one

SpaceEEC (approved):

iCrawl (approved):

vladfrangu (approved):

PR #7773, opened 2022-04-11
fix(Util): flatten ignoring certain fields

### LINKED ISSUE #7712: Message#toJSON() issue

discord.js

### Issue description

When message is containing embed with author and you call .toJSON on the message rather on the embed itself, and then access the embed property on the generated json, the author property is missing.
When you call toJSON on embed directly, the author property is there. I'd imagine that toJSON would call each objects toJSON separately but i guess thats not the case?

I'd like to request unban on my [main account](https://github.com/DaliborTrampota) from this repo. Its been a long time, I have not caused any issues and I've reported few bugs already. I've done my sentence 

### Code sample

```typescript
const e = new MessageEmbed().setAuthor({ name: 'hey' })
let msg = await message.channel.send({ embeds: [e] })
console.log(msg.toJSON())
console.log(msg.embeds[0].toJSON())
```

### Package version

13.6.0

### Node.js version

16.13.0

### Operating system

windows 10 pro

Low (slightly annoying)

No Partials

Guilds, GuildMessages

_No response_

newracket:
Fixes #7712 

Improves clarity of toJSON functions throughout. Currently, some fields are ignored in toJSON if the value tied to that field is an object (as pointed out by issue #7712). This change fixes issue #7712 by making it so fields tied to an object don't get ignored.


vladfrangu (approved):

iCrawl (approved):

kyranet:
One of my concerns here is... shouldn't we try to use `.toJSON` where possible?

With the switch to raw data, we may have nested data, which is what happens here in your sample:

```diff
  "embeds": [
    {
+     "data": {
        "type": "rich",
        "author": {
          "name": "hey"
        }
+     }
    }
  ],
```

If `.toJSON` were to be preferred and bypass the JSON serialization from the utility, `data` would get dropped, and the data would be "flat".

newracket:
@kyranet I made a second commit making it try to use `.toJSON` if it exists (both when the element's an object and when it's an array). The same `Message.toJSON()` from my previous commit now looks like this.

```json
{
  "channelId": "723654712410374194",
  "guildId": "605969074178228224",
  "id": "963209574242979900",
  "createdTimestamp": 1649717458068,
  "type": 0,
  "system": false,
  "content": "",
  "authorId": "605975599609020437",
  "pinned": false,
  "tts": false,
  "nonce": null,
  "embeds": [
    {
      "type": "rich",
      "author": {
        "name": "hey"
      }
    }
  ],
  "components": [],
  "attachments": [],
  "stickers": [],
  "editedTimestamp": null,
  "mentions": {
    "everyone": false,
    "users": [],
    "roles": [],
    "crosspostedChannels": [],
    "repliedUser": null,
    "members": [],
    "channels": []
  },
  "webhookId": null,
  "groupActivityApplicationId": null,
  "applicationId": null,
  "activity": null,
  "flags": 0,
  "reference": null,
  "interaction": null,
  "cleanContent": ""
}
```

kyranet (approved):

PR #7792, opened 2022-04-13
fix(ButtonBuilder): default constructor to empty object

monbrey:
Fixes a error thrown by deconstructing `property 'emoji' of 'undefined'` when calling `new ButtonBuilder()` with no argument.


imranbarbhuiya:
#7788

monbrey:
> #7788

oops, I even looked first but missed that somehow. Thanks!

PR #7795, opened 2022-04-13
feat(builders): improve embed errors and predicates

monbrey:
Improves the error output of some EmbedBuilder methods by using an object validator rather than validating params individually.

This provides much better output for the most common errors seen due to old v13 style code such as `<EmbedBuilder>.setAuthor("Monbrey")`, which will now throw `Expected the value to be an object, but received string instead`.

I didn't see anywhere else this needed doing but feel free to point out any spots and I'll do the same.

### DIFF
--- a/packages/builders/src/messages/embed/Assertions.ts
+++ b/packages/builders/src/messages/embed/Assertions.ts
@@ -33,7 +33,7 @@ export const urlPredicate = s.string.url({
 
 export const embedAuthorPredicate = s.object({
 	name: authorNamePredicate,
-	iconURL: urlPredicate,
+	iconURL: imageURLPredicate,
 	url: urlPredicate,
 });
 
@@ -49,7 +49,7 @@ export const footerTextPredicate = s.string.lengthGe(1).lengthLe(2048).nullable;
 
 export const embedFooterPredicate = s.object({
 	text: footerTextPredicate,
-	iconURL: urlPredicate,
+	iconURL: imageURLPredicate,
 });
 
 export const timestampPredicate = s.union(s.number, s.date).nullable;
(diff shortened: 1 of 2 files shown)


SpaceEEC (approved):
:+1:

iCrawl (approved):

packages/builders/src/messages/embed/Assertions.ts
```diff
@@ -41,6 +47,11 @@ export const descriptionPredicate = s.string.lengthGe(1).lengthLe(4096).nullable
 
 export const footerTextPredicate = s.string.lengthGe(1).lengthLe(2048).nullable;
 
+export const embedFooterPredicate = s.object({
+	text: footerTextPredicate,
+	iconURL: urlPredicate,
```
Syjalo, line 52:
```suggestion
	iconURL: imageURLPredicate,
```
Fixes #7800

Syjalo (changes requested):

packages/builders/src/messages/embed/Assertions.ts
```diff
@@ -31,6 +31,12 @@ export const urlPredicate = s.string.url({
 	allowedProtocols: ['http:', 'https:'],
 }).nullish;
 
+export const embedAuthorPredicate = s.object({
+	name: authorNamePredicate,
+	iconURL: urlPredicate,
```
Syjalo, line 36:
```suggestion
	iconURL: imageURLPredicate,
```
Also can be set by attachment

Syjalo (changes requested):

vladfrangu (approved):

PR #7804, opened 2022-04-16
docs(ApplicationCommand): fix and improve localization docs

ImRodry:

This PR:

- adds a `Locale` typedef with all the possible locales to make it easier to document them in various places
- fixes the documentation of the *Localizations and *Localized properties

### DIFF
--- a/packages/discord.js/src/structures/ApplicationCommand.js
+++ b/packages/discord.js/src/structures/ApplicationCommand.js
@@ -65,7 +65,7 @@ class ApplicationCommand extends Base {
     if ('name_localizations' in data) {
       /**
        * The name localizations for this command
-       * @type {?Object<string, string>}
+       * @type {?Object<Locale, string>}
        */
       this.nameLocalizations = data.name_localizations;
     } else {
@@ -75,7 +75,7 @@ class ApplicationCommand extends Base {
     if ('name_localized' in data) {
       /**
        * The localized name for this command
-       * @type {?Object<string, string>}
+       * @type {?string}
        */
       this.nameLocalized = data.name_localized;
     } else {
@@ -93,7 +93,7 @@ class ApplicationCommand extends Base {
     if ('description_localizations' in data) {
       /**
        * The description localizations for this command
-       * @type {?string}
+       * @type {?Object<Locale, string>}
        */
       this.descriptionLocalizations = data.description_localizations;
     } else {
@@ -169,9 +169,9 @@ class ApplicationCommand extends Base {
    * @typedef {Object} ApplicationCommandData
    * @property {string} name The name of the command, must be in all lowercase if type is
    * {@link ApplicationCommandType.ChatInput}
-   * @property {Object<string, string>} [nameLocalizations] The localizations for the command name
+   * @property {Object<Locale, string>} [nameLocalizations] The localizations for the command name
    * @property {string} description The description of the command, if type is {@link ApplicationCommandType.ChatInput}
-   * @property {Object<string, string>} [descriptionLocalizations] The localizations for the command description,
+   * @property {Object<Locale, string>} [descriptionLocalizations] The localizations for the command description,
    * if type is {@link ApplicationCommandType.ChatInput}
    * @property {ApplicationCommandType} [type=ApplicationCommandType.ChatInput] The type of the command
    * @property {ApplicationCommandOptionData[]} [options] Options for the command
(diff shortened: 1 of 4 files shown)


xHyroM:
I can do backport for v13

suneettipirneni:
> I tested with a bot token and the X-Discord-Locale header and they are not sent

I just tested this and it certainly sends back name_localized and description_localized when the header is set.

> changes the default of withLocalizations in ApplicationCommandManager#fetch to true in order to always ensure the user running this method gets all the data, seeing as providing it as false by default has no impact other than receiving the properties as null

This was brought up internally and the consensus was that it shouldn't return localizations by default, because localization JSON can result in *much* larger json payloads. If you need to do something like diff commands you can specify the option as true in those situations.

ImRodry:
> I just tested this and it certainly sends back name_localized and description_localized when the header is set.

Show me how you did it then because I tested this many times and never received those fields

> This was brought up internally and the consensus was that it shouldn't return localizations by default, because localization JSON can result in much larger json payloads. If you need to do something like diff commands you can specify the option as true in those situations.

That's inconsistent with other methods but if a maintainer wants that removed I'll remove it. I'll wait for now

suneettipirneni:
> Show me how you did it then because I tested this many times and never received those fields

I just did a get request to fetch all commands with my bot token in the header and the locale header set to the locale.

ImRodry:
Can't reproduce so unless someone else is able to and details how I'll keep these changes here. For reference I'm requesting all application commands without the with_localizations query parameter, a bot token and the X-Discord-Locale header set to "pt-BR"

kyranet:
> That's inconsistent with other methods but if a maintainer wants that removed I'll remove it. I'll wait for now

Hi, maintainer here. The behaviour and defaults of the method were chosen internally, and as such, we took part of it.

IMO the code changes in this PR should be reverted, the docs changes look better, though.

ImRodry:
> > That's inconsistent with other methods but if a maintainer wants that removed I'll remove it. I'll wait for now
> 
> Hi, maintainer here. The behaviour and defaults of the method were chosen internally, and as such, we took part of it.
> 
> IMO the code changes in this PR should be reverted, the docs changes look better, though.

I redid my tests and I'm now able to get the Localized properties so I have reverted the code changes and kept the documentation changes. Feel free to change the labels on this PR

packages/discord.js/src/structures/ApplicationCommand.js
```diff
@@ -458,6 +453,15 @@ class ApplicationCommand extends Base {
    * {@link ApplicationCommandOptionType.Number} option
    */
 
+  /**
+   * A choice for an application command option.
+   * @typedef {Object} ApplicationCommandOptionChoice
+   * @property {string} name The name of the choice
+   * @property {?string} nameLocalized The localized name of the choice in the provided locale, if any
```
vladfrangu, line 460: Isn't this also possibly undefined?

packages/discord.js/typings/index.d.ts
```diff
@@ -3595,10 +3595,6 @@ export interface ApplicationCommandOptionChoiceData {
   value: string | number;
 }
 
-export interface ApplicationCommandOptionChoice extends ApplicationCommandOptionChoiceData {
```
vladfrangu, line 3598: Is this not needed anymore?

vladfrangu (dismissed):

packages/discord.js/typings/index.d.ts
ImRodry: it never was

packages/discord.js/src/structures/ApplicationCommand.js
ImRodry: ? means undefined or null, but also no

vladfrangu: No, `?` means null, `[nameLocalized]` means undefined
Are you *sure* it can't be undefined too?

ImRodry: [nameLocalized] means optional (aka possibly absent, not present as undefined) and ? means nullable, aka null or undefined
And yes I am sure because the code sets the property to null if not sent by the API

kyranet: Nullable is `T | null`, null or undefined (`T | null | undefined`) is called "nullish".

ImRodry: But why does it matter? This will never be undefined and there’s no property on djs that haw undefined as an explicit possible type in the documentation

kyranet: You didn't say that. In this case, `?string` is correct.

kyranet (approved):

packages/discord.js/src/structures/ApplicationCommand.js
ImRodry: I did say that, twice

SpaceEEC (approved):

PR #7810, opened 2022-04-17
feat(MessageReaction): add react method

ImRodry:

This PR adds the MessageReaction#react method to make it easier to add an existing reaction to a message. Its purpose is to make it easier for bots to react with emojis they do not know in their cache. The name is up for debate if anyone has better ideas, this was the best I could come up with

### DIFF
--- a/packages/discord.js/src/structures/MessageReaction.js
+++ b/packages/discord.js/src/structures/MessageReaction.js
@@ -53,6 +53,14 @@ class MessageReaction {
   }
 
   /**
+   * Makes the client user react with this reaction
+   * @returns {Promise<MessageReaction>}
+   */
+  react() {
+    return this.message.react(this.emoji);
+  }
+
+  /**
    * Removes all users from this reaction.
    * @returns {Promise<MessageReaction>}
    */
--- a/packages/discord.js/typings/index.d.ts
+++ b/packages/discord.js/typings/index.d.ts
@@ -1883,6 +1883,7 @@ export class MessageReaction {
   public message: Message | PartialMessage;
   public get partial(): false;
   public users: ReactionUserManager;
+  public react(): Promise<MessageReaction>;
   public remove(): Promise<MessageReaction>;
   public fetch(): Promise<MessageReaction>;
   public toJSON(): unknown;


vladfrangu:
> This PR adds the MessageReaction#react method to make it easier to add an existing reaction to a message. Its purpose is to make it easier for bots to react with emojis they do not know in their cache.

I mean no, you know the emoji since you have the message reaction class :P (+ you don't neeeed the emoji cached to add reactions)
This is just a shortcut 😄

ImRodry:
Exactly, I said make it easier not make it possible :)

monbrey:
I don't quite understand this. It's a shortcut for the bot to react with the same emoji a user did? 

...I guess it's fine?

ImRodry:
Yep that’s pretty much it

suneettipirneni:
I don't understand why this method needs to be a part of discord.js. Making a utility method for this would be trivial:

```ts
const rereact = async (msg: Message, reaction: MessageReaction) => msg.react(reaction.emoji);
```

In addition, the use cases to "duplicate" a reaction on a message seem niche. Beyond that, this method makes no structural sense.

A message is *a parent* to its reactions, the reactions are *children* of the message. It doesn't make sense for a child to be able to manage its parent. It's the same reason there's no `.send()` method on `Message` which sends a "copy" of the message to the channel the original message was sent in. A message shouldn't have a method that implicitly manages its parent/owner.

These kinds of methods also lack readability.

```ts
messageRxn.react()
```

Are we reacting to a message reaction? How is that possible? Where's the arguments? Even with a better name the method will always feel out of place due to the fact that a single message reaction isn't responsible for modifying the collection of message reactions on a given message.

```ts
message.react(...)
```
This is much clearer to see we're reacting *to* a message here. And the method feels like it's in the right place because messages own reactions, not vice-versa.

ImRodry:
>  It's the same reason there's no .send() method on Message which sends a "copy" of the message to the channel the original message was sent in. A message shouldn't have a method that implicitly manages its parent/owner.

On the Discord client you can't click a message to send one similar to it and you can click a reaction to react with that emoji too, that was the rationale behind this

> Are we reacting to a message reaction? How is that possible? Where's the arguments?

You know that's not the case so why are you even making that an argument, that's just stupid. This is simply grabbing an existing reaction by someone else and adding a reaction from the client user similar to the one that was given, nothing more nothing less. If the maintainers don't want to add this they can close the PR but I'm tired of reading a thesis from you every time I open a PR, so please stop, you're not a maintainer

DraftProducts:
I find the addition of this *shortcut* relevant, especially in the case of reaction vents where `<MessageReaction>` is directly returned as a parameter. 

Reading is much easier in the second demo, and someone who has read the doc knows that reactionMessage represents an emoji linked to a message.
It is therefore understandable that you cannot add another emoji than the one concerned by interacting with this method.

```js
client.on('messageReactionAdd', async (reactionMessage, _user) => {
   await reactionMessage.message.react(reactionMessage.emoji);
})

client.on('messageReactionAdd', async (reactionMessage, _user) => {
   await reactionMessage.react();
})
```
If the lack of arguments to this method is problematic at this point, could a name change be considered? Something like `<MessageReaction>.reReact()`?

> The project is open-source, it is so that the community can give its opinion and contribute to it, just as much as you allow yourself to propose this PR, it is normal that @suneettipirneni has the right to bring an opinion opposed to yours. 😕

vladfrangu:
> If the lack of arguments to this method is problematic at this point, could a name change be considered? Something like `<MessageReaction>.reReact()`?

Personally I'd rather approve the PR with the current naming rather than going with `reReact`... The naming is something that can be improved (maybe? nobody has offered any decent alternatives for now) but I doubt it'll be a blocker

vladfrangu (approved):
Past the naming argument which I've mentioned like 3? times now, LGTM

BenPatersonxx:
@iCrawl @kyranet @SpaceEEC please review

ImRodry:
Just to be clear: if the maintainers agree on any better naming for this method I'll gladly change it but I'd like that opinion to come from them so we don't keep changing things back and forth. Appreciate the feedback so far though

SpaceEEC (approved):

iCrawl (approved):

kyranet (approved):
I was thinking of `reaction.users.addMe()`, but I *really* can't think of a better name.

PR #7812, opened 2022-04-17
fix(InteractionResponses): use optional chaining on nullable property

monbrey:
MessageComponents do not require an interaction to be sent - this was causing an `id of null` error.

### DIFF
--- a/packages/discord.js/src/structures/InteractionCollector.js
+++ b/packages/discord.js/src/structures/InteractionCollector.js
@@ -211,7 +211,7 @@ class InteractionCollector extends Collector {
       this.stop('messageDelete');
     }
 
-    if (message.interaction.id === this.messageInteractionId) {
+    if (message.interaction?.id === this.messageInteractionId) {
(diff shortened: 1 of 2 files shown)


Clauz07:
I found out that the same bug happens in `structures/InteractionCollector.js` on line 214, so I think this should be updated

vladfrangu (commented):
Won't this break things in InteractionResponse when the id is undefined?

monbrey:
> Won't this break things in InteractionResponse when the id is undefined?

I don't believe so. InteractionResponse nullish coalesces it if the second param isn't provided:
<https://github.com/discordjs/discord.js/blob/9ef7ffdc6dc8eb9466f77344f9356d9027bf7a19/packages/discord.js/src/structures/InteractionResponse.js#L24>

vladfrangu (approved):

iCrawl (approved):

iCrawl (approved):

PR #7827, opened 2022-04-22
fix(User): banner and accentColor undefined, use null instead

Mateo-tem:

Fixes the `banner` and `accentColor` properties of `User`, use null instead of undefined.

I don't refer that using undefined is wrong, but I think it is a matter of aspect because if there is no data, null is used always, but only on these properties, undefined is used.

### DIFF
--- a/packages/discord.js/src/structures/User.js
+++ b/packages/discord.js/src/structures/User.js
@@ -78,8 +78,8 @@ class User extends Base {
        * @type {?string}
        */
       this.banner = data.banner;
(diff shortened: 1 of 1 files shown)


BenjammingKirby:
hmm did you check https://github.com/discordjs/discord.js/pull/6721 ?

Mateo-tem:
Uhmm, no. I'm sorry.
Ok, so I will close this because now I see that have a valid reason. Thanks :)

PR #7851, opened 2022-04-26
docs: Added default boolean values

KonkenBonken:
Added default boolean values

- No code change, only docs

### DIFF
--- a/packages/discord.js/src/structures/ApplicationCommand.js
+++ b/packages/discord.js/src/structures/ApplicationCommand.js
@@ -192,10 +192,10 @@ class ApplicationCommand extends Base {
    * @property {Object<Locale, string>} [nameLocalizations] The name localizations for the option
    * @property {string} description The description of the option
    * @property {Object<Locale, string>} [descriptionLocalizations] The description localizations for the option
-   * @property {boolean} [autocomplete] Whether the autocomplete interaction is enabled for a
+   * @property {boolean} [autocomplete=false] Whether the autocomplete interaction is enabled for a
    * {@link ApplicationCommandOptionType.String}, {@link ApplicationCommandOptionType.Integer} or
    * {@link ApplicationCommandOptionType.Number} option
-   * @property {boolean} [required] Whether the option is required
+   * @property {boolean} [required=false] Whether the option is required
    * @property {ApplicationCommandOptionChoiceData[]} [choices] The choices of the option for the user to pick from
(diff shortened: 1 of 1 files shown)


packages/discord.js/src/structures/ApplicationCommand.js
```diff
@@ -192,10 +192,10 @@ class ApplicationCommand extends Base {
    * @property {Object<Locale, string>} [nameLocalizations] The name localizations for the option
    * @property {string} description The description of the option
    * @property {Object<Locale, string>} [descriptionLocalizations] The description localizations for the option
-   * @property {boolean} [autocomplete] Whether the autocomplete interaction is enabled for a
+   * @property {boolean} [autocomplete=false] Whether the autocomplete interaction is enabled for a
```
vladfrangu, line 195: We never set autocomplete to `false` by default (https://github.com/discordjs/discord.js/blob/df64d3ea382c07e66bc7cc8877ee430206c31d63/packages/discord.js/src/structures/ApplicationCommand.js#L511)

vladfrangu (changes requested):

packages/discord.js/src/structures/ApplicationCommand.js
KonkenBonken: `autocomplete` is a boolean value, so it will have a default value. If it's omitted, it will have exactly the same effect as if it's set to `false`, because it's the default in the Discord API.
Am I wrong?

BenjammingKirby: we don't have a default value explicitly so this documentation is wrong
and we don't need to hard input the default values the api sets, we recently removed all the redundant default value settings too

KonkenBonken: So we don't have any Discord API defaults in the Discord.js docs?
Ok, then I understand, thanks

KonkenBonken: Should I close this pull request?

And why don't we put the Discord API defaults in the docs?

BenjammingKirby: probably
and should we? imagine discord changes it, it's just more work on djs' end. And the way you're going about this makes it seem like the method itself has a default value which it doesn't

KonkenBonken: Understand that. But for everyone using Djs, where the defaulting happends has no effect at all.
But thanks for explaining :)

PR #7853, opened 2022-04-26
feat(REST): enable setting default authPrefix

### LINKED ISSUE #7823: set default `authPrefix` in REST's constructor

rest

### Feature

when using REST with Bearer tokens, i have to set `authPrefix` to `Bearer` on every request, while the actual bearer token is set right after constructing the object with `.setToken(...)`. it looks bad, and sometimes i forget to supply the parameter leading to errors and longer time to develop.

example of what this looks like in my code
```ts
const guilds = (await userDiscordRest.get(Routes.userGuilds(), {
    authPrefix: 'Bearer'
})) as APIGuild[];
```
prettier seems to want to wrap it to three lines which is anoying as well.

Allow the rest constructor to set the default `authPrefix`, reducing copied code when using the library.

```ts
const rest = new REST({ version: '10', authPrefix: "Bearer" });
```

Something else we could implement instead is an extra option to setToken so it would look like `.setToken(token, { authPrefix: "Bearer" })` or `.setToken(token, "Bearer")` or something similar.

### Other context

It should be a simple change to this line in RequestManager https://github.com/discordjs/discord.js/blob/585169f2f097ffb1940d17f549e4290aa55acde2/packages/rest/src/lib/RequestManager.ts#L354, letting it use a property passed to the constructor.

with-heart:

This PR closes #7823 by adding an `authPrefix` option to `RESTOptions`, with a default value of `Bot`. 

Previously, a user wanting to use `REST` with `Bearer` tokens had to set the `authPrefix` option on every request. With this change, users only need to initialize `REST` with `authPrefix: 'Bearer'`.

```diff
-const api = new REST({version: '10'}).setToken(token)
+const api = new REST({version: '10', authPrefix: 'Bearer'}).setToken(token)

-const me = await api.get(Routes.user(), {authPrefix: 'Bearer'})
+const me = await api.get(Routes.user())
```

`options.authPrefix` is used when setting `headers.Authorization` if `request.authPrefix` is not defined. If `options.authPrefix` isn't set, there is no change in behavior from how the library worked previously.

### DIFF
--- a/packages/rest/src/lib/REST.ts
+++ b/packages/rest/src/lib/REST.ts
@@ -30,6 +30,12 @@ export interface RESTOptions {
 	 */
 	api: string;
 	/**
+	 * The authorization prefix to use for requests, useful if you want to use
+	 * bearer tokens
+	 * @default 'Bot'
+	 */
+	authPrefix: 'Bot' | 'Bearer';
+	/**
 	 * The cdn path
 	 * @default 'https://cdn.discordapp.com'
 	 */
--- a/packages/rest/src/lib/RequestManager.ts
+++ b/packages/rest/src/lib/RequestManager.ts
@@ -351,7 +351,7 @@ export class RequestManager extends EventEmitter {
 				throw new Error('Expected token to be set for this request, but none was present');
 			}
 
-			headers.Authorization = `${request.authPrefix ?? 'Bot'} ${this.#token}`;
+			headers.Authorization = `${request.authPrefix ?? this.options.authPrefix} ${this.#token}`;
 		}
 
 		// If a reason was set, set it's appropriate header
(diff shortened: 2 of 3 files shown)


vladfrangu (approved):

kyranet (approved):

davecaruso:
thank you for implementing my suggestion, i was too scared to send a pr without prior approval on the issue, and code looks great.

with-heart:
> thank you for implementing my suggestion, i was too scared to send a pr without prior approval on the issue, and code looks great.

Yo! It's no problem but also you should never be scared!

What makes open source so great is our ability to share with and learn from each other and our ability to learn through experience.

If you have an idea and you feel like it's important or worth it for the project to support it, just go for it! At the worst, the maintainers decide not to merge it or go with a different approach, but either way you come away knowing you tried your best to give life to your idea, and you learned some things along the way 💖

iCrawl (approved):

PR #7869, opened 2022-04-30
enum fixes

ErdemGKSL:
All CONSTANT_CASE strings' turned into PascalCase strings


ImRodry:
Not only did you ignore any kind of commit/PR standards or guidelines you also didn’t understand where these are used. Constant strings like these are usually written in UPPER_CASE in JS so this PR is invalid

vladfrangu:
I'm going to be closing this as it's not valid. We have no plans of making the strings PascalCase for constants (If you want PascalCase, use the enums we re-export instead), and altering just the typings won't change the runtime behavior. These strings are just for convenience for those that don't want to use enums. :D

PR #7876, opened 2022-05-03
feat(EnumResolvers): remove Enumresolvers

KhafraDev:

👋 These were an okay idea when in planning when it seemed as though it would help transition from v13... or whatever their original intention was... but they're rather useless. Importing types directly from `discord-api-types/v10` is shorter & easier than looking up a 20+ digit function name.

Let's remove them before people actually start using them in v14...


petsdom:
hi

vladfrangu (approved):
I'm approving this but I'm neutral if it gets merged or not. All enums are re-exported in discord.js, so all users have to do is add them to the imports from discord.js

iCrawl (approved):

PR #7910, opened 2022-05-11
fix(AllowedImageSizes): add ALL image sizes

asdfisdffdg:
List of allowed image sizes was a random mishmash of whatever was seen by whatever people. Using the following code I found ALL the sizes that are actually available.
```js
var fetch = require("node-fetch");

(async function(){
	var sizes = [];
	for (let i = 1; i <= 4096; i++) {
		console.debug(i);
		var res = await fetch(`https://media.discordapp.net/emojis/672958233329270804.png?size=${i}`);
		if (res.ok) {
			console.log(i.toString());
			sizes.push(i);
		}
		else if (res.status != 400) console.warn(res.status);
	}
	console.log(JSON.stringify(sizes));
})();
```

**Status and versioning classification:** wat

### DIFF
--- a/src/util/Constants.js
+++ b/src/util/Constants.js
@@ -17,7 +17,7 @@ exports.WSCodes = {
 
 const AllowedImageFormats = ['webp', 'png', 'jpg', 'jpeg', 'gif'];
 
-const AllowedImageSizes = [16, 32, 56, 64, 96, 128, 256, 300, 512, 600, 1024, 2048, 4096];
+const AllowedImageSizes = [16, 20, 22, 24, 28, 32, 40, 44, 48, 56, 60, 64, 80, 96, 100, 128, 160, 240, 256, 300, 320, 480, 512, 600, 640, 1024, 1280, 1536, 2048, 3072, 4096];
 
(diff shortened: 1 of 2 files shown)


almostSouji:
Or, alternatively: <https://discord.com/developers/docs/reference#image-formatting>
> The returned size can be changed by appending a querystring of ?size=desired_size to the URL. Image size can be any power of two between 16 and 4096.

almostSouji:
We do not accept PRs to the stable branch of the library (`dev`, `v13` are the current development branches).
As we follow API spec as per Discord API documentation, you should suggest the change at <https://github.com/discord/discord-api-docs/issues> before.

asdfisdffdg:
@almostSouji If discord.js is to match the api doc then https://github.com/discordjs/discord.js/pull/6754 and https://github.com/discordjs/discord.js/pull/6844 should not have been merged?

But they have been, so now it is inconsistent. So should I pr to v13 branch or should those changes be reverted?

I will open the issue on the discord api docs.

almostSouji:
then feel free to re-target and re-open/redo as a backport and reference those two PRs in the opening post

PR #7912, opened 2022-05-11
fix(AllowedImageSizes): add ALL image sizes

asdfisdffdg:
same as https://github.com/discordjs/discord.js/pull/7910 but cherry-picked to the right branch.

https://github.com/discordjs/discord.js/pull/6754 and https://github.com/discordjs/discord.js/pull/6844 set precedent to include undocumented sizes as well, but has missed most of them.

### DIFF
--- a/src/util/Constants.js
+++ b/src/util/Constants.js
@@ -17,7 +17,7 @@ exports.WSCodes = {
 
 const AllowedImageFormats = ['webp', 'png', 'jpg', 'jpeg', 'gif'];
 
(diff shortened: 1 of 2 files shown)


imranbarbhuiya:
https://github.com/discord/discord-api-docs/issues/4031#issuecomment-963456127

asdfisdffdg:
I was just about to reference that. So maybe those two PRs should be reverted instead.

imranbarbhuiya:
> I was just about to reference that. So maybe those two PRs should be reverted instead.

Yeah

PR #7917, opened 2022-05-13
refactor: always return `Message` instances in interactions

### LINKED ISSUE #7956: Messages typings in DM

discord.js

### Issue description

Actually, working with DM messages (that can not satisfy the condition `inCachedGuild()`) is a pain in the ass since we have all API*** typings (like APIMessage).

image

After multiple tests, I never got a raw message identical to the APIMessage interface, is it possible ?
*Looks like that the only way to have a APIMessage is to not have a cached dm channel*

It would be nice (if the APIMessage is a structure that we can get in DM) if we could have a method to exclude this type of Typing.

### Code sample

```typescript
import { Client, IntentsBitField, Interaction } from 'discord.js';
const client = new Client({
    intents: IntentsBitField.Flags.DirectMessages,
});
client.on('interactionCreate', async (interaction: Interaction) => {
    if (!interaction.isMessageComponent()) return;
    const message = interaction.message;
    if (message.deletable) await message.delete();
});
```

### Package version

main

### Node.js version

v16.13.2 (typescript v4.6.4)

Medium (should be fixed soon)

suneettipirneni:

When editing or fetching replies, the code in this PR will no longer return `APIMessage` objects for bots that aren't present in a guild. Instead it will always return a `Message` instance by creating a new `Message` from the `APIMessage` data. As result the return type for the interaction response methods (when `fetchReply` is set to `true`) is no longer `Message | APIMessage`.

Closes #7956

### DIFF
--- a/packages/discord.js/src/structures/ContextMenuCommandInteraction.js
+++ b/packages/discord.js/src/structures/ContextMenuCommandInteraction.js
@@ -3,6 +3,9 @@
 const { ApplicationCommandOptionType } = require('discord-api-types/v10');
 const CommandInteraction = require('./CommandInteraction');
 const CommandInteractionOptionResolver = require('./CommandInteractionOptionResolver');
+const { lazy } = require('../util/Util');
+
+const getMessage = lazy(() => require('./Message').Message);
 
 /**
  * Represents a context menu interaction.
@@ -50,7 +53,7 @@ class ContextMenuCommandInteraction extends CommandInteraction {
         value: target_id,
         message:
           this.channel?.messages._add(resolved.messages[target_id]) ??
-          new (require('./Message').Message)(resolved.messages[target_id]),
+          new (getMessage())(this.client, resolved.messages[target_id]),
       });
     }
 
--- a/packages/discord.js/src/structures/ContextMenuCommandInteraction.js
+++ b/packages/discord.js/src/structures/ContextMenuCommandInteraction.js
@@ -48,7 +48,9 @@ class ContextMenuCommandInteraction extends CommandInteraction {
         name: 'message',
         type: '_MESSAGE',
         value: target_id,
-        message: this.channel?.messages._add(resolved.messages[target_id]) ?? resolved.messages[target_id],
+        message:
+          this.channel?.messages._add(resolved.messages[target_id]) ??
+          new (require('./Message').Message)(resolved.messages[target_id]),
       });
     }
 
--- a/packages/discord.js/src/structures/ModalSubmitInteraction.js
+++ b/packages/discord.js/src/structures/ModalSubmitInteraction.js
@@ -34,9 +34,9 @@ class ModalSubmitInteraction extends Interaction {
     if ('message' in data) {
       /**
        * The message associated with this interaction
-       * @type {?(Message|APIMessage)}
+       * @type {?Message}
        */
-      this.message = this.channel?.messages._add(data.message) ?? data.message;
+      this.message = this.channel?.messages._add(data.message) ?? new (require('./Message').Message)(data.message);
     } else {
       this.message = null;
     }
--- a/packages/discord.js/src/structures/ModalSubmitInteraction.js
+++ b/packages/discord.js/src/structures/ModalSubmitInteraction.js
@@ -4,6 +4,9 @@ const Interaction = require('./Interaction');
 const InteractionWebhook = require('./InteractionWebhook');
 const ModalSubmitFields = require('./ModalSubmitFields');
 const InteractionResponses = require('./interfaces/InteractionResponses');
+const { lazy } = require('../util/Util');
+
+const getMessage = lazy(() => require('./Message').Message);
 
 /**
  * @typedef {Object} ModalData
@@ -36,7 +39,7 @@ class ModalSubmitInteraction extends Interaction {
        * The message associated with this interaction
        * @type {?Message}
        */
-      this.message = this.channel?.messages._add(data.message) ?? new (require('./Message').Message)(data.message);
+      this.message = this.channel?.messages._add(data.message) ?? new (getMessage())(this.client, data.message);
     } else {
       this.message = null;
     }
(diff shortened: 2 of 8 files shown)


packages/discord.js/src/structures/Webhook.js
```diff
@@ -207,7 +207,10 @@ class Webhook {
 
     const { body, files } = await messagePayload.resolveFiles();
     const d = await this.client.rest.post(Routes.webhook(this.id, this.token), { body, files, query, auth: false });
-    return this.client.channels?.cache.get(d.channel_id)?.messages._add(d, false) ?? d;
+    return (
+      this.client.channels?.cache.get(d.channel_id)?.messages._add(d, false) ??
+      new (require('./Message'))(this.client, d)
```
almeidx, line 212:
```suggestion
      new (require('./Message').Message)(this.client, d)
```

packages/discord.js/src/structures/Webhook.js
```diff
@@ -207,7 +207,10 @@ class Webhook {
 
     const { body, files } = await messagePayload.resolveFiles();
     const d = await this.client.rest.post(Routes.webhook(this.id, this.token), { body, files, query, auth: false });
-    return this.client.channels?.cache.get(d.channel_id)?.messages._add(d, false) ?? d;
+    return (
+      this.client.channels?.cache.get(d.channel_id)?.messages._add(d, false) ??
+      new (require('./Message').Message)(this.client, d)
```
kyranet, line 212: Do `let Message;` above and then here (and everywhere else) `(Message ??= require('./Message').Message)`.

kyranet (changes requested):

packages/discord.js/src/structures/Webhook.js
```diff
@@ -7,6 +7,8 @@ const MessagePayload = require('./MessagePayload');
 const { Error } = require('../errors');
 const DataResolver = require('../util/DataResolver');
 
+let Message;
```
vladfrangu, line 10: Completely out of curiosity, can we not do something like 
```js
const Message = (() => require('./Message').Message)();
```
for lazy loading?

vladfrangu (approved):

packages/discord.js/src/structures/Webhook.js
Syjalo: Then can't we just do?
```js
const { Message } = require('./Message');
```

vladfrangu: If it's done like its currently done in the PR it's because there's circular imports at hand. So..no

suneettipirneni: > Completely out of curiosity, can we not do something like
> 
> ```js
> const Message = (() => require('./Message').Message)();
> ```
> 
> for lazy loading?

Doesn't seem to work, still gives me a circular import warning.

kyranet: Circular dependencies are triggered when A requires B, B requires A, and A is empty.

This is why many files define their requires at the bottom, right after `module.exports =` is defined, so A isn't empty, B can resolve A, and then A can resolve B.

A prettier alternative is to make a `lazy` method utility (similar to Rust's `lazy_static`, sorta):

```typescript
export function lazy<T>(cb: () => T) {
	let defaultValue: T;

	return () => (defaultValue ??= cb());
}
```

And then we do the following:

```js
const getMessage = lazy(() => require('./Message').Message);
```

Then once we do `getMessage();`, it gets cached and has no performance impact over the next few operations. This approach also deduplicates requires and is cleaner:

```diff
- let Message;
+ const getMessage = lazy(() => require('./Message').Message);
  // ...
- Message ??= require('./Message').Message;
+ const Message = getMessage();
```

suneettipirneni: Does it make any performance difference for node.js requires?

Node.js caches modules after the first time they're loaded anyway:

> Modules are cached after the first time they are loaded. This means (among other things) that every call to require('foo') will get exactly the same object returned, if it would resolve to the same file.

> Provided require.cache is not modified, multiple calls to require('foo') will not cause the module code to be executed multiple times.

DraftProducts (approved):

packages/discord.js/src/structures/Webhook.js
kyranet: Yes, because it needs to resolve a relative path into an absolute one, among many other operations.

suneettipirneni: Even with the `lazy()` function I still get a circular import error.

kyranet: Then you're using it wrong.

You're supposed to use it inside the code that needs it, not at the very top, which would be the same as having a top-level require.

kyranet (approved):

packages/discord.js/src/structures/ContextMenuCommandInteraction.js
```diff
@@ -48,7 +48,9 @@ class ContextMenuCommandInteraction extends CommandInteraction {
         name: 'message',
         type: '_MESSAGE',
         value: target_id,
-        message: this.channel?.messages._add(resolved.messages[target_id]) ?? resolved.messages[target_id],
+        message:
+          this.channel?.messages._add(resolved.messages[target_id]) ??
+          new (require('./Message').Message)(resolved.messages[target_id]),
```
imranbarbhuiya, line 53: this one isn't using lazy loading

imranbarbhuiya (changes requested):

iCrawl (approved):

SpaceEEC (approved):

packages/discord.js/src/structures/ModalSubmitInteraction.js
```diff
@@ -34,9 +34,9 @@ class ModalSubmitInteraction extends Interaction {
     if ('message' in data) {
       /**
        * The message associated with this interaction
-       * @type {?(Message|APIMessage)}
+       * @type {?Message}
        */
-      this.message = this.channel?.messages._add(data.message) ?? data.message;
+      this.message = this.channel?.messages._add(data.message) ?? new (require('./Message').Message)(data.message);
```
SerenModz21, line 39: lazy loading is also missing here

PR #7918, opened 2022-05-13
fix(sequentialhandler): account for 2xx status

nev-r:
correct an issue where SequentialHandler.runRequest verified success with a 200 http status code, but emoji creation returns a 201

standard emoji upload fails right now because it returns a 201 instead of 200, this causes case fallthrough until "null" is returned for the API request.

this results in `GuildEmojiCreateAction.handle` trying to access `createdEmoji.id` while `createdEmoji` is `nul`, and crashes

```
 const already = guild.emojis.cache.has(createdEmoji.id);
                                                        ^
TypeError: Cannot read properties of null (reading 'id')
    at GuildEmojiCreateAction.handle
```

repro script:
[emoji-failure.js.txt](https://github.com/discordjs/discord.js/files/8691314/emoji-failure.js.txt)

### DIFF
--- a/packages/rest/src/lib/handlers/SequentialHandler.ts
+++ b/packages/rest/src/lib/handlers/SequentialHandler.ts
@@ -391,7 +391,7 @@ export class SequentialHandler implements IHandler {
 			}
 		}
 
-		if (status === 200|| status === 201) {
+		if (status === 200 || status === 201) {
 			return parseResponse(res);
 		} else if (status === 429) {
 			// A rate limit was hit - this may happen if the route isn't associated with an official bucket hash yet, or when first globally rate limited
--- a/packages/rest/src/lib/handlers/SequentialHandler.ts
+++ b/packages/rest/src/lib/handlers/SequentialHandler.ts
@@ -391,7 +391,7 @@ export class SequentialHandler implements IHandler {
 			}
 		}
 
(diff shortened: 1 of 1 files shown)


nev-r:
i know i know. sorry

packages/rest/src/lib/handlers/SequentialHandler.ts
```diff
@@ -391,7 +391,7 @@ export class SequentialHandler implements IHandler {
 			}
 		}
 
-		if (status === 200) {
+		if (status === 200 || status === 201) {
```
suneettipirneni, line 394: For future proofing we should accept anything within the ok range, like we did before.

```suggestion
		if (status >= 200 && status <= 299) {
```

suneettipirneni (changes requested):

packages/rest/src/lib/handlers/SequentialHandler.ts
nev-r: sounds good since it previously relied on Response.ok which refers to a 2xx code.
to match the formatting of other status checks, seems like it should be a `< 300` though?

suneettipirneni: Yeah I've updated my diff.

PR #7919, opened 2022-05-13
fix: ok statusCode can be 200..299

KhafraDev:

Expands upon #7918 with a test.

Sorry I forgot <response>.ok accounted for status codes in the 2xx range.


iCrawl (approved):

vladfrangu (approved):
Make sure this still works with 204 no content, otherwise lgtm

KhafraDev:
This is just an issue regarding the switch from fetch -> undici's request. Before it used <response>.ok (status is between 200..299), but I accidentally only checked for 200 in my pr. Unless the 204 no content has an `application/json` content-type header, it will still parse the body as an empty array buffer.

SpaceEEC (approved):

PR #7925, opened 2022-05-15
feat: @discordjs/proxy

didinele:
Due to the unusual nature of this PR, I will be mostly opting out of the template.

This PR introduces a new package to the monorepo - `@discordjs/proxy`. As the name implies, this package would enable users to create proxies for their REST calls.

## Motivation and reasoning
Prior to #7747, there was no easy, non-hacky way to have "shared rate-limiting", which made things like offloading specific workloads to a different service or even just a `worker_thread` dangerous. Now that it's become possible, users might be already inclined to spin up a [twilight proxy](https://github.com/twilight-rs/http-proxy) (or other Discord HTTP proxies!) to send their requests through.

This is *fine*, but realistically, I feel like proxying will mostly be done by power users, who may not have their use cases fulfilled by existing proxies. It's also a big turn-off that twilight is the most recognizable one, but is written in Rust and as such can feel a bit daunting to fork and modify to one's needs.

## Goals
This is intended to be a library. It should most certainly not be specific to any HTTP framework, but rather expose middleware/request handler functions that would be compatible with most. Later down the road (probably out of scope for this PR) we could be somehow shipping "binaries" (i.e. docker images) with the more common defaults, making it extremely easy to spin a proxy up, while still enabling the creation of custom ones.

## How would this work?
Simply put, we would be using `@discordjs/rest`. Our request handler would essentially just forward whatever request comes through, strip away the rate-limit headers, and forward the response from Discord. Stripping rate-limit headers is needed so whatever REST handler receives this response thinks there are no rate limits, and as such doesn't try to deal with them.

## Current caveats
- Response headers are not forwarded for non-2xx responses.
- As `REST` does not support dictating the API version to use per-request, the `/api/v[x]` part of the URL is dropped and ignored by the current implementation, as such, the API version used is dictated solely by the proxy, not by the remote upstream.

**WIP**:
- [x] initialize package
- [x] finalize signature for `proxyRequests` - opted to only take in a `REST` instance
- [x] wait for #7929 to get merged
- [x] parse request URL properly, as `REST` always appends `/api/v[apiversion]` to it
- [x] figure out a more correct way to handle invalid methods - opted for just letting Discord throw Method Not Allowed
- [x] ~~header to dictate if 429s should be retried~~ - not possible with current REST, unsure if it'll still be implemented
- [x] better modularization, enabling easily writing custom handlers
- [x] write docs
- [x] write tests


packages/proxy/cliff.toml
```diff
@@ -0,0 +1,62 @@
...
+    { message = "^revert", skip = true},
+    { message = "^style", group = "Styling"},
+    { message = "^test", group = "Testing"},
+    { message = "^chore", skip = true},
+    { message = "^ci", skip = true},
+    { message = "^build", skip = true},
+    { body = ".*security", group = "Security"},
+]
+filter_commits = true
+tag_pattern = "@discordjs\\/rest@.*"
```
imranbarbhuiya, line 58:
```suggestion
tag_pattern = "@discordjs\\/proxy@.*"
```

imranbarbhuiya (changes requested):

packages/proxy/cliff.toml
didinele: good catch. i honestly just copy pasted those, lol.

packages/proxy/__tests__/index.test.ts
```diff
@@ -0,0 +1,83 @@
+import { createServer } from 'http';
```
SuperchupuDev, line 1:
```suggestion
import { createServer } from 'node:http';
```

packages/proxy/src/index.ts
```diff
@@ -0,0 +1,64 @@
+import type { IncomingMessage, ServerResponse } from 'http';
```
SuperchupuDev, line 1:
```suggestion
import type { IncomingMessage, ServerResponse } from 'node:http';
```

SuperchupuDev (commented):
really small thing but this is how native node modules are imported in every other djs package

packages/proxy/package.json
```diff
@@ -0,0 +1,88 @@
...
+		"eslint-config-marine": "^9.4.1",
+		"eslint-config-prettier": "^8.5.0",
+		"eslint-plugin-import": "^2.26.0",
+		"jest": "^27.5.1",
+		"prettier": "^2.6.2",
+		"supertest": "^6.2.3",
+		"tsup": "^5.12.5",
+		"typedoc": "^0.22.15",
+		"typescript": "^4.6.3",
+		"undici": "^5.2.0"
```
Syjalo, line 80: Shouldn't undici be in the main dependencies?

Syjalo (changes requested):

packages/proxy/package.json
didinele: Given that I only use it for types, no.

iCrawl: Only if it is planned to be used as a library, mind you.

didinele: > Only if it is planned to be used as a library, mind you.

right, forgot about this aspect, thanks.

didinele:
With #7929 merged and rebased into this branch, it's now ready for review.

packages/proxy/src/handlers/proxyRequests.ts
```diff
@@ -0,0 +1,46 @@
...
+
+		if (!method || !rawUrl) {
+			throw new TypeError(
+				'Invalid request. Missing method and/or url, implying that this is not a Server IncomingMesage',
+			);
+		}
+
+		// The 2nd parameter is here so the URL constructor doesn't complain about an "invalid url" when the origin is missing
+		// we don't actually care about the origin and the value passed is irrelevant
+		const url = new URL(rawUrl, 'http://noop').pathname.replace(/^\/api(\/v\d+)?/, '') as RouteLike;
```
Syjalo, line 22: It's a route, not a URL. I think the variable should be named correctly.

Syjalo (changes requested):

Milo123459:
Is this going to have support for disabling the ratelimiter?

monbrey:
> Is this going to have support for disabling the ratelimiter?

Why? So you can just get rate limited and throw errors?

Milo123459:
> > Is this going to have support for disabling the ratelimiter?
> 
> Why? So you can just get rate limited and throw errors?

Some proxies handle ratelimiting for you.

didinele:
> Some proxies handle ratelimiting for you.

Unsure what you mean with this in the given context - but no, the whole point of proxying is to have shared ratelimiting.

The only configurable behavior here comes from passing in your own `REST` instance, which will allow you to pick where you want ratelimit retries to take place.

Our default binary (once we start shipping those) will leave upstream to handle ratelimiting - i.e. the proxy will forward 429s immediately.

Milo123459:
Twilight proxy (the one you linked) supports ratelimiting.

ckohen:
> Is this going to have support for disabling the ratelimiter?

Which rate limiter are you talking about, think that's the miscommunication here.

Milo123459:
> > Is this going to have support for disabling the ratelimiter?
> 
> Which rate limiter are you talking about, think that's the miscommunication here.

Does discord.js not handle ratelimits? That's the one I'm talking about, the one that is built-in

packages/proxy/src/handlers/proxyRequests.ts
```diff
@@ -0,0 +1,52 @@
...
+		} catch (error) {
+			if (error instanceof DiscordAPIError) {
+				populateGeneralErrorResponse(res, error);
+			} else if (error instanceof RateLimitError) {
+				populateRatelimitErrorResponse(res, error);
+			} else {
+				// Unclear if there's better course of action here. Any web framework allows to pass in an error handler for something like this
+				// at which point the user could dictate what to do with the error - otherwise we could just 500
+				throw error;
+			}
```
kyranet, line 47: Other errors thrown here are `AbortError` (thrown by [`AbortController`](https://developer.mozilla.org/en-US/docs/Web/API/AbortController)) and [`HTTPError`](https://github.com/discordjs/discord.js/blob/main/packages/rest/src/lib/errors/HTTPError.ts).

What's the behaviour in the proxy when either of those two happen?

packages/proxy/__tests__/proxyRequests.test.ts
```diff
@@ -0,0 +1,83 @@
...
+					...responseOptions.headers,
+					'x-ratelimit-limit': '10',
+				},
+			},
+		}));
+
+	const res = await supertest(server).get('/api/v10/simpleGet');
+	const headers = res.headers as Record<string, string>;
+
+	expect(headers['content-type'].startsWith('application/json')).toBe(true);
```
kyranet, line 58: We can also use [`expect.stringContaining`](https://jestjs.io/docs/expect#expectstringcontainingstring) or [`expect.stringMatching`](https://jestjs.io/docs/expect#expectstringmatchingstring--regexp).

Should look like:

```typescript
expect(headers['content-type']).toEqual(expect.stringMatching(/^application\/json/));
```

This way, when the test fails, it shows a much better error.

packages/proxy/__tests__/proxyRequests.test.ts
```diff
@@ -0,0 +1,83 @@
...
+	// Ratelimit headers should be dropped
+	expect(headers['x-ratelimit-limit']).toBe(undefined);
```
kyranet, line 60: Likewise here, using [`expect.not.toHaveProperty`](https://jestjs.io/docs/expect#tohavepropertykeypath-value):

```typescript
expect(headers).not.toHaveProperty('x-ratelimit-limit');
```

packages/proxy/__tests__/proxyRequests.test.ts
kyranet: If this works, please also apply it on line 80 too 👀

kyranet (changes requested):
Just a few things, otherwise LGTM

packages/proxy/src/util/responseHelpers.ts
```diff
@@ -0,0 +1,46 @@
...
+
+/**
+ * Populates a server response with the data from a Discord 2xx REST response
+ * @param res The server response to populate
+ * @param data The data to populate the response with
+ */
+export async function populateOkResponse(res: ServerResponse, data: Dispatcher.ResponseData): Promise<void> {
+	res.statusCode = data.statusCode;
+
+	for (const header of Object.keys(data.headers)) {
```
vladfrangu, line 13: Use Object.entries

vladfrangu (changes requested):

packages/proxy/src/util/responseHelpers.ts
kyranet: If we always used the value, sure, but we're skipping them. As such, for performance reasons, it may be wise to keep this as is.

vladfrangu: Is there any actual performance hit tho?

kyranet: Considering it has to allocate an array for every property... most likely?

packages/proxy/src/handlers/proxyRequests.ts
didinele: For AbortError I'm unsure what the behavior should be, now that you mention it.

Right now the behavior would be essentially silent failure on the proxy end which would cause god knows what on the client side.

I think the fix is to make the proxy NOT abort (i.e. make the timeout fairly high) so the client can do it.

So say (just throwing numbers out, I don't know what the defaults are right now), the proxy aborts after 2 minutes, meaning the only supported behavior is to have clients abort in less than that.

No other good way I can think of to deal with this - other than intentionally letting the request hang when an AbortError is thrown - but that seems equally scuffed.


As per HTTPError, I missed that. When is it thrown? 5xx after multiple retries? I thought that'd fall under DiscordAPIError initially. Will cover this.

kyranet: AbortError may return 408 (Request Timeout), we will need to handle this in /rest. Likely also make a TimedOutError.

> The server timed out waiting for the request. According to HTTP specifications: "The client did not produce a request within the time that the server was prepared to wait. The client MAY repeat the request without modifications at any later time."

Although I think 504 (Gateway Timeout) would probably be better, but it mixes up with Discord's errors, which can return this.

> The server was acting as a gateway or proxy and did not receive a timely response from the upstream server.

There's also nginx's unofficial code, 444 (No Response):

> Used internally to instruct the server to return no information to the client and close the connection immediately.

And Cloudflare's 524 (A Timeout Occurred), but might mix up with Discord too:

> Cloudflare was able to complete a TCP connection to the origin server, but did not receive a timely HTTP response.

We could also use an error Discord uses and make /rest differentiate between the two via headers, but I don't think it's really a good idea.

---

HTTPError is thrown on 5XX, it's used specifically because those generally don't give us a JSON body.

Also, I was thinking of also using error.cause, which is available since Node.js v16.10.0, in case we want to expose the original error as well, or we could just do it, since it's backwards compatible.

SpaceEEC (approved):

packages/proxy/src/util/responseHelpers.ts
```diff
@@ -0,0 +1,54 @@
...
+		// Strip ratelimit headers
+		if (header.startsWith('x-ratelimit')) {
+			continue;
+		}
+
+		res.setHeader(header, data.headers[header]!);
+	}
+
+	const parsed = await parseResponse(data);
+	res.write(data.headers['content-type']?.startsWith('application/json') ? JSON.stringify(parsed) : parsed);
```
kyranet, line 23: I suppose... there's no way to pass raw data back? It would be even better to be able to pass the readable body, similar to this:

Assuming `data` is a `Readable`, which at least the body in HTTP requests are:
```javascript
for await (const chunk of data) res.write(chunk);
```

kyranet (changes requested):
Just two more doubts, otherwise LGTM.

packages/proxy/src/util/responseHelpers.ts
vladfrangu: Couldn't you just pipe the response directly?

didinele: I believe you mean `data.body`?

didinele: Oh! My bad. GitHub didn't show those comments re piping until I refreshed. Let me have a look at that.

kyranet (approved):

packages/proxy/README.md
```diff
@@ -0,0 +1,46 @@
...
+    <a href="https://discord.gg/djs"><img src="https://img.shields.io/discord/222078108977594368?color=5865F2&logo=discord&logoColor=white" alt="Discord server" /></a>
+    <a href="https://www.npmjs.com/package/@discordjs/proxy"><img src="https://img.shields.io/npm/v/@discordjs/proxy.svg?maxAge=3600" alt="npm version" /></a>
+    <a href="https://www.npmjs.com/package/@discordjs/proxy"><img src="https://img.shields.io/npm/dt/@discordjs/proxy.svg?maxAge=3600" alt="npm downloads" /></a>
+    <a href="https://github.com/discordjs/discord.js/actions"><img src="https://github.com/discordjs/discord.js/actions/workflows/test.yml/badge.svg" alt="Build status" /></a>
+  </p>
+</div>
+
+## About
+
+`@discordjs/proxy` is a powerful set of tools for running an HTTP proxy in front of Discord's API
```
vladfrangu, line 17: It's not a set of tools, it's...a wrapper for djs/rest to turn it into a proxy

vladfrangu (approved):
Small nit but otherwise LGTM

packages/proxy/src/handlers/proxyRequests.ts
```diff
@@ -0,0 +1,55 @@
...
+	populateSuccessfulResponse,
+	populateRatelimitErrorResponse,
+} from '../util/responseHelpers';
+import type { RequestHandler } from '../util/util';
+
+/**
+ * Creates an HTTP handler used to forward requests to Discord
+ * @param rest REST instance to use for the requests
+ */
+export function proxyRequests(rest: REST): RequestHandler {
```
splatterxl, line 15: Personally would prefer `createRequestProxy` here

iCrawl (approved):

PR #7929, opened 2022-05-15
feat: REST#raw

didinele:
This PR adds a new `REST#raw` method, enabling users to get raw response objects and consume the body themselves, also exposing the `parseResponse` function.

### DIFF
--- a/packages/rest/src/lib/REST.ts
+++ b/packages/rest/src/lib/REST.ts
@@ -13,6 +13,7 @@ import {
 import type { HashData } from './RequestManager';
 import type { IHandler } from './handlers/IHandler';
 import { DefaultRestOptions, RESTEvents } from './utils/constants';
+import { parseResponse } from './utils/utils';
 
 /**
  * Options to be passed when creating the REST instance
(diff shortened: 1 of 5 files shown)


packages/rest/src/lib/REST.ts
```diff
@@ -314,7 +315,14 @@ export class REST extends EventEmitter {
...
 	 * @param options Request options
 	 */
-	public request(options: InternalRequest) {
+	public async request(options: InternalRequest) {
+		return parseResponse(await this.requestManager.queueRequest(options));
+	}
+
+	/**
+	 * Runs a request from the api, yielding the raw Response object
+	 */
```
kyranet, line 324: Missing the doc explaining what the parameter is:
```suggestion
	 * Runs a request from the api, yielding the raw Response object
	 * @param options Request options
	 */
```

kyranet (changes requested):

Syjalo:
Wouldn't it be better to add `rawResponse` request option?

didinele:
> Wouldn't it be better to add rawResponse request option?

That was initially the plan and Vlad's suggestion, but Crawl told me to make it a method.

Syjalo:
With the option we will able to get raw response from any request method (`.request()`, `.get()`, `.post()`, etc) and we will be able to set default value for the option in REST options. Also `REST#raw()` doesn't say what the method does.

vladfrangu (approved):

iCrawl (approved):

PR #7954, opened 2022-05-20
Update bug_report.yml

AhmadShkour71:


ImRodry:
Usually when we make PRs we give them descriptions. Anyway, this PR is incorrect because version 13 is very much supported as it is the latest available in a stable manner

PR #7965, opened 2022-05-24
fix: typings

markisha64:
Typings were wrong for channel create/edit, some properties are nullable but weren't set as such. When using Typescript this could force workarounds like // @ts-ignore which go against the point of Typescript.

### DIFF
--- a/packages/discord.js/src/managers/GuildChannelManager.js
+++ b/packages/discord.js/src/managers/GuildChannelManager.js
@@ -99,7 +99,7 @@ class GuildChannelManager extends CachedManager {
   /**
    * Options used to create a new channel in a guild.
    * @typedef {CategoryCreateChannelOptions} GuildChannelCreateOptions
-   * @property {CategoryChannelResolvable} [parent] Parent of the new channel
+   * @property {CategoryChannelResolvable|null} [parent] Parent of the new channel
    */
 
   /**
--- a/packages/discord.js/src/managers/GuildChannelManager.js
+++ b/packages/discord.js/src/managers/GuildChannelManager.js
@@ -199,7 +199,7 @@ class GuildChannelManager extends CachedManager {
    * @property {string} [name] The name of the channel
    * @property {ChannelType} [type] The type of the channel (only conversion between text and news is supported)
    * @property {number} [position] The position of the channel
-   * @property {string} [topic] The topic of the text channel
+   * @property {string|null} [topic] The topic of the text channel
    * @property {boolean} [nsfw] Whether the channel is NSFW
    * @property {number} [bitrate] The bitrate of the voice channel
    * @property {number} [userLimit] The user limit of the voice channel
--- a/packages/discord.js/src/managers/GuildChannelManager.js
+++ b/packages/discord.js/src/managers/GuildChannelManager.js
@@ -199,7 +199,7 @@ class GuildChannelManager extends CachedManager {
    * @property {string} [name] The name of the channel
    * @property {ChannelType} [type] The type of the channel (only conversion between text and news is supported)
    * @property {number} [position] The position of the channel
-   * @property {string|null} [topic] The topic of the text channel
+   * @property {?(string|null)} [topic] The topic of the text channel
    * @property {boolean} [nsfw] Whether the channel is NSFW
    * @property {number} [bitrate] The bitrate of the voice channel
    * @property {number} [userLimit] The user limit of the voice channel
--- a/packages/discord.js/src/managers/GuildChannelManager.js
+++ b/packages/discord.js/src/managers/GuildChannelManager.js
@@ -99,7 +99,7 @@ class GuildChannelManager extends CachedManager {
   /**
    * Options used to create a new channel in a guild.
    * @typedef {CategoryCreateChannelOptions} GuildChannelCreateOptions
-   * @property {CategoryChannelResolvable|null} [parent] Parent of the new channel
+   * @property {?CategoryChannelResolvable} [parent] Parent of the new channel
    */
 
   /**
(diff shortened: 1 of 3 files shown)


packages/discord.js/src/managers/GuildManager.js
```diff
@@ -84,7 +84,7 @@ class GuildManager extends CachedManager {
    * @property {Snowflake|number} [parentId] The parent id for this channel
    * @property {ChannelType|number} [type] The type of the channel
    * @property {string} name The name of the channel
-   * @property {string} [topic] The topic of the text channel
+   * @property {string|null} [topic] The topic of the text channel
```
almeidx, line 87:
```suggestion
   * @property {?string} [topic] The topic of the text channel
```

packages/discord.js/src/managers/GuildChannelManager.js
```diff
@@ -99,7 +99,7 @@ class GuildChannelManager extends CachedManager {
   /**
    * Options used to create a new channel in a guild.
    * @typedef {CategoryCreateChannelOptions} GuildChannelCreateOptions
-   * @property {CategoryChannelResolvable} [parent] Parent of the new channel
+   * @property {CategoryChannelResolvable|null} [parent] Parent of the new channel
```
almeidx, line 102:
```suggestion
   * @property {?CategoryChannelResolvable} [parent] Parent of the new channel
```

ImRodry: is this correct though? The property is already optional, why would you pass it as null?

almeidx: To unset the parent

markisha64: Setting it to null clears it while omitting it or setting it to undefined just ignores it. For example, if you wanted to delete the current topic by passing in null, you can't do it, at the very least, not without typescripts errors, even though thats how it's supposed to be done.

ImRodry: clears what? This is the typedef for creating a channel. You can't clear the parent of a channel you just created, that doesn't make sense

Jiralite: I believe the same point may be made to the above review about a channel's topic during guild creation additionally?

ImRodry: yes, I only commented on one but it applies to the entire PR tbh

monbrey: Yes, but for a different reason. That one is `PartialChannelData` not `CategoryCreateChannelOptions`, but we don't explicitly set properties to null in a Partial. It's just there or it isn't.

markisha64: There's been a misunderstanding on my side. Here's a better example, a non-partial non-category channel's parent value is a CategoryChannel or it's null. If we wanted to save a channel as JSON and use it to create new channels we face the issue that parent accepts CategoryChannelResolvable | undefined. Of which our channel parent value might be neither because it could be null. Sure you could just transform it to undefined but to me it doesn't really make sense to not allow null (because it works and it's accepted by Discord API). It just seems inconsistent. As for the channel topic, as of now, there isn't a way to clear a topic programmatically according to discord.js typings because it doesn't accept null. (setting it to undefined doesn't remove it).

monbrey: `BaseGuildTextChannel#setTopic` does accept null: <https://github.com/discordjs/discord.js/blob/fdeac9d9fba06c532eca296ddd8479047bc732bf/packages/discord.js/typings/index.d.ts#L507>

Arguably, you could also change `ChannelData` to accept `string | null`, as this is the correct typedef for the `GuildChannel#edit` method. `PartialChannelData` is only used when creating channels in the `GuildCreateOptions` payload, so `null` does not belong here.

> Sure you could just transform it to undefined but to me it doesn't really make sense to not allow null 

`null` and `undefined` are not the same thing. Trying to treat them as the same thing would be inconsistent, not the other way around. Transform it.

packages/discord.js/src/managers/GuildManager.js
```diff
@@ -84,7 +84,7 @@ class GuildManager extends CachedManager {
    * @property {Snowflake|number} [parentId] The parent id for this channel
    * @property {ChannelType|number} [type] The type of the channel
    * @property {string} name The name of the channel
-   * @property {string} [topic] The topic of the text channel
+   * @property {?string} [topic] The topic of the text channel
```
vladfrangu, line 87: Why is it `?string` here but `string|null` in ChannelData

vladfrangu (changes requested):

packages/discord.js/src/managers/GuildManager.js
markisha64: Yeah you're right, the partial data can be `undefined | string | null` (`null` being no topic set). Is there a definition for Partial Channel Data? The developer portal doesn't say anything and it is inconsistently defined inside discord.js. The typedef says that topic is always string while TS typing says it can be undefined.
https://github.com/discordjs/discord.js/blob/68d5169f66c96f8fe5be17a1c01cdd5155607ab2/packages/discord.js/src/managers/GuildManager.js#L79-L96
https://github.com/discordjs/discord.js/blob/68d5169f66c96f8fe5be17a1c01cdd5155607ab2/packages/discord.js/typings/index.d.ts#L4825-L4847

Jiralite: > The typedef says that topic is always string while TS typing says it can be undefined.

Incorrect - if you look closely, the topic in the JSDoc is wrapped in `[this]` which indicates an optional parameter, which is reflected accordingly in the typings. Not sure about needing to pass `null`, though... why pass `null` to create a channel alongside a topic when one can omit it?

markisha64: > Here's a better example, a non-partial non-category channel's parent value is a CategoryChannel or it's null. If we wanted to save a channel as JSON and use it to create new channels we face the issue that parent accepts CategoryChannelResolvable | undefined. Of which our channel parent value might be neither because it could be null. Sure you could just transform it to undefined but to me it doesn't really make sense to not allow null (because it works and it's accepted by Discord API). It just seems inconsistent.
Same as this. Omitting it and setting it to null does the same thing during creation and that's why I find it sensible to allow both rather then omitting it every time that I'm making a channel with no topic.

SpaceEEC (approved):

vladfrangu (approved):

kyranet (approved):

iCrawl (approved):

PR #7970, opened 2022-05-26
docs: add Client#raw event

utyfua:

Added missing documentation for Client#raw event

### DIFF
--- a/packages/discord.js/src/client/websocket/WebSocketShard.js
+++ b/packages/discord.js/src/client/websocket/WebSocketShard.js
@@ -299,6 +299,24 @@ class WebSocketShard extends EventEmitter {
       this.manager.client.emit(Events.ShardError, err, this.id);
       return;
     }
+
+    /**
+     * Emitted whenever a raw payload is received from discord.
+     *
+     * `packet.s` and `packet.t` are `null` when `packet.op` is not `0` (Gateway Dispatch Opcode).
+     *
+     * @event Client#raw
+     * @param {Object} packet [Gateway Payload](https://discord.com/developers/docs/topics/gateway#payloads)
+     * @param {number} packet.op
+     * [opcode](https://discord.com/developers/docs/topics/opcodes-and-status-codes#gateway-gateway-opcodes)
+     * for the payload
+     * @param {any} packet.d mixed event data (any JSON value)
+     * @param {number | null} packet.s sequence number, used for resuming sessions and heartbeats
+     * @param {string | null} packet.t the event name for this payload
+     * @param {number} shardId The shard's id in the manager
+     *
+     * @see {@link https://discord.com/developers/docs/topics/gateway#payloads}
+     */
     this.manager.client.emit(Events.Raw, packet, this.id);
     if (packet.op === GatewayOpcodes.Dispatch) this.manager.emit(packet.t, packet.d, this.id);
     this.onPacket(packet);


monbrey:
This was intentionally undocumented as we do not support listening to raw events.

utyfua:
> This was intentionally undocumented as we do not support listening to raw events.

This looks reasonable, but why should the documentation be missing something?

Maybe I should add `@internal` and will be fine?

utyfua:
> If you want the raw payload you should instead listen to `client.ws.on("WS_EVENT_NAME", cb)` instead of the blanko raw event. There is almost no reason not to adapt the handling within and selectively listen to ws events over using the raw event (not last because of the absolute carnage that event is on even just a reasonably sized guild).

I am very sorry, BUT I didn't edit any logic at all. 
I just made some documentation. 
So my changes will not increase the load in any way.

almostSouji:
What I was trying to get at is that the raw event is an internal API which may change at any release, even patch. We are not documenting it because people are very much not supposed to use it, nor does it bear any useful use case, as you (or anyone thinking they need to use raw) can (and should) instead handle the raw websocket payloads at the `client.ws` level.

utyfua:
> What I was trying to get at is that the raw event is an internal API which may change at any release, even patch. We are not documenting it because people are very much not supposed to use it, nor does it bear any useful use case, as you (or anyone thinking they need to use raw) can (and should) instead handle the raw websocket payloads at the `client.ws` level.

Sorry but we can not just do any listeners for `client.ws` cuz there no any listeners pop up in this object(even if its still an instance of ` EventEmitter`).
In your case I should do own manager for the [shards](https://discord.js.org/#/docs/discord.js/stable/class/WebSocketManager?scrollTo=shards) and then add an listener for each of the shards.
Why cant we just put `@internal` instead of placing nothing here?

almostSouji:
<https://discord.js.org/#/docs/discord.js/stable/class/WebSocketManager> extends regular node.js EventEmitter with the following note on the documentation:

> This class forwards raw dispatch events, read more about it here https://discord.com/developers/docs/topics/gateway 

Not too sure what you mean by "popping up"

utyfua:
Okay. Lets review hw its works in reality.

We a getting connection and setting listeners [here](https://github.com/utyfua/discord.js/blob/c5fd477e959da9f5a584ecd097b3a9cbbac80e5a/packages/discord.js/src/client/websocket/WebSocketShard.js#L259).
We can catch `connection` variable only by next _private_ path `client.ws.shards[].connection`. 
So in this case I should keep all shards in check.

Okay, let's see where the raw messages pop up.
[Here](https://github.com/utyfua/discord.js/blob/c5fd477e959da9f5a584ecd097b3a9cbbac80e5a/packages/discord.js/src/client/websocket/WebSocketShard.js#L261) we are setting listener. 
Then we are listening the raw data [here](https://github.com/utyfua/discord.js/blob/c5fd477e959da9f5a584ecd097b3a9cbbac80e5a/packages/discord.js/src/client/websocket/WebSocketShard.js#L281) and then [the `raw` event throwing](https://github.com/utyfua/discord.js/blob/c5fd477e959da9f5a584ecd097b3a9cbbac80e5a/packages/discord.js/src/client/websocket/WebSocketShard.js#L320) directly to `this.manager.client` (which original `client` are).

I want to reminder one more time I did only documentation for already existing realization.
I agree to add `@internal` if that's the case.

vladfrangu:
Heyo, let me correct and clarify some things.

Before I begin, I agree with Souji here over the fact we will not document the `raw` event anytime soon. It's volatile structure, combined with the fact it's super spammy, and the fact you should **not** use it unless you know what you're doing working with raw data are just a few examples as to why it won't be documented.

On the WebSocketManager (`client.ws`), we emit events based on the packet's `t` value. For instance, `client.ws.on('GUILD_MEMBER_ADD', (packet) => {})` will work (provided you have the intent required + a member joins the guild). We do not provide intellisense for this, and we never will, as Discord can add new payload types we might not have documented at that time. The line below shows how we emit that.

https://github.com/discordjs/discord.js/blob/68d5169f66c96f8fe5be17a1c01cdd5155607ab2/packages/discord.js/src/client/websocket/WebSocketShard.js#L303

`client.ws.shards[].connection` is STRICTLY a direct WebSocket connection to Discord and you **should never interact with it at all unless you 1000% know what you're doing**.

Hope this clarifies it.

utyfua:
Thank you for the answer.

Lets review next case.
I wanna create own audit log. 
How should I listen events? 
Should I do subscriptions for each of them?

PR #7977, opened 2022-05-28
fix: add missing update method in types

yokkkoso:
This PR return missing `update` method in `ModalSubmitInteraction` typings


Jiralite:
I'm not sure this is actually correct. You need to let TypeScript know that the modal came from a message via `ModalSubmitInteraction#isFromMessage()`. Once TypeScript knows it came from a message, the `update()` method will be present.

yokkkoso:
> I'm not sure this is actually correct. You need to let TypeScript know that the modal came from a message via `ModalSubmitInteraction#isFromMessage()`. Once TypeScript knows it came from a message, the `update()` method will be present.

Oh, okay, thank you!

PR #7986, opened 2022-06-01
types: add types to `EventEmitter` static methods

suneettipirneni:

Previously the following code:

```ts

await client.login(process.env.DISCORD_TOKEN);
const [readyClient] = await once(client, 'ready');
```

Would improperly type `readyClient` as `any`

With this PR the types are properly provided so it resolves to `Client<true>`

### DIFF
--- a/packages/discord.js/typings/index.d.ts
+++ b/packages/discord.js/typings/index.d.ts
@@ -185,8 +185,8 @@ import {
 declare module 'node:events' {
   class EventEmitter {
     // Add type overloads for client events.
-    static once<K extends keyof ClientEvents>(eventEmitter: Client, eventName: K): Promise<ClientEvents[K]>;
-    static on<K extends keyof ClientEvents>(eventEmitter: Client, eventName: K): AsyncIterator<ClientEvents[K]>;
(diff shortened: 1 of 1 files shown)


packages/discord.js/typings/index.d.ts
```diff
@@ -182,6 +182,14 @@ import {
   RawWidgetMemberData,
 } from './rawDataTypes';
 
+declare module 'node:events' {
+  class EventEmitter {
```
imranbarbhuiya, line 186: If you are changing types for `node:events` it'll change for user's `EventEmitter` too. Is it intentional?

imranbarbhuiya (changes requested):

packages/discord.js/typings/index.d.ts
suneettipirneni: Yes, it's overloaded with the djs client type as the event emitter as the first param. It has no effect on the event emitter types for anything else and still supports custom client events.

vladfrangu (approved):

SpaceEEC (approved):

iCrawl (approved):

PR #8000, opened 2022-06-04
feat: proxy container

didinele:
As the next step to #7925, this PR introduces another "package" (it's more of a container), with the purpose of being an easy-to-deploy HTTP proxy built on top of `@discordjs/proxy`

Additionally, this PR:
- bac20e84f9475be49b7943c7c8685f6a2a6f792e: fixes a hard-crash from `@discordjs/proxy`, since apparently `error.message` would contain "unsupported characters"
- d1832eb5b148a0dd1468eed53ec36ca2194762f9: documents `populateAbortErrorResponse` (apparently it had no doc comment, something that was missed from #7925, oops)
- 6f83219f9f36458f2c6fab4ef36e3e58cce58004: notable fix, `proxyRequests` now properly forwards `Content-Type`, no longer causing errors on non-GET requests

TODO:
- [x] Usage section in the README
- [x] Dockerfile
- [x] Publish workflow


packages/proxy-container/src/index.ts
```diff
@@ -0,0 +1,9 @@
+import { createServer } from 'node:http';
+import { proxyRequests } from '@discordjs/proxy';
+import { REST } from '@discordjs/rest';
+
+// We want to let upstream handle retrying
+const api = new REST({ rejectOnRateLimit: () => true, retries: 0 });
```
S222em, line 6:
```suggestion
const rest = new REST({ rejectOnRateLimit: () => true, retries: 0 });
```

didinele: I took the naming from [REST's tests](https://github.com/discordjs/discord.js/blob/3401fd4eb605d093209c6d2d2e2db1f5efce9f52/packages/rest/__tests__/RequestManager.test.ts). This doesn't really matter anyways

.github/workflows/publish-dev-docker.yml
```diff
@@ -0,0 +1,25 @@
+name: Publish dev Docker Images
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: '0 */12 * * *'
+jobs:
+  deploy:
+    name: Deploy
```
iCrawl, line 8: Can you call those 2 `docker` and `Docker` respectively?

.github/workflows/publish-dev-docker.yml
```diff
@@ -0,0 +1,25 @@
...
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v2
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v1
+
+      - name: Login to DockerHub
+        run: docker login -u ${{ secrets.DOCKERHUB_USERNAME }} -p ${{ secrets.DOCKERHUB_TOKEN }}
```
iCrawl, line 19:
```suggestion
        run: docker login -u ${{ secrets.DOCKER_USERNAME }} -p ${{ secrets. DOCKER_ACCESS_TOKEN }}
```

.github/workflows/publish-docker.yml
```diff
@@ -0,0 +1,26 @@
...
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v2
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v1
+
+      - name: Login to DockerHub
+        run: docker login -u ${{ secrets.DOCKERHUB_USERNAME }} -p ${{ secrets.DOCKERHUB_TOKEN }}
```
iCrawl, line 17:
```suggestion
        run: docker login -u ${{ secrets.DOCKER_USERNAME }} -p ${{ secrets.DOCKER_ACCESS_TOKEN }}
```

.github/workflows/publish-docker.yml
```diff
@@ -0,0 +1,26 @@
+name: Publish Docker Images
+on:
+  workflow_dispatch:
+jobs:
+  deploy:
+    name: Deploy
```
iCrawl, line 6:
```suggestion
  docker:
    name: Docker
```

.github/workflows/publish-dev-docker.yml
iCrawl, line 8:
```suggestion
  docker:
    name: Docker
```

packages/proxy-container/tsup.config.ts
```diff
@@ -0,0 +1,19 @@
+import { defineConfig } from 'tsup';
+
+export default defineConfig({
+	clean: true,
+	entryPoints: ['src/index.ts'],
+	format: ['esm', 'cjs'],
```
kyranet, line 6: Do we even need a multi-format bundle? For what's worth, this is an app and not a library.

iCrawl: True, we could just bundle as `esm` only 👍

kyranet: Should we also use minify and other options to keep bundle size the smallest?

iCrawl: Yes sure, we could.

didinele: Anything else other than `minify`?

iCrawl: Well, you could update to use the root `tsup.config.ts` config, since it has defaults already.

kyranet: `skipNodeModulesBundle: false`, `minify: true`, and I'm unsure if we need source maps, probably?

iCrawl: `skipNodeModulesBundle: false` that doesn't exactly bundle `node_modules`, you'd have to use another option which I have to look up.

didinele:
One more issue actually, I need to disable dts.

iCrawl:
That’s fair, it’s not a library. Or do you mean in the library?

didinele:
Meant in the container, taken care of.

packages/proxy-container/package.json
```diff
@@ -0,0 +1,75 @@
...
+		"changelog": "git cliff --prepend ./CHANGELOG.md -u -c ./cliff.toml -r ../../ --include-path 'packages/proxy-container/*'"
+	},
+	"main": "./dist/index.js",
+	"module": "./dist/index.mjs",
+	"typings": "./dist/index.d.ts",
+	"exports": {
+		"import": "./dist/index.mjs",
+		"require": "./dist/index.js",
+		"types": "./dist/index.d.ts"
+	},
```
kyranet, line 19: No exports, only `main` is to be kept, also missing the type to run the ESM code:
```suggestion
	"type": "module",
```

packages/proxy-container/package.json
```diff
@@ -0,0 +1,75 @@
...
+	"bugs": {
+		"url": "https://github.com/discordjs/discord.js/issues"
+	},
+	"homepage": "https://discord.js.org",
+	"dependencies": {
+		"@discordjs/proxy": "^0.1.0-dev",
+		"tslib": "^2.4.0"
+	},
+	"devDependencies": {
+		"@discordjs/ts-docgen": "^0.4.1",
```
kyranet, line 56: Does it need the docgen, tho? 🤔

kyranet (dismissed):

.github/workflows/publish-dev-docker.yml
```diff
@@ -0,0 +1,25 @@
...
+        uses: actions/checkout@v2
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v1
+
+      - name: Login to DockerHub
+        run: docker login -u ${{ secrets.DOCKER_USERNAME }} -p ${{ secrets.DOCKER_TOKEN }}
+
+      - name: Build the image
+        run: docker build -t discordjs/proxy:latest -f packages/proxy-container/Dockerfile .
```
vladfrangu, line 22: Shouldn't this use a different tag (like dev)?

vladfrangu (dismissed):

.github/workflows/publish-dev-docker.yml
didinele: No, Docker images are basically always `:latest`

vladfrangu: That's not true, and that doesn't mean we should publish dev releases as *latest*.

didinele: ¯\_(ツ)_/¯ I honestly don't see a big issue with publishing them as latest. Crawl was also in favor of doing it this way when I went over the workflows with him yesterday

vladfrangu (approved):

packages/proxy-container/package.json
```diff
@@ -0,0 +1,68 @@
+{
+	"name": "@discordjs/proxy-container",
+	"version": "0.1.0-dev",
+	"description": "Lightweight HTTP proxy for Discord's API, brought to you as a container 📦",
+	"scripts": {
+		"build": "tsup && tsc --noEmit --incremental",
```
iCrawl, line 6: Should this use `--incremental`?

packages/proxy-container/package.json
```diff
@@ -0,0 +1,68 @@
...
+		"@types/node": "^16.11.38",
+		"@typescript-eslint/eslint-plugin": "^5.27.0",
+		"@typescript-eslint/parser": "^5.27.0",
+		"eslint": "^8.17.0",
+		"eslint-config-marine": "^9.4.1",
+		"eslint-config-prettier": "^8.5.0",
+		"eslint-plugin-import": "^2.26.0",
+		"prettier": "^2.6.2",
+		"tsup": "^6.0.1",
+		"typedoc": "^0.22.17",
```
iCrawl, line 59: This shouldn't be needed either.

packages/proxy-container/package.json
didinele: probably not

iCrawl (approved):

packages/proxy-container/tsup.config.ts
```diff
@@ -0,0 +1,3 @@
+import { createTsupConfig } from '../../tsup.config';
+
+export default createTsupConfig({ minify: true, dts: false });
```
kyranet, line 3: We should probably write a TODO specifying that we need to setup the options to bundle node_modules, so the Docker image is as small as possible.

packages/proxy-container/src/index.ts
```diff
@@ -0,0 +1,15 @@
...
+import { REST } from '@discordjs/rest';
+
+if (!process.env.DISCORD_TOKEN) {
+	throw new Error('A DISCORD_TOKEN env var is required');
+}
+
+// We want to let upstream handle retrying
+const api = new REST({ rejectOnRateLimit: () => true, retries: 0 }).setToken(process.env.DISCORD_TOKEN);
+// eslint-disable-next-line @typescript-eslint/no-misused-promises
+const server = createServer(proxyRequests(api));
```
kyranet, line 12: Where is the misused promise, and why can't we just `await` it, using TLA, which is supported in ESM?

didinele: Misused promise is because `createServer` takes `(req, res) => void` and `proxyRequests` returns `(req, res) => Promise<void>`. Nothing we can await here (not that there is much of a point, given on those node versions uncaught promises like that will just crash the process)

kyranet (approved):

JosPoka:
@SpaceEEC Please review

SpaceEEC (approved):

PR #8010, opened 2022-06-05
refact: Sticker's `tags` property

markisha64:
So, after working with stickers I've realized that the Sticker's `tags` property doesn't really make that much sense. As of now its type is `string[]` even though the Discord API returns `string`.

For every sticker the string value is parsed with `.split(', ')` which doesn't make sense considering that only standard stickers have `tags` as  a comma separated list of keywords. This adds overhead with no purpose as you can't edit standard stickers and guild stickers ignore the commas and use the whole string for suggestion. Further more, this makes sticker comparison slower because it's comparing arrays of strings instead of just strings (O(n*k) vs O(n)).

It's also inconsistent with the create/edit methods as they accept tags: `string` and nowhere is it mentioned why the tags are/would be an array of strings apart from the developer docs.

https://discord.com/developers/docs/resources/sticker#sticker-resource

If people want to iterate over standard sticker's `tags` (for whatever reason) they should split the string themselves, no reason to always parse the string.

### DIFF
--- a/packages/discord.js/src/structures/Sticker.js
+++ b/packages/discord.js/src/structures/Sticker.js
@@ -70,10 +70,10 @@ class Sticker extends Base {
 
     if ('tags' in sticker) {
       /**
-       * An array of tags for the sticker
-       * @type {?string[]}
+       * Autocomplete/suggestions for the sticker
+       * @type {?string}
        */
-      this.tags = sticker.tags.split(', ');
+      this.tags = sticker.tags;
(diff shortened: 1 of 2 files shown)


vladfrangu (approved):

iCrawl (approved):

advaith1:
just wanted to mention that the array code was written long before guild stickers existed, so it made more sense to always treat it as an array.

PR #8020, opened 2022-06-06
refactor(Users): remove global user cache

### LINKED ISSUE #7474: Roadmap: global user cache removal
Initially, we made a global user cache because they exist only once in Discord, however, that has brought many caching issues that eventually required fixes that damage the overall performance, see #6782.

With the release of intents, we can really limit the amount of `GuildMember` instances we have, and with it, also the amount of `User` instances.

There's also a huge memory leak that has been there since caching has been added to the library: if a bot leaves a guild, all the users that shared only said guild with the bot would sit in memory, forever, and eventually become stale or invalid were we to read their properties when they update something while we don't share a guild with them.

To mitigate this, #6013 added a cache system that allowed us to limit how many entries we store in our CachedManagers, and later #6825 improved the sweeping performance. However, by limiting or sweeping `User`s in `client.users.cache`, you're only removing one of the many references, in fact, they stay alive at the `GuildMember` instance: https://github.com/discordjs/discord.js/blob/7959a68d8ec600af248f5506f39871cae7eeeb04/packages/discord.js/src/structures/GuildMember.js#L60-L67 And if it's swept and the user joins another guild, they'll just get a second `User` instance, completely independent of the first one.

In fact, because `GuildMember` has a transient relationship with `User`, some funny things happen. For example when discord.js fetches guild members and the developer decides to specify `cache: false`, the `GuildMember`s are not cached but their respective `User`s are instantiated and cached regardless (see code above, thanks @suneettipirneni).

We could just make `GuildMember#user` a getter so we don't duplicate `User`s, **but**, not only that would worsen the developer experience (since it becomes nullable), but it also doesn't solve the memory leak at all.

Alternatively, we could also sweep all entries from `client.users.cache` when the bot leaves a guild, but for large internal sharded bots, that's going to be a huge performance hit, and definitely not something we should do.

And last but not least, what's the use case for `client.users.cache` to even exist? Most users use `client.users.fetch(UserResolvable)` but very often:
- Leave `cache` default, which means it'll always cache (as it defaults to `true`). This has a pitfall, if the user doesn't share a guild, it'll eventually be stale and incorrect, plus it'll be contributing to the memory leak.
- Have to use `force` due to the aforementioned issue. If you cache a user that doesn't share a guild with the bot (e.g. for a `whois` command), and they change their avatar, the library will, by default, retrieve the entry from the cache, which has a stale avatar, and as such, the command will fail to display correctly. Using `force` disregards the cache, and ends up making the cache useless again.

## A solution

Starting with #6013, Discord.js has a manager called `DataManager`, it is similar to `CachedManager`, but doesn't have a cache. We can use this so we're still able to do `client.users.resolve[Id](UserResolvable)`. We may also move `client.user` to `client.users.me` or `client.users.self`, so we don't have two different properties.

With this approach, we'll stop caching `User`s globally. Now we need to store them somewhere: inside `GuildMember`. This approach makes sense, because if the guild member leaves a guild, both the `GuildMember` and the `User` instance will be disposed from memory. Memory leak solved!

And because data duplication may be controversial (and some people may object to this proposal because of it), let me sum up some points:

- Sweeping users can and will result on data duplication due to hard references inside `GuildMember` and new ones not finding the previously created `User` instances.
- The issue from #6782 exists, and the approach it takes duplicates `User` instances anyways.
- As @suneettipirneni pointed out, if you specify `cache: false` in `GuildMemberManager#fetch`, members won't be cached, but users will.

There may be possible problems with the removal of the global cache, specially for guild data, since we won't be able to get the user data if the member leaves the guild. One place that would be very notable would be `Message#author`, however, it turns out that property [holds a hard reference](https://github.com/discordjs/discord.js/blob/7959a68d8ec600af248f5506f39871cae7eeeb04/packages/discord.js/src/structures/Message.js#L91-L99) just like `GuildMember#user` does.

ckohen:

> **Warning**
> This PR has not yet been fully tested, testing is still under way

- This PR is an attempt to satisfy the roadmap issue and will close #7474 

Most of what this PR achieves and why is outlined perfectly by that issue, differences and other notes here:
- `UserManager` extends `BaseManager` to not have a false `cache` property, but still retains `resolve(Id)`
- Members are the primary source of truth for user data
- Presences are a potential primary source of truth, but they create members anyways (and if you have the presence intent it seems highly likely you have the members intent as well)
- There's a lot of places like `Invite` where we're creating a `User` instance that can get stale quickly, is this okay? Yes, it is highly likely that the exact same thing was happening before but went unnoticed.
- `client.user` has been changed into a getter instead of removing it completely for now

### DIFF
--- a/packages/discord.js/src/client/actions/UserUpdate.js
+++ b/packages/discord.js/src/client/actions/UserUpdate.js
@@ -7,27 +7,21 @@ class UserUpdateAction extends Action {
   handle(data) {
     const client = this.client;
 
-    const newUser = data.id === client.user.id ? client.user : client.users.cache.get(data.id);
+    if (data.user.id !== client.users.me.id) return { old: null, updated: null };
+
+    const newUser = client.users.me;
     const oldUser = newUser._update(data);
 
-    if (!oldUser.equals(newUser)) {
-      /**
-       * Emitted whenever a user's details (e.g. username) are changed.
-       * Triggered by the Discord gateway events USER_UPDATE, GUILD_MEMBER_UPDATE, and PRESENCE_UPDATE.
-       * @event Client#userUpdate
-       * @param {User} oldUser The user before the update
-       * @param {User} newUser The user after the update
-       */
-      client.emit(Events.UserUpdate, oldUser, newUser);
-      return {
-        old: oldUser,
-        updated: newUser,
-      };
-    }
-
+    /**
+     * Emitted whenever the client user's details (e.g. username) are changed.
+     * @event Client#userUpdate
+     * @param {ClientUser} oldUser The user before the update
+     * @param {ClientUser} newUser The user after the update
+     */
+    client.emit(Events.UserUpdate, oldUser, newUser);
     return {
-      old: null,
-      updated: null,
+      old: oldUser,
+      updated: newUser,
     };
   }
 }
--- a/packages/discord.js/src/managers/UserManager.js
+++ b/packages/discord.js/src/managers/UserManager.js
@@ -1,7 +1,7 @@
 'use strict';
 
 const { ChannelType, Routes } = require('discord-api-types/v10');
-const CachedManager = require('./CachedManager');
+const BaseManager = require('./BaseManager');
 const { GuildMember } = require('../structures/GuildMember');
 const { Message } = require('../structures/Message');
 const ThreadMember = require('../structures/ThreadMember');
@@ -11,17 +11,16 @@ const User = require('../structures/User');
  * Manages API methods for users and stores their cache.
  * @extends {CachedManager}
  */
-class UserManager extends CachedManager {
-  constructor(client, iterable) {
-    super(client, User, iterable);
+class UserManager extends BaseManager {
+  constructor(client) {
+    super(client);
+
+    /**
+     * User that the client is logged in as
+     * @type {?ClientUser}
+     */
+    this.me = null;
   }
-
-  /**
-   * The cache of this manager
-   * @type {Collection<Snowflake, User>}
-   * @name UserManager#cache
-   */
-
   /**
    * Data that resolves to give a User object. This can be:
    * * A User object
@@ -77,28 +76,22 @@ class UserManager extends CachedManager {
   /**
    * Obtains a user from Discord, or the user cache if it's already available.
    * @param {UserResolvable} user The user to fetch
-   * @param {BaseFetchOptions} [options] Additional options for this fetch
    * @returns {Promise<User>}
    */
-  async fetch(user, { cache = true, force = false } = {}) {
+  async fetch(user) {
     const id = this.resolveId(user);
-    if (!force) {
-      const existing = this.cache.get(id);
-      if (existing && !existing.partial) return existing;
-    }
-
     const data = await this.client.rest.get(Routes.user(id));
-    return this._add(data, cache);
+    if (user instanceof User) return user._patch(data);
+    return new User(data);
   }
 
   /**
    * Fetches a user's flags.
    * @param {UserResolvable} user The UserResolvable to identify
-   * @param {BaseFetchOptions} [options] Additional options for this fetch
    * @returns {Promise<UserFlagsBitField>}
    */
-  async fetchFlags(user, options) {
-    return (await this.fetch(user, options)).flags;
+  async fetchFlags(user) {
+    return (await this.fetch(user)).flags;
   }
 
   /**
@@ -114,12 +107,15 @@ class UserManager extends CachedManager {
   /**
    * Resolves a {@link UserResolvable} to a {@link User} object.
    * @param {UserResolvable} user The UserResolvable to identify
+   * @param {GuildResolvable} [guild] The guild to search members for to find a user object
    * @returns {?User}
    */
-  resolve(user) {
+  resolve(user, guild) {
     if (user instanceof GuildMember || user instanceof ThreadMember) return user.user;
     if (user instanceof Message) return user.author;
-    return super.resolve(user);
+    if (user instanceof User) return User;
+    if (typeof user === 'string') return this.client.guilds.resolve(guild)?.members.cache.get(user)?.user ?? null;
+    return null;
   }
 
   /**
@@ -131,7 +127,22 @@ class UserManager extends CachedManager {
     if (user instanceof ThreadMember) return user.id;
     if (user instanceof GuildMember) return user.user.id;
     if (user instanceof Message) return user.author.id;
-    return super.resolveId(user);
+    if (user instanceof User) return user.id;
+    if (typeof user === 'string') return user;
+    return null;
+  }
+
+  /**
+   * Attempts to get a cached user from the appropriate guild and updates it
+   * OR creates a new user and returns that instance if there is no cached user
+   * @param {APIUser} data Raw data from discord used to update or create the user
+   * @param {BaseGuild|null} guild The guild to search for an existing cached user,
+   * accepts null for better support with uncached guilds
+   * @private
+   * @returns {User}
+   */
+  _obtain(data, guild) {
+    return guild?.members?.cache.get(data.id)?._patch(data) ?? new User(this.client, data);
   }
 }
 
--- a/packages/discord.js/src/managers/UserManager.js
+++ b/packages/discord.js/src/managers/UserManager.js
@@ -8,7 +8,7 @@ const ThreadMember = require('../structures/ThreadMember');
 const User = require('../structures/User');
 
 /**
- * Manages API methods for users and stores their cache.
+ * Manages API methods for users.
  * @extends {CachedManager}
  */
 class UserManager extends BaseManager {
@@ -82,7 +82,7 @@ class UserManager extends BaseManager {
     const id = this.resolveId(user);
     const data = await this.client.rest.get(Routes.user(id));
     if (user instanceof User) return user._patch(data);
-    return new User(data);
+    return new User(this.client, data);
   }
 
   /**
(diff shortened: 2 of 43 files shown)


packages/discord.js/src/util/Util.js
```diff
@@ -489,15 +489,15 @@ class Util extends null {
...
+          const user = channel.recipient;
           return user ? `@${user.username}` : input;
         }
 
         const member = channel.guild.members.cache.get(id);
         if (member) {
           return `@${member.displayName}`;
         } else {
-          const user = channel.client.users.cache.get(id);
+          const user = channel.guild?.members?.cache.get(id)?.user;
```
vladfrangu, line 500: This is literally the same call as above, only with extra `?.` to something we already know doesn't exist

packages/discord.js/typings/index.d.ts
```diff
@@ -775,8 +775,8 @@ export class Client<Ready extends boolean = boolean> extends BaseClient {
   public shard: ShardClientUtil | null;
   public token: If<Ready, string, string | null>;
   public get uptime(): If<Ready, number>;
-  public user: If<Ready, ClientUser>;
-  public users: UserManager;
+  public readonly user: If<Ready, ClientUser>;
```
vladfrangu, line 778:
```suggestion
  public get user(): If<Ready, ClientUser>;
```

vladfrangu (dismissed):

packages/discord.js/src/managers/UserManager.js
```diff
@@ -77,28 +76,22 @@ class UserManager extends CachedManager {
...
     const id = this.resolveId(user);
-    if (!force) {
-      const existing = this.cache.get(id);
-      if (existing && !existing.partial) return existing;
-    }
-
     const data = await this.client.rest.get(Routes.user(id));
-    return this._add(data, cache);
+    if (user instanceof User) return user._patch(data);
+    return new User(data);
```
Syjalo, line 85:
```suggestion
    return new User(this.client, data);
```

Syjalo (changes requested):

packages/discord.js/typings/index.d.ts
```diff
@@ -3379,13 +3377,19 @@ export class ThreadMemberManager extends CachedManager<Snowflake, ThreadMember,
   public remove(id: Snowflake | '@me', reason?: string): Promise<Snowflake>;
 }
 
-export class UserManager extends CachedManager<Snowflake, User, UserResolvable> {
-  private constructor(client: Client, iterable?: Iterable<RawUserData>);
+export class UserManager<Ready extends boolean> extends BaseManager {
```
kyranet, line 3380: So we can use it without having to specify:
```suggestion
export class UserManager<Ready extends boolean = boolean> extends BaseManager {
```

packages/discord.js/src/structures/Webhook.js
```diff
@@ -4,6 +4,7 @@ const { makeURLSearchParams } = require('@discordjs/rest');
 const { DiscordSnowflake } = require('@sapphire/snowflake');
 const { Routes, WebhookType } = require('discord-api-types/v10');
 const MessagePayload = require('./MessagePayload');
+const { User } = require('..');
```
kyranet, line 7:
```suggestion
const User = require('./User');
```

packages/discord.js/src/structures/IntegrationApplication.js
```diff
@@ -15,7 +16,7 @@ class IntegrationApplication extends Application {
        * The bot user for this application
        * @type {?User}
        */
-      this.bot = this.client.users._add(data.bot);
+      this.bot = this.bot ? this.bot._patch(data.bot) : new User(this.user, data.bot);
```
kyranet, line 19:
```suggestion
      this.bot = this.bot ? this.bot._patch(data.bot) : new User(this.client, data.bot);
```

packages/discord.js/src/structures/GuildTemplate.js
```diff
@@ -69,7 +70,7 @@ class GuildTemplate extends Base {
        * The user that created this template
        * @type {User}
        */
-      this.creator = this.client.users._add(data.creator);
+      this.creator = this.creator ? this.creator._patch(data.creator) : new User(this.client, data.creator);
```
kyranet, line 73: If you make `User#_patch` return something, we'll be able to clean this up:

```suggestion
      this.creator = this.creator?._patch(data.creator) ?? new User(this.client, data.creator);
```

packages/discord.js/src/structures/GuildMember.js
```diff
@@ -63,7 +64,7 @@ class GuildMember extends Base {
        * The user that this guild member instance represents
        * @type {?User}
        */
-      this.user = this.client.users._add(data.user, true);
+      this.user = this.user ? this.user._clone().patch(data.user) : new User(this.client, data.user);
```
kyranet, line 67: What's the justification for the clone here?

kyranet (changes requested):

Syjalo (changes requested):
image
Description of `Client#users` should be changed

packages/discord.js/src/structures/Webhook.js
ckohen: damn, thought I got it to do all of them correctly 😢

packages/discord.js/src/structures/GuildMember.js
ckohen: Was to ensure that cloned member instances had unique user instances, gonna handle this better in _clone now though

packages/discord.js/src/structures/GuildMember.js
```diff
@@ -63,7 +64,7 @@ class GuildMember extends Base {
        * The user that this guild member instance represents
        * @type {?User}
```
vladfrangu, line 65: When will this be nullable? Even if data.user would be undefined it will still have an (albeit invalid) user class

packages/discord.js/src/structures/Webhook.js
```diff
@@ -80,11 +81,15 @@ class Webhook {
...
-       */
-      this.owner = this.client.users?._add(data.user) ?? data.user;
+      if ('users' in this.client) {
+        this.owner = this.owner?._patch(data.user) ?? new User(this.client, data.user);
+      } else {
+        /**
+         * The owner of the webhook
+         * @type {?(User|APIUser)}
+         */
+        this.owner = this.owner ? { ...this.owner, ...data.user } : data.user;
```
vladfrangu, line 91: Couldn't we technically always create a user class? (we'd probably need to have a fake users manager on the webhook client I suppose..) Is that even worth it?

vladfrangu (dismissed):

packages/discord.js/src/structures/GuildAuditLogsEntry.js
```diff
@@ -102,8 +102,8 @@ class GuildAuditLogsEntry {
      */
     this.executor = data.user_id
       ? guild.client.options.partials.includes(Partials.User)
-        ? guild.client.users._add({ id: data.user_id })
-        : guild.client.users.cache.get(data.user_id)
+        ? guild.client.users._obtain({ id: data.user_id }, guild)
+        : logs.users.get(data.user_id)
```
ImRodry, line 106: What’s logs meant to be? I don’t see it declared anywhere

almeidx: Its the first parameter in the constructor

ImRodry: Ah I see
But then what’s the type? Is users.get a function?

packages/discord.js/src/structures/Webhook.js
ckohen: Can't be a fake manager, but also can't be a real one without a some rework, i.e. for createDM, deleteDM

packages/discord.js/src/structures/GuildMember.js
ckohen: it won't set the prop if data.user is undefined, there's a lot of places where we use ?Type to indicate undefined instead of explicitly making it null still.

packages/discord.js/src/structures/GuildAuditLogsEntry.js
Syjalo: https://github.com/ckohen/discord.js/blob/aa2bf30a975b47aeeb024b58ce48d130af3cbc9c/packages/discord.js/src/structures/GuildAuditLogs.js#L81

GravIos3:
@ckohen Can you resolve the conversations?
@vanishedvan @iCrawl @kyranet @SpaceEEC Please review

packages/discord.js/src/client/actions/UserUpdate.js
```diff
@@ -7,27 +7,21 @@ class UserUpdateAction extends Action {
   handle(data) {
     const client = this.client;
 
-    const newUser = data.id === client.user.id ? client.user : client.users.cache.get(data.id);
-    const oldUser = newUser._update(data);
+    if (data.user.id !== client.users.me.id) return { old: null, updated: null };
```
DraftProducts, line 10: Why couldn't the `userUpdate` event handle normal users ?

DraftProducts (changes requested):

packages/discord.js/src/client/actions/UserUpdate.js
ckohen: This is a safety net, the check shouldn't ever be false given discord only emits USER_UPDATE for the client user

ckohen:
After internal discussions, we are going to postpone this to the typescript rewrite after all since it will afford us a lot more flexibility in the implementation,

PR #8038, opened 2022-06-08
fix(webhooks): revert webhook caching (and returning Message)

nyapat:
webhooks inherently can't cache messages & the way the library currently works is by attempting to cache messages if an APIMessage was returned from using `send`, `editMessage` or `fetchMessage`, but incorrectly makes the assumption that the client is the actual bot client, not a WebhookClient, so caching doesn't work in the first place. (#7917) 

in the event that the bot client already has the channel cached, it will already cache a message because of the information provided by the gateway. basically it's messy

### DIFF
--- a/packages/discord.js/src/structures/InteractionWebhook.js
+++ b/packages/discord.js/src/structures/InteractionWebhook.js
@@ -29,10 +29,26 @@ class InteractionWebhook {
   /**
    * Sends a message with this webhook.
    * @param {string|MessagePayload|InteractionReplyOptions} options The content for the reply
-   * @returns {Promise<Message|APIMessage>}
+   * @returns {Promise<Message>}
    */
+
   send() {}
+
+  /**
+   * Gets a message that was sent by this webhook.
+   * @param {Snowflake|'@original'} message The id of the message to fetch
+   * @returns {Promise<Message>} Returns the message sent by this webhook
+   */
+
   fetchMessage() {}
+
+  /**
+   * Edits a message that was sent by this webhook.
+   * @param {MessageResolvable|'@original'} message The message to edit
+   * @param {string|MessagePayload|WebhookEditMessageOptions} options The options to provide
+   * @returns {Promise<Message>} Returns the message edited by this webhook
+   */
+
   editMessage() {}
   deleteMessage() {}
   get url() {}
--- a/packages/discord.js/src/structures/Webhook.js
+++ b/packages/discord.js/src/structures/Webhook.js
@@ -148,7 +148,7 @@ class Webhook {
   /**
    * Sends a message with this webhook.
    * @param {string|MessagePayload|WebhookMessageOptions} options The options to provide
-   * @returns {Promise<Message>}
+   * @returns {Promise<APIMessage>}
    * @example
    * // Send a basic message
    * webhook.send('hello!')
@@ -211,7 +211,9 @@ class Webhook {
 
     const { body, files } = await messagePayload.resolveFiles();
     const d = await this.client.rest.post(Routes.webhook(this.id, this.token), { body, files, query, auth: false });
-    return this.client.channels?.cache.get(d.channel_id)?.messages._add(d, false) ?? new (getMessage())(this.client, d);
+
+    if (!this.client.channels) return d;
+    return this.client.channels.cache.get(d.channel_id)?.messages._add(d, false) ?? new (getMessage())(this.client, d);
   }
 
   /**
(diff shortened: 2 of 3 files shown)


Syjalo:
Can't we just add optional chaining where it needs to be instead of reverting the pr?

packages/discord.js/src/structures/Webhook.js
```diff
@@ -211,7 +208,7 @@ class Webhook {
 
     const { body, files } = await messagePayload.resolveFiles();
     const d = await this.client.rest.post(Routes.webhook(this.id, this.token), { body, files, query, auth: false });
-    return this.client.channels?.cache.get(d.channel_id)?.messages._add(d, false) ?? new (getMessage())(this.client, d);
```
muchnameless, line 214: instead of always returning `d` could you try treating both cases of optional chaining differently? If the client doesn't have a channels property it's a WebhookClient -> return d, if it has a ChannelManager but the channel isn't cached construct a Message. That way interactions wouldn't have APIMessages (what the initial pr intended) while keeping WebhookClients as before

kyranet (approved):

packages/discord.js/typings/index.d.ts
```diff
@@ -3456,16 +3456,16 @@ export interface PartialWebhookFields {
...
+  fetchMessage(message: Snowflake | '@original', options?: WebhookFetchMessageOptions): Promise<APIMessage>;
+  send(options: string | MessagePayload | Omit<WebhookMessageOptions, 'flags'>): Promise<APIMessage>;
 }
 
 export interface WebhookFields extends PartialWebhookFields {
   get createdAt(): Date;
   get createdTimestamp(): number;
   delete(reason?: string): Promise<void>;
-  edit(options: WebhookEditData, reason?: string): Promise<Webhook>;
+  edit(options: WebhookEditData, reason?: string): Promise<APIMessage>;
```
vladfrangu, line 3468: This is not correct, Webhook#edit edits the webhook itself

vladfrangu (changes requested):

packages/discord.js/src/structures/InteractionWebhook.js
```diff
@@ -29,7 +29,21 @@ class InteractionWebhook {
...
+   * @returns {Promise<Message>} Returns the message sent by this webhook
+   */
+
+  /**
+   * Edits a message that was sent by this webhook.
+   * @param {MessageResolvable|'@original'} message The message to edit
+   * @param {string|MessagePayload|WebhookEditMessageOptions} options The options to provide
+   * @returns {Promise<Message>} Returns the message edited by this webhook
    */
   send() {}
```
almeidx, line 48: Shouldn't this comment be above the `editMessage()` method? And the same goes for the other comments above this one

almeidx: > to override the jsdoc and keep consistency

I understand that, but wouldn't the jsdocs have to be above each method?

packages/discord.js/src/structures/Webhook.js
```diff
@@ -148,7 +148,7 @@ class Webhook {
   /**
    * Sends a message with this webhook.
    * @param {string|MessagePayload|WebhookMessageOptions} options The options to provide
-   * @returns {Promise<Message>}
```
vladfrangu, line 151: But it can return both

packages/discord.js/src/structures/Webhook.js
```diff
@@ -286,17 +288,19 @@ class Webhook {
    * Gets a message that was sent by this webhook.
    * @param {Snowflake|'@original'} message The id of the message to fetch
    * @param {WebhookFetchMessageOptions} [options={}] The options to provide to fetch the message.
-   * @returns {Promise<Message>} Returns the message sent by this webhook
+   * @returns {Promise<APIMessage>} Returns the message sent by this webhook
```
vladfrangu, line 291: Ditto

packages/discord.js/src/structures/Webhook.js
```diff
@@ -305,7 +309,7 @@ class Webhook {
    * Edits a message that was sent by this webhook.
    * @param {MessageResolvable|'@original'} message The message to edit
    * @param {string|MessagePayload|WebhookEditMessageOptions} options The options to provide
-   * @returns {Promise<Message>} Returns the message edited by this webhook
+   * @returns {Promise<APIMessage>} Returns the message edited by this webhook
```
vladfrangu, line 312: Ditto

packages/discord.js/typings/index.d.ts
```diff
@@ -1562,6 +1562,11 @@ export class InteractionWebhook extends PartialWebhookMixin() {
   public constructor(client: Client, id: Snowflake, token: string);
   public token: string;
   public send(options: string | MessagePayload | InteractionReplyOptions): Promise<Message>;
+  public editMessage(
+    message: MessageResolvable,
+    options: string | MessagePayload | WebhookEditMessageOptions,
+  ): Promise<Message>;
```
vladfrangu, line 1568: This is technically not gonna be true if you use no intents in your bot (which is possible)...

vladfrangu (changes requested):

packages/discord.js/typings/index.d.ts
vladfrangu: But it won't have the channel in cache 👀

ChrisSch-dev:
Any ETA on when this will be merged?

Raiden-Ei:
> Any ETA on when this will be merged?

When It's ready™️

vladfrangu (approved):

SpaceEEC (approved):

PR #8050, opened 2022-06-09
docs(WebSocketOptions): add `version` to docs and typings

advaith1:
You can already set the gateway version in `ws`, but it's the only property missing from the typings, and it also isn't documented. Setting the rest version is documented for both [v13](https://discord.js.org/#/docs/discord.js/stable/typedef/HTTPOptions) and [djs/rest](https://discord.js.org/#/docs/rest/main/typedef/RESTOptions).


packages/discord.js/src/util/Options.js
```diff
@@ -62,6 +62,7 @@ const Transformers = require('./Transformers');
  * @typedef {Object} WebsocketOptions
  * @property {number} [large_threshold=50] Number of members in a guild after which offline users will no longer be
  * sent in the initial guild member list, must be between 50 and 250
+ * @property {number} [version=10] The Discord gateway version to use
```
kyranet, line 65: Setting `options.ws.version` is an advanced option and may break the library, since we generally need to refactor several things in order to bump the version. I believe this is the reason why it wasn't documented in the first place.

This PR is fine as-is, but we should add a warning here to let users know.

kyranet (approved):

vladfrangu (approved):

SpaceEEC (approved):

PR #8052, opened 2022-06-09
refactor(util): make utility functions top level

suneettipirneni:

Makes static utility functions within a class top level. This PR also exports channel-based typeguards as top level functions rather than instance methods.

### DIFF
--- a/packages/discord.js/src/index.js
+++ b/packages/discord.js/src/index.js
@@ -15,6 +15,7 @@ exports.ActivityFlagsBitField = require('./util/ActivityFlagsBitField');
 exports.ApplicationFlagsBitField = require('./util/ApplicationFlagsBitField');
 exports.BaseManager = require('./managers/BaseManager');
 exports.BitField = require('./util/BitField');
+__exportStar(require('./util/Channels.js'), exports);
 exports.Collection = require('@discordjs/collection').Collection;
 exports.Constants = require('./util/Constants');
 exports.Colors = require('./util/Colors');
@@ -34,7 +35,7 @@ exports.Sweepers = require('./util/Sweepers');
 exports.SystemChannelFlagsBitField = require('./util/SystemChannelFlagsBitField');
 exports.ThreadMemberFlagsBitField = require('./util/ThreadMemberFlagsBitField');
 exports.UserFlagsBitField = require('./util/UserFlagsBitField');
-exports.Util = require('./util/Util');
+__exportStar(require('./util/Util.js'), exports);
 exports.version = require('../package.json').version;
 
 // Managers
--- a/packages/discord.js/src/index.js
+++ b/packages/discord.js/src/index.js
@@ -15,7 +15,6 @@ exports.ActivityFlagsBitField = require('./util/ActivityFlagsBitField');
 exports.ApplicationFlagsBitField = require('./util/ApplicationFlagsBitField');
 exports.BaseManager = require('./managers/BaseManager');
 exports.BitField = require('./util/BitField');
-__exportStar(require('./util/Channels.js'), exports);
 exports.Collection = require('@discordjs/collection').Collection;
 exports.Constants = require('./util/Constants');
 exports.Colors = require('./util/Colors');
--- a/packages/discord.js/src/util/Channels.js
+++ b/packages/discord.js/src/util/Channels.js
@@ -1,0 +1,121 @@
+'use strict';
+
+const { ChannelType } = require('discord-api-types/v10');
+const { ThreadChannelTypes } = require('../util/Constants');
+
+let CategoryChannel;
+let DMChannel;
+let NewsChannel;
+let StageChannel;
+let TextChannel;
+let ThreadChannel;
+let VoiceChannel;
+let DirectoryChannel;
+
+/**
+ * Indicates whether this channel is a {@link ThreadChannel}
+ * @param {Channel} channel The channel to check against
+ * @returns {boolean}
+ */
+function isThread(channel) {
+  return ThreadChannelTypes.includes(channel.type);
+}
+
+/**
+ * Indicates whether this channel is {@link TextBasedChannels text-based}.
+ * @param {Channel} channel The channel to check against
+ * @returns {boolean}
+ */
+function isTextBased(channel) {
+  return 'messages' in channel;
+}
+
+/**
+ * Indicates whether this channel is DM-based (either a {@link DMChannel} or a {@link PartialGroupDMChannel}).
+ * @param {Channel} channel The channel to check against
+ * @returns {boolean}
+ */
+function isDMBased(channel) {
+  return [ChannelType.DM, ChannelType.GroupDM].includes(channel.type);
+}
+
+/**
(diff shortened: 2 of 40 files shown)


imranbarbhuiya:
finally top level functions in djs 🚀

DraftProducts:
Is that a good move ?
Having utilities top level functions make sense in an app or a project, in my opinion this kind of breaking change is not something that we should want for a library like djs. This only complicates the DX of the library.
Checking types methods is something that we need everywhere.
I can't imagine importing that stuff everywhere when it can be ready for use in classes 😕

DraftProducts:
I understand the move for these types of utilities, but why were these methods moved:
- `<Channel>.isThread()`
- `<Channel>isTextBased()`
- `<Channel>isDMBased()`
- `<Channel>isVoiceBased()`

nowifi4u:
> I understand the move for these types of utilities, but why were these methods moved:
> 
> * `<Channel>.isThread()`
> * `<Channel>isTextBased()`
> * `<Channel>isDMBased()`
> * `<Channel>isVoiceBased()`

Exactly, that will remove the possibility of separately typing them i.e. having `.isText()` return different types for `Channel` and `GuildChannel`

suneettipirneni:
> Exactly, that will remove the possibility of separately typing them i.e. having `.isText()` return different types for `Channel` and `GuildChannel`

The comment you were replying was talking about a different issue. You can still narrow types with the top level functions or the instance methods. The typings don't change here either way.

packages/discord.js/typings/index.d.ts
```diff
@@ -2632,19 +2626,28 @@ export interface MappedComponentTypes {
...
-  public static createComponentBuilder(data: APIMessageComponent | ComponentBuilder): ComponentBuilder;
+export interface CreateChannelOptions {
+  allowFromUnknownGuild?: boolean;
+  fromInteraction?: boolean;
 }
 
+export function isThread(channel: unknown): channel is ThreadChannel;
+export function isTextBased(channel: unknown): channel is TextBasedChannel;
+export function isDMBased(channel: unknown): channel is DMChannel | PartialDMChannel;
+export function isVoiceBased(channel: unknown): channel is VoiceBasedChannel;
```
Syjalo, line 2637: Why the `channel` type is `unknown` if it supposed to take any channel (`AnyChannel` type)?

packages/discord.js/src/index.js
```diff
@@ -15,6 +15,7 @@ exports.ActivityFlagsBitField = require('./util/ActivityFlagsBitField');
 exports.ApplicationFlagsBitField = require('./util/ApplicationFlagsBitField');
 exports.BaseManager = require('./managers/BaseManager');
 exports.BitField = require('./util/BitField');
+__exportStar(require('./util/Channels.js'), exports);
```
vladfrangu, line 18: We should not export createChannel to end users I'd say... Or should we?

packages/discord.js/src/util/Channels.js
```diff
@@ -0,0 +1,121 @@
...
+ */
+function createChannel(client, data, guild, { allowUnknownGuild, fromInteraction } = {}) {
+  CategoryChannel ??= require('./CategoryChannel');
+  DMChannel ??= require('./DMChannel');
+  NewsChannel ??= require('./NewsChannel');
+  StageChannel ??= require('./StageChannel');
+  TextChannel ??= require('./TextChannel');
+  ThreadChannel ??= require('./ThreadChannel');
+  VoiceChannel ??= require('./VoiceChannel');
+  DirectoryChannel ??= require('./DirectoryChannel');
```
kyranet, line 63: This should use the `lazy` utility, as well as use on demand, instead of always doing those operations on every run.

packages/discord.js/src/index.js
Jiralite: I say we shouldn't. It's for internal use only and the constructors are not documented for the individual channels.

packages/discord.js/typings/index.d.ts
```diff
@@ -2632,19 +2626,28 @@ export interface MappedComponentTypes {
...
+export interface CreateChannelOptions {
+  allowFromUnknownGuild?: boolean;
+  fromInteraction?: boolean;
 }
 
+export function isThread(channel: AnyChannel | Channel): channel is ThreadChannel;
+export function isTextBased(channel: AnyChannel | Channel): channel is TextBasedChannel;
+export function isDMBased(channel: AnyChannel | Channel): channel is DMChannel | PartialDMChannel;
+export function isVoiceBased(channel: AnyChannel | Channel): channel is VoiceBasedChannel;
+export function createChannel(client: Client, data: APIChannel, options?: CreateChannelOptions): Channel;
```
kyranet, line 2638: This shouldn't return `Channel`, but `AnyChannel`.

packages/discord.js/typings/index.d.ts
```diff
@@ -2632,19 +2626,28 @@ export interface MappedComponentTypes {
...
-  public static createComponentBuilder(data: APIMessageComponent | ComponentBuilder): ComponentBuilder;
+export interface CreateChannelOptions {
+  allowFromUnknownGuild?: boolean;
+  fromInteraction?: boolean;
 }
 
+export function isThread(channel: AnyChannel | Channel): channel is ThreadChannel;
+export function isTextBased(channel: AnyChannel | Channel): channel is TextBasedChannel;
+export function isDMBased(channel: AnyChannel | Channel): channel is DMChannel | PartialDMChannel;
+export function isVoiceBased(channel: AnyChannel | Channel): channel is VoiceBasedChannel;
```
kyranet, line 2637: I don't think we should have this union, and in fact, it should be just `AnyChannel`.

kyranet (changes requested):

packages/discord.js/typings/index.d.ts
suneettipirneni: I've tried this, but the TS complains that the type predicated isn't assignable to the parameter.

suneettipirneni:
> I understand the move for these types of utilities, but why were these methods moved:
> 
> * `<Channel>.isThread()`
> * `<Channel>isTextBased()`
> * `<Channel>isDMBased()`
> * `<Channel>isVoiceBased()`

This changes has been reverted in the latest commit.

kyranet (approved):

iCrawl:
Needs a rebase 👍

muchnameless:
two things I noticed
- you can completely remove the `Formatters` util class, the builder functions are already re-exported by __exportStar in index.js
- functions like createChannel should use the new ignore tag instead of private to not show up on the docs

suneettipirneni:
> two things I noticed
> 
> - you can completely remove the `Formatters` util class, the builder functions are already re-exported by __exportStar in index.js
> 
> - functions like createChannel should use the new ignore tag instead of private to not show up on the docs

The only reason I kept the formatter class was because it wouldn't be documented without creating the class again.

DraftProducts (approved):

vladfrangu (approved):

SpaceEEC (approved):

PR #8021, opened 2022-06-06
refactor(ThreadMemberManager): Consistent thread member fetching

Jiralite:
**Please describe the changes this PR makes and why it should be merged:**
`ThreadMemberManager#fetch()` has been made consistent with other managers, such as `MessageManager#fetch()` and `GuildBanManager#fetch()`. This also fixes an issue (by removing it) in which passing only a boolean was ignored and was deemed always `true`.

**Status and versioning classification:**
- Code changes have been tested against the Discord API, or there are no code changes
- I know how to update typings and have done so, or typings don't need updating
- This PR includes breaking changes (methods removed or renamed, parameters moved or removed)

PR #8060, opened 2022-06-10
types(ThreadMemberManager): Fix return type of fetching members with no arguments

Jiralite:
A regression from #8021: performing `<ThreadMemberManager>.fetch()` would yield a return type of a singular thread member. This should be a collection of thread members.

### DIFF
--- a/packages/discord.js/typings/index.test-d.ts
+++ b/packages/discord.js/typings/index.test-d.ts
@@ -126,6 +126,7 @@ import {
   MessageContextMenuCommandInteraction,
   UserContextMenuCommandInteraction,
   AnyThreadChannel,
+  ThreadMemberManager,
 } from '.';
 import { expectAssignable, expectNotAssignable, expectNotType, expectType } from 'tsd';
 import { UnsafeButtonBuilder, UnsafeEmbedBuilder, UnsafeSelectMenuBuilder } from '@discordjs/builders';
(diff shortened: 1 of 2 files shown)


DraftProducts (approved):

vladfrangu (approved):

packages/discord.js/typings/index.test-d.ts
```diff
@@ -566,9 +566,11 @@ client.on('guildCreate', async g => {
     expectType<ThreadMember>(fetchedMember);
     expectType<ThreadMember>(fetchedMember2);
     const fetchedMemberCol = await channel.members.fetch({ cache: true });
+    const fetchedMemberCol2 = await channel.members.fetch();
     expectType<Collection<Snowflake, ThreadMember>>(fetchedMemberCol);
+    expectType<Collection<Snowflake, ThreadMember>>(fetchedMemberCol2);
     // @ts-expect-error The `force` option cannot be used alongside fetching all thread members.
-    const fetchedMemberCol2 = await channel.members.fetch({ cache: true, force: false });
+    const fetchedMemberCol3 = await channel.members.fetch({ cache: true, force: false });
```
kyranet, line 573: We should either use meaningful names, or use blocks to declare variables without name conflicts. An example of the latter would be:

```ts
{
	const result = await channel.members.fetch();
	expectType<Collection<Snowflake, ThreadMember>>(result);
}
{
	const result = await channel.members.fetch({ cache: true });
	expectType<Collection<Snowflake, ThreadMember>>(result);
}
```

Also makes the separation between type tests clearer.

kyranet (changes requested):
Just a nit.

packages/discord.js/typings/index.test-d.ts
Jiralite: To be honest, I want to get rid of this entire `if` block. The `ThreadMemberManager` is only being tested here, but the other managers are on lines 940 to 1113. What would you say about just moving it there alongside other managers?

iCrawl (approved):

kyranet (approved):

SpaceEEC (approved):

PR #8061, opened 2022-06-10
types(GuildChannel): fix type of `.isText()` method

Syjalo:
This pr fixes type of `GuildChannel#isText()` method


nowifi4u:
Why even fix it when there is already correct typing for the method in class's parent?

Syjalo:
> @Syjalo Why even fix it when there is already correct typing for the method in class's parent? Simply removing the override will fix it

DM channels aren't a GuildChannel. That's why.

vladfrangu (approved):

SpaceEEC (approved):

PR #8063, opened 2022-06-10
typings: remove isAutocomplete typeguard from typings

imranbarbhuiya:
`isAutocomplete` was removed but it's still present in typings. So this pr removes it from typings. Also updated readme and example to use `isChatInputCommand` instead of `isCommand`.

### DIFF
--- a/packages/voice/examples/recorder/src/bot.ts
+++ b/packages/voice/examples/recorder/src/bot.ts
@@ -29,7 +29,7 @@ client.on(Events.MessageCreate, async (message) => {
 const recordable = new Set<string>();
 
 client.on(Events.InteractionCreate, async (interaction: Interaction) => {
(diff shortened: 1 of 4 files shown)


vladfrangu (approved):

iCrawl (approved):

packages/voice/examples/recorder/src/bot.ts
```diff
@@ -29,7 +29,7 @@ client.on(Events.MessageCreate, async (message) => {
 const recordable = new Set<string>();
 
 client.on(Events.InteractionCreate, async (interaction: Interaction) => {
-	if (!interaction.isCommand() || !interaction.guildId) return;
+	if (!interaction.isChatInputCommand() || !interaction.inGuild()) return;
```
almeidx, line 32: This is not correct, this example is still using discord.js v13.7.0

SpaceEEC (approved):

PR #8066, opened 2022-06-11
fix(GuildMemberManager): Allow setting own nickname

### LINKED ISSUE #7425: Interface with Modify Current User endpoint

discord.js

### Feature

I attempted to have my bot change its own username today in a server I'm in to enforce a nickname policy (I didn't want to bug the owner) and found to my utter surprise that I did not have the permissions to do so, despite having Manage Nicknames, Change Nickname, Administrator, and the highest role on the server. I did a little bit of research and found out about the [Manage Current Member](https://discord.com/developers/docs/resources/guild#modify-current-member) endpoint, and realized DJS apparently had no implementation of this. I'd like to see this implemented with DJS.

Oddly enough, others did not appear to have the same issue, but I've double checked everything, and I could not implement a fix on my own.

I can confirm that I was getting an HTTP 403 with a 50013 `DiscordAPIError: Missing Permissions` every time I attempted to do this, and no other syntatical errors. I've ensured both the guild is correct and that the methods I'm using are accurate, hence why this is not listed as a bugfix.

A new method on the Guild class, setOwnNickname. Accepts a valid nickname as an argument, then uses the Modify Current Member endpoint to update the bot's username.

Checking equivalence with the ClientUser when using GuildMember.setNickname. If they are the same, use the Modify Current Member endpoint. As it is, attempting to modify the bot's own nickname with this permission will always fail, as it has equivalent permissions/roles to itself.

### Other context

I used `<GuildManager>.fetch` to obtain the guild itself (but also tried using `<GuildManager>.cache.get`), and then `<Guild>.me.setNickname` to change the username. This seems like a very simple way to do it, so I have no clue what I can do from here. Others in the DJS server were also stumped.

Jiralite:
~~There was possible confusion about why a developer might not be able to edit their bot's nickname in a server. This is because it is not obvious (at all) that one should _only_ provide `nick` for the suitable endpoint to be used. Separating this endpoint out will help with future endpoint support, if any will be added, as well as make it clear on what the method is used for.~~

This will simply fix being able to set the client's own nickname in a guild.


packages/discord.js/src/managers/GuildMemberManager.js
```diff
@@ -254,6 +254,25 @@ class GuildMemberManager extends CachedManager {
...
+   * @property {?string} [nick] The nickname to set for the client user in a guild
+   * @property {string} [reason] The reason for editing the client user's data in a guild
+   */
+
+  /**
+   * Edits the client user in a guild.
+   * @param {GuildMemberEditMeData} data The data to edit the guild member belonging to the client user with
+   * @returns {Promise<GuildMember>}
+   */
+  async editMe({ nick, reason }) {
```
kyranet, line 269: If all properties are optional, shouldn't the object itself also be optional?

Jiralite: I could change it to `({ nick, reason }) = {}`, but the `edit()` method below is identical to this one. It also helps people to actually pass something to this method, as calling `editMe()` with no arguments would likely indicate a developer is doing something wrong (the method will throw additionally). `editMe({})` works, however, and is a no-op. That's why `data` is a required parameter here yet the properties in the object are optional.

vladfrangu (approved):

SpaceEEC:
Would it be reasonable to throw an error if one attempts to edit their own nick through `GuildMember#edit`, as that will yield you a confusing error still?

kyranet:
I don't think so, @SpaceEEC. One of the things worth considering is that Discord may change the behaviour from setting a bot's own nick through this endpoint and make it either throw a more helpful error, or just accept it.

packages/discord.js/src/structures/GuildMember.js
```diff
@@ -307,12 +307,12 @@ class GuildMember extends Base {
...
   /**
    * Sets the nickname for this member.
-   * @param {?string} nick The nickname for the guild member, or `null` if you want to reset their nickname
+   * @param {?string} nick The nickname for the guild member. Pass `null` to reset the nickname
    * @param {string} [reason] Reason for setting the nickname
    * @returns {Promise<GuildMember>}
    */
   setNickname(nick, reason) {
-    return this.edit({ nick, reason });
+    return this.id === this.client.user.id ? this.guild.members.editMe({ nick, reason }) : this.edit({ nick, reason });
```
ckohen, line 315: Is this the only thing specific to the clients guildMember? Just wondering if we may want a helper for this or an extended class like ClientUser (doubt)

ckohen (approved):
🧹 

One question though

packages/discord.js/src/structures/GuildMember.js
ckohen: Yeah, if this is the only thing neither was worth it. `ClientGuildMember` would only be warranted with 4+ unique traits IMO

GravIos3:
@Jiralite Can you resolve the conversation?
@iCrawl @SpaceEEC Please review

DraftProducts (approved):

kyranet (approved):

packages/discord.js/src/managers/GuildMemberManager.js
```diff
@@ -261,6 +261,25 @@ class GuildMemberManager extends CachedManager {
...
+    const data = await this.client.rest.patch(Routes.guildMember(this.guild.id), { body: { nick }, reason });
+    const clone = this.cache.get(this.client.user.id)?._clone();
```
almeidx, line 278: Can't you just use this?
```suggestion
    const clone = this.me?._clone();
```

SpaceEEC (approved):

iCrawl (approved):

PR #8067, opened 2022-06-11
fix(ApplicationCommand): fix default member permissions assignment

CodeGoat-dev:

Fixes default member permissions assignment when creating or editing application commands.

### DIFF
--- a/packages/discord.js/src/managers/ApplicationCommandManager.js
+++ b/packages/discord.js/src/managers/ApplicationCommandManager.js
@@ -246,7 +246,7 @@ class ApplicationCommandManager extends CachedManager {
       description_localizations: command.descriptionLocalizations ?? command.description_localizations,
       type: command.type,
       options: command.options?.map(o => ApplicationCommand.transformOption(o)),
-      default_member_permissions: default_member_permissions,
+      default_member_permissions: command.defaultMemberPermissions ? command.default_member_permissions,
       dm_permission: command.dmPermission ?? command.dm_permission,
     };
   }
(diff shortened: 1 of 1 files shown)


imranbarbhuiya:
This isn't incorrect. If the key and value name is same, then u can write in shorten form. I.e.:

```ts
const a = 1;
const b = { a };
// is same as 
const b = { a: a}
```

CodeGoat-dev:
> This isn't incorrect. If the key and value name is same, then u can write in shorten form. I.e.:
> 
> ```ts
> const a = 1;
> const b = { a };
> // is same as 
> const b = { a: a}
> ```

I accidentally submitted the wrong changes and only realised after your comment.
I guess I should have double checked what I was doing since I'm fully blind, but it's first thing in the morning. 😁

ckohen:
What is this fixing?

There's code above the line you edited to take into account the differences in types between the two potential values.

CodeGoat-dev:
> What is this fixing?
> 
> There's code above the line you edited to take into account the differences in types between the two potential values.

Using the latest dev release, application command creation and edit are throwing bigint serialisation errors for me regardless of whether my default member permissions are stringified or not.
This fixes my issue.

ckohen:
Ah, I see. Curious how I missed this in testing.
Also curious that you can test this at all considering latest dev crashes on running.

Anyways, the proper fix here is to add .toString() tp the aforementioned previous code points, lines 232 and 238

ckohen (approved):

CodeGoat-dev:
I resolved the crash in the latest dev release by removing the requirements for the missing integration scripts in client/websocket/handlers.

Anyway the recommended fix has been applied in the latest commit.

kyranet (approved):

kyranet:
That's fixed in #8065, @CodeGoat-dev.

SpaceEEC (approved):

iCrawl (approved):

PR #8069, opened 2022-06-11
types(AutocompleteOption): fix and improve types

ImRodry:

The type of AutocompleteOption#value was wrong so I went ahead and fixed that and also made the other properties more accurate (made focused not optional and type more specific).

### DIFF
--- a/packages/discord.js/typings/index.d.ts
+++ b/packages/discord.js/typings/index.d.ts
@@ -1006,8 +1006,8 @@ export class CommandInteractionOptionResolver<Cached extends CacheType = CacheTy
   ): NonNullable<CommandInteractionOption<Cached>['member' | 'role' | 'user']> | null;
   public getMessage(name: string, required: true): NonNullable<CommandInteractionOption<Cached>['message']>;
   public getMessage(name: string, required?: boolean): NonNullable<CommandInteractionOption<Cached>['message']> | null;
-  public getFocused(getFull: true): AutocompleteOption;
-  public getFocused(getFull?: boolean): string | number;
+  public getFocused(getFull: true): AutocompleteFocusedOption;
+  public getFocused(getFull?: boolean): string;
 }
 
 export class ContextMenuCommandInteraction<Cached extends CacheType = CacheType> extends CommandInteraction<Cached> {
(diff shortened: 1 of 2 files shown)


packages/discord.js/typings/index.d.ts
```diff
@@ -4063,7 +4063,14 @@ export interface CommandInteractionResolvedData<Cached extends CacheType = Cache
...
 
-export type AutocompleteOption = Pick<CommandInteractionOption, 'name' | 'type' | 'value' | 'focused'>;
+export type AutocompleteOption = Pick<CommandInteractionOption, 'name'> & {
+  type:
+    | ApplicationCommandOptionType.String
+    | ApplicationCommandOptionType.Integer
+    | ApplicationCommandOptionType.Number;
+  focused: boolean;
+  value: string | number;
+};
```
almeidx, line 4073:
```suggestion
export type AutocompleteOption = Required<Pick<CommandInteractionOption, 'name' | 'focused'>> & {
  type:
    | ApplicationCommandOptionType.String
    | ApplicationCommandOptionType.Integer
    | ApplicationCommandOptionType.Number;
  value: string | number;
};
```

ImRodry: sure ig?

Jiralite: `focused` is always a `true` here actually, so I don't think getting the optional property here would be accurate?

almeidx: Right, this is only used in `getFocused()`. The type could have a better name then

ImRodry: oh yeah I thought it would be false for autocomplete options that aren't focused. In that case I'll revert this and make it true

ImRodry:
In a weird twist of events it turns out value is always a string when an option is focused and, since this interface is only used for focused autocomplete options I went ahead and changed that type and its naem

Jiralite (approved):

Jiralite (changes requested):
`AutocompleteOption` JSDoc needs updating too!

vladfrangu (approved):

SpaceEEC (approved):

kyranet (approved):

PR #8075, opened 2022-06-12
chore: cancel previous run in workflow

imranbarbhuiya:
Cancel previously running tests when a new commit gets added.

### DIFF
--- a/.github/workflows/documentation.yml
+++ b/.github/workflows/documentation.yml
@@ -22,6 +22,11 @@ jobs:
       BRANCH_OR_TAG: ${{ steps.env.outputs.BRANCH_OR_TAG }}
       SHA: ${{ steps.env.outputs.SHA }}
     steps:
+      - name: Cancel Previous Runs
+        uses: styfle/cancel-workflow-action@0.9.1
+        with:
+          access_token: ${{ github.token }}
+
       - name: Checkout repository
         uses: actions/checkout@v2
         with:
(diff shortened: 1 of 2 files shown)


ImRodry:
You can just use this https://docs.github.com/en/actions/using-jobs/using-concurrency

.github/workflows/test.yml
```diff
@@ -5,6 +5,11 @@ jobs:
     name: Lint
     runs-on: ubuntu-latest
     steps:
+      - name: Cancel Previous Runs
+        uses: styfle/cancel-workflow-action@0.9.1
```
ArjunSharda, line 9:
```suggestion
        uses: styfle/cancel-workflow-action@0.10.0
```

.github/workflows/documentation.yml
```diff
@@ -22,6 +22,11 @@ jobs:
       BRANCH_OR_TAG: ${{ steps.env.outputs.BRANCH_OR_TAG }}
       SHA: ${{ steps.env.outputs.SHA }}
     steps:
+      - name: Cancel Previous Runs
+        uses: styfle/cancel-workflow-action@0.9.1
```
ArjunSharda, line 26:
```suggestion
        uses: styfle/cancel-workflow-action@0.10
```

ArjunSharda (changes requested):
I've added two suggestions, as the current version you are proposing for is outdated.

iCrawl:
I'll close this for now, this might conflict tags and branch docs. And I think we don't have a huge issue with too many CIs running.

PR #7909, opened 2022-05-10
chore: reexport all dapi-types

suneettipirneni:
**Please describe the changes this PR makes and why it should be merged:**

Simplifies using -types within discord.js.

**Status and versioning classification:**
- Code changes have been tested against the Discord API, or there are no code changes
- I know how to update typings and have done so, or typings don't need updating

PR #8085, opened 2022-06-13
chore: remove -types import in readme example

BaumianerNiklas:
As of #7909 discord.js re-exports everything from discord-api-types, so there's no need in installing it as a redundant dependency. 

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,16 +45,16 @@ pnpm add discord.js
 Install all required dependencies:
 
 ```sh-session
-npm install discord.js @discordjs/rest discord-api-types
-yarn add discord.js @discordjs/rest discord-api-types
-pnpm add discord.js @discordjs/rest discord-api-types
+npm install discord.js @discordjs/rest
+yarn add discord.js @discordjs/rest
+pnpm add discord.js @discordjs/rest
 ```
 
 Register a slash command against the Discord API:
 
 ```js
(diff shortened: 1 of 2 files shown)


vvito7 (commented):
There is also this one:
https://github.com/discordjs/discord.js/blob/96053babe1bd65ebe1fc6a261f5eb052906afdb9/packages/discord.js/README.md?plain=1#L47-L51

ImRodry:
/rest comes installed with djs so you can also remove that

Syjalo:
> /rest comes installed with djs so you can also remove that

Only the error. Not the `REST` class.
image

ImRodry:
It still comes installed so you don’t have to install it

Syjalo:
> It still comes installed so you don’t have to install it

If you use a package in your code you should install it as a dependency.

ImRodry:
Debatable because if you install a more recent version than the one djs has installed it might cause conflicts

didinele:
👎 Using a package you haven't explicitly installed is one of the worst practices, please don't tell people to do this. The ideal solution would be to change the command to like, `discord-api-types@^0.33.5` (assuming current d.js is on 0.35) and keep it up to date.

SpaceEEC (approved):

iCrawl (approved):

PR #8095, opened 2022-06-14
refactor: deprecate `$` prefix from `ws.properties` keys

SuperchupuDev:
- https://github.com/discordjs/discord.js/pull/8094#issuecomment-1155549091
- https://github.com/discord/discord-api-docs/pull/5067

### DIFF
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -6016,9 +6016,9 @@ export interface WebSocketOptions {
 }
 
 export interface WebSocketProperties {
-  $os?: string;
-  $browser?: string;
-  $device?: string;
+  os?: string;
+  browser?: string;
+  device?: string;
 }
 
 export interface WidgetActivity {
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -6016,6 +6016,9 @@ export interface WebSocketOptions {
 }
 
 export interface WebSocketProperties {
+  $os?: string;
+  $browser?: string;
+  $device?: string;
   os?: string;
   browser?: string;
   device?: string;
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -6019,11 +6019,11 @@ export interface WebSocketProperties {
   os?: string;
   browser?: string;
   device?: string;
-  /** @deprecated Use `os` instead */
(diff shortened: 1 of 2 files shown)


almeidx:
As is, this is a breaking change

iCrawl:
We should do this in a non-breaking way, yes. So accept both but only send the "new" way.

SuperchupuDev:
alright then

almeidx:
🤔 Can you deprecate the prefixed keys?

typings/index.d.ts
```diff
@@ -6016,8 +6016,14 @@ export interface WebSocketOptions {
 }
 
 export interface WebSocketProperties {
+  os?: string;
+  browser?: string;
+  device?: string;
+  /** @deprecated Use `os` instead */
```
Jiralite, line 6022:
```suggestion
  /** @deprecated Use {@link os} instead. */
```

typings/index.d.ts
```diff
@@ -6016,8 +6016,14 @@ export interface WebSocketOptions {
...
   $os?: string;
+  /** @deprecated Use `browser` instead */
```
Jiralite, line 6024:
```suggestion
  /** @deprecated Use {@link browser} instead. */
```

typings/index.d.ts
```diff
@@ -6016,8 +6016,14 @@ export interface WebSocketOptions {
...
   $browser?: string;
+  /** @deprecated Use `device` instead */
```
Jiralite, line 6026:
```suggestion
  /** @deprecated Use {@link device} instead. */
```

Jiralite (commented):
These `@link`s will point to the correct definition in the typings.

vladfrangu (approved):

iCrawl (approved):

kyranet (approved):

JosPoka:
@SuperchupuDev Could you check the failing check? (vercel)

kyranet:
@JosPoka Vercel only passes on PRs pointing to main, it'll always fail for PRs to v13, don't mind it.

vladfrangu (approved):

SpaceEEC (approved):

PR #8109, opened 2022-06-17
types(voice): bring back typed events

skick1234:
d4b41dd0815b493b599d4f4d1b6dd18cd99f91ea commits removed all typed events for compatible with `docgen` I guess. But `AudioPlayer` event types are broken. That's why I bring back all removed typed events.

### DIFF
--- a/packages/voice/src/VoiceConnection.ts
+++ b/packages/voice/src/VoiceConnection.ts
@@ -1,3 +1,4 @@
+/* eslint-disable @typescript-eslint/method-signature-style */
 import { EventEmitter } from 'node:events';
 import type { GatewayVoiceServerUpdateDispatchData, GatewayVoiceStateUpdateDispatchData } from 'discord-api-types/v10';
 import type { CreateVoiceConnectionOptions } from '.';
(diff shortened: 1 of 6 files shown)


packages/voice/src/VoiceConnection.ts
```diff
@@ -161,6 +162,32 @@ export type VoiceConnectionState =
 	| VoiceConnectionReadyState
 	| VoiceConnectionDestroyedState;
 
+export interface VoiceConnection extends EventEmitter {
```
vladfrangu, line 165: Why not use [tiny-typed-emitter](https://www.npmjs.com/package/tiny-typed-emitter) instead? Would make every method strictly typed then 👀

skick1234: @iCrawl removed this with https://github.com/discordjs/discord.js/commit/d4b41dd0815b493b599d4f4d1b6dd18cd99f91ea. My description mentioned it.

GravIos3:
@iCrawl @kyranet @SpaceEEC Please review
@skick1234 Can you resolve the conversation?

PR #8121, opened 2022-06-19
fix: vitest recursion

didinele:


didinele:
this will be remembered as the time where I force pushed all my commits away so the PR auto closed

PR #8033, opened 2022-06-07
refactor: move all the config files to root

imranbarbhuiya:
**Please describe the changes this PR makes and why it should be merged:**
This pr moves all the config file to root to reduce repetition.
**Status and versioning classification:**
- This PR **only** includes non-code changes, like changes to documentation, README, etc.

PR #8129, opened 2022-06-20
chore: move deps to root and some miscellaneous changes

imranbarbhuiya:
first sorry for doing all these in one pr. I just don't want to make these small changes in multiple prs.
1. followup of #8033 . Now that all the configs are present in root so we can move deps to root too.
2. change `prepublishOnly` to `prepack` [ref](https://yarnpkg.com/advanced/lifecycle-scripts)
3. remove some `yarn test` from `prepack` as these packages don't have any test script
4. enable cache in eslint and prettier. turbo cache checks for full folder changes but using eslint cache will check for single single file changes.
5. stage files in pre-commit


iCrawl:
Not interested in moving prettier to an eslint rule because: <https://prettier.io/docs/en/integrating-with-linters.html#notes>

imranbarbhuiya:
ok I'll revert it with `--cache` enabled

iCrawl:
I'm not sure if any of those changes are worth it. Esp hoisting eslint and `@types/node` (which is even the wrong move to hoist and shouldn't be hoisted).

imranbarbhuiya:
ci is failing but I don't think it's related to my changes as it was failing with same error in #8133

iCrawl (approved):

PR #8136, opened 2022-06-20
fix(vcs): nsfw property v13 backport

CodeGoat-dev:
**Please describe the changes this PR makes and why it should be merged:**

Backports #8132 to v13.

**Status and versioning classification:**

- Code changes have been tested against the Discord API, or there are no code changes
- I know how to update typings and have done so, or typings don't need updating

PR #8147, opened 2022-06-22
types: add missing nsfw type to voice channels

CodeGoat-dev:

Adds the missing nsfw type to the voice channel structure.

### DIFF
--- a/packages/discord.js/src/structures/VoiceChannel.js
+++ b/packages/discord.js/src/structures/VoiceChannel.js
@@ -63,6 +63,10 @@ class VoiceChannel extends BaseGuildVoiceChannel {
     }
 
     if ('nsfw' in data) {
+      /**
+       * If the guild considers this channel NSFW
+       * @type {boolean}
+       */
       this.nsfw = Boolean(data.nsfw);
     }
   }


packages/discord.js/src/structures/VoiceChannel.js
```diff
@@ -63,6 +63,10 @@ class VoiceChannel extends BaseGuildVoiceChannel {
     }
 
     if ('nsfw' in data) {
+      /**
+       * If the guild considers this channel NSFW
+       * @type {boolean}
+       */
```
almeidx, line 69: This is already in the constructor, and it shows in the documentation. I don't think this is necessary
https://github.com/discordjs/discord.js/blob/65d1879c0ae2d8d820323d0d835b1e5e3d071e57/packages/discord.js/src/structures/VoiceChannel.js#L23-L26

CodeGoat-dev: I only added it because This change was recommended in #8136 and all other properties have types.

almeidx: `nsfw` has a type too, but it's being set in the constructor. All other properties are being set in the `_patch()` method, `nsfw` is an exception which is being set in both the constructor and the `_patch()` method

CodeGoat-dev:
Should I close this and update #8136 accordingly?

Jiralite:
The type of this property is not missing. It is a couple of lines above where you have implemented this change.

PR #8159, opened 2022-06-22
revert: refactor: move eslint.tsconfig.json to root

ckohen:

.eslintrc.json is smart and loads from the local config based on the project parameter at root, tsconfig.eslint.json is not so smart, so it was ignoring the tsconfig from any given package

### DIFF
--- a/packages/actions/tsconfig.eslint.json
+++ b/packages/actions/tsconfig.eslint.json
@@ -1,3 +1,20 @@
 {
-	"extends": "../../tsconfig.eslint.json"
+	"extends": "./tsconfig.json",
+	"compilerOptions": {
+		"allowJs": true
+	},
+	"include": [
+		"**/*.ts",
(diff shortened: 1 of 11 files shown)


packages/actions/tsconfig.eslint.json
```diff
@@ -1,3 +1,20 @@
 {
-	"extends": "../../tsconfig.eslint.json"
+	"extends": "./tsconfig.json",
+	"compilerOptions": {
+		"allowJs": true
+	},
```
kyranet, line 5: I don't think we should need this option for TypeScript-only packages, only `discord.js` would use this.

kyranet (changes requested):

imranbarbhuiya:
You can make a js file just like `tsup.config.ts` and use that instead of copying full config file.

iCrawl (approved):

PR #8166, opened 2022-06-24
feat(util): parseWebhookURL

didinele:
This PR moves the URL parsing part out of the `WebhookClient` constructor and instead makes it a util

### DIFF
--- a/packages/discord.js/src/util/Util.js
+++ b/packages/discord.js/src/util/Util.js
@@ -528,6 +528,32 @@ function lazy(cb) {
   return () => (defaultValue ??= cb());
 }
 
+/**
+ * Represents the credentials used for a given webhook
+ * @typedef {object} WebhookCredentials
+ * @property {string} id The webhook's id
+ * @property {string} token The webhook's token
+ */
+
+/**
+ * Parses a webhook URL for the id and token
+ * @param {string} url The URL to parse
+ * @returns {?WebhookCredentials} Null if the URL is invalid, otherwise the id and the token
+ */
+function parseWebhookURL(url) {
+  const matches = url.match(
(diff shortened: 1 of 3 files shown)


packages/discord.js/src/util/Util.js
```diff
@@ -528,6 +528,32 @@ function lazy(cb) {
   return () => (defaultValue ??= cb());
 }
 
+/**
+ * Represents the credentials used for a given webhook
+ * @typedef {object} WebhookCredentials
```
Jiralite, line 533:
```suggestion
 * @typedef {Object} WebhookCredentials
```

Jiralite (changes requested):

SpaceEEC (approved):

packages/discord.js/typings/index.d.ts
```diff
@@ -2625,6 +2625,11 @@ export class UserFlagsBitField extends BitField<UserFlagsString> {
   public static resolve(bit?: BitFieldResolvable<UserFlagsString, number>): number;
 }
 
+export interface WebhookCredentials {
+  id: string;
+  token: string;
+}
```
Jiralite, line 2631: Should this not be a `Snowflake`? And then, wouldn't it be the same as this?

https://github.com/discordjs/discord.js/blob/ffc439e58e020b98f7d36bc47186778c3ce803eb/packages/discord.js/typings/index.d.ts#L5462-L5465

Jiralite (approved):

vladfrangu (approved):

kyranet (approved):
RegExp groups combined with `RegExp.prototype.exec` would be better, but that's out of scope for this PR, can be done in a follow-up.

PR #8170, opened 2022-06-24
Convert .login token parameter to an object

lewis45454:
This PR converts the .login function token parameter to an object.

### DIFF
--- a/README.md
+++ b/README.md
@@ -96,7 +96,7 @@ client.on('interactionCreate', async (interaction) => {
 	}
 });
 
-client.login('token');
+client.login({ token: 'bot token' });
 ```
 
 ## Links
--- a/packages/discord.js/src/client/Client.js
+++ b/packages/discord.js/src/client/Client.js
@@ -208,7 +208,7 @@ class Client extends BaseClient {
    * @param {string} [token=this.token] Token of the account to log in with
    * @returns {Promise<string>} Token of the account used
    * @example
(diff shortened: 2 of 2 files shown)


kyranet:
What's the reasoning behind this? There are no more options for `Client.login`, nor we plan to make everything that takes one parameter be an object.

kyranet:
By that logic, we should make everything take an object, but no, it does not make code more readable.

If you want to make your usage more readable, you can try something like:

```javascript
import { token } from './config.js';

await client.login(token);
```

Or if you use `.env`, you can just define the `DISCORD_TOKEN` variable, and do the following:

```javascript
await client.login();
```

As it's what the token defaults to:

https://github.com/discordjs/discord.js/blob/e24970e3c3d24f71ba711e59666cd8a49a33e33b/packages/discord.js/src/client/Client.js#L143-L150

lewis45454:
Would it be better if I changed "my token" to "bot token" for the jsdoc and revert the object back to a single value?

kyranet:
Furthermore, Visual Studio Code has this option: `javascript.inlayHints.parameterNames.enabled`

When enabled, it will show the name of the argument, as shown below:

image

---

About your question, I really have no preference, we use [`#library-discussion`](https://canary.discord.com/channels/222078108977594368/682166281826598932) in [discord.js](https://discord.gg/djs) to discuss changes in the library.

PR #8177, opened 2022-06-27
fix: Remove global flag on regular expressions

Jiralite:
**Please describe the changes this PR makes and why it should be merged:**
Resolves #8175 by removing the global flag on regular expressions. Some are now imported from discord-api-types and they all used named capturing groups now.

**Status and versioning classification:**
- Code changes have been tested against the Discord API, or there are no code changes
- I know how to update typings and have done so, or typings don't need updating
- This PR includes breaking changes (methods removed or renamed, parameters moved or removed)

PR #8178, opened 2022-06-27
fix: Use non-global flag whilst resolving regular expressions

Jiralite:
Backports #8177 to version 13.

### DIFF
--- a/src/util/DataResolver.js
+++ b/src/util/DataResolver.js
@@ -34,7 +34,7 @@ class DataResolver extends null {
    * @returns {string}
    */
   static resolveCode(data, regex) {
-    return data.matchAll(regex).next().value?.[1] ?? data;
+    return new RegExp(regex.source).exec(data)?.[1] ?? data;
   }
 
   /**


SpaceEEC (approved):

vladfrangu (approved):

src/util/DataResolver.js
```diff
@@ -34,7 +34,7 @@ class DataResolver extends null {
    * @returns {string}
    */
   static resolveCode(data, regex) {
-    return data.matchAll(regex).next().value?.[1] ?? data;
+    return new RegExp(regex.source).exec(data)?.[1] ?? data;
```
kyranet, line 37: Ideally, we should check for [`RegExp.prototype.global`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/global) before creating a clone.

kyranet (approved):
Approving since it indeed fixes the bug, but I'll add something that can further improve the method.

PR #8183, opened 2022-06-28
feat: add website documentation early mvp

suneettipirneni:

Adds routing to documentation with very early designs.

Entry point for documentation is`/docs/packages/discord.js`

Screen Shot 2022-06-28 at 12 51 55 PM

Screen Shot 2022-06-28 at 12 52 26 PM


packages/website/tsconfig.json
```diff
@@ -11,5 +11,5 @@
 			"~/*": ["./src/*"]
 		}
 	},
-	"include": ["src/**/*.ts"]
+	"include": ["src/**/*.ts", "src/model.server.tsx"]
```
almeidx, line 14:
```suggestion
	"include": ["src/**/*.ts", "src/**/*.tsx"]
```

almeidx: Asides from this, the [remix.env.d.ts](https://github.com/discordjs/discord.js/blob/main/packages/website/remix.env.d.ts) and [types.d.ts](https://github.com/discordjs/discord.js/blob/main/packages/website/types.d.ts) files on the website root should probably be included too

packages/website/src/model.server.tsx
```diff
@@ -0,0 +1,215 @@
+import path from 'path';
```
SuperchupuDev, line 1:
```suggestion
import path from 'node:path';
```

PR #8197, opened 2022-06-29
fix(GuildMemberManager): unable to change client's nickname

N1ckPro:
Currently `data.communication_disabled_until` is always being assigned (even with undefined).
That leads to the if check never being true as `Object.keys` will always return `communication_disabled_until`, even when trying to change just the client's nickname
https://github.com/discordjs/discord.js/blob/c4653f97b1529eb0b99fccdba67c37eb4f467ff9/packages/discord.js/src/managers/GuildMemberManager.js#L310-L311


kyranet:
This is already fixed with #8066

PR #8200, opened 2022-06-30
chore: revert pre-commit hook to not change state

didinele:
This PR reverts point 5 from #8129, which made our pre-commit hook ` && git add .` This is a *really* bad idea, as it makes it impossible to only commit select changes without skipping hooks, and generally speaking, pre-commit [should never change state](https://github.com/pre-commit/pre-commit/issues/806#issuecomment-408278608)

### DIFF
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,4 +1,4 @@
 #!/bin/sh
 . "$(dirname "$0")/_/husky.sh"
 
-yarn build && yarn format && git add .
+yarn build && yarn format


imranbarbhuiya:
> By that logic why should we even run yarn format? The changes would require a secondary commit

this was the only reason I added it. Why can't we use lint-stage?

didinele:
Last commit addresses this, we no longer run `format`, but `lint` instead. If that fails, you can run `yarn format && git add .` yourself.

PR #8202, opened 2022-06-30
fix(MessageMentions): `ignoreRepliedUser` option in `has()`

### LINKED ISSUE #8201: message.mentions.has ignoreRepliedUser bug

discord.js

### Issue description

Reproduce -
1. Put in the code sample with the token and channel_id
2. Reply to the message from the bot
3. bot replies back even though ignoreRepliedUser is true

if ignoreDirect is true and ignoreRepliedUser is true then the bot WON'T reply
if ignoreRepliedUser is true then the bot WILL reply
if ignoreDirect is true then the bot WILL reply

### Code sample

```typescript
const client = new Client({ intents: 513 });

client.on('ready', function(client) {
    let channel = client.channels.cache.get(channel_id);
    channel.send({content: 'test'});
})

client.on('messageCreate', function(message) {
    if(message.mentions.has(message.client.user, {ignoreRepliedUser: true})) {
        message.reply({content: 'test'});
    }
})

client.login(BOT_TOKEN);
```

### Package version

discord.js@13.8.1

### Node.js version

16.13.2

### Operating system

Windows 11

Low (slightly annoying)

No Partials

Guilds, GuildMessages

_No response_

almeidx:
Fixes #8201

- The `ignoreRepliedUser` option was not being respected when paired with `ignoreDirect: false`
- Added a `parsedUsers` getter, which parses user mentions from the message content to address the following comment:
  - https://github.com/discordjs/discord.js/pull/8202#issuecomment-1172209705
- Moved the `ignoreEveryone` condition to the top as the other options don't affect it

### DIFF
--- a/packages/discord.js/src/structures/MessageMentions.js
+++ b/packages/discord.js/src/structures/MessageMentions.js
@@ -134,6 +134,13 @@ class MessageMentions {
     this._channels = null;
 
     /**
+     * Cached users for {@link MessageMentions#parsedUsers}
+     * @type {?Collection<Snowflake, User>}
+     * @private
+     */
+    this._parsedUsers = null;
+
+    /**
      * Crossposted channel data.
      * @typedef {Object} CrosspostedChannel
      * @property {Snowflake} channelId The mentioned channel's id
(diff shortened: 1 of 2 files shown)


Syjalo:
If we'll mention the user in the content and by reply then it also will ignore the content mention. As I said in the issue we should remove the `ignoreRepliedUser` option or parse user mentions in the message content like channel ones.

DraftProducts (approved):

iCrawl (approved):

kyranet (approved):

SpaceEEC (approved):

PR #8210, opened 2022-07-01
refactor(Util): make single `replace` call in `cleanContent`

kyranet:

Doing multiple `.replace` risks subsequent calls from running in the previous replaces, and this also comes with an extra performance impact where the content is scanned 3 times, and then each mention is scanned and replaced *again* to get the ID. This is a much simpler approach that uses a single call and RegExp capture groups to avoid extra processing.

The PR is https://github.com/discordjs/discord.js/labels/semver%3Amajor because it changes the matching ID in mentions from `[0-9]+` to the much stricter and safer `\d{17,19}`, which may result on different results if somebody injects arbitrary IDs outside the valid range in Discord's API.

### DIFF
--- a/packages/discord.js/src/util/Util.js
+++ b/packages/discord.js/src/util/Util.js
@@ -474,6 +474,7 @@ function basename(path, ext) {
   const res = parse(path);
   return ext && res.ext.startsWith(ext) ? res.name : res.base.split('?')[0];
 }
+
 /**
  * The content to have all mentions replaced by the equivalent text.
  * @param {string} str The string to be converted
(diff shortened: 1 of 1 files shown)


packages/discord.js/src/util/Util.js
```diff
@@ -474,39 +474,45 @@ function basename(path, ext) {
...
+          return user ? `@${user.username}` : match;
+        }
+
+        const member = channel.guild.members.cache.get(id);
+        if (member) {
+          return `@${member.displayName}`;
+        } else {
+          const user = channel.client.users.cache.get(id);
+          return user ? `@${user.username}` : match;
+        }
```
almeidx, line 500:
```suggestion
        const member = channel.guild?.members.cache.get(id);
        if (channel.type === ChannelType.DM || !member) {
          const user = channel.client.users.cache.get(id);
          return user ? `@${user.username}` : match;
        }
        return `@${member.displayName}`;
```

almeidx: Or
```suggestion
        const member = channel.guild?.members.cache.get(id);
        if (member) {
          return `@${member.displayName}`;
        }
        
        const user = channel.client.users.cache.get(id);
        return user ? `@${user.username}` : match;
```

almeidx (changes requested):

packages/discord.js/src/util/Util.js
```diff
@@ -474,39 +474,40 @@ function basename(path, ext) {
...
  */
 function cleanContent(str, channel) {
-  str = str
-    .replace(/<@!?[0-9]+>/g, input => {
-      const id = input.replace(/<|!|>|@/g, '');
-      if (channel.type === ChannelType.DM) {
-        const user = channel.client.users.cache.get(id);
-        return user ? `@${user.username}` : input;
-      }
+  return str.replaceAll(/<(@[!&]?|#)(\d{17,19})>/g, (match, type, id) => {
```
vladfrangu, line 485: I think we should keep it like it was before with `[0-9]+`, instead of having to update the pattern every time ids bump a number (even if that would take many years) but I wouldn't call that a blocker, just a thought

vladfrangu (approved):

packages/discord.js/src/util/Util.js
imranbarbhuiya: `[0-9]+` can match invalid mentions too, we can also use `\d{17,}` so we don't need to increment it every time but it won't match ids with length <17

SpaceEEC (approved):

packages/discord.js/src/util/Util.js
kyranet: There'll be maximum a 20-digit ID, but that's assuming Discord's database uses `u64` (range up to 18,446,744,073,709,551,615, 20 digits) and not `i64` (range up to 9,223,372,036,854,775,807, 19 digits).

Either way, we would need to wait until 2074 (iirc from my old calculations) for us to run out of `i64` IDs and start needing `u64`.

A maximum range of 19 already surpasses the longevity of this version and many others.

iCrawl (approved):

PR #8245, opened 2022-07-06
feat(website): show package members in a sidebar

suneettipirneni:

Package items are now displayed in a sidebar with the item themselves as the content view.


packages/website/src/components/ItemSidebar.tsx
```diff
@@ -0,0 +1,31 @@
...
+	data: {
+		members: ReturnType<typeof getMembers>;
+	};
+}
+
+export function ItemSidebar({ packageName, data }: ItemListProps) {
+	return (
+		<div className="flex flex-col px-7">
+			<div className="flex items-center content-center">
+				<h1 className="px-2 font-mono">
```
iCrawl, line 16:
```suggestion
			<div>
				<h1 className="px-2 font-mono flex items-center content-center">
```

Should probably apply this here, otherwise the icons and the text are not vertically centered to each other.

packages/website/src/components/ItemSidebar.tsx
```diff
@@ -0,0 +1,31 @@
...
+					<VscPackage className="px-1" />
+					{`${packageName}`}
+				</h1>
+			</div>
+			{data.members.map((member, i) => (
+				<div key={i} className="flex mb-1 content-center items-center align-center">
+					<a className="font-mono no-underline break-all color-blue-500" href={member.path}>
```
iCrawl, line 23:
```suggestion
				<div key={i} className="mb-1">
					<a className="flex content-center items-center align-center font-mono no-underline break-all color-blue-500" href={member.path}>
```

Same as above.

iCrawl (changes requested):

packages/website/src/components/ItemSidebar.tsx
iCrawl: Reason: The div only has one child, while the `h1` is the one that has 2 that need to be centered to each other.

packages/website/src/components/ItemSidebar.tsx
iCrawl: Reason: same as above, but inside the `a` element instead.

iCrawl:
<details>
	<summary>We should think about doing something starting from the breakpoint 1020px in width:</summary>
	image
</details>

<details>
	<summary>Around 820px width:</summary>
	image
</details>

<details>
	<summary>Around 700px width:</summary>
	image
	image
</details>

<details>
	<summary>Finally at around 640px width:</summary>
	image
</details>

PR #8250, opened 2022-07-08
add bun support

North-West-Wind:
Supports Bun. the new fast Javascript runtime, by over-complicating a single-line of regex into 36 lines of spaghetti

### DIFF
--- a/packages/discord.js/src/util/Util.js
+++ b/packages/discord.js/src/util/Util.js
@@ -105,8 +105,42 @@ function escapeMarkdown(
       .join(codeBlock ? '\\`\\`\\`' : '```');
   }
   if (!inlineCodeContent) {
-    return text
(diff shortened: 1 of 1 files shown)


packages/discord.js/src/util/Util.js
```diff
@@ -105,8 +105,42 @@ function escapeMarkdown(
       .join(codeBlock ? '\\`\\`\\`' : '```');
   }
   if (!inlineCodeContent) {
-    return text
-      .split(/(?<=^|[^`])`(?=[^`]|$)/g)
+    const noTriple = text.split("```");
```
imranbarbhuiya, line 108: instead of changing it for everyone, u can check the `process.version`  and do it for only bun users.

imranbarbhuiya (changes requested):

PR #8254, opened 2022-07-08
refactor(Embed): Add all the types

Jiralite:
The type definitions for this class were missing. This pull request adds them all in.

Additionally, `Embed#fields` is now always an array and is no longer nullable.


packages/discord.js/src/structures/Embed.js
```diff
@@ -6,31 +6,26 @@ const isEqual = require('fast-deep-equal');
  * Represents an embed.
  */
 class Embed {
-  /**
-   * Creates a new embed object
-   * @param {APIEmbed} data API embed data
-   * @private
```
vladfrangu, line 12: Do we want to document this constructor? Iirc we don't document others, like Channels for example 👀

vladfrangu (approved):

packages/discord.js/src/structures/Embed.js
Jiralite: We can't document this as `@private` else it won't show in the sidebar in the documentation. If we publicly document it, well, they should be using `EmbedBuilder` instead.

iCrawl (approved):

PR #8265, opened 2022-07-11
fix: slash command name regex

ThaTiemsz:
Explicitly adds the letter unicode properties to the slash command name regex, because the negated `{Lu}` resulted in the regex matching _any_ character except uppercase letters.

**Old regex:**
image

**New regex:**
image

image


vladfrangu (approved):
Times like these I wish we had the python re so we could just use discord's regex

iCrawl (approved):

PR #8271, opened 2022-07-12
types(builders): allow action row builders to infer generic type

suneettipirneni:

Previously doing this:

```ts
const row = new ActionRowBuilder().addComponents([new TextInputComponentBuilder()]);
```
Would yield the type `ActionRowBuilder<AnyComponentType>` for `row`
With this PR, the generic parameter is now inferred so row would yield: `ActionRowBuilder<TextInputComponent>`.


imranbarbhuiya:
Why this one is closed?

PR #8280, opened 2022-07-13
feat: add `applicationCommandMention()` formatter

almeidx:
- Added `applicationCommandMention()` formatted in /builders
- Added `ApplicationCommand#toString()`, which will use the function mentioned above to return the commands mention

Upstream:
- discord/discord-api-docs#5186

### DIFF
--- a/packages/ws/README.md
+++ b/packages/ws/README.md
@@ -1,16 +1,19 @@
 <div align="center">
-  <br />
-  <p>
-    <a href="https://discord.js.org"><img src="https://discord.js.org/static/logo.svg" width="546" alt="discord.js" /></a>
-  </p>
-  <br />
-  <p>
-    <a href="https://discord.gg/djs"><img src="https://img.shields.io/discord/222078108977594368?color=5865F2&logo=discord&logoColor=white" alt="Discord server" /></a>
-    <a href="https://www.npmjs.com/package/@discordjs/ws"><img src="https://img.shields.io/npm/v/@discordjs/ws.svg?maxAge=3600" alt="npm version" /></a>
-    <a href="https://www.npmjs.com/package/@discordjs/ws"><img src="https://img.shields.io/npm/dt/@discordjs/ws.svg?maxAge=3600" alt="npm downloads" /></a>
-    <a href="https://github.com/discordjs/discord.js/actions"><img src="https://github.com/discordjs/discord.js/actions/workflows/test.yml/badge.svg" alt="Build status" /></a>
-    <a href="https://codecov.io/gh/discordjs/discord.js" ><img src="https://codecov.io/gh/discordjs/discord.js/branch/main/graph/badge.svg?precision=2&flag=ws" alt="Code coverage" /></a>
-  </p>
+	<br />
+	<p>
+		<a href="https://discord.js.org"><img src="https://discord.js.org/static/logo.svg" width="546" alt="discord.js" /></a>
+	</p>
+	<br />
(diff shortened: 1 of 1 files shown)


packages/discord.js/src/util/Formatters.js
```diff
@@ -24,6 +25,16 @@ const {
...
 class Formatters extends null {
+  /**
+   * Formats an application command name and id into an application command mention.
+   * @method applicationCommandMention
+   * @memberof Formatters
+   * @param {string} commandName The name of the application command.
+   * @param {string} commandId The id of the application command.
+   * @returns {string}
+   */
+  static applicationCommandMention = applicationCommandMention;
```
ImRodry, line 36: could you add this to ApplicationCommand#toString()?

almeidx: And how would subcommands (and subcommand groups) be handled? Would they just be ignored?

ImRodry: Well yeah there's no way to mention a subcommand

packages/builders/src/messages/formatters.ts
```diff
@@ -182,6 +182,75 @@ export function roleMention<C extends Snowflake>(roleId: C): `<@&${C}>` {
...
+ * @param subcommandGroupName - The subcommand group name to format
+ * @param subcommandName - The subcommand name to format
+ * @param commandId - The application command ID to format
+ */
+export function applicationCommandMention<N extends string, G extends string, S extends string, I extends Snowflake>(
+	commandName: N,
+	subcommandGroupName: G,
+	subcommandName: S,
+	commandId: I,
+): `</${N} ${G} ${S}:${I}>`;
```
ImRodry, line 198: You shouldn't add this imo. It doesn't actually work, it just displays the text nicely but you can't click it.

almeidx: It works on my end

packages/discord.js/src/util/Formatters.js
ImRodry: like I said you can write any text so I think toString should only put the name of the command

almeidx: 062bcccaf14d5c2a02693e3ece09dca1e8615af2

vladfrangu (approved):

kyranet:
This PR needs a rebase.

almeidx:
Rebased

iCrawl (approved):

thehackerboi69github:
This should also be used in https://github.com/discordjs/discord.js/blob/main/packages/discord.js/src/structures/ChatInputCommandInteraction.js

almeidx:
> This should also be used in https://github.com/discordjs/discord.js/blob/main/packages/discord.js/src/structures/ChatInputCommandInteraction.js

You cannot include options in command mentions

thehackerboi69github:
> > This should also be used in https://github.com/discordjs/discord.js/blob/main/packages/discord.js/src/structures/ChatInputCommandInteraction.js
> 
> You cannot include options in command mentions
> 

That's true, but now that there are attachment options, `ChatInputCommandInteraction#toString` is not 100% working anymore, and now that there is an official application command mention format, I feel like we should use the official method instead of a custom, not fully working one.

kyranet (approved):

almeidx:
Can't reopen this pull request now. Made a new one, #8546

PR #8281, opened 2022-07-13
feat(GuildMember): add dmChannel getter

ImRodry:

This PR adds the dmChannel getter to the GuildMember class, similar to the one on the User class. If there's the createDM and deleteDM methods it only makes sense to have this

### DIFF
--- a/packages/discord.js/src/structures/GuildMember.js
+++ b/packages/discord.js/src/structures/GuildMember.js
@@ -211,6 +211,15 @@ class GuildMember extends Base {
   }
 
   /**
+   * The DM between the client's user and this member
+   * @type {?DMChannel}
+   * @readonly
+   */
+  get dmChannel() {
+    return this.client.users.dmChannel(this.id);
+  }
+
+  /**
    * The nickname of this member, or their username if they don't have one
    * @type {?string}
    * @readonly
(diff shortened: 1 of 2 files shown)


packages/discord.js/src/structures/GuildMember.js
```diff
@@ -210,6 +210,15 @@ class GuildMember extends Base {
     return this.user.id;
   }
 
+  /**
+   * The DM between the client's user and this member
```
almeidx, line 214: Might be worth mentioning that it only returns the channel if it is cached (I know the description is the same as the User class and the private method, but perhaps those should be updated too)

almeidx: I don't know. One could assume it returns `null` if the channel does not exist at all

ImRodry: hmm not sure, it's still a getter and all getters rely on cache (or almost all) so I don't think it's necessary. If a maintainer does I can change this and the other descriptions so I'll keep this open until one of them resolves it

SpaceEEC (approved):

vladfrangu (approved):
Sure

iCrawl (approved):

PR #8282, opened 2022-07-14
fix(PermissionOverwriteManager): mutates user

AppEternal:
Prevent the mutation of the user/role as it's passed into upset. 

The user passed into upset would always be of type "User" preventing the use of "GuildMember" in upset. 

It's only noticed if you have User cache disabled and only wish to use guild member cache, this fixes that limitation by passing along the original data. 

### DIFF
--- a/src/managers/PermissionOverwriteManager.js
+++ b/src/managers/PermissionOverwriteManager.js
@@ -142,8 +142,9 @@ class PermissionOverwriteManager extends CachedManager {
    *   .catch(console.error);
    */
   edit(userOrRole, options, overwriteOptions) {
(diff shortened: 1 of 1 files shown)


Jiralite:
Could you change the branch this is targeting to v13, please?

Jiralite:
This likely exists in the main branch too, so it'd be appreciated if a pull request could be opened there too!

AppEternal:
> This likely exists in the main branch too, so it'd be appreciated if a pull request could be opened there too!

Done #8283

iCrawl (approved):

PR #8288, opened 2022-07-14
types(CategoryChannelChildManager): fix Holds type

### LINKED ISSUE #8287: Incorrect type for CategoryChannelChildManager#cache

discord.js

### Issue description

The type for CategoryChannelChildManager#cache is wrong as it says it can include thread channels, which is not true. The type for Holds on the manager is currently NonCategoryGuildBasedChannel which simply removes CategoryChannel from GuildBasedChannel, it should instead remove thread channels too.

### Code sample

```typescript
const category = guild.channels.cache.get("ID_HERE") as CategoryChannel
for (const child of category.children.cache.values())
	child.setParent("PARENT_ID", { reason: "Moving this channel" }) // Error: Property 'setParent' does not exist on type 'NonCategoryGuildBasedChannel'. Property 'setParent' does not exist on type 'PublicThreadChannel'. ts(2339)
```

### Package version

14.0.0-dev.1657411900-f0b68d5

### Node.js version

18.5.0

### Operating system

Windows 11

Medium (should be fixed soon)

f0b68d5

ImRodry:

Fixes #8287. If you have better suggestions for the type feel free to leave them here.

### DIFF
--- a/packages/discord.js/typings/index.d.ts
+++ b/packages/discord.js/typings/index.d.ts
@@ -3279,11 +3279,7 @@ export class BaseGuildEmojiManager extends CachedManager<Snowflake, GuildEmoji, 
   public resolveIdentifier(emoji: EmojiIdentifierResolvable): string | null;
 }
 
-export class CategoryChannelChildManager extends DataManager<
-  Snowflake,
-  NonCategoryGuildBasedChannel,
-  GuildChannelResolvable
-> {
+export class CategoryChannelChildManager extends DataManager<Snowflake, CategoryChildChannel, GuildChannelResolvable> {
   private constructor(channel: CategoryChannel);
 
   public channel: CategoryChannel;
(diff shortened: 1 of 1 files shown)


packages/discord.js/typings/index.d.ts
```diff
@@ -5433,7 +5429,7 @@ export type VoiceBasedChannel = Extract<Channel, { bitrate: number }>;
 
 export type GuildBasedChannel = Extract<Channel, { guild: Guild }>;
 
-export type NonCategoryGuildBasedChannel = Exclude<GuildBasedChannel, CategoryChannel>;
+export type CategoryChildChannel = Exclude<Extract<Channel, { parent: CategoryChannel | null }>, CategoryChannel>;
```
vladfrangu, line 5432: Is the Exclude necessary?

ImRodry: Yes otherwise the type would include CategoryChannel. The property is inherited from GuildChannel and even if you override it with null (which is the value it will always have) it would still show up, I tested that

vladfrangu (approved):

iCrawl (approved):

PR #8290, opened 2022-07-14
refactor(rest): add content-type(s) to uploads

### LINKED ISSUE #8285: Sticker uploading doesn't work in v14

discord.js

### Issue description

Sticker creation works in v13 but in v14, you get `DiscordAPIError[50046]: Invalid Asset`. I think it's an issue in djs/rest, but I'm not 100% sure.

This can be seen by simply running the first example in https://discord.js.org/#/docs/discord.js/main/class/GuildStickerManager?scrollTo=create.

When running identical code with `Routes.guildStickers` and `Routes.channelMessages`, the latter works but the former returns the error.

I've been debugging this for an hour but can't figure it out.

### Code sample

_No response_

### Package version

14.0.0-dev.1657757514-fe34f48

### Node.js version

v16.15.1

### Operating system

Windows 11 22H2 25151.1010

Medium (should be fixed soon)

_No response_

didinele:
This PR adds a `type` to our `Blob`s that we use when we send files. As it turns out, Discord does *NOT* check for magic bytes, so for endpoints such as sticker creation that check for a png/apng file, if the content-type isn't precisely set to `image/png`, they will yield a 400.

For ref on the sticker endpoint, this worked fine in v13 since we actually hard coded the Content-Type to be an image, but this is not a great solution. If an endpoint that takes, say, only gifs were to be added, we'd be running into essentially the same problem.

We try our best to infer this:
- when passing in a URL, we'll read the `Content-Type` from the response
- if a Buffer was passed in without a content type, we use a 3rd party pkg within `@discordjs/rest` to parse the MIME type based off of the magic bytes

This should cover most use cases and as far as I've thought about it, have no unwanted side effects, other than perhaps a website returning a miss-leading `Content-Type`, when you pass in a URL, but that is not our problem™️.

Also closes #8285

### DIFF
--- a/packages/discord.js/src/structures/MessagePayload.js
+++ b/packages/discord.js/src/structures/MessagePayload.js
@@ -255,8 +255,8 @@ class MessagePayload {
       name = fileLike.name ?? findName(attachment);
     }
 
-    const data = await DataResolver.resolveFile(attachment);
-    return { data, name };
+    const { data, contentType } = await DataResolver.resolveFile(attachment);
+    return { data, name, contentType };
   }
 
   /**
--- a/packages/discord.js/src/util/DataResolver.js
+++ b/packages/discord.js/src/util/DataResolver.js
@@ -101,31 +101,36 @@ class DataResolver extends null {
    */
 
   /**
+   * @typedef {Object} ResolvedFile
+   * @property {Buffer} data Buffer containing the file data
+   * @property {string} [contentType] Content type of the file
+   */
+
+  /**
    * Resolves a BufferResolvable to a Buffer.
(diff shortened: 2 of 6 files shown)


vladfrangu (approved):

advaith1:
can this be an issue? from file-type's readme:

> This package is for detecting binary-based file formats, not text-based formats like .txt, .csv, .svg, etc.

didinele:
No. It'll just return null in that case, causing us to set no type:
https://github.com/discordjs/discord.js/blob/2c1bcdb9c13b5936be4a8ed02edc62e984f15851/packages/rest/src/lib/RequestManager.ts#L414-L418

iCrawl (approved):

PR #8293, opened 2022-07-15
chore: ignore index files in coverage

almeidx:
Ignores all the `index.ts` files, as they only contain exports from other files

Because of that, I moved the Collection source into a separate file and exported it in `index.ts`. We could also make the main entry point the new file(?) instead of re-exporting in an index file, however, I don't think it matters given that `tsup` bundles everything into a single file anyway

cc: @ckohen 


vitest.config.ts
```diff
@@ -9,8 +9,12 @@ export default defineConfig({
 			all: true,
 			reporter: ['text', 'lcov', 'cobertura'],
 			include: ['src'],
-			// All ts files that only contain types, due to ALL
-			exclude: ['**/*.{interface,type,d}.ts'],
+			exclude: [
+				// All ts files that only contain types, due to ALL
+				'**/*.{interface,type,d}.ts',
+				// All index files that only contain exports from other files
+				'**/index.ts',
```
ckohen, line 16:
```suggestion
				// All index files that *should* only contain exports from other files
				'**/index.{js,ts}',
```

ckohen (changes requested):

vladfrangu (approved):

iCrawl (approved):

PR #8311, opened 2022-07-18
fix(ApplicationCommandManager): allow passing 0n to defaultMemberPermissions

almeidx:
`defaultMemberPermissions` can be `0n`, and it was being passed as a bigint to JSON.stringify(), which doesn't know how to serialize them


SpaceEEC (dismissed):
Shouldn't the same fix be made for `default_member_permissions` (a few lines above) and in `ApplicationCommand#equals` too?

Jiralite:
`default_member_permissions` should only take a string. It's meant to represent the raw API data, which doesn't take big integers.

As for `ApplicationCommand#equals()`, that would be an oversight that needs changing too.

PR #8326, opened 2022-07-21
feat(GuildBanManager#create): add `deleteMessageSeconds`

advaith1:
* Adds `deleteMessageSeconds`
* Marks `deleteMessageDays` as deprecated

### DIFF
--- a/packages/discord.js/src/managers/GuildBanManager.js
+++ b/packages/discord.js/src/managers/GuildBanManager.js
@@ -129,6 +129,9 @@ class GuildBanManager extends CachedManager {
    * Options used to ban a user from a guild.
    * @typedef {Object} BanOptions
    * @property {number} [deleteMessageDays] Number of days of messages to delete, must be between 0 and 7, inclusive
+   * (deprecated, use deleteMessageSeconds instead)
+   * @property {number} [deleteMessageSeconds] Number of seconds of messages to delete,
+   * must be between 0 and 604800 (7 days), inclusive
    * @property {string} [reason] The reason for the ban
    */
 
@@ -150,7 +153,10 @@ class GuildBanManager extends CachedManager {
     const id = this.client.users.resolveId(user);
     if (!id) throw new Error(ErrorCodes.BanResolveId, true);
     await this.client.rest.put(Routes.guildBan(this.guild.id, id), {
-      body: { delete_message_days: options.deleteMessageDays },
+      body: {
+        delete_message_days: options.deleteMessageDays,
+        delete_message_seconds: options.deleteMessageSeconds,
+      },
       reason: options.reason,
     });
     if (user instanceof GuildMember) return user;
--- a/packages/discord.js/src/structures/GuildMember.js
+++ b/packages/discord.js/src/structures/GuildMember.js
@@ -346,8 +346,8 @@ class GuildMember extends Base {
    * @param {BanOptions} [options] Options for the ban
    * @returns {Promise<GuildMember>}
    * @example
-   * // ban a guild member
-   * guildMember.ban({ deleteMessageDays: 7, reason: 'They deserved it' })
+   * // ban a guild member, deleting a week of messages
+   * guildMember.ban({ deleteMessageSeconds: 604800, reason: 'They deserved it' })
    *   .then(console.log)
    *   .catch(console.error);
    */
--- a/packages/discord.js/typings/index.d.ts
+++ b/packages/discord.js/typings/index.d.ts
@@ -3948,7 +3948,9 @@ export interface AwaitReactionsOptions extends ReactionCollectorOptions {
 }
 
 export interface BanOptions {
+  /** @deprecated */
   deleteMessageDays?: number;
+  deleteMessageSeconds?: number;
   reason?: string;
 }
 


packages/discord.js/src/managers/GuildBanManager.js
```diff
@@ -150,7 +153,10 @@ class GuildBanManager extends CachedManager {
     const id = this.client.users.resolveId(user);
     if (!id) throw new Error(ErrorCodes.BanResolveId, true);
     await this.client.rest.put(Routes.guildBan(this.guild.id, id), {
-      body: { delete_message_days: options.deleteMessageDays },
+      body: {
+        delete_message_days: options.deleteMessageDays,
+        delete_message_seconds: options.deleteMessageSeconds,
+      },
```
Jiralite, line 159: I would rather just transform `delete_message_days` to `delete_message_seconds`. Something like this:

```JavaScript
  const deleteMessageSeconds = typeof options.deleteMessageDays !== 'undefined' ? options.deleteMessageDays * 172800 : options.deleteMessageSeconds;

  await this.client.rest.put(Routes.guildBan(this.guild.id, id), {
    body: { delete_message_seconds: deleteMessageSeconds },
    reason: options.reason,
  });
```

packages/discord.js/src/structures/GuildMember.js
```diff
@@ -346,8 +346,8 @@ class GuildMember extends Base {
    * @param {BanOptions} [options] Options for the ban
    * @returns {Promise<GuildMember>}
    * @example
-   * // ban a guild member
-   * guildMember.ban({ deleteMessageDays: 7, reason: 'They deserved it' })
+   * // ban a guild member, deleting a week of messages
+   * guildMember.ban({ deleteMessageSeconds: 604800, reason: 'They deserved it' })
```
Jiralite, line 350: Just like the [timeout](https://discord.js.org/#/docs/discord.js/main/class/GuildMember?scrollTo=timeout) example:

```suggestion
   * // Ban a guild member, deleting a week's worth of messages
   * guildMember.ban({ deleteMessageSeconds: 60 * 60 * 24 * 7, reason: 'They deserved it' })
```

packages/discord.js/typings/index.d.ts
```diff
@@ -3948,7 +3948,9 @@ export interface AwaitReactionsOptions extends ReactionCollectorOptions {
 }
 
 export interface BanOptions {
+  /** @deprecated */
```
Jiralite, line 3951:
```suggestion
  /** @deprecated Use {@link deleteMessageSeconds} instead. */
```

packages/discord.js/src/managers/GuildBanManager.js
```diff
@@ -129,6 +129,9 @@ class GuildBanManager extends CachedManager {
    * Options used to ban a user from a guild.
    * @typedef {Object} BanOptions
    * @property {number} [deleteMessageDays] Number of days of messages to delete, must be between 0 and 7, inclusive
+   * (deprecated, use deleteMessageSeconds instead)
```
Jiralite, line 132:
```suggestion
   * <warn>This property is deprecated. Use `deleteMessageSeconds` instead.</warn>
```

Jiralite (changes requested):

packages/discord.js/src/managers/GuildBanManager.js
ImRodry: I think you should do 24 * 60 * 60 instead of that magic number

packages/discord.js/typings/index.d.ts
ImRodry: Does that link work?

packages/discord.js/src/managers/GuildBanManager.js
Jiralite: Edited that in. I think that'd look clearer.

packages/discord.js/typings/index.d.ts
Jiralite: Yes.

kyranet (approved):

Jiralite:
Can we get a deprecation warning emitting during runtime too?

packages/discord.js/src/managers/GuildBanManager.js
```diff
@@ -150,7 +153,12 @@ class GuildBanManager extends CachedManager {
     const id = this.client.users.resolveId(user);
     if (!id) throw new Error(ErrorCodes.BanResolveId, true);
     await this.client.rest.put(Routes.guildBan(this.guild.id, id), {
-      body: { delete_message_days: options.deleteMessageDays },
+      body: {
+        delete_message_seconds:
+          typeof options.deleteMessageSeconds !== 'undefined'
+            ? options.deleteMessageSeconds
+            : (options.deleteMessageDays ?? 0) * 24 * 60 * 60,
```
vladfrangu, line 160: Isn't this optional? If it is, we should respect that

vladfrangu (changes requested):

packages/discord.js/src/managers/GuildBanManager.js
CamiKaseM7: I think it's still optional, if you don't pass any argument to the banOptions `options.deleteMessageSeconds !== 'undefined'` will evaluate to false, and `options.deleteMessageDays ?? 0` to 0. Unless I'm missing something.

Jiralite: In the case you described, yes 0 is the result. That means upon providing nothing, `delete_message_seconds` will be set to `0` and sent to Discord. `delete_message_seconds` should be nothing (`undefined`) - not a set value by discord.js.

advaith1: it defaults to 0 so it doesn't matter. anyway we could change it although it doesn't make a difference; the typeof check seems to be useless tho. https://discord.com/channels/222078108977594368/222197033908436994/1015418401503580271

Jiralite: Yes, please change it, hence this requested change in the first place.

SpaceEEC (approved):

vladfrangu (approved):

PR #8345, opened 2022-07-24
fix(GuildAuditLogsEntry): replace OverwriteType with AuditLogOptionsType

thehairy:
- This PR fixes a bug with GuildAuditLogsEntry.extra, since the overwrite type is a string and the wrong enum was used.


thehairy:
since its an issue with discord-api-types this is obsolete and will be fixed there hopefully

Jiralite (approved):

Jiralite (changes requested):
The types need to be updated in `GuildAuditLogsEntryExtraField`:

https://github.com/discordjs/discord.js/blob/af837debe3d8c0ad9b729371d5f0ff4360f3b118/packages/discord.js/typings/index.d.ts#L4618-L4632

It seems the first two events still use the string variant and the third event uses `OverwriteType` instead of `AuditLogOptionsType`.

iCrawl (approved):

SpaceEEC (approved):

PR #8370, opened 2022-07-26
docs(WebhookClient): Make constructor a union

Jiralite:
`WebhookClientData` is now a union of two possible objects rather than one object containing all the properties. This reflects the typings exactly and also introduces `WebhookClientDataIdWithToken` & `WebhookClientDataURL` in the documentation which also reflects the typings exactly.

This also fixes an issue where the return type of the top-level function `parseWebhookURL()` documented the id as a `string` instead of a `Snowflake`.

### DIFF
--- a/packages/discord.js/src/client/WebhookClient.js
+++ b/packages/discord.js/src/client/WebhookClient.js
@@ -12,11 +12,21 @@ const { parseWebhookURL } = require('../util/Util');
  */
 class WebhookClient extends BaseClient {
   /**
-   * The data for the webhook client containing either an id and token or just a URL
-   * @typedef {Object} WebhookClientData
-   * @property {Snowflake} [id] The id of the webhook
-   * @property {string} [token] The token of the webhook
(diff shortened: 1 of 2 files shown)


packages/discord.js/src/client/WebhookClient.js
```diff
@@ -12,11 +12,21 @@ const { parseWebhookURL } = require('../util/Util');
...
+   * Represents the credentials used for a webhook in the form of its id and token.
+   * @typedef {Object} WebhookClientDataIdWithToken
+   * @property {Snowflake} id The webhook's id
+   * @property {string} token The webhook's token
+   */
+
+  /**
+   * Represents the credentials used for a webhook in the form of a URL.
+   * @typedef {Object} WebhookClientDataURL
+   * @property {string} [url] The full URL for the webhook
```
muchnameless, line 24:
```suggestion
   * @property {string} url The full URL for the webhook
```

iCrawl (approved):

PR #8373, opened 2022-07-26
refactor: Deprecate `Formatter` class

Jiralite:
Everything in this class is already exported at the top level. This just leads to code duplication.

- Some `string` types have been updated to `Snowflake` types
- Optional parameters were properly documented
- Documented all the methods as top-level functions. Previously, they weren't documented
- Removed `TimestampStylesString` from the class in the typings. It didn't even exist!
- Deprecated the class (with emitted warnings)


packages/discord.js/src/util/Formatters.js
```diff
@@ -20,180 +21,510 @@ const {
...
+let deprecatedHyperlink = false;
+let deprecatedInlineCode = false;
+let deprecatedItalic = false;
+let deprecatedQuote = false;
+let deprecatedRoleMention = false;
+let deprecatedSpoiler = false;
+let deprecatedStrikethrough = false;
+let deprecatedTime = false;
+let deprecatedUnderscore = false;
+let deprecatedUserMention = false;
```
imranbarbhuiya, line 39: use https://nodejs.org/api/util.html#utildeprecatefn-msg-code instead

imranbarbhuiya (changes requested):

packages/discord.js/src/util/Formatters.js
Jiralite: Should be done with 1ce3eba22752193e09fcbd52b7b944bd5ccf8f6c.

packages/discord.js/src/util/Formatters.js
```diff
@@ -20,180 +21,356 @@ const {
...
   /**
-   * Formats the content into a block quote. This needs to be at the start of the line for Discord to format it.
+   * Formats the content into a block quote.
+   * <info>This needs to be at the start of the line for Discord to format it.</info>
    * @method blockQuote
    * @memberof Formatters
-   * @param {string} content The content to wrap.
+   * @param {string} content The content to wrap
    * @returns {string}
+   * @deprecated Use this method as a top-level function instead.
```
ImRodry, line 168: Only gonna suggest this here but applies to all methods. I believe this should say something along the lines of “Import this method directly from discord.js instead” because using it at the top level is something completely different and you don’t have to do that

Jiralite: Should be done with 080622e58f4aa5391c6b90787e9c80035630b859.

PR #8377, opened 2022-07-27
feat(website): add icons for sections

suneettipirneni:

Adds icons next to each section.

### DIFF
--- a/packages/website/src/components/DocContainer.tsx
+++ b/packages/website/src/components/DocContainer.tsx
@@ -1,4 +1,5 @@
 import type { ReactNode } from 'react';
+import { VscListSelection, VscSymbolParameter } from 'react-icons/vsc';
 import { PrismAsyncLight as SyntaxHighlighter } from 'react-syntax-highlighter';
 import { vscDarkPlus } from 'react-syntax-highlighter/dist/cjs/styles/prism';
 import { CodeListingSeparatorType } from './CodeListing';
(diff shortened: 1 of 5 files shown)


packages/website/src/components/Section.tsx
```diff
@@ -35,16 +46,13 @@ export function Section({ title, children, className, defaultClosed }: SectionPr
 								open: {
 									opacity: 1,
 									height: 'auto',
-									paddingLeft: '1.75rem',
-									paddingRight: '1.75rem',
-									paddingBottom: '2.5rem',
 								},
-								collapsed: { opacity: 0, height: 0, paddingLeft: '1.75rem', paddingRight: '1.75rem', paddingBottom: 0 },
+								collapsed: { opacity: 0, height: 0 },
```
iCrawl, line 50: This was deliberate, because this was it animates nicely even if they are static compared to the static class on the element.

packages/website/src/components/DocContainer.tsx
```diff
@@ -48,11 +49,16 @@ export function DocContainer({ name, kind, excerpt, summary, typeParams, childre
 					</div>
 				) : null}
 				<div className="space-y-10">
-					<Section title="Summary" className="dark:text-white">
-						<p className="text-dark-100 dark:text-gray-300 m-0">{summary ?? 'No summary provided.'}</p>
+					<Section iconElement={<VscListSelection />} title="Summary" className="dark:text-white">
+						<p className="text-dark-100 dark:text-gray-300 m-0 mb-5">{summary ?? 'No summary provided.'}</p>
```
iCrawl, line 53: You already put `mb-5` on the section itself, putting it here again actually has no effect and it just collapses the margins making them only count as a single `mb-5` instead of 2 `mb-5`

PR #8381, opened 2022-07-28
chore: cleanup tsup configs

imranbarbhuiya:
#8324 changed build system so tsup configs are no longer needed.


packages/proxy-container/Dockerfile
```diff
@@ -3,7 +3,7 @@ FROM node:16-alpine
 WORKDIR /usr/proxy
 
 # First copy over dependencies separate from src for better caching
-COPY package.json yarn.lock tsconfig.json .yarnrc.yml tsup.config.ts ./
+COPY package.json yarn.lock tsconfig.json .yarnrc.yml ./
```
iCrawl, line 6: Might as well copy the `build.config.ts` instead here.

PR #8383, opened 2022-07-28
chore: build only actions

imranbarbhuiya:
build only actions when pushing docs changes.


iCrawl:
Don't remove the turbo stuff. We might move back once they fixed their issues.

PR #8391, opened 2022-07-29
chore: fix file path

imranbarbhuiya:
`unbuild` uses `.cjs` extension which is failing docs workflows. But how it was passing earlier?

### DIFF
--- a/packages/actions/src/formatTag/action.yml
+++ b/packages/actions/src/formatTag/action.yml
@@ -11,4 +11,4 @@ outputs:
     description: 'The semver string that was extracted from this tag'
 runs:
   using: node16
-  main: ../../dist/formatTag/index.cjs
+  main: ../../dist/formatTag/index.mjs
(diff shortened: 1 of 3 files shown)


iCrawl:
Looks like something else went wrong here anyway lol

```
Error: File not found: '/home/runner/work/discord.js/discord.js/./packages/actions/src/formatTag/../../dist/formatTag/index.js'
```

packages/actions/src/formatTag/action.yml
```diff
@@ -11,4 +11,4 @@ outputs:
     description: 'The semver string that was extracted from this tag'
 runs:
   using: node16
-  main: ../../dist/formatTag/index.js
+  main: ../../dist/formatTag/index.cjs
```
iCrawl, line 14:
```suggestion
  main: ../../dist/formatTag/index.mjs
```

Can't we use `.mjs` instead here?

imranbarbhuiya: yeah, we can. Btw do we need cjs build for `/actions`? iirc when using tsup, we were using mjs only.

iCrawl: Yeah we don't in this case.

PR #8393, opened 2022-07-29
feat: add `Collection#subtract()`

Syjalo:

This PR adds a new method. The subtract method returns a new structure containing items where the keys and values of original structure are not present in the other.
<details>
  <summary>Visualization</summary>

image

</details>


suneettipirneni:
This should be called `difference` instead of `missing` since it aligns with the mathematical terminology.

packages/collection/src/collection.ts
```diff
@@ -662,6 +662,21 @@ export class Collection<K, V> extends Map<K, V> {
 		return coll;
 	}
 
+	/**
+	 * The missing method returns a new structure containing items where the keys and values of original structure are not present in the other.
+	 *
+	 * @param other - The other Collection to filter against
```
ImRodry, line 668: nice copy and paste
```suggestion
	 * This method returns a collection containing the items that are missing from the other collection in comparison to the current one
	 *
	 * @param other The other Collection to compare against
```

Jiralite:
> This should be called `difference` instead of `missing` since it aligns with the mathematical terminology.

Actually, that method seems to already exist.

Syjalo:
> This should be called `difference` instead of `missing` since it aligns with the mathematical terminology.

The `difference` method is exist and it's coeerct: [see](https://en.wikipedia.org/wiki/Symmetric_difference)
Maybe name it as `complementing`?: [see](https://en.wikipedia.org/wiki/Complement_(set_theory)#Relative_complement)

suneettipirneni:
> > This should be called `difference` instead of `missing` since it aligns with the mathematical terminology.
> 
> The `difference` method is exist and it's correct: [see](https://en.wikipedia.org/wiki/Symmetric_difference) Maybe name it as `complement`?: [see](https://en.wikipedia.org/wiki/Complement_(set_theory)#Relative_complement)

The diagram in your PR description showed a difference operation, which made me think the function was implemented as a difference operation. 

As for the naming *technically* it's a "relative complement" because there's no concept of a universal set in our context. That being said, "complement" would be better because the lack of a universal set is implied.

packages/collection/src/collection.ts
almeidx: The hyphen separating the parameter name and description should stay
Ref: https://tsdoc.org/pages/tags/param/

1Computer1:
The problem here is that the existing `difference` method should be `symmetricDifference`, while this should be `difference`. But `relativeComplement` or similar is also valid; `missing` doesn't mean much. Also possibly valid is `minus` or `subtract`, since that's common notation for the set difference.

packages/collection/__tests__/collection.test.ts
```diff
@@ -418,6 +418,19 @@ describe('intersect() tests', () => {
 	});
 });
 
+describe('missing() tests', () => {
```
almeidx, line 421:
```suggestion
describe('complement() tests', () => {
```

almeidx (changes requested):

packages/collection/src/collection.ts
```diff
@@ -662,6 +662,21 @@ export class Collection<K, V> extends Map<K, V> {
...
 
+	/**
+	 * The complement method returns a new structure containing items where the keys and values of other structure are not present in the original.
+	 *
+	 * @param other - The other Collection to filter against
+	 */
+	public complement<T>(other: ReadonlyCollection<K, T>): Collection<K, T> {
+		const coll = new this.constructor[Symbol.species]<K, T>();
+		for (const [k, v] of other) {
+			if (!(this.has(k) || Object.is(v, this.get(k)))) {
```
almeidx, line 673:
```suggestion
			if (!this.has(k) && !Object.is(v, this.get(k))) {
```

packages/collection/__tests__/collection.test.ts
```diff
@@ -418,6 +418,19 @@ describe('intersect() tests', () => {
 	});
 });
 
+describe('complement() tests', () => {
```
almeidx, line 421: Can you sort this alphabetically? 🙏 (the `describe()`, by function name)

packages/collection/src/collection.ts
Syjalo: Isn't my one faster?

iCrawl: Shouldn't sacrifice readability for speed, except we talk like 10x or something.

packages/collection/src/collection.ts
```diff
@@ -662,6 +662,21 @@ export class Collection<K, V> extends Map<K, V> {
...
+	 *
+	 * @param other - The other Collection to filter against
+	 */
+	public complement<T>(other: ReadonlyCollection<K, T>): Collection<K, T> {
+		const coll = new this.constructor[Symbol.species]<K, T>();
+		for (const [k, v] of other) {
+			if (!this.has(k) && !Object.is(v, this.get(k))) {
+				coll.set(k, v);
+			}
+		}
```
Syjalo, line 676: What should `original.complement(other)` return?
<details>
  <summary>1</summary>

image

</details>
<details>
  <summary>2 (currently)</summary>

image

</details>

1Computer1: I'd prefer it to be `a.minus(b)` and result 1 myself. Simpler that way.

iCrawl: This was minus also sounds nicer and easier to grasp 👀

packages/collection/src/collection.ts
```diff
@@ -662,6 +662,21 @@ export class Collection<K, V> extends Map<K, V> {
 		return coll;
 	}
 
+	/**
+	 * The subtract method returns a new structure containing items where the keys and values of original structure are not present in the other.
```
Jiralite, line 666:
```suggestion
	 * The subtract method returns a new structure containing items where the keys and values of the original structure are not present in the other.
```

packages/collection/src/collection.ts
```diff
@@ -662,6 +662,21 @@ export class Collection<K, V> extends Map<K, V> {
...
 
+	/**
+	 * The subtract method returns a new structure containing items where the keys and values of the original structure are not present in the other.
+	 *
+	 * @param other - The other Collection to filter against
+	 */
+	public subtract<T>(other: ReadonlyCollection<K, T>): Collection<K, V> {
+		const coll = new this.constructor[Symbol.species]<K, V>();
+		for (const [k, v] of this) {
+			if (!other.has(k) && !Object.is(v, other.get(k))) {
```
vladfrangu, line 673: Shouldn't this be an `||` instead? Or alternatively call .get once and check if its undefined or else if it equals to v via object.is?

Syjalo: https://github.com/discordjs/discord.js/pull/8393#discussion_r933528705

kyranet, line 673: Hi, sorry for late review, I have checked this PR quite a few times, but I still can't wrap my head on this condition. Does it even work?

Let's debug this using our brains:

- `other.has(k)` returns `true`
  - It's negated, so `false && expr`, which always returns `false`.
- `other.has(k)` returns `false`
  - It's negated, so `true && expr`, which always returns `expr`.
  - `**other.get(k)` returns `V | undefined`, but because `other.has(k)` is `false`, then `other.get(k)` must always return `undefined`.**
  - That boils down to the constant `Object.is(v, undefined)`.
  - If `v` is always defined:
    - Then `Object.is(v, undefined)` is always `false`, and with the negation, it always returns `true`.
    - **And therefore, for a non-undefined `V`, the condition for substract is for `other` to not have the key.**
  - If `v` can be undefined:
    - **Then `Object.is(v, undefined)` may return `true`, but only if `v` is `undefined`.

I have written the logical flaws in bold for visibility.

---

I believe the right code would be `!other.has(k) || !Object.is(v, other.get(k))`, as `other.get` will be executed only if `other.has(k)` returns `true`.

So yeah, @vladfrangu's comment is right.

kyranet (changes requested):

kyranet (approved):

DraftProducts (approved):

SpaceEEC (approved):

vladfrangu (approved):

packages/collection/src/collection.ts
```diff
@@ -684,6 +684,21 @@ export class Collection<K, V> extends Map<K, V> {
...
+	 * @param other - The other Collection to filter against
+	 */
+	public subtract<T>(other: ReadonlyCollection<K, T>): Collection<K, V> {
+		const coll = new this.constructor[Symbol.species]<K, V>();
+		for (const [k, v] of this) {
+			if (!other.has(k) || !Object.is(v, other.get(k))) {
+				coll.set(k, v);
+			}
+		}
+		return coll;
```
kyranet, line 699: ESLint rules changed on main, this seems to be the fix:
```suggestion
		}

		return coll;
```

PR #8414, opened 2022-08-02
types: add generic to `ActionRowBuilder.from()`

Syjalo:
This PR adds generic type to `ActionRowBuilder.from()`. Currently we can create typed `ActionRowBuilder` using the `.from()` static method only these ways.
```ts
<ActionRowBuilder<ButtonBuilder>>ActionRowBuilder.from();
ActionRowBuilder.from() as ActionRowBuilder<ButtonBuilder>;
```
But these ways are wrong, because we can create an `ActionRowBuilder<ButtonBuilder>`, but type it as `ActionRowBuilder<SelectMenuBuilder>`. This PR adds a new correct way to do that.
```ts
ActionRowBuilder.from<ButtonBuilder>();
```
Now the return type is typed and you can't add other types of action rows to the `.from()` method.


Jiralite (commented):
Can you add tests?

packages/discord.js/typings/index.test-d.ts
```diff
@@ -1683,6 +1687,18 @@ EmbedBuilder.from(embedData);
...
+ActionRowBuilder.from(actionRowData);
+
+declare const actionRowComp: ActionRow<ActionRowComponent>;
+ActionRowBuilder.from(actionRowComp);
+
+declare const buttonsActionRowData: APIActionRowComponent<APIButtonComponent>;
+declare const buttonsActionRowComp: ActionRow<ButtonComponent>;
+
+expectType<ActionRowBuilder<ButtonBuilder>>(ActionRowBuilder.from<ButtonBuilder>(buttonsActionRowData));
+expectType<ActionRowBuilder<ButtonBuilder>>(ActionRowBuilder.from<ButtonBuilder>(buttonsActionRowComp));
```
imranbarbhuiya, line 1700: it should auto infer the type from the parameter. So can u add a new test for auto infer pls?

imranbarbhuiya (changes requested):

packages/discord.js/typings/index.test-d.ts
Syjalo: Didn't get what should I do

thehackerboi69github, line 1700:
```suggestion
expectType<ActionRowBuilder<ButtonBuilder>>(ActionRowBuilder.from(buttonsActionRowData));
expectType<ActionRowBuilder<ButtonBuilder>>(ActionRowBuilder.from(buttonsActionRowComp));
```

thehackerboi69github (changes requested):

packages/discord.js/typings/index.test-d.ts
Syjalo: In the tests generic is needed because we need to know that we have exactly passed an action row with buttons

vladfrangu: Add an identical test without the `<ButtonBuilder>`

Jiralite (changes requested):
This needs a rebase + [tests](https://github.com/discordjs/discord.js/pull/8414#discussion_r966908380).

kyranet (approved):

iCrawl (approved):

SpaceEEC (approved):

PR #8417, opened 2022-08-03
feat: @discordjs/structures

ckohen:

Introduces a generic structure implementation for the main lib (and others) to build on.

⚠️ This is currently in a very early proof of concept stage published as a PR to formulate ideas and solidify the intended structure (ha) of the module.

TODO:
- [ ] Application
- [ ] Audit Log
- [ ] Auto Moderation
- [ ] Channel
  - [x] Channels
  - [ ] Mixins
    - [ ] DM
    - [ ] Guild
    - [ ] Text
    - [ ] Permissions
    - [ ] Webhooks
    - [ ] Thread
    - [ ] ThreadOnly
    - [ ] Voice
  - [ ] Message
- [ ] Emoji
- [ ] Guild
  - [ ] Guild
  - [ ] Guild Preview
  - [ ] Guild Widget
  - [ ] Guild Member
  - [ ] Integration
  - [ ] Guild Ban
  - [ ] ~~Membership Screening~~ Oh wait...this is *still* undocumented
- [ ] Guild Scheduled Event
- [ ] Guild Template
- [ ] Interactions (Need more thoughts on this)
- [x] Invite
- [ ] Stage Instance
- [ ] Sticker
  - [ ] Sticker
  - [ ] Sticker Pack
- [ ] Teams
- [x] User
  - [x] User
  - [x] Connection
- [ ] Voice
  - [ ] Voice State
  - [ ] Voice Region
- [ ] Webhook

### DIFF
--- a/packages/structures/src/channels/Channel.ts
+++ b/packages/structures/src/channels/Channel.ts
@@ -1,6 +1,5 @@
 import { DiscordSnowflake } from '@sapphire/snowflake';
 import type { APIChannel, APIPartialChannel, ChannelType } from 'discord-api-types/v10';
-import type { StructureExtraOptions } from '../Structure.js';
 import { Structure } from '../Structure.js';
 import { kData } from '../utils/symbols.js';
 
@@ -13,6 +12,14 @@ export type ChannelDataType<Type extends ChannelType | 'unknown'> = Type extends
 	? APIChannel
 	: Extract<APIChannel, { type: Type }>;
 
+/**
+ * Represents any channel on Discord.
+ *
+ * @typeParam Type - Specify the type of the channel being constructed for more accurate data types
+ * @typeParam Omitted - Specify the propeties that will not be stored in the raw data field as a union, implement via `DataTemplate`
+ * @remarks While this class can be directly constructed for any channel type, this class is not intended to be used directly,
+ * preferring instead to create class that extend this class with the appropriate mixins for each channel type.
+ */
 export class Channel<
 	Type extends ChannelType | 'unknown' = 'unknown',
 	Omitted extends keyof ChannelDataType<Type> | '' = '',
@@ -20,26 +27,25 @@ export class Channel<
 	/**
 	 * The template used for removing data from the raw data stored for each Channel.
 	 *
-	 * @remarks This template will only apply to channels constructed
-	 * Use the appropriate subclass template to remove data from that channel type
+	 * @remarks This template is only guaranteed to apply to channels constructed directly via `new Channel()`.
+	 * Use the appropriate subclass template to remove data from that channel type.
 	 */
-	public static DataTemplate: Partial<APIChannel> = {};
+	public static override DataTemplate: Partial<APIChannel> = {};
 
 	public constructor(
 		/**
 		 * The raw data received from the API for the channel
 		 */
 		data: Omit<ChannelDataType<Type>, Omitted>,
-		extraOptions?: StructureExtraOptions,
 	) {
-		super(data as ChannelDataType<Type>, extraOptions ?? { template: Channel.DataTemplate });
+		super(data as ChannelDataType<Type>);
 	}
 
 	/**
 	 * {@inheritDoc Structure._patch}
 	 */
-	public override _patch(data: Partial<ChannelDataType<Type>>, extraOptions: StructureExtraOptions) {
-		return super._patch(data, extraOptions ?? { template: Channel.DataTemplate });
+	public override _patch(data: Partial<ChannelDataType<Type>>) {
+		return super._patch(data);
 	}
 
 	/**
@@ -104,7 +110,7 @@ export class Channel<
 	/**
 	 * Indiciates whether this channel is in a guild
 	 *
-	 * @privateRemarks OVerriden to `true` on `GuildChannelMixin`
+	 * @privateRemarks Overriden to `true` on `GuildChannelMixin`
 	 */
 	public isGuildBased() {
 		return false;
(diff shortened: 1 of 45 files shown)


packages/structures/src/users/User.ts
```diff
@@ -0,0 +1,193 @@
...
+
+	public override patch(data: Partial<APIUser>) {
+		return super.patch(data, { template: UserDataTemplate });
+	}
+
+	/**
+	 * The user's id
+	 */
+	public get id() {
+		return this[kData].id as 'id' extends Omitted ? never : APIUser['id'];
```
vladfrangu, line 48: If we'll have to add this return type for each getter
1. let's add it as the return type, if we can, else keep it as a cast
2. Let's make a utility type for this (`NeverIfOmitted<Omitted, 'id', APIUser['id']>`)

packages/structures/src/users/User.ts
```diff
@@ -0,0 +1,193 @@
...
+	 */
+	public get avatar() {
+		return this[kData].avatar as 'avatar' extends Omitted ? never : APIUser['avatar'];
+	}
+
+	/**
+	 * Whether the user is a bot
+	 */
+	public get bot() {
+		return (this[kData].bot ?? false) as 'bot' extends Omitted ? never : APIUser['bot'];
```
vladfrangu, line 83: If we have defaults, we'll never have this return `never`, even if omitted by users

vladfrangu (changes requested):

packages/structures/src/users/User.ts
ckohen: hmm, technically we should return undefined in those cases, but there's no way for us to know that. ngl, I kind of like lying about the type since it isn't guaranteed to be accurate

packages/structures/src/users/User.ts
ckohen: Definitely utility type. I wish it didn't have to cast, but [value of thing] is not assignable to never (or undefined). Maybe we can add `| never` to all the types in `Structure[kData]` though and then it works? I'll try that, cause it'd make it a lot easier if it were a merged type.

ckohen: Unfortunately not no. One, it's not as nice of an interface for consumers. Two, this doesn't allow us to convert from snake_case to camelCase.

I would love to do it this way, since it's a lot easier to maintain for us, but I think prioritizing the consumer interface is more important here. These base structures are a lot to implement at first, but shouldn't need to be touched super often once implemented.

packages/structures/src/users/User.ts
```diff
@@ -0,0 +1,177 @@
...
+	public get createdAt() {
+		return this.createdTimestamp ? new Date(this.createdTimestamp) : null;
+	}
+
+	/**
+	 * The hexadecimal version of the user accent color, with a leading hash
+	 * <info>This property is only set when the user was manually fetched</info>
+	 */
+	public get hexAccentColor() {
+		if (typeof this.accentColor !== 'number') return this.accentColor;
```
Syjalo, line 174: The type is incorrect. It should be casted or something
```suggestion
		if (typeof this.accentColor !== 'number') return this.accentColor as 'accent_color' extends Omitted ? undefined : null | undefined;
```

Syjalo (changes requested):

Shibukub234:
https://m.me/ch/AbZrqh1Ot_dQr-BZ/

รับ Outlook สำหรับ Android<https://aka.ms/AAb9ysg>
________________________________

packages/structures/src/invites/Invite.ts
```diff
@@ -0,0 +1,177 @@
...
+			(this[kCreatedTimestamp] && this.maxAge ? this[kCreatedTimestamp] + (this.maxAge as number) * 1_000 : null)
+		);
+	}
+
+	/**
+	 * The time the invite will expire at
+	 */
+	public get expiresAt() {
+		const expiresTimestamp = this.expiresTimestamp;
+		return expiresTimestamp && new Date(expiresTimestamp);
```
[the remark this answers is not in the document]
ckohen: (editing your comment to keep the cache of expiresTimestamp)

packages/structures/src/invites/Invite.ts
```diff
@@ -0,0 +1,177 @@
...
+	 */
+	public get createdTimestamp() {
+		return this[kCreatedTimestamp];
+	}
+
+	/**
+	 * The time the invite was created at
+	 */
+	public get createdAt() {
+		return this[kCreatedTimestamp] && new Date(this[kCreatedTimestamp]);
```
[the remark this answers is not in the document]
ckohen: I did do this intentionally, though I'm interested in hearing why you might want to access the getter here.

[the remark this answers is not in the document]
Syjalo: (editing your comment to keep the cache of createdTimestamp)

packages/structures/src/channels/Channel.ts
```diff
@@ -0,0 +1,157 @@
...
+	public isThread() {
+		return false;
+	}
+
+	/**
+	 * Indicates whether this channel can contain messages
+	 *
+	 * @privateRemarks Overriden to `true` on `TextChannelMixin`
+	 */
+	public iSTextBased() {
```
Syjalo, line 100:
```suggestion
	public isTextBased() {
```

Syjalo (changes requested):

ckohen:
Superseded by #10900

PR #8430, opened 2022-08-04
fix(MessageMentions): infinite loop in `parsedUsers` getter

### LINKED ISSUE #8426: Infinite loop on MessageMentions `parsedUsers` getter

discord.js

### Issue description

In the getter for the `MessageMentions#parsedUsers` when the `this._parsedUsers` is a falsy value what prevents this code from creating an infinite loop?

Since if the content has mentions the `this.constructor.UsersPattern.exec(this._content)` will always be truthy and `UsersPattern` isn't a global Regex so no `lastIndex`.

> **Warning**
> May cause memory leaks or stack overflow!

### Code sample

```typescript
/**
   * Any user mentions that were included in the message content
   * <info>Order as they appear first in the message content</info>
   * @type {Collection<Snowflake, User>}
   * @readonly
   */
  get parsedUsers() {
    if (this._parsedUsers) return this._parsedUsers;
    this._parsedUsers = new Collection();
    let matches;
    while ((matches = this.constructor.UsersPattern.exec(this._content)) !== null) {
      const user = this.client.users.cache.get(matches[1]);
      if (user) this._parsedUsers.set(user.id, user);
    }
    return this._parsedUsers;
  }

// https://github.com/discordjs/discord.js/blob/359f6885558fcfb3151971ab589077a89ee71a01/packages/discord.js/src/structures/MessageMentions.js#L224
```

### Package version

14.1.2

### Node.js version

Node v17.9.1 and Ts v4.7.4

### Operating system

Windows

Medium (should be fixed soon)

Not applicable (internal bug)

Not applicable (internal bug)

almeidx:
Fixes #8426 

### DIFF
--- a/packages/discord.js/typings/index.d.ts
+++ b/packages/discord.js/typings/index.d.ts
@@ -1880,12 +1880,12 @@ export class MessageMentions {
   public crosspostedChannels: Collection<Snowflake, CrosspostedChannel>;
   public toJSON(): unknown;
 
-  public static ChannelsPattern: typeof FormattingPatterns.Channel;
   private static GlobalChannelsPattern: RegExp;
+  private static GlobalUsersPattern: RegExp;
+  public static ChannelsPattern: typeof FormattingPatterns.Channel;
   public static EveryonePattern: RegExp;
   public static RolesPattern: typeof FormattingPatterns.Role;
   public static UsersPattern: typeof FormattingPatterns.User;
-  private static GlobalUsersPattern: RegExp;
 }
 
 export class MessagePayload {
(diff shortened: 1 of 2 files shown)


iCrawl (approved):

packages/discord.js/typings/index.d.ts
```diff
@@ -1885,6 +1885,7 @@ export class MessageMentions {
   public static EveryonePattern: RegExp;
   public static RolesPattern: typeof FormattingPatterns.Role;
   public static UsersPattern: typeof FormattingPatterns.User;
+  private static GlobalUsersPattern: RegExp;
```
SpaceEEC, line 1888: Mind moving this (and the `GlobalChannelsPattern`) above the `ChannelsPattern`?

SpaceEEC (changes requested):

vladfrangu (approved):

iCrawl (approved):

PR #8439, opened 2022-08-05
feat: split `CommandInteractionOptionResolver`

Syjalo:
This PR splits `CommandInteractionOptionResolver` to make methods visible in the docs only what are available in the context.

This is technically breaking change, but this doesn't change the library's interface, only removes what aren't available in the context and can't be used.


packages/discord.js/src/structures/ChatInputCommandInteractionOptionResolver.js
```diff
@@ -1,21 +1,16 @@
 'use strict';
 
 const { ApplicationCommandOptionType } = require('discord-api-types/v10');
-const { TypeError, ErrorCodes } = require('../errors');
+const InteractionOptionResolver = require('./InteractionOptionResolver');
+const { ErrorCodes } = require('../errors');
```
vladfrangu, line 5: You should still import the TypeError here!

vladfrangu (changes requested):

monbrey:
I don't see how this split is correct?

Why are the methods for subcommands, subcommand groups and the primitive types (string, number, integer, boolean) removed from the AutocompleteInteractionOptionResolver? They should also still work, as the value doesn't require any resolved structures.

I thought the issue was only that Discord doesn't send resolved data for Users/GuildMembers, Roles and Channels in AutocompleteInteraction.

packages/discord.js/src/structures/ChatInputCommandInteractionOptionResolver.js
Rebazkoye: Hi

Rebazkoye:
Sweden

packages/discord.js/src/structures/AutocompleteInteraction.js
```diff
@@ -53,7 +53,7 @@ class AutocompleteInteraction extends BaseInteraction {
      * The options passed to the command
      * @type {CommandInteractionOptionResolver}
```
MrMythicalYT, line 54:
```suggestion
     * @type {AutocompleteInteractionOptionResolver}
```

MrMythicalYT (changes requested):

Qjuh:
No idea if this PR is still meant to be implemented.
- If it is then I got the question why `MessageContextMenuInteraction` and `UserContextMenuInteraction` would still share the same `CommandInteractionOptionResolver` having both methods for `getUser()` and `getMessage()`, as the split seems to only be done partially.
- If it isn't then maybe this PR should be closed

PR #8446, opened 2022-08-06
Update Util.js

Alexandre887:

[One of the functions](https://github.com/discordjs/discord.js/blob/main/packages/discord.js/src/util/Util.js#L276-L284) in [this](https://github.com/discordjs/discord.js/blob/main/packages/discord.js/src/util/Util.js) file uses a non-existent object method `.hasOwn()`.


Jiralite:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/hasOwn

PR #8447, opened 2022-08-06
fix for "TypeError: Object.hasOwn is not a function"

Endxca:
change a function name (hasOwn to hasOwnProperty) in discord.js/src/util/Util.js , ligne 279 , col 16
to patch the error ```TypeError: Object.hasOwn is not a function```


SuperchupuDev:
Update to Node 16.9 or higher

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/hasOwn

PR #8454, opened 2022-08-09
docs: change registration example to use global commands

advaith1:

Global commands are more familiar to users and make more sense in an example with little context. With commands v2, global commands deploy instantly so there is no point in using guild commands.

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,7 +72,7 @@ const rest = new REST({ version: '10' }).setToken('token');
 	try {
 		console.log('Started refreshing application (/) commands.');
 
-		await rest.put(Routes.applicationGuildCommands(CLIENT_ID, GUILD_ID), { body: commands });
+		await rest.put(Routes.applicationCommands(CLIENT_ID), { body: commands });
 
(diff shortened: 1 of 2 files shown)


almostSouji:
I disagree that there is no point in guild commands. First-time developers should build the habit of deploying commands to a test server before releasing them into the wild, testing them to make sure they do what they expect them to.

That being said, I could see that covered as a note, or maybe even with both, annotated, in the same example.
A ping-pong demonstration of a proof of concept to be a global "finished" command seems slightly absurd to me.

vladfrangu:
I agree with @almostSouji, however that sort of content feels more appropriate for our guides (educating on the difference, and teaching users to start off with guild commands and then migrate to global ones). Personally I think having the ping pong command example as global is *fine*, but it should definitely tell the users to read more on our guides for more examples

iCrawl (approved):

vladfrangu (approved):

kyranet (approved):

PR #8443, opened 2022-08-05
types: fix missing types for mentionable options

janparisek:
**Please describe the changes this PR makes and why it should be merged:**
- Add `ApplicationCommandRoleOptionData` interface
- Add `ApplicationCommandRoleOption` interface
- Add `ApplicationCommandUserOptionData` interface
- Add `ApplicationCommandUserOption` interface
- Add `ApplicationCommandMentionableOptionData` interface
- Add `ApplicationCommandMentionableOption` interface
- Update `ApplicationCommandSubCommandData.options` union
- Update `ApplicationCommandOptionData` union
- Update `ApplicationCommandOption` union

Fixes #8440 

**Status and versioning classification:**

- Code changes have been tested against the Discord API, or there are no code changes
- I know how to update typings and have done so, or typings don't need updating
- This PR changes the library's interface (methods or parameters added)

PR #8456, opened 2022-08-09
types: change type of `ApplicationCommandSubCommand`.options

ZyXProFR:

This PR changes the type of the `ApplicationCommandSubCommand.options` field to `ApplicationCommandOption[]` in order to match the documentation and be able to use all the properties of this type (like `minValue`, `maxValue`, etc).

### DIFF
--- a/packages/discord.js/typings/index.d.ts
+++ b/packages/discord.js/typings/index.d.ts
@@ -3871,8 +3871,15 @@ export interface ApplicationCommandSubCommandData extends Omit<BaseApplicationCo
 
 export interface ApplicationCommandSubCommand extends Omit<BaseApplicationCommandOptionsData, 'required'> {
   type: ApplicationCommandOptionType.Subcommand;
-  options?: ApplicationCommandOption[];
-}
+  options?: (  
+    | ApplicationCommandNonOptions
+    | ApplicationCommandChannelOption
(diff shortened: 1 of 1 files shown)


Jiralite (changes requested):
This will introduce the issue where the type of the options will allow a subcommand group or a subcommand as a valid type. The current code does not allow this to happen as it is strongly typed as to what can be allowed. This needs to be modified - perhaps #8443 will help with this.

ZyXProFR:
> This will introduce the issue where the type of the options will allow a subcommand group or a subcommand as a valid type. The current code does not allow this to happen as it is strongly typed as to what can be allowed. This needs to be modified - perhaps #8443 will help with this.

Hmm can you confirm that what you meant was my last commit?

Jiralite:
It's better, but it's missing other option types now (which was always the case). #8443 partially fixes this, so either you can suggest the change there or you will have to wait until that pull request is merged, fetch the changes and add the missing types.

ZyXProFR:
> It's better, but it's missing other option types now (which was always the case). #8443 partially fixes this, so either you can suggest the change there or you will have to wait until that pull request is merged, fetch the changes and add the missing types.

Okay thanks :)

ZyXProFR:
Waiting for the merge of another pull request

PR #8457, opened 2022-08-09
refactor(codeBlock): swap order of arguments

ImRodry:

This PR swaps the order of the arguments in the codeBlock function. It also adds a deprecation warning if arguments are passed the old way and attempts to swap them. This isn't 100% bulletproof as there is no official list of languages so I think it's the best we can have. This was discussed in Discord too

### DIFF
--- a/packages/builders/src/messages/formatters.ts
+++ b/packages/builders/src/messages/formatters.ts
@@ -17,7 +17,9 @@ export function codeBlock<C extends string>(content: C): `\`\`\`\n${C}\n\`\`\``;
  */
 export function codeBlock<C extends string, L extends string>(content: C, language: L): `\`\`\`${L}\n${C}\n\`\`\``;
 export function codeBlock(content: string, language?: string): string {
-	if (language && language.length > 18) {
+	// unicorn-rails-log is the longest named language known to be supported at the time of this commit
+	// That's why we use 17 here, because it's the length of that language
+	if (language && language.length > 17) {
 		process.emitWarning(
 			'Passing the language as the first parameter in the codeBlock method is deprecated. Please pass it as the second parameter instead.',
 			'DeprecationWarning',
--- a/packages/builders/src/messages/formatters.ts
+++ b/packages/builders/src/messages/formatters.ts
@@ -1,3 +1,4 @@
+import process from 'node:process';
 import type { URL } from 'url';
 import type { Snowflake } from 'discord-api-types/globals';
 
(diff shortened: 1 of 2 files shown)


packages/builders/src/messages/formatters.ts
```diff
@@ -11,12 +12,21 @@ export function codeBlock<C extends string>(content: C): `\`\`\`\n${C}\n\`\`\``;
...
- * @param language - The language for the codeblock
  * @param content - The content to wrap
+ * @param language - The language for the codeblock
  */
-export function codeBlock<L extends string, C extends string>(language: L, content: C): `\`\`\`${L}\n${C}\n\`\`\``;
-export function codeBlock(language: string, content?: string): string {
-	return typeof content === 'undefined' ? `\`\`\`\n${language}\n\`\`\`` : `\`\`\`${language}\n${content}\n\`\`\``;
+export function codeBlock<C extends string, L extends string>(content: C, language: L): `\`\`\`${L}\n${C}\n\`\`\``;
+export function codeBlock(content: string, language?: string): string {
+	if (language && language.length > 18) {
```
iCrawl, line 20: Whats this `18` based on again?
I imagine its the longest possible language string but can we put a comment there or something? So we don't get confused at one point.

ImRodry: It's based on this message: https://canary.discord.com/channels/222078108977594368/1001614734203834499/1003131709803483156
I'll add that comment

vladfrangu:
Won't this break big time if you run `codeBlock('js', 'hi')` or something chaotic like ``codeBlock(`user-${user.id}-hidden-data-owo`, 'hey! Hidden data ahead!')``? You can't really handle deprecation warnings without having a list of all possible codeblock languages (which can actually be anything the user wants even if it might not do anything in the client)...

ImRodry:
The only reason I didn’t add the list of all languages is because that can change at any time and there’s no official source. Yes it will break in those scenarios but those sound stupid to begin with so they might as well. Why would you put something that isn’t a language in the language parameter? It doesn’t make sense. I know this solution isn’t perfect but it’s the best we’ve got if you don’t wanna add a list with potentially unreliable data

vladfrangu:
> Yes it will break in those scenarios but those sound stupid to begin with so they might as well. Why would you put something that isn’t a language in the language parameter? It doesn’t make sense

What you might consider stupid someone else could consider ingenious to store extra data in the message 😛 

---

That said, leaving the arguments of language aside, I personally think the code should stay the same! `codeBlock('hewwo')` works perfectly fine, and `codeBlock('js', '1')` works too (and is closer to what you'd write in the client. But I'll leave the rest of the @discordjs/the-big-4 to give their thoughts too! Thats just my two cents :D

JPBM135:
I really think that an type with the some of the supported languages should be added, there is so many languages that no one knows of and that does not necessarily means that it is all of them

ImRodry:
> > Yes it will break in those scenarios but those sound stupid to begin with so they might as well. Why would you put something that isn’t a language in the language parameter? It doesn’t make sense
> 
> What you might consider stupid someone else could consider ingenious to store extra data in the message 😛
> 
> That said, leaving the arguments of language aside, I personally think the code should stay the same! `codeBlock('hewwo')` works perfectly fine, and `codeBlock('js', '1')` works too (and is closer to what you'd write in the client. But I'll leave the rest of the @discordjs/the-big-4 to give their thoughts too! Thats just my two cents :D

like I said at the start of the thread on Discord, I understand why it was originally designed this way but it doesn't make sense to keep this since the language parameter is optional and it's counter-intuitive to have an optional parameter come before a required one. Once again, I realize this approach is not perfect, but it's the best that can be done without maintaining a list of all languages, which wouldn't be a good idea imo

DTrombett:
We should just stop lying... It's impossible to predict which one is the language and which one is the text. For example if I use `codeBlock("js", "py")` you'll never know if here the language is JS or PY.
If we want to go ahead with this change we must do it in a major version with a breaking change

kyranet:
Honestly for me, this is a big no. This is either to break or to leave as-is.

Dynamic argument ordering based on length is going to backfire easily, besides, types are now wrong.

`codeBlock('js', '240875059953139714')` is typed as `῾῾῾240875059953139714\njs\n῾῾῾`, however, the resulting type is `῾῾῾js\n240875059953139714\n῾῾῾`.

Now, the reason why I went for `codeBlock(language, content)`, it's because it's natural, when you make a codeblock, you write the language and then the content, all other formatters follow this ordering pattern for a reason.

vladfrangu:
I mean this back and forth is not productive to anyone really 😅 

If you really want to fix it, use an object for the properties. Can't confuse them then :D
Otherwise yeah..just leave as is..

ImRodry:
Yeah I’m definitely not doing an object for this PR, if anyone else wants to take it feel free to

PR #8464, opened 2022-08-10
fix(MessageDelete): when someone delete an embed message or empty message it returns empty value and it makes an error

dager-mohamed:

### DIFF
--- a/packages/discord.js/src/client/actions/MessageDelete.js
+++ b/packages/discord.js/src/client/actions/MessageDelete.js
@@ -12,7 +12,7 @@ class MessageDeleteAction extends Action {
       if (!channel.isTextBased()) return {};
 
       message = this.getMessage(data, channel);
-      if (message) {
+      if ((message && message.embeds.length > 0) || message.content.length > 0) {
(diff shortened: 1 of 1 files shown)


almeidx:
How can a string/array have a negative length?

kyranet:
Please provide a reproducible sample in which this would happen, I think the only way for this to happen is for a cosmic ray to flip up to 63 bits in a row 😅 

As aforementioned, arrays and strings cannot have negative length.

dager-mohamed:
> How can a string/array have a negative length?

no because if someone deleted an embed message it returns empty content or someone  deleted an empty message

kyranet:
Empty arrays and strings have a length of 0, not -1.

dager-mohamed:
> Empty arrays and strings have a length of 0, not -1.

mhmm ok so I should make if it is equal zero right?

kyranet:
No, you have yet to provide the error from discord.js (if any).

Otherwise, if the error is in the user's side (aka yours, and I think it most likely is), then this PR is invalid.

PR #8465, opened 2022-08-10
types(Message): correct `bulkDelete` return type

### LINKED ISSUE #8410: incorrect bulkDelete return type

discord.js

### Issue description

1. Bulk delete messages via ID that are not cached by the client
2. Observe `Snowflake => undefined` entries

### Code sample

```typescript
const deleted = await channel.bulkDelete([m1, m2, m3, m4], true);
console.log(deleted);
```

### Package version

14.1.1

### Node.js version

v18.7.0

### Operating system

_No response_

Medium (should be fixed soon)

_No response_

RedGuy12:

Resolves #8410 (Partially)
This implements the patch fix detailed in #8410. I updated both the typings and the docs.

### DIFF
--- a/packages/discord.js/typings/index.d.ts
+++ b/packages/discord.js/typings/index.d.ts
@@ -3658,7 +3658,7 @@ export interface TextBasedChannelFields extends PartialTextBasedChannelFields {
   bulkDelete(
     messages: Collection<Snowflake, Message> | readonly MessageResolvable[] | number,
     filterOld?: boolean,
-  ): Promise<Collection<Snowflake, Message | undefined>>;
+  ): Promise<Collection<Snowflake, Message | PartialMessage | undefined>>;
   createMessageComponentCollector<T extends MessageComponentType>(
     options?: MessageChannelCollectorOptionsParams<T, true>,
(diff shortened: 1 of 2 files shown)


vladfrangu (approved):

vladfrangu:
Actually this is kinda redundant considering Collection#get already returns T | undefined (in this case Message | undefined) 👀

iCrawl (approved):

kyranet (approved):

packages/discord.js/typings/index.d.ts
```diff
@@ -3658,7 +3658,7 @@ export interface TextBasedChannelFields extends PartialTextBasedChannelFields {
   bulkDelete(
     messages: Collection<Snowflake, Message> | readonly MessageResolvable[] | number,
     filterOld?: boolean,
-  ): Promise<Collection<Snowflake, Message>>;
+  ): Promise<Collection<Snowflake, Message | undefined>>;
```
Jiralite, line 3661: `PartialMessage` should be added here.

If we consider the case that a array of uncached message ids are passed to this method but we also specify the message partial, we will get back a `Message` structure but it will be partial.

PR #8479, opened 2022-08-12
fix(WebSocketShard#destroy): wait for close and cleanup listeners

didinele:
This is yet another attempt at fixing the exact same problem as #8399 and #8405


packages/ws/src/ws/WebSocketShard.ts
```diff
@@ -186,7 +186,18 @@ export class WebSocketShard extends AsyncEventEmitter<WebSocketShardEventsMap> {
 			this.connection &&
 			(this.connection.readyState === WebSocket.OPEN || this.connection.readyState === WebSocket.CONNECTING)
 		) {
+			// No longer need to listen to messages
+			this.connection.removeAllListeners('message');
```
vladfrangu, line 190: Use .off

vladfrangu (changes requested):

packages/ws/src/ws/WebSocketShard.ts
didinele: Is there a good reason for this?

vladfrangu: Consistency

didinele: Hm, I wonder if I have to do `.off('message', this.onMessage`) or `.off('message', this.onMessage.bind(this))` to match how I originally set up the listener. I'll mess with it

SpaceEEC (approved):

iCrawl (approved):

PR #8487, opened 2022-08-14
refactor: docs design

iCrawl:


packages/website/package.json
```diff
@@ -47,6 +47,13 @@
 	},
 	"homepage": "https://discord.js.org",
 	"dependencies": {
+		"@emotion/react": "^11.10.0",
+		"@emotion/server": "^11.10.0",
+		"@mantine/core": "^5.1.5",
+		"@mantine/hooks": "^5.1.5",
+		"@mantine/next": "^5.1.5",
```
imranbarbhuiya, line 54: Have you tried https://www.npmjs.com/package/nextjs-progressbar, it doesn't need any extra configurations. You just need to add `<NextNProgress />` in `_app.ts`.

imranbarbhuiya: I selected the wrong line number. This suggestion was for `@mantine/nprogress`

iCrawl: I'd rather stay in the same UI framework.

packages/website/src/components/SidebarItems.tsx
```diff
@@ -81,42 +65,54 @@ function resolveIcon(item: keyof GroupedMembers) {
...
-										>
-											{member.name}
-										</a>
-									</Link>
-									<div>
+				.map((group, idx) => (
+					<Section key={idx} title={group} icon={resolveIcon(group)}>
+						{groupItems[group].map((member, i) => (
+							<Link key={i} href={member.path} passHref>
+								<UnstyledButton className={classes.link} component="a" onClick={() => setOpened((o) => !o)}>
```
imranbarbhuiya, line 104: different color for the selected member is removed. Is it intentional?

imranbarbhuiya (changes requested):

packages/website/src/components/SidebarItems.tsx
iCrawl: For now, yes.

PR #8490, opened 2022-08-14
feat(Emoji): support more extensions in `url()` function

dager-mohamed:

### DIFF
--- a/packages/discord.js/src/structures/Emoji.js
+++ b/packages/discord.js/src/structures/Emoji.js
@@ -53,7 +53,7 @@ class Emoji extends Base {
    * @readonly
    */
   get url() {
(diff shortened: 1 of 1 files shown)


monbrey:
This is just incorrect. The line of code you modified is a ternary, providing .gif if animated and .png if not. 

It's not a list of options which to be selected from.

dager-mohamed:
so... , i understand only 50% of you said, anyway the problem is from my javascript code or all idea is wrong or what :rofl:

monbrey:
I understand the idea is to support additional formats, because static emojis aren't always PNG.

The code you wrote does not do that, at all. It will continue to always pass .PNG as it's the first one and it's a truthy value. Nothing you've written allows the user to select a format.

dager-mohamed:
> 

ahaa ok , i will delete this PR and when I finish I will do another PR

PR #8492, opened 2022-08-14
Add APIRole as a valid RoleResolvable type

CamiKaseM7:

`interface.options.getRole` return type is `Role | APIRole | Null`.  `APIRole` isn't include in `RoleResolvable` (which is defined as `Role |  Snowflake`) so when trying to implement a function like this:
```ts
async execute(interaction: CommandInteraction) {
        await interaction.deferReply({ ephemeral: true });

        const mutedRole = interaction.options.getRole("muted")!;

        const botHighestRole = interaction.guild!.me!.roles.highest;
        const canBotManageRole =
                botHighestRole.comparePositionTo(mutedRole) > 0;
        // ...
}
```

you get this type error in typescript:

```
Argument of type 'Role | APIRole' is not assignable to parameter of type 'RoleResolvable'.
  Type 'APIRole' is not assignable to type 'RoleResolvable'.
    Type 'APIRole' is missing the following properties from type 'Role': createdAt, createdTimestamp, deleted, editable, and 21 more.
```

The missing properties of `APIRole` aren't needed to resolve the role, what you only need is the id/Snowflake of the role which is present. There shouldn't be an error here so we should accept `APIRole` as a valid `RoleResolvable` type.

### DIFF
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -5774,7 +5774,7 @@ export interface RolePosition {
   position: number;
 }
 
-export type RoleResolvable = Role | Snowflake;
+export type RoleResolvable = Role | Snowflake | APIRole;
 
 export interface RoleTagData {
   botId?: Snowflake;


Jiralite:
> The missing properties of `APIRole` aren't needed to resolve the role, what you only need is the id/Snowflake of the role which is present. There shouldn't be an error here so we should accept `APIRole` as a valid `RoleResolvable` type.

`Role#comparePositionTo()` internally calls `RoleManager#comparePositions()`:

https://github.com/discordjs/discord.js/blob/30e89a401ddc095b3689e861c6104d49fb47c2d1/src/structures/Role.js#L223-L231

Here is `RoleManager#comparePositions()`:

https://github.com/discordjs/discord.js/blob/30e89a401ddc095b3689e861c6104d49fb47c2d1/src/managers/RoleManager.js#L290-L307

For your example, you would be passing an `APIRole` as `role2` here. discord.js would attempt to resolve a discord.js structure from this. How it does that is from the `DataManager`:

https://github.com/discordjs/discord.js/blob/30e89a401ddc095b3689e861c6104d49fb47c2d1/src/managers/DataManager.js#L34-L43

Since a `RoleManager` does not contain `APIRole`s and is definitely not a string, this will return `null`, which will throw a `TypeError`.

In other words, this is not correct. At least, not in its current state.

PR #8493, opened 2022-08-15
feat(Emoji): create `getURL()` function to get more extension of emoji

dager-mohamed:

discordjs support only `.gif` and `.png` for emoji extension, so I created the `getURL(ext)` function to make the user get any extension of emoji he wants.


packages/discord.js/src/structures/Emoji.js
```diff
@@ -47,13 +47,18 @@ class Emoji extends Base {
...
+   * @returns {string}
    */
-  get url() {
-    return this.id && this.client.rest.cdn.emoji(this.id, this.animated ? 'gif' : 'png');
+  url(extension) {
+    let final_result;
+    const extensions = ['webp', 'png', 'jpg', 'jpeg', 'gif'];
+    if (!extensions.includes(extension)) extension = this.animated ? 'gif' : 'png';
+    final_result = this.id && this.client.rest.cdn.emoji(this.id, extension);
+    return final_result;
```
thehackerboi69github, line 61:
```suggestion
    const extensions = ['webp', 'png', 'jpg', 'jpeg', 'gif'];
    if (!extensions.includes(extension)) extension = this.animated ? 'gif' : 'png';
    return this.id && this.client.rest.cdn.emoji(this.id, extension);
```
You're only assigning to `final_result` once.

thehackerboi69github (changes requested):
I'm not so sure about this one. This breaks consistency with other `url` getters in the library, and if users want to have another extension, they can use `client.rest.cdn.emoji(emoji.id, 'jpg')` for example. This function is a shortcut for that, but it breaks consistency, so I don't think it should be added like this.

packages/discord.js/src/structures/Emoji.js
dager-mohamed: Good job bro

jaw0r3k:
I agree with thehackerboi that .url should be a getter
So making function like getUrl would be better

dager-mohamed:
> I agree with thehackerboi that .url should be a getter So making a function like getUrl would be better

so I have an idea and I don't know if this idea is like you said or not,

I can create `getUrl(URL)` as you said function inside the class to check the URL of the emoji and then return the extension, 
and I will use the `getUrl()` function inside `url()` and it will be getter that you mean?

I hope that you mean.
سخ 
سخ

jaw0r3k:
Nah i just mean getURL(ext) which will return a url
and leave url as getter ( like everywhere else )

dager-mohamed:
> 

ahaa ok i understand now, i will do it

dager-mohamed:
now i added the `getUrl(ext)` function and `url` is a getter now.

packages/discord.js/src/structures/Emoji.js
```diff
@@ -56,6 +56,17 @@ class Emoji extends Base {
...
   }
 
+  /**
+   * @type {?string}
+   * @param {ImageExtension} ext the extension of emoji
+   * @returns {string}
+   */
+  getUrl(ext) {
+    const extensions = ['webp', 'png', 'jpg', 'jpeg', 'gif'];
+    if (!extensions.includes(ext) || !ext) return undefined;
```
Syjalo, line 66: Do we need that? The rest cdn getter throws an error if the extension is not valid.

Syjalo (changes requested):

packages/discord.js/src/structures/Emoji.js
dager-mohamed: > Do we need that? The rest cdn getter throws an error if the extension is not valid.

are you talking about the all `getUrl` function or you talking about this
```js
    const extensions = ['webp', 'png', 'jpg', 'jpeg', 'gif'];
    if (!extensions.includes(ext) || !ext) return undefined;
    ````
    
    anyway I made this to return undefined like `url` getter , anyway if it is not required we can remove it

File: packages/discord.js/src/structures/Emoji.js [hunk 3]
```diff
@@ -56,6 +56,17 @@ class Emoji extends Base {
...
+    return this.id && this.client.rest.cdn.emoji(this.id, ext);
```
thehackerboi69github, line 67: I think it's best to throw an error when the user provides an invalid extension.

packages/discord.js/src/structures/Emoji.js
```diff
@@ -56,6 +56,17 @@ class Emoji extends Base {
     return this.id && this.client.rest.cdn.emoji(this.id, this.animated ? 'gif' : 'png');
   }
 
+  /**
+   * @type {?string}
+   * @param {ImageExtension} ext the extension of emoji
+   * @returns {string}
+   */
+  getUrl(ext) {
```
thehackerboi69github, line 64:
```suggestion
  getURL(ext) {
```

packages/discord.js/src/structures/Emoji.js
dager-mohamed: you are right i will do it

dager-mohamed:
note: the rest.cdn is automatically throw the error at all

thehackerboi69github:
> note: the rest.cdn is automatically throw the error at all

In that case you don't have to return undefined if an invalid extension is provided.

thehackerboi69github (approved):

packages/discord.js/src/structures/Emoji.js
```diff
@@ -56,6 +56,15 @@ class Emoji extends Base {
...
   }
 
+  /**
+   * @type {?string}
+   * @param {ImageExtension} ext the extension of emoji
+   * @returns {string}
+   */
+  getURL(ext) {
+    return this.id && this.client.rest.cdn.emoji(this.id, ext);
+  }
```
NotSugden, line 66: i think the position of this function needs to be below all of the getter methods
```suggestion
  /**
   * @param {ImageExtension} ext the extension of emoji
   * @returns {?string}
   */
  getURL(ext) {
    return this.id && this.client.rest.cdn.emoji(this.id, ext);
  }
```
also corrected documentation

packages/discord.js/typings/index.d.ts
```diff
@@ -1070,6 +1070,7 @@ export class Emoji extends Base {
   public name: string | null;
   public get identifier(): string;
   public get url(): string | null;
+  public getURL(ext: ImageExtension): string;
```
NotSugden, line 1073:
```suggestion
  public getURL(ext: ImageExtension): string | null;
```
as per previous suggestion the method can return null

NotSugden (changes requested):

packages/discord.js/src/structures/Emoji.js
jaw0r3k:
```suggestion
  /**
   * @param {ImageExtension} ext The extension of the emoji
   * @returns {?string}
   */
  getURL(ext) {
    return this.id && this.client.rest.cdn.emoji(this.id, ext);
  }
```

Jiralite:
Superseded by #9788.

PR #8495, opened 2022-08-16
Fixed bug on TypeScript with the autocomplete in the options property

reallyslimy:
It changes some typings for the autocomplete property inside the package of discord.js in order to be usable by typescript.
With this change typescript users are now able to use the autocomplete inside options and change it to true | false and actually work with it.

discord.js version 14.2.0

- Code changes have been tested against the Discord API, or there are no code changes-

### DIFF
--- a/packages/discord.js/typings/index.d.ts
+++ b/packages/discord.js/typings/index.d.ts
@@ -3761,7 +3761,7 @@ export interface BaseApplicationCommandOptionsData {
   description: string;
   descriptionLocalizations?: LocalizationMap;
   required?: boolean;
-  autocomplete?: never;
+  autocomplete?: boolean;
 }
 
 export interface UserApplicationCommandData extends BaseApplicationCommandData {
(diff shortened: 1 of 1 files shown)


Jiralite:
This unfortunately is not the fix. Checking the `autocomplete` value of an attachment option is now a `boolean`. Before, the type of this was `never` which made sense as attachment options cannot have an `autocomplete` value.

reallyslimy:
Yeah it does makes sense, so instead, the `BaseApplicationCommandOptionsData` can be `autocomplete?: never` and only the `ApplicationCommandAutocompleteOption`, `ApplicationCommandChoicesOption` and `ApplicationCommandChoicesData` can be booleans instead, altho I'm not sure if `ApplicationCommandChoicesOption` and `ApplicationCommandChoicesData` should be booleans.

Jiralite:
> `ApplicationCommandAutocompleteOption`, `ApplicationCommandChoicesOption` and `ApplicationCommandChoicesData` can be booleans instead, altho I'm not sure if `ApplicationCommandChoicesOption` and `ApplicationCommandChoicesData` should be booleans.

There is no need to change `ApplicationCommandAutocompleteOption` to a boolean value. It cannot be `false`. Additionally, you cannot utilise choices and autocompletion in the same payload.

Jiralite:
Superseded by #8498.

PR #8504, opened 2022-08-16
feat: deprecate `ActionRow.from()`

Syjalo:
This PR adds a depreciation warning for `ActionRow.from()`

### DIFF
--- a/packages/discord.js/src/structures/ActionRow.js
+++ b/packages/discord.js/src/structures/ActionRow.js
@@ -1,8 +1,11 @@
 'use strict';
 
+const { emitWarning } = require('node:process');
 const { isJSONEncodable } = require('@discordjs/builders');
 const Component = require('./Component');
 const { createComponent } = require('../util/Components');
+
+let deprecationEmittedForFrom = false;
 
 /**
  * Represents an action row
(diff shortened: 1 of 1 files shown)


packages/discord.js/src/structures/ActionRow.js
```diff
@@ -24,8 +27,17 @@ class ActionRow extends Component {
    * Creates a new action row builder from JSON data
    * @param {JSONEncodable<APIActionRowComponent>|APIActionRowComponent} other The other data
    * @returns {ActionRowBuilder}
+   * @deprecated Use {@link ActionRowBuilder.from()} instead
```
Jiralite, line 30:
```suggestion
   * @deprecated Use {@link ActionRowBuilder.from()} instead.
```

Jiralite (changes requested):
Simply adding this after the class will do the trick:

```JavaScript
ActionRow.from = deprecate(
  ActionRow.from,
  'ActionRow.from() is deprecated. Use ActionRowBuilder.from() instead.',
);
```

Note the change of the deprecation message to be consistent with the `Formatters` class.

iCrawl (approved):

kyranet (approved):

PR #8507, opened 2022-08-17
chore: bump vitest and add @vitest/coverage-c8

imranbarbhuiya:
Bumped vitest to 0.22.0 and added `@vitest/coverage-c8`
ref: https://github.com/vitest-dev/vitest/releases/tag/v0.22.0

### DIFF
--- a/build.config.ts
+++ b/build.config.ts
@@ -37,6 +37,7 @@ export function createUnbuildConfig({
 	return defineBuildConfig({
 		entries,
 		clean: true,
+		failOnWarn: false,
 		rollup: {
 			esbuild: {
 				minify,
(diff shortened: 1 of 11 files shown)


build.config.ts
```diff
@@ -37,6 +37,7 @@ export function createUnbuildConfig({
 	return defineBuildConfig({
 		entries,
 		clean: true,
+		failOnWarn: false,
```
ckohen, line 40: This seems unrelated to this change?

ckohen (changes requested):
While not strictly necessary since c8 is default, maybe we should specify for clarity https://vitest.dev/guide/coverage.html#coverage-providers

iCrawl (approved):

PR #8510, opened 2022-08-17
feat(GuildMemberManager): addRole and removeRole

didinele:
Supersedes #8400.

### DIFF
--- a/packages/discord.js/src/managers/GuildMemberManager.js
+++ b/packages/discord.js/src/managers/GuildMemberManager.js
@@ -441,6 +441,42 @@ class GuildMemberManager extends CachedManager {
     return this.guild.bans.remove(user, reason);
   }
 
+  /**
+   * Options used for adding or removing a role from a member.
+   * @typedef {Object} AddOrRemoveGuildMemberRoleOptions
+   * @property {GuildMemberResolvable} user The user to add/remove the role from
+   * @property {RoleResolvable} role The role to add/remove
+   * @property {string} [reason] Reason for adding/removing the role
+   */
+
+  /**
+   * Adds a role to a member.
+   * @param {AddOrRemoveGuildMemberRoleOptions} options Options for adding the role
+   * @returns {Promise<GuildMember|User|Snowflake>}
+   */
+  async addRole(options) {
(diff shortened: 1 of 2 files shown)


packages/discord.js/src/managers/GuildMemberManager.js
```diff
@@ -441,6 +441,42 @@ class GuildMemberManager extends CachedManager {
...
+   * @typedef {Object} AddOrRemoveGuildMemberRoleOptions
+   * @property {GuildMemberResolvable} user The user to add/remove the role from
+   * @property {RoleResolvable} role The role to add/remove
+   * @property {string} [reason] Reason for adding/removing the role
+   */
+
+  /**
+   * Adds a role to a member.
+   * @param {AddOrRemoveGuildMemberRoleOptions} options Options for adding the role
+   * @returns {Promise<GuildMember|User|Snowflake>}
```
Jiralite, line 455: I personally don't like how we attempt to resolve the guild member as best as we can based on the developer's cache. I'd rather have just a `Snowflake` for which the developer can fetch/resolve to get what they want as they please. Wouldn't this also mean this "best as we can" approach would be expensive for big bots as it'd be iterating in two separate caches?

didinele: What I did is just consistent with all the other methods, though,

Jiralite: You could say that's consistently expensive then.

In any case, adding a description to the `@returns` describing how the different vales are returned would be ideal. For example:

https://github.com/discordjs/discord.js/blob/0f834029850d2448981596cf082ff59917018d66/packages/discord.js/src/managers/GuildBanManager.js#L135-L147

didinele: Regardless, considering those are just hashmap look-ups and not "iterations" over caches, its not all that bad all things considered.

As per the return doc, it'd be ideal if we did that for all the methods, but that'd probably be out of scope for this PR.

iCrawl (approved):

SpaceEEC (approved):

kyranet (approved):

PR #8513, opened 2022-08-17
fix(errors): add ClientMissingGuildIntent error

### LINKED ISSUE #8301: Insufficient error message for `member.displayHexColor` when GUILDS intent is missing

discord.js

### Issue description

1. Create a bot *without* member intent.
2. Get a guild member
3. call `member.displayHexColor`
4. Get an error

The problem here is not inherently with getting an error, but rather with the fact that the error is hard to understand for someone who has no idea that they are missing the GUILDS intent. They might try to console log, but will only get this error again, not undefined or anything. I fully believe this is very confusing and there should be a check for the guilds intent (or if the roles cache is empty) that throws with the message that the GUILDS intent is missing.
Screenshot from 2022-07-17 17-23-35

### Code sample

```typescript
const client = new Client({
  intents: [],
})

client.on('interactionCreate, async (interaction) => {
  if (!interaction.member || interaction.member === undefined) { return; }
  console.log(interaction?.member?.displayHexColor) // This will error
})
```

### Package version

13.8.1

### Node.js version

17.5.0

### Operating system

Linux Manjaro

Low (slightly annoying)

No Intents

_No response_

RedGuy12:
Fixes #8301 by throwing an error when the `Guilds` intent is not provided.

### DIFF
--- a/packages/discord.js/src/client/Client.js
+++ b/packages/discord.js/src/client/Client.js
@@ -4,6 +4,7 @@ const process = require('node:process');
 const { Collection } = require('@discordjs/collection');
 const { makeURLSearchParams } = require('@discordjs/rest');
 const { OAuth2Scopes, Routes } = require('discord-api-types/v10');
+const { GatewayIntentBits } = require('discord-api-types/v10');
 const BaseClient = require('./BaseClient');
 const ActionsManager = require('./actions/ActionsManager');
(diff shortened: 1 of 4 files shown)


monbrey:
This is a bad solution. We do not want to prevent discord.js from being used without the Guilds intent if the developer understands the implications.

PR #8515, opened 2022-08-17
types(ChannelManager): fetch cannot return null

ImRodry:

The types for ChannelManager#fetch were wrong as no fetch operation returns null, it rejects if the channel is not found


Jiralite:
This returns `null` if discord.js does not recognise the channel:

https://github.com/discordjs/discord.js/blob/74740260a783bd969ea8c49ceb6b795e0f4608c2/packages/discord.js/src/managers/ChannelManager.js#L50-L55

ImRodry:
That is stupid

PR #8519, opened 2022-08-18
feat(website): show parameter descriptions

suneettipirneni:

Previously param descriptions would always be "None" despite having descriptions in the source code. 

Additionally. this PR tweaks the table resizing. The columns that show types now break in a more legible way, previously the types would be smashed into 1-2 character lines which would render them unreadable in small breakpoints. Tables also now scroll horizonally when the content within them cannot be shrank any farther.

### DIFF
--- a/packages/website/src/components/DocContainer.tsx
+++ b/packages/website/src/components/DocContainer.tsx
@@ -43,7 +43,7 @@ export function DocContainer({
 
 	return (
 		<Group>
-			<Stack sx={{ flexGrow: 1 }} className="max-w-full">
+			<Stack sx={{ flexGrow: 1, maxWidth: '100%' }}>
 				<Title sx={{ wordBreak: 'break-all' }} order={2} ml="xs">
 					<Group>
 						{generateIcon(kind)}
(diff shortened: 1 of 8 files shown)


imranbarbhuiya:
image
it adds this scrollbar. Is it intentional?

suneettipirneni:
> image it adds this white scrollbar. Is it intentional?

Whoops, that should be fixed. The scroll bar should only show for the table at small breakpoints where the text can't fit.

packages/website/src/components/DocContainer.tsx
```diff
@@ -43,7 +43,7 @@ export function DocContainer({
 
 	return (
 		<Group>
-			<Stack sx={{ flexGrow: 1 }}>
+			<Stack sx={{ flexGrow: 1 }} className="max-w-full">
```
iCrawl, line 46:
```suggestion
			<Stack sx={{ flexGrow: 1, maxWidth: '100%' }}>
```

iCrawl (changes requested):

PR #8527, opened 2022-08-19
feat(website): render `@defaultValue` blocks

suneettipirneni:

Members with `@defaultValue` tags will show their default value in the docs.

### DIFF
--- a/packages/rest/src/lib/REST.ts
+++ b/packages/rest/src/lib/REST.ts
@@ -25,45 +25,45 @@ export interface RESTOptions {
 	agent: Dispatcher;
 	/**
 	 * The base api path, without version
-	 * @default 'https://discord.com/api'
+	 * @defaultValue `'https://discord.com/api'`
 	 */
 	api: string;
 	/**
 	 * The authorization prefix to use for requests, useful if you want to use
 	 * bearer tokens
 	 *
-	 * @default 'Bot'
+	 * @defaultValue `'Bot'`
 	 */
 	authPrefix: 'Bot' | 'Bearer';
 	/**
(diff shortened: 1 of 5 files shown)


Syjalo:
Shouldn't it show something like `Default value: {value}`?

suneettipirneni:
> Shouldn't it show something like `Default value: {value}`?

It does:
Screen Shot 2022-08-19 at 12 29 44 PM

I assume you're viewing the PR preview, which didn't have the updated api.json being used.

Syjalo:
> > Shouldn't it show something like `Default value: {value}`?
> 
> It does: Screen Shot 2022-08-19 at 12 29 44 PM
> 
> I assume you're viewing the PR preview, which didn't have the updated api.json being used.

Does this also apply to local launch?

suneettipirneni:
> Does this also apply to local launch?

Yes, if you build the relevant packages before running the website locally it should use the updated api.json.

Syjalo:
Double quotes (\`\`Node.js $process.version\`\`)
image

packages/rest/src/lib/REST.ts
```diff
@@ -25,45 +25,45 @@ export interface RESTOptions {
...
 	 *
-	 * @default 'Bot'
+	 * @defaultValue `'Bot'`
 	 */
 	authPrefix: 'Bot' | 'Bearer';
 	/**
 	 * The cdn path
 	 *
-	 * @default 'https://cdn.discordapp.com'
+	 * @defaultValue 'https://cdn.discordapp.com'
```
Syjalo, line 41: You forgot to put the quotes

PR #8541, opened 2022-08-22
types(webhook): avatar can be null

DTrombett:

`Webhook#avatar` can be null if the webhook hasn't any avatar. The documentation is correct, and `Webhook#avatarURL()` handles it correctly, but types were missing the `null` type.

### DIFF
--- a/packages/discord.js/typings/index.d.ts
+++ b/packages/discord.js/typings/index.d.ts
@@ -2811,7 +2811,7 @@ export class VoiceState extends Base {
 
 export class Webhook extends WebhookMixin() {
   private constructor(client: Client, data?: RawWebhookData);
-  public avatar: string;
+  public avatar: string | null;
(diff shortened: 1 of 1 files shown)


SpaceEEC (approved):

iCrawl (approved):

packages/discord.js/typings/index.d.ts
```diff
@@ -2811,7 +2811,7 @@ export class VoiceState extends Base {
 
 export class Webhook extends WebhookMixin() {
   private constructor(client: Client, data?: RawWebhookData);
-  public avatar: string;
+  public avatar: string | null;
```
Syjalo, line 2814: Shouldn't it be set to `null` in the code then?

Syjalo (changes requested):

vladfrangu (approved):

PR #8544, opened 2022-08-22
feat(website): show class name in constructor doc

imranbarbhuiya:


suneettipirneni (changes requested):
I understand what this PR is trying to accomplish, however I disagree with this. The documentation should show something resembling the *signature* of a property not it's usage. In this PR it's showing it's usage to initialize it. Usage of `new` should be in an `@example` block not within the constructor title.

PR #8547, opened 2022-08-22
feat(website): add index generation script

suneettipirneni:


packages/website/generateIndex.ts
```diff
@@ -0,0 +1,135 @@
...
+if (!fs.existsSync(dir)) {
+	fs.mkdirSync(dir);
+}
+
+const members = packageNames.reduce<MemberJSON[]>((acc, pkg) => {
+	model.loadPackage(path.join('..', pkg, 'docs', 'docs.api.json'));
+	return acc.concat(visitNodes(model.tryGetPackageByName(pkg)!.entryPoints[0]!));
+}, []);
+
+fs.writeFile(path.join('searchIndex', 'doc-index.json'), JSON.stringify(members, undefined, 2), (err) => {
```
Jiralite, line 131: Prefer `null` over `undefined` here? (Same behaviour really)

```suggestion
fs.writeFile(path.join('searchIndex', 'doc-index.json'), JSON.stringify(members, null, 2), (err) => {
```

PR #8556, opened 2022-08-25
feat: @discordjs/redis-collection

imranbarbhuiya:
redis collection


DraftProducts:
This is a really nice addition 🚀
This coordinates with @kyranet's efforts to reduce discord.js cache usage.
However, is there any particular reason for using [ioredis](https://github.com/luin/ioredis) rather than [node-redis](https://github.com/redis/node-redis) ?

kyranet:
I don't think, to be honest, that the Redis collection will actually save a significant amount of memory.

And the reason why ioredis over node-redis, is because there's literally no reason to use the latter. The former comes with massive performance improvements along several other enhancements.

kyranet:
Problem is that the data is stored in KV pairs, whereas V8's objects store data in value-only arrays alongside a pointer to a Shape.

While binary data storage can be compact, KV storage can make it use more memory than V8. If you want to do something even more efficient, you'll need a custom aligned value-only binary RW, which can be application specific. In this case, yes, there might be a benefit in memory usage if we store data in a layout similar to C. I have written said RW for an application of mine (and later @didinele tried it on her project), but I have yet to measure the memory usage of discord.js vs value storage.

The bar for saving memory is very high, and such we can't have savings unless we do some complicated low-level code.

sebasptsch:
This looks really helpful, I am wondering whether the [`cache-manager`](https://www.npmjs.com/package/cache-manager) package was being considered to make this a little more versatile?

KagChi:
Redis cluster support?

imranbarbhuiya:
Since some people have an interest in this pr, so please add a reason before closing.

iCrawl:
Because this is not actually what anyone wants. As much as you think they do.

1) This is a performance nightmare. All the serializing and deserializing happening here. If this were in the main lib to replace in-memory caching, you might as well run on an old Pentium 4 with a 1MB/s connection.
2) This does not utilize the JSON Redis module, which honestly it should because see point 1.
3) There are efficient solutions out there for how to store your data properly in your database and this should not be used as a "storing" layer for random things that should be saved properly. It just leads to a horrible "lead-by-example", because let's be real: Everyone would use this instead of storing stuff in a proper way (Whether that be actual K/V in Redis or SQL)

PR #8566, opened 2022-08-28
feat(NewsChannel)!: return followed channel data

Syjalo:
BREAKING CHANGE: `GuildChannelManager#addFollower` and `AnnouncementChannel#addFollower` now return `FollowedChannelData` instead of the webhook id.

This PR changes the return type of `NewsChannel#addFollower()` method to the followed channel.


Jiralite:
I agree with this change. Sadly, this is currently breaking.

What we can do in a non-breaking way is add this method to the `GuildChannelManager` as it doesn't exist (for some reason). That would not be breaking and the return type can be a `Snowflake`. In the breaking change, we would internally call the `GuildChannelManager`.

packages/discord.js/src/structures/NewsChannel.js
```diff
@@ -13,7 +13,7 @@ class NewsChannel extends BaseGuildTextChannel {
    * Adds the target to this channel's followers.
    * @param {TextChannelResolvable} channel The channel where the webhook should be created
    * @param {string} [reason] Reason for creating the webhook
-   * @returns {Promise<NewsChannel>}
+   * @returns {Promise<Snowflake>} Returns created target webhook id.
```
vladfrangu, line 16:
```suggestion
   * @returns {Promise<Snowflake>} Returns the created webhook id in the target channel.
```

vladfrangu (approved):

packages/discord.js/src/structures/NewsChannel.js
Jiralite: Seems like one might as well return the whole object instead of just the id?

Jiralite: discord.js doesn't cache them. Doesn't mean the user can do it themselves.

Syjalo: I don't see any way to do so, unless fetching the webhook, which is not ideal. Returning either the object or snowflake will force to check if it's one or not. If you have any code suggestions share them

almeidx (approved):

packages/discord.js/src/structures/NewsChannel.js
Jiralite: This is irrelevant to the point, and another point to raise is this is not a future-proof solution. If Discord decides to make a sparkling new feature related to this and more properties get returned, it would be a breaking change to change the return type. This should be an object.

Jiralite (changes requested):
This should be calling the guild channel manager's method and ideally have the changes made noted here: https://github.com/discordjs/discord.js/pull/8566#discussion_r1844954508.

packages/discord.js/src/structures/NewsChannel.js
Syjalo: I just realised what you mean

packages/discord.js/typings/index.d.ts
```diff
@@ -4211,6 +4211,11 @@ export class GuildApplicationCommandManager extends ApplicationCommandManager<Ap
   public set(commands: readonly ApplicationCommandDataResolvable[]): Promise<Collection<Snowflake, ApplicationCommand>>;
 }
 
+export interface FollowedChannel {
+  channelId: Snowflake;
+  webhookId: Snowflake;
+}
```
almeidx, line 4217: IMO we should pick a different name for this, since `FollowedChannel` follows the pattern we use for actual channels, i.e., `TextChannel`, which might mislead users

Something like `FollowedChannelData` could perhaps work

almeidx (approved):

Jiralite (approved):

PR #8575, opened 2022-08-30
feat(GuildBanManager): Add `deleteMessageSeconds`

Jiralite:
Backports #8326 to version 13.

### DIFF
--- a/src/managers/GuildBanManager.js
+++ b/src/managers/GuildBanManager.js
@@ -1,10 +1,13 @@
 'use strict';
 
+const process = require('node:process');
 const { Collection } = require('@discordjs/collection');
 const CachedManager = require('./CachedManager');
 const { TypeError, Error } = require('../errors');
 const GuildBan = require('../structures/GuildBan');
 const { GuildMember } = require('../structures/GuildMember');
+
+let deprecationEmittedForDays = false;
 
 /**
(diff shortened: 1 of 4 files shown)


kyranet (approved):

src/managers/GuildBanManager.js
```diff
@@ -142,15 +148,30 @@ class GuildBanManager extends CachedManager {
...
+      deprecationEmittedForDays = true;
+    }
+
     await this.client.api
       .guilds(this.guild.id)
       .bans(id)
       .put({
-        data: { delete_message_days: options.days },
+        data: {
+          delete_message_seconds:
```
vladfrangu, line 170: Same requested change as in the main PR (is it optional, if it is we shouldn't enforce it)

vladfrangu (changes requested):

src/managers/GuildBanManager.js
Jiralite: This would be a breaking change since it is a default.

Jiralite: [Internally resolved.](https://canary.discord.com/channels/222078108977594368/999796393810067607/1014275925812183150)

vladfrangu (approved):

PR #8594, opened 2022-09-04
feat(Message): add `position` for message

dager-mohamed:

### DIFF
--- a/packages/discord.js/src/structures/Message.js
+++ b/packages/discord.js/src/structures/Message.js
@@ -357,6 +357,15 @@ class Message extends Base {
     } else {
       this.interaction ??= null;
     }
+    if (data.position) {
+      /**
+       * The position of the message
+       * @type {?number}
+       */
+      this.position = data.position;
+    } else {
+      this.position ??= null;
+    }
   }
 
   /**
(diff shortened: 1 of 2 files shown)


packages/discord.js/src/structures/Message.js
```diff
@@ -357,6 +357,15 @@ class Message extends Base {
     } else {
       this.interaction ??= null;
     }
+    if (data.position) {
```
almeidx, line 360:
```suggestion

    if (typeof data.position !== 'undefined') {
```

almeidx (changes requested):

packages/discord.js/src/structures/Message.js
```diff
@@ -357,6 +357,16 @@ class Message extends Base {
     } else {
       this.interaction ??= null;
     }
+
+    if (typeof data.position !== 'undefined') {
```
almeidx, line 361: Actually, should be like this to be consistent with the rest of the properties
```suggestion
    if ('positon' in data) {
```

almeidx (changes requested):

vladfrangu (approved):

PR #8598, opened 2022-09-05
fix: `awaitMessageComponent` with `MessageComponentInteraction`s

### LINKED ISSUE #8501: InteractionResponse#awaitMessageComponent

discord.js

### Issue description

The awaitMessageComponent method doesn't work when called on an InteractionResponse despite it existing and being typed as valid. The method will never catch an interaction and will reject if a time period is specified. Not sure why this happens seeing as the InteractionCollector class accepts an interactionResponse in its constructor.

### Code sample

```typescript
if (interaction.isButton()) {
	const res = await interaction.update({ content: new Date().toISOString(), components: interaction.message.components }),
		int = await res.awaitMessageComponent({ time: 5_000 }).catch(() => null)
	console.log(int) // int will always be null, even if you reply before the 5 second limit
}
```

### Package version

14.2.0

### Node.js version

18.7.0

### Operating system

Windows 11

Medium (should be fixed soon)

_No response_

RedGuy12:

Fixes #8501
Check the message id when a `MessageComponentInteraction` is passed, and ignore the interaction id if it's not a slash command.


packages/discord.js/src/structures/InteractionCollector.js
```diff
@@ -150,7 +150,13 @@ class InteractionCollector extends Collector {
     if (this.interactionType && interaction.type !== this.interactionType) return null;
     if (this.componentType && interaction.componentType !== this.componentType) return null;
     if (this.messageId && interaction.message?.id !== this.messageId) return null;
-    if (this.messageInteractionId && interaction.message?.interaction?.id !== this.messageInteractionId) return null;
+    if (
+      this.messageInteractionId &&
+      interaction.message?.interaction?.id &&
+      interaction.message?.interaction?.id !== this.messageInteractionId
```
almeidx, line 156:
```suggestion
      interaction.message.interaction.id !== this.messageInteractionId
```

almeidx (changes requested):

packages/discord.js/typings/index.d.ts
```diff
@@ -483,7 +483,7 @@ export abstract class CommandInteraction<Cached extends CacheType = CacheType> e
 }
 
 export class InteractionResponse<Cached extends boolean = boolean> {
-  private constructor(interaction: Interaction, id?: Snowflake);
+  private constructor(interaction: Exclude<Interaction, AutocompleteInteraction>, id?: Snowflake);
   public interaction: Interaction<WrapBooleanCache<Cached>>;
```
Jiralite, line 487: Isn't the typings here not updated?

packages/discord.js/src/structures/InteractionCollector.js
```diff
@@ -150,7 +150,13 @@ class InteractionCollector extends Collector {
     if (this.interactionType && interaction.type !== this.interactionType) return null;
     if (this.componentType && interaction.componentType !== this.componentType) return null;
     if (this.messageId && interaction.message?.id !== this.messageId) return null;
-    if (this.messageInteractionId && interaction.message?.interaction?.id !== this.messageInteractionId) return null;
+    if (
+      this.messageInteractionId &&
+      interaction.message?.interaction?.id &&
```
vladfrangu, line 155: Is this needed?

packages/discord.js/typings/index.d.ts
RedGuy12: Reverted.

packages/discord.js/src/structures/InteractionCollector.js
RedGuy12: You're right, this isn't needed because if they're both falsey, the below !== comparison will return true anways. Removed.

packages/discord.js/src/structures/InteractionCollector.js
```diff
@@ -40,7 +40,7 @@ class InteractionCollector extends Collector {
      * The message from which to collect interactions, if provided
      * @type {?Snowflake}
      */
-    this.messageId = options.message?.id ?? null;
+    this.messageId = options.message?.id ?? options.interactionResponse?.message?.id ?? null;
```
SpaceEEC, line 43: `InteractionResponse#message` does not exist.
It is possible to access `message` through the interaction from `InteractionResponse#interaction`.

---

That wouldn't solve the issue however because the next property
https://github.com/discordjs/discord.js/blob/048e80a8224634f30d2431a488dccd4de713519f/packages/discord.js/src/structures/InteractionCollector.js#L49
is set and then the following check here
https://github.com/discordjs/discord.js/blob/048e80a8224634f30d2431a488dccd4de713519f/packages/discord.js/src/structures/InteractionCollector.js#L174
would fail as the message would not have an interaction (as it's not an interaction originating message).

---

(Wouldn't this fix also break if you were to use `ButtonInteraction#reply` and thus have a new message with a different id?)

SpaceEEC (dismissed):

ImRodry:
@RedGuy12 are you still maintaining this PR? Would be nice to have this fixed

packages/discord.js/src/structures/InteractionCollector.js
RedGuy12: I'm not sure why I removed this looking back. They can't both be falsey when the `this.messageInteractionId` check is there. If `interaction.message?.interaction?.id` is falsey, then they're not equal to each other, this will evaluate to true, and it will return null.

packages/discord.js/src/structures/InteractionCollector.js
almeidx, line 156: Seems like this change got rolled back?
```suggestion
      interaction.message.interaction.id !== this.messageInteractionId
```

almeidx (changes requested):

packages/discord.js/src/structures/InteractionCollector.js
RedGuy12: > (Wouldn't this fix also break if you were to use `ButtonInteraction#reply` and thus have a new message with a different id?)

I'm not sure what you mean by this. Could you elaborate?

SpaceEEC: The target message differs depending on which method you used:
- In case if `update`, you'd want the original interaction message.
- In case of `reply`, you'd want the new message.

RedGuy12: I don't think it's possible to get the new message's id without fetching it...that removes the whole point of this class...

kyranet (approved):

iCrawl (approved):

suneettipirneni (approved):

PR #8600, opened 2022-09-06
refactor: website components

iCrawl:


packages/website/package.json
```diff
@@ -41,52 +41,50 @@
...
-		"@mantine/nprogress": "^5.2.5",
-		"@mantine/spotlight": "^5.2.5",
 		"@microsoft/api-extractor-model": "7.24.0",
 		"@microsoft/tsdoc": "0.14.1",
 		"@vscode/codicons": "^0.0.32",
+		"ariakit": "^2.0.0-next.41",
 		"minisearch": "^5.0.0",
 		"next": "^12.2.5",
 		"next-mdx-remote": "^4.1.0",
+		"next-progress": "^2.2.0",
```
imranbarbhuiya, line 51: [nextjs-progressbar](https://www.npmjs.com/package/nextjs-progressbar) is a good alternative with 30x more downloads

imranbarbhuiya (changes requested):

packages/website/package.json
almeidx: 4460c573bfff08f38e19fc62af96c9552195fb25

iCrawl: Theres actually a few differences between the 2 packages, one would be allowing full custom css, which `nextjs-progressbar` doesn't. Plus its wrapped in a `memo`, which `nextjs-progressbar` is not.

packages/website/package.json
```diff
@@ -41,52 +41,50 @@
...
-		"@emotion/server": "^11.10.0",
-		"@mantine/core": "^5.2.5",
-		"@mantine/hooks": "^5.2.5",
-		"@mantine/next": "^5.2.5",
-		"@mantine/nprogress": "^5.2.5",
-		"@mantine/spotlight": "^5.2.5",
 		"@microsoft/api-extractor-model": "7.24.0",
 		"@microsoft/tsdoc": "0.14.1",
 		"@vscode/codicons": "^0.0.32",
+		"ariakit": "^2.0.0-next.41",
```
diegohaz, line 47: I recommend setting it to the exact version (without `^`) while it's in alpha so that you don't get breaking changes by mistake.

PR #8601, opened 2022-09-06
fix: fixed TS 4.8 compatibility

favna:

Currently, users of DiscordJS v13 and TypeScript 4.8 require using `"skipLibCheck": true` because the changed line raises type errors.

At the same time, however, fixing the issue is unfeasible because it would also require `0.33.6` of discord-api-types. I tried tracing the issue back to the source and the best I got went into `APITextInputComponent` and I can't go and change discord-api-types in node_modules.

Other than that I also cleaned up some unused imports that were showing up.


kyranet (approved):

vladfrangu (approved):

typings/index.d.ts
```diff
@@ -1632,6 +1627,8 @@ export class MessageActionRow<
     ? APIActionRowComponent<APIModalActionRowComponent>
     : APIActionRowComponent<APIMessageActionRowComponent>,
 > extends BaseMessageComponent {
+  // @ts-expect-error (TS:2344, Caused by TypeScript 4.8)
```
imranbarbhuiya, line 1630: Ig `ts-expect-error` will throw for ts<4.8 users as it's valid.  Instead use `ts-ignore` here

imranbarbhuiya (changes requested):

typings/index.d.ts
favna, line 1630:
```suggestion
  // @ts-ignore (TS:2344, Caused by TypeScript 4.8)
```

Am on work laptop w/o the repo cloned, this is faster to process @imranbarbhuiya's comment.

favna: Good point. Resolved.

PR #8613, opened 2022-09-11
chore: use webpack plugin for unocss

imranbarbhuiya:
I'm not sure why cli is used where we need to run both commands. Instead, we can just use the webpack plugin too
Also removed `fallbackNodePolyfills: true` as it defaults to true.


iCrawl:
It doesn’t work well with HMR and if you want to make it work you have to disable webpack caching

PR #8616, opened 2022-09-11
fix(ClientOptions): make ClientOptions#intents returns an IntentsBitField

Idris1401:

I think it's better than `ClientOptions#intents` returns an instance of IntentsBitField instead of just a number.


Idris1401:
forgot how "Status and versioning classification" works ._.

PR #8618, opened 2022-09-11
refactor(Client): make censoring token in debug log faster

That-Guy977:

This PR modifies the method used to censor tokens in debug logs within `Client#login`. From benchmarking, this is about 1/3 the duration (~`0.003` -> ~`0.001`) although both are very fast, so this may be an unnecessary change.

- Typings don't need updating


That-Guy977:
Sorry, I've gotten quite sloppy with my testing. The final difference is about `0.0002` per call, so this is very much unnecessary, and much less readable.

PR #8641, opened 2022-09-19
fix: Correct applied tags type

Jiralite:
This required a full object, but only the ids are necessary.

In addition:

- The return type is corrected on the helper method
- `ThreadEditData` now possesses this property
- The return type of the method in the typings no longer possibly returns a private thread


almeidx (approved):

jaw0r3k:
maybe it could accept Tag or id ( like resolver )

SpaceEEC (approved):

kyranet (approved):

PR #8680, opened 2022-09-25
ThreadChannel#setLocked fix

ErdemGKSL:
In order to lock the channel, you need to archive it at the same time according to discord API.

### DIFF
--- a/packages/discord.js/src/structures/ThreadChannel.js
+++ b/packages/discord.js/src/structures/ThreadChannel.js
@@ -344,7 +344,7 @@ class ThreadChannel extends BaseChannel {
     const newData = await this.client.rest.patch(Routes.channel(this.id), {
       body: {
         name: (data.name ?? this.name).trim(),
-        archived: data.archived,
+        archived: data.locked ? true : data.archived,
         auto_archive_duration: data.autoArchiveDuration,
         rate_limit_per_user: data.rateLimitPerUser,
         locked: data.locked,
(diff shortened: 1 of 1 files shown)


Jiralite:
I don't think this is correct.

You can lock a request, sure, and nothing happens visibly. However, closing it normally will actually lock the thread too because `locked` is `true`.

Look at what happens in the client. There's no "Lock" option, only "Close & Lock". Discord is setting both `archived` and `lock` to `true` here.

If you want to close and lock a thread, you should close and lock the thread - not only lock it.

ErdemGKSL:
So that would make more sense right? I was confused because I was judging the API by the frontend requests actually sorry for my mistake. Wouldn't such use be more logical so that others do not confuse it? Or it can be in setLockedAndArchived in order to maintain systematicity.

ImRodry:
If your thread is archived and you want to lock it, you need to unarchive and then archive and lock at the same time. If it is open and you want to lock it you have to archive and lock like is being done in this PR, I just don't know if this is all that useful, would just be another shortcut to something you could easily do yourself

jaw0r3k:
and also it need typings

packages/discord.js/src/structures/ThreadChannel.js
```diff
@@ -568,15 +584,15 @@ class ThreadChannel extends BaseChannel {
...
-  bulkDelete() {}
+  get lastMessage() { }
+  get lastPinAt() { }
+  send() { }
+  sendTyping() { }
+  createMessageCollector() { }
+  awaitMessages() { }
+  createMessageComponentCollector() { }
+  awaitMessageComponent() { }
+  bulkDelete() { }
```
NotSugden, line 595:
```suggestion
  get lastMessage() {}
  get lastPinAt() {}
  send() {}
  sendTyping() {}
  createMessageCollector() {}
  awaitMessages() {}
  createMessageComponentCollector() {}
  awaitMessageComponent() {}
  bulkDelete() {}
```

packages/discord.js/src/structures/ThreadChannel.js
```diff
@@ -422,6 +422,22 @@ class ThreadChannel extends BaseChannel {
...
+   * {@link PermissionFlagsBits.ManageThreads} permission can unarchive it.
+   * @param {string} reason Reason for locking or unlocking the thread
+   * @returns {Promise<ThreadChannel>}
+   * @example
+   * // Set the thread to locked and archived
+   * thread.closeAndLock()
+   *   .then(() => console.log(`Thread is now locked and archived.`))
+   *   .catch(console.error);
+   */
+  closeAndLock(reason) {
```
NotSugden, line 437: not a fan of the name.

i think a better soloution would be to add a `lock` parameter to setArchive?

NotSugden (changes requested):
typings required

packages/discord.js/src/structures/ThreadChannel.js
ckohen: I agree setArchived should have three potential signatures,
setArchived() => archives
setArchived(false, lock: never) => unarchived
setArchived(true, lock: boolean) => archive and set lock accordingly

Issue #8015, opened 2022-06-05
RFC(builders): The interface, the validation, and the pains you have had with them so far


vladfrangu:
*Preface: this RFC (Request For Comments) is meant to gather feedback and opinions for the 3 issues that will be mentioned down below. I understand builders tend to be quite a sensitive topic with everyone disagreeing on how it should be done, so please keep the feedback useful. Thanks!*

This RFC aims to solve (hopefully once and for all) the three pain points of builders as they currently are done. The end goal is to decide what the interface should look like for all current and possibly future builders! It is split into 3 separate parts, denoted by the headings.

Before you answer, please read through the entire RFC, and don’t jump to answering just one question only. You don’t have to answer question 3 if you have nothing to add to it, but the other two would be super appreciated to receive answers to. Thanks for reading this, and now let’s get started~!

---

# 1. The interface

One of the key issue with the current builders is the lack of a consistent interface between the Application Command builders, the Message Component builders and the Embed builder.

As it stands at the time of writing this RFC:

- Application command builders follow a nested approach, with `.add*` methods for each option type. The class’s fields represent the structure similar to how the api payload will look like.
- Message component builders follow part of the same structure as application command builders, with some caveats (`addComponents` takes in any component, whereas `add*` in application command builders add just that one kind of options), some options that take in arrays allow rest parameters too (something that in my opinion is fine except for `set*` methods that should take in an array only). The class stores all data in a `data` property, and accesses it for setting/reading the properties in it
- Embed builder also stores the actual api embed data in the `data` field. They also allow input in `camelCase` format, where no other builder allows (or should allow) that.

Several things need to be decided:

1. How should the classes store the data that will be returned from `toJSON()` calls? In a `data` field or as the class fields themselves?
2. Should `addComponents` be removed in favor of `addButton`, `addSelectMenu` and so on? Or should application command builders use `addOption` instead?
3. Should nested objects in any builder be a raw object or should they be miniature builders? Consider EmbedBuilder#author. Currently it’s a raw object. Should it be brought in line with the other two builders and nest it as a builder?

My opinion on those 3 questions are:

1. ~~The class fields should represent as closely what the toJSON will return. Calling `{ ...builder }` should return the same valid data as calling `builder.toJSON()`~~ Well I changed my mind, we should instead use the `data` field, however no getters should be added pointing to said field
2. ActionRowComponent should use the same interface as application command builders (one `add*` per component type)
[...]

PR #8682, opened 2022-09-26
feat(EmbedBuilder): add `.length`

### LINKED ISSUE #8647: Adding a string length property to embeds builders

discord.js

### Feature

In v13 you used to be able to use .length to get the total text length of all the text in an embed (author, title, description, fields, etc.) but in v14 this was not added to the EmbedBuilder, and thus there doesn't appear to be a way to get this information now without doing it manually. From what I've heard it seems to be still returned from the API, so it doesn't seem like it'd be too hard to add back in.

When using .length on an EmbedBuilder, it should return the total text length of the entire embed from the API.

_No response_

### Other context

_No response_

RedGuy12:
Resolves #8647 


RedGuy12:
Could I get a review on this?

iCrawl:
Wouldn't it be fine if users just do:

```ts
import { embedLength } from '@discordjs/builders`;

embedLength(YourEmbedBuilderInstance.toJSON());
```
?

RedGuy12:
> Wouldn't it be fine if users just do

Yes, but why does [Embed have a length property](https://github.com/discordjs/discord.js/blob/9d8179c6a78e1c7f9976f852804055964d5385d4/packages/discord.js/src/structures/Embed.js#L183) then?

kyranet:
I'm against this change, it goes against [RFC](https://github.com/discordjs/discord.js/issues/8015) about exposing a builder's data via getters:

> Well I changed my mind, we should instead use the `data` field, however no getters should be added pointing to said field.

Jiralite (changes requested):
Can you add typings?

vladfrangu (approved):

Jiralite (approved):

kyranet (approved):

PR #8688, opened 2022-09-30
chore: change token in readme to constant

jeongnaehyeok:
I was confused when I saw the guide code in README.md. Cuz **'CLIENT_ID'** is a constant expression but **'token'** is not, I pondered for a moment what was wrong. I think this should be expressed in the same way. So I modify this.
Modify **'token'** to **'TOKEN'** in all example codes


jaw0r3k:
um as u said to 'TOKEN'
not to variable

README.md
```diff
@@ -66,7 +66,7 @@ const commands = [
 	},
 ];
 
-const rest = new REST({ version: '10' }).setToken('token');
+const rest = new REST({ version: '10' }).setToken(TOKEN);
```
jaw0r3k, line 69:
```suggestion
const rest = new REST({ version: '10' }).setToken('TOKEN');
```

jaw0r3k: hmm idk

jeongnaehyeok: I think it would be good to use it as one expression. This confuses beginners.

iCrawl (approved):

kyranet (approved):

vladfrangu (approved):

SpaceEEC (approved):

PR #8701, opened 2022-10-04
feat(Util): escape more markdown characters

### LINKED ISSUE #8689: `escapeMarkdown` does not escape backslashes

discord.js

### Issue description

I wasn't sure if this counts as a bug or a feature suggestion, but I expected `escapeMarkdown` to also escape the `\` character, and it does not, leaving me to have to write `.replace(/\\/g, '\\\\')` everywhere I call `escapeMarkdown` (or a function of my own that does the same).

There should perhaps also be a separate function for only escaping backslashes, like `escapeEscapes`.

### Code sample

```typescript
const { escapeMarkdown } = require('discord.js');

console.log(escapeMarkdown('\\')); // Outputs '\\' instead of '\\\\'
```

### Package version

14.5.0

### Node.js version

16.13.2

### Operating system

Windows

Low (slightly annoying)

No Partials

No Intents

_No response_

RedGuy12:
Allows for escaping more characters in `escapeMarkdown` and adds the corresponding individual functions:
- escape characters (resolves #8689)
- bulleted and numbered lists, and headings (new markdown only supported in forum guidelines and starter messages; resolves https://discord.com/channels/222078108977594368/1019980824365506570)
- masked links (only supported in webhook messages)

I enabled the `escapeEscapes` option by default since you can add those to any message, but since lists, headings, and masked links are only supported in specific contexts I disabled them by default. The `escapeMarkdown` function now returns different information by default because `escapeEscapes` is enabled by default. If this is a problem, I can disable it by default until the next major release.

- This PR includes *potentially* breaking changes (methods removed or renamed, parameters moved or removed)


packages/discord.js/typings/index.d.ts
```diff
@@ -4659,8 +4664,13 @@ export interface EscapeMarkdownOptions {
   underline?: boolean;
   strikethrough?: boolean;
   spoiler?: boolean;
-  inlineCodeContent?: boolean;
```
RedGuy12, line 4662: I moved this down two lines so it can be in the same order as in `Util.js`. If wanted I can revert this.

packages/discord.js/src/util/Util.js
```diff
@@ -210,6 +235,51 @@ function escapeSpoiler(text) {
...
+}
+
+/**
+ * Escapes masked link characters in a string.
+ * @param {string} text Content to escape
+ * @returns {string}
+ */
+function escapeMaskedLink(text) {
+  return text.replaceAll(/\[.+\]\(.+\)/gm, '\\$0');
+}
```
Jiralite, line 281: There is no capturing group here, so this function effectively returns `$0` always.

Jiralite (changes requested):

packages/discord.js/src/util/Util.js
```diff
@@ -65,6 +65,11 @@ function flatten(obj, ...props) {
  * @property {boolean} [spoiler=true] Whether to escape spoilers or not
  * @property {boolean} [codeBlockContent=true] Whether to escape text inside code blocks or not
  * @property {boolean} [inlineCodeContent=true] Whether to escape text inside inline code or not
+ * @property {boolean} [escape=true] Whether to escape escape characters or not
+ * @property {boolean} [heading=false] Whether to escape headings or not
+ * @property {boolean} [bulletedList=false] Whether to escape bulleted lists or not
+ * @property {boolean} [numberedList=false] Whether to escape numbered lists or not
+ * @property {boolean} [maskedLink=false] Whether to escape masked links or not
```
Jiralite, line 72: Superfluous wording:

```suggestion
 * @property {boolean} [escape=true] Whether to escape escape characters
 * @property {boolean} [heading=false] Whether to escape headings
 * @property {boolean} [bulletedList=false] Whether to escape bulleted lists
 * @property {boolean} [numberedList=false] Whether to escape numbered lists
 * @property {boolean} [maskedLink=false] Whether to escape masked links
```

packages/discord.js/typings/index.d.ts
Jiralite: This is fine.

packages/discord.js/src/util/Util.js
imranbarbhuiya: adding a test for all these functions is needed ig?

packages/discord.js/src/util/Util.js
RedGuy12: This is for consistency with the above lines. Should I remove "or not" from those as well?

Jiralite: Yeah, sure.

kyranet (approved):

iCrawl (approved):

vladfrangu (approved):

Issue #8719, opened 2022-10-07
Error loading @discordjs/ws package on latest dev version


SammyWhamy:
### Which package is this bug report for?

ws

### Issue description

1. install `@discordjs/ws@0.3.1-dev.1665144455-9f68d73.0`
2. Import the package
3. Observe crash

### Code sample

```typescript
const ws = require("@discordjs/ws") // ESM Import has the same issue
```

### Package version

0.3.1-dev.1665144455-9f68d73.0

### Node.js version

16.13.0

### Operating system

Windows 10

### Priority this issue should have

Medium (should be fixed soon)

### Which partials do you have configured?

Not applicable (subpackage bug)

### Which gateway intents are you subscribing to?

Not applicable (subpackage bug)

### I have tested this issue on a development release

9f68d73

Jiralite:
What's the crash?

SammyWhamy:
Issue seems to originate from f79ccb4
Crash is:
```
node:fs:585
  handleErrorFromBinding(ctx);
  ^

Error: ENOENT: no such file or directory, open 'C:\project\node_modules\@discordjs\package.json'
    at Object.openSync (node:fs:585:3)
    at readFileSync (node:fs:453:35)
    at file:///C:/project/node_modules/@discordjs/ws/dist/index.mjs:283:19
    at ModuleJob.run (node:internal/modules/esm/module_job:185:25)
    at async Promise.all (index 0)
    at async ESMLoader.import (node:internal/modules/esm/loader:281:24)
    at async loadESM (node:internal/process/esm_loader:88:5)
    at async handleMainPromise (node:internal/modules/run_main:65:12) {
  errno: -4058,
  syscall: 'open',
  code: 'ENOENT',
  path: 'C:\\project\\node_modules\\@discordjs\\package.json'
}
```

PR #8720, opened 2022-10-07
fix: ws package.json path

imranbarbhuiya:
fixes: #8719 
After the build, `util/constants` files get bundled and become `dist/index`.  So removing a `..`  is the proper fix but that will fail the ci. So I disabled bundling in `/ws` package.

### DIFF
--- a/packages/ws/src/utils/constants.ts
+++ b/packages/ws/src/utils/constants.ts
@@ -20,7 +20,7 @@ export enum CompressionMethod {
 	ZlibStream = 'zlib-stream',
 }
 
-const packageJson = readFileSync(join(__dirname, '..', '..', 'package.json'), 'utf8');
+const packageJson = readFileSync(join(__dirname, '..', 'package.json'), 'utf8');
 const Package = JSON.parse(packageJson);
 
 export const DefaultDeviceProperty = `@discordjs/ws ${Package.version}`;
--- a/packages/ws/src/utils/constants.ts
+++ b/packages/ws/src/utils/constants.ts
@@ -1,9 +1,8 @@
-import { readFileSync } from 'node:fs';
-import { join } from 'node:path';
 import process from 'node:process';
 import { Collection } from '@discordjs/collection';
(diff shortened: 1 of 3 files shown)


iCrawl:
Would importing it fix this properly?

imranbarbhuiya:
No ig? because the file path will be diff before and after build

iCrawl:
Importing it should inline it

imranbarbhuiya:
> Importing it should inline it

Yeah it works

favna:
> > Importing it should inline it
> 
> Yeah it works

Can confirm

packages/ws/src/utils/constants.ts
```diff
@@ -1,9 +1,8 @@
-import { readFileSync } from 'node:fs';
-import { join } from 'node:path';
 import process from 'node:process';
 import { Collection } from '@discordjs/collection';
 import { lazy } from '@discordjs/util';
 import { APIVersion, GatewayOpcodes } from 'discord-api-types/v10';
+import packageJSON from '../../package.json';
```
almeidx, line 5:
```suggestion
import { version } from '../../package.json';
```

packages/ws/src/utils/constants.ts
```diff
@@ -20,10 +19,7 @@ export enum CompressionMethod {
 	ZlibStream = 'zlib-stream',
 }
 
-const packageJson = readFileSync(join(__dirname, '..', '..', 'package.json'), 'utf8');
-const Package = JSON.parse(packageJson);
-
-export const DefaultDeviceProperty = `@discordjs/ws ${Package.version}`;
+export const DefaultDeviceProperty = `@discordjs/ws ${packageJSON.version}`;
```
almeidx, line 22:
```suggestion
export const DefaultDeviceProperty = `@discordjs/ws ${version}`;
```

almeidx (changes requested):
This will inline the version itself instead of the whole package.json

iCrawl (approved):

PR #8735, opened 2022-10-11
refactor(Embed): use `embedLength` function from builders

almeidx:
Uses the [`embedLength()`](https://discordjs.dev/docs/packages/builders/main/embedLength:Function) function from /builders in `Embed#length` to reduce code duplication

### DIFF
--- a/packages/discord.js/src/structures/Embed.js
+++ b/packages/discord.js/src/structures/Embed.js
@@ -1,5 +1,6 @@
 'use strict';
 
+const { embedLength } = require('@discordjs/builders');
 const isEqual = require('fast-deep-equal');
 
 /**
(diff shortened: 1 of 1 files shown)


kyranet (approved):

packages/discord.js/src/structures/Embed.js
```diff
@@ -181,13 +182,7 @@ class Embed {
...
    */
   get length() {
-    return (
-      (this.data.title?.length ?? 0) +
-      (this.data.description?.length ?? 0) +
-      (this.data.fields?.reduce((prev, curr) => prev + curr.name.length + curr.value.length, 0) ?? 0) +
-      (this.data.footer?.text.length ?? 0) +
-      (this.data.author?.name.length ?? 0)
-    );
+    return embedLength(this.data);
```
imranbarbhuiya, line 185: imo this alias should be removed and `embedLength` should be used by the user. So this method should be deprecated. 
Also `length` prop in builder does make sense to check length before adding more fields but not not in received data

imranbarbhuiya (changes requested):

iCrawl (approved):

PR #8748, opened 2022-10-14
fix(BitField): remove hasParams parameters

almeidx:
`has()` never had any more parameters
https://github.com/discordjs/discord.js/blob/bf9aa1858dab2e1bca3be390ce2392b99d208dbf/packages/discord.js/src/util/BitField.js#L59-L62

Also cleaned up the `serialize()` method

### DIFF
--- a/packages/discord.js/src/util/BitField.js
+++ b/packages/discord.js/src/util/BitField.js
@@ -113,9 +113,7 @@ class BitField {
    * @returns {Object}
    */
   serialize() {
-    const serialized = {};
-    for (const [flag, bit] of Object.entries(this.constructor.Flags)) serialized[flag] = this.has(bit);
-    return serialized;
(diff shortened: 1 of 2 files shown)


iCrawl (approved):

kyranet:
`BitField` does not use `...hasParams`, however, `PermissionsBitField` (which extends is) does:

https://github.com/discordjs/discord.js/blob/4f3c13628ecaf8d22174a7a1e227bb7b8278dbb5/packages/discord.js/src/util/PermissionsBitField.js#L65-L101

The reason why `...hasParams` exists, is to reduce the amount of method overrides needed for `PermissionsBitField`.

Can you make sure `PermissionsBitField` works 100% as previously with this patch?

almeidx:
This is a breaking change since the `BitField` constructor is public and is documented, meaning one *could* extend it and have 
its functionally break. The only non-breaking change was the `serialize()` refactor, but it's probably not worth it

PR #8754, opened 2022-10-14
test(Range): improvement unit test in range func

juniorpb:
Improvement unit test in Range function with step greater than default


kyranet:
This test feels redundant with the previous one, which also tests the `step` parameter. I'm afraid this PR may not be accepted in the current state.

If you could add tests for structures that don't have any, specially in the new packages, we would appreciate it 🙏🏼

Issue #8486, opened 2022-08-13
Bot randomly exiting process or going offline/unresponsive


NeonWizard:
### Which package is this bug report for?

discord.js

### Issue description

After around 5-6 hours of running, my discord bot decides to randomly exit with an exit code of 0. There is nothing in my code that should cause that, and adding a debug event listener shows the bot failing around a websocket reacquisition near the 6 hour mark.

I have tried running my bot with typescript, I've tried compiling the code to javascript and running that, I've tried packaging it as a docker container and running that locally, I've tried packing it as a docker container and running that on my production server. Every environment and method results in the same issue.

![image](https://user-images.githubusercontent.com/1355586/184507417-9febf81a-cd78-4e3e-ae25-607be769b347.png)

### Bot code
https://github.com/NeonWizard/sockbot-discord

### Logs
```
[WS => Shard 0] Heartbeat acknowledged, latency of 86ms.
[WS => Shard 0] [HeartbeatTimer] Sending a heartbeat.
[WS => Shard 0] Heartbeat acknowledged, latency of 89ms.
[WS => Shard 0] [HeartbeatTimer] Sending a heartbeat.
[WS => Shard 0] Heartbeat acknowledged, latency of 85ms.
[WS => Shard 0] [HeartbeatTimer] Sending a heartbeat.
[WS => Shard 0] Heartbeat acknowledged, latency of 83ms.
[WS => Shard 0] [HeartbeatTimer] Sending a heartbeat.
[WS => Shard 0] Heartbeat acknowledged, latency of 84ms.
[WS => Shard 0] [RECONNECT] Discord asked us to reconnect
[WS => Shard 0] [DESTROY]
    Close Code    : 4000
    Reset         : false
    Emit DESTROYED: true
[WS => Shard 0] Clearing the heartbeat interval.
[WS => Shard 0] [WebSocket] Destroy: Attempting to close the WebSocket. | WS State: OPEN      
[WS => Shard 0] [WebSocket] Close: Tried closing. | WS State: CLOSING
[WS => Shard 0] [WebSocket] Adding a WebSocket close timeout to ensure a correct WS reconnect.
        Timeout: 5000ms
[WS => Shard 0] [WebSocket] Clearing the close timeout.
[WS => Shard 0] [WebSocket] Close Emitted: false
[WS => Shard 0] [WebSocket] did not close properly, assuming a zombie connection.
Emitting close and reconnecting again.
[WS => Shard 0] [CLOSE]
[...]

PR #8759, opened 2022-10-16
fix(WebSocketShard): Add RESUMED and READY timeout

gamer0mega:
This PR should solve the issue #8486, where the connection gets stuck on IDENTIFY or RESUME.

It adds new private properties of WebSocketShard: readyDispatchTimeout and resumedDispatchTimeout, as well as 2 more private methods: setReadyDispatchTimeout and setResumedDispatchTimeout.

It also removes one .unref() of a timeout as the process exited on it if that was the only shard.

I agree that the names could be confusing (readyTimeout and readyDispatchTimeout), feel free to change them before merging.

I ran a bot with 4 shards for 2 days and all the shards are still up and working fine, whereas before they stopped reconnecting after a few hours of running.


SpaceEEC (approved):

packages/discord.js/src/client/websocket/WebSocketShard.js
```diff
@@ -559,6 +583,48 @@ class WebSocketShard extends EventEmitter {
...
+        clearTimeout(this.resumedDispatchTimeout);
+        this.resumedDispatchTimeout = null;
+      }
+      return;
+    }
+    this.debug('Setting a RESUMED dispatch timeout for 30s.');
+    this.resumedDispatchTimeout = setTimeout(() => {
+      this.debug('Did not receive RESUMED in time. Destroying and connecting again.');
+      this.destroy({ reset: false, closeCode: 4009 });
+    }, 30_000).unref();
```
kyranet, line 604: I don't think we should call `.unref()` here, as this can potentially cause the process to exit early.

Also, why 30 seconds?

packages/discord.js/src/client/websocket/WebSocketShard.js
```diff
@@ -559,6 +583,48 @@ class WebSocketShard extends EventEmitter {
...
+        clearTimeout(this.readyDispatchTimeout);
+        this.readyDispatchTimeout = null;
+      }
+      return;
+    }
+    this.debug('Setting a READY dispatch timeout for 20s.');
+    this.readyDispatchTimeout = setTimeout(() => {
+      this.debug('Did not receive READY in time. Destroying and connecting again.');
+      this.destroy({ reset: true, closeCode: 4009 });
+    }, 20_000).unref();
```
kyranet, line 625: I don't think we should call `.unref()` here, as this can potentially cause the process to exit early.

Also, why 20 seconds?

kyranet (dismissed):

packages/discord.js/src/client/websocket/WebSocketShard.js
gamer0mega: > why 30 seconds?

As the RESUMED event is sent only after the missed events are replayed

packages/discord.js/src/client/websocket/WebSocketShard.js
gamer0mega: > why 20 seconds?

As the other timeout was 20 seconds. READY should be faster than RESUMED as there are no packets sent in between

packages/discord.js/src/client/websocket/WebSocketShard.js
vladfrangu: This is not a solution. Imagine if you resume and replay thousands of events... you'd end up in a case where the connection gets closed mid resume because of this interval!

gamer0mega: > This is not a solution! Imagine if you resume and replay thousands of events... you'd end up in a case where the connection gets closed mid resume because of this interval!

Fixing this would require either an interval to check the last packet timestamp, or resetting the timeout on every packet received until it resumed

iCrawl (approved):

l1v0n1 (commented):
п

DraftProducts:
This stuff looks great!
What did you test it on? @gamer0mega 
I realized that the issue can be increased when the network speed is low or not proportional to the amount of servers supported.
Did you compare the session start tokens used? If so, is there a significant difference?

I will test this fix on production 400 shards to see if it reduces or fixes the problem on my side 👍

packages/discord.js/src/client/websocket/WebSocketShard.js
```diff
@@ -565,7 +601,53 @@ class WebSocketShard extends EventEmitter {
...
+        clearTimeout(this.resumedDispatchTimeout);
+        this.resumedDispatchTimeout = null;
+      }
+      return;
+    }
+    this.debug('Setting a RESUMED dispatch timeout for 20s.');
+    this.resumedDispatchTimeout = setTimeout(() => {
+      if ((Date.now() - this.lastReplayedAt) < 20_000) {
+        this.debug('Received a message within the last 20s. Delaying RESUMED timeout.');
+        return this.setResumedDispatchTimeout();
```
DraftProducts, line 625: https://eslint.org/docs/latest/rules/consistent-return
```suggestion
        this.setResumedDispatchTimeout();
        return
```

DraftProducts (changes requested):

gamer0mega:
> This stuff looks great!
> What did you test it on? @gamer0mega 
> I realized that the issue can be increased when the network speed is low or not proportional to the amount of servers supported.
> Did you compare the session start tokens used? If so, is there a significant difference?
> 
> I will test this fix on production 400 shards to see if it reduces or fixes the problem on my side 👍

I didn't test it on a large bot, because it was a rewrite of me moving my main bot from eris as it is not supported anymore. I tested it on a combination of mobile data + mobile hotspot + an unstable Linux wi-fi driver, with 4 shards. This is very unstable and I found this issue quickly as the test bot randomly died and never connected back, I had 4 shards spread across 2 processes to make sure the fix worked, I always looked through the debug logs and seeing that it got READY was rare, most of the times it got RESUMED and after 10 days all the shards were up and working before I had to restart my pc. You can check the session starts yourself, because before it died on the second or even the first reconnect and I was unable to check that at all

DraftProducts:
Okay nice work 🔥 
After +10 hours of production test, I seem that I have significative improvements.
Could we implement this fix too ?
https://github.com/discordjs/discord.js/commit/2e1c68e28a3b7bb30d327cb265857d1c0c662a9e

It has a good impact on zombie connections, we just have to change the `reset` value to `false` on L607 to allow a RESUME on a zombie connection as it's documented in the discord api docs :
https://discord.com/developers/docs/topics/gateway#sending-heartbeats
image

DraftProducts (approved):

legendhimself:
I don't think we need this pr anymore since #8989 has already been fixed https://github.com/discordjs/discord.js/issues/8486
Correct me if I am wrong but we don't need the READY timeout since we already have a [timeout](https://discord.js.org/#/docs/discord.js/main/typedef/MultipleShardSpawnOptions) on the ShardingManager

Moreover, we shouldn't be handling resumed and ready event timeouts by ourselves since if any connection goes dead it's going to be picked up by zombie connection code in a few seconds [closeTimeout](https://discord.js.org/#/docs/discord.js/main/typedef/ClientOptions)

gamer0mega:
> I don't think we need this pr anymore since #8989 has fixed #8486 Correct me if I am wrong but we don't need the READY timeout since we already have a [timeout](https://discord.js.org/#/docs/discord.js/main/typedef/MultipleShardSpawnOptions) on the ShardingManager
> 
> Moreover, we shouldn't be handling resumed and ready event timeouts by ourselves since if any connection goes dead it's going to be picked up by zombie connection code in a few seconds [closeTimeout](https://discord.js.org/#/docs/discord.js/main/typedef/ClientOptions)

It can get stuck on RESUMING, where it should never heartbeat until RESUMED, so if it gets stuck there is no way to detect it

Qjuh:
> > I don't think we need this pr anymore since #8989 has fixed #8486 Correct me if I am wrong but we don't need the READY timeout since we already have a [timeout](https://discord.js.org/#/docs/discord.js/main/typedef/MultipleShardSpawnOptions) on the ShardingManager
> > Moreover, we shouldn't be handling resumed and ready event timeouts by ourselves since if any connection goes dead it's going to be picked up by zombie connection code in a few seconds [closeTimeout](https://discord.js.org/#/docs/discord.js/main/typedef/ClientOptions)
> 
> It can get stuck on RESUMING, where it should never heartbeat until RESUMED, so if it gets stuck there is no way to detect it

Can it really get stuck there though? None of the errors in the issue you linked had that as their cause, all were related to reconnecting over and over and then getting stuck because of that.

legendhimself:
> It can get stuck on RESUMING, where it should never heartbeat until RESUMED, so if it gets stuck there is no way to detect it

@gamer0mega If it gets stuck on resuming you need a better host for the bot. This is highly unlikely unless the host has a bad internet connection. Resuming -> Resumed depends on Our Server to Discord Server, if the resumed didn't happen even with a good connection (which is highly unlikely) its a discord issue.

gamer0mega:
> > It can get stuck on RESUMING, where it should never heartbeat until RESUMED, so if it gets stuck there is no way to detect it
> 
> @gamer0mega If it gets stuck on resuming you need a better host for the bot. This is highly unlikely unless the host has a bad internet connection. Resuming -> Resumed depends on Our Server to Discord Server, if the resumed didn't happen even with a good connection (which is highly unlikely) its a discord issue.

It is not a problem with the network, the socket gets disconnected, the library doesnt know that and keeps RESUMING for eternity

legendhimself:
@gamer0mega
Can you provide me with a repro
This repro should work even with a good connection as you just mentioned "It is not a problem with the network"

Jiralite:
Superseded by #9099.

PR #8760, opened 2022-10-16
feat: add `Message#bulkDeletable`

Syjalo:
This PR adds `Message#bulkDeletable`


Jiralite (commented):
I believe this is missing direct message channel checks.

Also, if possible, can you state a use case for this as the bulk delete method we have already filters out ids that cannot be bulk deleted.

packages/discord.js/src/structures/Message.js
```diff
@@ -611,6 +611,24 @@ class Message extends Base {
...
+   * @example
+   * // Filter for bulk deletable messages
+   * channel.bulkDelete(messages.filter(message => message.bulkDeletable));
+   */
+  get bulkDeletable() {
+    const permissions = this.channel?.permissionsFor(this.client.user);
+    return (
+      this.deletable &&
+      this.inGuild() &&
+      permissions.has(PermissionFlagsBits.ManageMessages) &&
```
wasdennnoch, line 627: Due to the optional chaining above, `permissions` is nullable.

jaw0r3k, line 627: But .deletable already checks that

Syjalo: It's nullable if the message from a DM. In guild it always will be there since the guild member of the client is cached.

vladfrangu (approved):

kyranet (changes requested):
Same comments as with https://github.com/discordjs/discord.js/pull/8761#pullrequestreview-1158057143, I'll approve once those requests are resolved on both PRs.

packages/discord.js/src/util/Constants.js
```diff
@@ -2,6 +2,12 @@
 
 const { ChannelType, MessageType } = require('discord-api-types/v10');
 
+/**
+ * Max bulk deletable message age
+ * @typedef {number} MaxBulkDeletableMessageAge
```
Jiralite, line 7: Did you forget to add this to the `@typedef` on line 123?

iCrawl (approved):

kyranet (approved):

PR #8761, opened 2022-10-16
feat: add `Message#bulkDeletable` (v13)

Syjalo:
Backports #8760 to version 13

### DIFF
--- a/src/structures/Message.js
+++ b/src/structures/Message.js
@@ -639,7 +639,8 @@ class Message extends Base {
       this.inGuild() &&
       Date.now() - this.createdTimestamp < MaxBulkDeletableMessageAge &&
       this.deletable &&
-      permissions?.has(Permissions.FLAGS.MANAGE_MESSAGES, false)
+      permissions?.has(Permissions.FLAGS.MANAGE_MESSAGES, false) ??
+      false
     );
   }
 
--- a/src/structures/Message.js
+++ b/src/structures/Message.js
@@ -633,8 +633,14 @@ class Message extends Base {
    * // Filter for bulk deletable messages
    * channel.bulkDelete(messages.filter(message => message.bulkDeletable));
    */
-   get bulkDeletable() {
-    return this.deletable && this.inGuild() && Date.now() - this.createdTimestamp < 1_209_600_000;
+  get bulkDeletable() {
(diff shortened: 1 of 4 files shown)


vladfrangu (approved):

src/structures/Message.js
```diff
@@ -625,6 +625,24 @@ class Message extends Base {
...
+   * channel.bulkDelete(messages.filter(message => message.bulkDeletable));
+   */
+  get bulkDeletable() {
+    const permissions = this.channel?.permissionsFor(this.client.user);
+    return (
+      this.deletable &&
+      this.inGuild() &&
+      permissions.has(Permissions.FLAGS.MANAGE_MESSAGES) &&
+      Date.now() - this.createdTimestamp < 1_209_600_000
+    );
```
kyranet, line 643: I would rearrange those checks in order of expensiveness:

- `this.inGuild()`
- `Date.now() - this.createdTimestamp < 1_209_600_000` (please make the magic number a constant)
- `this.deletable`
- `this.channel?.permissionsFor(this.client.user).has(Permissions.FLAGS.MANAGE_MESSAGES) ?? false`

> **Note**: Keep in mind that I'm not doing `permissionsFor` until very late in the chain, that's because we read it only once at the very end of the chain, so it's best to call it once we need to use it, rather than very early in the chain.

For reference, the last check is so `bulkDeletable` doesn't return `undefined`, however, because of the following line in `Message#deletable`:

https://github.com/discordjs/discord.js/blob/c312da795e5a1039a2174a5a36f462b8e30f4c61/src/structures/Message.js#L612-L613

You can pretty much assume `this.channel` won't be undefined by this point, simplifying the last line to:

- `this.channel.permissionsFor(this.client.user).has(Permissions.FLAGS.MANAGE_MESSAGES)`

Dropping both the optional chaining property and the default.

---

That being said, can you confirm this works with `MANAGE_MESSAGES` off but `ADMINISTRATOR` on? `Message#deletable` has a check specifically for this, but we also are dropping the `checkAdmin` in this method (which the other getter defines with `false`).

kyranet (changes requested):

Syjalo:
You need to merge this manually

PR #8762, opened 2022-10-17
refactor(InteractionResponses): make `InteractionResponses.followUp()` unusable after deferring

Idris1401:

I think it's better to make `InteractionResponses.followUp()` unusable after deferring.
*I'm not sure if I did it correctly. I wasn't sure if it was really needed to create another error code and if the error code's message is correctly written. And also I didn't touch at types or tests because firstly I think they don't need to get changed and secondly, I just don't know how to edit them.* ._.


monbrey:
This isn't correct though. In the Discord API you _can_ send a followUp after a defer. 

discord.js should not be opinionated and enforce a different standard.

Idris1401:
> This isn't correct though. In the Discord API you _can_ send a followUp after a defer. 
> 
> 
> 
> discord.js should not be opinionated and enforce a different standard.

I see. Thanks for the explanation!

PR #8770, opened 2022-10-20
Check for github action updates weekly

jlosito:

This will use dependabot to check any github action this project uses for updates and submit pull requests with version bumps in order to keep dependencies up-to-date.

https://docs.github.com/en/code-security/dependabot/working-with-dependabot/keeping-your-actions-up-to-date-with-dependabot

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,7 @@
+---
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"


imranbarbhuiya:
Does depandabot supports yarn v3?

jlosito:
@imranbarbhuiya According to this, it only supports v1.

imranbarbhuiya:
This repo uses yarn v3 so depandabot can't be used

jlosito:
This change would only check against the github action dependencies which this project uses, not the javascript libraries. Are you saying that you're not interested in dependabot since it doesn't handle all of the dependency management which this project uses?

jlosito:
@imranbarbhuiya dependabot can now support v3 starting today. A coincidence.

https://github.blog/changelog/2022-10-20-dependabot-can-now-generate-security-and-version-updates-for-yarn-v2-and-v3/

PR #8772, opened 2022-10-20
Check for dependency updates weekly

jlosito:

The pull request below was closed due to dependabot not supporting yarn v3. Now it does starting today. Therefore, I'm re-opening the pull request.

https://github.com/discordjs/discord.js/pull/8770
https://github.blog/changelog/2022-10-20-dependabot-can-now-generate-security-and-version-updates-for-yarn-v2-and-v3/

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,10 @@
+version: 2
+updates:
+  - package-ecosystem: github-actions
+    directory: /
+    schedule:
+      interval: weekly
+  - package-ecosystem: npm
+    directory: /
+    schedule:
+      interval: weekly


Syjalo (changes requested):
Use the same style as in other `.yml` files.

kyranet:
Since nobody has done so yet, I'll voice my concerns with Dependabot.

1. We used to use Dependabot in the past, but the massive PR spam it caused cause more of a burden than of a help.
    1. Schedules kind of help with this, however, Dependabot massively lags behind the competency, as it lacks of one fundamental feature: [grouped updates](https://github.com/dependabot/dependabot-core/issues/1190).
1. Likewise in the past, `discord-api-types` (and other dependencies) get a number of breaking changes that need us to intervene and do the necessary code changes to make the branch work.
    1. Contributors need to create a PR pointing to another PR to do said changes. This can limit the contributing factor as not everyone is well versed with contributing to non-main branches.
1. Dependabot re-creates PRs rather than updating them with a force-push, this makes the previous step harder as there's a chance it'll just drop all the progress because a new version just released.
1. I personally cannot trust in their capability to bring required updates for our tooling: Yarn v2 has been out for over 2 years, and Yarn v3 for over a year. However, they released compatibility with both of them just this week.
    1. Gets even more troublesome because Yarn v4 is releasing soon, and we will likely want to switch to it soon after it does, as it may bring improved workspace tools besides even better performance.
    1. If Dependabot fails to bring Yarn v4 support by then, we will be forced to either:
        1. Drop Dependabot so we can use Yarn v4.
        1. Keep Dependabot and continue using Yarn v3 against our needs.
1. Automatic PRs from Dependabot need manual reviews from at least 3 of the core developers (or @iCrawl's).

I would say [Renovate](https://www.mend.io/free-developer-tools/renovate/) would stand a much better chance at being approved for usage, as it supports scheduled grouped updates (solves `1.`), re-uses PRs (solves `3.`), added Yarn v2/v3 support very quickly (solves `4.`), and has auto-approval bots that trigger on CI status (solves `5.`), and has very fast and effective technical support, however, it (as well as any other tool) still fails to solve `2.`.

PR #8774, opened 2022-10-20
feat: allow deletion of ephemeral messages

monbrey:
* https://github.com/discord/discord-api-docs/pull/5570

Breaking because it no longer throws an error, but a bugfix because it shouldn't throw an error here anymore.

### DIFF
--- a/packages/discord.js/src/errors/ErrorCodes.js
+++ b/packages/discord.js/src/errors/ErrorCodes.js
@@ -128,6 +128,7 @@
  * @property {'InteractionAlreadyReplied'} InteractionAlreadyReplied
  * @property {'InteractionNotReplied'} InteractionNotReplied
  * @property {'InteractionEphemeralReplied'} InteractionEphemeralReplied
+ * <warn>This property is deprecated.</warn>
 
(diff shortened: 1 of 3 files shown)


SpaceEEC (approved):

kyranet (commented):
I'd argue that it's still https://github.com/discordjs/discord.js/labels/semver:major as-is because you're removing a property from `ErrorCodes`, which is a public and documented interface.

SpaceEEC:
It should probably be deprecated instead then. 😅

monbrey:
> I'd argue that it's still [ semver:major](https://github.com/discordjs/discord.js/labels/semver%3Amajor) as-is because you're removing a property from `ErrorCodes`, which is a public and documented interface.

I agree in theory and I'm happy to add it back, but the error can't be thrown by anything anymore.

kyranet:
That is correct, but functions that read that property (e.g. error handlers) will break with this PR, that's why I consider it a breaking change.

Deprecating it is fine, however.

vladfrangu (approved):

kyranet (approved):

Issue #8757, opened 2022-10-16
djs v14 silently fails if initial gateway connection is ratelimited


hackermondev:
### Which package is this bug report for?

discord.js

### Issue description

Discord.js v14 silents fails if initial gateway connection API endpoint (``/api/v9/gateway/bot``) is ratelimited.

Steps to reproduce:
1. connect to a server where the ip is currently ratelimited
```
{"code": 0, "message": "You are being blocked from accessing our API temporarily due to exceeding our rate limits frequently. Please read our docs at https://discord.com/developers/docs/topics/rate-limits to prevent this moving forward."}
```
2. run a basic bot

**Expected Result:** It should throw an error or print something to the console
**Actual Result:** the issue here is that discord.js doesn't return anything to the user and the program just ends abruptly.
**Note:** On discord.js v13, it doesn't print anything to the console either (unless you're logging ``debug`` events) but it waits out the timeout. Right now, djs v14 doesn't debug anything and it doesn't print anything to the console when its ratelimited (it just closes the program)

GIF EXAMPLE:
![OnUc3PL7o7](https://user-images.githubusercontent.com/60828015/196013347-2c79bb7d-7823-4cca-bdee-91bd2664f8be.gif)

### Code sample

```typescript
const { Client, GatewayIntentBits } = require('discord.js');
const client = new Client({ intents: [GatewayIntentBits.Guilds] });

client.on('ready', () => {
  console.log(`Logged in as ${client.user.tag}!`);
});

client.on('interactionCreate', async interaction => {
  if (!interaction.isChatInputCommand()) return;

  if (interaction.commandName === 'ping') {
    await interaction.reply('Pong!');
  }
});

client.login(process.env['DISCORD_TOKEN']);
```

### Package version

14.6.0

### Node.js version

16.15.0

### Operating system

Linux

### Priority this issue should have

Low (slightly annoying)

### Which partials do you have configured?

No Partials

### Which gateway intents are you subscribing to?

Guilds, GuildMembers, GuildMessages, MessageContent

didinele:
It doesn't "fail", it waits for the ratelimit and will eventually connect. This is intended default behavior. If you would like for it to throw, you can use the `rejectOnRateLimit` REST option.

hackermondev:
> It doesn't "fail", it waits for the ratelimit and will eventually connect. This is intended default behavior. If you would like for it to throw, you can use the `rejectOnRateLimit` REST option.

@didinele 
[...]

PR #8779, opened 2022-10-22
fix: make ratelimit timeout require event loop to be active

hackermondev:
related to #8757

The ``ref`` option doesn't require the Node.js event loop to remain active when waiting for the timeout causing the program to end if there isn't an active gateway connection.

### DIFF
--- a/packages/rest/src/lib/handlers/SequentialHandler.ts
+++ b/packages/rest/src/lib/handlers/SequentialHandler.ts
@@ -460,7 +460,7 @@ export class SequentialHandler implements IHandler {
 
 				this.#sublimitPromise?.resolve();
 				this.#sublimitPromise = null;
-				await sleep(sublimitTimeout, undefined);
+				await sleep(sublimitTimeout);
 				let resolve: () => void;
 				// eslint-disable-next-line promise/param-names, no-promise-executor-return
 				const promise = new Promise<void>((res) => (resolve = res));
--- a/packages/rest/src/lib/handlers/SequentialHandler.ts
+++ b/packages/rest/src/lib/handlers/SequentialHandler.ts
@@ -137,7 +137,7 @@ export class SequentialHandler implements IHandler {
 	 * @param time - The amount of time to delay all requests for
 	 */
 	private async globalDelayFor(time: number): Promise<void> {
-		await sleep(time, undefined, { ref: false });
+		await sleep(time, undefined);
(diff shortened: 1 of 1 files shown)


packages/rest/src/lib/handlers/SequentialHandler.ts
```diff
@@ -137,7 +137,7 @@ export class SequentialHandler implements IHandler {
 	 * @param time - The amount of time to delay all requests for
 	 */
 	private async globalDelayFor(time: number): Promise<void> {
-		await sleep(time, undefined, { ref: false });
+		await sleep(time, undefined);
```
Jiralite, line 140:
```suggestion
		await sleep(time);
```

packages/rest/src/lib/handlers/SequentialHandler.ts
```diff
@@ -460,7 +460,7 @@ export class SequentialHandler implements IHandler {
 
 				this.#sublimitPromise?.resolve();
 				this.#sublimitPromise = null;
-				await sleep(sublimitTimeout, undefined, { ref: false });
+				await sleep(sublimitTimeout, undefined);
```
Jiralite, line 463:
```suggestion
				await sleep(sublimitTimeout);
```

Jiralite (changes requested):

didinele (approved):

vladfrangu (approved):

ckohen (approved):
Adding context:
These were initially added so that a client that was .destroy()ed would not continue waiting for requests.

I think the chances of hitting a rate limit near the time of calling such method are actually quite low AND in the cases where destroy is called at all (cause lets be honest most people just kill the process) there will likely be additional cleanup happening anyways that takes some time.

Given that, I am okay removing these unrefs.

Worth considering whether this is really patch or major, because hitting cloudflare ratelimits on startup will have vastly different behavior now and this was an intentional design decision originally.

didinele:
> Worth considering whether this is really patch or major, because hitting cloudflare ratelimits on startup will have vastly different behavior now and this was an intentional design decision originally.

Considering its only now intended behavior, this should def. be a patch IMO. For most people the behavior won't even change - i.e. if they also have a postgres/redis/whatever connection open, which is probably why we didn't find this issue earlier.

kyranet (approved):

SpaceEEC (approved):

PR #8786, opened 2022-10-25
feat(ThreadChannel): add a helper for pin and unpin

Idris1401:

With the help of @monbrey and @almeidx.


almeidx (changes requested):

packages/discord.js/src/structures/ThreadChannel.js
```diff
@@ -456,6 +456,27 @@ class ThreadChannel extends BaseChannel {
   setAppliedTags(appliedTags, reason) {
     return this.edit({ appliedTags, reason });
   }
+  
+  /**
+   * Pins this thread channel from the forum channel.
```
Jiralite, line 461: Consistent suggestion as "thread channel" is written twice in the documentation:
```suggestion
   * Pins this thread from the forum channel.
```

packages/discord.js/src/structures/ThreadChannel.js
```diff
@@ -456,6 +456,27 @@ class ThreadChannel extends BaseChannel {
...
+   * @param {string} [reason] Reason for pinning
+   * @returns {Promise<ThreadChannel>}
+   */
+   pin(reason) {
+    if (this.parent.type !== ChannelType.ForumChannel) throw new DiscordjsError(ErrorCodes.CannotPinThreadOutOfForumChannel);
+    return this.edit({ flags: [ChannelFlags.Pinned], reason });
+  }
+  
+  /**
+   * Unpins this thread channel from the forum channel.
```
Jiralite, line 471: Consistent suggestion as "thread channel" is written twice in the documentation:
```suggestion
   * Unpins this thread from the forum channel.
```

packages/discord.js/src/structures/ThreadChannel.js
```diff
@@ -456,6 +456,27 @@ class ThreadChannel extends BaseChannel {
...
   }
+  
+  /**
+   * Pins this thread channel from the forum channel.
+   * @param {string} [reason] Reason for pinning
+   * @returns {Promise<ThreadChannel>}
+   */
+   pin(reason) {
+    if (this.parent.type !== ChannelType.ForumChannel) throw new DiscordjsError(ErrorCodes.CannotPinThreadOutOfForumChannel);
+    return this.edit({ flags: [ChannelFlags.Pinned], reason });
```
monbrey, line 467: My bad on this one since I gave you this line - this would remove `ChannelFlags.RequireTag` if it were currently set (and any other future tags. Should do something similar to how you handle remove.

Jiralite (changes requested):

packages/discord.js/src/structures/ThreadChannel.js
```diff
@@ -457,6 +457,24 @@ class ThreadChannel extends BaseChannel {
     return this.edit({ appliedTags, reason });
   }
 
+  /**
+   * Pins this thread from the forum channel.
+   * @param {string} [reason] Reason for pinning
+   * @returns {Promise<ThreadChannel>}
+   */
+  pin(reason) {
+    return this.edit({ flags: this.flags.add(ChannelFlags.Pinned), reason });
```
almeidx, line 466:
```suggestion
    return this.edit({ flags: this.flags.bitfield | ChannelFlags.Pinned, reason });
```

packages/discord.js/src/structures/ThreadChannel.js
```diff
@@ -457,6 +457,24 @@ class ThreadChannel extends BaseChannel {
...
+  }
+
+  /**
+   * Unpins this thread from the forum channel.
+   * @param {string} [reason] Reason for unpinning
+   * @returns {Promise<ThreadChannel>}
+   */
+  unpin(reason) {
+    return this.edit({ flags: this.flags.remove(ChannelFlags.Pinned), reason });
```
almeidx, line 475:
```suggestion
    return this.edit({ flags: this.flags.bitfield & ~ChannelFlags.Pinned, reason });
```

almeidx (changes requested):
Actually, probably shouldn't be mutating the flags bit field like that (in case of failure, etc)
Something like this should do the trick

packages/discord.js/src/structures/ThreadChannel.js
muchnameless: this.flags is frozen so it wouldn't be mutated

packages/discord.js/typings/index.d.ts
```diff
@@ -2640,6 +2640,8 @@ export class ThreadChannel<Forum extends boolean = boolean> extends TextBasedCha
   public setLocked(locked?: boolean, reason?: string): Promise<AnyThreadChannel>;
   public setName(name: string, reason?: string): Promise<AnyThreadChannel>;
   public setAppliedTags(appliedTags: Snowflake[], reason?: string): Promise<ThreadChannel<true>>;
+  public pin(reason?: string): Promise<ThreadChannel<true>>;
+  public unpin(reason?: string): Promise<ThreadChannel<true>>;
```
kyranet, line 2644: Since those methods return `this.edit()`, which is has a return type of `Promise<AnyThreadChannel>` (the other methods with the exception of `setAppliedTags`, for some reason, also follow this), the two new methods should also follow for consistency:
```suggestion
  public pin(reason?: string): Promise<AnyThreadChannel>;
  public unpin(reason?: string): Promise<AnyThreadChannel>;
```

kyranet (changes requested):

Idris1401:
@kyranet do I edit types for `<ThreadChannel>.setAppliedTags()`?

Idris1401:
> @kyranet do I edit types for `<ThreadChannel>.setAppliedTags()`?

ah you said except that one mb

kyranet:
I personally don't mind, but it's also out of this PR's scope, feel free to create a new one. This PR is fine as-is.

kyranet (approved):

Idris1401:
should I add a *note* saying that only forum threads can be pinned?
as `.setAppliedTags()` has

kyranet:
Good catch, please do so @Idris1401.

Idris1401:
also isn't this why `.setAppliedTags()` returns `Promise<ThreadChannel<true>>` and not `Promise<AnyThreadChannel>` ?

iCrawl (approved):

vladfrangu (approved):

PR #8790, opened 2022-10-27
fix(Transfomers): call `.toJSON` in `toSnakeCase`

### LINKED ISSUE #8742: Non-recursive InteractionResponses#showModal body transformation

discord.js

### Issue description

Basically, the typings do allow me to pass `ModalComponentData` into showModal method, which itself takes `JSONEncodable<APIActionRowComponent<APIModalActionRowComponent>> | ActionRowData<ModalActionRowComponentData>`, **though doesn't handle toJSON method**.

As turns out, in `InteractionResponses#showModal` for `body.data` transformation, this code does the thing:
```js
data: isJSONEncodable(modal) ? modal.toJSON() : this.client.options.jsonTransformer(modal)
```

So if you pass non-`JSONEncodable` modal data, it doesn't have `toJSON` method in it, which makes it to use default jsonTransformer - `toSnakeCase`, which doesn't use `toJSON` method in the child properties.

Would be good to either forbid to use `JSONEncodable` components in non-`JSONEncodable` modal data, or fix it in `InteractionResponses.showModal` method

### Code sample

```typescript
import {
  type ButtonInteraction,
  ActionRowBuilder,
  ComponentType,
  TextInputBuilder
} from "discord.js";

/* Just in case */
function showModal(interaction: ButtonInteraction) {
  interaction.showModal({
    customId: "whatever",
    title: "Example Modal",
    components: [
      new ActionRowBuilder({
        /* ... - doesn't really matter */
      })
    ]
  });

  /* second case */
  interaction.showModal({
    customId: "second",
    title: "Another Example Modal",
    components: [
      {
        type: ComponentType.ActionRow,
        components: [
          new TextInputComponent({
            /* ... */
          })
        ]
      }
    ]
  });
}
```

### Package version

14.6.0

### Node.js version

18.9.0

### Operating system

win32

Low (slightly annoying)

No Partials

No Intents

_No response_

RedGuy12:
Makes `toSnakeCase` call `.toJSON` functions if they exist.
Fixes #8742

I could make calling `.toJSON` togglable via a boolean parameter if that is preferred to avoid breaking changes.


packages/discord.js/src/util/Transformers.js
```diff
@@ -9,6 +10,7 @@ const snakeCase = require('lodash.snakecase');
  */
 function toSnakeCase(obj) {
   if (typeof obj !== 'object' || !obj) return obj;
+  if (isJSONEncodable(obj)) return obj.toJSON();
```
Jiralite, line 13: I feel like this check should be done beneath line 14 where it checks `instanceof Date` as that I believe is JSON-encodeable, no?

RedGuy12: I was not aware of that. Fixed.

kyranet (approved):
This makes the bold assumption that any JSON encodable will always yield snake_case properties, but that's (hopefully) almost always the case here, and for the places we use `toSnakeCase` for, it would be wrong to use JSON encodable classes that don't yield snake_case properties.

Either way, this seems to be triggered when using builders inside plain objects, which makes sense.

RedGuy12:
Would it be better if I called `toSnakeCase` on `toJSON`'s return value?

suneettipirneni:
> Would it be better if I called `toSnakeCase` on `toJSON`'s return value?

I would personally prefer this. 

Splitting functions by specific behaviors yields more flexible predictable code. `toSnakeCase` should only be concerned with transforming the input (as-is) to snake_case json, it shouldn't do anything else in addition to that.

vladfrangu (approved):

PR #8775, opened 2022-10-21
New select menus

jaw0r3k:
**Please describe the changes this PR makes and why it should be merged:**
New Selects
 Collaborating with @RedGuy12 

**Status and versioning classification:**
- [discord-api-types@0.37.15](https://github.com/discordjs/discord-api-types/releases/tag/0.37.15)
- discordjs/discord-api-types#602

PR #8793, opened 2022-10-28
feat: new select menus

### LINKED ISSUE #8753: Add new component types for User, Role, Mentionable, and Channel Select

discord.js

### Feature

Currently discord.js only supports Component Types 1 through 4: Action Row, Button, String Select, and Text Input. Discord added 4 more types, 5 through 8: User Select, Role Select, Mentionable Select, and Channel Select. They function nearly identically to String Select (Component Type 3), except they don't take options and they return not only a value but also a resolved object.

https://discord.com/developers/docs/interactions/message-components#component-object-component-types

https://discord.com/developers/docs/interactions/message-components#select-menu-types

Add the 4 new select menu types.

_No response_

### Other context

_No response_

didinele:
As discussed internally (see: https://canary.discord.com/channels/222078108977594368/999796393810067607/1035493023162830848), supersedes #8775

Things this PR does better:
- Makes 0 breaking changes; in case of sensical class/method renames, re-exports/aliases were added under the old names with deprecation notices on them
- After some discussion with other contributors, has all of the caching behavior set in stone - i.e. consistent with how slash commands behave
- Cleaner/finished typings for the main lib
- Split `SelectMenuComponent` into multiple classes
- ... few other things that no longer come to mind after the 3 hours this took

Closes #8753

Note: There is one caveat with the typings I wrote, `AnySelectMenuInteraction` is a confusing type name and something we moved away from in the past. Ideally this would just be named `SelectMenuInteraction` and be a union of all the possible select menu interaction classes, but this is not possible, as `SelectMenuInteraction` is the name of the old `StringSelectMenuInteraction` and is re-exported as such to prevent this PR from being breaking.


jaw0r3k:
Mhh

Qjuh:
> Its just copied lol

Did you actually read the bullet point list of five things this PR does better?

jaw0r3k:
> > Its just copied lol
> 
> Did you actually read the bullet point list of five things this PR does better?

The only thing i see better is 
- Cleaner/finished typings for the main lib

Syjalo (changes requested):
Add `BaseInteraction#isAnySelectMenu()` that will be changed to `BaseInteraction#isSelectMenu()` in v15

didinele:
> Add `BaseInteraction#isAnySelectMenu()` that will be changed to `BaseInteraction#isSelectMenu()` in v15

Not a bad idea, will do tomorrow.

packages/discord.js/src/structures/MentionableSelectMenuInteraction.js
```diff
@@ -0,0 +1,55 @@
...
+     * Collection of the selected users
+     * @type {Collection<Snowflake, User>}
+     */
+    this.users = new Collection();
+
+    /**
+     * Collection of the selected users
+     * @type {Collection<Snowflake, GuildMember|APIGuildMember>}
+     */
+    this.members = new Collection();
```
Syjalo, line 26: Shouldn't this be `null` if the interaction is from a dm based channel?

packages/discord.js/src/structures/MentionableSelectMenuInteraction.js
```diff
@@ -0,0 +1,55 @@
...
+
+    /**
+     * Collection of the selected roles
+     * @type {Collection<Snowflake, Role|APIRole>}
+     */
+    this.roles = new Collection();
```
Syjalo, line 32: ditto

packages/discord.js/src/structures/UserSelectMenuInteraction.js
```diff
@@ -0,0 +1,38 @@
...
+     * Collection of the selected users
+     * @type {Collection<Snowflake, User>}
+     */
+    this.users = new Collection();
+
+    /**
+     * Collection of the selected users
+     * @type {Collection<Snowflake, GuildMember|APIGuildMember>}
+     */
+    this.members = new Collection();
```
Syjalo, line 24: ditto

Syjalo (changes requested):

packages/discord.js/src/structures/MentionableSelectMenuInteraction.js
didinele: Unsure, this was also a dispute in the old PR - I opted t make it never null. cc @vladfrangu

Syjalo: If you want access the members and the interaction is from a dm based channel you get `GuildMember | APIGuildMember | undefined`. Wouldn't it be better to just make it `null` for dm based channels. You should use `.inGuild()` type guard anyway.

didinele: Type-wise you'll always get `GuildMember | APIGuildMember | undefined` anyway, `Collection#get` returns `V | undefined`.

Syjalo: Yes. I meant it looks like it is possible to get a `GuildMember` if the interaction is from a dm based channel, but that's not really the case.

packages/discord.js/typings/index.d.ts
```diff
@@ -615,7 +636,31 @@ export class SelectMenuBuilder extends BuilderSelectMenuComponent {
   public override setOptions(
     ...options: RestOrArray<BuildersSelectMenuOption | SelectMenuComponentOptionData | APISelectMenuOption>
   ): this;
-  public static from(other: JSONEncodable<APISelectMenuComponent> | APISelectMenuComponent): SelectMenuBuilder;
+  public static from(other: JSONEncodable<APISelectMenuComponent> | APISelectMenuComponent): StringSelectMenuBuilder;
+}
+
+export { StringSelectMenuBuilder as SelectMenuBuilder };
```
kyranet, line 642: We should add a `@deprecate` tsdoc here.

packages/discord.js/typings/index.d.ts
```diff
@@ -639,16 +684,31 @@ export class TextInputComponent extends Component<APITextInputComponent> {
...
   public get minValues(): number | null;
   public get customId(): string;
   public get disabled(): boolean | null;
+}
+
+export class StringSelectMenuComponent extends BaseSelectMenuComponent<APIStringSelectComponent> {
   public get options(): APISelectMenuOption[];
 }
 
+export { StringSelectMenuComponent as SelectMenuComponent };
```
kyranet, line 700: We should add a `@deprecate` tsdoc here.

packages/discord.js/typings/index.d.ts
```diff
@@ -1550,7 +1610,13 @@ export class BaseInteraction<Cached extends CacheType = CacheType> extends Base
   public isMessageContextMenuCommand(): this is MessageContextMenuCommandInteraction<Cached>;
   public isModalSubmit(): this is ModalSubmitInteraction<Cached>;
   public isUserContextMenuCommand(): this is UserContextMenuCommandInteraction<Cached>;
-  public isSelectMenu(): this is SelectMenuInteraction<Cached>;
+  /** @deprecated */
+  public isSelectMenu(): this is StringSelectMenuInteraction<Cached>;
```
kyranet, line 1614: Missing `isAnySelectMenu()` type guard here

packages/discord.js/typings/index.d.ts
```diff
@@ -2254,22 +2328,113 @@ export class Role extends Base {
...
   public values: string[];
-  public inGuild(): this is SelectMenuInteraction<'raw' | 'cached'>;
-  public inCachedGuild(): this is SelectMenuInteraction<'cached'>;
-  public inRawGuild(): this is SelectMenuInteraction<'raw'>;
+  public inGuild(): this is StringSelectMenuInteraction<'raw' | 'cached'>;
+  public inCachedGuild(): this is StringSelectMenuInteraction<'cached'>;
+  public inRawGuild(): this is StringSelectMenuInteraction<'raw'>;
+}
+
+export { StringSelectMenuInteraction as SelectMenuInteraction };
```
kyranet, line 2349: We should add a `@deprecate` tsdoc here.

kyranet (changes requested):
We should also add type tests for the type narrowing between different select menu types.

packages/discord.js/src/structures/MentionableSelectMenuInteraction.js
kyranet: > Shouldn't this be `null` if the interaction is from a dm based channel?

No, making it `null` would barely serve a benefit here, and it would also drive down a rabbit hole where we would need *even more* type guards.

Besides, `Collection#get` already returns `undefined` on missing entries, so there isn't really much of a need to make the entire thing nullable.

packages/builders/src/components/selectMenu/ChannelSelectMenu.ts
```diff
@@ -0,0 +1,61 @@
...
+
+	/**
+	 * {@inheritDoc ComponentBuilder.toJSON}
+	 */
+	public override toJSON(): APIChannelSelectComponent {
+		customIdValidator.parse(this.data.custom_id);
+
+		return {
+			...this.data,
+		} as APIChannelSelectComponent;
```
JulianVennen, line 59: You seemingly forgot to add the channel type to the JSON data here.
```suggestion
		return {
			...this.data,
			channel_types: this.channel_types
		} as APIChannelSelectComponent;
```

I'm not sure whether this should be null or an empty array if a user doesn't specify channel types. The API Documentation doesn't mention that, but both seem to do the same thing in my testing.

JulianVennen (changes requested):

packages/discord.js/src/structures/SelectMenuBuilder.js
```diff
@@ -1,78 +1,25 @@
 'use strict';
 
-const { SelectMenuBuilder: BuildersSelectMenu, isJSONEncodable, normalizeArray } = require('@discordjs/builders');
-const { toSnakeCase } = require('../util/Transformers');
-const { resolvePartialEmoji } = require('../util/Util');
+const { process } = require('node:process');
+const { StringSelectMenuBuilder } = require('./StringSelectMenuBuilder');
```
JulianVennen, line 4:
```suggestion
const process = require('node:process');
const StringSelectMenuBuilder = require('./StringSelectMenuBuilder');
```

packages/discord.js/src/structures/SelectMenuComponent.js
```diff
@@ -1,64 +1,24 @@
 'use strict';
 
-const Component = require('./Component');
+const { process } = require('node:process');
+const { StringSelectMenuComponent } = require('./StringSelectMenuComponent');
```
JulianVennen, line 4:
```suggestion
const process = require('node:process');
const StringSelectMenuComponent = require('./StringSelectMenuComponent');
```

packages/discord.js/src/structures/SelectMenuInteraction.js
```diff
@@ -1,20 +1,24 @@
 'use strict';
 
-const MessageComponentInteraction = require('./MessageComponentInteraction');
+const { process } = require('node:process');
+const { StringSelectMenuInteraction } = require('./StringSelectMenuInteraction');
```
JulianVennen, line 4:
```suggestion
const process = require('node:process');
const StringSelectMenuInteraction = require('./StringSelectMenuInteraction');
```

packages/discord.js/src/structures/SelectMenuOptionBuilder.js
```diff
@@ -1,50 +1,25 @@
 'use strict';
 
-const { SelectMenuOptionBuilder: BuildersSelectMenuOption, isJSONEncodable } = require('@discordjs/builders');
-const { toSnakeCase } = require('../util/Transformers');
-const { resolvePartialEmoji } = require('../util/Util');
+const { process } = require('node:process');
+const { StringSelectMenuOptionBuilder } = require('./StringSelectMenuOptionBuilder');
```
JulianVennen, line 4:
```suggestion
const process = require('node:process');
const StringSelectMenuOptionBuilder = require('./StringSelectMenuOptionBuilder');
```

JulianVennen (changes requested):
Some of your imports don't work correctly. You're using object destruction here, even though the classes are exported directly.

didinele:
Now also tested.

SpaceEEC (approved):

packages/discord.js/src/structures/MentionableSelectMenuInteraction.js
```diff
@@ -0,0 +1,65 @@
...
+
+    if (members) {
+      for (const [id, member] of Object.entries(members)) {
+        const user = users[id];
+        if (!user) {
+          this.client.emit(
+            Events.Debug,
+            `[MentionableSelectMenuInteraction] Recieved a member without a user, skipping ${id}`,
```
almeidx, line 41:
```suggestion
            `[MentionableSelectMenuInteraction] Received a member without a user, skipping ${id}`,
```

packages/discord.js/src/structures/UserSelectMenuInteraction.js
```diff
@@ -0,0 +1,49 @@
...
+      this.users.set(user.id, this.client.users._add(user));
+    }
+
+    if (data.data.resolved.members) {
+      for (const [id, member] of Object.entries(data.data.resolved.members)) {
+        const user = data.data.resolved.users[id];
+        if (!user) {
+          this.client.emit(
+            Events.Debug,
+            `[UserSelectMenuInteraction] Recieved a member without a user, skipping ${id}`,
```
almeidx, line 37:
```suggestion
            `[UserSelectMenuInteraction] Received a member without a user, skipping ${id}`,
```

almeidx (changes requested):

packages/discord.js/src/structures/UserSelectMenuInteraction.js
```diff
@@ -0,0 +1,49 @@
...
+    super(client, data);
+
+    /**
+     * Collection of the selected users
+     * @type {Collection<Snowflake, User>}
+     */
+    this.users = new Collection();
+
+    /**
+     * Collection of the selected users
```
almeidx, line 22:
```suggestion
     * Collection of the selected members
```

almeidx:
Maybe bump discord-api-types to the latest version (0.37.6)?

didinele:
> Maybe bump discord-api-types to the latest version (0.37.6)?

Feel like that's out of scope for me at this point unless we need something from that version/it has relevant bug fixes.

packages/builders/src/components/selectMenu/ChannelSelectMenu.ts
```diff
@@ -0,0 +1,61 @@
...
+	 * 	.addChannelTypes(ChannelType.GuildText, ChannelType.GuildAnnouncement)
+	 * 	.setMinValues(2)
+	 * ```
+	 */
+	public constructor(data?: Partial<APIChannelSelectComponent>) {
+		const { channel_types = [], ...initData } = data ?? {};
+		super({ ...initData, type: ComponentType.ChannelSelect, channel_types });
+	}
+
+	public addChannelTypes(...types: RestOrArray<ChannelType>) {
```
vladfrangu, line 36: Can this even be any channel or just guild channels?

packages/discord.js/src/structures/BaseInteraction.js
```diff
@@ -1,8 +1,10 @@
 'use strict';
 
+const { deprecate } = require('util');
```
vladfrangu, line 3:
```suggestion
const { deprecate } = require('node:util');
```

packages/discord.js/src/structures/BaseInteraction.js
```diff
@@ -268,12 +270,63 @@ class BaseInteraction extends Base {
     return this.type === InteractionType.MessageComponent && this.componentType === ComponentType.Button;
   }
 
+  // TODO: Replace with isAnySelectMenu in the next major - get rid of isAnySelectMenu
```
vladfrangu, line 273: Ngl I'd rather we removed this `isSelectMenu` and keep `isAnySelectMenu`

packages/discord.js/src/structures/ChannelSelectMenuInteraction.js
```diff
@@ -0,0 +1,25 @@
...
+  constructor(client, data) {
+    super(client, data);
+
+    /**
+     * Collection of the selected channels
+     * @type {Collection<Snowflake, Channel|APIChannel>}
+     */
+    this.channels = new Collection();
+    for (const channel of Object.values(data.data.resolved.channels)) {
+      this.channels.set(channel.id, this.client.channels._add(channel, this.guild) ?? channel);
```
vladfrangu, line 20: should we cache the channels received? I think not, right @discordjs/core?

packages/discord.js/src/structures/MentionableSelectMenuInteraction.js
```diff
@@ -0,0 +1,65 @@
...
+     */
+    this.roles = new Collection();
+
+    if (members) {
+      for (const [id, member] of Object.entries(members)) {
+        const user = users[id];
+        if (!user) {
+          this.client.emit(
+            Events.Debug,
+            `[MentionableSelectMenuInteraction] Received a member without a user, skipping ${id}`,
```
vladfrangu, line 41: Should we skip it? Should we add it with user null? @discordjs/core

packages/discord.js/src/structures/ChannelSelectMenuInteraction.js
didinele: Yes. I did it this way because so do slash commands - space agreed it's sound logic to make it consistent.

packages/discord.js/src/structures/MentionableSelectMenuInteraction.js
didinele: Skipping should be ok, it's actually theoretically impossible to not have the user in `users` - this condition is here just as a last resort sanity check to prevent a hard crash.

Also, from a bug I encountered earlier where I wasn't passing in the user at all, the manager was straight up throwing without it - it relies on `member.user.id` (and potentially some other things).

packages/discord.js/src/structures/BaseInteraction.js
Syjalo: But this would be a breaking change

packages/builders/src/components/selectMenu/ChannelSelectMenu.ts
didinele: Spoke about this internally - see https://github.com/discord/discord-api-docs/issues/5600

For now we'll keep it as-is. If Discord decides this is an issue and fixes it by the time this PR is merged, I'll narrow the types/validation, otherwise we'll do it as a patch once/if they do it.

packages/discord.js/src/structures/MentionableSelectMenuInteraction.js
```diff
@@ -0,0 +1,65 @@
...
+
+        this.members.set(id, this.guild?.members._add({ user, ...member }) ?? { user, ...member });
+      }
+    }
+
+    if (users) {
+      for (const user of Object.values(users)) {
+        this.users.set(user.id, this.client.users._add(user));
+      }
+    }
```
Syjalo, line 55:
```suggestion
    if (users) {
      for (const [id, user] of Object.entries(users)) {
        this.users.set(id, this.client.users._add(user));
        const member = members?.[id];
        if (member) {
          this.members.set(id, this.guild?.members._add({ user, ...member }) ?? { user, ...member });
        }
      }
    }
```
As an alternative to @vladfrangu's suggestion

Syjalo (changes requested):

packages/discord.js/src/structures/MentionableSelectMenuInteraction.js
didinele: In practice this does the same thing my current code does, it doesn't actually address Vlad's review.

Syjalo: > In practice this does the same thing my current code does

But with 1 loop. 2 times faster.

> It doesn't actually address Vlad's review

No skipping, no the Vlad's review

didinele: It.. still very much skips though in the impossible edge case that there's a guild member with no associated user, just that now there's no debug log/explicitness to it.

Syjalo: Discord gives us a promise that the member will be there with the user if the interaction is from a guild as it gives us 
a promise that by route `PATCH: /channels/{channel.id}/messages/{message.id}` we edit a message, not delete it.
Also that case isn't documented so I don't think we need that debug log.

didinele: I'm glad you're so optimistic, but regardless, the log was added under the guidance of fellow contribs/core maintainers, so it's already been decided that it's a necessary sanity check.

packages/discord.js/src/structures/BaseInteraction.js
vladfrangu: Not in this PR! In the future :D

packages/builders/src/components/selectMenu/ChannelSelectMenu.ts
```diff
@@ -0,0 +1,61 @@
...
+
+		(this.data.channel_types ??= []).push(...channelTypesValidator.parse(types));
+		return this;
+	}
+
+	public setChannelTypes(...types: RestOrArray<ChannelType>) {
+		// eslint-disable-next-line no-param-reassign
+		types = normalizeArray(types);
+
+		(this.data.channel_types ??= []).splice(0, this.data.channel_types.length, ...channelTypesValidator.parse(types));
```
vladfrangu, line 47: I love one liners like the next person but please separate these into two different lines

packages/builders/src/components/selectMenu/ChannelSelectMenu.ts
```diff
@@ -0,0 +1,61 @@
...
+	 */
+	public constructor(data?: Partial<APIChannelSelectComponent>) {
+		super({ ...data, type: ComponentType.ChannelSelect });
+	}
+
+	public addChannelTypes(...types: RestOrArray<ChannelType>) {
+		// eslint-disable-next-line no-param-reassign
+		types = normalizeArray(types);
+
+		(this.data.channel_types ??= []).push(...channelTypesValidator.parse(types));
```
vladfrangu, line 39: Ditto

vladfrangu (changes requested):

packages/builders/src/components/selectMenu/ChannelSelectMenu.ts
didinele: Do you mean the `??=` or the validator call?

vladfrangu (approved):

packages/builders/src/components/Assertions.ts
```diff
@@ -68,6 +68,8 @@ export function validateRequiredSelectMenuOptionParameters(label?: string, value
 	labelValueDescriptionValidator.parse(value);
 }
 
+export const channelTypesValidator = s.nativeEnum(ChannelType).array;
```
kyranet, line 71:
```suggestion
export const channelTypesValidator = s.nativeEnum(ChannelType).array.setValidationEnabled(isValidationEnabled);
```

packages/discord.js/src/structures/InteractionCollector.js
```diff
@@ -147,10 +147,17 @@ class InteractionCollector extends Collector {
...
      */
+
     if (this.interactionType && interaction.type !== this.interactionType) return null;
     if (this.componentType && interaction.componentType !== this.componentType) return null;
     if (this.messageId && interaction.message?.id !== this.messageId) return null;
-    if (this.messageInteractionId && interaction.message?.interaction?.id !== this.messageInteractionId) return null;
+    if (
+      this.messageInteractionId &&
+      interaction.message?.interaction?.id &&
+      interaction.message?.interaction?.id !== this.messageInteractionId
```
kyranet, line 157: Previous line makes sure none of those objects are nullish.
```suggestion
      interaction.message.interaction.id !== this.messageInteractionId
```

packages/discord.js/src/structures/MentionableSelectMenuInteraction.js
```diff
@@ -0,0 +1,65 @@
...
+
+/**
+ * Represents a {@link ComponentType.MentionableSelect} select menu interaction.
+ * @extends {MessageComponentInteraction}
+ */
+class MentionableSelectMenuInteraction extends MessageComponentInteraction {
+  constructor(client, data) {
+    super(client, data);
+
+    const { members, users, roles } = data.data.resolved || {};
```
kyranet, line 15:
```suggestion
    const { members, users, roles } = data.data.resolved ?? {};
```

kyranet (changes requested):

packages/discord.js/src/structures/InteractionCollector.js
didinele: Prior to this PR it was the way you're requesting it to be, however I was running into a `null !== undefined` edgecase that was causing it to break.

kyranet (approved):

PR #8803, opened 2022-11-01
Revert 1 main

Mohamadjaber96:


kyranet:
Given that the PR has a semver-major nature and does not justify any of it (please fill the template), while it also has a wide scope reverting several changes at once, I'll close this.

If you think this decision is wrong, please justify the changes and update the PR to have a single scope.

PR #8546, opened 2022-08-22
feat: add `chatInputApplicationCommandMention` formatter

almeidx:
**Please describe the changes this PR makes and why it should be merged:**
- Added `chatInputApplicationCommandMention()` formatted in /builders
- ~~Added `ApplicationCommand#toString()`, which will use the function mentioned above to return the commands mention~~

Not going to add the mention to `ApplicationCommand#toString()` due to not being able to have a consistent return type. (The mentions only apply to Chat Input commands)

Upstream:
- discord/discord-api-docs#5186

**Status and versioning classification:**
- This PR changes the library's interface (methods or parameters added)

PR #8818, opened 2022-11-07
feat(ApplicationCommand): add `toString()`

Squiddleton:
Adds `ApplicationCommand#toString()` in the format of `</name:id>` mentions.  Please note that this does not currently account for subcommands: Discord still highlights the mention if only the base command name is in the mentions, however it is unclickable.

### DIFF
--- a/packages/discord.js/src/structures/ApplicationCommand.js
+++ b/packages/discord.js/src/structures/ApplicationCommand.js
@@ -399,8 +399,9 @@ class ApplicationCommand extends Base {
   }
   
   /**
-   * When concatenated with a string, if this is a chat input command,
-   * this automatically returns the command's mention instead of the ApplicationCommand object.
+   * When concatenated with a string, this automatically returns the command's mention
+   * instead of the ApplicationCommand object.
+   * <info>This only works with {@link ChatInputCommandInteraction}s.</info>
    * @returns {string}
    * @example
    * // Logs: Command: </ping:123456789012345678>
--- a/packages/discord.js/src/structures/ApplicationCommand.js
+++ b/packages/discord.js/src/structures/ApplicationCommand.js
@@ -2,7 +2,7 @@
 
 const { chatInputApplicationCommandMention } = require('@discordjs/builders');
 const { DiscordSnowflake } = require('@sapphire/snowflake');
-const { ApplicationCommandOptionType } = require('discord-api-types/v10');
+const { ApplicationCommandOptionType, ApplicationCommandType } = require('discord-api-types/v10');
 const isEqual = require('fast-deep-equal');
 const Base = require('./Base');
 const ApplicationCommandPermissionsManager = require('../managers/ApplicationCommandPermissionsManager');
--- a/packages/discord.js/src/structures/ApplicationCommand.js
+++ b/packages/discord.js/src/structures/ApplicationCommand.js
@@ -399,7 +399,8 @@ class ApplicationCommand extends Base {
   }
   
   /**
-   * When concatenated with a string, if this is a chat input command, this automatically returns the command's mention instead of the ApplicationCommand object.
+   * When concatenated with a string, if this is a chat input command,
+   * this automatically returns the command's mention instead of the ApplicationCommand object.
    * @returns {string}
    * @example
    * // Logs: Command: </ping:123456789012345678>
--- a/packages/discord.js/src/structures/ApplicationCommand.js
+++ b/packages/discord.js/src/structures/ApplicationCommand.js
@@ -408,7 +408,7 @@ class ApplicationCommand extends Base {
   toString() {
     return this.type === ApplicationCommandType.ChatInput
       ? chatInputApplicationCommandMention(this.name, this.id)
-      : '[object Object]';
+      : super.toString();
   }
 
   /**
--- a/packages/discord.js/src/structures/ApplicationCommand.js
+++ b/packages/discord.js/src/structures/ApplicationCommand.js
@@ -396,6 +396,17 @@ class ApplicationCommand extends Base {
     }
     return true;
   }
+  
+  /**
+   * When concatenated with a string, this automatically returns the command's mention instead of the ApplicationCommand object.
+   * @returns {string}
+   * @example
+   * // Logs: Command: </ping:123456789012345678>
+   * console.log(`Command: ${command}`);
+   */
+  toString() {
+    return `</${this.name}:${this.id}>`
+  }
 
   /**
    * Recursively checks that all options for an {@link ApplicationCommand} are equal to the provided options.
--- a/packages/discord.js/src/structures/ApplicationCommand.js
+++ b/packages/discord.js/src/structures/ApplicationCommand.js
@@ -1,5 +1,6 @@
 'use strict';
 
+const { chatInputApplicationCommandMention } = require('@discordjs/builders');
 const { DiscordSnowflake } = require('@sapphire/snowflake');
 const { ApplicationCommandOptionType } = require('discord-api-types/v10');
 const isEqual = require('fast-deep-equal');
@@ -405,7 +406,7 @@ class ApplicationCommand extends Base {
    * console.log(`Command: ${command}`);
    */
   toString() {
-    return `</${this.name}:${this.id}>`
+    return chatInputApplicationCommandMention(`${this.name}`, this.id);
   }
 
   /**
--- a/packages/discord.js/src/structures/ApplicationCommand.js
+++ b/packages/discord.js/src/structures/ApplicationCommand.js
@@ -399,7 +399,7 @@ class ApplicationCommand extends Base {
   }
   
   /**
-   * When concatenated with a string, this automatically returns the command's mention instead of the ApplicationCommand object.
+   * When concatenated with a string, if this is a chat input command, this automatically returns the command's mention instead of the ApplicationCommand object.
    * @returns {string}
    * @example
    * // Logs: Command: </ping:123456789012345678>
--- a/packages/discord.js/src/structures/ApplicationCommand.js
+++ b/packages/discord.js/src/structures/ApplicationCommand.js
@@ -406,7 +406,7 @@ class ApplicationCommand extends Base {
    * console.log(`Command: ${command}`);
    */
   toString() {
-    return chatInputApplicationCommandMention(`${this.name}`, this.id);
+    return chatInputApplicationCommandMention(this.name, this.id);
   }
 
   /**
--- a/packages/discord.js/src/structures/ApplicationCommand.js
+++ b/packages/discord.js/src/structures/ApplicationCommand.js
@@ -406,7 +406,9 @@ class ApplicationCommand extends Base {
    * console.log(`Command: ${command}`);
    */
   toString() {
-    return chatInputApplicationCommandMention(this.name, this.id);
+    return this.type === ApplicationCommandType.ChatInput
+      ? chatInputApplicationCommandMention(this.name, this.id)
+      : '[object Object]';
   }
 
   /**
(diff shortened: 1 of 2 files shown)


Jiralite:
> Please note that this does not currently account for subcommand

Works for me (and has for a very long time):

https://user-images.githubusercontent.com/33201955/200207719-5052ac18-6487-4af8-bd07-d81a65ac064d.mp4

I think this was initially going to be added in #8546, but @almeidx decided against it... maybe they can shed more light onto this as to why.

Squiddleton:
> Works for me (and has for a very long time):

I was referring to if a command has subcommands, and you attempt to type out only the format `</name:id>`, clicking it won't do anything in the client.  This is the main downside of `toString()` without any parameters since the options are either 1) use the name without any subcommands (e.g. `</request:id>` in your case), or 2) pick the first/random subcommand name (e.g. `</request close:id>` which sacrifices consistency.

Jiralite:
Ah, thank you for clarifying.

Then, I wonder if this should be a method on the class vs. `toString()`. Then, I further wonder if that is necessary as discord.js already exports the application command mention formatter.

Squiddleton:
One middle ground between those would be a `toString()` with optional subcommand/subcommand group parameters.  Ignoring a `toString()` feels inconsistent with User/Role/Channel mentions, although adding parameters is inconsistent with native JS `Object#toString()`.

> discord.js already exports the application command mention formatter
Honestly, I had no idea this exists until now.  I completely understand disregarding this since that accomplishes the same purpose, although having some form of it used on the main class would still be appreciated.

almeidx:
> I think this was initially going to be added in #8546, but @almeidx decided against it... maybe they can shed more light onto this as to why.

This was because the ApplicationCommand includes all of the application commands (e.g. context menu commands), and the mention only works for Chat Input commands afaik

packages/discord.js/src/structures/ApplicationCommand.js
```diff
@@ -396,6 +396,17 @@ class ApplicationCommand extends Base {
...
+  
+  /**
+   * When concatenated with a string, this automatically returns the command's mention instead of the ApplicationCommand object.
+   * @returns {string}
+   * @example
+   * // Logs: Command: </ping:123456789012345678>
+   * console.log(`Command: ${command}`);
+   */
+  toString() {
+    return `</${this.name}:${this.id}>`
```
almeidx, line 408: If this is to be added, it should use the `chatInputApplicationCommandMention()` formatter from `@discordjs/builders`

almeidx (changes requested):

packages/discord.js/src/structures/ApplicationCommand.js
```diff
@@ -396,6 +397,17 @@ class ApplicationCommand extends Base {
...
+  
+  /**
+   * When concatenated with a string, this automatically returns the command's mention instead of the ApplicationCommand object.
+   * @returns {string}
+   * @example
+   * // Logs: Command: </ping:123456789012345678>
+   * console.log(`Command: ${command}`);
+   */
+  toString() {
+    return chatInputApplicationCommandMention(`${this.name}`, this.id);
```
Syjalo, line 409:
```suggestion
    return chatInputApplicationCommandMention(this.name, this.id);
```
Why do you convert the name string to a string?

Syjalo (changes requested):

packages/discord.js/src/structures/ApplicationCommand.js
Squiddleton: image
TS was complaining about the typings, so I wasn't sure what the best practice was here since the documentation clearly shows that the name is always present.  Is it best to just ignore TS here?

packages/discord.js/src/structures/ApplicationCommand.js
Syjalo: TypeScript? We are using JavaScript in discord.js and name property declared as a string the the d.ts.

packages/discord.js/src/structures/ApplicationCommand.js
Syjalo: https://github.com/discordjs/discord.js/pull/8818#discussion_r1015402809

packages/discord.js/src/structures/ApplicationCommand.js
Squiddleton: My apologies, I wasn't sure if this would error when building/testing.  Will keep that in mind for the future.

packages/discord.js/src/structures/ApplicationCommand.js
```diff
@@ -396,6 +397,17 @@ class ApplicationCommand extends Base {
...
+  
+  /**
+   * When concatenated with a string, this automatically returns the command's mention instead of the ApplicationCommand object.
+   * @returns {string}
+   * @example
+   * // Logs: Command: </ping:123456789012345678>
+   * console.log(`Command: ${command}`);
+   */
+  toString() {
+    return chatInputApplicationCommandMention(this.name, this.id);
```
Syjalo, line 409:
```suggestion
    return this.type === ApplicationCommandType.ChatInput
      ? chatInputApplicationCommandMention(this.name, this.id)
      : '[object Object]';
```
Shouldn't it check the type?

Syjalo (changes requested):

packages/discord.js/src/structures/ApplicationCommand.js
Syjalo: Did you forgot to change the return type in d.ts?

Squiddleton: Ah, right.  Would you suggest doing `ChatInputApplicationCommandMention | string`, or just `string`?  The former would make it more consistent with types that return UserMention, RoleMention, etc. although those can *only* return those types.

packages/discord.js/src/structures/ApplicationCommand.js
```diff
@@ -396,6 +397,19 @@ class ApplicationCommand extends Base {
...
+   * When concatenated with a string, this automatically returns the command's mention instead of the ApplicationCommand object.
+   * @returns {string}
+   * @example
+   * // Logs: Command: </ping:123456789012345678>
+   * console.log(`Command: ${command}`);
+   */
+  toString() {
+    return this.type === ApplicationCommandType.ChatInput
+      ? chatInputApplicationCommandMention(this.name, this.id)
+      : '[object Object]';
```
almeidx, line 411:
```suggestion
      : super.toString();
```

packages/discord.js/src/structures/ApplicationCommand.js
```diff
@@ -1,5 +1,6 @@
 'use strict';
 
+const { chatInputApplicationCommandMention } = require('@discordjs/builders');
 const { DiscordSnowflake } = require('@sapphire/snowflake');
 const { ApplicationCommandOptionType } = require('discord-api-types/v10');
```
almeidx, line 5:
```suggestion
const { ApplicationCommandOptionType, ApplicationCommandType } = require('discord-api-types/v10');
```

packages/discord.js/src/structures/ApplicationCommand.js
```diff
@@ -396,6 +397,19 @@ class ApplicationCommand extends Base {
     }
     return true;
   }
+  
+  /**
+   * When concatenated with a string, this automatically returns the command's mention instead of the ApplicationCommand object.
```
almeidx, line 402: This should mention that it will only return the mention if the command is a chat input command

almeidx (changes requested):

packages/discord.js/typings/index.d.ts
```diff
@@ -376,6 +376,7 @@ export class ApplicationCommand<PermissionsFetchType = {}> extends Base {
     command: ApplicationCommand | ApplicationCommandData | RawApplicationCommandData,
     enforceOptionOrder?: boolean,
   ): boolean;
+  public toString(): ChatInputApplicationCommandMention;
```
almeidx, line 379:
```suggestion
  public toString(): ChatInputApplicationCommandMention | string;
```

packages/discord.js/src/structures/ApplicationCommand.js
Squiddleton: Done!

packages/discord.js/src/structures/ApplicationCommand.js
```diff
@@ -396,6 +397,19 @@ class ApplicationCommand extends Base {
     }
     return true;
   }
+  
+  /**
+   * When concatenated with a string, if this is a chat input command, this automatically returns the command's mention instead of the ApplicationCommand object.
```
almeidx, line 402: This line is too long, eslint/prettier isn't going to like it
```suggestion
   * When concatenated with a string, if this is a chat input command,
   * this automatically returns the command's mention instead of the ApplicationCommand object.
```

almeidx (changes requested):

packages/discord.js/src/structures/ApplicationCommand.js
```diff
@@ -396,6 +397,20 @@ class ApplicationCommand extends Base {
     }
     return true;
   }
+  
+  /**
+   * When concatenated with a string, if this is a chat input command,
+   * this automatically returns the command's mention instead of the ApplicationCommand object.
```
Jiralite, line 403: The requirement would be better seen in `<info>`:

```suggestion
   * When concatenated with a string, this automatically returns the command's mention
   * instead of the ApplicationCommand object.
   * <info>This only works with {@link ChatInputCommandInteraction}s.</info>
```

Syjalo: This is not an interaction. I don't think `ChatInputCommandInteraction` is appropriate here.

packages/discord.js/src/structures/ApplicationCommand.js
```diff
@@ -396,6 +397,21 @@ class ApplicationCommand extends Base {
     }
     return true;
   }
+  
+  /**
+   * When concatenated with a string, this automatically returns the command's mention
+   * instead of the ApplicationCommand object.
+   * <info>This only works with {@link ChatInputCommandInteraction}s.</info>
```
Jiralite, line 404:
```suggestion
   * <info>This only works with chat input application commands.</info>
```

packages/discord.js/src/structures/ApplicationCommand.js
```diff
@@ -396,6 +397,21 @@ class ApplicationCommand extends Base {
...
+   * @returns {string}
+   * @example
+   * // Logs: Command: </ping:123456789012345678>
+   * console.log(`Command: ${command}`);
+   */
+  toString() {
+    return this.type === ApplicationCommandType.ChatInput
+      ? chatInputApplicationCommandMention(this.name, this.id)
+      : super.toString();
+  }
```
kyranet, line 414: A few things:

- This looks like a semver-major change.
- This does not account for chat commands with subcommands or subcommand groups and subcommands.
- This doesn't work for context menu commands, yet the class may be constructed by them as well.

kyranet (changes requested):

packages/discord.js/src/structures/ApplicationCommand.js
Squiddleton: Addressing these one at a time,

1. The only scenario where I see this as breaking is if the developer relies on `ApplicationCommand#toString()` returning `[Object object]` since both pre- and post-change return a string.  This seems like a very, *very* specific edge case, but is understandable nonetheless.
2. That was one of the main problems with not being able to use parameters for those options; as of now, Discord still highlights the mentions for commands with the formatting missing a required subcommand/group, but they are unclickable.  As such, whether or not it remains a "valid" mention in the form that the `toString()` change returns is debatable.
3. That's entirely correct.  By checking for the chat input type, this will still return a valid string for context menu commands, but it will remain inconsistent with the return value for chat input commands.

Overall, I see how this change feels inconsistent and much more iffy than the original intentions.  I have no qualms if this feels unfeasible and should be scrapped since the builder mention function serves the same purpose but with greater functionality/practicality.

net-tech:
@Squiddleton are you still working on this?

PR #8825, opened 2022-11-08
chore: use generator for range utility

suneettipirneni:

Aside from using a generator, this changes the range function in the following ways:

- `range` is no longer end-inclusive. This is because we mainly work with 0-based indexing which ends at length - 1.
- The function accepts an object or a just a number. The object contains the parameters for start, end, and step. The sole number parameter indicates the range 0...(n-1).

This is *techinically* a breaking change, but also not really because `range` is more for internal uses rather than it being a member of the public API.


packages/util/src/functions/range.ts
```diff
@@ -1,20 +1,55 @@
...
- * @param step - The amount to increment between each number
+ * @remarks
+ * This method is end-exclusive, for example the last number yielded by `range(5)` is 4. if you
+ * prefer add 1 to the range or `end` option.
+ * @param range - A number representing the the range to yield (exclusive) or an object with start, end and step
  * @example
  * Basic range
  * ```ts
- * range(3, 5); // [3, 4, 5]
+ * range(5); // [0, 1, 2, 3, 4]
```
almeidx, line 31: Isn't this comment inaccurate? Perhaps the example should be tweaked?

packages/util/src/functions/range.ts
```diff
@@ -1,20 +1,61 @@
...
+	/**
+	 * The amount to increment by
+	 *
+	 * @defaultValue `1`
+	 */
+	step?: number;
+}
+
+/**
+ * Yields the numbers in the given range as an array.
```
almeidx, line 22: Like in my previous review, this does not return an array

packages/util/__tests__/range.test.ts
```diff
@@ -2,11 +2,15 @@ import { describe, test, expect } from 'vitest';
 import { range } from '../src/index.js';
 
 describe('range', () => {
-	test('GIVEN valid range THEN valid array is returned', () => {
-		expect(range(0, 5)).toEqual([0, 1, 2, 3, 4, 5]);
+	test('GIVEN valid range and then valid numbers are returned', () => {
```
vladfrangu, line 5:
```suggestion
	test('GIVEN valid range THEN valid numbers are returned', () => {
```

vladfrangu (changes requested):

iCrawl (approved):

vladfrangu (approved):

SpaceEEC (approved):

PR #8829, opened 2022-11-09
fix: support `discordapp.com` webhook URLs

xTVaser:

Webhook URLs copied from Discord are no longer considered valid.  They are in the format `https://discordapp.com/api/webhooks/{ID}/{TOKEN}` now it seems.

This updates the regex to support such URLs and continue to parse it successfully.

### DIFF
--- a/packages/discord.js/src/util/Util.js
+++ b/packages/discord.js/src/util/Util.js
@@ -585,7 +585,7 @@ function cleanCodeBlockContent(text) {
  */
 function parseWebhookURL(url) {
   const matches = url.match(
-    /https?:\/\/(?:ptb\.|canary\.)?discord\.com\/api(?:\/v\d{1,2})?\/webhooks\/(\d{17,19})\/([\w-]{68})/i,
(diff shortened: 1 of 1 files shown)


almostSouji:
What client are you using to copy? I cannot reproduce it on any of mine.
As per spec `discordapp.com` is only used for the CDN these days.

Jiralite:
I am not able to reproduce this on all release channels (stable, PTB, canary) additionally.

ckohen:
Webhooks copied from discord use whatever url your client is currently using to access discord. I.e. if you use canary, it will copy canary.discord.com, ptb will copy ptb.canary.com. What this means for you is that your client is also using the unsupported url to make all its requests. On occassion this has been known to happen but it requires your client to be in an incredibly bad state.

TLDR this is a client issue and this regex should intentionally not support discordapp since discord does not.

xTVaser:
> TLDR this is a client issue and this regex should intentionally not support discordapp since discord does not.

I see, perhaps a better error message would be useful as this is the format for the URLs I get when copied straight out of the desktop client in multiple channels/servers.  I only figured out the problem by digging into the code itself, and would have no idea how or why my client is in an "incredibly bad state".

Sorry for the inconvenience!

thehackerboi69github:
I downloaded the Discord client for Linux from the official site and it gives me discordapp.com URLs

PR #8832, opened 2022-11-11
feat: Rename interaction methods

Syjalo:
This PR renames some interaction methods to not be confused.

| Method                                      | Description                                                                                                                                                   |
|---------------------------------------------|---------------------------------------------------------------------------------------------------------------------------------------------------------------|
| `.reply()`                                  | It is ok                                                                                                                                                      |
| `.deferReply()`                             | It is ok                                                                                                                                                      |
| `.fetchReply()` -> `.fetchResponse()`       | With this method we can fetch not only interaction replies. If we do `.update()` or `.deferUpdate()` and then `.fetchReply()`, we will get the target message |
| `.editReply()` -> `.editResponse()`         | The same with `.fetchReply()`                                                                                                                                 |
| `.deleteReply()` -> `.deleteResponse()`     | The same with `.fetchReply()`                                                                                                                                 |
| `.followUp()`                               | It is ok                                                                                                                                                      |
| `.update()` -> `.updateMessage()`           | When we call that method it looks like we update the interaction (`interaction.update()`), not the target message                                             |
| `.deferUpdate()` -> `.deferMessageUpdate()` | For consistency with `.update()`                                                                                                                              |

### DIFF
--- a/packages/discord.js/src/structures/CommandInteraction.js
+++ b/packages/discord.js/src/structures/CommandInteraction.js
@@ -205,15 +205,15 @@ class CommandInteraction extends BaseInteraction {
   /* eslint-disable no-empty-function */
   deferReply() {}
   reply() {}
-  fetchReply() {}
-  editReply() {}
-  deleteReply() {}
+  fetchResponse() {}
+  editResponse() {}
+  deleteResponse() {}
   followUp() {}
   showModal() {}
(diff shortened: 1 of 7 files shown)


cobaltt7:
now that we have `withResponse`, `fetchResponse` is unclear what it's doing

PR #8840, opened 2022-11-16
docs: describe InteractionEditReplyOptions

jaw0r3k:
@MrMythicalYT idea
Just a docs change


packages/discord.js/src/structures/interfaces/InteractionResponses.js
```diff
@@ -137,6 +137,7 @@ class InteractionResponses {
   }
 
   /**
+   * Options that can be passed into editReply.
```
kyranet, line 140: I think it's better to link the method rather than just mentioning it, at first, you can try this:
```suggestion
   * Options that can be passed into {@link editReply}.
```

If the website doesn't pick the link up, I suppose you can fall back to the following option:
```suggestion
   * Options that can be passed into {@link InteractionResponses.editReply}.
```

kyranet (changes requested):

kyranet (approved):

vladfrangu (approved):

iCrawl (approved):

PR #8841, opened 2022-11-16
fix(Activity): Fix equals() not checking for differing emoji

Eejit43:
This PR fixes `Activity.equals()` not checking for differing emojis, leading `presenceUpdate` to not fire if only a custom status emoji was updated.


packages/discord.js/src/structures/Presence.js
```diff
@@ -254,7 +254,8 @@ class Activity {
         this.type === activity.type &&
         this.url === activity.url &&
         this.state === activity.state &&
-        this.details === activity.details)
+        this.details === activity.details &&
+        this.emoji === activity.emoji)
```
almeidx, line 258: Something like this is probably a better idea
```suggestion
        this.emoji?.id === activity.emoji?.id)
```

almeidx (changes requested):

packages/discord.js/src/structures/Presence.js
Eejit43: Didn't even consider that, thanks! 👍

RedGuy12: don't built in emojis not have an id?

packages/discord.js/src/structures/Presence.js
```diff
@@ -254,7 +254,8 @@ class Activity {
         this.type === activity.type &&
         this.url === activity.url &&
         this.state === activity.state &&
-        this.details === activity.details)
+        this.details === activity.details &&
+        this.emoji?.id === activity.emoji?.id)
```
GoldenAngel2, line 258: This won't fix the overall problem, if the user changes their custom status from one unicode emoji to another then it won't fire the event either, so both the `emoji?.name` and `emoji?.id` should be checked

GoldenAngel2 (changes requested):

packages/discord.js/src/structures/Presence.js
jaw0r3k: ^ exactly

Eejit43: you're right... wonder why it worked for me in testing then.

however, what if the emoji's name is updated, won't that trigger a false update?

jaw0r3k: Your can't prevent that

Eejit43: Alright I'll just check for `name` & `id` then 👍

Eejit43: > Alright I'll just check for `name` & `id` then 👍

Better yet, might as well properly check for it: `this.emoji?.id ? this.emoji.id === activity.emoji?.id : this.emoji?name === activity.emoji?.name`

jaw0r3k: idi

Eejit43: > idk

Well I've tested it, it seems to work!

packages/discord.js/src/structures/Presence.js
```diff
@@ -254,7 +254,8 @@ class Activity {
         this.type === activity.type &&
         this.url === activity.url &&
         this.state === activity.state &&
-        this.details === activity.details)
+        this.details === activity.details &&
+        (this.emoji?.id ? this.emoji.id === activity.emoji?.id : this.emoji?.name === activity.emoji?.name))
```
Jiralite, line 258:
```suggestion
        this.emoji?.id === activity.emoji?.id &&
        this.emoji?.name === activity.emoji?.name)
```

Eejit43: Won't this falsely update if the emoji is a guild emoji and its name is updated?

Jiralite: How is that a false update?

Eejit43: I'm kinda wavering back and forth, but you would know better up to you

Jiralite: I would say to change it to this. By the way, modifying the name of a custom emoji will not fire a `PRESENCE_UPDATE` event.

jaw0r3k: but updating presence yes
and some old presence can have old name

kyranet (approved):

vladfrangu (approved):

iCrawl (approved):

PR #8842, opened 2022-11-16
fix(Activity): Fix equals() not checking for differing emoji (v13)

Eejit43:
This PR fixes `Activity.equals()` not checking for differing emojis, leading `presenceUpdate` to not fire if only a custom status emoji was updated.

### DIFF
--- a/src/structures/Presence.js
+++ b/src/structures/Presence.js
@@ -293,8 +293,7 @@ class Activity {
         this.url === activity.url &&
         this.state === activity.state &&
         this.details === activity.details &&
-        this.emoji?.id === activity.emoji?.id &&
-        this.emoji?.name === activity.emoji?.name)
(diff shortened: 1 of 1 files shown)


src/structures/Presence.js
```diff
@@ -292,7 +292,8 @@ class Activity {
         this.type === activity.type &&
         this.url === activity.url &&
         this.state === activity.state &&
-        this.details === activity.details)
+        this.details === activity.details &&
+        this.emoji?.id === activity.emoji?.id)
```
jaw0r3k, line 296:
```suggestion
        this.emoji?.id === activity.emoji?.id &&
        this.emoji?.name === activity.emoji?.name)
``` Also there it needs to check name

Eejit43: Note for others: currently discussing in #8841.

kyranet (approved):

vladfrangu (approved):

Issue #8084, opened 2022-06-13
RFC(sharder): Scaling from the threads to the cores and beyond


kyranet:
*Preface: this RFC (Request For Comments) is meant to gather feedback and opinions for the feature set and some implementation details for `@discordjs/sharder`! While a lot of the things mentioned in this RFC will be opinionated from my side (like what I’d like to see implemented in the module), not everything might be added in the initial version or some extra things might be added as extras.*

This RFC is split into several parts, described below. As an important note, through this entire specification, "shard" is **not** used to describe a gateway shard, those are handled by `@discordjs/ws` (see #8083). "Shard" is used to describe the client that the manager instantiates. Those could be (but are not limited to): workers, processes, or servers.

Before any work is done on this, I would like to ask them kindly to not take much if any inspiration from the current shard manager, and that this implementation needs to be as agnostic as possible so it can be used by bots made with discord.js's components (or even with older versions of discord.js).

*Yes, I'm also aware this is over-engineered, but everything is designed with composability in mind, and has been thought with experience from implementing this in #7204. Truth be told, without some of those components, certain things would be significantly harder to implement.*

---

## ShardManager

*Alternatively: `Sharder`, `ProcessManager`, `ClientManager`*

This is the entry point in using this module. Its job is to handle the creation of shards, and is the central place where shards can intercommunicate. This manager also sets several parameters via environment variables (if available) to configure the clients.

As a result, a `ShardManager` has access to one or more `ShardClient`s, and has a channel for each one of them. 

### API

```typescript
// Error handling: TBD
// Lifecycle handlers: TBD

class ShardManager {
	public constructor(options: ShardManagerOptions);
}

interface ShardManagerOptions {
	strategy?: ChannelHandler | string;
}
```

> **Note**: Shards will be configured in the strategy's options. The reasoning for this is simple, not all strategies have to be single-layer. There are also proxies which act as a second layer, 

### Strategies

This class also instantiates shards in one of the following four ways:

- **Worker**: using [`worker_threads`](https://nodejs.org/dist/latest-v18.x/docs/api/worker_threads.html#new-workerfilename-options), it has the benefit of faster IPC than process-based strategies.
  > **Warning**: If this option is chosen, you may not be able to use worker sharding in `@discordjs/ws`.
[...]

PR #8859, opened 2022-11-24
feat: added `@discordjs/sharder`

kyranet:

Supersedes #7204, implements part of #8084.

For more information about this sharder's architecture, please refer to the aforementioned RFC issue.

### Features

- [ ] ShardManager
    - [ ] Channel strategies (server)
        - [x] Cluster
        - [x] Fork (default)
        - [ ] WorkerThread
    - [ ] Events
        - [ ] Shard has been started
        - [ ] Shard has been restarted
        - [ ] Shard has been closed
        - [ ] Shard has signalled ready
        - [ ] Shard has signalled exit
        - [ ] Shard has sent an invalid message
        - [ ] Shard has sent a ping
        - [ ] Shard has not sent a ping
    - [x] Ping
- [ ] ShardClient
    - [ ] Channel strategies (client)
        - [ ] Cluster
        - [ ] Fork (default)
        - [ ] WorkerThread
    - [ ] Events
        - [ ] Ready
        - [ ] Exit
        - [ ] Starting
        - [ ] Restarting
- [ ] Shared
    - [ ] Message strategies
        - [x] JSON (default)
        - [ ] V8
    - [ ] Message transformers
        - [ ] ~~Gzip~~ (will not be included in initial PR)
        - [ ] ~~Brotli~~ (will not be included in initial PR)
- [ ] ~~ShardManagerProxy~~ (will not be included in initial PR)


packages/sharder/src/ShardManager.ts
```diff
@@ -0,0 +1,228 @@
...
+export class ShardManager<ChannelOptions extends {} = {}> extends AsyncEventEmitter {
+	public readonly respawns: number;
+
+	public readonly channelOptions: ChannelOptions;
+
+	public readonly pingOptions: ShardPingOptions;
+
+	public readonly spawnOptions: Required<ShardSpawnOptions>;
+
+	public readonly channels: IChannel[] = [];
```
DraftProducts, line 20: We could maybe choose another more conducive name for `IChannel` & `<ShardManager>.channels`, it could be confused with traditional Discord channels, maybe this is something to take into account? Unless the fact that it is internal allows us to derogate from this non-ambiguity?

DraftProducts (changes requested):
Nice supplant, excited to test cluster in production mode 🚀

packages/sharder/src/ShardManager.ts
kyranet: It's called channel because it's the... *channel* strategy both the manager and the clients use to communicate.

DraftProducts: Before the class that managed the shard like this was called "shard", that why I was hesitant about this new name.

kyranet: You said that, they're shards. But that's just djs's ShardingManager being too limited. Also, they're *literally* shards, and the way it instantiates shards is fixed by `child_process.fork` or `worker_threads.spawn`. Furthermore, they held a lot of Discord shard state.

Here, they're not just a single shard, but clusters of shards, they don't keep track of Discord shard state (in fact they're agnostic, this library can even be used for other libraries such as Eris, or even in stacks that have no relationship with Discord at all). And those clusters of shards are just the IDs, the classes' purpose is to define what's the way of communicating with the client, which is basically:

- ITC (inter-thread communication, `worker_threads`)
- IPC (inter-process communication, `child_process.fork`/`cluster.fork`)
- TCP (over network, `ws`/`http(s)`/`http2`/`quic`)
- RPC (over network, gRPC)
- Other (Redis streams, RabbitMQ)

Similarly, those keep the same analogy and are ported to the opposite direction, so basically something like this:

```mermaid
flowchart LR
  A[ShardManager]
  subgraph clients
  B[ShardClient]
  C[ShardClient]
  D[ShardClient]
  end
  A <-->|Channel| B
  A <-->|Channel| C
  A <-->|Channel| D
```

There needs to be an opposite for the `ShardManager --> ShardClient`, while `manager.shards` may be somewhat accurate, `client.shard` (for `ShardClient --> ShardManager`) wouldn't. Furthermore, the "shard" is at the process where `client` lies on, it's not on `client` nor is `client`.

That's why I chose the name "channel". I understand it may be confusing for some users, but also keep in mind that sharder has *no* relationship with Discord, and instead, channels refer to the channel on which messages are sent.

It also makes a great analogy with real life, here we're using GitHub as a channel to communicate, but we could also be using Discord, or phones, or pigeons, or we could fly somewhere and talk face to face; the end result is always the same, the means we use to communicate is called a "[communication channel](https://en.wikipedia.org/wiki/Communication_channel)", here, `IChannel` isn't that different.

PR #8860, opened 2022-11-24
docs(Status): Add op codes for shard status

Moquent:
- Added missing op codes reference link for the shard status

- This PR only includes non-code changes, like changes to documentation, README, etc.

### DIFF
--- a/packages/discord.js/src/util/Status.js
+++ b/packages/discord.js/src/util/Status.js
@@ -3,7 +3,7 @@
 const { createEnum } = require('./Enums');
 
 /**
- * @typedef {Object} Status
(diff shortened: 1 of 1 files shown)


packages/discord.js/src/util/Status.js
```diff
@@ -3,7 +3,7 @@
 const { createEnum } = require('./Enums');
 
 /**
- * @typedef {Object} Status
+ * @typedef {Object<GatewayOpcodes>} Status
```
Jiralite, line 6: This isn't exactly right as it's missing some members (see https://discord-api-types.dev/api/discord-api-types-v10/enum/GatewayOpcodes).

PR #8863, opened 2022-11-25
Update incorrect erlpack install command

CitTheDev:
This PR updates the `erlpack` installation command as it was incorrect. The hyperlink now leads to the npm page of the `erlpack` package instead of the Github Page

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,7 +39,7 @@ pnpm add discord.js
 ### Optional packages
 
 - [zlib-sync](https://www.npmjs.com/package/zlib-sync) for WebSocket data compression and inflation (`npm install zlib-sync`)
-- [erlpack](https://github.com/discord/erlpack) for significantly faster WebSocket data (de)serialisation (`npm install discord/erlpack`)
+- [erlpack](https://www.npmjs.com/package/erlpack) for significantly faster WebSocket data (de)serialisation (`npm install erlpack`)
 - [bufferutil](https://www.npmjs.com/package/bufferutil) for a much faster WebSocket connection (`npm install bufferutil`)
 - [utf-8-validate](https://www.npmjs.com/package/utf-8-validate) in combination with `bufferutil` for much faster WebSocket processing (`npm install utf-8-validate`)
 - [@discordjs/voice](https://www.npmjs.com/package/@discordjs/voice) for interacting with the Discord Voice API (`npm install @discordjs/voice`)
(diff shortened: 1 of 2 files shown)


imranbarbhuiya:
The package at npm is different from the github one. That's why github link was used

CitTheDev:
> The package at npm is different from the github one. That's why github link was used 

If that is the case I will change the npm link back to the GitHub link. Thanks for confirming

imranbarbhuiya:
> > The package at npm is different from the github one. That's why github link was used
> 
> If that is the case I will change the npm link back to the GitHub link. Thanks for confirming

The change is still invalid as the command will install erlpack from npm and not the girhub one

CitTheDev:
> > > The package at npm is different from the github one. That's why github link was used
> 
> > 
> 
> > If that is the case I will change the npm link back to the GitHub link. Thanks for confirming
> 
> 
> 
> The change is still invalid as the command will install erlpack from npm and not the girhub one 

Then what command should I put for the package installation?

Jiralite:
The existing command is fine. Can you elaborate on why it is incorrect? Did an error occur for you somehow?

CitTheDev:
> The existing command is fine. Can you elaborate on why it is incorrect? Did an error occur for you somehow?

imranbarbhuiya said that the reason why the GitHub page was linked in the README was because it was not synced with the actual package. After I reverted the hyperlink back to the GitHub link of the package, they said that the npm package was not synced _again_, which caused me some confusion. The reason for this pull request was to fix the incorrect npm command to download `erlpack`.

Jiralite:
> Can you elaborate on why it is incorrect?

CitTheDev:
I do not know, hence why I asked imranbarbhuiya. Maybe ask them what was wrong with my solution because I don't see anything wrong with it.

Jiralite:
To be clear, I'm asking why _you_ said:

> This PR updates the `erlpack` installation command as it was incorrect.

What's incorrect about it?

CitTheDev:
Oh sorry for the confusion. The thing that was wrong with the npm install command was that it was `npm I discord/erlpack` when it is either supposed to be `npm i @discord/erlpack` or `npm i erlpack`. Since the `@discord/erlpack` package did not exist and `erlpack` did I changed it to that.

imranbarbhuiya:
`discord/erlpack` is valid will be resolved to the github repo of erlpack. And it'll download the package directly from npm.

More info https://docs.npmjs.com/cli/v9/commands/npm-install

Jiralite:
Did you try running the supposedly incorrect command?

PR #8736, opened 2022-10-11
feat: Add `@discordjs/core`

suneettipirneni:
### A new library to interact with the discord API at a lower level.

`@discordjs/core` is a thinly abstracted wrapper on top of the discord api. It avoids much of the pre-processing a fully-fledged library such as discord.js performs. With `/core` events give the raw data directly from gateway events. It also includes a rest wrapper which takes raw data and performs queries. In other words the rest client and gateway logic are decoupled from the actual returned structures giving users maximum flexibility to make their own modifications.

TODO:
- [x] Implement files/form data uploading
- [x] Add automod API

### example usage:
```ts
import { REST } from '@discordjs/rest';
import { WebSocketManager } from '@discordjs/ws';
import { Gateway, GatewayDispatchEvents, GatewayIntentBits, InteractionType, MessageFlags, createClient } from '@discordjs/core';

// Create REST and web socket managers directly
const rest = new REST({ version: '10' }).setToken(token);
const ws = new WebSocketManager({
  token,
  intents: GatewayIntentBits.GuildMessages | GatewayIntentBits.MessageContent,
  rest,
});

// Create a client to emit relevant events.
const client = createClient({ rest, ws });

// Listen for interactions
// Each event contains an `api` prop along with the event data that allows you to interface with the discord REST API
client.on(GatewayDispatchEvents.InteractionCreate, async ({ data: interaction, api }) => {
  if (
    !(interaction.type === InteractionType.ApplicationCommand) ||
    interaction.data.name !== 'ping'
  ) {
    return;
  }

  api.interactions.reply(interaction.id, interaction.token, { content: 'Pong!', flags: MessageFlags.Ephemeral });
});

// Listen for the ready event
client.on(GatewayDispatchEvents.Ready, () => console.log('Ready!'));

// Start the ws connection.
ws.connect();
```

Similar to the raw ws events, you can also keep track of shards as each event includes a `shardId` prop:
```ts
client.on(GatewayDispatchEvents.InteractionCreate, async ({ data: interaction, api, shardId }) => {
  ...
  api.interactions.reply(interaction.id, interaction.token, `Pong! from shard ${shardId}`);
});
```

### Independent REST API usage
This library has REST API functions that be used without a live connection to the gateway:

```ts
// Create rest instance
const rest = new REST({ version: '10' }).setToken(token);

// Pass into api
const api = new API(rest);

// Fetch a guild using the api wrapper
const guild = await api.guilds.get('1234567891011');
```

[...]

PR #8868, opened 2022-11-26
refactor: update `makeURLSearchParams` to accept readonly non-`Record`s

kyranet:

I didn't realize how painful it was to use the utility until I saw #8736 using them extensively. This patch ensures that:

1. We can pass non-Records:
    ```diff
      declare options: RESTGetAPIApplicationCommandsQuery;
    - makeURLSearchParams(options as Record<string, unknown>);
    + makeURLSearchParams(options);
    ```
1. We can cast types:
    ```diff
      // Expects `RESTGetAPIApplicationCommandsQuery`:
    - makeURLSearchParams({ withLocalizations: true }); // Should be `with_localizations``, doesn't raise type error.
    + makeURLSearchParams<RESTGetAPIApplicationCommandsQuery>({ withLocalizations: true }); // Raises type error due to mismatching types.
    ```

### DIFF
--- a/packages/rest/src/lib/utils/utils.ts
+++ b/packages/rest/src/lib/utils/utils.ts
@@ -43,7 +43,7 @@ function serializeSearchParam(value: unknown): string | null {
  * @param options - The options to use
  * @returns A populated URLSearchParams instance
  */
-export function makeURLSearchParams(options?: Record<string, unknown>) {
+export function makeURLSearchParams<T extends object & {}>(options?: Readonly<T>) {
 	const params = new URLSearchParams();
 	if (!options) return params;
 
--- a/packages/rest/src/lib/utils/utils.ts
+++ b/packages/rest/src/lib/utils/utils.ts
@@ -43,7 +43,7 @@ function serializeSearchParam(value: unknown): string | null {
  * @param options - The options to use
  * @returns A populated URLSearchParams instance
  */
-export function makeURLSearchParams<T extends object & {}>(options?: Readonly<T>) {
+export function makeURLSearchParams<T extends object>(options?: Readonly<T>) {
 	const params = new URLSearchParams();
 	if (!options) return params;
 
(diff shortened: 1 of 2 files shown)


SpaceEEC (approved):

iCrawl (approved):

packages/rest/src/lib/utils/utils.ts
```diff
@@ -43,7 +43,7 @@ function serializeSearchParam(value: unknown): string | null {
  * @param options - The options to use
  * @returns A populated URLSearchParams instance
  */
-export function makeURLSearchParams(options?: Record<string, unknown>) {
+export function makeURLSearchParams<T extends object & {}>(options?: Readonly<T>) {
```
imranbarbhuiya, line 46: from ts 4.7 (or 8 ig) `T & {}` is `{}`. So this will become `object`. So do we need this intersection?

imranbarbhuiya (approved):

packages/rest/src/lib/utils/utils.ts
kyranet: Yes.

- `object` means any object value.
- `{}` means any non-nullish value (`unknown` sans `undefined | null`).

When combined, `object & {}`, it means that it's any non-nullish object.

imranbarbhuiya: check https://tsplay.dev/wE4D2W
when using `object`, it's already non-nullish and adding the intersection of `{}` with any value returns the same value like 
`number & {}` is also `number`

kyranet: Besides `type C = Test<null>` needing to be `type C = TestWithInt<null>` on line 6, yes, you're right.

I'll remove `{}` right away, thank you for letting me know, I didn't know the semantics changed in 4.8 (I checked the announcements for [4.7](https://devblogs.microsoft.com/typescript/announcing-typescript-4-7/), [4.8](https://devblogs.microsoft.com/typescript/announcing-typescript-4-8/), and [4.9](https://devblogs.microsoft.com/typescript/announcing-typescript-4-9/), but only found [Unconstrained Generics No Longer Assignable to `{}`](https://devblogs.microsoft.com/typescript/announcing-typescript-4-8/#unconstrained-generics-no-longer-assignable-to) inside the 4.8 announcement, it was a bit hidden to be fair). 🙏🏼

iCrawl (approved):

PR #8869, opened 2022-11-26
feat(website): add `app` dir

suneettipirneni:


apps/website/.vscode/settings.json
```diff
@@ -0,0 +1,4 @@
+{
+	"typescript.tsdk": "../../node_modules/typescript/lib",
```
imranbarbhuiya, line 2: Git ignore this file. The path is os specific and everytime a user uses the build command on windows, it'll change. I copied these 2 lines in the root of my monorepo and gitignored this file.

imranbarbhuiya (changes requested):
Also from v13, `prefetch=true` is the default so false is still a valid option. Did u removed it intentionally?

iCrawl:
> ~~Also from v13, `prefetch=true` is the default so false is still a valid option. Did u remove it intentionally?~~
> 
> Edit: just saw fa7967245d4cd33bef23f2227e4d354e5c501fcc commit

If you want some context on this, I'm still trying to figure out why prefetch is giving me a lot of 500 or 404s. Additionally prefetch seems to work better for v13 as its still not 100% stable.

PR #8885, opened 2022-11-30
feat(StageInstanceManager): add `guildScheduledEvent` to `create()`

almeidx:
Upstream:
- discord/discord-api-docs#5687


SpaceEEC (approved):

packages/discord.js/src/managers/StageInstanceManager.js
```diff
@@ -59,14 +61,17 @@ class StageInstanceManager extends CachedManager {
     const channelId = this.guild.channels.resolveId(channel);
     if (!channelId) throw new DiscordjsError(ErrorCodes.StageChannelResolve);
     if (typeof options !== 'object') throw new DiscordjsTypeError(ErrorCodes.InvalidType, 'options', 'object', true);
-    let { topic, privacyLevel, sendStartNotification } = options;
+    let { guildScheduledEvent, topic, privacyLevel, sendStartNotification } = options;
```
kyranet, line 64: I'm confused as to why this is a `let`, it doesn't seem to be reassigned anywhere.

packages/discord.js/src/managers/StageInstanceManager.js
```diff
@@ -59,14 +61,17 @@ class StageInstanceManager extends CachedManager {
...
+
+    const guildScheduledEventId = guildScheduledEvent && this.resolveId(guildScheduledEvent);
 
     const data = await this.client.rest.post(Routes.stageInstances(), {
       body: {
         channel_id: channelId,
         topic,
         privacy_level: privacyLevel,
         send_start_notification: sendStartNotification,
+        guild_scheduled_event_id: guildScheduledEventId ?? undefined,
```
kyranet, line 74: Can we future-proof this by not including `?? undefined`? `guildScheduledEventId` will be `undefined` if it's not provided.

kyranet (changes requested):

packages/discord.js/src/managers/StageInstanceManager.js
almeidx: It will be `null` if it is provided but not resolved (if the user passes something other than an id or a GuildScheduledEvent instance). Is that okay?

kyranet: I did the review taking that into account, so yes. Worst case scenario it'll throw an API error to signal the user they have given the wrong thing.

kyranet (approved):

vladfrangu (approved):

iCrawl (approved):

PR #8890, opened 2022-11-30
feat: align some methods with the Set Methods proposal

almeidx:
Stage 3 proposal:
- https://github.com/tc39/proposal-set-methods

Overview of the changes:
- `intersect` -> `intersection`
- `difference` -> `symmetricDifference`
- `subtract` -> `difference`
- Added `union`

### DIFF
--- a/packages/collection/src/collection.ts
+++ b/packages/collection/src/collection.ts
@@ -751,50 +751,100 @@ export class Collection<K, V> extends Map<K, V> {
 	}
 
 	/**
-	 * The intersect method returns a new structure containing items where the keys and values are present in both original structures.
+	 * The intersection method returns a new collection containing the items where the key is present in both collections.
 	 *
 	 * @param other - The other Collection to filter against
+	 * @example
+	 * ```ts
+	 * const col1 = new Collection([['a', 1], ['b', 2]]);
+	 * const col2 = new Collection([['a', 1], ['c', 3]]);
+	 * const intersection = col1.intersection(col2);
+	 * console.log(col1.intersection(col2));
+	 * // => Collection { 'a' => 1 }
+	 * ```
 	 */
-	public intersect<T>(other: ReadonlyCollection<K, T>): Collection<K, T> {
-		const coll = new this.constructor[Symbol.species]<K, T>();
-		for (const [key, value] of other) {
-			if (this.has(key) && Object.is(value, this.get(key))) {
-				coll.set(key, value);
-			}
-		}
-
-		return coll;
-	}
-
-	/**
-	 * The subtract method returns a new structure containing items where the keys and values of the original structure are not present in the other.
-	 *
-	 * @param other - The other Collection to filter against
-	 */
-	public subtract<T>(other: ReadonlyCollection<K, T>): Collection<K, V> {
-		const coll = new this.constructor[Symbol.species]<K, V>();
-		for (const [key, value] of this) {
-			if (!other.has(key) || !Object.is(value, other.get(key))) {
-				coll.set(key, value);
-			}
-		}
-
-		return coll;
-	}
-
-	/**
-	 * The difference method returns a new structure containing items where the key is present in one of the original structures but not the other.
-	 *
-	 * @param other - The other Collection to filter against
-	 */
-	public difference<T>(other: ReadonlyCollection<K, T>): Collection<K, T | V> {
+	public intersection<T>(other: ReadonlyCollection<K, T>): Collection<K, T | V> {
 		const coll = new this.constructor[Symbol.species]<K, T | V>();
-		for (const [key, value] of other) {
-			if (!this.has(key)) coll.set(key, value);
+
+		for (const [key, value] of this) {
+			if (other.has(key)) coll.set(key, value);
 		}
+
+		return coll;
+	}
+
+	/**
+	 * Returns a new collection containing the items where the key is present in either of the collections.
+	 *
+	 * @remarks
+	 *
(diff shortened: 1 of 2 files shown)


packages/collection/src/collection.ts
```diff
@@ -669,52 +669,93 @@ export class Collection<K, V> extends Map<K, V> {
...
-			}
+			if (!other.has(k)) coll.set(k, v);
 		}
 
 		return coll;
 	}
 
 	/**
-	 * The difference method returns a new structure containing items where the key is present in one of the original structures but not the other.
+	 * The difference method returns a new collection containing the items where the key is present in this collection and not the other, and vice-versa.
```
kyranet, line 736: This can be much to unpack for the average user, while the terminology is correct, it may be too technical.

We can draw some inspiration about the wording from other languages, e.g. .NET's [`HashSet<T>.SymmetricExceptWith`](https://learn.microsoft.com/en-us/dotnet/api/system.collections.generic.hashset-1.symmetricexceptwith?view=net-7.0):

```suggestion
	 * Returns a new collection containing only the items where the keys are present in either collection, but not both.
```

The description above fits better with the mathematical description too: https://en.wikipedia.org/wiki/Symmetric_difference

packages/collection/src/collection.ts
```diff
@@ -669,52 +669,93 @@ export class Collection<K, V> extends Map<K, V> {
...
+	 * console.log(union);
+	 * // => Collection { 'a' => 1, 'b' => 2, 'c' => 3 }
+	 * ```
+	 */
+	public union<T>(other: ReadonlyCollection<K, T>): Collection<K, T | V> {
+		return new this.constructor[Symbol.species]<K, T | V>([...this, ...other]);
+	}
+
+	/**
+	 * The difference method returns a new collection containing the items where the key is present in this collection and not the other.
```
kyranet, line 712:
```suggestion
	 * Returns a new collection containing the items where the key is present in this collection but not the other.
```

packages/collection/src/collection.ts
```diff
@@ -669,52 +669,93 @@ export class Collection<K, V> extends Map<K, V> {
...
+		for (const [k, v] of this) {
+			if (other.has(k)) coll.set(k, v);
 		}
 
 		return coll;
 	}
 
 	/**
-	 * The subtract method returns a new structure containing items where the keys and values of the original structure are not present in the other.
+	 * The union method returns a new collection containing the items where the key is present in either of the collections.
```
kyranet, line 695: At first I would suggest this to be more direct and consistent with my other suggestions, since this is the mathematical definition...
```suggestion
	 * Returns a new collection containing the items where the key is present in either of the collections.
```

But I don't feel like it's right, take the example:

```typescript
const col1 = new Collection([['a', 1], ['b', 2]]);
const col2 = new Collection([['a', 1], ['c', 3]]);
const union = col1.union(col2);
console.log(union);
// => Collection { 'a' => 1, 'b' => 2, 'c' => 3 }
```

This can be a given, but once we change things a little, it behaves in a way that is inconsistent with the other methods, namely, the fact that `B`'s values have precedence over `A`'s.

```typescript
const col1 = new Collection([['a', 1], ['b', 2]]);
const col2 = new Collection([['a', 4], ['c', 3]]);
const union = col1.union(col2);
console.log(union);
// => Collection { 'a' => 4, 'b' => 2, 'c' => 3 }
//                        ^ union.get('a') changed!
```

I believe this behaviour would suit more for a "merge" — which has its own [proposal](https://github.com/tc39/proposal-collection-methods) with a [polyfill](https://github.com/tc39/proposal-collection-methods/blob/99480f94980a69efd1f24bdd7381f571d688f53d/map-polyfill.js#L88-L92) — rather than an "union". In my head, the way to replicate an union behaviour is to skip setting B's values if their keys are already present, as shown by the [`Set.prototype.union` proposal](https://github.com/tc39/proposal-set-methods/blob/b67c7a22f713ac70e5fa7d596aa85490fad17d7e/spec/index.html#L25-L26).

kyranet (changes requested):
We should make a new PR to change the documentation a little and phase out `The X method returns...` from this package. It's verbose and we don't write like that anywhere else.

I'm also not fond of `describe('X() tests', cb);` and would rather prefer `describe('X', cb);`, but that'd be another PR. 😅 

Also adding the https://github.com/discordjs/discord.js/labels/blocked label to prevent accidental semver-major merges on main.

packages/collection/src/collection.ts
```diff
@@ -669,52 +669,102 @@ export class Collection<K, V> extends Map<K, V> {
...
+	 * ```ts
+	 * const col1 = new Collection([['a', 1], ['b', 2]]);
+	 * const col2 = new Collection([['a', 1], ['b', 3], ['c', 3]]);
+	 * const union = col1.union(col2);
+	 * console.log(union);
+	 * // => Collection { 'a' => 1, 'b' => 2, 'c' => 3 }
+	 * ```
+	 */
+	public union<T>(other: ReadonlyCollection<K, T>): Collection<K, T | V> {
+		const coll = new this.constructor[Symbol.species]<K, T | V>([...this]);
```
kyranet, line 711: I'm pretty sure `Map`'s constructor takes iterables, not just arrays, so this should just work.
```suggestion
		const coll = new this.constructor[Symbol.species]<K, T | V>(this);
```

kyranet (approved):

vladfrangu (approved):

SpaceEEC (approved):

iCrawl (approved):

PR #8917, opened 2022-12-09
feat(InteractionResponse): createdTimestamp

### LINKED ISSUE #8916: Add `createdTimestamp` property to `InteractionResponse`

discord.js

### Feature

As title. Since we do have an ID, the `createdTimestamp` property can be provided. Although the user can deconstruct the ID themselves, there is no reason to not have this property provided.

The property is added

_No response_

### Other context

_No response_

BenjammingKirby:
InteractionResponses have IDs, we can just destructure it and get createdAt and createdTimestamp, (it's a getter because other such properties on other classes were getters)
Close #8916 

### DIFF
--- a/packages/discord.js/src/structures/InteractionResponse.js
+++ b/packages/discord.js/src/structures/InteractionResponse.js
@@ -23,7 +23,7 @@ class InteractionResponse {
   }
 
   /**
-   * The timestamp the interaction's response was created at
+   * The timestamp the interaction response was created at
    * @type {number}
    * @readonly
    */
@@ -32,7 +32,7 @@ class InteractionResponse {
   }
 
   /**
-   * The time the interaction's response was created at
+   * The time the interaction response was created at
    * @type {Date}
    * @readonly
    */
--- a/packages/discord.js/src/structures/InteractionResponse.js
+++ b/packages/discord.js/src/structures/InteractionResponse.js
@@ -32,7 +32,7 @@ class InteractionResponse {
   }
 
   /**
-   * The time the interaction's resposne was created at
+   * The time the interaction's response was created at
    * @type {Date}
    * @readonly
    */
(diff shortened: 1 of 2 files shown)


packages/discord.js/src/structures/InteractionResponse.js
```diff
@@ -21,6 +22,24 @@ class InteractionResponse {
     this.client = interaction.client;
   }
 
+  /**
+   * The timestamp the interaction's response was created at
```
MrMythicalYT, line 26:
```suggestion
   * The timestamp the interaction response was created at
```

packages/discord.js/src/structures/InteractionResponse.js
```diff
@@ -21,6 +22,24 @@ class InteractionResponse {
...
+   * @type {number}
+   * @readonly
+   */
+  get createdTimestamp() {
+    return DiscordSnowflake.timestampFrom(this.id);
+  }
+
+  /**
+   * The time the interaction's response was created at
```
MrMythicalYT, line 35:
```suggestion
   * The time the interaction response was created at
```

MrMythicalYT (changes requested):

MaksiRose:
As far as I know, `createdTimestamp` is normally a normal property rather than a method. Is there a reason this is different here?

imranbarbhuiya:
This isn't a method but a getter that behaves like a property but u can return result dynamically

imranbarbhuiya:
using property will store the `Date` and `number` in memory and using getter, it'll generate when needed without storing. Ig this is the reason for not using property but not sure tho

Qjuh:
> using property will store the `Date` and `number` in memory and using getter, it'll generate when needed without storing. Ig this is the reason for not using property but not sure tho

That, but also that it doesn’t have to be calculated at all if it isn’t used.

kyranet (approved):

SpaceEEC (approved):

vladfrangu (approved):

PR #8928, opened 2022-12-14
refactor(sharding): use switch statement

twisttaan:

Internally we use switch else-where instead of `else if` statements, this also opens the gate up to more shard spawning implementations in the future without an `else if` mess.

- This PR **only** includes internal code changes.


SpaceEEC (approved):

vladfrangu (approved):

kyranet (approved):

Syjalo:
This needs a rebase

PR #8934, opened 2022-12-16
feat(CommandInteractionOptionResolver): add `channelTypes` option to `getChannel`

### LINKED ISSUE #8931: CommandInteractionOptionResolver#getChannel() channel type option

discord.js

### Feature

When a channel option's channel type has been restricted ([SlashCommandChannelOption#addChannelTypes()](https://discord.js.org/#/docs/builders/main/class/SlashCommandChannelOption?scrollTo=addChannelTypes)), the [CommandInteractionOptionResolver#getChannel()](https://discord.js.org/#/docs/discord.js/stable/class/CommandInteractionOptionResolver?scrollTo=getChannel) method may still return any of the channel types.

Similar to how you can already specify is the option was required, there should be a way to optionally specify what types of channels were allowed in the option.

_No response_

### Other context

_No response_

almeidx:
Resolves #8931


packages/discord.js/src/structures/CommandInteractionOptionResolver.js
```diff
@@ -142,12 +142,24 @@ class CommandInteractionOptionResolver {
    * Gets a channel option.
    * @param {string} name The name of the option.
    * @param {boolean} [required=false] Whether to throw an error if the option is not found.
+   * @param {ChannelType[]} [channelTypes=[]] The allowed types of channels. If empty, all channel types are allowed.
    * @returns {?(GuildChannel|ThreadChannel|APIChannel)}
    * The value of the option, or null if not set and not required.
    */
-  getChannel(name, required = false) {
+  getChannel(name, required = false, channelTypes = []) {
```
Jiralite, line 149: Thoughts on the second parameter being an object instead?

iCrawl: Breaking though

iCrawl (approved):

packages/discord.js/src/structures/CommandInteractionOptionResolver.js
Jiralite: We can implement it in a non-breaking way with a deprecation message emitting if one doesn't use an object... though this introduces some inconsistency in the second parameter of other methods.

packages/discord.js/typings/index.d.ts
```diff
@@ -1113,8 +1113,16 @@ export class CommandInteractionOptionResolver<Cached extends CacheType = CacheTy
   public getSubcommandGroup(required?: boolean): string | null;
   public getBoolean(name: string, required: true): boolean;
   public getBoolean(name: string, required?: boolean): boolean | null;
-  public getChannel(name: string, required: true): NonNullable<CommandInteractionOption<Cached>['channel']>;
-  public getChannel(name: string, required?: boolean): NonNullable<CommandInteractionOption<Cached>['channel']> | null;
+  public getChannel<T extends ChannelType = ChannelType>(
```
Jiralite, line 1116: How about `GuildChannelType` from discord-api-types instead?

almeidx: Resolved internally

kyranet (approved):

SpaceEEC (approved):

PR #8938, opened 2022-12-16
feat(core): add oauth2 api support

suneettipirneni:

Adds support for oauth token flows.


packages/core/src/api/oauth2.ts
```diff
@@ -0,0 +1,98 @@
...
+	}
+
+	/**
+	 * Fetches the bearer token for the current application
+	 *
+	 * @remarks
+	 * This is primarily used for testing purposes
+	 * @see {@link https://discord.com/developers/docs/topics/oauth2#client-credentials-grant}
+	 * @param options - The options for the client credentials grant request
+	 * @returns
```
almeidx, line 70:
```suggestion
```

almeidx (changes requested):

packages/core/src/api/index.ts
```diff
@@ -34,6 +36,8 @@ export class API {
 
 	public readonly invites: InvitesAPI;
 
+	public readonly oauth2: OAuth2API;
```
Syjalo, line 39: Shouldn't the property name be `oAuth2`?

Syjalo (changes requested):

packages/core/src/api/index.ts
suneettipirneni: While this is _technically_ correct, the word is most often used as a noun so we'll keep it as-is.

iCrawl (approved):

vladfrangu (approved):

kyranet (approved):

PR #7626, opened 2022-03-09
feat: backport handle zombie connection

legendhimslef:
**Please describe the changes this PR makes and why it should be merged:**
Backports the following pull requests to version 14: #7581 
**Status and versioning classification:**
- Code changes have been tested against the Discord API, or there are no code changes
- I know how to update typings and have done so, or typings don't need updating
- This PR changes the library's interface (methods or parameters added)

PR #8956, opened 2022-12-20
fix(WebSocketShard): either start close timeout or emit destroyed but never both

Qjuh:
#7626 introduced a `WebSocketShard#wsCloseTimeout()` to listen for zombie connections and reconnect when zombie connections get detected.

This fix was going a bit too far, as the timeout was also starting if `WebSocketShard#destroy()` was emitting the `WebSocketShard#destroyed` event. In those cases we don't need to listen for zombie connections anymore, as the `WebSocketManager`will reconnect the shard anyway.

But this lead to the Websocket connection being unneccessarily destroyed after the reconnect, because the timeout happened after the new connection already got made.

This PR now only starts the `wsCloseTimeout` if the `WebSocketShard#destroy()` method doesn't emit the `WebSocketShard#destroyed` event.

### DIFF
--- a/packages/discord.js/src/client/websocket/WebSocketShard.js
+++ b/packages/discord.js/src/client/websocket/WebSocketShard.js
@@ -840,8 +840,8 @@ class WebSocketShard extends EventEmitter {
         if (emit) {
           this._emitDestroyed();
         } else if (
-          this.connection?.readyState === WebSocket.CLOSING ||
-          this.connection?.readyState === WebSocket.CLOSED
+          this.connection.readyState === WebSocket.CLOSING ||
+          this.connection.readyState === WebSocket.CLOSED
         ) {
           this.closeEmitted = false;
           this.debug(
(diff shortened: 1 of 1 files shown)


packages/discord.js/src/client/websocket/WebSocketShard.js
```diff
@@ -835,23 +835,27 @@ class WebSocketShard extends EventEmitter {
...
           this.connection.terminate();
         }
+
         // Emit the destroyed event if needed
-        if (emit) this._emitDestroyed();
+        if (emit) {
+          this._emitDestroyed();
+        } else if (
+          this.connection?.readyState === WebSocket.CLOSING ||
+          this.connection?.readyState === WebSocket.CLOSED
```
legendhimself, line 844:
```suggestion
          this.connection.readyState === WebSocket.CLOSING ||
          this.connection.readyState === WebSocket.CLOSED
```
We don't need optional chaining because connection object is never null in that scope.

legendhimself (changes requested):

legendhimself:
These changes makes sense and looks good to me. Have you tested these changes?

SpaceEEC (approved):

vladfrangu (approved):

kyranet (approved):

PR #8971, opened 2022-12-25
refactor(Guild): Destructure object in guild editing

Jiralite:
This style of code isn't consistent with other parts of the library. I think destructuring the object here looks pretty neat and seems to result in code that's easier to look at. Personally, every time I've come across this, I think about doing this but end up not bothering. But now's the time, apparently!

I also reordered the parameters as they are in Discord's documentation.

There was also a pretty suspicious max-len rule modification here which I removed. It must have served an ideal purpose in the past, but presently, it doesn't seem to do much at all.

### DIFF
--- a/packages/discord.js/src/structures/Guild.js
+++ b/packages/discord.js/src/structures/Guild.js
@@ -740,24 +740,24 @@ class Guild extends AnonymousGuild {
    * @typedef {Object} GuildEditOptions
    * @property {string} [name] The name of the guild
    * @property {?GuildVerificationLevel} [verificationLevel] The verification level of the guild
+   * @property {?GuildDefaultMessageNotifications} [defaultMessageNotifications] The default message
+   * notification level of the guild
    * @property {?GuildExplicitContentFilter} [explicitContentFilter] The level of the explicit content filter
    * @property {?VoiceChannelResolvable} [afkChannel] The AFK channel of the guild
-   * @property {?TextChannelResolvable} [systemChannel] The system channel of the guild
    * @property {number} [afkTimeout] The AFK timeout of the guild
    * @property {?(BufferResolvable|Base64Resolvable)} [icon] The icon of the guild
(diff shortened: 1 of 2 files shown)


SpaceEEC (approved):

packages/discord.js/src/structures/Guild.js
```diff
@@ -788,51 +787,59 @@ class Guild extends AnonymousGuild {
...
+    banner,
+    systemChannel,
+    systemChannelFlags,
+    rulesChannel,
+    publicUpdatesChannel,
+    preferredLocale,
+    features,
+    description,
+    premiumProgressBarEnabled,
+    reason,
```
kyranet, line 810: There are some fields that are as-is with Discord's, we can rest those an map only the camelCase values (and `reason`).

Those would be `name`, `features`, and `description`.

packages/discord.js/src/structures/Guild.js
```diff
@@ -788,51 +787,59 @@ class Guild extends AnonymousGuild {
...
+  }) {
+    const {
+      client: { actions, channels, rest, users },
+      id,
+    } = this;
```
kyranet, line 815: ngl I'm not a fan of this, there are fields that can be lazily called below, and fields that are always called, but nevertheless, we don't do this pattern anywhere else.

kyranet (changes requested):

packages/discord.js/src/structures/Guild.js
Jiralite: Reverted in 5e088a929b3b151cf4a8aaa3e46f705c5bd85abe.

packages/discord.js/src/structures/Guild.js
Jiralite: Should be done with f9dcd8d5745d2a2cc9ec715b284dc7180799606d.

kyranet (approved):

vladfrangu (approved):

Issue #5267, opened 2021-01-26
Switch code patterns to go from mutating options, to keeping them read only


kyranet:
**Is your feature request related to a problem? Please describe.**
There are many issues that arise from mutating option objects, for example, they can mutate data in unexpected ways, see https://github.com/discordjs/discord.js/issues/5261#issuecomment-767810621. Furthermore, we do not document in which cases we mutate the options, and when we don't, which causes unconsistent behaviour.

We have been mutating options for a very long time, so the pattern is repeated and used across the entire library, which can take a lot of time to switch, and a single PR would probably be too big to test and review. However, it's also not consistent.

**Describe the ideal solution**
In fact, we also avoid mutating some options, which is the case for `Guild#edit`:

https://github.com/discordjs/discord.js/blob/9ffcd83027f0fc06d69df21475865ad55138de01/src/structures/Guild.js#L1032-L1080

For all intents and purposes, there are simpler alternatives which not only reduce code clutter, but also [optimise very well with V8](https://mathiasbynens.be/notes/shapes-ics). My favourite is to create a request object using the following pattern:

```javascript
edit(options, reason) {
	const data = {
		name: options.name,
		region: options.region,
		afk_timeout: data.afkTimeout,
		// ...
	};

	// Do something with data
}
```

This pattern is a huge simplification of the previous, which works flawlessly as undefined fields are not serialized into JSON:

```javascript
JSON.stringify({ foo: 'bar', hello: undefined });
// -> {"foo":"bar"}
```

Another pattern is to use object destructure, which is used in other places:

https://github.com/discordjs/discord.js/blob/9ffcd83027f0fc06d69df21475865ad55138de01/src/structures/GuildChannel.js#L476-L489

If we desire to future-proof and let users pass raw options (while we're at it), which can also work for properties that translate 1:1 with Discord's (in both key names and value types), the following pattern is also available, from a project I used to work on:

```javascript
async edit({ icon, splash, banner, ...options }, reason) {
	const data = {
		icon: icon ? await resolveImageToBase64(icon) : icon,
		splash: splash ? await resolveImageToBase64(splash) : splash,
		banner: banner ? await resolveImageToBase64(banner) : banner,
		// More properties that need translating
		...options
	};

	// Use data
}
```

[...]

PR #8972, opened 2022-12-25
refactor(GuildMemberManager): Tidy up fetching guild members

Jiralite:
This was refactored due to mutations & should help tackle #5267. Along the way, I noticed the following:

- `user`, `query`, and `nonce` in `FetchMembersOptions` were not marked as optional. All properties are optional as shown in the tests.
- `force` was documented in `FetchMembersOptions` which had no effect. There was likely confusion during implementation between `FetchMemberOptions` and `FetchMembersOptions`.
- `cache` was defaulted to `true`, but is redundant as it is already defaulted to `true` in `GuildMemberManager#_add()`.
- `_fetchMany()` sets `presences` to `false`, but [Discord already sets it to `false` for us](https://discord.com/developers/docs/topics/gateway-events#request-guild-members).

I moved the `_fetchSingle()` and `_fetchMany()` methods next to the `fetch()` method to make it easier to follow in the code.

Please let me know if anything is amiss!


packages/discord.js/src/managers/GuildMemberManager.js
```diff
@@ -206,18 +204,76 @@ class GuildMemberManager extends CachedManager {
...
-      if (!options.limit && !options.withPresences) return this._fetchSingle(options);
+    const { user, cache, force } = options;
+
+    const {
+      client: { users },
+    } = this;
+
+    const resolvedUser = users.resolveId(user ?? options);
+    if (resolvedUser) return this._fetchSingle({ user: resolvedUser, cache, force });
+    const resolvedUsers = user?.map(_user => users.resolveId(_user));
```
kyranet, line 215: If we don't destructure `users` out of `client`, we can give the variable a better name: `users`.

Also, this is quite assuming that if it couldn't resolve a userId, it's then an array. That won't translate well in TypeScript. 👀

packages/discord.js/src/managers/GuildMemberManager.js
```diff
@@ -206,18 +204,76 @@ class GuildMemberManager extends CachedManager {
...
+    limit = 0,
+    withPresences: presences,
+    users,
+    query,
+    time = 120e3,
+    nonce = DiscordSnowflake.generate().toString(),
+  } = {}) {
+    return new Promise((resolve, reject) => {
+      if (!query && !users) query = '';
+      if (nonce.length > 32) throw new DiscordjsRangeError(ErrorCodes.MemberFetchNonceLength);
```
kyranet, line 239: I never realised we're throwing the error incorrectly — it should be done with `return reject();`, or thrown in the outer scope with `Promise.reject`/`async`.

packages/discord.js/src/managers/GuildMemberManager.js
```diff
@@ -206,18 +204,76 @@ class GuildMemberManager extends CachedManager {
...
+          limit,
+        },
+      });
+      const fetchedMembers = new Collection();
+      let i = 0;
+      const handler = (members, _, chunk) => {
+        timeout.refresh();
+        if (chunk.nonce !== nonce) return;
+        i++;
+        for (const member of members.values()) {
```
kyranet, line 257: Isn't members an array? If so, wouldn't it suffice to iterate over `members` instead of `members.values()`?

packages/discord.js/src/managers/GuildMemberManager.js
```diff
@@ -206,18 +204,76 @@ class GuildMemberManager extends CachedManager {
...
+          query,
+          nonce,
+          limit,
+        },
+      });
+      const fetchedMembers = new Collection();
+      let i = 0;
+      const handler = (members, _, chunk) => {
+        timeout.refresh();
+        if (chunk.nonce !== nonce) return;
```
kyranet, line 255: I don't know if we should refresh the timeout on every event, rather than on every event sent with the same nonce.

packages/discord.js/src/managers/GuildMemberManager.js
```diff
@@ -206,18 +204,76 @@ class GuildMemberManager extends CachedManager {
...
+        for (const member of members.values()) {
+          fetchedMembers.set(member.id, member);
+        }
+        if (members.size < 1_000 || (limit && fetchedMembers.size >= limit) || i === chunk.count) {
+          clearTimeout(timeout);
+          this.client.removeListener(Events.GuildMembersChunk, handler);
+          this.client.decrementMaxListeners();
+          let fetched = fetchedMembers;
+          if (users && !Array.isArray(users) && fetched.size) fetched = fetched.first();
+          resolve(fetched);
```
kyranet, line 266: Can't we make this a ternary?

kyranet (changes requested):

packages/discord.js/src/managers/GuildMemberManager.js
Jiralite: I've removed the destructure with 207289fabb8067d1fb26a490566e42c95fbdc86f.

> Also, this is quite assuming that if it couldn't resolve a userId, it's then an array. That won't translate well in TypeScript. 👀

To be honest, this method performs a gateway request and a REST GET for a single guild member. To do both, `user` is either a single user or an array of users. It'd probably be simpler in the future to only have an array... or I was thinking to separate these two functionalities into their own methods.

packages/discord.js/src/managers/GuildMemberManager.js
Jiralite: This is actually a `Collection`:

https://github.com/discordjs/discord.js/blob/43ce2a572eb8977b6994680171ac0c5f9bda1703/packages/discord.js/src/client/websocket/handlers/GUILD_MEMBERS_CHUNK.js#L9-L11

packages/discord.js/src/managers/GuildMemberManager.js
Jiralite: Done with f1de458c111249b3e6fe861deae9f1434145c8b8.

packages/discord.js/src/managers/GuildMemberManager.js
Jiralite: Done with 6edd20402f629423bc994ef639ceb3f9764ffb5a.

packages/discord.js/src/managers/GuildMemberManager.js
Jiralite: I believe it should be confined at best to prevent unexpected behaviour. The `nonce` check has been prioritised with f06f747fad7f729f901eac0c7e8da0aed7903a66.

packages/discord.js/src/managers/GuildMemberManager.js
```diff
@@ -206,18 +204,70 @@ class GuildMemberManager extends CachedManager {
...
-        options.user = options.user.map(u => this.client.users.resolveId(u));
-        return this._fetchMany(options);
-      } else {
-        options.user = this.client.users.resolveId(options.user);
-      }
-      if (!options.limit && !options.withPresences) return this._fetchSingle(options);
+    const { user, limit, withPresences, cache, force } = options;
+    const resolvedUser = this.client.users.resolveId(user ?? options);
+    if (resolvedUser && !limit && !withPresences) return this._fetchSingle({ user: resolvedUser, cache, force });
+    const resolvedUsers = user?.map?.(users => this.client.users.resolveId(users)) ?? resolvedUser ?? undefined;
```
DraftProducts, line 210: Having "users" as param of a `.map` function looks weird, we should find another name. 
I propose "userId", a UserResolvable is not only "userId" but it looks better 🤔 
```suggestion
    const resolvedUsers = user?.map?.(userId => this.client.users.resolveId(userId)) ?? resolvedUser ?? undefined;
```

kyranet (approved):

packages/discord.js/src/managers/GuildMemberManager.js
```diff
@@ -206,18 +204,70 @@ class GuildMemberManager extends CachedManager {
...
-        options.user = options.user.map(u => this.client.users.resolveId(u));
-        return this._fetchMany(options);
-      } else {
-        options.user = this.client.users.resolveId(options.user);
-      }
-      if (!options.limit && !options.withPresences) return this._fetchSingle(options);
+    const { users, limit, withPresences, cache, force } = options;
+    const resolvedUser = this.client.users.resolveId(users ?? options);
+    if (resolvedUser && !limit && !withPresences) return this._fetchSingle({ user: resolvedUser, cache, force });
+    const resolvedUsers = users?.map?.(user => this.client.users.resolveId(user)) ?? resolvedUser ?? undefined;
```
vladfrangu, line 210:
```suggestion
    const resolvedUsers = users?.map?.(user => this.client.users.resolveId(user)) ?? [resolvedUser] ?? undefined;
```

packages/discord.js/src/managers/GuildMemberManager.js
```diff
@@ -206,18 +204,70 @@ class GuildMemberManager extends CachedManager {
...
+        timeout.refresh();
+        i++;
+        for (const member of members.values()) {
+          fetchedMembers.set(member.id, member);
+        }
+        if (members.size < 1_000 || (limit && fetchedMembers.size >= limit) || i === chunk.count) {
+          clearTimeout(timeout);
+          this.client.removeListener(Events.GuildMembersChunk, handler);
+          this.client.decrementMaxListeners();
+          resolve(users && !Array.isArray(users) && fetchedMembers.size ? fetchedMembers.first() : fetchedMembers);
```
vladfrangu, line 260: Hmm, does Discord support passing a string to user_ids? If yes, then disregard my suggested change above, but if not...this needs a fix

vladfrangu (changes requested):

packages/discord.js/src/managers/GuildMemberManager.js
Jiralite: It does! See https://discord.com/developers/docs/topics/gateway-events#request-guild-members-guild-request-members-structure.

SpaceEEC (approved):

vladfrangu (approved):

PR #8975, opened 2022-12-25
feat: Add `not_found` to guild member chunk data

Jiralite:
Implements `not_found` which was added in https://github.com/discord/discord-api-docs/pull/1078.

### DIFF
--- a/packages/discord.js/typings/index.d.ts
+++ b/packages/discord.js/typings/index.d.ts
@@ -229,6 +229,8 @@ declare module 'node:events' {
   }
 }
 
+type JSONValue = string | number | object | boolean | JSONValue[];
+
 //#region Classes
 
 export class Activity {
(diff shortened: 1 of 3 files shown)


packages/discord.js/typings/index.d.ts
```diff
@@ -229,6 +229,8 @@ declare module 'node:events' {
   }
 }
 
+type JSONValue = string | number | object | boolean | JSONValue[];
```
kyranet, line 232: Shouldn't this be exported?

Jiralite: Sure, I just assumed it was for internal use only.

kyranet: Well, it's documented in the docs without `@private`, but people may find it useful to have for handling the field's type when handling `notFound`.

Jiralite: Just exported the type and sorted it into the typedef region in the types!

kyranet (approved):

jaw0r3k:
Isnt it an array of ids? 
image

vladfrangu (approved):

SpaceEEC (approved):

PR #8981, opened 2022-12-26
fix(WebSocketShard): wsCloseTimeout in wrong if

Qjuh:
Fixes issues introduced by #8956 because it added the `wsCloseTimeout` even when the Websocket was already `Closed` before. This lead to the timeout trying to close the new connection after it successfully reconnected.
Also added a clearing of the `wsCloseTimeout` before making a new connection to prevent any other obscure occurences like this.

### DIFF
--- a/packages/discord.js/src/client/websocket/WebSocketShard.js
+++ b/packages/discord.js/src/client/websocket/WebSocketShard.js
@@ -820,14 +820,12 @@ class WebSocketShard extends EventEmitter {
       if (this.connection.readyState === WebSocket.OPEN) {
         this.connection.close(closeCode);
         this.debug(`[WebSocket] Close: Tried closing. | WS State: ${CONNECTION_STATE[this.connection.readyState]}`);
-        if (this.connection.readyState === WebSocket.CLOSING || this.connection.readyState === WebSocket.CLOSED) {
-          this.closeEmitted = false;
-          this.debug(
-            `[WebSocket] Adding a WebSocket close timeout to ensure a correct WS reconnect.
+        this.closeEmitted = false;
+        this.debug(
+          `[WebSocket] Adding a WebSocket close timeout to ensure a correct WS reconnect.
             Timeout: ${this.manager.client.options.closeTimeout}ms`,
-          );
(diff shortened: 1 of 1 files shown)


packages/discord.js/src/client/websocket/WebSocketShard.js
```diff
@@ -819,6 +820,12 @@ class WebSocketShard extends EventEmitter {
       if (this.connection.readyState === WebSocket.OPEN) {
         this.connection.close(closeCode);
         this.debug(`[WebSocket] Close: Tried closing. | WS State: ${CONNECTION_STATE[this.connection.readyState]}`);
+        this.closeEmitted = false;
+        this.debug(
+          `[WebSocket] Adding a WebSocket close timeout to ensure a correct WS reconnect.
+            Timeout: ${this.manager.client.options.closeTimeout}ms`,
+        );
+        this.setWsCloseTimeout(this.manager.client.options.closeTimeout);
```
legendhimself, line 828: These changes are very similar to the https://github.com/legendhimself/discord.js/blob/01797d3b0cd316a97853f34a971034c1904b4efb/packages/discord.js/src/client/websocket/WebSocketShard.js#L824
```js
this.connection.close(closeCode);
``` Once this ^ is invoked the connection is either CLOSING or CLOSED state.

Qjuh: It's only CLOSING state, never CLOSED (because it was OPEN before the .close() call), so yes it is similar, but not exactly the same.

legendhimself: `CLOSED` check was added for very large scaled bots (Just to cover every case). And after close() is invoked the ws waits for the close frame and then marks the connection `CLOSED`. Receiving that close frame differs (might be instant and can happen when the execution is in the current scope) hence that's why I had the check.

Qjuh: Websocket doesn’t wait for the close frame though… it’s an async callback. So that‘ll never happen while in this scope, status will always be CLOSING here…

DraftProducts:
This can be closed, it's superseded by #8989

PR #8982, opened 2022-12-26
typings(Attachment): make `attachment` private

net-tech:

Removes `<Attachment>.attachment`. This isn't documented anywhere in the [docs](https://discord.js.org/#/docs/discord.js/main/class/Attachment) and from my testing returns a URL like `https://cdn.discordapp.com/ephemeral-attachments/XXXXXXXXXXXXXXXXXX/XXXXXXXXXXXXXXXXXX/message.txt`. If this is here on purpose then the typing should be `string`, not `BufferResolvable | Stream`.


Jiralite:
I cannot seem to find where this is referenced:

https://github.com/discordjs/discord.js/blob/3160ab0fc75946041cbe7a093032ea10503dfd8c/packages/discord.js/src/structures/Attachment.js#L17

Perhaps it's remnants of an old refactor - possibly from being split into a builder? That property can be removed too. Pretty sure.

Jiralite:
Okay, I just found out how and where this is being used.

Let's say a message is created and it has an attachment. They would be created as an `Attachment`. The property `this.attachment = data.url;` is being defined here.

We modify that via `AttachmentBuilder.from()`. For example: `const attachment = AttachmentBuilder.from(message.attachments.first());`

Without that property, it looks like the `AttachmentBuilder` will not be properly created. Definitely something funny is going on here, but for now, I would keep that property as it is and remove only the typing.

net-tech:
@Jiralite Wait, revert structures change and keep the typing removed entirely or revert structures change and keep the typing but change it to `string`?

Jiralite:
Probably best to remove the typings altogether since there is no documentation for it.

Syjalo:
Isn't it better to make that property private?

net-tech:
> Isn't it better to make that property private?

Done.

Idris1401:
Maybe you should change the return type too, if it actually returns a `string` (or it doesn't matter?).

packages/discord.js/typings/index.d.ts
```diff
@@ -1972,7 +1972,7 @@ export class AttachmentBuilder {
 
 export class Attachment {
   private constructor(data: APIAttachment);
-  public attachment: BufferResolvable | Stream;
+  private attachment: BufferResolvable | Stream;
```
kyranet, line 1975: Considering it's initially set to `data.url`, which is a string:
```suggestion
  private attachment: BufferResolvable | Stream | string;
```

kyranet (changes requested):

kyranet (approved):

Jiralite (approved):
Do you mind also modifying the pull request title to reflect the changes too?

iCrawl (approved):

vladfrangu (approved):

SpaceEEC (approved):

PR #8988, opened 2022-12-28
chore: init @discordjs/next

ckohen:

Simple init of the typescript rewrite package to allow its parallel development.

All places that have per-package things have been updated, however

~~1. Will we actually have prereleases for this? If yes is it @discordjs/discord.js-next@dev or discord.js@next?~~
`@discordjs/next`
2. Do we actually want separate labels and such for this? I think yes at least for now, but thoughts appreciated.

~~Furthermore, one more little kink to iron out with this,
Every package re-exports -types causing a clash in reexporting these subpackages, solved at the moment by disabling import/export. This doesn't feel elegant, and there has to be a better way.
Furthermore, we cannot reexport both ws and core in full as star exports because core has an overwritten events map with the same name.~~

~~One solution to this is to reexport with named exports, so each subpackage has its name. Not perfect but it seems better. Again, thoughts appreciated.~~

*Solved via subpath exports*


packages/discord.js-next/package.json
```diff
@@ -0,0 +1,76 @@
+{
+	"name": "@discordjs/discord.js-next",
+	"version": "0.1.0",
+	"description": "A powerful typescript library for interacting with the Discord API",
```
SuperchupuDev, line 4:
```suggestion
	"description": "A powerful TypeScript library for interacting with the Discord API",
```

suneettipirneni:
Should TSR use node 18?

iCrawl:
@suneettipirneni Yes, 18 is LTS so we can def push for that.

packages/actions/src/uploadCoverage/action.yml
```diff
@@ -39,6 +39,12 @@ runs:
         files: ./packages/discord.js/coverage/cobertura-coverage.xml
         flags: discord.js
 
+    - name: Upload Next Coverage
+      uses: codecov/codecov-action@v3
+      with:
+        files: ./packages/next/coverage/cobertura-coverage.xml
+        flags: mext
```
JuanPablo2655, line 46:
```suggestion
        flags: next
```

JuanPablo2655 (changes requested):

packages/actions/src/uploadCoverage/action.yml
ckohen: beat me to it lol

packages/next/README.md
```diff
@@ -0,0 +1,60 @@
...
+See [the contribution guide][contributing] if you'd like to submit a PR.
+
+## Help
+
+If you don't understand something in the documentation, you are experiencing problems, or you just need a gentle
+nudge in the right direction, please don't hesitate to join our official [discord.js Server][discord].
+
+[website]: https://discord.js.org/
+[website-source]: https://github.com/discordjs/discord.js/tree/main/apps/website
+[documentation]: https://discord.js.org/#/docs/util
```
SuperchupuDev, line 51:
```suggestion
[documentation]: https://discord.js.org/#/docs/next
```

packages/next/README.md
```diff
@@ -0,0 +1,60 @@
...
+[guide]: https://discordjs.guide/
+[guide-source]: https://github.com/discordjs/guide
+[guide-update]: https://discordjs.guide/additional-info/changes-in-v14.html
+[discord]: https://discord.gg/djs
+[discord-api]: https://discord.gg/discord-api
+[source]: https://github.com/discordjs/discord.js/tree/main/packages/util
+[npm]: https://www.npmjs.com/package/@discordjs/util
```
SuperchupuDev, line 58:
```suggestion
[source]: https://github.com/discordjs/discord.js/tree/main/packages/next
[npm]: https://www.npmjs.com/package/@discordjs/next
```

PR #8996, opened 2022-12-31
fix(handlers): create burst handler for interaction callbacks

### LINKED ISSUE #9019: Request route for interaction responses includes the token

rest

### Issue description

The request route for an interaction reply of any kind includes the token, which makes it impossible to compare the route against a fixed string like we can do for others (which I believe is the point of having this route and not just the path). It would be nice to see the token be replaced with something like `:token` like is done for the id
image

### Code sample

```typescript
client.rest.on("response", req => console.log(req.route))
```

### Package version

1.4.0

### Node.js version

18.12.1

### Operating system

Windows 11

Low (slightly annoying)

_No response_

ckohen:

Interaction callbacks currently create a queue per callback, this is unnecessary and consumes excess memory for no reason.
The burst handler ignores all ratelimits (except for hitting 429s) because that's what callbacks do. 

What do we think of how the route is handled with this? As it currently is this resolves #9019 

_Also solves the testing issue on Node 18+_


packages/rest/src/lib/handlers/BurstHandler.ts
```diff
@@ -0,0 +1,184 @@
...
+import { RESTEvents } from '../utils/constants.js';
+import { onRateLimit, parseHeader, parseResponse, shouldRetry } from '../utils/utils.js';
+import type { IHandler, PolyFillAbortSignal } from './IHandler.js';
+
+/**
+ * The structure used to handle burst requests for a given bucket.
+ * Burst requests have no ratelimit handling but allow for pre- and post-processing
+ * of data in the same manner as sequentially queued requests.
+ *
+ * <info>This queue may still emit a rate limit error if an unexpected 429 is hit</info>
```
almeidx, line 17: Shouldn't this use a `@remark` instead of an info tag?

packages/rest/src/lib/handlers/BurstHandler.ts
```diff
@@ -0,0 +1,184 @@
...
+					`  Sublimit       : None`,
+				].join('\n'),
+			);
+
+			// We are bypassing all other limits, but an encountered limit should be respected (it's probably a non-punished ratelimit anyways)
+			await sleep(retryAfter);
+
+			// Since this is not a server side issue, the next request should pass, so we don't bump the retries counter
+			return this.runRequest(routeId, url, options, requestData, retries);
+		} else if (status >= 500 && status < 600) {
```
vladfrangu, line 158: While this is good to have, this is useless for interactions as if it times out, we're most likely past the 3s limit too. Leaving it up to you if we should keep this

ckohen: I left this in for the case where interaction callbacks aren't the only thing using this handler. That may never happen but if it does we let discord error which I think is better?

packages/rest/src/lib/handlers/BurstHandler.ts
```diff
@@ -0,0 +1,185 @@
...
+import { onRateLimit, parseHeader, parseResponse, shouldRetry } from '../utils/utils.js';
+import type { IHandler, PolyFillAbortSignal } from './IHandler.js';
+
+/**
+ * The structure used to handle burst requests for a given bucket.
+ * Burst requests have no ratelimit handling but allow for pre- and post-processing
+ * of data in the same manner as sequentially queued requests.
+ *
+ * @remarks
+ * This queue may still emit a rate limit error if an unexpected 429 is hit</info>
```
Jiralite, line 18:
```suggestion
 * This queue may still emit a rate limit error if an unexpected 429 is hit
```

packages/rest/src/lib/handlers/BurstHandler.ts
```diff
@@ -0,0 +1,185 @@
...
+		url: string,
+		options: Omit<Dispatcher.RequestOptions, 'method' | 'origin' | 'path'> &
+			Partial<Pick<Dispatcher.RequestOptions, 'method'>> & { dispatcher?: Dispatcher },
+		requestData: HandlerRequestData,
+	): Promise<Dispatcher.ResponseData> {
+		return this.runRequest(routeId, url, options, requestData);
+	}
+
+	/**
+	 * The method that actually makes the request to the api, and updates info about the bucket accordingly
```
Jiralite, line 67:
```suggestion
	 * The method that actually makes the request to the API and updates info about the bucket accordingly
```

packages/rest/src/lib/handlers/BurstHandler.ts
```diff
@@ -0,0 +1,185 @@
...
+	 *
+	 * @param routeId - The generalized api route with literal ids for major parameters
+	 * @param url - The fully resolved url to make the request to
```
Jiralite, line 70:
```suggestion
	 * @param url - The fully resolved URL to make the request to
```

packages/rest/src/lib/handlers/BurstHandler.ts
```diff
@@ -0,0 +1,185 @@
...
+			Partial<Pick<Dispatcher.RequestOptions, 'method'>> & { dispatcher?: Dispatcher },
+		requestData: HandlerRequestData,
+	): Promise<Dispatcher.ResponseData> {
+		return this.runRequest(routeId, url, options, requestData);
+	}
+
+	/**
+	 * The method that actually makes the request to the api, and updates info about the bucket accordingly
+	 *
+	 * @param routeId - The generalized api route with literal ids for major parameters
```
Jiralite, line 69:
```suggestion
	 * @param routeId - The generalized API route with literal ids for major parameters
```

packages/rest/src/lib/handlers/BurstHandler.ts
```diff
@@ -0,0 +1,185 @@
...
+					`  Bucket         : ${routeId.bucketRoute}`,
+					`  Major parameter: ${routeId.majorParameter}`,
+					`  Hash           : ${this.hash}`,
+					`  Limit          : ${Number.POSITIVE_INFINITY}`,
+					`  Retry After    : ${retryAfter}ms`,
+					`  Sublimit       : None`,
+				].join('\n'),
+			);
+
+			// We are bypassing all other limits, but an encountered limit should be respected (it's probably a non-punished ratelimit anyways)
```
Jiralite, line 154:
```suggestion
			// We are bypassing all other limits, but an encountered limit should be respected (it's probably a non-punished rate limit anyways)
```

didinele (approved):

packages/rest/src/lib/handlers/BurstHandler.ts
```diff
@@ -0,0 +1,147 @@
...
+	}
+
+	/**
+	 * {@inheritDoc IHandler.queueRequest}
+	 */
+	public async queueRequest(
+		routeId: RouteData,
+		url: string,
+		options: Omit<Dispatcher.RequestOptions, 'method' | 'origin' | 'path'> &
+			Partial<Pick<Dispatcher.RequestOptions, 'method'>> & { dispatcher?: Dispatcher },
```
vladfrangu, line 58:
```suggestion
		options: RequestOptions,
```

vladfrangu (changes requested):

vladfrangu (approved):

kyranet (approved):

PR #9004, opened 2023-01-02
feat(ws): custom workers

didinele:
Another big stepping stone in prod usability - this time for folks trying to scale.


packages/ws/src/utils/WorkerBootstrapper.ts
```diff
@@ -0,0 +1,160 @@
...
+					}
+				}
+			});
+	}
+
+	/**
+	 * Bootstraps the worker thread with the provided options
+	 */
+	public async bootstrap(options: BootstrapOptions = {}): Promise<void> {
+		options.forwardEvents ??= Object.values(WebSocketShardEvents);
```
kyranet, line 125: We shouldn't mutate options.

packages/ws/src/utils/WorkerBootstrapper.ts
```diff
@@ -0,0 +1,160 @@
...
+
+		if (isMainThread) {
```
kyranet, line 127: Can't we make this the first thing? Or even throw in the constructor?

kyranet (changes requested):
By the way, how do we use `defaultWorker`?

didinele:
> By the way, how do we use `defaultWorker`?

It's defaulted to here: https://github.com/discordjs/discord.js/blob/7156161d58054a69443addfade78a42f6a52c825/packages/ws/src/strategies/sharding/WorkerShardingStrategy.ts#L177

Reckon a way to explicitly provide it is needed as well?

kyranet:
Yes — if we use export paths, we can probably allow `@discordjs/ws/defaultWorker` and use `require.resolve` to get it if it's not a path. Might also be useful for third-party plugins that define a custom strategy.

packages/ws/src/utils/WorkerBootstrapper.ts
```diff
@@ -0,0 +1,160 @@
...
+	}
+
+	/**
+	 * Bootstraps the worker thread with the provided options
+	 */
+	public async bootstrap(options: BootstrapOptions = {}): Promise<void> {
+		// Start by initializing the shards
+		for (const shardId of this.data.shardIds) {
+			const shard = new WebSocketShard(new WorkerContextFetchingStrategy(this.data), shardId);
+			for (const event of (options.forwardEvents ??= Object.values(WebSocketShardEvents))) {
```
kyranet, line 134: You're still mutating. `??=` → `??`.

To be safe, we should do `Readonly<BootstrapOptions>`, akin to what we do in `/core`.

kyranet (changes requested):

packages/ws/src/strategies/sharding/WorkerShardingStrategy.ts
```diff
@@ -1,5 +1,5 @@
 import { once } from 'node:events';
-import { join } from 'node:path';
+import path, { join } from 'node:path';
```
kyranet, line 2: Why are you importing path instead of its individual methods? Considering you do that already for `join`

packages/ws/src/strategies/sharding/WorkerShardingStrategy.ts
```diff
@@ -179,6 +173,61 @@ export class WorkerShardingStrategy implements IShardingStrategy {
...
+			.on('message', async (payload: WorkerRecievePayload) => this.onMessage(worker, payload));
+
+		this.#workers.push(worker);
+		for (const shardId of workerData.shardIds) {
+			this.#workerByShardId.set(shardId, worker);
+		}
+	}
+
+	private resolveWorkerPath(): string {
+		if (!this.options.workerPath) {
```
kyranet, line 199: Instead of accessing the property over and over, we can define it as a variable.

packages/ws/src/utils/WorkerBootstrapper.ts
```diff
@@ -0,0 +1,160 @@
...
+		}
+	}
+
+	/**
+	 * Helper method to initiate a shard's connection process
+	 */
+	protected async connect(shardId: number): Promise<void> {
+		const shard = this.shards.get(shardId);
+		if (!shard) {
+			throw new Error(`Shard ${shardId} does not exist`);
```
kyranet, line 57: Shouldn't we use [`RangeError`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RangeError)?

kyranet (changes requested):

packages/ws/src/utils/WorkerBootstrapper.ts
didinele: We already decided against that - this same check takes places in many other places. It doesn't make a lot of sense to use here, it's confusing if say, the manager in question is responsible for shards `1`, `3`, and `5` and, say, an op is attempted on shard `2`, that's not really a "range" error, IMO.

kyranet: > The `RangeError` object indicates an error when __a value is not in the set__ or range __of allowed values__.

While the class name is misleading, it's still talking about a value within the allowed values, whether or not that's within a range or within a set of ranges.

kyranet (approved):

didinele:
rebased

iCrawl (approved):

SpaceEEC (approved):

PR #9016, opened 2023-01-06
refactor(GuildChannel)!: default setParent's lockPermissions to false

ImRodry:

This PR changes the default for lockPermissions on the GuildChannel#setParent method to false and also fixes the documentation. The reasoning for this change is that, when using this method, users are most likely only expecting to change the parent of the channel, as happens with all the other setX methods on the library as a whole. Giving them the permission to lockPermissions is nice but doing this silently might result in unintentional changes (has happened to me countless times) and I believe it's better to let the user explicitly add this in, but I'm open to feedback if you guys disagree


Idris1401:
When we create a channel on a category Discord is automatically syncing the permissions or when we move a channel to a category Discord puts more emphasis on the sync button so I would say that we should keep it to `true`.

ImRodry:
I know that, but my argument is simply that these set methods are supposed to change only what is in their name. Also even though the sync permissions button is given more emphasis on the client, it still required explicit action from the user, which is what I’m trying to do here. The goal here is to mainly prevent accidental changes for users who are unaware of this default

vladfrangu (approved):

Jiralite (approved):

packages/discord.js/src/structures/GuildChannel.js
```diff
@@ -327,11 +327,16 @@ class GuildChannel extends BaseChannel {
...
+   *   .then(channel => console.log(`New parent of ${channel.name}: ${channel.parent.name}`))
+   *   .catch(console.error);
+   * @example
+   * // Move a channel and sync its permissions with the parent
+   * message.channel.setParent('355908108431917066', { lockPermissions: true })
+   *   .then(channel => console.log(`Moved ${message.channel.name} to ${channel.parent.name}`))
    *   .catch(console.error);
    */
-  setParent(channel, { lockPermissions = true, reason } = {}) {
+  setParent(channel, { lockPermissions = false, reason } = {}) {
```
almeidx, line 339: What's Discord's default behaviour? Do we need this default value at all?

ImRodry: The default API behavior is what this PR is changing. Setting lockPermissions to true changes the permissionOverwrites as calculated by djs, setting it to false only changes the parent and nothing else. On the client, it usually prompts the user whether they want to sync or not.
In summary: false is the default behavior

kyranet (approved):

PR #9026, opened 2023-01-07
refactor: Use `deprecate()` directly

Jiralite:
Instead of modifying the class, it would probably be a good idea to just call `deprecate()` directly. The added tags are needed as it otherwise crashes the documentation generator. I also took this opportunity to stylise the other deprecation messages like these for consistency.


packages/discord.js/src/structures/BaseInteraction.js
```diff
@@ -270,16 +270,18 @@ class BaseInteraction extends Base {
...
+   * @memberof BaseInteraction
+   * @instance
    * @returns {boolean}
-   *
    * @deprecated Use {@link BaseInteraction#isStringSelectMenu} instead.
    */
-  isSelectMenu() {
-    return this.isStringSelectMenu();
-  }
+  isSelectMenu = deprecate(
```
kyranet, line 281: This is called on every single instance, which resets the deprecation warning state.

kyranet (changes requested):

packages/discord.js/src/structures/BaseInteraction.js
Jiralite: That's unfortunate. Undid this with ea9ca1ea528a0cf197baca8ced5794b39ecbba72.

kyranet (approved):

iCrawl (approved):

vladfrangu (approved):

SpaceEEC (approved):

PR #9027, opened 2023-01-07
refactor(website): extract layouts and use more server components

### LINKED ISSUE #8920: Render `@see` tags in documentation

N/A

### Feature

The `@see` tags need to be seen by those that wish to see them. Currently, no one can see the `@see` tags.

Parse `@see` tags so they can be viewed in the documentation.

_No response_

### Other context

_No response_

### LINKED ISSUE #8997: Documentation seemingly does not show methods or properties derived via mixins

N/A

### Feature

Mixins seem to be an exception in our documentation.

An example is the [`SlashCommandStringOption`](https://discordjs.dev/docs/packages/builders/main/SlashCommandStringOption:Class) class. This does not have the `addChoices()` method, but it does exist in the code:

https://github.com/discordjs/discord.js/blob/d3e9f2a355a1f5272d62a507eb6ecd8808904fff/packages/builders/src/interactions/slashCommands/options/string.ts#L10-L11

Mixins seem to be the commonality here, so I would want this to be supported through one shape or another.

_No response_

### Other context

_No response_

suneettipirneni:

This PR does a large amount of refactoring the website components and associated pages and layouts.

Firstly, server components are used where they are deemed effective. This is usually for components that are static and don't rely on client-side state. A positive side-effect of this change is the amount of serialization has been reduced dramatically. In our current website, we have to serialize *all* `ApiItems` directly when pages are generated. This was because the network boundary was between `page.tsx` and every other component. With the refactor the network boundary is between a server component and and client component which means we can pass class instances directly to server components. 

Secondly, layouts have been extracted or rather "moved up". Currently our main documentation layout rerenders anytime a new item is selected. This result in things like the sidebar and nav bar flickering and scroll state being reset. This occurs because we used [...slug] as a catch-all route, meaning next considered any changes to parameter in this route a signal to re-render the documentation shell. To fix this I simply pulled routes up that have persistent layouts (for example the documentation shell). This resulted in scroll states being persisted between elements and less re-renders for items that use the same props.

>  **Note**
> When looking at the code you may think that we are doing a lot of redundant fetches for each of the nested layouts, however this actually isn't the case. The same fetches use across layouts are deduped, so the requests only happen once as described [here](https://beta.nextjs.org/docs/data-fetching/fundamentals#automatic-fetch-request-deduping)

Finally, I added some requested changes to the website (scope creep). `@see` block rendering and proper documentation for classes that use mixins.

Closes #8920 
Closes #8997


packages/ui/vite.config.ts.timestamp-1672422498444.mjs
```diff
@@ -0,0 +1,31 @@
+// vite.config.ts
```
almeidx, line 1: This file shouldn't be here. Should probably also add it to the root .gitignore

packages/api-extractor-utils/src/tsdoc/LinkTagCommentNode.ts
```diff
@@ -23,6 +23,7 @@ export function genLinkToken(
 ): LinkTagCodeLink | null {
 	const item = model.resolveDeclarationReference(ref, context ?? undefined).resolvedApiItem ?? null;
 
+	console.log(item);
```
kyranet, line 26: 👀

apps/website/package.json
```diff
@@ -54,7 +54,7 @@
 		"ariakit": "^2.0.0-next.41",
 		"cmdk": "^0.1.20",
 		"meilisearch": "^0.30.0",
-		"next": "^13.0.7-canary.1",
+		"next": "latest",
```
kyranet, line 57: I don't think we should use `latest`, but rather, a specific version.

kyranet (dismissed):

apps/website/next.config.js
```diff
@@ -20,6 +20,7 @@ export default withBundleAnalyzer({
 	},
 	cleanDistDir: true,
 	outputFileTracing: true,
+	productionBrowserSourceMaps: true,
```
iCrawl, line 23: 🔕

apps/website/design.txt
```diff
@@ -0,0 +1,41 @@
+function DocumentationHeader(item={APIItem}) { // Server
```
iCrawl, line 1: Whats this

apps/website/src/app/docAPI.ts
```diff
@@ -0,0 +1,26 @@
+import { readFile } from 'node:fs/promises';
+import { join } from 'node:path';
+
+export async function fetchVersions(packageName: string): Promise<string[]> {
+	const response = await fetch(`https://docs.discordjs.dev/api/info?package=${packageName}`, {
+		// next: { revalidate: 3_600 },
```
iCrawl, line 6: No revalidation?

apps/website/package.json
```diff
@@ -54,7 +54,7 @@
 		"ariakit": "^2.0.0-next.41",
 		"cmdk": "^0.1.20",
 		"meilisearch": "^0.30.0",
-		"next": "^13.0.7-canary.1",
+		"next": "canary",
```
iCrawl, line 57: Should pin a specific canary version, if you aren't waiting for something specific that is.

apps/website/src/app/docs/packages/[package]/[version]/[item]/page.tsx
```diff
@@ -0,0 +1,104 @@
...
+import { Class } from '~/components/model/Class';
+import { Interface } from '~/components/model/Interface';
+import { TypeAlias } from '~/components/model/TypeAlias';
+import { Variable } from '~/components/model/Variable';
+import { Enum } from '~/components/model/enum/Enum';
+import { Function } from '~/components/model/function/Function';
+import { OVERLOAD_SEPARATOR } from '~/util/constants';
+import { findMember, findMemberByKey } from '~/util/model.server';
+
+export const dynamicParams = true;
```
Comment (iCrawl) [inline apps/website/src/app/docs/packages/[package]/[version]/[item]/page.tsx:26]: This is the default value for `dynamicParams` already.

apps/website/src/app/docs/packages/[package]/[version]/[item]/page.tsx
```diff
@@ -0,0 +1,104 @@
...
+		default:
+			return <div>Cannot render that item type</div>;
+	}
+}
+
+export default async function Page({ params }: { params: ItemRouteParams }) {
+	const member = await fetchMember(params);
+
+	return (
+		<div>
```
Comment (iCrawl) [inline apps/website/src/app/docs/packages/[package]/[version]/[item]/page.tsx:87]: Is that wrapper `div` necessary here?

apps/website/src/app/docs/packages/[package]/[version]/layout.tsx
```diff
@@ -0,0 +1,136 @@
+import { createApiModel } from '@discordjs/scripts';
+import type { ApiFunction, ApiItem } from '@microsoft/api-extractor-model';
+import Image from 'next/image';
+import type { PropsWithChildren } from 'react';
+import vercelLogo from '../../../../../assets/powered-by-vercel.svg';
```
Comment (iCrawl) [inline apps/website/src/app/docs/packages/[package]/[version]/layout.tsx:5]:
```suggestion
import vercelLogo from '~/assets/powered-by-vercel.svg';
```

Maybe?

apps/website/src/app/docs/packages/[package]/[version]/layout.tsx
```diff
@@ -0,0 +1,136 @@
...
+import { fetchModelJSON, fetchVersions } from '~/app/docAPI';
+import { Header } from '~/components/Header';
+import { Nav } from '~/components/Nav';
+import type { SidebarSectionItemData } from '~/components/Sidebar';
+import { resolveItemURI } from '~/components/documentation/util';
+import { N_RECENT_VERSIONS, PACKAGES } from '~/util/constants';
+
+export const dynamicParams = true;
```
Comment (iCrawl) [inline apps/website/src/app/docs/packages/[package]/[version]/layout.tsx:13]: Default value is true already.

apps/website/src/app/docs/packages/[package]/[version]/layout.tsx
```diff
@@ -0,0 +1,136 @@
...
+	const members = entry?.members.filter((member) => {
+		if (member.kind !== 'Function') {
+			return true;
+		}
+
+		return (member as ApiFunction).overloadIndex === 1;
+	});
+
+	return (
+		<div>
```
Comment (iCrawl) [inline apps/website/src/app/docs/packages/[package]/[version]/layout.tsx:60]: Wrapper `div` can be a `Fragment/<>` instead?

apps/website/src/app/docs/packages/[package]/page.tsx
```diff
@@ -2,15 +2,9 @@ import { VscArrowLeft } from '@react-icons/all-files/vsc/VscArrowLeft';
...
-import { notFound } from 'next/navigation';
-import { PACKAGES } from '~/util/constants';
 
 async function getData(pkg: string) {
-	if (!PACKAGES.includes(pkg)) {
-		notFound();
-	}
-
-	const res = await fetch(`https://docs.discordjs.dev/api/info?package=${pkg}`, { next: { revalidate: 3_600 } });
+	const res = await fetch(`https://docs.discordjs.dev/api/info?package=${pkg}`, {});
```
Comment (iCrawl) [inline apps/website/src/app/docs/packages/[package]/page.tsx:7]: No revalidation?

apps/website/src/app/docs/packages/[package]/page.tsx
```diff
@@ -30,7 +24,7 @@ export default async function Page({ params }: { params: { package: string } })
 				{data.map((version) => (
 					<Link
 						className="dark:bg-dark-400 dark:border-dark-100 dark:hover:bg-dark-300 dark:active:bg-dark-200 focus:ring-width-2 focus:ring-blurple flex h-11 transform-gpu cursor-pointer select-none appearance-none flex-col place-content-center rounded border border-neutral-300 bg-transparent p-4 text-base font-semibold leading-none text-black outline-0 hover:bg-neutral-100 focus:ring active:translate-y-px active:bg-neutral-200 dark:text-white"
-						href={`/docs/packages/${params.package}/${version}`}
+						href={`/docs/packages/${params.package}/${version}/`}
```
Comment (iCrawl) [inline apps/website/src/app/docs/packages/[package]/page.tsx:27]: Trailing slash?

apps/website/src/app/docs/packages/page.tsx
```diff
@@ -44,7 +27,7 @@ export default async function Page() {
 				{PACKAGES.map((pkg) => (
 					<Link
 						className="dark:bg-dark-400 dark:border-dark-100 dark:hover:bg-dark-300 dark:active:bg-dark-200 focus:ring-width-2 focus:ring-blurple flex h-11 transform-gpu cursor-pointer select-none appearance-none flex-row place-content-between rounded border border-neutral-300 bg-transparent p-4 text-base font-semibold leading-none text-black outline-0 hover:bg-neutral-100 focus:ring active:translate-y-px active:bg-neutral-200 dark:text-white"
-						href={`/docs/packages/${pkg}/${findLatestVersion(pkg)?.version ?? 'main'}`}
+						href={`/docs/packages/${pkg}/`}
```
iCrawl, line 30: Trailing slash?

apps/website/src/app/layout.tsx
```diff
@@ -7,6 +7,8 @@ import '../styles/unocss.css';
 import '../styles/cmdk.css';
 import '../styles/main.css';
 
+export const dynamicParams = true;
```
iCrawl, line 10: Default value is true.

apps/website/src/components/Anchor.tsx
```diff
@@ -0,0 +1,13 @@
+import { FiLink } from '@react-icons/all-files/fi/FiLink';
+
+export function Anchor({ href }: { href: string }) {
+	return (
+		<a
+			aria-label="Anchor"
```
iCrawl, line 6: I don't think that is necessary, native HTML elements already have proper ARIA tags applied.

apps/website/src/components/ExcerptText.tsx
```diff
@@ -0,0 +1,42 @@
...
+			{excerpt.spannedTokens.map((token) => {
+				if (token.kind === ExcerptTokenKind.Reference) {
+					const item = model.resolveDeclarationReference(token.canonicalReference!, model).resolvedApiItem;
+
+					if (!item) {
+						return token.text;
+					}
+
+					return (
+						<ItemLink className="color-blurple" itemURI={resolveItemURI(item)} key={item.containerKey}>
```
iCrawl, line 32:
```suggestion
						<ItemLink className="text-blurple" itemURI={resolveItemURI(item)} key={item.containerKey}>
```

iCrawl: Double check all instances where we use `color-` instead of `text-`

apps/website/src/components/Header.tsx
```diff
@@ -86,7 +86,7 @@ export function Header() {
 					>
 						<VscMenu size={24} />
 					</Button>
-					<div className="hidden md:flex md:flex-row md:overflow-hidden">{breadcrumbs}</div>
+					{/* <div className="hidden md:flex md:flex-row md:overflow-hidden">{breadcrumbs}</div> */}
```
iCrawl, line 89: ?

apps/website/src/components/Panel.tsx
```diff
@@ -0,0 +1,8 @@
+export function Panel({ children }: { children: React.ReactNode }) {
```
iCrawl, line 1:
```suggestion
import type { PropsWithChildren } from 'react';

export function Panel({ children }: PropsWithChildren) {
```

apps/website/src/components/Scrollbars.tsx
```diff
@@ -0,0 +1,8 @@
+'use client';
+
+import type { ScrollbarProps } from 'react-custom-scrollbars-2';
+import { Scrollbars as ReactScrollBars2 } from 'react-custom-scrollbars-2';
+
+export function Scrollbars(props: ScrollbarProps) {
+	return <ReactScrollBars2 {...props} />;
```
iCrawl, line 7:
```suggestion
import { Scrollbars as ReactScrollbars2 } from 'react-custom-scrollbars-2';

export function Scrollbars(props: ScrollbarProps) {
	return <ReactScrollbars2 {...props} />;
```

apps/website/src/components/documentation/ExtendsText.tsx
```diff
@@ -0,0 +1,51 @@
+import type { ApiInterface, Excerpt } from '@microsoft/api-extractor-model';
+import { type ApiClass, ApiItemKind } from '@microsoft/api-extractor-model';
```
iCrawl, line 2:
```suggestion
import type { ApiClass, ApiInterface, Excerpt } from '@microsoft/api-extractor-model';
import { ApiItemKind } from '@microsoft/api-extractor-model';
```

apps/website/src/components/documentation/ImplementsText.tsx
```diff
@@ -0,0 +1,3 @@
+import type { ApiClass } from '@microsoft/api-extractor-model';
+
+export function ImplementsText({ item }: { item: ApiClass }) {}
```
iCrawl, line 3: Is this incomplete still?

apps/website/src/components/model/method/MethodName.tsx
```diff
@@ -0,0 +1,68 @@
...
+				return `${prev}, ${cur.isOptional ? `${cur.name}?` : cur.name}`;
+			}, '')})`,
+		[],
+	);
+
+	return (
+		<div className="flex flex-col">
+			<div className="flex flex-col gap-2 md:-ml-9 md:flex-row md:place-items-center">
+				<a
+					aria-label="Anchor"
```
iCrawl, line 31: Mentioned above already, not needed.

iCrawl: Also I saw a component like this earlier, can probably re-use it?

iCrawl (changes requested):

PR #9039, opened 2023-01-10
Chane "PermissionOverwrites" to CamelCase naming

gorgulenkozxc:
https://discordjs.guide/additional-info/changes-in-v14.html#permissionoverwritesmanager


Jiralite:
Thank you for your contribution. However, we do not use the stable branch (this is soon to be removed).

We version our branches (version 14 is main, version 13, version 12 etc.), and this doesn't appear to be a problem on any.

PR #9044, opened 2023-01-11
feat(Collector): add lastCollectedTimestamp

ImRodry:

This PR adds a new property and a new getter to the Collector class to store the timestamp at which the collector last collected an item. This is useful for determining when the collector will end when an idle timeout is specified as there is currently no reliable way to get this information from Node without using private properties of the Timeout class and weird hacks


jaw0r3k:
you should chanhe the name

ImRodry:
The formatting errors are due to an update to prettier that changed a rule. Should I commit those changes?

packages/discord.js/src/structures/interfaces/Collector.js
```diff
@@ -97,6 +97,20 @@ class Collector extends EventEmitter {
...
 
+  /**
+   * The timestamp at which this collector last collected an item
+   * @type {number|null}
+   */
+  lastCollectedTimestamp = null;
+
+  /**
+   * The Date at which this collector last collected an item
+   * @type {Date|null}
```
jaw0r3k, line 108:
```suggestion
   * @type {?Date}
```

packages/discord.js/src/structures/interfaces/Collector.js
```diff
@@ -97,6 +97,20 @@ class Collector extends EventEmitter {
     if (options.idle) this._idletimeout = setTimeout(() => this.stop('idle'), options.idle).unref();
   }
 
+  /**
+   * The timestamp at which this collector last collected an item
+   * @type {number|null}
```
Syjalo, line 102:
```suggestion
   * @type {?number}
```

packages/discord.js/src/structures/interfaces/Collector.js
Syjalo, line 108:
```suggestion
   * @type {?Date}
```

Syjalo (changes requested):

packages/discord.js/src/structures/interfaces/Collector.js
ImRodry: Some properties on this class are typed as |null others with ? I’m not sure which one to use but |null seems more appropriate

jaw0r3k: Yeah also for me |null is more clear

Syjalo: `|null` is used only in 2 places. In all other places `?` is used.

https://github.com/KhafraDev/discord.js/blob/5befbc3773a1cf33614bf900f8301686130b33c7/packages/discord.js/src/structures/interfaces/Collector.js#L83

https://github.com/ckohen/discord.js/blob/8dfd0037b53ec6ef0900a53a0fb1b68fa2cbaf72/packages/discord.js/src/structures/ApplicationCommand.js#L147

ImRodry: Exactly, it’s used in this file. I think `?` is for optional properties which this one isn’t

kyranet: `?number` and `number|null` are equivalent, the `?` doesn't denote optionality (like in TypeScript), but nullability:

image

packages/discord.js/src/structures/interfaces/Collector.js
```diff
@@ -97,6 +97,20 @@ class Collector extends EventEmitter {
     if (options.idle) this._idletimeout = setTimeout(() => this.stop('idle'), options.idle).unref();
   }
 
+  /**
+   * The timestamp at which this collector last collected an item
+   * @type {?number}
+   */
+  lastCollectedTimestamp = null;
```
kyranet, line 104: All the properties are assigned in the constructor, not as fields, this should be moved inside the constructor.

kyranet (changes requested):

packages/discord.js/src/structures/interfaces/Collector.js
ImRodry: don't really get why but I'll change this

ImRodry:
would love to know how in the world I removed those review requests lol

kyranet (approved):

Syjalo:
Probably you can also add `endingAt`

ImRodry:
maybe a new pr

SpaceEEC (approved):

vladfrangu (approved):

PR #9048, opened 2023-01-12
feat(website): add support for source file links

suneettipirneni:

In addition to providing links for each item page, there have been a few changes to CI. Instead of generating documentation based on output provided by tsup, we instead use the build output of *only declaration files and declaration maps*. The reasoning for this is tsup doesn't support declaration maps, using `tsc` to add declarations after tsup has compiled overwrites the rolled up `index.d.ts` files and maps to `tsc` output not `tsup` output. This build output goes into `dist-docs` and api-extractor reads the model from there.


suneettipirneni:
Note, you won't see the links on the preview since the `api.json` models are older from the api.

PR #8567, opened 2022-08-28
feat(GuildChannelManager): add `.addFollower()` method

Syjalo:
**Please describe the changes this PR makes and why it should be merged:**
This PR adds `GuildChannelManager#addFollower()` method.

**Status and versioning classification:**
- Code changes have been tested against the Discord API, or there are no code changes
- I know how to update typings and have done so, or typings don't need updating
- This PR changes the library's interface (methods or parameters added)

PR #9050, opened 2023-01-12
feat v13: add `addFollower` on GuildChannelManager

jaw0r3k:
 Backports #8567
  It includes no braking changes, because `NewsChannel#addFolower` cant be changed to return webook id :(

### DIFF
--- a/src/structures/NewsChannel.js
+++ b/src/structures/NewsChannel.js
@@ -9,8 +9,8 @@ const { Error } = require('../errors');
  */
 class NewsChannel extends BaseGuildTextChannel {
   /**
-   * <info>This method returns itself. To get created webhook id, use `GuildChannelManager#addFolower`</info>
    * Adds the target to this channel's followers.
+   * <info>If you need the created webhook id, use {@link GuildChannelManager#addFollower}.</info>
    * @param {TextChannelResolvable} channel The channel where the webhook should be created
    * @param {string} [reason] Reason for creating the webhook
    * @returns {Promise<NewsChannel>}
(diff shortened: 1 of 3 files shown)


src/structures/NewsChannel.js
```diff
@@ -9,6 +9,7 @@ const { Error } = require('../errors');
  */
 class NewsChannel extends BaseGuildTextChannel {
   /**
+   * <info>This method returns itself. To get created webhook id, use `GuildChannelManager#addFolower`</info>
    * Adds the target to this channel's followers.
```
Jiralite, line 13: > This method returns itself.

I'm not sure how to describe it, but this sounds confusing. This should be rephrased and placed below the description.

```suggestion
   * Adds the target to this channel's followers.
   * <info>If you need the created webhook id, use {@link GuildChannelManager#addFolower}.</info>
```

Jiralite (changes requested):

src/structures/NewsChannel.js
```diff
@@ -10,6 +10,7 @@ const { Error } = require('../errors');
 class NewsChannel extends BaseGuildTextChannel {
   /**
    * Adds the target to this channel's followers.
+   * <info>If you need the created webhook id, use {@link GuildChannelManager#addFollower}.</info>
```
vladfrangu, line 13: Maybe instead mark this one as deprecated and point to the new method

vladfrangu (approved):

src/structures/NewsChannel.js
Jiralite: You mean mark the whole method as deprecated? I believe this method will stay as a helper, so I'm not sure that's the best approach.

SpaceEEC (approved):

kyranet (approved):

PR #9053, opened 2023-01-12
fix(next): fix type declarations

Syjalo:

This PR adds `typesVersions` property to `package.json` to make type declarations work as needed.

### DIFF
--- a/packages/next/package.json
+++ b/packages/next/package.json
@@ -28,6 +28,14 @@
 			"types": "./dist/exports/*.d.ts"
 		}
 	},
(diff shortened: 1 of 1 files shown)


ckohen:
For those that may come looking for why this "fix" wasn't merged:
This only poses an issue when typescript``moduleResolution` is set to `node`. The way that typescript resolves modules in this mode is *not* accurate to how node will in versions 12+, particularly with respect to the exports field.  You should instead use `node16` or `nodenext` based on your node version.

PR #9058, opened 2023-01-13
feat(Client): `guildAuditLogEntryCreate` event

almeidx:
Implements basic support for the `GUILD_AUDIT_LOG_ENTRY_CREATE` event

Changes:
- Add `Client#guildAuditLogEntryCreate` event
- `GuildAuditLogsEntry#targetId` and `GuildAuditLogsEntry#executorId` to address https://github.com/discordjs/discord.js/pull/9058#issuecomment-1382655421
- Moved `logs` to the last constructor parameter of `GuildAuditLogsEntry` (which is private) so as to not make it nullable (we don't received it via the gateway)

Upstream:
- discord/discord-api-docs#5849

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -120,7 +120,7 @@ body:
         - No Intents
         - Guilds
         - GuildMembers
-        - GuildBans
+        - GuildModeration
         - GuildEmojisAndStickers
         - GuildIntegrations
         - GuildWebhooks
--- a/packages/discord.js/src/client/actions/ActionsManager.js
+++ b/packages/discord.js/src/client/actions/ActionsManager.js
@@ -12,6 +12,7 @@ class ActionsManager {
     this.register(require('./ChannelCreate'));
     this.register(require('./ChannelDelete'));
     this.register(require('./ChannelUpdate'));
+    this.register(require('./GuildAuditLogEntryCreate'));
     this.register(require('./GuildBanAdd'));
     this.register(require('./GuildBanRemove'));
     this.register(require('./GuildChannelsPositionUpdate'));
--- a/packages/discord.js/src/client/actions/GuildAuditLogEntryCreate.js
+++ b/packages/discord.js/src/client/actions/GuildAuditLogEntryCreate.js
@@ -1,0 +1,29 @@
+'use strict';
+
+const Action = require('./Action');
+const GuildAuditLogsEntry = require('../../structures/GuildAuditLogsEntry');
+const Events = require('../../util/Events');
+
+class GuildAuditLogEntryCreateAction extends Action {
+  handle(data) {
+    const client = this.client;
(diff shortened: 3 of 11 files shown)


SpaceEEC (approved):

kyranet (approved):

Jiralite (commented):
Should update the list of intents too via the issue tracker.

packages/discord.js/src/client/actions/AuditLogEntryCreate.js
```diff
@@ -0,0 +1,29 @@
...
+    let auditLogEntry;
+
+    if (guild) {
+      auditLogEntry = new GuildAuditLogsEntry(guild, data);
+
+      /**
+       * Emitted whenever a guild audit log entry is created.
+       * @event Client#auditLogEntryCreate
+       * @param {Guild} guild The guild where the entry was created
+       * @param {AuditLogEntry} auditLogEntry The entry that was created
```
Jiralite, line 20: I believe `AuditLogEntry` is more important and should be emitted as the first parameter. It would also be consistent with other events.

Jiralite (changes requested):
This doesn't work because the gateway event is called `GUILD_AUDIT_LOG_ENTRY_CREATE` but this is looking for `AUDIT_LOG_ENTRY_CREATE`.

Jiralite:
I seem to have found a bug with this event. Use this very basic bot:

```TypeScript
import { Client, Events, GatewayIntentBits } from "discord.js";

const client = new Client({ intents: GatewayIntentBits.Guilds | GatewayIntentBits.GuildBans });

client.on("raw", console.log);
client.on(Events.AuditLogEntryCreate, console.log);

client.login();
```

Upon pinning a message, this event will be received:

```JavaScript
{
  t: 'GUILD_AUDIT_LOG_ENTRY_CREATE',
  s: 6,
  op: 0,
  d: {
    user_id: 'user_id',
    target_id: 'target_id',
    options: {
      message_id: 'message_id',
      channel_id: 'channel_id'
    },
    id: 'id',
    action_type: 74,
    guild_id: 'guild_id'
  }
}
```

When discord.js constructs a `GuildAuditLogsEntry` out of it, `GuildAuditLogsEntry#executor` and `GuildAuditLogsEntry#target` are both `undefined`.

You can see how these are defined here:

https://github.com/discordjs/discord.js/blob/13ce78af6e3aedc793f53a099a6a615df44311f7/packages/discord.js/src/structures/GuildAuditLogsEntry.js#L127-L131

https://github.com/discordjs/discord.js/blob/13ce78af6e3aedc793f53a099a6a615df44311f7/packages/discord.js/src/structures/GuildAuditLogsEntry.js#L292-L297

When audit logs are fetched (as they have been all this time), Discord sends across the user objects that the audit log references and we store them safely away. However, the gateway **does not** do this. These are now potentially `undefined` and the same may go for application commands, auto moderation rules, scheduled events, integrations, threads, and webhooks. Currently, these are typed as always present.

vladfrangu (approved):

PR #9059, opened 2023-01-13
feat(next): export missing packages

dager-mohamed:

I discovered that a number of packages were not properly exported to the `next` package in the project, so I created new files within the exports folder and exported the missing packages. This ensures that all necessary packages are easily accessible and can be utilized by developers.

**Please move lines that apply to you out of the comment:**

### DIFF
--- a/packages/next/package.json
+++ b/packages/next/package.json
@@ -59,7 +59,14 @@
 		"@discordjs/rest": "workspace:^",
 		"@discordjs/util": "workspace:^",
 		"@discordjs/ws": "workspace:^",
-		"discord-api-types": "^0.37.27"
+		"discord-api-types": "^0.37.27",
+		"@discordjs/actions": "workspace:^",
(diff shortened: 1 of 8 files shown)


Syjalo:
Some packages are private and doesn't published on npm. Others aren't needed while working with discord.js.

dager-mohamed:
> Some packages are private and aren't published on npm. Others aren't needed when working with discord.js.

I removed the private packages,
but some packages may not currently be needed, they may become necessary in the future

kyranet (commented):
For future reference, you must use `yarn` to update the `yarn.lock` and link dependencies properly.

The code is also not formatted (missing semicolons).

Also, `/next` is intended to be for packages `discord.js` will use, and those 3 (/brokers, /proxy, /voice) are not part of those — they're optional.

dager-mohamed:
> 

got it

PR #9062, opened 2023-01-13
fix(website): remove layout shift

suneettipirneni:


apps/website/src/components/documentation/section/DocumentationSection.tsx
```diff
@@ -3,17 +3,9 @@
 import type { SectionOptions } from '@discordjs/ui';
```
iCrawl, line 3: Remove `'use client';` maybe? Since we don't use hooks no more here?

iCrawl (approved):

Issue #9034, opened 2023-01-09
Investigate alternative event emitters


suneettipirneni:
### Which package is the feature request for?

ws

### Feature

I wanted to bring up the topic of event emitters. Currently ws/ uses `@vladfrangu/async_event_emitter` for emitting events. While this has the benefit of being web-compatible, it suffers *greatly* when it comes to performance. 

This was a benchmark putting async_event_emitter against nodejs's `EventEmitter`:
```
Starting benchmark emit-multiple-listeners.js

Node-Events x 3,492,877 ops/sec ±0.87% (88 runs sampled)
@vladfrangu/async_event_emitter x 38,362 ops/sec ±17.01% (54 runs sampled)
Fastest is [ 'Node-Events' ]

Starting benchmark hundreds.js

Node-events x 402,779 ops/sec ±1.01% (86 runs sampled)
@vladfrangu/async_event_emitter x 5,222 ops/sec ±13.08% (55 runs sampled)
Fastest is [ 'Node-events' ]

Starting benchmark listeners.js

Node-events x 16,571,252 ops/sec ±1.33% (81 runs sampled)
@vladfrangu/async_event_emitter x 9,233,496 ops/sec ±1.67% (82 runs sampled)
Fastest is [ 'Node-events' ]

Starting benchmark emit.js

Node-events x 15,450,927 ops/sec ±1.07% (89 runs sampled)
@vladfrangu/async_event_emitter x 136,042 ops/sec ±3.04% (69 runs sampled)
Fastest is [ 'Node-events' ]

Starting benchmark init.js

Node-events x 32,487,062 ops/sec ±2.57% (86 runs sampled)
@vladfrangu/async_event_emitter x 1,400,947 ops/sec ±2.59% (86 runs sampled)
Fastest is [ 'Node-events' ]

Starting benchmark remove-emit.js

Node-Events x 19,093,876 ops/sec ±0.71% (89 runs sampled)
@vladfrangu/async_event_emitter x 133,367 ops/sec ±2.17% (71 runs sampled)
Fastest is [ 'Node-Events' ]

Starting benchmark context.js

Node-Events x 15,035,032 ops/sec ±0.68% (91 runs sampled)
@vladfrangu/async_event_emitter x 134,265 ops/sec ±3.66% (76 runs sampled)
Fastest is [ 'Node-Events' ]

Starting benchmark once.js

Fastest is [ 'Node-Events' ]

Starting benchmark add-remove.js

[...]

PR #9068, opened 2023-01-15
refactor(ws): use evt instead of ee

didinele:
Related: #9034 #9036

### DIFF
--- a/packages/core/src/client.ts
+++ b/packages/core/src/client.ts
@@ -1,7 +1,8 @@
 import { setTimeout } from 'node:timers';
 import type { REST } from '@discordjs/rest';
 import { calculateShardId } from '@discordjs/util';
-import { WebSocketShardEvents, type WebSocketManager } from '@discordjs/ws';
+import type { WebSocketManager } from '@discordjs/ws';
+import { Events } from '@discordjs/ws';
 import { DiscordSnowflake } from '@sapphire/snowflake';
 import { AsyncEventEmitter } from '@vladfrangu/async_event_emitter';
 import {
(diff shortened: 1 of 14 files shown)


packages/ws/src/strategies/sharding/WorkerShardingStrategy.ts
```diff
@@ -41,9 +42,21 @@ export enum WorkerReceivePayloadOp {
...
+	Hello = 'Hello',
+	Ready = 'Ready',
+	Resumed = 'Resumed',
+}
+
 export type WorkerReceivePayload =
-	// Can't seem to get a type-safe union based off of the event, so I'm sadly leaving data as any for now
-	| { data: any; event: WebSocketShardEvents; op: WorkerReceivePayloadOp.Event; shardId: number }
+	// No way to get type safety on data
+	| { data: any; event: ShardEvents; op: WorkerReceivePayloadOp.Event }
```
almeidx, line 59: I know it was like this before, but perhaps it would be better to type `data` as `unknown`
```suggestion
	| { data: unkown; event: ShardEvents; op: WorkerReceivePayloadOp.Event }
```

packages/core/src/client.ts
```diff
@@ -1,7 +1,8 @@
 import { setTimeout } from 'node:timers';
 import type { REST } from '@discordjs/rest';
 import { calculateShardId } from '@discordjs/util';
-import { WebSocketShardEvents, type WebSocketManager } from '@discordjs/ws';
+import type { WebSocketManager } from '@discordjs/ws';
+import { Events } from '@discordjs/ws';
```
almeidx, line 5:
```suggestion
import { Events, type WebSocketManager } from '@discordjs/ws';
```

packages/ws/src/Events.ts
```diff
@@ -0,0 +1,64 @@
+import type { GatewayDispatchPayload, GatewayReadyDispatchData } from 'discord-api-types/v10';
+import type { Operator, ToNonPostableEvt } from 'evt';
+import { Evt } from 'evt';
```
almeidx, line 3:
```suggestion
import { Evt, type Operator, type ToNonPostableEvt } from 'evt';
```

almeidx (changes requested):

packages/ws/src/strategies/sharding/WorkerShardingStrategy.ts
didinele: Can't. Makes me unable to `Evt#post` (or `EventEmitter#emit` in the past) w/o a ts-expect-error or some other cast on that line. Probs prefer it as-is.

didinele:
after internal deliberation, we've decided against using evt.

Issue #9083, opened 2023-01-23
Command localizations don't use camelcase when creating a command with slashcommand builder


schblondie:
### Which package is this bug report for?

builders

### Issue description

When creating a command with the slashcommand builder the localizations have the key names `description_localizations` and `name_localizations`.
This works when creating a command but is annoying when you want to compare those values with the values from using `GuildApplicationCommandManager.fetch({ withLocalizations: true })` as it returns those keys with the correct camelcase.

### Code sample returns:

```
SlashCommandBuilder {
  options: [],
  name: 'hello',
  name_localizations: { de: 'hallo' },
  description: 'Replies with Hello World!',
  description_localizations: { de: 'Antwortet mit Hallo Welt!' },
  default_permission: undefined,
  default_member_permissions: undefined,
  dm_permission: undefined
}
Collection(1) [Map] {
  '1067037449294983168' => <ref *1> ApplicationCommand {
    id: '1067037449294983168',
    applicationId: '824536235523047434',
    guild: Guild {
      id: '1000066569776402492',
      name: undefined,
      icon: undefined,
      features: undefined,
      commands: [GuildApplicationCommandManager],
      members: [GuildMemberManager],
      channels: [GuildChannelManager],
      bans: [GuildBanManager],
      roles: [RoleManager],
      presences: PresenceManager {},
      voiceStates: [VoiceStateManager],
      stageInstances: [StageInstanceManager],
      invites: [GuildInviteManager],
      scheduledEvents: [GuildScheduledEventManager],
      autoModerationRules: [AutoModerationRuleManager],
      available: false,
      shardId: 0
    },
    guildId: '1000066569776402492',
    permissions: ApplicationCommandPermissionsManager {
      manager: [Circular *1],
      guild: [Guild],
      guildId: '1000066569776402492',
      commandId: '1067037449294983168'
    },
    type: 1,
    name: 'hello',
    nameLocalizations: { de: 'hallo' },
    nameLocalized: null,
    description: 'Replies with Hello World!',
    descriptionLocalizations: { de: 'Antwortet mit Hallo Welt!' },
    descriptionLocalized: null,
    options: [],
    defaultMemberPermissions: null,
    dmPermission: null,
    version: '1067038402886770698'
  }
}
```

### Code sample

```typescript
const { Client, SlashCommandBuilder } = require('discord.js')
const { REST } = require('@discordjs/rest')
const { Routes } = require('discord-api-types/v10')
const client = new Client({ intents: 0 })
[...]

PR #9084, opened 2023-01-23
Command localizations don't use camelcase when creating a command with slashcommand builder

schblondie:
See #9083 

- This PR includes breaking changes (methods removed or renamed, parameters moved or removed) - Not sure if breaking change, because technically object keys changed

### DIFF
--- a/packages/discord.js/src/structures/ApplicationCommand.js
+++ b/packages/discord.js/src/structures/ApplicationCommand.js
@@ -70,12 +70,12 @@ class ApplicationCommand extends Base {
       this.name = data.name;
     }
 
-    if ('name_localizations' in data) {
+    if ('nameLocalizations' in data) {
       /**
        * The name localizations for this command
        * @type {?Object<Locale, string>}
        */
-      this.nameLocalizations = data.name_localizations;
+      this.nameLocalizations = data.nameLocalizations;
     } else {
       this.nameLocalizations ??= null;
     }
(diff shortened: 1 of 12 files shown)


packages/discord.js/src/structures/ApplicationCommand.js
```diff
@@ -70,12 +70,12 @@ class ApplicationCommand extends Base {
...
     }
 
-    if ('name_localizations' in data) {
+    if ('nameLocalizations' in data) {
       /**
        * The name localizations for this command
        * @type {?Object<Locale, string>}
        */
-      this.nameLocalizations = data.name_localizations;
+      this.nameLocalizations = data.nameLocalizations;
```
almeidx, line 78: The Discord API data comes with camel-cased keys

Jiralite:
Builders is not solely used with discord.js, which makes this pull request incorrect. In addition, changing the casing in discord.js is also wrong, as we are receiving that data from Discord.

packages/discord.js/src/structures/ApplicationCommand.js
schblondie: This is true, I apologize.
But when you compare line 78 with line 88 you can see that the `this.nameLocalized` is also in correct camel case even when getting the non camel case value `data.name_localized`

schblondie:
### I adjusted the code making sure it doesn't conflict with the Discord API. 
Now the SlashCommandBuilder Data is in `camelCase` which doesn't matter when accessing the API as the functions `transformCommand()` (packages\discord.js\src\managers\ApplicationCommandManager.js - line 233) and `transformOption()` (packages\discord.js\src\structures\ApplicationCommand.js - line 549) make sure that the keys are in the appropriate case.

PR #9085, opened 2023-01-23
Make SlashCommandBuilder & ContextMenuBuilder return camelCase instead of snake_case

schblondie:
### Please describe the changes this PR makes and why it should be merged:
**Changes:**
This pull request makes changes in the `SlashCommandBuilder` and `ContextMenuBuilder` to make sure they return properties in `camelCase` as the rest of the `discord.js` library does. To ensure no conflicts with the API, functions were already implemented to transform the properties before creating the commands with the API. See [ApplicationCommandManager](https://github.com/discordjs/discord.js/blob/8b70f497a1207e30edebdecd12b926c981c13d28/packages/discord.js/src/managers/ApplicationCommandManager.js#L233-L262) & [ApplicationCommand](https://github.com/discordjs/discord.js/blob/8b70f497a1207e30edebdecd12b926c981c13d28/packages/discord.js/src/structures/ApplicationCommand.js#L549-L587)
**Reason:**
This makes is possible to compare the data from the Builders with the data from the [ApplicationCommandManager.fetch()](https://discord.js.org/#/docs/discord.js/main/class/ApplicationCommandManager?scrollTo=fetch) method, which returns a [ApplicationCommand](https://discord.js.org/#/docs/discord.js/main/class/ApplicationCommand) or a collection of those with `camelCase` properties

**See also:**
For further information see #9083 

Follow up from #9084 
_I made sure to only edit properties which are not accessed by the API, as I didn't use search&replace this time_

### DIFF
--- a/packages/builders/src/interactions/contextMenuCommands/ContextMenuCommandBuilder.ts
+++ b/packages/builders/src/interactions/contextMenuCommands/ContextMenuCommandBuilder.ts
@@ -24,7 +24,7 @@ export class ContextMenuCommandBuilder {
 	/**
 	 * The localized names for this command
 	 */
-	public readonly name_localizations?: LocalizationMap;
+	public readonly nameLocalizations?: LocalizationMap;
 
 	/**
 	 * The type of this context menu command
@@ -91,7 +91,7 @@ export class ContextMenuCommandBuilder {
 		// Assert the value matches the conditions
 		validateDefaultPermission(value);
 
-		Reflect.set(this, 'default_permission', value);
+		Reflect.set(this, 'defaultPermission', value);
 
 		return this;
 	}
@@ -108,7 +108,7 @@ export class ContextMenuCommandBuilder {
 		// Assert the value and parse it
 		const permissionValue = validateDefaultMemberPermissions(permissions);
 
-		Reflect.set(this, 'default_member_permissions', permissionValue);
+		Reflect.set(this, 'defaultMemberPermissions', permissionValue);
 
 		return this;
 	}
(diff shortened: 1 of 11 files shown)


almeidx:
You can use builders as a standalone package, that's why its separate from discord.js
The Discord API does not accept camel-cased keys, so it wouldn't make sense for `@discordjs/builders` to return that

Jiralite:
Following on from the above comment:

> This makes is possible to compare the data from the Builders with the data from the [ApplicationCommandManager.fetch()](https://discord.js.org/#/docs/discord.js/main/class/ApplicationCommandManager?scrollTo=fetch) method, which returns a [ApplicationCommand](https://discord.js.org/#/docs/discord.js/main/class/ApplicationCommand) or a collection of those with `camelCase` properties

I just commented what it sounds like you wanted, and this confirms it: this functionality already exists over at [`ApplicationCommand#equals()`](https://discord.js.org/#/docs/discord.js/main/class/ApplicationCommand?scrollTo=equals).

kyranet:
This is a semver-major breaking change, one that conflicts with what's discussed in #8015, regarding this point of data exposure:

> **How should the classes store the data that will be returned from `toJSON()` calls? In a `data` field or as the class fields themselves?**
>
> [...] we should instead use the `data` field, however no getters should be added pointing to said field

Because we will move all those properties to an internal `data` field, we won't have casing issues. Furthermore, builders are intended to be write-only, not read-write. In Java for example (possibly the language that uses the builder pattern the most), you cannot read the data builders write, only once you do `.build()`, and similar will happen here with `.toJSON()`. They are, after all, a middleman whose purpose is to build data in a structured way.

And because builders are intended to be used **for** Discord, and **not for** Discord.js, they must return the data that Discord expects and can process, which is `snake_case`. If you want to compare bodies, compare the result of doing `builder.toJSON()` with incoming objects from Discord, they'll match in format and structure.

PR #9096, opened 2023-01-28
feat: remove `BitField#toArray()`

Syjalo:

We don't need `BitField#toArray()` since we can do `Array.from(<BitField>)` or `[...<BitField>]`. Also there was `<Collection>.array()` which cached values, but the method was removed and not just the caching.

Users use this method to convert bits to human readable strings with a map.
```js
const permissionsMap = {
  'SendMessages': `Send Messages`,
}
```
They shouldn't use strings for keys, but enums.

```js
const permissionsMap = {
  [PermissionFlagsBits.SendMessages]: `Send Messages`,
}
```
So the iterator should return an array of bits instead of enum key names.

### DIFF
--- a/packages/discord.js/src/util/BitField.js
+++ b/packages/discord.js/src/util/BitField.js
@@ -65,10 +65,10 @@ class BitField {
    * Gets all given bits that are missing from the bitfield.
    * @param {BitFieldResolvable} bits Bit(s) to check for
    * @param {...*} hasParams Additional parameters for the has method, if any
-   * @returns {string[]}
+   * @returns {(number|bigint)[]}
    */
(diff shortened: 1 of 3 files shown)


ImRodry:
You don't need a map to convert the strings to a human-readable format, you can simply do that with regex replacement. The toArray() method is also an easy way to read a bitfield without having to do all of those conversions. Please don't merge this.

vladfrangu:
I feel like the sentiment for this PR from the two messages speak for themselves. While we'll be closing this PR in its current state, you're free to open a PR that adds a new method that returns only the set bits (for whichever use-case that could have)

Issue #9075, opened 2023-01-17
@discordjs/ws erlpack encoding support


Deivu:
### Which package is the feature request for?

ws

### Feature

Erlpack was not added on `@discordjs/ws` and the only encoding option is `json` with no support for `etf`

### Ideal solution or implementation

Implement erlpack encoding that is similar to the old websocket code, with the option to fallback to json for those who can't install it

### Other context

Been using erlpack for a long time now and using it gave me a lot of performance benefits over the years I used it. Would be nice to add support for it as well on `@discordjs/ws`

Deivu:
> We've explicitly opted against implementing it due to concerns with the erlpack lib that Discord offers. If a better alternative were to arise, we're more than willing to implement it.

If that's the case, then should d.js just develop or fork their own erlpack and fix the concerns they have over the Discord provided erlpack lib? Since the performance increase on using erlpack is really noticeable and would help us to deploy less threads due to lower cpu usage on decoding / encoding

didinele:
Not only do I think there's no one willing to do it, there's always the concern that if someone does it no one else will have a clue how to maintain it, making it dependant on that one person.

timotejroiko:
> We've explicitly opted against implementing it due to concerns with the erlpack lib that Discord offers. If a better alternative were to arise, we're more than willing to implement it.

what would be those concerns and what would be the requirements for a better alternative? I have implemented an ETF encoder/decoder in pure js for my library, i wouldn't mind making it available as a separate package and, if need be, modifying it to comply with your requirements if you guys are interested.

SuperchupuDev:
according to discord staff (ian 2) erlpack isn't really needed a anymore

![Screenshot_20230321-084829_Discord.jpg](https://user-images.githubusercontent.com/53496941/226556821-fe54a98b-752f-494b-8e7d-9c94db7f4295.jpg)

vladfrangu:
Realistically, if you care about ws performance, installing the 3 recommended optional packages would give a noticeable performance increase! It's one of the main reason I leaned against implementing etf support when we (or well, DD made) /ws.

didinele:
As stated, we don't intend to support erlpack in the near future.

PR #9099, opened 2023-01-30
refactor(WebSocketManager): use /ws package internally

### LINKED ISSUE #8259: max_concurrency support

discord.js

### Feature

automaticly using the max_concurreny would be a awesome feature to have in djs without having to build your own custom sharding manager

a option to pick your concurreny to use if available

writing my own custom sharding manager

### Other context

_No response_

Qjuh:

Use `@discordjs/ws` for WebSocketShard handling, while maintaining the same interface v14 uses to accomplish this in a non-breaking way.

Resolves #8259 

It worked in tests with both a sharded and a single-shard bot.


packages/discord.js/src/client/websocket/WebSocketManager.js
```diff
@@ -119,11 +93,11 @@ class WebSocketManager extends EventEmitter {
...
    * Emits a debug message.
    * @param {string} message The debug message
-   * @param {?WebSocketShard} [shard] The shard that emitted this message, if any
+   * @param {?number} [shardId] The id of the shard that emitted this message, if any
    * @private
    */
-  debug(message, shard) {
-    this.client.emit(Events.Debug, `[WS => ${shard ? `Shard ${shard.id}` : 'Manager'}] ${message}`);
+  debug(message, shardId) {
+    this.client.emit(Events.Debug, `[WS => ${shardId ? `Shard ${shardId}` : 'Manager'}] ${message}`);
```
almeidx, line 100:
```suggestion
    this.client.emit(Events.Debug, `[WS => ${typeof shardId === 'number' ? `Shard ${shardId}` : 'Manager'}] ${message}`);
```

almeidx (changes requested):

packages/discord.js/src/client/websocket/WebSocketManager.js
Qjuh: This is (just) a debug message, wouldn't it be correct to include wrongly passed stuff in the log here instead of only making it log if it's a number?

didinele (changes requested):
Other than what Almeida already pointed out, LGTM.

One thing you can consider is to simplify this enum: https://github.com/discordjs/discord.js/blob/78cc7cff1fb721355385869ca17ac1f110b724e0/packages/discord.js/src/util/Status.js#L23-L33

Perhaps make it an extension of the status enum exported by `/ws`?

JMTK:
Since the new ws package doesn't support erlpack (#9075) it should probably be removed from the readme/guide as an optional package with this PR

Deivu:
Personally I think `shard.getStatus()` is redundant, as you can mark shard.status based on what the discordjs/ws emits. 
It can possibly be a bottleneck in worker sharding, where you need to rely on ipc to get the data from the ws shards.

You can remove the promise and set the status on what the discordjs/ws emits.

discordjs/ws emits ready => set the proxy ws class to waiting for guilds => forward discordjs/ws dispatch events to that proxy ws class =>set status to ready fire ready once able.

discordjs/ws emits closed => set the proxy ws to reconnecting => then go upwards for ident flow

discordjs/ws emits resumed => set the proxy ws to ready => emit resumed.

No need for promises on that regard

Qjuh:
> Personally I think `shard.getStatus()` is redundant, as you can mark shard.status based on what the discordjs/ws emits. It can possibly be a bottleneck in worker sharding, where you need to rely on ipc to get the data from the ws shards.
> 
> You can remove the promise and set the status on what the discordjs/ws emits.
> 
> discordjs/ws emits ready => set the proxy ws class to waiting for guilds => forward discordjs/ws dispatch events to that proxy ws class =>set status to ready fire ready once able.
> 
> discordjs/ws emits closed => set the proxy ws to reconnecting => then go upwards for ident flow or downwards for the resumed flow
> 
> discordjs/ws emits resumed => set the proxy ws to ready => emit resumed.
> 
> No need for promises on that regard

You are right that this could be a bottleneck when using WorkerStrategy. It won't be for SimpleStrategy because the promises immediately resolve here. Changing the strategy isn't exposed though so isn't supported at the moment anyway and we talked about this internally that we saw problems if the status of /ws and proxy ws class got out-of-sync.

But after looking further into it this looks like it should work especially since we have different enums in /ws and main lib anyway, so we need to map the status and can't just pass it through. So it can and will never be in sync.

Deivu:
> > Personally I think `shard.getStatus()` is redundant, as you can mark shard.status based on what the discordjs/ws emits. It can possibly be a bottleneck in worker sharding, where you need to rely on ipc to get the data from the ws shards.
> > You can remove the promise and set the status on what the discordjs/ws emits.
> > discordjs/ws emits ready => set the proxy ws class to waiting for guilds => forward discordjs/ws dispatch events to that proxy ws class =>set status to ready fire ready once able.
> > discordjs/ws emits closed => set the proxy ws to reconnecting => then go upwards for ident flow or downwards for the resumed flow
> > discordjs/ws emits resumed => set the proxy ws to ready => emit resumed.
> > No need for promises on that regard
> 
> You are right that this could be a bottleneck when using WorkerStrategy. It won't be for SimpleStrategy because the promises immediately resolve here. Changing the strategy isn't exposed though so isn't supported at the moment anyway and we talked about this internally that we saw problems if the status of /ws and proxy ws class got out-of-sync.
> 
> But after looking further into it this looks like it should work especially since we have different enums in /ws and main lib anyway, so we need to map the status and can't just pass it through. So it can and will never be in sync.

I aleady have an implementation that is tested and working, if you want I can pr the changes on your fork

Qjuh:
> > > Personally I think `shard.getStatus()` is redundant, as you can mark shard.status based on what the discordjs/ws emits. It can possibly be a bottleneck in worker sharding, where you need to rely on ipc to get the data from the ws shards.
> > > You can remove the promise and set the status on what the discordjs/ws emits.
> > > discordjs/ws emits ready => set the proxy ws class to waiting for guilds => forward discordjs/ws dispatch events to that proxy ws class =>set status to ready fire ready once able.
> > > discordjs/ws emits closed => set the proxy ws to reconnecting => then go upwards for ident flow or downwards for the resumed flow
> > > discordjs/ws emits resumed => set the proxy ws to ready => emit resumed.
> > > No need for promises on that regard
> > 
> > 
> > You are right that this could be a bottleneck when using WorkerStrategy. It won't be for SimpleStrategy because the promises immediately resolve here. Changing the strategy isn't exposed though so isn't supported at the moment anyway and we talked about this internally that we saw problems if the status of /ws and proxy ws class got out-of-sync.
> > But after looking further into it this looks like it should work especially since we have different enums in /ws and main lib anyway, so we need to map the status and can't just pass it through. So it can and will never be in sync.
> 
> I aleady have an implementation that is tested and working, if you want I can pr the changes on your fork
> 
> Also I think you should not hardcode what worker strategy you should use, and let the users use the original configuration options available on the ws package as well

I‘ll happily look at your implementation, open a PR and I‘ll take a look. The important part is to do it in a non-breaking way.
I‘ll try to allow setting strategy too but that would need more extensive testing to make sure it‘s not breaking.

Deivu:
> > > > Personally I think `shard.getStatus()` is redundant, as you can mark shard.status based on what the discordjs/ws emits. It can possibly be a bottleneck in worker sharding, where you need to rely on ipc to get the data from the ws shards.
> > > > You can remove the promise and set the status on what the discordjs/ws emits.
> > > > discordjs/ws emits ready => set the proxy ws class to waiting for guilds => forward discordjs/ws dispatch events to that proxy ws class =>set status to ready fire ready once able.
> > > > discordjs/ws emits closed => set the proxy ws to reconnecting => then go upwards for ident flow or downwards for the resumed flow
> > > > discordjs/ws emits resumed => set the proxy ws to ready => emit resumed.
> > > > No need for promises on that regard
> > > 
> > > 
> > > You are right that this could be a bottleneck when using WorkerStrategy. It won't be for SimpleStrategy because the promises immediately resolve here. Changing the strategy isn't exposed though so isn't supported at the moment anyway and we talked about this internally that we saw problems if the status of /ws and proxy ws class got out-of-sync.
> > > But after looking further into it this looks like it should work especially since we have different enums in /ws and main lib anyway, so we need to map the status and can't just pass it through. So it can and will never be in sync.
> > 
> > 
> > I aleady have an implementation that is tested and working, if you want I can pr the changes on your fork
> > Also I think you should not hardcode what worker strategy you should use, and let the users use the original configuration options available on the ws package as well
> 
> I‘ll happily look at your implementation, open a PR and I‘ll take a look. The important part is to do it in a non-breaking way. I‘ll try to allow setting strategy too but that would need more extensive testing to make sure it‘s not breaking.

strategy should be non breaking. specially since the events emitted are the same

DraftProducts:
This PR should be tagged blocked and removed from v14.8 milestone until erlpack is implemented in @discordjs/ws otherwise this migration would be a downgrade.

legendhimself:
I presonally think its a bad idea to include these changes in v14 as the current WS implementation is the most stable it could have been with all the bugs and issues squashed. There are no more issues (that I am aware of) on the current WS implementation.

This pr tries to include the new `/ws` package in a non breaking way but replacing the old code with completely new code that is not even tested in production for large scale can cause several new issues.
The new `/ws` was tested for many small bots and there have been many issues/reports for the new `/ws` package which is totally fine but its a bad idea to have the new ws replace the old ws in **non-major release**.
I suggest we do this in v15.

didinele:
> This PR should be tagged blocked and removed from v14.8 milestone until issues are fixed and `erlpack` is implemented in `@discordjs/ws` otherwise this migration would be a downgrade.
> 
> **Edit:** The package has issues like #9103 with such a small users/devs base, so I can't imagine how many we could have in the main discord.js module if it would be migrated.

Changing internal implementation details is not major regardless of if it's a "downgrade". This can be merged just fine and be prepared for 14.8, we can iron out stability/bugs while its a `@dev` release.

didinele:
> I presonally think its a bad idea to include these changes in v14 as the current WS implementation is the most stable it could have been with all the bugs and issues squashed. There are no more issues (that I am aware of) on the current WS implementation.
> 
> This pr tries to include the new `/ws` package in a non breaking way but replacing the old code with completely new code that is not even tested in production for large scale can cause several new issues. The new `/ws` was tested for many small bots and there have been many issues/reports for the new `/ws` package which is totally fine but its a bad idea to have the new ws replace the old ws in **non-major release**. I suggest we do this in v15.

And you reckon rolling it out untested in v15 is any better? It doesn't matter what version its in so long as it had enough time to be toyed with as a dev release.

legendhimself:
> And you reckon rolling it out untested in v15 is any better

We have time before v15 and during this time the new `/ws` code can be tested more therefore we have more reported issues.
The `/ws` code is the most important part of the bot and I think its better to release in major release,
and if you want it to be included in v14, make it opt-in (with the client option maybe?)

DraftProducts:
> Changing internal implementation details is not major regardless of if it's a "downgrade". This can be merged just fine and be prepared for 14.8, we can iron out stability/bugs while its a `@dev` release.

I agree! 
But then why should this module be released in the main module when it :
Contains unresolved bugs issues
Does not have a 1.0.0 version yet.
Has been redone from 0 and doesn't take advantage of this opportunity to integrate all the features documented in the Discord api ([concurrency sharding](https://discord.com/developers/docs/topics/gateway#sharding-max-concurrency) (unless it concerns the new sharder module started by @kyranet ?))

vladfrangu:
Since we're all giving our two cents, here's my three cents:

- You are correct in the fact that `/ws` is _less_ tested than the current zombie that is the djs v14 ws system. But it's also a piece that we (or at least I) want to move away from and focus less time maintaining, and as we've all probably seen, a LOT of mini PRs fixing some edge case end up with code that is less than ideal
- Version number doesn't guarantee stability. Look at `-types`, it is still at `0.x.x` and its considered stable
- @DraftProducts, I don't think you checked the `/ws` implementation! `max_concurrency` has been implemented from the get-go in the module, since it was remade from scratch and ensures it hits everything from Discord's gateway docs. You can see it handled here: https://github.com/discordjs/discord.js/blob/main/packages/ws/src/utils/IdentifyThrottler.ts

There's also advantages by moving to `/ws`. Not only a more powerful and robust ws system (hammering pending), we can also release fixes *without* needing to release a full djs version too. 

And it's not like we're merging this tonight and releasing yesterday and the world's on fire! This still needs to be reviewed, tested, knock on wood it just works, then release. Instead of holding on to the old system, why not help us test the new one? 🙏 (just by using it raw works). More testing = more confirmations of it being fine / hopefully few to none reports of issues = more stable ws experience for everyone = a W its called.

DraftProducts:
Indeed, I didn't see that it had been implemented.
I agree with the rest that has been said.

didinele (approved):

DraftProducts (approved):

packages/discord.js/src/client/websocket/WebSocketManager.js
```diff
@@ -22,15 +26,13 @@ const BeforeReadyWhitelist = [
...
-const unrecoverableErrorCodeMap = {
-  [GatewayCloseCodes.AuthenticationFailed]: ErrorCodes.TokenInvalid,
-  [GatewayCloseCodes.InvalidShard]: ErrorCodes.ShardingInvalid,
-  [GatewayCloseCodes.ShardingRequired]: ErrorCodes.ShardingRequired,
-  [GatewayCloseCodes.InvalidIntents]: ErrorCodes.InvalidIntents,
-  [GatewayCloseCodes.DisallowedIntents]: ErrorCodes.DisallowedIntents,
-};
+const WaitingForGuildEvents = [GatewayDispatchEvents.GuildCreate, GatewayDispatchEvents.GuildDelete];
 
-const UNRESUMABLE_CLOSE_CODES = [1000, GatewayCloseCodes.AlreadyAuthenticated, GatewayCloseCodes.InvalidSeq];
```
didinele, line 33: Prefer using https://github.com/discordjs/discord.js/blob/dc142c47e4021d6d37712453cc2fad6e4fe4e9f7/packages/ws/src/ws/WebSocketShard.ts#L73-L76

packages/discord.js/src/client/websocket/WebSocketManager.js
```diff
@@ -132,11 +117,34 @@ class WebSocketManager extends EventEmitter {
    */
   async connect() {
     const invalidToken = new DiscordjsError(ErrorCodes.TokenInvalid);
+    const { shards, shardCount, intents, ws } = this.client.options;
+    if (this._ws && this._ws.options.token !== this.client.token) {
+      await this._ws.destroy({ code: 1000, reason: 'Login with differing token requested' });
```
didinele, line 122: here too

packages/discord.js/src/client/websocket/WebSocketManager.js
```diff
@@ -322,8 +283,7 @@ class WebSocketManager extends EventEmitter {
     // TODO: Make a util for getting a stack
     this.debug(`Manager was destroyed. Called by:\n${new Error().stack}`);
     this.destroyed = true;
-    this.shardQueue.clear();
-    for (const shard of this.shards.values()) shard.destroy({ closeCode: 1_000, reset: true, emit: false, log: false });
+    this._ws.destroy({ code: 1_000 });
```
didinele, line 286: here as well

didinele (changes requested):
this small nit + waiting on strategy support, otherwise LGTM

almeidx (changes requested):
Need to export all of the /ws types in the typings as well

packages/discord.js/typings/index.d.ts
```diff
@@ -3309,8 +3307,6 @@ export class WebSocketManager extends EventEmitter {
 
   private debug(message: string, shard?: WebSocketShard): void;
```
almeidx, line 3308:
```suggestion
  private debug(message: string, shardId?: number): void;
```

packages/discord.js/src/client/websocket/WebSocketManager.js
```diff
@@ -152,156 +162,108 @@ class WebSocketManager extends EventEmitter {
...
-         * @event Client#invalidated
+         * Emitted when a shard's WebSocket disconnects and will no longer reconnect.
+         * @event Client#shardDisconnect
+         * @param {CloseEvent} event The WebSocket close event
+         * @param {number} id The shard id that disconnected
          */
-        this.client.emit(Events.Invalidated);
-        // Destroy just the shards. This means you have to handle the cleanup yourself
-        this.destroy();
+        this.client.emit(Events.ShardDisconnect, { code, reason, wasClean: true }, shardId);
```
almeidx, line 218: According to the types, `CloseEvent` (the first event parameter) should also have a `target`

https://github.com/discordjs/discord.js/blob/79f664e1d93c38d4c80e82cfa0fa774e2c41f045/packages/discord.js/typings/index.d.ts#L4859-L4864

However, according to the JSDoc, which is just a link to MDN, it does not have a `target`
https://developer.mozilla.org/en-US/docs/Web/API/CloseEvent#instance_properties

One of those should probably be changed

packages/discord.js/src/client/websocket/WebSocketManager.js
```diff
@@ -132,11 +117,36 @@ class WebSocketManager extends EventEmitter {
...
+        version: ws.version,
+        shardIds: shards === 'auto' ? null : shards,
+        shardCount: shards === 'auto' ? null : shardCount,
+        initialPresence: ws.presence,
+        retrieveSessionInfo: shardId => this.shards.get(shardId).sessionInfo,
+        updateSessionInfo: (shardId, sessionInfo) => {
+          this.shards.get(shardId).sessionInfo = sessionInfo;
+        },
+      };
+      if (ws.buildStrategy) wsOptions.buildStrategy = ws.buildStrategy;
```
almeidx, line 140: Can't this be set directly on the object?

almeidx (changes requested):
Seems like, out of all of the `WebSocketShardEvents`, only these are being emitted:
- `WebSocketShardEvents.Ready`
- `WebSocketShardEvents.AllReady`
- `WebSocketShardEvents.Resumed`

https://github.com/discordjs/discord.js/blob/79f664e1d93c38d4c80e82cfa0fa774e2c41f045/packages/discord.js/src/util/WebSocketShardEvents.js#L18-L25

packages/discord.js/src/client/websocket/WebSocketManager.js
Qjuh: No, because else the default in `@discordjs/ws` won't be merged if it's not set, because the key would exist in the passed options. Unless there's a way to do that that I don't see...

Qjuh:
> Seems like, out of all of the `WebSocketShardEvents`, only these are being emitted:
> 
>     * `WebSocketShardEvents.Ready`
> 
>     * `WebSocketShardEvents.AllReady`
> 
>     * `WebSocketShardEvents.Resumed`

`WebSocketShardEvents.Close` can be added back rather easily. The other two events can't be emitted in a useful way because of the different approach of `@discordjs/ws` when it comes to handling InvalidSession replies by the API and destroying WebSocketShards.
They were meant to be only used internally by the WebSocketManager before, so I'll mark them as deprecated.

didinele (approved):

DraftProducts:
Please tell me if there is a better place to discuss the subject.
I noticed that it wasn't planned to integrate erlpack to this new ws package (#9075) (which is a breaking change).
Do we have a bench about performance's impact that it would have on a significative number of requests?
I read that the package had problems, but couldn't find them 🤔 
(I am currently using it in production)

didinele:
> (which is a breaking change).

it's not.

as per the erlpack stuff, after speaking internally a bit, we'll probably be writing our own implementation in Rust, but this PR will be merged even without it.

Kaisarion:
is there an update here? is it functional/stable enough for me to pull and attempt?

didinele:
yes, we're looking into merging it soon, just making sure we can delay the next minor for long enough for this to get appropriately tested while in a dev release

packages/discord.js/src/util/Options.js
```diff
@@ -65,6 +65,8 @@ const { version } = require('../../package.json');
  * sent in the initial guild member list, must be between 50 and 250
  * @property {number} [version=10] The Discord gateway version to use <warn>Changing this can break the library;
  * only set this if you know what you are doing</warn>
+ * @property {Function} [buildStrategy] Builds the strategy to use for sharding;
+ * example: `(manager) => new WorkerShardingStrategy(manager, { shardsPerWorker: 2 })`
```
Jiralite, line 69: Maybe use an `@example` tag?

TetieWasTaken (changes requested):
See #9313 and #9286

packages/discord.js/src/client/websocket/WebSocketManager.js
```diff
@@ -99,11 +98,16 @@ class WebSocketManager extends EventEmitter {
     this.destroyed = false;
 
     /**
-     * If this manager is currently reconnecting one or multiple shards
-     * @type {boolean}
+     * @external WSWebSocketManager
+     * @see {@link https://discord.js.org/#/docs/ws/main/class/WebSocketManager}
```
Jiralite, line 102:
```suggestion
     * @see {@link https://discord.js.org/docs/packages/ws/stable/WebSocketManager:Class}
```

TetieWasTaken (approved):

packages/discord.js/src/client/websocket/WebSocketManager.js
```diff
@@ -99,11 +98,16 @@ class WebSocketManager extends EventEmitter {
     this.destroyed = false;
 
     /**
-     * If this manager is currently reconnecting one or multiple shards
-     * @type {boolean}
+     * @external WSWebSocketManager
+     * @see {@link https://discord.js.org/docs/packages/ws/stable/WebSocketManager:Class}
```
kyranet, line 102: This is already included in `packages/discord.js/src/util/Options.js`

kyranet (changes requested):

kyranet (approved):

PR #9119, opened 2023-02-09
fix(WebSocketShard): proper error bubbling

### LINKED ISSUE #9103: @discordjs/ws Unexpected network conditions cause crash when reconnecting

ws

### Issue description

When a close event is emitted, the library attempts to reconnect via onClose -> destroy -> connect.

At this point, if there is a network issue or unexpected network condition, the function `waitForEvent` may throw an error, for example if `ws` emits an error event, which will bubble up back to `onClose` and cause a crash.

An example of a possible error:

```
Error: Unexpected server response: 520
  at ClientRequest.<anonymous> (/app/node_modules/ws/lib/websocket.js:888:7)
  at ClientRequest.emit (node:events:513:28)
  at HTTPParser.parserOnIncomingClient [as onIncoming] (node:_http_client:695:27)
  at HTTPParser.parserOnHeadersComplete (node:_http_common:117:17)
  at TLSSocket.socketOnData (node:_http_client:536:22)
  at TLSSocket.emit (node:events:513:28)
  at addChunk (node:internal/streams/readable:324:12)
  at readableAddChunk (node:internal/streams/readable:297:9)
  at Readable.push (node:internal/streams/readable:234:10)
  at TLSWrap.onStreamRead (node:internal/stream_base_commons:190:23)

node:internal/process/promises:288
  triggerUncaughtException(err, true /* fromPromise */);
```

Adding a `.catch` to either the `destroy` method when called from `onClose`, or the `connect` method when called from `destroy`, fixes the issue

### Code sample

_No response_

### Package version

@discordjs/ws@0.6.1-dev.1674129868-56ef3ad.0

### Node.js version

19.3.0

### Operating system

windows 10

Medium (should be fixed soon)

56ef3ad

didinele:
Closes #9103

### DIFF
--- a/packages/ws/src/ws/WebSocketShard.ts
+++ b/packages/ws/src/ws/WebSocketShard.ts
@@ -163,7 +163,8 @@ export class WebSocketShard extends AsyncEventEmitter<WebSocketShardEventsMap> {
 
 		this.sendRateLimitState = getInitialSendRateLimitState();
 
-		if ((await this.waitForEvent(WebSocketShardEvents.Hello, this.strategy.options.helloTimeout)) !== null) {
+		const waitPromise = this.waitForEvent(WebSocketShardEvents.Ready, this.strategy.options.readyTimeout);
+		if ((await this.bubbleWaitForEventError(waitPromise)) !== null) {
 			return;
 		}
 
@@ -243,8 +244,7 @@ export class WebSocketShard extends AsyncEventEmitter<WebSocketShardEventsMap> {
 		}
 	}
 
-	// returns null on success, or an error (unknown type) on failure
-	private async waitForEvent(event: WebSocketShardEvents, timeoutDuration?: number | null): Promise<unknown> {
+	private async waitForEvent(event: WebSocketShardEvents, timeoutDuration?: number | null): Promise<void> {
 		this.debug([`Waiting for event ${event} ${timeoutDuration ? `for ${timeoutDuration}ms` : 'indefinitely'}`]);
 		const controller = new AbortController();
 		const timeout = timeoutDuration ? setTimeout(() => controller.abort(), timeoutDuration).unref() : null;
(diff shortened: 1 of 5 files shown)


didinele:
@timotejroiko @Deivu Mind giving my branch a shot? Can give you a hand if you aren't familiar with testing d.js releases that aren't on NPM yet locally. I don't want you necessarily running this in prod or anything crazy like that, just mess around with it.

packages/ws/src/ws/WebSocketShard.ts
```diff
@@ -393,7 +452,9 @@ export class WebSocketShard extends AsyncEventEmitter<WebSocketShardEventsMap> {
 			this.inflate.push(Buffer.from(decompressable), flush ? zlib.Z_SYNC_FLUSH : zlib.Z_NO_FLUSH);
 
 			if (this.inflate.err) {
-				this.emit('error', `${this.inflate.err}${this.inflate.msg ? `: ${this.inflate.msg}` : ''}`);
+				this.emit(WebSocketShardEvents.Error, {
+					error: `${this.inflate.err}${this.inflate.msg ? `: ${this.inflate.msg}` : ''}`,
```
SpaceEEC, line 456: Maybe this it a bit outside of scope, maybe it's not, but anyway:
Shouldn't this be an `Error`? Throwing a string (see line 288) does not seem right to me.

SpaceEEC (approved):
This looks fine otherwise

packages/ws/src/ws/WebSocketShard.ts
didinele: You might be right, could construct an `Error` object here since its the only place that makes us use `unknown`/`any` for the event.

packages/ws/src/ws/WebSocketShard.ts
```diff
@@ -234,18 +244,58 @@ export class WebSocketShard extends AsyncEventEmitter<WebSocketShardEventsMap> {
...
+			}
+
+			// If the error is handled, we can just try to reconnect
+			await this.destroy({
+				code: CloseCodes.Normal,
+				reason: 'Something timed out',
+				recover: WebSocketShardDestroyRecovery.Reconnect,
+			});
+
+			return null;
```
timotejroiko, line 298: shouldn't this return something other than `null` to prevent the original connect method from continuing after the destroy method re-calls it, potentially identifying/resuming twice?

didinele: Yes. I think ultimately I'll opt for a different signaling system - probably a result type of sorts with a bool prop on it.

vladfrangu (approved):

kyranet (approved):

PR #9144, opened 2023-02-18
fix(snowflake): snowflakes length

DraftProducts:

Snowflakes length depends on the UInt64 max value length and the Discord epoch.
- `2^64` max length is 20.
- A snowflake created a week after Discord epoch is 16 length minimum.

This PR change all snowflakes regex to `\d{16,20}`.

### DIFF
--- a/packages/discord.js/src/managers/ThreadManager.js
+++ b/packages/discord.js/src/managers/ThreadManager.js
@@ -148,7 +148,7 @@ class ThreadManager extends CachedManager {
     let id;
     const query = makeURLSearchParams({ limit });
     if (typeof before !== 'undefined') {
-      if (before instanceof ThreadChannel || /^\d{16,20}$/.test(String(before))) {
+      if (before instanceof ThreadChannel || /^\d{17,19}$/.test(String(before))) {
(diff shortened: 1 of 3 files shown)


kyranet:
This isn't exactly correct, see:

- The earliest known ID in Discord is Stanislav's user ID, which is 17 digits long (created on 28th February).
- And new IDs will be 19 digits long until late century (`2090-07-20T17:49:51.015Z`). Any ID that's time-travelled over 67 years to the past (please let me know if Discord's still a thing!) will never resolve to a valid entity in Discord, for over many years.

Any ID outside that range is just invalid.

DraftProducts:
Yes ok 😅 
Should we use this PR to transform current 16-digit ID's to 17 or isn't it needed ?

- `ThreadManager.js` L151
- `RequestManager.ts` (3 times)

Jiralite:
I would say convert them, as the 16-digit ids are likely invalid.

vladfrangu (approved):

kyranet (approved):

SpaceEEC (approved):

PR #9134, opened 2023-02-15
feat(StageChannel): add messages

jaw0r3k:
**Please describe the changes this PR makes and why it should be merged:**

Properties were implemented directly into stage channel, ~~to keep it as simple it can be~~
Second idea was making common class for voice and stage channels 

**Status and versioning classification:**

Waiting for https://github.com/discordjs/discord-api-types/pull/721 to be merged

- Code changes have been tested against the Discord API
- I know how to update typings and have done so
- This PR changes the library's interface (methods or parameters added)

PR #9145, opened 2023-02-18
feat(StageChannel): v13 support messages

jaw0r3k:
Backports:
- #9134

And that was sorted alphabetically so that's why changes looks terrible

### DIFF
--- a/src/structures/BaseGuildVoiceChannel.js
+++ b/src/structures/BaseGuildVoiceChannel.js
@@ -163,11 +163,11 @@ class BaseGuildVoiceChannel extends GuildChannel {
    * Sets the bitrate of the channel.
    * @param {number} bitrate The new bitrate
    * @param {string} [reason] Reason for changing the channel's bitrate
-   * @returns {Promise<VoiceChannel>}
+   * @returns {Promise<BaseGuildVoiceChannel>}
    * @example
    * // Set the bitrate of a voice channel
-   * voiceChannel.setBitrate(48_000)
-   *   .then(vc => console.log(`Set bitrate to ${vc.bitrate}bps for ${vc.name}`))
+   * channel.setBitrate(48_000)
+   *   .then(channel => console.log(`Set bitrate to ${channel.bitrate}bps for ${channel.name}`))
    *   .catch(console.error);
    */
   setBitrate(bitrate, reason) {
@@ -194,11 +194,11 @@ class BaseGuildVoiceChannel extends GuildChannel {
    * Sets the user limit of the channel.
    * @param {number} userLimit The new user limit
    * @param {string} [reason] Reason for changing the user limit
-   * @returns {Promise<VoiceChannel>}
+   * @returns {Promise<BaseGuildVoiceChannel>}
    * @example
    * // Set the user limit of a voice channel
-   * voiceChannel.setUserLimit(42)
-   *   .then(vc => console.log(`Set user limit to ${vc.userLimit} for ${vc.name}`))
+   * channel.setUserLimit(42)
+   *   .then(channel => console.log(`Set user limit to ${channel.userLimit} for ${channel.name}`))
    *   .catch(console.error);
    */
   setUserLimit(userLimit, reason) {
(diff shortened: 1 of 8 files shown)


src/structures/BaseGuildVoiceChannel.js
```diff
@@ -20,20 +47,46 @@ class BaseGuildVoiceChannel extends GuildChannel {
...
+      /**
+       * The rate limit per user (slowmode) for this channel in seconds
        * @type {number}
        */
-      this.userLimit = data.user_limit;
+      this.rateLimitPerUser = data.rate_limit_per_user;
+    }
+
+    if ('nsfw' in data) {
+      this.nsfw = Boolean(data.nsfw);
```
NotSugden, line 89:
```suggestion
      this.nsfw = data.nsfw;
```
is it required to use the Boolean conversion here? the API would send this as a boolean if it was included

NotSugden (approved):

src/structures/BaseGuildVoiceChannel.js
jaw0r3k: 1. Old code was the same
2. It can be null, so we just narrow it to false

NotSugden: according to the [documentation](https://discord.com/developers/docs/resources/channel#channel-object-channel-structure) if `nsfw` is present then it will be a `boolean` and not null. You are right that when the channel is initialized it may not be a boolean. It may be worth a new PR for fixing this if this is indeed the case!

jaw0r3k: ups yeah

vladfrangu (approved):

src/structures/BaseGuildVoiceChannel.js
```diff
@@ -80,6 +135,45 @@ class BaseGuildVoiceChannel extends GuildChannel {
...
+   *   .then(invite => console.log(`Created an invite with a code of ${invite.code}`))
+   *   .catch(console.error);
+   */
+  createInvite(options) {
+    return this.guild.invites.create(this.id, options);
+  }
+
+  /**
+   * Fetches a collection of invites to this guild channel.
+   * Resolves with a collection mapping invites by their codes.
```
kyranet, line 154: This is not included in the v14 PR.

src/structures/BaseGuildVoiceChannel.js
```diff
@@ -80,6 +135,45 @@ class BaseGuildVoiceChannel extends GuildChannel {
...
+
+  /**
+   * Sets the bitrate of the channel.
+   * @param {number} bitrate The new bitrate
+   * @param {string} [reason] Reason for changing the channel's bitrate
+   * @returns {Promise<VoiceChannel>}
+   * @example
+   * // Set the bitrate of a voice channel
+   * voiceChannel.setBitrate(48_000)
+   *   .then(vc => console.log(`Set bitrate to ${vc.bitrate}bps for ${vc.name}`))
```
kyranet, line 170: Do not change upstream's names.
```suggestion
   *   .then(channel => console.log(`Set bitrate to ${channel.bitrate}bps for ${channel.name}`))
```

src/structures/VoiceChannel.js
```diff
@@ -117,75 +59,52 @@ class VoiceChannel extends BaseGuildVoiceChannel {
...
+   * @param {string} [reason] The reason for modifying this region.
+   * @returns {Promise<VoiceChannel>}
+   * @example
+   * // Set the RTC region to sydney
+   * voiceChannel.setRTCRegion('sydney');
+   * @example
+   * // Remove a fixed region for this channel - let Discord decide automatically
+   * voiceChannel.setRTCRegion(null, 'We want to let Discord decide.');
+   */
   /**
```
kyranet, line 88:
```suggestion
   */

  /**
```

kyranet (changes requested):

src/structures/BaseGuildVoiceChannel.js
jaw0r3k: It's not, but some formatting should be done...

kyranet (approved):

iCrawl (approved):

PR #9160, opened 2023-02-22
feat: Audit logs manager

jaw0r3k:
1. Caches entries e.g. to check the admin who is banning a user *without unnecessary fetching*


packages/discord.js/src/managers/GuildAuditLogManager.js
```diff
@@ -0,0 +1,86 @@
...
+      }
+      const clone = existing._clone();
+      clone._patch(data);
+      return clone;
+    }
+
+    const entry = this.holds ? new this.holds(this.guild, data, ...extras) : data;
+    if (cache) this.cache.set(id ?? entry.id, entry);
+    return entry;
+  }
```
Syjalo, line 45: Why did you add this? `CachedManager` already has this method.

packages/discord.js/src/managers/GuildAuditLogManager.js
```diff
@@ -0,0 +1,86 @@
...
+const CachedManager = require('./CachedManager');
+const { DiscordjsTypeError, ErrorCodes } = require('../errors');
+const GuildAuditLogs = require('../structures/GuildAuditLogs');
+const GuildAuditLogsEntry = require('../structures/GuildAuditLogsEntry');
+
+/**
+ * An extension for guild-specific application commands.
+ * @extends {CachedManager}
+ */
+class GuildAuditLogManager extends CachedManager {
```
Syjalo, line 14: Shouldn't the name be `GuildAuditLogEntryManager` since it holds `GuildAuditLogsEntry`?
Btw `GuildAuditLogsEntry` should be renamed to `GuildAuditLogEntry` in v15.

packages/discord.js/src/structures/Guild.js
```diff
@@ -39,6 +38,12 @@ class Guild extends AnonymousGuild {
   constructor(client, data) {
     super(client, data, false);
 
+    /**
+     * A manager of the guild audit logs of this guild
+     * @type {GuildApplicationCommandManager}
+     */
+    this.auditLogs = new GuildAuditLogManager(this);
```
Syjalo, line 45:
```suggestion
    this.auditLogEntries = new GuildAuditLogEntryManager(this);
```
Ditto

packages/discord.js/src/managers/GuildAuditLogManager.js
```diff
@@ -0,0 +1,86 @@
...
+    if (user) {
+      const userId = this.client.users.resolveId(user);
+      if (!userId) throw new DiscordjsTypeError(ErrorCodes.InvalidType, 'user', 'UserResolvable');
+      query.set('user_id', userId);
+    }
+
+    const data = await this.client.rest.get(Routes.guildAuditLog(this.id), { query });
+    return new GuildAuditLogs(this, data);
+  }
+}
```
Syjalo, line 84: I don't know if this should be here since this returns an `AuditLogs`, but the manager holds `AuditLogsEntry`.

Syjalo (changes requested):

packages/discord.js/src/structures/Guild.js
jaw0r3k: not so sure about naming

packages/discord.js/src/managers/GuildAuditLogManager.js
jaw0r3k: Like it takes some objects like commands 
which Entity uses
possibly they can be placed in a separate param
idk

jaw0r3k: maybe we can even delete GuildAuditLogs as it only has entires property
+ collections of objects

jaw0r3k: e90c94958761cbf7ba74139b912c5052b4951e48

packages/discord.js/src/managers/GuildAuditLogEntryManager.js
```diff
@@ -0,0 +1,120 @@
...
+
+    if (user) {
+      const userId = this.client.users.resolveId(user);
+      if (!userId) throw new DiscordjsTypeError(ErrorCodes.InvalidType, 'user', 'UserResolvable');
+      query.set('user_id', userId);
+    }
+
+    const data = await this.client.rest.get(Routes.guildAuditLog(this.id), { query });
+
+    let notCachableObjects = {
```
jaw0r3k, line 86: i am not too sure about it

packages/discord.js/src/managers/GuildAuditLogEntryManager.js
```diff
@@ -0,0 +1,120 @@
...
+
+    if (user) {
+      const userId = this.client.users.resolveId(user);
+      if (!userId) throw new DiscordjsTypeError(ErrorCodes.InvalidType, 'user', 'UserResolvable');
+      query.set('user_id', userId);
+    }
+
+    const data = await this.client.rest.get(Routes.guildAuditLog(this.id), { query });
+
+    const notCachableObjects = {
```
jaw0r3k, line 86: i am not too sure about it

NotSugden: nonCacheableObjects?
structures?
uncacheable?

kyranet (commented):
I'll be honest here, but this is a no for me.

With TSR, I would personally like to move away from the excessive caching, as users can easily cache data on their way (plus, patching data with TSR would be even easier).

Besides, just like the message cache, the audit log one builds up, and can do so quite quickly, which increases memory usage. Users are aware of the settings to personalize the message cache, but they wouldn't be aware of this new manager, it can come up as unexpected and cause unexpected outages.

tl;dr: I don't see the benefit of this, you can cache this data easily by yourself.

jaw0r3k:
> I'll be honest here, but this is a no for me.
> 
> With TSR, I would personally like to move away from the excessive caching, as users can easily cache data on their way (plus, patching data with TSR would be even easier).
> 
> Besides, just like the message cache, the audit log one builds up, and can do so quite quickly, which increases memory usage. Users are aware of the settings to personalize the message cache, but they wouldn't be aware of this new manager, it can come up as unexpected and cause unexpected outages.
> 
> tl;dr: I don't see the benefit of this, you can cache this data easily by yourself.

I agree

PR #9165, opened 2023-02-23
feat(Collector): add iterators and getters

Syjalo:
This PR adds new iterators and getters that allow you use collectors without deep nesting.

### DIFF
--- a/packages/discord.js/src/structures/interfaces/Collector.js
+++ b/packages/discord.js/src/structures/interfaces/Collector.js
@@ -157,7 +157,7 @@ class Collector extends EventEmitter {
   }
 
   /**
-   * Returns a promise that resolves with the next collected element;
+   * Returns a promise that resolves with the next collected, disposed or ignored elements;
    * rejects with collected elements if the collector finishes without receiving a next element
    * @type {Promise}
    * @readonly
@@ -171,12 +171,24 @@ class Collector extends EventEmitter {
 
       const cleanup = () => {
         this.removeListener('collect', onCollect);
+        this.removeListener('dispose', onDispose);
+        this.removeListener('ignore', onIgnore);
         this.removeListener('end', onEnd);
       };
 
-      const onCollect = item => {
+      const onCollect = (...items) => {
         cleanup();
-        resolve(item);
+        resolve(['collecting', ...items]);
+      };
+
+      const onDispose = (...items) => {
+        cleanup();
+        resolve(['disposing', ...items]);
+      };
+
+      const onIgnore = (...items) => {
+        cleanup();
+        resolve(['ignoring', ...items]);
       };
 
       const onEnd = () => {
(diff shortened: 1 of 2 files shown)


packages/discord.js/src/structures/interfaces/Collector.js
```diff
@@ -251,13 +364,17 @@ class Collector extends EventEmitter {
   }
 
   /**
-   * Allows collectors to be consumed with for-await-of loops
+   * Allows collector to be consumed with for-await-of loop for collected, disposed and ignored elements
```
Jiralite, line 367:
```suggestion
   * Allows collectors to be consumed with for-await-of loop for collected, disposed and ignored elements
```

packages/discord.js/src/structures/interfaces/Collector.js
```diff
@@ -157,7 +157,7 @@ class Collector extends EventEmitter {
   }
 
   /**
-   * Returns a promise that resolves with the next collected element;
+   * Returns a promise that resolves with the next collected, disposed or ignored elements;
```
Jiralite, line 160:
```suggestion
   * Returns a promise that resolves with the next collected, disposed, or ignored elements;
```

SpaceEEC (commented):
The `next_` and the new async iterators look quite similar, these could probably be abstracted, if it doesn't hurt readability too much :eyes:

almeidx:
This needs a rebase

packages/discord.js/src/structures/interfaces/Collector.js
```diff
@@ -171,12 +171,59 @@ class Collector extends EventEmitter {
...
+    });
+  }
+
+  /**
+   * Returns a promise that resolves with the next collected elements;
+   * rejects with collected elements if the collector finishes without receiving a next element
+   * @type {Promise}
+   * @readonly
+   */
+  get nextCollecting() {
```
kyranet, line 212: I feel like we shouldn't have async getters. What about `waitForCollect()`? Same for the other methods.

I also think that it'd be useful to accept an `AbortSignal` so we can cancel and unregister the listeners earlier.

packages/discord.js/src/structures/interfaces/Collector.js
```diff
@@ -268,16 +385,114 @@ class Collector extends EventEmitter {
           await new Promise(resolve => {
             const tick = () => {
               this.removeListener('collect', tick);
+              this.removeListener('dispose', tick);
               this.removeListener('end', tick);
```
kyranet, line 389: This isn't removing the `ignore` listener that's registering in line 394.

kyranet (changes requested):

kyranet:
Also, regarding this description:
> This PR adds new iterators and getters that allow you use collectors without deep nesting.

Technically you can achieve the helper iterators by filtering the main iterator:
```ts
for await (const [type, value] of collector) {
	if (type !== ...) continue;
}
```

However, they would use more resources than the dedicated ones, as it registers 2 more events. And the next one can achieved with similar code.

Praying to get [async iterator helpers](https://github.com/tc39/proposal-async-iterator-helpers) soon so we can do this 🙏🏼:
```ts
const next = await collector.filter(([type]) => type === ...).next();
```

PR #9180, opened 2023-02-27
feat: v13 add `InteractionResponse`

jaw0r3k:
 Adds `InteractionResponse` class


suneettipirneni:
While this is a feature in v14 the addition of this that was made was flawed. The main issue is that interaction responses only work for creating collectors on interaction message replies and don't work for any other kinds of interactions like message component interactions. Ideally, this functionality would be removed in future major versions of djs, due to the gaps in its functionality. Therefore I don't support this being added to v13.

imranbarbhuiya:
also, this will be a breaking change bcz of the change in return type

PR #9183, opened 2023-02-28
Add property `type` to SlashCommandBuilder

immjs:
Might help if people want to categorize their commands, instead of checking if the builder is a SlashCommandBuilder, they can simply check the `type` property of builder, very much like how `ContextMenuCommandBuilder` does it

### DIFF
--- a/packages/builders/src/interactions/slashCommands/SlashCommandBuilder.ts
+++ b/packages/builders/src/interactions/slashCommands/SlashCommandBuilder.ts
@@ -66,6 +66,11 @@ export class SlashCommandBuilder {
 	public readonly dm_permission: boolean | undefined = undefined;
 
 	/**
+	 * The type of this command
+	 */
+	public readonly type: ContextMenuCommandType = 1;
+
+	/**
 	 * Whether this command is NSFW
 	 */
 	public readonly nsfw: boolean | undefined = undefined;
--- a/packages/builders/src/interactions/slashCommands/SlashCommandBuilder.ts
+++ b/packages/builders/src/interactions/slashCommands/SlashCommandBuilder.ts
@@ -68,7 +68,7 @@ export class SlashCommandBuilder {
 	/**
 	 * The type of this command
 	 */
-	public readonly type: ContextMenuCommandType = 1;
+	public readonly type = 1;
 
 	/**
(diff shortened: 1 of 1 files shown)


Qjuh:
This is not consistent builders design. You use the builder while you want to add to it/change it. When you‘re done you can `.toJSON()` it and will have an object having that type set already.

immjs:
I had spent a few minutes looking at the toJSON and there is nothing adding `type` automagically. Simply, it is not required by the API.

If you want to talk about design, we can also talk about how ContextMenuCommandBuilder does have a `type` property yet SlashCommandBuilder doesn't, even though they end up in the same API call.

packages/builders/src/interactions/slashCommands/SlashCommandBuilder.ts
```diff
@@ -65,6 +65,11 @@ export class SlashCommandBuilder {
 	 */
 	public readonly dm_permission: boolean | undefined = undefined;
 
+	/**
+	 * The type of this command
+	 */
+	public readonly type = 1;
```
jaw0r3k, line 71: Also you shouldn't use magic numbers, as they aren't readable

immjs: Alright, added the relevant type

monbrey:
Sorry, I have to disagree. `ContextMenuCommandBuilder` has a type property because there are two types that have no other differences between them, and it's required by the API to define an application command as those types.

For the use-case you described developers can simply add a `type` property, `category` property, or literally any other property they like alongside the builder in the exports.

immjs:
> Sorry, I have to disagree. `ContextMenuCommandBuilder` has a type property because there are two types that have no other differences between them, and it's required by the API to define an application command as those types.
> 
> For the use-case you described developers can simply add a `type` property, `category` property, or literally any other property they like alongside the builder in the exports.

Well, what are the downsides?

jaw0r3k:
> Well, what are the downsides?

There is just no need to add it

kyranet:
The `ContextMenuCommandBuilder` class has a `type` property because it uses the old builders design, which we will phase out on builders v2 based on the [RFC](https://github.com/discordjs/discord.js/issues/8015).

Issue #9185, opened 2023-03-01
VoiceConnection stuck in signalling state about a minute after creating it


Theiremi:
### Which package is this bug report for?

voice

### Issue description

I've noticed a bug that seems to affect also some other users (see #8482).
All the VoiceConnection works as intended the first minute, and then stop playing anything. The player subscribed switch in AutoPaused and the connection stays signalling state indefinitely.

After looking at the code (https://github.com/discordjs/discord.js/blob/main/packages/voice/src/VoiceConnection.ts), it seems that the only way for it to go back to ready state is to receive a message from discord api, that then create events that can put status back in ready state.
But, if I understand correctly, this discord message never comes(line 335), so an event is not created through the function `configureNetworking`(line 338, 420), and the state cannot change to ready(line 484).
That's a messy explanation, and maybe false btw, but I've tried my best to understand.

### Code sample

```typescript
let voice_connection = Voice.joinVoiceChannel({
	adapterCreator: guild.voiceAdapterCreator,
	guildId: guild_id,
	channelId: channel_id,
	selfDeaf: true,
	selfMute: false
});

let player = new Voice.AudioPlayer({noSubscriber: Voice.NoSubscriberBehavior.Pause});
voice_connection.subscribe(player);

let resource = Voice.createAudioResource(local_file_or_stream);
player.play(resource);
```

### Package version

14.7.0 

### Node.js version

18.7.0; 18.14.2; 19.7.0 and maybe others

### Operating system

Windows, WSL, Ubuntu, Debian...

### Priority this issue should have

Medium (should be fixed soon)

### Which partials do you have configured?

Not applicable (subpackage bug)

### Which gateway intents are you subscribing to?

Guilds, GuildVoiceStates

Theiremi:
I've also found a workaround to this problem :

Add `this.configureNetworking();` on the line 355 of this file : https://github.com/discordjs/discord.js/blob/main/packages/voice/src/VoiceConnection.ts
_Or_
Add `this.configureNetworking();` on the line 1361 of the file located at node_modules/@discordjs/voice/dist/index.js

FloPMT:
> I've also found a workaround to this problem :
> 
> Add `this.configureNetworking();` on the line 355 of this file : https://github.com/discordjs/discord.js/blob/main/packages/voice/src/VoiceConnection.ts _Or_ Add `this.configureNetworking();` on the line 1361 of the file located at node_modules/@discordjs/voice/dist/index.js

[...]

PR #9195, opened 2023-03-03
Fix for voice connection issues (heartbeating, signalling stall)

davidzeng0:
VoiceUDPSocket: Fix for udp heartbeating. Discord servers only recognize 0x1337cafe as the header prefix

The following is source code from the discord app's discord_voice.node:
image

VoiceConnection: Fix reconnecting after disconnect

It seems like discord doesn't send a new server packet if nothing else
has changed. Reconnect if we already have a server packet and session_id
matches

### DIFF
--- a/packages/voice/src/VoiceConnection.ts
+++ b/packages/voice/src/VoiceConnection.ts
@@ -352,8 +352,6 @@ export class VoiceConnection extends EventEmitter {
 	 * @param packet - The received `VOICE_STATE_UPDATE` packet
 	 */
 	private addStatePacket(packet: GatewayVoiceStateUpdateDispatchData) {
-		this.packets.state = packet;
-
 		if (typeof packet.self_deaf !== 'undefined') this.joinConfig.selfDeaf = packet.self_deaf;
 		if (typeof packet.self_mute !== 'undefined') this.joinConfig.selfMute = packet.self_mute;
 		if (packet.channel_id) this.joinConfig.channelId = packet.channel_id;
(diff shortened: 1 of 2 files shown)


vladfrangu:
I personally would make an issue on Discord's api documentation github (https://github.com/discord/discord-api-docs) to ask them about UDP keep alive first! Using random magic numbers extracted from their client's voice implementation sounds like a bad idea

dzlandis:
Just connecting this PR to the issue.
This should hopefully resolve: https://github.com/discordjs/discord.js/issues/9185

davidzeng0:
> I personally would make an issue on Discord's api documentation github (https://github.com/discord/discord-api-docs) to ask them about UDP keep alive first! Using random magic numbers extracted from their client's voice implementation sounds like a bad idea

I think it's fine, see https://github.com/discordjs/discord.js/issues/9185#issuecomment-1454105863
that, and originally the UDP heartbeat was never documented in the first place.

I highly doubt this magic constant will ever change and it's 100% compliant with discord's own client.

vladfrangu:
You should still make an issue about UDP heartbeating in general! Getting an official stance on this (should we even do it, if yes how, etc) is much better than reverse engineering their client for magic values 😅

davidzeng0:
> You should still make an issue about UDP heartbeating in general! Getting an official stance on this (should we even do it, if yes how, etc) is much better than reverse engineering their client for magic values sweat_smile

Fair enough, though my work here is done.
I'll leave the official documentation issuing to someone else
who is more familiar with the discord ecosystem.

Issue #8482, opened 2022-08-12
VoiceConnection disconnects after roughly 1 minute every time


Beriu:
### Which package is this bug report for?

voice

### Issue description

I created a music bot while following the most recent guide and the voice connection dies out at around the same time. No errors are thrown.
After roughly one minute the voiceConnection changes state from connecting to signaling and freezes indefinitely until I restart, sometimes the bot doesn't even disconnect. Please ignore the `ytdl` package, I have tested this with a normal local file and it does the same thing.

I've logged the state changes because those are the only things that change, as I said no errors are thrown and voiceConnection decides to change from `Ready` to `Connecting` to `Signalling` and stays that way.

The project can be found at [https://github.com/Beriu/boomtsy](https://github.com/Beriu/boomtsy)

```
--------------------------------------------------
Core Dependencies
- @discordjs/voice: 0.11.0
- prism-media: 1.3.4

Opus Libraries
- @discordjs/opus: 0.8.0
- opusscript: 0.0.8

Encryption Libraries
- sodium-native: 3.4.1
- sodium: 3.0.2
- libsodium-wrappers: 0.7.10
- tweetnacl: 1.0.3

FFmpeg
- version: 5.0.1-static https://johnvansickle.com/ffmpeg/
- libopus: yes
--------------------------------------------------
```

### Code sample

```typescript
Bot is running...
Trying to create new voice connection: 777883619157999646
[17:11:13][VoiceConnection][signalling->connecting]
[17:11:13][VoiceConnection][connecting->connecting]
[17:11:13][VoiceConnection][connecting->connecting]
[17:11:13][VoiceConnection][connecting->connecting]
[17:11:13][VoiceConnection][connecting->ready]
[17:11:13][VoiceConnection][ready->ready]
Add command Queue size: 0
Start command Queue size: 1
[17:11:15][AudioPlayer][idle->buffering]
[17:11:15][AudioPlayer][buffering->playing]
[17:12:13][VoiceConnection][ready->connecting]
[17:12:13][AudioPlayer][playing->idle]
[17:12:13][VoiceConnection][connecting->signalling]
```

### Package version

14.2.1

### Node.js version

v16.16.0 (Typescript 4.7.4)

### Operating system

WSL 2 (Windows subsystem for Linux, Ubuntu)

### Priority this issue should have

Low (slightly annoying)

### Which partials do you have configured?

No Partials

### Which gateway intents are you subscribing to?

Guilds, GuildVoiceStates

[...]

PR #9202, opened 2023-03-05
fix(Voice): send keep alives without awaiting a response

vladfrangu:

Supersedes #9195 
Takes a part in solving part of #8482 and #9185 as recommended in DDevs. While UDP keepAlives aren't documented (and technically not needed either), we need to keep them in order to ensure voice receivers still work.

Tested too on a bot, and this change helped the audio not cut out after 1 minute (ala 12 keep alives packets sent)

### DIFF
--- a/packages/voice/src/networking/VoiceUDPSocket.ts
+++ b/packages/voice/src/networking/VoiceUDPSocket.ts
@@ -79,6 +79,8 @@ export class VoiceUDPSocket extends EventEmitter {
 
 	/**
 	 * The time taken to receive a response to keep alive messages.
+	 *
+	 * @deprecated This field is no longer updated as keep alive messages are no longer tracked.
 	 */
 	public ping?: number;
 
--- a/packages/voice/src/networking/VoiceUDPSocket.ts
+++ b/packages/voice/src/networking/VoiceUDPSocket.ts
@@ -12,11 +12,6 @@ export interface SocketConfig {
 	port: number;
 }
 
-interface KeepAlive {
-	timestamp: number;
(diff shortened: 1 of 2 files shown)


packages/voice/src/networking/VoiceUDPSocket.ts
```diff
@@ -128,16 +105,6 @@ export class VoiceUDPSocket extends EventEmitter {
 	 * @param buffer - The received buffer
 	 */
 	private onMessage(buffer: Buffer): void {
-		// Handle keep alive message
-		if (buffer.length === 8) {
-			const counter = buffer.readUInt32LE(0);
-			const index = this.keepAlives.findIndex(({ value }) => value === counter);
-			if (index === -1) return;
-			this.ping = Date.now() - this.keepAlives[index]!.timestamp;
```
kyranet, line 136: With this gone, wouldn't that make `ping` never assigned?

vladfrangu: 🤷‍♂️ Yes, but there's nothing we can do about it, and removing it would be semver major

Jiralite: Could at least deprecate it then?

Jiralite: Just realised this package is in 0-ver and I don't believe any other subpackage depends on this package. It should be fine to wholly remove?

kyranet (approved):

iCrawl (approved):

PR #9216, opened 2023-03-09
fix(Message#editable): update editable check in threads locked

ErwanGit:

[Recent changes](https://discord.com/developers/docs/change-log#update-to-locked-threads) regarding locked threads were rolled out this week.
If a bot does not have permission to manage threads, it will not be able to edit its own message even if the thread is not archived.
I've also added a check that was probably forgotten, a bot can't edit its own message if the thread is archived.

### DIFF
--- a/packages/discord.js/src/structures/Message.js
+++ b/packages/discord.js/src/structures/Message.js
@@ -604,7 +604,7 @@ class Message extends Base {
     const precheck = Boolean(this.author.id === this.client.user.id && (!this.guild || this.channel?.viewable));
 
     // Regardless of permissions thread messages cannot be edited if
-    // the thread is archived or the thread is locked and the bot does not have permission to manage threads
+    // the thread is archived or the thread is locked and the bot does not have permission to manage threads.
     if (this.channel?.isThread()) {
       if (this.channel.archived) return false;
       if (this.channel.locked) {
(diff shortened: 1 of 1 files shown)


jaw0r3k:
what? is doesn't make any sense
1. you deleted checking precheck
2. as I understood well locked still means locked
so you can edit messages in archived thread but not in locked 
3. only thing to leave is checking perms

jaw0r3k:
yeah the logic doesn't make any sense
as even in thread isn't archived it will check for the permissions

ErwanGit:
I re-tested before submitting this PR and concluded that : 
- if the thread is archived : impossible to modify the message without unarchiving the thread
- if the thread is locked : it is possible to modify the message only if the bot has the permission to manage the threads in the channel

I could make changes to check permissions only if the thread is locked, this is a mistake on my part 
But after that, I don't see what I could change, the current behavior of the editable check on locked threads is not good at the moment

packages/discord.js/src/structures/Message.js
```diff
@@ -602,11 +602,17 @@ class Message extends Base {
    */
   get editable() {
     const precheck = Boolean(this.author.id === this.client.user.id && (!this.guild || this.channel?.viewable));
+
     // Regardless of permissions thread messages cannot be edited if
-    // the thread is locked.
+    // the thread is archived or the thread is locked and the bot does not have permission to manage threads
```
jaw0r3k, line 607:
```suggestion
    // the thread is archived or the thread is locked and the bot does not have permission to manage threads.
```

DraftProducts (approved):

jaw0r3k:
still, you can edit messages in archived channel....

ErwanGit:
> still, you can edit messages in archived channel....

I just tried again and it seems that it is not possible to edit a post in an archived thread
image

SpaceEEC (approved):

vladfrangu (approved):

PR #9219, opened 2023-03-10
fix: add support for new guild feature `GUILD_WEB_PAGE_VANITY_URL`

pkdev08:

This will fix the following error when trying to fetch vanity data when the guild has the new web page feature enabled:

`Error [VANITY_URL]: This guild does not have the VANITY_URL feature enabled.`

This seems like the best fix at the moment as the `GUILD_WEB_PAGE_VANITY_URL` feature is not officially documented yet.


jaw0r3k:
You should just check both
But now error isn't used so delete it

kyranet (approved):

vladfrangu (approved):

SpaceEEC (approved):

PR #9220, opened 2023-03-11
docs(MessageManager): add clarification to fetch messages

andre-paulo98:
**Please describe the changes this PR makes and why it should be merged:** [Discord's documentation](https://discord.com/developers/docs/resources/channel#get-channel-messages) mentions that you can only use `before`, `after` or `around` when fetching messages. This is not mentioned in D.JS documentation, and after some time trying to figure out why I was getting more messages than I should, decided to read the documentation.
This only changes the documentation.

- Code changes have been tested against the Discord API, or there are **no code changes**
- I know how to update typings and have done so, **or typings don't need updating**

### DIFF
--- a/packages/discord.js/src/managers/MessageManager.js
+++ b/packages/discord.js/src/managers/MessageManager.js
@@ -43,6 +43,8 @@ class MessageManager extends CachedManager {
 
   /**
    * Options used to fetch a message.
+   * <info>The `before`, `after`, and `around` parameters are mutually exclusive (on Discord API side),
+   * only one may be passed at a time.</info>
    * @typedef {BaseFetchOptions} FetchMessageOptions
    * @property {MessageResolvable} message The message to fetch
    */


packages/discord.js/src/managers/MessageManager.js
```diff
@@ -43,6 +43,8 @@ class MessageManager extends CachedManager {
 
   /**
    * Options used to fetch a message.
+   * <info>The `before`, `after`, and `around` parameters are mutually exclusive (on Discord API side),
```
jaw0r3k, line 46: any sense of mentioning that it's exclusive on discord API side?

Jiralite: This is on the wrong type definition. `FetchMessageOptions` is only for fetching a message. It should be on the one directly beneath this.



> any sense of mentioning that it's exclusive on _discord API side?_

That should be removed.

andre-paulo98: you're right. I also messed up something on my side while commiting, so I'll close this and resubmit the PR

Issue #9182, opened 2023-02-28
typings(webhooks): Incoming webhook


DraftProducts:
### Which package is this bug report for?

discord.js

### Issue description

Actually, we have 3 types of webhooks :
- **Incoming**: Incoming Webhooks can post messages to channels with a generated token
- **ChannelFollower**: Channel Follower Webhooks are internal webhooks used with Channel Following to post new messages into channels
- **Application**: Application webhooks are webhooks used with Interactions

The method `<TextBasedChannel>.createWebhook()` returns an Incoming Webhook, but an `.isIncoming()` call is needed to apply correct typings.
The Webhook typing structure should have an Incoming default flag (like Cached flag in interactions) and could be changed internally for Interactions usage (or something similar like Types extend in a union).

I can assume the fix if this suggestion is accepted :v:

### Code sample

```typescript
Here an example of what could be done.

export type WebhookTypeReducer<
  State extends WebhookType,
  Target extends WebhookType,
  Type,
  Fallback = null,
> = State extends Target
  ? Type
  : State extends Exclude<WebhookType, Target>
  ? null
  : Type | null;

export class Webhook<Type extends WebhookType = WebhookType> extends WebhookMixin() {
  // [...]
  public sourceGuild: WebhookTypeReducer<Type, WebhookType.ChannelFollower, Guild | APIPartialGuild>;
  public sourceChannel: WebhookTypeReducer<Type, WebhookType.ChannelFollower, NewsChannel | APIPartialChannel>;
  public token: WebhookTypeReducer<Type, WebhookType.Incoming, string>;
  public applicationId: WebhookTypeReducer<Type, WebhookType.Application, Snowflake>;
}
```
```ts
export interface TextBasedChannelFields {
  // [...]
  createWebhook(options: ChannelWebhookCreateOptions): Promise<Webhook<WebhookType.Incoming>>;
  fetchWebhooks(): Promise<Collection<Snowflake, Webhook>>;
}
```

### Package version

main

### Priority this issue should have

Low (slightly annoying)

PR #9225, opened 2023-03-11
types: split `Webhook` into typed classes

jaw0r3k:
Resolves: 
- #9182

- This PR **only** includes non-code changes

### DIFF
--- a/packages/discord.js/typings/index.d.ts
+++ b/packages/discord.js/typings/index.d.ts
@@ -3257,6 +3257,8 @@ export class BaseWebhook extends WebhookMixin() {
 export class IncomingWebhook extends BaseWebhook {
   public type: WebhookType.Incoming;
   public applicationId: Snowflake | null;
+  public sourceGuild: null;
+  public sourceChannel: null;
   public token: string;
 }
 
@@ -3268,15 +3270,17 @@ export class UserCreatedWebhook extends IncomingWebhook {
 export class ApplicationCreatedWebhook extends BaseWebhook {
   public type: WebhookType.Application;
   public applicationId: Snowflake;
+  public sourceGuild: null;
+  public sourceChannel: null;
   public token: string;
 }
 
 export class ChannelFollowerWebhook extends BaseWebhook {
   public type: WebhookType.ChannelFollower;
+  public applicationId: null;
   public sourceGuild: Guild | APIPartialGuild;
   public sourceChannel: NewsChannel | APIPartialChannel;
   public token: null;
-  public applicationId: null;
 }
 
 export type Webhook = UserCreatedWebhook | ApplicationCreatedWebhook | IncomingWebhook | ChannelFollowerWebhook;
--- a/packages/discord.js/typings/index.d.ts
+++ b/packages/discord.js/typings/index.d.ts
@@ -3241,7 +3241,6 @@ export class BaseWebhook extends WebhookMixin() {
   public name: string;
   public owner: User | APIUser | null;
   public type: WebhookType;
-  public applicationId: Snowflake | null;
   public get channel(): TextChannel | VoiceChannel | NewsChannel | ForumChannel | StageChannel | null;
   public isUserCreated(): this is UserCreatedWebhook;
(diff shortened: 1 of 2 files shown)


Idris1401:
not sure but shouldn't type be changed for things in /src folder too?

jaw0r3k:
> not sure but shouldn't type be changed for things in /src folder too?

what?

NotSugden (changes requested):
You have removed the typing for the `Webhook` class and created a new typing for the export of `BaseWebhook` which is not exported from DJS. this would cause an error should someone try to import thos class and use it. and similarly any TS code that imports this class would break

jaw0r3k:
> You have removed the typing for the export `Webhook` class and created a new typing for the export of `BaseWebhook` which is not exported from DJS. this would cause an error should someone try to import thos class and use it. and similarly any TS code that imports this class would break
> 
> This PR needs a different soloution.

Webhook is still exported 
now just as an union of other types, which allows to typeguard it

NotSugden:
> > You have removed the typing for the export `Webhook` class and created a new typing for the export of `BaseWebhook` which is not exported from DJS. this would cause an error should someone try to import thos class and use it. and similarly any TS code that imports this class would break
> > This PR needs a different soloution.
> 
> Webhook is still exported now just as an union of other types, which allows to typeguard it
> 
> well yeah other classes have to be _abstract_

the class is still exported yes. But if you were do to

```ts
import { Webhook } from 'discord.js';

console.log(Webhook);
```

ts would tell you that you are unable to do this because as far it TS is concerned `Webhook` is not exported as a class. Its type is exported which would allow you to

```ts
import { Webhook } from 'discord.js'
const webhook = null as unknown as Webhook; // just to show that the type is imported

console.log(webhook);
```

and a user would also be able to do
```ts
import { BaseWebhook } from 'discord.js';

console.log(BaseWebhook);
```
in which TS would see it as fine but upon running this code they would get undefined as their result because BaseWebhook is not exported.

packages/discord.js/typings/index.d.ts
```diff
@@ -3231,44 +3231,21 @@ export class VoiceState extends Base {
   public edit(options: VoiceStateEditOptions): Promise<this>;
 }
 
-export class Webhook extends WebhookMixin() {
-  private constructor(client: Client<true>, data?: RawWebhookData);
+export class BaseWebhook extends WebhookMixin() {
```
NotSugden, line 3234: see here for previous comment. `export class BaseWebhook`
image
but we have no export

packages/discord.js/typings/index.d.ts
```diff
@@ -3277,6 +3254,37 @@ export class Webhook extends WebhookMixin() {
...
+  public token: string;
+}
+
+export class ChannelFollowerWebhook extends BaseWebhook {
+  public type: WebhookType.ChannelFollower;
+  public applicationId: null;
+  public sourceGuild: Guild | APIPartialGuild;
+  public sourceChannel: NewsChannel | APIPartialChannel;
+  public token: null;
+}
```
NotSugden, line 3284: This is also a similar issue. None of these classes are exported but as far as TS is concerned they are exported. Which is now more of a problem than before as someone may do

```ts
import { ChannelFollowerWebhook } from 'discord.js';

const webhook = [they got a webhook from somewhere];

if (webhook instanceof ChannelFollowerWebhook) {
  // do something
}
```
and their code will make a runtime error because you can't instanceof undefined

packages/discord.js/typings/index.d.ts
```diff
@@ -3277,6 +3254,37 @@ export class Webhook extends WebhookMixin() {
...
+
+export type Webhook = UserCreatedWebhook | ApplicationCreatedWebhook | IncomingWebhook | ChannelFollowerWebhook;
```
NotSugden, line 3286: here we export `Webhook` as a `type` and not as a `class` so the user cannot import it and use it as a class.

NotSugden (changes requested):

jaw0r3k:
Yeah this pr needs a remake

PR #9243, opened 2023-03-17
fix(rest): remove `const enum`s in favour of regular enums

favna:

The motivation is that `const enum` produces ambient const enums when compiling which in turn causes issues with TypeScript 5.x when `verbatimModuleSyntax` is enabled.

Furthermore, the generally accepted best practice is to avoid `const enum`s when writing libraries. Can back this up with statements from TS maintainers if needed, I know they made them, I just can't be bothered to find the GitHub links lmao. @vladfrangu will probably be able to find those links much easier than me as it was also the motivation to remove `const enum`'s from discord-api-types.

### DIFF
--- a/packages/rest/src/lib/RequestManager.ts
+++ b/packages/rest/src/lib/RequestManager.ts
@@ -113,7 +113,7 @@ export interface RequestHeaders {
 /**
  * Possible API methods to be used when doing requests
  */
-export const enum RequestMethod {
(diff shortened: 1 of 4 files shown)


vladfrangu (approved):

kyranet (approved):

SpaceEEC (approved):
🎉

PR #9251, opened 2023-03-19
docs: differ `User#send`

### LINKED ISSUE #9250: Examples and description for `User#send()` should differ from `TextBasedChannel#send()`
### Which package is this feature request for?

discord.js

### Feature

At the moment, [`User#send()`](https://discord.js.org/#/docs/discord.js/main/class/User?scrollTo=send) documents how to send messages to a [`TextBasedChannel`](https://github.com/discordjs/discord.js/blob/229ad077ff52d8706d68ed4d31983619a32eba45/packages/discord.js/src/structures/interfaces/TextBasedChannel.js#L16). Although identical, doesn't mean it wouldn't confuse people looking at the documentation. Especially since it's a pretty popular method.

Points of interest:
- "Sends a message to this channel."
- All examples, most notably:
```ts
// Send a basic message
channel.send('hello!')
```

A solution would be to make separate examples and a separate description for `User#send`. Therefore it wouldn't confuse people anymore as to why it says "Sends a message to this channel".

_No response_

### Other context

_No response_

jaw0r3k:

Resolves #9250
Also ith GuildMember example

Clears also textbasedchannel#send doc

-  There are no code changes
- Typings don't need updating


packages/discord.js/src/structures/User.js
```diff
@@ -305,12 +305,60 @@ class User extends Base {
...
+ * @example
+ * // Send an embed with a local image inside
+ * user.send({
+ *   content: 'This is an embed',
+ *   embeds: [
+ *     {
+ *       thumbnail: {
+ *         url: 'attachment://file.jpg'
+ *       }
+ *     }
```
jaw0r3k, line 350: As in original i put there *raw* object
but should we show `EmbedBuilder`? the same with attachment

TetieWasTaken: I think the raw is fine, since we're documenting `User#send()`. And not how to use an `EmbedBuilder`. Although on the other hand it might be better to point new users to a better direction with a comment

monbrey: Four examples is a lot of bloat to be filling the docs page with. Is it really necessary? Do we have four on the other send examples?

TetieWasTaken: The changes only differ in variables and comments. [`TextBasedChannel#send()`](https://github.com/discordjs/discord.js/blob/229ad077ff52d8706d68ed4d31983619a32eba45/packages/discord.js/src/structures/interfaces/TextBasedChannel.js#L115) also has four examples. It should be shortened to 1-2 imo.

Jiralite: Why not generalise the existing documentation to account for both users and channels?

Jiralite:
You must use a closing keyword following by the issue for it to link!

packages/discord.js/src/structures/User.js
jaw0r3k: > Why not generalise the existing documentation to account for both users and channels?

Is possible, but maybe misleading as not everybody thinks about a user like a channel

packages/discord.js/src/structures/interfaces/TextBasedChannel.js
```diff
@@ -135,25 +135,6 @@ class TextBasedChannel {
    * })
    *   .then(console.log)
    *   .catch(console.error);
-   * @example
```
jaw0r3k, line 138: Example is shown in guide, so its deleted

kyranet (approved):

packages/discord.js/src/structures/interfaces/TextBasedChannel.js
vladfrangu: I don't like the removal of this example, both bc its outside the scope of the PR and also because its a useful example to have

jaw0r3k: then delete sending files?

Qjuh: No, why would you? The whole example is useful and correct.

jaw0r3k: https://github.com/discordjs/discord.js/pull/9251#discussion_r1141489850

vladfrangu (approved):

Jiralite:
What about `UserManager#send()`? Would that need updating too?

TetieWasTaken:
> What about [`UserManager#send()`](https://discord.js.org/#/docs/discord.js/main/class/UserManager?scrollTo=send)? Would that need updating too?

I don't think that's within the scope of this PR. But more examples are almost always appreciated—so I do think it should be featured in a different PR.

SpaceEEC (approved):

PR #9262, opened 2023-03-21
fix: resolving string bitfield

### LINKED ISSUE #9247: Incorrect resolve of bits in `BitField`

discord.js

### Issue description

Due to the use of enums flags have reverse values. If we use string bit filed the bit resolves to the key instead of the value.
https://github.com/discordjs/discord.js/blob/98420826bc2296fc392f17e8254cf4ad743ff5af/packages/discord.js/src/util/BitField.js#L166-L169

### Code sample

```typescript
import { GuildMemberFlagsBitField } from 'discord.js';

const permission1 = new GuildMemberFlagsBitField('1');
const permission2 = new GuildMemberFlagsBitField('2');
const permission3 = new GuildMemberFlagsBitField('3');

console.log(permission1.bitfield); // 'DidRejoin'
console.log(permission2.bitfield); // 'CompletedOnboarding'
console.log(permission3.bitfield); // 3
console.log(GuildMemberFlagsBitField.Flags);
/**
{
  '1': 'DidRejoin',
  '2': 'CompletedOnboarding',
  '4': 'BypassesVerification',
  '8': 'StartedOnboarding',
  DidRejoin: 1,
  CompletedOnboarding: 2,
  BypassesVerification: 4,
  StartedOnboarding: 8
}
**/
```

### Package version

14.8.0

### Node.js version

18.15.0

### Operating system

Win 11

Medium (should be fixed soon)

No Partials

No Intents

_No response_

jaw0r3k:
Well the solution it is really simple... just change the order
 Fixes #9247
- Typings don't need updating


kyranet (approved):
This comes with quite the performance impact, a correct solution would be to remove the two-way binding from the enums, but I'm afraid that would be a breaking change.

SpaceEEC (approved):

PR #9267, opened 2023-03-23
chore: Include discord.js in the user agent string

Jiralite:
This includes discord.js in the user agent string, even if a custom appendix is supplied.


vladfrangu (approved):

packages/discord.js/src/client/BaseClient.js
```diff
@@ -14,6 +14,12 @@ class BaseClient extends EventEmitter {
   constructor(options = {}) {
     super({ captureRejections: true });
 
+    if (options.rest?.userAgentAppendix !== undefined) {
+      // Merging the default options when a custom user agent appendix is supplied
+      // Replaces the discord.js string. Enforce it.
+      options.rest.userAgentAppendix = `${Options.userAgentAppendix} ${options.rest?.userAgentAppendix}`;
```
kyranet, line 20: The existence of the field is already checked by line 17.
```suggestion
      options.rest.userAgentAppendix = `${Options.userAgentAppendix} ${options.rest.userAgentAppendix}`;
```

packages/discord.js/src/client/BaseClient.js
```diff
@@ -14,6 +14,12 @@ class BaseClient extends EventEmitter {
   constructor(options = {}) {
     super({ captureRejections: true });
 
+    if (options.rest?.userAgentAppendix !== undefined) {
```
kyranet, line 17: This should be below line 23 so it checks for `options`'s type first. Otherwise passing `null` will result on a different error.

kyranet (changes requested):

packages/discord.js/src/client/BaseClient.js
```diff
@@ -18,6 +18,12 @@ class BaseClient extends EventEmitter {
       throw new DiscordjsTypeError(ErrorCodes.InvalidType, 'options', 'object', true);
     }
 
+    if (options.rest?.userAgentAppendix !== undefined) {
```
almeidx, line 21: Appending an empty string is probably not very useful
```suggestion
    if (options.rest?.userAgentAppendix) {
```

packages/discord.js/src/util/Options.js
```diff
@@ -70,6 +71,14 @@ const { toSnakeCase } = require('./Transformers');
  * Contains various utilities for client options.
  */
 class Options extends null {
+  /**
+   * The default user agent appendix.
+   * @type {string}
+   * @memberof Options
+   * @private
+   */
+  static userAgentAppendix = `discord.js/${version} ${DefaultUserAgentAppendix}`.trim();
```
almeidx, line 80: With the changes to `DefaultUserAgentAppendix`, there are no extra spaces here anymore
```suggestion
  static userAgentAppendix = `discord.js/${version} ${DefaultUserAgentAppendix}`;
```

almeidx (changes requested):
As a side note, shouldn't /rest's default user agent URL point to the new docs (`https://discordjs.dev/docs/packages/rest` or something), or use `@discordjs/rest@<version>` in the version itself? Currently, it's set to `DiscordBot (https://discord.js.org, 1.6.0)`

packages/discord.js/src/util/Options.js
almeidx: oops, yeah. I mistook it for `DefaultUserAgent`. `trimEnd()` can be used here instead then

Jiralite:
From https://discord.com/developers/docs/reference#user-agent-user-agent-example, there doesn't seem to be a place to include `@discordjs/rest`—supposedly, the URL accounts for that. Changing the URL to the new documentation site would probably be best done in another pull request where we change everything over.

SpaceEEC (approved):

PR #9282, opened 2023-03-27
refactor(WebSocketShard): waitForEvent and its error handling

didinele:
yet again, because this stuff was a mess.

all in all, this PR does the following things:
- `waitForEvent` now properly cancels if the connection closes while waiting (oops!!! this was a giant oversight that caused niche bugs)
- got rid of the separate bubble error method, since it wasn't all that great
- `connect` no longer throws if something times out during the initial connect, it'll just keep trying until we get Ready (unless something is hard-thrown, like no-auth)
- deals with sessions better, no longer caching them on the class itself
  - adds a nice debug log that should help users identify if their `updateSessionInfo` callback is being nasty

tagged as draft until @vladfrangu messes with it.

### DIFF
--- a/packages/ws/src/ws/WebSocketShard.ts
+++ b/packages/ws/src/ws/WebSocketShard.ts
@@ -101,14 +101,12 @@ export class WebSocketShard extends AsyncEventEmitter<WebSocketShardEventsMap> {
 
 	private lastHeartbeatAt = -1;
 
-	private session: SessionInfo | null = null;
-
 	// Indicates whether the shard has already resolved its original connect() call
 	private initialConnectResolved = false;
 
 	private readonly sendQueue = new AsyncQueue();
 
-	private readonly timeouts = new Collection<WebSocketShardEvents, NodeJS.Timeout>();
+	private readonly timeoutAbortControllers = new Collection<WebSocketShardEvents, AbortController>();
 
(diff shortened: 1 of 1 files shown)


packages/ws/src/ws/WebSocketShard.ts
```diff
@@ -127,6 +125,14 @@ export class WebSocketShard extends AsyncEventEmitter<WebSocketShardEventsMap> {
 	}
 
 	public async connect() {
+		const promise = this.initialConnectResolved ? Promise.resolve() : once(this, WebSocketShardEvents.Ready);
+		void this.internalConnect();
+
+		await promise;
+		this.initialConnectResolved = true;
```
vladfrangu, line 132: This is never reset

vladfrangu (changes requested):

packages/ws/src/ws/WebSocketShard.ts
didinele: Why would it? We only need it to be `false` *once*, during the initial connect. A shard that is completely `destroy`ed is cleared/re-spawned either way if re-connection is requested later for the sake of keeping things simple.

vladfrangu (approved):

kyranet (approved):

SpaceEEC (approved):

PR #9288, opened 2023-03-28
feat: Voice Channel Send Effects

Jiralite:

- https://github.com/discord/discord-api-docs/pull/6025
- https://github.com/discordjs/discord-api-types/pull/739


jaw0r3k:
may you add channel and guild getter?

jaw0r3k:
also shouldn't it be an actions, which handlers just handle
like every other event

packages/discord.js/src/client/websocket/handlers/VOICE_CHANNEL_EFFECT_SEND.js
```diff
@@ -0,0 +1,10 @@
+'use strict';
+
+const VoiceChannelEffect = require('../../../structures/VoiceChannelEffect');
+const Events = require('../../../util/Events');
+
+module.exports = (client, { d: data }) => {
+  const guild = client.guilds.cache.get(data.guild_id);
+  if (!guild) return;
+  client.emit(Events.VoiceChannelEffectSend, new VoiceChannelEffect(data, guild));
```
jaw0r3k, line 9:
```suggestion
  /**
   * Emmited when someone sends an effect, such as an emoji reaction,
   * in a voice channel the current user is connected to.
   * @event Client#voiceChannelEffectSend
   * @param {VoiceChannelEffect} voiceChannelEffect The sent voice channel effect
   */
  client.emit(Events.VoiceChannelEffectSend, new VoiceChannelEffect(data, guild));
```

packages/discord.js/src/structures/VoiceChannelEffect.js
```diff
@@ -0,0 +1,58 @@
+'use strict';
+
+const { Emoji } = require('./Emoji');
+
+/**
+ * Represents an effect used in a {@link VoiceChannel}.
+ */
+class VoiceChannelEffect {
```
jaw0r3k, line 8: Shouldnt it extend `Base`
> like tetie said

Jiralite:
> may you add channel getter?

This seems to already be done.

Jiralite:
> also shouldn't it be an action, which handlers just handle like every other event?

https://github.com/discordjs/discord.js/blob/b27c1d0819ce408899fd0ced404c0a3dd6b19be9/packages/discord.js/src/client/actions/Action.js#L5-L15

My take on this is until there are any REST events that affect this event or we need to manually call this code, there is no need to add it. It is a very simple structure: receive a WebSocket event then emit it. Also, not every event has an action. So no, it shouldn't be an action.

packages/discord.js/src/structures/VoiceChannelEffect.js
Jiralite: Probably not. Extending `Base` would mean this would (not publically) add `_clone()`, `_patch()`, and `_update()` to the class which are all irrelevant methods. In addition, `valueOf()` for this structure would return `undefined` as there is no `id` sent in voice channel effects.

PR #9309, opened 2023-04-01
chore: remove `GuildMember` id getters

TetieWasTaken:
This PR removes `id` getters for the `GuildMember` and `TeamMember` structures

### Why?
These properties/getters imply that guild members have ids, which is a bad practise and misleading. Even though it's a useful getter I don't believe we should prefer readability over actual truthfulness. Especially since the descriptions state: "The Member's id". It isn't a very complicated property either, the only change people have to make is:
```diff
- <GuildMember>.id
+ <GuildMember>.user?.id
```

### DIFF
--- a/packages/discord.js/src/client/actions/ThreadListSync.js
+++ b/packages/discord.js/src/client/actions/ThreadListSync.js
@@ -29,7 +29,7 @@ class ThreadListSyncAction extends Action {
 
     for (const rawMember of Object.values(data.members)) {
       // Discord sends the thread id as id in this object
-      const thread = client.channels.cache.get(rawMember.id);
+      const thread = client.channels.cache.get(rawMember.user.id);
       if (thread) {
         thread.members._add(rawMember);
       }
--- a/packages/discord.js/src/client/actions/ThreadListSync.js
+++ b/packages/discord.js/src/client/actions/ThreadListSync.js
@@ -29,7 +29,7 @@ class ThreadListSyncAction extends Action {
 
     for (const rawMember of Object.values(data.members)) {
       // Discord sends the thread id as id in this object
-      const thread = client.channels.cache.get(rawMember.user.id);
+      const thread = client.channels.cache.get(rawMember.id);
       if (thread) {
         thread.members._add(rawMember);
       }
--- a/packages/docgen/src/documentation.ts
+++ b/packages/docgen/src/documentation.ts
@@ -123,7 +123,7 @@ export class Documentation {
 
 					case 'Method': {
 						const event = prop?.groups?.find((group) => group.title === 'Events');
-						if ((event?.children as unknown as number[])?.includes(member.id)) {
+						if ((event?.children as unknown as number[])?.includes(member.user.id)) {
 							item = new DocumentedEvent(member, this.config);
 							break;
 						}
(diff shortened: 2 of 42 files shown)


jaw0r3k:
For average user it's easier to just use `.id` 
so i don't agree with that pr

TetieWasTaken:
> For average user it's easier to just use `.id` so i don't agree with that pr

Yes, I agree that it's easier. But it still implies that guild members have ids—and even though it's a small thing this could lead to even more misleading properties/methods. And, it's harder for people to differentiate users and guild members.

TetieWasTaken:
Uncertain about the possibility of `member#user` being null, we'll have to check for that aswell

gamer0mega:
> > For average user it's easier to just use `.id` so i don't agree with that pr
> 
> Yes, I agree that it's easier. But it still implies that guild members have ids—and even though it's a small thing this could lead to even more misleading properties/methods. And, it's harder for people to differentiate users and guild members. It's comparable to if `GuildMember#createdTimestamp` would exist.

By this logic, GuildMember#send(), GuildMember#createDM(), GuildMember#deleteDM() and GuildMember#dmChannel should be removed aswell, as they imply that the client messages the member and not the user, while internally just executing the methods/getters on this.user.

Jiralite:
> By this logic, `GuildMember#send()`, `GuildMember#createDM()`, `GuildMember#deleteDM()` and `GuildMember#dmChannel` should be removed aswell, as they imply that the client messages the member and not the user, while internally just executing the methods/getters on this.user.

Just so you know, they are also helpers on the `User` class, so they should be removed there too.

packages/docgen/src/documentation.ts
```diff
@@ -123,7 +123,7 @@ export class Documentation {
 
 					case 'Method': {
 						const event = prop?.groups?.find((group) => group.title === 'Events');
-						if ((event?.children as unknown as number[])?.includes(member.id)) {
+						if ((event?.children as unknown as number[])?.includes(member.user.id)) {
```
Jiralite, line 126: This is wrong.

TetieWasTaken:
> By this logic, GuildMember#send(), GuildMember#createDM(), GuildMember#deleteDM() and GuildMember#dmChannel should be removed aswell [...]

I understand what you mean, but I think those methods have a bigger purpose than just `#id`. I do agree with your view but I think `GuildMember#id` is there for no reason—if that makes sense...

packages/docgen/src/documentation.ts
TetieWasTaken: I'm assuming this has to be reverted—I think I accidentally missed the fact that this was not an actual `GuildMember`

packages/discord.js/src/client/actions/ThreadListSync.js
```diff
@@ -29,7 +29,7 @@ class ThreadListSyncAction extends Action {
 
     for (const rawMember of Object.values(data.members)) {
       // Discord sends the thread id as id in this object
-      const thread = client.channels.cache.get(rawMember.id);
+      const thread = client.channels.cache.get(rawMember.user.id);
```
Qjuh, line 32: This is also wrong. Don’t just add .user.id behind everything that has the name member. This is a threadMember from the API and the id is the id of the ThreadChannel, nothing to do with GuildMember or User.

Qjuh (changes requested):

monbrey:
I don't see any actual benefit in removing these.

Yes, the id originates from the User object, but I wouldn't consider it "misleading" in any way for these structures to also have the property. It's entirely reasonable and understandable to say that a GuildMember has an id. It's a transitive property, as all GuildMembers have only a single corresponding User.

TetieWasTaken:
You're very right

PR #9354, opened 2023-04-09
fix(ExcerptText): Use `resolveItemURI()` in excerpts

Jiralite:
After #9351, I noticed excerpt texts were also being incorrectly linked. An example can be found [here](https://discord.js.org/docs/packages/voice/main/AudioPlayerPausedState:Interface#status) where clicking either link of the `AudioPlayerStatus` enumeration would lead to a dead page. Simply using the fixed utility function resolves this.


apps/website/src/components/ExcerptText.tsx
```diff
@@ -30,7 +31,7 @@ export function ExcerptText({ model, excerpt }: ExcerptTextProps) {
 					return (
 						<ItemLink
 							className="text-blurple"
-							itemURI={`${item.displayName}:${item.kind}`}
+							itemURI={`${resolveItemURI(item)}`}
```
iCrawl, line 34:
```suggestion
							itemURI={resolveItemURI(item)}
```

iCrawl (approved):

Issue #9350, opened 2023-04-08
BitField#toArray() returning the bitfield with the bitfield name


joeyk710:
### Which package is this bug report for?

discord.js

### Issue description

- When accessing the method [BitField#toArray()](https://old.discordjs.dev/#/docs/discord.js/main/class/BitField?scrollTo=toArray), this is suppose to return **only** the Array of BitField *names* based on the bit given. The actual result returns the bitfield itself as a string as well as the bitfield name, which is the expected result.

- The PR #9262, and the issue #9247 both attempt to address this issue, but does not occur on `v14.8.0` and is shown below. I am unsure why this change was necessary since I was not experiencing these issues at all in previous versions. 
https://github.com/discordjs/discord.js/blob/98420826bc2296fc392f17e8254cf4ad743ff5af/packages/discord.js/src/util/BitField.js#L166-L169

- On `v14.9.0`, this has line 167 moved to line 168 (shown below), which is indeed causing an issue with the bitfield and bitfield name showing together.
https://github.com/discordjs/discord.js/blob/1b9d07f941dc3edf20cd5e65c56b70d009880893/packages/discord.js/src/util/BitField.js#L166-L169

Here is the result I got on `v14.9.0` accessing [UserFlagsBitField#toArray()](https://old.discordjs.dev/#/docs/discord.js/main/class/UserFlagsBitField?scrollTo=toArray)

![image](https://user-images.githubusercontent.com/71621973/230742148-4c41177c-28d3-4998-96b8-be30206d6ac4.png)

This is the *temporary* workaround I made so the bitfield is removed from the array:
```ts
const flagStrings = message.author.flags.toArray().map((flagName) => flagName);

const flagSplice = flagStrings.splice(0, flagStrings.length / 2, ...flagStrings.splice(0, flagStrings.length / 2).reverse());
```
<img width="825" alt="image" src="https://user-images.githubusercontent.com/71621973/230742805-09078a18-bb42-41cb-8a4c-3fe536478868.png">

- It would appear that [PermissionsBitField#toArray()](https://old.discordjs.dev/#/docs/discord.js/main/class/PermissionsBitField?scrollTo=toArray) does not have this issue from testing, but other extended BitField classes do. This does show the expected result with the permission bitfield **name** only.
<img width="545" alt="image" src="https://user-images.githubusercontent.com/71621973/230744032-25e8e763-2ed3-4230-aa56-da9ed84b42fe.png">

### Code sample

```typescript
[...]

PR #9357, opened 2023-04-09
fix(BitField): use only enum names in iterating

jaw0r3k:
Resolves: #9350
The change may seem weird, but its working.
As enums are reverse mapped, getting keys will not only give names but values
The shown exception was permissions, but its just an object

- Typings don't need updating


packages/discord.js/src/util/BitField.js
```diff
@@ -140,7 +140,7 @@ class BitField {
 
   *[Symbol.iterator](...hasParams) {
     for (const bitName of Object.keys(this.constructor.Flags)) {
-      if (this.has(bitName, ...hasParams)) yield bitName;
+      if (typeof bitName === 'string' && this.has(bitName, ...hasParams)) yield bitName;
```
vladfrangu, line 143: I don't see how this fixes anything, considering the keys will always be strings 
Code - Insiders - 2023-04-09 at 18 57 24@2x

vladfrangu (changes requested):

packages/discord.js/src/util/BitField.js
jaw0r3k: Oh sorry if forgot to change the commit 
Now its working fine 
image

vladfrangu (approved):
sure

kyranet (approved):

Jiralite:
It looks like this needs backporting.

jaw0r3k:
> It looks like this needs backporting, would you mind doing it?

Sure <3

SpaceEEC (approved):

SpaceEEC (changes requested):
On a second look, this needs to be applied to other places traversing `Flags` too.
This only seems to be `serialize` (currently).

---

```js
> new ActivityFlagsBitField(511).serialize()
{
  '1': true,
  '2': true,
  '4': true,
  '8': true,
  '16': true,
  '32': true,
  '64': true,
  '128': true,
  '256': true,
  Instance: true,
  Join: true,
  Spectate: true,
  JoinRequest: true,
  Sync: true,
  Play: true,
  PartyPrivacyFriends: true,
  PartyPrivacyVoiceChannel: true,
  Embedded: true
}
```

SpaceEEC (approved):

PR #9358, opened 2023-04-09
fix(BaseSelectMenuBuilder): Modify class to be `abstract`

Jiralite:
The purpose of this class is to share methods and properties for select menu builders. It would be a mistake to instantiate this class. Therefore, it should be appropriately marked as `abstract`.

- Code changes have been tested, or there are no code changes


packages/builders/src/index.ts
```diff
@@ -13,7 +13,6 @@ export * as TextInputAssertions from './components/textInput/Assertions.js';
 export * from './interactions/modals/Modal.js';
 export * as ModalAssertions from './interactions/modals/Assertions.js';
 
-export * from './components/selectMenu/BaseSelectMenu.js';
```
SpaceEEC, line 16: Why not export it?
You should still be able to use `instanceof` with the base class.

SpaceEEC (changes requested):

packages/builders/src/index.ts
Jiralite: @vladfrangu

Jiralite: We opted to keep this.

kyranet (approved):

iCrawl (approved):

PR #9373, opened 2023-04-13
Revert "chore: disable scope-case rule for commitlint"

Sxxvega:
Reverts discordjs/discord.js#7507


TetieWasTaken:
Why should this be reverted?

PR #9271, opened 2023-03-23
chore: Miscellaneous fixes

Jiralite:
Backports the following pull requests to version 13:

- #8879
- #9144
- #9146
- #9216
- #9219
- #9238
- #9239
- #9240
- #9241
- #9251
- #9252
- #9259
- #9262
- #9263
- #9270
- #9272
- #9293

PR #9377, opened 2023-04-13
fix(BitField): use only enum names in iterating

jaw0r3k:
Backports to v13:
- #9357 

- Typings don't need updating
- This PR only includes non-code changes, like changes to documentation, README, etc.


vladfrangu:
Just curious, does this version even have the same bug? 👁️

Jiralite:
Probably? https://github.com/discordjs/discord.js/pull/9271/commits/630d4f33cd1a18a10b371233fe74a081da9f2d8e

SpaceEEC (changes requested):
I'm unable to see anything to fix here, we don't seem to use any TS-enums in v13. 😅 
(Otherwise this would probably need to be applied to `serialize` too.)

jaw0r3k:
Well you are right, v13 uses only pure objects

PR #9383, opened 2023-04-13
refactor(GuildMember): make _roles a getter and roles a normal prop

Idris1401:

I think it would be better to have `GuildMember#roles` to be defined as a "normal" property (dunno what's the right term) instead of a getter and make `GuildMember#_roles` a getter.
People are often `console.log`ing data to get available properties and `._roles` does appear in the log and `.roles` doesn't which makes people using a private property.
I was also told that this change could cause performance issue but I still wanted to make this PR to know everyone's thought about it.


kyranet:
You messed up an update very badly. Also, there are reasons why `roles` is a getter: `_roles` is the raw data from Discord, and the manager is very memory heavy.

PR #9387, opened 2023-04-14
refactor(GuildMember): make `_roles` property non-enumerable

Idris1401:

I think it would be better to have `GuildMember#roles` to be defined as a "normal" property (dunno what's the right term) instead of a getter and make `GuildMember#_roles` a getter.
People are often `console.log`ing data to get available properties and `._roles` does appear in the log and `.roles` doesn't which makes people using a private property.
I was also told that this change could cause performance issue but I still wanted to make this PR to know everyone's thought about it.

*I made another PR because I fucked up the branch in the old one.*


SpaceEEC (changes requested):
I'd prefer `_roles` to be made non-enumerable over this change here, which would require further considerations (e.g. updates for all members when a role is deleted)

Jiralite:
> People are often `console.log`ing data to get available properties and `._roles` does appear in the log and `.roles` doesn't which makes people using a private property.

Frankly, that's their fault. Additionally, the property is prefixed with `_` which is the convention for something that is private. I do not see this as a valid reason for this change. It instead seems like a reason to cater towards individuals that do not know what they are doing or know what they are doing is wrong.

Since #9529 landed, I am pretty sure this reverts the performance improvements.

Idris1401:
> I'd prefer `_roles` to be made non-enumerable over this change here, which would require further considerations (e.g. updates for all members when a role is deleted)

what does "non-enumerable" mean?

SpaceEEC:
The `descriptor` parameter of [`Object.defineProperty`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/defineProperty) has an `enumerable` property, setting that to `false` is what I meant.
I think we are doing this with the `client` field in the `Base` class.

Idris1401:
> The `descriptor` parameter of [`Object.defineProperty`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/defineProperty) has an `enumerable` property, setting that to `false` is what I meant. I think we are doing this with the `client` field in the `Base` class.

based on MDN, `enumerable` is default set to `false` so I didn't add it

packages/discord.js/src/structures/GuildMember.js
```diff
@@ -56,10 +56,12 @@ class GuildMember extends Base {
 
     /**
      * The role ids of the member
+     * @name GuildMember#_roles
      * @type {Snowflake[]}
      * @private
      */
-    this._roles = [];
+    Object.defineProperty(this, '_roles', { value: [] });
```
SpaceEEC, line 63: I think you also need to add `writable` there, otherwise assigning to it will throw. 😅

SpaceEEC (changes requested):

vladfrangu (approved):

kyranet (approved):

SpaceEEC (approved):

PR #9416, opened 2023-04-18
refactor(rest): switch api to fetch-like and provide strategies

ckohen:

Undici request was originally chosen for performance reasons. ~~At the scale of discord bots, this performance difference is irrelevant.~~ This instead is the first step towards cross environment compatability.

This refactor also changes the actual network request portion into a strategy. By default, this will use `undici.request`. This strategy has also been provided and exported as /undiciRequest

### DIFF
--- a/packages/proxy/src/util/responseHelpers.ts
+++ b/packages/proxy/src/util/responseHelpers.ts
@@ -1,7 +1,7 @@
 import type { ServerResponse } from 'node:http';
+import { Readable } from 'node:stream';
 import { pipeline } from 'node:stream/promises';
-import { DiscordAPIError, HTTPError, RateLimitError } from '@discordjs/rest';
-import type { Dispatcher } from 'undici';
+import { DiscordAPIError, HTTPError, RateLimitError, type ResponseLike } from '@discordjs/rest';
 
 /**
  * Populates a server response with the data from a Discord 2xx REST response
@@ -9,19 +9,21 @@ import type { Dispatcher } from 'undici';
  * @param res - The server response to populate
  * @param data - The data to populate the response with
  */
-export async function populateSuccessfulResponse(res: ServerResponse, data: Dispatcher.ResponseData): Promise<void> {
-	res.statusCode = data.statusCode;
+export async function populateSuccessfulResponse(res: ServerResponse, data: ResponseLike): Promise<void> {
+	res.statusCode = data.status;
 
-	for (const header of Object.keys(data.headers)) {
+	for (const [header, value] of data.headers) {
 		// Strip ratelimit headers
-		if (header.startsWith('x-ratelimit')) {
+		if (/^x-ratelimit/i.test(header)) {
 			continue;
 		}
 
-		res.setHeader(header, data.headers[header]!);
+		res.setHeader(header, value);
 	}
 
-	await pipeline(data.body, res);
+	if (data.body) {
+		await pipeline(data.body instanceof Readable ? data.body : Readable.fromWeb(data.body), res);
+	}
 }
 
 /**
--- a/packages/rest/src/lib/REST.ts
+++ b/packages/rest/src/lib/REST.ts
@@ -1,6 +1,8 @@
 import { EventEmitter } from 'node:events';
+import type { Readable } from 'node:stream';
+import type { ReadableStream } from 'node:stream/web';
 import type { Collection } from '@discordjs/collection';
-import type { request, Dispatcher } from 'undici';
+import type { Dispatcher, RequestInit, Response } from 'undici';
 import { CDN } from './CDN.js';
 import {
 	RequestManager,
@@ -11,7 +13,7 @@ import {
 	type RequestData,
 	type RouteLike,
 } from './RequestManager.js';
-import type { IHandler } from './handlers/IHandler.js';
+import type { IHandler } from './interfaces/Handler.js';
 import { DefaultRestOptions, RESTEvents } from './utils/constants.js';
 import { parseResponse } from './utils/utils.js';
 
@@ -22,7 +24,7 @@ export interface RESTOptions {
 	/**
 	 * The agent to set globally
 	 */
-	agent: Dispatcher;
+	agent: Dispatcher | null;
 	/**
 	 * The base api path, without version
 	 *
@@ -79,6 +81,13 @@ export interface RESTOptions {
 	 * @defaultValue `0`
 	 */
 	invalidRequestWarningInterval: number;
+	/**
+	 * The method called to perform the actual HTTP request given a url and web `fetch` options
+	 * For example, to use global fetch, simply provide `makeRequest: fetch`
+	 *
+	 * @defaultValue `undici.request`
+	 */
+	makeRequest(url: string, init: RequestInit): Promise<ResponseLike>;
 	/**
 	 * The extra offset to add to rate limits in milliseconds
 	 *
@@ -179,7 +188,7 @@ export interface APIRequest {
 	/**
 	 * Additional HTTP options for this request
 	 */
-	options: RequestOptions;
+	options: RequestInit;
 	/**
 	 * The full path used to make the request
 	 */
@@ -192,6 +201,11 @@ export interface APIRequest {
 	 * The API route identifying the ratelimit for this request
 	 */
 	route: string;
+}
+
+export interface ResponseLike
+	extends Pick<Response, 'arrayBuffer' | 'bodyUsed' | 'headers' | 'json' | 'ok' | 'status' | 'text'> {
+	body: Readable | ReadableStream | null;
 }
 
 export interface InvalidRequestWarningData {
@@ -212,7 +226,7 @@ export interface RestEvents {
 	newListener: [name: string, listener: (...args: any) => void];
 	rateLimited: [rateLimitInfo: RateLimitData];
 	removeListener: [name: string, listener: (...args: any) => void];
-	response: [request: APIRequest, response: Dispatcher.ResponseData];
+	response: [request: APIRequest, response: ResponseLike];
 	restDebug: [info: string];
 }
 
@@ -232,8 +246,6 @@ export interface REST {
 	removeAllListeners: (<K extends keyof RestEvents>(event?: K) => this) &
 		(<S extends string | symbol>(event?: Exclude<S, keyof RestEvents>) => this);
 }
-
-export type RequestOptions = Exclude<Parameters<typeof request>[1], undefined>;
 
 export class REST extends EventEmitter {
 	public readonly cdn: CDN;
--- a/packages/rest/src/lib/global/fetch.d.ts
+++ b/packages/rest/src/lib/global/fetch.d.ts
@@ -1,0 +1,5 @@
+import type * as undici from 'undici';
+
+declare global {
+	export const { fetch, FormData, Headers, Request, Response }: typeof undici;
+}
--- a/packages/rest/src/lib/utils/constants.ts
+++ b/packages/rest/src/lib/utils/constants.ts
@@ -1,7 +1,11 @@
 import process from 'node:process';
+import { lazy } from '@discordjs/util';
 import { APIVersion } from 'discord-api-types/v10';
-import { Agent } from 'undici';
 import type { RESTOptions } from '../REST.js';
+
+const getUndiciRequest = lazy(async () => {
+	return import('../../strategies/undiciRequest.js');
+});
 
 export const DefaultUserAgent =
 	`DiscordBot (https://discord.js.org, [VI]{{inject}}[/VI])` as `DiscordBot (https://discord.js.org, ${string})`;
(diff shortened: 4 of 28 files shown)


packages/proxy/__tests__/proxyRequests.test.ts
```diff
@@ -55,6 +55,8 @@ test('simple GET', async () => {
 	const res = await supertest(server).get('/api/v10/simpleGet');
 	const headers = res.headers as Record<string, string>;
 
+	console.log(headers);
```
didinele: leftover
```suggestion
```

packages/rest/src/lib/REST.ts
```diff
@@ -11,7 +11,7 @@ import {
 	type RequestData,
 	type RouteLike,
 } from './RequestManager.js';
-import type { IHandler } from './handlers/IHandler.js';
+import type { IHandler } from './handlers/Handler.interface.js';
```
didinele, line 14: p sure I remember Crawl saying he's not a fan of doing `.interface.ts`

didinele (changes requested):

Jiralite (dismissed):
Make sure you update the respective README.md files about the Node.js bump!

packages/rest/src/lib/REST.ts
iCrawl: Yes I have to concur here, this pattern was only really used in the docgen, and that code is anything but what we should move forward with.

ckohen: using .d.ts instead!

.github/workflows/documentation.yml
```diff
@@ -95,7 +95,7 @@ jobs:
       - name: Install node.js v16
```
jaw0r3k, line 95:
```suggestion
      - name: Install node.js v18
```

vladfrangu (changes requested):
Also what [jaw0r3k](https://github.com/jaw0r3k) said

packages/rest/src/lib/global/fetch.d.ts
```diff
@@ -0,0 +1,5 @@
+import type * as undici from 'undici';
```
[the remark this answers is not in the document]
ckohen: I tried it and it worked fine for me (cause I was worried it would be a problem). I suppose we can mark it as a normal dep and it can either get shook away or simply not actually imported at runtime.

packages/rest/src/lib/utils/constants.ts
```diff
@@ -12,13 +11,7 @@ export const DefaultUserAgent =
...
 
 export const DefaultRestOptions = {
-	get agent() {
-		return new Agent({
-			connect: {
-				timeout: 30_000,
-			},
-		});
-	},
+	agent: null,
```
kyranet, line 14: I'm a bit concerned about this change getting forgotten in the future, while we're removing `undici` from `@discordjs/core`, are we also going to remove it from `discord.js`? If not, should we add a note (e.g. documenting this change) or a TODO so we don't forget to re-add this in the future?

ckohen: This change is a no-op for anyone not explicitly importing `DefaultRestOptions`.

https://github.com/discordjs/discord.js/blob/c6ca5a83e7b72613f95c2145606a1330f64ae894/packages/rest/src/lib/RequestManager.ts#L226-L231

It never gets used here. (though it is technically in RequestManager#options)
And furthermore, this default is useless unless someone sets timeout to a value greater than 30 seconds.

kyranet (approved):

packages/rest/src/lib/utils/constants.ts
ckohen: Added to PR description as an example anyways so the changelog will show it

KhafraDev:
This is a horrible, horrible change. I chose undici.request because fetch is not ready for production use, but mostly so people's bots aren't 6x+++ slower in every single http request.

1. It's slow. Web streams are also slow(er than node streams).
    -- https://github.com/nodejs/undici/issues/1203
    -- https://github.com/nodejs/performance/issues/11
    -- https://github.com/nodejs/undici/issues/1203#issuecomment-1100969210
2. It's marked as [experimental](https://nodejs.org/dist/latest-v20.x/docs/api/globals.html#fetch) in node, because of performance. Even if undici is used the same sentiment still applies. Particularly the part where it says stage 1 experimental features are "... not recommended in production environments."

> At the scale of discord bots, this performance difference is irrelevant

This is honestly just absurd, unsubstantiated, and completely wrong, lol. Maybe having 6x less requests "doesn't matter" (spoiler alert: it does, this is exactly the type of change that forces large bot creators to use eris) but you also don't account for increased memory, higher CPU usage, more GC cycles (which are slow), more TCP retransmissions (bug in fetch; packets are being dropped), and *finally*, general [performance concerns](https://github.com/nodejs/performance/issues/9) in web streams...

> This instead is the first step towards cross environment compatability.

At the expense of a majority of discord.js' users... Why is supporting both Deno users and the few bun users more important than the node.js ones?

---

If you really want to make it compatible with other environments, add in a hook system where users can use their own http client. That way platforms that offer faster alternatives (like node & deno) don't have to sacrifice performance for compatibility. Having just the url and request options, and returning a Response-like object would be enough to plug in `undici.request`, and thus cancel out my concerns.

```ts
import { Client } from 'discord.js'
import { request, type RequestInit, Headers } from 'undici'

interface ResponseLike {
  json (): Promise<unknown>
  headers: Headers, // Headers are fast in undici
  status: number
}

async function makeRequest (url: string, init: RequestInit): Promise<ResponseLike> {
  const { statusCode, body, headers } = await request(url, transformOpts(init))

  return {
    status: statusCode,
    headers: new Headers(headers),
    async json () {
      return await body.json()
    }
  }
}

// The default makeRequest would be
// async function makeRequest (url: string, init: RequestInit): Promise<ResponseLike> {
//    return await fetch(url, init) satisfies Promise<ResponseLike> // idk if satisfies is needed
// }

const client = new Client({
  rest: {
    makeRequest
  }
})

```

ckohen:
> Why is supporting both Deno users and the few bun users more important than the node.js ones?

That's not the type of cross-env we're talking about. It is far more important to have a rest implementation that works in edge at this point than supporting deno / bun (though deno is looking ever more appealing).

> This is honestly just absurd, unsubstantiated, and completely wrong, lol. Maybe having 6x less requests "doesn't matter"

The majority of discord.js users have a hard 50/second cap, from discord, nowhere near the throughput of either request or fetch. Even very large bots usually only get bumped to 500/second. (Remember, there's a good portion of users still on v13, which uses node-fetch).

> add in a hook system where users can use their own http client

I do like this, we'd probably export (subpath export so undici isn't imported by default), though you make it sound so much easier than it is.
Notably that entire `resolveBody` util which you handily gloss over in your example with the unexplained `transformOpts()`.
Additionally, `ResponseLike` needs to be a *lot* closer to `Response` than you made it, and quite importantly, `body` will have to be typed as `unknown`, and `clone()` couldn't exist.

> This is a horrible, horrible change.

I appreciate the information on why you chose request (which I did know), and clearly you have a good idea of alternative ways to make this work. I especially appreciate the explanation as to why fetch is still stage 1 since it wasn't anywhere we could find and we figured it wasn't actually unstable since last we heard it wasn't going to land in core until it was stable.

I am happy to take suggestions (part of why the blocked label was added to this PR), but I would appreciate if you were a little less harsh. I understand this isn't what you want to see, but there are better ways to approach that (much of which you do, just after you express your anger)

Also, in case you were wondering, this won't make it to djs v14.

KhafraDev:
When I said that "this" is a horrible change I meant switching to fetch, which is what I've implemented a majority of. It was meant as a critique of my own work, rather than this PR.

I have a draft PR implement the hook but I can't test it out because of <issues>. We can continue in Discord if you want?

ckohen:
The latest commit will need to be squashed before this merges, leaving it separate for ease of review for now.

packages/rest/src/lib/utils/constants.ts
```diff
@@ -27,6 +27,9 @@ export const DefaultRestOptions = {
 	hashSweepInterval: 14_400_000, // 4 Hours
 	hashLifetime: 86_400_000, // 24 Hours
 	handlerSweepInterval: 3_600_000, // 1 Hour
+	async makeRequest(...args) {
+		return fetch(...args);
```
didinele, line 31: Don't we want to default to undici request given its the best performing? It's a matter of deciding which one is worse: "by default, the package does not work on web without specifying this option" or "by default, the performance of this package is significantly worse for the majority of our userbase (nodejs) without specifying this option".

Personally, I feel like the former should be the way to go.

packages/rest/src/strategies/undiciRequest.ts
```diff
@@ -0,0 +1,78 @@
+import { Buffer } from 'node:buffer';
+import { URLSearchParams } from 'node:url';
+import { types } from 'node:util';
+import { type RequestInit, request } from 'undici';
+import type { ResponseLike } from '../index.js';
+
+export type RequestOptions = Exclude<Parameters<typeof request>[1], undefined>;
+
+export async function makeRequest(url: string, init: RequestInit): Promise<ResponseLike> {
+	// The cast is necessary because headers and method are not of the same type, though are request path guarantees they
```
didinele, line 10: `though our`*...? can't tell what you were trying to say here

didinele (changes requested):

packages/rest/src/strategies/undiciRequest.ts
```diff
@@ -0,0 +1,78 @@
...
+		const length = chunks.reduce((a, b) => a + b.length, 0);
+
+		const uint8 = new Uint8Array(length);
+		let lengthUsed = 0;
+
+		return chunks.reduce((a, b) => {
+			a.set(b, lengthUsed);
+			lengthUsed += b.length;
+			return a;
+		}, uint8);
```
kyranet, line 66: I just realised, does `Buffer.concat` not work for this?

kyranet (changes requested):

packages/rest/src/lib/utils/constants.ts
```diff
@@ -34,6 +27,9 @@ export const DefaultRestOptions = {
...
+	},
```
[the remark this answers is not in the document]
KhafraDev: note that this bumps the minimum node version to v16.15.0

packages/rest/src/lib/REST.ts
```diff
@@ -194,6 +202,11 @@ export interface APIRequest {
 	route: string;
 }
 
+export interface ResponseLike
+	extends Pick<Response, 'arrayBuffer' | 'bodyUsed' | 'headers' | 'json' | 'ok' | 'status' | 'text'> {
```
KhafraDev, line 206:
```suggestion
	extends Pick<Response, 'arrayBuffer' | 'headers' | 'json' | 'ok' | 'status' | 'text'> {
```

bodyUsed isn't used anywhere.

packages/rest/src/lib/utils/constants.ts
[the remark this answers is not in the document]
kyranet: This is semver-major, and if we bump the Node.js version, chances are we'll target v18.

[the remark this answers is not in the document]
ckohen: This PR already bumps the node req to 18.12 in all related packages

packages/rest/src/lib/utils/constants.ts
ckohen: I would love to do that, but we simply can't. If we export with a default of undici, then it imports it unnecessarily in edge / web which though it is possible it can get tree-shook, would likely end up throwing.
Also, in all likelihood, we will make undici the default in /next, because that is totally possible while leaving this the way it is

didinele: You can certainly do it by making undici a dynamic `import()` wrapped with `lazy`, and accessing it that way.

packages/rest/src/lib/REST.ts
ckohen: I decided to hoist it since it felt like it  might be useful to consumers that use .raw or the Response event

didinele (approved):

vladfrangu (approved):

packages/proxy/src/util/responseHelpers.ts
```diff
@@ -1,27 +1,29 @@
...
-export async function populateSuccessfulResponse(res: ServerResponse, data: Dispatcher.ResponseData): Promise<void> {
-	res.statusCode = data.statusCode;
+export async function populateSuccessfulResponse(res: ServerResponse, data: ResponseLike): Promise<void> {
+	res.statusCode = data.status;
 
-	for (const header of Object.keys(data.headers)) {
+	for (const [header, value] of data.headers) {
 		// Strip ratelimit headers
-		if (header.startsWith('x-ratelimit')) {
+		if (header.toLowerCase().startsWith('x-ratelimit')) {
```
kyranet, line 17: We should use `/^x-ratelimit/i` instead, it's significantly faster than lowercasing the entire string and then checking if it starts with something.

kyranet (changes requested):

iCrawl (approved):
🚀

kyranet (approved):

PR #9422, opened 2023-04-19
refactor(proxy): rely on auth header instead

didinele:
This more or less acts as a fix for `auth: false` being completely disregarded by the proxy if a client were to use it, which can run us into the nasty 401 bug.

### <samp>🤖 Generated by Copilot at eb414a9</samp>

> _We are the proxy of doom, we forward the headers of fate_
> _We don't need a token of lies, we break the chains of the gate_
> _We handle the errors of wrath, we populate the response of hate_
> _We fix the imports of death, we are the proxy of doom_

### DIFF
--- a/packages/proxy-container/README.md
+++ b/packages/proxy-container/README.md
@@ -23,7 +23,7 @@
 
 Quickly spin up an instance:
 
-`docker run -d --restart unless-stopped --name proxy -p 127.0.0.1:8080:8080 -e DISCORD_TOKEN=abc discordjs/proxy`
+`docker run -d --restart unless-stopped --name proxy -p 127.0.0.1:8080:8080 discordjs/proxy`
 
 Use it:
 
@@ -47,6 +47,9 @@ const rest = new REST({
 	api: 'http://localhost:8080/api',
 });
 ```
+
+**Do note that you should not use the proxy with multiple bots. We cannot gurantee you won't hit ratelimits.
+Webhooks or things that don't include the Authorization header are okay, though!**
 
 ## Links
 
--- a/packages/proxy-container/README.md
+++ b/packages/proxy-container/README.md
@@ -48,7 +48,7 @@ const rest = new REST({
 });
 ```
 
-**Do note that you should not use the proxy with multiple bots. We cannot guarantee you won't hit rate limits.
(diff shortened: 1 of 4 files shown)


iCrawl (approved):

SpaceEEC (approved):

packages/proxy-container/README.md
```diff
@@ -48,6 +48,9 @@ const rest = new REST({
 });
 ```
 
+**Do note that you should not use the proxy with multiple bots. We cannot guarantee you won't hit rate limits.
```
kyranet, line 51: This could be clarified that there's no problem using [multiple] proxies for multiple bots, but rather that there is when the same proxy is reused for multiple bots.
```suggestion
**Do note that you should not use the same proxy with multiple bots. We cannot guarantee you won't hit rate limits.
```

kyranet (changes requested):

packages/proxy-container/README.md
```diff
@@ -48,6 +48,9 @@ const rest = new REST({
 });
 ```
 
+**Do note that you should not use the same proxy with multiple bots. We cannot guarantee you won't hit rate limits.
+Webhooks or things that don't include the Authorization header are okay, though!**
```
ckohen, line 52:
```suggestion
Webhooks with tokens or other requests that don't include the Authorization header are okay, though!**
```

ckohen (changes requested):

ckohen (approved):

packages/proxy/src/util/responseHelpers.ts
```diff
@@ -59,3 +59,24 @@ export function populateAbortErrorResponse(res: ServerResponse): void {
...
 }
+
+/**
+ * Tries to populate a server response from an error object
+ *
+ * @param res - The server response to populate
+ * @param error - The error to check and use
+ * @returns - True if the error is known and the response object was populated, otherwise false
+ */
+export async function populateErrorResponse(res: ServerResponse, error: unknown): Promise<boolean> {
```
vladfrangu, line 70: Why is it async? 👀

didinele:
sorry about that, sorta messed up a rebase.

vladfrangu (approved):

Jiralite (approved):

kyranet (approved):

Issue #8837, opened 2022-11-15
Confusing error messages being produced by lack of null checking in builders/.../Assertions.ts


LeoDog896:
### Which package is this bug report for?

builders

### Issue description

In `Assertions.ts` for the builders package, there is a lack of null checking which causes confusing errors to occur. For example, omitting `TextInputBuilder#setLabel` gives you the following error message:

```bash
ValidationError > s.string
  Expected a string primitive

  Received:
  | undefined
```

This can be fixed by providing more context when the necessary parameters are passed as null:

```ts
export function validateRequiredParameters(customId?: string, style?: TextInputStyle, label?: string) {
        // I am unfamiliar with any code styling guides for discord.js, so this is just an example.
        if (customId === null || customId === undefined) throw Error("CustomID not provided in builder.")
	customIdValidator.parse(customId);
        // and so on ...
}
```

### Package version

14

### Node.js version

18

### Operating system

Ubuntu Linux

### Priority this issue should have

Low (slightly annoying)

### Which partials do you have configured?

Not applicable (subpackage bug), No Partials

### Which gateway intents are you subscribing to?

Not applicable (subpackage bug)

jaw0r3k:
but it needs label 🤔

LeoDog896:
Yes, and so a descriptive error message should be given whenever label is missing, not the one I sent above: (even the stacktrace is confusing)

kyranet:
A label system for shapeshift is tracked at https://github.com/sapphiredev/shapeshift/issues/201

LeoDog896:
Interesting! I'll see if I can work on the feature there.

willuhmjs:
It would probably be easier to move from [@sapphire/shapeshift](https://github.com/sapphiredev/shapeshift) to [zod](https://github.com/colinhacks/zod) to support validation error messages.

For instance, you could change:
```js
export const buttonLabelValidator = s.string
	.lengthGreaterThanOrEqual(1)
	.lengthLessThanOrEqual(80)
	.setValidationEnabled(isValidationEnabled);
```
to
```js
const buttonLabelValidator = z.string()
	.min(1, { message: "Must be 1 or more characters long" } )
	.max(80, { message: "Must be 80 or fewer characters long" } )
```

monbrey:
We moved off zod some time ago due to it's terrible error messages and comparatively poor performance. Very unlikely we'll go back.

imranbarbhuiya:
[...]

PR #9468, opened 2023-04-28
chore(SlashCommandBuilder): improve message when missing required params

wwselleck:
When you try to call `toJSON` on a `SlashCommandBuilder` that has not had a name or description set, or has any options that are missing a name or description, you get the following error:

```
.../node_modules/@sapphire/shapeshift/src/validators/StringValidator.ts:73
                return this.ip(6);
                  ^
Error: Expected a string primitive
    at StringValidator.handle (.../node_modules/@sapphire/shapeshift/src/validators/StringValidator.ts:73:19)
```

In my opinion, this error message is a bit cryptic and makes it difficult to figure out what the issue is. This PR adds explicit checks for missing command/options names/descriptions, and throws a more useful error message.

  - It changes the type of error that is thrown by SlashCommandBuilder::toJSON
  - If any dependents are making checks against the error thrown by SlashCommandBuilder::toJSON, this PR would break that

### DIFF
--- a/packages/builders/src/interactions/slashCommands/Assertions.ts
+++ b/packages/builders/src/interactions/slashCommands/Assertions.ts
@@ -5,6 +5,11 @@ import type { ToAPIApplicationCommandOptions } from './SlashCommandBuilder.js';
 import type { SlashCommandSubcommandBuilder, SlashCommandSubcommandGroupBuilder } from './SlashCommandSubcommands.js';
 import type { ApplicationCommandOptionBase } from './mixins/ApplicationCommandOptionBase.js';
 
+export class MissingRequiredParameterError extends Error {
+	public constructor(missingParameter: string) {
+		super(`Required parameter "${missingParameter}" is missing`);
+	}
+}
 const namePredicate = s.string
 	.lengthGreaterThanOrEqual(1)
 	.lengthLessThanOrEqual(32)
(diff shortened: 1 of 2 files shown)


packages/builders/src/interactions/slashCommands/Assertions.ts
```diff
@@ -35,14 +40,22 @@ export function validateMaxOptionsLength(options: unknown): asserts options is T
 }
 
 export function validateRequiredParameters(
-	name: string,
-	description: string,
+	name: string | undefined,
```
wwselleck, line 43: `name` and `description` are casted as non-undefined on SlashCommandBuilder, but I figure they should be treated as potentially undefined wherever they can be, since they can be undefined.

packages/builders/src/interactions/slashCommands/Assertions.ts
```diff
@@ -5,6 +5,11 @@ import type { ToAPIApplicationCommandOptions } from './SlashCommandBuilder.js';
 import type { SlashCommandSubcommandBuilder, SlashCommandSubcommandGroupBuilder } from './SlashCommandSubcommands.js';
 import type { ApplicationCommandOptionBase } from './mixins/ApplicationCommandOptionBase.js';
 
+export class MissingRequiredParameterError extends Error {
+	public constructor(missingParameter: string) {
+		super(`Required parameter "${missingParameter}" is missing`);
+	}
+}
```
kyranet, line 12: You don't need to create this (and make the function return non-shapeshift errors), you can use the [`MissingPropertyError`](https://github.com/sapphiredev/shapeshift/blob/main/src/lib/errors/MissingPropertyError.ts) error class from `@sapphire/shapeshift`, which is [exported](https://github.com/sapphiredev/shapeshift/blob/main/src/index.ts#L12).

Using the aforementioned class:
```ts
import { MissingPropertyError } from '@sapphire/shapeshift';

throw new MissingPropertyError('name');
```

kyranet (changes requested):

favna:
Duplicate of https://github.com/discordjs/discord.js/issues/8837

Jiralite:
Closing in favour of #10448.

PR #9469, opened 2023-04-28
feat: guide and website issues

jaw0r3k:

Adds labels to make request / reports for missing packages, website or guide

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -22,6 +22,8 @@ body:
         - proxy
         - proxy-container
         - rest
+        - ui
+        - util
         - voice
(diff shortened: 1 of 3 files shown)


TetieWasTaken:
Could you also add the other missing packages?

TetieWasTaken:
I think *all* packages should be there, so that also includes `actions`, `api-extractor-utils`, `scripts` and `docgen`. But maybe someone else has a different opinion? I don't see a reason why not.

jaw0r3k:
> I think _all_ packages should be there, so that also includes `actions`, `api-extractor-utils`, `scripts` and `docgen`. But maybe someone else has a different opinion? I don't see a reason why not.

I dont think an average user will use the maintaining packages
And they even dont have pr label

PR #9485, opened 2023-05-01
chore: "hide" `./defaultWorker` entrypoint with `types: null`

Andarist:
I moved `types` condition to the front. `package.json#exports` are **order-sensitive** - they are always matched from the top to the bottom. When a match is found then it should be used and no further matching should occur. 

Right now, the current setup works in TypeScript but it's considered a bug and it should not be relied upon, see the thread and the comment [here](https://github.com/microsoft/TypeScript/issues/50762#issuecomment-1528318260). For that reason, I would like to fix all popular packages that misconfigured their `exports` this way so the bug can be fixed in TypeScript.

⚠️ this PR focuses solely on fixing "🐛 Used fallback condition" problem but the "🎭 Masquerading as CJS" remains here. You can check how such an error is reported [here](https://arethetypeswrong.github.io/?p=%40discordjs%2Fbuilders%401.6.1)

This is basically the same kind of a PR that was already merged in here: https://github.com/discordjs/discord-api-types/pull/763


packages/ws/package.json
```diff
@@ -18,11 +18,12 @@
...
 	"exports": {
 		".": {
+			"types": "./dist/index.d.ts",
 			"import": "./dist/index.mjs",
-			"require": "./dist/index.js",
-			"types": "./dist/index.d.ts"
+			"require": "./dist/index.js"
 		},
 		"./defaultWorker": {
+			"types": "./dist/defaultWorker.d.ts",
```
didinele, line 26: I'm not sure we want this. This file is not meant to be imported - it's just a worker thread script. We have it in our `exports` to ensure users can still manually specify it, like so:
```ts
new WorkerShardingStrategy({ workerPath: '@discordjs/ws/defaultWorker' });
```

didinele (changes requested):

iCrawl:
I think I just did that on latest main, around 2h ago 😅
A rebase should uncover any changes I didn't do, which this PR might or might not cover.

didinele:
> I think I just did that on latest main, around 2h ago 😅 A rebase should uncover any changes I didn't do, which this PR might or might not cover.

48cab844d4b3e71a08f4076aa5f1cbc0431ab734

packages/ws/package.json
Andarist: Isn't this just a different way of importing though? :p

didinele: No. What I'm saying is, I'd prefer not allowing this:
```ts
import '@discordjs/ws/defaultWorker';
```

though I'm unsure if it currently throws a type error without the `types` field, but I deliberately left it out either way

Andarist: Hm, ye - this case is a little bit convoluted. In theory, `workerPath` wouldn't even have to use `exports` to resolve this thing but I guess that it might be doing that as a convenience. I wonder if it would be capable of loading something outside of `exports`...

> though I'm unsure if it currently throws a type error without the types field, but I deliberately left it out either way

It does throw smth like that today in the [TS playground](https://www.typescriptlang.org/play?moduleResolution=3&target=99&module=1&ts=5.0.4#code/PTAEAEFsHsBMFcA2BTASsgztR8AuBLaAOwC5Qi5kBGANgCgQIYEUyBjaSGIgKwzrr5IAB2gAnXKADeAX1AAzMZ1ABycLHwYOY2H2AB3DMFjJ5AQyS4A6uIDWyMStCNcAT2GYyABgB0ADh8AJgEQoA):
> Cannot find module '@discordjs/ws/defaultWorker' or its corresponding type declarations.(2307)

However, I think that TS playground might be wired differently because the same code **doesn't** error locally and it resolves successfully through the `require` condition and its sibling file (`.d.ts` file next to `.js`, with the same filename)

I feel like you should be able to use `types: null` to "forbid it" at type-level but it currently doesn't work because of the same issue that I'm fixing here ("used fallback condition"). I opened a PR with a failing test case about this in TS: https://github.com/microsoft/TypeScript/pull/54096

To sum it up somehow... this change here doesn't really change how it behaves in TypeScript today. It worked and it still works with it, regardless of the original intention. To fix this today you would have to actively remove `./dist/defaultWorker.d.ts` from disk.

Andarist: a follow up question to this - where do you expect to actually consume this particular entry point? and what's exactly `WorkerShardingStrategy`? what kind of rules govern it?

didinele: Pretty much, our manager class deals with a bunch of "shards" - websocket connections -  those strategies are used to dictate how to spawn them.

By default, we do in-process, but users can set the strategy to the worker_thread one.

This is where we create the worker threads:
https://github.com/discordjs/discord.js/blob/1b7981e4696b8ae74ceffa876fc0a9be7d0b9f93/packages/ws/src/strategies/sharding/WorkerShardingStrategy.ts#L215

You can see the file path resolution process here:
https://github.com/discordjs/discord.js/blob/1b7981e4696b8ae74ceffa876fc0a9be7d0b9f93/packages/ws/src/strategies/sharding/WorkerShardingStrategy.ts#L236-L256

We already default to that `defaultWorker` file, its in the `exports` field just to ensure the user can also explicitly pass it in.

didinele: Yeah, that sounds good.

Andarist:
@didinele adjusted the content and the PR's title, would you mind taking another look?

didinele (approved):

vladfrangu (approved):

kyranet (approved):

PR #9490, opened 2023-05-01
perf(Role): Use iterable and exit early if found

### LINKED ISSUE #3925: Sorting members by highest role position takes extremely long in large servers (10k+ members)

I made a function for the bot that finds members by their username, useful for moderation commands and such. Members with higher roles are prioritized, so the member list is sorted by highest role position to lowest. This process was near instant in v11, but ever since updating to v12 it has been taking significantly longer, sometimes up to **three minutes** in servers with over 10k members. The bot freezes during this process and no event handlers fire.

```js
let startTime = Date.now()
let sortedMembers = message.guild.members.cache.sort(function(x, y){return y.roles.highest.position - x.roles.highest.position})
console.log(Date.now() - startTime)  // 180117 (180 seconds)
console.log(sortedMembers.size)      // 13210
```

**Further details:**

- discord.js version: 12.0.2
- Node.js version: 12.6.1
- Priority this issue should have – please be realistic and elaborate if possible: Somewhat low

JMTK:
This improves performance by not copying the role sorted collection into an array, which is how the spread syntax treats it. This should help with both memory and CPU since it creates a copy of the array. This does not fully address #3925 but would help mitigate some performance issues related to access the position property frequently. 

Some profiling:
```
23000 iterations
Old: 66.575ms
New: 30.693ms
```

```
1 million iterations
Old: 761.495ms
New: 344.201ms
```

```
10 million iterations
Old: 7.964s
New: 3.588s
```

Patch

### DIFF
--- a/packages/discord.js/src/structures/Role.js
+++ b/packages/discord.js/src/structures/Role.js
@@ -177,7 +177,7 @@ class Role extends Base {
    */
   get editable() {
     if (this.managed) return false;
-    const clientMember = this.guild.members.resolve(this.client.user);
+    const clientMember = this.guild.members.me;
     if (!clientMember.permissions.has(PermissionFlagsBits.ManageRoles)) return false;
     return clientMember.roles.highest.comparePositionTo(this) > 0;
   }
@@ -189,7 +189,15 @@ class Role extends Base {
    */
   get position() {
     const sorted = this.guild._sortedRoles();
-    return [...sorted.values()].indexOf(sorted.get(this.id));
+    let positionIndex;
+    for (let roleId of sorted.keys()) {
+      if (roleId === this.id) {
+        positionIndex = positionIndex;
+        break;
+      }
+      positionIndex++;
+    }
+    return positionIndex;
   }
 
   /**
--- a/packages/discord.js/src/structures/Role.js
+++ b/packages/discord.js/src/structures/Role.js
@@ -189,16 +189,13 @@ class Role extends Base {
    */
   get position() {
     const sorted = this.guild._sortedRoles();
-    let positionIndex;
-    let ind = 0;
-    for (let roleId of sorted.keys()) {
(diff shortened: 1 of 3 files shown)


packages/discord.js/src/structures/Role.js
```diff
@@ -189,7 +189,16 @@ class Role extends Base {
...
+    let positionIndex;
+    let ind = 0;
+    for (let roleId of sorted.keys()) {
+      if (roleId === this.id) {
+        positionIndex = ind;
+        break;
+      }
+      ind++;
+    }
+    return positionIndex;
```
kyranet, line 201:
```suggestion
    let index = 0;
    for (const roleId of sorted.keys()) {
      if (roleId === this.id) return index;
      index++;
    }

    return -1;
```

kyranet (changes requested):

packages/discord.js/src/structures/Role.js
```diff
@@ -177,7 +177,7 @@ class Role extends Base {
    */
   get editable() {
     if (this.managed) return false;
-    const clientMember = this.guild.members.resolve(this.client.user);
+    const clientMember = this.guild.members.me;
```
kyranet, line 180: This should be reverted, it's unrelated to the PR.

kyranet (changes requested):
Otherwise LGTM.

packages/discord.js/src/structures/Role.js
JMTK: done

kyranet (approved):

JMTK:
> Couldn't this also be applied to `GuildChannel#position`? Or is it not relevant

Yeah definitely could. In my own use case I had a similar issue as the open issue on the repo, but it could definitely have the same type of problems if someone was looping through the channels

iCrawl (approved):

PR #9493, opened 2023-05-01
perf(Role): linear speed position getter

### LINKED ISSUE #3925: Sorting members by highest role position takes extremely long in large servers (10k+ members)

kyranet:

Alternative to #9490, fixes #3925.

| Iterations |      Old |      New | Difference |
| :--------: | :------: | :------: | :--------: |
|     10_000 | 631.24ms | 174.44ms |     27.63% |
|    100_000 |    8.01s |    1.89s |     23.59% |

Basically skips the `sorted` call, which is an algorithm with an O(n * log(n)) cost, as well as the indexOf, which is a O(n) call. With this, not only it's not longer allocating a Collection nor an Array (making the only allocation be the iterator used in reduce), but the performance also goes down to a single O(n).

### <samp>🤖 Generated by Copilot at 8b28bc4</samp>

> _`position` getter_
> _improved with reduce function_
> _snowflakes in winter_


jaw0r3k:
Can you make the same for channels?

kyranet:
Yes, in a follow-up PR, I wanted to push this one first to get reviews in case the code can be improved any further.

jaw0r3k:
> Yes, in a follow-up PR, I wanted to push this one first to get reviews in case the code can be improved any further.

Idk if using simply for loop will be a little bit faster, as not calling the function everytime
Except that it looks good

packages/discord.js/src/structures/Role.js
```diff
@@ -188,8 +188,14 @@ class Role extends Base {
    * @readonly
    */
   get position() {
-    const sorted = this.guild._sortedRoles();
```
vladfrangu, line 191: Doesn't this mean we no longer need _sortedRoles? 👀 🧹

jaw0r3k: It is used in setting position

kyranet: We still do:

https://github.com/discordjs/discord.js/blob/b327f4925ff557e0aa8273d7f017aa616226ca06/packages/discord.js/src/managers/RoleManager.js#L252

iCrawl (approved):
🐎

SpaceEEC (approved):

vladfrangu (approved):

PR #9497, opened 2023-05-01
perf(Channel): linear speed position getter

kyranet:

Similar to #9493, but for channels.

|   Runs |                                        Old |                                           New |                             Difference |
| :----: | :----------------------------------------: | :-------------------------------------------: | :------------------------------------: |
|  10000 |  83.95ms (m: 3.772μs M: 1.06ms ~: 6.902μs) |  31.84ms (m: 0.838μs M: 327.167μs ~: 1.801μs) | 37.93% (m: 22.22% M: 30.86% ~: 26.09%) |
| 100000 | 708.62ms (m: 3.702μs M: 0.46ms ~: 5.764μs) | 301.75ms (m: 0.489μs M:  38.975μs ~: 1.767μs) | 42.58% (m: 13.21% M:  8.47% ~: 30.66%) |

<details>
<summary>Labels</summary>

- `m`: minimum time per run
- `M`: maximum time per run
- `~`: mean time per run

</details>

Basically skips the `sorted` call, which is an algorithm with an O(n * log(n)) cost, a filter O(n), as well as the indexOf, which is another O(n) call. With this, not only it's not longer allocating 2 Collections nor an Array (making the only allocation be the iterator used in reduce), but the performance also goes down to a single O(n).

Also removed a repeated `indexOf` call for the target channel on every iteration, providing an extra performance boost. This code should see a performance boost between x4 and x5.

I also added `GuildForum` as a compatible type, and renamed variables to be more understandable of the actual behaviour they see.

### <samp>🤖 Generated by Copilot at e0c74a8</samp>

> _We're sailing on the Discord sea, with channels of all kinds_
> _We've got to sort them by their type, and keep them well aligned_
> _So heave away, me hearties, heave away with all your might_
> _We'll use the `ChannelType` enum to get the order right_


packages/discord.js/src/structures/GuildChannel.js
```diff
@@ -145,8 +147,20 @@ class GuildChannel extends BaseChannel {
...
    */
   get position() {
-    const sorted = this.guild._sortedChannels(this);
-    return [...sorted.values()].indexOf(sorted.get(this.id));
+    const selfIsCategory = this.type === ChannelType.GuildCategory;
+    const types = getSortableGroupTypes(this.type);
+
+    let count = 0;
+    for (const channel of this.guild.channels.cache.values()) {
+      if (!types.includes(channel.type) || !(selfIsCategory || channel.parentId === this.parentId)) continue;
```
jaw0r3k, line 155:
```suggestion
      if (!(types.includes(channel.type) && (selfIsCategory || channel.parentId === this.parentId))) continue;
```

well a little bit better

almeidx: IMO this reads slightly better, but the current code is fine as well

```suggestion
      if (!types.includes(channel.type) || (!selfIsCategory && channel.parentId !== this.parentId)) continue;
```

jaw0r3k: not too sure about the performance of yours
I have tested mine and it it is 25% better

kyranet: I think you may need to do more runs, 25% is an impressive performance gap after all. I'll also see if I can open source the benchmark suite I'm using, which uses `Histogram` and `performance.timerify`.

jaw0r3k: Well only the if is 25% better lol
I am not speaking of the whole function
> btw nice that you will publish it

Syjalo: https://github.com/discordjs/discord.js/pull/8393#discussion_r933528705

jaw0r3k: okey then take almeidx one

iCrawl (approved):

SpaceEEC (approved):

vladfrangu (approved):

packages/discord.js/src/structures/GuildChannel.js
kyranet: I ran the benchmarks, it seems @almeidx's is the fastest after all, testing with my support guild's channels (38 in total with multiple categories):

image

However, it's important to note that this performance varies from the case-to-case.

PR #9506, opened 2023-05-03
feat: add silent option to ShardingManager

MovementGH:

Currently discord.js does not give the option to users to fork shards silently. This PR adds that option, setting the default as "false" to maintain current behavior if users do not explicitly set it to true.


Idris1401:
I think you'll need to add types here too:
https://github.com/discordjs/discord.js/blob/main/packages/discord.js/typings/index.d.ts#L6106-6114

MovementGH:
I updated the type that you linked, let me know anything else. This is my first contribution so i'm not sure exactly where to look for everything or how you do things.

MovementGH:
Updated

vladfrangu (approved):

SpaceEEC (approved):

kyranet (approved):

PR #9517, opened 2023-05-05
fix(WebSocketShard): wait a little before reconnecting

### LINKED ISSUE #9510: Infinite loop of WebSocket reconnection

ws

### Issue description

This can be reproduced with the following steps:  
1. Run the code sample.
2. Disconnect the execution environment from the Internet.
3. Observe the output log.(WebSocket tries to reconnect with unusual frequency.)
  
When I observed it, I was getting an infinite loop like this:  
1. An exception occurs in the try syntax of `WebSocketShard#waitForEvent()`.
2. `WebSocketShard#destroy()` is called within the catch syntax.
5. `WebSocketShard#internalConnect()` is called in `destroy()`.
6. `WebSocketShard#waitForEvent()` is called in `internalConnect()`.
  
This bug consumes system resources and brings the system down.

### Code sample

```typescript
import { Client, Events } from 'discord.js';

const client = new Client({ intents: ['Guilds'] });

client.on(Events.Debug, (message) => console.log(message));

client.login('DISCORD_TOKEN')
  .catch(console.error);
```

### Versions

- discord.js 14.10.2
- Node.js v18.16.0

### Issue priority

Not applicable

Guilds

14.11.0-dev.1683201849-51d53f5.0

didinele:
Closes #9510 

### DIFF
--- a/packages/ws/src/ws/WebSocketShard.ts
+++ b/packages/ws/src/ws/WebSocketShard.ts
@@ -259,6 +259,9 @@ export class WebSocketShard extends AsyncEventEmitter<WebSocketShardEventsMap> {
 		this.#status = WebSocketShardStatus.Idle;
 
 		if (options.recover !== undefined) {
+			// There's cases (like no internet connection) where we immediately fail to connect,
+			// causing a very fast and draining reconnection loop.
+			await sleep(500);
 			return this.internalConnect();
 		}
 	}
--- a/packages/ws/__tests__/ws/WebSocketManager.test.ts
+++ b/packages/ws/__tests__/ws/WebSocketManager.test.ts
@@ -229,7 +229,6 @@ test('strategies', async () => {
 	await manager.destroy(destroyOptions);
 	expect(strategy.destroy).toHaveBeenCalledWith(destroyOptions);
 
-	// eslint-disable-next-line id-length
 	const send: GatewaySendPayload = {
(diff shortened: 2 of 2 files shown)


Jiralite (approved):

packages/ws/src/ws/WebSocketShard.ts
```diff
@@ -259,6 +259,9 @@ export class WebSocketShard extends AsyncEventEmitter<WebSocketShardEventsMap> {
 		this.#status = WebSocketShardStatus.Idle;
 
 		if (options.recover !== undefined) {
+			// There's cases (like no internet connection) where we immediately fail to connect,
+			// causing a very fast and draining reconnection loop.
+			await sleep(500);
```
vladfrangu, line 264: I'd say let's add a debug message for this so people aren't left in the dark, but it's optional from my pov

vladfrangu (approved):

didinele:
> It would be nice to be able to set the reconnection time as an option.😄

I don't see any case where that could ever be helpful

packages/ws/src/ws/WebSocketShard.ts
didinele: yeah not fond, esp. given it's consistent and not all that long.

kyranet (approved):

PR #9519, opened 2023-05-05
fix(WebSocketManager): await WebSocket destroy

Mogyuchi:
Fixed to wait for WebSocket destroy to complete.
This is because ['@discordjs/ws' destroy](https://github.com/discordjs/discord.js/blob/3665ad65d0bd25c067fcbbc90ad12acf01896a75/packages/ws/src/ws/WebSocketManager.ts#L310-L312) returns Awaitable.


TetieWasTaken:
This looks like a breaking change to me, could you add the correct versioning classification?

Syjalo:
> Although technically a fix it could lead to unexpected race conditions.

Well, it's already not awaited in the library

Qjuh:
Awaiting the `WebsocketManager#destroy()` here isn’t needed. Before the switch to using `@discordjs/ws` the disconnect of the Websocket connection wasn’t awaited too, it just wasn’t that apparent because there were no async functions involved.

However since a destroyed Client should not be used for anything else afterwards (since it‘s destroyed) I don’t see what race conditions would be solved by awaiting this. Especially since this is a breaking change.

didinele:
What exactly does this fix? Did you actually run into a case where this caused issues?

femshima:
A possible usecase here.

To ensure that bot goes offline and also the process stops after sending SIGINT, it is necessary to handle SIGINT and call `process.exit();`(in case uncancelled `setTimeout` or something like that prevents the process from stopping).

However, in the current implementation, calling `process.exit()` immediately after calling `client.destroy()` leaves the bot shown online on the client.
```js
process.on('SIGINT', () => {
  client.destroy();
  process.exit();
})
```

If `client.destroy()` returns `Promise` that waits until the disconnecting completes, it will be possible to ensure process stops after bot's websocket is normally closed.
```js
process.on('SIGINT', async () => {
  await client.destroy();
  process.exit();
})
```

Mogyuchi:
[My bot did exactly that](https://github.com/kazukazu123123/pow/blob/d1311ef26e9c8bba102523728aca93524634afc0/src/index.ts#L140-L153) and worked as expected until 14.9. However, when I updated to 14.10.2, the bot would not go offline or exit VoiceChannel instantly.

vladfrangu (approved):

kyranet (approved):

SpaceEEC (approved):

PR #9533, opened 2023-05-07
chore(deps): pin `@discordjs/ws` version in discord.js

SuperchupuDev:
#9416 pinned the version of `/rest` in discord.js but `/ws` also had the node version bump and it's being used in discord.js, which is a breaking change

image

image


jaw0r3k (approved):

iCrawl:
A breaking change would be released as 0.9.x which would get updated with ^

vladfrangu:
> A breaking change would be released as 0.9.x which would get updated with ^

Except it won't because 0ver

iCrawl (approved):

PR #9535, opened 2023-05-07
types(Message): mark author as nullable

ItzDerock:

Currently, the discord.js docs site shows [`<Message>.author` as `?User`](https://discord.js.org/#/docs/discord.js/main/class/Message?scrollTo=author), but the TypeScript types do not reflect this. 

Just ran into an instance where not having this field marked as nullable caused a runtime error that could have been prevented.

### DIFF
--- a/packages/discord.js/typings/index.d.ts
+++ b/packages/discord.js/typings/index.d.ts
@@ -1936,7 +1936,7 @@ export class Message<InGuild extends boolean = boolean> extends Base {
   public activity: MessageActivity | null;
   public applicationId: Snowflake | null;
   public attachments: Collection<Snowflake, Attachment>;
-  public author: User;
+  public author: User | null;
   public get bulkDeletable(): boolean;
   public get channel(): If<InGuild, GuildTextBasedChannel, TextBasedChannel>;
   public channelId: Snowflake;


jaw0r3k:
Wasn't it when message is partial?

ItzDerock:
I think it does have something to do with partial messages.
but somewhere there's a type issue as someone received the following Message object: 
```js
<ref *12> Message {
    channelId: '816665532157132820',
    guildId: '692128535090823319',
    id: '1104772260918726697',
    createdTimestamp: 1683468632679,
    system: null,
    type: null,
    content: null,
    author: null,
    pinned: null,
    tts: null,
    nonce: null,
    embeds: [],
    components: [],
    attachments: Collection(0) [Map] {},
    stickers: Collection(0) [Map] {},
    position: null,
    roleSubscriptionData: null,
    editedTimestamp: null,
    reactions: ReactionManager { message: [Circular *12] },
    mentions: MessageMentions {
      everyone: false,
      users: Collection(0) [Map] {},
      roles: Collection(0) [Map] {},
      _members: null,
      _channels: null,
      _parsedUsers: null,
      crosspostedChannels: Collection(0) [Map] {},
      repliedUser: null
    },
    webhookId: null,
    groupActivityApplication: null,
    applicationId: null,
    activity: null,
    flags: MessageFlagsBitField { bitfield: 0 },
    reference: null,
    interaction: null
  }
```
and that caused issues as my package wasn't expecting `<Message>.author` to be null.

Jiralite:
A partial message is defined as such:

https://github.com/discordjs/discord.js/blob/7ce6bb99e9fa593e2dfff5695b0f24e71de179bf/packages/discord.js/src/structures/Message.js#L407-L414

Per this, your message above is partial. Now, I don't know what you are doing, but if you received a runtime error, then it really, really looks like you did not check if this message was partial in the first place.

ItzDerock:
I see, I was unaware that was something I needed to check.

Jiralite:
> Currently, the discord.js docs site shows [`<Message>.author` as `?User`](https://discord.js.org/#/docs/discord.js/main/class/Message?scrollTo=author), but the TypeScript types do not reflect this.

They do, once a message is known to be partial:

https://github.com/discordjs/discord.js/blob/7ce6bb99e9fa593e2dfff5695b0f24e71de179bf/packages/discord.js/typings/index.d.ts#L6011-L6012

Jiralite:
> I see, I was unaware that was something I needed to check.

That's actually a post check because TypeScript would yell before that:

image


In a message update event, you may possibly receive an uncached message, so the `author` is `User | null`. TypeScript should fail to compile if you're relying on this to always be present. I'm not sure what you're doing, but you're absolutely welcome to join the [Discord server](https://discord.com/invite/djs) for guidance.

PR #9556, opened 2023-05-09
Fix: djs package button + redundant divs

brynpttrsn:
This brings djs package button inline with buttonVariants and removes redundant divs.


Jiralite:
Can the same fix be done for the version selection, or is that unrelated?

brynpttrsn:
> Can the same fix be done for the version selection, or is that unrelated?

Different issue but similar enough to be addressed here. :)

iCrawl (approved):

iCrawl (approved):

PR #9565, opened 2023-05-12
fix(AutocompleteInteraction): Prevent snake casing of locales

Jiralite:
The `jsonTransformer()` was converting locales like `zh-TW` to `zh_TW` resulting in an API error being returned.

### DIFF
--- a/packages/discord.js/src/structures/AutocompleteInteraction.js
+++ b/packages/discord.js/src/structures/AutocompleteInteraction.js
@@ -87,9 +87,9 @@ class AutocompleteInteraction extends BaseInteraction {
       body: {
         type: InteractionResponseType.ApplicationCommandAutocompleteResult,
         data: {
-          choices: options.map(option => ({
+          choices: options.map(({ nameLocalizations, ...option }) => ({
             ...this.client.options.jsonTransformer(option),
-            name_localizations: option.nameLocalizations,
+            name_localizations: nameLocalizations,
           })),
         },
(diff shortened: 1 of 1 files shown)


vladfrangu (approved):

packages/discord.js/src/structures/AutocompleteInteraction.js
```diff
@@ -86,7 +86,12 @@ class AutocompleteInteraction extends BaseInteraction {
     await this.client.rest.post(Routes.interactionCallback(this.id, this.token), {
       body: {
         type: InteractionResponseType.ApplicationCommandAutocompleteResult,
-        data: { choices: this.client.options.jsonTransformer(options) },
+        data: {
+          choices: options.map(option => ({
+            ...this.client.options.jsonTransformer(option),
+            name_localizations: option.nameLocalizations,
+          })),
```
kyranet, line 93: It's best to avoid transforming `option.nameLocalizations` to reduce overhead in this call:
```suggestion
          choices: options.map(({ nameLocalizations, ...option }) => ({
            ...this.client.options.jsonTransformer(option),
            name_localizations: nameLocalizations,
          })),
```

Alternatively, we can just transform the fields that are objects, rather than the top-level object.

kyranet (changes requested):

iCrawl (approved):

kyranet (approved):

PR #9581, opened 2023-05-18
refactor(collection): reduce `reduce`'s code

kyranet:

The code was checking an `if` condition on every iteration when `initialValue` was not provided, which leads to a performance impact, furthermore, the code is duplicated. The new code is a lot shorter and avoids the conditional branch, bringing in more performance.

I also made `T` default to `V` so `collection.reduce((acc, v) => acc + v)` becomes valid (`acc`'s type is `V` unless defined otherwise).


ckohen (approved):
🧹

vladfrangu (approved):

SpaceEEC (approved):

Issue #9579, opened 2023-05-17
Documentation for ClientOptions#sweepers is incorrect


ImRodry:
### Which package is this bug report for?

discord.js

### Issue description

The documentation for ClientOptions#sweepers mentions that the default value for this option is an empty object, which is not true. The default can be found below
https://github.com/discordjs/discord.js/blob/5d6eed64140029837043cf537033b97c40f39607/packages/discord.js/src/util/Options.js#L183-L199

### Versions

Djs v14.10.0

### Issue priority

Low (slightly annoying)

### Which partials do you have configured?

Not applicable

### Which gateway intents are you subscribing to?

Not applicable

Ryoh827:
I'd like to try this issue.

ImRodry:
@Jiralite good first issue? Lol

PR #9591, opened 2023-05-20
docs(ClientOptions): change default value of sweepers in docs

Ryoh827:

As described in #9579, the default value for `ClientOptions#sweepers` is `{threads: {interval: 3600, lifetime: 14400}}`, not `{}`.
So I have changed the documentation as follows

before
```
@property {SweeperOptions} [sweepers={}] Options for cache sweeping
```

After
```
@property {SweeperOptions} [sweepers=this.DefaultSweeperSettings] Options for cache sweeping
```


jaw0r3k:
Um worst way to do it ever
I would make a new typedef or smth but not like that

packages/discord.js/src/util/Options.js
```diff
@@ -35,7 +35,7 @@ const { version } = require('../../package.json');
  * @property {IntentsResolvable} intents Intents to enable for this connection
  * @property {number} [waitGuildTimeout=15_000] Time in milliseconds that clients with the
  * {@link GatewayIntentBits.Guilds} gateway intent should wait for missing guilds to be received before being ready.
- * @property {SweeperOptions} [sweepers={}] Options for cache sweeping
+ * @property {SweeperOptions} [sweepers={threads: {interval: 3600, lifetime: 14400}}] Options for cache sweeping
```
jaw0r3k, line 38:
```suggestion
 * @property {SweeperOptions} [sweepers=DefaultSweeperSettings] Options for cache sweeping=
```

the shortest way

Ryoh827: Sorry.

Thank you for the suggestion.
I will modify it to your form.

jaw0r3k: Btw you can just commit suggestion on github site

Ryoh827: I've modified it by adding an additional 'this.' What do you think?

I used the following code and others as reference
https://github.com/discordjs/discord.js/blob/main/packages/discord.js/src/sharding/ShardingManager.js#L177

Sorry if this is wrong.

thehairy: > the shortest way, but i am not too sure if its right

If that links properly to what it returns in the docs, then it's alright.
If not I think the original way is appropiate.

jaw0r3k: Nah docs shows just `default` so lnks wony work there

jaw0r3k: But this long syntax looks horrible in docs

Jiralite (approved):

kyranet (approved):

vladfrangu (approved):

PR #9612, opened 2023-05-29
types(MessageManager): Allow comparison of messages again

### LINKED ISSUE #8545: MessageManager#fetch has signatures, but none of those signatures are compatible with each other.

discord.js

### Issue description

On v14.3.0, when you fetch the messages of a channel through `message.channel.messages.fetch()` without narrowing down the type of message.channel you get a Collection of Message<true> or Message<false>. Upon trying to call .filter on this collection, you will get the following error:
This expression is not callable.
  Each member of the union type '{ <K2 extends string>(fn: (value: Message<true>, key: string, collection: Collection<string, Message<true>>) => key is K2): Collection<K2, Message<true>>; <V2 extends Message<...>>(fn: (value: Message<...>, key: string, collection: Collection<...>) => value is V2): Collection<...>; (fn: (value: Message<...>, key: st...' has signatures, but none of those signatures are compatible with each other. ts(2349)

This is because, since the message manager can come from multiple types of channels, the boolean type parameter can be true or false, and those two are not compatible with each other, thus TS cannot infer the type of the resulting filtered Collection

### Code sample

```typescript
client.on("message", async message => {
    (await message.channel.messages.fetch()).filter(m => m.author.id === message.author.id)
})
```

### Package version

14.3.0

### Node.js version

18.7.0

### Operating system

Windows 11

_No response_

Jiralite:

Resolves #8545 by loosening the strict message types for direct messages. Instead of `false`, they are now `boolean`, which is the prior behaviour. This means `Message<true>` and `Message<boolean>` are now comparable with each other in the overloads.

In addition, `DMMessageManager` and `GuildMessageManager` have been created. `crosspost()` cannot be used in a direct message channel, hence the need for the separation.


jaw0r3k:
isn't crosspost only working on announcement channel?

jaw0r3k (approved):
Looks little bit unnecessary
but its working then fine

vladfrangu (approved):
I feel like the creation of the JS classes to aid in TS strictness isn't _needed_ but LGTM I suppose

iCrawl (approved):

SpaceEEC (approved):

PR #9613, opened 2023-05-29
feat: add new markdown formatters

### LINKED ISSUE #9607: Support new markdown features

formatters

### Feature

Discord added headers and lists, so we can support them in formatters:
https://support.discord.com/hc/en-us/articles/210298617-Markdown-Text-101-Chat-Formatting-Bold-Italic-Underline-

Add functions in formatters like inlineCode, bold or italic

I can prepare PR, but I'm wondering about the function names:
- bigHeader
- header
- smallHeader
- listOption

We can also add support for indenting list, maybe by a second argument not required?

### Other context

_No response_

sdanialraza:

Resolves #9607 

### Summary
📝🧪📊

This pull request introduces the `formatters` module to the discord.js library, which provides functions to format content into headings, ordered lists, and unordered lists. It also adds unit tests for these functions in the `formatters.test.ts` file. The `formatters.ts` file exports the `heading`, `orderedList`, and `unorderedList` functions, as well as the `HeadingLevel` enum and the `RecursiveArray` type.

> _To format content with ease_
> _They added some functions and enums_
> _With `heading` and `listCallback`_
> _And `orderedList` and `unorderedList`_
> _They wrote some unit tests for them_

### Walkthrough
*  Implement and export `heading`, `orderedList`, and `unorderedList` functions, `HeadingLevel` enum, `RecursiveArray` type, and `listCallback` function in `formatters.ts` ([link](https://github.com/discordjs/discord.js/pull/9613/files?diff=unified&w=0#diff-d4285fce25b5b471efe732dc937d52a20472ece1405e460c72b7c100ce2ad119L376-R457))
* Import and test `heading`, `orderedList`, and `unorderedList` functions and `HeadingLevel` enum in `formatters.test.ts` ([link](https://github.com/discordjs/discord.js/pull/9613/files?diff=unified&w=0#diff-68b629777993c90a2cd96f8e967852f3508703d5128d0e0e785a7d813f853787R13-R14), [link](https://github.com/discordjs/discord.js/pull/9613/files?diff=unified&w=0#diff-68b629777993c90a2cd96f8e967852f3508703d5128d0e0e785a7d813f853787R20), [link](https://github.com/discordjs/discord.js/pull/9613/files?diff=unified&w=0#diff-68b629777993c90a2cd96f8e967852f3508703d5128d0e0e785a7d813f853787R28), [link](https://github.com/discordjs/discord.js/pull/9613/files?diff=unified&w=0#diff-68b629777993c90a2cd96f8e967852f3508703d5128d0e0e785a7d813f853787R209-R244))


packages/formatters/src/formatters.ts
```diff
@@ -373,6 +373,76 @@ export function messageLink<C extends Snowflake, M extends Snowflake, G extends
...
+}
+
+/**
+ * Formats the elements in the array to an indented ordered list.
+ *
+ * @typeParam C - This is inferred by the supplied content
+ * @param array - The array of elements to list
+ */
+export function indentedOrderedList<C extends string[]>(...array: C) {
+	return array.map((element, index) => ` ${index + 1} ${element}`).join('\n');
```
NotSugden, line 423:
```suggestion
	return array.map((element, index) => ` ${index + 1}. ${element}`).join('\n');
```
by the looks of it you need a . here as well

NotSugden (changes requested):

packages/formatters/src/formatters.ts
```diff
@@ -373,6 +373,76 @@ export function messageLink<C extends Snowflake, M extends Snowflake, G extends
...
+
+/**
+ * Formats the elements in the array to an ordered list.
+ *
+ * @typeParam C - This is inferred by the supplied content
+ * @param array - The array of elements to list
+ */
+export function orderedList<C extends string[]>(...array: C) {
+	return array.map((element, index) => `${index + 1}. ${element}`).join('\n');
+}
```
almeidx, line 414: Don't need to keep track of the index

```suggestion
export function orderedList<C extends string[]>(...items: C) {
	return items.map((item) => `1. ${item}`).join('\n');
}
```

packages/formatters/src/formatters.ts
```diff
@@ -373,6 +373,76 @@ export function messageLink<C extends Snowflake, M extends Snowflake, G extends
...
+
+/**
+ * Formats the elements in the array to an unordered list.
+ *
+ * @typeParam C - This is inferred by the supplied content
+ * @param array - The array of elements to list
+ */
+export function unorderedList<C extends string[]>(...array: C) {
+	return array.map((element) => `- ${element}`).join('\n');
+}
```
almeidx, line 434:
```suggestion
export function unorderedList<C extends string[]>(...items: C) {
	return items.map((item) => `- ${item}`).join('\n');
}
```

almeidx (changes requested):
Why are there 2 functions for each list type? Couldn't you add a parameter for the indentation (number) instead?

almeidx (changes requested):
Why are there 2 functions for each list type? Couldn't you add a parameter for the indentation (number) instead?

sdanialraza:
@almeidx 
> Why are there 2 functions for each list type? Couldn't you add a parameter for the indentation (number) instead?

That's a good question, I'll work on this change, do you think the same should be done for headings? A level parameter that accepts either 1, 2, or 3?

Syjalo:
I think the best API would be providing a multi-dimensional array.
```js
list([
  '1',
  '2',
  [
    '3',
    [
      '4',
    ],
  ],
])
```
What will be formatted to
- 1
- 2
  - 3
    - 4

sdanialraza:
> Do you think the same should be done for headings? A level parameter that accepts either 1, 2, or 3?

Just got done with this, as for the multi-dimensional array, I am not exactly sure how I'd do that, I've tried a few things, didn't seem to work, so some help would be nice, @Syjalo

packages/formatters/src/formatters.ts
```diff
@@ -373,6 +373,82 @@ export function messageLink<C extends Snowflake, M extends Snowflake, G extends
...
+
+/**
+ * Formats the content into a heading level three.
+ *
+ * @typeParam C - This is inferred by the supplied content
+ * @param content - The content to wrap
+ */
+export function heading<C extends string>(content: C, level: 3): `### ${C}`;
+
+export function heading(content: string, level?: 1 | 2 | 3) {
```
Syjalo, line 400: Add `HeadingLevel` enum

packages/formatters/src/formatters.ts
```diff
@@ -373,6 +373,82 @@ export function messageLink<C extends Snowflake, M extends Snowflake, G extends
...
+	if (level === 3) {
+		return `### ${content}`;
+	}
+
+	if (level === 2) {
+		return `## ${content}`;
+	}
+
+	return `# ${content}`;
```
Syjalo, line 409: Use switch case

Syjalo (changes requested):

packages/formatters/src/formatters.ts
```diff
@@ -373,6 +373,87 @@ export function messageLink<C extends Snowflake, M extends Snowflake, G extends
 	return `${guildId === undefined ? channelLink(channelId) : channelLink(channelId, guildId)}/${messageId}`;
 }
 
+enum HeadingLevel {
```
jaw0r3k, line 376:
```suggestion
export enum HeadingLevel {
```

packages/formatters/src/formatters.ts
```diff
@@ -373,6 +373,87 @@ export function messageLink<C extends Snowflake, M extends Snowflake, G extends
...
+export function heading(content: string, level?: HeadingLevel) {
+	switch (level) {
+		case HeadingLevel.Three:
+			return `### ${content}`;
+		case HeadingLevel.Two:
+			return `## ${content} `;
+		default:
+			return `# ${content}`;
+	}
+}
```
NotSugden, line 415: so bit of a weird suggestion but
```suggestion
export function heading(content: string, level?: HeadingLevel) {
	return `${'#'.repeat(level)} ${content}`;
}
```

not really required though.

jaw0r3k: 1. it works for more 
2. not efficient

sdanialraza: Yeah, don't think that's needed

NotSugden: yeah just me being weird

sdanialraza: I see where you're coming from, and it'd be a good idea in some cases, not in this case though

Syjalo:
> I am not exactly sure how I'd do that

Something like this

```ts
type ListElement = Array<string | ListElement>;

function list(elements: ListElement) {
  const callback = (element: ListElement[number], depth: number = 0): string => {
    if (Array.isArray(element)) {
      return element.flatMap((element) => callback(element, depth++)).join('\n');
    }

    return `${'  '.repeat(depth)}- ${element}`;
  }

  return callback(elements);
}
```

packages/formatters/src/formatters.ts
```diff
@@ -373,6 +373,92 @@ export function messageLink<C extends Snowflake, M extends Snowflake, G extends
...
+	if (startNumber < 0) newStartNumber = 1;
+
+	const processElement = (element: RecursiveArray<string>, depth: number = 0): string => {
+		if (Array.isArray(element)) {
+			const results = element.map((element) => processElement(element, depth + 1));
+			return results.join('\n');
+		}
+
+		return `${'  '.repeat(depth - 1)}${newStartNumber}. ${element}`;
+	};
```
Jiralite, line 439: This seems almost identical to the one in `unorderedList()`. Perhaps it can be brought to the outer scope, marked as `@internal`, and deduplicated?

almeidx:
It would be great if you could add tests

packages/formatters/src/formatters.ts
sdanialraza: I was going to put it in a `util` file or folder but formatters doesn't have something like that, can you please explain what you mean by marked as `@internal`?

Jiralite: You could probably create a util folder then.

[`@internal`](https://tsdoc.org/pages/tags/internal) from TSDoc marks the symbol for internal usage only. Example:

https://github.com/discordjs/discord.js/blob/7295a3a94a853086f781e7810b37c77d0c7e5283/packages/rest/src/lib/handlers/Shared.ts#L20-L25

It won't appear on the documentation. It's like JSDoc's [`@private`](https://jsdoc.app/tags-private.html).

sdanialraza: Alright, sounds good, and thank you for the explanation, I understand now ❤️

packages/formatters/src/formatters.ts
```diff
@@ -373,6 +373,88 @@ export function messageLink<C extends Snowflake, M extends Snowflake, G extends
...
+}
+
+export type RecursiveArray<T> = readonly (RecursiveArray<T> | T)[];
+
+/**
+ * Callback function for list formatters.
+ *
+ * @internal
+ */
+const listCallback = (element: RecursiveArray<string>, depth: number = 0, startNumber?: number): string => {
```
Jiralite, line 427: Make this a function definition like the others.

Jiralite (changes requested):

packages/formatters/src/formatters.ts
```diff
@@ -373,6 +373,88 @@ export function messageLink<C extends Snowflake, M extends Snowflake, G extends
...
+}
+
+export type RecursiveArray<T> = readonly (RecursiveArray<T> | T)[];
+
+/**
+ * Callback function for list formatters.
+ *
+ * @internal
+ */
+function listCallback(element: RecursiveArray<string>, startNumber?: number, depth: number = 0): string {
```
Jiralite, line 427:
```suggestion
function listCallback(element: RecursiveArray<string>, startNumber?: number, depth = 0): string {
```

packages/formatters/src/formatters.ts
```diff
@@ -419,26 +419,31 @@ export function heading(content: string, level?: HeadingLevel) {
...
 
+/**
+ * Callback function for list formatters.
+ *
+ * @internal
+ */
+const listCallback = (element: RecursiveArray<string>, depth: number = 0, startNumber?: number): string => {
+	if (Array.isArray(element)) {
+		const results = element.map((element) => listCallback(element, depth + 1, startNumber));
+		return results.join('\n');
```
NotSugden:
```suggestion
		return element.map((element) => listCallback(element, depth + 1, startNumber)).join('\n');
```

packages/formatters/src/formatters.ts
```diff
@@ -373,6 +373,87 @@ export function messageLink<C extends Snowflake, M extends Snowflake, G extends
 	return `${guildId === undefined ? channelLink(channelId) : channelLink(channelId, guildId)}/${messageId}`;
 }
 
+export enum HeadingLevel {
```
Jiralite, line 376: I'd prefer if there were some basic documentation for this. Something like "Heading levels used for expanded markdown" and then "The first heading level", "The second heading level", etc. Otherwise, on the documentation website, it might show up blank. Basic documentation is better than nothing.

packages/formatters/src/formatters.ts
```diff
@@ -373,6 +373,87 @@ export function messageLink<C extends Snowflake, M extends Snowflake, G extends
...
+		case HeadingLevel.Three:
+			return `### ${content}`;
+		case HeadingLevel.Two:
+			return `## ${content}`;
+		default:
+			return `# ${content}`;
+	}
+}
+
+export type RecursiveArray<T> = readonly (RecursiveArray<T> | T)[];
```
Jiralite, line 420: Basic documentation should be added to this, even if it's as simple as "A type that recursively traverses into arrays."

packages/formatters/src/formatters.ts
sdanialraza: I didn't think about that, thank you for letting me know! I've made the changes.

packages/formatters/src/formatters.ts
```diff
@@ -373,6 +373,102 @@ export function messageLink<C extends Snowflake, M extends Snowflake, G extends
...
+ * Formats the elements in the array to an ordered list.
+ *
+ * @param list - The array of elements to list
+ * @param startNumber - The starting number for the list
+ */
+export function orderedList(list: RecursiveArray<string>, startNumber = 1): string {
+	let newStartNumber = startNumber;
+	if (startNumber <= 0) newStartNumber = 1;
+
+	return listCallback(list, newStartNumber);
```
Jiralite, line 460:
```suggestion
	if (startNumber <= 0) newStartNumber = 1;
	return listCallback(list, newStartNumber);
```

Jiralite (approved):

packages/formatters/src/formatters.ts
```diff
@@ -373,6 +373,101 @@ export function messageLink<C extends Snowflake, M extends Snowflake, G extends
...
+/**
+ * Formats the elements in the array to an ordered list.
+ *
+ * @param list - The array of elements to list
+ * @param startNumber - The starting number for the list
+ */
+export function orderedList(list: RecursiveArray<string>, startNumber = 1): string {
+	let newStartNumber = startNumber;
+	if (startNumber <= 0) newStartNumber = 1;
+	return listCallback(list, newStartNumber);
```
almeidx, line 459:
```suggestion
	return listCallback(list, Math.max(newStartNumber, 1));
```

almeidx (approved):

SpaceEEC (approved):

PR #9626, opened 2023-06-06
feat: Add message to send resume event to shard

GrapeColor:
After a shard reconnects, the `Shard#ready` flag remains `false` because the ready event is not fired.
This PR solves this by firing `Events#ShardReady` when `WebSocketShard#Ready` fires.


Qjuh:
This would make the `Client#shardReady` event emit as soon as the Gateway READY payload is received. This is not what the prior implementation did, this event is meant to be emitted when all Guild payloads did arrive or the guildTimeout was reached.

So this is a breaking change that should not be merged.

But I would like to know what issue there actually was that needed this fix in your eyes. Can you please open a Issue here on GitHub with reproducible code that shows what needs fixing? Because in the current implementation the `Client#shardReady` should be emitted once all Guild payloads were received or the timeout to wait for unavailable Guilds ended. Because it’s the same code that does this successfully on first login call that also is responsible for this on reconnect.

GrapeColor:
I understood the problem and fixed the change.
However, it seems that the intention of the problem is not conveyed, so I will create an issue.

Qjuh:
Okay, now I see what you‘re getting at. #9474 introduced this issue by listening to `ShardReconnecting` and `ShardDisconnecting` event and setting ready to false in those cases.

Solving this by emitting a `ShardReady` event in places it wasn’t emitted before is still a breaking change, because other code listening to that event will now also be called on resume while it wasn’t before.

I‘ll have a look at this issue later to see how this could be fixed in a non-breaking way.

GrapeColor:
Thank you for showing me your approach.
I'll try it that way.

GrapeColor:
I fixed the change.
Request a review.

GrapeColor:
> That does indeed look good now. Did you test that it actually solves your initial issue too?

I have confirmed that the `ready` flag becomes `true` after a shard resumes.
So the original problem is solved.

vladfrangu (approved):

iCrawl (approved):

Jiralite (approved):

PR #9512, opened 2023-05-04
feat: support new username system

Syjalo:
**Please describe the changes this PR makes and why it should be merged:**

This PR adds support for the new username system

* https://github.com/discord/discord-api-docs/pull/6218

### Summary
🌐🖼️📝

This pull request adds support for the new global name feature for users, which allows them to have a different display name across servers. It deprecates the `tag` property of the `User` class and replaces it with the `displayName` property. It also updates the default avatar logic and the documentation of some properties and methods in the `User`, `GuildMember`, `CDN`, and `@discordjs/rest` modules.

> _Sing, O Muse, of the glorious pull request_
> _That changed the `displayName` of the users_
> _And made them shine like stars in different skies_
> _With global names bestowed by Zeus himself._

### Walkthrough
*  Deprecate `tag` and add `globalName` and `displayName` properties to `User` class ([link](https://github.com/discordjs/discord.js/pull/9512/files?diff=unified&w=0#diff-b7b221539f62dcbc8c29a6cad99d732320243e3bcc5fab3e4fec4383ac4d6bd1L3-R5), [link](https://github.com/discordjs/discord.js/pull/9512/files?diff=unified&w=0#diff-b7b221539f62dcbc8c29a6cad99d732320243e3bcc5fab3e4fec4383ac4d6bd1R11-R12), [link](https://github.com/discordjs/discord.js/pull/9512/files?diff=unified&w=0#diff-b7b221539f62dcbc8c29a6cad99d732320243e3bcc5fab3e4fec4383ac4d6bd1R48-R57), [link](https://github.com/discordjs/discord.js/pull/9512/files?diff=unified&w=0#diff-b7b221539f62dcbc8c29a6cad99d732320243e3bcc5fab3e4fec4383ac4d6bd1L56-R71), [link](https://github.com/discordjs/discord.js/pull/9512/files?diff=unified&w=0#diff-b7b221539f62dcbc8c29a6cad99d732320243e3bcc5fab3e4fec4383ac4d6bd1L157-R173), [link](https://github.com/discordjs/discord.js/pull/9512/files?diff=unified&w=0#diff-b7b221539f62dcbc8c29a6cad99d732320243e3bcc5fab3e4fec4383ac4d6bd1L191-R236))
[...]

PR #9632, opened 2023-06-09
chore: unpin rest, unbump node version req

ckohen:

#9512 added to /rest and used that method within djs. Upon publishing, this would break unless we unpinned the version or manually published a version of /rest with that change.

Looking at the breaking changes that made us pin, the most important was the node version bump, which after the change in scope of #9416 was not actually necessary. I've reverted the node version bump and unpinned /rest and /ws in discord.js.

This is an acknowledgement that there will be breaking changes, caveat being that an overwhelmingly large portion of the userbase will not actually be impacted whatsoever as they'd have to be using relatively very low level apis to be impacted.


vladfrangu (approved):
👀

didinele (approved):

SuperchupuDev (commented):
you forgot to update the /rest and /ws readmes, ~~maybe changing it to clarify what requires 18.12.0 would be nice~~

iCrawl (approved):

PR #9633, opened 2023-06-10
types(REST): Add body type info (for post, delete, patch, put)

waki285:

Currently `RequestData.body` is unknown without BodyInit. This change will allow users of the library to check for the type of body by putting the type of body in the generics.


ckohen:
This doesn't seem helpful to me at all. In 99% of cases typescript will just infer the type (which does nothing else).

It's not casting anything, so it doesn't really do any harm, but if you want to check the type of something you should use the (relatively new) `satisfies` keyword, not this.

iCrawl:
This won't be done for the same reasons the fetch interface doesn't get a generic to "determine" the return type.

PR #9648, opened 2023-06-13
fix(ThreadManager): Fix crash

RedGuy12:

`threads.get(raw.id)` can be `undefined` when `raw` is not a child thread of `parent`.

### DIFF
--- a/packages/discord.js/src/managers/ThreadManager.js
+++ b/packages/discord.js/src/managers/ThreadManager.js
@@ -189,7 +189,10 @@ class ThreadManager extends CachedManager {
 
     // Discord sends the thread id as id in this object
     const threadMembers = rawThreads.members.reduce(
-      (coll, raw) => coll.set(raw.user_id, threads.get(raw.id).members._add(raw)),
+      (coll, raw) => {
+        const thread = threads.get(raw.id);
+        return thread ? coll.set(raw.user_id, thread.members._add(raw)) : coll;
+      },
       new Collection(),
     );
 


Jiralite:
> `threads.get(raw.id)` can be `undefined` when `raw` is not a child thread of `parent`.

This explanation doesn't make sense to me: `raw` is a thread member object. How and why could that be or not be a child _thread_?

Can you share a code example of how this crashes?

RedGuy12:
> This explanation doesn't make sense to me: `raw` is a thread member object. How and why could that be or not be a child _thread_?

Sorry, I mean that `raw` could be a member from a thread that isn't a child of `parent`.

> Can you share a code example of how this crashes?

This crashes when calling `channel.threads.fetchActive()` in a server that has threads in channels other than `channel`.
- `rawThreads.threads` contains threads from the whole server
- We iterate over all threads and insert all *that are children of `parent`* into a new Collection: https://github.com/discordjs/discord.js/blob/main/packages/discord.js/src/managers/ThreadManager.js#L184-188
- `rawThreads.members` contains members of any thread in any channel in the whole server
- We iterate over all members and add them to the thread's MemberManager: https://github.com/discordjs/discord.js/blob/main/packages/discord.js/src/managers/ThreadManager.js#L191-194
- `raw` could be a member from a thread *that isn't a child of `parent`*, in which case `threads.get(raw.id)` returns `undefined`, which is not checked for

This bug was introduced in #9164, but is only reproducible in `@dev` due to #9568.

Jiralite (approved):

vladfrangu (approved):

PR #9652, opened 2023-06-15
chore(guide): update username system references

uhKevinMC:
This PR changes references of `.tag` and `.discriminator` in the guide to alternatives in anticipation of those properties being deprecated.

### DIFF
--- a/apps/guide/src/content/05-additional-info/02-collections.mdx
+++ b/apps/guide/src/content/05-additional-info/02-collections.mdx
@@ -28,8 +28,8 @@ Many of the methods on _`Collection`_ correspond to their namesake in _`Array`_.
 
 ```js
 // Assume we have an array of users and a collection of the same users.
-array.find((u) => u.discriminator === '1000');
-collection.find((u) => u.discriminator === '1000');
+array.find((user) => user.bot === true);
+collection.find((user) => user.bot === true);
 ```
 
 </CH.Code>
@@ -100,7 +100,7 @@ collection.last(2);
 
 // Removes anything that meets the condition from the collection.
 // Sort of like `filter`, but in-place.
-collection.sweep((user) => user.username === 'Bob');
+collection.sweep((user) => user.displayName === 'Bob');
 ```
 
 </CH.Code>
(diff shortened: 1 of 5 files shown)


uhKevinMC:
Should I also make changes to the jsdocs in this PR or should that be separate?

almostSouji:
we are not yet even sure if we want to/need to deprecate  #tag
if we decide to keep it (un-deprecated because apparently bots keep discriminators for now) using it in the context presented in the guide seems valid, yet could maybe benefit from a note on the defaulting

apps/guide/src/content/05-additional-info/02-collections.mdx
```diff
@@ -28,8 +28,8 @@ Many of the methods on _`Collection`_ correspond to their namesake in _`Array`_.
 
 ```js
 // Assume we have an array of users and a collection of the same users.
-array.find((u) => u.discriminator === '1000');
-collection.find((u) => u.discriminator === '1000');
+array.find((user) => user.bot === true);
+collection.find((user) => user.bot === true);
```
NotSugden, line 32:
```suggestion
array.find((user) => user.bot);
collection.find((user) => user.bot);
```

NotSugden (changes requested):

apps/guide/src/content/05-additional-info/02-collections.mdx
jaw0r3k: Idk if looking for bots is a good example
user.username.includes("cat") looks cooler

uhKevinMC: This code chunk was originally to filter for users with the discriminator of `#1000`, so I think @jaw0r3k's suggestion would be better suited for this use case..

Still waiting to see if this PR will be considered necessary since there is a discussion about `.tag` being "un-deprecated".

jaw0r3k: Disciminators wont be used by regular users
So this example should be changed
But leave tag code as bots still uses it

PR #9653, opened 2023-06-15
fix(Action): Use existing recipients if available

### LINKED ISSUE #9624: DMChannel#fetch() does not work on partial channels

discord.js

### Issue description

1. Enable the Channel partial for your client
2. Create a slash command that can work in DMs (I just had mine send a basic success embed)
3. In this slash command, attempt to fetch the DM channel (because it may be partial) with:
```js
if (interaction.channel.partial) await interaction.channel.fetch();
```
4. Log the channel before and after to confirm that it is no longer partial afterwards.
5. Use the slash command in your DM channel with the bot.

Expected behavior:
Bot sends success embed, and in the console you can see that the first time the channel is logged it is partial and the second is not.

Actual behavior:
Code throws error:
```
DiscordAPIError[50035]: Invalid Form Body
0[CHANNEL_RECIPIENT_REQUIRED]: At least one channel recipient is required.
    at handleErrors (/Users/oirnoir/Documents/Projects/Bots/node_modules/@discordjs/rest/dist/index.js:640:13)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async SequentialHandler.runRequest (/Users/oirnoir/Documents/Projects/Bots/node_modules/@discordjs/rest/dist/index.js:1021:23)
    at async SequentialHandler.queueRequest (/Users/oirnoir/Documents/Projects/Bots/node_modules/@discordjs/rest/dist/index.js:862:14)
    at async REST.request (/Users/oirnoir/Documents/Projects/Bots/node_modules/@discordjs/rest/dist/index.js:1387:22)
    at async UserManager.createDM (/Users/oirnoir/Documents/Projects/Bots/node_modules/discord.js/src/managers/UserManager.js:60:18)
```

I believe this is because partial DM channels no longer include the DMChannel#recipientId property, which is used in the DMChannel#fetch() method when it calls createDM. I can work around this issue by instead fetching the channel by its ID:
```js
await client.channels.fetch(interaction.channelId);
```

### Code sample

```typescript
"use strict";

const Discord = require('discord.js');
	partials: [
		Discord.Partials.Channel
	],
	intents: [
		Discord.IntentsBitField.Flags.DirectMessages
	]
});

client.on('interactionCreate', async interaction => {
	if (!interaction.isChatInputCommand()) return;
	console.log(interaction.channel);
	if (interaction.channel.partial) await interaction.channel.fetch(); // This line fails
	console.log(interaction.channel);
	await interaction.reply({content: "Success!"});
});

client.login(process.env.DISCORD_TOKEN);
```

### Versions

- discord.js 14.11.0
- Node.js 18.16.0
- MacOS Ventura 13.4 (Dev env)
- Linux Mint 21 (Prod env)

### Issue priority

Low (slightly annoying)

Channel

DirectMessages

df8b6e9

Jiralite:
This resolves #9624.

The tl;dr of that issue is that when receiving an interaction from a direct message channel with appropriate partials, the `recipients` array seemed to be absent and fetching the channel caused an internal crash.

I found that strange, so I looked further into this. Discord _does_ send a `recipients` array:

```JavaScript
{
  type: 1,
  recipients: [
    {
      username: 'jiralite',
      public_flags: 4210944,
      id: '618976181026422814',
      global_name: null,
      discriminator: '0',
      avatar_decoration: null,
      avatar: '6df091f5ebb7f9b9f1b021834f9888e1'
    }
  ],
  last_message_id: '1118964852661158049',
  id: '1074403651738669266',
  flags: 0
}
```

However, when `getChannel()` is called in `GenericAction`:

https://github.com/discordjs/discord.js/blob/b61e4fb0dcb9e5d43a7626c5a4760f5801e06fae/packages/discord.js/src/client/actions/Action.js#L42

We only look for an author, a user, or otherwise grab the user id, all of which does not exist on this payload. Checking for `data.recipients` fixed this issue and `DMChannel` is instantiated with the received `recipients` array.

<hr>

There is another case where this could fail with the same intents and partials from the issue, and that's from sending an interaction after deleting a message!

When you delete a message, you receive a very simple payload:

```JavaScript
{ id: '1118964852661158049', channel_id: '1074403651738669266' }
```

This gets included in the cache. Unfortunately, upon sending an interaction, discord.js would crash with the same error due to caching:

https://github.com/discordjs/discord.js/blob/b61e4fb0dcb9e5d43a7626c5a4760f5801e06fae/packages/discord.js/src/client/actions/Action.js#L26-L32

`existing` is truthy as the above payload is in the cache. Therefore, `._add()` was not called with the received `recipients` array. To mitigate this, I've removed the `!existing` check on the `if` statement so `_add()` is always called if a relevant partial is specified. I'm not sure this is the best solution, but it works.

<hr>

Finally, `last_message_id` was added to the `getPayload()` call. We receive this as seen from the payload above and as such, makes `DMChannel` no longer partial from interactions.


jaw0r3k (approved):

OIRNOIR (approved):

SpaceEEC (approved):

packages/discord.js/src/client/actions/Action.js
```diff
@@ -25,7 +25,7 @@ class GenericAction {
 
   getPayload(data, manager, id, partialType, cache) {
     const existing = manager.cache.get(id);
-    if (!existing && this.client.options.partials.includes(partialType)) {
+    if (this.client.options.partials.includes(partialType)) {
       return manager._add(data, cache);
     }
     return existing;
```
kyranet, line 31: Without the `existing` check, the line 27 is used only below, it can be inlined, although we can also make it a ternary:
```javascript
return this.client.options.partials.includes(partialType)
  ? manager._add(data, cache)
  : manager.cache.get(id);
```

kyranet (changes requested):

packages/discord.js/src/client/actions/Action.js
Jiralite: Resolved with 88b9dcc877e09ddc5c81fbe4122eb97f39145fed.

vladfrangu (approved):

kyranet (approved):

PR #9660, opened 2023-06-20
refactor(User): remove deprecation warning from tag

ImRodry:

This PR removes the deprecation warning from User#tag, simply because, as Discord stated, bots will continue on the legacy username system for the time being, so it's still needed, as well as with users that haven't completed the migration yet/haven't had accesses to it, thus there is no reasonable way a developer can remove all access to the tag property just yet. I would even argue that there's no need to check for discriminators equal to "0" here and let devs do that check on their end, but that's out of the scope of this PR and something I'd rather not touch on.


ImRodry:
Just to clarify, I definitely think this warning should be added back but in a future version, once the discriminator (and thus the tag) is of no use anymore.

jaw0r3k (approved):

kyranet (approved):

SpaceEEC (approved):

vladfrangu (approved):

PR #9672, opened 2023-06-26
Updated Default Avatar URL

Faiyaz2006:
Fixed error calculateUserDefaultAvatarIndex is not a function at line 172.


Jiralite:
If you are seeing this error, you are not running REST with its respective version.

PR #9673, opened 2023-06-27
feat(User): add `isPomelo` function

WilliamDavidHarrison:
Adds a new function called `isPomelo`, which returns `true` if the user uses the new username system (aka Pomelo). This can be helpful for bots if they want to quickly check if a user is using the new username system or not.

### DIFF
--- a/packages/discord.js/src/structures/User.js
+++ b/packages/discord.js/src/structures/User.js
@@ -224,6 +224,11 @@ class User extends Base {
       : null;
   }
 
+  get isPomelo() {
+    if(this.tag.endsWith("#0")) return true;
+    return false;
+  }
+
   /**
    * The global name of this user, or their username if they don't have one
    * @type {?string}
--- a/packages/discord.js/src/structures/User.js
+++ b/packages/discord.js/src/structures/User.js
@@ -224,8 +224,14 @@ class User extends Base {
       : null;
   }
 
+  /**
+   * Whether the user uses the new username system
+   * <info>Returns true if the user uses the new username system</info>
+   * @type {?boolean}
+   * @readonly
+   */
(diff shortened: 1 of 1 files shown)


Jiralite:
Pretty sure this method should be considered deprecated as soon as it is added. In addition, this getter is very basic—developers can implement this themselves?

packages/discord.js/src/structures/User.js
```diff
@@ -224,6 +224,17 @@ class User extends Base {
...
 
+  /**
+   * Whether the user uses the new username system
+   * <info>Returns true if the user uses the new username system</info>
+   * @type {?boolean}
+   * @readonly
+   */
+  get isPomelo() {
+    if(this.discriminator === "0") return true;
+    return false;
```
megatank58, line 235:
```suggestion
    return this.discriminator === "0";
```

packages/discord.js/src/structures/User.js
```diff
@@ -224,6 +224,17 @@ class User extends Base {
       : null;
   }
 
+  /**
+   * Whether the user uses the new username system
+   * <info>Returns true if the user uses the new username system</info>
```
jaw0r3k, line 229:
```suggestion
```
its already mentioned above

packages/discord.js/src/structures/User.js
```diff
@@ -224,6 +224,17 @@ class User extends Base {
...
+   * @type {?boolean}
+   * @readonly
+   */
+  get isPomelo() {
```
jaw0r3k, line 233: it is not too intuitive 
maybe `is migrated` is better

iCrawl:
Not needed.

PR #9676, opened 2023-06-30
fix: revalidate cached members on GUILD_CREATE

WhatCats:
According to the Discord docs regarding the `GUILD_CREATE` event:
If your bot does not have the `GUILD_PRESENCES` Gateway Intent, or if the guild has over 75k members, **members and presences returned in this event will only contain your bot and users in voice channels**.

This removes the line of code that clears the guild's members cache in `_patch()` (which is called by the `GUILD_CREATE` event handler) to save the cached members in the event that you do not have the `GUILD_PRESENCES` Gateway Intent, or if the guild has over 75k members and Discord sends a `GUILD_CREATE` event for whatever reason.

- typings don't need updating

### DIFF
--- a/packages/discord.js/src/structures/Guild.js
+++ b/packages/discord.js/src/structures/Guild.js
@@ -2,7 +2,7 @@
 
 const { Collection } = require('@discordjs/collection');
 const { makeURLSearchParams } = require('@discordjs/rest');
-const { ChannelType, GuildPremiumTier, Routes, GuildFeature } = require('discord-api-types/v10');
+const { ChannelType, GuildPremiumTier, Routes, GuildFeature, GatewayIntentBits } = require('discord-api-types/v10');
 const AnonymousGuild = require('./AnonymousGuild');
 const GuildAuditLogs = require('./GuildAuditLogs');
 const GuildPreview = require('./GuildPreview');
@@ -25,6 +25,7 @@ const RoleManager = require('../managers/RoleManager');
 const StageInstanceManager = require('../managers/StageInstanceManager');
 const VoiceStateManager = require('../managers/VoiceStateManager');
 const DataResolver = require('../util/DataResolver');
+const Events = require('../util/Events');
 const Status = require('../util/Status');
 const SystemChannelFlagsBitField = require('../util/SystemChannelFlagsBitField');
 const { discordSort } = require('../util/Util');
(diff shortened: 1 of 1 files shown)


WhatCats:
> If a guild goes on an outage, you won't receive updates to their presence or whether they still remain in the guild. This code change introduces potentially invalid guild members, right?

Both invalid GuildMembers and missing GuildMembers is not ideal.
It seems like the one way to avoid this issue completly is by using the `GUILD_PRESENCES` intent for now.

The other question is when is the `GUILD_CREATE` event fired?
Is it just
- on inintial connect after Ready packet (for each Guild)
- when bot joins Guild
- when Guild becomes available again

Or will reconnecting for example also trigger it?

Qjuh:
After full reconnect it will be triggered too, but then you also could have missed members leaving and presences being updated during the reconnect. So the same issue remains. The „fix“ for what you describe as problem here would be to fetch all members in guildCreate event and have a ready event that doesn’t run once but with on on every ready that does that for guilds you need it in. There’s no way for discord.js to validate the cache against what exists on discord other than replacing it completely in those cases.

WhatCats:
> The „fix“ for what you describe as problem here

Yes a "fix" because there is generally nothing else fetching the guild members again, leaving the cache in an invalid state, without there being any obvious signs/warnings of it. This can lead to very weird, random seeming and frusturating behaviour.

Also missed presence updates is not a problem since this problem only exists when there is no `GUILD_PRESENCES` intent, in which case presence updates are not being received anyways.

Qjuh:
And now you introduced a rather breaking change by fetching all members there. Which discord.js never does, it‘s upon the library user to decide whether they want to fetch all members or not. As already stated, the premise of your PR is wrong. If you need all members fetched you can do so yourself by listening to the respective event and fetching them again on a reconnect. It’s *not* the job of the library to do that for you.

WhatCats:
Your right, what I did here does not make much sense.

WhatCats:
If I am seeing this right though the GUILD_CREATE handler
```js
let guild = client.guilds.cache.get(data.id);
  if (guild) {
    if (!guild.available && !data.unavailable) {
      // A newly available guild
      guild._patch(data);
    }
  } else {
    // A new guild
    data.shardId = shard.id;
    guild = client.guilds._add(data);
    if (client.ws.status === Status.Ready) {
      /**
       * Emitted whenever the client joins a guild.
       * @event Client#guildCreate
       * @param {Guild} guild The created guild
       */
      client.emit(Events.GuildCreate, guild);
    }
  }
```
 will not update the guild if a full reconnect occurs and the GUILD_CREATE's are sent again?

WhatCats:
Let's say someone did want to keep the members fetched.
I can't find an event that would tigger when a guild becomes available again (after being marked as unavailable) or for when a full reconnect (not a resume) occurs.

PR #9678, opened 2023-07-01
types(BaseMessageOptions): Add ActionRowBuilder<AnyComponentBuilder>

Tirithel:
Small type change that allows for ActionRowBuilder to be used in a BaseMessage's components.

- I _"know"_ how to update typings and have done so, or typings don't need updating

minor change - bug fix. This resolves an issue where one is unable to use ActionRowBuilder in a message. Example:

```typescript
   // ...
    const buttons = new ActionRowBuilder().addComponents(button1, button2);

    await interaction.reply({
      components: [buttons],
    });
```

The above code would/will fail due to a typing issue.

### DIFF
--- a/packages/discord.js/typings/index.d.ts
+++ b/packages/discord.js/typings/index.d.ts
@@ -5834,6 +5834,7 @@ export interface BaseMessageOptions {
     | JSONEncodable<APIActionRowComponent<APIMessageActionRowComponent>>
     | ActionRowData<MessageActionRowComponentData | MessageActionRowComponentBuilder>
     | APIActionRowComponent<APIMessageActionRowComponent>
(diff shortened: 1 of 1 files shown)


Jiralite:
This is not an issue, you should use `ActionRowBuilder`'s type parameter to choose what builder(s) it holds.

Tirithel:
Isn't type meant to be optional? It [looks to me](https://discordjs.guide/message-components/buttons.html#sending-buttons) like it should be able to send it without a type specified, no?

Qjuh:
But your change would allow it to send TextInputBuilders in a message too, which is invalid. Which is why the types are restricted like they are

PR #9680, opened 2023-07-03
fix(core): fix inconsistencies on `core`

### LINKED ISSUE #9618: Fix `core`'s inconsistencies

core

### Feature

Currently API methods get added to different API classes based on three things:

1. Its topic on the Discord API docs, e.g. `WebhooksAPI#create` is part of the [Webhook](https://discord.com/developers/docs/resources/webhook#create-webhook) topic on the Discord docs, but uses the `/channels` route.
2. Its route on the Discord API, e.g. `ChannelsAPI#getWebhooks` uses the `/channels` route, but is part of the [Webhook](https://discord.com/developers/docs/resources/webhook#get-channel-webhooks) topic on the Discord docs.
3. Convenience, e.g. `ThreadsAPI` really just contains methods which use the Channels API, but having a different class can be more convenient when wording the methods, and as a result the parent class becomes smaller.

The problem is mostly with the first two, because notice how similar the two examples I gave are. They both use the `/channels` route, both appear in the Webhook topic, but somehow end up being in different API classes.

`ThreadsAPI` also has a bit of a problem:

- `ThreadsAPI#get` is literally the same as `ChannelsAPI#get` in terms of code after #8897, the only difference is that `ThreadsAPI#get` has a different return type, which isn't actually guaranteed (it doesn't necessarily return a thread, but a channel instead).
- `ThreadsAPI#create` and `ThreadsAPI#createForumThread` suffer from a problem similar but not exactly identical to the aforementioned one about routes and topics. This time they both appear in the same topic, route, and API class, but the action itself is done on a channel, not a thread, similar to `ChannelsAPI#createInvite` and the rest of the thread-related methods in `ChannelsAPI`. I guess you could say that the topic of the methods are the threads, but the actions themselves are done on channels.

In addition there's a minor problem where the documentation uses multiple different ways to word stuff such as parameters, for example:

- `The <parameter> for ...`
- `The <parameter> to use when ...`

I believe that adding methods to classes based on their route makes more sense.

When it comes to `ThreadsAPI`, I think its existence does make sense, but it still is just the Channels API and could be merged into `ChannelsAPI`.
Other than that, at least `ThreadsAPI#create`, and `ThreadsAPI#createForumThread` should be moved to `ChannelsAPI`, if we are going to choose to group methods by routes instead of topics.
`ThreadsAPI#get` should be completely removed. Or, instead, it may also check whether the channel is actually a thread or not, but I think the library users are the ones who should be doing that.

When it comes to the documentation, I think it should use the more concise phrasing.

_No response_

### Other context

[#9578 (review)](https://github.com/discordjs/discord.js/pull/9578#pullrequestreview-1432883615)

I can make the PR for this, but I wanted to spark a conversation first to agree on a decision.

suneettipirneni:

This PR implements the changes outlined in #9618.

Additionally, I've removed `createForumThread` as it was essentially an alias for `createThread` just with types constrained to forum thread JSON payload options.

Closes #9618

### DIFF
--- a/packages/core/src/api/channel.ts
+++ b/packages/core/src/api/channel.ts
@@ -30,7 +30,13 @@ import {
 	type Snowflake,
 	type RESTPostAPIChannelThreadsJSONBody,
 	type RESTPostAPIChannelThreadsResult,
+	type APIThreadChannel,
+	type RESTPostAPIGuildForumThreadsJSONBody,
 } from 'discord-api-types/v10';
+
+export interface StartForumThreadOptions extends RESTPostAPIGuildForumThreadsJSONBody {
+	message: RESTPostAPIGuildForumThreadsJSONBody['message'] & { files?: RawFile[] };
+}
 
 export class ChannelsAPI {
 	public constructor(private readonly rest: REST) {}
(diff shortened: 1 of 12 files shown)


vladfrangu:
So what's the replacement of `createForumThread` now? Code wise that is

packages/core/src/api/channel.ts
```diff
@@ -456,13 +511,33 @@ export class ChannelsAPI {
...
 
+	/**
+	 * Creates a new webhook
+	 *
+	 * @see {@link https://discord.com/developers/docs/resources/webhook#create-webhook}
+	 * @param channelId - The id of the channel to create the webhook in
+	 * @param body - The data for creating the webhook
+	 * @param options - The options for creating the webhook
+	 */
+	public async create(
```
vladfrangu, line 522: Shouldn't this be createWebhook?

vladfrangu (changes requested):

Jiralite (approved):

kyranet (approved):

PR #9682, opened 2023-07-03
refactor(rest): add web compatibility

suneettipirneni:

Allow /rest to run in edge/web environments.

- `node:` prefixes for *value* imports have been removed in favor of global objects.
- `ResponseLike` now has an additional property `statusText` as `STATUS_CODES` is only in node and `fetch`'s statusText seems to provide equivalent functionality
- Replace `EventEmitter` with `AsyncEventEmitter`.
- `file-type` has been removed due to dependence on node and `magic-bytes.js` is it's replacement
- We now build two entrypoints for both web and node environments. When importing the package in a node env, it will default to using the undici implementation. When imported in an edge/web env it will default to global fetch.


package.json
```diff
@@ -78,5 +78,8 @@
 		"apps/*",
 		"packages/*"
 	],
-	"packageManager": "yarn@3.5.1"
+	"packageManager": "yarn@3.5.1",
+	"dependencies": {
```
vladfrangu, line 82: Remove this from the root 🙏

packages/rest/package.json
```diff
@@ -4,7 +4,7 @@
 	"description": "The REST API for discord.js",
 	"scripts": {
 		"test": "vitest run",
-		"build": "tsup",
+		"build": "tsup --external=../../strategies/undiciRequest.js",
```
vladfrangu, line 7: Can't this be added as an option in the config file instead?

packages/rest/src/lib/RequestManager.ts
```diff
@@ -425,18 +423,9 @@ export class RequestManager extends EventEmitter {
...
-					let contentType = file.contentType;
-					if (!contentType) {
-						const parsedType = (await fileTypeFromBuffer(file.data))?.mime;
-						if (parsedType) {
-							contentType = OverwrittenMimeTypes[parsedType as keyof typeof OverwrittenMimeTypes] ?? parsedType;
-						}
-					}
 
-					formData.append(fileKey, new Blob([file.data], { type: contentType }), file.name);
+				if (typeof Buffer !== 'undefined' && Buffer.isBuffer(file.data) && isAPNG(file.data)) {
```
vladfrangu, line 427: But what ifff the data is Uint8Array/ArrayBuffer? It'll get stringified, which it shouldn't be. You also completely changed the logic here, the content type should be overriden only for APNG, otherwise... I guess we could try defaulting to just application/octet-stream?

packages/rest/src/lib/utils/constants.ts
```diff
@@ -34,7 +48,12 @@ export const DefaultRestOptions = {
 	handlerSweepInterval: 3_600_000, // 1 Hour
 	async makeRequest(...args) {
 		const strategy = await getUndiciRequest();
```
vladfrangu, line 50: Didn't I change this in the branch? o.o

packages/rest/src/lib/utils/utils.ts
```diff
@@ -121,3 +120,14 @@ export async function onRateLimit(manager: RequestManager, rateLimitData: RateLi
...
 	return Number(BigInt(userId) >> 22n) % 6;
 }
+
+/**
+ * Sleeps for a given amount of time.
+ *
+ * @param ms - The amount of time (in milliseconds) to sleep for
+ */
+export async function sleep(ms: number): Promise<void> {
+	void new Promise<number>((resolve) => {
```
vladfrangu, line 130:
```suggestion
	return new Promise<number>((resolve) => {
```

vladfrangu (changes requested):

packages/rest/.eslintrc.json
```diff
@@ -1,3 +1,11 @@
 {
-	"extends": "../../.eslintrc.json"
+	"extends": "../../.eslintrc.json",
+	"rules": {
+		"n/prefer-global/url": "off",
+		"n/prefer-global/url-search-params": "off",
+		"n/prefer-global/buffer": "off",
```
ckohen, line 6: maybe we should leave this one on so it's obvious that we are using node only APIs?

packages/rest/src/lib/REST.ts
```diff
@@ -247,7 +247,7 @@ export interface REST {
 		(<S extends string | symbol>(event?: Exclude<S, keyof RestEvents>) => this);
 }
 
-export class REST extends EventEmitter {
+export class REST extends AsyncEventEmitter {
```
ckohen, line 250: Should get rid of our EE hacks for exposing events and use the much better types of AEE

packages/rest/src/lib/RequestManager.ts
```diff
@@ -182,7 +170,7 @@ export interface RequestManager {
 /**
  * Represents the class that manages handlers for endpoints
  */
-export class RequestManager extends EventEmitter {
+export class RequestManager extends AsyncEventEmitter {
```
ckohen, line 173: Ditto

packages/rest/src/lib/RequestManager.ts
```diff
@@ -269,7 +257,12 @@ export class RequestManager extends EventEmitter {
 
 				// Fire event
 				this.emit(RESTEvents.HashSweep, sweptHashes);
-			}, this.options.hashSweepInterval).unref();
+			}, this.options.hashSweepInterval);
+
+			// Only unref in node envs
+			if (typeof this.hashTimer !== 'number') {
+				this.hashTimer = this.hashTimer.unref();
```
ckohen, line 264: No need to reassign here, just call unref. Almost makes sense to just .unref?.() instead.

packages/rest/src/lib/RequestManager.ts
```diff
@@ -292,7 +285,12 @@ export class RequestManager extends EventEmitter {
 
 				// Fire event
 				this.emit(RESTEvents.HandlerSweep, sweptHandlers);
-			}, this.options.handlerSweepInterval).unref();
+			}, this.options.handlerSweepInterval);
+
+			// Only unref in node envs
+			if (typeof this.handlerTimer !== 'number') {
+				this.handlerTimer = this.handlerTimer.unref();
```
ckohen, line 292: Ditto

packages/rest/src/lib/handlers/Shared.ts
```diff
@@ -135,7 +133,7 @@ export async function handleErrors(
 		}
 
 		// We are out of retries, throw an error
-		throw new HTTPError(status, method, url, requestData);
+		throw new HTTPError(res.status, res.statusText, method, url, requestData);
```
ckohen, line 136: no need to add res. for status here.

packages/rest/src/lib/utils/constants.ts
```diff
@@ -34,7 +48,12 @@ export const DefaultRestOptions = {
...
-		return strategy.makeRequest(...args);
+
+		if (strategy === null) {
+			return fetch(...args);
```
ckohen, line 53: This won't work in node 16. It's probably not possible to reach, but stupider stuff has happened.

packages/rest/tsup.config.ts
```diff
@@ -2,6 +2,6 @@ import { esbuildPluginVersionInjector } from 'esbuild-plugin-version-injector';
 import { createTsupConfig } from '../../tsup.config.js';
 
 export default createTsupConfig({
-	entry: ['src/index.ts', 'src/strategies/*.ts'],
+	entry: ['src/index.ts', 'src/strategies/undiciRequest.ts'],
```
ckohen, line 5: Why are you specifically calling this out?

ckohen (changes requested):

packages/rest/src/lib/RequestManager.ts
ckohen: Also, isAPNG does the previous checks anyways (its types should probably accept any since it thoroughly checks), so they're happening twice.

And actually lettign isAPNG handle that would solve the problem with Uint8Arrays right?

packages/rest/src/lib/errors/HTTPError.ts
```diff
@@ -12,18 +11,19 @@ export class HTTPError extends Error {
...
 	/**
 	 * @param status - The status code of the response
+	 * @param statusText - The status text of the response
 	 * @param method - The method of the request that erred
 	 * @param url - The url of the request that erred
 	 * @param bodyData - The unparsed data for the request that errored
 	 */
 	public constructor(
 		public status: number,
+		public statusText: string,
```
ckohen, line 21:
```suggestion
		statusText: string,
```

This is the message of the base errror, no need to double up.

ckohen (changes requested):

ckohen (approved):

packages/rest/src/lib/utils/constants.ts
```diff
@@ -41,7 +55,12 @@ export const DefaultRestOptions = {
 		}
 
 		const strategy = await getUndiciRequest();
-		return strategy.makeRequest(...args);
+
+		if (strategy === null) {
```
vladfrangu, line 59: again, didn't I change this in the branch? Why is it undone?

packages/rest/src/lib/utils/constants.ts
```diff
@@ -13,7 +16,18 @@ export const DefaultUserAgent =
 /**
  * The default string to append onto the user agent.
  */
-export const DefaultUserAgentAppendix = process.release?.name === 'node' ? `Node.js/${process.version}` : '';
+export const DefaultUserAgentAppendix =
```
vladfrangu, line 19: Should probably implement a method in discordjs/util to get this info

vladfrangu (changes requested):

packages/util/src/functions/userAgent.ts
```diff
@@ -0,0 +1,18 @@
...
+	// Most (if not all) edge environments will have `process` defined. Withing a web browser we'll extract it using `navigator.userAgent`.
+	return typeof process === 'object'
+		? process.release?.name === 'node'
+			? `Node.js/${process.version}`
+			: ''
+		: // @ts-expect-error web env
+		typeof window === 'object'
+		? // @ts-expect-error web env
+		  window.navigator.userAgent
+		: null;
```
vladfrangu, line 17: This should also handle deno and bun for the UA

```suggestion
	return typeof globalThis.process === 'object'
		? globalThis.process.release?.name === 'node'
			? `Node.js/${globalThis.process.version}`
			: ''
		: // @ts-expect-error web env
		typeof globalThis.window === 'object'
		? // @ts-expect-error web env
		  globalThis.window.navigator.userAgent
		: null;
```

vladfrangu (changes requested):

packages/rest/package.json
```diff
@@ -14,19 +14,19 @@
...
-	"module": "./dist/index.mjs",
-	"typings": "./dist/index.d.ts",
+	"types": "./dist/node.d.ts",
 	"exports": {
 		".": {
-			"types": "./dist/index.d.ts",
-			"import": "./dist/index.mjs",
-			"require": "./dist/index.js"
-		},
-		"./*": {
```
ckohen, line 26: do not remove this

ckohen (changes requested):

packages/rest/tsup.config.ts
```diff
@@ -2,6 +2,6 @@ import { esbuildPluginVersionInjector } from 'esbuild-plugin-version-injector';
 import { createTsupConfig } from '../../tsup.config.js';
 
 export default createTsupConfig({
-	entry: ['src/index.ts', 'src/strategies/*.ts'],
```
ckohen, line 5: again, don't remove strategies here

ckohen (changes requested):

packages/rest/tsup.config.ts
```diff
@@ -2,6 +2,12 @@ import { esbuildPluginVersionInjector } from 'esbuild-plugin-version-injector';
...
+	entry: [
+		'src/agnostic.ts',
+		'src/web.ts',
+		'src/strategies/undiciRequest.ts',
```
ckohen, line 8: no need to call out each strategy individually (alos fetchRequest.ts doesn't exist lol)

ckohen (changes requested):

packages/rest/src/agnostic.ts
```diff
@@ -0,0 +1,7 @@
+import { shouldUseGlobalFetchAndWebSocket } from '@discordjs/util';
+import { setDefaultStrategy } from './environment.js';
+import { makeRequest } from './strategies/undiciRequest.js';
```
kyranet, line 3: This is importing Node packages, shouldn't it be lazy loaded if anything?

packages/rest/src/lib/RequestManager.ts
```diff
@@ -425,14 +403,18 @@ export class RequestManager extends EventEmitter {
...
 				// The Blob constructor accepts TypedArray/ArrayBuffer, strings, and Blobs.
-				if (Buffer.isBuffer(file.data)) {
+				if (isBufferLike(file.data)) {
 					// Try to infer the content type from the buffer if one isn't passed
-					const { fileTypeFromBuffer } = await getFileType();
 					let contentType = file.contentType;
+
 					if (!contentType) {
-						const parsedType = (await fileTypeFromBuffer(file.data))?.mime;
+						const parsedType = filetypeinfo(file.data)[0];
```
kyranet, line 411:
```suggestion
						const [parsedType] = filetypeinfo(file.data);
```

packages/rest/src/lib/utils/utils.ts
```diff
@@ -121,3 +120,28 @@ export async function onRateLimit(manager: RequestManager, rateLimitData: RateLi
...
+}
+
+/**
+ * Verifies that a value is a buffer-like object.
+ *
+ * @param value - The value to check
+ */
+export function isBufferLike(value: unknown): value is ArrayBuffer | Buffer | Uint8Array | Uint8ClampedArray {
+	return (
+		(typeof Buffer !== 'undefined' && Buffer.isBuffer(value)) ||
```
kyranet, line 142: One question, do we strictly need it to be a `Buffer`? Is `Uint8Array` (which `Buffer` extends) not enough?

In fact we check this on line 144, so this check is a bit redundant.

kyranet (changes requested):
I'm also not fully convinced about the `setTimeout().unref?.()` since on Deno and Bun, it'll read properties that don't exist, which carries a performance penalty.

Maybe an utility named `setUnrefTimeout()`?

ckohen:
The performance impact on `.unref?.()` is irrelevant, it's called once on instantiation of `REST` (technically whenever sweepers are setup, but that's never going to be frequent)

packages/rest/src/agnostic.ts
vladfrangu: Agnostic is meant for envs like node/deno/bun, all of which can import node apis just fine (and wont necessarily use them!). @ckohen can you check if for CF workers or vercel, the web import is used? 🙏

suneettipirneni: I've verified that *only* the web entry is used for cloudflare workers.

vladfrangu: What about vercel edge?

packages/util/src/functions/userAgent.ts
```diff
@@ -0,0 +1,18 @@
...
+
+/**
+ * Resolves the user agent string for the current environment.
+ */
+export function getUserAgent(): string | null {
+	// Most (if not all) edge environments will have `process` defined. Withing a web browser we'll extract it using `navigator.userAgent`.
+	return typeof globalThis.process === 'object'
+		? globalThis.process.release?.name === 'node'
+			? `Node.js/${globalThis.process.version}`
+			: ''
```
vladfrangu, line 12: This still needs to handle deno/bun please!

vladfrangu (changes requested):

packages/rest/src/agnostic.ts
suneettipirneni: I've verified it selects the web entrypoint for vercel edge functions as well.

vladfrangu (approved):

packages/rest/src/lib/utils/constants.ts
```diff
@@ -33,15 +29,7 @@ export const DefaultRestOptions = {
...
-		const defaultToFetch = shouldUseGlobalFetchAndWebSocket();
-
-		if (defaultToFetch) {
-			// @ts-expect-error We know this is defined based on the check above
-			return globalThis.fetch(...args);
-		}
-
-		const strategy = await getUndiciRequest();
-		return strategy.makeRequest(...args);
+		return getDefaultStrategy()(...args);
```
vladfrangu, line 32: cache this method in a variable

vladfrangu (changes requested):

vladfrangu (approved):

kyranet (changes requested):

packages/rest/src/lib/REST.ts
```diff
@@ -256,9 +241,13 @@ export class REST extends EventEmitter {
 		super();
 		this.cdn = new CDN(options.cdn ?? DefaultRestOptions.cdn);
 		this.requestManager = new RequestManager(options)
+			// @ts-expect-error For some reason ts can't infer these types
```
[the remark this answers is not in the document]
suneettipirneni: This properly infers on the application developers side, for some reason it only has an issue in this specific scenario.

kyranet (approved):

packages/rest/src/lib/REST.ts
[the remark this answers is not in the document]
ckohen: The problem is incomplete inference of the generics on `.bind` (it incorrectly assumes all parameters are of the same type by default for some reason)  and all this event hoisting will be removed in a follow-up PR anyways.

ckohen (approved):

PR #9683, opened 2023-07-03
feat: no-de-no-de, now with extra buns

vladfrangu:

This PR brings in out-of-the-box support for:
- Deno
- Bun
- Cloudflare Workers
- Vercel Functions

and whatever else floats your boat. 

*note:* Bun is currently blocked with this PR due to https://github.com/oven-sh/bun/issues/3392, but 🤞 they get that fixed

### DIFF
--- a/packages/rest/.eslintrc.json
+++ b/packages/rest/.eslintrc.json
@@ -1,3 +1,11 @@
 {
-	"extends": "../../.eslintrc.json"
+	"extends": "../../.eslintrc.json",
+	"rules": {
+		"n/prefer-global/url": "off",
+		"n/prefer-global/url-search-params": "off",
+		"n/prefer-global/buffer": "off",
+		"n/prefer-global/process": "off",
+		"no-restricted-globals": "off",
+		"unicorn/prefer-node-protocol": "off"
+	}
 }
--- a/packages/rest/.eslintrc.json
+++ b/packages/rest/.eslintrc.json
@@ -1,11 +1,11 @@
 {
 	"extends": "../../.eslintrc.json",
 	"rules": {
-		"n/prefer-global/url": "off",
-		"n/prefer-global/url-search-params": "off",
-		"n/prefer-global/buffer": "off",
-		"n/prefer-global/process": "off",
-		"no-restricted-globals": "off",
-		"unicorn/prefer-node-protocol": "off"
+		"n/prefer-global/url": 0,
+		"n/prefer-global/url-search-params": 0,
+		"n/prefer-global/buffer": 0,
+		"n/prefer-global/process": 0,
+		"no-restricted-globals": 0,
+		"unicorn/prefer-node-protocol": 0
 	}
 }
--- a/packages/rest/.lintstagedrc.js
+++ b/packages/rest/.lintstagedrc.js
@@ -1,1 +1,4 @@
-module.exports = require('../../.lintstagedrc.json');
+module.exports = {
+	...require('../../.lintstagedrc.json'),
+	'src/**.ts': 'vitest related --run',
+};
--- a/packages/rest/.lintstagedrc.js
+++ b/packages/rest/.lintstagedrc.js
@@ -1,4 +1,4 @@
 module.exports = {
 	...require('../../.lintstagedrc.json'),
-	'src/**.ts': 'vitest related --run',
+	'src/**.ts': 'vitest related --run --config ./vitest.config.ts',
 };
--- a/packages/rest/src/lib/handlers/Shared.ts
+++ b/packages/rest/src/lib/handlers/Shared.ts
@@ -1,5 +1,3 @@
-import { setTimeout, clearTimeout } from 'node:timers';
-import { Response } from 'undici';
 import type { RequestInit } from 'undici';
 import type { ResponseLike } from '../REST.js';
 import type { HandlerRequestData, RequestManager, RouteData } from '../RequestManager.js';
@@ -65,7 +63,7 @@ export async function makeNetworkRequest(
 	retries: number,
 ) {
 	const controller = new AbortController();
-	const timeout = setTimeout(() => controller.abort(), manager.options.timeout).unref();
+	const timeout = setTimeout(() => controller.abort(), manager.options.timeout);
 	if (requestData.signal) {
 		// If the user signal was aborted, abort the controller, else abort the local signal.
 		// The reason why we don't re-use the user's signal, is because users may use the same signal for multiple
@@ -135,7 +133,7 @@ export async function handleErrors(
 		}
 
 		// We are out of retries, throw an error
-		throw new HTTPError(status, method, url, requestData);
+		throw new HTTPError(status, res.statusText, method, url, requestData);
 	} else {
 		// Handle possible malformed requests
 		if (status >= 400 && status < 500) {
--- a/packages/rest/tsup.config.ts
+++ b/packages/rest/tsup.config.ts
@@ -2,6 +2,6 @@ import { esbuildPluginVersionInjector } from 'esbuild-plugin-version-injector';
 import { createTsupConfig } from '../../tsup.config.js';
 
 export default createTsupConfig({
-	entry: ['src/agnostic.ts', 'src/web.ts', 'src/strategies/*.ts'],
+	entry: ['src/agnostic.ts', 'src/web.ts'],
 	esbuildPlugins: [esbuildPluginVersionInjector()],
 });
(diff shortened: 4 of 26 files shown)


packages/util/src/functions/runtime.ts
```diff
@@ -0,0 +1,12 @@
+export function shouldUseGlobalFetchAndWebSocket() {
+	// Browser env and deno when ran directly
+	if (typeof globalThis.process === 'undefined') {
+		return 'fetch' in globalThis && 'WebSocket' in globalThis;
```
Jarred-Sumner, line 4: For bun, you could probably check for these two as well and remove the `globalThis.process` check. Bun implements `fetch` and `WebSocket` as globals

packages/ws/src/ws/WebSocketShard.ts
```diff
@@ -461,17 +489,23 @@ export class WebSocketShard extends AsyncEventEmitter<WebSocketShardEventsMap> {
...
-		const decompressable = new Uint8Array(data);
-
+	private async unpackMessage(data: Data, isBinary: boolean): Promise<GatewayReceivePayload | null> {
 		// Deal with no compression
 		if (!isBinary) {
-			return JSON.parse(this.textDecoder.decode(decompressable)) as GatewayReceivePayload;
+			try {
+				return JSON.parse(data as string) as GatewayReceivePayload;
+			} catch {
+				// Either bun emulates the rejection as a message event or discord sent us invalid JSON, either way, we can't do anything with it
```
Jarred-Sumner, line 498: To clarify: Bun doesn't implement compression for client-side WebSocket yet. Only on the server

vladfrangu: That comment is related to the linked issue in the PR description (https://github.com/oven-sh/bun/issues/3392) 😄

vladfrangu:
For now, to not block this PR moving forward, /ws will keep using the ws module. Once the bun issue is solved, we'll enable using global ws instead

suneettipirneni (approved):

packages/ws/src/ws/WebSocketShard.ts
```diff
@@ -83,6 +83,12 @@ export interface SendRateLimitState {
 export class WebSocketShard extends AsyncEventEmitter<WebSocketShardEventsMap> {
 	private connection: WebSocket | null = null;
 
+	// TODO(vladfrangu): enable this once https://github.com/oven-sh/bun/issues/3392 is solved
+	// private WebSocketConstructor: typeof WebSocket = shouldUseGlobalFetchAndWebSocket()
+	// 	? (globalThis as any).WebSocket
+	// 	: WebSocket;
+	private WebSocketConstructor: typeof WebSocket = WebSocket;
```
kyranet, line 90: Shouldn't we make this static?

vladfrangu:
Yes, yes we should make it static instead, or even a variable outside the class. Good catch

packages/ws/src/ws/WebSocketShard.ts
```diff
@@ -261,13 +281,23 @@ export class WebSocketShard extends AsyncEventEmitter<WebSocketShardEventsMap> {
...
-				await once(this.connection, 'close');
+
+				let outerResolve: () => void;
+				const promise = new Promise<void>((resolve) => {
+					outerResolve = resolve;
+				});
+
+				this.connection.onclose = () => {
+					outerResolve();
+				};
```
kyranet, line 292: Wouldn't it be cleaner to do this?
```suggestion
				this.connection.onclose = outerResolve;
```

kyranet (changes requested):

packages/ws/src/ws/WebSocketShard.ts
didinele: Yeah, please do.

packages/rest/tsup.config.ts
```diff
@@ -2,6 +2,6 @@ import { esbuildPluginVersionInjector } from 'esbuild-plugin-version-injector';
 import { createTsupConfig } from '../../tsup.config.js';
 
 export default createTsupConfig({
-	entry: ['src/index.ts', 'src/strategies/*.ts'],
```
ckohen, line 5: you guessed it, n-n-n-n-nope

ckohen (changes requested):

kyranet (approved):

didinele (approved):

ckohen:
lgtm, just need the last commit message to have the breaking changes list

Bogdan1001:
I hope the changes will be applied asap :) I really need to use it with supabase.

vladfrangu:
> lgtm, just need the last commit message to have the breaking changes list

I'll just mark as blocked so kodiak doesn't do the auto merge but can't I specify the changelog and breaking entries when merging the PR myself?

kyranet:
You can, but you can also squash the commits with the breaking change list in the commit description.

7xa5h:
hm does this also fix the undici issue with windows?

Jiralite:
That sounds like something to raise to Undici.

suneettipirneni:
> hm does this also fix the undici issue with windows?

It should remedy this as it’ll opt to use the native fetch in deno instead of using undici.request.

7xa5h:
> > hm does this also fix the undici issue with windows?
> 
> It should remedy this as it’ll opt to use the native fetch in deno instead of using undici.request.

gr8, thanks for the reply!

ckohen (approved):

packages/rest/src/lib/handlers/Shared.ts
```diff
@@ -65,7 +63,7 @@ export async function makeNetworkRequest(
 	retries: number,
 ) {
 	const controller = new AbortController();
-	const timeout = setTimeout(() => controller.abort(), manager.options.timeout).unref();
+	const timeout = setTimeout(() => controller.abort(), manager.options.timeout);
```
KhafraDev, line 66:
```suggestion
	const timeout = setTimeout(() => controller.abort(), manager.options.timeout);
	if (typeof timeout !== 'number') timeout.unref()
```

that way the current behavior stays the same for node users?

ckohen: It's effectively a no-op 
https://github.com/discordjs/discord.js/pull/9682#discussion_r1252393899

Erisfiregamer1 (approved):
What's blocking this from being merged, Bun?

7xa5h (approved):

7xa5h:
> What's blocking this from being merged, Bun?

im p sure iCrawl has to approve first

packages/rest/.eslintrc.json
```diff
@@ -1,3 +1,11 @@
 {
-	"extends": "../../.eslintrc.json"
+	"extends": "../../.eslintrc.json",
+	"rules": {
+		"n/prefer-global/url": "off",
+		"n/prefer-global/url-search-params": "off",
+		"n/prefer-global/buffer": "off",
+		"n/prefer-global/process": "off",
+		"no-restricted-globals": "off",
+		"unicorn/prefer-node-protocol": "off"
```
iCrawl, line 9: 0 on all of those

packages/rest/.lintstagedrc.js
```diff
@@ -1 +1,4 @@
-module.exports = require('../../.lintstagedrc.json');
+module.exports = {
+	...require('../../.lintstagedrc.json'),
+	'src/**.ts': 'vitest related --run',
```
iCrawl, line 3: Doesn't this in the root:
`"src/**.ts": "vitest related --run --config ../../vitest.config.ts"` already do the same thing? Since it runs in each file path

packages/rest/tsup.config.ts
```diff
@@ -2,6 +2,6 @@ import { esbuildPluginVersionInjector } from 'esbuild-plugin-version-injector';
...
+	entry: ['src/agnostic.ts', 'src/web.ts', 'src/strategies/*.ts'],
```
iCrawl, line 5: Why the rename to agnostic? This can just stay as `index` no? The default therefore is agnostic.

iCrawl (dismissed):

packages/rest/.lintstagedrc.js
vladfrangu: @ckohen

packages/rest/tsup.config.ts
vladfrangu: Can rename it back to index, sure

packages/rest/.lintstagedrc.js
suneettipirneni: This change was made because the /rest directory has it's own vitest config. If this wasn't included it would use the root config which would break the tests.

iCrawl: Then let's be a bit more verbose and include loading the config as an additional parameter instead of it looking in the file-tree

Erisfiregamer1:
@vladfrangu you might need to merge with main, you're out of date

iCrawl (approved):

PR #9702, opened 2023-07-12
feat: add profile url property

MRDGH2821:

Added profile url. This will be useful for moderator bots which need to give profile url to the discord mod to check user's profile.
Other use cases include giving author url in embeds.


Syjalo:
This isn't something from the Bot API and this isn't in the API docs, so I don't think it should in discord.js.

almostSouji:
We do not plan to add undocumented protocols and links to the library, as they are not subject to semver and may break or change at any moment.

PR #9704, opened 2023-07-12
fix(WebSocketShard): close errors not being catchable

### LINKED ISSUE #9621: Regression starting from v14.10.0: "disallowed intents" is not catchable/handlable anymore.

discord.js

### Issue description

If you run `client.login()` specifying privileged intents (like `GuildMembers`) while the bot doesn't have permission for that, Node will crash due to an unhandled rejection.  
Up until v14.9.0, `client.login().catch()` would be able to catch this error and handle it accordingly, and I have tested and this is still an issue with `14.11.1-dev.1685793842-df8b6e9.0`.  
A few days ago I talked about it on the DJS discord and a few people acknowledge this issue, but I decided to create the GH issue to keep track of it since this ends up being a breaking issue in my use case where the bot is just one part of my OSS project.  
  
Test output:  
- Expected: https://media.discordapp.net/attachments/824411059443204127/1112593310398287933/Code_ZAk80nPgvh.png
- Actual: https://media.discordapp.net/attachments/824411059443204127/1112593310641553488/Code_1PKZOAcfgC.png

### Code sample

```typescript
import { Client, GatewayIntentBits } from 'discord.js';

const client = new Client({
    intents: [
        GatewayIntentBits.Guilds,
        GatewayIntentBits.GuildMembers,
    ]
})

client.on('ready', () => {
    console.error('client > ready fired!');
});

client.on('warn', (msg) => {
    console.error('client > warn fired!');
    console.warn(msg);
});
client.on('error', (msg) => {
    console.error('client > error fired!');
    console.error(msg);
});

client.login(process.env.DJS_TOKEN).catch((error) => {
    console.error(`handled login() error: ${error.message}`);
});
```

### Versions

Relevant versions listed above.  
This is TS/OS independent.

### Issue priority

Not applicable

Guilds, GuildMembers

df8b6e9

D4isDAVID:

Resolves #9621.

~~It seems like the errors could only occur when using `WebSocketShard#connect` (and I didn't want to modify much code), so my approach was to listen to the close event and throw the errors directly from `connect`, therefore making them catchable.~~

Swaps all `throw`s on close errors with emitting `WebSocketShardEvents.Error`, then catching and throwing them directly inside of `connect`.

### DIFF
--- a/packages/ws/src/ws/WebSocketShard.ts
+++ b/packages/ws/src/ws/WebSocketShard.ts
@@ -142,7 +142,9 @@ export class WebSocketShard extends AsyncEventEmitter<WebSocketShardEventsMap> {
 		void this.internalConnect();
 
 		try {
-			await promise;
+			await promise?.catch(({ error }) => {
+				throw error;
+			});
 		} finally {
 			// cleanup hanging listeners
 			controller.abort();
@@ -710,7 +712,9 @@ export class WebSocketShard extends AsyncEventEmitter<WebSocketShardEventsMap> {
 			}
 
 			case GatewayCloseCodes.AuthenticationFailed: {
-				throw new Error('Authentication failed');
+				return this.emit(WebSocketShardEvents.Error, {
+					error: new Error('Authentication failed'),
+				});
 			}
 
 			case GatewayCloseCodes.AlreadyAuthenticated: {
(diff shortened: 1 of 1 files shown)


didinele (changes requested):
Not fond of this approach and how it separates the close code handling at all.

I think I would much rather we keep it in the `onClose` method, swap out all the `throw`s for emitting `WebSocketShardEvents.Error` and instead leveraging `waitForEvent` inside of `connect`.

D4isDAVID:
> I think I would much rather we keep it in the `onClose` method, swap out all the `throw`s for emitting `WebSocketShardEvents.Error` and instead leveraging `waitForEvent` inside of `connect`.

Using `waitForEvent` would be a bit tricky no? It seems like other *recoverable* events can also happen when using `connect` and I don't see a way to use this function to ensure that the error is a close error. Should I maybe stick to using `once` from `node:events`?

didinele:
Yeah, just use `once`.

D4isDAVID:
I've ran into another issue where emitting `WebSocketShardEvents.Error` already causes `connect` to throw a catchable error (and as a result doesn't let me listen to it), but the error is for some reason `undefined`. I'll look into it later today.

D4isDAVID:
Turns out `once` handles error events, so I'll just use that.

packages/ws/src/ws/WebSocketShard.ts
```diff
@@ -142,7 +142,9 @@ export class WebSocketShard extends AsyncEventEmitter<WebSocketShardEventsMap> {
 		void this.internalConnect();
 
 		try {
-			await promise;
+			await promise?.catch(({ error }) => {
```
didinele, line 145: This is... very silly, we're in a `try` block and attach a `.catch` to the promise, just use a `catch` block instead.

packages/ws/src/ws/WebSocketShard.ts
```diff
@@ -710,7 +712,9 @@ export class WebSocketShard extends AsyncEventEmitter<WebSocketShardEventsMap> {
 			}
 
 			case GatewayCloseCodes.AuthenticationFailed: {
-				throw new Error('Authentication failed');
+				return this.emit(WebSocketShardEvents.Error, {
```
didinele, line 715: Should emit, and then destroy with no recovery - same as all the other emits.

didinele (changes requested):

didinele (approved):

vladfrangu:
Does this work with the worker strategy too?

didinele:
Definitely not:

https://github.com/discordjs/discord.js/blob/8d9701745840e23854e8f0b057d21cb10e7d1d54/packages/ws/src/strategies/sharding/WorkerShardingStrategy.ts#L221-L224

But that is way out of scope for this PR, I'll tackle that myself in the near future - lots of things the workers currently don't really forward or leave room to handle in the parent thread.

vladfrangu (approved):

kyranet (approved):

PR #9713, opened 2023-07-14
feat(ClientApplication): Approximate guild count and new `GET` route

Jiralite:

- https://github.com/discord/discord-api-docs/pull/5971
- https://github.com/discordjs/discord-api-types/pull/728

### DIFF
--- a/packages/discord.js/src/structures/ClientApplication.js
+++ b/packages/discord.js/src/structures/ClientApplication.js
@@ -69,6 +69,32 @@ class ClientApplication extends Application {
       this.flags = new ApplicationFlagsBitField(data.flags).freeze();
     }
 
+    if ('owner' in data) {
+      this._owner = this.client.users._add(data.owner);
+    } else {
+      this._owner ??= null;
+    }
+
+    if ('team' in data) {
+      /**
+       * The team that holds this application.
+       * @type {?Team}
+       */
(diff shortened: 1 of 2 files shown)


vladfrangu (approved):

packages/discord.js/src/structures/ClientApplication.js
```diff
@@ -118,16 +154,24 @@ class ClientApplication extends Application {
...
-      ? new Team(this.client, data.team)
-      : data.owner
-      ? this.client.users._add(data.owner)
-      : this.owner ?? null;
+  /**
+   * The owner of this application.
+   * @type {?(User|Team)}
+   * @readonly
+   */
+  get owner() {
```
SpaceEEC, line 164: This is technically a breaking change.
I'm also not sure why this is being changed, the behavior looks the same (except it now being read-only).

SpaceEEC (changes requested):

packages/discord.js/src/structures/ClientApplication.js
Jiralite: I could revert this whole change, actually. It was just something lingering from when there was a possible overwrite due to using two different routes that contained potentially different data.

kyranet (approved):

PR #9723, opened 2023-07-17
fix(REST): strip webhook tokens

### LINKED ISSUE #9415: Request route for interaction response deletions includes the token

rest

### Issue description

Follow-up of #9019, this time it's DELETE `/webhooks/:id/:token/@original`

### Code sample

```typescript
client.rest.on("response", req => console.log(req.route))
```

### Package version

1.7.0

### Node.js version

18.15.0

### Operating system

Windows 11

Low (slightly annoying)

_No response_

D4isDAVID:

Strip the token from webhook routes.

Fixes #9415.


packages/rest/src/lib/RequestManager.ts
```diff
@@ -509,7 +509,9 @@ export class RequestManager extends AsyncEventEmitter<RestEventsMap> {
 			// Strip out all ids
 			.replaceAll(/\d{17,19}/g, ':id')
 			// Strip out reaction as they fall under the same bucket
-			.replace(/\/reactions\/(.*)/, '/reactions/:reaction');
+			.replace(/\/reactions\/(.*)/, '/reactions/:reaction')
+			// Strip out webhook tokens
+			.replace(/\/webhooks\/:id\/[^/]+(?<slash>\/)?/, '/webhooks/:id/:token$<slash>');
```
ImRodry, line 514: There’s no need to use named groups imo, you can just use the regular $1

D4isDAVID: ESLint gave me a warning idk

ImRodry: Ahh ok then, that’s a new one

ckohen (changes requested):
This actually does change behavior significantly in the supposedly impossible case where two webhooks share an id (since webhooks *can* be keyed on id + token). Notably this is how interaction followups work (using the application id instead of the interaction id). 

What you've done is put all interaction followup requests (with the same method) in the same bucket!

We kinda got lucky with our implementation that happened to split out handlers because the route was different (because it had token in it) rather than because the major id was different.

Bonus to handling this properly would be that we don't create a bunch of duplicate hash entries in our hash cache.

In addition to what you've done (which is in the section that makes a route more generic to match how discord hashes routes), you need to edit the regex used to match major ids to include webhook tokens, so that we don't lose the uniqueness.

packages/rest/src/lib/REST.ts
```diff
@@ -434,16 +434,20 @@ export class REST extends AsyncEventEmitter<RestEventsMap> {
 			};
 		}
 
-		const majorIdMatch = /^\/(?:channels|guilds|webhooks)\/(\d{17,19})/.exec(endpoint);
+		const majorIdMatch = /(?:^\/webhooks\/(\d{17,19}\/[^/]+))|(?:^\/(?:channels|guilds|webhooks)\/(\d{17,19}))/.exec(
```
ckohen, line 437:
```suggestion
		const majorIdMatch = /(?:^\/webhooks\/(\d{17,19}\/[^/?]+))|(?:^\/(?:channels|guilds|webhooks)\/(\d{17,19}))/.exec(
```

Shouldn't match query params either

packages/rest/src/lib/REST.ts
```diff
@@ -434,16 +434,20 @@ export class REST extends AsyncEventEmitter<RestEventsMap> {
...
+			endpoint,
+		);
 
-		// Get the major id for this route - global otherwise
-		const majorId = majorIdMatch?.[1] ?? 'global';
+		// Get the major id or id + token for this route - global otherwise
+		const majorId = majorIdMatch?.[1] ?? majorIdMatch?.[2] ?? 'global';
```
ckohen, line 442:
```suggestion
		const majorId = majorIdMatch?.[2] ?? majorIdMatch?.[1] ?? 'global';
```

Flipping these because the standard match will happen far more often and when matching a webhook it stops processing after the first match

packages/rest/src/lib/REST.ts
```diff
@@ -434,16 +434,20 @@ export class REST extends AsyncEventEmitter<RestEventsMap> {
...
 
 		const baseRoute = endpoint
 			// Strip out all ids
 			.replaceAll(/\d{17,19}/g, ':id')
 			// Strip out reaction as they fall under the same bucket
-			.replace(/\/reactions\/(.*)/, '/reactions/:reaction');
+			.replace(/\/reactions\/(.*)/, '/reactions/:reaction')
+			// Strip out webhook tokens
+			.replace(/\/webhooks\/:id\/[^/]+/, '/webhooks/:id/:token');
```
ckohen, line 450:
```suggestion
			.replace(/\/webhooks\/:id\/[^/?]+/, '/webhooks/:id/:token');
```

ckohen (changes requested):

ckohen (approved):

Jiralite (approved):

suneettipirneni (approved):

PR #9750, opened 2023-08-04
fix: Set target to ES2021 for Node.js v16 compatibility

Tynopia:
This pull request resolves a compatibility issue between the discord.js library and Node.js v16.x. The change modifies the build process to target ES2021 rather than the more recent ES2022 standard.

## Issue Description

The issue arises from a new feature introduced in ES2022, the `static { ... }` initialization blocks in class bodies. While this feature is quite useful, it isn't supported in Node.js v16.x. Given that Node.js v16.x is still in wide use, despite Node.js v18.x being available, it's important to ensure that discord.js remains compatible with it.

Node.js v16.x supports a form of static initialization blocks: `static() { ... }`. However, this is not the same as the ES2022 feature of `static { ... }` initialization blocks. If we use ES2022 as our target in the build process, the TypeScript compiler will produce output code containing `static { ... }` blocks, which are not compatible with Node.js v16.x.

image

## The Solution

To resolve this issue, I have changed the `target` option in `tsup.config.js` from `ES2022` to `ES2021`. This ensures that the TypeScript compiler does not output `static { ... }` blocks in the transpiled code, making it compatible with Node.js v16.x.

While ES2021 does not support `static { ... }` blocks, the absence of these blocks does not adversely affect the functionality of discord.js. It only affects the version of JavaScript that the output code targets, ensuring that the library can be used with Node.js v16.x.

## Conclusion

Ensuring compatibility with Node.js v16.x is crucial for a sizable group of discord.js library users who have not yet upgraded to v18.x. By targeting ES2021 in our build process, we can maintain this compatibility while waiting for more Node.js users to migrate to versions that support the ES2022 features.

### DIFF
--- a/.github/workflows/publish-release.yml
+++ b/.github/workflows/publish-release.yml
@@ -1,4 +1,4 @@
-name: Publish Release
+name: Publish Release (CyberPurge)
 on:
   push:
   #release:
--- a/.github/workflows/publish-release.yml
+++ b/.github/workflows/publish-release.yml
@@ -1,15 +1,16 @@
 name: Publish Release
 on:
-  release:
-    types: [released]
+  push:
+  #release:
+  #  types: [released]
 jobs:
   npm-publish:
     name: npm publish
     runs-on: ubuntu-latest
-    env:
-      TURBO_TOKEN: ${{ secrets.TURBO_TOKEN }}
-      TURBO_TEAM: ${{ secrets.TURBO_TEAM }}
-    if: github.repository_owner == 'discordjs'
+    #env:
+    #  TURBO_TOKEN: ${{ secrets.TURBO_TOKEN }}
+    #  TURBO_TEAM: ${{ secrets.TURBO_TEAM }}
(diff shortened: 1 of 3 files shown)


Jiralite:
You keep saying this isn't supported in 16.x, but it is in 16.11?

ckohen:
No, you should update to Node.js 16.13 or later because that is the first version of Node 16 that is marked as LTS. Using 16.9 today is strongly recommended against due to security issues.

Further, this one ES2022 feature is not the only one we use, there are others that have much greater impact.

Sidenote: not sure why esbuild injects the static name blocks the way they do, it's an incredibly annoying way of solving an issue that should be solved much differently.

Tynopia:
@ckohen Unfortunately I can't update because I use NodeJS in conjunction with my FiveM server. CFX.RE currently does not plan to change its NodeJS version: https://github.com/citizenfx/node

Tynopia:
Do you possibly know the best way to upload my fork to my GitHub packages? I have now tried the last few commits a few things but I can not build more, ...

PR #9755, opened 2023-08-06
fix(Action): Do not set `undefined` values

### LINKED ISSUE #9754: Interaction channel loses guildId

discord.js

### Issue description

1. Run below code snippet
2. Trigger an interaction in a guild TextChannel (I used slash commands)
3. Notice how the channel.guildId returns undefined after deferReply

Both the Channel partial as well as the DirectMessages gateway intent are required for this issue to occur. The defer reply also need to be ephemeral. Changing any of those points prevents the issue from happening.

### Code sample

```typescript
import 'dotenv/config';
import Discord from 'discord.js';

const client = new Discord.Client( {
	intents: [
		Discord.GatewayIntentBits.Guilds,
		Discord.GatewayIntentBits.DirectMessages
	],
	partials: [
		Discord.Partials.Channel
	]
} );

client.on( Discord.Events.InteractionCreate, interaction => {
	console.log('Partial channel?', interaction.channel?.partial);
	if ( interaction.channel?.partial ) return interaction.channel.fetch().then( () => {
		return interactionCreate(interaction);
	}, log_error );
	return interactionCreate(interaction);
} );

function interactionCreate(interaction) {
	console.log('Before deferReply:', interaction.channel.guildId, interaction.channel.guild.name);
	interaction.deferReply( {ephemeral: true} ).then( () => {
		console.log('After deferReply: ', interaction.channel.guildId, interaction.channel.guild.name);
	} );
}

client.login(process.env.token);
```

### Versions

- discord.js 14.12.1
- Node.js 18.16.0
- NPM 9.8.1
- Windows 10

### Issue priority

Medium (should be fixed soon)

Channel

Guilds, DirectMessages

_No response_

Jiralite:
Resolves #9754.

The object being sent to `GenericAction#getPayload()` might contain keys that were `undefined`. In this specific case, an ephemeral message with the direct messages intent would not contain a `guild_id`.[^1] In combination with the partial, this would lead to `manager._add(data, cache)` which would then lead to here:

https://github.com/discordjs/discord.js/blob/7295a3a94a853086f781e7810b37c77d0c7e5283/packages/discord.js/src/structures/GuildChannel.js#L66-L68

Causing `guildId` to lose its value.

[^1]: https://discord.com/developers/docs/topics/gateway-events#messages

### DIFF
--- a/packages/discord.js/src/client/actions/Action.js
+++ b/packages/discord.js/src/client/actions/Action.js
@@ -28,20 +28,15 @@ class GenericAction {
   }
 
   getChannel(data) {
+    const payloadData = { recipients: data.recipients ?? [data.author ?? data.user ?? { id: data.user_id }] };
     const id = data.channel_id ?? data.id;
+    if (id !== undefined) payloadData.id = id;
+    if ('guild_id' in data) payloadData.guild_id = data.guild_id;
+    if ('last_message_id' in data) payloadData.last_message_id = data.last_message_id;
+
     return (
       data[this.client.actions.injectedChannel] ??
-      this.getPayload(
-        {
-          id,
-          guild_id: data.guild_id,
-          recipients: data.recipients ?? [data.author ?? data.user ?? { id: data.user_id }],
-          last_message_id: data.last_message_id,
-        },
-        this.client.channels,
-        id,
-        Partials.Channel,
-      )
+      this.getPayload(payloadData, this.client.channels, id, Partials.Channel)
     );
   }
 


sdanialraza (approved):

Markus-Rost:
I hope this can be merged in a patch release soon, per typings `guild_id` should never be undefined for TextChannel class.

vladfrangu (approved):

SpaceEEC (approved):

kyranet (approved):

PR #9756, opened 2023-08-08
feat(Util): add HSV and HSL color format support

codeblitz97:
In this pull request, I've made significant enhancements to the `resolveColor` function, empowering it to now handle HSL (Hue, Saturation, Lightness) and HSV (Hue, Saturation, Value) color representations. This addition not only broadens the spectrum of colors we can work with but also opens up exciting creative possibilities.

As someone who's passionate about colors, I truly believe that this enhancement will add a new layer of versatility to our projects. Whether you're aiming for a vibrant and eye-catching design or a more subtle and elegant palette, the flexibility of using HSL and HSV can take your color schemes to the next level.

To showcase the power of these additions, I've included usage examples below:

Using HSV on an embed:
```js
let embed = new EmbedBuilder().setColor({ h: 30, s: 100, v: 100 }); // Achieve that vivid orange using HSV
```

Or experimenting with HSL:
```js
let embed = new EmbedBuilder().setColor({ h: 30, s: 100, l: 50 }); // Capture the essence of orange using HSL
```

I'm confident that this enhancement will enhance our color-handling capabilities and elevate the visual aesthetics of our projects.

Looking forward to your feedback and suggestions as we continue to make strides in delivering exceptional experiences through our codebase. Thanks for considering these changes!

### DIFF
--- a/packages/discord.js/src/util/Util.js
+++ b/packages/discord.js/src/util/Util.js
@@ -235,6 +235,14 @@ function verifyString(
  * ```js
  * [255, 0, 255] // purple
  * ```
+ * or HSV object like:
+ * ```js
+ * { h: 30, s: 100, v: 100 }
+ * ```
+ * or HSL object like:
+ * ```js
+ * { h: 30, s: 100, l: 50 }
+ * ```
  * or one of the following strings:
  * - `Default`
  * - `White`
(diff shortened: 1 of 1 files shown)


kyranet:
I'm opposed to supporting this, for starters, users could expect the array overload (always RGB) to be HSL or HSV.

However, later we would face the issue where people want to use full names (`hue`, `saturation`, and `luminance`) over the short names (`h`, `s`, `l`).

And that's not to mention the users who would like to use other colour formats, e.g. [CIELUV](https://en.wikipedia.org/wiki/CIELUV), [CIELAB](https://en.wikipedia.org/wiki/CIELAB_color_space), [HSLuv](https://en.wikipedia.org/wiki/HSLuv), LCh (which is a cylindrical transformation of the two first), and so on. At that point, it's better to just use a dedicated color library such as [`colord`](https://www.npmjs.com/package/colord), which supports all the listed formats and even more.

vladfrangu:
I definitely concur with @kyranet's main point. I don't think this should be added to .setColor (not in builders, nor in discord.js), and instead utility methods should be used

codeblitz97:
You all think this should not be added?

iCrawl:
I'm not entirely against it. I think it has its place.

codeblitz97:
So I'll just close the PR?

PR #9761, opened 2023-08-10
fix(types): fixed CachedManager constructor arguments in type

LilyBergonzat:
The CachedManager's constructor has three arguments in the JS files but only two in the type declaration. I fixed that so it matches.

- There are no code changes

### DIFF
--- a/packages/discord.js/typings/index.d.ts
+++ b/packages/discord.js/typings/index.d.ts
@@ -3734,7 +3734,7 @@ export abstract class DataManager<K, Holds, R> extends BaseManager {
 }
 
 export abstract class CachedManager<K, Holds, R> extends DataManager<K, Holds, R> {
-  protected constructor(client: Client<true>, holds: Constructable<Holds>);
+  protected constructor(client: Client<true>, holds: Constructable<Holds>, iterable?: Iterable<Holds>);
   private readonly _cache: Collection<K, Holds>;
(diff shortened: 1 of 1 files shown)


packages/discord.js/typings/index.d.ts
```diff
@@ -3734,7 +3734,7 @@ export abstract class DataManager<K, Holds, R> extends BaseManager {
 }
 
 export abstract class CachedManager<K, Holds, R> extends DataManager<K, Holds, R> {
-  protected constructor(client: Client<true>, holds: Constructable<Holds>);
+  protected constructor(client: Client<true>, holds: Constructable<Holds>, iterable: Iterable<Holds>);
```
jaw0r3k, line 3737:
```suggestion
  protected constructor(client: Client<true>, holds: Constructable<Holds>, iterable?: Iterable<Holds>);
```
as it is not required

LilyBergonzat: Good catch, thank you!

kyranet (approved):

sdanialraza (approved):

SpaceEEC (approved):
https://github.com/discordjs/discord.js/blob/d8e37551ceefe9f82566e3f45edc69bb7f9d1463/packages/discord.js/src/managers/CachedManager.js#L11

PR #9766, opened 2023-08-12
feat(Presence): Expose sync_id in Activity

almostSouji:

Exposes `sync_id` via the `Activity` API, which is present in raw ws updates but not currently made available by discord.js. This represents the track id for Spotify activities, which is necessary to build Spotify links and request data from the Spotify API. (I am not sure if it has any use in non-Spotify activities, hence the generic name following API conventions with an info field in the documentation seemed appropriate)


Syjalo:
The property was removed in #7844 because it's undocumented

almostSouji:
The sync id represents the track id in spotify, which is integral to retrieving additional information or interacting with the spotify API. This cannot be retrieved from another property. Not exposing it means denying all that functionality.

The only alternative to exposing this on the class I see is the wider application of an opt-in raw data store within classes, which enables users to handle features we have not (yet, or due to lack of documentation) implemented.

advaith1:
the removal of this field is a reason I stuck with a custom presence cache and didn't use the d.js presence cache for discordprofiles/statusbadges, since I rely on this field (though now I use djs/core since I don't need any djs caching)

packages/discord.js/src/structures/Presence.js
```diff
@@ -216,6 +216,13 @@ class Activity {
      */
     this.party = data.party ?? null;
 
+    /**
+     * The sync id of the activity
+     * <info>This represents the track id in spotify activities.</info>
```
vladfrangu, line 221: Not sure if we should mention this or let users figure it out

vladfrangu (approved):

packages/discord.js/src/structures/Presence.js
almostSouji: that is the prime (if not only) use case for it at the moment, seems worth mentioning IMO for ease of use

kyranet (approved):

Jiralite:
It would probably be a good idea regardless to state this property is not documented. Just in case.

SpaceEEC (approved):

PR #9768, opened 2023-08-12
feat(Message): add 'referencedMessage' property

Pavel-Boyazov:
Allow to get referenced message synchronously without a need to fetch it from Discord API

### DIFF
--- a/packages/discord.js/src/structures/Message.js
+++ b/packages/discord.js/src/structures/Message.js
@@ -2,6 +2,7 @@
 
 const { messageLink } = require('@discordjs/builders');
 const { Collection } = require('@discordjs/collection');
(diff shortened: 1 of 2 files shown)


almostSouji:
`fetchReference` already attempts to get from cache before hitting the API (the resolve step below):
https://github.com/discordjs/discord.js/blob/9ed1b59df6acb6356d5950b43d04885d5e692887/packages/discord.js/src/structures/Message.js#L684-L691

and the referenced message is added to the cache here:
https://github.com/discordjs/discord.js/blob/9ed1b59df6acb6356d5950b43d04885d5e692887/packages/discord.js/src/structures/Message.js#L368-L370

PR #9771, opened 2023-08-13
feat(Client): AsyncIterator-returning method for fetching members

### LINKED ISSUE #9017: `/core`: add an AsyncIterator-returning method for fetching members

core

### Feature

The method would return an async iterator, which we would be able to use to consume data and process them as we receive them, rather than collecting all values into a collection and process them all at once.

The async iterator comes with a lot of benefits, one is that it allows us to process batches of up to 1000 members, distributing the load more evenly, e.g. creating classes from the members (e.g. discord.js), caching them somewhere (e.g. redis) or checking data from them (e.g. dehoisters, name search, etc).

```typescript
for await (const member of client.requestGuildMembers({ guild_id })) {
	// ...
}
```

Extend the work done in https://github.com/discordjs/discord.js/pull/8941, this time by implementing a method that returns an async iterator (we can have a private — or public — method that iterates over the raw batches, so we can minimize code duplication in the collect-all and the iterable method while maximizing performance.

The current method, `requestGuildMembers`, may be renamed.

N/a.

### Other context

https://github.com/discordjs/discord.js/pull/8941#pullrequestreview-1229770370

almeidx:
Resolves #9017

Not sure what to name the new method, open to suggestions

### Summary
:sparkles::recycle::books:

Added a new way to fetch guild members data using an async iterator. Improved the existing method to use the new iterator. Changed `packages/core/src/client.ts`.

> _Sing, O Muse, of the skillful `Client` who devised_
> _A new way to fetch the members of many guilds_
> _With an async iterator, a wondrous tool_
> _Like the golden fleece that Jason sought with his crew_

### Walkthrough
*  Add `requestGuildMembersIterator` method to `Client` class to return an async iterator for guild members data ([link](https://github.com/discordjs/discord.js/pull/9771/files?diff=unified&w=0#diff-858339789c798051d9c13620bffce8ce860e7ad8ba18dba1c8711eb1f7d51a55L248-R271), [link](https://github.com/discordjs/discord.js/pull/9771/files?diff=unified&w=0#diff-858339789c798051d9c13620bffce8ce860e7ad8ba18dba1c8711eb1f7d51a55L1-R2))
*  Refactor `requestGuildMembers` method to use `requestGuildMembersIterator` internally and collect the data into a promise ([link](https://github.com/discordjs/discord.js/pull/9771/files?diff=unified&w=0#diff-858339789c798051d9c13620bffce8ce860e7ad8ba18dba1c8711eb1f7d51a55L208-R216))
*  Import `on` and `clearTimeout` functions from `node:events` and `node:timers` modules for `requestGuildMembersIterator` ([link](https://github.com/discordjs/discord.js/pull/9771/files?diff=unified&w=0#diff-858339789c798051d9c13620bffce8ce860e7ad8ba18dba1c8711eb1f7d51a55L1-R2))


jaw0r3k (approved):

packages/core/src/client.ts
```diff
@@ -205,36 +206,13 @@ export class Client extends AsyncEventEmitter<ManagerShardEventsMap> {
...
-					this.off(GatewayDispatchEvents.GuildMembersChunk, handler);
-					resolve({ members, nonce, notFound, presences });
-				}
-			};
-
-			this.on(GatewayDispatchEvents.GuildMembersChunk, handler);
-		});
+		const timer = setTimeout(() => {
+			throw new Error('Request timed out');
+		}, timeout);
```
kyranet, line 215: This won't work by the way, you need to use an `AbortController` and pass its `signal` to [`events.on`](https://nodejs.org/dist/latest-v20.x/docs/api/events.html#eventsonemitter-eventname-options)'s options.

kyranet (changes requested):
We should also change the tsdoc for `requestGuildMembersIterator` to clarify it's a method that returns batches, preferably with an `@example` as well.

packages/core/src/client.ts
```diff
@@ -245,7 +235,63 @@ export class Client extends AsyncEventEmitter<ManagerShardEventsMap> {
...
+				if (data.chunk_index >= data.chunk_count - 1) {
+					break;
+				} else {
+					timer = createTimer();
+				}
+			}
+		} catch (error) {
+			if (error instanceof Error && error.name === 'AbortError') {
+				throw new Error('Request timed out');
+			}
```
kyranet, line 264: We shouldn't do this, technically `controller.abort(new Error('Request timed out'));` should achieve the same effect, although I can see stack traces dying there... maybe create the error before the timer? 🤔 

We should handle the case for this block:
```js
for await (const _ of client.requestGuildMembersIterator({ guild_id: '...' })) {
	AbortSignal.abort().throwIfAborted();
}
```

If my memory on `for await...of` is still correct, throwing an error inside a `for...of` (or a `for await...of`) causes `yield` to throw an error, which in this case it'll call this `catch`, and this piece will overwrite the user's error, which is undesirable and not intended.

packages/core/src/client.ts
```diff
@@ -245,7 +235,63 @@ export class Client extends AsyncEventEmitter<ManagerShardEventsMap> {
...
+
+			throw error;
+		}
```
kyranet, line 267: We should add a `finally {}` to clear the timer in case the user `throw`s or `return`s inside the loop.

kyranet (changes requested):

packages/core/src/client.ts
```diff
@@ -1,4 +1,5 @@
-import { setTimeout } from 'node:timers';
+import { on } from 'node:events';
```
vladfrangu, line 1: Use AsyncEventEmitter.on

packages/core/src/client.ts
vladfrangu: I agree, we should just remove this entire catch block and let errors bubble up

vladfrangu (dismissed):

packages/core/src/client.ts
almeidx: I had already sent this in Discord, but I'll send it here as well just for transparency

> We shouldn't do this, technically `controller.abort(new Error('Request timed out'));` should achieve the same effect

Doing this makes it throw an AbortError with its `cause` set to `new Error('Request timed out')`, hence why I was catching & rethrowing, e.g.:
image

> If my memory on `for await...of` is still correct, throwing an error inside a `for...of` (or a `for await...of`) causes `yield` to throw an error, which in this case it'll call this `catch`, and this piece will overwrite the user's error, which is undesirable and not intended.

This does not appear to be the case:

image
image

almeidx: After some internal discussion, we're keeping this as it is

vladfrangu (approved):

kyranet (approved):

PR #9777, opened 2023-08-16
feat(Client): add deleteWebhook method

jaw0r3k:

### <samp>🤖 Generated by Copilot at 3652f05</samp>

Add `deleteWebhook` method to `Client` class to enable deleting webhooks from Discord. Define `WebhookDeleteOptions` typedef for the method's options parameter.


packages/discord.js/src/client/Client.js
```diff
@@ -253,6 +253,23 @@ class Client extends BaseClient {
     this.rest.setToken(null);
   }
 
+  /**
+   * Options used when deleting a webhook from Discord.
```
Jiralite, line 257: The "from Discord" part is superfluous. Everything in this library is for Discord. No need to state "Create an event in Discord", "Delete a channel from Discord", etc.

```suggestion
   * Options used for deleting a webhook.
```

jaw0r3k: Well every jsdoc here uses that
But yeah i agree

https://github.com/discordjs/discord.js/blob/5de608a576a0b390e45074bed16b8dd47710b38e/packages/discord.js/src/client/Client.js#L281

Jiralite: We should move away from those patterns. I mean, that's what this library is for, after all.

Jiralite (commented):
`Webhook#delete()` can call this method internally to avoid duplication:

https://github.com/discordjs/discord.js/blob/24fbb11ba2f7e8f8f604752159d2053f2cee16ec/packages/discord.js/src/structures/Webhook.js#L365-L375

sdanialraza (approved):

SpaceEEC (approved):

kyranet (approved):

packages/discord.js/src/client/Client.js
advaith1: that's probably to clarify that it does an API request rather than using cache

Issue #9785, opened 2023-08-20
Unable to delete webhook when instantiated with WebhookClient


GodderE2D:
### Which package is this bug report for?

discord.js

### Issue description

When using `new WebhookClient()` and calling `.delete()` on it, it will error with this:

```js
TypeError: this.client.deleteWebhook is not a function
    at WebhookClient.delete (/home/ubuntu/.../.../node_modules/discord.js/src/structures/Webhook.js:371:24)
```

This issue happens in 14.13.0 but not in 14.12.1, likely caused by https://github.com/discordjs/discord.js/commit/d90ba8dce8de630db14b77764ec35201998c7ce7.

A workaround is to use [`Client.deleteWebhook()`](https://old.discordjs.dev/#/docs/discord.js/14.13.0/class/Client?scrollTo=deleteWebhook) however this requires a client and bot user to work.

### Code sample

```typescript
const webhookClient = new WebhookClient({ url: 'https://discord.com/api/webhooks/<channel>/<token>' });

await webhookClient.delete();
```

### Versions

- discord.js 14.13.0
- Node.js 18.16.0
- macOS 13.3

### Issue priority

Medium (should be fixed soon)

### Which partials do you have configured?

No Partials

### Which gateway intents are you subscribing to?

Guilds

### I have tested this issue on a development release

8eb978d32

PR #9786, opened 2023-08-20
fix(Webhook): do not call `client.deleteWebhook` in `delete`

kyranet:

~~Partially reverts #9777 which caused a regression (#9785) when using `WebhookClient`.~~

Moves the `deleteWebhook` client from `Client` to `BaseClient` so both `Client` and `WebhookClient` can use it.


SpaceEEC (approved):

sdanialraza (approved):

Jiralite (approved):

Jiralite (changes requested):
On second thought, why not simply move the implementation from `Client` to `BaseClient`? There will be no need to duplicate the implementation in Webhook.js since `WebhookClient` and `Client` both extend `BaseClient`.

Jiralite (approved):

vladfrangu (approved):

PR #9787, opened 2023-08-20
feat(CDN): Support emoji size

Jiralite:
`size` may affect emojis:

- https://cdn.discordapp.com/emojis/489818863341535247.gif?size=256
- https://cdn.discordapp.com/emojis/1117578868140806164.webp?size=16


didinele (approved):

SpaceEEC (approved):

sdanialraza (approved):

packages/rest/src/lib/CDN.ts
```diff
@@ -154,14 +157,39 @@ export class CDN {
...
-		return this.makeURL(`/emojis/${emojiId}`, { extension });
+	// eslint-disable-next-line @typescript-eslint/unified-signatures
+	public emoji(emojiId: string, extension?: ImageExtension): string;
+
+	public emoji(emojiId: string, options?: ImageExtension | Readonly<BaseImageURLOptions>): string {
+		let resolvedOptions;
+
+		if (typeof options === 'string') {
+			if (!deprecationEmittedForEmoji) {
+				deprecationWarning('Passing a string for CDN#emoji() is deprecated. Use an object instead.');
```
ckohen, line 183: This isn't right is it? It's specifically for the second param, which this doesn't indicate to me

ckohen (changes requested):

packages/rest/src/lib/CDN.ts
Jiralite: Would "Passing a string for the extension..." be better?

ckohen: I don't think so, cause you still pass a string to the extension, it'd be more along the lines of "Passing an extension as the second parameter" or something (I don't like what I said there even cause it kinda suggests you can't pass an extensions at all)

Jiralite: This?

```suggestion
				deprecationWarning(
					'Passing a string for the second parameter of CDN#emoji() is deprecated. Use a sole object for the first parameter instead.',
				);

```

ckohen: Am I misunderstanding something, the first parameter is still the emojiId regardless right?

Jiralite: You're right! I've edited the suggestion.

ckohen: Yeah that's probably fine.

ckohen (approved):

PR #9788, opened 2023-08-20
feat(Emoji): Add `imageURL()`

Jiralite:
Supersedes #8493.

This allows both `extension` and `size`. The getter has been deprecated in favour of the method, which is consistent with every other image URL method. `forceStatic` cannot be used as that option relies on a hash which does not exist for emojis.

Requires:

- #9787

### DIFF
--- a/packages/discord.js/src/structures/BaseGuildEmoji.js
+++ b/packages/discord.js/src/structures/BaseGuildEmoji.js
@@ -53,4 +53,23 @@ class BaseGuildEmoji extends Emoji {
   }
 }
 
+/**
+ * Returns a URL for the emoji.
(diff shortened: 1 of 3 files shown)


vladfrangu (approved):

SpaceEEC:
The upstream PR https://github.com/discordjs/discord.js/pull/9787 seems to not have been released yet. 👀

SpaceEEC (approved):

Jiralite:
I believe we can merge this still. It only means /rest would have to be released before this package, and builds will work because of the workspace resolution. The only caveat is that development releases won't install workspace resolutions.

didinele (approved):

PR #9789, opened 2023-08-20
refactor(CommandInteractionOptionResolver): remove getFull from getFocused()

ImRodry:
This PR removes the argument for getFull in CommandInteractionOptionResolver#getFocused and instead makes the function always return the full object. This change was discussed in a [thread in the DJS Discord server](https://discord.com/channels/222078108977594368/1140704683472728145) and the reason I am opening this PR is because I personally have never found a use case where I don't need the name of the option, and everyone should always check that property in their code in order to future-proof it in the event that a command ever gets 2 options with autocomplete with different names and different behaviors. I am also open to feedback as another option I considered in that thread was to change the default of this parameter to true instead of false, thus leaving this shortcut for the people that want to be less safe about their code but still making the default the recommended and more common behavior.


packages/discord.js/src/structures/CommandInteractionOptionResolver.js
```diff
@@ -294,14 +294,13 @@ class CommandInteractionOptionResolver {
 
   /**
    * Gets the focused option.
-   * @param {boolean} [getFull=false] Whether to get the full option object
    * @returns {string|AutocompleteFocusedOption}
```
Jiralite, line 297:
```suggestion
   * @returns {AutocompleteFocusedOption}
```

sdanialraza (approved):

SpaceEEC (approved):

vladfrangu (approved):

Jiralite (approved):

PR #9797, opened 2023-08-22
feat: Implement `GET` current application

Jiralite:

- https://github.com/discord/discord-api-docs/pull/5971


suneettipirneni (approved):

packages/core/src/api/index.ts
```diff
@@ -28,6 +30,8 @@ export * from './voice.js';
 export * from './webhook.js';
 
 export class API {
+	public readonly application: ApplicationsAPI;
```
suneettipirneni, line 33: This should be probably be pluralized to match with the other fields.
```suggestion
	public readonly applications: ApplicationsAPI;
```

suneettipirneni (changes requested):

packages/core/src/api/index.ts
Jiralite: I thought about this too, but there are many application commands, channels, guilds, interactions, invites, etc. but only _one_ application to do things against. Should I still change it?

suneettipirneni: When I used the plural version, it was more along the lines of "this is the API for {X}s" rather than it being an indication if multiple versions of X are available. I think this and the change of the name to `getCurrent()` would fit better.

packages/core/src/api/application.ts
```diff
@@ -0,0 +1,18 @@
...
+export class ApplicationsAPI {
+	public constructor(private readonly rest: REST) {}
+
+	/**
+	 * Fetches the application associated with the requesting bot user.
+	 *
+	 * @see {@link https://discord.com/developers/docs/resources/application#get-current-application}
+	 * @param options - The options for editing the application
+	 */
+	public async get({ signal }: Pick<RequestData, 'signal'> = {}) {
```
suneettipirneni, line 15: This would make the method more consistent with the [`UsersAPI` method](https://github.com/discordjs/discord.js/blob/f6c7c14c9e28c544050b396a4374fdacb0e25fa4/packages/core/src/api/user.ts#L43C3-L43C3).

```suggestion
	public async getCurrent({ signal }: Pick<RequestData, 'signal'> = {}) {
```

suneettipirneni (approved):

vladfrangu (approved):

SpaceEEC (approved):

PR #9800, opened 2023-08-23
Update wording for GuildMember.timeout parameter documentation

jeremy-rifkin:
The current wording for the `timeout` parameter of `GuildMember.timeout` strikes me as easily misinterpretable as the end-time at which the member's communications will be disable until, rather than a duration. This is a tiny documentation pr to improve the wording.

### DIFF
--- a/packages/discord.js/src/structures/GuildMember.js
+++ b/packages/discord.js/src/structures/GuildMember.js
@@ -425,8 +425,8 @@ class GuildMember extends Base {
 
   /**
    * Times this guild member out.
-   * @param {number|null} timeout The time in milliseconds
-   * for the member's communication to be disabled until. Provide `null` to remove the timeout.
+   * @param {number|null} timeout The duration in milliseconds
+   * for the member's communication to be disabled. Provide `null` to remove the timeout.
    * @param {string} [reason] The reason for this timeout.
    * @returns {Promise<GuildMember>}
    * @example


packages/discord.js/src/structures/GuildMember.js
```diff
@@ -425,8 +425,9 @@ class GuildMember extends Base {
 
   /**
    * Times this guild member out.
-   * @param {number|null} timeout The time in milliseconds
-   * for the member's communication to be disabled until. Provide `null` to remove the timeout.
+   * @param {number|null} timeout The duration in milliseconds
+   * for the member's communication to be disabled. The maximum duration allowed by discord is 28 days.
```
Jiralite, line 429: I agree with the duration wording. However, we shouldn't document API limits in case they change (the version of the documentation may still reference incorrect limits).

jeremy-rifkin: Thank you for the quick reply. Would it be ok to mention that a limit exists (or may exist) and link to discord's documentation?

Jiralite: Since we don't do this anywhere else, I think it'd be best to leave for future discussion rather than now.

SpaceEEC (approved):

vladfrangu (approved):

Jiralite (approved):

PR #9828, opened 2023-09-10
types(Client): fix isReady narrowing

### LINKED ISSUE #9827: Client.isReady() improperly narrows type to never

discord.js

### Issue description

When `Client.isReady()` returns false, the type of Client in that scope is narrowed to `never`. That doesn't seem like it should be the case, as you should be able to do something like this:

```typescript
if (!client.isReady()) {
  // Error: Property 'once' does not exist on type 'never'.
  client.once(Events.ClientReady, () => {
        // do stuff
  })
}
```

As a workaround, you can look at `readyAt`, which will keep the type of Client as `Client<boolean>`:

```typescript
if (!client.readyAt) {
  client.once(Events.ClientReady, () => {
        // do stuff
  })
}
```

### Code sample

```typescript
import { Client, Events } from "discord.js"

const client = new Client(clientOptions)

if (!client.isReady()) {
  client.once(Events.ClientReady, () => {
    // do stuff
  })
}
```

### Versions

- discord.js 14.13.0
- typescript 5.2.2

### Issue priority

Low (slightly annoying)

Not applicable

Not applicable

_No response_

N1ckPro:
Fixes #9827 

### DIFF
--- a/packages/discord.js/typings/index.d.ts
+++ b/packages/discord.js/typings/index.d.ts
@@ -955,6 +955,7 @@ export class Client<Ready extends boolean = boolean> extends BaseClient {
   private _eval(script: string): unknown;
   private _validateOptions(options: ClientOptions): void;
   private get _censoredToken(): string | null;
+  private readonly _ready: Ready;
 
   public application: If<Ready, ClientApplication>;
   public channels: ChannelManager;
--- a/packages/discord.js/typings/index.d.ts
+++ b/packages/discord.js/typings/index.d.ts
@@ -955,6 +955,7 @@ export class Client<Ready extends boolean = boolean> extends BaseClient {
   private _eval(script: string): unknown;
   private _validateOptions(options: ClientOptions): void;
   private get _censoredToken(): string | null;
+  // This a technique used to brand the ready state. Or else we'll get `never` errors on typeguard checks.
   private readonly _ready: Ready;
 
   public application: If<Ready, ClientApplication>;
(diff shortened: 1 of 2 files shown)


packages/discord.js/typings/index.d.ts
```diff
@@ -955,6 +955,7 @@ export class Client<Ready extends boolean = boolean> extends BaseClient {
   private _eval(script: string): unknown;
   private _validateOptions(options: ClientOptions): void;
   private get _censoredToken(): string | null;
+  private readonly _ready: Ready;
```
Jiralite, line 958: A comment would be ideal as to why this is needed and not present in JavaScript like so:

https://github.com/discordjs/discord.js/blob/c50809e20648cacea99f5450e8073d960ff8aa39/packages/discord.js/typings/index.d.ts#L1799-L1800

Jiralite (commented):
Will you add tests, please?

packages/discord.js/typings/index.test-d.ts
```diff
@@ -197,6 +197,10 @@ const client: Client = new Client({
   }),
 });
 
+if (client.isReady()) {
```
didinele, line 200: Please also add one for the opposite case, since that was actually the problematic one.

didinele (changes requested):

didinele (approved):

Jiralite (approved):

jaw0r3k (approved):

SpaceEEC (approved):

PR #9831, opened 2023-09-11
types: don't include dom types

### LINKED ISSUE #9737: Don't include dom types

discord.js

### Issue description

discord.js currently includes the dom types lib in its main types.d.ts
https://github.com/discordjs/discord.js/blob/8f4256db8a52ac08359d0b3436f41b641ac4e382/packages/discord.js/typings/index.d.ts#L1-L2
This leads to anyone importing discord.js into their project having dom types globally available (such as `document`, `window`, `HTMLDivElement`, ...)  
These types are of course false, since most discord.js projects will run in node and not in the browser.  
Looking at the comment ` // DOM types required for undici`, it seems to reference [the experimental `fetch` global](https://nodejs.org/api/globals.html#fetch) that was [recently added](https://github.com/nodejs/node/pull/41749) to nodejs.  
This fetch global internally uses `undici`, which is what that comment is likely referencing, since [the undici npm package](https://www.npmjs.com/package/undici) is properly typed out of the box.   

I could not find any references to the global fetch in the code from a very brief search, so I'm assuming a bit here - however since you already depend on undici anyway:
https://github.com/discordjs/discord.js/blob/8f4256db8a52ac08359d0b3436f41b641ac4e382/packages/discord.js/package.json#L65
using the global fetch would not make much sense. You are using an untyped experimental api that is only available in recent node versions, even though you have that same fetch function already installed as a dependency.

To clarify what difference this one line ` /// <reference lib="dom" />` makes, here are autocompletion suggestions when importing discord.js vs when not importing it:
image
image
As soon as discord.js is imported, typescript/vscode will start suggesting browser-only variables and classes. This makes this issue unavoidable, since it occours in your whole project as soon as you simply import discord.js, which is why I'd argue that this is not low priority.

Relevant:
https://github.com/discordjs/discord.js/commit/d0c82561b8a1765a1daa362ca903f3ffb3fa33ee by @favna introduced this issue
[[A](https://ptb.discord.com/channels/222078108977594368/1134145116844605531/1134145116844605531)] [[B](https://ptb.discord.com/channels/222078108977594368/937420598383108167/1131543523313848390)] Some brief conversation regarding this issue on the discord.js discord server

### Code sample

_No response_

### Versions

- discord.js v14.11.0
- NodeJS v18.16.1
- Typescript v5.1.6

### Issue priority

Medium (should be fixed soon)

Not applicable

Not applicable

_No response_

SuperchupuDev:

> As soon as discord.js is imported, typescript/vscode will start suggesting browser-only variables and classes. This makes this issue unavoidable, since it occours in your whole project as soon as you simply import discord.js, which is why I'd argue that this is not low priority.

There isn't really a reason to include DOM types for undici, and the commit that introduced it doesn't explain any further behind the reasons for this change

Closes #9737

### DIFF
--- a/packages/discord.js/typings/index.d.ts
+++ b/packages/discord.js/typings/index.d.ts
@@ -170,7 +170,7 @@ import { ChildProcess } from 'node:child_process';
 import { EventEmitter } from 'node:events';
 import { Stream } from 'node:stream';
 import { MessagePort, Worker } from 'node:worker_threads';
-import * as WebSocket from 'ws';
+import { Data as WebSocketData, WebSocket } from 'ws';
 import {
   RawActivityData,
   RawAnonymousGuildData,
@@ -5895,7 +5895,7 @@ export type MessageChannelComponentCollectorOptions<T extends CollectedMessageIn
 >;
 
 export interface MessageEvent {
-  data: WebSocket.Data;
+  data: WebSocketData;
   type: string;
   target: WebSocket;
 }
--- a/packages/discord.js/typings/index.d.ts
+++ b/packages/discord.js/typings/index.d.ts
@@ -1,6 +1,3 @@
-// DOM types required for undici
-/// <reference lib="dom" />
-
 import {
   ActionRowBuilder as BuilderActionRow,
   MessageActionRowComponentBuilder,


SuperchupuDev:
oh well, looks like it was needed :P, gonna to try going for a different solution

image

SuperchupuDev:
There are two possible solutions to this problem, I commited the latter but it's up to the maintainers really

- Leave the `ws` import as it is and replace all `WebSocket` mentions with `WebSocket.WebSocket`
- Change the import to `import { Data as WebSocketData, WebSocket }` and replace the only mention of `WebSocket.Data` with `WebSocketData`

SpaceEEC (approved):

didinele (approved):

iCrawl (approved):

PR #9835, opened 2023-09-15
types(Partials): add toString() method to supported Partials

### LINKED ISSUE #9833: Missing overrides for .toString() in partial types

discord.js

### Issue description

I've discovered incorrect typings in some of the partial types. Particularly, the `.toString()` method supposedly always evaluates to `"[object Object]"`, despite this never being the case during runtime.

Why I believe I'm correct:
- `<GuildMember>.toString()` returns a `UserMention` which is (ultimately) based on `<GuildMember>.user.id`
- `<GuildMember>.id` returns `<GuildMember>.user.id`
- The `id` property is guaranteed to exist on all partial types (see `Partialize` type)
- Therefore, `<PartialGuildMember>.user.id` is also guaranteed to exist
- **That is to say: `PartialGuildMember`s are always mentionable**
- Consequently, `<PartialGuildMember>.toString()` should always yield a string like the non-partial variant does

During runtime, `PartialGuildMember` always has a `.toString()` method attached.
However, typescript-eslint (extending `plugin:@typescript-eslint/strict-type-checked`) tells me that the latter may evaluate to "[object Object]" (@typescript-eslint/no-base-to-string).

This problem also exists on other partial types. I have yet to look a bit further into this, but so far I've identified this to potentially be the case with `User`, `Channel`, `MessageReaction` and `ThreadMember`.

### Code sample

```typescript
/* eslint-disable @typescript-eslint/no-unused-vars */

import { GuildMember, PartialGuildMember } from "discord.js";

// No linting error here
function test1(member: GuildMember) {
  console.log(`${member.toString()}`);
}

// Throws @typescript-eslint/no-base-to-string with typescript-eslint
function test2(member: PartialGuildMember) {
  console.log(`${member.toString()}`);
}
```

### Versions

- discord.js 14.13.0
- Node.js 18.17.1
- TypeScript 5.2.2
- Windows 10 Pro N 22H2 Build 19045.3086 with Experience 1000.19041.1000.0

- eslint 8.48.0
- @types/node 18.17.1
- @typescript-eslint/eslint-plugin 6.6.0
- @typescript-eslint/parser 6.6.0

### Issue priority

Low (slightly annoying)

GuildMember

GuildMembers

0d40855be.0

janparisek:
Add `toString()` method to `PartialDMChannel`, `PartialGuildMember` and `PartialUser` in typings.
The following Partials do not support a `toString()` method and thus remain unchanged:
- `PartialMessage`, since `<PartialMessage>.content` is always null
- `PartialMessageReaction`, since `MessageReaction` has no `toString()` method
- `PartialThreadMember`, since `ThreadMember` has no `toString()` method

Fixes #9833 

### DIFF
--- a/packages/discord.js/typings/index.d.ts
+++ b/packages/discord.js/typings/index.d.ts
@@ -6115,11 +6115,15 @@ export type Partialize<
   [K in keyof Omit<T, 'client' | 'id' | 'partial' | E>]: K extends N ? null : K extends M ? T[K] | null : T[K];
 };
 
-export interface PartialDMChannel extends Partialize<DMChannel, null, null, 'lastMessageId'> {
+export interface PartialDMChannel extends Partialize<DMChannel, null, null, 'lastMessageId' | 'toString'> {
   lastMessageId: undefined;
+  toString(): UserMention;
 }
 
-export interface PartialGuildMember extends Partialize<GuildMember, 'joinedAt' | 'joinedTimestamp' | 'pending'> {}
+export interface PartialGuildMember
+  extends Partialize<GuildMember, 'joinedAt' | 'joinedTimestamp' | 'pending', null, 'toString'> {
+  toString(): UserMention;
+}
 
 export interface PartialMessage
   extends Partialize<Message, 'type' | 'system' | 'pinned' | 'tts', 'content' | 'cleanContent' | 'author'> {}
@@ -6149,7 +6153,9 @@ export enum Partials {
   ThreadMember,
 }
 
-export interface PartialUser extends Partialize<User, 'username' | 'tag' | 'discriminator'> {}
+export interface PartialUser extends Partialize<User, 'username' | 'tag' | 'discriminator', null, 'toString'> {
+  toString(): UserMention;
+}
 
 export type PresenceStatusData = ClientPresenceStatus | 'invisible';
 


Jiralite (commented):
Unsure if I'm being too pedantic, but I would type them as `ReturnType<Structure["toString"]>` since the type will always be the same as the type it comes from.

Would it be possible to add tests too?

Jiralite:
Just realised that another solution could be this:

```TypeScript
export type Partialize<
  T extends AllowedPartial,
  N extends keyof T | null = null,
  M extends keyof T | null = null,
  E extends keyof T | '' = '',
> = {
  [K in keyof Omit<T, E>]: K extends 'partial' ? true : K extends N ? null : K extends M ? T[K] | null : T[K];
};
```

`client` and `id` were always typed as `Client<true>` and `Snowflake` respectfully so these could be removed, leaving a simple `K extends 'partial' ? true : ...`. Since there's no intersection with an object, `toString()` should be fine now. Also fixes the issue where `PartialMessageReaction` had `id` typed when it has no id.

janparisek:
Aight, I'm currently working on it.

TODO:
- [x] Utilize `ReturnType<Structure['toString']>`
- [x] Write tests
- [ ] Refactoring `Partialize`

janparisek:
Regarding the tests: I was not sure how many tests would be sufficient and at what point to draw the line. Let me know if I should add checks for additional keys.

packages/discord.js/typings/index.d.ts
```diff
@@ -6107,22 +6107,28 @@ export interface PartialChannelData {
...
-} & {
-  [K in keyof Omit<T, 'client' | 'id' | 'partial' | E>]: K extends N ? null : K extends M ? T[K] | null : T[K];
+  [K in keyof Omit<T, OverridableKeys>]: K extends 'partial'
+    ? true
+    : K extends NulledKeys
+    ? null
+    : K extends NullableKeys
+    ? T[K] | null
+    : T[K];
 };
```
Jiralite, line 6121: When I said "another solution", I didn't mean to apply it to the existing solution. I just proposed an alternative way of doing it, so it's one or the other. If you apply the solution I proposed, all the changes below can be reverted as they are unnecessary.

Jiralite (approved):

vladfrangu (approved):

didinele (approved):

PR #9864, opened 2023-10-01
feat: expose Retry-After and sublimit timeouts in RatelimitData

vladfrangu:

Expose `Retry-After` header in RatelimitData/RatelimitError, as well as sublimit timeouts, if any are present, which can be used in cases to report to users they'll need to wait (think for the /guilds/:id/emojis route and its ratelimit)

Also brings RatelimitData in sync with what we log for the debug event (unintentional enhancement)


packages/rest/src/lib/utils/types.ts
```diff
@@ -155,10 +155,20 @@ export interface RateLimitData {
...
 	/**
 	 * The route being hit in this request
 	 */
 	route: string;
+	/**
+	 * The time, in milliseconds, that will need to pass before this request can be retried
+	 *
+	 * This is only present on certain sublimits, and `0` otherwise
+	 */
+	sublimitTimeout: number;
```
ckohen, line 171: why `0` and not null for non-existent?

packages/rest/src/lib/utils/types.ts
```diff
@@ -155,10 +155,20 @@ export interface RateLimitData {
...
+	/**
+	 * The time, in milliseconds, that will need to pass before this request can be retried
+	 */
+	retryAfter: number;
 	/**
 	 * The route being hit in this request
 	 */
 	route: string;
+	/**
+	 * The time, in milliseconds, that will need to pass before this request can be retried
```
ckohen, line 167:
```suggestion
	 * The time, in milliseconds, that will need to pass before this specific request can be retried
```

vladfrangu: This is very much redundant to add 👀

packages/rest/src/lib/utils/types.ts
vladfrangu: Mostly for consistency (and because it will always be present when called during a 429 encounter, but never present before the request gets sent because sublimits are already handled)

vladfrangu: We could even make it be same as retryAfter in pre-request onRatelimited scenario (which then would make it equal to post-429 scenario)

packages/rest/src/lib/utils/types.ts
ckohen: you need to somehow identify the difference between this, `retryAfter`, and `timeToReset`
`timeToReset` is pretty good already, but `retryAfter` and `sublimitTimeout` are not the same. A `sublimitTimeout` is guaranteed to be per-request, whereas `retryAfter` could be for the whole queue or per-request. In fact, I'd say we should change to `isSublimit: boolean`, since it'll either be the same value as `retryAfter` or not sublimited

vladfrangu: I just added all fields that also get logged in the debug message. In terms of pretty good already, for a 429 on the emojis route i mentioned, `timeToReset` was way wrong (always at around 120) while `retryAfter` and sublimit one were the correct, very long number.

ckohen: because as mentioned, the request-lock isn't tied to `retryAfter`. This is expected behavior; the number isn't wrong, it's the wrong number. On 429's, the current request is stalled, which will *technically* stall the whole queue. On encountering a sublimit, it *does not* stall the whole queue, because the next request is allowed to pass (if we can identify why we hit the sublimit at least)

packages/rest/src/lib/utils/types.ts
ckohen: leaving as is for this. Further sublimited events *should* emit with this (and possibly emit at queue time), but currently do not. Will need to be fixed in a follwup PR later.

Jiralite (approved):

packages/rest/src/lib/utils/types.ts
```diff
@@ -155,12 +155,23 @@ export interface RateLimitData {
...
 	/**
-	 * The time, in milliseconds, until the request-lock is reset
+	 * The time, in milliseconds, that will need to pass before the sublimit lock for the route resets, and routes that fall under a sublimit
+	 * can be retried
+	 *
+	 * This is only present on certain sublimits, and `0` otherwise
+	 */
+	sublimitTimeout: number;
+	/**
+	 * The time, in milliseconds, until the global route request-lock is reset
```
ckohen, line 174:
```suggestion
	 * The time, in milliseconds, until this routes request-lock is reset
```
Don't love global here cause it will probably make people think of the 50/s global limit

packages/rest/src/lib/utils/types.ts
```diff
@@ -155,12 +155,23 @@ export interface RateLimitData {
...
+	 * The time, in milliseconds, that will need to pass before this specific request can be retried
+	 */
+	retryAfter: number;
 	/**
 	 * The route being hit in this request
 	 */
 	route: string;
 	/**
-	 * The time, in milliseconds, until the request-lock is reset
+	 * The time, in milliseconds, that will need to pass before the sublimit lock for the route resets, and routes that fall under a sublimit
```
ckohen, line 167:
```suggestion
	 * The time, in milliseconds, that will need to pass before the sublimit lock for the route resets, and requests that fall under a sublimit
```

ckohen (changes requested):
So close

packages/rest/src/lib/utils/types.ts
kyranet: Alternative spelling:
```suggestion
	 * The time in milliseconds until this route's request lock is reset
```

kyranet (approved):

ckohen (approved):

PR #9867, opened 2023-10-03
feat: default select menu values

jaw0r3k:
Q: Should discord.js also allow to pass structures (overwrite that like in case of setting embed colors)?

Adds:
- https://github.com/discord/discord-api-docs/pull/6442
- https://github.com/discordjs/discord-api-types/pull/824

Adds methods to every auto populated select builder like
- `addDefaultUsers`
- or `setDefaultRoles`

Explicitly Mentionable Select has 
setDefaultValues as it supports roles and users


Jiralite (commented):
Instead of adding the same two methods to four classes, perhaps a base automatically-populated select menu class should be made for them to inherit from? So, those named methods could be called `setDefaultValues()`, as that's the property they are updating.

jaw0r3k:
And then force users to use full objects?

jaw0r3k:
A bit complicated as ts confuses to merge types together as they are split on api types
image

packages/discord.js/typings/index.d.ts
```diff
@@ -586,9 +586,6 @@ export abstract class CommandInteraction<Cached extends CacheType = CacheType> e
     option: APIApplicationCommandOption,
     resolved: APIApplicationCommandInteractionData['resolved'],
   ): CommandInteractionOption<Cached>;
-  private transformResolved(
```
vladfrangu, line 589: lets export this method still, as it was technically exported before

vladfrangu (changes requested):

packages/discord.js/typings/index.d.ts
Jiralite: Resolved with 420288d90e94f218f780323bd85930d20b6132e3.

vladfrangu (approved):

SpaceEEC (approved):

packages/builders/src/components/selectMenu/MentionableSelectMenu.ts
```diff
@@ -31,4 +38,55 @@ export class MentionableSelectMenuBuilder extends BaseSelectMenuBuilder<APIMenti
...
+		this.data.default_values.push(...normalizedValues);
+		return this;
+	}
+
+	/**
+	 * Sets default values to this auto populated select menu.
+	 *
+	 * @param values - The values to set
+	 */
+	public setDefaultValues(
```
kyranet, line 81: Why is there no `addDefaultValues`?

packages/builders/src/components/selectMenu/ChannelSelectMenu.ts
```diff
@@ -59,6 +64,39 @@ export class ChannelSelectMenuBuilder extends BaseSelectMenuBuilder<APIChannelSe
...
+	 */
+	public addDefaultChannels(...channels: RestOrArray<Snowflake>) {
+		const normalizedValues = normalizeArray(channels).map((id) => ({
+			id,
+			type: SelectMenuDefaultValueType.Channel as const,
+		}));
+
+		this.data.default_values ??= [];
+		optionsLengthValidator.parse(this.data.default_values.length + normalizedValues.length);
+		this.data.default_values.push(...normalizedValues);
```
kyranet, line 80: Changing the operation order to not map or ensure `default_values` will reduce overhead in the case of the validator throwing an error.

- normalizeArray → validate length → ensure array → map values → push to array

kyranet (dismissed):

packages/builders/src/components/selectMenu/MentionableSelectMenu.ts
Jiralite: Resolved with 1126cd91c4fe9d7910a364a593913de7d87b78fe.

packages/discord.js/typings/index.d.ts
```diff
@@ -6363,6 +6365,12 @@ export interface StageInstanceEditOptions {
   privacyLevel?: StageInstancePrivacyLevel;
 }
 
+export interface SupportingInteractionResolvedData {
+  client: Client;
+  guild?: Guild;
+  channel?: Channel;
```
vladfrangu, line 6371: it needs to be a channel with messages in it

vladfrangu (changes requested):

packages/builders/src/components/selectMenu/ChannelSelectMenu.ts
Jiralite: Should be done with fa4833ceec160487b6ea423548fdac317874f7f1.

packages/discord.js/typings/index.d.ts
Jiralite: Done with de41e9540f9cead1708dd1ed2aab741a6a132aef.

vladfrangu (approved):

kyranet (approved):

PR #9888, opened 2023-10-15
feat(User): add `avatarDecorationData`

almeidx:
Upstream:
- https://github.com/discord/discord-api-docs/pull/6464


didinele (approved):

packages/rest/src/lib/CDN.ts
```diff
@@ -108,8 +117,18 @@ export class CDN {
 		userId: string,
 		userAvatarDecoration: string,
 		options?: Readonly<BaseImageURLOptions>,
+	): string;
+
+	public avatarDecoration(
+		userIdOrAsset: string,
+		assetOrOptions?: Readonly<ImageURLOptions> | string,
+		options?: Readonly<BaseImageURLOptions>,
```
ckohen, line 125: Is there a reason you used `BaseImageURLOptions` in the implementation but `ImageURLOptions` in the overload?

almeidx: The new overload uses the `dynamicMakeURL()` method, which takes `ImageURLOptions`

ckohen: Both overloads use `ImageURLOptions`, this is the only place `BaseImageURLOptions` is used

ckohen (approved):

packages/rest/src/lib/CDN.ts
```diff
@@ -97,9 +97,18 @@ export class CDN {
...
+	 *
+	 * @param asset - The avatar decoration hash
+	 * @param options - Optional options for the avatar decoration preset
+	 */
+	public avatarDecoration(asset: string, options?: Readonly<ImageURLOptions>): string;
+
 	/**
 	 * Generates a user avatar decoration URL.
 	 *
+	 * @deprecated This overload is deprecated. Pass an hash instead.
```
sdanialraza, line 111:
```suggestion
	 * @deprecated This overload is deprecated. Pass a hash instead.
```

Jiralite:
In discord.js, should this be checked for in `equals()`?

kyranet (approved):

packages/discord.js/typings/index.d.ts
```diff
@@ -3317,7 +3322,9 @@ export class User extends Base {
 
   public accentColor: number | null | undefined;
   public avatar: string | null;
+  /** @deprecated Use {@link avatarDecorationData} instead */
```
Jiralite, line 3325:
```suggestion
  /** @deprecated Use {@link User.avatarDecorationData} instead */
```

PR #9906, opened 2023-10-22
fix: use `super` on webhook delete

jaw0r3k:
Supersedes:
- https://github.com/discordjs/discord.js/pull/9786


Jiralite:
There is no `super`. This will not work.

Jiralite:
This is not going to work either because it will be overridden by `Webhook`'s symbol. Even if you remove it, it won't work as `super.delete` is not a function.

jaw0r3k:
oh yeah it just implements webhook...

PR #9918, opened 2023-11-02
refactor: Move `getNode`/`canEnableFFmpegOptimizations` into a lazy loaded call

### LINKED ISSUE #9859: Defer calling canEnableFFmpegOptimizations on module load

@discord.js/voice

### Issue description

On app startup, I noticed that there were calls to `canEnableFFmpegOptimizations()`, despite not actually doing anything with voice for some testing I was doing. I do include discord.js/voice code elsewhere. It seems like in https://github.com/discordjs/voice/blob/309ac8596cac422cf22e51331869e011c720124c/src/audio/TransformerGraph.ts#L169, it is calling it just when the module is require/imported. I think this could be moved elsewhere somehow to only call it the first time the nodes are actually referenced somewhere, which should improve app startup time.

### Code sample

1. bot.js:

```js
import Discord from 'discord.js';
import * as DiscordVoice from '@discord.js/voice';
```

2. Run `node --prof bot.js`

3. Run `node --prof-process isolate...log > processed.txt`:
```
 [C++ entry points]:
   ticks    cpp   total   name
    138    1.3%    JS: ~spawnSync node:internal/child_process:1109:19
    138  100.0%      JS: ~spawnSync node:child_process:831:19
    138  100.0%        JS: ~getInfo /discordbot/node_modules/prism-media/src/core/FFmpeg.js:123:17
    138  100.0%          JS: ~canEnableFFmpegOptimizations /discordbot/node_modules/@discordjs/voice/dist/index.js:2234:38
    138  100.0%            JS: ~<anonymous> /discordbot/node_modules/@discordjs/voice/dist/index.js:1:1
```

### Versions

- discord.js: 14.13.0
- @discordjs/voice: 0.16.0

### Issue priority

Low (slightly annoying)

Not applicable

Not applicable

_No response_

JMTK:
Lazy calls the functions in the root of `TransformerGraph.ts`, which are called the first time the file is imported(which is through any time **@discordjs/voice** is imported. More importantly, the call to `canEnableFFmpegOptimizations`, [which is a loop that synchronously spawns child processes](https://github.com/amishshah/prism-media/blob/4ef1d6f9f53042c085c1f68627e889003e248d77/src/core/FFmpeg.js#L123C1-L123C1) is now deferred until the first time you actually attempt to play audio. 

This improves bot startup time by 100ms for people importing @discordjs/voice:
Before:
image
After:
image

I tested this locally and can confirm audio plays correctly and the call is only called then. 

Resolves #9859 

- Patch

### DIFF
--- a/packages/voice/src/audio/TransformerGraph.ts
+++ b/packages/voice/src/audio/TransformerGraph.ts
@@ -123,7 +123,7 @@ function canEnableFFmpegOptimizations(): boolean {
 	return false;
 }
 
-function initializeNodes() : Map<StreamType, Node> {
+function initializeNodes(): Map<StreamType, Node> {
 	let nodes = new Map<StreamType, Node>();
 	for (const streamType of Object.values(StreamType)) {
 		nodes.set(streamType, new Node(streamType));
--- a/packages/voice/src/audio/TransformerGraph.ts
+++ b/packages/voice/src/audio/TransformerGraph.ts
@@ -124,7 +124,7 @@ function canEnableFFmpegOptimizations(): boolean {
 }
 
 function initializeNodes(): Map<StreamType, Node> {
-	let nodes = new Map<StreamType, Node>();
+	const nodes = new Map<StreamType, Node>();
 	for (const streamType of Object.values(StreamType)) {
 		nodes.set(streamType, new Node(streamType));
 	}
--- a/packages/voice/src/audio/TransformerGraph.ts
+++ b/packages/voice/src/audio/TransformerGraph.ts
@@ -101,8 +101,7 @@ export class Node {
 }
 
 // Create a node for each stream type
-const NODES = new Map<StreamType, Node>();
-let hasInitializedNodes = false;
+let NODES : (Map<StreamType, Node> | null);
 
 /**
  * Gets a node from its stream type.
@@ -110,16 +109,7 @@ let hasInitializedNodes = false;
  * @param type - The stream type of the target node
  */
 export function getNode(type: StreamType) {
-	if (!hasInitializedNodes) {
-		hasInitializedNodes = true;
-		for (const streamType of Object.values(StreamType)) {
-			NODES.set(streamType, new Node(streamType));
-		}
-
-		initializeNodeEdges();
-	}
-
-	const node = NODES.get(type);
+	const node = (NODES ??= initializeNodes()).get(type);
 	if (!node) throw new Error(`Node type '${type}' does not exist!`);
 	return node;
 }
@@ -133,7 +123,12 @@ function canEnableFFmpegOptimizations(): boolean {
 	return false;
 }
 
-function initializeNodeEdges() {
+function initializeNodes() : Map<StreamType, Node> {
+	let nodes = new Map<StreamType, Node>();
+	for (const streamType of Object.values(StreamType)) {
+		nodes.set(streamType, new Node(streamType));
+	}
+
 	getNode(StreamType.Raw).addEdge({
 		type: TransformerType.OpusEncoder,
 		to: getNode(StreamType.Opus),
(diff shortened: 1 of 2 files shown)


SpaceEEC (approved):
👍

vladfrangu (approved):

packages/voice/src/audio/TransformerGraph.ts
```diff
@@ -102,70 +102,28 @@ export class Node {
...
  * Gets a node from its stream type.
  *
  * @param type - The stream type of the target node
  */
 export function getNode(type: StreamType) {
+	if (!hasInitializedNodes) {
+		hasInitializedNodes = true;
+		for (const streamType of Object.values(StreamType)) {
+			NODES.set(streamType, new Node(streamType));
+		}
```
kyranet, line 117: If we rename `initializeNodeEdges` to `initializeNodes`, we can move this loop into the init to simplify this function.

kyranet (commented):
Alternative idea, what if we make `NODES` nullable and initialize using nullish assignment?

Similar to:
```typescript
const node = (NODES ??= initializeNodes()).get(type);
```

(Or splitting the expression in two lines). Obviously this change would make `initializeNodes()` return a newly constructed map.

The reasoning behind this idea is so `NODES`, being null rather than an empty map guards against the possible edge case of using an unitialized map elsewhere.

It would also remove the need for an extra variable and extra operations.

packages/voice/src/audio/TransformerGraph.ts
JMTK: I'll give it a shot!

JMTK: I pushed up the changes. I need to test this some more but I think one of the issues I had with doing something similar when I first tackled this was that `initializeNodes` will be calling `getNode` internally but won't have set the map yet. What do you think?

kyranet: In that case, call `nodes.get` rather than `getNode`.

packages/voice/src/audio/TransformerGraph.ts
```diff
@@ -101,71 +101,19 @@ export class Node {
 }
 
 // Create a node for each stream type
-const NODES = new Map<StreamType, Node>();
-for (const streamType of Object.values(StreamType)) {
-	NODES.set(streamType, new Node(streamType));
-}
+let NODES : (Map<StreamType, Node> | null);
```
kyranet, line 104:
```suggestion
let NODES: Map<StreamType, Node> | null = null;
```

kyranet (changes requested):
Needed for the ESLint check to pass:

packages/voice/src/audio/TransformerGraph.ts
```diff
@@ -101,71 +101,19 @@ export class Node {
...
+let NODES: Map<StreamType, Node> | null = null;
 
 /**
  * Gets a node from its stream type.
  *
  * @param type - The stream type of the target node
  */
 export function getNode(type: StreamType) {
-	const node = NODES.get(type);
+	const node = (NODES ??= initializeNodes()).get(type);
```
vladfrangu, line 112: This looks like it will no longer work, as it will just constantly recurse 👀

vladfrangu (changes requested):

packages/voice/src/audio/TransformerGraph.ts
kyranet: Not if it uses the local map instead of calling this helper, will also improve efficiency a little. https://github.com/discordjs/discord.js/pull/9918#discussion_r1382622692

packages/voice/src/audio/TransformerGraph.ts
JMTK: Doh 🤦 thank you I was overthinking it. Fixed and pushed. Thank you for fixing the ESLint issue!

packages/voice/src/audio/TransformerGraph.ts
JMTK: Fixed now!

Jiralite:
Please run Prettier!

JMTK:
> Please run Prettier!

Done! Didn't realize that space was still there

vladfrangu (approved):

kyranet (approved):

PR #9921, opened 2023-11-04
fix(GuildManager#fetch): inject shard id

Syjalo:

### Summary
:sparkles::wrench::floppy_disk:

This PR adds required for the Guild injection of the shard id

> _To handle more guilds in the bot_
> _We added some sharding support_
> _We imported `ShardClientUtil`_
> _To assign each guild a `shardId` too_
> _And cached them for better report_

### Walkthrough
* Import `ShardClientUtil` module to calculate shard IDs for guilds ([link](https://github.com/discordjs/discord.js/pull/9921/files?diff=unified&w=0#diff-d88269307bf05a7fe080d383536a9a0772ae1887a3c55d8178e72694c89c5ed4R9))
* Assign `data.shardId` the value of the shard ID for the guild ID ([link](https://github.com/discordjs/discord.js/pull/9921/files?diff=unified&w=0#diff-d88269307bf05a7fe080d383536a9a0772ae1887a3c55d8178e72694c89c5ed4R276))

### DIFF
--- a/packages/discord.js/src/managers/GuildManager.js
+++ b/packages/discord.js/src/managers/GuildManager.js
@@ -6,6 +6,7 @@ const { Collection } = require('@discordjs/collection');
 const { makeURLSearchParams } = require('@discordjs/rest');
 const { Routes, RouteBases } = require('discord-api-types/v10');
 const CachedManager = require('./CachedManager');
+const ShardClientUtil = require('../sharding/ShardClientUtil');
 const { Guild } = require('../structures/Guild');
 const GuildChannel = require('../structures/GuildChannel');
 const GuildEmoji = require('../structures/GuildEmoji');
@@ -272,6 +273,7 @@ class GuildManager extends CachedManager {
       const data = await this.client.rest.get(Routes.guild(id), {
         query: makeURLSearchParams({ with_counts: options.withCounts ?? true }),
       });
+      data.shardId = ShardClientUtil.shardIdForGuildId(id, this.client.options.shardCount);
       return this._add(data, options.cache);
     }
 


didinele:
What problem is this actually solving?

Syjalo:
> What problem is this actually solving?

Getting the shard in the Guild class

Jiralite:
I'm assuming this is to cater towards those without the `GUILDS` intent?

Syjalo:
Discord doesn't include the shard id in the response, discord.js injects it. And there it's missing. So currently `<Guild>.shard` returning `undefined` for the fetched guild.
https://github.com/discordjs/discord.js/blob/a9e1f6026c8d978442d67cb79b0c4545a49df6d2/packages/discord.js/src/structures/Guild.js#L134

didinele:
> Discord doesn't include the shard id in the response, discord.js injects it. And there it's missing. So currently <Guild>.shard returning undefined for the fetched guild.

This is what I asking, you tagged this as `fix` but it was completely unclear if there was an error or bug you were actually addressing; LGTM.

didinele (approved):

kyranet (approved):
I'm neutral about this change, because `Guild#shard[Id]`, while not documented as such, always meant what shard the guild came from, rather than being something we calculate.

If calculating the id of the shard is as reliable as the id of the shard the GUILD_CREATE payload came from, then we can probably look into making `shardId` a getter.

LGTM nevertheless, just wanted to note this.

vladfrangu:
> If calculating the id of the shard is as reliable as the id of the shard the GUILD_CREATE payload came from, then we can probably look into making `shardId` a getter

It'll probably be reliable for enough use cases to not cause issues.

vladfrangu (approved):

PR #9913, opened 2023-10-29
feat: link basic types to MDN/TS documentation

almeidx:
**Please describe the changes this PR makes and why it should be merged:**
Fixes #9496

**Status and versioning classification:**

PR #9932, opened 2023-11-08
fix: treat primitives as references instead of content

suneettipirneni:
This transforms primitives that are considered keywords to psuedo-referenced types in the api-extractor. This allows for downstream tasks of syntax highlighting and linking for primitives rather than it being merged with surrounding plain text.

Before api-extractor would give the following tokens for `Promise<string>`
```
[
	{
		text: "Promise"
		kind: Reference,
	},
	{
		text: "<string>",
		kind: Content
	}
]
```

Now it more accurately gives:
```
[
	{
		text: "Promise"
		kind: Reference,
	},
	{
		text: "<",
		kind: Content
	},
	{
		text: "string",
		kind: Reference
	},
	{
		text: ">",
		kind: Content
	},
]
```

This should prevent the need for a lot of the edge cases found in #9913

### DIFF
--- a/packages/api-extractor/src/generators/ExcerptBuilder.ts
+++ b/packages/api-extractor/src/generators/ExcerptBuilder.ts
@@ -147,6 +147,7 @@ export class ExcerptBuilder {
 			case ts.SyntaxKind.SymbolKeyword:
 			case ts.SyntaxKind.UnknownKeyword:
 			case ts.SyntaxKind.VoidKeyword:
+			case ts.SyntaxKind.TypeOfKeyword:
 				return true;
 			default:
 				return false;
--- a/packages/api-extractor/src/generators/ExcerptBuilder.ts
+++ b/packages/api-extractor/src/generators/ExcerptBuilder.ts
@@ -135,6 +135,24 @@ export class ExcerptBuilder {
 		return { startIndex: 0, endIndex: 0, typeParameters: [] };
 	}
 
+	public static isPrimitiveKeyword(node: ts.Node): boolean {
+		switch (node.kind) {
+			case ts.SyntaxKind.AnyKeyword:
+			case ts.SyntaxKind.BigIntKeyword:
+			case ts.SyntaxKind.BooleanKeyword:
+			case ts.SyntaxKind.NeverKeyword:
+			case ts.SyntaxKind.NumberKeyword:
+			case ts.SyntaxKind.ObjectKeyword:
+			case ts.SyntaxKind.StringKeyword:
+			case ts.SyntaxKind.SymbolKeyword:
+			case ts.SyntaxKind.UnknownKeyword:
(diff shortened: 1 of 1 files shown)


packages/api-extractor/src/generators/ExcerptBuilder.ts
```diff
@@ -135,6 +135,24 @@ export class ExcerptBuilder {
...
+			case ts.SyntaxKind.AnyKeyword:
+			case ts.SyntaxKind.BigIntKeyword:
+			case ts.SyntaxKind.BooleanKeyword:
+			case ts.SyntaxKind.NeverKeyword:
+			case ts.SyntaxKind.NumberKeyword:
+			case ts.SyntaxKind.ObjectKeyword:
+			case ts.SyntaxKind.StringKeyword:
+			case ts.SyntaxKind.SymbolKeyword:
+			case ts.SyntaxKind.UnknownKeyword:
+			case ts.SyntaxKind.VoidKeyword:
```
Qjuh, line 149: Please add
```
			case ts.SyntaxKind.NullKeyword:
			case ts.SyntaxKind.UndefinedKeyword:
```
too here
other than that looking good

packages/api-extractor/src/generators/ExcerptBuilder.ts
```diff
@@ -162,6 +183,8 @@ export class ExcerptBuilder {
 
 			if (canonicalReference) {
 				ExcerptBuilder._appendToken(excerptTokens, ExcerptTokenKind.Reference, span.prefix, canonicalReference);
+			} else if (ExcerptBuilder.isPrimitiveKeyword(span.node)) {
+				ExcerptBuilder._appendToken(excerptTokens, ExcerptTokenKind.Reference, span.prefix);
```
Qjuh, line 187: Currently the website assumes all Reference ExcerptTokens have a canonicalReference. So this would either need to now be checked for in ExcerptText.tsx or we need canonicalReferences for primite keywords

suneettipirneni: I would lean towards doing it on the website. We can assume that a reference node with a nullish canonicle reference is considered a primitive.

Qjuh: I wouldn't "assume" it, but try to look it up in the link table of #9913 and if it's not in there handle it like Content.

suneettipirneni: This is what I meant. Instead of performing a check on each content node we can narrow it down only to nodes that are both references and have no canonical reference *and only* then check it against the table of keywords to see if it will be linked or not.

Qjuh (approved):

iCrawl (approved):

PR #9937, opened 2023-11-10
chore: use descriptive type parameter names

almeidx:

### <samp>🤖 Generated by Copilot at 75ff168</samp>

This pull request renames generic type parameters in various TypeScript files to improve readability and clarity of the code. It also adds a new ESLint rule for type parameter naming conventions and updates the linting scripts and configuration accordingly. The changes affect the `ItemLink`, `resolveMembers`, `Broker`, `RPCRedis`, `ActionRowBuilder`, `ApplicationCommandOptionWithChoicesAndAutocompleteMixin`, `normalizeArray`, `Collection`, `Client`, and `Assertions` classes and functions, as well as the `typings` folder and the `docgen` package.


vladfrangu (approved):

Jiralite (approved):

suneettipirneni (approved):

packages/rest/src/lib/utils/utils.ts
```diff
@@ -33,7 +33,7 @@ function serializeSearchParam(value: unknown): string | null {
  * @param options - The options to use
  * @returns A populated URLSearchParams instance
  */
-export function makeURLSearchParams<T extends object>(options?: Readonly<T>) {
+export function makeURLSearchParams<Params extends object>(options?: Readonly<Params>) {
```
ckohen, line 36: Not going to block on this cause I don't know what else to call it, but calling this Params makes me feel like it wants to be passed some variant of URLSearchParams and not any generic (pun not intended) interface. It's also mostly for inference to work better since satisfies exists but ya know..

didinele (approved):

packages/rest/src/lib/utils/utils.ts
Qjuh: How about simply `OptionsType`?

ckohen: Sure, why not. I never really liked this generic anyways, but it does make stuff play better together, so that seems pretty clear to describe what's going on.

kyranet: Since `satisfies` is now a thing (it wasn't when I wrote it), can't we drop this generic altogether and change the input to `object`?

ckohen: Funnily enough I think I do use satisfies in places where I call this in app code rn lol. I was wondering why it wasn't that way, but I thought satisfies was out when I looked at the blame time. Guess not. I'd say just leave this for now then cause that will affect a lot of stuff very out of scope for this PR but can be removed soon (debating whether that's major or not).

kyranet: That's fine by me.

As for the breakability, I think types are always semver-patch? They are in the main package, after all.

packages/builders/src/components/Components.ts
```diff
@@ -55,21 +55,23 @@ export interface MappedComponentTypes {
 /**
  * Factory for creating components from API data.
  *
- * @typeParam T - The type of component to use
+ * @typeParam Type - The type of component to use
  * @param data - The API data to transform to a component class
  */
-export function createComponentBuilder<T extends keyof MappedComponentTypes>(
+export function createComponentBuilder<Type extends keyof MappedComponentTypes>(
```
kyranet, line 61: Wouldn't `ComponentType` be a more suitable name?

packages/builders/src/components/Components.ts
```diff
@@ -55,21 +55,23 @@ export interface MappedComponentTypes {
...
 
 /**
  * Factory for creating components from API data.
  *
- * @typeParam C - The type of component to use
+ * @typeParam Builder - The type of component to use
  * @param data - The API data to transform to a component class
  */
-export function createComponentBuilder<C extends MessageComponentBuilder | ModalComponentBuilder>(data: C): C;
+export function createComponentBuilder<Builder extends MessageComponentBuilder | ModalComponentBuilder>(
```
kyranet, line 72: Similarly here, `ComponentBuilder`.

packages/rest/src/lib/utils/utils.ts
vladfrangu: Dropping the generic isn't really sem patch... Changing its type to fix a bug is tho

packages/util/src/types.ts
```diff
@@ -1,4 +1,4 @@
 /**
  * Represents a type that may or may not be a promise
  */
-export type Awaitable<T> = PromiseLike<T> | T;
+export type Awaitable<Result> = PromiseLike<Result> | Result;
```
kyranet, line 4: Wouldn't `Value` be more appropiate? As a `Promise` wraps a `Value` inside a result.

packages/discord.js/typings/index.d.ts
```diff
@@ -720,11 +722,11 @@ export type AnyComponent =
   | APIModalComponent
   | APIActionRowComponent<APIMessageActionRowComponent | APIModalActionRowComponent>;
 
-export class Component<T extends AnyComponent = AnyComponent> {
-  public readonly data: Readonly<T>;
-  public get type(): T['type'];
-  public toJSON(): T;
-  public equals(other: this | T): boolean;
+export class Component<RawType extends AnyComponent = AnyComponent> {
```
kyranet, line 725: Perhaps `RawComponentData` (or `ComponentData`) would be more fitting?

packages/discord.js/typings/index.d.ts
```diff
@@ -6581,17 +6636,17 @@ export interface ClientApplicationInstallParams {
...
-  : T extends number | string | boolean | undefined
-  ? T
-  : T extends { toJSON(): infer R }
-  ? R
-  : T extends ReadonlyArray<infer V>
-  ? Serialized<V>[]
-  : T extends ReadonlyMap<unknown, unknown> | ReadonlySet<unknown>
+  : Value extends number | string | boolean | undefined
+  ? Value
+  : Value extends { toJSON(): infer Result }
```
kyranet, line 6643: Rather than `Result` (which is too generic and a bit hard to understand what `Result` is in the following line), maybe `JSONResult` or `ToJSONReturnValue`?

kyranet (dismissed):

kyranet (approved):

iCrawl (approved):

PR #9942, opened 2023-11-12
fix(Client): Ensure destroyed connections are not ready

### LINKED ISSUE #9941: Calling client.destroy() doesn't make client.isReady() return false

discord.js

### Issue description

1. Log in with a Client with the `login()` function.
2. Then call the `destroy()` function to this same client.
3. And check that the `isReady()` function still returns `true` even though the `client.ws.destroyed` property is `true`.

### Code sample

```typescript
const { Client, GatewayIntentBits } = require('discord.js');

const client = new Client({
	intents: [
		GatewayIntentBits.Guilds,
		GatewayIntentBits.GuildMessages,
		GatewayIntentBits.MessageContent,
		GatewayIntentBits.GuildMessageReactions,
	],
});

const main = async () => {
	await client.login();
	console.log(client.isReady());
	console.log(client.ws.destroyed);
	await client.destroy();
	console.log(client.isReady());
	console.log(client.ws.destroyed);
};

main();
```

### Versions

- discord.js@14.13.0
- node v20.9.0
- Linux cursed 6.2.0-36-generic #37~22.04.1-Ubuntu SMP PREEMPT_DYNAMIC Mon Oct  9 15:34:04 UTC 2 x86_64 x86_64 x86_64 GNU/Linux

### Issue priority

Low (slightly annoying)

No Partials

Guilds, GuildMessages, GuildMessageReactions, MessageContent

_No response_

Jiralite:
Resolves #9941.

### DIFF
--- a/packages/discord.js/src/client/Client.js
+++ b/packages/discord.js/src/client/Client.js
@@ -239,7 +239,7 @@ class Client extends BaseClient {
    * @returns {boolean}
    */
   isReady() {
-    return this.ws.destroyed === false && this.ws.status === Status.Ready;
+    return !this.ws.destroyed && this.ws.status === Status.Ready;
   }
 
   /**
--- a/packages/discord.js/src/client/Client.js
+++ b/packages/discord.js/src/client/Client.js
@@ -239,7 +239,7 @@ class Client extends BaseClient {
    * @returns {boolean}
    */
   isReady() {
-    return this.ws.status === Status.Ready;
+    return this.ws.destroyed === false && this.ws.status === Status.Ready;
   }
 
   /**


packages/discord.js/src/client/Client.js
```diff
@@ -239,7 +239,7 @@ class Client extends BaseClient {
    * @returns {boolean}
    */
   isReady() {
-    return this.ws.status === Status.Ready;
+    return this.ws.destroyed === false && this.ws.status === Status.Ready;
```
kyranet, line 242: I don't think we do `=== false` unless there's also another value with a different meaning. Negating the value should be fine:
```suggestion
    return !this.ws.destroyed && this.ws.status === Status.Ready;
```

kyranet (changes requested):

kyranet (approved):

vladfrangu (approved):

SpaceEEC (approved):
Might be nicer to have a `Destroyed` state on the WS, but well

PR #9964, opened 2023-11-13
fix: correctly import from ESM generated types

vladfrangu:

Fixes an issue where the following code wouldn't work:

```ts
import { API, Client } from '@discordjs/core';
import { REST } from '@discordjs/rest';

const rest = new REST().setToken('');
new API(rest);
```

This was due to the fact that `core` and `rest` get a "common" declaration file generated at build time which TS interprets as `CJS` regardless of the mode you are running in.

### DIFF
--- a/scripts/fixMtsImports.mjs
+++ b/scripts/fixMtsImports.mjs
@@ -1,0 +1,52 @@
+import { readdir, readFile, writeFile, opendir } from 'node:fs/promises';
+import process from 'node:process';
+
+const importPattern = process.argv[2];
+
+const files = await readdir('dist');
+
+const filesToCorrect = files.filter((file) => file.startsWith(importPattern));
+
+if (filesToCorrect.length === 0) {
+	process.exit(0);
+}
+
+if (filesToCorrect.length > 1) {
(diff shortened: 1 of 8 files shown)


scripts/fixMtsImports.mjs
```diff
@@ -0,0 +1,52 @@
...
+
+if (filesToCorrect.length > 1) {
+	console.warn(
+		`Found ${filesToCorrect.length} files to correct. This is unexpected. Please make the import pattern more specific. Received: ${importPattern}`,
+	);
+	console.warn('Files to correct:', filesToCorrect);
+	console.warn('Correcting first file only.');
+}
+
+const fileToCorrect = filesToCorrect[0];
```
kyranet, line 22: Can't we just loop thru them? It's fine if it doesn't use `Promise.all`.

kyranet (changes requested):

vladfrangu:
No longer needed as of https://github.com/discordjs/discord.js/commit/8f432400d817bc90b390ee9a07123d85c0bc3228

PR #9973, opened 2023-11-16
feat: present x-ratelimit-scope for 429s hit

vladfrangu:

Just useful for debugging purposes, it's not something we need to handle as our current ratelimit impl already takes care of it


ckohen (approved):
I'm not sure I like the scope being at the end of the debug string, but the rest of the order is kind random too, so that's out of scope (haha) for this PR

Jiralite (approved):

kyranet (approved):

PR #9974, opened 2023-11-16
refactor: Document relevant types as `@internal`

Jiralite:
Marking types as `@internal` where appropriate. Also decided to make `DataResolver` a file of functions instead of a class that had only static symbols.

### DIFF
--- a/packages/discord.js/src/index.js
+++ b/packages/discord.js/src/index.js
@@ -26,7 +26,7 @@ exports.ChannelFlagsBitField = require('./util/ChannelFlagsBitField');
 exports.Collection = require('@discordjs/collection').Collection;
 exports.Constants = require('./util/Constants');
 exports.Colors = require('./util/Colors');
-exports.DataResolver = require('./util/DataResolver');
+__exportStar(require('./util/DataResolver.js'), exports);
 exports.Events = require('./util/Events');
 exports.Formatters = require('./util/Formatters');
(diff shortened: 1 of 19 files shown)


packages/discord.js/src/index.js
```diff
@@ -26,7 +26,7 @@ exports.ChannelFlagsBitField = require('./util/ChannelFlagsBitField');
 exports.Collection = require('@discordjs/collection').Collection;
 exports.Constants = require('./util/Constants');
 exports.Colors = require('./util/Colors');
-exports.DataResolver = require('./util/DataResolver');
+__exportStar(require('./util/DataResolver.js'), exports);
```
kyranet, line 29: Isn't this a breaking change? Also, is the `.js` intended? 👀

Jiralite: This class is private. We can do what we like. We can even remove the export.

Jiralite: On that note, why is it even being exported? It's been `@private` since implementation (#1892). Shall we just remove it?

kyranet: Well, it's not like we can ask the author or the reviewers anymore about it... 😅

kyranet (approved):

packages/discord.js/src/index.js
vladfrangu: > Isn't this a breaking change (not exporting `DataResolver`)?

Yep, it is

almeidx (approved):

PR #9977, opened 2023-11-17
Fix: TypeScript thinks channel.send doesn't exist

Lev-Shapiro:
Before this PR, calling channel.send using TS would result in getting error that such function doesn't exist ( which is obviously incorrect, because it does ). This PR will change it.

### DIFF
--- a/packages/discord.js/typings/index.d.ts
+++ b/packages/discord.js/typings/index.d.ts
@@ -6472,23 +6472,29 @@ export interface LimitedCollectionOptions<Key, Value> {
   keepOverLimit?: (value: Value, key: Key, collection: LimitedCollection<Key, Value>) => boolean;
 }
 
+type ChannelWithSend = 
+  { send: (message: unknown) => Promise<Message<boolean>> } &
+  (
+    | CategoryChannel
+    | DMChannel
+    | PartialDMChannel
+    | PartialGroupDMChannel
+    | NewsChannel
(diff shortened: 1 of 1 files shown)


RedGuy12:
it does not exist on `ForumChannel` or `MediaChannel`

Lev-Shapiro:
Oh, understood, sec.

jaw0r3k:
You have to further typeguard your class, that it becomes TextChannel etc

Lev-Shapiro:
Will this solve the problem?

Jiralite:
This is not possible: `TextBasedChannel` resolves to `DMChannel | PartialDMChannel | NewsChannel | StageChannel | TextChannel | PrivateThreadChannel | PublicThreadChannel<...> | VoiceChannel`, all of which have a `send()` method. The current state of this pull request achieves nothing too. If you have an issue, you can seek help in the [Discord server](https://discord.gg/djs).

Lev-Shapiro:
But how it's invalid if there is no such a thing as channel.send in TypeScript?

Lev-Shapiro:
Oh, K, understood. Making it from instanceof would work, oops

Jiralite:
The answer to your question is just a repeat of what I said. It's invalid because it does exist and therefore this pull request archives nothing, and again, you can seek help in the Discord server where people who know the library are able to help you.

PR #9994, opened 2023-11-22
fix(website): doc links to other packages

Qjuh:

Assures links to other packages of the monorepo link to the appropriate version used by the currently viewed version.

Also removes inherited methods from the api.json, as they get correctly parsed from inheritance tree and caused issues by being duplicated in the api.json causing wrong type links and missing inheritance information.

### DIFF
--- a/apps/website/src/components/documentation/tsdoc/TSDoc.tsx
+++ b/apps/website/src/components/documentation/tsdoc/TSDoc.tsx
@@ -52,7 +52,7 @@ export function TSDoc({ item, tsdoc }: { readonly item: ApiItem; readonly tsdoc:
 
 						const declarationReference = item.getAssociatedModel()?.resolveDeclarationReference(codeDestination, item);
 						const foundItem = declarationReference?.resolvedApiItem;
-						const resolved = resolveCanonicalReference(codeDestination);
+						const resolved = resolveCanonicalReference(codeDestination, item.getAssociatedPackage());
 
 						if (!foundItem && !resolved) return null;
 
@@ -62,6 +62,12 @@ export function TSDoc({ item, tsdoc }: { readonly item: ApiItem; readonly tsdoc:
 								itemURI={resolveItemURI(foundItem ?? resolved!.item)}
 								key={idx}
 								packageName={resolved?.package ?? item.getAssociatedPackage()?.displayName.replace('@discordjs/', '')}
+								version={
+									resolved?.package
+										? // eslint-disable-next-line unicorn/better-regex
+										  item.getAssociatedPackage()?.dependencies?.[resolved.package]?.replace(/[~^]/, '')
+										: undefined
+								}
 							>
 								{linkText ?? foundItem?.displayName ?? resolved!.item.displayName}
 							</ItemLink>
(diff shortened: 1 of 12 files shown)


packages/api-extractor/src/generators/ApiModelGenerator.ts
```diff
@@ -1802,7 +1806,7 @@ export class ApiModelGenerator {
 			text: `${
 				method.scope === 'global'
 					? `export function ${method.name}(`
-					: `${method.access}${method.scope === 'static' ? ' static' : ''} ${method.name}(`
+					: `${`${method.access} ` ?? ''}${method.scope === 'static' ? 'static ' : ''}${method.name}(`
```
almeidx, line 1809:
```suggestion
					: `${method.access ? `${method.access} ` : ''}${method.scope === 'static' ? 'static ' : ''}${method.name}(`
```

apps/website/src/components/documentation/tsdoc/TSDoc.tsx
```diff
@@ -84,6 +84,12 @@ export function TSDoc({ item, tsdoc }: { readonly item: ApiItem; readonly tsdoc:
 								itemURI={resolveItemURI(foundItem ?? resolved!.item)}
 								key={idx}
 								packageName={resolved?.package ?? item.getAssociatedPackage()?.displayName.replace('@discordjs/', '')}
+								version={
+									resolved?.package
+										? // eslint-disable-next-line unicorn/better-regex
+										  item.getAssociatedPackage()?.dependencies?.[resolved.package]?.replace(/[~^]/, '')
```
almeidx, line 90: `item.getAssociatedPackage()` is called 3 times in this block, do you think it is necessary to cache the result of this?

<details>
<summary>Example</summary>

```ts
						let associatedPackage: ApiPackage | null | undefined = null;

						// eslint-disable-next-line no-inner-declarations
						function getAssociatedPackage() {
							if (associatedPackage !== null) return associatedPackage;
							return (associatedPackage = item.getAssociatedPackage());
						}
```
</details>

packages/api-extractor/src/generators/ApiModelGenerator.ts
```diff
@@ -1853,8 +1857,8 @@ export class ApiModelGenerator {
 						?.map((param) => ` * @param ${param.name} - ${this._fixLinkTags(param.description) ?? ''}\n`)
 						.join('') ?? ''
 				}${
-					method.returns?.length && !Array.isArray(method.returns[0])
-						? ` * @returns ${this._fixLinkTags(method.returns[0]!.description) ?? ''}\n`
+					method.returns?.length && !Array.isArray(method.returns[0]) && method.returns[0]?.description
+						? ` * @returns ${this._fixLinkTags(method.returns[0].description) ?? ''}`
```
almeidx, line 1861: This is already covered by the length check. Alternatively, you could get rid of the length check... or just leave it as it is
```suggestion
					method.returns?.length && !Array.isArray(method.returns[0]) && method.returns[0]!.description
						? ` * @returns ${this._fixLinkTags(method.returns[0]!.description) ?? ''}`
```

almeidx (changes requested):

apps/website/src/components/documentation/tsdoc/TSDoc.tsx
Qjuh: Might be a good idea. But if we talk about caching stuff it would be way better to cache the ApiModel/ApiPackage in a more central place, because currently the website rebuilds the Model a lot...

iCrawl (approved):

PR #9996, opened 2023-11-24
fix: InteractionWebhooknot returning a webhook

GreenHolland:
The client doesn't get defined in the property when thus line gets called:
`Object.defineProperty(this, 'client', { value: client });` 
because the InteractionWebhook class doesn't inherit client class. 

### DIFF
--- a/src/structures/InteractionWebhook.js
+++ b/src/structures/InteractionWebhook.js
@@ -1,18 +1,21 @@
 'use strict';
 
+const BaseClient = require('../client/BaseClient');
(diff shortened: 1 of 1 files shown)


SpaceEEC:
Can you explain what is wrong, how you determined that, and how this PR solves this?

I can't follow what is happening here. 😅

GreenHolland:
The InteractionWebhook function works as intended, if one pays attention to the docs and calls the send function instead of followUp for example on the returned object.

PR #10000, opened 2023-11-25
refactor(ThreadMemberManager): #remove accepts UserResolvable

Renegade334:

Adds the ability to pass a resolvable to `ThreadMemberManager#remove`, in keeping with other manager functions.

Note that this deliberately accepts a UserResolvable to pass to `UserManager#resolveId`, rather than a ThreadMemberResolvable to pass to `ThreadMemberManager#resolveId`. There are two reasons for this:
- `UserManager#resolveId` accepts a ThreadMember as an argument anyway;
- If the client lacks the privileged GUILD_MEMBERS intent, then the ThreadMemberManager cache is never populated, so `ThreadMemberManager#resolveId` always returns null. ([ref](https://discord.com/developers/docs/topics/gateway-events#thread-members-update))

The new argument type is a superset of the old one, and so this is suitable for semver:patch.

**Discussion:**

- Both `ThreadMemberManager#add` and `#remove` accept a reason argument, but neither API endpoint supports the X-Audit-Log-Reason header, as neither event is loggable. I've left them there, as removing the arguments would technically be a breaking change, but these should probably be removed at some point.

### DIFF
--- a/packages/discord.js/src/managers/ThreadMemberManager.js
+++ b/packages/discord.js/src/managers/ThreadMemberManager.js
@@ -103,11 +103,13 @@ class ThreadMemberManager extends CachedManager {
 
   /**
    * Remove a user from the thread.
-   * @param {Snowflake|'@me'} id The id of the member to remove
+   * @param {UserResolvable|'@me'} member The member to remove
    * @param {string} [reason] The reason for removing this member from the thread
    * @returns {Promise<Snowflake>}
    */
-  async remove(id, reason) {
+  async remove(member, reason) {
+    const id = member === '@me' ? member : this.client.users.resolveId(member);
+    if (!id) throw new DiscordjsTypeError(ErrorCodes.InvalidType, 'member', 'UserResolvable');
     await this.client.rest.delete(Routes.threadMembers(this.thread.id, id), { reason });
     return id;
   }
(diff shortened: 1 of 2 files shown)


packages/discord.js/src/managers/ThreadMemberManager.js
```diff
@@ -103,11 +103,13 @@ class ThreadMemberManager extends CachedManager {
...
    * Remove a user from the thread.
-   * @param {Snowflake|'@me'} id The id of the member to remove
+   * @param {UserResolvable|'@me'} member The member to remove
    * @param {string} [reason] The reason for removing this member from the thread
    * @returns {Promise<Snowflake>}
    */
-  async remove(id, reason) {
+  async remove(member, reason) {
+    const id = member === '@me' ? member : this.client.users.resolveId(member);
+    if (!id) throw new DiscordjsTypeError(ErrorCodes.InvalidType, 'member', 'UserResolvable');
```
Jiralite, line 112: I am pretty sure introducing an error here where one was not before is a breaking change. If an `id` is `undefined`, one would handle the error from Discord, but all of a sudden it will no longer work as discord.js will be throwing a custom one.

vladfrangu: I disagree, this brings it in line with the `add` method

Renegade334: Although versioning is a discussion for the project team, my own 2¢ is that whether this is breaking depends on which definition is used for the API standard: what's documented, or what's implemented.

Thrown error types aren't part of the API documentation, so there's nothing in the formal library specification that implies that this function _shouldn't_ throw a DiscordjsError if the function is called with the wrong signature (as per other library functions), nor that one should expect to be able to narrow-catch a DiscordAPIError if the function is called with the wrong signature, even if that's how the current implementation behaves.

iCrawl (approved):

Jiralite (approved):

PR #10001, opened 2023-11-26
feat(website): show union members of type aliases

Qjuh:

Adds a `Union Members` section on Type Alias documentation pages that references the different types the Type Alias is a union of. This section only shows if the Type Alias is a union.

Also refactored some `type`s in mainlib typings to `interface`s for better documentation purposes.

Removed the unused `RecursiveArray` type, which was identical to the `RecursiveReadonlyArray` type but never used.


apps/website/src/components/UnionMemberList.tsx
```diff
@@ -0,0 +1,22 @@
...
+						<ExcerptText
+							excerpt={new Excerpt(member, { startIndex: 0, endIndex: member.length })}
+							apiPackage={item.getAssociatedPackage()!}
+						/>
+					</span>
+				</div>
+			)),
+		[item, members],
+	);
+}
```
iCrawl, line 22: If this isn't reused, just slap it into the UnionMembersSection

apps/website/src/components/documentation/section/UnionMembersSection.tsx
```diff
@@ -0,0 +1,35 @@
...
+						<ExcerptText
+							excerpt={new Excerpt(member, { startIndex: 0, endIndex: member.length })}
+							apiPackage={item.getAssociatedPackage()!}
+						/>
+					</span>
+				</div>
+			)),
+		[item, members],
+	);
+	return (
```
iCrawl, line 30:
```suggestion
	);

	return (
```

PR #10011, opened 2023-12-01
Update index.d.ts

Abdelrahmanwalidhassan:
before
**Please describe the changes this PR makes and why it should be merged:** Added types to BaseMessageOptions to correct types of components option when using ActionRowBuilder

Code changes have been tested against the Discord API
I know how to update typings and have done so


Abdelrahmanwalidhassan:
> isn't this wrong? you just need to add the generic when building the `ActionRowBuilder` like this: `new ActionRowBuilder<ButtonBuilder>()` (I think)

yes you can

Idris1401:
your pr is invalid then ig

PR #10014, opened 2023-12-02
fix: don't mutate user provided array

imranbarbhuiya:

Calling `embed.setFields` twice will mutate the array passed in the first call. This PR fixes it.

I was going to change it in `spliceFields`only but thought this issue might be present in other builders too so doing it in `normalizeArray` makes more sense


Syjalo:
`.addFields()`, not `.setFields`
```js
const embed = new EmbedBuilder();
const fields1 = [{ name: '1', value: '1' }, { name: '2', value: '2' }, { name: '3', value: '3' }];
const fields2 = [{ name: '4', value: '4' }, { name: '5', value: '5' }, { name: '6', value: '6' }];

embed.addFields(fields1);
embed.addFields(fields2);

console.log(fields1);
/**
[
  { name: '1', value: '1' },
  { name: '2', value: '2' },
  { name: '3', value: '3' },
  { name: '4', value: '4' },
  { name: '5', value: '5' },
  { name: '6', value: '6' }
]
**/
console.log(fields2);
/**
[
  { name: '4', value: '4' },
  { name: '5', value: '5' },
  { name: '6', value: '6' }
]
**/
```

vladfrangu:
This is an incomplete PR i'd say - it needs to also clone the array in the other case

imranbarbhuiya:
> This is an incomplete PR i'd say - it needs to also clone the array in the other case

all the methods uses `normalizeArray` so ig it covers all the cases?

imranbarbhuiya:
> `.addFields()`, not `.setFields()`
> 
> ```js
> const embed = new EmbedBuilder();
> const fields1 = [{ name: '1', value: '1' }, { name: '2', value: '2' }, { name: '3', value: '3' }];
> const fields2 = [{ name: '4', value: '4' }, { name: '5', value: '5' }, { name: '6', value: '6' }];
> 
> embed.addFields(fields1);
> embed.addFields(fields2);
> 
> console.log(fields1);
> /**
> [
>   { name: '1', value: '1' },
>   { name: '2', value: '2' },
>   { name: '3', value: '3' },
>   { name: '4', value: '4' },
>   { name: '5', value: '5' },
>   { name: '6', value: '6' }
> ]
> **/
> console.log(fields2);
> /**
> [
>   { name: '4', value: '4' },
>   { name: '5', value: '5' },
>   { name: '6', value: '6' }
> ]
> **/
> ```

calling `setFields` twice or using `addFields` all has this issue but this pr should fix both cases

imranbarbhuiya:
> I asked for tests because I couldn't reproduce this via `setFields()`, by after @Syjalo's comment, I was able to reproduce with `addFields()`. Seeing the tests should be fairly obvious as to what the expected and current behaviour is!

it happens when u use `setFields` first with `a`, run again with a new value, log `a` and it's mutated. I added test for `normalizeArray` to return closed array, should I also add for embed fields?

.vscode/settings.json
```diff
@@ -9,9 +9,9 @@
 	"editor.defaultFormatter": "esbenp.prettier-vscode",
 	"editor.formatOnSave": true,
 	"editor.codeActionsOnSave": {
-		"source.organizeImports": false,
-		"source.fixAll.eslint": true,
-		"source.fixAll": true
+		"source.organizeImports": "never",
+		"source.fixAll.eslint": "explicit",
+		"source.fixAll": "explicit"
```
Jiralite, line 14: Why has this changed?

Jiralite:
> it happens when u use `setFields` first with `a`, run again with a new value, log `a` and it's mutated.

Do you mind sharing a code sample—not sure I'm getting it.

.vscode/settings.json
imranbarbhuiya: oh vscode autochanged , I didnt noticed reverting back

almeidx (changes requested):
More often than not, the output of this function is already going to be cloned (`.map()`, spread, etc.). Perhaps we should instead fix the `Embed.addFields()` method, which seems to be the only place where the value is not being cloned.

However, if you choose to keep it as it is, we should remove this clone:
https://github.com/discordjs/discord.js/blob/54453b04e561343af900f9da546c7ba0a607cca1/packages/builders/src/components/selectMenu/MentionableSelectMenu.ts#L113

packages/builders/__tests__/util.test.ts
```diff
@@ -12,3 +12,18 @@ describe('validation', () => {
...
+describe('normalizeArray', () => {
+	test('normalizes an array or array (when input is an array)', () => {
+		expect(normalizeArray([[1, 2, 3]])).toEqual([1, 2, 3]);
+	});
+
+	test('normalizes an array (when input is rest parameter)', () => {
+		expect(normalizeArray([1, 2, 3])).toEqual([1, 2, 3]);
+	});
+
+	test('always returns a clone', () => {
```
vladfrangu, line 25: Add a test case for rest too

vladfrangu (changes requested):

packages/builders/__tests__/util.test.ts
imranbarbhuiya: Rest doesnt gets cloned in normalizeArray as the original input wasn't an array but rest params

packages/builders/__tests__/util.test.ts
```diff
@@ -12,3 +12,18 @@ describe('validation', () => {
...
+		const arr = [1, 2, 3];
+		expect(normalizeArray([arr])).not.toBe(arr);
```
kyranet, line 27: I think this should check `toEqual(arr)` to make sure the output is the same as the input followed by the `not.toBe(arr)` to signal that the arrays are equal but they're not the same instance.

packages/builders/src/util/normalizeArray.ts
```diff
@@ -5,7 +5,7 @@
  * @param arr - The (possibly variadic) data to normalize
  */
 export function normalizeArray<ItemType>(arr: RestOrArray<ItemType>): ItemType[] {
-	if (Array.isArray(arr[0])) return arr[0];
+	if (Array.isArray(arr[0])) return [...arr[0]];
```
NotSugden, line 8: just wondering if it is worth updating the documentation to specify that it returns a cloned array

vladfrangu (approved):

almeidx (approved):

PR #10017, opened 2023-12-03
types: change Awaitable<void> to void

advaith1:

Currently, event listeners require a function that returns `Awaitable<void>`. However, this prevents returning a function call that returns a value for early execution (e.g. `return interaction.reply('error')`). This PR changes `=> Awaitable<void>` to `=> void`, since [void functions can return anything](<https://www.typescriptlang.org/docs/handbook/2/functions.html#assignability-of-functions>). This will remove the restriction and allow any function, bringing it in line with Node's standard EventEmitters, as well as previous discord.js behavior.


Qjuh:
`return void interaction.reply(…);` would be the correct fix for your use-case. As returning something from an event listener while it will never reach any place that uses it should not be done imho.

Qjuh:
async_event_emitter returns `Awaitable<void>` too though.

Qjuh:
Oh, my bad, I was looking at the Listener type instead of the type used in on(…) assuming it was used there…

kyranet (approved):

SpaceEEC (approved):

PR #10021, opened 2023-12-03
fix(InteractionResponses): check if ephemeral message flag is used

### LINKED ISSUE #10020: Specifying `flags: MessageFlags.Ephemeral` in interaction responses does not update the `ephemeral` property on the interaction

discord.js

### Issue description

When specifying `ephemeral: true`, discord.js updates the `ephemeral` property on the interaction:

https://github.com/discordjs/discord.js/blob/54453b04e561343af900f9da546c7ba0a607cca1/packages/discord.js/src/structures/interfaces/InteractionResponses.js#L103

However, it does not account for when `flags` are utilised (`flags: MessageFlags.Ephemeral`).

In the below code sample, uncomment `ephemeral: true` and see the logged interaction's `ephemeral` property. Then, uncomment `flags: MessageFlags.Ephemeral` and compare the property. `ephemeral: true` will be logged in the first instance and `ephemeral: false` will be logged in the second instance.

### Code sample

```typescript
import { Client, Events, MessageFlags } from "discord.js";

const client = new Client({ intents: 0 });

client.on(Events.InteractionCreate, async (interaction) => {
	if (!interaction.isChatInputCommand()) return;

	await interaction.reply({
		content: "Hello world!",
		// ephemeral: true,
		// flags: MessageFlags.Ephemeral,
	});

	console.log(interaction);
});

void client.login();
```

### Versions

discord.js 54453b04e561343af900f9da546c7ba0a607cca1

### Issue priority

Low (slightly annoying)

Not applicable

Not applicable

54453b04e561343af900f9da546c7ba0a607cca1

sdanialraza:
Resolves #10020


vladfrangu (changes requested):
deferReply needs this treatment too

packages/discord.js/src/structures/interfaces/InteractionResponses.js
```diff
@@ -100,7 +101,7 @@ class InteractionResponses {
    */
   async reply(options) {
     if (this.deferred || this.replied) throw new DiscordjsError(ErrorCodes.InteractionAlreadyReplied);
-    this.ephemeral = options.ephemeral ?? false;
+    this.ephemeral = options.ephemeral ?? new MessageFlagsBitField(options.flags).has(MessageFlags.Ephemeral);
```
didinele, line 104: mmmmh, this is already fully computed here:
https://github.com/discordjs/discord.js/blob/67360a9e0bae361d2465379cc754554a791a1b9c/packages/discord.js/src/structures/MessagePayload.js#L149-L164

I'd love to avoid the double instantiation of `MessageFlagsBitFields`, but that doesn't seem easy in this case. At most though, I feel like we should remove this line all-together (since `MessagePayload` checks for `options.ephemeral` anyways, see line 162) and instead just do `this.ephemeral = new MessageFlagsBitField(body.flags).has(MessageFlags.Ephemeral);` just a bit later to avoid duplicating any logic.

didinele (changes requested):

didinele (changes requested):
You haven't addressed Vlad's comment

vladfrangu:
> You haven't addressed Vlad's comment

Nothing to address haha, iirc deferReply doesnt support all flags

vladfrangu (approved):

sdanialraza:
Yeah, it was addressed internally, there's no flags property for options in deferReply

didinele (approved):

Jiralite (approved):

PR #10044, opened 2023-12-10
refactor: use interfaces for AsyncEventEmitter event maps

Qjuh:

Turn the mapped types for events in `rest/REST`, `ws/WebSocketManager,` `ws/WebSocketShard` and `core/Client` into interfaces for better documentation purposes.

Also turns `WebSocketManagerOptions` into an interface for the same reason.

Some interfaces are simple aliases of already existing interfaces, to keep the named export and make this possible as a non-breaking change even in TS land.

`@discordjs/brokers` still uses types for event maps, but they are dynamic so not as easily adjustable.

### DIFF
--- a/packages/ws/src/ws/WebSocketManager.ts
+++ b/packages/ws/src/ws/WebSocketManager.ts
@@ -9,11 +9,13 @@ import {
 	type RESTGetAPIGatewayBotResult,
 	type GatewayIntentBits,
 	type GatewaySendPayload,
+	type GatewayDispatchPayload,
+	type GatewayReadyDispatchData,
 } from 'discord-api-types/v10';
 import type { IShardingStrategy } from '../strategies/sharding/IShardingStrategy.js';
 import type { IIdentifyThrottler } from '../throttling/IIdentifyThrottler.js';
 import { DefaultWebSocketManagerOptions, type CompressionMethod, type Encoding } from '../utils/constants.js';
-import type { WebSocketShardDestroyOptions, WebSocketShardEventsMap } from './WebSocketShard.js';
+import type { WebSocketShardDestroyOptions, WebSocketShardEvents } from './WebSocketShard.js';
 
 /**
  * Represents a range of shard ids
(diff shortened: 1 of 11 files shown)


vladfrangu (approved):

packages/ws/src/ws/WebSocketManager.ts
```diff
@@ -178,13 +180,24 @@ export interface OptionalWebSocketManagerOptions {
 	version: string;
 }
 
-export type WebSocketManagerOptions = OptionalWebSocketManagerOptions & RequiredWebSocketManagerOptions;
+export interface WebSocketManagerOptions extends OptionalWebSocketManagerOptions, RequiredWebSocketManagerOptions {}
 
-export type ManagerShardEventsMap = {
```
didinele, line 183: Honestly, I prefer this be kept dynamic over the docs benefit here - esp. considering this isn't that important of a type docs-wise. I've broken things before when it comes to those EE payloads, this removed safety net doesn't help.

didinele (changes requested):

packages/ws/src/ws/WebSocketManager.ts
vladfrangu: What safety net? Outside the fact we now need to manually keep these two in sync (which iirc is already needed at the code layer)

Also this type can definitely be used for documentation purposes, even if it's not crystal clear what it represents (although most can _hopefully_ infer that its `eventName: [event argument]`

didinele: Doesn't AEE make it so events not in the map type can just be fired with `any` args? If so, the sync isn't type-required at the code level, this is also trivial to miss in code review.

vladfrangu: I don't think I allow emitting random events, but tbf I haven't tested that. Lets see

vladfrangu: Code - Insiders - 2023-12-10 at 20 53 27

So, yeah, shouldn't be an issue

didinele: sure, whatever then

didinele (approved):

ckohen (changes requested):
Unless you had a specific reason in mind for not doing this, please also change the generics in extends to the initial interface so the empty extends can be removed easily in the next major.

packages/ws/src/ws/WebSocketManager.ts
Qjuh: Added a type test to ensure those two event maps keep in sync, as suggested by @almeidx .

ckohen (approved):

PR #10045, opened 2023-12-10
types: use readonly array / collection types for user input

almeidx:
Added an ESLint rule to enforce the use of readonly Array and Collection types in function/method parameters and interface properties

### DIFF
--- a/packages/discord.js/typings/index.d.ts
+++ b/packages/discord.js/typings/index.d.ts
@@ -2609,7 +2609,7 @@ export class ReactionCollector extends Collector<Snowflake | string, MessageReac
     event: 'end',
     listener: (collected: ReadonlyCollection<Snowflake, MessageReaction>, reason: string) => void,
   ): this;
-  public on(event: string, listener: (...args: readonly any[]) => void): this;
+  public on(event: string, listener: (...args: any[]) => void): this;
 
   public once(
     event: 'collect' | 'dispose' | 'remove' | 'ignore',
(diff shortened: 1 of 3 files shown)


packages/discord.js/typings/index.d.ts
```diff
@@ -1008,7 +1016,7 @@ export class Client<Ready extends boolean = boolean> extends BaseClient {
   ): this;
   public on<Event extends string | symbol>(
     event: Exclude<Event, keyof ClientEvents>,
-    listener: (...args: any[]) => Awaitable<void>,
+    listener: (...args: readonly any[]) => Awaitable<void>,
```
vladfrangu, line 1019: Can you check this one specific change won't break anything? It also diverts from the types in node and AEE

Jiralite:
Is there one for `const` type parameters?

packages/discord.js/typings/index.d.ts
almeidx: I'll just make an exception for `on` and `once` methods as to keep it the same as node's types

kyranet (approved):

Jiralite (approved):

vladfrangu (approved):
Hopefully this won't break user code

vladfrangu:
Also this needs a rebase

ThinkThoughtThunk:
This is a breaking change and should have been saved for a new major version number. This ended up breaking my codebase everywhere I was directly referencing one of the "user input" message / response types since I wasn't explicitly declaring the data objects as `readonly`.

PR #10049, opened 2023-12-13
fix(Webhook): check whether user is null during patching

rockerbacon:
While fetching a webhook using the `Client#fetchWebhook` method, it is possible for the Discord API to send a response with some properties set to null and others not defined.
While creating a new instance of the Webhook class, the _patch method will check whether or not a property was returned using the `in` operator. An issue then arises when the property "user" is null: the check `'user' in data` evaluates to `true` (the property is indeed in data, it's just set to the value "null") and a null value is then sent to `CachedManager#_add`, which attempts to access the property `id` in a null value.

The following is the relevant stack trace of the error I encountered:
```
TypeError: Cannot read properties of null (reading 'id')
    at UserManager._add (/home/loka/Documents/discord-bots/greatreads-discord-bot/node_modules/discord.js/src/managers/CachedManager.js:47:48)
    at Webhook._patch (/home/loka/Documents/discord-bots/greatreads-discord-bot/node_modules/discord.js/src/structures/Webhook.js:91:39)
    at new Webhook (/home/loka/Documents/discord-bots/greatreads-discord-bot/node_modules/discord.js/src/structures/Webhook.js:25:20)
    at Client.fetchWebhook (/home/loka/Documents/discord-bots/greatreads-discord-bot/node_modules/discord.js/src/client/Client.js:313:12)
```
And this is the offending Discord API response, obtained by logging the immediate response ([Client.js ln312](https://github.com/discordjs/discord.js/blob/f2138bb5a85d55be1c5419da607c085c368ece96/packages/discord.js/src/client/Client.js#L312)) to the console:
```
{
  data: {
    application_id: '<redacted>',
    avatar: null,
    channel_id: null,
    guild_id: null,
    id: '1184125670163488828',
    name: '<redacted>',
    type: 3,
    user: null
  }
}
```

### DIFF
--- a/packages/discord.js/src/structures/Webhook.js
+++ b/packages/discord.js/src/structures/Webhook.js
@@ -83,7 +83,7 @@ class Webhook {
       this.channelId = data.channel_id;
     }
 
-    if ('user' in data) {
+    if ('user' in data && data.user !== null) {
       /**
        * The owner of the webhook
        * @type {?(User|APIUser)}


Jiralite:
Pretty sure this is just a bug with Discord's API. The user of a webhook is [not documented](https://discord.com/developers/docs/resources/webhook#webhook-object-webhook-structure) to be `null`...

So, in what case are you seeing this?

rockerbacon:
Thanks for the quick response.

If it is of any relevance: my use case is editing a message sent by my own bot at the end of an OAuth2 authorization process. Since the authorization system is designed to be a separate distributed system, our strategy was to store the message id, webhook id and webhook token in a central database and retrieve them later.
I worked around the issue by instantiating a WebhookClient directly, as documented in the discord.js guide.

Let me know if you'd require any further assistance with the issue.

Qjuh:
You could also just instantiate an InteractionWebhook instance directly, no need to fetch it or make another Client (which your WebhookClient is) if you already have a logged in Client.

Jiralite:
Looks like we got an official response: https://github.com/discord/discord-api-docs/issues/6576#issuecomment-1855851412

Fetching these kinds of webhooks should return a 404, which means we shouldn't be encountering a `null` `user`. I'll be closing this then.

PR #10050, opened 2023-12-14
SetID Method for EmbedBuilder

Bluley:
## Changes Made:
This Pull Request introduces a new feature to the EmbedBuilder in the Discord.js library. A setId method has been added, enabling the assignment of a unique identification ID to an embed. This feature is particularly useful for identifying and managing embeds more effectively in complex applications.

## Reason for Merging:
The inclusion of the setId method enhances the capabilities of EmbedBuilder, providing developers with greater flexibility and control over the creation and management of embeds. This can be particularly beneficial in scenarios where multiple embeds are used and require individual identification.

## Code changes have been tested against the Discord API.
~~Typings do not need updating or have been updated.~~
This PR includes changes to the library's interface (methods or parameters added).
There are no significant breaking changes (methods removed or renamed, parameters moved or removed).

## Technical Details:
The main modification is in the setID method, which stores a user-provided ID within the EmbedBuilder object. This ID can later be used for referencing or manipulating the embed. Additionally, the toJSON method has been updated to include this new ID field when an embed is converted to JSON. This functionality does not alter the existing functionality of EmbedBuilder but adds a new capability.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -9,9 +9,9 @@
 	"editor.defaultFormatter": "esbenp.prettier-vscode",
 	"editor.formatOnSave": true,
 	"editor.codeActionsOnSave": {
-		"source.organizeImports": false,
-		"source.fixAll.eslint": true,
-		"source.fixAll": true
+		"source.organizeImports": "never",
+		"source.fixAll.eslint": "explicit",
+		"source.fixAll": "explicit"
 	},
 	"editor.trimAutoWhitespace": false,
 	"files.associations": {
--- a/packages/discord.js/src/structures/EmbedBuilder.js
+++ b/packages/discord.js/src/structures/EmbedBuilder.js
@@ -12,6 +12,7 @@ const { resolveColor } = require('../util/Util');
 class EmbedBuilder extends BuildersEmbed {
   constructor(data) {
     super(toSnakeCase(data));
+    this._id = null; //aqui inicializamos o valor como null
   }
 
   /**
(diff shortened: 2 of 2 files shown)


.vscode/settings.json
```diff
@@ -9,9 +9,9 @@
 	"editor.defaultFormatter": "esbenp.prettier-vscode",
 	"editor.formatOnSave": true,
 	"editor.codeActionsOnSave": {
-		"source.organizeImports": false,
-		"source.fixAll.eslint": true,
-		"source.fixAll": true
+		"source.organizeImports": "never",
+		"source.fixAll.eslint": "explicit",
+		"source.fixAll": "explicit"
```
almostSouji, line 14: if you make local changes to your configuration files, please revert them before committing your PR.

packages/discord.js/src/structures/EmbedBuilder.js
```diff
@@ -23,6 +24,13 @@ class EmbedBuilder extends BuildersEmbed {
     return super.setColor(color && resolveColor(color));
   }
 
+  setID(id){
```
almostSouji, line 27: - needs docs string

packages/discord.js/src/structures/EmbedBuilder.js
```diff
@@ -40,6 +48,14 @@ class EmbedBuilder extends BuildersEmbed {
   get length() {
     return embedLength(this.data);
   }
+  toJSON() {
```
almostSouji, line 51: - empty line between methods
- requires docs string

Syjalo:
Just make an array of your embed ids and when you're done map it to your embeds. No need to add this to the library.

Qjuh:
The EmbedBuilder is for building JSON of an embed that then can be sent to discord’s API. Your PR adds a property to that JSON that the API doesn’t expect/take. That alone makes your PR invalid here.

Even if it was a valid approach (which it is not) you use two different properties, one named `_id` and one named `id` which makes it clear you didn’t even test your PR.

almostSouji (changes requested):
nevermind, not considering the questionable use, this isn't even correct or, as qjuh pointed out already, tested

PR #10059, opened 2023-12-28
docs: remove utf-8-validate

Zamiell:
utf-8-validate is only for legacy versions of Node.js, so having it in the readme like this is a footgun and will cause the ecosystem to install useless dependencies.


vladfrangu:
> utf-8-validate is only for legacy versions of Node.js

Where does it say that?

But you are right in the sense this isn't needed for node >=18.14, so we can consider removing this from our documentation

Zamiell:
> Where does it say that?

it says it on [the readme](https://github.com/websockets/ws/blob/master/README.md#legacy-opt-in-for-performance)

kyranet (approved):

didinele (approved):

vladfrangu (approved):

PR #10063, opened 2023-12-29
feat: add support for `using` keyword on discord.js `Client` and `WebSocketManager`

### LINKED ISSUE #10057: BaseClient/Client: Add support for 'using' keyword (Explicit Resource Management)

discord.js

### Feature

Right now, as I understand it, the documented way to close out a [`Client`](https://discord.js.org/docs/packages/discord.js/main/Client:Class) after it's done is [`destroy()`](https://discord.js.org/docs/packages/discord.js/main/Client:Class#destroy):

```ts
const client = new Client({ intents: [/* ... */] });

await client.login(/* ... */);

// ... (use the client) ...

await client.destroy();
```

One gotcha is that if the `( use the client )` bit throws, the `await client.destroy()` won't run - unless you `try`/`catch` or `try`/(`catch`/)`finally`). This is a common gotcha in async code.

[Explicit Resource Management](https://github.com/tc39/proposal-explicit-resource-management) is a TC39 proposal to add explicit syntax for situations where a resource should have some method called upon disposal. It's stage 3 now and [supported in TypeScript >=5.2](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-5-2.html#using-declarations-and-explicit-resource-management).

Proposal: can the `BaseClient` class be given a `[Symbol.asyncDispose]` method that calls `this.destroy()`?

```ts
class BaseClient<Ready extends boolean = boolean> {
  // ... (existing class fields) ...

  async [Symbol.asyncDispose]() {
    await this.destroy();
  }
}
```

That way we'd be able to use clients like:

```ts
await using client = new Client({ intents: [/* ... */] });
```

...and be more assured that the `client`'s `.destroy()` will be called and awaited when done.

For now, we can somewhat shim this in with:

```ts
class DisposableClient<Ready extends boolean = boolean> extends Client<Ready> {
	async [Symbol.asyncDispose]() {
		await this.destroy();
	}
}
```

```ts
await using client = new DisposableClient({
    intents: [GatewayIntentBits.Guilds],
});
```

### Other context

I made a small standalone package containing this suggested `DisposableClient`: https://github.com/JoshuaKGoldberg/disposable-discord-client -> https://www.npmjs.com/package/disposable-discord-client

suneettipirneni:

This adds support for [explicit resource management](https://github.com/tc39/proposal-explicit-resource-management) for the `Client` instances in discord.js, as well as the `WebSocketManager` in `ws`. This is done by polyfilling `Symbol.dispose` and `Symbol.asyncDispose` (if not already provided by the runtime).

Users who support the correct TypeScript project configuration can use the following syntax to clean up resources after the client goes out of scope or if it errors out:

```ts
await using client = new Client(...); 
```

Or if you're using the `WebSocketManager` from `@discordjs/ws` directly:

```ts
await using ws = new WebSocketManager(...);
```

Closes #10057


packages/util/src/functions/polyfillDispose.ts
```diff
@@ -0,0 +1,14 @@
+/**
+ * Polyfill for `Symbol.dispose` and `Symbol.asyncDispose` which is used as a part of
+ * {@link https://github.com/tc39/proposal-explicit-resource-management}. Node versions below 18.x
+ * don't have these symbols by default, so we need to polyfill them.
+ */
+export function polyfillDispose() {
```
didinele, line 6: Isn't this wrong? If there's a different tool polyfilling this and then discord.js is imported, we overwrite the other polyfill with a completely different symbol - pain to debug.

didinele (changes requested):

packages/util/src/functions/polyfillDispose.ts
suneettipirneni: Discord.js wouldn't write polyfill in that situation because Symbol.dispose would already be defined.

didinele: Oh sorry, I completely blanked on `??=`

didinele (changes requested):
Could you maybe look into doing this in WS as well? `WebSocketManager` specifically

didinele (approved):

packages/util/src/functions/polyfillDispose.ts
```diff
@@ -0,0 +1,14 @@
...
+	// Polyfill for `Symbol.dispose` and `Symbol.asyncDispose` if not available.
+	// Taken from https://www.typescriptlang.org/docs/handbook/release-notes/typescript-5-2.html#using-declarations-and-explicit-resource-management
+
+	// @ts-expect-error This is a polyfill, so it's fine to write
+	Symbol.dispose ??= Symbol('Symbol.dispose');
```
vladfrangu, line 11: Is this the same polyfill as node does?

suneettipirneni: No, this symbol is typescript-specific. In node the symbol is defined as `Symbol.dispose = Symbol(nodejs.dispose)`. But regardless, it will always prefer to use the symbol from the runtime and only use this impl when it's not provided.

vladfrangu (approved):

SpaceEEC (approved):

packages/ws/src/ws/WebSocketManager.ts
```diff
@@ -199,7 +203,7 @@ export interface ManagerShardEventsMap {
 	];
 }
 
-export class WebSocketManager extends AsyncEventEmitter<ManagerShardEventsMap> {
+export class WebSocketManager extends AsyncEventEmitter<ManagerShardEventsMap> implements AsyncDisposable {
```
Qjuh, line 206: Should that `implements AsnycDisposable` also be added to mainlib typings of BaseClient then?

Qjuh (approved):

PR #10072, opened 2024-01-06
fix: Add name back in Emoji#toString()

### LINKED ISSUE #10067: Emoji.toString() doesn't contain the name anymore

formatters

### Issue description

I noticed that in [this commit](https://github.com/discordjs/discord.js/commit/40726db722c7851f4096cda29667ea4ee89da98b#diff-294a8b034d842f958de630a98185dd9bf0a58d688934373c8702ededb302e211) the `toString()` implementation for the Emoji class has changed, in order to use the provided formatting functions by the `formatters` package

This also means that this function will never return anything other than `<:_:emoji_id>` or `<a:_:emoji_id>`. The way I understand it, the name can actually be missing under certain circumstances, but it's available 99% of the time so I don't see why it shouldn't be included in the string representation by default. 

I originally wanted to just make a PR for this but I'm not quite sure what this is supposed to look like so I figured I should probably open an issue about it first.

### Code sample

```typescript
import Fuse from "fuse.js";

const emojis = await message.guild?.emojis.fetch();
const emojiStrings = Array.from(emojis.map((x) => x.toString());
const fuse = new Fuse(emojiStrings, {
    shouldSort: true,
    threshold: 0.3,
});
const matchedEmojis = fuse.search(searchTerm).map((x) => x.item);
const outputString = matchedEmojis.join(" ");
await message.channel.send(outputString);
```

### Versions

- discord.js 14.14.1
- node 21.5.0
- Windows 11

### Issue priority

Medium (should be fixed soon)

Channel

Guilds, GuildMembers, GuildEmojisAndStickers, GuildMessages, GuildMessageReactions, DirectMessages, MessageContent

_No response_

Anonymous4078:
Fixes #10067 

### DIFF
--- a/packages/discord.js/src/structures/Emoji.js
+++ b/packages/discord.js/src/structures/Emoji.js
@@ -1,7 +1,6 @@
 'use strict';
 
 const process = require('node:process');
-const { formatEmoji } = require('@discordjs/formatters');
 const { DiscordSnowflake } = require('@sapphire/snowflake');
 const Base = require('./Base');
 
@@ -98,7 +97,7 @@ class Emoji extends Base {
    * reaction.message.channel.send(`The emoji used was: ${reaction.emoji}`);
    */
   toString() {
-    return this.id ? formatEmoji(this.id, this.animated) : this.name;
+    return this.id ? `<${this.animated ? 'a' : ''}:${this.name}:${this.id}>` : this.name;
   }
 
   toJSON() {
--- a/packages/discord.js/src/structures/Emoji.js
+++ b/packages/discord.js/src/structures/Emoji.js
@@ -1,6 +1,7 @@
 'use strict';
 
 const process = require('node:process');
+const { formatEmoji } = require('@discordjs/formatters');
 const { DiscordSnowflake } = require('@sapphire/snowflake');
 const Base = require('./Base');
 
@@ -97,7 +98,7 @@ class Emoji extends Base {
    * reaction.message.channel.send(`The emoji used was: ${reaction.emoji}`);
    */
   toString() {
-    return this.id ? `<${this.animated ? 'a' : ''}:${this.name}:${this.id}>` : this.name;
+    return this.id ? formatEmoji(this.id, this.animated, this.name) : this.name;
   }
 
   toJSON() {
--- a/packages/formatters/src/formatters.ts
+++ b/packages/formatters/src/formatters.ts
@@ -336,11 +336,30 @@ export function formatEmoji<EmojiId extends Snowflake>(
 	animated?: boolean,
 ): `<:_:${EmojiId}>` | `<a:_:${EmojiId}>`;
 
-export function formatEmoji<EmojiId extends Snowflake>(
+/**
+ * Formats an emoji id into a fully qualified emoji identifier.
+ *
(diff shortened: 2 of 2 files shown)


jaw0r3k:
well name is not necessary, so idk if we even should do anything about that
if so then somehow change the formatEmoji method instead of making duplicated code

Anonymous4078:
> well name is not necessary, so idk if we even should do anything about that if so then somehow change the formatEmoji method instead of making duplicated code

Wouldn't it be a breaking change to add a new parameter to `formatEmoji()`?
Should I make a separate PR for that?

sdanialraza:
> Wouldn't it be a breaking change to add a new parameter to `formatEmoji()`? Should I make a separate PR for that?

Not necessarily, you'd just add another overload and use `typeof` to differentiate it from the current uses, it'd be better to overload an object moving forward.

Anonymous4078:
> > Wouldn't it be a breaking change to add a new parameter to `formatEmoji()`? Should I make a separate PR for that?
> 
> Not necessarily, you'd just add another overload and use `typeof` to differentiate it from the current uses, it'd be better to overload an object moving forward.

Created #10073 , would you mind taking a look?

sdanialraza:
> Created #10073 , would you mind taking a look?

That's not what I said, you are changing the existing overloads there and adding a parameter so that would be a breaking change. There's also a lot wrong there which would cause the tests to fail, please read [this](https://github.com/discordjs/discord.js/blob/main/.github/CONTRIBUTING.md#setup) and make sure to test things locally before opening a PR.

Anonymous4078:
@sdanialraza Sorry, is it ok now. It is my first time making PR in discord.js so I am a bit confused

packages/discord.js/src/structures/Emoji.js
```diff
@@ -98,7 +97,7 @@ class Emoji extends Base {
    * reaction.message.channel.send(`The emoji used was: ${reaction.emoji}`);
    */
   toString() {
-    return this.id ? formatEmoji(this.id, this.animated) : this.name;
```
jaw0r3k, line 101: revert it and just add name param

packages/formatters/src/formatters.ts
```diff
@@ -336,11 +336,30 @@ export function formatEmoji<EmojiId extends Snowflake>(
...
 	emojiId: EmojiId,
 	animated = false,
+	name: Name
 ): `<:_:${EmojiId}>` | `<a:_:${EmojiId}>` {
-	return `<${animated ? 'a' : ''}:_:${emojiId}>`;
+	if (name === undefined) {
+	        return `<${animated ? 'a' : ''}:_:${emojiId}>`;
+	}
+
+	return `<${animated ? 'a' : ''}:${name}:${emojiId}>`;
```
jaw0r3k, line 362:
```suggestion
	return `<${animated ? 'a' : ''}:${name ?? '_'}:${emojiId}>`;
```

Jiralite (changes requested):
I feel it would be better to utilise an object approach because the name of an emoji does not depend on whether it is animated.

jaw0r3k:
> I feel it would be better to utilise an object approach because the name of an emoji does not depend on whether it is animated.

That would be braking
Then you have to support both boolean or object

PR #10099, opened 2024-01-23
feat(REST): dynamic rate limit offsets

### LINKED ISSUE #9762: Custom REST rate limit offsets for specific routes

rest

### Feature

To help prevent ratelimits and be more safe, discord.js gives you the option to set a ratelimit offset for all routes. But, some routes are heavily used by some bots, perhaps more than other routes, so providing a one size fits all may not be ideal all the time.

For example what if you have a reactions heavy bot, that adds reactions very often, and you want to be more cautious of the frequent reactions your bot is adding and want to avoid 429s, but you don't want it to make other actions slower than preferred, like editing messages.

Perhaps the `rateLimitOffset` option in the client options could optionally be a function, similar to `rejectOnRatelimit`. This could take in the route, or extra data, whichever works best, and return the number the offset should be in this context.

The alternative is to implement your own external rate limiting, but that can become messy or repetitive depending on the bot.

### Other context

_No response_

didinele:
Closes #9762

### DIFF
--- a/packages/rest/src/lib/handlers/SequentialHandler.ts
+++ b/packages/rest/src/lib/handlers/SequentialHandler.ts
@@ -4,7 +4,7 @@ import type { REST } from '../REST.js';
 import type { IHandler } from '../interfaces/Handler.js';
 import { RESTEvents } from '../utils/constants.js';
 import type { RateLimitData, ResponseLike, HandlerRequestData, RouteData } from '../utils/types.js';
-import { hasSublimit, onRateLimit, sleep } from '../utils/utils.js';
+import { hasSublimit, normalizeRateLimitOffset, onRateLimit, sleep } from '../utils/utils.js';
 import { handleErrors, incrementInvalidCount, makeNetworkRequest } from './Shared.js';
 
 const enum QueueType {
@@ -104,8 +104,9 @@ export class SequentialHandler implements IHandler {
 	/**
 	 * The time until queued requests can continue
 	 */
-	private get timeToReset(): number {
-		return this.reset + this.manager.options.offset - Date.now();
+	private getTimeToReset(routeId: RouteData): number {
+		const offset = normalizeRateLimitOffset(this.manager.options.offset, routeId.bucketRoute);
+		return this.reset + offset - Date.now();
 	}
 
 	/**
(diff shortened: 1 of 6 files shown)


vladfrangu (approved):
Up for debate if we should maybe make GetRateLimitOffsetFunction async compat or not, I'd say nah, but otherwise LGTM

packages/rest/src/lib/utils/types.ts
```diff
@@ -191,6 +191,11 @@ export interface RateLimitData {
  */
 export type RateLimitQueueFilter = (rateLimitData: RateLimitData) => Awaitable<boolean>;
 
+/**
+ * A function that determines the ratelimit offset for a given request.
```
Jiralite, line 195:
```suggestion
 * A function that determines the rate limit offset for a given request.
```

packages/rest/src/lib/handlers/SequentialHandler.ts
```diff
@@ -209,9 +210,11 @@ export class SequentialHandler implements IHandler {
 			let delay: Promise<void>;
 
 			if (isGlobal) {
+				const offset = normalizeRateLimitOffset(this.manager.options.offset, routeId.bucketRoute);
+
 				// Set RateLimitData based on the global limit
 				limit = this.manager.options.globalRequestsPerSecond;
-				timeout = this.manager.globalReset + this.manager.options.offset - Date.now();
+				timeout = offset - Date.now();
```
ckohen, line 217:
```suggestion
				timeout = this.manager.globalReset + offset - Date.now();
```

You removed a bit too much. 👋 Global ratelimit

packages/rest/src/lib/handlers/SequentialHandler.ts
```diff
@@ -341,13 +346,15 @@ export class SequentialHandler implements IHandler {
 			let timeout: number;
 
 			if (isGlobal) {
+				const offset = normalizeRateLimitOffset(this.manager.options.offset, routeId.bucketRoute);
+
 				// Set RateLimitData based on the global limit
 				limit = this.manager.options.globalRequestsPerSecond;
-				timeout = this.manager.globalReset + this.manager.options.offset - Date.now();
+				timeout = offset - Date.now();
```
ckohen, line 353:
```suggestion
				timeout = this.manager.globalReset + offset - Date.now();
```

Ditto

ckohen (changes requested):

ckohen (approved):

Jiralite (approved):

PR #10117, opened 2024-02-04
refactor: use zod in builders

Qjuh:

Replaces `shapeshift` with `zod` and `zod-validation-error` in `@discordjs/builders`.


packages/builders/src/util/validation.ts
```diff
@@ -24,3 +27,20 @@ export function disableValidators() {
...
+ * Parses a value with a given validator
+ *
+ * @param validator - Tthe zod validator to use
+ * @param value - The value to parse
+ * @returns The result from parsing
+ * @internal
+ */
+export function parse<Validator extends ZodTypeAny>(validator: Validator, value: unknown): output<Validator> {
+	const result = validator.safeParse(value);
+	if (isValidationEnabled() && !result.success) {
```
vladfrangu, line 41: Shouldn't validationEnabled be checked before everything else? 👀

vladfrangu (dismissed):
Also, this needs a rebase

packages/builders/src/util/validation.ts
Qjuh: Afaik shapeshift doesn’t do it either, because we also parse for example nativeEnums and replace them with their enum values if the string key is passed. That’s why we need the parse result even if validation is disabled.

favna:
What is the motivation for this PR? Builders [used to use Zod](https://github.com/discordjs/discord.js/pull/7547) but there were many many complaints about it. That was after /builders initially used the `ow` library as well. It was changed to the at the time freshly made shapeshift. One of the reasons for us to even make shapeshift was improvements to /builders. 

If this is related to [confusing errors](https://github.com/discordjs/discord.js/issues/8837) I strongly recommend you instead contribute to Shapeshift by fulfilling [this issue](https://github.com/sapphiredev/shapeshift/issues/201) for which most work is already finished in [this pr](https://github.com/sapphiredev/shapeshift/pull/231) and only unit tests need to be updated. Once that issue is resolved custom messages can be assigned in /builders and get very clear errors.

Also relevant:

- https://github.com/discordjs/discord.js/pull/7067

didinele:
This has all already been discussed in internals, albeit rather scattered, see:

- https://discord.com/channels/222078108977594368/999796393810067607/1201209906112503818

there's probably more if you search for "shapeshift".

favna:
Copying my message from Discord to here after reading that conversation:

> So I understand the argument of "it's still a draft", but there's also a good reason for that, there's simply still tasks to do. I've buggered @vladfrangu and @kyranet for (checks date of PR) 1,5 years now to help me with the PR because I'm inclined to think that I overcorrected in the source code as evident from what I see get returned in the unit tests, but I've never received any. Ergo, PR is stuck.

Adding to it, technically I could finish that PR in a couple of days if I really sit down and just finish it. I've just been lacking the motivation to do so. Once it's finished though, adding custom messages is a singe.

favna:
Dumping a comment on here as well as follow-up to my reply above, the shapeshfit version with custom messages has been released. We can review where we want to take this PR further now.

didinele:
I'm personally of the opinion that we should go further with zod after all.

I understand that an effort was made for that shapeshift release, but it's too little too late, on top of other accumulating issues. I know we keep on going back and forth on this, but the argument of "just wait until zod has an issue and we decide to switch back" really falls apart considering zod is a massive project that iterates quicker than shapeshift ever could - on top of, at least at this time, to my knowledge, having no big issues for us, while shapeshift has been problematic in various ways since the moment we introduced it.

didinele (approved):

appellation (approved):

kyranet (approved):

Qjuh:
Superseded by #10448

Issue #10062, opened 2023-12-29
LimitedCollection's maxSize also counts items that pass keepOverLimit


ImRodry:
### Which package is this bug report for?

discord.js

### Issue description

1. Create a LimitedCollection with maxSize of 2 and a keepOverLimit function
2. Add 2 elements that pass keepOverLimit
3. Add one that doesn't
4. Check to see that all 3 elements were correctly inserted
5. Add a fourth element that doesn't pass keepOverLimit
6. Check to see that the first element you added that didn't pass keepOverLimit has now been removed in order for your new element to be inserted, leaving you with 3 elements instead of 4.

The way I interpret this keepOverLimit function is that "any element added that passes this condition isn't counted towards the total size limit". I understand there may be other interpretations to this and that the current implementation may be right in the eyes of who interpreted this criteria, but I think this is also a valid way of seeing it and I would like to know if it is possible to change the current implementation to support this.

### Versions

- discord.js v14.14.1
- Node.js v20.10.0

### Issue priority

Low (slightly annoying)

### Which partials do you have configured?

Not applicable

### Which gateway intents are you subscribing to?

Not applicable

Qjuh:
> A function, which is passed the value and key of an entry, ran to decide to keep an entry past the maximum size
That’s what keepOverLimit is documented to be. And that’s what it is. It decides whether to keep an element although it‘s past the maxSize. Nowhere does it say that it doesn’t count towards the maxSize nor was it ever intended to be that way. So I doubt anyone would implement a breaking change into how this property acts without any reason to do so other than you misunderstanding what it‘s supposed to do.

ImRodry:
It's exactly because it doesn't say what happens to maxSize that I think it's reasonable to assume it doesn't count, just like it would be reasonable to assume that it does. If you're keeping an element over the limit, why would it count towards that limit and take up the space that could be taken up by other values that don't pass the requirements of that function?

p4bes:
Hi guys! In my understanding it is ok like @Qjuh described the handling of the count / maxSize. 

But I found another strange behaviour here: 
The docs say "A function called to check if an entry should be kept when the Collection is at max size."

To me it's strange that the first elements of the collection get cut off in favour of the latest ones, even when they are over limit and not to be kept over limit.

Example:
1. set maxSize to 2
2. add one element which doesn't pass keepOverLimit
3. add one element which pass keepOverLimit
4. Add one element which doesn't pass keepOverLimit

[...]

PR #10120, opened 2024-02-07
fix(LimitedCollection): fix keepOverLimit behavior  #10062

p4bes:
Original Issue: https://github.com/discordjs/discord.js/issues/10062

* always allow updating of elements even when maxSize is 0
* keep elements in the collection and disallow adding new ones if non matching the keepOverLimit expression
* do not remove any non-matching element from the collection when a new one is tried to add

The behavior of the set function of this class was inconistent depending on the maxSize and keepOverLimit attributes. Also the documentation is misleading any may need an update.

- Code changes have been tested against the Discord API, or there are no code changes (Tests can be found here: https://github.com/p4bes/discordjs-bot/blob/main/test/LimitedCollection.test.js )

### DIFF
--- a/packages/discord.js/src/util/LimitedCollection.js
+++ b/packages/discord.js/src/util/LimitedCollection.js
@@ -47,14 +47,10 @@ class LimitedCollection extends Collection {
   }
 
   set(key, value) {
-    if (this.maxSize === 0 && !this.keepOverLimit?.(value, key, this)) return this;
     if (this.size >= this.maxSize && !this.has(key)) {
-      for (const [k, v] of this.entries()) {
-        const keep = this.keepOverLimit?.(v, k, this) ?? false;
-        if (!keep) {
-          this.delete(k);
-          break;
-        }
+      const keep = this.keepOverLimit?.(value, key, this) ?? false;
+      if (!keep) {
+        return this;
       }
     }
     return super.set(key, value);


Qjuh:
This PR fixes nothing because there‘s nothing to be fixed. What you describe is intended behavior. A LimitedCollection is meant to work like a FIFO Queue that‘s also a Map. So if you add a value that makes it exceed the size the oldest element should be deleted (especially in terms of djs cache nothing else would make sense). And if the size is 0 and you try to add/update an element that doesn’t satisfy keepOverLimit there‘s nothing to be done. Because it won‘t make it into the collection either way.

so this PR is invalid as is the original issue

p4bes:
> This PR fixes nothing because there‘s nothing to be fixed. What you describe is intended behavior. A LimitedCollection is meant to work like a FIFO Queue that‘s also a Map. So if you add a value that makes it exceed the size the oldest element should be deleted (especially in terms of djs cache nothing else would make sense). And if the size is 0 and you try to add/update an element that doesn’t satisfy keepOverLimit there‘s nothing to be done. Because it won‘t make it into the collection either way.
> 
> so this PR is invalid as is the original issue

@Qjuh  Ok, I got your point regarding the FIFO behavior - that was not clear to me.

Still one last thing regarding the line 50:
Bildschirmfoto 2024-02-07 um 16 05 47

We do not think that this line is needed because:
1. If the size is 1 or greater but the maxSize is 0, the map element value cannot be updated (if it is not fullfilling the keepOverLimit expression)
2. In any scenario where maxSize is greater 0, it's possible to update those values regardless of the keepOverLimit fulfillment

What is the expected behavior? This might not be completely consistent.

Qjuh:
No, it’s not. If maxSize is 0 and keepOverLimit returns false the element won‘t be added to the collection at all in the first place. Because only elements with keepOverLimit true would be in it to begin with.

PR #10124, opened 2024-02-09
feat(Managers): update cache when use edit method

Pavel-Boyazov:
After updating something, you have to get information from Discord again, since using the information in the cache becomes impossible due to its irrelevance. I think this is wrong and suggest updating the cache with every valid change to something like in GuildChannelManager, instead of cloning.

### DIFF
--- a/packages/discord.js/src/managers/GuildEmojiManager.js
+++ b/packages/discord.js/src/managers/GuildEmojiManager.js
@@ -140,9 +140,8 @@ class GuildEmojiManager extends BaseGuildEmojiManager {
     });
     const existing = this.cache.get(id);
     if (existing) {
-      const clone = existing._clone();
-      clone._patch(newData);
-      return clone;
+      existing._patch(newData);
+      return existing; 
     }
     return this._add(newData);
   }
--- a/packages/discord.js/src/managers/GuildStickerManager.js
+++ b/packages/discord.js/src/managers/GuildStickerManager.js
@@ -115,9 +115,8 @@ class GuildStickerManager extends CachedManager {
 
     const existing = this.cache.get(stickerId);
     if (existing) {
-      const clone = existing._clone();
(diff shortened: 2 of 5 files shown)


vladfrangu:
> since using the information in the cache becomes impossible due to its irrelevance.

Cached data gets updated from the websocket, unless you don't have the intents required to get the events. What use case do you have that encounters this 👀? Or is this just to bring consistency from other places that do this?

Pavel-Boyazov:
Moreover, the message (and other) is added to the cache when it is changed, regardless of whether there is an intent. So why not keep it relevant?

Qjuh:
The issue with that is that for people that have the intent the messageUpdate event needs to emit with the oldMessage and newMessage. That is impossible to do if the message already was updated in cache before.

Pavel-Boyazov:
> The issue with that is that for people that have the intent the messageUpdate event needs to emit with the oldMessage and newMessage. That is impossible to do if the message already was updated in cache before.

Yeah, you're right. I also checked the implementation of like in GuildChannelManager, but it works every other time (at the same time, my solution never works. The difference is literally in tenths of milliseconds), while in GuildChannelManager event arrives stably before there is a change in the edit method. I'm stuck...

Pavel-Boyazov:
What about to check provided intents and if no GuildMessages intent - update cache?

Qjuh:
Changing the behavior of cache conditionally on set intents is not good DX.

kyranet:
I'll be closing this PR due to the reasons highlighted by @Qjuh.

If you don't have the intents, Discord.js won't receive the events. However, if the issue is the message being updated but not reflecting as such in the cache due to the lack of WebSocket events, you can force-fetch it `force: true` in the options to ensure it's always up-to-date rather than retrieving possibly stale data from cache.

PR #10128, opened 2024-02-13
refactor: remove usage of mixin on error classes

almeidx:
Removed the internal mixin function `DiscordjsErrorMixin` in order to provide better documentation
<https://discord.js.org/docs/packages/discord.js/14.14.1/DiscordjsError:Class>

image

### DIFF
--- a/packages/discord.js/typings/index.d.ts
+++ b/packages/discord.js/typings/index.d.ts
@@ -3915,23 +3915,23 @@ export enum DiscordjsErrorCodes {
   EntitlementCreateInvalidOwner = 'EntitlementCreateInvalidOwner',
 }
 
-/** @internal */
-export interface DiscordjsErrorFields<Name extends string> {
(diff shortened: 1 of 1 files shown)


vladfrangu (approved):

Jiralite (approved):

didinele (approved):

ssilve1989:
Why did the constructor become private on the typings? A private constructor would imply some other way of creating the object, via some static method, but the internal code still creates it via `new`. This change to making the constructor private breaks matching on instance type with tools like `ts-pattern`. 

Ex.
```typescript
    return match(error)
      .with(
        P.instanceOf(DiscordjsError),
        ({ code }) => code === DiscordjsErrorCodes.InteractionCollectorError,
        () => { /* some handler code */ }
      )
```
now breaks with 
> Argument of type 'typeof DiscordjsError' is not assignable to parameter of type 'AnyConstructor'.
> Cannot assign a 'private' constructor type to a 'public' constructor type.

PR #10136, opened 2024-02-18
refactor(ThreadChannel): use single thread member endpoint

sdanialraza:
This refactors the `fetchOwner` method to use the single thread member endpoint.


packages/discord.js/src/structures/ThreadChannel.js
```diff
@@ -293,15 +293,11 @@ class ThreadChannel extends BaseChannel {
...
-      if (existing) return existing;
-    }
+  async fetchOwner(options) {
+    if (!this.ownerId) return null;
 
-    // We cannot fetch a single thread member, as of this commit's date, Discord API responds with 405
-    const members = await this.members.fetch({ cache });
-    return members.get(this.ownerId) ?? null;
+    const member = await this.members.fetch({ ...options, member: this.ownerId }).catch(() => null);
+    return member;
```
jaw0r3k, line 300:
```suggestion
    return await this.members.fetch({ ...options, member: this.ownerId }).catch(() => null);
 ```
 
Also is this catch desired here?

Qjuh: Yes it is, so the method returns null in case the owner of the thread left the guild.

Jiralite: Whilst correct, I disagree this should be present. Silently catching errors and doing nothing here is not a good solution. For example, this removes the possibility of returning abortion errors.

There is no documentation to inform the condition when this returns `null`, so perhaps we could use that to our advantage. Line 297 could stay and `.catch(() => null)` could be removed. Then, we document this behaviour.

kyranet (approved):

SpaceEEC (approved):

PR #10149, opened 2024-02-23
feat(websocket): use proxy for websocket connections

NewMeeh:

The PR adds a way to use discord.js through a proxy.

I tested it and it works for my configuration. 
I am not sure how to update typings but I'll be happy to do so :) 
This PR adds one parameter but it is optional.
I will also be happy to add some documentation if needed.

### DIFF
--- a/packages/ws/src/ws/WebSocketShard.ts
+++ b/packages/ws/src/ws/WebSocketShard.ts
@@ -187,6 +187,7 @@ export class WebSocketShard extends AsyncEventEmitter<WebSocketShardEventsMap> {
 
 		const connection = new WebSocketConstructor(url, {
 			handshakeTimeout: this.strategy.options.handshakeTimeout ?? undefined,
+			agent: this.strategy.options.agent,
 		});
 
 		connection.binaryType = 'arraybuffer';
--- a/packages/ws/src/ws/WebSocketShard.ts
+++ b/packages/ws/src/ws/WebSocketShard.ts
@@ -20,6 +20,7 @@ import {
 	type GatewayReceivePayload,
 	type GatewaySendPayload,
 } from 'discord-api-types/v10';
+import { HttpsProxyAgent } from 'https-proxy-agent';
 import { WebSocket, type Data } from 'ws';
 import type { Inflate } from 'zlib-sync';
 import type { IContextFetchingStrategy } from '../strategies/context/IContextFetchingStrategy.js';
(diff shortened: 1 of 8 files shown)


almeidx:
Documentation for this option needs to be added to:
- https://github.com/NewMeeh/discord.js/blob/ce08bca2e16537cf537c8e987db79d9fdc7e97f0/packages/discord.js/src/util/Options.js#L87
- https://github.com/NewMeeh/discord.js/blob/ce08bca2e16537cf537c8e987db79d9fdc7e97f0/packages/discord.js/typings/index.d.ts#L6779

didinele (changes requested):
The key problem here is that your new agent option is not serializable over `worker_thread`s. This completely breaks the worker sharding strategy.

NewMeeh:
@didinele I see the problem. I don't really know how threads works in Javascript, if there's no way to share the agent option between threads, should I construct the `agent` inside the `WebSocketShard` but this means adding a dependency 🤷‍♂️

didinele:
I'm not confident we should go forward with this. Thoughts? @vladfrangu

didinele:
Yeah. Frankly, if the agent API was universal and we found a nice interface for users to pass an instance even in worker mode, I'd be a lot more in favor of this, but as it is it feels whacky at best.

packages/ws/src/ws/WebSocketShard.ts
```diff
@@ -185,8 +186,14 @@ export class WebSocketShard extends AsyncEventEmitter<WebSocketShardEventsMap> {
 
 		this.debug([`Connecting to ${url}`]);
 
+		const { proxy, httpsProxyAgentOptions } = this.strategy.options.proxyAgentOptions ?? {
+			proxy: undefined,
+			httpsProxyAgentOptions: undefined,
+		};
+		const agent = proxy ? new HttpsProxyAgent(proxy, httpsProxyAgentOptions) : undefined;
```
NotSugden, line 193:
```suggestion
		const { proxy, httpsProxyAgentOptions } = this.strategy.options.proxyAgentOptions ?? {
			proxy: undefined,
			httpsProxyAgentOptions: undefined,
		};
		const agent = proxy && new HttpsProxyAgent(proxy, httpsProxyAgentOptions);
```

didinele:
Hi. Especially considering the direction we're taking of ensuring the package works in non-node envs, on top of everything else already discussed in this thread, we won't be going forward with this.

PR #10152, opened 2024-02-24
fix(Embed): address `equals` method issue

### LINKED ISSUE #9095: `Embed#equals()` will never equal `EmbedBuilder#data`

discord.js

### Issue description

Here is the definition for `Embed#equals()`:

https://github.com/discordjs/discord.js/blob/8b70f497a1207e30edebdecd12b926c981c13d28/packages/discord.js/src/structures/Embed.js#L207-L217

This may take an `APIEmbed`. Looking at `EmbedBuilder#data`:

https://github.com/discordjs/discord.js/blob/8b70f497a1207e30edebdecd12b926c981c13d28/packages/builders/src/messages/embed/Embed.ts#L43-L47

We can see this is an `APIEmbed`. Therefore, these should be comparable. However, comparisons here will never be equal, as received embeds from Discord have a `type` property which is not present during construction.

The code sample contains a `console.log()`, of which the output is:

```xl
false { type: 'rich', description: 'test' } { description: 'test' }
```

The internally-called `isEqual()` method is derived from fast-deep-equal which compares the number of object keys, thus introducing this bug (view source code [here](https://github.com/epoberezkin/fast-deep-equal/blob/d807ffc5013e710deb1c63d463a03f729bcd144d/src/index.jst#L54-L56)).

An important mention here is that this is **not** limited to Discord's returned `type` property. Field `inline` values are optional whereas Discord will always return them, causing a mismatch of the length of keys. Additionally, what if Discord creates a new property on an embed? As soon as their API deployment is done, this method will begin failing despite no changes in discord.js. 

### Code sample

```typescript
import { Client, Events, GatewayIntentBits, EmbedBuilder } from "discord.js";
import { inspect } from "node:util";

const client = new Client({ intents: GatewayIntentBits.Guilds });
const embed = new EmbedBuilder().setDescription("test");

client.on(Events.ClientReady, async () => {
	const channel = client.channels.resolve("channel_id");
	const message = await channel.send({ embeds: [embed] });

	console.log(
		message.embeds[0].equals(embed.data),
		inspect(message.embeds[0].data, false, Number.POSITIVE_INFINITY, true),
		inspect(embed.data, false, Number.POSITIVE_INFINITY, true),
	);
});

client.login();
```

### Package version

14.7.2-dev

### Node.js version

Node.js 18.13.0 & typescript 4.9.4

### Operating system

macOS Ventura 13.1

Medium (should be fixed soon)

No Partials

Guilds

8b70f497a1207e30edebdecd12b926c981c13d28

sdanialraza:
Resolves #9095


packages/discord.js/src/structures/Embed.js
```diff
@@ -211,9 +211,30 @@ class Embed {
...
     }
-    return isEqual(other, this.data);
+
+    return (
+      this.color === (other.color ?? null) &&
+      this.description === (other.description ?? null) &&
+      this.title === (other.title ?? null) &&
+      this.url === (other.url ?? null) &&
+      (this.timestamp ? new Date(this.timestamp).getTime() : null) ===
+        (other.timestamp ? new Date(other.timestamp).getTime() : null) &&
```
almeidx, line 223:
```suggestion
      this.timestamp === (other.timestamp ?? null) &&
```

packages/discord.js/src/structures/Embed.js
```diff
@@ -211,9 +211,30 @@ class Embed {
...
+      this.image?.url === other.image?.url &&
+      this.thumbnail?.url === other.thumbnail?.url &&
+      this.video?.url === other.video?.url &&
```
almeidx, line 226: All of these are missing comparisons on height, width, and proxy URL

packages/discord.js/src/structures/Embed.js
```diff
@@ -211,9 +211,30 @@ class Embed {
...
+      isEqual(this.fields, other.fields?.map(field => ({ ...field, inline: field.inline ?? false })) ?? []) &&
+      isEqual(this.provider, other.provider ?? null) &&
+      isEqual(
+        { icon_url: this.author?.iconURL, name: this.author?.name, url: this.author?.url },
+        { icon_url: other.author?.icon_url, name: other.author?.name, url: other.author?.url },
+      ) &&
+      isEqual(
+        { icon_url: this.footer?.iconURL, text: this.footer?.text },
+        { icon_url: other.footer?.icon_url, text: other.footer?.text },
+      )
```
almeidx, line 236: Missing proxy URL comparisons. Is there a point in using `isEqual()` on 2 manually constructed objects here instead of direct comparisons?

almeidx (changes requested):

packages/discord.js/src/structures/Embed.js
sdanialraza: They're in different string formats so can't directly compare:
- 2024-02-26T04:52:15.531000+00:00 
- 2024-02-26T04:52:15.531Z

packages/discord.js/src/structures/Embed.js
sdanialraza: I've been using the code snippet from Jiralite's issue but with different embeds and those properties are never present when using `<Message>.embeds[0].data` so that's why I didn't add these comparisons

packages/discord.js/src/structures/Embed.js
almeidx:
```suggestion
      (this.timestamp && Date.parse(this.timestamp)) === (other.timestamp ? Date.parse(other.timestamp) : null) &&
```

almeidx (approved):

vladfrangu (approved):

SpaceEEC (approved):

PR #10156, opened 2024-02-27
fix(Faces): Escape backslash in `Shrug`

RedGuy12:

On stable, the value of `Shrug` is `¯_(ツ)_/¯`, which is incorrect. This PR fixes the error by escaping the backslash.

### DIFF
--- a/packages/formatters/__tests__/formatters.test.ts
+++ b/packages/formatters/__tests__/formatters.test.ts
@@ -314,12 +314,9 @@ describe('Message formatters', () => {
 	});
 
 	describe('Faces', () => {
-		// prettier-ignore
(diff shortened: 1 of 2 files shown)


Jiralite (dismissed):
How does this work for you? The extra `\` will escape the escape, resulting in an italicised face?

RedGuy12:
It looks like Discord's Markdown parser must have some special case for `/shrug`... I used the command, then used a bot to get the raw message content, and the underscores are not escaped.
image
image

Removing any one character from the emoticon causes Discord to parse it as proper Markdown again (note the greyed-out backslash)
image
image

almeidx (approved):

Jiralite (approved):

PR #10181, opened 2024-03-18
types: Use `ThreadChannel` and `AnyThreadChannel` consistently

cobaltt7:

I've been having issues with some methods returning `ThreadChannel` and some returning `AnyThreadChannel`. This PR makes how we use each consistent.
- If a more specific type can be determined (ie, threads in `ThreadOnlyChannel` will always be `PublicThreadChannel<true>`), that is used.
- Property values, return values, and anything else used as a value in userland are the more specific `AnyThreadChannel` type.
- If it's something the user passes in to Discord.js, or if it is used as a type guard, it uses the broader `ThreadChannel` type. Typically, the user shouldn't have to use `ThreadChannel` in their codebase at all (since all userland-facing values have been changed to `AnyThreadChannel`), but it doesn't hurt to allow it here.


packages/discord.js/typings/index.d.ts
```diff
@@ -3207,28 +3207,28 @@ export class ThreadChannel<ThreadOnly extends boolean = boolean> extends BaseCha
...
+  public setInvitable(invitable?: boolean, reason?: string): Promise<this>;
+  public setLocked(locked?: boolean, reason?: string): Promise<this>;
+  public setName(name: string, reason?: string): Promise<this>;
   // The following 3 methods can only be run on forum threads.
-  public setAppliedTags(appliedTags: readonly Snowflake[], reason?: string): Promise<ThreadChannel<true>>;
-  public pin(reason?: string): Promise<ThreadChannel<true>>;
-  public unpin(reason?: string): Promise<ThreadChannel<true>>;
+  public setAppliedTags(appliedTags: readonly Snowflake[], reason?: string): Promise<If<ThreadOnly, this, never>>;
+  public pin(reason?: string): Promise<If<ThreadOnly, this, never>>;
+  public unpin(reason?: string): Promise<If<ThreadOnly, this, never>>;
```
vladfrangu, line 3231: Doesn't this assert the type based on... if an error is thrown? 👀

packages/discord.js/typings/index.d.ts
```diff
@@ -4999,20 +4999,20 @@ export interface Caches {
   // TODO: GuildManager: [manager: typeof GuildManager, holds: typeof Guild];
   GuildMemberManager: [manager: typeof GuildMemberManager, holds: typeof GuildMember];
   GuildBanManager: [manager: typeof GuildBanManager, holds: typeof GuildBan];
-  GuildForumThreadManager: [manager: typeof GuildForumThreadManager, holds: typeof ThreadChannel<true>];
+  GuildForumThreadManager: [manager: typeof GuildForumThreadManager, holds: AnyThreadChannel<true>];
```
vladfrangu, line 5002: Pretty sure this still needs typeof

packages/discord.js/typings/index.d.ts
```diff
@@ -4999,20 +4999,20 @@ export interface Caches {
...
   GuildInviteManager: [manager: typeof GuildInviteManager, holds: typeof Invite];
   GuildMessageManager: [manager: typeof GuildMessageManager, holds: typeof Message<true>];
   GuildScheduledEventManager: [manager: typeof GuildScheduledEventManager, holds: typeof GuildScheduledEvent];
   GuildStickerManager: [manager: typeof GuildStickerManager, holds: typeof Sticker];
-  GuildTextThreadManager: [manager: typeof GuildTextThreadManager, holds: typeof ThreadChannel<false>];
+  GuildTextThreadManager: [manager: typeof GuildTextThreadManager, holds: AnyThreadChannel<false>];
```
vladfrangu, line 5007: Pretty sure this still needs typeof

packages/discord.js/typings/index.d.ts
```diff
@@ -4999,20 +4999,20 @@ export interface Caches {
...
   MessageManager: [manager: typeof MessageManager, holds: typeof Message];
   // TODO: PermissionOverwriteManager: [manager: typeof PermissionOverwriteManager, holds: typeof PermissionOverwrites];
   PresenceManager: [manager: typeof PresenceManager, holds: typeof Presence];
   ReactionManager: [manager: typeof ReactionManager, holds: typeof MessageReaction];
   ReactionUserManager: [manager: typeof ReactionUserManager, holds: typeof User];
   // TODO: RoleManager: [manager: typeof RoleManager, holds: typeof Role];
   StageInstanceManager: [manager: typeof StageInstanceManager, holds: typeof StageInstance];
-  ThreadManager: [manager: typeof ThreadManager, holds: typeof ThreadChannel];
+  ThreadManager: [manager: typeof ThreadManager, holds: AnyThreadChannel];
```
vladfrangu, line 5015: Pretty sure this still needs typeof

vladfrangu (dismissed):
Also don't forget to lint your codes

packages/discord.js/typings/index.d.ts
RedGuy12: Nope! `ThreadChannel` is an actual class that exists at runtime, that's why we need `typeof`. `AnyThreadChannel` is a type alias for different types of threads.

packages/discord.js/typings/index.d.ts
RedGuy12: Yeah. These methods exist at runtime on this class, so it'd be incorrect to move them elsewhere, and we couldn't move them in the code since there aren't any other classes, just interfaces. However, they only work in forum threads, as noted in the comment. They are guaranteed to throw elsewhere. There's no good way to restrict usage of these outside forum threads, but I figured we might as well flag the possible mistake in the return value.
Typically TypeScript recognizes functions that return `undefined` and flags unreachable code after them. Unfortunately, it doesn't yet do that for asynchronous functions, at least currently. But it still doesn't hurt to add it.
image
image

Also, the return type will be `this` even if `ThreadOnly` is just `boolean`. `never` is removed from unions, so the return type will only be `never` if it's guaranteed to be a non-forum thread throw.

RedGuy12: These functions aren't actually marked as async themselves, they just return the promise from `ThreadChannel#edit`. So if we wanted to, we could add our own runtime check for these methods (currently it just relies on the Discord API to fail). Then we could also change this return type to `If<ThreadOnly, Promise<this>, never>`, allowing TS's unreachability detection to work here.

packages/discord.js/typings/index.d.ts
vladfrangu: Can you add a test please that this doesn't break anything 🙏

packages/discord.js/typings/index.d.ts
```diff
@@ -968,7 +968,7 @@ export abstract class BaseChannel extends Base {
   public get url(): string;
   public delete(): Promise<this>;
   public fetch(force?: boolean): Promise<this>;
-  public isThread(): this is AnyThreadChannel;
+  public isThread(): this is ThreadChannel;
```
Qjuh, line 971: Wouldn't it be to better to keep this to the union type instead of the class to be able to narrow afterwards?

Qjuh (changes requested):

packages/discord.js/typings/index.test-d.ts
```diff
@@ -1624,8 +1630,10 @@ declare const guildChannelManager: GuildChannelManager;
 
 declare const threadManager: ThreadManager;
 {
-  expectType<Promise<AnyThreadChannel | null>>(threadManager.fetch('12345678901234567'));
-  expectType<Promise<AnyThreadChannel | null>>(threadManager.fetch('12345678901234567', { cache: true, force: false }));
```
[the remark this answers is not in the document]
RedGuy12: It's functionally the same, but unfortunately I couldn't keep the exact types the same. `ForumThreadChannel | TextThreadChannel` (the new return value when the Manager isn't narrowed to be more specific) resolves to `PublicThreadChannel<true> | PublicThreadChannel<false> | PrivateThreadChannel`. The old return value, `AnyThreadChannel`, now resolves to `PublicThreadChannel<boolean> | PrivateThreadChannel`. TypeScript doesn't consider these types to be the same.

However, the base `ThreadManager` class isn't exposed anywhere but Caches. All other references already (on `main`) use the more specific `GuildTextThreadManager` or `GuildForumThreadManager`, and their resolved return values remain the same.

packages/discord.js/typings/index.d.ts
RedGuy12: Yes, I missed that. Thanks for pointing it out.

packages/discord.js/typings/index.test-d.ts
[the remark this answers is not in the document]
Qjuh: And you just described why it's a breaking change... maybe should make `AnyThreadChannel = TextThreadChannel | ForumThreadChannel` to account for that?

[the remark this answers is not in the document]
RedGuy12: Yeah that makes sense, not sure why I didn't do that before, thanks

packages/discord.js/typings/index.d.ts
NotSugden: could we overwtite the functions at runtime. Is there something in the jsdoc as well to advise of this.

packages/discord.js/typings/index.d.ts
```diff
@@ -5006,20 +5010,20 @@ export interface Caches {
...
+  GuildTextThreadManager: [manager: typeof GuildTextThreadManager, holds: TextThreadChannel];
   MessageManager: [manager: typeof MessageManager, holds: typeof Message];
   // TODO: PermissionOverwriteManager: [manager: typeof PermissionOverwriteManager, holds: typeof PermissionOverwrites];
   PresenceManager: [manager: typeof PresenceManager, holds: typeof Presence];
   ReactionManager: [manager: typeof ReactionManager, holds: typeof MessageReaction];
   ReactionUserManager: [manager: typeof ReactionUserManager, holds: typeof User];
   // TODO: RoleManager: [manager: typeof RoleManager, holds: typeof Role];
   StageInstanceManager: [manager: typeof StageInstanceManager, holds: typeof StageInstance];
-  ThreadManager: [manager: typeof ThreadManager, holds: typeof ThreadChannel];
+  ThreadManager: [manager: typeof ThreadManager, holds: AnyThreadChannel];
```
vladfrangu, line 5026: My hunch was correct: this creates a discrepancy between what the makeCache method gets (manager, class, manager) and what the types imply when a threadmanager is brought into the mix (which implies the class is an instance). This still needs a typeof in some form

vladfrangu (changes requested):

packages/discord.js/typings/index.d.ts
RedGuy12: Just reverted the changes here since there's only ever the one constructor at runtime.

vladfrangu (approved):

Qjuh (approved):

SpaceEEC (approved):

PR #10185, opened 2024-03-21
feat: polls

almeidx:
Added support for polls.

Decided to take an opinionated stance with the poll results here. Instead of having a separate structure ([`results`](https://discord.com/developers/docs/resources/poll#poll-results-object-poll-results-object-structure)) like Discord, the answer vote count has been inlined directly into PollAnswer (e.g. `poll.results.answer_counts[].count` -> `poll.answers[].voteCount`), as it seemed to provide better DX to me. Additionally, also flattened all Poll Media objects (e.g. `poll.question.text` -> `poll.question`)

Upstream:
 - https://github.com/discord/discord-api-docs/pull/6746

Depends on:
- https://github.com/discordjs/discord-api-types/pull/925
- #10226


packages/discord.js/src/structures/Poll.js
```diff
@@ -0,0 +1,106 @@
...
+     * @readonly
+     */
+
+    Object.defineProperty(this, 'message', { value: message });
+
+    /**
+     * The question text of this poll
+     * @type {string}
+     */
+    this.question = data.question.text;
```
vladfrangu, line 31: I'm torn on this. On one hand, I get why you'd raise the text up, but on the other... What if Discord ends up adding more fields to question in the future? 😅 
Leaving others to give their thoughts too, I'm not gonna block on this

packages/discord.js/src/structures/Poll.js
```diff
@@ -0,0 +1,106 @@
...
+    if (data.results) {
+      /**
+       * Whether this poll's results have been precisely counted
+       * @type {boolean}
+       */
+      this.resultsFinalized = data.results.is_finalized;
+
+      for (const answerResult of data.results.answer_counts) {
+        const answer = this.answers.get(answerResult.id);
+        answer._patch(answerResult);
```
vladfrangu, line 73: Might be best to add a nullish access just in case

```suggestion
        answer?._patch(answerResult);
```

packages/discord.js/src/structures/Poll.js
```diff
@@ -0,0 +1,106 @@
...
+  async end() {
+    if (Date.now() > this.expiresTimestamp) {
+      throw new DiscordjsError(ErrorCodes.PollAlreadyExpired);
+    }
+
+    const message = await this.client.rest.post(Routes.expirePoll(this.message.channel.id, this.message.id));
+
+    this.message._patch(message);
+
+    return this.message;
```
vladfrangu, line 102: Won't this cause the GW event to emit a wrong messageUpdate? 👀 
We might need to clone this.message and patch the clone instead?

packages/discord.js/src/structures/Poll.js
RedGuy12: This field could be named `questionText` maybe, then if Discord adds more fields we could add a `question` object and deprecate `questionText`

vladfrangu: At that point just keeping it an object makes more sense 😄

vladfrangu (approved):

iCrawl (approved):

Jiralite (approved):

PR #10195, opened 2024-03-29
feat(Interactions): member objects for uncached guilds

### LINKED ISSUE #10010: Return a d.js object instead of a raw member object for interactions from uncached guilds

discord.js

### Feature

Currently, interaction.member can either be a discord.js member object or a raw member object:

https://github.com/discordjs/discord.js/blob/e412a22ceb92f142fbeddb6b9330e046bec92c69/packages/discord.js/src/structures/BaseInteraction.js#L63-L67

this is a painful API because these two objects are very different. I understand that d.js [GuildMember](https://old.discordjs.dev/#/docs/discord.js/main/class/GuildMember) requires a guild object, but a partial object that doesn't require a guild object would be much better than falling back to a completely separate format.

most d.js bots would probably error on accessing member data if they get added to a guild without the bot scope, since people are probably not handling the raw member case.

Introducing a new partial guild member class that takes a guild ID rather than a full guild object. It should support as much of GuildMember's interface as is possible to implement without having a full guild object. interaction.member should create an object of this class if the guild is uncached. 

_No response_

### Other context

this is currently an issue for adding apps with only the applications.commands scope, and it is also an issue for user apps, which will probably be used a lot more than the former.

advaith1:
fixes #10010
[Discord thread](https://discord.com/channels/222078108977594368/1179951996841771109)

Currently, when an interaction is received from a guild that isn't cached*, raw API member objects are returned instead of a d.js GuildMember object. This PR creates a new UncachedGuildMember object that is returned instead of raw API member objects.

This new object tries to be compatible with GuildMember, but retains backwards compatibility with APIGuildMember as to not make this a breaking change. Therefore it is still somewhat awkward and not great, but it is much better than just receiving an APIGuildMember. Hopefully in v15 we can remove the backwards compatibility and align it better with GuildMember.

**This has not been fully tested and likely needs changes - please thoroughly review!**

\* This was always possible by adding apps with only the applications.commands scope, but user apps makes this a much more common appearance. This PR is based on the discord.js patch AdvaithBot is running to properly handle user app commands.

### DIFF
--- a/packages/discord.js/src/structures/UncachedGuildMember.js
+++ b/packages/discord.js/src/structures/UncachedGuildMember.js
@@ -1,0 +1,332 @@
+'use strict';
+
+const Base = require('./Base');
+const TextBasedChannel = require('./interfaces/TextBasedChannel');
+const { GuildMemberFlagsBitField } = require('../util/GuildMemberFlagsBitField');
+const PermissionsBitField = require('../util/PermissionsBitField');
+
+/**
+ * Represents a member of a guild on Discord. Used in interactions from guilds that aren't cached.
+ * Backwards compatible with {@link APIGuildMember}.
+ * @implements {TextBasedChannel}
+ * @extends {Base}
+ */
+class UncachedGuildMember extends Base {
+  constructor(client, data, guildId) {
+    super(client);
+
+    /**
+     * The guild that this member is part of
+     * @type {Guild}
+     */
+    this.guildId = guildId;
+
+    if ('user' in data) {
+      /**
+       * The user that this guild member instance represents
+       * @type {?User}
+       */
+      this.user = this.client.users._add(data.user, true);
+    }
+
+    /**
+     * The nickname of this member, if they have one
+     * @type {?string}
+     */
+    this.nickname = data.nick;
+
+    /**
+     * The guild member's avatar hash
+     * @type {?string}
+     */
+    this.avatar = data.avatar ?? null;
+
+    /**
+     * The role ids of the member
+     * @name UncachedGuildMember#roleIds
+     * @type {Snowflake[]}
+     */
+    this.roleIds = data.roles;
+
+    /**
+     * The timestamp the member joined the guild at
+     * @type {number}
+     */
+    this.joinedTimestamp = Date.parse(data.joined_at);
+
+    /**
+     * The timestamp the member joined the guild at, as an ISO8601 timestamp
+     * @type {string}
+     * @deprecated Use {@link UncachedGuildMember#joinedTimestamp} or {@link UncachedGuildMember#joinedAt} instead
+     */
+    this.joined_at = data.joined_at;
+
+    /**
+     * The last timestamp this member started boosting the guild
+     * @type {?number}
+     */
+    this.premiumSinceTimestamp = data.premium_since ? Date.parse(data.premium_since) : null;
+
+    /**
+     * The last timestamp this member started boosting the guild, as an ISO8601 timestamp
+     * @type {?string}
+     * @deprecated Use {@link UncachedGuildMember#premiumSinceTimestamp}
+     * or {@link UncachedGuildMember#premiumSince} instead
+     */
+    this.premium_since = data.premium_since;
+
+    /**
+     * Whether the user is deafened in voice channels
+     * @type {?boolean}
+     * @deprecated
+     */
+    this.deaf = data.deaf;
+
+    /**
+     * Whether the user is muted in voice channels
+     * @type {?boolean}
+     * @deprecated
+     */
+    this.mute = data.mute;
+
+    /**
+     * The flags of this member
+     * @type {Readonly<GuildMemberFlagsBitField>}
+     */
+    this.parsedFlags = new GuildMemberFlagsBitField(data.flags).freeze();
+
+    /**
+     * The raw flags of this member
+     * @type {number}
+     * @deprecated Use {@link UncachedGuildMember#parsedFlags} instead.
+     * This field will be replaced with parsedFlags in the future.
+     */
+    this.flags = data.flags;
+
+    /**
+     * Whether this member has yet to pass the guild's membership gate
+     * @type {?boolean}
+     */
+    this.pending = data.pending;
+
+    /**
+     * The total permissions of the member in this channel, including overwrites
+     * @type {Readonly<PermissionsBitField>}
+     */
+    this.parsedPermissions = new PermissionsBitField(data.permissions).freeze();
+
+    /**
+     * Raw total permissions of the member in this channel, including overwrites
+     * @type {?string}
+     * @deprecated Use {@link UncachedGuildMember#parsedPermissions} instead.
+     * This field will be replaced with parsedPermissions in the future.
+     */
+    this.permissions = data.permissions;
+
+    /**
+     * The timestamp this member's timeout will be removed
+     * @type {?number}
+     */
+    this.communicationDisabledUntilTimestamp =
+      data.communication_disabled_until && Date.parse(data.communication_disabled_until);
+
+    /**
+     * The timestamp this member's timeout will be removed, as an ISO8601 timestamp
+     * @type {?string}
+     * @deprecated Use {@link UncachedGuildMember#communicationDisabledUntilTimestamp}
+     * or {@link UncachedGuildMember#communicationDisabledUntil} instead
+     */
+    this.communication_disabled_until = data.communication_disabled_until;
+
+    /**
+     * Whether this UncachedGuildMember is a partial (always true, as it is a partial GuildMember)
+     * @type {boolean}
+     * @readonly
+     */
+    this.partial = true;
+  }
+
+  /**
+   * A link to the member's guild avatar.
+   * @param {ImageURLOptions} [options={}] Options for the image URL
+   * @returns {?string}
+   */
+  avatarURL(options = {}) {
+    return this.avatar && this.client.rest.cdn.guildMemberAvatar(this.guildId, this.id, this.avatar, options);
+  }
+
+  /**
+   * A link to the member's guild avatar if they have one.
+   * Otherwise, a link to their {@link User#displayAvatarURL} will be returned.
+   * @param {ImageURLOptions} [options={}] Options for the Image URL
+   * @returns {string}
+   */
+  displayAvatarURL(options) {
+    return this.avatarURL(options) ?? this.user.displayAvatarURL(options);
+  }
+
+  /**
+   * The time this member joined the guild
+   * @type {?Date}
+   * @readonly
+   */
+  get joinedAt() {
+    return this.joinedTimestamp && new Date(this.joinedTimestamp);
+  }
+
+  /**
+   * The time this member's timeout will be removed
+   * @type {?Date}
+   * @readonly
+   */
+  get communicationDisabledUntil() {
+    return this.communicationDisabledUntilTimestamp && new Date(this.communicationDisabledUntilTimestamp);
+  }
+
+  /**
+   * The last time this member started boosting the guild
+   * @type {?Date}
+   * @readonly
+   */
+  get premiumSince() {
+    return this.premiumSinceTimestamp && new Date(this.premiumSinceTimestamp);
+  }
+
+  /**
+   * The member's id
+   * @type {Snowflake}
+   * @readonly
+   */
+  get id() {
+    return this.user.id;
+  }
+
+  /**
+   * The DM between the client's user and this member
+   * @type {?DMChannel}
+   * @readonly
+   */
+  get dmChannel() {
+    return this.client.users.dmChannel(this.id);
+  }
+
+  /**
+   * The nickname of this member, or their user display name if they don't have one
+   * @type {?string}
+   * @readonly
+   */
+  get displayName() {
+    return this.nickname ?? this.user.displayName;
+  }
+
+  /**
+   * The nickname of the member
+   * @name UncachedGuildMember#nick
+   * @type {?string}
+   * @deprecated Use {@link UncachedGuildMember#nickname} instead
+   */
+  get nick() {
+    return this.nickname;
+  }
+
+  /**
+   * The role ids of the member
+   * @name UncachedGuildMember#roles
+   * @type {Snowflake[]}
+   * @deprecated Use {@link UncachedGuildMember#roleIds} instead
+   */
+  get roles() {
+    return this.roleIds;
+  }
+
+  /**
+   * Whether this member is currently timed out
+   * @returns {boolean}
+   */
+  isCommunicationDisabled() {
+    return this.communicationDisabledUntilTimestamp > Date.now();
+  }
+
+  /**
+   * Creates a DM channel between the client and this member.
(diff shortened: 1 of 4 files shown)


packages/discord.js/src/structures/UncachedGuildMember.js
```diff
@@ -0,0 +1,332 @@
...
+ * @implements {TextBasedChannel}
+ * @extends {Base}
+ */
+class UncachedGuildMember extends Base {
+  constructor(client, data, guildId) {
+    super(client);
+
+    /**
+     * The guild that this member is part of
+     * @type {Guild}
```
vladfrangu, line 20: Its a string, not a Guild. You can add in a guild getter

packages/discord.js/src/structures/UncachedGuildMember.js
```diff
@@ -0,0 +1,332 @@
...
+     */
+    this.guildId = guildId;
+
+    if ('user' in data) {
```
vladfrangu, line 24: Can user ever be missing? 👀

packages/discord.js/src/structures/UncachedGuildMember.js
```diff
@@ -0,0 +1,332 @@
...
+     * @type {?string}
+     * @deprecated Use {@link UncachedGuildMember#premiumSinceTimestamp}
+     * or {@link UncachedGuildMember#premiumSince} instead
+     */
+    this.premium_since = data.premium_since;
+
+    /**
+     * Whether the user is deafened in voice channels
+     * @type {?boolean}
+     * @deprecated
```
vladfrangu, line 81: Why is this deprecated, and what't the alternative?

packages/discord.js/typings/index.d.ts
```diff
@@ -1642,6 +1642,38 @@ export class GuildMember extends Base {
   public valueOf(): string;
 }
 
+export class UncachedGuildMember extends Base {
```
vladfrangu, line 1645: It doesn't extend just base, you probably need to extend similar to GuildMember/User

vladfrangu (changes requested):
There's a LOT of inconsistencies between trying to make this a non-semver major change (by also having snake_case properties) and a semver-major change (by not exposing all values under the same name). Let's pick one side (and gut says you want this in a sem-minor 👀) and roll with it!

- `nick` isn't exposed as `nick`, only `nickname`
- `data.roles` isn't exposed at all, only as `roleIds`
	- this one I understand why, but I'd at least try to make uncachedMember.roles be our role manager (it _should_ work)

packages/discord.js/src/structures/UncachedGuildMember.js
advaith1: I was going off the member docs where it's optional, but actually as far as we are concerned here it'll always be present. will change that

packages/discord.js/typings/index.d.ts
advaith1: the GuildMember class definition also extends base, but that also has an interface so ig that's what's missing? I'm not sure why that has both an interface and a class

vladfrangu: We probably use the interface to make mixins easier, not sure 😓

packages/discord.js/src/structures/UncachedGuildMember.js
advaith1: oops will fix

packages/discord.js/typings/index.d.ts
Qjuh: Yep, it‘s the improved mixin typings using interface merging. The old Mixins were a bad fit for documentation purposes

Qjuh (changes requested):

packages/discord.js/src/structures/UncachedGuildMember.js
```diff
@@ -0,0 +1,332 @@
...
+      member instanceof this.constructor &&
+      this.id === member.id &&
+      this.partial === member.partial &&
+      this.guildId === member.guildId &&
+      this.joinedTimestamp === member.joinedTimestamp &&
+      this.nickname === member.nickname &&
+      this.avatar === member.avatar &&
+      this.pending === member.pending &&
+      this.communicationDisabledUntilTimestamp === member.communicationDisabledUntilTimestamp &&
+      this.parsedFlags.bitfield === member.parsedFlags.bitfield &&
```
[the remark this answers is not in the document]
advaith1: .flags is deprecated so I didn't want to use it; I want to rename parsedFlags to flags in v15

packages/discord.js/src/structures/UncachedGuildMember.js
```diff
@@ -0,0 +1,332 @@
...
+ * @param {string|MessagePayload|MessageCreateOptions} options The options to provide
+ * @returns {Promise<Message>}
+ * @example
+ * // Send a direct message
+ * UncachedGuildMember.send('Hello!')
+ *   .then(message => console.log(`Sent message: ${message.content} to ${guildMember.displayName}`))
+ *   .catch(console.error);
+ */
+
+TextBasedChannel.applyToClass(UncachedGuildMember);
```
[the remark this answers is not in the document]
advaith1: ah thanks, will look into that

packages/discord.js/src/structures/UncachedGuildMember.js
```diff
@@ -0,0 +1,332 @@
...
+     * The last timestamp this member started boosting the guild
+     * @type {?number}
+     */
+    this.premiumSinceTimestamp = data.premium_since ? Date.parse(data.premium_since) : null;
+
+    /**
+     * The last timestamp this member started boosting the guild, as an ISO8601 timestamp
+     * @type {?string}
+     * @deprecated Use {@link UncachedGuildMember#premiumSinceTimestamp}
+     * or {@link UncachedGuildMember#premiumSince} instead
```
NotSugden, line 74:
```suggestion
     * @deprecated Use {@link UncachedGuildMember#premiumSinceTimestamp} or {@link UncachedGuildMember#premiumSince} instead
```

packages/discord.js/src/structures/UncachedGuildMember.js
```diff
@@ -0,0 +1,332 @@
...
+  }
+
+  /**
+   * The member's id
+   * @type {Snowflake}
+   * @readonly
+   */
+  get id() {
+    return this.user.id;
+  }
```
NotSugden, line 202: does this need to be a getter. Why can't we just assign it statically as user ids don't change.

packages/discord.js/src/structures/UncachedGuildMember.js
```diff
@@ -0,0 +1,332 @@
...
+ * @example
+ * // Send a direct message
+ * UncachedGuildMember.send('Hello!')
+ *   .then(message => console.log(`Sent message: ${message.content} to ${guildMember.displayName}`))
+ *   .catch(console.error);
+ */
+
+TextBasedChannel.applyToClass(UncachedGuildMember);
+
+exports.UncachedGuildMember = UncachedGuildMember;
```
NotSugden, line 332: is there a reason we don't just export the class directly since its the only thing in this file

NotSugden (changes requested):

packages/discord.js/src/structures/UncachedGuildMember.js
Richy79: Richardkouakouchangeascommit.12fr

packages/discord.js/src/structures/UncachedGuildMember.js
tipakA: Because if it isn't a getter you're storing the id twice in each member object

packages/discord.js/src/structures/UncachedGuildMember.js
advaith1: idk, GuildMember does the same thing. should I change it here? is there a reason GuildMember does it?

packages/discord.js/src/structures/UncachedGuildMember.js
advaith1: this is what I had originally, but it fails prettier due to the line length, so I split it up.

packages/discord.js/src/structures/UncachedGuildMember.js
Qjuh: Historical reasons only. GuildMember used to export two things until removed in #7092 
So yeah, you should export as default here

vladfrangu:
Marking the PR as both sem minor and major until I test it out and ensure this _is_ a sem-minor PR 👀

packages/discord.js/src/structures/UncachedGuildMember.js
```diff
@@ -0,0 +1,328 @@
...
+     * The ID of the guild that this member is part of
+     * @type {string}
+     */
+    this.guildId = guildId;
+
+    /**
+     * The user that this guild member instance represents
+     * @type {User}
+     */
+    this.user = this.client.users._add(data.user, true);
```
Qjuh, line 28: Isn’t this a breaking change? `APIInteractionGuildMember#user` is `APIUser` and thus has snakecase properties. Now it doesn’t anymore.

vladfrangu: Yep, that is a breaking change indeed... @advaith1 can you solve that too?

packages/discord.js/src/structures/UncachedGuildMember.js
```diff
@@ -0,0 +1,328 @@
...
+   * @type {?DMChannel}
+   * @readonly
+   */
+  get dmChannel() {
+    return this.client.users.dmChannel(this.id);
+  }
+
+  /**
+   * The nickname of this member, or their user display name if they don't have one
+   * @type {?string}
```
NotSugden, line 211: would they not always have a _display_ name, or is this something to do with partials?

packages/discord.js/src/structures/UncachedGuildMember.js
```diff
@@ -0,0 +1,328 @@
...
+ * Sends a message to this user.
+ * @method send
+ * @memberof UncachedGuildMember
+ * @instance
+ * @param {string|MessagePayload|MessageCreateOptions} options The options to provide
+ * @returns {Promise<Message>}
+ * @example
+ * // Send a direct message
+ * UncachedGuildMember.send('Hello!')
+ *   .then(message => console.log(`Sent message: ${message.content} to ${guildMember.displayName}`))
```
NotSugden, line 322:
```suggestion
 *   .then(message => console.log(`Sent message: ${message.content} to ${uncachedGuildMember.displayName}`))
```
Feels a bit silly but eh

packages/discord.js/typings/index.d.ts
```diff
@@ -1642,6 +1642,46 @@ export class GuildMember extends Base {
...
+export interface UncachedGuildMember extends PartialTextBasedChannelFields<false> {}
+export class UncachedGuildMember extends Base {
+  private constructor(
+    client: Client<true>,
+    data: APIInteractionGuildMember | APIInteractionDataResolvedGuildMember,
+    guildId: Snowflake,
+  );
+  public avatar: string | null;
+  public get dmChannel(): DMChannel | null;
+  public get displayName(): string;
```
NotSugden, line 1654: i note infact from my previous comment that this is marked as always present but in the jsdoc it is not

NotSugden (changes requested):

packages/discord.js/src/structures/UncachedGuildMember.js
advaith1: hm yeah, why is this nullable in GuildMember? also user is nullable in GuildMember but id (which just returns this.user.id) isn't
are both of those just incorrectly marked as nullable

NotSugden: it would seem so

tipakA: Is the user being nullable related to the
> The field `user` won't be included in the member object attached to `MESSAGE_CREATE` and `MESSAGE_UPDATE` gateway events.

note that is under the Guild Member object table [in the DAPI docs](https://discord.com/developers/docs/resources/guild#guild-member-object)?

NotSugden:
this sounds like something we need to investigate.... possibly seperately
but that would mean the docs on here could be wrong until investigation
complete

packages/discord.js/src/structures/UncachedGuildMember.js
NotSugden: this sounds like something we need to investigate.... possibly seperately
but that would mean the docs on here could be wrong until investigation
complete

_re posted because email replies dont work properly_

Qjuh: In those two events the `message.author` field contains the `user` and is patched into the `GuildMember` accordingly so won’t be `null` either https://github.com/discordjs/discord.js/blob/de14c92c1158d3e5d7d87d29d2fe9d99eb407df5/packages/discord.js/src/structures/Message.js#L339

NotSugden: so in conclusion: the docs are wrong

packages/discord.js/src/structures/UncachedGuildMember.js
```diff
@@ -0,0 +1,328 @@
...
+
+  /**
+   * The nickname of the member
+   * @name UncachedGuildMember#nick
+   * @type {?string}
+   * @deprecated Use {@link UncachedGuildMember#nickname} instead
+   */
+  get nick() {
+    return this.nickname;
+  }
```
Jiralite, line 226: Why is this here?

Qjuh: To be non-breaking for https://discord-api-types.dev/api/discord-api-types-v10/interface/APIInteractionGuildMember#nick

vladfrangu (commented):
Thoughts on an `AnyGuildMember` type which is union of both Uncached and GuildMember and we have functions that assert whether a member instance is the uncached one?

advaith1:
Since this PR targets d.js v15 now, I've done some changes:

- renamed UncachedGuildMember to MinimalGuildMember
- removed backwards compatibility with the API object (since it was incorrect for the user field)
- GuildMember now extends MinimalGuildMember and adds getters/functions that need the guild object
- added member.isInCachedGuild() typeguard that assets it is a GuildMember

packages/discord.js/src/managers/GuildMemberRoleManager.js
```diff
@@ -154,7 +154,7 @@ class GuildMemberRoleManager extends DataManager {
       await this.client.rest.put(Routes.guildMemberRole(this.guild.id, this.member.id, resolvedRoleId), { reason });
 
       const clone = this.member._clone();
-      clone._roles = [...this.cache.keys(), resolvedRoleId];
+      clone.roleIds = [...this.cache.keys(), resolvedRoleId];
```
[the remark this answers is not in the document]
advaith1: This is not a logic change so if there is a bug here it sounds pre-existing and out of scope for this PR.

PR #10217, opened 2024-04-15
fix(api-extractor): link class in docs if interface merging happened

Qjuh:

Fixes parameter types linking to the the class instead of the (merged) interface for cases where interface merging happened. (e.g. GuildMember, User, DMChannel, TextChannel, Webhook, ...)


vladfrangu (approved):
If it works LGTM

Qjuh:
It works for js packages with manual ts typings (so just mainlib atm) under the assumption that the class declaration is always the last of the merged declarations. Which currently is the case. I‘m not sure if making it more complicated by finding a class declaration and if none exists falling back to other types of declaration would be beneficial here.

iCrawl (approved):

PR #9279, opened 2023-03-26
refactor: remove zlib-sync

didinele:
**Please describe the changes this PR makes and why it should be merged:**
draft because vlad wants us to keep zlib-sync & the new pure-native node:zlib impl together for a few dev releases

**Status and versioning classification:**

PR #10243, opened 2024-05-03
refactor: native zlib support

didinele:
#9279 again, but supports both

### DIFF
--- a/packages/ws/README.md
+++ b/packages/ws/README.md
@@ -50,7 +50,10 @@ const manager = new WebSocketManager({
 	intents: 0, // for no intents
 	rest,
 	// uncomment if you have zlib-sync installed and want to use compression
-	// compression: CompressionMethod.ZlibStream,
+	// compression: CompressionMethod.ZLibSync,
+
+	// alternatively, we now support compression using node's native `node:zlib` module:
+	// compression: CompressionMethod.ZLibNative,
 });
 
 manager.on(WebSocketShardEvents.Dispatch, (event) => {
--- a/packages/ws/README.md
+++ b/packages/ws/README.md
@@ -52,7 +52,7 @@ const manager = new WebSocketManager({
 	// uncomment if you have zlib-sync installed and want to use compression
 	// compression: CompressionMethod.ZLibSync,
 
-	// alternatively, we now support compression using node's native `node:zlib` module:
+	// alternatively, we support compression using node's native `node:zlib` module:
 	// compression: CompressionMethod.ZLibNative,
 });
 
--- a/packages/ws/README.md
+++ b/packages/ws/README.md
@@ -50,10 +50,10 @@ const manager = new WebSocketManager({
 	intents: 0, // for no intents
 	rest,
 	// uncomment if you have zlib-sync installed and want to use compression
-	// compression: CompressionMethod.ZLibSync,
+	// compression: CompressionMethod.ZlibSync,
 
 	// alternatively, we support compression using node's native `node:zlib` module:
-	// compression: CompressionMethod.ZLibNative,
+	// compression: CompressionMethod.ZlibNative,
 });
 
 manager.on(WebSocketShardEvents.Dispatch, (event) => {
--- a/packages/ws/src/utils/constants.ts
+++ b/packages/ws/src/utils/constants.ts
@@ -18,12 +18,18 @@ export enum Encoding {
  * Valid compression methods
  */
 export enum CompressionMethod {
-	ZlibStream = 'zlib-stream',
+	ZLibNative,
+	ZLibSync,
 }
 
 export const DefaultDeviceProperty = `@discordjs/ws [VI]{{inject}}[/VI]` as `@discordjs/ws ${string}`;
 
 const getDefaultSessionStore = lazy(() => new Collection<number, SessionInfo | null>());
+
+export const CompressionParameterMap = {
+	[CompressionMethod.ZLibNative]: 'zlib-stream',
+	[CompressionMethod.ZLibSync]: 'zlib-stream',
+} as const satisfies Record<CompressionMethod, string>;
 
 /**
  * Default options used by the manager
@@ -45,7 +51,7 @@ export const DefaultWebSocketManagerOptions = {
 	},
 	version: APIVersion,
 	encoding: Encoding.JSON,
-	transportCompression: null,
+	compression: null,
 	useIdentifyCompression: false,
 	retrieveSessionInfo(shardId) {
 		const store = getDefaultSessionStore();
--- a/packages/ws/src/utils/constants.ts
+++ b/packages/ws/src/utils/constants.ts
@@ -18,8 +18,8 @@ export enum Encoding {
  * Valid compression methods
  */
 export enum CompressionMethod {
-	ZLibNative,
-	ZLibSync,
+	ZlibNative,
+	ZlibSync,
 }
 
 export const DefaultDeviceProperty = `@discordjs/ws [VI]{{inject}}[/VI]` as `@discordjs/ws ${string}`;
@@ -27,8 +27,8 @@ export const DefaultDeviceProperty = `@discordjs/ws [VI]{{inject}}[/VI]` as `@di
 const getDefaultSessionStore = lazy(() => new Collection<number, SessionInfo | null>());
 
 export const CompressionParameterMap = {
-	[CompressionMethod.ZLibNative]: 'zlib-stream',
-	[CompressionMethod.ZLibSync]: 'zlib-stream',
+	[CompressionMethod.ZlibNative]: 'zlib-stream',
+	[CompressionMethod.ZlibSync]: 'zlib-stream',
 } as const satisfies Record<CompressionMethod, string>;
 
 /**
--- a/packages/ws/src/utils/constants.ts
+++ b/packages/ws/src/utils/constants.ts
@@ -45,7 +45,8 @@ export const DefaultWebSocketManagerOptions = {
 	},
 	version: APIVersion,
 	encoding: Encoding.JSON,
-	compression: null,
+	transportCompression: null,
+	useIdentifyCompression: false,
 	retrieveSessionInfo(shardId) {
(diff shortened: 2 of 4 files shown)


packages/ws/src/ws/WebSocketShard.ts
```diff
@@ -543,41 +580,61 @@ export class WebSocketShard extends AsyncEventEmitter<WebSocketShardEventsMap> {
...
+				) as GatewayReceivePayload;
+			}
 		}
 
 		this.debug([
 			'Received a message we were unable to decompress',
 			`isBinary: ${isBinary.toString()}`,
-			`useIdentifyCompress: ${this.useIdentifyCompress.toString()}`,
-			`inflate: ${Boolean(this.inflate).toString()}`,
+			`useIdentifyCompress: ${this.useIdentifyCompression.toString()}`,
```
SuperchupuDev, line 636:
```suggestion
			`useIdentifyCompression: ${this.useIdentifyCompression.toString()}`,
```
same thing i suggested in the previous pr :P

packages/ws/src/ws/WebSocketShard.ts
```diff
@@ -543,41 +580,61 @@ export class WebSocketShard extends AsyncEventEmitter<WebSocketShardEventsMap> {
...
+			`inflate: ${Boolean(this.nativeInflate).toString()}`,
```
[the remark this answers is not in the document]
didinele: oh, thanks. will do this a little different though

packages/ws/README.md
```diff
@@ -50,7 +50,10 @@ const manager = new WebSocketManager({
 	intents: 0, // for no intents
 	rest,
 	// uncomment if you have zlib-sync installed and want to use compression
-	// compression: CompressionMethod.ZlibStream,
+	// compression: CompressionMethod.ZLibSync,
+
+	// alternatively, we now support compression using node's native `node:zlib` module:
```
Jiralite, line 55: Words like "now" don't bode well in the future.

```suggestion
	// alternatively, we support compression using node's native `node:zlib` module:
```

packages/ws/src/utils/constants.ts
```diff
@@ -18,13 +18,19 @@ export enum Encoding {
...
 }
 
 export const DefaultDeviceProperty = `@discordjs/ws [VI]{{inject}}[/VI]` as `@discordjs/ws ${string}`;
 
 const getDefaultSessionStore = lazy(() => new Collection<number, SessionInfo | null>());
 
+export const CompressionParameterMap = {
+	[CompressionMethod.ZLibNative]: 'zlib-stream',
+	[CompressionMethod.ZLibSync]: 'zlib-stream',
+} as const satisfies Record<CompressionMethod, string>;
```
Jiralite, line 32: Isn't this a `Readonly<Record<...>>`?

almeidx: `satisfies` does not alter the type of the variable, it is merely used for validating that the object conforms to that type. Besides, `as const` already makes the type readonly.

Jiralite: What I meant was wouldn't it be more ideal to satisfy something that's `as const` (which is immutable) with something that is not immutable? So it should be like:

```suggestion
} as const satisfies Readonly<Record<CompressionMethod, string>>;
```

didinele: it's not any better IMO, just more verbose for no good reason

packages/ws/src/utils/constants.ts
```diff
@@ -18,13 +18,19 @@ export enum Encoding {
  * Valid compression methods
  */
 export enum CompressionMethod {
-	ZlibStream = 'zlib-stream',
+	ZLibNative,
```
vladfrangu, line 21: Why the capitalized `L` 😢

packages/ws/src/utils/constants.ts
vladfrangu: +1, the exported type is still going to be marked as readonly

packages/ws/src/ws/WebSocketShard.ts
```diff
@@ -451,28 +485,29 @@ export class WebSocketShard extends AsyncEventEmitter<WebSocketShardEventsMap> {
 			`shard id: ${this.id.toString()}`,
 			`shard count: ${this.strategy.options.shardCount}`,
 			`intents: ${this.strategy.options.intents}`,
-			`compression: ${this.inflate ? 'zlib-stream' : this.useIdentifyCompress ? 'identify' : 'none'}`,
+			`compression: ${this.strategy.options.compression === null ? (this.useIdentifyCompression ? 'identify' : 'none') : 'zlib-stream'}`,
```
vladfrangu, line 488: This should use the compression map, not a hardcoded string for the non-useIdentifyCompression path

packages/ws/src/ws/WebSocketShard.ts
```diff
@@ -543,41 +580,61 @@ export class WebSocketShard extends AsyncEventEmitter<WebSocketShardEventsMap> {
...
+				decompressable.length >= 4 &&
+				decompressable.at(-4) === 0x00 &&
+				decompressable.at(-3) === 0x00 &&
+				decompressable.at(-2) === 0xff &&
+				decompressable.at(-1) === 0xff;
 
-			const zlib = (await getZlibSync())!;
-			this.inflate.push(Buffer.from(decompressable), flush ? zlib.Z_SYNC_FLUSH : zlib.Z_NO_FLUSH);
+			if (this.nativeInflate) {
+				this.nativeInflate!.write(decompressable, 'binary');
```
vladfrangu, line 592: And if that `!` is needed, I will yell at TS

```suggestion
				this.nativeInflate.write(decompressable, 'binary');
```

packages/ws/src/ws/WebSocketShard.ts
```diff
@@ -543,41 +580,61 @@ export class WebSocketShard extends AsyncEventEmitter<WebSocketShardEventsMap> {
...
+			}
 		}
 
 		this.debug([
 			'Received a message we were unable to decompress',
 			`isBinary: ${isBinary.toString()}`,
-			`useIdentifyCompress: ${this.useIdentifyCompress.toString()}`,
-			`inflate: ${Boolean(this.inflate).toString()}`,
+			`useIdentifyCompression: ${this.useIdentifyCompression.toString()}`,
+			`inflate: ${this.nativeInflate ? 'node:zlib' : 'zlib-sync'} (zlib-stream)`,
```
vladfrangu, line 637: Should make this a map too ideally, so if/when we add zstd-stream support in we dont need a ton of extra changes

vladfrangu (changes requested):

packages/ws/src/utils/constants.ts
didinele: dunno. felt right to me. if you want, we can change it back

packages/ws/src/ws/WebSocketShard.ts
didinele: I have a different idea for this, let me know if you think this is ok.

packages/ws/src/utils/constants.ts
didinele: We settled on `Zlib`, vote was here:

- https://canary.discord.com/channels/222078108977594368/1209960895141388389/1236249988871229461

packages/ws/src/ws/WebSocketShard.ts
```diff
@@ -161,21 +179,63 @@ export class WebSocketShard extends AsyncEventEmitter<WebSocketShardEventsMap> {
...
+						console.warn('WebSocketShard: Compression is set to zlib-sync, but it is not installed.');
+						params.delete('compress');
+					}
+
+					break;
+				}
+			}
+		}
+
+		if (useIdentifyCompression) {
```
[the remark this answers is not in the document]
didinele: good catch

packages/ws/src/ws/WebSocketShard.ts
```diff
@@ -120,6 +126,18 @@ export class WebSocketShard extends AsyncEventEmitter<WebSocketShardEventsMap> {
...
+	private identifyCompressionEnabled = false;
+
+	/**
+	 * @privateRemarks
+	 *
+	 * This is needed because `this.strategy.options.compression` is not an actual reflection of the compression method
+	 * used, but rather the compression method that the user wants to use. This is because the libraries could just be missing.
+	 */
+	private get transportCompressionEnabled() {
+		return this.strategy.options.compression !== null && (Boolean(this.nativeInflate) || Boolean(this.zLibSyncInflate));
```
kyranet, line 138: Single `Boolean` call:
```suggestion
		return this.strategy.options.compression !== null && Boolean(this.nativeInflate || this.zLibSyncInflate);
```

Alternatively:
```suggestion
		return this.strategy.options.compression !== null && (this.nativeInflate ?? this.zLibSyncInflate) !== null;
```

kyranet (commented):
Just one small thing, not blocking

kyranet (approved):

vladfrangu (approved):

packages/ws/src/ws/WebSocketShard.ts
```diff
@@ -542,42 +615,50 @@ export class WebSocketShard extends AsyncEventEmitter<WebSocketShardEventsMap> {
...
+				return this.parseInflateResult(result);
+			}
 		}
 
 		this.debug([
 			'Received a message we were unable to decompress',
 			`isBinary: ${isBinary.toString()}`,
-			`useIdentifyCompress: ${this.useIdentifyCompress.toString()}`,
-			`inflate: ${Boolean(this.inflate).toString()}`,
+			`useIdentifyCompression: ${this.identifyCompressionEnabled.toString()}`,
```
SuperchupuDev, line 660:
```suggestion
			`identifyCompressionEnabled: ${this.identifyCompressionEnabled.toString()}`,
```
missed it again :P

SpaceEEC (approved):

PR #10249, opened 2024-05-04
fix(MessageManager): poll methods don't need a channel id

didinele:
average discord.js release (blundered)


vladfrangu (approved):

iCrawl (approved):

almeidx (changes requested):
- Missing type updates
- `fetchPollAnswerVoters` needs to be updated too
- both of these methods are being used in the Poll and PollAnswer classes

vladfrangu (approved):

almeidx (approved):

PR #10260, opened 2024-05-07
refactor(collection): `!` overrides for unchecked iterator access

Renegade334:

Currently, for backward compatibility reasons, when manually traversing builtin iterators, the type of `Iterator<T>.next().value` resolves to `T | any`, or in other words, just `any`.

Under changes to iterators in TypeScript 5.6, the resolved value type will change from `any` to `T | undefined` in strict mode.

There are a few instances in collection.ts where a result value no longer being `any` will cause issues. A test build shows the following errors:

```
src/collection.ts:90:54 - error TS2322: Type 'Value | undefined' is not assignable to type 'Value'.
  'Value' could be instantiated with an arbitrary type which could be unrelated to 'Value | undefined'.

90   return Array.from({ length: amount }, (): Value => iter.next().value);
                                                        ~~~~~~~~~~~~~~~~~

src/collection.ts:107:52 - error TS2322: Type 'Key | undefined' is not assignable to type 'Key'.
  'Key' could be instantiated with an arbitrary type which could be unrelated to 'Key | undefined'.

107   return Array.from({ length: amount }, (): Key => iter.next().value);
                                                       ~~~~~~~~~~~~~~~~~

src/collection.ts:515:10 - error TS2488: Type '[Key, Value] | undefined' must have a '[Symbol.iterator]()' method that returns an iterator.

515    const [key, value] = iter.next().value;
             ~~~~~~~~~~~~

src/collection.ts:629:4 - error TS2322: Type 'Value' is not assignable to type 'InitialValue'.
  'InitialValue' could be instantiated with an arbitrary type which could be unrelated to 'Value'.

629    accumulator = iterator.next().value[1];
       ~~~~~~~~~~~

  src/collection.ts:27:34
    27 export interface Collection<Key, Value> extends Map<Key, Value> {
                                        ~~~~~
    This type parameter might need an `extends InitialValue` constraint.

src/collection.ts:629:18 - error TS2532: Object is possibly 'undefined'.

629    accumulator = iterator.next().value[1];
                     ~~~~~~~~~~~~~~~~~~~~~


Found 5 errors in the same file, starting at: src/collection.ts:90
```

All of these errors occur in places where the length of the iterator has already been checked, so we know that the iterator will never be "done" and will therefore always return a value of type `T`. It's therefore safe to use `!` in these instances. Additionally, under `reduce()`, if there's no `initialValue` provided then `InitialValue = Value`, so it's safe to force the accumulator type (in a similar vein to what already occurs in `reduceRight()`).

- These are changes related to compilation only, and have no effects on the shape or behaviour of the API.
- These changes are fully build-compatible with current versions of TypeScript.


Renegade334:
Merged as #10394.

PR #10277, opened 2024-05-14
refactor(GuildChannelManager): improve addFollower errors

Renegade334:

This appears to be the last incidence of a naked `Error` being thrown within the d.js library. Changed to throw `DiscordjsTypeError` on failed resolution, along with individualising the parameter checks for easier debugging.

Adds a `NewsChannelResolvable` type for semantic purposes. This is a non-breaking type change.

### DIFF
--- a/packages/discord.js/src/managers/GuildChannelManager.js
+++ b/packages/discord.js/src/managers/GuildChannelManager.js
@@ -99,16 +99,28 @@ class GuildChannelManager extends CachedManager {
   }
 
   /**
+   * Data that can be resolved to a News Channel object. This can be:
+   * * A NewsChannel object
+   * * A Snowflake
+   * @typedef {NewsChannel|Snowflake} NewsChannelResolvable
+   */
+
+  /**
    * Adds the target channel to a channel's followers.
-   * @param {NewsChannel|Snowflake} channel The channel to follow
+   * @param {NewsChannelResolvable} channel The channel to follow
    * @param {TextChannelResolvable} targetChannel The channel where published announcements will be posted at
    * @param {string} [reason] Reason for creating the webhook
    * @returns {Promise<Snowflake>} Returns created target webhook id.
    */
   async addFollower(channel, targetChannel, reason) {
     const channelId = this.resolveId(channel);
+    if (!channelId) {
+      throw new DiscordjsTypeError(ErrorCodes.InvalidType, 'channel', 'NewsChannelResolvable');
+    }
     const targetChannelId = this.resolveId(targetChannel);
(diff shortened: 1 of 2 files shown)


packages/discord.js/src/managers/GuildChannelManager.js
```diff
@@ -98,17 +98,29 @@ class GuildChannelManager extends CachedManager {
...
+   * @param {NewsChannelResolvable} channel The channel to follow
    * @param {TextChannelResolvable} targetChannel The channel where published announcements will be posted at
    * @param {string} [reason] Reason for creating the webhook
    * @returns {Promise<Snowflake>} Returns created target webhook id.
    */
   async addFollower(channel, targetChannel, reason) {
     const channelId = this.resolveId(channel);
+    if (!channelId) {
+      throw new DiscordjsTypeError(ErrorCodes.InvalidType, 'channel', 'NewsChannelResolvable');
+    }
```
monbrey, line 119: I thought that it had been clearly stated in the [#djs-in-dev-version](https://discord.com/channels/222078108977594368/937420598383108167) channel that we would not be performing internal validations on API calls.

This introduces an additional reliance on the channel being cached - while this is usually the norm and is generally required for many aspects of discord.js to function as expected, its not a useful dependency here.

Renegade334: The resolution has been left as-is, though? It's only the thrown error that's been changed.

Jiralite: Won't throwing a different kind of error be unsuspecting, or possible breaking?

If a wrong channel type is passed, a `DiscordAPIError` would be thrown. Now, it'll throw `DiscordjsTypeError`. The `instanceof` check would no longer work.

Renegade334: This doesn't add a check for the channel type, though.

If the parameter is a GuildChannelResolvable (regardless of channel type) then it'll be resolved successfully by the GuildChannelManager and passed to the API call, just the same as before. If the channel isn't a NewsChannel, it'll still be resolved, and it'll still result in an API error.

In other words, anything that would have thrown a DiscordAPIError before, will still do so. The only change is that resolve errors (ie. passing something that isn't a GuildChannelResolvable) will now result in a DiscordjsTypeError being thrown, rather than a naked Error.

`NewsChannel | Snowflake` is already the parameter type, so aliasing it to NewsChannelResolvable is purely decorative. The only reason I've added this is to the DiscordjsTypeError message cleaner. It doesn't imply that the channel type is validated by the library, just the same as any other specific channel type Resolvables.

didinele (approved):

SpaceEEC (approved):

kyranet (approved):

PR #10283, opened 2024-05-18
feat(Message): add `call`

sdanialraza:
- https://github.com/discord/discord-api-docs/pull/6864


packages/discord.js/src/structures/Message.js
```diff
@@ -417,6 +417,28 @@ class Message extends Base {
...
+     * @property {Snowflake[]} participants The ids of the users that participated in the call
+     */
+
+    if (data.call) {
+      /**
+       * The call associated with the message
+       * @type {?MessageCall}
+       */
+      this.call = {
+        endedAt: data.call.ended_timestamp && new Date(data.call.ended_timestamp),
```
vladfrangu, line 435: This should work, but test it first!

```suggestion
        get endedAt() {
        	return this.endedTimestamp && new Date(this.endedTimestamp)
        },
```

benjGam (approved):

packages/discord.js/typings/index.d.ts
```diff
@@ -2025,6 +2025,12 @@ export class LimitedCollection<Key, Value> extends Collection<Key, Value> {
   public keepOverLimit: ((value: Value, key: Key, collection: this) => boolean) | null;
 }
 
+export interface MessageCall {
+  get endedAt(): Readonly<Date> | null;
```
Jiralite, line 2029: This is technically not a `Readonly<Date>`, is it? I think this should be removed. In JSDoc, it was added because it's a getter.

packages/discord.js/typings/index.d.ts
```diff
@@ -2025,6 +2025,12 @@ export class LimitedCollection<Key, Value> extends Collection<Key, Value> {
   public keepOverLimit: ((value: Value, key: Key, collection: this) => boolean) | null;
 }
 
+export interface MessageCall {
```
vladfrangu, line 2028: You need to add this property on message too

vladfrangu (dismissed):

packages/discord.js/src/structures/Message.js
```diff
@@ -417,6 +417,30 @@ class Message extends Base {
...
+     * @property {Snowflake[]} participants The ids of the users that participated in the call
+     */
+
+    if (data.call) {
+      /**
+       * The call associated with the message
+       * @type {?MessageCall}
+       */
+      this.call = {
+        endedTimestamp: data.call.ended_timestamp && Date.parse(data.call.ended_timestamp),
```
almeidx, line 435: According to the docs, this might also missing

```suggestion
        endedTimestamp: data.call.ended_timestamp ? Date.parse(data.call.ended_timestamp) : null,
```

packages/discord.js/src/structures/Message.js
```diff
@@ -417,6 +417,30 @@ class Message extends Base {
     } else {
       this.poll ??= null;
     }
+
+    /**
+     * The call associated with the message
```
almeidx, line 422:
```suggestion
     * A call associated with a message
```

packages/discord.js/src/structures/Message.js
```diff
@@ -417,6 +417,30 @@ class Message extends Base {
...
+     * @typedef {Object} MessageCall
+     * @property {?Readonly<Date>} endedAt The time the call ended
```
almeidx, line 424:
```suggestion
     * @property {Readonly<?Date>} endedAt The time the call ended
```

almeidx (changes requested):

packages/discord.js/src/structures/Message.js
Jiralite: I believe "the" is more appropriate since we are inside the message class (ergo looking at a singular message) and there can only be one call on a message, right?

Jiralite: Ah, understood. Then I agree.

Jiralite (approved):

almeidx (approved):

vladfrangu (approved):

PR #10295, opened 2024-05-21
fix: Throw error on no message id for `Message#fetchReference()`

### LINKED ISSUE #10294: `Message#fetchReference()` returns a non-message on certain message types

discord.js

### Issue description

[`fetchReference()`](https://github.com/discordjs/discord.js/blob/b36ec983828c7001e47debcd435592ea026768d5/packages/discord.js/src/structures/Message.js#L708) checks that the message has a reference and that the channel can be resolved, but doesn't check if `this.reference.messageId` exists.

Some message types (like [Thread Created Messages](https://discord.com/developers/docs/resources/channel#message-types-thread-created-messages)) only include `channel_id` and `guild_id`. Trying to `fetchReference()` on those messages causes a call to [`await channel.messages.fetch(messageId = undefined);`](https://github.com/discordjs/discord.js/blob/b36ec983828c7001e47debcd435592ea026768d5/packages/discord.js/src/structures/Message.js#L713) which then calls [`_fetchMany()`](https://github.com/discordjs/discord.js/blob/b36ec983828c7001e47debcd435592ea026768d5/packages/discord.js/src/managers/MessageManager.js#L94) and causes a `Promise<Collection<Snowflake, Message>>` to be returned instead of a `Promise<Message>`.

The documentation and types specify `Promise<Message>` so it's possible to write TS code that happily compiles and runs until you get a Thread Created Message and try to access a message property and everything crashes (how i figured this out lol). Ideally it should throw an error instead of silently returning the wrong type.

### Code sample

```typescript
const channelId = '<some channel ID>'
const channel = client.channels.cache.get(channelId)
if (!(channel instanceof BaseGuildTextChannel))
  throw new Error('Channel is not text-based')

await channel.threads.create({
  name: 'Example thread',
})

const latestMessage = await channel.messages.fetch({ limit: 1 })
const lm = latestMessage.first()
if (lm) {
  const notAMessage = await lm.fetchReference()
  // TS treats `notAMessage` as `Message<true>`
  console.log(`Latest message type: ${lm.type}`)
  console.log(`Referenced message: ${notAMessage instanceof Message} ${notAMessage}`)
  // Latest message type: 18
  // Referenced message: false [object Map] 

  // TS accepts this despite the property access failing at runtime
  console.log(`This will error at runtime: ${notAMessage.author.id}`)
  // Error: Cannot read properties of undefined (reading 'id')
}
```

### Versions

- discord.js 14.15.2
- Node.js 22.0.0
- typescript 5.4.5

### Issue priority

Low (slightly annoying)

Not applicable

Not applicable

b36ec9838

Jiralite:
Resolves #10294.

Simply put: if there's no message id, throw `ErrorCodes.MessageReferenceMissing`.

Also, the JSDoc had these properties as `?`. `?` does not indicate `undefined`—it indicates `<type> | null`.[^1]

[^1]: https://jsdoc.app/tags-type


packages/discord.js/src/structures/Message.js
```diff
@@ -354,15 +354,15 @@ class Message extends Base {
...
+     * * {@link MessageType.ChannelFollowAdd}
      * * {@link MessageType.Reply}
      * * {@link MessageType.ThreadStarterMessage}
      * @see {@link https://discord.com/developers/docs/resources/channel#message-types}
      * @typedef {Object} MessageReference
-     * @property {Snowflake} channelId The channel's id the message was referenced
-     * @property {?Snowflake} guildId The guild's id the message was referenced
-     * @property {?Snowflake} messageId The message's id that was referenced
+     * @property {Snowflake} channelId The channel id that was referenced
+     * @property {Snowflake|undefined} guildId The guild id that was referenced
```
Qjuh, line 364:
```suggestion
     * @property {Snowflake} [guildId] The guild id that was referenced
```

packages/discord.js/src/structures/Message.js
```diff
@@ -354,15 +354,15 @@ class Message extends Base {
...
+     * @property {Snowflake|undefined} messageId The message id that was referenced
```
Qjuh, line 365:
```suggestion
     * @property {Snowflake} [messageId] The message id that was referenced
```

Qjuh (changes requested):

packages/discord.js/src/structures/Message.js
Jiralite: Decided to leave them as-is as `[]` would indicate optionality but the properties are always present.

vladfrangu (approved):

kyranet (approved):

didinele (approved):

Issue #10273, opened 2024-05-13
Redis ReplyError when pulling with xreadgroup on empty subscribed events


nizans:
### Which package is this bug report for?

brokers

### Issue description

Hey,
First of all, Thanks for this package!

I noticed that `PubSubRedisBroker` errors when it is trying to pull using `xreadgroup` without subscribed events.

The error is a Redis `ReplyError` because the command is incomplete because it has no streams keys in it
Heres the error:
```bash
ReplyError: ERR syntax error
    at parseError (C:\dev\playground\discord-js-redis-broker-reproduce\node_modules\redis-parser\lib\parser.js:179:12)
    at parseType (C:\dev\playground\discord-js-redis-broker-reproduce\node_modules\redis-parser\lib\parser.js:302:14) {
  command: {
    name: 'xreadgroup',
    args: [
      'GROUP',
      'channel',
      '7b60fb0c836ce1d6320712e4b74642c30a9c942d',
      'COUNT',
      '10',
      'BLOCK',
      '5000',
      'STREAMS'
    ]
  }
}
```

The error shows clearly that the command simply missing the stream keys, which at that point don't exists.

Am I doing something wrong here?

If not I think this error can be avoided easily by checking if there's any subscribed events in before the next poll
 
Ill gladly open a PR for that.

Thanks

### Code sample

```typescript
import { PubSubRedisBroker } from "@discordjs/brokers";
import { Redis } from "ioredis";

const broker = new PubSubRedisBroker({ redisClient: new Redis() });

await broker.subscribe("channel", ["event"]);

broker.on("event", (message) => {
  console.log(message);
});

broker.on("error", (error) => {
  // This will be a ReplyError on the next poll
  console.error(error);
});

await broker.publish("event", "message");

// This will make the BaseRedis subscribedEvents empty
// But BaseRedis listen will continue to poll until errors
await broker.unsubscribe("channel", ["event"]);
```

### Versions

- "@discordjs/brokers": "^0.2.2"
- Node 18
- "ioredis": "^5.3.2"

### Issue priority

Low (slightly annoying)

### Which partials do you have configured?

Not applicable

### Which gateway intents are you subscribing to?

Not applicable

didinele:
> If not I think this error can be avoided easily by checking if there's any subscribed events in before the next poll

yup. odd bug. can I ask what the use case here is? are you dynamically subscribing/unsubscribing to certain keys? I'd never see myself running into this personally

[...]

PR #10297, opened 2024-05-23
refactor(brokers): re-design API to make groups a constructor option

nizans:
~~A small fix~~ A complete refactor to address the issue described in https://github.com/discordjs/discord.js/issues/10273 where the broker keeps pooling even when there's no subscribed events causing a `Redis` `ReplyError`.

This PR fixes this by stop pooling when there's no subscriptions and adds a cleanup to the event emitter listeners on the destroy as discussed in https://github.com/discordjs/discord.js/issues/10273 

### DIFF
--- a/packages/brokers/src/brokers/Broker.ts
+++ b/packages/brokers/src/brokers/Broker.ts
@@ -42,7 +42,7 @@ export type ToEventMap<
 
 export interface IBaseBroker<TEvents extends Record<string, any>> {
 	/**
-	 * Subscribes to the given events, grouping them by the given group name
+	 * Subscribes to the given events
 	 */
 	subscribe(events: (keyof TEvents)[]): Promise<void>;
 	/**
--- a/packages/brokers/src/brokers/Broker.ts
+++ b/packages/brokers/src/brokers/Broker.ts
@@ -44,11 +44,11 @@ export interface IBaseBroker<TEvents extends Record<string, any>> {
 	/**
 	 * Subscribes to the given events, grouping them by the given group name
 	 */
-	subscribe(group: string, events: (keyof TEvents)[]): Promise<void>;
+	subscribe(events: (keyof TEvents)[]): Promise<void>;
 	/**
 	 * Unsubscribes from the given events - it's required to pass the same group name as when subscribing for proper cleanup
 	 */
-	unsubscribe(group: string, events: (keyof TEvents)[]): Promise<void>;
+	unsubscribe(events: (keyof TEvents)[]): Promise<void>;
 }
 
 export interface IPubSubBroker<TEvents extends Record<string, any>>
--- a/packages/brokers/src/brokers/Broker.ts
+++ b/packages/brokers/src/brokers/Broker.ts
@@ -46,7 +46,7 @@ export interface IBaseBroker<TEvents extends Record<string, any>> {
 	 */
 	subscribe(events: (keyof TEvents)[]): Promise<void>;
 	/**
-	 * Unsubscribes from the given events - it's required to pass the same group name as when subscribing for proper cleanup
+	 * Unsubscribes from the given events
 	 */
 	unsubscribe(events: (keyof TEvents)[]): Promise<void>;
 }
--- a/packages/brokers/src/brokers/redis/BaseRedis.ts
+++ b/packages/brokers/src/brokers/redis/BaseRedis.ts
@@ -197,7 +197,6 @@ export abstract class BaseRedisBroker<TEvents extends Record<string, any>>
 		await this.unsubscribe(Array.from(this.subscribedEvents));
 		this.streamReadClient.disconnect();
 		this.redisClient.disconnect();
-		this.removeAllListeners();
 	}
 
 	/**
--- a/packages/brokers/src/brokers/redis/BaseRedis.ts
+++ b/packages/brokers/src/brokers/redis/BaseRedis.ts
@@ -23,10 +23,19 @@ export interface RedisBrokerOptions extends BaseBrokerOptions {
 	 * How long to block for messages when polling
 	 */
 	blockTimeout?: number;
+
+	/**
+	 * Consumer group name to use for this broker
+	 *
+	 * @see {@link https://redis.io/commands/xreadgroup/}
+	 */
+	group: string;
+
 	/**
 	 * Max number of messages to poll at once
 	 */
 	maxChunk?: number;
+
 	/**
 	 * Unique consumer name.
 	 *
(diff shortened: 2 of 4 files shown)


packages/brokers/src/brokers/redis/BaseRedis.ts
```diff
@@ -187,6 +187,7 @@ export abstract class BaseRedisBroker<TEvents extends Record<string, any>>
 	public async destroy() {
 		this.streamReadClient.disconnect();
```
didinele, line 188: This doesn't accomplish what we want here. 

For totally correct cleanup, what we have to do is keep track of everything we've `subscribe`'d to and call `unsubscribe` before destroying the actual redis clients.

The main problem with this is we don't have access to the `group`s; we should start storing a map in `subscribe` - was my first thought, but...

Looking at the implementation right now, I'm noticing that we don't really support having multiple groups set up at the same time anyway (and it doesn't make a lot sense to) - as in you can do it, but it will just break in unexpected ways because we don't track what event belongs to what group.

Let's just break the whole API instead; take `group` as a constructor option and reference it that way instead, remove it as a parameter from all the methods - that's a lot more correct and also allows us to do this here:

```ts
this.unsubscribe(...this.subscribedEvents);
```

didinele (changes requested):
Thanks for this PR! I haven't had the time.

packages/brokers/src/brokers/redis/BaseRedis.ts
```diff
@@ -187,6 +187,7 @@ export abstract class BaseRedisBroker<TEvents extends Record<string, any>>
...
 		this.redisClient.disconnect();
+		this.removeAllListeners();
```
didinele, line 190: This doesn't accomplish what we want here. 

For totally correct cleanup, what we have to do is keep track of everything we've `subscribe`'d to and call `unsubscribe` before destroying the actual redis clients.

The main problem with this is we don't have access to the `group`s; we should start storing a map in `subscribe` - was my first thought, but...

Looking at the implementation right now, I'm noticing that we don't really support having multiple groups set up at the same time anyway (and it doesn't make a lot sense to) - as in you can do it, but it will just break in unexpected ways because we don't track what event belongs to what group.

Let's just break the whole API instead; take `group` as a constructor option and reference it that way instead, remove it as a parameter from all the methods - that's a lot more correct and also allows us to do this here:

```ts
this.unsubscribe(...this.subscribedEvents);
```

nizans: Sounds good! 

I think setting the group as a constructor parameter makes more sense. In my case I even created a wrapper around it that takes the group and wraps the subscribe/unsubscribe methods.

I pushed a commit LMK what you think.

Not sure what `removeAllListeners` does exactly but should we keep it as well?
There's noting else that removes the listeners

packages/brokers/src/brokers/Broker.ts
```diff
@@ -44,11 +44,11 @@ export interface IBaseBroker<TEvents extends Record<string, any>> {
 	/**
 	 * Subscribes to the given events, grouping them by the given group name
 	 */
-	subscribe(group: string, events: (keyof TEvents)[]): Promise<void>;
+	subscribe(events: (keyof TEvents)[]): Promise<void>;
 	/**
 	 * Unsubscribes from the given events - it's required to pass the same group name as when subscribing for proper cleanup
 	 */
-	unsubscribe(group: string, events: (keyof TEvents)[]): Promise<void>;
+	unsubscribe(events: (keyof TEvents)[]): Promise<void>;
```
Jiralite, line 51: Is "required to pass the same group name" still relevant documentation?

packages/brokers/src/brokers/Broker.ts
```diff
@@ -44,11 +44,11 @@ export interface IBaseBroker<TEvents extends Record<string, any>> {
 	/**
 	 * Subscribes to the given events, grouping them by the given group name
```
nizans, line 45:
```suggestion
	 * Subscribes to the given events```

File: packages/brokers/src/brokers/redis/BaseRedis.ts (hunk 2 above)
Comment (didinele) [reply packages/brokers/src/brokers/redis/BaseRedis.ts:200]: No, cleaning up listeners I'd argue is unexpected behavior, IMO. The user should do that on their own if they want to.

File: packages/brokers/src/brokers/Broker.ts (hunk 3 above)
Comment (nizans) [reply packages/brokers/src/brokers/Broker.ts]: No
I removed those
Ty

File: packages/brokers/src/brokers/redis/BaseRedis.ts [hunk 5]
```diff
@@ -185,8 +194,10 @@ export abstract class BaseRedisBroker<TEvents extends Record<string, any>>
 	 * Destroys the broker, closing all connections
 	 */
 	public async destroy() {
+		await this.unsubscribe(Array.from(this.subscribedEvents));
 		this.streamReadClient.disconnect();
 		this.redisClient.disconnect();
+		this.removeAllListeners();
```
nizans, line 200:
```suggestion

```

packages/brokers/src/brokers/redis/BaseRedis.ts
nizans: OK sound right.
Removed then

packages/brokers/src/brokers/redis/BaseRedis.ts
```diff
@@ -185,6 +194,7 @@ export abstract class BaseRedisBroker<TEvents extends Record<string, any>>
 	 * Destroys the broker, closing all connections
 	 */
 	public async destroy() {
+		await this.unsubscribe(Array.from(this.subscribedEvents));
```
didinele, line 197:
```suggestion
		await this.unsubscribe(...this.subscribedEvents);
```

think we don't really use `Array.from`.

didinele (changes requested):

packages/brokers/src/brokers/redis/BaseRedis.ts
nizans: then it has to be 
`await this.unsubscribe([...this.subscribedEvents]);`
`unsubscribe` expecting an array 
pushed a commit

nizans: Or we can change `unsubscribe` to spread the `args` for a nicer API IMO.

didinele: Not a fan of spread.

> then it has to be
await this.unsubscribe([...this.subscribedEvents]);
unsubscribe expecting an array
pushed a commit

ah, yup.

kyranet (approved):

iCrawl (approved):

PR #10314, opened 2024-06-02
revert: refactor: native zlib support

Jiralite:
This reverts commit 20258f94bf1a62ec1989ef04d839e2800f6e6d28.

This is to ensure discord.js gets #10271 through a non-breaking version.

PR #10316, opened 2024-06-02
refactor: native zlib support

sdanialraza:
This reverts #10314 which was a revert of #10243

### DIFF
--- a/packages/ws/README.md
+++ b/packages/ws/README.md
@@ -50,7 +50,10 @@ const manager = new WebSocketManager({
 	intents: 0, // for no intents
 	rest,
 	// uncomment if you have zlib-sync installed and want to use compression
-	// compression: CompressionMethod.ZlibStream,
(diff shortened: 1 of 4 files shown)


vladfrangu (approved):

almeidx (approved):

Jiralite (approved):

dzlandis:
How are users of the regular `discord.js` package supposed to enable native zlib support?
Previously it was auto enabled (at least tmk) when the user had the zlib-sync package installed. But how does one enable it with the native package?

Qjuh:
This will be in a new major version of `@discordjs/ws` because of breaking changes. And because of that won’t be in mainlib discord.js before the next major version either. And by then the `@discordjs/ws` would be used directly in mainlib, not Frankensteined in like I did for the non-breaking change in v14. So then you‘d just set it with the ws options on ClientOptions

PR #10317, opened 2024-06-02
chore: Revamp of labels

Jiralite:
> [!Important]  
> The label synchronisation workflow must be temporarily disabled for this to merge.

- Rather than the "interactions" and "threads" labels which were the only kind of feature-related labels we had, they are now a "feature" label. Added as organised from the core package with a few others
  - "api changes" and "api support" have been deduplicated to "t:discord" (an update to the Discord API may require a change or a new feature to support)
- Added descriptions to some labels
- We will use issue types!

Some labels were removed. Examples:
- "discussion" was removed as GitHub Discussions exist
- "question (please use Discord instead)" was removed as GitHub Discussions exist
  - If a discussion is not required, then this should be labelled as invalid
- "bug" and "feature request" have been replaced with organisation-wide issue types
- Etc.

<hr>

### If `interactions` was renamed to `f:interactions`, wouldn't that remove all the labels from previous issues and pull requests?

Yes. However, that can be avoided with the following steps:

1. Disable the label synchronisation workflow before this pull request is merged (this is harmless)
2. Rename the label
3. Enable the workflow
4. Optionally run the workflow

crazy-max/ghaction-github-labeler will then identify that `f:interactions` exists and will not delete any labels, but make amendments, if any.

### How will the deduplication of "api changes" and "api support" work?

Similarily to the above question, we can rename "api changes" to "t:discord" and those with that label will be updated. However, we cannot rename "api support" to "t:discord" as label names must be unique.

We can automatically modify all those issues and pull requests. For example, with issues:

```
gh issue list --json number --label "api support" --limit 1000 --state all | jq --raw-output '.[].number' | while read -r issue_number; do
  gh issue edit $issue_number --add-label "t:discord"
  gh issue edit $issue_number --remove-label "api support"
done
```

Or, we could do nothing. Both are effortless.

<hr>

I'll be doing the manual work once this merges. No pressure on anyone else!

This is a time to make any other label changes! Open to suggestions.

### DIFF
--- a/.github/ISSUE_TEMPLATE/01-package_bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/01-package_bug_report.yml
@@ -1,6 +1,6 @@
 name: Bug report
 description: Report an issue with discord.js or another package.
-labels: [bug, need repro]
+labels: [t:bug, t:need reproduction]
 body:
   - type: markdown
     attributes:
--- a/.github/ISSUE_TEMPLATE/01-package_bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/01-package_bug_report.yml
@@ -1,6 +1,8 @@
 name: Bug report
 description: Report an issue with discord.js or another package.
-labels: [t:bug, t:need reproduction]
+labels:
+  - t:need reproduction
+type: Bug
 body:
   - type: markdown
     attributes:
--- a/.github/labels.yml
+++ b/.github/labels.yml
@@ -1,126 +1,149 @@
-- name: api changes
-  color: '5663e9'
-- name: api support
-  color: '5663e9'
-- name: apps:guide
+- name: a:api-extractor
+  description: Affects @discordjs/api-extractor.
   color: fbca04
-- name: apps:website
+- name: a:api-extractor-model
+  description: Affects @discordjs/api-extractor-model.
   color: fbca04
-- name: backlog
-  color: 7ef7ef
-- name: backport
-  color: 88aabb
-- name: backport-candidate
-  color: 0075ca
-- name: blocked
-  color: fc1423
-- name: bug
-  color: d73a4a
-- name: caching
-  color: 80c042
-- name: chore
-  color: ffffff
-- name: ci
-  color: 0075ca
-- name: dependencies
-  color: 276bd1
-- name: discord
-  color: '5663e9'
-- name: discussion
+- name: a:brokers
+  description: Affects @discordjs/brokers.
+  color: fbca04
+- name: a:builders
+  description: Affects @discordjs/builders.
+  color: fbca04
+- name: a:collection
+  description: Affects @discordjs/collection.
+  color: fbca04
+- name: a:core
+  description: Affects @discordjs/core.
+  color: fbca04
+- name: a:create-discord-bot
+  description: Affects create-discord-bot.
+  color: fbca04
+- name: a:discord.js
+  description: Affects discord.js.
+  color: fbca04
+- name: a:docgen
+  description: Affects @discordjs/docgen.
+  color: fbca04
+- name: a:formatters
+  description: Affects @discordjs/formatters.
+  color: fbca04
+- name: a:guide
+  description: Affects @discordjs/guide.
+  color: fbca04
+- name: a:next
+  description: Affects @discordjs/next.
+  color: fbca04
+- name: a:proxy
+  description: Affects @discordjs/proxy.
+  color: fbca04
+- name: a:proxy-container
+  description: Affects @discordjs/proxy-container.
+  color: fbca04
+- name: a:rest
+  description: Affects @discordjs/rest.
+  color: fbca04
+- name: a:structures
+  description: Affects @discordjs/structures.
+  color: fbca04
+- name: a:ui
+  description: Affects @discordjs/ui.
+  color: fbca04
+- name: a:util
+  description: Affects @discordjs/util.
+  color: fbca04
+- name: a:voice
+  description: Affects @discordjs/voice.
+  color: fbca04
+- name: a:website
+  description: Affects @discordjs/website.
+  color: fbca04
+- name: a:ws
+  description: Affects @discordjs/ws.
+  color: fbca04
+- name: f:application commands
+  description: Related to the application commands feature.
   color: b6b1f9
-- name: documentation
+- name: f:applications
+  description: Related to the applications feature.
+  color: b6b1f9
+- name: f:channel
+  description: Related to the channel feature.
+  color: b6b1f9
+- name: f:guild
+  description: Related to the guild feature.
+  color: b6b1f9
+- name: f:interactions
+  description: Related to the interactions feature.
+  color: b6b1f9
+- name: f:invite
+  description: Related to the invite feature.
+  color: b6b1f9
+- name: f:monetization
+  description: Related to the monetization feature.
+  color: b6b1f9
+- name: f:poll
+  description: Related to the poll feature.
+  color: b6b1f9
+- name: f:role connections
+  description: Related to the role connections feature.
+  color: b6b1f9
+- name: f:stage instances
+  description: Related to the stage instances feature.
+  color: b6b1f9
+- name: f:stickers
+  description: Related to the stickers feature.
+  color: b6b1f9
+- name: f:threads
+  description: Related to the threads feature.
+  color: b6b1f9
+- name: f:user
+  description: Related to the user feature.
+  color: b6b1f9
+- name: f:voice
+  description: Related to the voice feature.
+  color: b6b1f9
+- name: f:webhook
+  description: Related to the webhook feature.
+  color: b6b1f9
+- name: s:major
+  description: This is a major change.
+  color: c10f47
+- name: s:minor
+  description: This is a minor change.
+  color: e4f486
+- name: s:patch
+  description: This is a patch change.
+  color: e8be8b
+- name: t:backport
+  color: 88aabb
+- name: t:backport-candidate
   color: 0075ca
-- name: duplicate
+- name: t:bug
+  description: Something is not working as expected.
+  color: '870808'
+- name: t:duplicate
   color: cfd3d7
-- name: error handling
-  color: 80c042
-- name: feature request
+- name: t:feature request
   color: fcf95a
-- name: gateway
-  color: 80c042
+- name: t:invalid
+  description: This is not related to discord.js.
+  color: e4e669
+- name: t:need reproduction
+  color: c66037
+- name: t:won't fix
+  description: No work will be done towards this issue.
+  color: cfd3d7
+- name: blocked
+  description: Generic label to stop Kodiak from merging a pull request.
+  color: fc1423
+- name: discord
+  description: This is a Discord-related issue.
+  color: 5865f2
 - name: good first issue
   color: 7057ff
 - name: help wanted
   color: '008672'
-- name: in progress
-  color: ffccd7
 - name: in review
+  description: This pull request is currently in review; halts Kodiak.
   color: aed5fc
-- name: interactions
-  color: 80c042
-- name: invalid
-  color: e4e669
-- name: need repro
-  color: c66037
-- name: packages:api-extractor
-  color: fbca04
-- name: packages:api-extractor-model
-  color: fbca04
-- name: packages:brokers
-  color: fbca04
-- name: packages:builders
-  color: fbca04
-- name: packages:collection
-  color: fbca04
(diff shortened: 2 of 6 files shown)


ckohen:
Please PR an *add* to the label possibilities in the middleware first. We can let it run checking for both sets of labels until this is all settled, and another PR can be made to remove the old labels at that point.

Jiralite:
Opened https://github.com/discordjs/gh-discord-middleware/pull/23.

packages/scripts/turbo/generators/config.ts
```diff
@@ -71,7 +70,7 @@ export default function generator(plop: PlopTypes.NodePlopAPI): void {
 				path: `${plop.getDestBasePath()}/../../.github/labeler.yml`,
 				transform(content, answers) {
 					const labelerYAML = parseYAML(content) as Record<string, string[]>;
-					labelerYAML[`packages:${answers.name}`] = [`packages/${answers.name}/*`, `packages/${answers.name}/**/*`];
+					labelerYAML[`package:${answers.name}`] = [`package/${answers.name}/*`, `package/${answers.name}/**/*`];
```
ckohen, line 73:
```suggestion
					labelerYAML[`package:${answers.name}`] = [`packages/${answers.name}/*`, `packages/${answers.name}/**/*`];
```

These paths stay the same??

ckohen (dismissed):

packages/scripts/turbo/generators/config.ts
Jiralite: Actually, both are wrong due to #10272. I'll commit this suggestion, but fixing this would ideally be done in another pull request. It's pretty low priority.

ckohen: ah, right

Jiralite: #10332

packages/brokers/cliff.toml
```diff
@@ -45,7 +45,6 @@ commit_parsers = [
 	{ message = "^docs", group = "Documentation"},
 	{ message = "^perf", group = "Performance"},
 	{ message = "^refactor", group = "Refactor"},
-	{ message = "^typings", group = "Typings"},
```
vladfrangu, line 48: This removal feels out of scope for this PR 👀

Jiralite: Added back with 030d60e22864445ff39c6553b56434ffd647b86c.

.github/labels.yml
```diff
@@ -1,56 +1,31 @@
...
-- name: caching
-  color: 80c042
-- name: chore
-  color: ffffff
-- name: ci
-  color: 0075ca
-- name: dependencies
-  color: 276bd1
+  description: Something is not working as expected.
+  color: '000000'
```
ckohen, line 10: Is this color change intentional?

.github/labels.yml
```diff
@@ -88,37 +63,88 @@
...
+  color: b6b1f9
+- name: type:user
+  color: b6b1f9
+- name: type:voice
+  color: b6b1f9
+- name: type:webhook
+  color: b6b1f9
+- name: won't fix
+  description: No work will be done towards this issue.
+  color: cfd3d7
```
ckohen, line 150: this is the same color as duplicate, is that right? (very well could make sense)

Jiralite: Yeah, I intentionally made them the same colour.

.github/labels.yml
Jiralite: I feel black would stand out more as a bug. Welcome to feedback on this.

ckohen: I don't think i've actually seen black in the ui, so....what's it look like 🤷

didinele (approved):
not sure on the black for bug but otherwise lgtm

.github/labels.yml
iCrawl: Absolutely not

.github/labels.yml
```diff
@@ -1,126 +1,149 @@
...
+- name: t:need reproduction
+  color: c66037
+- name: t:won't fix
+  description: No work will be done towards this issue.
+  color: cfd3d7
+- name: blocked
+  description: Generic label to stop Kodiak from merging a pull request.
+  color: fc1423
+- name: discord
+  description: This is a Discord-related issue.
```
almeidx, line 131: What does this even mean? Isn't everything on this repository related to discord in some way 😅

.github/labels.yml
```diff
@@ -1,126 +1,149 @@
...
-- name: wontfix
-  color: ffffff
+- name: t:backport
+  color: 88aabb
+- name: t:backport-candidate
+  color: 0075ca
+- name: t:bug
+  description: Something is not working as expected.
+  color: '870808'
+- name: t:duplicate
```
almeidx: Is it really that important that we label things as "duplicate"s? Since there are now different types of issue closure types, including one for duplicates (not only that, but still)

.github/labels.yml
```diff
@@ -1,126 +1,149 @@
...
+  color: cfd3d7
+- name: t:feature request
+  color: fcf95a
+- name: t:invalid
```
almeidx: Similar to `duplicate` label

.github/labels.yml
```diff
@@ -1,126 +1,149 @@
...
+  color: fc1423
+- name: discord
+  description: This is a Discord-related issue.
+  color: 5865f2
+- name: good first issue
+  color: 7057ff
+- name: help wanted
+  color: '008672'
+- name: in review
+  description: This pull request is currently in review; halts Kodiak.
```
almeidx, line 138: Despite this description not being wrong, I think it could better describe the reasoning behind this label.

How about something like:

```suggestion
  description: Pull requests that require more than the usual number of reviewers; blocks automatic merge.
```

.github/labels.yml
```diff
@@ -1,126 +1,149 @@
...
+- name: blocked
+  description: Generic label to stop Kodiak from merging a pull request.
+  color: fc1423
+- name: discord
+  description: This is a Discord-related issue.
+  color: 5865f2
+- name: good first issue
+  color: 7057ff
+- name: help wanted
+  color: '008672'
```
almeidx, line 136: I've never seen us use this label. Quick search reveals there are no issues/pull requests with it.
Might as well get rid of it

```suggestion
```

.github/labels.yml
```diff
@@ -1,126 +1,149 @@
...
+- name: t:invalid
+  description: This is not related to discord.js.
+  color: e4e669
+- name: t:need reproduction
+  color: c66037
+- name: t:won't fix
+  description: No work will be done towards this issue.
+  color: cfd3d7
+- name: blocked
+  description: Generic label to stop Kodiak from merging a pull request.
```
almeidx, line 128:
```suggestion
  description: Pull requests that are not ready to be merged; blocks automatic merge.
```

.github/labels.yml
```diff
@@ -1,126 +1,149 @@
...
+- name: f:interactions
+  description: Related to the interactions feature.
+  color: b6b1f9
+- name: f:invite
+  description: Related to the invite feature.
+  color: b6b1f9
+- name: f:monetization
+  description: Related to the monetization feature.
+  color: b6b1f9
+- name: f:poll
```
almeidx, line 85: I think this and stickers could instead be grouped under a "message" label

.github/labels.yml
```diff
@@ -1,126 +1,149 @@
...
+  description: Related to the poll feature.
+  color: b6b1f9
+- name: f:role connections
+  description: Related to the role connections feature.
+  color: b6b1f9
+- name: f:stage instances
```
almeidx, line 91: There isn't a whole lot about this specifically. I think it could be grouped under channel

.github/labels.yml
```diff
@@ -1,126 +1,149 @@
...
-- name: api support
-  color: '5663e9'
-- name: apps:guide
+- name: a:api-extractor
+  description: Affects @discordjs/api-extractor.
   color: fbca04
-- name: apps:website
+- name: a:api-extractor-model
+  description: Affects @discordjs/api-extractor-model.
   color: fbca04
```
almeidx, line 6: What about api-extractor-utils?

IMO we could group all these in a single api-extractor label.

.github/labels.yml
```diff
@@ -1,126 +1,149 @@
...
+- name: a:next
+  description: Affects @discordjs/next.
   color: fbca04
-- name: packages:next
+- name: a:proxy
+  description: Affects @discordjs/proxy.
   color: fbca04
-- name: packages:proxy
+- name: a:proxy-container
+  description: Affects @discordjs/proxy-container.
```
almeidx, line 41: IMO these could be grouped under a single "proxy" label

.github/ISSUE_TEMPLATE/01-package_bug_report.yml
```diff
@@ -1,6 +1,6 @@
 name: Bug report
 description: Report an issue with discord.js or another package.
-labels: [bug, need repro]
+labels: [t:bug, t:need reproduction]
```
almeidx: Is there a point in keeping the `bug` label? Seems like [issue types](https://docs.github.com/en/issues/tracking-your-work-with-issues/using-issues/managing-issue-types-in-an-organization) are the built-in replacement for it (and the "feature request" label)

almeidx (changes requested):
In general I'd be in favor of less clutter on the labels. Especially like the shorter package labels.

IMO I'd prefer not having prefixes for the remaining labels. The package labels are different since having more than one is common, so it's nice that they are grouped. This is not really the case with the other labels. Not sure what others think about this

.github/labels.yml
Jiralite: Anything that is a Discord issue. Does this help?

https://github.com/discordjs/discord.js/issues?q=sort:updated-desc%20is:issue%20state:closed%20label:discord

For example, opening an issue that turns out to be a Discord issue instead. It may be useful to know!

.github/labels.yml
Jiralite: Can we search for them? If so, then maybe not. If we cannot, then it could have some merit?

.github/labels.yml
Jiralite: It's a special label by GitHub, actually. It shows up in search and whatnot that there are X issues that need help. Example is discord-api-types:

image

It could be useful in the future. The reason why you might not see it labelled currently is because it was removed from issues it once was on.

.github/labels.yml
Jiralite: Probably a good idea!

.github/ISSUE_TEMPLATE/01-package_bug_report.yml
Jiralite: ~~Would love to hear further input on this because I am not sure.~~ https://canary.discord.com/channels/222078108977594368/999796393810067607/1414001885102215279

.github/labels.yml
Jiralite: We could, but we do have packages/core/src/api/stageInstances.ts. So whilst that file still exists, this label does have a use!

.github/labels.yml
Jiralite: packages/core/src/api/poll.ts exists, and it's a separate resource on Discord's documentation so...

Same with stickers.

.github/labels.yml
almeidx: yes 
image

.github/labels.yml
almeidx: Fair enough

.github/labels.yml
almeidx: Disregard as it's a separate resource on discord's end

.github/labels.yml
almeidx: > separate resource on Discord's documentation

Ah, that's the reasoning for it.. Didn't realize it

.github/labels.yml
almeidx: I don't think the label is very self-explanatory and the description as is seems weird.
Perhaps removing "-related" would make it more clear (as you just said in the reply), or adding some more context

Jiralite: I'm thinking of our existing api-changes and api support labels. We could rename this label to "t:discord" and then use it for both issues (they opened an issue but it's actually for Discord) and pull requests (the reason for the pull request is a change in Discord). Or, we could have 2 labels: 1 for issues and 1 for pull requests. Thoughts? Thinking how best to say an issue is actually an upstream issue and the reason for a pull request is because Discord changed something (whether that is adding something new or a surprise breaking change).

PR #10325, opened 2024-06-05
ci: better release workflow

ckohen:

Creates a proper release workflow, works in ci as well as locally.

This 
- determines the release tree
- releases all packages in scope, or all packages necessary to release the specified package
- if a package is already released at a version, the registry check will allow it to skip publishing that package
-  sanity checks publish completion before moving to the next branch of the release tree
- creates the tag and publishes a release on github

### DIFF
--- a/.github/workflows/publish-dev.yml
+++ b/.github/workflows/publish-dev.yml
@@ -3,6 +3,10 @@ on:
   schedule:
     - cron: '0 */12 * * *'
   workflow_dispatch:
+    inputs:
+      dry_run:
+        type: boolean
+        default: false
 jobs:
   npm-publish:
     name: npm publish
@@ -55,32 +59,18 @@ jobs:
           node-version: 22
           registry-url: https://registry.npmjs.org/
 
-      - name: Check the current development version
-        id: release-check
-        run: |
-          if [[ $(npm view ${{ matrix.package }}@dev version | grep -e "$(git rev-parse --short HEAD)") ]]; \
-            then echo "RELEASE=0" >> "$GITHUB_OUTPUT"; \
-            else echo "RELEASE=1" >> "$GITHUB_OUTPUT"; \
-          fi
-
       - name: Install dependencies
-        if: steps.release-check.outputs.release == '1'
         uses: ./packages/actions/src/pnpmCache
 
       - name: Build dependencies
-        if: steps.release-check.outputs.release == '1'
         run: pnpm run build
 
-      - name: Publish package
-        if: steps.release-check.outputs.release == '1'
-        run: |
-          pnpm --filter=${{ matrix.package }} run release --preid "dev.$(date +%s)-$(git rev-parse --short HEAD)" --skip-changelog
-          pnpm --filter=${{ matrix.package }} publish --provenance --no-git-checks --tag dev || true
+      - name: Pubish packages
+        uses: ./packages/actions/src/releasePackages
+        with:
+          exclude: 'create-discord-bot,@discordjs/docgen'
+          dry: ${{ inputs.dry_run }}
+          dev: true
         env:
           NODE_AUTH_TOKEN: ${{ secrets.NPM_PUBLISH_TOKEN }}
-
-      - name: Deprecate prior development releases
-        if: steps.release-check.outputs.release == '1'
-        run: pnpm exec npm-deprecate --name "*dev*" --message "This version is deprecated. Please use a newer version." --package ${{ matrix.package }}
-        env:
-          NODE_AUTH_TOKEN: ${{ secrets.NPM_PUBLISH_TOKEN }}
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/publish-dev.yml
+++ b/.github/workflows/publish-dev.yml
@@ -68,7 +68,7 @@ jobs:
       - name: Pubish packages
         uses: ./packages/actions/src/releasePackages
         with:
-          exclude: 'create-discord-bot,@discordjs/docgen'
+          exclude: '@discordjs/docgen'
           dry: ${{ inputs.dry_run }}
           dev: true
         env:
--- a/.github/workflows/publish-dev.yml
+++ b/.github/workflows/publish-dev.yml
@@ -65,7 +65,7 @@ jobs:
       - name: Build dependencies
         run: pnpm run build
 
-      - name: Pubish packages
+      - name: Publish packages
         uses: ./packages/actions/src/releasePackages
         with:
           exclude: '@discordjs/docgen'
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -26,6 +26,9 @@ on:
         required: false
         type: string
         default: '@discordjs/docgen,@discordjs/next'
+      dry_run:
+        type: boolean
+        default: false
 jobs:
   npm-publish:
     name: npm publish
@@ -50,13 +53,12 @@ jobs:
       - name: Install dependencies
         uses: ./packages/actions/src/pnpmCache
 
-      - name: Build dependencies
-        run: pnpm run build
-
       - name: Release packages
         uses: ./packages/actions/src/releasePackages
         with:
           package: ${{ inputs.package }}
+          exclude: ${{ inputs.exclude }}
+          dry: ${{ inputs.dry_run }}
         env:
           NODE_AUTH_TOKEN: ${{ secrets.NPM_PUBLISH_TOKEN }}
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,56 @@
+name: Release
+on:
+  workflow_dispatch:
+    inputs:
+      package:
+        description: 'The published name of a single package to release'
+        type: choice
+        options:
+          - discord.js
+          - '@discordjs/brokers'
+          - '@discordjs/builders'
+          - '@discordjs/collection'
+          - '@discordjs/core'
+          - '@discordjs/create-discord-bot'
+          - '@discordjs/docgen'
+          - '@discordjs/formatters'
+          - '@discordjs/next'
+          - '@discordjs/proxy'
+          - '@discordjs/rest'
+          - '@discordjs/util'
+          - '@discordjs/voice'
+          - '@discordjs/ws'
+jobs:
+  npm-publish:
+    name: npm publish
+    runs-on: ubuntu-latest
+    permissions:
+      id-token: write
+      contents: write
+    env:
+      TURBO_TOKEN: ${{ secrets.TURBO_TOKEN }}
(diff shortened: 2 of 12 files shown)


packages/actions/src/releasePackages/generateReleaseTree.ts
```diff
@@ -0,0 +1,125 @@
...
+	version?: string;
+}
+
+export interface ReleaseEntry {
+	dependsOn?: string[];
+	name: string;
+	version: string;
+}
+
+async function getReleaseEntries() {
```
almeidx, line 19: This function could leverage the `pnpm list` command instead of manually finding the packages and their dependencies, e.g.:

```typescript
function getReleaseEntries(): ReleaseEntry[] {
	const pkgList = JSON.parse(execSync("pnpm list --recursive --only-projects --prod --json").toString());
	return pkgList
		.filter((pkg) => !pkg.private && !nonNodePackages.has(pkg.name))
		.map((pkg) => ({
			name: pkg.name,
			version: pkg.version,
			dependsOn: pkg.dependencies ? Object.keys(pkg.dependencies) : [],
		}));
}
```

ckohen: I do like this, but I really don't like that it gives me `@discordjs/discordjs` (the monorepo), and the apps directory. Right now, they're all marked private appropriately, so it'd work just fine, just rubs me  the wrong way a bit.

ckohen: oh! I can still use --filter with this, works great

.github/workflows/release.yml
```diff
@@ -0,0 +1,56 @@
...
+      package:
+        description: 'The published name of a single package to release'
+        type: choice
+        options:
+          - discord.js
+          - '@discordjs/brokers'
+          - '@discordjs/builders'
+          - '@discordjs/collection'
+          - '@discordjs/core'
+          - '@discordjs/create-discord-bot'
```
Jiralite, line 14: create-discord-bot isn't scoped under discord.js.

packages/actions/src/releasePackages/index.ts
```diff
@@ -0,0 +1,18 @@
+import process from 'node:process';
+import { getInput } from '@actions/core';
+import { generateReleaseTree } from './generateReleaseTree.js';
+import { releasePackage } from './releasePackage.js';
+
+const argInput = process.argv[2];
+const useableArg = argInput === 'discord.js' || argInput?.startsWith('@discordjs/');
+const packageName = getInput('package') || useableArg ? argInput : '';
```
Jiralite, line 8: Is it actually possible that `getInput('package')` will be falsy? Also, should `useableArg` account for create-discord-bot?

ckohen: yes, getInput returns an empty string when no input is passed (so when run from the command line and when you don't select a specific package in the workflow dispatch)

.github/workflows/release.yml
```diff
@@ -0,0 +1,62 @@
...
+          - '@discordjs/voice'
+          - '@discordjs/ws'
+      exclude:
+        description: 'Comma separated list of packages to exclude from release (if not depended upon)'
+        required: false
+        type: string
+        default: '@discordjs/docgen,@discordjs/next'
+jobs:
+  npm-publish:
+    name: npm publish
```
Jiralite, line 31: Would probably just say "publish" as we use pnpm and not npm.

packages/actions/src/releasePackages/index.ts
Jiralite, line 1: Is it worth using chalk for the colours in logging? Not a big thing though.

.github/workflows/release.yml
ckohen: it's publishing to npm

packages/actions/src/releasePackages/index.ts
ckohen: not in ci, could potentially see it used if you use the manual cli version, but don't think colors belong in ci

Jiralite (approved):

packages/actions/src/releasePackages/releasePackage.ts
```diff
@@ -0,0 +1,72 @@
...
+
+	const before = performance.now();
+
+	// Poll registry to ensure next publishes won't fail
+	await new Promise<void>((resolve) => {
+		const interval = setInterval(async () => {
+			if (await checkRegistry(release)) {
+				clearInterval(interval);
+				resolve();
+			}
```
kyranet, line 64: This is missing a `return`, so the code below isn't called:
```suggestion
				resolve();
				return;
			}
```

packages/actions/src/releasePackages/releasePackage.ts
```diff
@@ -0,0 +1,72 @@
...
+
+			if (performance.now() > before + 5 * 60 * 1_000) {
+				clearInterval(interval);
+				throw new Error(`Release for ${release.name} failed.`);
```
kyranet, line 68: Shouldn't the error be passed to the `reject()` callback? It's inside a `setInterval` callback so I doubt it'll bubble up to the `Promise`

kyranet (approved):

kyranet (dismissed):
Sorry, pressed the wrong button 😅

packages/actions/src/releasePackages/releasePackage.ts
ckohen: functionally, the idea is to cause the process to crash, but in case we use this elsewhere that would be the proper way to do it yeah

packages/actions/src/releasePackages/releasePackage.ts
```diff
@@ -0,0 +1,73 @@
...
+
+async function checkRegistry(release: ReleaseEntry) {
+	const res = await fetch(`https://registry.npmjs.org/${release.name}/${release.version}`);
+	return res.ok;
+}
+
+async function execPromise(execString: string) {
+	const child = exec(execString);
+	child.stdout?.pipe(process.stdout);
+	child.stderr?.pipe(process.stderr);
```
vladfrangu, line 21: Use the exec options to automatically map these please

vladfrangu (dismissed):

packages/actions/src/releasePackages/releasePackage.ts
```diff
@@ -0,0 +1,77 @@
...
+async function checkRegistry(release: ReleaseEntry) {
+	const res = await fetch(`https://registry.npmjs.org/${release.name}/${release.version}`);
+	return res.ok;
+}
+
+async function gitTagAndRelease(release: ReleaseEntry, dry: boolean) {
+	const tagName = `${release.name === 'discord.js' ? `` : `${release.name}@`}${release.version}`;
+	// Don't throw, if this exits non-zero it's probably because the tag already exists
+	await $`git tag ${tagName}`.nothrow();
+	if (dry) {
```
vladfrangu, line 23:
```suggestion
	await $`git tag ${tagName}`.nothrow();

	if (dry) {
```

packages/actions/src/releasePackages/releasePackage.ts
```diff
@@ -0,0 +1,77 @@
...
+		info(`[DRY] Tag "${tagName}" created, skipping push and release creation.`);
+		return;
+	}
+
+	await $`git push --tags`;
+	try {
```
vladfrangu, line 29:
```suggestion
	await $`git push --tags`;

	try {
```

vladfrangu (changes requested):

packages/actions/src/releasePackages/releasePackage.ts
```diff
@@ -0,0 +1,86 @@
...
+			}
+		}, 15_000);
+	});
+
+	if (dev) {
+		// Send and forget, deprecations are less important than releasing other dev versions and can be done manually
+		void $`pnpm exec npm-deprecate --name "*dev*" --message "This version is deprecated. Please use a newer version." --package ${release.name}`
+			.nothrow()
+			// eslint-disable-next-line promise/prefer-await-to-then
+			.then(() => {});
```
almeidx, line 84: This can probably be removed?
```suggestion
			.nothrow();
```

ckohen: Either have to await or call then (which await does). This isn't a real promise

packages/actions/src/releasePackages/index.ts
```diff
@@ -0,0 +1,47 @@
+import { getInput, startGroup, endGroup, getBooleanInput, info } from '@actions/core';
+import { program } from 'commander';
+import { generateReleaseTree } from './generateReleaseTree.js';
+import { releasePackage } from './releasePackage.js';
+
+const excludeInput = getInput('exclude');
```
vladfrangu, line 6: Won't this also throw like devInput/dryRun, and thus would need a try catch>

ckohen: Only booleans throw. This returns an empty string when not present (and the Boolean input actually uses this internally)

.github/workflows/publish-dev.yml
```diff
@@ -53,32 +29,18 @@ jobs:
...
-        env:
-          NODE_AUTH_TOKEN: ${{ secrets.NPM_PUBLISH_TOKEN }}
-
-      - name: Deprecate prior development releases
-        if: steps.release-check.outputs.release == '1'
-        run: pnpm exec npm-deprecate --name "*dev*" --message "This version is deprecated. Please use a newer version." --package ${{ matrix.package }}
+      - name: Pubish packages
+        uses: ./packages/actions/src/releasePackages
+        with:
+          exclude: 'create-discord-bot,@discordjs/docgen'
```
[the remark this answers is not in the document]
ckohen: good question, this is the existing workflow though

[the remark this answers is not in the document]
Jiralite: It was leaned away from because it's not a package like the others in a traditional sense and it has no documentation. I would be impartial to it having development releases.

vladfrangu (approved):

.github/workflows/publish-dev.yml
```diff
@@ -53,32 +29,18 @@ jobs:
...
-        run: |
-          pnpm --filter=${{ matrix.package }} run release --preid "dev.$(date +%s)-$(git rev-parse --short HEAD)"
-          pnpm --filter=${{ matrix.package }} publish --provenance --no-git-checks --tag dev || true
-        env:
-          NODE_AUTH_TOKEN: ${{ secrets.NPM_PUBLISH_TOKEN }}
-
-      - name: Deprecate prior development releases
-        if: steps.release-check.outputs.release == '1'
-        run: pnpm exec npm-deprecate --name "*dev*" --message "This version is deprecated. Please use a newer version." --package ${{ matrix.package }}
+      - name: Pubish packages
```
Qjuh, line 38:
```suggestion
      - name: Publish packages
```

Qjuh (changes requested):

Qjuh (approved):

iCrawl (approved):

Issue #10326, opened 2024-06-06
Emit poll events on uncached messages


uhKevinMC:
### Which package is this bug report for?

discord.js

### Issue description

No poll events are emitted when the message attached to the poll is uncached (regardless of partials). There was a discussion about it in the discord server [here](https://discord.com/channels/222078108977594368/1247758115340947510) and the solution for now was to use the `raw` event to fetch the message, then resolve the poll data.

*Slightly* inconvenient for now as this temporary fix is not particularly type-friendly with my events framework and there is some slightly unexpected behavior from the event not being emitted, even with partials enabled.

### Code sample

```ts
//pretend we received a message that has a poll object, now we go to sweep the messages cache
client.channels.cache.get('insertIdHere').messages.cache.sweep(() => true)

//vote on the poll attached to a sweeped, uncached message

client.on(Events.MessagePollVoteAdd, (answer: PollAnswer, userId: Snowflake) => {
  console.log(answer, userId)
  //does not emit
})
```

### Versions

-discord.js v14.15.3

### Issue priority

Low (slightly annoying)

### Which partials do you have configured?

User, Channel, GuildMember, Message

### Which gateway intents are you subscribing to?

Guilds, GuildMembers, GuildMessages, DirectMessages, MessageContent, GuildMessagePolls, DirectMessagePolls

### I have tested this issue on a development release

94cc02a2580496774d75673abc0caabc765d9ee0

PR #10328, opened 2024-06-06
feat: polls overhaul

uhKevinMC:

This PR resolves Issue #10326 and adds methods and properties found in similarly structured classes. This also creates a new manager, dubbed the `PollAnswerVoterManager` to store users that voted for the poll answer, similar to `ReactionUserManager` for users that reacted to a reaction.

This PR also patches a bug related to poll expiry dates being nullish in the future when polls have the option not to expire.

*please do give feedback on this PR, I don't bite*

- *[x] Add Poll Answer Voter/User Manager
- *[x] Add partials for `Poll` and `PollAnswer`
- *[x] Add support for partials `Poll` and `PollAnswer` in respective events
- *[x] Add typings for new properties, methods, classes, and partials
- *[x] Add tests
- *[x] Case testing
- *[x] Sanity check
- *[x] Finalize PR


uhKevinMC:
I'm not able to test this locally, gotta love yarn..

Feedback on this would be appreciated!

packages/discord.js/src/client/actions/MessagePollVoteAdd.js
```diff
@@ -13,9 +16,26 @@ class MessagePollVoteAddAction extends Action {
 
     const { poll } = message;
 
-    const answer = poll?.answers.get(data.answer_id);
-    if (!answer) return false;
+    const includePollPartial = this.client.options.partials.includes(Partials.Poll);
+    if (message.partial && !includePollPartial) return false;
 
+    if (!poll && includePollPartial) {
+      message.poll = new Poll(this.client, { ...data, partial: true }, message, channel);
```
Jiralite, line 23: `partial: true` appears to be unused.

uhKevinMC: This *should* be fixed now, added logic in `_patch()` to handle `partial` as a property in `data`

packages/discord.js/src/structures/Poll.js
```diff
@@ -48,9 +69,9 @@ class Poll extends Base {
 
     /**
      * The timestamp when this poll expires
-     * @type {number}
+     * @type {?number}
      */
-    this.expiresTimestamp = Date.parse(data.expiry);
+    this.expiresTimestamp = Date.parse(data.expiry) ?? null;
```
uhKevinMC, line 74: This change here is to prevent a hard crash once polls without expiry dates roll out, and it is also marked as nullish on the API documentation

packages/discord.js/src/client/actions/MessagePollVoteAdd.js
```diff
@@ -11,11 +14,33 @@ class MessagePollVoteAddAction extends Action {
...
+        channel,
+      );
+    }
+
+    const includePollAnswerPartial = this.client.options.partials.includes(Partials.PollAnswer);
+    if (message.partial && includePollPartial && !includePollAnswerPartial) return false;
+
+    let answer = message.poll?.answers?.get(data.answer_id);
+    if (!answer && message.poll) {
+      const pollAnswer = new PollAnswer(this.client, data, message.poll);
```
uhKevinMC, line 34: I'll combine these two poll partials `if` statements so there isn't any unnecessary extra work when either partial is disabled.

packages/discord.js/src/managers/PollAnswerVoterManager.js
```diff
@@ -0,0 +1,62 @@
...
+
+  /**
+   * Options used to fetch voters who voted for this poll answer.
+   * @typedef {Object} FetchPollAnswerVotersOptions
+   * @property {number} [limit] The maximum amount of users to fetch
+   * @property {Snowflake} [after] The user id to fetch voters after
+   */
+
+  /**
+   * Fetches all the voters that voted on this poll answer. Resolves with a collection of users, mapped by their ids.
```
almeidx, line 38:
```suggestion
   * Fetches the users that voted on this poll answer. Resolves with a collection of users, mapped by their ids.
```

packages/discord.js/src/managers/PollAnswerVoterManager.js
```diff
@@ -0,0 +1,62 @@
...
+  }
+
+  /**
+   * The cache of this manager
+   * @type {Collection<Snowflake, User>}
+   * @name PollAnswerVoterManager#cache
+   */
+
+  /**
+   * Options used to fetch voters who voted for this poll answer.
```
almeidx, line 31:
```suggestion
   * Options used to fetch users who voted for this poll answer.
```

packages/discord.js/src/managers/PollAnswerVoterManager.js
```diff
@@ -0,0 +1,62 @@
...
+   * @typedef {Object} FetchPollAnswerVotersOptions
```
almeidx, line 32: This definition appears to be the same as `BaseFetchPollAnswerVotersOptions`

packages/discord.js/src/structures/Poll.js
```diff
@@ -79,6 +108,21 @@ class Poll extends Base {
         const answer = this.answers.get(answerResult.id);
         answer?._patch(answerResult);
       }
+    } else if (data.question) {
```
almeidx, line 111: I don't believe these should be `else if`s, but rather regular `if`s after the previous if-else that was here

almeidx (changes requested):

packages/discord.js/src/managers/PollAnswerVoterManager.js
uhKevinMC: Looks like I accidentally overrode a type with this one 😅

packages/discord.js/src/structures/Poll.js
```diff
@@ -93,6 +137,33 @@ class Poll extends Base {
...
+   * Whether or not this poll is a partial
+   * @type {boolean}
+   * @readonly
+   */
+  get partial() {
+    return (
+      typeof this.question.text !== 'string' ||
+      typeof this.layoutType !== 'number' ||
+      typeof this.allowMultiselect !== 'boolean' ||
+      this._partial
```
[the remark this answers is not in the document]
uhKevinMC: I removed the `typeof this.question.text` check since it needs to be populated always (even if it is partial, where `...question.text` is populated with `''`), for compatibility reasons.

The other typeof checks however are not nullable on the API, therefore it would make the poll partial if they aren't there, and I can't supply it in the action without making the data inaccurate.

By default, `this._partial` will always be true if the poll was created as a partial poll in the action. I did move `this._partial` to the top of the checks though

packages/discord.js/src/client/actions/MessagePollVoteAdd.js
```diff
@@ -11,11 +14,25 @@ class MessagePollVoteAddAction extends Action {
...
+
+    if (!message.poll && includePollPartial && includePollAnswerPartial) {
+      message.poll = new Poll(this.client, { ...data, answers: [], partial: true }, message, channel);
+
+      const pollAnswer = new PollAnswer(this.client, data, message.poll);
+
+      message.poll.answers.set(data.answer_id, pollAnswer);
+    }
+
+    const answer = message.poll.answers.get(data.answer_id);
```
almeidx, line 29: This is not resolved. This logic is duplicated in the MessagePollVoteRemove action, so move this to a method in the Action class instead.

packages/discord.js/src/structures/Message.js
```diff
@@ -409,11 +409,17 @@ class Message extends Base {
     }
 
     if (data.poll) {
+      const existing = this.poll;
+
+      if (existing && data.poll) {
+        existing._patch({ ...existing, ...data.poll, partial: false });
+      }
```
almeidx, line 416:
```suggestion
      if (this.poll) {
        this.poll._patch({ ...data.poll, partial: false });
      }
```

packages/discord.js/src/structures/Poll.js
```diff
@@ -82,15 +111,55 @@ class Poll extends Base {
...
       this.resultsFinalized ??= false;
     }
+
+    if (data.question) {
+      this.question = {
+        text: data.question.text,
+      };
+    }
+
+    if (data.answers) {
```
almeidx, line 121: Remove the `this.answers` update from the constructor in favor of this one. (You'll also need to move the initial collection set here)

packages/discord.js/src/structures/Poll.js
```diff
@@ -82,15 +111,55 @@ class Poll extends Base {
...
+   * @type {?Date}
    * @readonly
    */
   get expiresAt() {
-    return new Date(this.expiresTimestamp);
+    return this.expiresTimestamp && new Date(this.expiresTimestamp);
+  }
+
+  /**
+   * Whether or not this poll is a partial
```
almeidx, line 147:
```suggestion
   * Whether this poll is a partial
```

packages/discord.js/src/structures/PollAnswer.js
```diff
@@ -50,8 +57,12 @@ class PollAnswer extends Base {
        * @type {number}
        */
       this.voteCount = data.count;
+    } else if (data.poll_media?.text) {
```
almeidx, line 60: Like in the other `_patch()`, these should not be `else if`s

packages/discord.js/src/managers/PollAnswerVoterManager.js
```diff
@@ -0,0 +1,50 @@
...
+   * @type {Collection<Snowflake, User>}
+   * @name PollAnswerVoterManager#cache
+   */
+
+  /**
+   * Fetches the users that voted on this poll answer. Resolves with a collection of users, mapped by their ids.
+   * @param {BaseFetchPollAnswerVotersOptions} [options] Options for fetching the users
+   * @returns {Promise<Collection<Snowflake, User>>}
+   */
+  async fetch({ after, limit }) {
```
almeidx, line 35:
```suggestion
  async fetch({ after, limit } = {}) {
```

packages/discord.js/src/managers/PollAnswerVoterManager.js
```diff
@@ -0,0 +1,50 @@
...
+
+  /**
+   * The cache of this manager
+   * @type {Collection<Snowflake, User>}
+   * @name PollAnswerVoterManager#cache
+   */
+
+  /**
+   * Fetches the users that voted on this poll answer. Resolves with a collection of users, mapped by their ids.
+   * @param {BaseFetchPollAnswerVotersOptions} [options] Options for fetching the users
```
almeidx, line 32:
```suggestion
   * @param {BaseFetchPollAnswerVotersOptions} [options={}] Options for fetching the users
```

packages/discord.js/typings/index.d.ts
```diff
@@ -2609,16 +2609,27 @@ export interface PollQuestionMedia {
   text: string;
 }
 
+export class PollAnswerVoterManager extends CachedManager<Snowflake, User, UserResolvable> {
+  public constructor(answer: PollAnswer, iterable: Iterable<APIUser>);
+  public answer: PollAnswer;
+  public fetch(): Promise<Collection<Snowflake, User>>;
```
almeidx, line 2615:
```suggestion
  public fetch(options?: BaseFetchPollAnswerVotersOptions): Promise<Collection<Snowflake, User>>;
```

packages/discord.js/typings/index.d.ts
```diff
@@ -2630,11 +2641,13 @@ export interface BaseFetchPollAnswerVotersOptions {
...
   private _emoji: APIPartialEmoji | null;
-  public readonly poll: Poll;
+  public readonly poll: Poll | PartialPoll;
   public id: number;
   public text: string | null;
   public voteCount: number;
+  public voters: PollAnswerVoterManager;
   public get emoji(): GuildEmoji | Emoji | null;
+  public get partial(): false;
   public fetchVoters(options?: BaseFetchPollAnswerVotersOptions): Promise<Collection<Snowflake, User>>;
```
almeidx, line 2651: Needs to be marked as deprecated here too

packages/discord.js/typings/index.d.ts
```diff
@@ -2609,16 +2609,27 @@ export interface PollQuestionMedia {
...
-  public expiresTimestamp: number;
-  public get expiresAt(): Date;
+  public answers: Collection<number, PollAnswer | PartialPollAnswer>;
+  public expiresTimestamp: number | null;
+  public get expiresAt(): Date | null;
   public allowMultiselect: boolean;
   public layoutType: PollLayoutType;
   public resultsFinalized: boolean;
+  public get partial(): false;
+  public fetch(): Promise<Poll>;
```
almeidx, line 2632:
```suggestion
  public fetch(): Promise<this>;
```

packages/discord.js/src/structures/Poll.js
```diff
@@ -82,15 +111,55 @@ class Poll extends Base {
...
+          existing._patch(answer);
+        } else {
+          this.answers.set(answer.answer_id, new PollAnswer(this.client, answer, this));
+        }
+      }
+    }
+
+    if (data.partial) {
+      Object.defineProperty(this, '_partial', { value: data.partial });
+    }
```
almeidx, line 134: I believe this `_partial` property could just be removed. We should be able to keep only the `typeof` checks in the `partial` getter, which is also what's done in the rest of the entities that have partials

almeidx (changes requested):

packages/discord.js/src/client/actions/MessagePollVoteAdd.js
```diff
@@ -26,6 +43,8 @@ class MessagePollVoteAddAction extends Action {
      */
     this.client.emit(Events.MessagePollVoteAdd, answer, data.user_id);
 
+    const { poll } = message;
```
NotSugden, line 46: this feels a bit silly
```suggestion
    return { poll: message.poll };
```

packages/discord.js/src/client/actions/MessagePollVoteRemove.js
```diff
@@ -26,6 +44,8 @@ class MessagePollVoteRemoveAction extends Action {
      */
     this.client.emit(Events.MessagePollVoteRemove, answer, data.user_id);
 
+    const { poll } = message;
```
NotSugden, line 47: same

packages/discord.js/src/structures/Poll.js
```diff
@@ -11,9 +11,30 @@ const { ErrorCodes } = require('../errors/index');
...
+    this.channelId = data.channel_id;
+
+    /**
+     * The channel that this poll is in
+     * @name Poll#channel
+     * @type {TextBasedChannel}
+     * @readonly
+     */
+
+    Object.defineProperty(this, 'channel', { value: channel });
```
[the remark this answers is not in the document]
NotSugden: why isn't this a getter function ?

packages/discord.js/src/structures/PollAnswer.js
```diff
@@ -50,8 +57,12 @@ class PollAnswer extends Base {
...
+      this.text = data.poll_media.text;
+    } else if (data.poll_media?.emoji) {
+      Object.defineProperty(this, '_emoji', { value: data.poll_media.emoji });
     } else {
-      this.voteCount ??= 0;
+      this.voteCount ??= this.voters.cache.size ?? 0;
```
NotSugden, line 65:
```suggestion
      this.voteCount ??= this.voters.cache.size;
```

packages/discord.js/src/structures/Message.js
```diff
@@ -409,11 +409,15 @@ class Message extends Base {
...
+      if (this.poll) {
+        this.poll._patch({ ...data.poll });
+      }
+
       /**
        * The poll that was sent with the message
        * @type {?Poll}
        */
-      this.poll = new Poll(this.client, data.poll, this);
+      this.poll ??= new Poll(this.client, data.poll, this, this.channel);
```
almeidx, line 420: Maybe move this into an `else` block on the `if` above and drop the `??` to avoid evaluating `this.poll` twice in this branch

packages/discord.js/src/structures/Poll.js
```diff
@@ -26,31 +47,22 @@ class Poll extends Base {
...
-
     /**
      * The timestamp when this poll expires
-     * @type {number}
+     * @type {?number}
      */
-    this.expiresTimestamp = Date.parse(data.expiry);
+    this.expiresTimestamp = data.expiry && Date.parse(data.expiry);
 
     /**
```
almeidx, line 67: Due to partials, I believe the `allowMultiselect` and `layoutType` updates below should be moved into `_patch()`

packages/discord.js/src/client/actions/Action.js
```diff
@@ -63,6 +65,22 @@ class GenericAction {
     );
   }
 
+  getPoll(data, message, channel) {
+    const includePollPartial = this.client.options.partials.includes(Partials.Poll);
+    const includePollAnswerPartial = this.client.options.partials.includes(Partials.PollAnswer);
+    if (message.partial && (!includePollPartial || !includePollAnswerPartial)) return null;
+
+    if (!message.poll && includePollPartial && includePollAnswerPartial) {
+      message.poll = new Poll(this.client, { ...data, answers: [] }, message, channel);
```
almeidx, line 74: Not needed anymore
```suggestion
      message.poll = new Poll(this.client, data, message, channel);
```

packages/discord.js/typings/index.d.ts
```diff
@@ -6542,6 +6558,14 @@ export interface PartialMessage
 
 export interface PartialMessageReaction extends Partialize<MessageReaction, 'count'> {}
 
+export interface PartialPoll
+  extends Partialize<
+    Poll,
+    'allowMultiselect' | 'answers' | 'layoutType' | 'expiresTimestamp' | 'question' | 'resultsFinalized' | 'message'
+  > {}
```
almeidx, line 6565: I believe this matches the current behavior. Type tests will need to be updated

```suggestion
export interface PartialPoll
  extends Partialize<
    Poll,
    'allowMultiselect' | 'layoutType' | 'expiresTimestamp',
    null,
    'question' | 'message'
  > {
    question: { text: null };
    message: PartialMessage;
  }
```

packages/discord.js/typings/index.d.ts
```diff
@@ -6542,6 +6558,14 @@ export interface PartialMessage
...
+
+export interface PartialPollAnswer extends Partialize<PollAnswer, 'voteCount' | 'emoji' | 'text'> {}
```
almeidx, line 6567:
```suggestion
export interface PartialPollAnswer extends Partialize<PollAnswer, 'emoji' | 'text', null, 'poll'> {
  readonly poll: PartialPoll;
}
```

almeidx (changes requested):

packages/discord.js/src/structures/Message.js
```diff
@@ -409,11 +409,15 @@ class Message extends Base {
     }
 
     if (data.poll) {
+      if (this.poll) {
+        this.poll._patch({ ...data.poll });
```
almeidx, line 413: Unnecessary clone
```suggestion
        this.poll._patch(data.poll);
```

almeidx:
Can you integrate these changes in this pull request too?
- https://github.com/discord/discord-api-docs/pull/6968

almeidx (approved):

packages/discord.js/src/structures/PollAnswer.js
```diff
@@ -64,6 +75,15 @@ class PollAnswer extends Base {
     return this.client.emojis.resolve(this._emoji.id) ?? new Emoji(this.client, this._emoji);
   }
 
+  /**
+   * Whether this poll answer is a partial.
+   * @type {boolean}
+   * @readonly
+   */
+  get partial() {
+    return this.poll.partial || (this.text === null && this.emoji === null);
```
vladfrangu, line 84: Isn't the check for emoji === null redundant? Answers can have no emoji, but we know for a fact that its a partial one if we have no cached text

packages/discord.js/typings/index.d.ts
```diff
@@ -6542,6 +6558,23 @@ export interface PartialMessage
...
 
+export interface PartialPoll
+  extends Partialize<
+    Poll,
+    'allowMultiselect' | 'layoutType' | 'expiresTimestamp',
+    null,
+    'question' | 'message' | 'answers'
+  > {
+  question: { text: null };
+  message: PartialMessage;
```
vladfrangu, line 6569: Can't message also be.. a full message?

packages/discord.js/typings/index.d.ts
```diff
@@ -6542,6 +6558,23 @@ export interface PartialMessage
...
+  // eslint-disable-next-line no-restricted-syntax
```
vladfrangu, line 6570: Not a blocker but a question, what does this comment disable, I see no restricted syntax 😅 👀

vladfrangu (dismissed):

packages/discord.js/typings/index.d.ts
almeidx: Because of the rule added in this pull request, but the rule shouldn't apply to this interface as it's not user input
- #10045

packages/discord.js/typings/index.d.ts
almeidx: If you had a full message, you'd also have a full poll, as the poll is part of the message (goes both ways)

packages/discord.js/src/structures/PollAnswer.js
almeidx: Presumably, poll answers might only contain an emoji in the future(?). Just future-proofing I guess
> `text` should always be non-null for both questions and answers, but please do not depend on that in the future.

https://discord.com/developers/docs/resources/poll#poll-media-object-poll-media-object-structure

packages/discord.js/src/structures/Poll.js
```diff
@@ -82,15 +62,89 @@ class Poll extends Base {
     } else {
       this.resultsFinalized ??= false;
     }
+
+    /**
+     * Whether this poll allows multiple answers
+     * @type {boolean}
```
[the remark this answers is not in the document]
almeidx: Only applies to partial polls. Full polls always have this field

packages/discord.js/src/structures/Poll.js
```diff
@@ -82,15 +62,89 @@ class Poll extends Base {
...
+     */
+    this.allowMultiselect ??= data.allow_multiselect ?? null;
+
+    /**
+     * The layout type of this poll
+     * @type {PollLayoutType}
```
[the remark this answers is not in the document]
almeidx: Same as above

vladfrangu:
This needs a rebase

packages/discord.js/src/structures/PollAnswer.js
```diff
@@ -75,14 +95,10 @@ class PollAnswer extends Base {
    * Fetches the users that voted for this answer.
    * @param {BaseFetchPollAnswerVotersOptions} [options={}] The options for fetching voters
    * @returns {Promise<Collection<Snowflake, User>>}
+   * @deprecated Use {@link PollAnswerVoterManager#fetch} instead
```
vladfrangu, line 98: Since this targets main, you can do a breaking change and drop the method, and it can be backported with a deprecation warning 👀

vladfrangu (approved):

packages/discord.js/src/structures/PollAnswer.js
almeidx: This can be done in a follow-up PR. So this PR can be backported as-is

Qjuh (approved):

jvrdelafuente:
Quick question, I opened an [issue](https://github.com/discordjs/discord.js/issues/11017) and I was pointed to here. These changes are not added into 14 version. Is there any plan to include it? Thanks.

Jiralite:
We welcome for pull requests that backport this (hence https://github.com/discordjs/discord.js/labels/backport-candidate). It is simply waiting for someone to do it!

PR #10352, opened 2024-06-17
fix: package gen script

didinele:
Unfortunately, some files can't be `.hbs` due to this bug https://github.com/plopjs/plop/issues/370

That is fine for now.


suneettipirneni (approved):

iCrawl (approved):

packages/scripts/turbo/generators/config.ts
```diff
@@ -41,14 +41,14 @@ export default function generator(plop: PlopTypes.NodePlopAPI): void {
 			{
 				type: 'addMany',
 				destination: `${plop.getDestBasePath()}/../{{name}}`,
-				templateFiles: ['templates/**'],
+				templateFiles: ['templates/**', '!templates/default/cliff.toml'],
```
ckohen, line 44: Can you add a comment here to make sure no one removes this in the future accidentally. I presume it's because the toml replacer syntax cliff uses causes problems with the default plop template replacer....

ckohen (approved):

PR #10360, opened 2024-06-21
types(Client): `EventEmitter` static method overrides

### LINKED ISSUE #10358: Conflict between discord.js + @nodes/types after upgrading

discord.js

### Issue description

Hi there, approximately two hours ago @nodes/types upgraded their package. I use GitHub Actions to build and test my discord bot, and I didn't realize I was out-of-date until I got conflicting errors that were very confusing.

As far as I can tell, I believe that @ nodes/types did some important changes and now discord.js is conflicting with those changes, and this is an issue with discord.js but I'm not sure. I'm not overly a huge expert on these types of things.

You can replicate this issue by installing the latest @nodes/types and the latest development build of discord.js. Upon attempting to build your project, you will be greeted with this error:

```
Error: node_modules/@types/node/events.d.ts(519,30): error TS2300: Duplicate identifier 'EventEmitter'.
Error: node_modules/@types/node/stream.d.ts(53,11): error TS2420: Class 'ReadableBase' incorrectly implements interface 'ReadableStream'.
  Type 'ReadableBase' is missing the following properties from type 'ReadableStream': off, removeAllListeners, setMaxListeners, getMaxListeners, and 4 more.
Error: node_modules/@types/node/stream.d.ts(662,11): error TS2420: Class 'WritableBase' incorrectly implements interface 'WritableStream'.
  Type 'WritableBase' is missing the following properties from type 'WritableStream': off, removeAllListeners, setMaxListeners, getMaxListeners, and 4 more.
Error: node_modules/@types/node/test.d.ts(380,11): error TS2420: Class 'TestsStream' incorrectly implements interface 'ReadableStream'.
  Type 'TestsStream' is missing the following properties from type 'ReadableStream': off, removeAllListeners, setMaxListeners, getMaxListeners, and 4 more.
Error: node_modules/discord.js/typings/index.d.ts(246,[9](https://github.com/Polytoria/CommunityBot/actions/runs/9605412040/job/26492987136#step:6:10)): error TS2300: Duplicate identifier 'EventEmitter'.
Error: Process completed with exit code 2.
```

### Code sample

```typescript
// <reference path="index.d.ts"/>
import { Client, GatewayIntentBits, ActivityType, Events, Collection, BaseInteraction } from 'discord.js'
import dotenv from 'dotenv'
import { success, warning } from './utils/log.js'
import commandsData from './commandsData.js'
import { IConfiguration } from '../types'

// Initialize .env file.
dotenv.config()

const configuration: IConfiguration = {
  token: process.env.TOKEN,
  clientId: process.env.CLIENTID,
  coolDown: 3
}

const client = new Client({
  intents: [
    GatewayIntentBits.GuildMessages,
    GatewayIntentBits.Guilds,
    GatewayIntentBits.GuildMembers,
    GatewayIntentBits.MessageContent
  ]
})

// @ts-expect-error
client.commands = new Collection()

commandsData.forEach((commandData, index) => {
  // @ts-expect-error
  client.commands.set(commandData.data.name, commandData)
})

client.on('ready', () => {
  // @ts-expect-error
  client.user.setActivity(`${client.guilds.cache.reduce((a, guild) => a + guild.memberCount, 0)} Users`, { type: ActivityType.Watching })

  setInterval(function () {
    // @ts-expect-error
    client.user.setActivity(`${client.guilds.cache.reduce((a, guild) => a + guild.memberCount, 0)} Users`, { type: ActivityType.Watching })
  }, 60000)
})
success({ context: '[Bot]', message: 'Bot succesfully started.' })

client.on('messageCreate', async (message) => {
  if (message.author.bot) return
  if (!message.content.startsWith('p!')) return
  if (!message.inGuild) return

  await message.reply('The Polytoria Community Bot has switched to slash commands!')
})

client.on(Events.InteractionCreate, async (interaction:BaseInteraction) => {
  if (!interaction.isCommand()) {
    return
  }

  // @ts-expect-error
  const command:any = interaction.client.commands.get(interaction.commandName)

  if (!command) {
    interaction.reply("Command doesn't exist")
    return
  }

  success({
    context: '[Client]',
    message: 'Command registered.'
  })

  success({
    context: '[Bot]',
    message: 'Running command ' + command.data.name
  })

  try {
    if (command.constructor.name === 'AsyncFunction') {
      await command.execute(interaction)
    } else {
      command.execute(interaction)
    }
  } catch (error: any) {
    if (interaction.replied) {
      await interaction.followUp('Failed to execute command: ' + error)
    } else {
      await interaction.reply('Failed to execute command: ' + error)
    }
    warning({
      context: '[Bot]',
      message: error.toString()
    })
  }
})

// Handle Promise Rejection
process.on('unhandledRejection', (reason, p) => {
  console.error(reason, 'Unhandled Rejection at Promise', p)
})

process.on('uncaughtException', err => {
  console.error(err)
  process.exit(1)
})

success({ context: '[Bot]', message: 'Bot succesfully logged in.' })

client.login(configuration.token)
```

### Versions

typescript - 5.5.2
discord-js - latest development build
@types/node - 20.14.7

Node - Latest v20

### Issue priority

No Partials

Guilds, GuildMembers, GuildMessages, MessageContent

_No response_

Renegade334:

Fixes a build failure in TypeScript 5.5.
Resolves #10358.

See discussion at the associated issue.

Currently, the static `on()` and `once()` methods on Node's `EventEmitter` are overloaded in the discord.js typings to allow for type-checked event handling.

TypeScript 5.5.1-rc fixed the bug that allowed class merging here, and this approach is no longer possible. The `EventEmitter` class declaration in typings.d.ts causes breaking errors for `node:events` module imports in TS >=5.5.1, and needs to be removed.

It would still be useful for these methods to have access to this type-safe behaviour for client events, but we can't add static method overloads to `EventEmitter` itself.

Instead, we can override those methods further down the inheritance chain, and provide `Client.on()` and `Client.once()` as the type-safe methods for those who wish to use them. In other words:

```ts
// old
EventEmitter.on(client, 'messageCreate'); // AsyncIterableIterator<[message: Message<boolean>]>
Client.on(client, 'messageCreate'); // AsyncIterableIterator<[message: Message<boolean>]>
// new
EventEmitter.on(client, 'messageCreate'); // AsyncIterableIterator<any[]>
Client.on(client, 'messageCreate'); // AsyncIterableIterator<[message: Message<boolean>]>

// old
EventEmitter.once(client, 'interactionCreate'); // Promise<[interaction: Interaction<CacheType>]>
Client.once(client, 'interactionCreate'); // Promise<[interaction: Interaction<CacheType>]>
// new
EventEmitter.once(client, 'interactionCreate'); // Promise<any[]>
Client.once(client, 'interactionCreate'); // Promise<[interaction: Interaction<CacheType>]>
```

This won't break any existing builds that use `EventEmitter.on()` or `EventEmitter.once()`, as any resolved event argument type will now just become `any`. However, in order to revert back to the type-safe behaviour, they'd need to use the static methods on `Client` rather than on `EventEmitter`.

This patch also adds the optional `options` parameter (Node >=14) to the method overrides.

### DIFF
--- a/packages/discord.js/typings/index.d.ts
+++ b/packages/discord.js/typings/index.d.ts
@@ -243,20 +243,6 @@ import {
   RawWidgetMemberData,
 } from './rawDataTypes.js';
 
-declare module 'node:events' {
-  class EventEmitter {
-    // Add type overloads for client events.
-    public static once<Emitter extends EventEmitter, Event extends keyof ClientEvents>(
-      eventEmitter: Emitter,
-      eventName: Emitter extends Client ? Event : string,
-    ): Promise<Emitter extends Client ? ClientEvents[Event] : any[]>;
-    public static on<Emitter extends EventEmitter, Events extends keyof ClientEvents>(
-      eventEmitter: Emitter,
-      eventName: Emitter extends Client ? Events : string,
-    ): AsyncIterableIterator<Emitter extends Client ? ClientEvents[Events] : any>;
-  }
-}
-
 //#region Classes
 
 export class Activity {
@@ -996,6 +982,18 @@ export class Client<Ready extends boolean = boolean> extends BaseClient {
   private get _censoredToken(): string | null;
   // This a technique used to brand the ready state. Or else we'll get `never` errors on typeguard checks.
   private readonly _ready: Ready;
+
+  // Override inherited static EventEmitter methods, with added type checks for Client events.
+  public static once<Emitter extends EventEmitter, Event extends keyof ClientEvents>(
+    eventEmitter: Emitter,
+    eventName: Emitter extends Client ? Event : string | symbol,
+    options?: { signal?: AbortSignal | undefined },
+  ): Promise<Emitter extends Client ? ClientEvents[Event] : any[]>;
+  public static on<Emitter extends EventEmitter, Event extends keyof ClientEvents>(
+    eventEmitter: Emitter,
+    eventName: Emitter extends Client ? Event : string | symbol,
+    options?: { signal?: AbortSignal | undefined },
+  ): AsyncIterableIterator<Emitter extends Client ? ClientEvents[Event] : any[]>;
 
   public application: If<Ready, ClientApplication>;
   public channels: ChannelManager;
--- a/packages/discord.js/typings/index.test-d.ts
+++ b/packages/discord.js/typings/index.test-d.ts
@@ -1294,6 +1294,10 @@ client.on('guildCreate', async g => {
   );
 });
 
+// EventEmitter static method overrides
+expectType<Promise<[Client<true>]>>(Client.once(client, 'ready'));
+expectType<AsyncIterableIterator<[Client<true>]>>(Client.on(client, 'ready'));
+
 client.login('absolutely-valid-token');
 
 declare const loggedInClient: Client<true>;


vladfrangu:
This is a breaking change, as before the point was to also support EventEmitter.on/once returning strictly typed results. Can't we use namespace merging maybe to add overloads to those functions?

Renegade334:
> This is a breaking change, as before the point was to also support EventEmitter.on/once returning strictly typed results. Can't we use namespace merging maybe to add overloads to those functions?

Namespace function declarations and class static method declarations aren't compatible with each other, so this isn't an option.

Merging regular class methods/properties is perfectly possible (via `interface ClassName {}`), but there's no equivalent for static methods/properties. The question has been asked before (microsoft/TypeScript#39870) but there's just no valid way to do it in TS as things stand. We got away with it previously due to a TypeScript bug, but now that it's been patched, we have no choice but to remove the `EventEmitter` overloads.

This patch isn't truly "breaking", in the sense that any usage of `EventEmitter.on()` or `EventEmitter.once()` will still work and compile just fine, just without type narrowing. This is a change to existing behaviour, but it's completely unavoidable given the above limitations.

Renegade334:
More reports of breaking are coming in as people are attempting to upgrade to TS 5.5, so I'm going to mark this as ready for review.

As a minimum changeset to unbreak TS 5.5 builds, the `EventEmitter` definitions need to be removed – as discussed, there simply is no way of getting around this.

As for how best to provide the type-safe DJS versions of on/once, I think that adding overloads for the inherited static methods `Client.on()` and `Client.once()` is the cleanest solution, unless anyone can come up with a better option!

StarManTheGamer (approved):
I think this is a good solution. Sucks that it has to be removed but I don’t see any other ways around it

ckohen:
The more appropriate issue seems to be https://github.com/microsoft/TypeScript/issues/2957 which is still open and might see some more attention now that the class merging "bug" has been "fixed".

The other thing worthy of note is https://github.com/DefinitelyTyped/DefinitelyTyped/pull/68313 which added strict typing to `EventEmitter` natively, but neglected to do the static members. I've asked if this was intentional, and if not I assume we can fasttrack getting that in, which solves this.

I'm suggesting we block this until we get an answer on the static members getting generics. The only concern I have is the potential incompatibility between our `ClientEvents` interface and what EE takes as a generic, but that is likely fixable via a mapped type if it is an issue (to maintain `ClientEvents` to be non-breaking)

Renegade334:
@ckohen: Generics would be ideal, but the question is whether or not it's a good idea to leave the repo in an unbuildable state until such a time as that feature gets added to DT and then makes it into a @types/node package build.

The TS syntax proposal is nearly a decade old, so while re-opening the discussion is a good idea, I wouldn't be holding out for it.

almeidx (approved):

voltflake:
Why is this still not reviewed? Even default vscode typescript version now screams at discordjs.

kyranet (approved):
The new code is fine, however, we want to wait for https://github.com/DefinitelyTyped/DefinitelyTyped/pull/69997 first so it's correctly updated.

jamezrin:
I don't think the DT team is looking to merge https://github.com/DefinitelyTyped/DefinitelyTyped/pull/69997, at least for now. Could this PR get merged since it fixes usage with TS 5.5+?

ckohen:
DT will not merge it as is but refuses to respond to say how it can be merged, so it's on them. They broke it in the first place and are unwilling to fix it.

Regardless, no matter how we do this it's a breaking change, so we're heavily considering fix-breaking this in a more sustainable way.

ckohen:
This is the least breaking change (even in the case where EE has full generic support) so we're going to move forward with this for v14 at least.

ckohen (approved):

vladfrangu (approved):

PR #10381, opened 2024-07-09
feat: Add support for Automated Message nonce handling

### LINKED ISSUE #10323: Automated Message nonce handling

discord.js

### Feature

Discord recently [added support for enforcing the nonce parameter](https://discord.com/developers/docs/change-log#enforced-nonces-on-create-message-endpoint) when creating a Message, to be able to prevent duplicate messages in a short amount of time.

When using this, currently developers have to specify both a random `nonce` and `enforceNonce: true` either in every `<BaseChannel>.send()` function in the payload, or extend the class(es) themselves to handle it everywhere.

A client option, disabled by default, to automatically generate a nonce (e.g. using `SnowflakeUtil.generate().toString()`) and set both `nonce` and (while it's still required) `enforceNonce` when creating a Message, unless one or both of those parameters are already present in the payload.

A Message payload parameter like `sendNonce` which does the same as above, but it's per Message instead of globally.

### Other context

See #10301 for why this could be useful.

JMTK:
Took a stab at doing this, let me know if there's any issues! Fixes #10323 

Only thing I thought about was to instead make this a function you can pass in to generate nonces, so something like

```js
if (typeof this.client.options.generateNonce === 'function') {
    options.nonce = this.client.options.generateNonce();
}
```

But the users may run into issues, such as the 25 character limit, generating high collision nonces, etc

Minor


packages/discord.js/src/structures/interfaces/TextBasedChannel.js
```diff
@@ -165,6 +165,11 @@ class TextBasedChannel {
       return dm.send(options);
     }
 
+    if (!options.nonce && this.client.options.automatedMessageNonce) {
```
Jiralite, line 168: A nonce can be `0`, so this may incorrectly pass the check.

packages/discord.js/typings/index.d.ts
```diff
@@ -5311,6 +5311,7 @@ export interface ClientOptions {
   ws?: WebSocketOptions;
   rest?: Partial<RESTOptions>;
   jsonTransformer?: (obj: unknown) => unknown;
+  automatedMessageNonce?: boolean;
```
Jiralite, line 5314: This is missing in the documentation.

Also, I think it would be best to name this `enforceNonce`, much like how `failIfNotExists` matches the actual property name.

Jiralite (changes requested):

packages/discord.js/typings/index.d.ts
JMTK: Updated

packages/discord.js/src/structures/interfaces/TextBasedChannel.js
JMTK: Updated

packages/discord.js/src/structures/interfaces/TextBasedChannel.js
```diff
@@ -165,6 +165,11 @@ class TextBasedChannel {
       return dm.send(options);
     }
 
+    if (typeof options.nonce === 'undefined' && this.client.options.enforceNonce) {
```
jaw0r3k, line 168: There is no `undefined` type, you should check if the typeof is object and then if nonce isnt undefined

Jiralite, line 168: See #9191:

```suggestion
    if (options.nonce === undefined && this.client.options.enforceNonce) {
```

Jiralite (dismissed):
Should we handle when `enforceNonce: false` is explicitly passed?

```JavaScript
const createdMessage = await message.channel.send({
	content: "hello",
	enforceNonce: false,
});

// `createdMessage` has a nonce which is enforced despite `enforceNonce` being `false`.
```

packages/discord.js/src/structures/interfaces/TextBasedChannel.js
```diff
@@ -165,6 +165,14 @@ class TextBasedChannel {
       return dm.send(options);
     }
 
+    /**
+     * If `nonce` is not provided generate one & set `enforceNonce`, unless `enforceNonce` is explicitly set to `false`
+     */
+    if (options.nonce === undefined && options.enforceNonce !== false && this.client.options.enforceNonce) {
```
almeidx, line 171: This should be in `MessagePayload#resolveBody`

almeidx (dismissed):

packages/discord.js/src/structures/interfaces/TextBasedChannel.js
JMTK: Good catch! I had to make a couple of small logic changes to account for some of the existing code

packages/discord.js/src/structures/MessagePayload.js
```diff
@@ -134,8 +135,18 @@ class MessagePayload {
...
+     * If `nonce` isn't provided, generate one & set `enforceNonce`
+     * Unless `enforceNonce` is explicitly set to `false`(not just falsy)
+     */
+    if (nonce === undefined) {
+      if (this.options.enforceNonce !== false && this.client.options.enforceNonce) {
+        this.options.nonce = DiscordSnowflake.generate().toString();
+        this.options.enforceNonce = true;
+      } else if (enforce_nonce) {
+        throw new DiscordjsError(ErrorCodes.MessageNonceRequired);
+      }
```
almeidx, line 149: IMO we should just get rid of this[^1] and automatically generate a nonce if none is passed when `enforceNonce` is set to `true`.
Probably results in better DX and simplifies the logic a bit 

[^1]: meaning we stop throwing the error and deprecate the error code

packages/discord.js/src/structures/MessagePayload.js
```diff
@@ -134,8 +135,18 @@ class MessagePayload {
     }
 
     const enforce_nonce = Boolean(this.options.enforceNonce);
-    if (enforce_nonce && nonce === undefined) {
-      throw new DiscordjsError(ErrorCodes.MessageNonceRequired);
+
+    /**
+     * If `nonce` isn't provided, generate one & set `enforceNonce`
+     * Unless `enforceNonce` is explicitly set to `false`(not just falsy)
+     */
```
almeidx, line 142: I don't think this should be a JSDoc 
```suggestion
    // If `nonce` isn't provided, generate one & set `enforceNonce`
    // Unless `enforceNonce` is explicitly set to `false`(not just falsy)
```

packages/discord.js/src/structures/MessagePayload.js
JMTK: While I don't disagree with you, I think that would end up changing the semver for this particular PR. What do you think?

packages/discord.js/src/structures/MessagePayload.js
```diff
@@ -134,8 +135,16 @@ class MessagePayload {
...
     const enforce_nonce = Boolean(this.options.enforceNonce);
-    if (enforce_nonce && nonce === undefined) {
-      throw new DiscordjsError(ErrorCodes.MessageNonceRequired);
+
+    // If `nonce` isn't provided, generate one & set `enforceNonce`
+    // Unless `enforceNonce` is explicitly set to `false`(not just falsy)
+    if (nonce === undefined) {
+      if (this.options.enforceNonce !== false && this.client.options.enforceNonce) {
+        this.options.nonce = DiscordSnowflake.generate().toString();
+        this.options.enforceNonce = true;
```
vladfrangu, line 144: Changing just `this.options` is useless since later at line 223-224 the variables above are re-used. You'll wanna alter the variables (`nonce =` from line 129 instead of `this.options.nonce` and `enforce_nonce` above

packages/discord.js/src/structures/MessagePayload.js
vladfrangu: Lets not change the current behavior, although technically not breaking in this case.

vladfrangu (dismissed):

RedGuy12:
@JMTK you force pushed, which (i assume unintentionally) closed the PR
you may find it useful to learn how to use branches

JMTK:
oh woops, i thought i had it on a branch. i was working on creating a new PR and accidentally did that

kyranet (approved):

didinele (approved):

vladfrangu (approved):

PR #10384, opened 2024-07-12
feat(GuildMember): Banners

Jiralite:

- https://github.com/discord/discord-api-docs/pull/6986


packages/discord.js/src/structures/GuildMember.js
```diff
@@ -492,6 +523,7 @@ class GuildMember extends Base {
     });
     json.avatarURL = this.avatarURL();
     json.displayAvatarURL = this.displayAvatarURL();
+    json.displayBannerURL = this.displayBannerURL();
```
almeidx, line 526: Any reason not to add `bannerURL` together with `displayBannerURL`, just like with the avatar?

packages/discord.js/src/structures/GuildMember.js
```diff
@@ -84,6 +84,17 @@ class GuildMember extends Base {
...
     }
+
+    if ('banner' in data) {
+      /**
+       * The guild member's banner hash.
+       * @type {?string}
+       */
+      this.banner = data.banner;
+    } else {
+      this.banner ??= null;
```
benjGam, line 95: Hi, is this a mistype or this operator actually works ? 

If it's not a mistype, could you explain me that operator (??=, i know that ?? is null coalescence but i'd never saw ??= before)

Thanks !

benjGam (approved):

packages/discord.js/src/structures/GuildMember.js
almostSouji: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_assignment
please use your favourite search engine for questions, instead of code reviews

sdanialraza: It's not a mistype, you can learn about it [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_assignment)

didinele (approved):

almeidx (approved):

vladfrangu (approved):

PR #10394, opened 2024-07-18
fix(build): update to support strictBuiltinIteratorReturn

rbuckton:

This updates the build to be forwards compatible with https://github.com/microsoft/TypeScript/pull/58243, which uses a stricter return type for built-in iterators under the new `strictBuiltinIteratorReturn` option, which is enabled by default when building with `strict`.

### DIFF
--- a/packages/collection/src/collection.ts
+++ b/packages/collection/src/collection.ts
@@ -87,7 +87,7 @@ export class Collection<Key, Value> extends Map<Key, Value> {
 		if (amount < 0) return this.last(amount * -1);
 		amount = Math.min(this.size, amount);
 		const iter = this.values();
-		return Array.from({ length: amount }, (): Value => iter.next().value);
+		return Array.from({ length: amount }, (): Value => iter.next().value!);
 	}
 
 	/**
@@ -104,7 +104,7 @@ export class Collection<Key, Value> extends Map<Key, Value> {
 		if (amount < 0) return this.lastKey(amount * -1);
 		amount = Math.min(this.size, amount);
 		const iter = this.keys();
-		return Array.from({ length: amount }, (): Key => iter.next().value);
+		return Array.from({ length: amount }, (): Key => iter.next().value!);
 	}
 
 	/**
@@ -512,7 +512,7 @@ export class Collection<Key, Value> extends Map<Key, Value> {
 		if (thisArg !== undefined) fn = fn.bind(thisArg);
 		const iter = this.entries();
 		return Array.from({ length: this.size }, (): NewValue => {
-			const [key, value] = iter.next().value;
+			const [key, value] = iter.next().value!;
 			return fn(value, key, this);
 		});
 	}
@@ -626,7 +626,7 @@ export class Collection<Key, Value> extends Map<Key, Value> {
 		const iterator = this.entries();
 		if (initialValue === undefined) {
 			if (this.size === 0) throw new TypeError('Reduce of empty collection with no initial value');
-			accumulator = iterator.next().value[1];
+			accumulator = iterator.next().value![1];
 		} else {
 			accumulator = initialValue;
 		}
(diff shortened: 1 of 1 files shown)


rbuckton:
I've filed this as a Draft PR in the event there are any other changes to https://github.com/microsoft/TypeScript/pull/58243 that must be addressed. Once that PR has merged, I will mark this PR as "Ready for review".

Renegade334:
Duplicate of #10260. I don't mind which is kept.

ckohen (approved):

packages/collection/src/collection.ts
```diff
@@ -626,7 +626,7 @@ export class Collection<Key, Value> extends Map<Key, Value> {
 		const iterator = this.entries();
 		if (initialValue === undefined) {
 			if (this.size === 0) throw new TypeError('Reduce of empty collection with no initial value');
-			accumulator = iterator.next().value[1];
+			accumulator = iterator.next().value![1];
```
Renegade334, line 629:
```suggestion
			accumulator = iterator.next().value![1] as unknown as InitialValue;
```

As per #10260. This is forced to overcome an assignability error once the `value` type is no longer `any`.

Renegade334 (changes requested):

almeidx (approved):

packages/collection/src/collection.ts
didinele: I did look at #10260 and I don't really follow why `!`isn't okay here. Could you elaborate?

Renegade334: Apologies.

This becomes directly equivalent to [the corresponding statement](https://github.com/discordjs/discord.js/blob/2fcf1910ee30935d27624a00fa2330b4983d4b43/packages/collection/src/collection.ts#L659) in `reduceRight()`, where this coercion is already used. Essentially, the value of `iterator.next().value![1]` becomes `Value`, not `InitialValue`, and TypeScript can't infer that these types should be the same at this point. ([TSP](https://www.typescriptlang.org/play/?ts=5.6.0-dev.20240724#code/MYGwhgzhAEDCD2IQFNgBcCW8B2AeA0sgJ4A00AamCAK7IB80yAHmstgCYwCyYADgcTKUa9aAG8AUAEhe1AEYgMwaACdk7asGS4AktgyYqw2tAC8FKrToAKaVIBm2AFzRrYYMGoBbauDTwVFz0DDCNLZDIAN3CXYwjoAGtiF0JSaGBEFHQsZ2g0AAsMCABKMwZgwxA4kjsMfUq4gH4g+tCq8JqpYpaQsJFxWvtXNCJeZHghx2gAQlNzAHJ7amxsnHnSgpV4AHdobGRdgBVR5ABRFS2VawADABIxRwBfaCK9+DRoMGgllcwca+KAG47CgPu5PD4-AFpj0GuFgXYMtgIB8DMgVGB-CozHlChAAHRsNAqDDICDWIGDVx1XrtfpzczLdjIex1dSlSRSLkYIbWApFfEQDAAL2QZjm0AADBt8ltdvsjidzpdrPMAErqTRiiaMLy8EbpTKoP7YaDbAz5N4vVpUaDRETrYFcqTg7y+TEBHFojFY-H7FgU-H22jTADaAEYALpOqTPZAgCBizlc12Qj3Y8w0uEiGOPCR2eye6xIlHQUNJNLB5CR6A673pjl2F0eN1QjPfbBuFtprFRDqJQS4oqUrl5uxqNDUFSm1PurFOvN5oA))

ckohen: Really wish we could constrain InitialValue based on the passed parameters (as TS correctly does for the callback). I don't love this solution, but I can't think of a "more correct" solution off the top of my head. Might have something to do with the type of the accumulator or something like that.

didinele (changes requested):
I'd like to see this addressed before merging. https://github.com/discordjs/discord.js/pull/10394#discussion_r1690308415

didinele (approved):

vladfrangu (approved):

kyranet (approved):

suneettipirneni (approved):

SpaceEEC (approved):

PR #10411, opened 2024-07-26
feat(ws): Allow passing in CompressionMethod to ClientOptions to use native node zlib

JMTK:
#10316 added support for native zlib support to @discordjs/ws, but there's not currently a way to pass it in to a client(see https://github.com/discordjs/discord.js/blob/bf6761a44adec1fe5017f6bf5d8bc0734916961f/packages/discord.js/src/client/websocket/WebSocketManager.js#L156). 

This PR now first checks if you passed in a CompressionMethod, and if not checks if you have `zlib-sync` installed, if not then no compression

Minor

### DIFF
--- a/packages/builders/__tests__/components/button.test.ts
+++ b/packages/builders/__tests__/components/button.test.ts
@@ -50,6 +50,11 @@ describe('Button Components', () => {
 				button.toJSON();
 			}).not.toThrowError();
 
+			expect(() => {
+				const button = buttonComponent().setSKUId('123456789012345678').setStyle(ButtonStyle.Premium);
+				button.toJSON();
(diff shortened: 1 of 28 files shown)


Qjuh:
While that PR was merged it was done in a semver major way and `@discordjs/ws@2.0.0` isn‘t released. So this PR needs to
- be blocked until that happens
- address any other changes that major update might bring, including the event change already merged in it
- best case that being done in a non-breaking way because otherwise if it would be a semver major the /ws incorporation would be redone entirely anyway

almeidx:
Achieving this in a non-breaking way doesn't seem feasible, meaning this pull request is superseded by #10420

PR #10412, opened 2024-07-28
fix(collection)!: default sort comparison algorithm

Renegade334:

The documentation for `Collection#sort` and `Collection#toSorted` implies that if no comparison function is passed to the method, the default comparison behaviour mirrors that of `Array.prototype.sort` &ndash; in other words, stringwise comparison by UTF-16 code points.

As things stand, that's not the case. There are a couple of significant issues with the current implementation of `defaultSort`:
- The function doesn't implement stringwise comparison but instead just uses the `>` operator, which preferentially performs numeric comparison of its operands.
  - For example, `'1.0'` is sorted after `1` with stringwise comparison, but `'1.0' > 1` is false.
- The function returns -1 (_ie._ A before B) if A is neither greater than B nor strict-equal to B. However, this isn't logically correct in a sort comparison context, and leads to violations of the required properties of a reliable comparator.
  - For example:
    ```js
    const a = { a: 1 };
    const b = { b: 2 };
    defaultSort(a, b); // -1
    defaultSort(b, a); // also -1
    ```

I'm assuming that the intention is for the default behaviour to mirror that of `Array.prototype.sort`, and so this PR rewrites the function to follow the [ECMA-262 specification](https://tc39.es/ecma262/multipage/indexed-collections.html#sec-comparearrayelements) for the default comparison algorithm. The docblocks are also updated to reflect this more accurately.

### DIFF
--- a/packages/collection/__tests__/collection.test.ts
+++ b/packages/collection/__tests__/collection.test.ts
@@ -829,6 +829,33 @@ describe('sort() tests', () => {
 			const coll = createCollectionFrom(['a', 5], ['b', 3], ['c', 1]);
 			expect(coll.sort()).toStrictEqual(createCollectionFrom(['c', 1], ['b', 3], ['a', 5]));
 		});
+
+		describe('returns correct sort order for test values', () => {
+			const defaultSort = Collection['defaultSort']; // eslint-disable-line @typescript-eslint/dot-notation
+			const testDefaultSortOrder = (firstValue: any, secondValue: any, result: number) => {
+				expect(defaultSort(firstValue, secondValue)).toStrictEqual(result);
+				expect(defaultSort(secondValue, firstValue)).toStrictEqual(result ? result * -1 : 0);
+			};
+
+			test('correctly evaluates sort order of undefined', () => {
+				testDefaultSortOrder(undefined, undefined, 0);
+				testDefaultSortOrder(0, undefined, -1);
+			});
+
+			test('correctly evaluates numeric values stringwise', () => {
+				testDefaultSortOrder(-1, -2, -1); // "-1" before "-2"
+				testDefaultSortOrder(1, '1', 0); // "1" equal to "1"
+				testDefaultSortOrder(1, '1.0', -1); // "1" before "1.0"
+				testDefaultSortOrder(1.1, '1.1', 0); // "1.1" equal to "1.1"
+				testDefaultSortOrder('01', 1, -1); // "01" before "1"
+				testDefaultSortOrder(1, 1n, 0); // "1" equal to "1"
+				testDefaultSortOrder(Number.NaN, 'NaN', 0); // "NaN" equal to "NaN"
+			});
+
+			test('evaluates object literals as equal', () => {
+				testDefaultSortOrder({ a: 1 }, { b: 2 }, 0);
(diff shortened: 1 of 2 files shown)


packages/collection/src/collection.ts
```diff
@@ -988,8 +989,20 @@ export class Collection<Key, Value> extends Map<Key, Value> {
...
+	 * {@link https://tc39.es/ecma262/multipage/indexed-collections.html#sec-comparearrayelements | CompareArrayElements}
+	 * operation with arguments `firstValue`, `secondValue` and `undefined`.
+	 */
 	private static defaultSort<Value>(firstValue: Value, secondValue: Value): number {
-		return Number(firstValue > secondValue) || Number(firstValue === secondValue) - 1;
+		if (firstValue === undefined) return secondValue === undefined ? 0 : 1;
+		if (secondValue === undefined) return -1;
+
+		const x = String(firstValue);
+		const y = String(secondValue);
```
Renegade334, line 1002: The only difference between the ECMA-262 `ToString` abstract operation and calling the `String()` constructor here, is that the `ToString` operation throws if its argument is a symbol.

I didn't think it was worth adding checks here – if a user is creating a collection of symbols and then trying to sort it, then they probably have bigger issues...

packages/collection/__tests__/collection.test.ts
```diff
@@ -770,6 +770,33 @@ describe('sort() tests', () => {
 			const coll = createCollectionFrom(['a', 5], ['b', 3], ['c', 1]);
 			expect(coll.sort()).toStrictEqual(createCollectionFrom(['c', 1], ['b', 3], ['a', 5]));
 		});
+
+		describe('returns correct sort order for test values', () => {
+			const defaultSort = Reflect.get(Collection, 'defaultSort') as (typeof Collection)['defaultSort'];
```
kyranet, line 775: You should be able to do this
```suggestion
			const defaultSort = Collection['defaultSort'];
```

kyranet (changes requested):

packages/collection/src/collection.ts
kyranet: I think that's a reasonable compromise, we don't have to follow the specification to a T, and if we added every single thing, the library would probably become a bit big.

kyranet:
I'm labelling this as semver-major because it changes the behaviour of `Collection#sort` without arguments, which may have been desirable to users (and not longer is anymore).

packages/collection/__tests__/collection.test.ts
Renegade334: Was concerned that using `[]` to bypassing access checks for private properties was possibly unintentional and could be removed in the future. However, it looks like it was confirmed to be intentional by the TS team previously, so should be OK. 👍

vladfrangu (approved):

didinele (approved):

kyranet (approved):

Jiralite:
Please resolve conflicts!

Also, would you mind writing a BREAKING CHANGE entry in a commit or in the pull request's body somewhere (https://github.com/discordjs/discord.js/blob/main/.github/COMMIT_CONVENTION.md)? Thank you!

PR #10419, opened 2024-07-31
chore: deprecate client options presence

didinele:


kyranet (approved):

kyranet:
One little thing, you should mark this field as deprecated in the typings:
https://github.com/discordjs/discord.js/blob/5eabec14d45ef7bdd7f610e84234eb63e726eacd/packages/discord.js/typings/index.d.ts#L5307

Jiralite (approved):

Jiralite:
Maybe deprecate the specific option in the typings too?

vladfrangu (dismissed):
Requesting changes to the PR so the typing change is applied

packages/discord.js/typings/index.d.ts
```diff
@@ -5304,6 +5304,7 @@ export interface ClientOptions {
   allowedMentions?: MessageMentionOptions;
   partials?: readonly Partials[];
   failIfNotExists?: boolean;
+  /** @deprecated In the next release, use {@link ClientOptions.ws.initialPresence} instead */
```
kyranet, line 5307: By the time users see this, it'll be already that release:
```suggestion
  /** @deprecated Use {@link ClientOptions.ws.initialPresence} instead */
```

kyranet (changes requested):

packages/discord.js/typings/index.d.ts
didinele: > By the time users see this, it'll be already that release:

No...? Vlad said we're deprecating in 14 and removing completely in 15.

didinele: In retrospect though, wsOptions.presence is literally not a thing right now. It's probably best if we started supporting it in 14 right now.

Jiralite: This is landing in the next version 14 release. The message they'll see is "In the next release use...", but it should be only "Use".

packages/discord.js/src/client/Client.js
```diff
@@ -547,6 +557,12 @@ class Client extends BaseClient {
     if (typeof options.ws !== 'object' || options.ws === null) {
       throw new DiscordjsTypeError(ErrorCodes.ClientInvalidOption, 'ws', 'an object');
     }
+    if (
+      (typeof options.ws === 'object' && typeof options.ws.initialPresence !== 'object') ||
+      options.ws.initialPresence === null
```
almeidx, line 562:
```suggestion
      typeof options.ws === 'object' &&
      (typeof options.ws.initialPresence !== 'object' || options.ws.initialPresence === null)
```

almeidx (dismissed):

kyranet (approved):

almeidx (approved):

PR #10420, opened 2024-07-31
refactor!: fully integrate /ws into mainlib

Qjuh:

This removes the compatibility layer of WebSocketManager to integrate `@discordjs/ws` fully into mainlib discord.js without wrapping it.

BREAKING CHANGE: `Client#ws` is now a `@discordjs/ws#WebSocketManager`
BREAKING CHANGE: `WebSocketManager` and `WebSocketShard` are now re-exports from `@discordjs/ws`
BREAKING CHANGE: Removed the `WebSocketShardEvents` enum
BREAKING CHANGE: Renamed the `Client#ready` event to `Client#clientReady` event to not confuse it with the gateway `READY`event
BREAKING CHANGE: Added `Client#ping` to replace the old `WebSocketManager#ping`
BREAKING CHANGE: Removed the `Shard#reconnecting` event which wasn’t emitted anymore since 14.8.0 anyway
BREAKING CHANGE: Removed `ShardClientUtil#ids` and `ShardClientUtil#count` in favor of `Client#ws#getShardIds()` and `Client#ws#getShardCount()`
BREAKING CHANGE: `ClientUser#setPresence()` and `ClientPresence#set()` now return a Promise which resolves when the gateway call was sent successfully
BREAKING CHANGE: Removed `Guild#shard` as `WebSocketShard`s are now handled by `@discordjs/ws`
BREAKING CHANGE: Removed the following deprecated `Client` events: `raw`, `shardDisconnect`, `shardError`, `shardReady`, `shardReconnecting`, `shardResume` in favor of events from `@discordjs/ws#WebSocketManager`
BREAKING CHANGE: Removed `ClientOptions#shards` and `ClientOptions#shardCount` in favor of `ClientOptions#ws#shardIds` and `ClientOptions#ws#shardCount`


packages/discord.js/src/client/Client.js
```diff
@@ -154,6 +148,35 @@ class Client extends BaseClient {
       this.token = null;
     }
 
+    /**
+     * The presence of the Client
+     * @private
+     * @type {ClientPresence}
+     */
+    this.presence = new ClientPresence(this, this.options.presence ?? this.options.ws.initialPresence);
```
didinele: We said we'd remove `ClientOptions#presence`

packages/discord.js/src/client/Client.js
```diff
@@ -154,6 +148,35 @@ class Client extends BaseClient {
...
+
+    if (this.options.presence) {
+      this.options.ws.initialPresence = this.presence._parse(this.options.presence);
```
didinele: ditto

packages/discord.js/src/client/Client.js
```diff
@@ -166,11 +189,33 @@ class Client extends BaseClient {
...
      */
     this.readyTimestamp = null;
+
+    /**
+     * An array of queued events before this Client became ready
+     * @type {Object[]}
+     * @private
+     * @name Client#packetQueue
+     */
+    Object.defineProperty(this, 'packetQueue', { value: [] });
```
didinele, line 213: I'd opt for a different name here, packet is quite ambiguous, at a glance it feels like it could mean "send" as well.

Maybe `incomingEventQueue`

packages/discord.js/src/client/Client.js
```diff
@@ -233,13 +275,149 @@ class Client extends BaseClient {
...
+  }
+
+  /**
+   * Broadcasts a packet to every shard of this client handles.
+   * @param {Object} packet The packet to send
+   * @private
+   */
+  _broadcast(packet) {
+    this.ws.getShardIds().then(shardIds => {
+      for (const shardId of shardIds) this.ws.send(shardId, packet);
```
didinele, line 380: Is it intended this `_broadcast` method resolves before the work is done? `send` is "maybe async", but with current implementations it always is. Don't we want like a `Promise.all`-esque thing going on here?

packages/discord.js/src/sharding/ShardClientUtil.js
```diff
@@ -67,20 +62,18 @@ class ShardClientUtil {
 
   /**
    * Array of shard ids of this client
-   * @type {number[]}
-   * @readonly
+   * @returns {Promise<number[]>}
    */
-  get ids() {
-    return this.client.options.shards;
+  async ids() {
```
didinele, line 67: Are those helpers really needed? 🤔

packages/discord.js/src/sharding/ShardClientUtil.js
```diff
@@ -67,20 +62,18 @@ class ShardClientUtil {
...
   /**
    * Total number of shards
-   * @type {number}
-   * @readonly
+   * @returns {Promise<number>}
    */
-  get count() {
-    return this.client.options.shardCount;
+  async count() {
+    return this.client.ws.getShardCount();
```
didinele, line 76: ditto

packages/discord.js/src/structures/ClientPresence.js
```diff
@@ -22,13 +22,13 @@ class ClientPresence extends Presence {
     const packet = this._parse(presence);
     this._patch(packet);
     if (presence.shardId === undefined) {
-      this.client.ws.broadcast({ op: GatewayOpcodes.PresenceUpdate, d: packet });
+      this.client._broadcast({ op: GatewayOpcodes.PresenceUpdate, d: packet });
     } else if (Array.isArray(presence.shardId)) {
       for (const shardId of presence.shardId) {
-        this.client.ws.shards.get(shardId).send({ op: GatewayOpcodes.PresenceUpdate, d: packet });
+        this.client.ws.send(shardId, { op: GatewayOpcodes.PresenceUpdate, d: packet });
```
didinele, line 28: async, do we want to await?

packages/discord.js/src/structures/ClientPresence.js
```diff
@@ -22,13 +22,13 @@ class ClientPresence extends Presence {
...
       }
     } else {
-      this.client.ws.shards.get(presence.shardId).send({ op: GatewayOpcodes.PresenceUpdate, d: packet });
+      this.client.ws.send(presence.shardId, { op: GatewayOpcodes.PresenceUpdate, d: packet });
```
didinele, line 31: ditto

packages/discord.js/src/structures/Guild.js
```diff
@@ -1418,8 +1409,8 @@ class Guild extends AnonymousGuild {
       this.client.voice.adapters.set(this.id, methods);
       return {
         sendPayload: data => {
-          if (this.shard.status !== Status.Ready) return false;
-          this.shard.send(data);
+          if (this.client.status !== Status.Ready) return false;
+          this.client.ws.send(this.shardId, data);
```
didinele, line 1413: ditto

didinele (changes requested):

packages/discord.js/src/client/Client.js
didinele: this was resolved by the time I submitted my changes

packages/discord.js/src/client/Client.js
Qjuh: `_broadcast` is only really used in `ClientUser#setPresence()`, which immediately patches the presence locally. I'm fine with turning all of those into async to be awaitable along the chain, but that means if one shard fails to send the presence we end with... what exactly? Because some shards successfully set the presence while others didn't. Not to mention that our ClientPresence class doesn't support having different presences per shard stored locally.

packages/discord.js/src/structures/Guild.js
Qjuh: we can't really change this into an async method without also changing `@discoerdjs/voice` because that's where these get used. So I'd say that's outside of the scope of my PR...

packages/discord.js/src/client/Client.js
didinele: > Because some shards successfully set the presence while others didn't.

What do you mean by this? `send` never throws (unless the throttler/whatever has a bug). if setting the presence "fails" it's either silently dropped by Discord or we get an error event

didinele:
@Qjuh You should also search for "deprecate" in the codebase and remove what is related to WS. I've been working on removing stuff from other parts of the codebase, but it's probably best if stuff like https://github.com/discordjs/discord.js/blob/5eabec14d45ef7bdd7f610e84234eb63e726eacd/packages/discord.js/src/errors/ErrorCodes.js was addressed by this PR.

didinele:
For future ref, this should be merged after #10421 - probably cleaner that way.

packages/discord.js/src/client/Client.js
Qjuh: Then what good would awaiting the `send` do? As far as I can tell all that would do would be waiting for the websocket ratelimit to pass if we currently had hit the limit. Because we don't return/await the actual websocket sending in it anyway. Unless that is a sync operation which I hope it isn't.

didinele: Yeah, exactly, it would wait for the message to have actually been sent. I feel like it's just bad design to have an async method that queues load you cannot possibly wait for. We should make it so the user can *choose* to wait (for whatever reason, like they *want* to make sure the presence updates were sent during startup) and as such, `await` whatever calls `_broadcast` (in this case setPresence) or.. not, if they don't care. cc @vladfrangu if you have diff. thoughts

Qjuh: The thing is that the `ClientUser#setPresence()` call returns the ClientPresence which gets immediately patched locally anyway. Making that async would mean we'd have a `Promise<ClientPresence>` although the return value doesn't change after the initial sync call...
And as I said awaiting the `WebSocketManager#send()` call wouldn't make sure the payload actually was sent because it's not awaiting (and can't await) the `WebSocket#send()` call (because the WebSocket API doesn't support it). So if anything making them async/await would give the users a false sense of being able to actually wait for the payload to be sent successfully.

didinele: Re that last part, while true, it's about the queue API we have. Regardless though, I digress. You make a fair point as well, let's let others decide

vladfrangu: Splitting my answers into two parts:
- _broadcast, should it stay, should become async/await and properly handle that
- ClientUser#setPresence becoming async is 🤷‍♂️. Imo we can either make it so (in which case that'll cascade for setActivity, and technically make it consistent with other methods), or call `void broadcast`. Both are fine _with me_, but maybe others disagree. @discordjs/core ☝️ thoughts?

Qjuh: Since _broadcast literally only is used for ClientPresence#set I'll just make all of them async/await. Users of discord.js can still decide whether they want to await it or not i guess but you convinced me that it's the better approach.

packages/discord.js/src/util/Events.js
```diff
@@ -99,7 +94,7 @@ module.exports = {
   ChannelDelete: 'channelDelete',
   ChannelPinsUpdate: 'channelPinsUpdate',
   ChannelUpdate: 'channelUpdate',
-  ClientReady: 'ready',
+  ClientReady: 'clientReady',
```
TAEMBO, line 97: If this is going to be changed from `ready` to `clientReady`, shouldn't the types for [`Events.ClientReady`](https://github.com/discordjs/discord.js/blob/main/packages/discord.js/typings/index.d.ts#L5433) and [`ClientEvents`](https://github.com/discordjs/discord.js/blob/main/packages/discord.js/typings/index.d.ts#L5253) be updated as well?

TAEMBO (approved):

didinele (approved):

packages/discord.js/src/util/Events.js
```diff
@@ -148,12 +143,6 @@ module.exports = {
   MessageReactionRemoveEmoji: 'messageReactionRemoveEmoji',
   MessageUpdate: 'messageUpdate',
   PresenceUpdate: 'presenceUpdate',
-  Raw: 'raw',
-  ShardDisconnect: 'shardDisconnect',
-  ShardError: 'shardError',
-  ShardReady: 'shardReady',
-  ShardReconnecting: 'shardReconnecting',
-  ShardResume: 'shardResume',
```
vladfrangu, line 156: Why are we removing some of these? We can still emit them from ws, no?

vladfrangu (changes requested):

packages/discord.js/src/client/Client.js
```diff
@@ -243,13 +273,148 @@ class Client extends BaseClient {
...
+   * @private
+   */
+  _attachEvents() {
+    this.ws.on(WebSocketShardEvents.Debug, (message, shardId) =>
+      this.emit(Events.Debug, `[WS => ${typeof shardId === 'number' ? `Shard ${shardId}` : 'Manager'}] ${message}`),
+    );
+    this.ws.on(WebSocketShardEvents.Dispatch, this._handlePacket.bind(this));
+
+    this.ws.on(WebSocketShardEvents.Ready, data => {
+      this.expectedGuilds = new Set([...this.expectedGuilds, ...data.guilds.map(guild => guild.id)]);
```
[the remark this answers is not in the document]
didinele: Just from a perf. standpoint it's more natural to use a for loop just adding the guilds

packages/discord.js/src/client/Client.js
```diff
@@ -219,21 +260,10 @@ class Client extends BaseClient {
...
-
-        deprecationEmittedForClientPresence = true;
-      }
-
-      this.options.ws.presence = this.presence._parse(this.options.presence);
-    }
-
     this.emit(Events.Debug, 'Preparing to connect to the gateway...');
 
+    if (this.ws.options.token !== this.token) this.ws.setToken(this.token);
```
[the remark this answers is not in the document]
vladfrangu: Then this code will throw anyways if I pass in ws.tokens in client options and login with a different token 👁️

packages/discord.js/src/util/Events.js
didinele: Use `client.ws.on('resume')` (etc) instead, no real reason to re-emit here.

vladfrangu: True

packages/discord.js/src/client/Client.js
```diff
@@ -243,13 +273,148 @@ class Client extends BaseClient {
...
+
+    this.ws.on(WebSocketShardEvents.Ready, data => {
+      this.expectedGuilds = new Set([...this.expectedGuilds, ...data.guilds.map(guild => guild.id)]);
+      this.status = Status.WaitingForGuilds;
+      this._checkReady();
+    });
+
+    this.ws.on(WebSocketShardEvents.HeartbeatComplete, ({ heartbeatAt, latency }, shardId) => {
+      this.emit(Events.Debug, `[Shard ${shardId}] Heartbeat acknowledged, latency of ${latency}ms.`);
+      this.lastPingTimestamp = heartbeatAt;
```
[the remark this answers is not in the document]
Qjuh: This depends on how this field is interpreted I guess. How I implemented it currently is the last time any ping from discord was received. Independent of which shard received it. I'm fine with dropping or turning into a Collection keyed by shardId though.

packages/discord.js/src/client/Client.js
```diff
@@ -243,13 +273,148 @@ class Client extends BaseClient {
...
+    }
+    // Step 1. If we don't have any other guilds pending, we are ready
+    if (
+      !this.expectedGuilds.size &&
+      (await this.ws.fetchStatus()).every(status => status === WebSocketShardStatus.Ready)
+    ) {
+      this.emit(Events.Debug, 'Client received all its guilds. Marking as fully ready.');
+      this.status = Status.Ready;
+
+      this._triggerClientReady();
```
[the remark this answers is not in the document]
Qjuh: Already there in line 283

vladfrangu:
This needs a rebase

packages/discord.js/src/client/Client.js
```diff
@@ -231,13 +273,150 @@ class Client extends BaseClient {
...
+        setImmediate(() => {
+          this._handlePacket(item.packet, item.shardId);
+        }).unref();
+      }
+
+      if (packet && PacketHandlers[packet.t]) {
+        PacketHandlers[packet.t](this, packet, shardId);
+      }
+
+      if (this.status === Status.WaitingForGuilds && WaitingForGuildEvents.includes(packet.t)) {
```
kyranet, line 369: Considering `packet` is optional in the method's definition, and is checked accordingly for truthiness in all other branches, this line may throw an error were `packet` to be nullish.

kyranet (changes requested):

kyranet (approved):

iCrawl (approved):

PR #10424, opened 2024-08-02
fix(scripts): show name of inheriting class on search index

Qjuh:

With #10415 the name on search index includes the name of the class now, but for inherited properties it's the name of the defining class instead of the one the entry links to.

This PR fixes this and also changes the delimiter so the name for https://discord.js.org/docs/packages/discord.js/14.15.3/ClientUser:Class#bot would be `ClientUser#bot` instead of `User.bot`now.

### DIFF
--- a/packages/scripts/src/generateIndex.ts
+++ b/packages/scripts/src/generateIndex.ts
@@ -129,7 +129,7 @@ export function visitNodes(item: ApiItem, tag: string) {
 
 		members.push({
 			id: idx++,
-			name: (inherited
+			name: (inherited && member.parent
 				? member.getScopedNameWithinPackage().replace(new RegExp(`^${member.parent?.displayName}`), item.displayName)
 				: member.getScopedNameWithinPackage()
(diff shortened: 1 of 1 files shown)


vladfrangu (approved):

packages/scripts/src/generateIndex.ts
```diff
@@ -129,7 +129,10 @@ export function visitNodes(item: ApiItem, tag: string) {
 
 		members.push({
 			id: idx++,
-			name: member.getScopedNameWithinPackage(),
+			name: (inherited
+				? member.getScopedNameWithinPackage().replace(new RegExp(`^${member.parent?.displayName}`), item.displayName)
```
ckohen, line 133: edge case, but this regex will match a class named undefined if no parent...I think checking inherited is supposed to be our guarantee this doesn't happen right?. You could also just check both

ckohen (approved):

almeidx (approved):

PR #10426, opened 2024-08-04
revert: chore: deprecate client options presence

almeidx:
This reverts commit 8f97d2bacf78544e60935405316024a6bc4a4a50.

After internal discussion, we decided to revert this change. See https://github.com/discordjs/discord.js/pull/10426#discussion_r1703174908 for more info

### DIFF
--- a/packages/discord.js/src/client/Client.js
+++ b/packages/discord.js/src/client/Client.js
@@ -32,7 +32,6 @@ const Status = require('../util/Status');
 const Sweepers = require('../util/Sweepers');
 
 let deprecationEmittedForPremiumStickerPacks = false;
-let deprecationEmittedForClientPresence = false;
 
 /**
  * The main hub for interacting with the Discord API, and the starting point for any bot.
@@ -140,7 +139,7 @@ class Client extends BaseClient {
      * @private
      * @type {ClientPresence}
      */
-    this.presence = new ClientPresence(this, this.options.ws.presence ?? this.options.presence);
+    this.presence = new ClientPresence(this, this.options.presence);
 
     Object.defineProperty(this, 'token', { writable: true });
     if (!this.token && 'DISCORD_TOKEN' in process.env) {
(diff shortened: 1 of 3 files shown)


didinele (approved):

kyranet (approved):

packages/discord.js/typings/index.d.ts
```diff
@@ -6878,7 +6877,7 @@ export interface WebhookMessageCreateOptions extends Omit<MessageCreateOptions,
 export interface WebSocketOptions {
   large_threshold?: number;
   version?: number;
-  presence?: GatewayPresenceUpdateData;
+  presence?: PresenceData;
```
Qjuh, line 6880: Why was this changed? The goal here is to have those types being forward compatible and `@discordjs/ws` expects the previous type.

packages/discord.js/src/util/Options.js
```diff
@@ -91,7 +91,7 @@ const { version } = require('../../package.json');
  * only set this if you know what you are doing</warn>
  * @property {BuildStrategyFunction} [buildStrategy] Builds the strategy to use for sharding
  * @property {IdentifyThrottlerFunction} [buildIdentifyThrottler] Builds the identify throttler to use for sharding
- * @property {PresenceData} [presence={}] Presence data to use upon login
+ * @property {PresenceData} [presence] Presence data to use upon login
```
Qjuh, line 94: We should add `initialPresence` here because that’s what /ws uses and should be the property we suggest as alternative to the deprecated `ClientOptions#presence`.

packages/discord.js/src/client/Client.js
```diff
@@ -230,6 +230,8 @@ class Client extends BaseClient {
       }
 
       this.options.ws.presence = this.presence._parse(this.options.presence);
+    } else if (this.options.ws.presence) {
+      this.options.ws.presence = this.presence._parse(this.options.ws.presence);
```
Qjuh, line 234: This should parse `this.options.ws.initialPresence` with the ws.presence as fallback with nullish coalesce.

Qjuh (changes requested):

vladfrangu (dismissed):
Requesting changes too in case Qjuhs doesn't count

packages/discord.js/typings/index.d.ts
almeidx: I changed it to match the type in the JSDoc. I'll revert this and change it in the JSDoc instead

almeidx: Changing this to the gateway type forces users to pass all of its properties at all times, e.g.:
image

IMO it doesn't hurt to keep the parsing on our side... Thoughts? @discordjs/core

didinele: I'm fine with this as long as we do it exactly as we're doing intents (see Qjuh's /ws integration PR).

didinele (approved):

vladfrangu (approved):

PR #10427, opened 2024-08-04
chore: pin /ws version in discord.js

almeidx:
There have been breaking changes in /ws which discord.js doesn't support, so pinning the version to not break potential future v14 releases

### DIFF
--- a/packages/discord.js/package.json
+++ b/packages/discord.js/package.json
@@ -70,7 +70,7 @@
     "@discordjs/formatters": "workspace:^",
     "@discordjs/rest": "workspace:^",
     "@discordjs/util": "workspace:^",
-    "@discordjs/ws": "workspace:^",
+    "@discordjs/ws": "1.1.1",
     "@sapphire/snowflake": "3.5.3",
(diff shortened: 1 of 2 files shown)


packages/discord.js/package.json
```diff
@@ -70,7 +70,7 @@
     "@discordjs/formatters": "workspace:^",
     "@discordjs/rest": "workspace:^",
     "@discordjs/util": "workspace:^",
-    "@discordjs/ws": "workspace:^",
+    "@discordjs/ws": "1.1.1",
```
vladfrangu, line 73: We can use a wider range

```suggestion
    "@discordjs/ws": "^1.1.1",
```

vladfrangu (approved):

didinele (approved):
what vlad said

packages/discord.js/package.json
almeidx: Every other dependency is pinned without ranges, including `@discordjs/collection`. Don't see why this one should be different
cc @iCrawl

iCrawl (approved):

PR #10429, opened 2024-08-04
fix(Shard): add env, execArgv, and argv for worker-based shards

catplvsplus:
Adds `env`, `execArgv`, `argv` to worker thread creation options. This allows the [`ShardingManagerOptions#execArgv`](https://discordjs.dev/docs/packages/discord.js/14.15.3/ShardingManagerOptions:Interface#execArgv), [`ShardingManagerOptions#shardArgs`](https://discordjs.dev/docs/packages/discord.js/14.15.3/ShardingManagerOptions:Interface#shardArgs) and [`Shard#env`](https://discordjs.dev/docs/packages/discord.js/14.15.3/Shard:Class#env) to be used not only by the child-process based shards.


almeidx (dismissed):
These doc comments need updating

https://github.com/discordjs/discord.js/blob/f623e7a315321fdde498140277c5d5f0eb447de7/packages/discord.js/src/sharding/Shard.js#L51-L61

packages/discord.js/src/sharding/Shard.js
```diff
@@ -136,7 +136,12 @@ class Shard extends EventEmitter {
           .on('exit', this._exitListener);
         break;
       case 'worker':
-        this.worker = new Worker(path.resolve(this.manager.file), { workerData: this.env })
+        this.worker = new Worker(path.resolve(this.manager.file), {
+          workerData: this.env,
+          env: this.env,
+          execArgv: this.execArgv,
+          argv: this.args,
+        })
```
Qjuh, line 144: As seen here the `env` already gets passed to the workers just fine. And the other two options are literally meant for being passed to the node executable when running the child process. Which simply doesn’t happen for worker threads, so I don’t see why you would want to pass them along (unless if you abuse them to have three independent variables you can pass stuff in).

Qjuh: And especially for the `env` the default value is already `process.env` which is actually better than `this.env` because the additional properties we set in `this.env` are only meant for the Shard‘s Client and not the whole process.

catplvsplus: > And the other two options are literally meant for being passed to the node executable when running the child process. Which simply doesn’t happen for worker threads, so I don’t see why you would want to pass them along (unless if you abuse them to have three independent variables you can pass stuff in).

The `execArgv` and `argv` serve a similar purpose as child process's `execArgv` and `args`. I always thought that those options were previously undefined due to lack of support in older Node.js versions, but given that discord.js now requires Node.js version `>=16.11.0`, incorporating these options should be feasible.

> As seen here the env already gets passed to the workers just fine.

> And especially for the env the default value is already process.env which is actually better than this.env because the additional properties we set in this.env are only meant for the Shard‘s Client and not the whole process.

Since `this.env` is a copy of `process.env`, I see no problem in utilizing it. It could provide information from the sharding manager, as opposed to accessing it through the Shard's client or `workerData`. This approach ensures consistency with the env of `process`-based shards.

SpaceEEC (approved):

packages/discord.js/src/sharding/Shard.js
vladfrangu: If `env` is supposed to be a copy of process.env, please use the `workerThreads.SHARE_ENV` constant instead

PR #10434, opened 2024-08-07
refactor(actions): safer getChannel calls

### LINKED ISSUE #10393: TypeError: thread.members._add is not a function

discord.js

### Issue description

Creating this issue for posterity, bulk of the discussion and fix were noted here: https://discord.com/channels/222078108977594368/1248260322163294238

After a recent update to @discordjs/ws, it seems like Messages are being instantiated as channels in some very specific cases. 

Full trace of the issue is here:
```js
Trace
    at TextChannel._patch (/app/node_modules/discord.js/src/structures/TextChannel.js:15:15)
    at ChannelManager._add (/app/node_modules/discord.js/src/managers/ChannelManager.js:42:27)
    at MessageCreateAction.getPayload (/app/node_modules/discord.js/src/client/actions/Action.js:27:73)
    at MessageCreateAction.getChannel (/app/node_modules/discord.js/src/client/actions/Action.js:44:12)
    at MessageCreateAction.handle (/app/node_modules/discord.js/src/client/actions/MessageCreate.js:9:26)
    at module.exports [as MESSAGE_CREATE] (/app/node_modules/discord.js/src/client/websocket/handlers/MESSAGE_CREATE.js:4:32)
    at WebSocketManager.handlePacket (/app/node_modules/discord.js/src/client/websocket/WebSocketManager.js:355:31)
    at WebSocketManager.<anonymous> (/app/node_modules/discord.js/src/client/websocket/WebSocketManager.js:239:12)
    at WebSocketManager.emit (/app/node_modules/@vladfrangu/async_event_emitter/dist/index.cjs:282:31)
    at WebSocketShard.<anonymous> (/app/node_modules/@discordjs/ws/dist/index.js:1190:51)
```

Sample "TextChannel" that is actually a "thread":
```js
<ref *2> TextChannel {
  type: 0,
  guild: <ref *1> Guild {
    id: '755765126140919881',
    name: '彩虹研究社:rainbow:',
    icon: 'b2ea11110ab979ddda535b7c4479efc9',
    features: [
      'AUTO_MODERATION',
      'ROLE_ICONS',
      'SEVEN_DAY_THREAD_ARCHIVE',
      'ANIMATED_BANNER',
      'NEWS',
      'SOUNDBOARD',
      'BANNER',
      'ANIMATED_ICON',
      'INVITE_SPLASH',
      'CHANNEL_ICON_EMOJIS_GENERATED',
      'THREE_DAY_THREAD_ARCHIVE',
      'MEMBER_PROFILES',
      'COMMUNITY',
      'VANITY_URL',
      'PRIVATE_THREADS'
    ],
    (TRUNCATING SO I CAN PASTE IN DISCORD)
  },
  guildId: '755765126140919881',
  permissionOverwrites: PermissionOverwriteManager { channel: [Circular *2] },
  messages: GuildMessageManager { channel: [Circular *2] },
  threads: GuildTextThreadManager { channel: [Circular *2] },
  nsfw: false,
  flags: ChannelFlagsBitField { bitfield: 0 },
  id: '1262790110647156736',
  parentId: null,
  defaultThreadRateLimitPerUser: null,
  memberCount: 7,
  rawPosition: 4,
  lastMessageId: '1262806103016607784'
}
```
(Note the `memberCount` property here means we know it's a thread)

I think the possible problem commit was: https://github.com/discordjs/discord.js/pull/10278

### Code sample

```typescript
Not sure on exactly what the code sample is to reproduce this yet
```

### Versions

- discord.js v14.15.3
- nodejs v21.7.1

### Issue priority

Medium (should be fixed soon)

User, Channel, GuildMember, Message, Reaction, ThreadMember

Guilds, GuildMembers, GuildPresences, GuildMessages, DirectMessages

_No response_

didinele:
Fixes #10393 and also does a little spring cleaning on areas that could have also been dangerous in the same way.

The gist of #10393 was that it was especially happening when a message sent unarchived a thread, and yet it only *sometimes* caused the error. It seems that sometimes, Discord was sending the `MESSAGE_CREATE` payload before the `THREAD_UPDATE`. The latter normally comes first and populates our cache, which yields to the subsequent `getChannel` caused by the `MESSAGE_CREATE` to return from cache.

In cases when the opposite was happening, before #10278 we were unable to resolve the channel type and would simply skip it, not adding it to cache at all (and also skipping the messageCreate) as a result. **This is a fundamental flaw of how those partials work, and should be addressed with either the next major, or the library's rewrite**. There is no easy way to otherwise fix this right now.

After #10278, we were essentially passing full message objects from `MESSAGE_CREATE` to `getChannel`, messages, which, have a `type` property, often `0`, which lead to a successful channel creation with jank data.

With this fix, we're back to the old behavior of dropping those events. Regarding the properties I chose to forward, I followed the payload docs found here https://discord.com/developers/docs/topics/gateway-events#gateway-events


kyranet (approved):

packages/discord.js/src/client/actions/MessageReactionAdd.js
```diff
@@ -22,7 +22,7 @@ class MessageReactionAdd extends Action {
     if (!user) return false;
 
     // Verify channel
-    const channel = this.getChannel(data);
+    const channel = this.getChannel({ id: data.channel_id, guild_id: data.guild_id, user_id: data.user_id });
```
Qjuh, line 25: Channels have no user_id
```suggestion
    const channel = this.getChannel({ id: data.channel_id, guild_id: data.guild_id });
```

packages/discord.js/src/client/actions/MessageReactionRemove.js
```diff
@@ -19,7 +19,7 @@ class MessageReactionRemove extends Action {
     if (!user) return false;
 
     // Verify channel
-    const channel = this.getChannel(data);
+    const channel = this.getChannel({ id: data.channel_id, guild_id: data.guild_id, user_id: data.user_id });
```
Qjuh, line 22: Same
```suggestion
    const channel = this.getChannel({ id: data.channel_id, guild_id: data.guild_id });
```

Qjuh (changes requested):

packages/discord.js/src/client/actions/MessageReactionAdd.js
didinele: https://github.com/discordjs/discord.js/blob/432e9b8425311f132e3970d238005c1166215dbb/packages/discord.js/src/client/actions/Action.js#L36

Qjuh: Oh true, my bad…

Qjuh (approved):

vladfrangu (approved):

SpaceEEC (approved):

PR #10451, opened 2024-08-19
feat(voice)!: add new encryption methods, remove old methods

nyapat:

- [x] aes256gcm encryption
- [x] xchacha20 encryption
- [x] aes256gcm decryption
- [x] xchacha20 decryption
- [x] test if alternative sodiums work for above
- [x] alternative for tweetnacl
- [x] remove old encryption, decryption methods
- [x] tests (waiting on #10472)

Added the new encryption methods that Discord will enforce for all voice users in November. Since the older methods are being [deprecated entirely](https://discord.com/developers/docs/change-log#voice-encryption-modes), I am removing them as well. Because tweetnacl does not support the cipher methods necessary for xchacha, I have removed it and replaced it with @stablelib/xchacha20poly1305 (aes256gcm does not require anything other than node:crypto, so that could theoretically be the only one needed if the vc always requested that)

Testing has been added for the encryption methods aswell, and I have changed some older tests to properly remove the RTP header based on the original encrypted data.


packages/ws/src/ws/WebSocketShard.ts
Qjuh, line 1: This is the `WebSocket` connection for the bot gateway, completely unrelated to the `@discordjs/voice` websocket or the change this PR addresses. Please revert as this is out of scope of the PR.

Qjuh (changes requested):

packages/voice/src/receive/VoiceReceiver.ts
```diff
@@ -87,22 +92,43 @@ export class VoiceReceiver {
...
+		const header = buffer.slice(0, headerSize);
+
+		// Encrypted contains the extension, if any, the opus packet, and the auth tag
+		const encrypted = buffer.slice(headerSize, buffer.length - AUTH_TAG_LENGTH - UNPADDED_NONCE_LENGTH);
+		const authTag = buffer.slice(
+			buffer.length - AUTH_TAG_LENGTH - UNPADDED_NONCE_LENGTH,
+			buffer.length - UNPADDED_NONCE_LENGTH,
+		);
+
+		if (mode === 'aead_aes256_gcm_rtpsize') {
```
vladfrangu, line 112: Ditto

packages/voice/src/util/Secretbox.ts
```diff
@@ -33,16 +63,57 @@ const libs = {
...
+		) => {
+			return sodium.crypto_aead_xchacha20poly1305_ietf_decrypt(null, message, additionalData, nonce, key);
+		},
 	}),
 	tweetnacl: (tweetnacl: any): Methods => ({
 		open: tweetnacl.secretbox.open,
 		close: tweetnacl.secretbox,
 		random: tweetnacl.randomBytes,
+
+		crypto_aead_xchacha20poly1305_ietf_encrypt: () => {
```
vladfrangu, line 111: If tweetnacl doesn't support this, lets deprecate or remove it and release a major version for voice. or see if theres an alternative to tweetnacl

packages/voice/src/util/Secretbox.ts
```diff
@@ -1,70 +1,108 @@
...
+	crypto_aead_xchacha20poly1305_ietf_encrypt: fallbackError,
+	crypto_aead_xchacha20poly1305_ietf_decrypt: fallbackError,
 };
 
 void (async () => {
 	for (const libName of Object.keys(libs) as (keyof typeof libs)[]) {
 		try {
-			// eslint-disable-next-line @typescript-eslint/no-require-imports, @typescript-eslint/no-var-requires
-			const lib = require(libName);
+			const lib = await import(libName);
```
vladfrangu, line 105: Any particular reason for this change? Do we have any way to _not_ require this to become async (as it can come with its own can of worms)... Maybe we should make this secretbox module a lazy loaded thing thats awaitable (export the two functions that handle imports themselves the first time)

vladfrangu (changes requested):

packages/voice/src/util/Secretbox.ts
almeidx: You can create the `require` function [in ESM] by using [`createRequire()`](https://nodejs.org/api/module.html#modulecreaterequirefilename) from `node:module`

vladfrangu: Realistically its still something we ought to change, this random `void Promise` thing is hella janky imo (and prone to race conditions or worse)

didinele: I'm for a `lazy` top level.

vladfrangu: Would it cause issues to export en/decrypt methods top level that are async and that handle the imports themselves? 👀 (or that are sync and export an `initCrypto` or w/e function that gets called first thing on connect, assuming then that connect is async)

Alternatively similar to what dd said (which we already do in /ws for zlib-sync)

vladfrangu: Was the move to `import` required? Otherwise we should probably leave it be as is unless others have a better idea

vladfrangu: I mean you didn't show what the alternative is 😅, so 🤷‍♂️

vladfrangu: Sure! But would love the thoughts of @didinele and/or @discordjs/core about this too (if we should just leave it as is or not)

kyranet: [`@noble/ciphers`](https://www.npmjs.com/package/@noble/ciphers) looks good to me 👍🏼

packages/voice/src/receive/VoiceReceiver.ts
```diff
@@ -87,22 +92,50 @@ export class VoiceReceiver {
...
+		const header = buffer.subarray(0, headerSize);
+
+		// Encrypted contains the extension, if any, the opus packet, and the auth tag
+		const encrypted = buffer.subarray(headerSize, buffer.length - AUTH_TAG_LENGTH - UNPADDED_NONCE_LENGTH);
+		const authTag = buffer.subarray(
+			buffer.length - AUTH_TAG_LENGTH - UNPADDED_NONCE_LENGTH,
+			buffer.length - UNPADDED_NONCE_LENGTH,
+		);
+
+		let decrypted;
```
SpaceEEC, line 112: This variable seems unused.
```suggestion
```

packages/voice/src/receive/VoiceReceiver.ts
```diff
@@ -87,22 +92,50 @@ export class VoiceReceiver {
...
+		switch (mode) {
+			case 'aead_aes256_gcm_rtpsize': {
+				const decipheriv = crypto.createDecipheriv('aes-256-gcm', secretKey, nonce);
+				decipheriv.setAAD(header);
+				decipheriv.setAuthTag(authTag);
+
+				decrypted = Buffer.concat([decipheriv.update(encrypted), decipheriv.final()]);
+				return decrypted;
```
SpaceEEC, line 120:
```suggestion
				return Buffer.concat([decipheriv.update(encrypted), decipheriv.final()]);
```

SpaceEEC (changes requested):
Just a small nit of mine, otherwise this looks good.
Thank you for this PR. 🙇

packages/voice/src/util/Secretbox.ts
vladfrangu: I'll take a look at helping out with this 🫡

SpaceEEC (approved):

packages/voice/src/util/Secretbox.ts
```diff
@@ -1,75 +1,135 @@
...
+
 			Object.assign(methods, libs[libName](lib));
+
 			break;
 		} catch {}
 	}
-})();
+
+	resolve();
+});
```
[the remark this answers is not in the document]
vladfrangu: I have tested this locally with both noble and @stablelib/xchacha20poly1305 and have had 0 issues (and I forced it to use the secretbox methods and not AES by removing AES locally as a supported type)

packages/voice/src/util/Secretbox.ts
Qjuh: Shouldn’t this be reverted to require then (since that was the whole reason for choosing that lib)?

packages/voice/src/util/Secretbox.ts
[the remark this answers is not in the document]
vladfrangu: Annoyingly, we don't have a place where we can safely call this, outside of making `joinVoiceChannel` async and awaiting the promise there (which we can do but it'd be sem major then)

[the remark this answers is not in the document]
vladfrangu: iTerm2 - 2024-11-17 at 22 44 01

It's still started when creating! But it's definitely obscure 😓, and I'd be down to make it handled in joinVoiceChannel (calling load), and throw if no crypto.AES is available and no secretbox either

iCrawl (approved):

kyranet (approved):

PR #10471, opened 2024-09-02
chore: update cliff configs

ckohen:

Not all footers are breaking changes
since version 2 of git-cliff the tag_pattern option is a regex, fixes discord.js tag matching 

### DIFF
--- a/packages/builders/cliff.toml
+++ b/packages/builders/cliff.toml
@@ -29,8 +29,10 @@ body = """
 		  {% endif %}\
 			{{ commit.message | upper_first }} ([{{ commit.id | truncate(length=7, end="") }}]({{ self::remote_url() }}/commit/{{ commit.id }}))\
 		{% if commit.breaking %}\
-			{% for breakingChange in commit.footers %}\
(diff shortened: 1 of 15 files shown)


Jiralite (approved):

packages/builders/cliff.toml
```diff
@@ -29,8 +29,10 @@ body = """
 		  {% endif %}\
 			{{ commit.message | upper_first }} ([{{ commit.id | truncate(length=7, end="") }}]({{ self::remote_url() }}/commit/{{ commit.id }}))\
 		{% if commit.breaking %}\
-			{% for breakingChange in commit.footers %}\
```
vladfrangu, line 32: You could also use the `| filter(breaking=true)` filter or w/e it is

vladfrangu (changes requested):

packages/builders/cliff.toml
ckohen: I chose this way because it allows us to do other things with other footers easily if we wanted. Also reads a bit cleaner to me.

vladfrangu (approved):

didinele (approved):

iCrawl (approved):

PR #10477, opened 2024-09-03
feat: Add ApplicationEmoji to EmojiResolvable and MessageReaction#emoji

Moebits:

When using typescript you can't react to a message with an `ApplicationEmoji` because `EmojiResolvable` lacks the type.

Edit: I also updated BaseGuildEmojiManager.js and added `ApplicationEmoji` to the `MessageReaction#emoji` getter.

### DIFF
--- a/packages/discord.js/src/structures/MessageReaction.js
+++ b/packages/discord.js/src/structures/MessageReaction.js
@@ -121,18 +121,17 @@ class MessageReaction {
     if (this._emoji instanceof ApplicationEmoji) return this._emoji;
     // Check to see if the emoji has become known to the client
     if (this._emoji.id) {
+      const applicationEmojis = this.message.client.application.emojis.cache;
+      if (applicationEmojis.has(this._emoji.id)) {
+        const emoji = applicationEmojis.get(this._emoji.id);
+        this._emoji = emoji;
+        return emoji;
+      }
       const emojis = this.message.client.emojis.cache;
       if (emojis.has(this._emoji.id)) {
         const emoji = emojis.get(this._emoji.id);
         this._emoji = emoji;
         return emoji;
-      } else {
-        const applicationEmojis = this.message.client.application.emojis.cache;
-        if (applicationEmojis.has(this._emoji.id)) {
-          const emoji = applicationEmojis.get(this._emoji.id);
-          this._emoji = emoji;
-          return emoji;
-        }
       }
     }
     return this._emoji;
--- a/packages/discord.js/src/structures/MessageReaction.js
+++ b/packages/discord.js/src/structures/MessageReaction.js
@@ -1,6 +1,7 @@
 'use strict';
 
 const { Routes } = require('discord-api-types/v10');
+const ApplicationEmoji = require('./ApplicationEmoji');
 const GuildEmoji = require('./GuildEmoji');
 const ReactionEmoji = require('./ReactionEmoji');
 const ReactionUserManager = require('../managers/ReactionUserManager');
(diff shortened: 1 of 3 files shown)


imnaiyar:
This may also need updating 
https://github.com/discordjs/discord.js/blob/main/packages%2Fdiscord.js%2Fsrc%2Fmanagers%2FBaseGuildEmojiManager.js#L23-L28

Moebits:
Thanks I updated BaseGuildEmojiManager.js and it seemed to fix the equality comparison with `MessageReaction#emoji` as well.

vladfrangu (approved):

packages/discord.js/typings/index.d.ts
```diff
@@ -2429,13 +2429,13 @@ export class MessagePayload {
...
   private constructor(client: Client<true>, data: RawMessageReactionData, message: Message);
-  private _emoji: GuildEmoji | ReactionEmoji;
+  private _emoji: GuildEmoji | ReactionEmoji | ApplicationEmoji;
 
   public burstColors: string[] | null;
   public readonly client: Client<true>;
   public count: number;
   public countDetails: ReactionCountDetailsData;
-  public get emoji(): GuildEmoji | ReactionEmoji;
+  public get emoji(): GuildEmoji | ReactionEmoji | ApplicationEmoji;
```
TAEMBO, line 2438: In its current state, it is not possible for `MessageReaction#emoji` along with its associated private property to return an ApplicationEmoji, which would make these typings inaccurate.

Moebits: I think it makes sense for a MessageReaction done by the bot with an ApplicationEmoji to return ApplicationEmoji for its emoji property. Do you know where I can make these changes?

vladfrangu: Since application emojis need to be fetched to be had, I wouldn't actually try to squeeze that in. it's fine to leave it as before

Qjuh: why not add a cache check for ApplicationEmoji here?
https://github.com/discordjs/discord.js/blob/4594896b5404c6a34e07544951c59ff8f3657184/packages/discord.js/src/structures/MessageReaction.js#L126

vladfrangu: Oof, _that_ getter. I'd maybe add that logic into the getter to store app emojis if we *really* wanna go there

Moebits: I have added `ApplicationEmoji` lookup to the getter.

Qjuh: That now changes the scope of the PR to not be types only. Please edit the PR title and description accordingly and make sure you actually tested that change.

TAEMBO (approved):

packages/discord.js/src/structures/MessageReaction.js
```diff
@@ -108,21 +109,30 @@ class MessageReaction {
...
+    if (this._emoji instanceof ApplicationEmoji) return this._emoji;
     // Check to see if the emoji has become known to the client
     if (this._emoji.id) {
       const emojis = this.message.client.emojis.cache;
       if (emojis.has(this._emoji.id)) {
         const emoji = emojis.get(this._emoji.id);
         this._emoji = emoji;
         return emoji;
+      } else {
+        const applicationEmojis = this.message.client.application.emojis.cache;
```
vladfrangu, line 130: Should be moved under line 124

vladfrangu (changes requested):

vladfrangu (approved):

didinele (approved):

kyranet (approved):

PR #10478, opened 2024-09-03
types: export GroupDM helper type

Qjuh:

Export the helper type used to remove  `PartialGroupDMChannel` as possible value for `Structure#channel` as it’s also useful for users.


packages/discord.js/typings/index.d.ts
```diff
@@ -5270,7 +5270,7 @@ export interface GuildMembersChunk {
   nonce: string | undefined;
 }
 
-type NonPartialGroupDMChannel<Structure extends { channel: Channel }> = Structure & {
+export type NonPartialGroupDMChannel<Structure extends { channel: Channel }> = Structure & {
```
vladfrangu, line 5273: If we want to expose this lets give this a better name. As it stands, this name is very... confuse

vladfrangu (approved):

almeidx (approved):

Jiralite (approved):

PR #10480, opened 2024-09-03
docs(ApplicationEmojiManager): fix fetch example

sdanialraza:
Fixes the `ApplicationEmojiManager#fetch` example, `<Message>.application` doesn't exist

### DIFF
--- a/packages/discord.js/src/managers/ApplicationEmojiManager.js
+++ b/packages/discord.js/src/managers/ApplicationEmojiManager.js
@@ -65,12 +65,12 @@ class ApplicationEmojiManager extends CachedManager {
    * @returns {Promise<ApplicationEmoji|Collection<Snowflake, ApplicationEmoji>>}
    * @example
    * // Fetch all emojis from the application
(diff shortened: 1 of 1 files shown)


vladfrangu (approved):
Could've also just dropped message from there

packages/discord.js/src/managers/ApplicationEmojiManager.js
```diff
@@ -65,12 +65,12 @@ class ApplicationEmojiManager extends CachedManager {
    * @returns {Promise<ApplicationEmoji|Collection<Snowflake, ApplicationEmoji>>}
    * @example
    * // Fetch all emojis from the application
-   * message.application.emojis.fetch()
+   * message.client.application.emojis.fetch()
```
almeidx, line 68: This should be consistent with the `create` method examples. Either use `client.application` or just `application` on both

almeidx (changes requested):

almeidx (approved):

PR #10482, opened 2024-09-04
fix: type guard for sendable text-based channels

vladfrangu:

Adds in `BaseChannel#isSendable`, which is a mix of `isTextBased()` and checking if you can actually send messages in the channel. This way, we can have the correct types for interaction.channel, and you can assert whether you got the interaction from a Group DM Channel or not

### DIFF
--- a/packages/discord.js/src/structures/BaseChannel.js
+++ b/packages/discord.js/src/structures/BaseChannel.js
@@ -156,11 +156,11 @@ class BaseChannel extends Base {
   }
 
   /**
-   * Indicates whether this channel is {@link TextBasedChannels text-based} and can have messages sent in it.
+   * Indicates whether this channel is sendable.
    * @returns {boolean}
    */
   isSendable() {
-    return this.isTextBased() && 'send' in this;
+    return 'send' in this;
   }
 
   toJSON(...props) {
--- a/packages/discord.js/src/structures/BaseChannel.js
+++ b/packages/discord.js/src/structures/BaseChannel.js
@@ -155,6 +155,14 @@ class BaseChannel extends Base {
     return 'availableTags' in this;
   }
 
+  /**
+   * Indicates whether this channel is {@link TextBasedChannels text-based} and can have messages sent in it.
+   * @returns {boolean}
+   */
+  isSendable() {
+    return this.isTextBased() && 'send' in this;
+  }
+
   toJSON(...props) {
     return super.toJSON({ createdTimestamp: true }, ...props);
   }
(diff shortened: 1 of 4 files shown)


imnaiyar:
IG this (and the JSDoc) can also be updated to include PartialGroupDMChannel since it's considered text-based now
https://github.com/discordjs/discord.js/blob/4594896b5404c6a34e07544951c59ff8f3657184/packages/discord.js/src/util/Constants.js#L122

packages/discord.js/src/structures/BaseChannel.js
```diff
@@ -155,6 +155,14 @@ class BaseChannel extends Base {
     return 'availableTags' in this;
   }
 
+  /**
+   * Indicates whether this channel is {@link TextBasedChannels text-based} and can have messages sent in it.
+   * @returns {boolean}
+   */
+  isSendable() {
+    return this.isTextBased() && 'send' in this;
+  }
```
Qjuh, line 164: Is the first check here needed? If a channel had a send method it should be considered sendable. Even if for whatever reason it‘s not TextBased (although no channel type that does that currently exists).

packages/discord.js/typings/index.d.ts
```diff
@@ -6878,11 +6880,15 @@ export type Channel =
 
 export type TextBasedChannel = Exclude<Extract<Channel, { type: TextChannelType }>, ForumChannel | MediaChannel>;
 
+export type SendableTextBasedChannels = Exclude<TextBasedChannel, PartialGroupDMChannel>;
```
Qjuh, line 6883: If TextBased check is dropped then this would simply be `SendableChannels`.

Qjuh (changes requested):

packages/discord.js/src/structures/BaseChannel.js
vladfrangu: I'd rather keep it in, if in the future for god-knows what reason we get a channel with a .send that doesn't do messages XD

Qjuh: Then you can simply chain both typeguards as a user… but I‘m good with either way

packages/discord.js/typings/index.test-d.ts
```diff
@@ -2593,3 +2594,16 @@ declare const poll: Poll;
...
+  if (!interaction.channel) {
+    return;
+  }
+
+  // @ts-expect-error
+  interaction.channel.send();
+
+  if (interaction.channel.isSendable()) {
+    expectType<SendableTextBasedChannels>(interaction.channel);
+  }
```
Qjuh, line 2608: To not only check typeguard with type but also actual purpose, maybe add: 
```suggestion
    interaction.channel.send()
  }
```

Qjuh (approved):

packages/discord.js/src/structures/BaseChannel.js
Jiralite: Leaning towards @Qjuh, but I am majorly impartial here.

Jiralite (approved):

didinele (approved):

SpaceEEC (approved):

Qjuh (approved):

PR #10448, opened 2024-08-15
refactor: builders

didinele:
**Please describe the changes this PR makes and why it should be merged:**
Closes #8015

Change list/resolution points:
- [x] remove `@discordjs/formatters` re-export (deprecated currently)
- [x] remove `SelectMenuBuilder` and `SelectMenuOptionBuilder` aliases (deprecated currently, left-over from before we had different types of select menus)
- [x] all builders now store data inside of a `data` field
- [x] `EmbedBuilder` no longer takes in camalCase options
- [x] `EmbedBuilder` now has "nested builders" implemented for things like the author and footer
- [x] removed API-deprecated features such as `.setDMPermission`
- [x] `ActionRowBuilder` now has specialized `[add/set]X` methods as opposed to the current `[add/set]Components`
- [x] validation no longer occurs during the actual "build process", but only during conversion to API data (done strictly with `toJSON()`)
- [x] all builders now use `structuredClone`
- [x] introduced explicit `clearX()` methods as opposed to taking previously inconsistent `null` or `undefined` to clear
- [x] removed inconsistent `equals` methods, if they're needed & desired, we can easily re-implement them (and this time on all classes too)
- [x] moved from `shapeshift` to `zod`, supersedes #10117

---

BREAKING CHANGE: formatters export removed (prev. deprecated)
BREAKING CHANGE: `SelectMenuBuilder` and `SelectMenuOptionBuilder` have been removed (prev. deprecated)
BREAKING CHANGE: `EmbedBuilder` no longer takes camalCase options
BREAKING CHANGE: `ActionRowBuilder` now has specialized `[add/set]X` methods as opposed to the current `[add/set]Components`
BREAKING CHANGE: Removed `equals` methods
BREAKING CHANGE: Sapphire -> zod for validation
BREAKING CHANGE: Removed the ability to pass `null`/`undefined` to clear fields, use `clearX()` instead
BREAKING CHANGE: Renamed all "slash command" symbols to instead use "chat input command"
BREAKING CHANGE: Removed `ContextMenuCommandBuilder` in favor of `MessageCommandBuilder` and `UserCommandBuilder`
BREAKING CHANGE: Removed support for passing the "string key"s of enums
BREAKING CHANGE: Removed `Button` class in favor for specialized classes depending on the style
BREAKING CHANGE: Removed nested `addX` styled-methods in favor of plural `addXs`

PR #10491, opened 2024-09-06
feat: SlashCommandOption#setType and SharedSlashCommandOptions#addOption

Moebits:

**The problem:**

It has been tedious for me to add SlashCommandOption's with every type being it's own class. This usually results in bloated imports that must be changed between commands

```ts
import { SlashCommandStringOption, SlashCommandChannelOption, 
SlashCommandBooleanOption, SlashCommandNumberOption, 
SlashCommandMentionableOption, SlashCommandRoleOption, 
SlashCommandAttachmentOption } from "discord.js" // imports are tedious

const userOption = new SlashCommandMentionableOption()
    .setName("user")
    .setDescription("User")
            
const subcommand = new SlashCommandSubcommandBuilder()
    .setName("subcommand")
    .setDescription("description")
    .addMentionableOption(userOption) // different method for each type
```

To improve ease of use, I propose adding two changes: 
- A `SlashCommandOption` helper class that can be set to any type in `.setType()`
- Addition of `SharedSlashCommandOptions#addOption` that supports adding option of any type

Now with my proposed changes, the code will look like this:
```ts
import { SlashCommandOption } from "discord.js" // cleaner imports

const userOption = new SlashCommandOption()
    .setType("mentionable")
    .setName("user")
    .setDescription("User")
            
const subcommand = new SlashCommandSubcommandBuilder()
    .setName("subcommand")
    .setDescription("description")
    .addOption(userOption) // supports any type
```

`SlashCommandOption#setType` supports passing a string or `ApplicationCommandOptionType`.

This also works: 
```ts
const userOption = new SlashCommandOption()
    .setType(ApplicationCommandOptionType.Mentionable)
```

**No breaking changes**

This PR does not rename or remove any old methods so it doesn't have breaking changes. If people prefer the old method they can continue to use it. 

**Use cases**

I believe this makes it easier for me to create SlashCommandOptions and quickly copy code over to a different command without having to deal with changing imports and methods.

### DIFF
--- a/packages/builders/src/index.ts
+++ b/packages/builders/src/index.ts
@@ -37,6 +37,7 @@ export * from './components/selectMenu/UserSelectMenu.js';
 export * as SlashCommandAssertions from './interactions/slashCommands/Assertions.js';
 export * from './interactions/slashCommands/SlashCommandBuilder.js';
 export * from './interactions/slashCommands/SlashCommandSubcommands.js';
+export * from './interactions/slashCommands/options/all.js';
 export * from './interactions/slashCommands/options/boolean.js';
 export * from './interactions/slashCommands/options/channel.js';
 export * from './interactions/slashCommands/options/integer.js';
--- a/packages/builders/src/interactions/slashCommands/mixins/SharedSlashCommandOptions.ts
+++ b/packages/builders/src/interactions/slashCommands/mixins/SharedSlashCommandOptions.ts
@@ -12,6 +12,46 @@ import { SlashCommandUserOption } from '../options/user.js';
 import type { ApplicationCommandOptionBase } from './ApplicationCommandOptionBase.js';
 
 /**
+ * Represents the slash command option type.
+ */
+export type SlashCommandOptionType =
+	| SlashCommandAttachmentOption
+	| SlashCommandBooleanOption
+	| SlashCommandChannelOption
+	| SlashCommandIntegerOption
+	| SlashCommandMentionableOption
+	| SlashCommandNumberOption
+	| SlashCommandRoleOption
+	| SlashCommandStringOption
+	| SlashCommandUserOption;
+
+/**
+ * Represents the slash command option type constructors.
+ */
+type SlashCommandOptionTypeConstructor =
+	| typeof SlashCommandAttachmentOption
+	| typeof SlashCommandBooleanOption
+	| typeof SlashCommandChannelOption
+	| typeof SlashCommandIntegerOption
+	| typeof SlashCommandMentionableOption
+	| typeof SlashCommandNumberOption
+	| typeof SlashCommandRoleOption
+	| typeof SlashCommandStringOption
+	| typeof SlashCommandUserOption;
+
+const optionMap = new Map<string, SlashCommandOptionTypeConstructor>([
+	['SlashCommandAttachmentOption', SlashCommandAttachmentOption],
+	['SlashCommandBooleanOption', SlashCommandBooleanOption],
+	['SlashCommandChannelOption', SlashCommandChannelOption],
+	['SlashCommandIntegerOption', SlashCommandIntegerOption],
+	['SlashCommandMentionableOption', SlashCommandMentionableOption],
+	['SlashCommandNumberOption', SlashCommandNumberOption],
+	['SlashCommandRoleOption', SlashCommandRoleOption],
+	['SlashCommandStringOption', SlashCommandStringOption],
+	['SlashCommandUserOption', SlashCommandUserOption],
+]);
+
+/**
  * This mixin holds symbols that can be shared in slash command options.
  *
  * @typeParam TypeAfterAddingOptions - The type this class should return after adding an option.
@@ -20,6 +60,33 @@ export class SharedSlashCommandOptions<
 	TypeAfterAddingOptions extends SharedSlashCommandOptions<TypeAfterAddingOptions>,
 > {
 	public readonly options!: ToAPIApplicationCommandOptions[];
+
+	/**
+	 * Adds an option of any type.
+	 *
+	 * @param input - A function that returns an option builder or an already built builder
+	 */
+	public addOption(input: SlashCommandOptionType | ((builder: SlashCommandOptionType) => SlashCommandOptionType)) {
+		if (typeof input === 'function') {
+			for (const OptionConstructor of optionMap.values()) {
+				try {
+					const instance = new OptionConstructor();
+					const result = input(instance);
+					const OptionBuilder = optionMap.get(result.constructor.name);
+					if (OptionBuilder) return this._sharedAddOptionMethod(input, OptionBuilder);
+				} catch {
+					// Ignore errors from passing incorrect arguments to input
+				}
+			}
+
+			throw new Error(`Unsupported option type returned from function input to addOption()`);
+		}
+
+		const OptionBuilder = optionMap.get(input?.constructor.name);
+		if (OptionBuilder) return this._sharedAddOptionMethod(input, OptionBuilder);
+
+		throw new Error(`Unsupported option type passed to addOption(): ${input?.constructor.name}`);
+	}
 
 	/**
 	 * Adds a boolean option.
--- a/packages/builders/src/interactions/slashCommands/options/all.ts
+++ b/packages/builders/src/interactions/slashCommands/options/all.ts
@@ -28,7 +28,7 @@ export type ApplicationCommandOptionStringType =
  *
  * @typeParam OptionType - The type of option
  */
-export type ApplicationCommandOptionEnumTypeMap<OptionType extends ApplicationCommandOptionType | undefined> =
+export type ApplicationCommandOptionEnumTypeMap<OptionType extends ApplicationCommandOptionType> =
 	OptionType extends ApplicationCommandOptionType.Attachment
 		? SlashCommandAttachmentOption
 		: OptionType extends ApplicationCommandOptionType.Boolean
@@ -54,7 +54,7 @@ export type ApplicationCommandOptionEnumTypeMap<OptionType extends ApplicationCo
  *
  * @typeParam OptionType - The type of option
  */
-export type ApplicationCommandOptionStringTypeMap<OptionType extends ApplicationCommandOptionStringType | undefined> =
+export type ApplicationCommandOptionStringTypeMap<OptionType extends ApplicationCommandOptionStringType> =
 	OptionType extends 'attachment'
 		? SlashCommandAttachmentOption
 		: OptionType extends 'boolean'
--- a/packages/builders/src/interactions/slashCommands/options/all.ts
+++ b/packages/builders/src/interactions/slashCommands/options/all.ts
@@ -30,43 +30,26 @@ export type ApplicationCommandOptionStringType =
  */
 export type ApplicationCommandOptionTypeMap<
 	OptionType extends ApplicationCommandOptionStringType | ApplicationCommandOptionType | undefined,
-> = OptionType extends ApplicationCommandOptionType.Attachment
-	? SlashCommandAttachmentOption
-	: OptionType extends ApplicationCommandOptionType.Boolean
-		? SlashCommandBooleanOption
-		: OptionType extends ApplicationCommandOptionType.Channel
-			? SlashCommandChannelOption
-			: OptionType extends ApplicationCommandOptionType.Integer
-				? SlashCommandIntegerOption
-				: OptionType extends ApplicationCommandOptionType.Mentionable
-					? SlashCommandMentionableOption
-					: OptionType extends ApplicationCommandOptionType.Number
-						? SlashCommandNumberOption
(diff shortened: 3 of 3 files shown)


vladfrangu:
You should use the callback syntax if importing the classes is too "tedious"

Moebits:
> You should use the callback syntax if importing the classes is too "tedious"

I think that the callback syntax is too confusing. I like to create all the options separate and attach them to their subcommand.

tipakA:
> I think that the callback syntax is too confusing.

Could you elaborate on why is that? Maybe we could look into addressing this pain point instead.

Here's a slightly cut-down snippet from my current code. There's a single import from `discord.js` related to this entire builder.
```ts
import { SlashCommandBuilder } from 'discord.js';

const deployData = new SlashCommandBuilder()
  .setName(commandName)
  .setDescription('Fix something')
  .addSubcommandGroup(group => { // member
    return group
      .setName('member')
      .setDescription('Fix Member')
      .addSubcommand(sub => { // separator-roles
        return sub
          .setName('separator-roles')
          .setDescription('Fix member\'s separator roles')
          .addUserOption(option => { // member
            return option
              .setName('member')
              .setDescription('Member whose roles to fix')
              .setRequired(true);
          })
          .addBooleanOption(option => { // show
            return option
              .setName('show')
              .setDescription('Should the response be visible?');
          });
      });
  });
  ```
  
Admittedly, I agree that I could switch this to implicit return instead, but using explicit one makes this look better for me. Not like it matters overall though, this code was written once and collapsed both within the blocks (that's why there are name comments on the block starts), as well as with a `// #region SlashCommandBuilder` wrapping it. So far I didn't have to edit it since I wrote it 2 months ago.

---
<details>

<summary>If you're still curious about what else this 537 LOC file imports from discord.js, here:</summary>

```ts
import {
  ButtonInteraction,
  ChannelType,
  ChatInputCommandInteraction,
  Collection,
  EmbedBuilder,
  GuildBasedChannel,
  GuildMember,
  PermissionFlagsBits,
  PermissionsString,
  PrivateThreadChannel,
  PublicThreadChannel,
  RoleMention,
  SlashCommandBuilder,
  Snowflake,
} from 'discord.js';
```

</details>

As far as I'm aware, you would not be cutting anything out of this list regardless of how much you'd change up the Builder. By using JSON object directly instead of `deployData.toJSON()` for the command class, I would save on exactly 1 import. And this isn't even a file with the most imports, other commands of mine also import `ButtonBuilder`s and several types of `SelectMenuBuilder`s.

Moebits:
My issue with that is mainly the same reasons I prefer using async/await over callbacks. It results in a giant "pyramid" that becomes very hard to read the more subcommands/options you add. 

The way I do it by separating the options and attaching them later is much easier to read at a glance for me. It's easier to change/remove individual options without going into the pyramid and trying to figure out which block corresponds to which option. 
```ts
const showOption = SlashCommandOption.createOption("boolean")
    .setName('show')
    .setDescription('Should the response be visible?');

const memberOption = SlashCommandOption.createOption("user")
    .setName('member')
    .setDescription('Member whose roles to fix')
    .setRequired(true)

const sub = new SlashCommandSubcommandBuilder()
    .setName('separator-roles')
    .setDescription('Fix member\'s separator roles')
    .addOption(memberOption)
    .addOption(showOption)

const group = new SlashCommandSubcommandGroupBuilder()
    .setName('member')
    .setDescription('Fix Member')
    .addSubcommand(sub)

const deployData = new SlashCommandBuilder()
    .setName(commandName)
    .setDescription('Fix something')
    .addSubcommandGroup(group)
```

almostSouji:
This seems like an awfully convoluted PR to support a single use case that's based on style preference.
I understand the excitement of wanting to improve handling to what feels more comfortable to use for you, but as a library we have to factor in maintainability, which this seems to negatively affect, since this likely won't be a one-off change if Discord introduces more option types.

Builders exist to ease people into the concept, provide autocomplete and optionally validation.
If you are looking to support advanced use cases and imports are that much of an issue, potentially consider supplying objects directly instead of using builders .

imnaiyar:
> The way I do it by separating the options and attaching them later is much easier to read at a glance for me.

You can still do this somewhat tho.

Consider the following snippet;

```js

const command = new SlashCommandBuilder()
.setName("name")
.setDescription("description");


// add a boolean option
command.addBooleanOption(option => option.setName("bool").setDescription("desc"))

// add string option
command.addStringOption(...)

// or for a subcommand
const sub = new SlashCommandSubcommnadBuilder()
.setName(...)
.setDescription(...);

// add an option to it
sub.addX(...)
sub.addY(...)

// add subcommand to the command
command.addSubcommand(sub);

// similary for subcommand group
```

There's nothing pyramind going here

vladfrangu:
Souji basically said my feelings about this PR in its current form. I also don't understand why exactly you have an extra function to do something that can be done by just exporting an object with all builders in it and using that (not that it'd make my feelings towards this much better)

Ultimately, if this is something others want I will fold, but imo this is one of those abstractions better fit in your own code. Or just split your declarations up like shown in the previous comment by @imnaiyar.

Moebits:
Ok I found the interface I am happy with which is to export a `SlashCommandOption` class with a `setType` method this way it's more consistent with how you create the other parts of a slash command.

This is what usage looks like:
```ts
const questionOption = new SlashCommandOption()
    .setType("string")
    .setName("question")
    .setDescription("question")
    .setRequired(true)
```

I should note that the linter formats the code very weirdly and ruins the spacing on the conditional types. I tried to fix the spacing but it made the test fail so I just left it that way.

If you want me to be honest, I don't really get the point of having so many type classes when the code between them is 90% similar and you could easily combine them into one class. I think there's a lot of unnecessary duplication in the code. However, I got very lukewarm responses to combining them in the discord so this is basically my way of still getting the functionality I wanted without a breaking change.

Yes, I could just implement this into my code. I already did that. But by opening this PR, I hoped that my solution could be helpful to anyone else in the community by being integrated into the library. Am I really the only one that found importing all the option type classes a bit tedious?..

didinele:
> If you want me to be honest, I don't really get the point of having so many type classes when the code between them is 90% similar and you could easily combine them into one class. I think there's a lot of unnecessary duplication in the code. However, I got very lukewarm responses to combining them in the discord so this is basically my way of still getting the functionality I wanted without a breaking change.

It was already explained to you multiple times. On top of the maintenance concerns, you legitimately wouldn't even be able to combine them completely because type-guarding away methods based off of a generic is a nightmare-ish task that probably makes you create separate classes internally, even if to the user it seems like one.

> Yes, I could just implement this into my code. I already did that. But by opening this PR, I hoped that my solution could be helpful to anyone else in the community by being integrated into the library. Am I really the only one that found importing all the option type classes a bit tedious?..

Yes, use a modern code editor that imports for you automatically, or, for the 10th time, use the callback API.

You're not fixing a legitimate problem with this PR, you're introducing more tech debt than builders has seen in its entire lifetime. `SlashCommandOption#setType` is terribly slow and extremely prone to breaking and the types are a complete nightmare and looks exactly like the sort of thing we've regretted every time in the past. 

I, much like @vladfrangu and @almostSouji am against this.

Moebits:
> Yes, use a modern code editor that imports for you automatically, or, for the 10th time, use the callback API.

The library provides two separate methods of adding slash command options: 1) importing the type classes and 2) using the`.addX()` callbacks. My concern is only on improving (1).

My main question is... why do I have to import so many different types? Exporting a simple helper to get any of them is a basic functionality (that should preferably be implemented by the library, not by me).

I don't feel like the issue is so much maintainability, but a clash of style preference. I don't like to use the callback syntax or import so many classes in my code, and this is a suitable solution for me. But it seems like no one else likes my style so they don't want to add this. Claiming that the issue is maintainability when I only added one more file you have to modify just doesn't seem right to me. 

Is it really that difficult to add another slot to the conditional types when (once every blue moon) discord adds a new option type? And eventually they will stop adding more option types, there are only so many that they can add. Interface changes like adding a new method to an option type will not break my code.

didinele:
Builders is opinionated. Wherever we provide multiple ways to do something it's because we think it's an improvement to the developer experience that doesn't compromise the set of patterns we've adopted.

Your proposed changes do not fit the patterns in question at all, so it's not even an issue of "style", but rather consistency. Even if this wasn't a consideration, your DX improvement here is "you don't have to import classes anymore", which we already provide a solution to. "I don't like your solution" is not a good enough reason to add another one, especially with how convoluted it is. I' not even certain why you're using builders, it seems like you would benefit from raw objects a lot more.

As per your comment on maintainability, I've raised multiple points that you haven't addressed & there's more to it than that anyway.

Moebits:
> As per your comment on maintainability, I've raised multiple points that you haven't addressed & there's more to it than that anyway.

If complexity is really your concern then I can export a simple object as someone mentioned before. I prefer the call interface less, but on your end it does remove the need for conditional types.

```ts
export const SlashCommandOption = {
	attachment: SlashCommandAttachmentOption,
	boolean: SlashCommandBooleanOption,
	channel: SlashCommandChannelOption,
	integer: SlashCommandIntegerOption,
	mentionable: SlashCommandMentionableOption,
	number: SlashCommandNumberOption,
	role: SlashCommandRoleOption,
	string: SlashCommandStringOption,
	user: SlashCommandUserOption
}
```

But something tells me that it wouldn't make you happy because maintainability was never the real issue you had. Maybe library authors get defensive after I implied there was something wrong with their design choices.

You mentioned consistency but I looked at the rest of the builders and I just find the consistency to be... kinda off? I expect each of the builders to each follow the same interface designs but it seems like every single builder is doing things their own way.

This is how you add things to an `ActionRowBuilder` or `ModalBuilder`. It is not `addButton()` or `addRow()`, but they accept a type generic `addComponents()` method and you have to set a generic type to define the type of builder.

```ts
const buttonRow = new ActionRowBuilder<ButtonBuilder>()
buttonRow.addComponents(button1, button2, button3)
```

The `StringSelectBuilder` accepts a `addOptions()` that can take an arbitrary number of options. It makes sense for these, but why doesn't `addX()` in SlashCommandOptions accept multiple options?
```ts
const option1  = new StringSelectMenuOptionBuilder()
    .setLabel("label")
    .setValue("value")

const select = new StringSelectMenuBuilder()
    .addOptions(option1)
```

Then there is a weird choice of renaming `MessageEmbed` to `EmbedBuilder` and also putting it here, even though a message embed is not an interaction component like the rest of the builders.

The `AttachmentBuilder` is not even in the builders package, but in core discord.js! And that's not an interaction either.

To me it looks like adding a `SlashCommandOption` export would be one of your least consistency issues.

didinele:
You're now back-peddling and looking for straws to pick. I thought I was having a good faith conversation & went out of my way to tell you why your changes are not in-line with our vision for the package.

> This is how you add things to an ActionRowBuilder or ModalBuilder. It is not addButton() or addRow(), but they accept a type generic addComponents() method and you have to set a generic type to define the type of builder.

#10448 

> The StringSelectBuilder accepts a addOptions() that can take an arbitrary number of options. It makes sense for these, but why doesn't addX() in SlashCommandOptions accept multiple options?

Because that'd make for a bad API that encourages more complex code updates when, say, changing the type of an option. When I cite "consistency" this doesn't mean everything is going to be the exact same without regard for other factors. For example, if we were to introduce more builders, one which takes a bunch of "options" of the same type (e.g. a bunch of strings), and one which is similar to the slash command options, then the former would be consistent with the string select menu builder, while the latter would be consistent with the slash command options.

> Then there is a weird choice of renaming MessageEmbed to EmbedBuilder and also putting it here, even though a message embed is not an interaction component like the rest of the builders.

Quote from the package description:
> @discordjs/builders is a utility package for easily building Discord API payloads.

Not sure where else you'd want this to be? Of course all the other builders are related to interactions, because the API doesn't really have any other constructs that mandate a builder... The rename is not at all a "weird choice".

> The AttachmentBuilder is not even in the builders package, but in the discord.js package! And that's not an interaction either.

Yes, it's not really a builder, it's poorly named, and quite frankly, there are issues with it, we're already aware of this and addressing it after #10448 

---

Multiple maintainers have come out of their way to provide you with solutions and explanations for why we're not happy with this pull request. If you'd like to take this free time that we spent here (as opposed to simply closing the PR with a single sentence on the topic) and going out of your way to make a personal attack like

> Maybe library authors get defensive after I implied there was something wrong with their design choices.

and further implying that this is simply some sort of egoistical behavior on our part is honestly deeply inappropriate and bordering a violation of our code of conduct.

Your arguments have left me unconvinced about these changes and I believe at this point we'll just be going in circles if we continue, so I'll be resigning from this conversation. If other maintainers wish to go further with this, great.

almostSouji:
Thanks for the suggestion. As various people have iterated on above we do not think this is feasible to introduce on the library level and are not going forward with a merge.
If you find it worth the effort, consider wrapping this in userland or consider forking the library to make it fit your specific use case.

Moebits:
I expressed my dissatisfaction using all the import classes and the callback interface, but if you don't want to add my solution there isn't much to be helped. I will implement it into my own code.

PR #10502, opened 2024-09-13
feat(PartialGroupDMChannel): add missing properties

imnaiyar:
Props Added:
- ownerId
- lastMessageId
- lastPinTimestamp
- lastMessage
- lastPinAt
- awaitMessageComponent
- createMessageComponentCollector

~~I noticed `ownerId`, `last_message_id` and `last_pin_timestamp` was missing from `PartialGroupDMChannel`. This PR fixes that.~~

~~While looking at the api documentation alone, it is hard to determine for me if the `owner_id` can be missing for Group DM Channel's, but from my testing so far, I found it to be always present. So taking that at face value~~


imnaiyar:
Yeah it's optional because it can be missing for other channel types, my thinking was it would always be present for GroupDMs, but now that you mention it, looking at other djs structures (ThreadChannel) it accounts for that.so I think I should stick to the library behavior

imnaiyar:
I noticed `last_message_id` and `last_pin_timestamp` is also missing, i wondering if it was left out for a reason or I should add those too

vladfrangu:
> I noticed `last_message_id` and `last_pin_timestamp` is also missing, I am wondering if it was left out for a reason or I should add those too

If they are returned, feel free to add them in your PR and update the title/description

imnaiyar:
Added, `awaitMessageComponent` & `createMessageComponentCollector`, noticed them missing, but there was no issue on creating collector on group dm for components.
image

vladfrangu:
> Added, `awaitMessageComponent` & `createMessageComponentCollector`, noticed them missing, but there was no issue on creating collector on group dm for components. I see no reasons to exclude it. image

Do these actually work 👀?

imnaiyar:
> Do these actually work 👀?

Yes, I've tested it. The channel object returned in the payload contains the group dm's id, which is the most you would need for a channel component collector. At least I have not encountered any edge case where it wouldn't work. As long as you have `Channel` partial, it should work

packages/discord.js/src/structures/PartialGroupDMChannel.js
```diff
@@ -44,6 +46,36 @@ class PartialGroupDMChannel extends BaseChannel {
...
+    } else {
+      this.lastMessageId ??= null;
+    }
+
+    if ('last_pin_timestamp' in data) {
+      /**
+       * The timestamp when the last pinned message was pinned, if there was one
+       * @type {?number}
+       */
+      this.lastPinTimestamp = Date.parse(data.last_pin_timestamp);
```
vladfrangu, line 75: This will return `NaN` for null last_pin_timestamp

```suggestion
      this.lastPinTimestamp = data.last_pin_timestamp ? Date.parse(data.last_pin_timestamp) : null;
```

vladfrangu (dismissed):

vladfrangu (approved):

SpaceEEC (approved):

didinele (approved):

PR #10503, opened 2024-09-13
chore: fix typos

HigherOrderLogic:
Fix typos in various places.


Qjuh:
Thank you for the suggestion. But since this is a fork from https://github.com/microsoft/rushstack/tree/main/libraries/api-extractor-model you might want to suggest those typo fixes upstream instead.

PR #10504, opened 2024-09-14
feat: install button with dropdown for npm, pnpm, yarn and bun

mathisdev7:
This PR modifies a documentation component to enhance the install button functionality. The dropdown now includes `npm`, `pnpm`, `yarn`, and `bun`, allowing users to easily copy the installation command for their preferred package manager. This update improves the flexibility and user experience for developers consulting the documentation.

### DIFF
--- a/apps/website/src/components/ui/InstallButton.tsx
+++ b/apps/website/src/components/ui/InstallButton.tsx
@@ -1,15 +1,30 @@
 'use client';
 
-import { ChevronDown, Copy, CopyCheck } from 'lucide-react';
+import { Copy, CopyCheck } from 'lucide-react';
 import { useEffect, useState } from 'react';
 import { useCopyToClipboard } from 'usehooks-ts';
+
+const getCommand = (manager: string) => {
+	switch (manager) {
+		case 'pnpm':
+			return 'pnpm install discord.js';
+		case 'yarn':
+			return 'yarn add discord.js';
+		case 'bun':
+			return 'bun add discord.js';
+		default:
+			return 'npm install discord.js';
+	}
+};
 
 export function InstallButton({ className = '' }: { readonly className?: string }) {
 	const [interacted, setInteracted] = useState(false);
 	const [copiedText, copyToClipboard] = useCopyToClipboard();
 	const [showDropdown, setShowDropdown] = useState(false);
+	const [selectedPackageManager, setSelectedPackageManager] = useState('npm');
 
-	const handleCopy = async (command: string) => {
+	const handleCopy = async (manager: string) => {
+		const command = getCommand(manager);
 		setInteracted(true);
 		await copyToClipboard(command);
 		setShowDropdown(false);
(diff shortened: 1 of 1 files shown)


apps/website/src/components/ui/InstallButton.tsx
```diff
@@ -1,33 +1,83 @@
...
-			type="button"
-		>
-			<span className="font-semibold text-blurple">{'>'}</span> npm install discord.js{' '}
+		<div className={`relative inline-block ${className}`}>
+			<button
+				className="cursor-pointer rounded-md border border-neutral-300 bg-white px-4 py-2 font-mono hover:bg-neutral-200 dark:border-neutral-700 dark:bg-transparent dark:hover:bg-neutral-800"
+				onClick={() => setShowDropdown(!showDropdown)}
+				type="button"
+			>
+				<span className="font-semibold text-blurple">{'>'}</span> npm install discord.js
```
vladfrangu, line 30: This makes it look really weird, `npm` is always shown but dropdown select makes it copy a diff package manager

mathisdev7:
Reviewed the PR, can you check if it is good please, thanks.

apps/website/src/components/ui/InstallButton.tsx
```diff
@@ -4,30 +4,110 @@ import { Copy, CopyCheck } from 'lucide-react';
...
+		case 'pnpm':
+			return 'pnpm install discord.js';
+		case 'yarn':
+			return 'yarn add discord.js';
+		case 'bun':
+			return 'bun add discord.js';
+		default:
+			return 'npm install discord.js';
+	}
+};
```
almeidx, line 18: You could just make an array of `{ packageManager: string; command: string }` or something like that, and then loop over that to generate the list items

apps/website/src/components/ui/InstallButton.tsx
```diff
@@ -4,30 +4,110 @@ import { Copy, CopyCheck } from 'lucide-react';
...
+				{copiedText && interacted ? (
+					<CopyCheck aria-hidden size={20} className="ml-2 inline-block text-green-500" />
+				) : (
+					<Copy aria-hidden size={20} className="ml-2 inline-block" />
+				)}
+			</button>
+
+			{showDropdown && (
+				<div className="absolute left-0 mt-2 w-full rounded-md border border-neutral-300 bg-white shadow-lg dark:border-neutral-700 dark:bg-neutral-800">
+					<ul className="py-1">
```
almeidx, line 55: Maybe use [Menu from react-aria-components](https://react-spectrum.adobe.com/react-aria/Menu.html) instead

almeidx (changes requested):
From a usability standpoint, doesn't make much sense that for npm, I have to open the select and then select npm (which was already selected) for it to copy the command. Perhaps we should rethink the way this works.

We could keep the button as it was before to copy the command, and have a menu to the side which selects the package manager?

PR #10505, opened 2024-09-14
fix: `GuildChannel#guildId` not being patched to `undefined`

### LINKED ISSUE #10495: interaction.channel.guildId undefined after ephemeral reply

discord.js

### Issue description

Responding to interactions with an ephemeral message sometimes sets GuildChannel.guildId to undefined.

After updating to v14.16.1 I've noticed that sometimes when responding to an interaction with an ephemeral message, the guildId of the channel in the cache becomes set to undefined. I was able to create a minimum reproducible example below, but the behavior is still inconsistent. The issue can occur on the first interaction of a session or in subsequent ones.

Logging output 1:
image

Logging output 2:
image

From my testing it seems to require both the DirectMessages intent and the Channel partial. The application is installed in the guild and its bot user has view message permissions in the channel.

In a help [thread](https://discord.com/channels/222078108977594368/1282367088605921371) I created, Syjalo identified the issue was in GuildChannel.js:68:
> The issue is in GuildChannel.js:68
> Sometimes the data have guild_id property, but with value of undefined.

I've put the issue as "medium" priority since it broke existing v14.15.3 code, but does have workarounds like using GuildChannel.guild.id instead.

TIA.

### Code sample

```typescript
const { Client, Events, GatewayIntentBits, Partials } = require("discord.js")
const { token } = require("./config.json")

const client = new Client({
    intents: [
        GatewayIntentBits.DirectMessages,
        GatewayIntentBits.Guilds,
    ],
    partials: [
        Partials.Channel,
    ],
})

client.on(Events.InteractionCreate, async interaction => {
    console.log(interaction.channel.guildId) // pass
    await interaction.reply({ content: "content", ephemeral: true })
    console.log(interaction.channel.guildId) // sometimes undefined
    console.log(interaction.client.channels.resolve(interaction.channelId).guildId) // same
})

client.login(token)
```

### Versions

discord.js v14.16.1
node v20.17.0
TypeScript N/A
OS Windows 11 23H2

### Issue priority

Medium (should be fixed soon)

Channel

Guilds, DirectMessages

1725667851-8a74f144a

Qjuh:

Fixes #10495.

`guild_id` only gets passed to `Action#getChannel()` if it exists in the payload as key.

### DIFF
--- a/packages/discord.js/src/client/actions/MessageCreate.js
+++ b/packages/discord.js/src/client/actions/MessageCreate.js
@@ -6,7 +6,11 @@ const Events = require('../../util/Events');
 class MessageCreateAction extends Action {
   handle(data) {
     const client = this.client;
-    const channel = this.getChannel({ id: data.channel_id, guild_id: data.guild_id, author: data.author });
+    const channel = this.getChannel({
+      id: data.channel_id,
+      author: data.author,
+      ...(data.guildId && { guild_id: data.guild_id }),
+    });
     if (channel) {
       if (!channel.isTextBased()) return {};
 
--- a/packages/discord.js/src/client/actions/MessageCreate.js
+++ b/packages/discord.js/src/client/actions/MessageCreate.js
@@ -9,7 +9,7 @@ class MessageCreateAction extends Action {
     const channel = this.getChannel({
       id: data.channel_id,
       author: data.author,
-      ...(data.guildId && { guild_id: data.guild_id }),
+      ...('guild_id' in data && { guild_id: data.guild_id }),
     });
     if (channel) {
       if (!channel.isTextBased()) return {};
--- a/packages/discord.js/src/client/actions/MessageDelete.js
+++ b/packages/discord.js/src/client/actions/MessageDelete.js
@@ -6,7 +6,7 @@ const Events = require('../../util/Events');
 class MessageDeleteAction extends Action {
   handle(data) {
     const client = this.client;
-    const channel = this.getChannel({ id: data.channel_id, guild_id: data.guild_id });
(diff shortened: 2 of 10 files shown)


packages/discord.js/src/client/actions/MessageCreate.js
```diff
@@ -6,7 +6,11 @@ const Events = require('../../util/Events');
 class MessageCreateAction extends Action {
   handle(data) {
     const client = this.client;
-    const channel = this.getChannel({ id: data.channel_id, guild_id: data.guild_id, author: data.author });
+    const channel = this.getChannel({
+      id: data.channel_id,
+      author: data.author,
+      ...(data.guildId && { guild_id: data.guild_id }),
```
vladfrangu, line 12: this will not include it if guildId is null
also, data.guild_id not data.guildId

vladfrangu (changes requested):

packages/discord.js/src/client/actions/MessageCreate.js
Syjalo, line 12: This behavior is documented only for `MESSAGE_CREATE`. I don't think it's needed for other events

Syjalo: Or it should be applied for each property

Qjuh: It's documented for MESSAGE_UPDATE too, for MESSAGE_DELETE implicitly. For the others it's also documented to be optional.

vladfrangu (approved):

didinele (approved):

SpaceEEC (approved):

PR #10499, opened 2024-09-12
feat(InteractionResponses)!: support with_response query parameter

monbrey:
**Please describe the changes this PR makes and why it should be merged:**
Closes #10498

/core done in #10512

Upstream: https://github.com/discord/discord-api-docs/pull/7106

**Status and versioning classification:**
- Code changes have been tested against the Discord API, or there are no code changes
- I know how to update typings and have done so, or typings don't need updating
- This PR changes the library's interface (methods or parameters added)

PR #10512, opened 2024-09-18
feat(interactions): support with_response query parameter in core

monbrey:
#10499 but for /core


vladfrangu (approved):

packages/core/src/api/interactions.ts
```diff
@@ -213,19 +382,56 @@ export class InteractionsAPI {
 	public async createAutocompleteResponse(
 		interactionId: Snowflake,
 		interactionToken: string,
-		callbackData: APICommandAutocompleteInteractionResponseCallbackData,
+		callbackData: APICommandAutocompleteInteractionResponseCallbackData &
+			RESTPostAPIInteractionCallbackQuery & { we_response?: false },
```
kyranet, line 386:
```suggestion
			RESTPostAPIInteractionCallbackQuery & { with_response?: false },
```

kyranet (approved):

kyranet (changes requested):
Wrong button

packages/core/src/api/interactions.ts
```diff
@@ -238,14 +444,29 @@ export class InteractionsAPI {
 	public async createModal(
 		interactionId: Snowflake,
 		interactionToken: string,
-		callbackData: APIModalInteractionResponseCallbackData,
+		body: APIModalInteractionResponseCallbackData & RESTPostAPIInteractionCallbackQuery & { with_response?: false },
```
almeidx, line 447: Either revert the parameter name change or update the other overloads and TSDocs to match. Perhaps use the same name for this parameter in all methods too

almeidx (dismissed):

packages/core/src/api/interactions.ts
```diff
@@ -32,65 +51,166 @@ export class InteractionsAPI {
...
-			files,
-			auth: false,
-			body: {
-				type: InteractionResponseType.ChannelMessageWithSource,
-				data,
-			},
-			signal,
-		});
+		return with_response
+			? this.rest.post(Routes.interactionCallback(interactionId, interactionToken), {
```
kyranet, line 79: If `with_response` isn't a truthy value, no API call will be done. Same for all the other methods 😅

kyranet (changes requested):

packages/core/src/api/interactions.ts
```diff
@@ -32,65 +51,166 @@ export class InteractionsAPI {
...
+
+	/**
+	 * Replies to an interaction
+	 *
+	 * @see {@link https://discord.com/developers/docs/interactions/receiving-and-responding#create-interaction-response}
+	 * @param interactionId - The id of the interaction
+	 * @param interactionToken - The token of the interaction
+	 * @param body - The callback data for replying
+	 * @param options - The options for replying
+	 */
```
Jiralite, line 67: Everywhere there is an overload for a method, you can remove any documentation on its implementation signature. For example, look at https://github.com/discordjs/discord.js/blob/main/packages/formatters/src/formatters.ts. It's because IntelliSense or the documentation will not pick it up any more.

packages/core/src/api/interactions.ts
monbrey: Wow, good catch. Not sure what I was doing here.

kyranet (approved):

Jiralite (approved):

PR #10521, opened 2024-09-28
refactor(GuildAuditLogsEntry)!: add type guard for narrowing

Qjuh:

Adds the `GuildAuditLogsEntry#isAction(action: AuditLogEvent)` typeguard to narrow the entry's property types based on the action type.

Before this it wasn't possible to narrow the target type or similar in `Client#guildAuditLogEntryCreate` event based on the actionType.

Because `GuildAuditLogsEntry` is a class and not a discriminated union this is the only way I could think of to achieve at least some sort of narrowing.

In the process also cleaned up the typings a bit.

BREAKING CHANGE: removes `GuildAuditLogsEntry.Targets.All` which wasn’t used anywhere anyway


packages/discord.js/src/structures/GuildAuditLogsEntry.js
```diff
@@ -14,6 +14,7 @@ const Partials = require('../util/Partials');
 const { flatten } = require('../util/Util');
 
 const Targets = {
+  /** @deprecated this was never used or a possible value to begin with */
```
Jiralite, line 17: If this is to target main, this can just be removed and a separate pull request can be made to backport it into v14.

Otherwise, we can keep this as-is, change the base branch to v14, and open another pull request for main.

vladfrangu (approved):
*technically* we could just make GuildAuditLogsEntry be a union based on the type, but this is probs good enough for now

didinele (approved):

PR #10528, opened 2024-10-01
refactor(PermissionOverwrites)!: cache-independent resolve

### LINKED ISSUE #10450: Type error with permissionOverwrites

discord.js

### Issue description

With an existing guild text channel, an error is thrown when setting the `permissionOverwrites` with the required types. 

Here is an example

```ts
const channel = await client.channels.fetch('some ID that exists as a text channel')

channel.permissionOverwrites
            .set([
              {
                id: client.user.id,
                allow: [
                  'SendMessages',
                ],
              },
            ]);
```
An error will be thrown
```
TypeError [InvalidType]: Supplied parameter is not a User nor a Role.
```

And the typings for this method I am using have `id` as a `Snowflake`
image

When using the `client.user` instead, there is no issue and the permissions are correctly changed.

### Code sample

_No response_

### Versions

- discord.js 14.15.3
- typescript 5.4.5
- node 20.13.1

### Issue priority

Low (slightly annoying)

Not applicable

Not applicable

_No response_

Renegade334:

Resolves #10450.

The linked issue relates to the current cache-dependent behaviour of `PermissionOverwrites.resolve()`, specifically that passing identical parameters to `.resolve()` can conditionally pass or fail validation depending on the state of the cache.

This isn't ideal, and the function would be better placed resolving the target via `.resolveId()` in a cache-independent manner.

This approach doesn't allow for `overwrite.type` to be optional if `overwrite.id` is a Snowflake, since it would no longer attempt to seek the respective member/role object from the cache to identify it as one or the other. There are two options for how to handle this:
1. always enforce `overwrite.type` as mandatory, even if `overwrite.id` is a member object or role object;
2. keep `overwrite.type` as optional if `overwrite.id` is a member object or role object (current behaviour), but enforce it as mandatory if `overwrite.id` is a Snowflake.

This PR implements the latter approach. It validates `overwrite.id` and `overwrite.type` as `GuildMemberResolvable|RoleResolvable` and `OverwriteType|undefined` respectively, then:
- if `overwrite.id` is a Snowflake, it sets the type of the result to `overwrite.type`, throwing an error if it's not defined;
- if `overwrite.id` is a member/role object, it sets the type of the result to the OverwriteType corresponding to that object.
  - There's then the question of what to do if `overwrite.type` is provided, but doesn't match the type of `overwrite.id`. This changeset plumps for throwing a validation error if that occurs, since silently overruling the type provided to the function without warning the user could lead to some hard-to-debug quirks.

The same cache-dependent `type` behaviour is also present on `PermissionOverwriteManager#upsert()`, and the behaviour of `PermissionOverwriteManager#edit()` is also seemingly unnecessarily cache-dependent. Think these should also be updated, but will wait for feedback on the PR so far.

### DIFF
--- a/packages/discord.js/src/structures/PermissionOverwrites.js
+++ b/packages/discord.js/src/structures/PermissionOverwrites.js
@@ -160,7 +160,7 @@ class PermissionOverwrites extends Base {
    * @property {GuildMemberResolvable|RoleResolvable} id Member or role this overwrite is for
    * @property {PermissionResolvable} [allow] The permissions to allow
    * @property {PermissionResolvable} [deny] The permissions to deny
-   * @property {OverwriteType} [type] The type of this OverwriteData
+   * @property {OverwriteType} [type] The type of this OverwriteData (mandatory if `id` is a Snowflake)
    */
 
   /**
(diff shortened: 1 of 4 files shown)


packages/discord.js/src/structures/PermissionOverwrites.js
```diff
@@ -171,21 +171,32 @@ class PermissionOverwrites extends Base {
...
+        type = overwrite.type;
+      } else {
+        throw new DiscordjsTypeError(ErrorCodes.PermissionOverwritesTypeMandatory);
+      }
+    } else {
+      type = overwrite.id instanceof Role ? OverwriteType.Role : OverwriteType.Member;
+      if (overwrite.type !== undefined && type !== overwrite.type) {
+        throw new DiscordjsTypeError(ErrorCodes.PermissionOverwritesTypeMismatch, OverwriteType[type]);
+      }
+    }
```
benjGam, line 196: Hi, this should be refactored.

Great work anyway.

vladfrangu (approved):

Jiralite (approved):

Renegade334:
If this approach is kosher, then just wanted to check the following:

- `PermissionOverwriteManager#create()` and `PermissionOverwriteManager#edit()` have the same cache-dependent erroring if a snowflake is passed, via the logic here: https://github.com/discordjs/discord.js/blob/e2df0e0dbc224cf8915e30b4906b832e36a6840a/packages/discord.js/src/managers/PermissionOverwriteManager.js#L96-L103
  Should this behaviour also be altered to mirror this change?
- `PermissionOverwriteManager#edit()` uses the cache to generate the `existing` permission bitfields, which are used as a base onto which the changes are added: https://github.com/discordjs/discord.js/blob/e2df0e0dbc224cf8915e30b4906b832e36a6840a/packages/discord.js/src/managers/PermissionOverwriteManager.js#L146-L149
  If a snowflake is passed here and there is a cache miss, then the "base" bitfields will end up being empty, and any permissions that aren't explicitly modified will be silently truncated on the target. Does this also need some dependent resolution logic?

didinele (approved):
> May or may not be out of scope

I think both things you mentioned are valid concerns that should probably be looked into, but are def. out of scope for this PR.

Also, this needs a rebase and it can be merged.

PR #10529, opened 2024-10-02
refactor!: remove discord.js builders

didinele:
\>:)

BREAKING CHANGE: Removed all builder extensions within discord.js

Upcoming follow-ups:
- re-think discord.js' jank `AttachmentBuilder` and potentially move it, in some form, to `/builders`
- implement a `MessageBuilder` within builders
- as of this PR, users would get the following diff:
```diff
-   .setEmoji('😏')
+  .setEmoji(resolvePartialEmoji('😏'))
```
we can:
1. stick to that
2. move those functions to `/builders` (still stick to that API, though)
3. along with `2.`, allow those simplified types in `setX()` methods within builders


Syjalo (changes requested):
Bump the version to 15.0.0

packages/discord.js/typings/index.test-d.ts
```diff
@@ -2370,33 +2371,33 @@ expectType<
...
 
-new ButtonBuilder({
+new PrimaryButtonBuilder({
   style: ButtonStyle.Danger,
 })
-  .setEmoji('<a:foo:123>')
-  .setEmoji('<:foo:123>')
-  .setEmoji('foobar:123')
-  .setEmoji('😏')
+  .setEmoji(resolvePartialEmoji('<a:foo:123>'))
```
Qjuh, line 2397: This is merely testing the return type of resolveColor, so should just do that. The builder itself gets tested in /builders.

packages/discord.js/typings/index.test-d.ts
```diff
@@ -2406,7 +2407,7 @@ new ButtonBuilder({
 // @ts-expect-error
 new EmbedBuilder().setColor('abc');
 
-new EmbedBuilder().setColor('#ffffff');
+new EmbedBuilder().setColor(resolveColor('#ffffff'));
```
Qjuh, line 2410: In context of mainlib discord.js this would be a test for the return type of resolveColor, the builder‘s methods get tested in `@discordjs/builders` already.

the failing test above this one is also redundant.

packages/discord.js/src/structures/MessagePayload.js
```diff
@@ -1,9 +1,9 @@
 'use strict';
 
 const { Buffer } = require('node:buffer');
+const { ActionRowBuilder } = require('@discordjs/builders');
```
Qjuh, line 4: This should be removed completely, since we don‘t want to make new builders in resolveBody() anymore and just keep json as is.

packages/discord.js/typings/index.test-d.ts
```diff
@@ -2370,33 +2371,33 @@ expectType<
...
   style: ButtonStyle.Danger,
 })
-  .setEmoji('<a:foo:123>')
-  .setEmoji('<:foo:123>')
-  .setEmoji('foobar:123')
-  .setEmoji('😏')
+  .setEmoji(resolvePartialEmoji('<a:foo:123>'))
+  .setEmoji(resolvePartialEmoji('<:foo:123>'))
+  .setEmoji(resolvePartialEmoji('foobar:123'))
+  .setEmoji(resolvePartialEmoji('😏'))
```
Qjuh, line 2400: If this will be the recommended way to do that in mainlib then the `resolvePartialEmoji` function should either be renamed to something easier and/or at least not be marked as private/internal in https://github.com/discordjs/discord.js/blob/39eadce9cbf28de9e8217056bc6aef41032f2ab9/packages/discord.js/src/util/Util.js#L115 and https://github.com/discordjs/discord.js/blob/39eadce9cbf28de9e8217056bc6aef41032f2ab9/packages/discord.js/typings/index.d.ts#L3406-L3408 anymore.

Qjuh (changes requested):

packages/discord.js/typings/index.test-d.ts
didinele: Yeah for sure, all though for unicode emotes you can also do `{ name: '<emote>' }` very easily (on the topic of recommendations)

packages/discord.js/src/structures/MessagePayload.js
```diff
@@ -132,10 +131,6 @@ class MessagePayload {
       }
     }
 
-    const components = this.options.components?.map(component =>
-      (isJSONEncodable(component) ? component : new ActionRowBuilder(component)).toJSON(),
```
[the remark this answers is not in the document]
didinele: ah okay, I miss-understood initially

packages/discord.js/src/util/Components.js
Jiralite, line 1: There are still imports to this file in ActionRow.js and Message.js.

Jiralite (changes requested):

Jiralite:
It seems there's a circular dependency somewhere:

`import "discord.js";`

```
(node:41103) Warning: Accessing non-existent property 'createComponent' of module exports inside circular dependency
```

didinele:
> It seems there's a circular dependency somewhere:

should be fixed

Jiralite (dismissed):

kyranet (approved):
🧹

packages/discord.js/src/util/Components.js
```diff
@@ -100,56 +35,14 @@ function createComponent(data) {
   }
 }
 
-/**
- * Transforms API data into a component builder
- * @param {APIMessageComponent|ComponentBuilder} data The data to create the component from
- * @returns {ComponentBuilder}
- * @ignore
- */
-function createComponentBuilder(data) {
```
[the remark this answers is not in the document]
Jiralite: Nu-uh.

packages/discord.js/src/util/Components.js
```diff
@@ -1,72 +1,7 @@
 'use strict';
 
-const { ComponentBuilder } = require('@discordjs/builders');
 const { ComponentType } = require('discord-api-types/v10');
 
-/**
```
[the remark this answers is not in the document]
Jiralite: I think `ComponentData` in the typings is not actually attached to anything in JavaScript. That or in JSDoc, it is a different type.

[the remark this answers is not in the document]
didinele: > I think ComponentData in the typings is not actually attached to anything in JavaScript. That or in JSDoc, it is a different type.

The type is unused in the typings as well, good catch.

Jiralite (approved):

packages/discord.js/typings/index.d.ts
```diff
@@ -6346,7 +6230,7 @@ export interface BaseMessageOptions {
     | AttachmentPayload
   )[];
   components?: readonly (
-    | JSONEncodable<APIActionRowComponent<APIMessageActionRowComponent>>
+    | JSONEncodable<APIActionRowComponent<APIActionRowComponentTypes>>
```
vladfrangu, line 6233: This is a mistake, it would now allow modal components too

vladfrangu (dismissed):
After fixing this, it can be merged without needing my approval again 🫡

didinele:
After a multitude of internal discussions and attempts, Vlad's request will ultimately be left unresolved for the time being.

PR #10543, opened 2024-10-08
fix(GuildScheduledEvent): handle null recurrence_rule

Eejit43:

This PR handles a null `recurrence_rule` value in `GuildScheduledEvent`s. If there is no rule, the value appears to always be null rather than not present.


packages/discord.js/src/structures/GuildScheduledEvent.js
```diff
@@ -213,7 +213,7 @@ class GuildScheduledEvent extends Base {
      * @property {GuildScheduledEventRecurrenceRuleWeekday} day The day within the week to recur on
      */
 
-    if ('recurrence_rule' in data) {
+    if ('recurrence_rule' in data && data.recurrence_rule) {
```
almeidx, line 216:
```suggestion
    if (data.recurrence_rule) {
```

almeidx (approved):

Jiralite (changes requested):
When the recurrence rule is removed, Discord will send `null`. This means the proposed condition change will result in `false` where it goes to the `else` clause where nothing happens. The cached scheduled event is not updated and will be incorrect.

A proposed fix would be to leave the condition as it is and instead do this:

```JavaScript
this.recurrenceRule = data.recurrenceRule && {...};
```

When `null` is received, it will be correctly set to `null`.

vladfrangu (approved):

PR #10548, opened 2024-10-10
chore: make semver:major block kodiak

vladfrangu:

This is realistically a proposal and can be closed if we don't do it, but as we tend to want to merge major PRs manually (to ensure the commit has changelog entry correct), I think we should skip adding in an extra label.


Jiralite (approved):
Whilst I believe pull requests should not be approved without all the desires we want, I believe it is sensible anyway to disallow automatically merging something that is not backwards compatible. Just in case. So ultimately, I'm more leaning towards accepting this.

kyranet (approved):

didinele (approved):

iCrawl (approved):

PR #10550, opened 2024-10-11
refactor: Deprecate fetching user flags

Jiralite:
Deprecates the fetching of user flags that were removed with #8755.

### DIFF
--- a/packages/discord.js/src/structures/User.js
+++ b/packages/discord.js/src/structures/User.js
@@ -346,6 +346,8 @@ class User extends Base {
    * Fetches this user's flags.
    * @param {boolean} [force=false] Whether to skip the cache check and request the API
    * @returns {Promise<UserFlagsBitField>}
+   * @deprecated <warn>This method is deprecated and will be removed in the next major version.
+   * Flags may still be retrieved via {@link User#fetch}.</warn>
    */
   fetchFlags(force = false) {
     return this.client.users.fetchFlags(this.id, { force });
--- a/packages/discord.js/src/structures/User.js
+++ b/packages/discord.js/src/structures/User.js
@@ -351,7 +351,7 @@ class User extends Base {
    * Flags may still be retrieved via {@link User#fetch}.</warn>
    */
   fetchFlags(force = false) {
-    emitDeprecationWarningForUserFetchFlags(false);
+    emitDeprecationWarningForUserFetchFlags(this.constructor.name);
     return this.client.users.fetchFlags(this.id, { force });
   }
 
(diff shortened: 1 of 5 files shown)


packages/discord.js/typings/index.d.ts
```diff
@@ -4693,6 +4694,7 @@ export class UserManager extends CachedManager<Snowflake, User, UserResolvable>
   public createDM(user: UserResolvable, options?: BaseFetchOptions): Promise<DMChannel>;
   public deleteDM(user: UserResolvable): Promise<DMChannel>;
   public fetch(user: UserResolvable, options?: BaseFetchOptions): Promise<User>;
+  /** @deprecated This method is deprecated and will be removed in the next major version. Flags may still be retrieved via {@link User.fetch} */
```
almeidx, line 4697:
```suggestion
  /** @deprecated This method is deprecated and will be removed in the next major version. Flags may still be retrieved via {@link UserManager.fetch} */
```

vladfrangu (approved):
It's not really a 1:1 migration so the deprecation message is misleading but it's 🤷‍♂️

packages/discord.js/src/structures/User.js
```diff
@@ -346,6 +346,8 @@ class User extends Base {
    * Fetches this user's flags.
    * @param {boolean} [force=false] Whether to skip the cache check and request the API
    * @returns {Promise<UserFlagsBitField>}
+   * @deprecated <warn>This method is deprecated and will be removed in the next major version.
+   * Flags may still be retrieved via {@link User#fetch}.</warn>
    */
   fetchFlags(force = false) {
```
almeidx, line 352: Runtime deprecate as well?

almeidx (changes requested):

packages/discord.js/src/structures/User.js
Jiralite: That would produce 2 deprecation warnings since this calls the method from the user manager where there is already a deprecation warning. I am not immediately sure how to make it "one or the other".

didinele: Probably shared variable re the dep. warning? Just export it from util or something like that.

```js
let xOrYDeprecationWarningEmitted = false;

function emitXorYDeprecationWarning() {
  if (xOrYDeprecationWarningEmitted) return;

  xOrYDeprecationWarningEmitted = true;
  process.emitWarning('bleh');
}
```

Jiralite: Should be done with 925398fb19a1d3415c19add7b7c515fd1636b807.

packages/discord.js/src/util/Util.js
```diff
@@ -499,6 +502,18 @@ function resolveSKUId(resolvable) {
   return null;
 }
 
+/**
+ * Deprecation function for fetching user flags.
+ * @param {boolean} userManager Whether the class name is the user manager
+ * @private
+ */
+function emitDeprecationWarningForUserFetchFlags(userManager) {
```
didinele, line 510: would probably be better to take name as a string param here, but it's fine

didinele (approved):

almeidx (approved):

PR #10552, opened 2024-10-11
perf(collection): optimisations

Renegade334:
#### `.first()`, `.firstKey()`
The callback variant of `Array.from()` is slow in V8. Changing the method of array generation from:
```js
Array.from({ length: amount }, () => ...)
```
to
```js
const results = new Array(amount)
for (let index = 0; index < amount; index++) results[index] = ...
```
produces a ~90-95% reduction in execution time, when tested with all current LTS versions of Node.

Note that this is deliberately `new Array(n)` and not `Array.from({ length: n })`:
- The first of these just creates an empty array with a length property of `n`, and is an O(1) operation.
- The second of these is an O(n) operation: it creates an empty array of length `n`, then for all `i` in `0 <= i < n`, it reads the property `obj[i]` from the object parameter (which in this case will always be undefined), and sets `array[i]` to that value. This is significantly slower for large arrays, and not what we want here.

Additionally, switches to just using iterable-to-array in the case where `amount >= this.size`, which is a fast operation for builtin iterables. This further halves the execution time when compared with the above method.

#### `.last()`, `.lastKey()`
Only carry out the memory-expensive `[...this.<iterable>()]` if needed.

Previously, passing `amount <= 0` would still copy the entire collection into a temporary array, even though it's never referenced. This significantly speeds up execution in those cases.

Uses the new `.at()` or `.keyAt()` (as below) to fetch the last element in the collection in the case where `amount === undefined`. This is around 10-20% faster than the previous method of copying into a temporary array and selecting the last element.

#### `.at()`, `.keyAt()`
Manually iterate to the target index, instead of generating a full array copy of the collection to call `Array.prototype.at()`.

The performance of the previous implementation was essentially linear on the size of the collection, and independent of the index being fetched, as the whole collection was copied into a temporary array regardless.

The performance of the new implementation is linear on the index being fetched, and results in a performance improvement of >90% when the target index is close to the start of the collection. In the worst case when the target index is close to the end of the collection, it still performs well (around 10-20% faster in Node v22, and around 50% faster in Node v18). It also avoids the memory cost of copying the collection into a temporary array.

**This contains a small off-by-one fix which represents a technical breaking change.** Previously, the implementations of `.at()` and `.keyAt()` were not compliant with the standard for `Array.prototype.at()` specifically when passing a negative non-integer index: the standard dictates that these are truncated (ie. rounded _towards_ zero), whereas the previous implementation used `Math.floor()` (ie. rounded these _away from_ zero).

#### `.random()`, `.randomKey()`
The previous implementation used repeated calls to `Array.prototype.splice()` to extract individual random elements from a temporary array copy of the collection. This is a very slow process for larger arrays, since each splice necessitates a memory move of all elements above the extracted element, so performing repeated splices for each random element becomes very inefficient.

The new implementation uses an in-place Durstenfeld shuffle, stopping after enough elements have been randomised, and then returns the shuffled elements using `Array.prototype.slice()`. This is about 40% faster for small collections, and >90% faster for large ones.

Additionally, skips copying the collection into a temporary array if `amount === 0`, and uses the new `.at()` or `.keyAt()` (as above) to fetch a single element in the case where `amount === undefined`.

#### `.map()`
Changes the method of array generation to pushing to a new array, instead of passing a callback to `Array.from()`.

Same rationale as `.first()` above. This change produces a ~80% reduction in execution time when tested with all current LTS versions of Node.

#### `.merge()`
Adjusts the logic for checking `hasInSelf` and `hasInOther`. Previously, each of these boolean conditions was evaluated twice; this makes a small change to the control flow to deduplicate the checks.

#### `.toSorted()`
Removes the redundant closure wrapping `compareFunction` and just passes it directly, eliminating a needless call from the stack.

### Status and versioning classification:

### DIFF
--- a/packages/collection/src/collection.ts
+++ b/packages/collection/src/collection.ts
@@ -154,10 +154,19 @@ export class Collection<Key, Value> extends Map<Key, Value> {
 	 *
 	 * @param index - The index of the element to obtain
 	 */
-	public at(index: number) {
-		index = Math.floor(index);
-		const arr = [...this.values()];
-		return arr.at(index);
+	public at(index: number): Value | undefined {
+		index = Math.trunc(index);
+		if (index >= 0) {
+			if (index >= this.size) return undefined;
+		} else {
+			if (index < this.size * -1) return undefined;
+			index += this.size;
+		}
+		const iter = this.values();
+		for (let skip = 0; skip < index; skip++) {
+			iter.next();
+		}
+		return iter.next().value!;
 	}
 
 	/**
@@ -167,10 +176,19 @@ export class Collection<Key, Value> extends Map<Key, Value> {
 	 *
 	 * @param index - The index of the key to obtain
 	 */
-	public keyAt(index: number) {
-		index = Math.floor(index);
-		const arr = [...this.keys()];
-		return arr.at(index);
+	public keyAt(index: number): Key | undefined {
+		index = Math.trunc(index);
+		if (index >= 0) {
+			if (index >= this.size) return undefined;
+		} else {
+			if (index < this.size * -1) return undefined;
+			index += this.size;
+		}
+		const iter = this.keys();
+		for (let skip = 0; skip < index; skip++) {
+			iter.next();
+		}
+		return iter.next().value!;
 	}
 
 	/**
--- a/packages/collection/src/collection.ts
+++ b/packages/collection/src/collection.ts
@@ -117,11 +117,14 @@ export class Collection<Key, Value> extends Map<Key, Value> {
 	public last(): Value | undefined;
 	public last(amount: number): Value[];
 	public last(amount?: number): Value | Value[] | undefined {
-		const arr = [...this.values()];
-		if (amount === undefined) return arr[arr.length - 1];
-		if (amount < 0) return this.first(amount * -1);
+		if (amount === undefined) {
+			const arr = [...this.values()];
+			return arr[arr.length - 1];
+		}
 		if (!amount) return [];
-		return arr.slice(-amount);
+		if (amount < 0) return this.first(amount * -1);
+		const arr = [...this.values()];
+		return arr.slice(amount * -1);
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


almeidx:
Should we add benchmarks for this? Would be easier to see the performance enhancements.

packages/collection/src/collection.ts
```diff
@@ -148,10 +162,21 @@ export class Collection<Key, Value> extends Map<Key, Value> {
...
-		const arr = [...this.values()];
-		return arr.at(index);
+	public at(index: number): Value | undefined {
+		index = Math.trunc(index);
+		if (index >= 0) {
+			if (index >= this.size) return undefined;
+		} else {
+			if (index < this.size * -1) return undefined;
+			index += this.size;
+		}
```
Syjalo, line 172: Less checks, better readability
```suggestion
		if (index >= this.size) return undefined;
		if (index < 0) {
			index += this.size;
			if (index < 0) return undefined;
		}
```

packages/collection/src/collection.ts
```diff
@@ -161,10 +186,21 @@ export class Collection<Key, Value> extends Map<Key, Value> {
...
-		const arr = [...this.keys()];
-		return arr.at(index);
+	public keyAt(index: number): Key | undefined {
+		index = Math.trunc(index);
+		if (index >= 0) {
+			if (index >= this.size) return undefined;
+		} else {
+			if (index < this.size * -1) return undefined;
+			index += this.size;
+		}
```
Syjalo, line 196: Ditto
```suggestion
		if (index >= this.size) return undefined;
		if (index < 0) {
			index += this.size;
			if (index < 0) return undefined;
		}
```

packages/collection/src/collection.ts
Syjalo: I think making one addition is faster then multiplication + addition

Syjalo: Yeah, it's 14 8% faster for negative indexes

vladfrangu: Addition is always going to be faster than multiplication, indeed

Renegade334: It's a single operation per method call rather than something involved in the iteration, so the cost is negligible either way. Originally written like this so that the behaviour is obvious at a glance, but there is scope to tweak it.

kyranet (approved):

didinele (approved):

ckohen (approved):

PR #10557, opened 2024-10-13
fix(ThreadManager): match parent ID when fetching a single thread

imnaiyar:
Currently as it stands, when fetching a single thread, discord.js simply returns `ChannelManager#fetch`, so ideally, one can fetch any channel of a guild, not just a thread in the channel of the manager, this is misleading as the docs state the return type to be a `ThreadChannel` (or equivalent). This pr adds in a check to match the parent id of the returned channel with the id of manager's channel and throws an error if it doesn't match. It also makes sense to have it this way, as one would expect to get a thread of the channel the manager is of, similar to how `fetchActive` (and others) do.

I don't think it's breaking in any way as it doesn't change the return type or anything, just adds in a way to ensure it returns a thread of the parent channel.


vladfrangu (approved):
Unfortunately sem major :(

didinele (approved):

Jiralite (approved):

PR #10567, opened 2024-10-20
refactor(formatters): Change `:_:` emoji name placeholder

cobaltt7:

Discord doesn't verify emoji names, meaning you only actually need to include a valid ID for an emoji to display properly. Discord.js uses `_` as a fallback name when formatting emojis with just an ID. However, since `_` is not a valid emoji name, if you were to parse that emoji, it would fail, which could be unexpected behavior.
image
image

Additionally, if you were to have multiple emojis with the `_` name placeholder and invalid IDs (or if the bot doesn't have access to the emoji), they fail to render in Discord and the text between them becomes unintentionally italicized. For example, `<:_:123678901234578> foobar <:_:123678901234578>`
image
This is another unintentional side-effect of using `:_:` as a placeholder emoji name that end-users currently have to look out for.

To fix this, I changed the `:_:` placeholder to `:emoji:`. This way it is a valid emoji name that parses correctly and doesn't interfere with any other Markdown features. Originally, I was going to change it to `:__:`, which would parse correctly, but causes text between invalid emojis to be underlined.

In my opinion, `parseEmoji` should also be changed to properly support emojis that used `:_:` in the past, because even though it's not a valid emoji name, it does still render correctly, but people in the support server disagreed
image

### DIFF
--- a/packages/formatters/__tests__/formatters.test.ts
+++ b/packages/formatters/__tests__/formatters.test.ts
@@ -179,15 +179,21 @@ describe('Message formatters', () => {
 		});
 
 		test('GIVEN static emojiId WITH animated explicitly false THEN returns "<:emoji:[emojiId]>"', () => {
-			expect<`<:emoji:851461487498493952>`>(formatEmoji('851461487498493952', false)).toEqual('<:emoji:851461487498493952>');
+			expect<`<:emoji:851461487498493952>`>(formatEmoji('851461487498493952', false)).toEqual(
+				'<:emoji:851461487498493952>',
+			);
 		});
 
 		test('GIVEN animated emojiId THEN returns "<a:emoji:${emojiId}>"', () => {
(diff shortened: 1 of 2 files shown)


Jiralite (commented):
I personally would rather keep this as-is for the current minor version and make `name` required in the future. Thoughts? @vladfrangu @almeidx

vladfrangu:
> I personally would rather keep this as-is for the current minor version and make `name` required in the future. Thoughts? @vladfrangu @almeidx

🤷. We can also just have a fallback on `unknown_emoji`, and recommend users pass in the emoji name (but theres plenty times where you only have an ID but can still react). I _personally_ don't think we need to make it required

vladfrangu (approved):

didinele (approved):

Jiralite (approved):

PR #10421, opened 2024-07-31
chore!: remove all deprecated features/props

didinele:

PR #10579, opened 2024-10-24
fix(ThreadChannel): Address parameter type on `fetchOwner()`

### LINKED ISSUE #10578: Wrong fetch options interface - ThreadChannel#fetchOwner()
### Which application is this bug report for?

Documentation

### Issue description

I opened an [issue](https://github.com/discordjs/docs/issues/13) on the wrong repo, so I'm reposting this here with more info

Original issue:
> https://discord.js.org/docs/packages/discord.js/14.16.3/ThreadChannel:Class#fetchOwner
> 
> As you can see it's
> fetchOwner(options?: [BaseFetchOptions](https://discord.js.org/docs/packages/discord.js/14.16.3/BaseFetchOptions:Interface))
> instead of
> fetchOwner(options?: [FetchThreadMemberOptions](https://discord.js.org/docs/packages/discord.js/14.16.3/FetchThreadMemberOptions:Interface))

Update w/ more info:

`Thread#fetchOwner()` returns a `ThreadMember` but with `guildMember` property `null` unless you do `Thread#fetchOwner({withMember:true})`

Turns out (according to Qjuh) that only `withMember` can be passed in options of `Thread#fetchOwner()`.  
So supposedly `FetchThreadMemberOptions` isn't still really the right one.

_Also there's no indication that you should pass this option if you don't want the member to be null (and need it to use .guildMember after 🙃)_

### Steps to Reproduce

1. Go on https://discord.js.org/docs/packages/discord.js/14.16.3/ThreadChannel:Class#fetchOwner
2. Read :)

### Versions

**Doesn't matter, it's a content related thing**
- Chrome Version 130.0.6723.69
- Updated Win 10 

### Issue priority

Low (slightly annoying)

Jiralite:
Resolves #10578.

In addition, this method cannot return `null`, so the typings and documentation had `null` removed (remnant of #10421).

### DIFF
--- a/packages/discord.js/src/structures/ThreadChannel.js
+++ b/packages/discord.js/src/structures/ThreadChannel.js
@@ -288,10 +288,16 @@ class ThreadChannel extends BaseChannel {
   }
 
   /**
+   * Options used to fetch a thread member owner.
+   * @typedef {BaseFetchOptions} FetchThreadOwnerOptions
+   * @property {boolean} [withMember] Whether to also return the guild member associated with this thread member
+   */
+
+  /**
    * Fetches the owner of this thread. If the thread member object isn't needed,
    * use {@link ThreadChannel#ownerId} instead.
-   * @param {BaseFetchOptions} [options] The options for fetching the member
-   * @returns {Promise<?ThreadMember>}
+   * @param {FetchThreadOwnerOptions} [options] Options for fetching the owner
+   * @returns {Promise<ThreadMember>}
    */
   async fetchOwner(options) {
     if (!this.ownerId) {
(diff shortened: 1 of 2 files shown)


vladfrangu (approved):

didinele (approved):

packages/discord.js/src/structures/ThreadChannel.js
```diff
@@ -287,11 +287,17 @@ class ThreadChannel extends BaseChannel {
     return this.parent?.permissionsFor(memberOrRole, checkAdmin) ?? null;
   }
 
+  /**
+   * Options used to fetch a thread member owner.
```
almeidx, line 291:
```suggestion
   * Options used to fetch a thread owner.
```

almeidx (changes requested):

almeidx (approved):

Issue #10577, opened 2024-10-24
After upgrading @types/node to v22.7.9, an incompatible type error occurs in the “on” method of the client class


book000:
### Which package is this bug report for?

discord.js

### Issue description

After upgrading @types/node v22.7.9, the following error occurs when executing the tsc command.

```
node_modules/.pnpm/discord.js@14.16.3/node_modules/discord.js/typings/index.d.ts(986,14): error TS2417: Class static side 'typeof Client' incorrectly extends base class static side 'typeof BaseClient'.
The types returned by 'on(...)' are incompatible between these types.
   Property '[Symbol.asyncDispose]' is missing in type 'AsyncIterableIterator<any>' but required in type 'AsyncIterator<any[], any, any>'.
```

It could be a problem with @types/node, but this type error appears to be occurring in discord.js, so we will raise it here first.

### Code sample

```typescript
https://github.com/jaoafa/discord-new-threads/actions/runs/11490768941/job/31982043563?pr=1179
```

### Versions

- discord.js v14.16.3
- Node.js v20.18.0
- pnpm v9.12.2
- TypeScript 5.6.3

### Issue priority

Medium (should be fixed soon)

### Which partials do you have configured?

Not applicable

### Which gateway intents are you subscribing to?

Not applicable

Renegade334:
Will currently affect builds using the latest @types/node releases in the following circumstances:

- TS >= 5.6
- esnext.disposable library loaded
- skipLibCheck off

Will look at sorting later.

cobaltt7:
> * esnext.disposable library loaded

I had this same issue, and it was fixed by adding `esnext.disposable` to my `lib`. Looking at the linked repo, it looks like they only have `esnext.asyncdisposable`, which I'm not sure exists.

@book000 Try changing `esnext.asyncdisposable` to `esnext.disposable` in your tsconfig.json `lib`

book000:
@cobaltt7 The following changes have been made, but have not improved the error.

- Pull request: https://github.com/jaoafa/discord-new-threads/pull/1196
- CI log: https://github.com/jaoafa/discord-new-threads/actions/runs/11508694464/job/32037393529?pr=1196

JstnMcBrd:
Just tested on the new `discord.js@14.18.0` release, and the type error is not fixed. Could we reopen this issue?

```json
  "dependencies": {
    "discord.js": "^14.18.0"
  },
  "devDependencies": {
    "@types/node": "^22.13.4"
  },
[...]

PR #10581, opened 2024-10-24
feat(Client): add typed wrapper methods for `EventEmitter.{on,once}`

Renegade334:
So...

Historically, the discord.js typings have overloaded `EventEmitter.on()` and `EventEmitter.once()` with signatures that return typed tuples from ClientEvents, if the target is a client object and the event is in ClientEvents. This allowed users to, for example, do things like `const [ role ] = await EventEmitter.once(client, 'roleCreate')` and have the yielded values be automatically typed without needing any manual coercion.

These overloads had to be removed, due to TypeScript patching out the bug that was being exploited to add them to EventEmitter itself. They were moved downstream to `Client.{on,once}` in #10360 as a temporising measure, but the overload signatures themselves remained the same.

Unfortunately, `EventEmitter.on()` returns an async iterator.

Iterators are in a state of heavy flux within the TS library at the moment, and hence also within @types/node, which recently changed the types of its returned iterators across the entire library. These were formerly `[Async]IterableIterator`s, but are now internal interfaces which extend TypeScript's `[Async]IteratorObject` in TS 5.6+.

Since _those_ interfaces change shape depending on which TypeScript libs are loaded, this has the potential to cause subtle assignability errors with the Client overloads (_eg._ #10577) that are conditional on both TS and @types/node versions and also on which libs are loaded, and are therefore not picked up within the library's build tests. Even if the Client overloads are fixed now, they're going to be vulnerable to further breakage in the future if those interfaces continue to evolve.

However, there is a different approach entirely: **declare new client _instance_ methods, which wrap the EventEmitter static methods.**

Instead of playing with EventEmitter definitions, wrapper methods are declared on the BaseClient prototype (naming up for discussion):
- `.eventIterator(eventName[, options])` as a wrapper for `EventEmitter.on(client, eventName[, options])`
- `.awaitEvent(eventName[, options])` as a wrapper for `EventEmitter.once(client, eventName[, options])`

This has a number of advantages:
- Offers users a stable API, any changes to which can respect discord.js semantic versioning.
- Offers a simpler call signature: `client.awaitEvent(event)` rather than `EventEmitter.once(client, event)`.
- Makes the type definitions independent of users' TypeScript and @types/node versions (since they only depend on `Promise` and `AsyncIterableIterator`).
- Obviates the need to mess with the @types/node library definitions altogether, which removes the potential for further breakage in the future.

Removing the client-specific `Client.{on,once}` overload definitions from the .d.ts won't break any existing builds, since the only potential change from reverting to the native @types/node signatures is to _widen_ the types of the yielded/resolved arrays (from typed tuples to `any[]`).

Resolves: #10577

### DIFF
--- a/packages/discord.js/src/client/BaseClient.js
+++ b/packages/discord.js/src/client/BaseClient.js
@@ -103,6 +103,26 @@ class BaseClient extends EventEmitter {
     }
   }
 
+  /**
+   * Shorthand for `EventEmitter.on(client, ...)`
+   * @param {string} eventName The name of the client event to listen for
+   * @param {Object} [options] Passed as the `options` argument to `EventEmitter.on()`
+   * @returns {AsyncIterator} An async iterator that yields on each emitted event
+   */
+  eventIterator(eventName, options = {}) {
+    return super.constructor.on(this, eventName, options);
+  }
+
+  /**
+   * Shorthand for `EventEmitter.once(client, ...)`
+   * @param {string} eventName The name of the client event to listen for
+   * @param {Object} [options] Passed as the `options` argument to `EventEmitter.once()`
+   * @returns {Promise<Array<*>>} A promise that resolves to the next emitted event
+   */
+  awaitEvent(eventName, options = {}) {
+    return super.constructor.once(this, eventName, options);
+  }
+
   toJSON(...props) {
     return flatten(this, ...props);
   }
--- a/packages/discord.js/typings/index.d.ts
+++ b/packages/discord.js/typings/index.d.ts
@@ -168,7 +168,7 @@ import {
   GatewayDispatchPayload,
 } from 'discord-api-types/v10';
 import { ChildProcess } from 'node:child_process';
-import { EventEmitter } from 'node:events';
+import { Abortable, EventEmitter } from 'node:events';
 import { Stream } from 'node:stream';
 import { MessagePort, Worker } from 'node:worker_threads';
 import {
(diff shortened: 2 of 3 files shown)


almeidx:
Instead of adding yet another workaround, why don't we teach users that they can just use this instead?

```ts
const [role] = await Client.once(client, 'roleCreate');
```

[ts playground](<https://www.typescriptlang.org/play/?#code/JYWwDg9gTgLgBAbzgYQDbAKYDt4F84BmUEIcARACbADOAxtBQHQBW1ZAUOxRragIZQMcelmrxemHAC4U6bDADcnEWLgBtOMVRCAunAC8cPgHc+weGkkxGELLQwAKCfIA0cAORaMyQXxgZ3AEolAHoQuAjIgD0Afk4CaCNTczgnW1U1Lz0IAlkrGywnORw3TwhtHww-AMDAxHYIsMjmlojY9lx2IA>)

Renegade334:
> Instead of adding yet another workaround, why can't users just use this instead?
> 
> ```ts
> const [role] = await Client.once(client, 'roleCreate');
> ```
> 
> [ts playground](https://www.typescriptlang.org/play/?#code/JYWwDg9gTgLgBAbzgYQDbAKYDt4F84BmUEIcARACbADOAxtBQHQBW1ZAUOxRragIZQMcelmrxemHAC4U6bDADcnEWLgBtOMVRCAunAC8cPgHc+weGkkxGELLQwAKCfIA0cAORaMyQXxgZ3AEolAHoQuAjIgD0Afk4CaCNTczgnW1U1Lz0IAlkrGywnORw3TwhtHww-AMDAxHYIsMjmlojY9lx2IA)

This definition is still fine for downstream users, _providing_ they use TS <5.6, or avoid the esnext libs, or have `skipLibCheck` on. Those that don't are still receiving compilation errors (#10577).

It's worth noting that this issue is also blocking d.js from upgrading its TypeScript dependency from 5.5, as the .d.ts tests fail with a compilation error in newer TS versions, due to `AsyncIterableIterator` incorrectly extending TS 5.6's  `AsyncIteratorObject`.

There isn't a particularly elegant solution to this that works for all build environments. (This proposal would work with all build environments, regardless of TS and @types/node versions, but is certainly not elegant.)

The alternatives would be to use TS 5.6's `AsyncIteratorObject` or @types/node's `NodeJS.AsyncIterator` types in the definitions, but this would require downstream users to ensure they meet those dependencies for their own builds (although could be appropriate for v15?)

PR #10583, opened 2024-10-26
types(Message): correctly mark `system` as nullable

sdanialraza:
`Message#system` was marked nullable in JSDoc and implementation but not in typings, this PR fixes that


Jiralite:
The reason for this is because of partials. It may be `null` in a partial message, but it will not be on a full structure.

PR #10584, opened 2024-10-31
types: bump `@types/node` version, use EventEmitter generic

Qjuh:

Bumps `@types/node` to latest version for the respective node version used in each package but `@discordjs/rest`, because there's a clash on the types for global `fetch()` function that needs to be resolved.

Adapts mainlib `discord.js` typings to use the EventEmitter generic instead of manual overloads for EventEmitter methods.

Supersedes #10581.

### DIFF
--- a/apps/guide/package.json
+++ b/apps/guide/package.json
@@ -69,7 +69,7 @@
 		"@testing-library/react": "^15.0.7",
 		"@testing-library/user-event": "^14.5.2",
 		"@types/html-escaper": "^3.0.2",
-		"@types/node": "^18.19.45",
+		"@types/node": "^18.19.62",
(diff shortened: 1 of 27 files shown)


iCrawl (approved):

Renegade334 (commented):
Suggest this should be semver:major, as it will cause new compile errors for downstream projects (in the absence of skipLibCheck) if their installed @types/node version dates from prior to EventEmitter taking a type parameter.

The @types/node patch that allowed for event-mapped returns from top-level `node:events` functions has had to be reverted, so whether to provide a typed solution for `events.on()` and `events.once()` probably still needs to be considered.

Qjuh:
Superseded by #10710

PR #10590, opened 2024-11-05
feat: add soundboard

sdanialraza:
**Please describe the changes this PR makes and why it should be merged:**
Adds soundboard for mainlib

**Status and versioning classification:**
- Code changes have been tested against the Discord API, or there are no code changes
- I know how to update typings and have done so, or typings don't need updating
- This PR changes the library's interface (methods or parameters added)

PR #10591, opened 2024-11-05
fix(GuildAuditLogEntry)!: Fix some incorrect types and runtime logic

cobaltt7:

Fixes some audit log types and runtime logic.
 - The target of `MessageBulkDelete` is a channel, not a guild
   - Runtime logic already resolved it as a channel
   - Added a typings test
   - BREAKING CHANGE: It also doesn't have a `options.channel_id`, so I stopped `.extra.channel` from being set to `{ id: undefined }`
 - AutoModeration trigger logs have the violating `User` as a target, not the `AutoModerationRule`
   - BREAKING CHANGE: Fixed both types and runtime logic here, it previously created a broken `AutoModerationRule`
 - `GuildOnboardingCreate` and `GuildOnboardingUpdate` do not have any target
   - BREAKING CHANGE: Removed `Targets.GuildOnboarding`, it will fallback to `Targets.Unknown` and generate a placeholder `target` from the `changes`
   - Added a note to the types mentioning them and other events that do not have targets
 - Handle `SoundboardSound` as much as possible right now, add placeholders for after #10590
 - Properly resolve the target type of `AutoModerationRule` rules
 - Include `PartialUser` in the types where a partial user is possible
 - Properly type the values of the `GuildAuditLogsEntry.Target` static object
   - Previously was typed as:
      ```js
      {
        Guild: "Guild" | "Channel" | "User" | ...;
        Channel: "Guild" | "Channel" | "User" | ...;
        User: "Guild" | "Channel" | "User" | ...;
        ...
      }
      ```
      Instead of:
      ```js
      {
        Guild: "Guild";
        Channel: "Channel";
        User: "User";
        ...
      }
      ```
 - Hardcode `Role` and `Emoji` targets at runtime
   - They were already hardcoded in the types, although they were missing the `{ id: Snowflake }` alternative for if they were uncached

There are some breaking changes, but they all are minor things that never worked properly.


Jiralite:
Please resolve conflicts!

cobaltt7:
Resolved conflicts and removed the overlapping changes. Sorry, I hadn't realized that PR existed!

SpaceEEC (approved):

vladfrangu (approved):

didinele (approved):

packages/discord.js/src/structures/GuildAuditLogsEntry.js
```diff
@@ -391,9 +397,10 @@ class GuildAuditLogsEntry {
     if (target < 110) return Targets.GuildScheduledEvent;
     if (target < 120) return Targets.Thread;
     if (target < 130) return Targets.ApplicationCommand;
-    if (target >= 140 && target < 150) return Targets.AutoModeration;
+    if (target < 140) return Targets.SoundboardSound;
+    if (target >= 140 && target <= 142) return Targets.AutoModeration;
+    if (target >= 143 && target <= 145) return Targets.User;
```
almeidx, line 402: We already know the first conditions are true at this point.
Also updated to use `<` to more or less keep the pattern that was being followed on this method.

```suggestion
    if (target < 143) return Targets.AutoModeration;
    if (target < 146) return Targets.User;
```

almeidx (changes requested):

packages/discord.js/src/structures/GuildAuditLogsEntry.js
cobaltt7: Thanks @almeidx, fixed

almeidx (approved):
🤠

PR #10595, opened 2024-11-12
docs(channel): Clarify emoji parameter

Jiralite:
For these methods, the emoji must be encoded which is done via `encodeURIComponent()`. However, there is no indication that discord.js handles this internally without checking the source code. Therefore, it is plausible to assume that one must encode their emoji before passing it into the method, resulting in a double encoding. This pull request clarifies the internal behaviour and provides examples.

### DIFF
--- a/packages/core/src/api/channel.ts
+++ b/packages/core/src/api/channel.ts
@@ -92,7 +92,7 @@ export class ChannelsAPI {
 	 * @see {@link https://discord.com/developers/docs/resources/message#get-reactions}
 	 * @param channelId - The id of the channel the message is in
 	 * @param messageId - The id of the message to get the reactions for
-	 * @param emoji - The emoji to get the reactions for. Encoding happens internally
+	 * @param emoji - The emoji to get the reactions for. URL encoding happens internally
 	 * @param query - The query options for fetching the reactions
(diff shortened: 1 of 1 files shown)


packages/core/src/api/channel.ts
```diff
@@ -92,9 +92,17 @@ export class ChannelsAPI {
 	 * @see {@link https://discord.com/developers/docs/resources/message#get-reactions}
 	 * @param channelId - The id of the channel the message is in
 	 * @param messageId - The id of the message to get the reactions for
-	 * @param emoji - The emoji to get the reactions for
+	 * @param emoji - The emoji to get the reactions for. Encoding happens internally
```
vladfrangu, line 95: I'd make it clearer, what kind of encoding?
```suggestion
	 * @param emoji - The emoji to get the reactions for. URL encoding happens internally
```

vladfrangu (approved):

packages/core/src/api/channel.ts
Jiralite: Resolved with a3fc5bf40befa9a65efa3ccb7925c5b678b796a5.

didinele (approved):

almeidx (approved):

PR #10597, opened 2024-11-16
fix(MessageReaction): Address `undefined` burst properties

Jiralite:

When receiving a non-burst reaction over the gateway, we get this (example) payload:
```JavaScript
{
  user_id: '618976181026422814',
  type: 0,
  message_id: '1307361477182754877',
  message_author_id: '618976181026422814',
  member: {
    user: {
      username: 'jiralite',
      public_flags: 4210944,
      id: '618976181026422814',
      global_name: 'Jiralite',
      display_name: 'Jiralite',
      discriminator: '0',
      clan: null,
      bot: false,
      avatar_decoration_data: null,
      avatar: '096240dfad1c926c33f0a92b111aaaeb'
    },
    roles: [],
    premium_since: null,
    pending: false,
    nick: null,
    mute: false,
    joined_at: '2024-06-08T17:57:51.599000+00:00',
    flags: 0,
    deaf: false,
    communication_disabled_until: null,
    banner: null,
    avatar: null
  },
  emoji: { name: 'SkyLook', id: '649013670625542155' },
  channel_id: '1089254519268974712',
  burst: false,
  guild_id: '1089217648996401242'
}
```

This data is shoved into the cache here:
<https://github.com/discordjs/discord.js/blob/c45d912c98e3373d1f78ce4e10380e704eee1b45/packages/discord.js/src/client/actions/MessageReactionAdd.js#L42-L47>

`burst_colors` is not here, but we make sure the property exists. This means here, it's set to `undefined`:
<https://github.com/discordjs/discord.js/blob/c45d912c98e3373d1f78ce4e10380e704eee1b45/packages/discord.js/src/structures/MessageReaction.js#L55-L61>

`me_burst` is also not here, which means this is `undefined` too:
<https://github.com/discordjs/discord.js/blob/c45d912c98e3373d1f78ce4e10380e704eee1b45/packages/discord.js/src/structures/MessageReaction.js#L39>

Interestingly, there does not seem to be a way for bots to burst react, so doing `Boolean(data.me_burst)` should be enough. For `burst_colors`... since that is always present, I've added an array check. If it's not an array, then it's likely missing and should be left alone.

Tested also with no partials and `[Partials.Channel, Partials.GuildMember, Partials.GuildScheduledEvent, Partials.Message, Partials.Reaction, Partials.ThreadMember, Partials.User]`.


vladfrangu (approved):

packages/discord.js/src/structures/MessageReaction.js
```diff
@@ -52,7 +52,7 @@ class MessageReaction {
   }
 
   _patch(data) {
-    if ('burst_colors' in data) {
+    if ('burst_colors' in data && Array.isArray(data.burst_colors)) {
```
almeidx, line 55:
```suggestion
    if (data.burst_colors) {
```

almeidx (approved):

PR #10608, opened 2024-11-19
feat(Client): add request soundboard sounds

sdanialraza:
Adds iterator and non-iterator variants for requesting soundboard sounds for core


packages/core/src/client.ts
```diff
@@ -270,11 +272,9 @@ export class Client extends AsyncEventEmitter<MappedEvents> {
...
 				};
 
-				if (data.chunk_index >= data.chunk_count - 1) {
-					break;
-				} else {
-					timer = createTimer();
-				}
+				if (data.chunk_index >= data.chunk_count - 1) break;
+
+				timer = createTimer(controller, timeout);
```
vladfrangu, line 277: Oof, this shows a bittt of an issue with our current code, which is that we keep creating timers and _not_ clearing out previous ones... We should probs solve that too

packages/core/src/client.ts
```diff
@@ -316,6 +316,103 @@ export class Client extends AsyncEventEmitter<MappedEvents> {
...
+			for await (const [{ data }] of iterator) {
+				clearTimeout(timer);
+				timer = undefined;
+
+				yield {
+					guildId: data.guild_id,
+					soundboardSounds: data.soundboard_sounds,
+				};
+
+				guildIds.delete(data.guild_id);
```
vladfrangu, line 370: Shouldn't we yield only if the guild id is in the list of guild ids the user provided?

packages/core/src/client.ts
```diff
@@ -316,6 +316,103 @@ export class Client extends AsyncEventEmitter<MappedEvents> {
...
+
+				if (guildIds.size === 0) break;
+
+				timer = createTimer(controller, timeout);
```
vladfrangu, line 374: Same issue as in members iterator, needs to clear previous timeout

vladfrangu (changes requested):

packages/core/src/client.ts
sdanialraza: When would the guild id not be in the list of guild ids the user provided though?

vladfrangu: Multiple requests?

vladfrangu (approved):

packages/util/src/functions/groupBy.ts
```diff
@@ -0,0 +1,40 @@
...
+ * 	{ name: 'Alice', age: 20 },
+ * 	{ name: 'Bob', age: 20 },
+ * 	{ name: 'Charlie', age: 30 },
+ * ];
+ *
+ * groupBy(items, item => item.age);
+ * // Map { 20 => [ { name: 'Alice', age: 20 }, { name: 'Bob', age: 20 } ], 30 => [ { name: 'Charlie', age: 30 } ] }
+ * ```
+ */
+export function groupBy<Key, Value>(
```
didinele, line 20: is there... a reason for this? MDN says it's in node 21+

didinele (changes requested):

packages/util/src/functions/groupBy.ts
Jiralite: Our minimum required Node.js version is 20. Unless we do 22, this polyfill is needed.

didinele: right. well... if we're doing v15 we.. could?

packages/core/src/client.ts
```diff
@@ -316,6 +316,105 @@ export class Client extends AsyncEventEmitter<MappedEvents> {
...
+	 * @param options - The options for the request
+	 * @param timeout - The timeout for waiting for each soundboard sounds
+	 * @example
+	 * Requesting soundboard sounds for specific guilds
+	 * ```ts
+	 * for await (const { soundboardSounds } of this.requestSoundboardSoundsIterator({
+	 *	guild_ids: ['1234567890', '9876543210'],
+	 * 	})) {
+	 *	console.log(soundboardSounds);
+	 *}
```
almeidx, line 332:
```suggestion
	 * for await (const { guildId, soundboardSounds } of this.requestSoundboardSoundsIterator({
	 *	guild_ids: ['1234567890', '9876543210'],
	 * })) {
	 *	console.log(`Soundboard sounds for guild ${guildId}:`, soundboardSounds);
	 * }
```

packages/util/src/functions/groupBy.ts
Jiralite: #10726

didinele: We're going through with the node bump, this function can be removed in favor of native Map.prototype.groupBy.

sdanialraza: Done via [`e990e3f` (#10608)](https://github.com/discordjs/discord.js/pull/10608/commits/e990e3fc169418c2366e59bc5f3b5a9c992abb33)

didinele (approved):

Jiralite (approved):

PR #10613, opened 2024-11-23
refactor(Emoji)!: make imageURL() change extension dynamically

Idris1401:
BREAKING CHANGE: The `REST#emoji` method now has a `animated` required parameter.

I think it'd be better if `Emoji#imageURL()` changes the extension dynamically as others `...URL()` functions.

### DIFF
--- a/packages/discord.js/src/structures/Emoji.js
+++ b/packages/discord.js/src/structures/Emoji.js
@@ -46,7 +46,7 @@ class Emoji extends Base {
    * @returns {?string}
    */
   imageURL(options) {
-    return this.id && this.client.rest.cdn.emoji(this.id, options);
+    return this.id && this.client.rest.cdn.emoji(this.id, this.animated ? { ...options, extension: 'gif' } : options);
   }
 
   /**
--- a/packages/discord.js/src/structures/Emoji.js
+++ b/packages/discord.js/src/structures/Emoji.js
@@ -42,11 +42,11 @@ class Emoji extends Base {
 
   /**
    * Returns a URL for the emoji or `null` if this is not a custom emoji.
-   * @param {BaseImageURLOptions} [options] Options for the image URL
+   * @param {ImageURLOptions} [options={}] Options for the image URL
    * @returns {?string}
    */
-  imageURL(options) {
-    return this.id && this.client.rest.cdn.emoji(this.id, this.animated ? { extension: 'gif', ...options } : options);
+  imageURL(options = {}) {
+    return this.id && this.client.rest.cdn.emoji(this.id, this.animated, options);
   }
 
   /**
--- a/packages/discord.js/src/structures/Emoji.js
+++ b/packages/discord.js/src/structures/Emoji.js
@@ -46,7 +46,7 @@ class Emoji extends Base {
    * @returns {?string}
    */
   imageURL(options) {
-    return this.id && this.client.rest.cdn.emoji(this.id, this.animated ? { ...options, extension: 'gif' } : options);
+    return this.id && this.client.rest.cdn.emoji(this.id, this.animated ? { extension: 'gif', ...options } : options);
   }
 
   /**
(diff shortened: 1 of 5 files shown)


packages/discord.js/src/structures/Emoji.js
```diff
@@ -46,7 +46,7 @@ class Emoji extends Base {
    * @returns {?string}
    */
   imageURL(options) {
-    return this.id && this.client.rest.cdn.emoji(this.id, options);
+    return this.id && this.client.rest.cdn.emoji(this.id, this.animated ? { ...options, extension: 'gif' } : options);
```
kyranet, line 49: This change makes it so when an emoji is animated, passing `extension: 'png'` will still return a GIF.

However, if you swap the order, `...options` will be able to override the `extension` property:
```suggestion
    return this.id && this.client.rest.cdn.emoji(this.id, this.animated ? { extension: 'gif', ...options } : options);
```

kyranet (dismissed):

packages/discord.js/src/structures/Emoji.js
kyranet: I would also document this dynamic default so users know what to expect.

almeidx (approved):

packages/discord.js/src/structures/Emoji.js
```diff
@@ -46,7 +46,7 @@ class Emoji extends Base {
    * @returns {?string}
    */
   imageURL(options) {
-    return this.id && this.client.rest.cdn.emoji(this.id, options);
+    return this.id && this.client.rest.cdn.emoji(this.id, this.animated ? { extension: 'gif', ...options } : options);
```
almeidx, line 49: Actually, while this works, it might be worth adding support for the `forceStatic` option like the other dynamic image URL methods

You'd have to change the parameter type to `ImageURLOptions` and handle the `forceStatic` option here as well

Idris1401: what about adding an `animated` option for the `emoji` function in /rest?

almeidx: That's fine by me

Idris1401:
idk if the docs/typings for `BaseGuildEmoji#imageURL()` should also be updated or no

almeidx (changes requested):

packages/rest/src/lib/CDN.ts
```diff
@@ -163,8 +163,8 @@ export class CDN {
 	 * @param emojiId - The emoji id
 	 * @param options - Optional options for the emoji
 	 */
-	public emoji(emojiId: string, options?: Readonly<BaseImageURLOptions>): string {
-		return this.makeURL(`/emojis/${emojiId}`, options);
+	public emoji(emojiId: string, animated: boolean, options?: Readonly<ImageURLOptions>): string {
```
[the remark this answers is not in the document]
Idris1401: what exactly needs to be changed?

packages/rest/src/lib/CDN.ts
```diff
@@ -161,10 +161,11 @@ export class CDN {
 	 * Generates an emoji's URL.
 	 *
 	 * @param emojiId - The emoji id
+  	 * @param animated - Indicates whether the emoji is animated or not
```
almeidx, line 164:
```suggestion
	 * @param animated - Whether the emoji is animated
```

almeidx (changes requested):

SpaceEEC (approved):

almeidx (approved):

vladfrangu (approved):

Jiralite (changes requested):
Tests need updating! The second parameter of `emoji()` in REST is no longer an object, but it still is in the tests.

Jiralite (approved):

PR #10616, opened 2024-11-25
build!: Bump Node.js to 20

Jiralite:
All packages will now require Node.js 20 or above.


vladfrangu (approved):
Let's bump this across packages tbh, node 18 is EOL

Renegade334:
Should this also bump the `@types/node` build dependency?

didinele (approved):

Jiralite:
> Should this also bump the `@types/node` build dependency?

I have bumped this dependency everywhere but for packages/rest and apps/guide. It seems doing this causes many errors that should be resolved separately.

PR #10624, opened 2024-11-28
refactor(commands): use `Locale`

Jiralite:
`LocaleString` is deprecated. Switching to `Locale`.


vladfrangu (approved):
Technically this might be a major change since now it requires using the enum, but then again we also validate against it sooo 🤷, leaving it to @didinele to decide

PR #10626, opened 2024-11-28
refactor: use `cache.get()` for snowflakes, `resolve()` otherwise

sdanialraza:
This makes it consistent to use `cache.get()` for snowflakes and `resolve()` for everything else.

### DIFF
--- a/packages/discord.js/src/managers/GuildBanManager.js
+++ b/packages/discord.js/src/managers/GuildBanManager.js
@@ -158,7 +158,7 @@ class GuildBanManager extends CachedManager {
       reason: options.reason,
     });
     if (user instanceof GuildMember) return user;
-    const _user = this.client.users.cache.get(id) ?? null;
+    const _user = this.client.users.cache.get(id);
     if (_user) {
       return this.guild.members.resolve(_user) ?? _user;
     }
--- a/packages/discord.js/src/managers/GuildBanManager.js
+++ b/packages/discord.js/src/managers/GuildBanManager.js
@@ -158,7 +158,7 @@ class GuildBanManager extends CachedManager {
       reason: options.reason,
     });
     if (user instanceof GuildMember) return user;
-    const _user = this.client.users.resolve(id);
+    const _user = this.client.users.cache.get(id) ?? null;
     if (_user) {
       return this.guild.members.resolve(_user) ?? _user;
     }
--- a/packages/discord.js/src/managers/PresenceManager.js
+++ b/packages/discord.js/src/managers/PresenceManager.js
@@ -38,8 +38,8 @@ class PresenceManager extends CachedManager {
   resolve(presence) {
     const presenceResolvable = super.resolve(presence);
     if (presenceResolvable) return presenceResolvable;
-    const userResolvable = this.client.users.resolveId(presence);
-    return super.cache.get(userResolvable) ?? null;
+    const userId = this.client.users.resolveId(presence);
+    return super.cache.get(userId) ?? null;
   }
 
   /**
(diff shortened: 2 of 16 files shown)


didinele (dismissed):
surely this won't somehow break anything

packages/discord.js/src/managers/ThreadMemberManager.js
```diff
@@ -72,7 +72,7 @@ class ThreadMemberManager extends CachedManager {
     const memberResolvable = super.resolve(member);
     if (memberResolvable) return memberResolvable;
     const userResolvable = this.client.users.resolveId(member);
-    if (userResolvable) return super.resolve(userResolvable);
+    if (userResolvable) return super.cache.get(userResolvable);
```
Jiralite, line 75: It looks more appropriate to rename the variable to `userId`.

packages/discord.js/src/managers/PresenceManager.js
```diff
@@ -38,8 +38,8 @@ class PresenceManager extends CachedManager {
   resolve(presence) {
     const presenceResolvable = super.resolve(presence);
     if (presenceResolvable) return presenceResolvable;
-    const UserResolvable = this.client.users.resolveId(presence);
-    return super.resolve(UserResolvable);
+    const userResolvable = this.client.users.resolveId(presence);
+    return super.cache.get(userResolvable) ?? null;
```
Jiralite, line 42: It looks more appropriate to rename the variable to `userId`.

packages/discord.js/src/structures/ThreadMember.js
```diff
@@ -87,7 +87,7 @@ class ThreadMember extends Base {
    * @readonly
    */
   get user() {
-    return this.client.users.resolve(this.id);
+    return this.client.users.cache.get(this.id);
```
Jiralite, line 90:
```suggestion
    return this.client.users.cache.get(this.id) ?? null;
```

Jiralite (changes requested):

Jiralite (approved):

packages/discord.js/src/managers/GuildBanManager.js
```diff
@@ -158,7 +158,7 @@ class GuildBanManager extends CachedManager {
       reason: options.reason,
     });
     if (user instanceof GuildMember) return user;
-    const _user = this.client.users.resolve(id);
+    const _user = this.client.users.cache.get(id) ?? null;
```
vladfrangu, line 161: This is one of the cases, like in `packages/discord.js/src/client/actions/ThreadListSync.js`, where you don't need `?? null`

vladfrangu (approved):

SpaceEEC (approved):

DJj123dj:
This change DID indeed break a lot of bots. It might not look like it at a first glance, but the `cache.get()` method only allows ID's/snowflakes. It returns `undefined` when trying to resolve a `Role` or `User` class in the `PermissionOverwrites`.

The param type still allows the `Role` & `User` class to be used even if they aren't processed and will just return `undefined`.

😄 But no worries, in the latest version of discord.js is this code already overwritten/modified & now it works again.

Jiralite:
Code talks better! What was the issue? I want to understand what you are talking about.

DJj123dj:
> Code talks better! What was the issue? I want to understand what you are talking about.

Yea, the issue was exactly what is described in pull request #10686 as mentioned by @sdanialraza. The fact that it would return `undefined` when using a `discorde.Role` or `discord.User` in the `id` of the `PermissionOverwrites` instead of a `string` (snowflake)

Jiralite:
We had decoded your message by then! Please open an issue instead of commenting on pull requests in the future with a reproducible code sample next time.

PR #10631, opened 2024-12-03
feat: Undesired Properties

TempVoice:
Cache only the properties you need and remove the undesired one. This greatly improves the memory footprint.

For example, the `channel.topic` property of channels which consumes dozens of bytes. If a bot operates in thousands of servers this can waste unnecessary memory.
```js
const client = new Client({
	undesiredProperties: {
		ChannelManager: ["topic"]
	}
})
```

I am open for what you guys think about it. I am willing to battle test this for a bot with >150k servers.

### DIFF
--- a/packages/discord.js/src/client/Client.js
+++ b/packages/discord.js/src/client/Client.js
@@ -77,6 +77,8 @@ class Client extends BaseClient {
       ];
     }
 
+    this.undesiredProperties = this.options.undesiredProperties;
+
     this._validateOptions();
 
     /**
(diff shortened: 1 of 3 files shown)


TempVoice:
Got feedback that `delete` shouldn’t be used to remove properties. Instead will set the properties to undefined.

Will update changes later.

Jiralite:
For the record, you should open pull requests to main. The non-default branches should not be touched apart from backports from pull requests to main or specific fixes and other miscellaneous stuff.

TempVoice:
Alright thanks for letting me know.

Amgelo563:
Why is it called `undesiredProperties` and takes a `ChannelManager`, if that's the only key it checks either way? It pretty much only is a `undesiredChannelManagerProperties`.

TempVoice:
@Amgelo563 It’s meant to cover caching from other managers too. Agree I could have declared that more.

I know this pr could be controversial so I love feedback about it before adding support for more managers.

Maybe someone has an more ideal way to implement this before I „waste“ time.

PR #10633, opened 2024-12-04
feat: allow removing properties from cached objects

TempVoice:
Cache only the properties you need and remove the undesired one. This greatly improves the memory footprint.

For example, the `channel.topic` property of channels which consumes dozens of bytes. If a bot operates in thousands of servers this can waste unnecessary memory.
```js
const client = new Client({
	undesiredProperties: {
		ChannelManager: ["topic"]
	}
})
```

Goal is to support caching from other managers too.
I am open for what you guys think about it. I am willing to battle test this for a bot with >150k servers.

### DIFF
--- a/packages/discord.js/src/client/Client.js
+++ b/packages/discord.js/src/client/Client.js
@@ -72,6 +72,13 @@ class Client extends BaseClient {
      * @type {ClientPresence}
      */
     this.presence = new ClientPresence(this, this.options.ws.initialPresence ?? this.options.presence);
+    
+    /**
+     * The presence of the Client
+     * @private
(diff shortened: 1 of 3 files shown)


balkrishan99:
Feature Description:
The PR introduces an option to exclude specific properties from being cached. For example, removing the channel.topic property could save memory, especially in bots operating across thousands of servers.

Implementation:
The feature uses a new configuration option, undesiredProperties, in the client configuration:

javascript
Copy code
const client = new Client({
    undesiredProperties: {
        ChannelManager: ["topic"]
    }
});

PR #10640, opened 2024-12-07
feat: entry-point command

imnaiyar:
For mainlib
Depends on:
- https://github.com/discordjs/discord-api-types/pull/1166

### DIFF
--- a/packages/discord.js/src/structures/ApplicationCommand.js
+++ b/packages/discord.js/src/structures/ApplicationCommand.js
@@ -174,6 +174,19 @@ class ApplicationCommand extends Base {
       this.contexts ??= null;
     }
 
+    if ('handler' in data) {
+      /**
+       * Determines whether the interaction is handled by the app's interactions handler or by Discord.
+       * <info>Only available for {@link ApplicationCommandType.PrimaryEntryPoint} command
+       * and for apps with `EMBEDDED` flag (i.e, applications that have an Activity)
+       * </info>
+       * @type {?EntryPointCommandHandlerType}
+       */
+      this.handler = data.handler;
+    } else {
+      this.handler ??= null;
+    }
+
     if ('version' in data) {
       /**
        * Autoincrementing version identifier updated during substantial record changes
@@ -419,7 +432,8 @@ class ApplicationCommand extends Base {
         this.descriptionLocalizations ?? {},
       ) ||
       !isEqual(command.integrationTypes ?? command.integration_types ?? [], this.integrationTypes ?? []) ||
-      !isEqual(command.contexts ?? [], this.contexts ?? [])
+      !isEqual(command.contexts ?? [], this.contexts ?? []) ||
+      ('handler' in command && command.handler !== this.handler)
     ) {
       return false;
     }
--- a/packages/discord.js/src/structures/ApplicationCommand.js
+++ b/packages/discord.js/src/structures/ApplicationCommand.js
@@ -229,15 +229,20 @@ class ApplicationCommand extends Base {
    * @property {string} name The name of the command, must be in all lowercase if type is
    * {@link ApplicationCommandType.ChatInput}
    * @property {Object<Locale, string>} [nameLocalizations] The localizations for the command name
-   * @property {string} description The description of the command, if type is {@link ApplicationCommandType.ChatInput}
+   * @property {string} description The description of the command,
+   * if type is {@link ApplicationCommandType.ChatInput} or {@link ApplicationCommandType.PrimaryEntryPoint}
    * @property {boolean} [nsfw] Whether the command is age-restricted
    * @property {Object<Locale, string>} [descriptionLocalizations] The localizations for the command description,
-   * if type is {@link ApplicationCommandType.ChatInput}
+   * if type is {@link ApplicationCommandType.ChatInput} or {@link ApplicationCommandType.PrimaryEntryPoint}
    * @property {ApplicationCommandType} [type=ApplicationCommandType.ChatInput] The type of the command
    * @property {ApplicationCommandOptionData[]} [options] Options for the command
    * @property {?PermissionResolvable} [defaultMemberPermissions] The bitfield used to determine the default permissions
    * a member needs in order to run the command
-   * @property {boolean} [dmPermission] Whether the command is enabled in DMs
+   * @property {boolean} [dmPermission] Whether the command is enabled in DMs.
+   * Deprecated, use {@link ApplicationCommandData#contexts} instead.
+   * @property {ApplicationIntegrationType[]} [integrationTypes] Installation context(s) where the command is available
+   * @property {InteractionContextType[]} [contexts] Interaction context(s) where the command can be used
+   * @property {EntryPointCommandHandlerType} [handler] Determines whether the interaction is handled by the app's
    */
 
   /**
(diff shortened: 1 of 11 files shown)


packages/discord.js/typings/index.d.ts
```diff
@@ -579,6 +565,8 @@ export abstract class CommandInteraction<Cached extends CacheType = CacheType> e
   public reply(
     options: string | MessagePayload | InteractionReplyOptions,
   ): Promise<InteractionResponse<BooleanCache<Cached>>>;
+  public launchActivity(options: LaunchActivityOptions & { withResponse: true }): Promise<InteractionCallbackResponse>;
+  public launchActivity(options?: LaunchActivityOptions): Promise<undefined>;
```
SpaceEEC, line 569: `LaunchActivityOptions` may have `withResponse: true` without TypeScript being able to infer that.
This overload should return `InteractionCallbackResponse | undefined`.
You could another overload like the one above, but with `false` returning `undefined`.

The same goes for similar overloads.

SpaceEEC (changes requested):

packages/discord.js/src/structures/PrimaryEntryPointCommandInteraction.js
```diff
@@ -0,0 +1,10 @@
+'use strict';
+
+const { CommandInteraction } = require('./CommandInteraction.js');
+/**
+ * Represents a primary entry point command interaction.
+ * @extends {CommandInteraction}
+ */
+class PrimaryEntryPointCommandInteraction extends CommandInteraction {}
```
Qjuh, line 8: Isn’t this a somewhat breaking change in the sense that prior to this all instances of CommandInteraction automatically had an .options property (as evident by the needed change in typings)? Since this PR targets `main` anyway that wouldn’t be an issue, just curious about the semver label here.

imnaiyar: I already discussed this internally [here](https://discord.com/channels/222078108977594368/1209960895141388389/1314869981145595954) and vlad said it isn't breaking unless it's same in the runtime, which `options` isn't for `CommandInteraction`. Technically CommandInteraction is abstract class and it's upto parent class to implement it, so if we think it loke that it is kinda breaking. This is kinda gray area for me, probably needs to discuss what we wanna do

Qjuh (approved):

packages/discord.js/src/structures/ApplicationCommand.js
```diff
@@ -595,3 +613,8 @@ exports.ApplicationCommand = ApplicationCommand;
  * @external ApplicationCommandOptionAllowedChannelTypes
  * @see {@link https://discord.js.org/docs/packages/builders/stable/ApplicationCommandOptionAllowedChannelTypes:TypeAlias}
  */
+
```
vladfrangu, line 616: This goes in the [APITypes](https://github.com/discordjs/discord.js/blob/main/packages/discord.js/src/util/APITypes.js) file

vladfrangu (dismissed):

sdanialraza (approved):

packages/discord.js/src/structures/ApplicationCommand.js
```diff
@@ -162,6 +162,19 @@ class ApplicationCommand extends Base {
       this.contexts ??= null;
     }
 
+    if ('handler' in data) {
+      /**
+       * Determines whether the interaction is handled by the app's interactions handler or by Discord.
+       * <info>Only available for {@link ApplicationCommandType.PrimaryEntryPoint} command
+       * and for apps with {@link ApplicationFlags.Embedded} flag (i.e, applications that have an Activity)
```
almeidx, line 169:
```suggestion
       * <info>Only available for {@link ApplicationCommandType.PrimaryEntryPoint} commands on
       * applications with the {@link ApplicationFlags.Embedded} flag (i.e, those that have an Activity)</info>
```

packages/discord.js/src/structures/ApplicationCommand.js
```diff
@@ -204,14 +217,18 @@ class ApplicationCommand extends Base {
...
    * @property {boolean} [nsfw] Whether the command is age-restricted
    * @property {Object<Locale, string>} [descriptionLocalizations] The localizations for the command description,
-   * if type is {@link ApplicationCommandType.ChatInput}
+   * if type is {@link ApplicationCommandType.ChatInput} or {@link ApplicationCommandType.PrimaryEntryPoint}
    * @property {ApplicationCommandType} [type=ApplicationCommandType.ChatInput] The type of the command
    * @property {ApplicationCommandOptionData[]} [options] Options for the command
    * @property {?PermissionResolvable} [defaultMemberPermissions] The bitfield used to determine the default permissions
    * a member needs in order to run the command
+   * @property {ApplicationIntegrationType[]} [integrationTypes] Installation context(s) where the command is available
+   * @property {InteractionContextType[]} [contexts] Interaction context(s) where the command can be used
```
almeidx, line 230:
```suggestion
   * @property {ApplicationIntegrationType[]} [integrationTypes] Installation contexts where the command is available
   * @property {InteractionContextType[]} [contexts] Interaction contexts where the command can be used
```

packages/discord.js/src/structures/ApplicationCommand.js
```diff
@@ -204,14 +217,18 @@ class ApplicationCommand extends Base {
...
+   * @property {EntryPointCommandHandlerType} [handler] Determines whether the interaction is handled by the app's
```
almeidx, line 231:
```suggestion
   * @property {EntryPointCommandHandlerType} [handler] Whether the interaction is handled by the app's
   * interactions handler or by Discord.
```

packages/discord.js/src/structures/PrimaryEntryPointCommandInteraction.js
```diff
@@ -0,0 +1,10 @@
+'use strict';
+
+const { CommandInteraction } = require('./CommandInteraction.js');
+/**
```
almeidx, line 4:
```suggestion

/**
```

packages/discord.js/src/structures/interfaces/InteractionResponses.js
```diff
@@ -265,6 +271,25 @@ class InteractionResponses {
     return options.withResponse ? new InteractionCallbackResponse(this.client, response) : undefined;
   }
 
+  /**
+   * Launches activity, if enabled
```
almeidx, line 275:
```suggestion
   * Launches this application's activity, if enabled
```

almeidx (approved):

packages/discord.js/typings/index.d.ts
```diff
@@ -508,23 +510,6 @@ export type BooleanCache<Cached extends CacheType> = Cached extends 'cached' ? t
 export abstract class CommandInteraction<Cached extends CacheType = CacheType> extends BaseInteraction<Cached> {
   public type: InteractionType.ApplicationCommand;
   public get command(): ApplicationCommand | ApplicationCommand<{ guild: GuildResolvable }> | null;
-  public options: Omit<
```
didinele, line 511: Am I missing something? This was only added back to `ContextMenuCommandInteraction` but not `ChatInputCommandInteraction`

At which point, I'd also ask you extract it into a helper type.

Qjuh: `ChatInputCommandInteraction` already had it defined, since it overwrote the type of it.

didinele: Ahh, I had even looked at the full file but didn't realize. If it was properly `override` it would've showed up in this diff too and been more obvious.

didinele (approved):

vladfrangu (approved):

PR #10646, opened 2024-12-08
feat(interactions): add launchActivity method

imnaiyar:
 #10640 but for core

Depends on
- https://github.com/discordjs/discord-api-types/pull/1166


packages/core/src/api/interactions.ts
```diff
@@ -429,4 +429,54 @@ export class InteractionsAPI {
 
 		return with_response ? response : undefined;
 	}
+
+	/**
+	 * Launches activity and returns an interaction callback object
```
Jiralite, line 434:
```suggestion
	 * Launches an activity and returns an interaction callback object
```

packages/core/src/api/interactions.ts
```diff
@@ -429,4 +429,54 @@ export class InteractionsAPI {
...
+	 */
+	public async launchActivity(
+		interactionId: Snowflake,
+		interactionToken: string,
+		body?: RESTPostAPIInteractionCallbackQuery & { with_response: true },
+		options?: Pick<RequestData, 'signal'>,
+	): Promise<RESTPostAPIInteractionCallbackWithResponseResult>;
+
+	/**
+	 * Launches activity
```
Jiralite, line 450:
```suggestion
	 * Launches an activity
```

Jiralite (changes requested):

Jiralite (approved):

packages/core/src/api/interactions.ts
```diff
@@ -429,4 +429,54 @@ export class InteractionsAPI {
...
+	 * @param interactionToken - The token of the interaction
+	 * @param body - The callback data for launching the activity
+	 * @param options - The options for launching the activity
+	 */
+	public async launchActivity(
+		interactionId: Snowflake,
+		interactionToken: string,
+		body?: RESTPostAPIInteractionCallbackQuery & { with_response?: false },
+		options?: Pick<RequestData, 'signal'>,
+	): Promise<undefined>;
```
SpaceEEC, line 463: This overload will be selected when passing a `RESTPostAPIInteractionCallbackQuery` where `with_response` is `boolean`.
This could be a `true` in runtime and then return resolve to a `RESTPostAPIInteractionCallbackWithResponseResult`.

You could add another overload with `{ with_response: false }` where the function resolves to `undefined` and then the catch-all with an union of both.

SpaceEEC (changes requested):

packages/core/src/api/interactions.ts
Jiralite: Adding onto this, it could prove useful to add tests into https://github.com/discordjs/discord.js/blob/main/packages/core/__tests__/types.test.ts.

SpaceEEC (approved):

didinele (approved):

PR #10655, opened 2024-12-11
feat(GuildForumThreadManager): add pinned getter

SSardorf:
Given that there can only be one pinned thread per guild forum, a helper getter is added to provide an easier way to retrieve it.

### DIFF
--- a/packages/discord.js/src/managers/GuildForumThreadManager.js
+++ b/packages/discord.js/src/managers/GuildForumThreadManager.js
@@ -1,6 +1,6 @@
 'use strict';
 
-const { Routes } = require('discord-api-types/v10');
+const { Routes, ChannelFlags } = require('discord-api-types/v10');
 const ThreadManager = require('./ThreadManager');
 const { DiscordjsTypeError, ErrorCodes } = require('../errors');
 const MessagePayload = require('../structures/MessagePayload');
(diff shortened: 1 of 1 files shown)


Jiralite:
Pinned posts are decided with a channel flag—it would be very easy for Discord to say "Hey, let's make the maximum pinned posts 2!" and then suddenly this will be the source of confusion as it may return one or the other post only, which will result in an eventual deprecation and breaking change to remove it in the future.

I just don't see this as something the library should implement.

Jiralite:
In the event of an eventual change in the maximum amount of pinned posts, modifying `find()` to `filter()` is a breaking change. The return type changing to a possible thread to a collection is not backwards-compatible.

If this is to happen, it needs to be a filter now. That being said, a filter returning only one result is strange.

I am thinking long-term here, and it seems something better left to consumers on how they want to get a pinned post.

didinele:
On top of Jira's concerns, which I also agree with, I stand against adding a `get`ter that runs an O(n) computation. There's a few others in the library that I'm particularly vocal about, and they're massive performance hazards. If this were to be added into the interface *at all*, it should be a method to clearly signal that it's a computational task and not just an accessor shortcut.

Jiralite:
There is no consensus to move forward with this.

PR #10671, opened 2024-12-31
build: bump dependencies

almeidx:
Regenerated the lockfile and upgraded pnpm, which also solved the [/voice CI error](<https://github.com/discordjs/discord.js/actions/runs/12449063853/job/34754324830#step:7:105>)


package.json
```diff
@@ -89,13 +89,10 @@
 			"@contentlayer/utils>@opentelemetry/sdk-trace-base": "^1.15.1",
 			"@contentlayer/utils>@opentelemetry/sdk-trace-node": "^1.15.1",
 			"@contentlayer/utils>@opentelemetry/semantic-conventions": "^1.15.1"
-		},
-		"patchedDependencies": {
```
vladfrangu, line 93: Is this intended to be removed?

packages/discord.js/typings/index.d.ts
```diff
@@ -598,7 +598,7 @@ export abstract class CommandInteraction<Cached extends CacheType = CacheType> e
   ): Promise<ModalSubmitInteraction<Cached>>;
   private transformOption(
     option: APIApplicationCommandOption,
-    resolved: APIApplicationCommandInteractionData['resolved'],
+    resolved: Extract<APIApplicationCommandInteractionData, { resolved: unknown }>['resolved'],
```
vladfrangu, line 601: Shouldn't we use any? Does it work still?

iCrawl (approved):

PR #10672, opened 2025-01-01
fix(PresenceUpdate): correctly add user regardless of their properties

Digital39999:
Using [this](https://sourceb.in/3tgsknkS3S) code, after 1 minute (sweeper time), any event for users that were cached and then swiped will not be fired anymore, this is due to user not being created in cache at all even though it can be fully partial as types for `newPresence` state, this adds user regardless of the data and allows event to be fired (similarly how it adds for guild emojis).


packages/discord.js/src/client/actions/PresenceUpdate.js
```diff
@@ -6,7 +6,7 @@ const Events = require('../../util/Events');
 class PresenceUpdateAction extends Action {
   handle(data) {
     let user = this.client.users.cache.get(data.user.id);
-    if (!user && data.user.username) user = this.client.users._add(data.user);
+    if (!user && data.user) user = this.client.users._add(data.user);
```
Qjuh, line 9: This should check for enabled partial and if no User partial it should still require the username to be present

Digital39999: Indeed, realized that and was working on that.

SpaceEEC (approved):

monbrey (approved):

vladfrangu (approved):

PR #10682, opened 2025-01-03
feat(rest)!: allow passing tokens per request

ckohen:
BREAKING CHANGE: `RequestData.authPrefix` has been removed in favor of `RequestData.auth.prefix`

This allows passing a token when making a request (mostly useful for OAuth endpoints) and moves the token prefix into the same object as it didn't make sense having it separate.

This separates handlers based on the token used to isolate (non-global) ratelimits. There is a chance that ratelimits are hit if a bucket spans tokens. It does respect bucket hashes, so after the first request with each token they will be bucketed together.

Fixes DJS-41


vladfrangu (approved):

packages/rest/src/lib/utils/types.ts
```diff
@@ -278,17 +291,11 @@ export interface RequestData {
 	 */
 	appendToFormData?: boolean;
 	/**
-	 * If this request needs the `Authorization` header
+	 * Alternate authorization data to use for this request only, or `null` to disable the Authorizaiton header
```
Jiralite, line 294:
```suggestion
	 * Alternate authorization data to use for this request only, or `null` to disable the Authorization header
```

Jiralite (approved):

packages/rest/src/lib/utils/types.ts
```diff
@@ -278,17 +291,11 @@ export interface RequestData {
 	 */
 	appendToFormData?: boolean;
 	/**
-	 * If this request needs the `Authorization` header
+	 * Alternate authorization data to use for this request only, or `null` to disable the Authorization header
```
Jiralite, line 294:
```suggestion
	 * Alternate authorization data to use for this request only, or `false` to disable the Authorization header
```

Jiralite (changes requested):

Jiralite (approved):

SpaceEEC (approved):

almeidx (approved):

iCrawl (approved):

PR #10686, opened 2025-01-04
fix: use `resolve()` for `PermissionOverwrites`

sdanialraza:
This fixes an issue where `cache.get()` was used because of #10626, causing `PermissionOverwrites#resolve` to return undefined when `overwrite.id` may be a Role or User instance. Also updated `super.cache.get()` to `this.cache.get()` for clarity.

### DIFF
--- a/packages/discord.js/src/structures/PermissionOverwrites.js
+++ b/packages/discord.js/src/structures/PermissionOverwrites.js
@@ -180,7 +180,7 @@ class PermissionOverwrites extends Base {
       };
     }
 
-    const userOrRole = guild.roles.cache.get(overwrite.id) ?? guild.client.users.cache.get(overwrite.id);
+    const userOrRole = guild.roles.cache.resolve(overwrite.id) ?? guild.client.users.resolve(overwrite.id);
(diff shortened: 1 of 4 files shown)


vladfrangu (approved):

almeidx (approved):

packages/discord.js/src/structures/PermissionOverwrites.js
```diff
@@ -180,7 +180,7 @@ class PermissionOverwrites extends Base {
       };
     }
 
-    const userOrRole = guild.roles.cache.get(overwrite.id) ?? guild.client.users.cache.get(overwrite.id);
+    const userOrRole = guild.roles.cache.resolve(overwrite.id) ?? guild.client.users.resolve(overwrite.id);
```
Renegade334, line 183:
```suggestion
    const userOrRole = guild.roles.resolve(overwrite.id) ?? guild.client.users.resolve(overwrite.id);
```

Renegade334 (changes requested):

Jiralite (approved):

Issue #10658, opened 2024-12-15
Lost thread update when calling <ThreadChannel>.edit()


Apokalypt:
### Which package is this bug report for?

discord.js

### Issue description

When someone create a post on a server and put a very bad title like ``question``, our bot sends a button to edit the title post. When the user clicks on the button a modal is shown and the user has to enter the new title. When he submit the modal form I retrieve the input value and call ``thread.edit({ name: newNameFromInput })``

However, in some cases, the ``threadUpdate`` event received has the same object in the argument "old" and "new" and doesn't show that the new name is different from the old one as expected.
After digging in the code I saw that in the [method](https://github.com/discordjs/discord.js/blob/main/packages/discord.js/src/structures/ThreadChannel.js#L358) we are calling ``this.client.actions.ChannelUpdate.handle(newData);`` which update the data in the cache. Therefore, if this line executes before the event is received from the gateway the emitted event on the client will not expose the old name

### Code sample

```typescript
import { Client, Events, GatewayIntentBits, Partials } from 'discord.js';
import { token, thread } from './config.json';

const client = new Client({
    intents: [
        GatewayIntentBits.Guilds,
        GatewayIntentBits.GuildMembers
    ],
    partials: [Partials.GuildMember, Partials.ThreadMember, Partials.Channel]
});
client.once(Events.ClientReady, readyClient => {
	console.log(`Ready! Logged in as ${readyClient.user.tag}`);
});
client.on(Events.ThreadUpdate,(old, new) => {
	console.log("Thread update event received", old, new);
});
await client.login(token);

// As this bug is random and depends on when each network event occurs (gateway + API call)
// You have to execute the below code multiple times
const thread = await client.channels.fetch(thread);
await thread.edit({ name: "aaaa" + Math.round(Math.random() * 100) })

// To reproduce this bug easily you can modify the file "discord.js/src/client/websocket/handlers/THREAD_UPDATE.js" by putting this code
'use strict';

const Events = require('../../../util/Events');

module.exports = (client, packet) => {
    // Delay code execution to force order of code execution
    setTimeout(() => {
        const { old, updated } = client.actions.ChannelUpdate.handle(packet.d);
        if (old && updated) {
            /**
             * Emitted whenever a thread is updated - e.g. name change, archive state change, locked state change.
             * @event Client#threadUpdate
             * @param {ThreadChannel} oldThread The thread before the update
             * @param {ThreadChannel} newThread The thread after the update
             */
[...]

PR #10690, opened 2025-01-04
fix(ChannelUpdateAction): emit client event in handle to not miss change

Apokalypt:
This PR corrects the bug reported in #10658 by moving the emission of events inside the action so that no changes are zapped.
Currently, the `<ChannelUpdateAction>.handle` method is called within the `ThreadChannel` and `GuildChannelManager` classes. This call could result in a modification to the cached instance before the event was received from Discord, causing the event to be lost with the actual modification.

### DIFF
--- a/packages/discord.js/src/client/actions/ChannelUpdate.js
+++ b/packages/discord.js/src/client/actions/ChannelUpdate.js
@@ -2,6 +2,7 @@
 
 const Action = require('./Action');
 const { createChannel } = require('../../util/Channels');
+const Events = require('../../util/Events');
 
 class ChannelUpdateAction extends Action {
   handle(data) {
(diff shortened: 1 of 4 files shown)


SpaceEEC:
Wouldn't this change cause the update events to be emitted twice when first calling the action from within `edit` and then from the WebSocket handler?

Apokalypt:
> Wouldn't this change cause the update events to be emitted twice when first calling the action from within `edit` and then from the WebSocket handler?

It will, exactly like the method `<Guild>.edit()`

In my opinion, it's better to receive two events (one with old/new channels with the same data) than to lose the real modification. On the other hand, I agree with you that it's not ideal and that we should perhaps think about a better solution in general, as other events are subject to the same behavior.

PR #10697, opened 2025-01-11
fix(actions): respect ratelimits on split docs upload

ckohen:

with the massive amount of files we're uploading to blob storage we're running into ratelimits. This is making a conservative estimate of the ratelimit to avoid hitting it but also handles hitting it better

### DIFF
--- a/packages/actions/src/uploadSplitDocumentation/index.ts
+++ b/packages/actions/src/uploadSplitDocumentation/index.ts
@@ -1,15 +1,16 @@
 import { readFile } from 'node:fs/promises';
 import { basename, dirname, relative, sep } from 'node:path';
 import { cwd } from 'node:process';
+import { setTimeout as sleep } from 'node:timers/promises';
 import { getInput } from '@actions/core';
 import { create } from '@actions/glob';
 import { put } from '@vercel/blob';
-import pLimit from 'p-limit';
+import PQueue from 'p-queue';
 
 const pkg = getInput('package') || '*';
 const version = getInput('version') || 'main';
 
-const limit = pLimit(10);
+const queue = new PQueue({ concurrency: 10, interval: 60_000, intervalCap: 1_000 });
 const promises = [];
 
 const globber = await create(`packages/${pkg}/docs/${pkg}/split/*.api.json`);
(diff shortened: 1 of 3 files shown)


packages/actions/src/uploadSplitDocumentation/index.ts
```diff
@@ -20,13 +21,31 @@ for await (const file of globber.globGenerator()) {
...
+						}
+
+						if (typeof error === 'object' && error && 'retryAfter' in error && typeof error.retryAfter === 'number') {
+							await sleep(error.retryAfter * 1_000);
+							return upload(retries + 1);
+						} else {
+							console.error(`Could not upload ${file}`, error);
+						}
+					}
+				}
```
Jiralite, line 46: Could we do something like https://docs.github.com/en/actions/sharing-automations/creating-actions/setting-exit-codes-for-actions#setting-a-failure-exit-code-in-a-javascript-action here?

vladfrangu:
```suggestion
					} catch (error) {
						if (retries > 3) {
							console.error(`Could not upload ${file} after 3 retries`, error);
							return;
						}

						if (typeof error === 'object' && error && 'retryAfter' in error && typeof error.retryAfter === 'number') {
							await sleep(error.retryAfter * 1_000);
							return upload(retries + 1);
						} else {
							console.error(`Could not upload ${file}`, error);
							process.exitCode = 1;
						}
					}
				}
```

iCrawl: We already have `@actions/core` so we might as well use `core.setFailed`

ckohen: yeah that makes sense, we can set failed for either case here, good shout

packages/actions/src/uploadSplitDocumentation/index.ts
```diff
@@ -36,6 +58,9 @@ for await (const file of globber.globGenerator()) {
 
 try {
 	await Promise.all(promises);
+	if (failedUploads.length) {
+		setFailed(`Failed to upload ${failedUploads.length} files: ${failedUploads.join(', ')}`);
+	}
 } catch (error) {
 	console.log(error);
 }
```
Jiralite, line 66: What about the `console.log(error);` here?

ckohen: I can do all the other places if you want (old docs upload and search indices for example), just felt like I was tackling one thing in particular here

iCrawl (approved):

PR #10704, opened 2025-01-15
refactor!: Return only snowflakes for bulk delete

Jiralite:
We had to add `undefined` as a return type per #8410. This was because when message ids are supplied and deleted, they are attempted to be resolved from the cache. This removes the cache hit, and will now only return the ids that were removed. Users will have access to the messages they have cached before running this method anyway.

### DIFF
--- a/packages/discord.js/src/structures/interfaces/TextBasedChannel.js
+++ b/packages/discord.js/src/structures/interfaces/TextBasedChannel.js
@@ -294,7 +294,7 @@ class TextBasedChannel {
    * @example
    * // Bulk delete messages
    * channel.bulkDelete(5)
-   *   .then(messages => console.log(`Bulk deleted ${messages.size} messages`))
+   *   .then(messages => console.log(`Bulk deleted ${messages.length} messages`))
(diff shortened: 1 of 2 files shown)


packages/discord.js/src/structures/interfaces/TextBasedChannel.js
```diff
@@ -286,56 +286,44 @@ class TextBasedChannel {
...
       let messageIds =
         messages instanceof Collection ? [...messages.keys()] : messages.map(message => message.id ?? message);
+
       if (filterOld) {
         messageIds = messageIds.filter(
           id => Date.now() - DiscordSnowflake.timestampFrom(id) < MaxBulkDeletableMessageAge,
         );
       }
+
       if (messageIds.length === 0) return new Collection();
```
almeidx, line 311:
```suggestion
      if (messageIds.length === 0) return [];
```

almeidx (approved):

SpaceEEC (approved):

vladfrangu (approved):

PR #10709, opened 2025-01-18
docs(Message): improve message snapshots description

Amgelo563:

This improves the `Message#messageSnapshots` documentation, since the current one doesn't mention the snapshots themselves. Mainly opening the PR to gather comments on it if necessary.


packages/discord.js/src/structures/Message.js
```diff
@@ -425,7 +425,7 @@ class Message extends Base {
 
     if (data.message_snapshots) {
       /**
-       * The message associated with the message reference
+       * The message snapshots associated with the message reference. Currently Discord only supports one snapshot
```
Jiralite, line 428:
```suggestion
       * The message snapshots associated with the message reference. Currently, only one snapshot is supported
```

Jiralite (approved):

packages/discord.js/src/structures/Message.js
vladfrangu: I'm not a fan of this either, because we support an infinite amount of snapshots, Discord's the one providing only one right now...

Amgelo563: That was pretty much my reasoning, I felt like users may be misled into thinking it's on djs' side

Jiralite: > Either we drop the callout entirely, or we keep it and mention that its not our fault

I'm with dropping it. It keeps it simple.

Amgelo563: I feel the callout makes it clear how you can use it. Currently one may ask themselves how do you get the key of the forwarded message or under what situations there may be more than one entry (since it's a Collection it sounds like it can hold multiple values). With the callout it's clear you only need a `.first()` to take the only entry.

vladfrangu: We don't have such callouts for other collection-backed fields that may have only one entry, I don't see why we should for this one.

Also adding in a singular field is a no from me, it wouldn't bring any benefit, but maybe others disagree

Jiralite: Discord not providing a singular field is the reason we should not too. It's not future-proof, and will be inevitably removed.

Amgelo563: Sorry for late response, been busy with IRL stuff. I still think that the callout is worth the trouble, but given it's a majority I'll change it then.

almeidx (approved):

didinele (approved):

vladfrangu (approved):

kyranet (approved):

PR #10710, opened 2025-01-18
refactor!: use `AsyncEventEmitter` instead of `EventEmitter`

### LINKED ISSUE #10577: After upgrading @types/node to v22.7.9, an incompatible type error occurs in the “on” method of the client class

almeidx:
BREAKING CHANGE: The `BaseClient`, `Shard`, `ShardingManager`, and `Collector` classes now extend `AsyncEventEmitter` instead of `EventEmitter`.

Besides the obvious change, I also renamed `ClientEvents` to `ClientEventTypes` for consistency with the rest of the emitters.

Fixes #10577
Supersedes #10581, and #10584 partially

### DIFF
--- a/packages/discord.js/typings/index.d.ts
+++ b/packages/discord.js/typings/index.d.ts
@@ -959,7 +959,7 @@ export type If<Value extends boolean, TrueResult, FalseResult = null> = Value ex
     ? FalseResult
     : TrueResult | FalseResult;
 
-export class Client<Ready extends boolean = boolean> extends BaseClient<ClientEvents> {
+export class Client<Ready extends boolean = boolean> extends BaseClient<ClientEventTypes> {
   public constructor(options: ClientOptions);
   private actions: unknown;
   private expectedGuilds: Set<Snowflake>;
@@ -1098,9 +1098,12 @@ export interface CollectorEventTypes<Key, Value, Extras extends unknown[] = []> 
   end: [collected: ReadonlyCollection<Key, Value>, reason: string];
 }
 
-export abstract class Collector<Key, Value, Extras extends unknown[] = []> extends AsyncEventEmitter<
-  CollectorEventTypes<Key, Value, Extras>
-> {
+export abstract class Collector<
+  Key,
+  Value,
+  Extras extends unknown[] = [],
+  EventTypes extends {} = CollectorEventTypes<Key, Value, Extras>,
+> extends AsyncEventEmitter<EventTypes> {
   protected constructor(client: Client<true>, options?: CollectorOptions<[Value, ...Extras]>);
   private _timeout: NodeJS.Timeout | null;
   private _idletimeout: NodeJS.Timeout | null;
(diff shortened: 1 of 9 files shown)


Jiralite (approved):

kyranet (approved):
🎉

packages/discord.js/typings/index.d.ts
```diff
@@ -2783,26 +2726,6 @@ export class ReactionCollector extends Collector<Snowflake | string, MessageReac
   public collect(reaction: MessageReaction, user: User): Snowflake | string | null;
   public dispose(reaction: MessageReaction, user: User): Snowflake | string | null;
   public empty(): void;
-
-  public on(
```
Qjuh, line 2787: This is now missing the `remove` event in the typings.

packages/discord.js/typings/index.d.ts
```diff
@@ -2783,26 +2726,6 @@ export class ReactionCollector extends Collector<Snowflake | string, MessageReac
...
-    event: 'collect' | 'dispose' | 'remove' | 'ignore',
-    listener: (reaction: MessageReaction, user: User) => void,
-  ): this;
-  public on(
-    event: 'end',
-    listener: (collected: ReadonlyCollection<Snowflake, MessageReaction>, reason: string) => void,
-  ): this;
-  public on(event: string, listener: (...args: any[]) => void): this;
-
-  public once(
```
Qjuh, line 2797: Ditto

packages/discord.js/typings/index.d.ts
```diff
@@ -2031,19 +1987,6 @@ export class InteractionCollector<Interaction extends CollectedInteraction> exte
   public collect(interaction: Interaction): Snowflake;
   public empty(): void;
   public dispose(interaction: Interaction): Snowflake;
-  public on(event: 'collect' | 'dispose' | 'ignore', listener: (interaction: Interaction) => void): this;
```
Qjuh, line 2034: This did set the emitted variables correctly although the `Extras` type parameter in the extends was wrong. There are no extras emitted from this class, they typings were always wrong it seems. Need to remove the `[Collection<Snowflake, Interaction>]` on the extends of this class to be correct again.

Qjuh (changes requested):

Qjuh (approved):

vladfrangu (approved):

iCrawl (approved):

ckohen (approved):

PR #10718, opened 2025-01-19
refactor!: consolidate parameters into a single `options` object

sdanialraza:
Removes `options` as the second parameter. The first parameter now accepts `id` or `options` which can also have the `id` property. Internal discussion [here](https://canary.discord.com/channels/222078108977594368/1209960895141388389/1330584992564379648)

BREAKING CHANGE: `ApplicationCommandManager#fetch` and `GuildApplicationCommandManager#fetch` no longer accept 2 parameters. Instead, the first parameter accepts `id` or `options` which can now also have the `id` property.

### DIFF
--- a/packages/discord.js/src/managers/ApplicationCommandManager.js
+++ b/packages/discord.js/src/managers/ApplicationCommandManager.js
@@ -118,7 +118,7 @@ class ApplicationCommandManager extends CachedManager {
     const { cache, force, guildId, id, locale, withLocalizations } = options;
 
     if (typeof options === 'string' || typeof id === 'string') {
-      return this._fetchSingle({ cache, force, guildId, id: options.id ?? options });
+      return this._fetchSingle({ cache, force, guildId, id: id ?? options });
     }
 
     return this._fetchMany({ cache, guildId, locale, withLocalizations });
--- a/packages/discord.js/src/managers/ApplicationCommandManager.js
+++ b/packages/discord.js/src/managers/ApplicationCommandManager.js
@@ -89,39 +89,59 @@ class ApplicationCommandManager extends CachedManager {
 
   /**
    * Obtains one or multiple application commands from Discord, or the cache if it's already available.
-   * @param {FetchApplicationCommandOptions} [options] Options for fetching application command(s)
+   * @param {Snowflake|FetchApplicationCommandOptions} [options] Options for fetching application command(s)
    * @returns {Promise<ApplicationCommand|Collection<Snowflake, ApplicationCommand>>}
    * @example
    * // Fetch a single command
-   * client.application.commands.fetch({ id: '123456789012345678' })
+   * client.application.commands.fetch('123456789012345678')
    *   .then(command => console.log(`Fetched command ${command.name}`))
    *   .catch(console.error);
    * @example
    * // Fetch all commands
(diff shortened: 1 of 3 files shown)


vladfrangu (approved):

Jiralite (commented):
What about leaving the `id` overload (passing the id)? That is a common pattern for managers fetching.

OfficialSirH (approved):
Muy bueno 👌

didinele (approved):

Jiralite (changes requested):
I believe the breaking change entry needs amending. Would you mind doing that?

sdanialraza:
> I believe the breaking change entry needs amending. Would you mind doing that?

Done

packages/discord.js/src/managers/ApplicationCommandManager.js
```diff
@@ -98,28 +98,50 @@ class ApplicationCommandManager extends CachedManager {
...
-      if (!force) {
-        const existing = this.cache.get(id);
-        if (existing) return existing;
-      }
-      const command = await this.client.rest.get(this.commandPath({ id, guildId }));
-      return this._add(command, cache);
+  async fetch(options) {
+    if (!options) return this._fetchMany();
+
+    const { cache, force, guildId, id, locale, withLocalizations } = options;
```
almeidx, line 118: This might technically work but, if an id is passed instead of an object all these variables will be undefined..
I'd rather check if `typeof options === 'string'` before this and handle accordingly.

almeidx (changes requested):

packages/discord.js/src/managers/ApplicationCommandManager.js
sdanialraza: Fixed via [`59b7252` (#10718)](https://github.com/discordjs/discord.js/pull/10718/commits/59b72527358db4cfcdf94b1ca143c9374f47ba34)

packages/discord.js/typings/index.d.ts
```diff
@@ -3913,14 +3913,16 @@ export class ApplicationCommandManager<
...
-  public fetch(options: FetchApplicationCommandOptions): Promise<Collection<Snowflake, ApplicationCommandScope>>;
-  public fetch(id: Snowflake, options?: FetchApplicationCommandOptions): Promise<ApplicationCommandScope>;
   public fetch(
-    id?: Snowflake,
-    options?: FetchApplicationCommandOptions,
+    options: Omit<FetchApplicationCommandOptions, 'id'> & { guildId: Snowflake },
+  ): Promise<Collection<Snowflake, ApplicationCommandScope>>;
+  public fetch(
+    options?: Omit<FetchApplicationCommandOptions, 'id' | 'guildId'>,
   ): Promise<Collection<Snowflake, ApplicationCommandScope>>;
```
almeidx, line 3926: `guildId` is optional in `FetchApplicationCommandOptions`. We can just make the `options` parameter optional to merge both these overloads.

```suggestion
    options?: Omit<FetchApplicationCommandOptions, 'id'>,
  ): Promise<Collection<Snowflake, ApplicationCommandScope>>;
```

almeidx (changes requested):

almeidx (approved):

PR #10723, opened 2025-01-21
feat(collection): honour subclassing via `@@species` in static methods

Renegade334:

Collection methods feature subclassing support via `Symbol.species`. This currently doesn't apply to the one static method, which calls the Collection constructor specifically.

This change brings the behaviour of static methods in line with instance methods, such that `DerivedCollection.combineEntries(...)` returns a DerivedCollection, not a Collection.

### DIFF
--- a/packages/collection/src/collection.ts
+++ b/packages/collection/src/collection.ts
@@ -11,11 +11,11 @@ export type ReadonlyCollection<Key, Value> = Omit<
 
 export interface Collection<Key, Value> {
 	/**
-	 * Ambient declaration to allow `this.constructor[@@species]` in class methods.
+	 * Ambient declaration to allow references to `this.constructor` in class methods.
 	 *
 	 * @internal
 	 */
-	constructor: typeof Collection & { readonly [Symbol.species]: typeof Collection };
+	constructor: typeof Collection;
 }
 
 /**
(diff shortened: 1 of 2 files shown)


packages/collection/src/collection.ts
```diff
@@ -11,11 +11,11 @@ export type ReadonlyCollection<Key, Value> = Omit<
...
 export interface Collection<Key, Value> {
 	/**
-	 * Ambient declaration to allow `this.constructor[@@species]` in class methods.
+	 * Ambient declaration to allow references to `this.constructor` in class methods.
 	 *
 	 * @internal
 	 */
-	constructor: typeof Collection & { readonly [Symbol.species]: typeof Collection };
+	constructor: typeof Collection;
 }
```
Renegade334, line 19: In-class `declare` syntax specifically balks at a property named `constructor`, so this declaration has to remain in its own interface rather than being merged into the class body.

almeidx:
Can you add tests for this?

Renegade334:
Subclassing is currently not tested at all, which is why I didn't add any. If desirable here, then they probably also need adding for the other methods as well?

almeidx (approved):

Jiralite (approved):

benjGam (approved):

SpaceEEC (approved):

PR #10754, opened 2025-02-12
refactor(Client)!: Remove emojis getter

Wolvinny:

With the introductions of application emojis, accessed using `<ClientApplication>.emojis`, this naming has become confusing, since it only includes guild emojis.
Adding application emojis to Client#emojis is not an option since those require fetching.
Other factors are that it can cause performance issues, and, well, the fact app emojis now exist.


packages/discord.js/src/util/Util.js
```diff
@@ -122,6 +122,29 @@ function resolvePartialEmoji(emoji) {
   return { id, name, animated: Boolean(animated) };
 }
 
+/**
+ * Resolve an emoji object from a Guild
+ * @param {Client} client The Client
+ * @param {string} emojiId The id of the emoji
+ * @returns {GuildEmoji | undefined}
```
sdanialraza, line 129:
```suggestion
 * Resolves a {@link GuildEmoji} from an emoji id.
 * @param {Client} client The client to use to resolve the emoji
 * @param {Snowflake} emojiId The emoji id to resolve
 * @returns {?GuildEmoji}
```

packages/discord.js/typings/index.d.ts
```diff
@@ -3477,6 +3477,8 @@ export function resolveColor(color: ColorResolvable): number;
 export function resolvePartialEmoji(emoji: Snowflake): PartialEmojiOnlyId;
 /** @internal */
 export function resolvePartialEmoji(emoji: Emoji | EmojiIdentifierResolvable): PartialEmoji | null;
+/** @internal */
+export function resolveGuildEmoji(client: Client, emojiId: string): GuildEmoji | null;
```
sdanialraza, line 3481:
```suggestion
export function resolveGuildEmoji(client: Client, emojiId: Snowflake): GuildEmoji | null;
```

sdanialraza (changes requested):

kyranet (approved):

sdanialraza (approved):

didinele (approved):

PR #10758, opened 2025-02-14
feat(ws): zstd streaming support

vladfrangu:

Brings in native zstd streaming support for gateway decompression in environments that have it

That said, I'm leaving this as a draft for now, and seeing how zstd support progresses in node (specifically the backporting)

If someone wants to test it locally with their bots that have bigger payloads, you can use a script like this, after cloning this PR and building the repo

```js
import { REST } from '@discordjs/rest';
import { CompressionMethod, WebSocketManager } from '@discordjs/ws';

const token = '';

const rest = new REST({}).setToken(token);

const manager = new WebSocketManager({
	intents: 0,
	rest,
	token,
	shardCount: 16,
	shardIds: [0],
	helloTimeout: 10_000,
	compression: CompressionMethod.ZstdNative,
});

manager.on('debug', console.log);
manager.on('error', console.error);

await manager.connect();
console.log('Connected');
```

Runnable with `volta run --node 23 node script.mjs` or whichever tool floats your boat.

Ideally we'd have some non native zstd module too but there doesn't seem to be any that exist right now 🙃

### DIFF
--- a/packages/ws/src/utils/constants.ts
+++ b/packages/ws/src/utils/constants.ts
@@ -20,6 +20,7 @@ export enum Encoding {
 export enum CompressionMethod {
 	ZlibNative,
 	ZlibSync,
+	ZstdStream,
 }
 
 export const DefaultDeviceProperty = `@discordjs/ws [VI]{{inject}}[/VI]` as `@discordjs/ws ${string}`;
@@ -29,6 +30,7 @@ const getDefaultSessionStore = lazy(() => new Collection<number, SessionInfo | n
 export const CompressionParameterMap = {
 	[CompressionMethod.ZlibNative]: 'zlib-stream',
 	[CompressionMethod.ZlibSync]: 'zlib-stream',
+	[CompressionMethod.ZstdStream]: 'zstd-stream',
 } as const satisfies Record<CompressionMethod, string>;
 
 /**
--- a/packages/ws/src/ws/WebSocketShard.ts
+++ b/packages/ws/src/ws/WebSocketShard.ts
@@ -222,7 +222,7 @@ export class WebSocketShard extends AsyncEventEmitter<WebSocketShardEventsMap> {
 
 						this.nativeInflate = inflate;
 					} else {
-						console.warn('WebSocketShard: Compression is set to native but node:zlib is not available.');
+						console.warn('WebSocketShard: Compression is set to native zlib but node:zlib is not available.');
 						params.delete('compress');
 					}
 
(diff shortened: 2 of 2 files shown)


packages/ws/src/ws/WebSocketShard.ts
```diff
@@ -632,12 +661,14 @@ export class WebSocketShard extends AsyncEventEmitter<WebSocketShardEventsMap> {
...
 		// Deal with transport compression
 		if (this.transportCompressionEnabled) {
+			// Each WS message received is a full gateway message for zstd streaming, but for zlib it's chunked
 			const flush =
-				decompressable.length >= 4 &&
-				decompressable.at(-4) === 0x00 &&
-				decompressable.at(-3) === 0x00 &&
-				decompressable.at(-2) === 0xff &&
-				decompressable.at(-1) === 0xff;
+				this.strategy.options.compression === CompressionMethod.ZstdNative ||
```
didinele, line 666: I know this is a draft, but flagging now

This check isn't correct, because there's cases where we don't uphold the compression option passes (i.e. the logic in the constructor). I think generally the other code relies on the presence of the Inflate objects specific to that method.

didinele (dismissed):

packages/ws/src/ws/WebSocketShard.ts
didinele: Ohhh okay, I didn't notice you shared the `nativeInflate` prop, also forgot what outter if with the getter, I see.

Yeah, I guess this works fine.

packages/ws/src/utils/constants.ts
```diff
@@ -20,6 +20,7 @@ export enum Encoding {
 export enum CompressionMethod {
```
JMTK, line 20: Just out of curiousity, could you do

```ts
export enum CompressionMethod {
	ZlibNative = 'zlib-stream',
	ZlibSync = 'zlib-stream',
	ZstdNative = 'zstd-stream',
}
```

and just remove the `CompressionParameterMap` or was there a functional reason not to use it?

didinele: We used to do this, but I changed it in #10243; memory is doing me dirty right now so I wouldn't be able to tell you why, but I definitely remember the enum values were terribly annoying for some reason.

didinele: Got it. It's because of the switch statements where we do `case CompressionMethod.ZlibNative` and such. We need to actually distinguish between ALL of the types, which an enum of that sort would make impossible

almeidx:
This branch is available for testing at `@discordjs/ws@3.0.0-dev.1746803407-02175c0dc`

packages/ws/src/ws/WebSocketShard.ts
```diff
@@ -243,6 +243,35 @@ export class WebSocketShard extends AsyncEventEmitter<WebSocketShardEventsMap> {
...
+
+				case CompressionMethod.ZstdNative: {
+					const zlib = await getNativeZlib();
+					if (zlib && 'createZstdDecompress' in zlib) {
+						this.inflateBuffer = [];
+
+						// @ts-expect-error node 23 called, it wants its code back
+						const inflate = zlib.createZstdDecompress({
+							chunkSize: 65_535,
+						}) as nativeZlib.Inflate;
```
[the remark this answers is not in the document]
vladfrangu: Last I tested this, it worked fine without that 👀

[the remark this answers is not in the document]
Renegade334: I think it might make a difference for low-traffic websockets, as incoming messages might end up getting stuck in the decompression buffer and not passed on to the client until the buffer fills up enough for zstd to close the frame.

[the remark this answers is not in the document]
vladfrangu: Well, each ws message is a full payload compressed with zstd, and I've also tested this on a bot in a guild that was getting _many_ payloads/sec with 0 issues. Coulddd you test the hypothesis please and let us know results 🙏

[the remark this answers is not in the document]
Renegade334: Looks like this applies only to the compression stream, not decompression. Ignore me! 👍

almeidx:
@discord-js-bot pack this

almeidx (approved):

didinele (approved):

PR #10765, opened 2025-02-19
fix(MessagePayload): preserve existing flags when editing

sdanialraza:
Fixes regression from #10660

### DIFF
--- a/packages/discord.js/src/structures/MessagePayload.js
+++ b/packages/discord.js/src/structures/MessagePayload.js
@@ -149,15 +149,10 @@ class MessagePayload {
 
     let flags;
     if (
-      this.options.flags !== undefined ||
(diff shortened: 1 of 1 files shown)


vladfrangu (approved):

Jiralite (dismissed):
Can we not replace it all with this?

```JavaScript
if (
  this.options.flags != null
) {
  flags = new MessageFlagsBitField(this.options.flags);
}
```

#7514 added the `this.options.reply` check, but that code path included using the target's bit field. This code is simply "if you have a bit field, we use it, else we do not".

Jiralite (approved):

didinele (approved):

vladfrangu (approved):

PR #10784, opened 2025-02-28
types(GuildMember): change partial typing

Pavel-Boyazov:
[Discord API's Guild Member object](https://discord.com/developers/docs/resources/guild#guild-member-object) has isn't nullable `joinedAt` property.
This also confirms the [partial property set to false](https://github.com/discordjs/discord.js/blob/main/packages/discord.js/typings/index.d.ts#L1627), which depends on the [nullability of `joinedAt`](https://github.com/discordjs/discord.js/blob/88bfeaab224b896147483e59013874c0de2dfbac/packages/discord.js/src/structures/GuildMember.js#L136) and the [use of `Partialize`](https://github.com/discordjs/discord.js/blob/88bfeaab224b896147483e59013874c0de2dfbac/packages/discord.js/typings/index.d.ts#L6579) for this property to create the `PartialGuildMember` interface.

### DIFF
--- a/packages/discord.js/typings/index.d.ts
+++ b/packages/discord.js/typings/index.d.ts
@@ -1616,8 +1616,8 @@ export class GuildMember extends Base {
   public get communicationDisabledUntil(): Date | null;
   public communicationDisabledUntilTimestamp: number | null;
   public flags: Readonly<GuildMemberFlagsBitField>;
-  public get joinedAt(): Date | null;
-  public joinedTimestamp: number | null;
+  public get joinedAt(): Date;
+  public joinedTimestamp: number;
   public get kickable(): boolean;
   public get manageable(): boolean;
   public get moderatable(): boolean;
--- a/packages/discord.js/typings/index.d.ts
+++ b/packages/discord.js/typings/index.d.ts
@@ -1622,7 +1622,10 @@ export class GuildMember extends Base {
   public get manageable(): boolean;
   public get moderatable(): boolean;
   public nickname: string | null;
-  public get partial(): false;
+  public get partial(): this is GuildMember & {
+    joinedTimestamp: null;
+    readonly joinedAt: null;
+  };
   public get permissions(): Readonly<PermissionsBitField>;
   public get premiumSince(): Date | null;
   public premiumSinceTimestamp: number | null;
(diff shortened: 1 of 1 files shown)


Qjuh:
The reason why all `GuildMember` in cache can have those values nullable is because of
https://github.com/discordjs/discord.js/blob/88bfeaab224b896147483e59013874c0de2dfbac/packages/discord.js/src/client/websocket/handlers/PRESENCE_UPDATE.js#L24-L28
Here we add partial `GuildMember` data to the cache, because the `PRESENCE_UPDATE` payload does not contain the `member` data but only the `user`.
So with the current state of the code this typing change would be incorrect.

Pavel-Boyazov:
Then, we need to change type of `partial` property and add type guard for `joinedAt` based on this.
At the moment, its type is always `false`, although in the scenario you described, it may correspond to `true`.

Pavel-Boyazov:
> The reason why all `GuildMember` in cache can have those values nullable is because of
> 
> https://github.com/discordjs/discord.js/blob/88bfeaab224b896147483e59013874c0de2dfbac/packages/discord.js/src/client/websocket/handlers/PRESENCE_UPDATE.js#L24-L28
> 
> 
> Here we add partial `GuildMember` data to the cache, because the `PRESENCE_UPDATE` payload does not contain the `member` data but only the `user`.
> So with the current state of the code this typing change would be incorrect.

How about to use `PartialGuildMember` type in cache? Why client partial not affect on this?

Pavel-Boyazov:
The simplest solution would be to change the type of `partial` to `boolean`, but it seems to me that it would be good to use the existing type of `PartialGuildMember` in places where it can be (like cache)

Qjuh:
The better solution would be to fix the behavior to not put the GuildMember into cache if it's partial like that and no GuildMember partial was set in client constructor. Instead have the `Presence#member` return a PartialGuildMember constructed on the fly if not cached (or simply return `null` and add a `Presence#user` getter too).

Because the way it currently is means *almost all* places using GuildMember can theoretically get passed such a partial GuildMember if they rely on cache.

For other structures we make sure to not have them ever be put into cache as partial structures if the partial wasn't provided in client.

Pavel-Boyazov:
Ok, thanks for your answer. From the very beginning, it seemed strange to me to use partial structures despite the client's settings. I'll be change that in other PR.

PR #10785, opened 2025-03-01
fix(GuildMember): cache based on partials options

Pavel-Boyazov:
This PR based on [previous PR's](https://github.com/discordjs/discord.js/pull/10784) conversation.

[Discord API's Guild Member object](https://discord.com/developers/docs/resources/guild#guild-member-object) has isn't nullable `joinedAt` property.
This also confirms the [partial property set to false](https://github.com/discordjs/discord.js/blob/main/packages/discord.js/typings/index.d.ts#L1627), which depends on the [nullability of `joinedAt`](https://github.com/discordjs/discord.js/blob/88bfeaab224b896147483e59013874c0de2dfbac/packages/discord.js/src/structures/GuildMember.js#L136) and the [use of `Partialize`](https://github.com/discordjs/discord.js/blob/88bfeaab224b896147483e59013874c0de2dfbac/packages/discord.js/typings/index.d.ts#L6579) for this property to create the `PartialGuildMember` interface.
It's also incorrect that caching of a partial member is performed despite the provided partials options.

### DIFF
--- a/packages/discord.js/src/client/websocket/handlers/PRESENCE_UPDATE.js
+++ b/packages/discord.js/src/client/websocket/handlers/PRESENCE_UPDATE.js
@@ -21,11 +21,14 @@ module.exports = (client, { d: data }) => {
 
   let member = guild.members.cache.get(user.id);
   if (!member && data.status !== 'offline') {
-    member = guild.members._add({
-      user,
-      deaf: false,
-      mute: false,
-    });
+    member = guild.members._add(
+      {
+        user,
+        deaf: false,
+        mute: false,
+      },
+      client.options.partials.includes(Partials.GuildMember),
+    );
 
     client.emit(Events.GuildMemberAvailable, member);
   }
--- a/packages/discord.js/src/managers/GuildMemberManager.js
+++ b/packages/discord.js/src/managers/GuildMemberManager.js
@@ -31,7 +31,7 @@ class GuildMemberManager extends CachedManager {
 
   /**
    * The cache of this Manager
-   * @type {Collection<Snowflake, GuildMember>}
+   * @type {Collection<Snowflake, GuildMember | PartialGuildMember>}
    * @name GuildMemberManager#cache
    */
 
@@ -42,7 +42,7 @@ class GuildMemberManager extends CachedManager {
   /**
    * Resolves a {@link UserResolvable} to a {@link GuildMember} object.
    * @param {UserResolvable} member The user that is part of the guild
-   * @returns {?GuildMember}
+   * @returns {?(GuildMember | PartialGuildMember)}
    */
   resolve(member) {
     const memberResolvable = super.resolve(member);
@@ -132,7 +132,7 @@ class GuildMemberManager extends CachedManager {
 
   /**
    * The client user as a GuildMember of this guild
-   * @type {?GuildMember}
+   * @type {?(GuildMember | PartialGuildMember)}
    * @readonly
    */
   get me() {
--- a/packages/discord.js/src/managers/GuildMemberRoleManager.js
+++ b/packages/discord.js/src/managers/GuildMemberRoleManager.js
@@ -16,7 +16,7 @@ class GuildMemberRoleManager extends DataManager {
 
     /**
      * The GuildMember this manager belongs to
-     * @type {GuildMember}
+     * @type {GuildMember | PartialGuildMember}
      */
     this.member = member;
 
--- a/packages/discord.js/src/structures/Presence.js
+++ b/packages/discord.js/src/structures/Presence.js
@@ -63,7 +63,7 @@ class Presence extends Base {
 
   /**
    * The member of this presence
-   * @type {?GuildMember}
+   * @type {?(GuildMember | PartialGuildMember)}
    * @readonly
    */
   get member() {
--- a/packages/discord.js/typings/index.d.ts
+++ b/packages/discord.js/typings/index.d.ts
@@ -677,7 +677,7 @@ export class BaseGuildVoiceChannel extends GuildChannel {
   public bitrate: number;
   public get full(): boolean;
   public get joinable(): boolean;
-  public get members(): Collection<Snowflake, GuildMember>;
+  public get members(): Collection<Snowflake, GuildMember | PartialGuildMember>;
   public nsfw: boolean;
   public rateLimitPerUser: number | null;
   public rtcRegion: string | null;
@@ -1541,7 +1541,10 @@ export class GuildBan extends Base {
 
 export abstract class GuildChannel extends BaseChannel {
   public constructor(guild: Guild, data?: RawGuildChannelData, client?: Client<true>, immediatePatch?: boolean);
-  private memberPermissions(member: GuildMember, checkAdmin: boolean): Readonly<PermissionsBitField>;
+  private memberPermissions(
+    member: GuildMember | PartialGuildMember,
+    checkAdmin: boolean,
+  ): Readonly<PermissionsBitField>;
   private rolePermissions(role: Role, checkAdmin: boolean): Readonly<PermissionsBitField>;
   public get createdAt(): Date;
   public get createdTimestamp(): number;
@@ -1550,7 +1553,7 @@ export abstract class GuildChannel extends BaseChannel {
   public guild: Guild;
   public guildId: Snowflake;
   public get manageable(): boolean;
-  public get members(): Collection<Snowflake, GuildMember>;
+  public get members(): Collection<Snowflake, GuildMember | PartialGuildMember>;
   public name: string;
   public get parent(): CategoryChannel | null;
   public parentId: Snowflake | null;
@@ -1565,7 +1568,10 @@ export abstract class GuildChannel extends BaseChannel {
   public edit(options: GuildChannelEditOptions): Promise<this>;
   public equals(channel: GuildChannel): boolean;
   public lockPermissions(): Promise<this>;
-  public permissionsFor(memberOrRole: GuildMember | Role, checkAdmin?: boolean): Readonly<PermissionsBitField>;
+  public permissionsFor(
+    memberOrRole: GuildMember | PartialGuildMember | Role,
+    checkAdmin?: boolean,
+  ): Readonly<PermissionsBitField>;
   public permissionsFor(
     memberOrRole: UserResolvable | RoleResolvable,
     checkAdmin?: boolean,
@@ -1618,8 +1624,8 @@ export class GuildMember extends Base {
   public get communicationDisabledUntil(): Date | null;
   public communicationDisabledUntilTimestamp: number | null;
   public flags: Readonly<GuildMemberFlagsBitField>;
-  public get joinedAt(): Date | null;
-  public joinedTimestamp: number | null;
+  public get joinedAt(): Date;
+  public joinedTimestamp: number;
   public get kickable(): boolean;
   public get manageable(): boolean;
   public get moderatable(): boolean;
@@ -2135,7 +2141,7 @@ export class Message<InGuild extends boolean = boolean> extends Base {
   public get hasThread(): boolean;
   public id: Snowflake;
   public interactionMetadata: MessageInteractionMetadata | null;
-  public get member(): GuildMember | null;
+  public get member(): GuildMember | PartialGuildMember | null;
   public mentions: MessageMentions<InGuild>;
   public nonce: string | number | null;
   public get partial(): false;
@@ -2370,7 +2376,7 @@ export class MessageMentions<InGuild extends boolean = boolean> {
   );
   private _channels: Collection<Snowflake, Channel> | null;
   private readonly _content: string;
-  private _members: Collection<Snowflake, GuildMember> | null;
+  private _members: Collection<Snowflake, GuildMember | PartialGuildMember> | null;
   private _parsedUsers: Collection<Snowflake, User> | null;
 
   public get channels(): Collection<Snowflake, Channel>;
@@ -2378,7 +2384,7 @@ export class MessageMentions<InGuild extends boolean = boolean> {
   public everyone: boolean;
   public readonly guild: If<InGuild, Guild>;
   public has(data: UserResolvable | RoleResolvable | ChannelResolvable, options?: MessageMentionsHasOptions): boolean;
-  public get members(): If<InGuild, Collection<Snowflake, GuildMember>>;
+  public get members(): If<InGuild, Collection<Snowflake, GuildMember | PartialGuildMember>>;
   public get parsedUsers(): Collection<Snowflake, User>;
   public repliedUser: User | null;
   public roles: Collection<Snowflake, Role>;
@@ -2693,7 +2699,7 @@ export class Presence extends Base {
   public activities: Activity[];
   public clientStatus: ClientPresenceStatusData | null;
   public guild: Guild | null;
-  public get member(): GuildMember | null;
+  public get member(): GuildMember | PartialGuildMember | null;
   public status: PresenceStatus;
   public get user(): User | null;
   public userId: Snowflake;
@@ -2803,7 +2809,7 @@ export class Role extends Base {
   public hoist: boolean;
   public id: Snowflake;
   public managed: boolean;
-  public get members(): Collection<Snowflake, GuildMember>;
+  public get members(): Collection<Snowflake, GuildMember | PartialGuildMember>;
   public mentionable: boolean;
   public name: string;
   public permissions: Readonly<PermissionsBitField>;
@@ -3355,7 +3361,10 @@ export class ThreadChannel<ThreadOnly extends boolean = boolean> extends BaseCha
   public edit(options: ThreadEditOptions): Promise<this>;
   public join(): Promise<this>;
   public leave(): Promise<this>;
-  public permissionsFor(memberOrRole: GuildMember | Role, checkAdmin?: boolean): Readonly<PermissionsBitField>;
+  public permissionsFor(
+    memberOrRole: GuildMember | PartialGuildMember | Role,
+    checkAdmin?: boolean,
+  ): Readonly<PermissionsBitField>;
   public permissionsFor(
     memberOrRole: UserResolvable | RoleResolvable,
     checkAdmin?: boolean,
@@ -3403,7 +3412,7 @@ export class Typing extends Base {
   public startedTimestamp: number;
   public get startedAt(): Date;
   public get guild(): Guild | null;
-  public get member(): GuildMember | null;
+  public get member(): GuildMember | PartialGuildMember | null;
   public inGuild(): this is this & {
     channel: TextChannel | AnnouncementChannel | ThreadChannel;
     get guild(): Guild;
@@ -3586,7 +3595,7 @@ export class VoiceState extends Base {
   public get deaf(): boolean | null;
   public guild: Guild;
   public id: Snowflake;
-  public get member(): GuildMember | null;
+  public get member(): GuildMember | PartialGuildMember | null;
   public get mute(): boolean | null;
   public selfDeaf: boolean | null;
   public selfMute: boolean | null;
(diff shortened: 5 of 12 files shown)


packages/discord.js/src/managers/GuildMemberManager.js
```diff
@@ -31,7 +31,7 @@ class GuildMemberManager extends CachedManager {
 
   /**
    * The cache of this Manager
-   * @type {Collection<Snowflake, GuildMember>}
+   * @type {Collection<Snowflake, GuildMember | PartialGuildMember>}
```
Qjuh, line 34: The whole idea of fixing this was that there *won‘t* be `PartialGuildMember` in the cache. And there is no `PartialGuildMember` in JS land anyway, so this would only make sense in the typings (but is wrong there for the above reason).
Same applies for all other JSDoc comments you added `PartialGuildMember` to

packages/discord.js/src/client/websocket/handlers/PRESENCE_UPDATE.js
```diff
@@ -21,11 +21,14 @@ module.exports = (client, { d: data }) => {
...
-      deaf: false,
-      mute: false,
-    });
+    member = guild.members._add(
+      {
+        user,
+        deaf: false,
+        mute: false,
+      },
+      client.options.partials.includes(Partials.GuildMember),
```
Qjuh, line 30: This needs more/better logic. The member should always be patched with that data, if cached before, but should never be newly added to cache. The `guildMemberAvaliable` event here should only emit if the partial is set. So add the partial check to the if conditions surrounding this block instead.

packages/discord.js/src/structures/Presence.js
```diff
@@ -63,7 +63,7 @@ class Presence extends Base {
 
   /**
    * The member of this presence
-   * @type {?GuildMember}
+   * @type {?(GuildMember | PartialGuildMember)}
```
Qjuh, line 66: Here `this.member` should be turned into a property created from data passed in constructor instead of a getter. That way it could contain a partial guild member without putting that partial data into cache.

packages/discord.js/typings/index.d.ts
```diff
@@ -677,7 +677,7 @@ export class BaseGuildVoiceChannel extends GuildChannel {
   public bitrate: number;
   public get full(): boolean;
   public get joinable(): boolean;
-  public get members(): Collection<Snowflake, GuildMember>;
+  public get members(): Collection<Snowflake, GuildMember | PartialGuildMember>;
```
Qjuh, line 680: Since this is filtered from cache it won't contain `PartialGuildMember` anymore.

packages/discord.js/typings/index.d.ts
```diff
@@ -1541,7 +1541,10 @@ export class GuildBan extends Base {
 
 export abstract class GuildChannel extends BaseChannel {
   public constructor(guild: Guild, data?: RawGuildChannelData, client?: Client<true>, immediatePatch?: boolean);
-  private memberPermissions(member: GuildMember, checkAdmin: boolean): Readonly<PermissionsBitField>;
+  private memberPermissions(
+    member: GuildMember | PartialGuildMember,
```
Qjuh, line 1545: This doesn't work with a `PartialGuildMember`, since that won't have accurate `roles` data.

packages/discord.js/typings/index.d.ts
```diff
@@ -1550,7 +1553,7 @@ export abstract class GuildChannel extends BaseChannel {
   public guild: Guild;
   public guildId: Snowflake;
   public get manageable(): boolean;
-  public get members(): Collection<Snowflake, GuildMember>;
+  public get members(): Collection<Snowflake, GuildMember | PartialGuildMember>;
```
Qjuh, line 1556: See above

packages/discord.js/typings/index.d.ts
```diff
@@ -1565,7 +1568,10 @@ export abstract class GuildChannel extends BaseChannel {
   public edit(options: GuildChannelEditOptions): Promise<this>;
   public equals(channel: GuildChannel): boolean;
   public lockPermissions(): Promise<this>;
-  public permissionsFor(memberOrRole: GuildMember | Role, checkAdmin?: boolean): Readonly<PermissionsBitField>;
+  public permissionsFor(
```
Qjuh, line 1571: See above

packages/discord.js/typings/index.d.ts
```diff
@@ -2135,7 +2141,7 @@ export class Message<InGuild extends boolean = boolean> extends Base {
   public get hasThread(): boolean;
   public id: Snowflake;
   public interactionMetadata: MessageInteractionMetadata | null;
-  public get member(): GuildMember | null;
+  public get member(): GuildMember | PartialGuildMember | null;
```
Qjuh, line 2144: see above

packages/discord.js/typings/index.d.ts
```diff
@@ -2370,15 +2376,15 @@ export class MessageMentions<InGuild extends boolean = boolean> {
   );
   private _channels: Collection<Snowflake, Channel> | null;
   private readonly _content: string;
-  private _members: Collection<Snowflake, GuildMember> | null;
+  private _members: Collection<Snowflake, GuildMember | PartialGuildMember> | null;
```
Qjuh, line 2379: This is also filtered from cache, so no

packages/discord.js/typings/index.d.ts
```diff
@@ -2370,15 +2376,15 @@ export class MessageMentions<InGuild extends boolean = boolean> {
...
   private _parsedUsers: Collection<Snowflake, User> | null;
 
   public get channels(): Collection<Snowflake, Channel>;
   public readonly client: Client;
   public everyone: boolean;
   public readonly guild: If<InGuild, Guild>;
   public has(data: UserResolvable | RoleResolvable | ChannelResolvable, options?: MessageMentionsHasOptions): boolean;
-  public get members(): If<InGuild, Collection<Snowflake, GuildMember>>;
+  public get members(): If<InGuild, Collection<Snowflake, GuildMember | PartialGuildMember>>;
```
Qjuh, line 2387: See above

packages/discord.js/typings/index.d.ts
```diff
@@ -2693,7 +2699,7 @@ export class Presence extends Base {
   public activities: Activity[];
   public clientStatus: ClientPresenceStatusData | null;
   public guild: Guild | null;
-  public get member(): GuildMember | null;
+  public get member(): GuildMember | PartialGuildMember | null;
```
Qjuh, line 2702: This should be turned into an actual property, not a getter.

packages/discord.js/typings/index.d.ts
```diff
@@ -2803,7 +2809,7 @@ export class Role extends Base {
   public hoist: boolean;
   public id: Snowflake;
   public managed: boolean;
-  public get members(): Collection<Snowflake, GuildMember>;
+  public get members(): Collection<Snowflake, GuildMember | PartialGuildMember>;
```
Qjuh, line 2812: See above

packages/discord.js/typings/index.d.ts
```diff
@@ -3355,7 +3361,10 @@ export class ThreadChannel<ThreadOnly extends boolean = boolean> extends BaseCha
   public edit(options: ThreadEditOptions): Promise<this>;
   public join(): Promise<this>;
   public leave(): Promise<this>;
-  public permissionsFor(memberOrRole: GuildMember | Role, checkAdmin?: boolean): Readonly<PermissionsBitField>;
+  public permissionsFor(
```
Qjuh, line 3364: See above

packages/discord.js/typings/index.d.ts
```diff
@@ -3403,7 +3412,7 @@ export class Typing extends Base {
   public startedTimestamp: number;
   public get startedAt(): Date;
   public get guild(): Guild | null;
-  public get member(): GuildMember | null;
+  public get member(): GuildMember | PartialGuildMember | null;
```
Qjuh, line 3415: See above

packages/discord.js/typings/index.d.ts
```diff
@@ -3586,7 +3595,7 @@ export class VoiceState extends Base {
   public get deaf(): boolean | null;
   public guild: Guild;
   public id: Snowflake;
-  public get member(): GuildMember | null;
+  public get member(): GuildMember | PartialGuildMember | null;
```
Qjuh, line 3598: See above

packages/discord.js/typings/index.d.ts
```diff
@@ -4224,10 +4233,10 @@ export interface AddOrRemoveGuildMemberRoleOptions {
   reason?: string;
 }
 
-export class GuildMemberManager extends CachedManager<Snowflake, GuildMember, UserResolvable> {
+export class GuildMemberManager extends CachedManager<Snowflake, GuildMember | PartialGuildMember, UserResolvable> {
```
Qjuh, line 4236: `PartialGuildMember` should never be added to cache, that's the main purpose of this PR after all.

packages/discord.js/typings/index.d.ts
```diff
@@ -4224,10 +4233,10 @@ export interface AddOrRemoveGuildMemberRoleOptions {
...
   private constructor(guild: Guild, iterable?: Iterable<RawGuildMemberData>);
   public guild: Guild;
-  public get me(): GuildMember | null;
+  public get me(): GuildMember | PartialGuildMember | null;
```
Qjuh, line 4239: We always get the full `GuildMember` data for the `ClientUser` on `GUILD_CREATE`, so this can never be a `PartialGuildMember` to begin with.

packages/discord.js/typings/index.d.ts
```diff
@@ -4314,14 +4323,14 @@ export class GuildStickerManager extends CachedManager<Snowflake, Sticker, Stick
...
-  private constructor(member: GuildMember);
+  private constructor(member: GuildMember | PartialGuildMember);
   public get hoist(): Role | null;
   public get icon(): Role | null;
   public get color(): Role | null;
   public get highest(): Role;
   public get premiumSubscriberRole(): Role | null;
   public get botRole(): Role | null;
-  public member: GuildMember;
+  public member: GuildMember | PartialGuildMember;
```
Qjuh, line 4333: While technically correct this degrades UX. Probably better to add a `Partialize`d version of this manager overriding this property and overriding the `roles` getter on `PartialGuildMember` with that partial manager.

packages/discord.js/typings/index.d.ts
```diff
@@ -5959,7 +5968,15 @@ export interface GuildMemberEditOptions {
   reason?: string;
 }
 
-export type GuildResolvable = Guild | NonThreadGuildBasedChannel | GuildMember | GuildEmoji | Invite | Role | Snowflake;
+export type GuildResolvable =
+  | Guild
+  | NonThreadGuildBasedChannel
+  | GuildMember
+  | PartialGuildMember
```
Qjuh, line 5975: I don't think it makes much sense to resolve a Guild from a partial member, but that one's up to debate I'd say

Qjuh (changes requested):

packages/discord.js/typings/index.d.ts
Pavel-Boyazov: I looked at the wrong type, thanks

packages/discord.js/src/client/websocket/handlers/PRESENCE_UPDATE.js
Pavel-Boyazov: > The `guildMemberAvaliable` event here should only emit if the partial is set.
I think it also needs to be emit if member object cached before. Didn't?

packages/discord.js/typings/index.d.ts
Pavel-Boyazov: `GUILD_CREATE ` not emit if `Guilds` intent not set

packages/discord.js/src/client/websocket/handlers/PRESENCE_UPDATE.js
Qjuh: No, if it's cache before then the `!member` condition would be `false` anyway.

packages/discord.js/typings/index.d.ts
Qjuh: Without `Guilds` intent you wouldn't have any `GuildMembers` cached and the whole `GuildMemberManager` would be inaccessible because you don't have a `Guild` having it as property to begin with.

Pavel-Boyazov: What prevents us from using `fetch` to get and cache guilds with specific members exclude client member?

Pavel-Boyazov: https://github.com/discordjs/discord.js/blob/a30a749599382ba02554d1baa7f2e10bcf26e96e/packages/discord.js/src/managers/GuildMemberManager.js#L138-L145
And if it like you say, why exist this functionality?

Qjuh: Okay, that's a good catch indeed. Guess `GuildMemberManager#me` can be a `PartialGuildMember` after all, if the `GuildMemberManager` cache is limited to 0 without exception for the ClientUser.
But about no `Guilds` intent: most of the typings rely on that being present. Because several helper methods need it to work anyway.

packages/discord.js/typings/index.d.ts
Pavel-Boyazov: ~~How we can get actual property? Presence update not send actual member info, we can get this only for cache (which is almost the same as it's now)~~
UPD: Ok, I understand

packages/discord.js/typings/index.d.ts
```diff
@@ -4551,6 +4552,11 @@ export class GuildMemberRoleManager extends DataManager<Snowflake, Role, RoleRes
   ): Promise<GuildMember>;
 }
 
+export interface PartialGuildMemberRoleManager extends Partialize<GuildMemberRoleManager, null, null, 'member'> {
+  get partial(): true;
+  member: PartialGuildMember;
+}
```
Pavel-Boyazov, line 4630: I don't understand need I and how to override constructor

Pavel-Boyazov:
That's about it. If I forgot something, please remind me.

packages/discord.js/src/structures/Presence.js
```diff
@@ -56,6 +56,17 @@ class Presence extends Base {
...
+    if (this.guild) {
+      /**
+       * The partial member of this presence
+       *
+       * @type {?PartialGuildMember}
+       */
+      this.member = this.guild.members._add({ user: data.user, guild: data.guild }, false);
+    } else {
+      this.member = null;
+    }
```
Qjuh, line 68: I know that I was the one suggesting to turn `member` into a property in https://github.com/discordjs/discord.js/pull/10784#issuecomment-2692177930 but after reading this now again I believe this causes more issues than it solves. Instead we should add the `user` property like I suggested in that comment in the parantheses and keep `member` as a getter.

packages/discord.js/src/client/websocket/handlers/PRESENCE_UPDATE.js
```diff
@@ -20,11 +20,14 @@ module.exports = (client, { d: data }) => {
 
   let member = guild.members.cache.get(user.id);
   if (!member && data.status !== 'offline') {
```
Qjuh, line 22: I'd say this should either never emit here or at least check for `Partial.GuildMember` being set in `<Client>.options.partials`. We won't ever have a full GuildMember here.

packages/discord.js/src/managers/GuildMemberRoleManager.js
```diff
@@ -18,7 +18,7 @@ class GuildMemberRoleManager extends DataManager {
     /**
      * The GuildMember this manager belongs to
      *
-     * @type {GuildMember}
+     * @type {GuildMember | PartialGuildMember}
```
Qjuh, line 21: JSDoc has no notion of `PartialGuildMember`, since that only exists in TS typings.

packages/discord.js/src/managers/GuildMemberRoleManager.js
```diff
@@ -118,6 +118,16 @@ class GuildMemberRoleManager extends DataManager {
...
+  /**
+   * Whether this GuildMemberRoleManager is a partial
+   *
+   * @type {boolean}
+   * @readonly
+   */
+  get partial() {
+    return this.member.partial;
+  }
+
```
Qjuh, line 130: Since a `GuildMemberRoleManager` can only ever be accessed from the `GuildMember#roles` property of the partial `GuildMember` I'd say this getter is rather useless and redundant. you'd check the `GuildMember#partial` directly instead.

packages/discord.js/typings/index.d.ts
```diff
@@ -4623,6 +4624,11 @@ export class GuildMemberRoleManager extends DataManager<Snowflake, Role, RoleRes
   ): Promise<GuildMember>;
 }
 
+export interface PartialGuildMemberRoleManager extends Partialize<GuildMemberRoleManager, null, null, 'member'> {
```
Qjuh, line 4627: This doesn't need the Partialize, since you only narrow the type of the getter here.
```suggestion
export interface PartialGuildMemberRoleManager extends GuildMemberRoleManager {
```

packages/discord.js/typings/index.d.ts
```diff
@@ -4921,6 +4927,7 @@ export interface AddGuildMemberOptions {
 export type AllowedPartial =
   | Channel
   | GuildMember
+  | GuildMemberRoleManager
```
Qjuh, line 4930: This isn't needed then either.

packages/discord.js/typings/index.d.ts
```diff
@@ -6949,7 +6957,15 @@ export interface PartialDMChannel extends Partialize<DMChannel, null, null, 'las
...
 }
 
-export interface PartialGuildMember extends Partialize<GuildMember, 'joinedAt' | 'joinedTimestamp' | 'pending'> {}
+export interface PartialGuildMember extends Partialize<
+  GuildMember,
+  'joinedAt' | 'joinedTimestamp' | 'pending',
+  null,
+  'roles'
+> {
+  get partial(): true;
```
Qjuh, line 6966: That's already done by `Partialize<>`
```suggestion
```

packages/discord.js/typings/index.d.ts
```diff
@@ -4606,6 +4606,7 @@ export class GuildMemberRoleManager extends DataManager<Snowflake, Role, RoleRes
   public get highest(): Role;
   public get premiumSubscriberRole(): Role | null;
   public get botRole(): Role | null;
+  public get partial(): false;
```
Qjuh, line 4609: Not needed, see other comment.
```suggestion
```

Qjuh (changes requested):

packages/discord.js/typings/index.d.ts
Pavel-Boyazov: Why? `GuildMemberRoleManager` is the base for `PartialGuildMemberRoleManager`, like `Poll` for `PartialPoll`, `PollAnswer` for `PartialPollAnswer` and other

packages/discord.js/typings/index.d.ts
Pavel-Boyazov: What getter? We specify member property without getter

Qjuh: You are right, I meant the type of the property.

packages/discord.js/typings/index.d.ts
Qjuh: Because `AllowedPartial` is only used as a type for Partialize. Since we don't need Partialize we don't need to add that here.

packages/discord.js/typings/index.d.ts
Pavel-Boyazov: > You are right, I meant the type of the property.

But member in `GuildMemberRoleManager` has different type and we get error without Partialize. I should to bypass it with Omit or extend type of `<GuildMemberRoleManager>.member` if not use Partialize?

PR #10786, opened 2025-03-01
feat(Typing): cache member object

Pavel-Boyazov:
Discord API [provides member object](https://discord.com/developers/docs/events/gateway-events#typing-start) with `TYPING_START` event in guild. We can use this for cache update and getting value without getter

### DIFF
--- a/packages/discord.js/src/structures/Typing.js
+++ b/packages/discord.js/src/structures/Typing.js
@@ -33,6 +33,14 @@ class Typing extends Base {
        */
       this.startedTimestamp = data.timestamp * 1_000;
     }
+
+    if ('member' in data) {
+      /**
+       * The member that this guild typing instance represents
+       * @type {?GuildMember}
+       */
+      this.member = this.channel.guild.members._add(data.member, true);
+    }
   }
 
   /**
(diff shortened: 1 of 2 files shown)


packages/discord.js/src/structures/Typing.js
```diff
@@ -33,6 +33,14 @@ class Typing extends Base {
...
       this.startedTimestamp = data.timestamp * 1_000;
     }
+
+    if ('member' in data) {
+      /**
+       * The member that this guild typing instance represents
+       * @type {?GuildMember}
+       */
+      this.member = this.channel.guild.members._add(data.member, true);
+    }
```
vladfrangu, line 43: This PR makes the member only set when its provided, otherwise it will be undefined (not null). Either we keep the getter and add the caching logic in too, or we define a fallback that is the getter

vladfrangu (changes requested):

packages/discord.js/src/structures/Typing.js
Pavel-Boyazov: This can be done without getters. If the `member` property was not set when the object was created, then there can be no conditions when it will be set later (except manually). I can use the ternary operator to set null if there is no data.
What do you think?

Qjuh: https://github.com/discordjs/discord.js/blob/d1f56ffb2a663ec654d7457489b22e2c401669eb/packages/discord.js/src/client/actions/TypingStart.js#L17 already adds the member to cache, so this is redundant. The getter was fine as is.

Qjuh:
https://github.com/discordjs/discord.js/pull/10786#discussion_r1976598318 still stands, this PR is adding nothing, the cache already gets updated and the getter is fine.

PR #10788, opened 2025-03-03
feat: components v2 in builders 

vladfrangu:

🤫 

	- This PR includes breaking changes (methods removed or renamed, parameters moved or removed) [tbd]

### DIFF
--- a/packages/builders/src/components/Component.ts
+++ b/packages/builders/src/components/Component.ts
@@ -9,6 +9,24 @@ import type { APIBaseComponent, ComponentType } from 'discord-api-types/v10';
 export abstract class ComponentBuilder<Component extends APIBaseComponent<ComponentType>>
 	implements JSONEncodable<Component>
 {
+	protected abstract readonly data: { id?: number | undefined };
+
+	/**
+	 * Sets the id of this component.
+	 *
+	 * @param id - The id to use
+	 */
+	public setId(id: number) {
+		this.data.id = id;
+	}
+
+	/**
+	 * Clears the id of this component, defaulting to a default incremented id.
+	 */
+	public clearId() {
+		this.data.id = undefined;
+	}
+
 	/**
 	 * Serializes this builder to API-compatible JSON data.
 	 *
--- a/packages/builders/src/components/Component.ts
+++ b/packages/builders/src/components/Component.ts
@@ -1,17 +1,19 @@
 import type { JSONEncodable } from '@discordjs/util';
-import type { APIActionRowComponent, APIComponentInActionRow } from 'discord-api-types/v10';
+import type { APIActionRowComponent, APIComponentInActionRow, APIMessageComponent } from 'discord-api-types/v10';
 
 /**
  * Any action row component data represented as an object.
  */
-export type AnyAPIActionRowComponent = APIActionRowComponent<APIComponentInActionRow> | APIComponentInActionRow;
+export type AnyMessageComponent =
+	// this action row type represents modals as wel
+	APIActionRowComponent<APIActionRowComponentTypes> | APIActionRowComponentTypes | APIMessageComponent;
 
 /**
  * The base component builder that contains common symbols for all sorts of components.
  *
  * @typeParam Component - The type of API data that is stored within the builder
  */
-export abstract class ComponentBuilder<Component extends AnyAPIActionRowComponent> implements JSONEncodable<Component> {
+export abstract class ComponentBuilder<Component extends AnyMessageComponent> implements JSONEncodable<Component> {
 	/**
 	 * Serializes this builder to API-compatible JSON data.
 	 *
--- a/packages/builders/src/components/Component.ts
+++ b/packages/builders/src/components/Component.ts
@@ -1,6 +1,10 @@
 import type { JSONEncodable } from '@discordjs/util';
 import type { APIBaseComponent, ComponentType } from 'discord-api-types/v10';
 
+export interface ComponentBuilderBaseData {
+	id?: number | undefined;
+}
+
 /**
  * The base component builder that contains common symbols for all sorts of components.
  *
@@ -9,7 +13,7 @@ import type { APIBaseComponent, ComponentType } from 'discord-api-types/v10';
 export abstract class ComponentBuilder<Component extends APIBaseComponent<ComponentType>>
 	implements JSONEncodable<Component>
 {
-	protected abstract readonly data: { id?: number | undefined };
+	protected abstract readonly data: ComponentBuilderBaseData;
 
 	/**
 	 * Sets the id of this component.
--- a/packages/builders/src/components/Component.ts
+++ b/packages/builders/src/components/Component.ts
@@ -6,7 +6,7 @@ import type { APIActionRowComponent, APIComponentInActionRow, APIMessageComponen
  */
 export type AnyMessageComponent =
 	// this action row type represents modals as wel
-	APIActionRowComponent<APIActionRowComponentTypes> | APIActionRowComponentTypes | APIMessageComponent;
+	APIActionRowComponent<APIComponentInActionRow> | APIComponentInActionRow | APIMessageComponent;
 
 /**
  * The base component builder that contains common symbols for all sorts of components.
--- a/packages/builders/src/components/Component.ts
+++ b/packages/builders/src/components/Component.ts
@@ -1,19 +1,14 @@
 import type { JSONEncodable } from '@discordjs/util';
-import type { APIActionRowComponent, APIComponentInActionRow, APIMessageComponent } from 'discord-api-types/v10';
-
-/**
- * Any action row component data represented as an object.
- */
-export type AnyMessageComponent =
-	// this action row type represents modals as wel
-	APIActionRowComponent<APIComponentInActionRow> | APIComponentInActionRow | APIMessageComponent;
+import type { APIBaseComponent, ComponentType } from 'discord-api-types/v10';
 
 /**
  * The base component builder that contains common symbols for all sorts of components.
  *
  * @typeParam Component - The type of API data that is stored within the builder
  */
-export abstract class ComponentBuilder<Component extends AnyMessageComponent> implements JSONEncodable<Component> {
+export abstract class ComponentBuilder<Component extends APIBaseComponent<ComponentType>>
+	implements JSONEncodable<Component>
+{
 	/**
 	 * Serializes this builder to API-compatible JSON data.
 	 *
--- a/packages/builders/src/components/Component.ts
+++ b/packages/builders/src/components/Component.ts
@@ -22,6 +22,7 @@ export abstract class ComponentBuilder<Component extends APIBaseComponent<Compon
 	 */
 	public setId(id: number) {
 		this.data.id = id;
+		return this;
 	}
 
 	/**
@@ -29,6 +30,7 @@ export abstract class ComponentBuilder<Component extends APIBaseComponent<Compon
 	 */
 	public clearId() {
 		this.data.id = undefined;
+		return this;
 	}
 
 	/**
--- a/packages/builders/src/components/selectMenu/BaseSelectMenu.ts
+++ b/packages/builders/src/components/selectMenu/BaseSelectMenu.ts
@@ -11,9 +11,10 @@ export abstract class BaseSelectMenuBuilder<Data extends APISelectMenuComponent>
 	extends ComponentBuilder<Data>
 	implements JSONEncodable<APISelectMenuComponent>
 {
-	protected abstract readonly data: Partial<
+	protected abstract override readonly data: Partial<
 		Pick<Data, 'custom_id' | 'disabled' | 'max_values' | 'min_values' | 'placeholder'>
-	>;
+	> &
+		Pick<Data, 'id'>;
 
 	/**
 	 * Sets the placeholder for this select menu.
--- a/packages/builders/src/components/selectMenu/BaseSelectMenu.ts
+++ b/packages/builders/src/components/selectMenu/BaseSelectMenu.ts
@@ -12,9 +12,8 @@ export abstract class BaseSelectMenuBuilder<Data extends APISelectMenuComponent>
 	implements JSONEncodable<APISelectMenuComponent>
 {
 	protected abstract override readonly data: Partial<
-		Pick<Data, 'custom_id' | 'disabled' | 'max_values' | 'min_values' | 'placeholder'>
-	> &
-		Pick<Data, 'id'>;
+		Pick<Data, 'custom_id' | 'disabled' | 'id' | 'max_values' | 'min_values' | 'placeholder'>
+	>;
 
 	/**
 	 * Sets the placeholder for this select menu.
--- a/packages/builders/src/components/selectMenu/StringSelectMenuOption.ts
+++ b/packages/builders/src/components/selectMenu/StringSelectMenuOption.ts
@@ -7,7 +7,7 @@ import { selectMenuStringOptionPredicate } from '../Assertions.js';
  * A builder that creates API-compatible JSON data for string select menu options.
  */
 export class StringSelectMenuOptionBuilder implements JSONEncodable<APISelectMenuOption> {
-	private readonly data: Partial<APISelectMenuOption>;
+	protected readonly data: Partial<APISelectMenuOption>;
 
 	/**
 	 * Creates a new string select menu option from API data.
--- a/packages/builders/src/components/selectMenu/StringSelectMenuOption.ts
+++ b/packages/builders/src/components/selectMenu/StringSelectMenuOption.ts
@@ -7,7 +7,7 @@ import { selectMenuStringOptionPredicate } from '../Assertions.js';
  * A builder that creates API-compatible JSON data for string select menu options.
  */
 export class StringSelectMenuOptionBuilder implements JSONEncodable<APISelectMenuOption> {
-	protected readonly data: Partial<APISelectMenuOption>;
+	private readonly data: Partial<APISelectMenuOption>;
 
 	/**
 	 * Creates a new string select menu option from API data.
--- a/packages/builders/src/components/v2/Assertions.ts
+++ b/packages/builders/src/components/v2/Assertions.ts
@@ -6,12 +6,26 @@ const unfurledMediaItemPredicate = z.object({
 		.string()
 		.url()
 		.refine(refineURLPredicate(['http:', 'https:', 'attachment:']), {
-			message: 'Invalid protocol for icon URL. Must be http:, https:, or attachment:',
+			message: 'Invalid protocol for media URL. Must be http:, https:, or attachment:',
 		}),
 });
 
 export const thumbnailPredicate = z.object({
 	media: unfurledMediaItemPredicate,
 	description: z.string().min(1).max(1_024).nullish(),
+	spoiler: z.boolean().optional(),
+});
+
+const unfurledMediaItemAttachmentOnlyPredicate = z.object({
+	url: z
+		.string()
+		.url()
+		.refine(refineURLPredicate(['attachment:']), {
+			message: 'Invalid protocol for file URL. Must be attachment:',
+		}),
+});
+
+export const filePredicate = z.object({
+	file: unfurledMediaItemAttachmentOnlyPredicate,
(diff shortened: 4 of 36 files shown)


packages/builders/src/components/v2/Section.ts
```diff
@@ -0,0 +1,206 @@
...
+
+	public get components(): readonly TextDisplayBuilder[] {
+		return this.data.components;
+	}
+
+	public constructor({ components = [], accessory, ...rest }: Partial<APISectionComponent> = {}) {
+		super();
+		this.data = {
+			...structuredClone(rest),
+			// Nasty use of `!` here, but we will validate this in toJSON anyways
```
didinele, line 45: Yeah I don't like this. The pattern I adapt generally is making the prop optional in the `XData` interface and casting in `validate`

packages/builders/src/components/v2/MediaGalleryItem.ts
```diff
@@ -0,0 +1,62 @@
...
+		this.data.description = undefined;
+		return this;
+	}
+
+	/**
+	 * Sets the spoiler status of this thumbnail.
+	 *
+	 * @param spoiler - The spoiler status to use
+	 */
+	public setSpoiler(spoiler: boolean) {
```
didinele, line 48: Might've missed other places but this one jumped to me, `setX` methods for bools should almost always be `= true`

didinele (changes requested):
didn't look at your tests; otherwise, so far mostly LGTM

packages/builders/src/components/v2/Section.ts
didinele: I mean, I'd say "we assert something that's not true" is much less true when it's just the cast after `validate`, given it'll throw if our data doesn't match the type we cast to.

packages/builders/src/components/v2/Thumbnail.ts
```diff
@@ -0,0 +1,66 @@
+import type { APIThumbnailComponent } from 'discord-api-types/v10';
+import { ComponentType } from 'discord-api-types/v10';
+import { validate } from '../../util/validation.js';
+import { ComponentBuilder } from '../Component.js';
+import { thumbnailPredicate } from './Assertions.js';
+
+export class ThumbnailBuilder extends ComponentBuilder<APIThumbnailComponent> {
+	private readonly data: Partial<APIThumbnailComponent>;
+
```
Qjuh, line 9:
```suggestion

	/**
	 * Creates a new thumbnail from API data.
	 *
	 * @param data - The API data to create this thumbnail with
	 * @example
	 * Creating a thumbnail from an API data object:
	 * ```ts
	 * const thumbnaik = new ThumbnailBuilder({
	 * 	description: 'some text',
	 *  media: {
	 *      url: 'https://cdn.discordapp.com/embed/assets/4.png',
	 *  },
	 * });
	 * ```
	 * @example
	 * Creating a thumbnail using setters and API data:
	 * ```ts
	 * const thumbnail = new ThumbnailBuilder({
	 * 	media: {
	 *      url: 'attachment://image.png',
	 *  },
	 * })
	 * 	.setDescription('alt text');
	 * ```
	 */
```

packages/builders/src/components/v2/Separator.ts
```diff
@@ -0,0 +1,55 @@
...
+	 */
+	public setSpacing(spacing: SeparatorSpacingSize) {
+		this.data.spacing = spacing;
+		return this;
+	}
+
+	/**
+	 * Clears the spacing of this separator.
+	 */
+	public clearSpacing() {
```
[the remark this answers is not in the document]
didinele: Yeah, the convention is that all optional props should have a `clearX` method, generally.

packages/builders/src/components/v2/Assertions.ts
```diff
@@ -0,0 +1,78 @@
...
+				mediaGalleryPredicate,
+				sectionPredicate,
+				separatorPredicate,
+				textDisplayPredicate,
+			]),
+		)
+		.min(1)
+		.max(10),
+	spoiler: z.boolean().optional(),
+	accent_color: z.number().int().min(0).max(0xffffff).nullish(),
```
Qjuh, line 77: numbers are usually `.optional()` not `.nullish()` in other places, `Embed#color` or `BaseSelectMenu#min_values` for example.
```suggestion
	accent_color: z.number().int().min(0).max(0xffffff).optional(),
```

packages/builders/src/components/v2/Container.ts
```diff
@@ -0,0 +1,184 @@
...
+			APITextDisplayComponent | TextDisplayBuilder | ((builder: TextDisplayBuilder) => TextDisplayBuilder)
+		>
+	): this {
+		const normalized = normalizeArray(input);
+		const resolved = normalized.map((component) => resolveBuilder(component, TextDisplayBuilder));
+
+		this.data.components.push(...resolved);
+		return this;
+	}
+
```
Qjuh, line 171:
```suggestion

	/**
	 * Removes, replaces, or inserts components for this container
	 *
	 * @remarks
	 * This method behaves similarly
	 * to {@link https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/splice | Array.prototype.splice()}.
	 *
	 * It's useful for modifying and adjusting order of the already-existing components of a container.
	 * @example
	 * Remove the first component:
	 * ```ts
	 * container.spliceComponents(0, 1);
	 * ```
	 * @example
	 * Remove the first n components:
	 * ```ts
	 * const n = 4;
	 * container.spliceComponents(0, n);
	 * ```
	 * @example
	 * Remove the last component:
	 * ```ts
	 * container.spliceComponents(-1, 1);
	 * ```
	 * @param index - The index to start at
	 * @param deleteCount - The number of components to remove
	 * @param components - The replacing component objects
	 */
	public spliceComponents(index: number, deleteCount: number, ...components: AnyActionRowComponentBuilder[]): this {
		this.data.components.splice(index, deleteCount, ...components);
		return this;
	}

	/**
	 * {@inheritDoc ComponentBuilder.toJSON}
	 */
```

Qjuh (dismissed):

packages/builders/src/components/selectMenu/BaseSelectMenu.ts
```diff
@@ -11,9 +11,10 @@ export abstract class BaseSelectMenuBuilder<Data extends APISelectMenuComponent>
 	extends ComponentBuilder<Data>
 	implements JSONEncodable<APISelectMenuComponent>
 {
-	protected abstract readonly data: Partial<
+	protected abstract override readonly data: Partial<
 		Pick<Data, 'custom_id' | 'disabled' | 'max_values' | 'min_values' | 'placeholder'>
-	>;
+	> &
+		Pick<Data, 'id'>;
```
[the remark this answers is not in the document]
vladfrangu: Had issues due to partial, will have to recheck again if it works

packages/builders/src/components/v2/Assertions.ts
vladfrangu: The type says nullish, I respect the types

Qjuh: Oh, true, it does indeed behave differently than `Embed#color`. Disregard then.

packages/builders/src/components/v2/Separator.ts
```diff
@@ -0,0 +1,76 @@
...
+			type: ComponentType.Separator,
+		};
+	}
+
+	/**
+	 * Sets whether this separator should show a divider line.
+	 *
+	 * @param divider - Whether to show a divider line
+	 */
+	public setDivider(divider: boolean) {
```
didinele, line 44:
```suggestion
	public setDivider(divider = true) {
```

packages/builders/src/interactions/modals/Modal.ts
```diff
@@ -24,7 +24,7 @@ export class ModalBuilder implements JSONEncodable<APIModalInteractionResponseCa
 	/**
 	 * The API data associated with this modal.
 	 */
-	private readonly data: ModalBuilderData;
+	protected readonly data: ModalBuilderData;
```
didinele, line 27: again, this doesn't extend ComponentBuilder, does it need to be protected?

packages/builders/src/messages/embed/Embed.ts
```diff
@@ -25,7 +25,7 @@ export class EmbedBuilder implements JSONEncodable<APIEmbed> {
 	/**
 	 * The API data associated with this embed.
 	 */
-	private readonly data: EmbedBuilderData;
+	protected readonly data: EmbedBuilderData;
```
didinele, line 28: ditto

packages/builders/src/messages/embed/EmbedAuthor.ts
```diff
@@ -6,7 +6,7 @@ import { embedAuthorPredicate } from './Assertions.js';
  * A builder that creates API-compatible JSON data for the embed author.
  */
 export class EmbedAuthorBuilder {
-	private readonly data: Partial<APIEmbedAuthor>;
+	protected readonly data: Partial<APIEmbedAuthor>;
```
didinele, line 9: ditto

packages/builders/src/messages/embed/EmbedField.ts
```diff
@@ -6,7 +6,7 @@ import { embedFieldPredicate } from './Assertions.js';
  * A builder that creates API-compatible JSON data for embed fields.
  */
 export class EmbedFieldBuilder {
-	private readonly data: Partial<APIEmbedField>;
+	protected readonly data: Partial<APIEmbedField>;
```
didinele, line 9: ditto

packages/builders/src/messages/embed/EmbedFooter.ts
```diff
@@ -6,7 +6,7 @@ import { embedFooterPredicate } from './Assertions.js';
  * A builder that creates API-compatible JSON data for the embed footer.
  */
 export class EmbedFooterBuilder {
-	private readonly data: Partial<APIEmbedFooter>;
+	protected readonly data: Partial<APIEmbedFooter>;
```
didinele, line 9: ditto

didinele (changes requested):

packages/builders/src/components/selectMenu/StringSelectMenuOption.ts
```diff
@@ -7,7 +7,7 @@ import { selectMenuStringOptionPredicate } from '../Assertions.js';
  * A builder that creates API-compatible JSON data for string select menu options.
  */
 export class StringSelectMenuOptionBuilder implements JSONEncodable<APISelectMenuOption> {
-	private readonly data: Partial<APISelectMenuOption>;
+	protected readonly data: Partial<APISelectMenuOption>;
```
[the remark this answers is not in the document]
Qjuh: Because it extends the data on ComponentBuilder now (asked the same before)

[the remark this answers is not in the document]
vladfrangu: Because I found all instances and replaced them 💀, i can undo it in certain cases.

packages/builders/src/components/Component.ts
```diff
@@ -1,17 +1,32 @@
...
 /**
  * The base component builder that contains common symbols for all sorts of components.
  *
  * @typeParam Component - The type of API data that is stored within the builder
  */
-export abstract class ComponentBuilder<Component extends AnyAPIActionRowComponent> implements JSONEncodable<Component> {
+export abstract class ComponentBuilder<Component extends APIBaseComponent<ComponentType>>
+	implements JSONEncodable<Component>
+{
+	protected abstract readonly data: { id?: number | undefined };
```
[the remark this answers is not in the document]
vladfrangu: Sure, but in this case all types have this field anyways

packages/builders/src/messages/Assertions.ts
```diff
@@ -73,5 +82,47 @@ export const messagePredicate = z
...
+		{ message: 'Messages must have content, embeds, a poll, attachments, components OR stickers' },
 	);
+
+const allTopLevelComponentsPredicate = z
+	.union([
+		basicActionRowPredicate,
+		z.object({
+			type: z.union([
+				// Components v1
+				z.literal(ComponentType.Button),
```
Qjuh, line 94: These can't exist as top level components, only in ActionRows or as accessory of a Section. SelectMenus can only exist in ActionRows.

packages/builders/src/messages/Assertions.ts
```diff
@@ -73,5 +82,47 @@ export const messagePredicate = z
...
+	.max(10);
+
+const messageComponentsV2Predicate = baseMessagePredicate.extend({
+	components: allTopLevelComponentsPredicate,
+	flags: z.number().refine((flags) => {
+		return (flags & MessageFlags.IsComponentsV2) === MessageFlags.IsComponentsV2;
+	}),
+	// These fields cannot be set
+	content: z.string().length(0).nullish(),
+	embeds: z.array(embedPredicate).max(0).nullish(),
```
Qjuh, line 123:
```suggestion
	embeds: z.array(z.never()).nullish(),
```

packages/builders/src/messages/Assertions.ts
```diff
@@ -73,5 +82,47 @@ export const messagePredicate = z
...
+	sticker_ids: z.array(z.string()).max(0).nullish(),
```
Qjuh, line 124:
```suggestion
	sticker_ids: z.array(z.never()).nullish(),
```

packages/builders/src/messages/Message.ts
```diff
@@ -271,7 +273,7 @@ export class MessageBuilder implements JSONEncodable<RESTPostAPIChannelMessageJS
 	public addComponents(
```
Qjuh, line 273:
```suggestion
	public addActionRowComponents(
```

and then needs more methods for the other top-level component types

packages/builders/src/messages/Message.ts
```diff
@@ -31,7 +33,7 @@ export interface MessageBuilderData
 	> {
 	allowed_mentions?: AllowedMentionsBuilder;
 	attachments: AttachmentBuilder[];
-	components: ActionRowBuilder[];
+	components: MessageComponentBuilder[];
```
Qjuh, line 36: This also includes Buttons and SelectMenus. Those are not top-level components. We need another `MessageTopLevelComponentBuilder` type without those for this.

packages/builders/src/messages/Message.ts
```diff
@@ -335,7 +337,7 @@ export class MessageBuilder implements JSONEncodable<RESTPostAPIChannelMessageJS
 	public setComponents(
 		...components: RestOrArray<
 			| ActionRowBuilder
-			| APIActionRowComponent<APIMessageActionRowComponent>
+			| APIActionRowComponent<APIComponentInMessageActionRow>
```
Qjuh, line 340: Either needs to accept all top-level component types (ugly) or simply be removed. Neither `ActionRowBuilder` nor `ChatInputCommandBuilder` have a `set` method for "replacing" the existing components/options. That can already be achieved with `spliceComponents()` after all.

Qjuh (changes requested):

packages/builders/src/components/v2/Container.ts
```diff
@@ -0,0 +1,230 @@
...
+	/**
+	 * Adds media gallery components to this container.
+	 *
+	 * @param input - The media gallery components to add
+	 */
+	public addMediaGalleryComponents(
+		...input: RestOrArray<MediaGalleryBuilder | ((builder: MediaGalleryBuilder) => MediaGalleryBuilder)>
```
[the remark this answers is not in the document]
vladfrangu: Yes, I messed it up lol. Good catch with this and the ret, tyyy

packages/builders/src/messages/Assertions.ts
vladfrangu: Wouldn't max(0) still be needed

packages/builders/src/messages/Message.ts
vladfrangu: I could see us supporting a setComponents with all types but I feel like that's just horrible, and should probably stick to add (maybe add in a clear method so you can start with a blank slate)?

But I'd ask @didinele's thoughts too on this. I could easily implement a setComponents i think

packages/builders/src/messages/Assertions.ts
didinele: Doubt it, you can't pass a value that satisfies `never`, so the predicate should insta reject on the first value passed in.

packages/builders/src/messages/Message.ts
didinele: Sorry, I'm a bit lost on what the problem is here

Qjuh: The problem is that this is a generic "set the components array to this arbitrary list of components" when we have specific `addXComponent` methods for all the other ways. As already stated we don't have a set method in ActionRowBuilder or ChatInputCommandBuilder or ContainerBuilder or SectionBuilder. So why would we suddenly have one in MessageBuilder?

didinele: I see. I added it with the thought that there's only ActionRows so it sort of made sense. I'm for removing it in this case.

packages/builders/src/messages/Assertions.ts
```diff
@@ -27,40 +27,49 @@ export const messageReferencePredicate = z.object({
...
-		enforce_nonce: z.boolean().optional(),
 		poll: pollPredicate.optional(),
+		components: basicActionRowPredicate.array().max(5).optional(),
+		flags: z
+			.number()
+			.optional()
+			.refine((flags) => {
+				// If we have flags, ensure we don't have the ComponentsV2 flag
+				if (flags) {
+					return (flags & MessageFlags.IsComponentsV2) !== MessageFlags.IsComponentsV2;
```
Jiralite, line 68: This feels more idiomatic as it is not an inverse check:

```suggestion
					return (flags & MessageFlags.IsComponentsV2) === 0
```

Qjuh (approved):

packages/builders/src/components/v2/MediaGallery.ts
```diff
@@ -0,0 +1,118 @@
...
+
+	/**
+	 * {@inheritDoc ComponentBuilder.toJSON}
+	 */
+	public override toJSON(validationOverride?: boolean): APIMediaGalleryComponent {
+		const { items, ...rest } = this.data;
+
+		const data = {
+			...structuredClone(rest),
+			items: items.map((item) => item.toJSON(validationOverride)),
```
didinele, line 111:
```suggestion
			items: items.map((item) => item.toJSON(false)),
```

Whenever sub-items are defined in the root predicate, we pass `false`


https://github.com/discordjs/discord.js/blob/12d2d12262d1195599329b7b1f9e7ceee5696084/packages/builders/src/components/v2/Assertions.ts#L51

packages/builders/src/components/v2/Container.ts
```diff
@@ -0,0 +1,232 @@
...
+
+	/**
+	 * {@inheritDoc ComponentBuilder.toJSON}
+	 */
+	public override toJSON(validationOverride?: boolean): APIContainerComponent {
+		const { components, ...rest } = this.data;
+
+		const data = {
+			...structuredClone(rest),
+			components: components.map((component) => component.toJSON(validationOverride)),
```
didinele, line 225:
```suggestion
			components: components.map((component) => component.toJSON(false)),
```

packages/builders/src/components/v2/Section.ts
```diff
@@ -0,0 +1,257 @@
...
+
+	/**
+	 * {@inheritDoc ComponentBuilder.toJSON}
+	 */
+	public override toJSON(validationOverride?: boolean): APISectionComponent {
+		const { components, accessory, ...rest } = this.data;
+
+		const data = {
+			...structuredClone(rest),
+			components: components.map((component) => component.toJSON(validationOverride)),
```
didinele, line 249:
```suggestion
			components: components.map((component) => component.toJSON(false)),
```

`accessory` is correct since `sectionPredicate`'s `accessory` only validates the `type`

didinele (approved):

Jiralite (approved):

PR #10791, opened 2025-03-04
feat: override `groupBy` to return `Collection`

sdanialraza:
The `groupBy` method inherited would return a `Map`. This overrides it to return a `Collection` instead.


vladfrangu (approved):

almeidx (approved):

packages/collection/src/collection.ts
```diff
@@ -1088,6 +1088,17 @@ export class Collection<Key, Value> extends Map<Key, Value> {
...
 
+	/**
+	 * Identical to {@link https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map/groupBy | Map.groupBy()}
+	 * but returns a Collection instead of a Map.
+	 */
+	public static override groupBy<Key, Item>(
+		items: Iterable<Item>,
+		keySelector: (item: Item, index: number) => Key,
+	): Collection<Key, Item[]> {
+		return new this[Symbol.species]<Key, Item[]>(Map.groupBy(items, keySelector));
```
Renegade334, line 1099: This creates a new throwaway Map, then immediately copies all its contents into a new Collection.

Would it not make more sense just to alter the prototype of the generated Map object?

PR #10797, opened 2025-03-09
fix: 🔧 don't error out if no options are provided

jsJack:
This commit stops calls to `options.withResponse`, etc erroring out when `interaction.update();` is called alone with no params.

This allows users to do `await interaction.update()` on something like a `StringSelectMenuInteraction` where the goal isn't to update the message in any way, just to send a blank object to refresh the message state and reset the Select Menu so it shows the placeholder again and not the last-selected `StringSelectMenuOption`.

### DIFF
--- a/packages/discord.js/src/structures/interfaces/InteractionResponses.js
+++ b/packages/discord.js/src/structures/interfaces/InteractionResponses.js
@@ -242,7 +242,7 @@ class InteractionResponses {
    *   .then(console.log)
    *   .catch(console.error);
    */
-  async update(options) {
+  async update(options = {}) {
     if (this.deferred || this.replied) throw new DiscordjsError(ErrorCodes.InteractionAlreadyReplied);
 
     let messagePayload;


jsJack:
Apologies if types aren't correct on last 2 commits, I don't do loads of Typescript.

vladfrangu:
Is there any reason you use update _without_ any data instead of deferUpdate? It'd replicate the same behavior tmk

jsJack:
@vladfrangu - From my testing, passing `update()` with no params and `deferUpdate()` displayed different behaviour.

Once of them completed the interaction and left the Select Menu in it's just-used state with the last option selected.
The other (`update()`) refreshed the Message State and reset the select menu, showing the placeholder again.

I needed the select menu to reset because it's a 'control panel' for tickets which allows you to reuse the most common actions a few times without having to refresh Discord ideally.

If you'd like I can record a video showing the differences.

almeidx:
> Once of them completed the interaction and left the Select Menu in it's just-used state with the last option selected.
> The other (`update()`) refreshed the Message State and reset the select menu, showing the placeholder again.

This is indeed the behaviour I'm seeing too.

PS: `update()` resets the state back to what it was before, so either the placeholder or the default value, if there is one.

almeidx (approved):

Jiralite (approved):

sdanialraza (approved):

vladfrangu (approved):

PR #10798, opened 2025-03-10
fix: invalid token wrong recognition

josejooj:
The following code does not correctly handle cases where the bot token is invalid: https://github.com/discordjs/discord.js/blob/ab6a69401e7de81a7c619d82b9f00741dfae3af4/packages/rest/src/lib/handlers/Shared.ts#L141-L143

Other requests may return a 401 status, which does not mean that the token is actually invalid. The following request is an example, where the webhook token is invalid, not the bot's token:

```js
DiscordAPIError[50027]: Invalid Webhook Token
    at handleErrors (/application/node_modules/@discordjs/rest/dist/index.js:727:13)
    at process.processTicksAndRejections (node:internal/process/task_queues:105:5)
    at async SequentialHandler.runRequest (/application/node_modules/@discordjs/rest/dist/index.js:1128:23)
    at async SequentialHandler.queueRequest (/application/node_modules/@discordjs/rest/dist/index.js:959:14)
    at async _REST.request (/application/node_modules/@discordjs/rest/dist/index.js:1272:22)
    at async Object.f (/application/dist/util/tasks/carrinhos/treatments/cancelled/cancelled.js:4:320)
    at async C (/application/dist/util/tasks/carrinhos/main.js:1:2251) {
  requestBody: { files: undefined, json: { embeds: [Array], components: [] } },
  rawError: { message: 'Invalid Webhook Token', code: 50027 },
  code: 50027,
  status: 401,
  method: 'PATCH',
  url: 'https://discord.com/api/v10/webhooks/1332931122040143913/aW50ZXJhY3Rpb246MTM0ODY5NDU4MTYyNTIyOTQyNDozODF5NlBJTHYxZ09XWWtjdTl5ZlhCNU5qU2E0djI5SXhWV3psUUk0SFlKaXY4cmh0VXRoY1NKQjRKdmxIcmVEMDhEeG5WZ2pXUmNKaUc5Y1dXaW03WnlKYVJOU1RhaHJsSXZmdExJbTRxMjE1V1JBY2t5NE4yYTF0UDRBdU12YQ/messages/@original'
}
```

Remembering that you don't necessarily need to have a webhook to receive this error, it just takes a long time to respond to an interaction, so this wrong condition ends up affecting several bots, especially in Discord API instabilities.

This commit fixes invalid token wrong detection cases like these, only handling the case where the token is actually invalid, which is when receiving code 0 from the Discord API:

```json
{ "message": "401: Unauthorized", "code": 0 }
```

### DIFF
--- a/packages/rest/src/lib/handlers/Shared.ts
+++ b/packages/rest/src/lib/handlers/Shared.ts
@@ -137,15 +137,17 @@ export async function handleErrors(
 	} else {
 		// Handle possible malformed requests
 		if (status >= 400 && status < 500) {
-			// If we receive this status code, it means the token we had is no longer valid.
-			if (status === 401 && requestData.auth === true) {
+			// The request will not succeed for some reason, parse the error returned from the api
+			const data = (await parseResponse(res)) as DiscordErrorData | OAuthErrorData;
+			const isDiscordError = 'code' in data;
+
+			// If we receive the 401 status with 0 code, it means the token we had is no longer valid.
+			if (status === 401 && requestData.auth === true && isDiscordError && data.code === 0) {
 				manager.setToken(null!);
 			}
 
-			// The request will not succeed for some reason, parse the error returned from the api
-			const data = (await parseResponse(res)) as DiscordErrorData | OAuthErrorData;
 			// throw the API error
-			throw new DiscordAPIError(data, 'code' in data ? data.code : data.error, status, method, url, requestData);
+			throw new DiscordAPIError(data, isDiscordError ? data.code : data.error, status, method, url, requestData);
 		}
 
 		return res;
--- a/packages/rest/src/lib/handlers/Shared.ts
+++ b/packages/rest/src/lib/handlers/Shared.ts
@@ -140,9 +140,10 @@ export async function handleErrors(
 			// The request will not succeed for some reason, parse the error returned from the api
 			const data = (await parseResponse(res)) as DiscordErrorData | OAuthErrorData;
 			const isDiscordError = 'code' in data;
+			const isOauthError = 'error' in data;
 
 			// If we receive the 401 status with 0 code, it means the token we had is no longer valid.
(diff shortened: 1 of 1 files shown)


didinele (commented):
> Remembering that you don't necessarily need to have a webhook to receive this error, it just takes a long time to respond to an interaction, so this wrong condition ends up affecting several bots, especially in Discord API instabilities.

FWIW I'm pretty sure this is wrong. When you respond too slow you get "Unknown interaction", which I *assume* is a 404, otherwise our issue tracker would probably be getting blown up if we invalidated the token every time you were too slow to respond...

Otherwise, for those webhook routes, the whole point is to pass in `auth: false` for routes that don't require the bot's token, i.e. see

https://github.com/discordjs/discord.js/blob/ab6a69401e7de81a7c619d82b9f00741dfae3af4/packages/discord.js/src/structures/Webhook.js#L228-L233

and similar callsites from the main library.

All that said, I was not aware of this code 0, but this seems like a pretty nice additional measure that I'm not immediately against.

josejooj:
> FWIW I'm pretty sure this is wrong. When you respond too slow you get "Unknown interaction", which I _assume_ is a 404, otherwise our issue tracker would probably be getting blown up if we invalidated the token every time you were too slow to respond...

I apologize, I really expressed myself wrong, this error does not happen when you take too long to respond to the interaction, but when you try to edit an interaction response that has been going on for a long time (15 minutes on my tests), here are 7 interactions routes that use the `/webhooks` route: [Discord API Docs](https://discord.com/developers/docs/interactions/receiving-and-responding#get-original-interaction-response)

I didn't know about the existence of auth: false, it probably solves the problem, but I leave my suggestion that code 0 be read to determine if the token is really invalid, we don't know which other routes may end up returning 401 without actually representing an invalid token, I work developing several bots, and some receive this error without actually having an invalid token, I continue to look for the causes, but treating code 0 should solve it.

almeidx:
Superceded by #11066

PR #10801, opened 2025-03-15
fix(website): correctly link type parameters in docs

Qjuh:

Links to Type Parameters in docs were using the wrong suffix so they 404d.

For example the link to CachedType or RawType on https://discord.js.org/docs/packages/discord.js/14.18.0/CacheTypeReducer:TypeAlias

### DIFF
--- a/packages/scripts/src/generateSplitDocumentation.ts
+++ b/packages/scripts/src/generateSplitDocumentation.ts
@@ -295,15 +295,15 @@ function itemExcerptText(excerpt: Excerpt, apiPackage: ApiPackage, parent?: ApiT
 			}
 
 			if (parent?.typeParameters.some((type) => type.name === token.text)) {
-				const [packageName, parentItem] = parent.canonicalReference.toString().split('!');
+				const resolvedParent = resolveCanonicalReference(parent.canonicalReference, apiPackage);
 				return {
 					text: token.text,
 					resolvedItem: {
 						kind: 'TypeParameter',
 						displayName: token.text,
 						containerKey: `${parent.containerKey}|${token.text}`,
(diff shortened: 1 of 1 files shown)


vladfrangu (approved):

packages/scripts/src/generateSplitDocumentation.ts
```diff
@@ -295,15 +295,15 @@ function itemExcerptText(excerpt: Excerpt, apiPackage: ApiPackage, parent?: ApiT
...
+				const resolvedParent = resolveCanonicalReference(parent.canonicalReference, apiPackage);
 				return {
 					text: token.text,
 					resolvedItem: {
 						kind: 'TypeParameter',
 						displayName: token.text,
 						containerKey: `${parent.containerKey}|${token.text}`,
-						uri: `${parentItem}#${token.text}`,
-						packageName: packageName?.replace('@discordjs/', ''),
+						uri: `${resolveItemURI(parent)}#${token.text}`,
```
almeidx, line 305: Not sure if this is possible, but if this branch is ever reached this would include 2 fragments on the uri

https://github.com/discordjs/discord.js/blob/09beb8a6a095eadbba33c3478d05796d6de10649/packages/scripts/src/generateSplitDocumentation.ts#L226

almeidx (approved):

packages/scripts/src/generateSplitDocumentation.ts
Qjuh: That shouldn’t ever happen, since type parameters like that are only documented separately on top-level stuff like class, type alias, interface or top-level function.

Jiralite (approved):

PR #10802, opened 2025-03-15
fix: message builders

almeidx:
- Added `clearParse`, `clearRoles`, and `clearUsers` methods to the `AllowedMentionsBuilder`, since passing an empty array and omitting the these fields behave differently
- Strictened assertions
- Removed `AttachmentBuilder#clearId`, as it is a required field
- Added missing `MessageBuilder#setEmbeds`
- Added missing `MessageReferenceBuilder#setFailIfNotExists`
- Improve/fix documentation
- Consistency:tm:


packages/builders/src/messages/Message.ts
```diff
@@ -197,6 +212,15 @@ export class MessageBuilder implements JSONEncodable<RESTPostAPIChannelMessageJS
 		return this;
 	}
 
+	/**
+	 * Sets the embeds for this message.
+	 *
+	 * @param embeds - The embeds to set
+	 */
+	public setEmbeds(...embeds: RestOrArray<APIEmbed | EmbedBuilder | ((builder: EmbedBuilder) => EmbedBuilder)>): this {
+		return this.spliceEmbeds(0, this.embeds.length, ...normalizeArray(embeds));
```
vladfrangu, line 221: I tested it and while it works, I would much rather stick to explicitly overriding the array instead on relying on splicing to replicate a set operator

vladfrangu (changes requested):

packages/builders/src/messages/Message.ts
almeidx: This (and the other changes like this) fall under the "Consistency:tm:" part of my description

vladfrangu: Well I am against relying on splice to do a set 🤷, would love to hear other's thoughts.

didinele: I don't see a problem with using the splice method, but if we're not going to, then we should update it everywhere else too.

didinele (approved):
thanks

packages/builders/src/messages/Message.ts
almeidx: I think this can be decided and updated in another pull request. Doesn't seem like a good reason to block merging this

Qjuh (approved):

packages/builders/src/messages/Message.ts
```diff
@@ -218,7 +242,7 @@ export class MessageBuilder implements JSONEncodable<RESTPostAPIChannelMessageJS
 	 * @param updater - The function to update the allowed mentions with
 	 */
 	public updateAllowedMentions(updater: (builder: AllowedMentionsBuilder) => AllowedMentionsBuilder): this {
-		this.data.allowed_mentions = updater(this.data.allowed_mentions ?? new AllowedMentionsBuilder());
+		updater((this.data.allowed_mentions ??= new AllowedMentionsBuilder()));
```
Qjuh, line 245: On second thought this relies on the assumption that the passed function returns the same builder it gets passed. While that *should* normally the case we don‘t actually require it anywhere yet. If you create a new Builder inside the callback and return that it should work as intended, but wouldn’t with this change.
Same applies to all the other places this was done

Qjuh (changes requested):

packages/builders/src/messages/Message.ts
vladfrangu: +1, this should be reverted to what it was before

almeidx: Changed the return type to `void`.

almeidx: https://github.com/discordjs/discord.js/pull/10802#discussion_r1997308456

almeidx: > you‘re now restricting people from returning stuff from the callback. Why?

It's an update function. If you want to create a new builder, use *set*

> What’s the benefit?

Consistency

> This seems to me like a classic example of „breaking change just because we can“

This is not released.

> Especially since it‘s not consistent with the rest of the builders anymore

Every single update* function:

https://github.com/discordjs/discord.js/blob/abc5d99ce89bad64d0ad0d7d68310e610efdca76/packages/builders/src/messages/embed/Embed.ts#L159-L162
https://github.com/discordjs/discord.js/blob/abc5d99ce89bad64d0ad0d7d68310e610efdca76/packages/builders/src/messages/embed/Embed.ts#L225-L228
https://github.com/discordjs/discord.js/blob/abc5d99ce89bad64d0ad0d7d68310e610efdca76/packages/builders/src/messages/poll/Poll.ts#L163-L166
https://github.com/discordjs/discord.js/blob/abc5d99ce89bad64d0ad0d7d68310e610efdca76/packages/builders/src/messages/poll/PollAnswer.ts#L47-L50

Qjuh: Apologies, you‘re right. I misunderstood what this was all about.

Qjuh (approved):

vladfrangu (approved):

PR #10808, opened 2025-03-18
chore(WebSocketShard): improve zlib-sync errors

LJNeon:
The current error thrown from `zlib-sync` [isn't great](https://github.com/discordjs/discord.js/blob/main/packages/ws/src/ws/WebSocketShard.ts#L672), and doesn't match what errors from `node:zlib` look like. This brings them more in line with `node:zlib` errors.


packages/ws/src/ws/WebSocketShard.ts
```diff
@@ -59,6 +59,17 @@ export enum WebSocketShardDestroyRecovery {
...
+enum ZlibErrorCodes {
+	Z_NEED_DICT = 2,
+	Z_STREAM_END = 1,
+	Z_ERRNO = -1,
+	Z_STREAM_ERROR = -2,
+	Z_DATA_ERROR = -3,
+	Z_MEM_ERROR = -4,
+	Z_BUF_ERROR = -5,
+	Z_VERSION_ERROR = -6,
+}
```
almeidx, line 71: Having these hardcoded here is not ideal.. Especially since zlib-sync exports each of these codes as constants

We could import them and construct this object using them
https://github.com/abalabahaha/zlib-sync/blob/7f1b404ba9e1bb9af1f47ca395f33e4fdfc2299c/src/zlib_sync.cc#L220-L228

(This would imply moving this object creation into the `unpackMessage` method as that's where zlib is being imported)

almeidx: Something like this perhaps

```ts
					enum ZlibErrorCodes {
						Z_NEED_DICT = zLibSync.Z_NEED_DICT,
						Z_STREAM_END = zLibSync.Z_STREAM_END,
						Z_ERRNO = zLibSync.Z_ERRNO,
						Z_STREAM_ERROR = zLibSync.Z_STREAM_ERROR,
						Z_DATA_ERROR = zLibSync.Z_DATA_ERROR,
						Z_MEM_ERROR = zLibSync.Z_MEM_ERROR,
						Z_BUF_ERROR = zLibSync.Z_BUF_ERROR,
						Z_VERSION_ERROR = zLibSync.Z_VERSION_ERROR,
					}

					error.code = ZlibErrorCodes[this.zLibSyncInflate.err];
					this.emit(WebSocketShardEvents.Error, error);
```

vladfrangu: Just turn the enum into a normal object please, and move it to a constants / utils place

packages/ws/src/utils/constants.ts
```diff
@@ -31,6 +32,17 @@ export const CompressionParameterMap = {
 	[CompressionMethod.ZlibSync]: 'zlib-stream',
 } as const satisfies Record<CompressionMethod, string>;
 
+export const ZlibErrorCodes = {
+	[ZlibSync.Z_NEED_DICT]: 'Z_NEED_DICT',
```
almeidx, line 36: Can't import zlib-sync on the top level at runtime. It's being lazy loaded conditionally since it's usage is opt-in
https://github.com/discordjs/discord.js/blob/cf87cbb04103feccb7bd515d89a820adddbe7d8f/packages/ws/src/ws/WebSocketShard.ts#L667

Refer to the parenthesis i included in https://github.com/discordjs/discord.js/pull/10808#discussion_r2071456416

almeidx (changes requested):

packages/ws/src/utils/constants.ts
LJNeon: I moved it, does that look good to you now?

iCrawl (approved):

didinele (approved):

packages/ws/src/ws/WebSocketShard.ts
```diff
@@ -667,10 +667,23 @@ export class WebSocketShard extends AsyncEventEmitter<WebSocketShardEventsMap> {
 				this.zLibSyncInflate.push(Buffer.from(decompressable), flush ? zLibSync.Z_SYNC_FLUSH : zLibSync.Z_NO_FLUSH);
 
 				if (this.zLibSyncInflate.err) {
-					this.emit(
-						WebSocketShardEvents.Error,
-						new Error(`${this.zLibSyncInflate.err}${this.zLibSyncInflate.msg ? `: ${this.zLibSyncInflate.msg}` : ''}`),
-					);
+					// Must be here because zlib-sync is lazily loaded
+					const ZlibErrorCodes = {
```
vladfrangu, line 672: Cache this in a variable outside the class please, so its computed only once

LJNeon: This code path is only executed when zlib-sync is used and a zlib error occurs when inflating a payload. Is this a code path that needs that kind of optimization for a small & simple object?

vladfrangu: Be that as it may, you can still cache it, but meh, this should be a code path hit once a blue moon, so it's fine

vladfrangu (approved):

PR #10814, opened 2025-03-22
fix: all shards spawning on shard 0 with sharding manager

almeidx:
`process.env.SHARDS` only has the shard id (not an array). This meant that we were passing `shardIds: 0` to WebSocketManager, which fails this the following `if` statement, making /ws fetch get all shards and start them on the current process. 

https://github.com/discordjs/discord.js/blob/4d1942681077c5c49abaf90873cf1cafb21de84f/packages/ws/src/ws/WebSocketManager.ts#L346-L347

image

### DIFF
--- a/packages/discord.js/src/client/Client.js
+++ b/packages/discord.js/src/client/Client.js
@@ -58,7 +58,8 @@ class Client extends BaseClient {
     const defaults = Options.createDefault();
 
     if (this.options.ws.shardIds === defaults.ws.shardIds && 'SHARDS' in data) {
-      this.options.ws.shardIds = JSON.parse(data.SHARDS);
+      const shards = JSON.parse(data.SHARDS);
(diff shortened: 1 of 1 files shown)


vladfrangu:
Doesn't this get set to `0,1,2` when having multiple shards per process?

Qjuh:
Current `ShardingManager` has no notion of multiple shards per process. But I agree that it should be possible to pass both array of ids and plain number to it, same way it was possible in v14. So instead of this fix it should be added as conversion to array if typeof is number

Qjuh (approved):

Qjuh:
This is not an issue in v14, as it‘s correctly converted to an array there. So nothing to backport.

vladfrangu (approved):

sdanialraza (approved):

PR #10824, opened 2025-04-05
chore: upgrade deps

iCrawl:


iCrawl:
The guide deploy fails because I nuked the whole folder and only afterwards deleted the vercel project.

PR #10826, opened 2025-04-07
fix(WebsocketManager): always emit shardDisconnect on unresumable close

Qjuh:

Previously `Client#shardDisconnect` was only emitted if the cause was (through a chain of method calls) `Client#destroy()` being called. This was missing the cases where Discord caused the unresumable disconnect from the gateway, e.g. token reset.


didinele (approved):

iCrawl (approved):

packages/discord.js/src/client/websocket/WebSocketManager.js
```diff
@@ -242,7 +242,7 @@ class WebSocketManager extends EventEmitter {
     this._ws.on(WSWebSocketShardEvents.Closed, ({ code, shardId }) => {
       const shard = this.shards.get(shardId);
       shard.emit(WebSocketShardEvents.Close, { code, reason: reasonIsDeprecated, wasClean: true });
-      if (UNRESUMABLE_CLOSE_CODES.includes(code) && this.destroyed) {
+      if (UNRESUMABLE_CLOSE_CODES.includes(code)) {
```
vladfrangu, line 245: The array also includes `1000` - a normal close that _will_ try to reconnect unless the ws is destroyed. We'll need to handle that manually

vladfrangu (changes requested):

vladfrangu (approved):

PR #10827, opened 2025-04-07
fix: generateSplitDocumentation for external docs on main

Qjuh:


packages/scripts/src/generateSplitDocumentation.ts
```diff
@@ -522,8 +522,8 @@ function itemInfo(item: ApiDeclaredItem) {
 	const isAbstract = ApiAbstractMixin.isBaseClassOf(item) && item.isAbstract;
 	const isOptional = ApiOptionalMixin.isBaseClassOf(item) && item.isOptional;
 	const isDeprecated = Boolean(item.tsdocComment?.deprecatedBlock);
-	const isExternal = Boolean(item.sourceLocation.fileUrl?.includes('node_modules'));
-
+	const isExternal = Boolean(sourceLine === undefined);
+	if (item.displayName === 'MessageActionRowComponentBuilder') console.log(sourceURL);
```
almeidx, line 526: Is this supposed to be here?

packages/scripts/src/generateSplitDocumentation.ts
```diff
@@ -587,6 +586,19 @@ function resolveFileUrl(item: ApiDeclaredItem) {
 				sourceURL: href,
 			};
 		}
+	} else if (fileUrl?.includes('/dist/') && fileUrl.includes('/main/packages/')) {
+		const [, pkg] = fileUrl.split('/main/packages/');
+		const pkgName = pkg?.split('/')[0];
```
almeidx, line 591:
```suggestion
		const pkgName = pkg!.split('/')[0];
```

almeidx (approved):

packages/api-extractor/extends/tsdoc-base.json
```diff
@@ -71,6 +71,7 @@
 
 		"@betaDocumentation": true,
 		"@internalRemarks": true,
+		"@mixes": true,
```
almeidx, line 74: This can probably be removed now.

https://github.com/Qjuh/discord.js/blob/scripts/external-docs-main-branch/packages/builders/tsdoc.json

Can you maybe also add `@unstable` here, because of discord-api-types?

Qjuh: `@unstable` should be defined in `tsdoc.json` for `discord-api-types` not here. The only reason `@mixes` got added here is because our `api-extractor` actually uses that tag for documentation decisions.
Alternatively the logic what to do with `@unstable` would have to be added to `api-extractor` too.

Qjuh: And the tsdoc.json in builders is still needed for eslint to recognize the `@mixes` tag. But it should simply extend the `api-extractor` version.

iCrawl (approved):

didinele (approved):

Amgelo563 (approved):

PR #10857, opened 2025-04-26
fix(Guild): cache soundboard sounds when patching

N1ckPro:
Guild's soundboard sounds are sent with the guild create event but are never cached before fetching them.

### DIFF
--- a/packages/discord.js/src/structures/Guild.js
+++ b/packages/discord.js/src/structures/Guild.js
@@ -499,6 +499,13 @@ class Guild extends AnonymousGuild {
     } else {
       this.incidentsData ??= null;
     }
+
+    if (data.soundboard_sounds) {
+      this.soundboardSounds.cache.clear();
+      for (const soundboardSound of data.soundboard_sounds) {
+        this.soundboardSounds._add(soundboardSound);
+      }
+    }
   }
 
   /**


packages/discord.js/src/structures/Guild.js
```diff
@@ -499,6 +499,13 @@ class Guild extends AnonymousGuild {
     } else {
       this.incidentsData ??= null;
     }
+
+    if (data.soundboard_sounds) {
+      this.soundboardSounds.cache.clear();
```
Jiralite, line 504: If `soundboard_sounds` is only sent on `GUILD_CREATE`, then isn't this line useless? What will it clear apart from nothing?

Qjuh: unavailable Guild becoming available again?

Jiralite: It is not done for all similar properties.

I have to wonder why it's in `_patch()`. No gateway events will be received if the guild is unavailable nor can members do anything. Seems more sensible to be done in the constructor, no?

N1ckPro: Seems like `guild_scheduled_events`, `stage_instances`, `presences` and maybe more fields do have a similar check in `_patch()` while only being sent on `GUILD_CREATE` too.

Qjuh (approved):

Jiralite (approved):

sdanialraza (approved):

PR #10888, opened 2025-05-07
refactor(ClientFetchInviteOptions): configurable withCounts

Jiralite:
Depending on the invite type, an invite class will be created, similar to components. This has the added benefit of stronger typings: group direct message invites cannot be deleted, they do not have `target_application`, etc.

The separation of properties was done per the API specification:

- [`GuildInvite`](https://github.com/discord/discord-api-spec/blob/507909d1ea5a921064d599e1f3b57163af2037d6/specs/openapi.json#L20138)
- [`GroupDMInvite`](https://github.com/discord/discord-api-spec/blob/507909d1ea5a921064d599e1f3b57163af2037d6/specs/openapi.json#L19243)

[`FriendInvite`](https://github.com/discord/discord-api-spec/blob/507909d1ea5a921064d599e1f3b57163af2037d6/specs/openapi.json#L18485) (https://github.com/discord/discord-api-docs/pull/3659) was not created as applications can no longer seem to fetch them (returns unknown invite).

No other properties were added as they were not in the documentation.

Also contains miscellaneous fixes:
- This nullish coalescing operator here was redundant, as adding channels already has a cache check.  https://github.com/discordjs/discord.js/blob/432cdbe88ad6a0888b499466d54ab82735ba539c/packages/discord.js/src/structures/Invite.js#L181-L183
- Deleting an invite was typed to return the invite, but didn't
- Approximate member and presence counts were typed to always return a number, but this is not true
- `with_counts` is now configurable when fetching an invite


packages/discord.js/src/structures/GuildInvite.js
```diff
@@ -0,0 +1,183 @@
...
+      /**
+       * The guild the invite is for. May include welcome screen data.
+       * @type {?(Guild|InviteGuild)}
+       */
+      this.guild = this.client.guilds.cache.get(data.guild.id) ?? new InviteGuild(this.client, data.guild);
+    } else {
+      this.guild ??= null;
+    }
+
+    super._patch(data, this.guild);
```
Qjuh, line 55: This is a confusing approach. Super constructor calls this._patch which actually uses the implementation here, which then (after assigning the guild from data) call super._patch(), passing that guild. BaseInvite should have no business with that guild, since only GuildInvites have them.

packages/discord.js/src/managers/GuildInviteManager.js
```diff
@@ -53,12 +53,12 @@ class GuildInviteManager extends CachedManager {
    */
 
   /**
-   * Resolves an InviteResolvable to an Invite object.
+   * Resolves an InviteResolvable to a `GuildInvite` object.
```
Qjuh, line 56: Use GuildInviteResolvable here

packages/discord.js/src/structures/BaseInvite.js
```diff
@@ -0,0 +1,182 @@
...
+       */
+      this.channelId = data.channel_id;
+    }
+
+    if ('channel' in data) {
+      /**
+       * The channel this invite is for.
+       * @type {?BaseChannel}
+       */
+      this.channel = this.client.channels._add(data.channel, guild, { cache: false });
```
Qjuh, line 85:
```suggestion
      this.channel = this.client.channels._add(data.channel, guild, { cache: false }) ?? this.client.channels.resolve(this.channelId);
```

If `data.channel` is null but `data.channel_id` exists this would result in `this.channel` being null although it might exist in cache. This happens in `inviteDelete` event for example.

packages/discord.js/src/managers/GuildInviteManager.js
```diff
@@ -33,7 +33,7 @@ class GuildInviteManager extends CachedManager {
   }
 
   /**
-   * Data that resolves to give an Invite object. This can be:
+   * Data that resolves to give a `GuildInvite`. This can be:
```
Qjuh, line 36: We have this type defined twice (here and in DataResolver) which is also used in `Client#fetchInvite()` where it‘s not explicitly for only GuildInvite, but GroupDMInvite too. So this one here should be renamed to `GuildInviteResolvable`.

Qjuh (changes requested):

packages/discord.js/src/structures/BaseInvite.js
Jiralite: Since the documentation does not state the channel object is emitted, I have done this.

packages/discord.js/src/managers/GuildInviteManager.js
```diff
@@ -89,7 +89,7 @@ class GuildInviteManager extends CachedManager {
   /**
    * Fetches invite(s) from Discord.
    * @param {InviteResolvable|FetchInviteOptions|FetchInvitesOptions} [options] Options for fetching guild invite(s)
-   * @returns {Promise<Invite|Collection<string, Invite>>}
```
Qjuh, line 92: Should use GuildInviteResolvable as type here too (in the line before)

packages/discord.js/src/structures/GroupDMInvite.js
```diff
@@ -0,0 +1,34 @@
...
+   * @type {?number}
+   */
+
+  _patch(data) {
+    super._patch(data);
+
+    if ('channel' in data) {
+      /**
+       * The channel this invite is for.
+       * @type {?BaseChannel}
```
Qjuh, line 23: Shouldn’t this be PartialGroupDMChannel here?

packages/discord.js/src/structures/GuildInvite.js
```diff
@@ -0,0 +1,195 @@
...
+       */
+      this.guild = this.client.guilds.cache.get(data.guild.id) ?? new InviteGuild(this.client, data.guild);
+    } else {
+      this.guild ??= null;
+    }
+
+    if ('channel' in data) {
+      /**
+       * The channel this invite is for.
+       * @type {?BaseChannel}
```
Qjuh, line 60: GuildInvitableChannel or something like that if it exists. Or at least GuildChannel?

packages/discord.js/src/util/DataResolver.js
```diff
@@ -33,13 +33,13 @@ function resolveCode(data, regex) {
 }
 
 /**
- * Resolves InviteResolvable to an invite code.
- * @param {InviteResolvable} data The invite resolvable to resolve
+ * Resolves GuildInviteResolvable to an invite code.
```
Qjuh, line 36: These should stay InviteResolvable, as they can be used on both types.

packages/discord.js/src/util/Invites.js
```diff
@@ -0,0 +1,29 @@
...
+};
+
+/**
+ * Creates an invite.
+ * @param {Client} client The client
+ * @param {Object} data The data
+ * @returns {BaseInvite}
+ * @ignore
+ */
+function createInvite(client, data) {
```
Qjuh, line 25: This means it could happen to have an actual instance of BaseInvite, not a child class. But BaseInvite is documented to be abstract.

packages/discord.js/typings/index.d.ts
```diff
@@ -1925,37 +1929,49 @@ export class InteractionWebhook {
...
+  public readonly guildId: Snowflake;
+  public channel: NonThreadGuildBasedChannel | null;
+  public targetType: InviteTargetType | null;
+  public targetUser: User | null;
+  public targetApplication: IntegrationApplication | null;
   public guildScheduledEvent: GuildScheduledEvent | null;
+  public uses: number | null;
+  public maxUses: number | null;
+  public temporary: boolean | null;
+  public approximatePresenceCount: WithCounts extends true ? number : number | null;
```
Qjuh, line 1963:
```suggestion
  public approximatePresenceCount: WithCounts extends true ? number : null;
```

packages/discord.js/typings/index.d.ts
```diff
@@ -1925,37 +1929,49 @@ export class InteractionWebhook {
...
-  public maxUses: number | null;
-  public memberCount: number;
-  public presenceCount: number;
-  public targetApplication: IntegrationApplication | null;
-  public targetUser: User | null;
-  public targetType: InviteTargetType | null;
-  public temporary: boolean | null;
-  public type: InviteType;
+  public readonly channelId: Snowflake | null;
+  public approximateMemberCount: WithCounts extends true ? number : number | null;
```
Qjuh, line 1944:
```suggestion
  public approximateMemberCount: WithCounts extends true ? number : null;
```

packages/discord.js/typings/index.d.ts
```diff
@@ -5030,7 +5046,7 @@ export interface Caches {
   // TODO: GuildChannelManager: [manager: typeof GuildChannelManager, holds: typeof GuildChannel];
   GuildEmojiManager: [manager: typeof GuildEmojiManager, holds: typeof GuildEmoji];
   GuildForumThreadManager: [manager: typeof GuildForumThreadManager, holds: typeof ThreadChannel<true>];
-  GuildInviteManager: [manager: typeof GuildInviteManager, holds: typeof Invite];
+  GuildInviteManager: [manager: typeof GuildInviteManager, holds: GuildInvite];
```
Qjuh, line 5049:
```suggestion
  GuildInviteManager: [manager: typeof GuildInviteManager, holds: typeof GuildInvite];
```

Qjuh (changes requested):

Qjuh (approved):

packages/discord.js/src/structures/GuildInvite.js
```diff
@@ -0,0 +1,211 @@
...
+ * A channel invite leading to a guild.
+ *
+ * @extends {BaseInvite}
+ */
+class GuildInvite extends BaseInvite {
+  constructor(client, data) {
+    super(client, data);
+
+    // Type may be missing from audit logs.
+    this.type ??= InviteType.Guild;
```
vladfrangu: Might as well always set it fwiw

packages/discord.js/src/structures/GuildInvite.js
```diff
@@ -0,0 +1,211 @@
...
+   *
+   * @type {boolean}
+   * @readonly
+   */
+  get deletable() {
+    const guild = this.guild;
+    if (!guild || !this.client.guilds.cache.has(guild.id)) return false;
+    if (!guild.members.me) throw new DiscordjsError(ErrorCodes.GuildUncachedMe);
+    return Boolean(
+      this.channel?.permissionsFor(this.client.user).has(PermissionFlagsBits.ManageChannels, false) ||
```
vladfrangu, line 182: Why do we pass false for checkAdmin? If admin -> you can delete

```suggestion
      this.channel?.permissionsFor(this.client.user).has(PermissionFlagsBits.ManageChannels) ||
```

packages/discord.js/src/util/DataResolver.js
```diff
@@ -43,7 +43,7 @@ function resolveCode(data, regex) {
  * @private
  */
 function resolveInviteCode(data) {
-  return resolveCode(data, Invite.InvitesPattern);
+  return resolveCode(data, GuildInvite.InvitesPattern);
```
vladfrangu: This regex should be on the base class and probably used from there too

packages/discord.js/typings/index.d.ts
```diff
@@ -2001,37 +2005,49 @@ export class InteractionWebhook {
...
+  public get deletable(): boolean;
+  public delete(reason?: string): Promise<void>;
+}
+
+export class GroupDMInvite<WithCounts extends boolean = boolean> extends BaseInvite<WithCounts> {
+  public readonly type: InviteType.GroupDM;
+  public channel: PartialGroupDMChannel | null;
 }
 
+export type Invite<WithCounts extends boolean = boolean> = GroupDMInvite<WithCounts> | GuildInvite<WithCounts>;
```
vladfrangu, line 2066: Let's name this AnyInvite

packages/discord.js/typings/index.d.ts
```diff
@@ -914,6 +914,10 @@ export class Client<Ready extends boolean = boolean> extends BaseClient<ClientEv
   public destroy(): Promise<void>;
   public deleteWebhook(id: Snowflake, options?: WebhookDeleteOptions): Promise<void>;
   public fetchGuildPreview(guild: GuildResolvable): Promise<GuildPreview>;
+  public fetchInvite(
+    invite: InviteResolvable,
+    options: ClientFetchInviteOptions & { withCounts: true },
```
vladfrangu, line 936: For the future, lets just set this to true always for users XD

packages/discord.js/src/structures/GuildInvite.js
Qjuh: Because `permissionsFor()` already does the `checkAdmin=true` by default, so the returned PermissionsBitField would already have all permissions, no need to check for admin again.

packages/discord.js/src/util/DataResolver.js
Jiralite: Actually, the regular expression is already on the base class. I will switch this to `BaseInvite`.

vladfrangu (approved):

PR #10900, opened 2025-05-16
feat: @discordjs/structures

Qjuh:

Supersedes #8417

Introduces a generic structure implementation for the main lib (and others) to build on.

TODO:
- [ ] Application
- [ ] Audit Log
- [ ] Auto Moderation
- [ ] Channel
  - [x] Channels
  - [x] PermissionOverwrite
  - [x] ForumTag
  - [x] ThreadMetadata
  - [x] Mixins
    - [x] DM
    - [x] Guild
    - [x] Text
    - [x] Owner
    - [x] Parent
    - [x] Permissions
    - [x] Pin
    - [x] Slowmode
    - [x] Topic
    - [x] Webhooks
    - [x] Thread
    - [x] ThreadOnly
    - [x] Voice
  - [ ] Message
- [ ] Emoji
- [ ] Guild
  - [ ] Guild
  - [ ] Guild Preview
  - [ ] Guild Widget
  - [ ] Guild Member
  - [ ] Integration
  - [ ] Guild Ban
  - [ ] ~~Membership Screening~~ Oh wait...this is *still* undocumented
- [ ] Guild Scheduled Event
- [ ] Guild Template
- [ ] Interactions (Need more thoughts on this)
- [x] Invite
- [ ] Stage Instance
- [ ] Sticker
  - [ ] Sticker
  - [ ] Sticker Pack
- [ ] Teams
- [x] User
  - [x] User
    - [x] AvatarDecorationData
  - [x] Connection
- [ ] Voice
  - [ ] Voice State
  - [ ] Voice Region
- [ ] Webhook

### DIFF
--- a/apps/guide/content/docs/legacy/additional-info/meta.json
+++ b/apps/guide/content/docs/legacy/additional-info/meta.json
@@ -1,3 +1,3 @@
 {
-	"pages": ["async-await", "collections", "es6-syntax", "notation", "rest-api"]
+	"pages": ["async-await", "collections", "es6-syntax", "notation", "rest-api", "proxy"]
 }
--- a/apps/guide/content/docs/legacy/additional-info/proxy.mdx
+++ b/apps/guide/content/docs/legacy/additional-info/proxy.mdx
@@ -1,0 +1,73 @@
+---
+title: Using a proxy
+---
+
+This guide will show you how to set up a proxy with discord.js. This may be necessary if you are deploying your bot to a server with a firewall only allowing outside traffic through the proxy.
+
+Proxying discord.js requires two components: a REST proxy and a WebSocket proxy.
+
+## Prerequisites
+
+To achieve these two components you can utilize the `undici` and `global-agent` packages:
+
+```sh tab="npm"
+npm install undici global-agent
+```
+
+```sh tab="yarn"
+yarn add undici global-agent
+```
+
+```sh tab="pnpm"
+pnpm add undici global-agent
+```
+
+```sh tab="bun"
+bun add undici global-agent
+```
+
+## Setting up the proxy for REST calls
+
+The `@discordjs/rest` package handling HTTP requests in discord.js uses the `undici` package. Accordingly, you can provide a custom `ProxyAgent` configuration to the client constructor:
+
+```js title="index.js" lineNumbers
+const { ProxyAgent } = require('undici'); // [!code word:ProxyAgent]
+const { Client } = require('discord.js');
+
+const client = new Client({
+	// ...other client options
+	rest: {
+		agent: new ProxyAgent('http://my-proxy-server:port'),
+	},
+});
+
+client.login('your-token-goes-here');
+```
+
+<Callout>
+	For further information on the `undici` `ProxyAgent`, please refer to the [undici
+	documentation](https://undici.nodejs.org/#/docs/api/ProxyAgent.md).
+</Callout>
+
+## Setting up the proxy for the WebSocket connection
+
+To set up a proxy for WebSocket, you can use the `global-agent` package. You will need to import and call the `bootstrap()` function and set the required `GLOBAL_AGENT` globals as shown below:
+
+```js title="index.js" lineNumbers
+const { ProxyAgent } = require('undici');
+const { Client } = require('discord.js');
+const { bootstrap } = require('global-agent'); // [!code ++:5]
+
+bootstrap(); // [!code word:bootstrap]
+global.GLOBAL_AGENT.HTTP_PROXY = 'http://my-proxy-server:port';
+global.GLOBAL_AGENT.HTTPS_PROXY = 'https://my-proxy-server:port';
+
+const client = new Client({
+	// ...other client options
+	rest: {
+		agent: new ProxyAgent('http://my-proxy-server:port'),
+	},
+});
+
+client.login('your-token-goes-here');
+```
--- a/apps/guide/content/docs/legacy/oauth2/oauth2.mdx
+++ b/apps/guide/content/docs/legacy/oauth2/oauth2.mdx
@@ -75,56 +75,11 @@ The `identify` scope will allow your application to get basic user information f
 
 ### Implicit grant flow
 
-You have your website, and you have a URL. Now you need to use those two things to get an access token. For basic applications like [SPAs](https://en.wikipedia.org/wiki/Single-page_application), getting an access token directly is enough. You can do so by changing the `response_type` in the URL to `token`. However, this means you will not get a refresh token, which means the user will have to explicitly re-authorize when this access token has expired.
-
-After you change the `response_type`, you can test the URL right away. Visiting it in your browser, you will be directed to a page that looks like this:
-
-![Authorization Page](./images/authorize-app-page.png)
-
-You can see that by clicking `Authorize`, you allow the application to access your username and avatar. Once you click through, it will redirect you to your redirect URL with a [fragment identifier](https://en.wikipedia.org/wiki/Fragment_identifier) appended to it. You now have an access token and can make requests to Discord's API to get information on the user.
-
-Modify `index.html` to add your OAuth2 URL and to take advantage of the access token if it exists. Even though [`URLSearchParams`](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams) is for working with query strings, it can work here because the structure of the fragment follows that of a query string after removing the leading "#".
-
-```html title="index.html" lineNumbers=11
-<div id="info">Hoi!</div>
-<a id="login" style="display: none;" href="your-oauth2-URL-here">Identify Yourself</a>
-<script>
-	window.onload = () => {
-		const fragment = new URLSearchParams(window.location.hash.slice(1));
-		const [accessToken, tokenType] = [fragment.get('access_token'), fragment.get('token_type')];
-
-		if (!accessToken) {
-			return (document.getElementById('login').style.display = 'block');
-		}
-
-		fetch('https://discord.com/api/users/@me', {
-			headers: {
-				authorization: `${tokenType} ${accessToken}`,
-			},
-		})
-			.then((result) => result.json())
-			.then((response) => {
-				const { username, discriminator } = response;
-				document.getElementById('info').innerText += ` ${username}#${discriminator}`;
-			})
-			.catch(console.error);
-	};
-</script>
-```
-
-Here you grab the access token and type from the URL if it's there and use it to get info on the user, which is then used to greet them. The response you get from the [`/api/users/@me` endpoint](https://discord.com/developers/docs/resources/user#get-current-user) is a [user object](https://discord.com/developers/docs/resources/user#user-object) and should look something like this:
-
-```json
-{
-	"id": "123456789012345678",
-	"username": "User",
-	"discriminator": "0001",
-	"avatar": "1cc0a3b14aec3499632225c708451d67",
-	...
-}
-```
-
-In the following sections, we'll go over various details of Discord and OAuth2.
+<Callout type="error">
+	Implicit grant flow, as previously covered in this section, is vulnerable to token leakage and replay attacks. Please
+	use the **authorization grant** flow instead. You can find out more about the best implementation practices in the
+	[Oauth2 RFC](https://datatracker.ietf.org/doc/html/rfc9700).
+</Callout>
 
 ## More details
 
--- a/apps/guide/content/docs/legacy/sharding/additional-information.mdx
+++ b/apps/guide/content/docs/legacy/sharding/additional-information.mdx
@@ -79,14 +79,24 @@ You can access them later as usual via `process.argv`, which contains an array o
 
 ## Eval arguments
 
-There may come the point where you will want to pass arguments from the outer scope into a `.broadcastEval()` call.
+There may come the point where you will want to pass arguments from the outer scope into a `.broadcastEval()` call. The `context` parameter is useful for this purpose.
 
 ```js
-function funcName(c, { arg }) {
+// [!code word:context]
+function funcName(client, context) {
 	// ...
 }
 
-client.shard.broadcastEval(funcName, { context: { arg: 'arg' } });
+// Evaluate on all shards
+client.shard.broadcastEval(funcName, {
+	context: { arg: 'arg' },
+});
+
+// Evaluate on a specific shard
+client.shard.broadcastEval(funcName, {
+	shard: 0,
+	context: { arg: 'arg' },
+});
 ```
 
 The `BroadcastEvalOptions` typedef was introduced in discord.js v13 as the second parameter in `.broadcastEval()`. It accepts two properties: `shard` and `context`. The `context` property will be sent as the second argument to your function.
--- a/apps/guide/content/docs/legacy/slash-commands/permissions.mdx
+++ b/apps/guide/content/docs/legacy/slash-commands/permissions.mdx
@@ -14,7 +14,7 @@ The slash command permissions for guilds are defaults only and can be altered by
 
 ## Member permissions
 
-You can use `SlashCommandBuilder#setDefaultMemberPermissions` to set the default permissions required for a member to run the command. Setting it to `0` will prohibit anyone in a guild from using the command unless a specific overwrite is configured or the user has the Administrator permission flag.
+You can use `SlashCommandBuilder#setDefaultMemberPermissions` to set the default permissions required for a member to run the command. Setting it to `0` will hide the command from and prohibit anyone in a guild from using the command unless a specific overwrite is configured or the user has the Administrator permission flag.
 
 For this, you'll introduce two common and simple moderation commands: `ban` and `kick`. For a ban command, a sensible default is to ensure that users already have the Discord permission `BanMembers` in order to use it.
 
--- a/apps/guide/content/docs/legacy/slash-commands/response-methods.mdx
+++ b/apps/guide/content/docs/legacy/slash-commands/response-methods.mdx
@@ -49,14 +49,12 @@ After you've sent an initial response, you may want to edit that response for va
 </Callout>
 
 ```js
-const wait = require('node:timers/promises').setTimeout;
-
 client.on(Events.InteractionCreate, async (interaction) => {
 	if (!interaction.isChatInputCommand()) return;
 
 	if (interaction.commandName === 'ping') {
 		await interaction.reply('Pong!');
-		await wait(2_000);
+		// do something that requires time (database queries, api requests, ...)
 		await interaction.editReply('Pong again!'); // [!code word:editReply]
 	}
 });
@@ -71,16 +69,13 @@ As previously mentioned, Discord requires an acknowledgement from your bot withi
 In this case, you can make use of the `ChatInputCommandInteraction#deferReply()` method, which triggers the `<application> is thinking...` message. This also acts as the initial response, to confirm to Discord that the interaction was received successfully and gives you a **15-minute timeframe to complete your tasks** before responding.
 
 ```js
-const wait = require('node:timers/promises').setTimeout;
-
 client.on(Events.InteractionCreate, async (interaction) => {
 	if (!interaction.isChatInputCommand()) return;
 
 	if (interaction.commandName === 'ping') {
 		await interaction.deferReply(); // [!code word:deferReply]
-		// you can take up to 15 minutes! We take 4 seconds to demonstrate this
-		// since it is barely above the 3-second threshold for the initial response
-		await wait(4_000);
+		// you can do things that take time here (database queries, api requests, ...) that you need for the initial response
+		// you can take up to 15 minutes, then the interaction token becomes invalid!
 		await interaction.editReply('Pong!'); // [!code word:editReply]
 	}
 });
--- a/apps/guide/content/docs/voice/meta.json
+++ b/apps/guide/content/docs/voice/meta.json
@@ -8,7 +8,8 @@
 		"life-cycles",
 		"voice-connections",
 		"audio-player",
-		"audio-resources"
+		"audio-resources",
+		"seeking"
 	],
 	"root": true,
 	"icon": "Volume2"
--- a/apps/guide/content/docs/voice/seeking.mdx
+++ b/apps/guide/content/docs/voice/seeking.mdx
@@ -1,0 +1,64 @@
+---
+title: Seeking
+---
+
+There is no built-in method for seeking audio resources in discord.js. However external libraries such as `prism-media` can be used to tackle this issue.
+
+## Seeking in a stream
+
+To seek resource, you can create a new function with the following code:
+
+```js
+// Require necessary package
+const prism = require('prism-media'); // [!code word:prism]
+
+function createFFmpegStream(stream, seek) {
+	let seekPosition = '0';
+	if (seek) seekPosition = String(seek);
+	const transcoder = new prism.FFmpeg({
+		args: [
+			'-analyzeduration',
+			'0',
+			'-loglevel',
+			'0',
+			'-f',
+			's16le',
+			'-ar',
+			'48000',
+			'-ac',
+			'2',
+			'-ss',
+			seekPosition,
+			'-ab',
+			'320',
+		],
+	});
+	const s16le = stream.pipe(transcoder);
+	const opus = s16le.pipe(new prism.opus.Encoder({ rate: 48000, channels: 2, frameSize: 960 }));
+	return opus;
+}
+```
+
+This function takes two arguments: the audio stream and the desired seek position, expressed as duration within the duration of the full stream. It returns the seeked stream.
+
+<Callout>
+	You can find configuration options in the [prism media documentation](https://amishshah.github.io/prism-media/).
+</Callout>
+
+## Using seek with the audio player
+
+```js
+const { createAudioResource, createAudioPlayer } = require('@discordjs/voice');
+const fs = require('fs');
+
+const player = createAudioPlayer();
+const normalAudioResource = createAudioResource('Your audio file path');
+
+player.play(normalAudioResource);
+
+// [!code word:createFFmpegStream]
+const seekedAudioStream = createFFmpegStream(fs.createReadStream('Your audio file path'), 10); // Seek to 10s
+const seekedAudioResource = createAudioResource(seekedAudioStream);
+
+player.play(seekedAudioResource);
+```
--- a/packages/core/src/api/channel.ts
+++ b/packages/core/src/api/channel.ts
@@ -284,9 +284,14 @@ export class ChannelsAPI {
 	public async edit(
 		channelId: Snowflake,
 		body: RESTPatchAPIChannelJSONBody,
-		{ auth, signal }: Pick<RequestData, 'auth' | 'signal'> = {},
+		{ auth, signal, reason }: Pick<RequestData, 'auth' | 'reason' | 'signal'> = {},
 	) {
-		return this.rest.patch(Routes.channel(channelId), { auth, body, signal }) as Promise<RESTPatchAPIChannelResult>;
+		return this.rest.patch(Routes.channel(channelId), {
+			auth,
+			reason,
+			body,
+			signal,
+		}) as Promise<RESTPatchAPIChannelResult>;
 	}
 
 	/**
@@ -296,8 +301,11 @@ export class ChannelsAPI {
 	 * @param channelId - The id of the channel to delete
 	 * @param options - The options for deleting the channel
 	 */
-	public async delete(channelId: Snowflake, { auth, signal }: Pick<RequestData, 'auth' | 'signal'> = {}) {
-		return this.rest.delete(Routes.channel(channelId), { auth, signal }) as Promise<RESTDeleteAPIChannelResult>;
+	public async delete(
+		channelId: Snowflake,
+		{ auth, signal, reason }: Pick<RequestData, 'auth' | 'reason' | 'signal'> = {},
+	) {
+		return this.rest.delete(Routes.channel(channelId), { auth, signal, reason }) as Promise<RESTDeleteAPIChannelResult>;
 	}
 
 	/**
@@ -511,12 +519,13 @@ export class ChannelsAPI {
 		channelId: Snowflake,
 		body: RESTPostAPIChannelThreadsJSONBody,
 		messageId?: Snowflake,
-		{ auth, signal }: Pick<RequestData, 'auth' | 'signal'> = {},
+		{ auth, signal, reason }: Pick<RequestData, 'auth' | 'reason' | 'signal'> = {},
 	) {
 		return this.rest.post(Routes.threads(channelId, messageId), {
 			auth,
 			body,
 			signal,
+			reason,
 		}) as Promise<RESTPostAPIChannelThreadsResult>;
 	}
 
@@ -531,7 +540,7 @@ export class ChannelsAPI {
 	public async createForumThread(
 		channelId: Snowflake,
 		{ message, ...optionsBody }: StartForumThreadOptions,
-		{ auth, signal }: Pick<RequestData, 'auth' | 'signal'> = {},
+		{ auth, signal, reason }: Pick<RequestData, 'auth' | 'reason' | 'signal'> = {},
 	) {
 		const { files, ...messageBody } = message;
 
@@ -540,7 +549,13 @@ export class ChannelsAPI {
 			message: messageBody,
 		};
 
-		return this.rest.post(Routes.threads(channelId), { auth, files, body, signal }) as Promise<APIThreadChannel>;
+		return this.rest.post(Routes.threads(channelId), {
+			auth,
+			files,
+			body,
+			reason,
+			signal,
+		}) as Promise<APIThreadChannel>;
 	}
 
 	/**
--- a/packages/core/src/api/guild.ts
+++ b/packages/core/src/api/guild.ts
@@ -187,11 +187,8 @@ export class GuildsAPI {
 	 * @param guildId - The id of the guild to delete
 	 * @param options - The options for deleting this guild
 	 */
-	public async delete(
-		guildId: Snowflake,
-		{ auth, reason, signal }: Pick<RequestData, 'auth' | 'reason' | 'signal'> = {},
-	) {
-		await this.rest.delete(Routes.guild(guildId), { auth, reason, signal });
+	public async delete(guildId: Snowflake, { auth, signal }: Pick<RequestData, 'auth' | 'signal'> = {}) {
+		await this.rest.delete(Routes.guild(guildId), { auth, signal });
 	}
 
 	/**
--- a/packages/discord.js/src/managers/GuildMemberRoleManager.js
+++ b/packages/discord.js/src/managers/GuildMemberRoleManager.js
@@ -72,7 +72,7 @@ class GuildMemberRoleManager extends DataManager {
    * @readonly
    */
   get color() {
-    const coloredRoles = this.cache.filter(role => role.color);
+    const coloredRoles = this.cache.filter(role => role.colors.primaryColor);
     if (!coloredRoles.size) return null;
     return coloredRoles.reduce((prev, role) => (role.comparePositionTo(prev) > 0 ? role : prev));
   }
--- a/packages/discord.js/src/managers/RoleManager.js
+++ b/packages/discord.js/src/managers/RoleManager.js
@@ -110,11 +110,21 @@ class RoleManager extends CachedManager {
    */
 
   /**
+   * @typedef {Object} RoleColorsResolvable
+   * @property {ColorResolvable} primaryColor The primary color of the role
+   * @property {ColorResolvable} [secondaryColor] The secondary color of the role.
+   * This will make the role a gradient between the other provided colors
+   * @property {ColorResolvable} [tertiaryColor] The tertiary color of the role.
+   * When sending `tertiaryColor` the API enforces the role color to be a holographic style with values of `primaryColor = 11127295`, `secondaryColor = 16759788`, and `tertiaryColor = 16761760`.
+   * These values are available as a constant: `Constants.HolographicStyle`
+   */
+
+  /**
    * Options used to create a new role.
    *
    * @typedef {Object} RoleCreateOptions
    * @property {string} [name] The name of the new role
-   * @property {ColorResolvable} [color] The data to create the role with
+   * @property {RoleColorsResolvable} [colors] The colors to create the role with
    * @property {boolean} [hoist] Whether or not the new role should be hoisted
    * @property {PermissionResolvable} [permissions] The permissions for the new role
    * @property {number} [position] The position of the new role
@@ -141,16 +151,30 @@ class RoleManager extends CachedManager {
    * // Create a new role with data and a reason
    * guild.roles.create({
    *   name: 'Super Cool Blue People',
-   *   color: Colors.Blue,
    *   reason: 'we needed a role for Super Cool People',
+   *   colors: {
+   *     primaryColor: Colors.Blue,
+   *   },
+   * })
+   *   .then(console.log)
+   *   .catch(console.error);
+   * @example
+   * // Create a role with holographic colors
+   * guild.roles.create({
+   *   name: 'Holographic Role',
+   *   reason: 'Creating a role with holographic effect',
+   *   colors: {
+   *     primaryColor: Constants.HolographicStyle.Primary,
+   *     secondaryColor: Constants.HolographicStyle.Secondary,
+   *     tertiaryColor: Constants.HolographicStyle.Tertiary,
+   *   },
    * })
    *   .then(console.log)
    *   .catch(console.error);
    */
   async create(options = {}) {
-    let { color, permissions, icon } = options;
+    let { permissions, icon } = options;
     const { name, hoist, position, mentionable, reason, unicodeEmoji } = options;
-    color &&= resolveColor(color);
     if (permissions !== undefined) permissions = new PermissionsBitField(permissions);
     if (icon) {
       const guildEmojiURL = this.guild.emojis.resolve(icon)?.imageURL();
@@ -158,10 +182,16 @@ class RoleManager extends CachedManager {
       if (typeof icon !== 'string') icon = undefined;
     }
 
+    const colors = options.colors && {
+      primary_color: resolveColor(options.colors.primaryColor),
+      secondary_color: options.colors.secondaryColor && resolveColor(options.colors.secondaryColor),
+      tertiary_color: options.colors.tertiaryColor && resolveColor(options.colors.tertiaryColor),
+    };
+
     const data = await this.client.rest.post(Routes.guildRoles(this.guild.id), {
       body: {
         name,
-        color,
+        colors,
         hoist,
         permissions,
         mentionable,
@@ -212,9 +242,15 @@ class RoleManager extends CachedManager {
       if (typeof icon !== 'string') icon = undefined;
     }
 
+    const colors = options.colors && {
+      primary_color: resolveColor(options.colors.primaryColor),
+      secondary_color: options.colors.secondaryColor && resolveColor(options.colors.secondaryColor),
+      tertiary_color: options.colors.tertiaryColor && resolveColor(options.colors.tertiaryColor),
+    };
+
     const body = {
       name: options.name,
-      color: options.color === undefined ? undefined : resolveColor(options.color),
+      colors,
       hoist: options.hoist,
       permissions: options.permissions === undefined ? undefined : new PermissionsBitField(options.permissions),
       mentionable: options.mentionable,
--- a/packages/discord.js/src/structures/Role.js
+++ b/packages/discord.js/src/structures/Role.js
@@ -57,13 +57,27 @@ class Role extends Base {
       this.name = data.name;
     }
 
-    if ('color' in data) {
+    /**
+     * @typedef {Object} RoleColors
+     * @property {number} primaryColor The primary color of the role
+     * @property {?number} secondaryColor The secondary color of the role.
+     * This will make the role a gradient between the other provided colors
+     * @property {?number} tertiaryColor The tertiary color of the role.
+     * When sending `tertiaryColor` the API enforces the role color to be a holographic style with values of `primaryColor = 11127295`, `secondaryColor = 16759788`, and `tertiaryColor = 16761760`.
+     * These values are available as a constant: `Constants.HolographicStyle`
+     */
+
+    if ('colors' in data) {
       /**
-       * The base 10 color of the role
+       * The colors of the role
        *
-       * @type {number}
+       * @type {RoleColors}
        */
-      this.color = data.color;
+      this.colors = {
+        primaryColor: data.colors.primary_color,
+        secondaryColor: data.colors.secondary_color,
+        tertiaryColor: data.colors.tertiary_color,
+      };
     }
 
     if ('hoist' in data) {
@@ -257,7 +271,7 @@ class Role extends Base {
    *
    * @typedef {Object} RoleData
    * @property {string} [name] The name of the role
-   * @property {ColorResolvable} [color] The color of the role, either a hex string or a base 10 number
+   * @property {RoleColorsResolvable} [colors] The colors of the role
    * @property {boolean} [hoist] Whether or not the role should be hoisted
    * @property {number} [position] The position of the role
    * @property {PermissionResolvable} [permissions] The permissions of the role
@@ -315,19 +329,28 @@ class Role extends Base {
   }
 
   /**
-   * Sets a new color for the role.
+   * Sets new colors for the role.
    *
-   * @param {ColorResolvable} color The color of the role
-   * @param {string} [reason] Reason for changing the role's color
+   * @param {RoleColorsResolvable} colors The colors of the role
+   * @param {string} [reason] Reason for changing the role's colors
    * @returns {Promise<Role>}
    * @example
-   * // Set the color of a role
-   * role.setColor('#FF0000')
-   *   .then(updated => console.log(`Set color of role to ${updated.color}`))
+   * // Set the colors of a role
+   * role.setColors({ primaryColor: '#FF0000', secondaryColor: '#00FF00', tertiaryColor: '#0000FF' })
+   *   .then(updated => console.log(`Set colors of role to ${updated.colors}`))
+   *   .catch(console.error);
+   * @example
+   * // Set holographic colors using constants
+   * role.setColors({
+   *   primaryColor: Constants.HolographicStyle.Primary,
+   *   secondaryColor: Constants.HolographicStyle.Secondary,
+   *   tertiaryColor: Constants.HolographicStyle.Tertiary,
+   * })
+   *   .then(updated => console.log(`Set holographic colors for role ${updated.name}`))
    *   .catch(console.error);
    */
-  async setColor(color, reason) {
-    return this.edit({ color, reason });
+  async setColors(colors, reason) {
+    return this.edit({ colors, reason });
   }
 
   /**
@@ -475,7 +498,9 @@ class Role extends Base {
       role &&
       this.id === role.id &&
       this.name === role.name &&
-      this.color === role.color &&
+      this.colors.primaryColor === role.colors.primaryColor &&
+      this.colors.secondaryColor === role.colors.secondaryColor &&
+      this.colors.tertiaryColor === role.colors.tertiaryColor &&
       this.hoist === role.hoist &&
       this.position === role.position &&
       this.permissions.bitfield === role.permissions.bitfield &&
--- a/packages/discord.js/src/util/Constants.js
+++ b/packages/discord.js/src/util/Constants.js
@@ -223,6 +223,21 @@ exports.StickerFormatExtensionMap = {
 };
 
 /**
+ * Holographic color values for role styling.
+ * When using `tertiaryColor`, the API enforces these specific values for holographic effect.
+ *
+ * @typedef {Object} HolographicStyle
+ * @property {number} Primary 11127295 (0xA9FFFF)
+ * @property {number} Secondary 16759788 (0xFFCCCC)
+ * @property {number} Tertiary 16761760 (0xFFE0A0)
+ */
+exports.HolographicStyle = {
+  Primary: 11_127_295,
+  Secondary: 16_759_788,
+  Tertiary: 16_761_760,
+};
+
+/**
  * @typedef {Object} Constants Constants that can be used in an enum or object-like way.
  * @property {number} MaxBulkDeletableMessageAge Max bulk deletable message age
  * @property {SweeperKey[]} SweeperKeys The possible names of items that can be swept in sweepers
@@ -232,4 +247,5 @@ exports.StickerFormatExtensionMap = {
  * @property {VoiceBasedChannelTypes} VoiceBasedChannelTypes The types of channels that are voice-based
  * @property {SelectMenuTypes} SelectMenuTypes The types of components that are select menus.
  * @property {Object} StickerFormatExtensionMap A mapping between sticker formats and their respective image formats.
+ * @property {HolographicStyle} HolographicStyle Holographic color values for role styling.
  */
--- a/packages/discord.js/typings/index.d.ts
+++ b/packages/discord.js/typings/index.d.ts
@@ -2823,9 +2823,21 @@ export class RichPresenceAssets {
   public smallImageURL(options?: ImageURLOptions): string | null;
 }
 
+export interface RoleColors {
+  primaryColor: number;
+  secondaryColor: number | null;
+  tertiaryColor: number | null;
+}
+
+export interface RoleColorsResolvable {
+  primaryColor: ColorResolvable;
+  secondaryColor?: ColorResolvable;
+  tertiaryColor?: ColorResolvable;
+}
+
 export class Role extends Base {
   private constructor(client: Client<true>, data: APIRole, guild: Guild);
-  public color: number;
+  public colors: RoleColors;
   public get createdAt(): Date;
   public get createdTimestamp(): number;
   public get editable(): boolean;
@@ -2853,7 +2865,7 @@ export class Role extends Base {
     channel: NonThreadGuildBasedChannel | Snowflake,
     checkAdmin?: boolean,
   ): Readonly<PermissionsBitField>;
-  public setColor(color: ColorResolvable, reason?: string): Promise<Role>;
+  public setColors(colors: RoleColorsResolvable, reason?: string): Promise<Role>;
   public setHoist(hoist?: boolean, reason?: string): Promise<Role>;
   public setMentionable(mentionable?: boolean, reason?: string): Promise<Role>;
   public setName(name: string, reason?: string): Promise<Role>;
@@ -3808,6 +3820,11 @@ export type UndeletableMessageType =
 
 export const Constants: {
   GuildTextBasedChannelTypes: GuildTextBasedChannelTypes[];
+  HolographicStyle: {
+    Primary: 11_127_295;
+    Secondary: 16_759_788;
+    Tertiary: 16_761_760;
+  };
   MaxBulkDeletableMessageAge: 1_209_600_000;
   NonSystemMessageTypes: NonSystemMessageType[];
   SelectMenuTypes: SelectMenuType[];
@@ -6836,7 +6853,7 @@ export interface ResolvedOverwriteOptions {
 }
 
 export interface RoleData {
-  color?: ColorResolvable;
+  colors?: RoleColorsResolvable;
   hoist?: boolean;
   icon?: Base64Resolvable | BufferResolvable | EmojiResolvable | null;
   mentionable?: boolean;
--- a/packages/rest/__tests__/CDN.test.ts
+++ b/packages/rest/__tests__/CDN.test.ts
@@ -8,7 +8,7 @@ const hash = 'abcdef';
 const animatedHash = 'a_bcdef';
 const defaultAvatar = 1_234 % 5;
 
-const cdn = new CDN(baseCDN, baseMedia);
+const cdn = new CDN({ cdn: baseCDN, mediaProxy: baseMedia });
 
 test('appAsset default', () => {
 	expect(cdn.appAsset(id, hash)).toEqual(`${baseCDN}/app-assets/${id}/${hash}.webp`);
--- a/packages/rest/src/lib/CDN.ts
+++ b/packages/rest/src/lib/CDN.ts
@@ -84,13 +84,35 @@ interface MakeURLOptions {
 }
 
 /**
+ * Options for initializing the {@link CDN} class.
+ */
+export interface CDNOptions {
+	/**
+	 * The base URL for the CDN.
+	 *
+	 * @defaultValue `DefaultRestOptions.cdn`
+	 */
+	cdn?: string | undefined;
+	/**
+	 * The base URL for the media proxy.
+	 *
+	 * @defaultValue `DefaultRestOptions.mediaProxy`
+	 */
+	mediaProxy?: string | undefined;
+}
+
+/**
  * The CDN link builder
  */
 export class CDN {
-	public constructor(
-		private readonly cdn: string = DefaultRestOptions.cdn,
-		private readonly mediaProxy: string = DefaultRestOptions.mediaProxy,
-	) {}
+	private readonly cdn: string;
+
+	private readonly mediaProxy: string;
+
+	public constructor({ cdn, mediaProxy }: CDNOptions = {}) {
+		this.cdn = cdn ?? DefaultRestOptions.cdn;
+		this.mediaProxy = mediaProxy ?? DefaultRestOptions.mediaProxy;
+	}
 
 	/**
 	 * Generates an app asset URL for a client's asset.
--- a/packages/rest/src/lib/REST.ts
+++ b/packages/rest/src/lib/REST.ts
@@ -78,7 +78,7 @@ export class REST extends AsyncEventEmitter<RestEvents> {
 
 	public constructor(options: Partial<RESTOptions> = {}) {
 		super();
-		this.cdn = new CDN(options.cdn ?? DefaultRestOptions.cdn, options.mediaProxy ?? DefaultRestOptions.mediaProxy);
+		this.cdn = new CDN(options);
 		this.options = { ...DefaultRestOptions, ...options };
 		this.globalRemaining = Math.max(1, this.options.globalRequestsPerSecond);
 		this.agent = options.agent ?? null;


packages/structures/src/channels/Channel.ts
```diff
@@ -0,0 +1,203 @@
...
+	? APIThreadChannel
+	: Extract<APIChannel, { type: Type }>; // TODO: remove special handling once dtypes PR for thread channel types releases
+
+/**
+ * Represents any channel on Discord.
+ *
+ * @typeParam Type - Specify the type of the channel being constructed for more accurate data types
+ * @typeParam Omitted - Specify the propeties that will not be stored in the raw data field as a union, implement via `DataTemplate`
+ * @remarks While this class can be directly constructed for any channel type, this class is not intended to be used directly,
+ * preferring instead to create class that extend this class with the appropriate mixins for each channel type.
```
almeidx, line 41:
```suggestion
 * @remarks Although this class _can_ be instantiated directly for any channel type,
 * it's intended to be subclassed with the appropriate mixins for each channel type.
```

packages/structures/src/channels/Channel.ts
```diff
@@ -0,0 +1,203 @@
...
+	| ChannelType.PrivateThread
+	| ChannelType.PublicThread
+	? APIThreadChannel
+	: Extract<APIChannel, { type: Type }>; // TODO: remove special handling once dtypes PR for thread channel types releases
+
+/**
+ * Represents any channel on Discord.
+ *
+ * @typeParam Type - Specify the type of the channel being constructed for more accurate data types
+ * @typeParam Omitted - Specify the propeties that will not be stored in the raw data field as a union, implement via `DataTemplate`
```
almeidx, line 39:
```suggestion
 * @typeParam Omitted - Specify the properties that will not be stored in the raw data field as a union, implement via `DataTemplate`
```

packages/structures/src/channels/ForumTag.ts
```diff
@@ -0,0 +1,52 @@
+import type { APIGuildForumTag } from 'discord-api-types/v10';
+import { Structure } from '../Structure.js';
+import { kData } from '../utils/symbols.js';
+
+/**
+ * Represents metadata of a thread channel on Discord.
+ *
+ * @typeParam Omitted - Specify the propeties that will not be stored in the raw data field as a union, implement via `DataTemplate`
```
almeidx, line 8:
```suggestion
 * @typeParam Omitted - Specify the properties that will not be stored in the raw data field as a union, implement via `DataTemplate`
```

packages/structures/src/channels/PermissionOverwrite.ts
```diff
@@ -0,0 +1,87 @@
+import type { APIOverwrite } from 'discord-api-types/v10';
+import { Structure } from '../Structure.js';
+import { kAllow, kData, kDeny } from '../utils/symbols.js';
+
+/**
+ * Represents metadata of a thread channel on Discord.
+ *
+ * @typeParam Omitted - Specify the propeties that will not be stored in the raw data field as a union, implement via `DataTemplate`
```
almeidx, line 8:
```suggestion
 * @typeParam Omitted - Specify the properties that will not be stored in the raw data field as a union, implement via `DataTemplate`
```

packages/structures/src/invites/Invite.ts
```diff
@@ -0,0 +1,218 @@
...
+import { kData, kExpiresTimestamp, kCreatedTimestamp } from '../utils/symbols.js';
+
+// TODO: use an actual dtypes type instead if it exists
+export interface APIActualInvite extends APIInvite, Partial<Omit<APIExtendedInvite, keyof APIInvite>> {}
+
+/**
+ * Represents an invitation to a discord channel
+ *
+ * @typeParam Omitted - Specify the propeties that will not be stored in the raw data field as a union, implement via `DataTemplate`
+ * @typeParam Extended - Whether the invite is a full extended invite
```
almeidx, line 12:
```suggestion
 * @typeParam Omitted - Specify the properties that will not be stored in the raw data field as a union, implement via `DataTemplate`
```

packages/structures/src/users/Connection.ts
```diff
@@ -0,0 +1,91 @@
+import type { APIConnection } from 'discord-api-types/v10';
+import { Structure } from '../Structure.js';
+import { kData } from '../utils/symbols.js';
+
+/**
+ * Represents a user's connection on Discord.
+ *
+ * @typeParam Omitted - Specify the propeties that will not be stored in the raw data field as a union, implement via `DataTemplate`
```
almeidx, line 8:
```suggestion
 * @typeParam Omitted - Specify the properties that will not be stored in the raw data field as a union, implement via `DataTemplate`
```

packages/structures/src/users/User.ts
```diff
@@ -0,0 +1,192 @@
+import { DiscordSnowflake } from '@sapphire/snowflake';
+import type { APIUser } from 'discord-api-types/v10';
+import { Structure } from '../Structure.js';
+import { kData } from '../utils/symbols.js';
+
+/**
+ * Represents any user on Discord.
+ *
+ * @typeParam Omitted - Specify the propeties that will not be stored in the raw data field as a union, implement via `DataTemplate`
```
almeidx, line 9:
```suggestion
 * @typeParam Omitted - Specify the properties that will not be stored in the raw data field as a union, implement via `DataTemplate`
```

packages/structures/src/channels/ThreadMetadata.ts
```diff
@@ -0,0 +1,121 @@
+import type { APIThreadMetadata } from 'discord-api-types/v10';
+import { Structure } from '../Structure.js';
+import { kArchiveTimestamp, kCreatedTimestamp, kData } from '../utils/symbols.js';
+
+/**
+ * Represents metadata of a thread channel on Discord.
+ *
+ * @typeParam Omitted - Specify the propeties that will not be stored in the raw data field as a union, implement via `DataTemplate`
```
almeidx, line 8:
```suggestion
 * @typeParam Omitted - Specify the properties that will not be stored in the raw data field as a union, implement via `DataTemplate`
```

packages/structures/src/users/User.ts
```diff
@@ -0,0 +1,192 @@
...
+	}
+
+	/**
+	 * The hexadecimal version of the user accent color, with a leading hash
+	 *
+	 * @remarks This property is only set when the user was manually fetched
+	 */
+	public get hexAccentColor() {
+		if (typeof this.accentColor !== 'number') return this.accentColor;
+		return `#${this.accentColor.toString(16).padStart(6, '0')}`;
```
almeidx, line 190:
```suggestion
		const accentColor = this.accentColor;
		if (typeof accentColor !== 'number') return accentColor;
		return `#${accentColor.toString(16).padStart(6, '0')}`;
```

almeidx (dismissed):

packages/structures/src/Structure.ts
```diff
@@ -0,0 +1,133 @@
...
+	 */
+	protected [kMixinConstruct]?(data: Partial<DataType>): void;
+
+	/**
+	 * The template used for removing data from the raw data stored for each Structure.
+	 *
+	 * @remarks This template should be overriden in all subclasses to provide more accurate type information.
+	 * The template in the base {@link Structure} class will have no effect on most subclasses for this reason.
+	 */
+	protected static DataTemplate: Record<string, unknown> = {};
```
didinele: perhaps I'm wrong, but I don't see us mutating this anywhere

```suggestion
	protected readonly static DataTemplate: Record<string, unknown> = {};
```

or sorry, I forget which one our linter likes, might be
```suggestion
	protected static readonly DataTemplate: Record<string, unknown> = {};
```

packages/structures/src/Structure.ts
```diff
@@ -0,0 +1,133 @@
...
+
+	/**
+	 * Creates a new structure to represent API data
+	 *
+	 * @param data - the data from the API that this structure will represent
+	 * @param extraOptions - any additional options needed to appropriately construct a structure from the data
+	 * @remarks To be made public in subclasses
+	 * @internal
+	 */
+	protected constructor(data: Readonly<Partial<DataType>>, ..._rest: unknown[]) {
```
didinele: Semantically, I think I'd prefer if we went with public here. Protected constructors feel a bit silly when the class is already abstract and requires extending.

```suggestion
	public constructor(data: Readonly<Partial<DataType>>, ..._rest: unknown[]) {
```

packages/structures/src/Structure.ts
```diff
@@ -0,0 +1,133 @@
...
+	/**
+	 * Patches the raw data of this object in place
+	 *
+	 * @param data - the updated data from the API to patch with
+	 * @param extraOptions - any additional options needed to appropriately patch the data
+	 * @remarks To be made public in subclasses
+	 * @returns this
+	 * @internal
+	 */
+	protected _patch(data: Readonly<Partial<DataType>>): this {
```
didinele: So, are we going to stick with `_` prefixes for private/protected methods? As far as I know, none of the new codebases really do this (and if some do, we should agree on something and stick to it everywhere/refactor accordingly)

packages/structures/src/bitfields/BitField.ts
```diff
@@ -0,0 +1,205 @@
...
+
+/**
+ * Data structure that makes it easy to interact with a bitfield.
+ */
+export abstract class BitField<Flags extends string, Type extends bigint | number = number> {
+	/**
+	 * Numeric bitfield flags.
+	 * <info>Defined in extension classes</info>
+	 */
+	public static Flags: EnumLike<unknown, bigint | number> = {};
```
didinele: What was the thought process behind `DefaultBit` being typed like this (I guess Flags you can sort of justify it, though I'd probably go `Type | string` there too) despite the `Type` generic? I do see the way we handle it in `resolve` but I feel like it's begging for trouble? Imagine ending up with this by refactoring

```ts
class Something extends BitField<Flags, bigint> { // just changed to bigint
  public static override Flags = whatever; // contains a bigint that would be too big for number representation
  public static DefaultBit = 123; // uh-oh! forgot to change this. everything breaks because the runtime checks for `typeof DefaultBit` and coerces things in `Flags` to that type
}
```

packages/structures/src/Structure.ts
```diff
@@ -0,0 +1,133 @@
...
+	 * For typescript users, mixins must use the closest ancestors access modifier.
+	 * @remarks Automatically called in {@link Structure#_patch} but must be called manually in the constructor
+	 * of any class implementing this method.
+	 * @remarks Additionally, when implementing, ensure to call `super._optimizeData` if any class in the super chain aside
+	 * from Structure contains an implementation.
+	 * Note: mixins do not need to call super ever as the process of mixing walks the prototype chain.
+	 * @virtual
+	 * @internal
+	 */
+	protected _optimizeData(_data: Partial<DataType>) {}
```
didinele: Why not `abstract`? So classes that don't have any work to do here don't need to implement? Not sure I love that, we could go this route instead if we want to avoid empty implementations everywhere:
```suggestion
	protected _optimizeData?(_data: Partial<DataType>): void;
```
and calls do `this._optimizeData.?();` (`kMixinConstruct` also follows this pattern)

(notice the lack of an `abstract` keyword, that actually requires explicitly specifying it in the subclass even if you don't want to implement it)

packages/structures/src/Structure.ts
```diff
@@ -0,0 +1,133 @@
...
+	 * automatically called by `JSON.stringify()` when this structure is stringified
+	 *
+	 * @remarks
+	 * The type of this data is determined by omissions at runtime and is only guaranteed for default omissions
+	 * @privateRemarks
+	 * When omitting properties at the library level, this must be overriden to re-add those properties
+	 */
+	public toJSON(): DataType {
+		// This will be DataType provided nothing is omitted, when omits occur, subclass needs to overwrite this.
+		return { ...this[kData] } as DataType;
```
didinele: So, we have no guarantee this is actually a proper clone, right? i.e. `kData` can have objects as properties

Doesn't this fundamentally break, well, everything? `_clone()` will end up yielding references for some of those props and mutations to a totally new `_clone()`d instance will carry over to the current class as well.

We probably want as we do in builders:
```suggestion
		return structuredClone(this[kData]) as DataType;
```

packages/structures/src/bitfields/BitField.ts
didinele: readonly!

packages/structures/src/bitfields/BitField.ts
```diff
@@ -0,0 +1,205 @@
...
+
+	protected static DefaultBit: bigint | number = 0;
```
didinele: readonly!

packages/structures/src/bitfields/PermissionsBitField.ts
```diff
@@ -0,0 +1,78 @@
...
+	 * Numeric permission flags.
+	 *
+	 * @see {@link https://discord.com/developers/docs/topics/permissions#permissions-bitwise-permission-flags}
+	 */
+	public static override Flags = PermissionFlagsBits;
+
+	/**
+	 * Bitfield representing every permission combined
+	 */
+	public static All = Object.values(PermissionFlagsBits).reduce((all, perm) => all | perm, 0n);
```
didinele: ditto

packages/structures/src/bitfields/PermissionsBitField.ts
```diff
@@ -0,0 +1,78 @@
...
+
+	/**
+	 * Bitfield representing the default permissions for users
+	 */
+	public static Default = BigInt(104_324_673);
```
didinele: ditto

packages/structures/src/bitfields/PermissionsBitField.ts
```diff
@@ -0,0 +1,78 @@
...
+
+	/**
+	 * Bitfield representing the permissions required for moderators of stage channels
+	 */
+	public static StageModerator =
```
didinele: ditto

packages/structures/src/utils/types.ts
```diff
@@ -0,0 +1,32 @@
...
+		? MergePrototype<Class1, MergePrototypes<Rest>>
+		: never;
+
+export interface RecursiveReadonlyArray<ItemType> extends ReadonlyArray<ItemType | RecursiveReadonlyArray<ItemType>> {}
+
+export type EnumLike<Enum, Value> = Record<keyof Enum, Value>;
+
+export type If<Check, Value, True, False = never> = Check extends Value ? (Value extends Check ? True : False) : False;
+
+// TODO: remove helper type once dtypes PR for thread channel types releases
```
didinele: 👀

didinele (changes requested):
about half the files viewed, more to come later

packages/structures/src/bitfields/BitField.ts
Qjuh: static properties can't use the generic, as it's only present on instances, not the class itself.

packages/structures/src/Structure.ts
Qjuh: This method gets defined/overridden in the Mixin process. Making it optional instead of an empty function body would mean that every implementation of a Structure would have to do the optional chaining of that method call in their constructor. `kMixinConstruct` works completely different, as it's not a method that gets overridden but instead all of them get combined into a single function that gets called in the `Structure` base class here if present. Implementers don't need to worry about that optionality at all.

packages/structures/src/bitfields/BitField.ts
didinele: Right. That's quite annoying. Anyone else have ideas on how to improve the safety here?

packages/structures/src/Structure.ts
didinele: alright, I'm split on it then; up to others

packages/structures/src/Structure.ts
Qjuh: Ideally substructures (which objects as properties technically are) would be cloned in their `_toJSON()` applied after this in the Mixin or in an overridden `toJSON()` method in the extending class. When that happens the `structuredClone` of that object would be completely replaced/overwritten again and we have unneeded overhead. So the real question is if we want to rely on implementation always overriding in Mixin or extended class if there are substructures or if we want to go the safe (but less performant) route of always deep cloning (structuredClone).

didinele: If that's the design pattern we're opting for and we can accomplish this without `structuredClone`, then I think with some decent unit test coverage we can afford going the the override route

packages/structures/src/bitfields/BitField.ts
vladfrangu: Maybe inspire from [@sapphire/bitfield](https://www.npmjs.com/package/@sapphire/bitfield) and make the enum be passed in? But then we'd basically have a factory of bitfields 🫠...

packages/structures/README.md
```diff
@@ -0,0 +1,68 @@
...
+	</p>
+</div>
+
+## About
+
+`@discordjs/structures` is a low level wrapper around Discord JSON Objects, meant to be a foundation to build upon in a higher level library.
+
+## Installation
+
+**Node.js 18.17.1 or newer is required.**
```
vladfrangu: Nit: bump to 20+

packages/structures/package.json
```diff
@@ -0,0 +1,97 @@
...
+		"lib": "src",
+		"test": "__tests__"
+	},
+	"files": [
+		"dist"
+	],
+	"contributors": [
+		"Crawl <icrawltogo@gmail.com>",
+		"SpaceEEC <spaceeec@yahoo.com>",
+		"Vlad Frangu <kingdgrizzle@gmail.com>",
```
vladfrangu: Nit 2: emails need to update from other pkgs

packages/structures/src/Structure.ts
Qjuh: You're right, only `@discordjs/voice` does and only for private methods, and even there it's probably more historic reasons and not a current decision. Will remove the prefix.

packages/structures/src/Mixin.ts
```diff
@@ -0,0 +1,194 @@
...
+			const usingDescriptors: { [prop: string]: PropertyDescriptor } = {};
+			for (const [prop, descriptor] of Object.entries(originalDescriptors)) {
+				// Drop constructor
+				if (['constructor'].includes(prop)) {
+					continue;
+				}
+
+				// Special case for optimize function, we want to combine these
+				if (prop === OptimizeDataPropertyName) {
+					if (typeof descriptor.value !== 'function') return;
```
vladfrangu: Returning with no output means this will never finish, this sounds more like a throw RangeError situation

packages/structures/src/Mixin.ts
```diff
@@ -0,0 +1,194 @@
...
+					dataOptimizations.push(descriptor.value);
+					continue;
+				} else if (prop === EnrichToJSONPropertyName) {
+					if (typeof descriptor.value !== 'function') return;
```
vladfrangu: Ditto

packages/structures/src/Structure.ts
```diff
@@ -0,0 +1,133 @@
...
+/**
+ * Additional options needed to appropriately construct / patch data
+ *
+ * @internal
+ */
+// eslint-disable-next-line @typescript-eslint/no-empty-interface
+export interface StructureExtraOptions {}
+
+export const DataTemplatePropertyName = 'DataTemplate';
+export const OptimizeDataPropertyName = '_optimizeData';
```
vladfrangu: What if we also make these symbols instead of strings? Just a thought, not a request

packages/structures/src/Structure.ts
```diff
@@ -0,0 +1,133 @@
+/* eslint-disable jsdoc/check-param-names */
+import { kData, kMixinConstruct } from './utils/symbols.js';
+import type { ReplaceOmittedWithUnknown } from './utils/types.js';
+
+/**
+ * Additional options needed to appropriately construct / patch data
+ *
+ * @internal
+ */
+// eslint-disable-next-line @typescript-eslint/no-empty-interface
```
vladfrangu: I'd disable this rule in eslint for this entire package

packages/structures/src/bitfields/BitField.ts
```diff
@@ -0,0 +1,205 @@
...
+	/**
+	 * Bitfield of the packed bits
+	 */
+	public bitfield: Type;
+
+	/**
+	 * @param bits - Bit(s) to read from
+	 */
+	public constructor(
+		bits: BitFieldResolvable<Flags, Type> = (this.constructor as typeof BitField<Flags, Type>).DefaultBit as Type,
```
vladfrangu: pro tip: you can define `['constructor']: typeof BitField<Flags, Type>` on the class level to skip all this casting, or make a getter function for it (https://www.typescriptlang.org/play/?#code/MYGwhgzhAECiAeYC2AHEBTAPAdQPYHdp14AXdAOwBMYByMG6AH2hoCMaA+aAbwCgBISulBgATumgQSYEgEtgRRKgwAFUbhQAuaCVEBXdAG5eAoSPHQA2jWC5yU-cBK5RNALraSATxTpcAMzglNCw8fA5jU2FwC29fSgA1MBADbTDI-n89cgAKAEoeAX5ZQJySAAtZCAA6OPRE5INoAF5WlnoCvn5uiqrq23tdPScXauJkELUNIv4AelnoRaXllYA9AH4igF8TbvnoAEEQMlFyGVkAN3Rd-l6akdEx4NV1FCL9laWNgR2BAHN0CRoA98oVuuISHpTjpKvc7A5hs5RNBIDofH5AggJhgcAQIj9eFsgA)

packages/structures/src/bitfields/BitField.ts
```diff
@@ -0,0 +1,205 @@
...
+ * Data structure that makes it easy to interact with a bitfield.
+ */
+export abstract class BitField<Flags extends string, Type extends bigint | number = number> {
+	/**
+	 * Numeric bitfield flags.
+	 * <info>Defined in extension classes</info>
+	 */
+	public static readonly Flags: EnumLike<unknown, bigint | number> = {};
+
+	protected static readonly DefaultBit: bigint | number = 0;
```
vladfrangu: Random thought, cc @didinele, what if we just always enforce bigints for bitfields? And we convert to string/number in djs as needed? Makes it far easier to work with in the future if an enum moves from numbers to bigints

packages/structures/src/bitfields/PermissionsBitField.ts
```diff
@@ -0,0 +1,78 @@
...
+
+	/**
+	 * Bitfield representing every permission combined
+	 */
+	public static readonly All = Object.values(PermissionFlagsBits).reduce((all, perm) => all | perm, 0n);
+
+	/**
+	 * Bitfield representing the default permissions for users
+	 */
+	public static readonly Default = BigInt(104_324_673);
```
vladfrangu:
```suggestion
	public static readonly Default = 104_324_673n;
```

packages/structures/src/bitfields/PermissionsBitField.ts
```diff
@@ -0,0 +1,78 @@
...
+
+	/**
+	 * Bitfield representing the permissions required for moderators of stage channels
+	 */
+	public static readonly StageModerator =
+		PermissionFlagsBits.ManageChannels | PermissionFlagsBits.MuteMembers | PermissionFlagsBits.MoveMembers;
+
+	protected static override DefaultBit = BigInt(0);
```
vladfrangu:
```suggestion
	protected static override DefaultBit = 0n;
```

packages/structures/src/channels/Channel.ts
```diff
@@ -0,0 +1,205 @@
...
+
+/**
+ * @internal
+ */
+export type ChannelDataType<Type extends ChannelType> = Type extends
+	| ChannelType.AnnouncementThread
+	| ChannelType.PrivateThread
+	| ChannelType.PublicThread
+	? APIThreadChannel<Type>
+	: Extract<APIChannel, { type: Type }>; // TODO: this still breaks without the (new) special handling, find out why
```
vladfrangu: Confused about the todo and why we don't just do `APIChannel & { type: Type }`

packages/structures/src/Structure.ts
Qjuh: Just realized that the `_toJSON()` pattern in the mixins need a slight adjustment to get it in line with `[kMixinConstruct]` and the omission of the _ prefix on protected methods.

packages/structures/src/users/User.ts
```diff
@@ -0,0 +1,178 @@
...
+	 */
+	public get premiumType() {
+		return this[kData].premium_type;
+	}
+
+	/**
+	 * The timestamp the user was created at
+	 */
+	public get createdTimestamp() {
+		return typeof this.id === 'string' || typeof this.id === 'bigint' ? DiscordSnowflake.timestampFrom(this.id) : null;
```
vladfrangu: If this becomes a common check, please make a typeguard function for this and reuse it

```ts
function isIdSet(id: unknown): id is string | bigint {
	return typeof id === 'string' || typeof id === 'bigint';
}
```

vladfrangu (changes requested):

packages/structures/src/Structure.ts
vladfrangu: Circling back to this, I'm starting to lean more towards using symbols for all internal functions (even patch) but I'm not fully set to just doing that... discussion needed

packages/structures/src/bitfields/BitField.ts
didinele: Well, I was going to initially say, is there a reason why we want `number`s at all? Is the BigInt overhead *that big* for numbers that would fit in `number` anyway?

packages/structures/src/Structure.ts
didinele: I'm not big on this at all. Is there a good reason why we'd want to do this? Even with the properties I'm a little iffy, I think that's a justified evil because you can't have `#member`s accessible in child classes

I think with properties where we absolutely know the user won't want to meddle it makes sense, but methods don't really fall under that, it also makes implementation much more annoying.

We have to think here from the perspective that this package has 2 layers of users:
- implementors (ourselves, and let's not forget that we want to allow others as well)
- actual end-users that interact with class instances

packages/structures/src/bitfields/BitField.ts
didinele: I'm a big fan of sapphire/bitfield 😅, if we can accustom our users to it (and it doesn't make some sort of abstraction very difficult), I personally wouldn't be against just using it in our stuff.

packages/structures/src/bitfields/BitField.ts
Qjuh: This works nicely for properties, but fails on `new this.constructor(...)` because `BitField` is abstract. Any ideas for those cases too? (https://www.typescriptlang.org/play/?ssl=23&ssc=1&pln=24&pc=1#code/IYIwzgLgTsDGEAJYBthjAgogD2AWwAdkBTAHgHUB7AdwWOwmIDsATDAcmHYQB8F2Q7AHwIA3gCgAkC2IpgUYgkjAIAS1h1chEgAUolAgC4E0AK7EA3OKky5ChAG12sSk0hRT8SlHYBdYxAAngTElABmWFpEZFTUQlY2sqj2QSEsAGrAyObGsQmSYaZMABQAlGJSkqoRxRAAFqpgAHSpxBlZ5ggAvD38XOUSkkP1jU0ubmZeUE30+NF6BpWSAPTLCOsbm1sAegD8lQC+1kOrCACCyIxQTCqqAG7Ex5IjzVMzUbr6BJWnWxt7TyYxFoLzGrncngg3jKVhOywBkiOUgA5sREFMyhUhgoIKZriYGq9wZMoVAEGgTMFQhEcHMSBQaPEpEcDkA)

packages/structures/src/channels/Channel.ts
Qjuh: I'm confused too, hence the "find out why". But I assume it is because `APIPublicThreadChannel` etc. are types and not interfaces. Will try something and maybe PR another dtypes if that's the issue.

packages/structures/src/users/User.ts
Qjuh: as standalone function in utils/ you mean?

packages/structures/src/bitfields/BitField.ts
Qjuh: We do need a way to know how to JSONify it (as string or number). If it's not based on type of the `bitfield` property then we need to store that information separately.

packages/structures/src/bitfields/BitField.ts
didinele: Initially I came up with
```ts
type NonAbstract<Constructor> = Constructor extends abstract new (...args: infer Args) => infer Instance ? new (...args: Args) => Instance : never; 
```

but that got rid of static props, so with some tweaking I got this:
https://www.typescriptlang.org/play/?#code/C4TwDgpgBAcg9gOwIICMDOwBOBDAxsAHgBUoIAPYCBAEzSm3Sz2CgQgHcoAKAOj+0wBzNAC56CEAEooAXgB84kApkAoKFBLlKNOgww58rDtz48BwsQEsEAMwiYoSIWmnyo1uw4CSCDNgS4EGrqUAD83GycvPzOYk7Crgo+fgEQ0gBkUAAKlrgA1sQANFB5ECBwNhpywepibABu9gDcKq16TIa4ADbYaHQAomTYALZgXRAEAOpwnFpUtFAA5NiLUAA+SyiLCgDeKgCQ1BDdAtB+wLmkQ6PjWZhwYGJYAK4QLQdHJ5jQANqLuIh9M98HBMIsALpieDIRgGQigSAVKCDEZjCbTdhyOTvQ7HHrfKAIiDUABq2C6rzEGJxNmeCC40j2+32lkqXGAAAtLGgeETSeTXrIZDIlitGQdmZzuTwAb4XiDMDxyKjbvcwBKAL6tZmRQlcnmyoEKhktfZajVAA

packages/structures/src/Structure.ts
Qjuh: Imho the main advantage of having them be symbols is to not accidentally have an actual property on a structure that has the same name. So it makes sense to use symbols for anything we only use internally (here and in mainlib). But as soon as we want end users to also use it (which is the case for DataTemplate and optimizeData afaict) they shouldn't be symbols.

.github/workflows/documentation.yml
```diff
@@ -219,7 +219,7 @@ jobs:
       - name: Move docs to correct directory
         if: ${{ env.REF_TYPE == 'branch' }}
         run: |
-          declare -a PACKAGES=("brokers" "builders" "collection" "core" "discord.js" "formatters" "next" "proxy" "rest" "util" "voice" "ws")
+          declare -a PACKAGES=("brokers" "builders" "collection" "core" "discord.js" "formatters" "next" "proxy" "rest" "structures" "util" "voice" "ws")
```
Jiralite, line 222: Also needs adding on the step "Build docs with main api-extractor" above.

packages/actions/src/uploadCoverage/action.yml
```diff
@@ -88,6 +88,15 @@ runs:
         flags: rest
         token: ${{ inputs.CODECOV_TOKEN }}
 
+    - name: Upload Structures Coverage
+      if: ${{ hashFiles('packages/structures/coverage/cobertura-coverage.xml') != '' }}
+      uses: codecov/codecov-action@v4
+      with:
+        files: ./packages/structures/coverage/cobertura-coverage.xml
+        disable_search: true
+        flags: rest
```
Jiralite:
```suggestion
        flags: structures
```

Jiralite (changes requested):

packages/structures/src/Structure.ts
ckohen: _patch is a specific case because some structures do benefit from having a patch() method that is essentially a direct call to the associated patch route

Qjuh: If that's the intention then naming it `_patch()` while allowing a `patch()` is a horrible approach imho, as that leads to bad maintainability, as those two are easily confusable.

packages/structures/src/Structure.ts
ckohen: worthy of note that it is entirely possible to have references when you are *not* constructing sub-structures, when you are, that data is theoretically not even present in kData anymore (to avoid double-cache)

packages/structures/src/bitfields/BitField.ts
Qjuh: I'm open for any way to do this here. The current implementation is just a 1:1 port of current mainlib BitField and its manual typings.

packages/structures/src/Structure.ts
Qjuh: So we should `structuredClone()` after all?

packages/structures/package.json
```diff
@@ -0,0 +1,97 @@
...
+	},
+	"files": [
+		"dist"
+	],
+	"contributors": [
+		"Crawl <icrawltogo@gmail.com>",
+		"SpaceEEC <spaceeec@yahoo.com>",
+		"Vlad Frangu <me@vladfrangu.dev>",
+		"Aura Román <kyradiscord@gmail.com>",
+		"Chai Kohem <chaikohen@gmail.com>"
```
ckohen: That's embarrassing....
```suggestion
		"Chai Kohen <chaikohen@gmail.com>"
```

packages/structures/src/Structure.ts
```diff
@@ -0,0 +1,144 @@
...
+	/**
+	 * Patches the raw data of this object in place
+	 *
+	 * @param data - the updated data from the API to patch with
+	 * @param extraOptions - any additional options needed to appropriately patch the data
+	 * @remarks To be made public in subclasses
+	 * @returns this
+	 * @internal
+	 */
+	protected patch(data: Readonly<Partial<DataType>>): this {
```
ckohen: Yeah I'm not a huge fan of this not having the _ prefix, it's public everywhere else because it has to be called outside of the structures by the caching library but really isn't meant to be called by the end consumer. It's workable but I'd suggest one of the following
1. Leave it protected everywhere else and force `<Structure>['patch']`
2. Revert to _patch

packages/structures/src/Structure.ts
```diff
@@ -0,0 +1,144 @@
...
+ * kData stores properties as `unknown` when it is omitted, which allows accessing the property in getters even when it may not actually be present.
+ * This is the most technically correct way of representing the value, especially since there is no way to guarantee runtime matches the "type cast."
+ */
+export abstract class Structure<DataType, Omitted extends keyof DataType | '' = ''> {
+	/**
+	 * A construct function used when mixing to allow mixins to set optimized property defaults
+	 *
+	 * @internal
+	 * @remarks This should only be used to set defaults, setting optimized values should be done
+	 * in the mixins `_optimizeData` method, which will be called automatically.
```
ckohen:
```suggestion
	 * in the mixins `optimizeData` method, which will be called automatically.
```

packages/structures/__tests__/invite.test.ts
```diff
@@ -0,0 +1,89 @@
...
+		created_at: '2020-10-10T13:50:17.209Z',
+		max_age: 12,
+		max_uses: 34,
+		temporary: false,
+		uses: 5,
+	};
+
+	test('Invite has all properties', () => {
+		const instance = new Invite(data);
+		// expect(instance.type).toBe(data.type);
```
ckohen: Why is this commented? I noticed type is not included in the structure, but I couldn't tell you why (except maybe it didn't exist when I wrote it?)

packages/structures/src/Mixin.ts
```diff
@@ -0,0 +1,179 @@
...
+					optimization.call(this, data);
+				}
+			},
+		});
+	}
+
+	if (enrichToJSONs.length > 0) {
+		Object.defineProperty(destination.prototype, kMixinToJSON, {
+			writable: true,
+			enumerable: true,
```
ckohen: Are you sure you want this enumerable?

packages/structures/src/invites/Invite.ts
```diff
@@ -0,0 +1,207 @@
...
+
+// TODO: use an actual dtypes type instead if it exists
+export interface APIActualInvite extends APIInvite, Partial<Omit<APIExtendedInvite, keyof APIInvite>> {}
+
+/**
+ * Represents an invitation to a discord channel
+ *
+ * @typeParam Omitted - Specify the properties that will not be stored in the raw data field as a union, implement via `DataTemplate`
+ */
+export class Invite<Omitted extends keyof APIActualInvite | '' = ''> extends Structure<APIActualInvite, Omitted> {
```
ckohen: The default is to omit these from raw data and optimize them instead
```suggestion
export class Invite<Omitted extends keyof APIActualInvite | '' = 'created_at' | 'expires_at'> extends Structure<APIActualInvite, Omitted> {
```

packages/structures/src/utils/symbols.ts
```diff
@@ -0,0 +1,12 @@
+export const kData = Symbol.for('djs.structures.data');
+export const kExpiresTimestamp = Symbol.for('djs.structures.expiresTimestamp');
+export const kCreatedTimestamp = Symbol.for('djs.structures.createdTimestamp');
+export const kArchiveTimestamp = Symbol.for('djs.structures.archiveTimestamp');
+export const kThreadMetadata = Symbol.for('djs.structures.threadMetadata');
```
ckohen: Unused
```suggestion
```

packages/structures/src/utils/symbols.ts
```diff
@@ -0,0 +1,12 @@
...
+export const kAllow = Symbol.for('djs.structures.allow');
+export const kDeny = Symbol.for('djs.structures.deny');
+export const kPermissionOverwrite = Symbol.for('djs.structures.permissionOverwrite');
```
ckohen:
```suggestion
```

packages/structures/src/users/User.ts
```diff
@@ -0,0 +1,178 @@
...
+
+	/**
+	 * The username of the user
+	 */
+	public get username() {
+		return this[kData].username;
+	}
+
+	/**
+	 * The user's 4 digit tag, if a bot or not migrated to unique usernames
```
ckohen: I think it's just bots now
```suggestion
	 * The user's 4 digit tag, if a bot
```

packages/structures/src/users/User.ts
```diff
@@ -0,0 +1,178 @@
+import { DiscordSnowflake } from '@sapphire/snowflake';
+import type { APIUser } from 'discord-api-types/v10';
+import { Structure } from '../Structure.js';
+import { kData } from '../utils/symbols.js';
+
+/**
+ * Represents any user on Discord.
+ *
+ * @typeParam Omitted - Specify the properties that will not be stored in the raw data field as a union, implement via `DataTemplate`
+ * @remarks has a substructure `AvatarDecorationData`, which needs to be instantiated and stored by an extending class using it
```
ckohen, line 12: I appreciate notating like this, and we should definitely do it.
I had wondered if maybe there should be a little bit more of a write up in this package's readme about a couple of the intracacies like this (and to help indicate it's not meant for direct use in most cases?)

packages/structures/src/users/AvatarDecorationData.ts
```diff
@@ -0,0 +1,27 @@
...
+	 * The id of the SKU this avatar decoration is part of.
+	 */
+	public get skuId() {
+		return this[kData].sku_id;
+	}
+
+	/**
+	 * The asset of this avatar decoration.
+	 */
+	public get name() {
```
ckohen: Why name instead of asset?

packages/structures/src/bitfields/index.ts
```diff
@@ -0,0 +1,2 @@
+export * from './BitField.js';
+export * from './ChannelFlagsBitField.js';
```
ckohen, line 3:
```suggestion
export * from './ChannelFlagsBitField.js';
export * from './PermissionsBitField.js';
```

packages/structures/src/channels/Channel.ts
```diff
@@ -0,0 +1,201 @@
...
+/**
+ * Represents any channel on Discord.
+ *
+ * @typeParam Type - Specify the type of the channel being constructed for more accurate data types
+ * @typeParam Omitted - Specify the properties that will not be stored in the raw data field as a union, implement via `DataTemplate`
+ * @remarks Although this class _can_ be instantiated directly for any channel type,
+ * it's intended to be subclassed with the appropriate mixins for each channel type.
+ */
+export class Channel<
+	Type extends ChannelType = ChannelType,
```
ckohen: This needs the `| 'unknown'` to allow future channel types to not break and be useable (which makes the TODO comment above irrelevant cause you'll need the conditional anyways)

I think it should be the default but there's probably some good reasons for it to not be

packages/structures/src/channels/AnnouncementChannel.ts
```diff
@@ -0,0 +1,42 @@
...
+			TextChannelMixin<ChannelType.GuildAnnouncement>,
+			ChannelParentMixin<ChannelType.GuildAnnouncement>,
+			ChannelPermissionMixin<ChannelType.GuildAnnouncement>,
+			ChannelPinMixin<ChannelType.GuildAnnouncement>,
+			ChannelSlowmodeMixin<ChannelType.GuildAnnouncement>,
+			ChannelTopicMixin<ChannelType.GuildAnnouncement>,
+		]
+	> {}
+
+export class AnnouncementChannel<Omitted extends keyof APINewsChannel | '' = ''> extends Channel<
```
ckohen, line 29: Should probably notate that these are only meant to be used by direct end consumers and are really samples for implementation

packages/structures/src/channels/GroupDMChannel.ts
```diff
@@ -0,0 +1,47 @@
...
+> {
+	public constructor(data: Omit<APIGroupDMChannel, Omitted>) {
+		super(data);
+		this.optimizeData(data);
+	}
+
+	/**
+	 * The icon hash of the group DM.
+	 */
+	public get icon() {
```
ckohen: Since these classes aren't being used in implementing libraries maybe we should pull these out into a "mixin" even though it's for a single channel type?

packages/structures/src/channels/PublicThreadChannel.ts
```diff
@@ -0,0 +1,50 @@
...
+> {
+	public constructor(data: Omit<APIPublicThreadChannel, Omitted>) {
+		super(data);
+		this.optimizeData(data);
+	}
+
+	/**
+	 * The IDs of the set of tags that have been applied to a thread in a {@link ForumChannel} or a {@link MediaChannel}.
+	 */
+	public get appliedTags() {
```
ckohen: This may technically be under caching but also ditto to having this not be in something we don't get to use in mainlib

packages/structures/src/channels/ThreadMetadata.ts
```diff
@@ -0,0 +1,122 @@
...
+	Omitted
+> {
+	protected [kArchiveTimestamp]: number | null;
+
+	protected [kCreatedTimestamp]: number | null;
+
+	public constructor(data: Omit<APIThreadMetadata, Omitted>) {
+		super(data);
+		this.optimizeData(data);
+		this[kArchiveTimestamp] ??= null;
```
ckohen: You can set these normally too

packages/structures/src/channels/PermissionOverwrite.ts
```diff
@@ -0,0 +1,95 @@
...
+	Omitted
+> {
+	protected [kAllow]: bigint | null;
+
+	protected [kDeny]: bigint | null;
+
+	public constructor(data: Omit<APIOverwrite, Omitted>) {
+		super(data);
+		this.optimizeData(data);
+		this[kAllow] ??= null;
```
ckohen: ditto

packages/structures/src/Structure.ts
ckohen: I'd say it makes sense to but only in cases where we need to? The problem is detecting that, it's still faster than just blindly strucutredCloning to check through values, but I don't love it

packages/structures/src/bitfields/BitField.ts
ckohen: I don't see why we shouldn't use sapphire if it has what we want, haven't looked into it but this is part of a rather large changeset so as long as it's close we can notate the minor differences in the upgrade guide

(I wouldn't block merging this on that, we can flip to sapphire later if we find it desireable)

packages/structures/src/channels/Channel.ts
Qjuh: But what would be the correct ChannelDataType for that 'unknown' then? Because dtypes won't allow an unknown type in the generic of `APIChannelBase<T>` which would be the best fit imho

ckohen: APIPartialChannel I guess, the only thing different is the type guarantee and flags....and given that you have to skirt type safety on this anyways it's probably fine?

packages/structures/src/Structure.ts
```diff
@@ -0,0 +1,148 @@
...
+		return this;
+	}
+
+	/**
+	 * Creates a clone of this structure
+	 *
+	 * @returns a clone of this
+	 * @internal
+	 */
+	protected clone(patchPayload: Readonly<Partial<DataType>>): typeof this {
```
ckohen:
```suggestion
	protected clone(patchPayload?: Readonly<Partial<DataType>> = {}): typeof this {
```

packages/structures/src/channels/ThreadMetadata.ts
```diff
@@ -0,0 +1,121 @@
...
+
+/**
+ * Represents metadata of a thread channel on Discord.
+ *
+ * @typeParam Omitted - Specify the properties that will not be stored in the raw data field as a union, implement via `DataTemplate`
+ */
+export class ThreadMetadata<Omitted extends keyof APIThreadMetadata | '' = ''> extends Structure<
+	APIThreadMetadata,
+	Omitted | 'archive_timestamp' | 'create_timestamp'
+> {
```
ckohen:
```suggestion
export class ThreadMetadata<Omitted extends keyof APIThreadMetadata | '' = 'archive_timestamp' | 'create_timestamp'> extends Structure<
	APIThreadMetadata,
	Omitted
> {
```

packages/structures/src/channels/mixins/index.ts
```diff
@@ -0,0 +1,13 @@
+export * from './ChannelOwnerMixin.js';
```
ckohen, line 2:
```suggestion
export * from './AppliedTagsMixin.js';
export * from './ChannelOwnerMixin.js';
```

packages/structures/src/channels/mixins/index.ts
```diff
@@ -0,0 +1,13 @@
...
+export * from './ChannelParentMixin.js';
+export * from './ChannelPermissionMixin.js';
+export * from './ChannelPinMixin.js';
+export * from './ChannelSlowmodeMixin.js';
+export * from './ChannelTopicMixin.js';
+export * from './ChannelWebhookMixin.js';
+export * from './DMChannelMixin.js';
```
ckohen, line 9:
```suggestion
export * from './DMChannelMixin.js';
export * from './GroupDMMixin.js';
```

packages/structures/src/channels/mixins/ThreadChannelMixin.ts
```diff
@@ -0,0 +1,45 @@
+import type { ThreadChannelType } from 'discord-api-types/v10';
+import { kData } from '../../utils/symbols.js';
+import type { Channel, ChannelDataType } from '../Channel.js';
+
+export interface ThreadChannelMixin<
+	Type extends ThreadChannelType = ThreadChannelType,
+	Omitted extends keyof ChannelDataType<Type> | '' = '',
```
ckohen: This is the only mixin that has omitted, I can't remember if it actually is necessary to have it in the mixins or not.

packages/structures/src/Structure.ts
didinele: so what's the consensus here; I see it was resolved but we kept it as `_patch`

Qjuh: It was kept as _patch because of https://github.com/discordjs/discord.js/pull/10900#discussion_r2173485976 as it is public in other classes so it needs another indicator of not being intended for end user usage.

ckohen (approved):
Looks pretty good as a starting point. There's probably a few things we'll discover as we develop the rest of the module but it will be easier to deal with in smaller PRs from this point.

packages/structures/src/channels/mixins/VoiceChannelMixin.ts
```diff
@@ -0,0 +1,48 @@
...
+
+	/**
+	 * The voice region id for this channel, automatic when set to null.
+	 */
+	public get rtcRegion() {
+		return this[kData].rtc_region!;
+	}
+
+	/**
+	 * The camera video quality mode of the voice channel, 1 when not present.
```
Jiralite: Instead of "1", can we reference `VideoQualityMode.Auto`?

packages/structures/cliff.toml
```diff
@@ -0,0 +1,80 @@
...
+conventional_commits = true
+filter_unconventional = true
+commit_parsers = [
+	{ message = "^feat", group = "Features"},
+	{ message = "^fix", group = "Bug Fixes"},
+	{ message = "^docs", group = "Documentation"},
+	{ message = "^perf", group = "Performance"},
+	{ message = "^refactor", group = "Refactor"},
+	{ message = "^typings", group = "Typings"},
+	{ message = "^types", group = "Typings"},
```
Jiralite: #10738

```suggestion
	{ message = "^types", group = "Typings"},
```

packages/structures/tsconfig.json
```diff
@@ -0,0 +1,9 @@
+{
+	"$schema": "https://json.schemastore.org/tsconfig.json",
+	"extends": "../../tsconfig.json",
+	"include": ["src/**/*.ts", "src/**/*.js", "src/**/*.cjs", "src/**/*.mjs", "bin"],
+	"exclude": ["node_modules"],
+	"compilerOptions": {
+		"experimentalDecorators": false
```
Jiralite, line 7: Why is this here? Surely it's not `true` by default?

Jiralite (changes requested):

packages/structures/tsconfig.json
Qjuh: No, but it is true in our base tsconfig.json which this extends.

packages/structures/src/channels/ForumTag.ts
```diff
@@ -0,0 +1,57 @@
...
+
+	/**
+	 * The name of the tag.
+	 */
+	public get name() {
+		return this[kData].name;
+	}
+
+	/**
+	 * Whether this tag can only be added to or removed from threads by a member with the {@link discord-api-types/v10#PermissionFlagsBits.ManageThreads} permission.
```
Jiralite: How come this is `{@link discord-api-types/v10#PermissionFlagsBits.ManageThreads}` but line 83 in packages/structures/src/channels/ThreadMetadata.ts has `{@link PermissionFlagsBits.ManageThreads}`? I assume one of them is wrong.

Jiralite (approved):

packages/structures/src/Structure.ts
```diff
@@ -0,0 +1,139 @@
...
+
+	/**
+	 * Patches the raw data of this object in place
+	 *
+	 * @param data - the updated data from the API to patch with
+	 * @remarks To be made public in subclasses
+	 * @returns this
+	 * @internal
+	 */
+	protected _patch(data: Readonly<Partial<DataType>>): this {
```
vladfrangu: So I'll open the wound again. What do we think about making all of these methods (excluding toJSON) symbols? I can esp think `clone()` could at some point be used to represent something else other than the re-create this class?

packages/structures/src/bitfields/BitField.ts
```diff
@@ -0,0 +1,203 @@
...
+	public static readonly Flags: EnumLike<unknown, bigint | number> = {};
+
+	public static readonly DefaultBit: bigint = 0n;
+
+	/**
+	 * Bitfield of the packed bits
+	 */
+	public bitfield: bigint;
+
+	declare public ['constructor']: NonAbstract<typeof BitField<Flags>>;
```
vladfrangu, line 38: Do you need the NonAbstract type at all?

packages/structures/src/bitfields/BitField.ts
```diff
@@ -0,0 +1,203 @@
...
+	 */
+	public toArray(...hasParams: readonly unknown[]) {
+		return [...this[Symbol.iterator](...hasParams)];
+	}
+
+	public toJSON(asNumber?: boolean) {
+		if (asNumber) {
+			if (this.bitfield > Number.MAX_SAFE_INTEGER) {
+				throw new RangeError(
+					`Cannot convert bitfield value ${this.bitfield} to number, as it is bigger than Number.MAX_SAFE_INTEGER`,
```
vladfrangu: Might as well

```suggestion
					`Cannot convert bitfield value ${this.bitfield} to number, as it is bigger than ${Number.MAX_SAFE_INTEGER}`,
```

packages/structures/src/bitfields/PermissionsBitField.ts
```diff
@@ -0,0 +1,78 @@
...
+	 */
+	public static readonly Default = 104_324_673n;
+
+	/**
+	 * Bitfield representing the permissions required for moderators of stage channels
+	 */
+	public static readonly StageModerator =
+		PermissionFlagsBits.ManageChannels | PermissionFlagsBits.MuteMembers | PermissionFlagsBits.MoveMembers;
+
+	public static override readonly DefaultBit = 0n;
```
vladfrangu: Yeet it, yeet it

packages/structures/src/channels/Channel.ts
```diff
@@ -0,0 +1,203 @@
...
+	Type extends ChannelType | 'unknown' = ChannelType,
+	Omitted extends keyof ChannelDataType<Type> | '' = '',
+> extends Structure<ChannelDataType<Type>, Omitted> {
+	/**
+	 * The template used for removing data from the raw data stored for each Channel.
+	 *
+	 * @remarks This template is only guaranteed to apply to channels constructed directly via `new Channel()`.
+	 * Use the appropriate subclass template to remove data from that channel type.
+	 */
+	public static override DataTemplate: Partial<APIChannel> = {};
```
vladfrangu: Ngl, DataTemplate should be a symbol

packages/structures/src/channels/Channel.ts
```diff
@@ -0,0 +1,203 @@
...
+		const createdTimestamp = this.createdTimestamp;
+		return createdTimestamp ? new Date(createdTimestamp) : null;
+	}
+
+	/**
+	 * Indicates whether this channel is a thread channel
+	 *
+	 * @privateRemarks Overriden to `true` on `ThreadChannelMixin`
+	 */
+	public isThread(): this is ThreadChannelMixin<Extract<Type, ThreadChannelType>> {
```
vladfrangu: Shouldn't these assertions be `this is this & MixInClass`? Cause otherwise won't TS narrow the type to ONLY those properties?

packages/structures/src/channels/PermissionOverwrite.ts
```diff
@@ -0,0 +1,94 @@
...
+			this[kDeny] = BigInt(data.deny);
+		}
+	}
+
+	/**
+	 * The permission bit set allowed by this overwrite.
+	 */
+	public get allow() {
+		const allow = this[kAllow];
+		return allow ? new PermissionsBitField(allow) : null;
```
vladfrangu: If allow is 0n, this will wrongfully return null

packages/structures/src/channels/mixins/AppliedTagsMixin.ts
```diff
@@ -0,0 +1,14 @@
...
+import type { Channel } from '../Channel.js';
+
+export interface AppliedTagsMixin extends Channel<ChannelType.PublicThread> {}
+
+export class AppliedTagsMixin {
+	/**
+	 * The IDs of the set of tags that have been applied to a thread in a {@link (ForumChannel:class)} or a {@link (MediaChannel:class)}.
+	 */
+	public get appliedTags() {
+		return Array.isArray(this[kData].applied_tags) ? [...this[kData].applied_tags] : null;
```
vladfrangu: Cloning every time the getter is called sounds like a side effect that should
- be removed
- be documented

packages/structures/src/invites/Invite.ts
```diff
@@ -0,0 +1,221 @@
+import { type APIInvite, type APIExtendedInvite, RouteBases } from 'discord-api-types/v10';
+import { Structure } from '../Structure.js';
+import { kData, kExpiresTimestamp, kCreatedTimestamp } from '../utils/symbols.js';
+import type { Partialize } from '../utils/types.js';
+
+// TODO: use an actual dtypes type instead if it exists
```
vladfrangu: What type is missing?

packages/structures/src/invites/Invite.ts
```diff
@@ -0,0 +1,221 @@
...
+export interface APIActualInvite extends APIInvite, Partial<Omit<APIExtendedInvite, keyof APIInvite>> {}
+
+/**
+ * Represents an invitation to a discord channel
```
vladfrangu: `s/discord/Discord` in comments

packages/structures/src/invites/Invite.ts
```diff
@@ -0,0 +1,221 @@
...
+	public get type() {
+		return this[kData].type;
+	}
+
+	/**
+	 * The approximate number of online members of the guild this invite is for
+	 *
+	 * @remarks Only available when the invite was fetched from `GET /invites/<code>` with counts
+	 */
+	public get presenceCount() {
```
vladfrangu:
```suggestion
	public get approximatePresenceCount() {
```

packages/structures/src/invites/Invite.ts
```diff
@@ -0,0 +1,221 @@
...
+		return this[kData].approximate_presence_count;
+	}
+
+	/**
+	 * The approximate total number of members of the guild this invite is for
+	 *
+	 * @remarks Only available when the invite was fetched from `GET /invites/<code>` with counts
+	 */
+	public get memberCount() {
```
vladfrangu:
```suggestion
	public get approximateMemberCount() {
```

vladfrangu (changes requested):

packages/structures/src/channels/Channel.ts
almeidx: What's the motivation for that? This is meant to be used by the end-user to allow them to remove unwanted properties from the structure

At first glance making it a symbol only seems to make it more obscure and potentially harder to use

packages/structures/src/bitfields/BitField.ts
almeidx: For an uninformed user that will just look like a "random" number though...
While as it is currently, it shows them where the value comes from.

Jiralite: How about both? `[...] bigger than ${Number.MAX_SAFE_INTEGER} (the maximum safe integer)`?

packages/structures/src/channels/Channel.ts
vladfrangu: The point is this isn't for users to just _define_ unless they know what they're doing... I mean this ties into https://github.com/discordjs/discord.js/pull/10900#discussion_r2181068448

packages/structures/src/channels/mixins/AppliedTagsMixin.ts
```diff
@@ -0,0 +1,14 @@
+import type { ChannelType } from 'discord-api-types/v10';
+import { kData } from '../../utils/symbols.js';
+import type { Channel } from '../Channel.js';
+
+export interface AppliedTagsMixin extends Channel<ChannelType.PublicThread> {}
+
+export class AppliedTagsMixin {
+	/**
+	 * The IDs of the set of tags that have been applied to a thread in a {@link (ForumChannel:class)} or a {@link (MediaChannel:class)}.
```
Jiralite:
```suggestion
	 * The ids of the set of tags that have been applied to a thread in a {@link (ForumChannel:class)} or a {@link (MediaChannel:class)}.
```

packages/structures/src/bitfields/BitField.ts
Qjuh: Yes, else `new this.constructor(...)` errors because can't instantiate abstract classes.

packages/structures/src/Structure.ts
Qjuh: the `_patch()` maybe could/should be a symbol now with all the reasoning for protected/public and _ prefix shenanigans. For `clone()` I'm indifferent. To me `Class#clone()` should always re-create the class, anything else would be counter-intuitive for anyone using it (and in other OOP languages even is enforced to do so) so I don't quite get what you're thinkinh of there.

packages/structures/src/invites/Invite.ts
Qjuh: The one defined here. `APIInvite` but with all additional properties of `APIExtendedInvite` as optional.

packages/structures/src/channels/mixins/AppliedTagsMixin.ts
Qjuh: The intention here was that Structures should not be mutable. Unless I'm confused returning the array directly would mean the user could mutate it. I'm fine with changing it either way you deem appropriate though.

packages/structures/src/channels/Channel.ts
Qjuh: Tried applying that change, but this causes an OOM error on running `tsc`, probably because the `this` reference in the typeguard makes it quirky in the mixins. This might need some more thought, happy if someone has an idea how to do it correctly without v8 (apparently) running into an endless recursion loop. Or if this even is a typescript bug and shouldn't happen in the first place

Qjuh: Good news: it's not an infinite recursion
bad news: with more than 5 of these typeguards v8 reaches its limits on my M2 with 8GB RAM

packages/structures/src/channels/mixins/AppliedTagsMixin.ts
didinele: I agree, though this is a case where there's fair runtime penalty for enforcing immutability.

I'd say it'd be good if we enforced it type-level anyways, and probably settle for just that (i.e. return readonly T[])

packages/structures/src/channels/Channel.ts
didinele: mmm, not good
the typeguard in its current form is also not really acceptable, since I'd imagine it leads into weird cases like
```ts
if (!channel.isThread()) return;

// can't use the normal `Channel` props now?
```

Qjuh: better news: I think I was able to solve it by simplifying the typeguards.

packages/structures/src/channels/Channel.ts
Qjuh: It's meant to be `readonly` (and was in many places, but forgotten in others). Does that satisfy your worry about users defining it?

packages/structures/src/Structure.ts
vladfrangu: I don't know any OOP language that enforces clone to return a new instance of the class, and I can absolutely imagine somebody implementing `clone()` as `send a discord request that recreates this structure with a new id`, like cloning a text channel for example. It's why I'd rather it also be a symbol like patch, but if others don't agree then c'est la vie, let it be

packages/structures/src/channels/Channel.ts
vladfrangu: Not really, imo still should be a symbol but 🤷

packages/structures/src/invites/Invite.ts
vladfrangu: Does that exist at all, returned from Discord? 👀

packages/structures/src/Structure.ts
almeidx: > I can absolutely imagine somebody implementing `clone()` as `send a discord request that recreates this structure with a new id`, like cloning a text channel for example.

You don't have to imagine it as that's precisely what discord.js does. [TextChannel#clone()](<https://discord.js.org/docs/packages/discord.js/14.21.0/TextChannel:Class#clone>)

I think we should rename `clone()` back to `_clone()`, as it also is done in discord.js currently.

Qjuh: Java: https://docs.oracle.com/javase/7/docs/api/java/lang/Object.html#clone()
C#: https://learn.microsoft.com/en-us/dotnet/api/system.icloneable?view=netcore-3.1
C++: uses copy constructors instead, so fair, they don't
Ruby: https://docs.ruby-lang.org/en/3.4/Kernel.html#method-i-clone
Python: they use __deepcopy__ and __copy__ instead of clone indeed
PHP: they use the __clone() method https://www.php.net/manual/en/language.oop5.cloning.php
I could go on but I guess you see what i mean

vladfrangu: Since we need to rename it anyways, lets also symbol it, consistency with patch

didinele: I agree to symbol it in this case

packages/structures/src/channels/Channel.ts
Qjuh: I'm with almeidx here, making it a symbol defeats the whole purpose of it being usable by the end user.

vladfrangu: How? They just import the symbol from structures and define it on their class, similar to patch or clone? Am I missing something?

almeidx: Discoverability for one. It's also much worse DX.

There's basically 2 levels of end-user. The actual end-users, i.e. users of discord.js, and other libs extending structures, like discord.js itself will

Patch and clone I think would realistically only be used by the latter, which *maybe* justifies "hiding" them behind symbols. DataTemplate is meant to be used by the actual end users, and in my eyes is one of the selling points of /structures

packages/structures/src/Structure.ts
```diff
@@ -0,0 +1,139 @@
...
+	/**
+	 * Creates a clone of this structure
+	 *
+	 * @returns a clone of this
+	 * @internal
+	 */
+	protected [kClone](patchPayload?: Readonly<Partial<DataType>>): typeof this {
+		const clone = this.toJSON();
+		// @ts-expect-error constructor is of abstract class is unknown
+		return new this.constructor(patchPayload ? Object.assign(clone, patchPayload) : clone);
```
almeidx: This `@ts-expect-error` comment is actually suppressing 2 errors, one of which seems unexpected as it isn't related to the reasoning in the comment.

Moving the arguments into a new line reveals the second error, on the `Object.assign()` call, specifically the `clone`:

```ts
No overload matches this call.
  Overload 1 of 4, '(target: {}, source: Readonly<Partial<DataType>>): Readonly<Partial<DataType>>', gave the following error.
    Argument of type 'DataType' is not assignable to parameter of type '{}'.
  Overload 2 of 4, '(target: object, ...sources: any[]): any', gave the following error.
    Argument of type 'DataType' is not assignable to parameter of type 'object'.ts(2769)
Structure.ts(22, 33): This type parameter might need an `extends {}` constraint.
Structure.ts(22, 33): This type parameter might need an `extends object` constraint.
```

Doesn't seem like such a bad idea to add this constraint to `DataType` (It requires more changes after that however)

packages/structures/src/Structure.ts
```diff
@@ -0,0 +1,139 @@
...
+	 */
+	protected [kMixinToJSON]?(data: Partial<DataType>): void;
+
+	/**
+	 * The template used for removing data from the raw data stored for each Structure.
+	 *
+	 * @remarks This template should be overridden in all subclasses to provide more accurate type information.
+	 * The template in the base {@link Structure} class will have no effect on most subclasses for this reason.
+	 */
+	protected static readonly DataTemplate: Record<string, unknown> = {};
```
vladfrangu, line 49: Symbol's not wanted but we should use `DataTemplatePropertyName` / `OptimizeDataPropertyName` when defining this field, lest we end up renaming it at some point and miss this 🙃

packages/structures/src/Structure.ts
```diff
@@ -0,0 +1,139 @@
...
+	 * For typescript users, mixins must use the closest ancestors access modifier.
+	 * @remarks Automatically called in Structure[kPatch] but must be called manually in the constructor
+	 * of any class implementing this method.
+	 * @remarks Additionally, when implementing, ensure to call `super._optimizeData` if any class in the super chain aside
+	 * from Structure contains an implementation.
+	 * Note: mixins do not need to call super ever as the process of mixing walks the prototype chain.
+	 * @virtual
+	 * @internal
+	 */
+	protected optimizeData(_data: Partial<DataType>) {}
```
vladfrangu, line 121: Ditto

packages/structures/src/Structure.ts
```diff
@@ -0,0 +1,139 @@
...
+	 * The type of this data is determined by omissions at runtime and is only guaranteed for default omissions
+	 * @privateRemarks
+	 * When omitting properties at the library level, this must be overridden to re-add those properties
+	 */
+	public toJSON(): DataType {
+		// This will be DataType provided nothing is omitted, when omits occur, subclass needs to overwrite this.
+		const data = (
+			Object.values(this[kData]).some((value) => typeof value === 'object' && value !== null)
+				? structuredClone(this[kData])
+				: { ...this[kData] }
```
vladfrangu: What exactly is the point of all this logic? Why would we use structuredClone just if we have nested objects? Not a blocker for the PR but I wanna know XD

packages/structures/src/channels/Channel.ts
```diff
@@ -0,0 +1,186 @@
...
+		const flags =
+			'flags' in this[kData] && typeof this[kData].flags === 'number' ? (this[kData].flags as ChannelFlags) : null;
+		return flags ? new ChannelFlagsBitField(flags) : null;
+	}
+
+	/**
+	 * The timestamp the channel was created at
+	 */
+	public get createdTimestamp() {
+		return ['string', 'bigint'].includes(typeof this.id)
```
vladfrangu: Use the id function you wrote for another place

vladfrangu (approved):

packages/structures/src/Structure.ts
Qjuh: Based on the benchmark and internal discussion here https://discord.com/channels/222078108977594368/1145525943411155045/1388626168391139388

almeidx: Wouldn't hurt to add a comment explaining that

packages/structures/src/Structure.ts
almeidx: To prevent errors like this, we could do something like:

```suggestion
		// @ts-expect-error constructor is of abstract class is unknown
		return new this.constructor( // Ensure the ts-expect-error only applies to the constructor call
			patchPayload ? Object.assign(clone, patchPayload) : clone
		);
```

packages/structures/src/Structure.ts
Qjuh: I understand the idea, but that would mean that users implementing this would either
- also do `protected static override readonly [DataTemplatePropertyName]: ...` making it same/worse than symbols
- use `protected static override readonly DataTemplate: ...` and not match how it's written in the base class, which is also weird (and face the same issue if renamed)

The purpose of the variables for `DataTemplatePropertyName` is merely to be used in the `Mixin` function, which is in another file. Not to confuse the users

vladfrangu: 1 is not applicable and 2 is fine because the override flag would throw an error if we ever rename the field... I don't understand whats wrong with that?

Qjuh: I just find it weird to have users extend and override a field but expect them to write it differently than how it is written in the base class property they extend/override.

Qjuh: Maybe the added tests suffice to ensure what you're worried about instead?

packages/structures/README.md
```diff
@@ -0,0 +1,69 @@
...
+```
+
+## Links
+
+- [Website][website] ([source][website-source])
+- [Documentation][documentation]
+- [Guide][guide] ([source][guide-source])
+  Also see the v13 to v14 [Update Guide][guide-update], which includes updated and removed items from the library.
+- [discord.js Discord server][discord]
+- [Discord API Discord server][discord-api]
```
Jiralite:
```suggestion
- [Discord Developers Discord server][discord-developers]
```

packages/structures/README.md
```diff
@@ -0,0 +1,69 @@
...
+If you don't understand something in the documentation, you are experiencing problems, or you just need a gentle nudge in the right direction, please don't hesitate to join our official [discord.js Server][discord].
+
+[website]: https://discord.js.org
+[website-source]: https://github.com/discordjs/discord.js/tree/main/apps/website
+[documentation]: https://discord.js.org/docs/packages/structures/stable
+[guide]: https://discordjs.guide/
+[guide-source]: https://github.com/discordjs/guide
+[guide-update]: https://discordjs.guide/additional-info/changes-in-v14.html
+[discord]: https://discord.gg/djs
+[discord-api]: https://discord.gg/discord-api
```
Jiralite:
```suggestion
[discord-developers]: https://discord.gg/discord-developers
```

iCrawl (approved):

almeidx (approved):

PR #10912, opened 2025-05-25
fix: remove incorrect nullables from ApplicationEmoji and GuildEmoji, add ApplicationEmoji#managed

### LINKED ISSUE #10858: Incorrect typing for `ApplicationEmoji`

discord.js

### Issue description

The property `ApplicationEmoji#name` is never null since that is only possible for reaction emojis
The property `ApplicationEmoji#animated` is never null since that is only possible for unicode reaction emojis

The property `ApplicationEmoji#author` cannot be null ddocs state that:
> Includes a user object for the team member that uploaded the emoji from the app's settings, or for the bot user if uploaded using the API.

I'm not really sure whether this information is correct but from my point of view, these properties, and maybe some more could never be null. If I'm wrong please correct me.

### Code sample

```typescript

```

### Versions

- discord.js@14.19.1
- typescript@5.8.3

### Issue priority

Low (slightly annoying)

Not applicable

Not applicable

_No response_

Amgelo563:

ApplicationEmoji and BaseGuildEmoji inherit some incorrect nullables from Emoji (since they are never null for their case). Also BaseGuildEmoji#url wasn't marked as deprecated in the index.d.ts and ApplicationEmoji#available was missing (though always true, added for consistency with the rest). Internal discussion at https://discord.com/channels/222078108977594368/1366526457064591440/1366526457064591440

This also addresses the pending changes at https://github.com/discordjs/discord.js/pull/10905

Fixes #10858 

### Typings
- Remove `| null` from `ApplicationEmoji#author`, `managed`, and `requiresColons`
- Change `ApplicationEmoji#managed`'s type to `false` and `requiresColons` to `true`
- Add `ApplicationEmoji#available: true`
- Type override `ApplicationEmoji#name`, `animated`, `imageURL`, `url`, `createdAt` and `createdTimestamp` to make them non nullable
- Add `@deprecated` tag for `BaseGuildEmoji#url`
- Type override `BaseGuildEmoji#name`, `animated`, and `managed` to make them non nullable (the other properties were already non nullable)
### JS
- Add jsdoc comments overriding `ApplicationEmoji#name`, `animated`, `imageURL`, `url`, `createdAt` and `createdTimestamp`'s to remove the "or null if unicode" mentions
- Add jsdoc comments overriding `BaseGuildEmoji#name`, `animated`, `createdAt` and `createdTimestamp`'s to remove the "or null if unicode" mentions (`url` and `imageURL` already have them)
- Add `available` on `ApplicationEmoji`:
  - Set to `null` in constructor
  - Set in `_patch()`
  - Add check to `equals()`

- - I know how to update typings and have done so, or typings don't need updating

### DIFF
--- a/packages/discord.js/src/structures/ApplicationEmoji.js
+++ b/packages/discord.js/src/structures/ApplicationEmoji.js
@@ -18,7 +18,7 @@ class ApplicationEmoji extends Emoji {
 
     /**
      * The user who created this emoji
-     * @type {?User}
+     * @type {User}
      */
     this.author = null;
 
@@ -35,7 +35,8 @@ class ApplicationEmoji extends Emoji {
     if ('managed' in data) {
       /**
        * Whether this emoji is managed by an external service
-       * @type {?boolean}
+       * @remarks Always false for application emojis
+       * @type {false}
        */
       this.managed = data.managed;
     }
@@ -43,7 +44,8 @@ class ApplicationEmoji extends Emoji {
     if ('require_colons' in data) {
       /**
        * Whether or not this emoji requires colons surrounding it
-       * @type {?boolean}
+       * @remarks Always true for application emojis
+       * @type {true}
        */
       this.requiresColons = data.require_colons;
     }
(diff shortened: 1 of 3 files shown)


Qjuh (changes requested):

packages/discord.js/src/structures/ApplicationEmoji.js
```diff
@@ -35,18 +36,29 @@ class ApplicationEmoji extends Emoji {
     if ('managed' in data) {
       /**
        * Whether this emoji is managed by an external service
-       * @type {?boolean}
+       * @remarks Always false for application emojis
+       * @type {false}
```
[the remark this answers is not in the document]
Amgelo563: Resolved internally

packages/discord.js/typings/index.d.ts
```diff
@@ -704,13 +704,16 @@ export abstract class BaseGuild extends Base {
 export class BaseGuildEmoji extends Emoji {
   protected constructor(client: Client<true>, data: RawGuildEmojiData, guild: Guild | GuildPreview);
   public imageURL(options?: BaseImageURLOptions): string;
+  /** @deprecated Use {@link BaseGuildEmoji#imageURL} instead */
```
Qjuh, line 707:
```suggestion
  /** @deprecated Use {@link BaseGuildEmoji.imageURL} instead */
```

packages/discord.js/typings/index.d.ts
```diff
@@ -1490,10 +1493,18 @@ export class ApplicationEmoji extends Emoji {
...
-  public managed: boolean | null;
-  public requiresColons: boolean | null;
+  public managed: false;
+  public requiresColons: true;
+  public name: string;
+  public animated: boolean;
+  public available: true;
+  public get createdAt(): Date;
+  public get createdTimestamp(): number;
+  /** @deprecated Use {@link ApplicationEmoji#imageURL} instead */
```
Qjuh, line 1505:
```suggestion
  /** @deprecated Use {@link ApplicationEmoji.imageURL} instead */
```

Qjuh (changes requested):
Just two small nits.

PR #10922, opened 2025-06-06
feat!: use zod v4

almeidx:
Updated builders to use zod v4 ~~which is actually v3.25~~
Replaced `zod-validation-error` with `z.prettifyError()`

<https://zod.dev/v4>


packages/builders/src/Assertions.ts
```diff
@@ -1,18 +1,16 @@
...
-			Locale,
-			z.ZodOptional<z.ZodString>
-		>,
-	)
-	.strict();
+export const localeMapPredicate = z.strictObject(
+	Object.fromEntries(Object.values(Locale).map((loc) => [loc, z.string().optional()])) as Record<
+		Locale,
+		z.ZodOptional<z.ZodString>
+	>,
```
vladfrangu, line 12: Do we really need the cast or can we just cast the returned array `as const` and achieve the same thing?

packages/builders/src/messages/Assertions.ts
```diff
@@ -77,13 +77,9 @@ const messageNoComponentsV2Predicate = baseMessagePredicate
...
-			.refine((flags) => {
-				// If we have flags, ensure we don't have the ComponentsV2 flag
-				if (flags) {
-					return (flags & MessageFlags.IsComponentsV2) === 0;
-				}
-
-				return true;
+			// If we have flags, ensure we don't have the ComponentsV2 flag
+			.refine((flags) => !flags || (flags & MessageFlags.IsComponentsV2) === 0, {
+				error: 'ComponentsV2 flag cannot be set on messages without components v2',
```
vladfrangu: This is absolutely not true, you CAN set the flag and not use components v2 (aka send a message with just `components: [actionRowsHere]`)

packages/builders/src/messages/Assertions.ts
```diff
@@ -77,13 +77,9 @@ const messageNoComponentsV2Predicate = baseMessagePredicate
 		flags: z
 			.number()
```
vladfrangu: flags are ints, idk how we didn't catch this before

vladfrangu (dismissed):

colinhacks:
Those errors are fixed, upgrade to 3.25.56 👍

packages/builders/src/components/Assertions.ts
```diff
@@ -26,31 +25,22 @@ const buttonCustomIdPredicateBase = buttonPredicateBase.extend({
...
-			.string()
-			.url()
-			.refine(refineURLPredicate(['http:', 'https:', 'discord:'])),
-		emoji: emojiPredicate.optional(),
-		label: labelPredicate,
-	})
-	.strict();
+const buttonLinkPredicate = buttonPredicateBase.extend({
+	style: z.literal(ButtonStyle.Link),
+	url: z.url().refine(refineURLPredicate(['http:', 'https:', 'discord:'])),
```
colinhacks: This is a fine solution but there is first-party regex-based hostname & protocol validation if you want it:

https://zod.dev/api?id=urls

packages/builders/src/Assertions.ts
almeidx: Yes, we do need the cast. This is because the return type of `Object.fromEntries()` disregards the type of the keys.
https://github.com/microsoft/TypeScript/blob/51dcd9001d9dceb98d248f574f39f57497a44968/src/lib/es2019.object.d.ts#L8

packages/builders/src/components/Assertions.ts
almeidx: This is great, thanks for bringing it up!

packages/builders/src/messages/Assertions.ts
Qjuh: But then the cv2 predicate would be true in the union below and this error wouldn't propagate. If it does propagate there was more than just components set.

almeidx: Made the error message more clear https://github.com/discordjs/discord.js/pull/10922/commits/a5acb4d0469f6f8b00b51e6a07698972e227a17d

packages/builders/src/components/Assertions.ts
```diff
@@ -26,31 +25,22 @@ const buttonCustomIdPredicateBase = buttonPredicateBase.extend({
...
-			.string()
-			.url()
-			.refine(refineURLPredicate(['http:', 'https:', 'discord:'])),
-		emoji: emojiPredicate.optional(),
-		label: labelPredicate,
-	})
-	.strict();
+const buttonLinkPredicate = buttonPredicateBase.extend({
+	style: z.literal(ButtonStyle.Link),
+	url: z.url({ protocol: /^(?:https?|attachment)$/ }),
```
Qjuh:
```suggestion
	url: z.url({ protocol: /^(?:https?|discord)$/ }),
```

Qjuh (approved):

packages/builders/src/util/ValidationError.ts
```diff
@@ -0,0 +1,21 @@
...
+export class ValidationError extends Error {
+	/**
+	 * The underlying cause of the validation error.
+	 */
+	public override readonly cause: z.ZodError;
+
+	/**
+	 * @internal
+	 */
+	public constructor(error: z.ZodError) {
```
Jiralite, line 15: Should this be `private` then?

didinele: No, considering you wouldn't be able to construct it then:

https://github.com/discordjs/discord.js/blob/1c277eef744d9267ea5a16c756ba94da4d58dfa9/packages/builders/src/util/validation.ts#L54

I guess the class should be @ internal if anything? Though I imagine the intent is to tell users they can `catch` this error and to expect it from our API, but they shouldn't try to construct it/we can change the constructor without semver breaking.

packages/builders/src/interactions/commands/chatInput/Assertions.ts
```diff
@@ -24,24 +26,19 @@ const sharedNameAndDescriptionPredicate = z.object({
...
 	channel_types: z
-		.union(
-			ApplicationCommandOptionAllowedChannelTypes.map((type) => z.literal(type)) as unknown as [
-				ZodTypeAny,
-				ZodTypeAny,
-				...ZodTypeAny[],
-			],
-		)
+		// Zod doesn't allow passing readonly arrays :/
+		.literal(ApplicationCommandOptionAllowedChannelTypes as unknown as ApplicationCommandOptionAllowedChannelType[])
```
vladfrangu: https://github.com/colinhacks/zod/pull/4643

packages/builders/src/util/ValidationError.ts
vladfrangu: You could make the constructor private/protected, but then you'd need to hack around it with `Reflect.construct()` calls. Imo it's fine if the constructor is public, if someone wants to build the class for testing or w/e they should be able to

Jiralite (approved):

vladfrangu (approved):

PR #10923, opened 2025-06-06
types: reverse interface inheritance to omit Omit

Qjuh:

For documentation purposes it is better to have regular inheritance, without Omit<> or Pick<> types in the inheritance chain.
While implementing this change I also fixed some bugs in typings, for example:
- Interaction#awaitModalSubmit(..) had too many allowed options in typings, that wouldn't do anything
- ForumChannel#lastMessageId doesn't exist


vladfrangu (approved):
Sure

didinele (approved):

PR #10961, opened 2025-07-01
feat!: role gradient colors

sdanialraza:

- https://github.com/discord/discord-api-docs/pull/7549
- https://github.com/discord/discord-api-docs/pull/7661

### DIFF
--- a/packages/discord.js/src/managers/RoleManager.js
+++ b/packages/discord.js/src/managers/RoleManager.js
@@ -115,7 +115,8 @@ class RoleManager extends CachedManager {
    * @property {ColorResolvable} [secondaryColor] The secondary color of the role.
    * This will make the role a gradient between the other provided colors
    * @property {ColorResolvable} [tertiaryColor] The tertiary color of the role.
-   * When sending `tertiaryColor` the API enforces the role color to be a holographic style with values of `primaryColor = 11127295`, `secondaryColor = 16759788`, and `tertiaryColor = 16761760`
+   * When sending `tertiaryColor` the API enforces the role color to be a holographic style with values of `primaryColor = 11127295`, `secondaryColor = 16759788`, and `tertiaryColor = 16761760`.
+   * These values are available as a constant: `Constants.HolographicStyle`
    */
 
   /**
@@ -153,6 +154,19 @@ class RoleManager extends CachedManager {
    *   reason: 'we needed a role for Super Cool People',
    *   colors: {
    *     primaryColor: Colors.Blue,
+   *   },
+   * })
+   *   .then(console.log)
+   *   .catch(console.error);
+   * @example
+   * // Create a role with holographic colors
+   * guild.roles.create({
+   *   name: 'Holographic Role',
+   *   reason: 'Creating a role with holographic effect',
+   *   colors: {
+   *     primaryColor: Constants.HolographicStyle.Primary,
+   *     secondaryColor: Constants.HolographicStyle.Secondary,
+   *     tertiaryColor: Constants.HolographicStyle.Tertiary,
    *   },
    * })
    *   .then(console.log)
--- a/packages/discord.js/src/managers/RoleManager.js
+++ b/packages/discord.js/src/managers/RoleManager.js
@@ -123,8 +123,6 @@ class RoleManager extends CachedManager {
    *
    * @typedef {Object} RoleCreateOptions
    * @property {string} [name] The name of the new role
-   * @property {ColorResolvable} [color] The color to create the role with.
-   * This will still be returned by the API, but using the `colors` field is recommended when doing requests
    * @property {RoleColors} [colors] The colors to create the role with
    * @property {boolean} [hoist] Whether or not the new role should be hoisted
    * @property {PermissionResolvable} [permissions] The permissions for the new role
@@ -152,16 +150,17 @@ class RoleManager extends CachedManager {
    * // Create a new role with data and a reason
    * guild.roles.create({
    *   name: 'Super Cool Blue People',
-   *   color: Colors.Blue,
    *   reason: 'we needed a role for Super Cool People',
+   *   colors: {
+   *     primaryColor: Colors.Blue,
+   *   },
    * })
    *   .then(console.log)
    *   .catch(console.error);
    */
   async create(options = {}) {
-    let { color, permissions, icon } = options;
+    let { permissions, icon } = options;
     const { name, hoist, position, mentionable, reason, unicodeEmoji } = options;
-    color &&= resolveColor(color);
     if (permissions !== undefined) permissions = new PermissionsBitField(permissions);
     if (icon) {
       const guildEmojiURL = this.guild.emojis.resolve(icon)?.imageURL();
@@ -178,7 +177,6 @@ class RoleManager extends CachedManager {
     const data = await this.client.rest.post(Routes.guildRoles(this.guild.id), {
       body: {
         name,
-        color,
         colors,
         hoist,
         permissions,
@@ -238,7 +236,6 @@ class RoleManager extends CachedManager {
 
     const body = {
       name: options.name,
-      color: options.color === undefined ? undefined : resolveColor(options.color),
       colors,
       hoist: options.hoist,
       permissions: options.permissions === undefined ? undefined : new PermissionsBitField(options.permissions),
--- a/packages/discord.js/src/managers/RoleManager.js
+++ b/packages/discord.js/src/managers/RoleManager.js
@@ -169,14 +169,11 @@ class RoleManager extends CachedManager {
       if (typeof icon !== 'string') icon = undefined;
     }
 
-    let colors;
-    if (options.colors) {
-      colors = {
-        primary_color: resolveColor(options.colors.primaryColor),
-        secondary_color: options.colors.secondaryColor && resolveColor(options.colors.secondaryColor),
-        tertiary_color: options.colors.tertiaryColor && resolveColor(options.colors.tertiaryColor),
-      };
-    }
+    const colors = options.colors && {
+      primary_color: resolveColor(options.colors.primaryColor),
+      secondary_color: options.colors.secondaryColor && resolveColor(options.colors.secondaryColor),
+      tertiary_color: options.colors.tertiaryColor && resolveColor(options.colors.tertiaryColor),
+    };
 
     const data = await this.client.rest.post(Routes.guildRoles(this.guild.id), {
       body: {
@@ -233,14 +230,11 @@ class RoleManager extends CachedManager {
       if (typeof icon !== 'string') icon = undefined;
     }
 
-    let colors;
-    if (options.colors) {
-      colors = {
-        primary_color: resolveColor(options.colors.primaryColor),
-        secondary_color: options.colors.secondaryColor && resolveColor(options.colors.secondaryColor),
-        tertiary_color: options.colors.tertiaryColor && resolveColor(options.colors.tertiaryColor),
-      };
-    }
+    const colors = options.colors && {
+      primary_color: resolveColor(options.colors.primaryColor),
+      secondary_color: options.colors.secondaryColor && resolveColor(options.colors.secondaryColor),
+      tertiary_color: options.colors.tertiaryColor && resolveColor(options.colors.tertiaryColor),
+    };
 
     const body = {
       name: options.name,
--- a/packages/discord.js/src/managers/RoleManager.js
+++ b/packages/discord.js/src/managers/RoleManager.js
@@ -110,11 +110,22 @@ class RoleManager extends CachedManager {
    */
 
   /**
+   * @typedef {Object} RoleColors
+   * @property {number} primaryColor The primary color of the role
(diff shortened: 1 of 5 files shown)


packages/discord.js/src/managers/RoleManager.js
```diff
@@ -212,9 +233,19 @@ class RoleManager extends CachedManager {
...
     }
 
+    let colors;
+    if (options.colors) {
+      colors = {
+        primary_color: resolveColor(options.colors.primaryColor),
+        secondary_color: options.colors.secondaryColor && resolveColor(options.colors.secondaryColor),
+        tertiary_color: options.colors.tertiaryColor && resolveColor(options.colors.tertiaryColor),
+      };
+    }
```
Jiralite: Would `const colors = options.colors && { ... };` not be fine?

packages/discord.js/src/managers/RoleManager.js
```diff
@@ -109,12 +109,23 @@ class RoleManager extends CachedManager {
...
+   */
+
   /**
    * Options used to create a new role.
    *
    * @typedef {Object} RoleCreateOptions
    * @property {string} [name] The name of the new role
-   * @property {ColorResolvable} [color] The data to create the role with
+   * @property {ColorResolvable} [color] The color to create the role with.
+   * This will still be returned by the API, but using the `colors` field is recommended when doing requests
```
Qjuh: I know this is the wording in the docs, but to me it reads like this should get a `deprecated` treatment or (since this targets main) simply be removed since djs shouldn't maintain both fields in the future.

packages/discord.js/src/managers/RoleManager.js
```diff
@@ -158,10 +169,17 @@ class RoleManager extends CachedManager {
...
+      primary_color: resolveColor(options.colors.primaryColor),
+      secondary_color: options.colors.secondaryColor && resolveColor(options.colors.secondaryColor),
+      tertiary_color: options.colors.tertiaryColor && resolveColor(options.colors.tertiaryColor),
+    };
+
     const data = await this.client.rest.post(Routes.guildRoles(this.guild.id), {
       body: {
         name,
         color,
+        colors,
```
Qjuh, line 194: We shouldn't send color and colors here. Instead (if we keep the color option, which we shouldn't imho, see above) colors should be assigned the value `colors ??= {primary_color: color}` or rather as ternary when assigned.

packages/discord.js/src/managers/RoleManager.js
```diff
@@ -212,9 +230,16 @@ class RoleManager extends CachedManager {
...
+    const colors = options.colors && {
+      primary_color: resolveColor(options.colors.primaryColor),
+      secondary_color: options.colors.secondaryColor && resolveColor(options.colors.secondaryColor),
+      tertiary_color: options.colors.tertiaryColor && resolveColor(options.colors.tertiaryColor),
+    };
+
     const body = {
       name: options.name,
       color: options.color === undefined ? undefined : resolveColor(options.color),
+      colors,
```
Qjuh, line 253: Ditto

packages/discord.js/src/structures/Role.js
```diff
@@ -66,6 +66,19 @@ class Role extends Base {
       this.color = data.color;
     }
```
Qjuh, line 69: Remove (for backport deprecate) the color property in favor of colors imho

packages/discord.js/src/structures/Role.js
```diff
@@ -257,7 +270,9 @@ class Role extends Base {
    *
    * @typedef {Object} RoleData
    * @property {string} [name] The name of the role
-   * @property {ColorResolvable} [color] The color of the role, either a hex string or a base 10 number
+   * @property {ColorResolvable} [color] The color of the role, either a hex string or a base 10 number.
+   * This will still be returned by the API, but using the `colors` field is recommended when doing requests
```
Qjuh: Ditto

packages/discord.js/src/structures/Role.js
```diff
@@ -330,6 +345,22 @@ class Role extends Base {
...
+   * @param {RoleColors} colors The colors of the role
+   * @param {string} [reason] Reason for changing the role's colors
+   * @returns {Promise<Role>}
+   * @example
+   * // Set the colors of a role
+   * role.setColors({ primaryColor: '#FF0000', secondaryColor: '#00FF00', tertiaryColor: '#0000FF' })
+   *   .then(updated => console.log(`Set colors of role to ${updated.colors}`))
+   *   .catch(console.error);
+   */
+  async setColors(colors, reason) {
```
Qjuh, line 352: setColor() should call this internally, instead of using the old color property in edit()

packages/discord.js/src/structures/Role.js
```diff
@@ -476,6 +507,9 @@ class Role extends Base {
       this.id === role.id &&
       this.name === role.name &&
       this.color === role.color &&
+      this.colors.primaryColor === role.colors.primaryColor &&
```
Qjuh, line 501: The equality check on color should be replaced by these, not added on top.

packages/discord.js/typings/index.d.ts
```diff
@@ -2834,9 +2834,16 @@ export class RichPresenceAssets {
...
+export interface RoleColors {
+  primaryColor: number;
+  secondaryColor: number | null;
+  tertiaryColor: number | null;
+}
+
 export class Role extends Base {
   private constructor(client: Client<true>, data: APIRole, guild: Guild);
   public color: number;
+  public colors: RoleColors;
```
Qjuh, line 2840: remove color here and the other places in typings too then.

Qjuh (changes requested):

packages/discord.js/src/managers/RoleManager.js
vladfrangu: > since djs shouldn't maintain both fields in the future

If both are supported at the API level we'll support both

Qjuh: Message#interaction is also "supported" at the API level, yet we removed it in main and deprecated in v14, because it's deprecated on API level...

vladfrangu: Color isn't deprecated tho, is it?

Jiralite: It is not marked as deprecated in the documentation.

Qjuh: > color will still be returned by the API, but using the colors field is recommended when doing requests.

This is straight from documentation though. So we shouldn't send `color` anymore.

vladfrangu: Recommended != deprecated, we should keep both until its fully deprecated

Qjuh: The way I read it: color will stay in the received payload (because removing it would be a breaking change) but you should not send it in a request anymore. So why would we continue sending it when there's a recommended alternative? It doesn't need to be restricted for users (although I'm even in favor of doing that for `main`) but at least the REST payload we construct and send should use the recommended structure.

vladfrangu: I don't know how you're reading that like `you should not use it`, when you absolutely can use it 😕. Its a recommendation to use `colors` instead, not a requirement 🤷. `color` is not deprecated, we keep it and support sending it

vladfrangu: After discussing internally some more, we agreed that we will possibly keep the `color` property (maybe turning it into a getter, tbd), and will pass `color` as primary_color on edit

packages/discord.js/src/managers/RoleManager.js
```diff
@@ -109,12 +109,21 @@ class RoleManager extends CachedManager {
...
+   * When sending `tertiaryColor` the API enforces the role color to be a holographic style with values of `primaryColor = 11127295`, `secondaryColor = 16759788`, and `tertiaryColor = 16761760`
+   */
+
   /**
    * Options used to create a new role.
    *
    * @typedef {Object} RoleCreateOptions
    * @property {string} [name] The name of the new role
-   * @property {ColorResolvable} [color] The data to create the role with
+   * @property {RoleColors} [colors] The colors to create the role with
```
Qjuh:
```suggestion
   * @property {RoleColorsResolvable} [colors] The colors to create the role with
```

packages/discord.js/src/structures/Role.js
```diff
@@ -257,7 +261,7 @@ class Role extends Base {
    *
    * @typedef {Object} RoleData
    * @property {string} [name] The name of the role
-   * @property {ColorResolvable} [color] The color of the role, either a hex string or a base 10 number
+   * @property {RoleColors} [colors] The colors of the role
```
Qjuh:
```suggestion
   * @property {RoleColorsResolvable} [colors] The colors of the role
```

packages/discord.js/src/structures/Role.js
```diff
@@ -315,19 +319,19 @@ class Role extends Base {
   }
 
   /**
-   * Sets a new color for the role.
+   * Sets new colors for the role.
    *
-   * @param {ColorResolvable} color The color of the role
-   * @param {string} [reason] Reason for changing the role's color
+   * @param {RoleColors} colors The colors of the role
```
Qjuh:
```suggestion
   * @param {RoleColorsResolvable} colors The colors of the role
```

packages/discord.js/typings/index.d.ts
```diff
@@ -2864,7 +2870,7 @@ export class Role extends Base {
     channel: NonThreadGuildBasedChannel | Snowflake,
     checkAdmin?: boolean,
   ): Readonly<PermissionsBitField>;
-  public setColor(color: ColorResolvable, reason?: string): Promise<Role>;
+  public setColors(colors: RoleColors, reason?: string): Promise<Role>;
```
```suggestion
  public setColors(colors: RoleColorsResolvable, reason?: string): Promise<Role>;
```

packages/discord.js/typings/index.d.ts
```diff
@@ -6797,7 +6803,7 @@ export interface ResolvedOverwriteOptions {
 }
 
 export interface RoleData {
-  color?: ColorResolvable;
+  colors?: RoleColors;
```
```suggestion
  colors?: RoleColorsResolvable;
```

packages/discord.js/typings/index.d.ts
```diff
@@ -2834,9 +2834,15 @@ export class RichPresenceAssets {
   public smallImageURL(options?: ImageURLOptions): string | null;
 }
 
+export interface RoleColors {
+  primaryColor: number;
+  secondaryColor: number | null;
+  tertiaryColor: number | null;
+}
+
```
Qjuh, line 2831:
```suggestion

export interface RoleColorsResolvable {
  primaryColor: ColorResolvable;
  secondaryColor?: ColorResolvable
  tertiaryColor?: ColorResolvable;
}

```

packages/discord.js/src/managers/RoleManager.js
```diff
@@ -109,12 +109,21 @@ class RoleManager extends CachedManager {
...
 
+  /**
+   * @typedef {Object} RoleColors
+   * @property {number} primaryColor The primary color of the role
+   * @property {?number} secondaryColor The secondary color of the role.
+   * This will make the role a gradient between the other provided colors
+   * @property {?number} tertiaryColor The tertiary color of the role.
+   * When sending `tertiaryColor` the API enforces the role color to be a holographic style with values of `primaryColor = 11127295`, `secondaryColor = 16759788`, and `tertiaryColor = 16761760`
+   */
+
```
Qjuh, line 121:
```suggestion

  /**
   * @typedef {Object} RoleColorsResolvable
   * @property {ColorResolvable} primaryColor The primary color of the role
   * @property {ColoResolvable} [secondaryColor] The secondary color of the role.
   * This will make the role a gradient between the other provided colors
   * @property {ColorResolvable} [tertiaryColor] The tertiary color of the role.
   * When sending `tertiaryColor` the API enforces the role color to be a holographic style with values of `primaryColor = 11127295`, `secondaryColor = 16759788`, and `tertiaryColor = 16761760`
   */

```

packages/discord.js/src/structures/Role.js
```diff
@@ -57,13 +57,26 @@ class Role extends Base {
...
     }
 
-    if ('color' in data) {
+    /**
+     * @typedef {Object} RoleColors
+     * @property {number} primaryColor The primary color of the role
+     * @property {?number} secondaryColor The secondary color of the role.
+     * This will make the role a gradient between the other provided colors
+     * @property {?number} tertiaryColor The tertiary color of the role.
+     * When sending `tertiaryColor` the API enforces the role color to be a holographic style with values of `primaryColor = 11127295`, `secondaryColor = 16759788`, and `tertiaryColor = 16761760`
```
vladfrangu: Let's expose this color triplet via constants I'd say?

vladfrangu (approved):

SpaceEEC (approved):
Using decimal numbers for colors hurts me a bit. 😬

vladfrangu (approved):

PR #10962, opened 2025-07-02
feat: support user guilds

sdanialraza:
- https://github.com/discord/discord-api-docs/pull/6836


packages/discord.js/src/structures/User.js
```diff
@@ -151,6 +151,30 @@ class User extends Base {
     } else {
       this.avatarDecorationData = null;
     }
+
+    /**
+     * @typedef {Object} UserPrimaryGuild
```
Jiralite, line 185: I believe it may be fine to call this `PrimaryGuild`. Thoughts?

packages/discord.js/src/structures/User.js
```diff
@@ -246,6 +270,19 @@ class User extends Base {
...
+   * A link to the user's guild tag badge.
+   *
+   * @param {ImageURLOptions} [options={}] Options for the image URL
+   * @returns {?string}
+   */
+  guildTagBadgeURL(options = {}) {
+    return (
+      this.primaryGuild?.badge &&
+      this.client.rest.cdn.guildTagBadge(this.primaryGuild.identityGuildId, this.primaryGuild.badge, options)
+    );
```
almeidx: This would return `undefined` otherwise.

```suggestion
    return this.primaryGuild?.badge
      ? this.client.rest.cdn.guildTagBadge(this.primaryGuild.identityGuildId, this.primaryGuild.badge, options)
      : null;
```

packages/discord.js/src/structures/User.js
```diff
@@ -151,6 +151,30 @@ class User extends Base {
...
+     * @property {?string} badge - The server tag badge hash
+     */
+
+    if (data.primary_guild) {
+      /**
+       * The primary guild of the user
+       *
+       * @type {?UserPrimaryGuild}
+       */
+      this.primaryGuild = {
```
almeidx: Typically we'd also have some sort of getter for the `Guild` instance.
But since the name of this is `primaryGuild`... Not sure what to do about that

packages/discord.js/src/structures/User.js
```diff
@@ -246,6 +270,19 @@ class User extends Base {
...
+   *
+   * @param {ImageURLOptions} [options={}] Options for the image URL
+   * @returns {?string}
+   */
+  guildTagBadgeURL(options = {}) {
+    return (
+      this.primaryGuild?.badge &&
+      this.client.rest.cdn.guildTagBadge(this.primaryGuild.identityGuildId, this.primaryGuild.badge, options)
+    );
+  }
```
Jiralite: `UserPrimaryGuild` prefers "server tag", but this prefers "guild tag". We should probably choose one and stick to it.

I see the CDN route is `guild-tag-badges`, but the descriptions written in the documentation use "server tag". I'd personally stick with "guild tags" because the CDN probably will not change.

packages/discord.js/typings/index.d.ts
```diff
@@ -3534,12 +3541,14 @@ export class User extends Base {
   public get hexAccentColor(): HexColorString | null | undefined;
   public id: Snowflake;
   public get partial(): false;
+  public primaryGuild: UserPrimaryGuild;
```
```suggestion
  public primaryGuild: UserPrimaryGuild | null;
```

packages/rest/src/lib/CDN.ts
```diff
@@ -318,6 +318,17 @@ export class CDN {
 		return `${this.cdn}${CDNRoutes.soundboardSound(soundId)}`;
 	}
 
+	/**
+	 * Generates a URL for a guild tag badge.
+	 *
+	 * @param guildId - The guild id
+	 * @param badgeHash - The hash provided by Discord for this badge
```
Jiralite: I mean, where else would the badge be from?

```suggestion
	 * @param badgeHash - The hash of the badge
```

almeidx (changes requested):

packages/discord.js/src/structures/User.js
Jiralite: What if we make `PrimaryGuild` as a class which has these properties? Then we could include a getter to the guild.

almeidx: Would we then also duplicate the `guildTagBadgeURL` method there? It's currently directly on the `User` class

packages/discord.js/src/structures/User.js
sdanialraza: I think `UserPrimaryGuild` helps avoid confusion that it's related to `Guild`

packages/rest/src/lib/CDN.ts
sdanialraza: That's the description for every hash parameter in this class

Jiralite: I'd say it's superfluous.

packages/discord.js/typings/index.d.ts
```diff
@@ -31,10 +31,7 @@ import {
   APIComponentInModalActionRow,
   APIContainerComponent,
   APIEmbed,
-  APIEmbedAuthor,
```
Qjuh, line 34: why wasn't this caught by the linter...

packages/discord.js/src/structures/User.js
```diff
@@ -151,6 +151,30 @@ class User extends Base {
...
+       *
+       * @type {?UserPrimaryGuild}
+       */
+      this.primaryGuild = {
+        identityGuildId: data.primary_guild.identity_guild_id,
+        identityEnabled: data.primary_guild.identity_enabled,
+        tag: data.primary_guild.tag,
+        badge: data.primary_guild.badge,
+      };
+    } else {
```
Qjuh, line 208: if a patch call doesn't include primary_guild this will unset it although it should remain unchanged. Same applies to avatar_decoration_data though, no idea if that was discussed for that before. Imho it should get the same treatment as banner or accent_color did, unless we got something that guarantees this behaves differently.

vladfrangu: _If_ it can happen, the if check should check key in data

packages/discord.js/src/structures/User.js
```diff
@@ -138,18 +138,42 @@ class User extends Base {
      * @property {Snowflake} skuId The id of the avatar decoration's SKU
      */
 
-    if (data.avatar_decoration_data) {
+    if ('avatar_decoration_data' in data) {
```
Qjuh, line 142: This will now cause a payload of `{avatar_decoration_data: null}` to cause `User#avatarDecorationData` to be `{asset: undefined, skuId: undefined}` which is not valid according to the types. So should probably keep both the `in` check and the `value is falsey`.

packages/discord.js/src/structures/User.js
```diff
@@ -138,18 +138,42 @@ class User extends Base {
...
+
+    /**
+     * @typedef {Object} UserPrimaryGuild
+     * @property {?Snowflake} identityGuildId - The id of the user's primary guild
+     * @property {?boolean} identityEnabled - Whether the user is displaying the primary guild's tag
+     * @property {?string} tag - The user's guild tag. Limited to 4 characters
+     * @property {?string} badge - The guild tag badge hash
+     */
+
+    if ('primary_guild' in data) {
```
Qjuh, line 192: Ditto here, with the difference that the (wrong) value here would actually be allowed by the types.

Qjuh (approved):

packages/discord.js/src/structures/User.js
```diff
@@ -176,6 +180,34 @@ class User extends Base {
     } else {
       this.collectibles = null;
     }
+
+    /**
+     * @typedef {Object} UserPrimaryGuild
+     * @property {?Snowflake} identityGuildId - The id of the user's primary guild
+     * @property {?boolean} identityEnabled - Whether the user is displaying the primary guild's tag
+     * @property {?string} tag - The user's guild tag. Limited to 4 characters
+     * @property {?string} badge - The guild tag badge hash
```
Jiralite:
```suggestion
     * @property {?Snowflake} identityGuildId The id of the user's primary guild
     * @property {?boolean} identityEnabled Whether the user is displaying the primary guild's tag
     * @property {?string} tag The user's guild tag. Limited to 4 characters
     * @property {?string} badge The guild tag badge hash
```

Jiralite (approved):

PR #10969, opened 2025-07-06
refactor: Deprecate `ready` event in favor of `clientReady`

Jiralite:
`clientReady` is what should be used moving forward.

### DIFF
--- a/packages/discord.js/src/client/websocket/WebSocketManager.js
+++ b/packages/discord.js/src/client/websocket/WebSocketManager.js
@@ -387,7 +387,7 @@ class WebSocketManager extends EventEmitter {
       deprecationEmitted = true;
 
       process.emitWarning(
-        'The ready event has been renamed to clientReady. Please use that instead',
+        'The ready event has been renamed to clientReady to distinguish it from the gateway READY event and will only emit under that name in v15. Please use clientReady instead.',
         'DeprecationWarning',
       );
     }
--- a/packages/discord.js/src/client/websocket/WebSocketManager.js
+++ b/packages/discord.js/src/client/websocket/WebSocketManager.js
@@ -19,6 +19,7 @@ const Status = require('../../util/Status');
 const WebSocketShardEvents = require('../../util/WebSocketShardEvents');
 
 let zlib;
+let deprecationEmitted = false;
 
 try {
   zlib = require('zlib-sync');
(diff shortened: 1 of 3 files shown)


packages/discord.js/src/client/websocket/WebSocketManager.js
```diff
@@ -379,6 +380,17 @@ class WebSocketManager extends EventEmitter {
     /**
      * Emitted when the client becomes ready to start working.
      * @event Client#ready
+     * @deprecated Use {@link Client#event:clientReady} instead.
+     * @param {Client} client The client
+     */
+    if (this.client.emit('ready', this.client) && !deprecationEmitted) {
+      deprecationEmitted = true;
+      process.emitWarning('The ready event is deprecated. Use clientReady instead.', 'DeprecationWarning');
```
vladfrangu:
```suggestion
      process.emitWarning('The ready event has been renamed to clientReady. Please use that instead', 'DeprecationWarning');
```

vladfrangu (dismissed):

didinele (approved):

packages/discord.js/src/util/Events.js
```diff
@@ -108,7 +108,7 @@ module.exports = {
   ChannelDelete: 'channelDelete',
   ChannelPinsUpdate: 'channelPinsUpdate',
   ChannelUpdate: 'channelUpdate',
-  ClientReady: 'ready',
+  ClientReady: 'clientReady',
```
Qjuh, line 111: This feels to me like a breaking change.

didinele: By some technicality you're right, but we historically treated it as not one when we made identical changes since this way by using `Events.ClientReady` you won't get the warning and realistically, what code are we actually breaking? A conditional that asserts the value of the constant?

Qjuh: We didn't do this for Events.WebhooksUpdate which still has the value webhookUpdate in v14 despite it being marked as deprecated the same way.

vladfrangu: Its not really a breaking change unless you, for whatever reason, did `Events.ClientReady === "ready"`, which at that point I'd question what you're doing. In fact I'd even argue its a correction, all other enum names can easily be converted to enum keys, not this one

Qjuh: And GuildEmojiCreate/Delete/Update and GuildRoleCreate/Delete/Update and (sorry to repeat myself) Webhook**s**Update

packages/discord.js/src/client/websocket/WebSocketManager.js
```diff
@@ -379,6 +380,21 @@ class WebSocketManager extends EventEmitter {
...
      * Emitted when the client becomes ready to start working.
      * @event Client#ready
+     * @deprecated Use {@link Client#event:clientReady} instead.
+     * @param {Client} client The client
+     */
+    if (this.client.emit('ready', this.client) && !deprecationEmitted) {
+      deprecationEmitted = true;
+
+      process.emitWarning(
+        'The ready event has been renamed to clientReady. Please use that instead',
```
Qjuh:
```suggestion
        'The ready event has been renamed to clientReady to distinguish it from the gateway READY event and will only emit under that name in v15. Please use clientReady instead.',
```
mostly to reduce the outrage this might cause in a minor (not really a patch is it?) release

Qjuh (changes requested):

PR #10983, opened 2025-07-12
feat: implement structures in mainlib

Qjuh:

This is a very early proof-of-concept stage to discuss how `@discordjs/structures` might get implemented in next discord.js version.

### DIFF
--- a/packages/next/src/Client.ts
+++ b/packages/next/src/Client.ts
@@ -1,0 +1,50 @@
+import process from 'node:process';
+import { Collection } from '@discordjs/collection';
+import type { RESTGetAPIGatewayBotResult, Snowflake } from '@discordjs/core';
+import { Client as CoreClient, Routes, GatewayDispatchEvents } from '@discordjs/core';
+import { REST } from '@discordjs/rest';
+import type { Structure } from '@discordjs/structures';
+import { WebSocketManager } from '@discordjs/ws';
+import type { CacheConstructor, Cache } from './util/types';
+
+export class Client {
+	private core!: CoreClient;
+
+	public options: { token: string | undefined };
+
+	#token: string;
+
+	public CacheConstructor<Value extends Structure<object>>(_value: new (...args: any[]) => Value) {
+		return new this.cacheConstructor() as Cache<Snowflake, Value>;
+	}
+
+	protected cacheConstructor: CacheConstructor;
+
+	public constructor({ token = process.env.DISCORD_TOKEN }: { token?: string }, cache: CacheConstructor = Collection) {
+		this.options = { token };
+		this.#token = token!;
+		this.cacheConstructor = cache;
+	}
+
+	public async login(token?: string) {
+		if (token) {
+			this.#token = token;
+		}
+
+		const rest = new REST().setToken(this.#token);
+		const gateway = new WebSocketManager({
+			fetchGatewayInformation: async () => rest.get(Routes.gatewayBot()) as Promise<RESTGetAPIGatewayBotResult>,
+			intents: 0,
+			token: this.#token,
+		});
+		this.core = new CoreClient({
+			gateway,
+			rest,
+		});
+
+		// TODO: remove once we actually implement anything using this.core
+		this.core.on(GatewayDispatchEvents.Ready, console.log);
+
+		await gateway.connect();
+	}
+}
--- a/packages/next/src/structures/mixins/BaseChannelMixin.ts
+++ b/packages/next/src/structures/mixins/BaseChannelMixin.ts
@@ -1,0 +1,23 @@
+/* eslint-disable @typescript-eslint/dot-notation */
+import { equal } from 'node:assert';
+import type { ChannelType } from '@discordjs/core';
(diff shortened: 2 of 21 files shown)


packages/next/src/structures/mixins/BaseChannelMixin.ts
```diff
@@ -0,0 +1,24 @@
...
+
+	public async delete() {
+		await container.get('client')['core'].api.channels.delete(this.id);
+		return this;
+	}
+
+	public async fetch() {
+		const data = await container.get('client')['core'].api.channels.get(this.id);
+
+		equal(data.type, this.type);
```
kyranet, line 21: This can throw an error when a text channel is toggled into an announcement channel

packages/next/src/util/container.ts
```diff
@@ -0,0 +1,30 @@
+import { Collection } from '@discordjs/collection';
+import type { Client } from '../Client';
+
+export class Container extends Collection<ContainerKey, ContainerValue> {}
```
didinele: Are the semantics of a collection that much more desired here rather than a plain object? Even with plain module augmentation I still think objects are a bit more desirable.

There's a silly argument to be made about performance, assuming [this](https://v8.dev/blog/fast-properties) is still valid, but other than that, I just find the semantics to be better?

As it stands, you're overloading `get` and `has`, which goes to imply that's the main interaction we're expecting users to have with the container, presumably other than `set` to.. add stuff.

Is there situations where we want users removing stuff? I can't really imagine that, feels to me the container should only be added upon. All the other collection methods are also useless.

At least with an object you're cleanly doing `container.Client` or `const { Client } = container;`. Added benefit, we can probably get away with

```ts
interface Container {
  readonly client: Client;
}
```

and encourage users to augment in a similar fashion, to avoid accidental mutation (holding onto my instinct that once something is added to the container, it's there to stay as-is). Yes, forces us to implement a lie (expose a function that initially sets `client` despite its readonly state/tell users to ignore the compiler error/use `Reflect.set`), but I think that's alright for just the initialization stage.

packages/next/src/Client.ts
```diff
@@ -0,0 +1,54 @@
...
+import type { RESTGetAPIGatewayBotResult, Snowflake } from '@discordjs/core';
+import { Client as CoreClient, Routes, GatewayDispatchEvents } from '@discordjs/core';
+import { REST } from '@discordjs/rest';
+import type { Structure } from '@discordjs/structures';
+import { WebSocketManager } from '@discordjs/ws';
+import { CollectionCache, type StructureCreator, type CacheConstructor, type StructureCache } from './util/cache.js';
+import { container } from './util/container.js';
+
+export class Client {
+	private core!: CoreClient;
```
didinele: I'm not 100% confident we want this. We're not gonna be plainly re-emitting events from the `CoreClient` anyway. There's a case to be made it's cleaner/easier for us to plainly handle a `Gateway` (imported from `Core`) and a `API` (which perhaps we can construct ourselves from a `REST` instance. depends what we want the user to pass)

packages/next/src/util/cache.ts
```diff
@@ -0,0 +1,53 @@
+import { Collection } from '@discordjs/collection';
+import { kPatch, type Structure } from '@discordjs/structures';
+import type { Awaitable } from '@discordjs/util';
+import type { Snowflake } from 'discord-api-types/globals';
+
+export type RawAPIType<Value extends Structure<{}>> = Value extends Structure<infer Type> ? Type : never;
+
+export interface StructureCache<
+	Value extends Structure<{ id: Snowflake }>,
```
didinele: Are we sure we want to provide this abstraction? Considering how easy we've made the structure <-> api data conversion, can't we justify keeping the caching stuff simpler and making it only interact with raw API data, so the caller is responsible for conversion?

packages/next/src/Client.ts
vladfrangu: One thing I absolutely want (and I'd even want to do in current djs) is to always have a `client['api']` which is the /core API class. I also agree that maybe we should have two separate properties for the gateway/rest, since the idea with next is that you should be able to wire up your own gw implementation too

vladfrangu: Also it should be protected, not private (so we can do major bumps without needing major releases on this module's end but users can also use it should they desire)

ckohen: Just saying it to say it. I was under the assumption that we would have core (or at least the api half of core) as a prop like this on the client, and we'd use that in all of [whatever managers becomes]. Totally fine with me to access it with client['api'] on internals.

packages/next/src/util/cache.ts
Qjuh: Wouldn't that mean that we slow down collection cache by always converting from/to JSON for the benefit of implementers of alternative caching solutions (pretty sure that's not the majority of our users) only having to handle the JSON?

packages/next/src/structures/mixins/BaseChannelMixin.ts
Qjuh: That is a general issue with patching that kind of channel then and would need to be addressed as special case later anyway.

PR #10911, opened 2025-05-25
feat!: Support animated WebP

Jiralite:
**Please describe the changes this PR makes and why it should be merged:**

- https://github.com/discord/discord-api-docs/pull/7404
- https://github.com/discord/discord-api-docs/pull/7568

BREAKING CHANGE: `"webp?animated=true"` is now the default for animated assets.

**Status and versioning classification:**

- Code changes have been tested against the Discord API, or there are no code changes
- I know how to update typings and have done so, or typings don't need updating
- This PR includes breaking changes (methods removed or renamed, parameters moved or removed)

PR #10987, opened 2025-07-14
feat: Support animated WebP

Jiralite:
Backports #10911.

### DIFF
--- a/packages/rest/__tests__/CDN.test.ts
+++ b/packages/rest/__tests__/CDN.test.ts
@@ -54,12 +54,20 @@ test('discoverySplash default', () => {
 	expect(cdn.discoverySplash(id, hash)).toEqual(`${baseCDN}/discovery-splashes/${id}/${hash}.webp`);
 });
 
-test('emoji default', () => {
+test('emoji', () => {
 	expect(cdn.emoji(id)).toEqual(`${baseCDN}/emojis/${id}.webp`);
 });
 
 test('emoji gif', () => {
 	expect(cdn.emoji(id, 'gif')).toEqual(`${baseCDN}/emojis/${id}.gif`);
+});
+
+test('emoji animated', () => {
+	expect(cdn.emoji(id, { animated: true })).toEqual(`${baseCDN}/emojis/${id}.webp?animated=true`);
+});
+
+test('emoji with GIF format', () => {
+	expect(cdn.emoji(id, { extension: 'gif' })).toEqual(`${baseCDN}/emojis/${id}.gif`);
 });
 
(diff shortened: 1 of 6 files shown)


packages/rest/__tests__/CDN.test.ts
```diff
@@ -54,14 +54,22 @@ test('discoverySplash default', () => {
...
 	expect(cdn.emoji(id, 'gif')).toEqual(`${baseCDN}/emojis/${id}.gif`);
 });
 
+test('emoji animated', () => {
+	expect(cdn.emoji(id, { animated: true })).toEqual(`${baseCDN}/emojis/${id}.webp?animated=true`);
+});
+
+test('emoji with GIF format', () => {
+	expect(cdn.emoji(id, { extension: 'gif' })).toEqual(`${baseCDN}/emojis/${id}.gif`);
+});
```
vladfrangu, line 71: What happens if you provide both?

vladfrangu (approved):

packages/rest/__tests__/CDN.test.ts
Jiralite: `"https://cdn-discord.com/emojis/123456.gif?animated=true"`.

It should be noted that the typings disallow this:

```
No overload matches this call.
  Overload 1 of 2, '(emojiId: string, options?: Readonly<EmojiURLOptions> | undefined): string', gave the following error.
    Object literal may only specify known properties, and 'animated' does not exist in type 'Readonly<EmojiURLOptionsNotWebp>'.
  Overload 2 of 2, '(emojiId: string, extension?: "gif" | "webp" | "png" | "jpg" | "jpeg" | undefined): string', gave the following error.
    Argument of type '{ extension: string; animated: boolean; }' is not assignable to parameter of type '"gif" | "webp" | "png" | "jpg" | "jpeg" | undefined'.ts(2769)
```

packages/rest/src/lib/CDN.ts
```diff
@@ -14,39 +14,60 @@ import { deprecationWarning } from './utils/utils.js';
...
-	 * Whether or not to prefer the static version of an image asset.
+	 * Whether to prefer the static asset.
 	 */
 	forceStatic?: boolean;
 }
 
 /**
  * The options to use when making a CDN URL
  */
-export interface MakeURLOptions {
```
Qjuh: not exporting this anymore is a breaking change, isn't it?

Jiralite: Good catch. This will be reverted.

Qjuh (approved):

PR #11000, opened 2025-07-17
fix: Channel Name ratelimit make channel.edit functions freeze

ErdemGKSL:
**Description of Changes and Justification:**

This PR adjusts the request handler to avoid unnecessary sublimit blocking when editing a channel. Currently, editing a channel name three times triggers a sublimit rate limit. Following that, all subsequent `channel.edit()` calls (even ones *not* modifying the name) are queued behind the rate limit, which is incorrect.

This PR ensures that `channel.edit()` requests which do *not* modify the `name` field are not blocked by the name change sublimit. This improves compliance with Discord's rate-limiting behavior and prevents unrelated modifications (e.g., permissions, user limits, etc.) from being delayed.

* Code changes have been tested against the Discord API
* Typings don't need updating
* This PR changes the library's interface: **No**
* This PR includes breaking changes: **No**
* This PR only includes non-code changes: **No**

### DIFF
--- a/packages/rest/src/lib/handlers/SequentialHandler.ts
+++ b/packages/rest/src/lib/handlers/SequentialHandler.ts
@@ -158,9 +158,10 @@ export class SequentialHandler implements IHandler {
 				const wait = queue.wait();
 				this.#asyncQueue.shift();
 				await wait;
+			} else if (this.#sublimitPromise) {
+				// Stall requests while the sublimit queue gets processed
+				await this.#sublimitPromise.promise;
 			}
-			// Note: Removed the "else if (this.#sublimitPromise)" block to allow non-sublimited requests
-			// to proceed independently of the sublimit queue
 		}
 
 		try {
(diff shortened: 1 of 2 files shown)


ckohen:
This is working exactly as designed. the sublimitPromise only exists when a sublimited request is actively on the wire, which means we should not be sending another request. The real reason you encounter this is because disord.js *always* passes name and topic via that edit function (on the wire undefined goes away, but the body is checked before that, it might be worth considering checking if those properties are non-undefined, but I don't love that risk). See below

https://github.com/discordjs/discord.js/blob/43a995bef0dbe7c18631a20816ef6438f5c2cc51/packages/discord.js/src/managers/GuildChannelManager.js#L339-L343

Qjuh:
So I guess the real fix would be to dynamically build the body in the edit method to only include those properties if they are present in options.

vladfrangu:
> So I guess the real fix would be to dynamically build the body in the edit method to only include those properties if they are present in options.

Which we should've done from the get-go but yes

PR #11001, opened 2025-07-18
refactor!: make RedisBroker require consumer name

didinele:
In anticipation for TSR, I'm cooking some brokers stuff up again for the future.

While working on the Redis brokers, I realized that it's non-sense to default `name`, it was a wrong assumption I came with from RabbitMQ.

To spare some docs reading, a Redis *consumer group* (what we use under the hood) is used when you want to have multiple clients that see a *sub set* of the messages (evidently what we want with our brokers). The `name` here (or consumer name) identifies each one. If a client were to crash and restart automatically, with the user not providing the `name` argument, Redis ends up potentially leaving some messages in purgatory that were already assigned to the old name.

There's some more resilience I've noted we can add (like checking if there's messages in purgatory caused by a service downscale), but considering those tasks are relatively complex (low LOC diffs, but can have deep impact), I'm gonna split my PRs.

I've also used this one to update the outdated README examples.

### DIFF
--- a/packages/brokers/README.md
+++ b/packages/brokers/README.md
@@ -88,7 +88,7 @@ import { RPCRedisBroker } from '@discordjs/brokers';
 import Redis from 'ioredis';
 
 const broker = new RPCRedisBroker(new Redis(), {
-	// Equivelent to the group/name in pubsub, refer to the previous example.
+	// Equivalent to the group/name in pubsub, refer to the previous example.
 	group: 'responders',
 	name: `consumer-${UNIQUE_ID}`,
 });


packages/brokers/README.md
```diff
@@ -77,14 +87,18 @@ await broker.destroy();
 import { RPCRedisBroker } from '@discordjs/brokers';
 import Redis from 'ioredis';
 
-const broker = new RPCRedisBroker(new Redis());
+const broker = new RPCRedisBroker(new Redis(), {
+	// Equivelent to the group/name in pubsub, refer to the previous example.
```
iCrawl:
```suggestion
	// Equivalent to the group/name in pubsub, refer to the previous example.
```

packages/brokers/src/brokers/redis/BaseRedis.ts
```diff
@@ -23,36 +22,32 @@ export interface RedisBrokerOptions extends BaseBrokerOptions {
...
 /**
  * Default broker options for redis
  */
 export const DefaultRedisBrokerOptions = {
 	...DefaultBrokerOptions,
-	name: randomBytes(20).toString('hex'),
 	maxChunk: 10,
 	blockTimeout: 5_000,
-} as const satisfies Required<Omit<RedisBrokerOptions, 'group'>>;
+} as const satisfies Required<Omit<RedisBrokerOptions, 'group' | 'name'>>;
```
Qjuh, line 50: From a docs perspective it would be better to have name and group be in their own interface and have the options interfaces extend both interfaces and have another interface only extending the one contained in these defaults. Omit is always bad for docs. Apart from that this LGTM

Jiralite (approved):

iCrawl (approved):

PR #11003, opened 2025-07-18
feat(RedisBroker): recover unacked events

didinele:
Makes it so `listen` first checks if there are unacked events our consumer is responsible for before entering the regular polling loop.


didinele:
Superseded by #11004

PR #11007, opened 2025-07-20
fix: Remove trailing `color` references

Jiralite:

- Updated the documentation of `RoleManager#fetch()` to use `colors`
- `GuildMember#displayColor` was always returning `0` as it referenced a non-existent property
- Getting the `hexColor` of a role will cause a crash as `this.color` does not exist

### DIFF
--- a/packages/discord.js/src/managers/RoleManager.js
+++ b/packages/discord.js/src/managers/RoleManager.js
@@ -61,7 +61,7 @@ class RoleManager extends CachedManager {
    * @example
    * // Fetch a single role
    * message.guild.roles.fetch('222078108977594368')
-   *   .then(role => console.log(`The role color is: ${role.color}`))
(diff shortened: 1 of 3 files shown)


sdanialraza (approved):

vladfrangu (approved):
Does hexColor make sense to keep in its current form?

almeidx:
This is missing. And the typings do not correspond to the jsdoc / runtime implementation

https://github.com/discordjs/discord.js/blob/a7f711b30f3089a155a0ee721e2eff511d4759fb/packages/discord.js/src/managers/GuildManager.js#L201

Qjuh:
> This is missing. And the typings do not correspond to the jsdoc / runtime implementation
> 
> https://github.com/discordjs/discord.js/blob/a7f711b30f3089a155a0ee721e2eff511d4759fb/packages/discord.js/src/managers/GuildManager.js#L201

That's irrelevant as the route that uses is no longer available and errors either way.

PR #11020, opened 2025-07-30
feat(GuildMember): add 'clearTimeout' method

etoile2feu:
This pull request add ``clearTimeout`` method to guildMember.

Actually, you have to set the duration on **null** to cancel the timeout of a guild member.
The clearTimeout method would make the code more easy to read.

### DIFF
--- a/packages/discord.js/src/structures/GuildMember.js
+++ b/packages/discord.js/src/structures/GuildMember.js
@@ -530,7 +530,7 @@ class GuildMember extends Base {
    * Times this guild member out.
    *
    * @param {?number} timeout The duration in milliseconds
-   * for the member's communication to be disabled. Provide `null` to remove the timeout.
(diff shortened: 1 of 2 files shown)


Tristan-GPT:
I think it can be a good add. This method can make the code easier to read and a little bit more logic because call timeout function to remove the timeout is not very logic.

didinele (commented):
I am against this. Methods that hardcode a value and pass it to the actual API call method subtly make users make unnecessary calls

etoile2feu:
> I am against this. Methods that hardcode a value and pass it to the actual API call method subtly make users make unnecessary calls

Yes, I didn't think about that.
Should I close this PR ?

Jiralite:
Closing for the reason listed above. This is generally why helper methods should be avoided.

Regardless, thank you for your efforts!

PR #11026, opened 2025-08-02
feat: rpc

OfficialSirH:
TypeScript rewrite of the discord-rpc library moved to the monorepo


packages/rpc/.gitignore
Qjuh, line 1: Should already be handled by the top-level file

packages/rpc/examples/boops-javascript/.eslintrc.cjs
```diff
@@ -0,0 +1,3 @@
+module.exports = {
+  extends: ['eslint:recommended', '@electron-toolkit', '@electron-toolkit/eslint-config-prettier']
```
Qjuh: Probably should use the top-level eslint config and eslint-config-neon instead...

packages/rpc/package.json
```diff
@@ -0,0 +1,86 @@
...
+	"version": "0.1.0",
+	"description": "A simple RPC client for Discord",
+	"scripts": {
+		"build": "tsc --noEmit && tsup",
+		"build:docs": "tsc -p tsconfig.docs.json",
+		"test": "vitest run --config ../../vitest.config.ts",
+		"lint": "prettier --check . && cross-env TIMING=1 eslint --format=pretty src",
+		"format": "prettier --write . && cross-env TIMING=1 eslint --fix --format=pretty src",
+		"fmt": "pnpm run format",
+		"docs": "pnpm run build:docs && api-extractor run --local",
```
Qjuh:
```suggestion
		"docs": "pnpm run build:docs && api-extractor run --local --minify",
```

packages/rpc/package.json
```diff
@@ -0,0 +1,86 @@
...
+	},
+	"files": [
+		"dist"
+	],
+	"contributors": [
+		"Crawl <icrawltogo@gmail.com>",
+		"SpaceEEC <spaceeec@yahoo.com>",
+		"Vlad Frangu <me@vladfrangu.dev>",
+		"Aura Román <kyradiscord@gmail.com>"
+	],
```
Qjuh, line 44: Pretty sure you deserve to be in here...

Jiralite (dismissed):
Missing stuff:

https://github.com/discordjs/discord.js/blob/943c39e4c068447c968faa3814d3622e02dca3fb/.github/ISSUE_TEMPLATE/01-package_bug_report.yml#L11-L32

https://github.com/discordjs/discord.js/blob/943c39e4c068447c968faa3814d3622e02dca3fb/.github/ISSUE_TEMPLATE/03-feature_request.yml#L11-L33

https://github.com/discordjs/discord.js/blob/943c39e4c068447c968faa3814d3622e02dca3fb/.github/workflows/deprecate-version.yml#L5-L23

https://github.com/discordjs/discord.js/blob/943c39e4c068447c968faa3814d3622e02dca3fb/.github/workflows/documentation.yml#L86-L96

https://github.com/discordjs/discord.js/blob/943c39e4c068447c968faa3814d3622e02dca3fb/.github/workflows/documentation.yml#L219-L232

https://github.com/discordjs/discord.js/blob/943c39e4c068447c968faa3814d3622e02dca3fb/.github/workflows/release.yml#L9-L29

https://github.com/discordjs/discord.js/blob/943c39e4c068447c968faa3814d3622e02dca3fb/apps/website/package.json#L8

https://github.com/discordjs/discord.js/blob/943c39e4c068447c968faa3814d3622e02dca3fb/apps/website/src/util/constants.ts#L1-L16

https://github.com/discordjs/discord.js/blob/943c39e4c068447c968faa3814d3622e02dca3fb/packages/actions/src/uploadCoverage/action.yml#L1-L134

https://github.com/discordjs/discord.js/blob/943c39e4c068447c968faa3814d3622e02dca3fb/packages/scripts/src/shared.ts#L3-L18

The README.md at the root also needs updating (the source of RPC should ideally be removed and @discordjs/rpc should be added to the list of packages).

devsnek:
rpc code is licensed MIT and has several contributors. you'll need to either keep the MIT license (at the top of each file, for example) or reach out to them all to relicense it.

vladfrangu:
Wouldn't it be easier to keep RPC MIT licensed by keeping the license file as is? cc @discordjs/core

packages/rpc/__tests__/.gitkeep
Jiralite, line 1: Unsure if the stylistic preference is to remove this directory (which is fine). @vladfrangu @almeidx

almeidx: Will this package realistically have tests? If not, I'd rather get rid of the folder, script and test-related dependencies

vladfrangu: Hard to test RPC without mocking a lot of obscure whoknowswhats, i'm ok with dropping tests

packages/rpc/src/index.ts
```diff
@@ -0,0 +1,12 @@
+export * from './client.js';
+export * from './constants.js';
+export * from './ipc.js';
+export * from './RPCEventError.js';
+export * from './util.js';
+
+/**
+ * The {@link https://github.com/discordjs/RPC#readme | discord-rpc} version
```
Jiralite:
```suggestion
 * The {@link https://github.com/discordjs/discord.js/blob/main/packages/rpc#readme | @discordjs/rpc} version
```

packages/rpc/examples/boops-javascript/package.json
```diff
@@ -0,0 +1,32 @@
+{
+  "name": "boops-javascript",
+  "version": "1.0.0",
+  "description": "A minimal Electron application using discord-rpc with JavaScript",
```
Jiralite:
```suggestion
  "description": "A minimal Electron application using @discordjs/rpc with JavaScript.",
```

packages/rpc/examples/boops-typescript/package.json
```diff
@@ -0,0 +1,38 @@
+{
+  "name": "boops-typescript",
+  "version": "1.0.0",
+  "description": "A minimal Electron application using discord-rpc with TypeScript",
```
Jiralite:
```suggestion
  "description": "A minimal Electron application using @discordjs/rpc with TypeScript.",
```

packages/rpc/examples/boops-typescript/README.md
```diff
@@ -0,0 +1,34 @@
+# boops-typescript
+
+A minimal Electron application using discord-rpc with TypeScript
```
Jiralite:
```suggestion
A minimal Electron application using @discordjs/rpc with TypeScript.
```

packages/rpc/examples/boops-javascript/README.md
```diff
@@ -0,0 +1,34 @@
+# boops-javascript
+
+A minimal Electron application using discord-rpc with JavaScript
```
Jiralite:
```suggestion
A minimal Electron application using @discordjs/rpc with JavaScript.
```

Jiralite (dismissed):
Update this too:

https://github.com/discordjs/discord.js/blob/fdc305dd6a2b78879df0c15751037a894adedf3d/packages/discord.js/README.md#L145

I believe Vitest can be removed as a dependency now.

0sapphy (approved):

packages/rpc/src/client.ts
```diff
@@ -0,0 +1,580 @@
...
+
+	/**
+	 * Move the user to a voice channel
+	 *
+	 * @param id - Id of the voice channel
+	 * @param options - Options
+	 * @param options.timeout - Timeout for the command
+	 * @param options.force - Force this move. This should only be done if you
+	 *
+	 * have explicit permission from the user.
```
Jiralite: Should this be here? It looks awkward!

packages/rpc/src/client.ts
```diff
@@ -0,0 +1,580 @@
...
+	}
+
+	/**
+	 * Move the user to a text channel
+	 *
+	 * @param id - Id of the voice channel
+	 * @param options - Options
+	 * @param options.timeout - Timeout for the command
+	 *
+	 * have explicit permission from the user.
```
Jiralite: Should this be here? It looks awkward!

Jiralite (dismissed):

packages/rpc/LICENSE
Jiralite, line 1: Do we need to do something about the license at the root of the monorepo as this license differs from it?

packages/rpc/examples/boops-javascript/package-lock.json
kyranet, line 1: The example already has a pnpm-lock.yaml, is this lockfile necessary?

packages/rpc/examples/boops-javascript/src/main/index.js
kyranet, line 1: Is boops-javascript needed when we already have a nearly 1:1 TypeScript example?

packages/rpc/src/client.ts
```diff
@@ -0,0 +1,593 @@
...
+		super();
+
+		this.options = options;
+
+		this.accessToken = '';
+		this.clientId = '';
+		this.clientSecret = '';
+		this.ready = false;
+		this.readyAt = new Date();
+		this.readyTimestamp = 0;
```
kyranet: `ready` being `false` and `readyAt` being not only always a non-null value, but also `readonly` and mismatching with `readyTimestamp` is misleading.

A better solution would be setting `readyTimestamp` as `null` and make `readyAt` a getter that constructs a date with the aforementioned value.

packages/rpc/src/client.ts
```diff
@@ -0,0 +1,593 @@
...
+	public application: RPCOAuth2Application | null;
+
+	public user: APIUser | null;
+
+	public transport: IPCTransport;
+
+	/**
+	 * Map of nonces being expected from the transport
+	 */
+	readonly #expected_nonces: Map<
```
kyranet: Field names should be camelCase, so `#expectedNonces`

packages/rpc/src/client.ts
```diff
@@ -0,0 +1,593 @@
...
+	 *
+	 * @param cmd - Command
+	 * @param args - Arguments
+	 * @param evt - Event
+	 */
+	async #request<Cmd extends RPCCallableCommands = RPCCallableCommands>(
+		cmd: Cmd,
+		args?: MappedRPCCommandsArgs[Cmd],
+		evt?: RPCEvents,
+	): Promise<MappedRPCCommandsResultsData[Cmd]>;
```
kyranet: It would be nice to support `options: RequestOptions` being it an interface with `signal?: AbortSignal`, so all endpoints can be called with a `signal` and be able to abort requests early

packages/rpc/src/client.ts
```diff
@@ -0,0 +1,593 @@
...
+
+		this.transport.once('close', onClose);
+
+		const timeout = setTimeout(() => {
+			this.removeListener(RPCEvents.Ready, onReady);
+			this.transport.removeListener('close', onClose);
+
+			reject(new Error('RPC_CONNECTION_TIMEOUT'));
+		}, 10e3);
+		timeout.unref();
```
kyranet: As per other suggestions, this should use an `AbortSignal`. The `.login()` method could also pass its signal to this method so to reuse it. It'll also solve the FP lint issue regarding no-use-before-define

packages/rpc/src/client.ts
```diff
@@ -0,0 +1,593 @@
...
+	 *
+	 * @param event - Name of event e.g. `MESSAGE_CREATE`
+	 * @param args - Args for event e.g. `{ channel_id: '1234' }`
+	 */
+	public async subscribe<Evt extends RPCEvents>(
+		...[event, args]: EventAndArgsParameters<Evt>
+	): Promise<{ unsubscribe(): Promise<RPCUnsubscribeResultData> }> {
+		await this.#request(RPCCommands.Subscribe, args, event);
+		return {
+			unsubscribe: async () => this.#request(RPCCommands.Unsubscribe, args, event),
```
kyranet: This should be available as a standalone method too

packages/rpc/src/ipc.ts
```diff
@@ -0,0 +1,195 @@
...
+		this.emit('open');
+
+		socket.write(
+			encode(OPCodes.Handshake, {
+				v: 1,
+				client_id: this.client.clientId,
+			}),
+		);
+
+		socket.pause();
```
kyranet: What's this pause for?

packages/rpc/src/ipc.ts
```diff
@@ -0,0 +1,195 @@
...
+		const { promise, resolve, reject } = Promise.withResolvers();
+
+		this.once('close', (error) => {
+			// eslint-disable-next-line @typescript-eslint/no-use-before-define
+			timeout.close();
+			resolve(error);
+		});
+
+		const timeout = setTimeout(reject, 10e3);
+		timeout.unref();
```
kyranet: `AbortSignal` support here

packages/rpc/src/util.ts
```diff
@@ -0,0 +1,24 @@
...
+		} catch {
+			return false;
+		}
+	}
+})();
+
+export function getPid() {
+	// eslint-disable-next-line n/prefer-global/process
+	if (typeof globalThis.process !== 'undefined') {
+		// eslint-disable-next-line n/prefer-global/process
```
kyranet, line 20: Wait, does this line actually do something? It's using global process

README.md
kyranet: You'll have to edit those too

kyranet (changes requested):

packages/rpc/examples/boops-javascript/package-lock.json
OfficialSirH: nope, gonna delete that

packages/rpc/examples/boops-javascript/src/main/index.js
OfficialSirH: nah

packages/rpc/src/client.ts
OfficialSirH: I don't remember why I did this a year ago. Ask my past self. (I'll fix this)

packages/rpc/src/ipc.ts
OfficialSirH: I have no clue, was there from the original source code and decided that it must've been important enough so I left it alone.

OfficialSirH: Oh, nevermind, figured it out, it's to enable manual reads as seen in the decode method.

packages/rpc/src/util.ts
OfficialSirH: It does. I genuinely don't understand it either lmao

packages/rpc/src/ipc.ts
kyranet: Can you add a comment or something so it's documented? (Also 5 years younger me would have loved to know this tip)

packages/rpc/examples/boops/src/preload/index.d.ts
kyranet, line 1: This can be moved to the /boops/src/preload/index.ts, it'll also resolve the errors you're needing `@ts-ignore` for.

packages/rpc/examples/boops/src/renderer/src/renderer.ts
```diff
@@ -0,0 +1,24 @@
+function init(): void {
+  window.addEventListener('DOMContentLoaded', () => {
+    const snek = document.querySelector<HTMLDivElement>('#snek')!
+    const counter = document.querySelector<HTMLHeadElement>('#boops')!
+
+    let boops = 0
+
+    function boop(): void {
+      boops += 1
+      counter.innerHTML = `${boops} BOOPS`
```
kyranet: Keeps coderabbitai happy and it's a very small trivial change
```suggestion
      counter.textContent = `${boops} BOOPS`
```

packages/rpc/src/client.ts
```diff
@@ -0,0 +1,667 @@
...
+		const { promise, resolve, reject } = Promise.withResolvers<RPCClient>();
+		const signal = options?.signal ?? AbortSignal.timeout(10e3);
+
+		const onReady = () => {
+			this.readyTimestamp = Date.now();
+			resolve(this);
+		};
+
+		const onClose = (data: unknown) => {
+			for (const exp_nonce of this.#expectedNonces.values()) {
```
kyranet: Another snake_case here

packages/rpc/src/client.ts
```diff
@@ -0,0 +1,667 @@
...
+	 * Request but public
+	 */
+	public async request<Cmd extends RPCCallableCommands>(
+		cmd: Cmd,
+		args: MappedRPCCommandsArgs[Cmd],
+		evt?: RPCEvents,
+		options?: RequestOptions,
+	): Promise<MappedRPCCommandsResultsData[Cmd]> {
+		return this.#request(cmd, args, options, evt);
+	}
```
kyranet: This is working as basically an alias with different argument order. We can move the `#request` method into this one.

packages/rpc/src/client.ts
```diff
@@ -0,0 +1,667 @@
...
+			client_secret: this.clientSecret,
+			code,
+			grant_type: 'authorization_code',
+		};
+
+		if (this.options.redirectUri) {
+			jsonBody.redirect_uri = this.options.redirectUri;
+		}
+
+		const response = (await fetch(`https://discord.com/api${Routes.oauth2TokenExchange()}`, {
```
kyranet: You can use this, will also append the missing explicit API version (`/v10`):
```suggestion
		const response = (await fetch(`${RouteBases.api}${Routes.oauth2TokenExchange()}`, {
```

packages/rpc/src/client.ts
```diff
@@ -0,0 +1,667 @@
...
+		activity.instance = Boolean(activity.instance);
+
+		if (activity.timestamps) {
+			if ('start' in activity.timestamps && activity.timestamps.start > 2_147_483_647_000) {
+				throw new RangeError('timestamps.start must fit into a unix timestamp');
+			}
+
+			if ('end' in activity.timestamps && activity.timestamps.end > 2_147_483_647_000) {
+				throw new RangeError('timestamps.end must fit into a unix timestamp');
+			}
```
kyranet: If we're validating upper ranges, we should validate the lower ranges as well as asserting the number is an integer

packages/rpc/src/ipc.ts
```diff
@@ -0,0 +1,198 @@
...
+
+async function getIPC(id = 0, { promise, resolve, reject } = Promise.withResolvers<Socket>()): Promise<Socket> {
+	const path = await getIPCPath(id);
+
+	const onError = async () => {
+		if (id < 10) {
+			return getIPC(id + 1, { promise, resolve, reject });
+		}
+
+		reject(new Error('Could not connect'));
```
kyranet: I have a concern with the recursion (over using a loop) here, if it retries a few times, this will fill the stack trace with `getIPC` calls, making it slightly harder to trace errors.

packages/rpc/src/ipc.ts
```diff
@@ -0,0 +1,198 @@
...
+		const signal = options?.signal ?? AbortSignal.timeout(10e3);
+
+		this.once('close', (error) => {
+			resolve(error);
+		});
+
+		signal.addEventListener('abort', (_: Event, reason?: string) => {
+			if (this.listenerCount('close') === 0) return;
+			reject(reason);
+		});
```
kyranet: One way to make this more robust this is to declare an `onAbort` method and attach it as a listener for the `signal`'s abort event, then in `this.once('close', ...)` remove said listener. With this you won't need to use the conditional to check if it called close, as it won't fire.

kyranet (changes requested):
The Electron code being completely semi-less is something I'm not very keen of considering it clashes with the style of the rest of the repo, but I also assume that's the "recommended" style from Electron.

Anyways, more requested changes:

packages/rpc/src/ipc.ts
OfficialSirH: I solved the recursion but there's another nuisance now if you'd like to look at it here: 
https://github.com/discordjs/discord.js/pull/11026/changes#diff-14907beb21f0364c1101b8158a70a388d1a27e95cf020cc95dc024b897f7bb0aR94-R103

packages/rpc/src/ipc.ts
```diff
@@ -0,0 +1,201 @@
...
+		let socket: Error | Socket | undefined;
+		for (let id = 0; id < 10; id++) {
+			socket = await getIPC(id).catch((error: Error) => error);
+			if (socket instanceof Socket) break;
+		}
+
+		// NOTE: I wouldn't need to do this if TS' type-checker could understand that socket IS being asigned to before use,
+		// but unfortunately, `undefined` has to be included in the variable's type.
+		// eslint-disable-next-line @typescript-eslint/only-throw-error
+		if (!(socket instanceof Socket)) throw socket;
```
Qjuh:
```suggestion
		let socket: Socket | undefined;
		for (let id = 0; id < 10; id++) {
			try {
				socket = await getIPC(id);
			} catch (error: Error) {
				if (id === 9) throw error;
			}
		}

		// NOTE: I wouldn't need to do this if TS' type-checker could understand that socket IS being asigned to before use,
		// but unfortunately, `undefined` has to be included in the variable's type.
```
and then cast as Socket on assign.

packages/rpc/examples/boops/pnpm-lock.yaml
Qjuh, line 1: Shouldn't this be either handled by top-level lockfile or not have a committed lockfile at all?

packages/rpc/examples/boops/.gitignore
Qjuh, line 1: Already handled by top-level .gitignore

packages/rpc/examples/boops/.editorconfig
Qjuh, line 1: Do we really want/need this in the example repo?

packages/rpc/examples/boops/.eslintignore
Qjuh, line 1: Should also be handled by top-level file imho

packages/rpc/examples/boops/package.json
```diff
@@ -0,0 +1,38 @@
...
+  "name": "boops",
+  "version": "1.0.0",
+  "description": "A minimal Electron application using @discordjs/rpc",
+  "main": "./out/main/index.js",
+  "scripts": {
+    "format": "prettier --write .",
+    "lint": "eslint . --ext .js,.jsx,.cjs,.mjs,.ts,.tsx,.cts,.mts --fix",
+    "typecheck:node": "tsc --noEmit -p tsconfig.node.json --composite false",
+    "typecheck:web": "tsc --noEmit -p tsconfig.web.json --composite false",
+    "typecheck": "npm run typecheck:node && npm run typecheck:web",
```
Qjuh: This example repo is inconsistent. Does it use `pnpm` or `npm`? The lockfile suggests pnpm, the scripts and .npmrc file suggest npm. Please choose one.

.github/workflows/documentation.yml
```diff
@@ -202,6 +202,30 @@ jobs:
           CF_R2_DOCS_BUCKET: ${{ secrets.CF_R2_DOCS_BUCKET }}
         uses: ./main/packages/actions/src/uploadSplitDocumentation
 
+      - name: Move docs to correct directory
```
Qjuh: This was removed on purpose in the monorepo. Please don't add it back.

Qjuh (changes requested):

packages/rpc/examples/boops-javascript/.eslintrc.cjs
OfficialSirH: How would I properly make use of the top-level eslint/neon?

Qjuh: It should already apply. At least if you remove this config. Specific config overrides needed for this example repo would go into the base eslint.config.js in root folder.

packages/rpc/examples/boops/src/main/index.ts
```diff
@@ -0,0 +1,112 @@
...
+  if (process.platform !== 'darwin') {
+    app.quit()
+  }
+})
+
+// In this file you can include the rest of your app"s specific main process
+// code. You can also put them in separate files and require them here.
+
+// Set this to your Client ID.
+const clientId = '1100471145871966258'
```
kyranet: The Electron example should use the base eslint configuration (so at very least the code comes with semicolons), and also replace this with ENV

packages/rpc/src/client.ts
```diff
@@ -0,0 +1,686 @@
...
+
+	public clientSecret: string;
+
+	public readyTimestamp: number | null;
+
+	public application: RPCOAuth2Application | null;
+
+	public user: APIUser | null;
+
+	public transport: IPCTransport;
```
kyranet, line 111: Class fields should be documented

packages/rpc/src/client.ts
```diff
@@ -0,0 +1,686 @@
...
+	}
+
+	public get readyAt() {
+		if (this.readyTimestamp === null) return null;
+		return new Date(this.readyTimestamp);
+	}
+
+	public get ready() {
+		return this.readyTimestamp !== null;
+	}
```
kyranet, line 153: Getters should be documented (check other packages for examples)

packages/rpc/src/client.ts
```diff
@@ -0,0 +1,686 @@
...
+				throw new RangeError('timestamps.start must fit into a unix timestamp');
+			}
+
+			if (
+				'end' in activity.timestamps &&
+				Number.isInteger(activity.timestamps.end) &&
+				Math.min(Math.max(0, activity.timestamps.end), 2_147_483_647_000) !== activity.timestamps.end
+			) {
+				throw new RangeError('timestamps.end must fit into a unix timestamp');
+			}
```
kyranet: If you make an `isValidTimestamp` function, you'll be able to reduce the code to:
```ts
if ('start' in activity.timestamps && !isValidTimestamp(activity.timestamps.start)) {
  throw new RangeError('timestamps.start must fit into a unix timestamp');
}
```
And reduce code duplication with the `end` branch.

By the way it's more efficient for the second comparison to be:
```ts
value >= 0 && value <= 2_147_483_647_000
```

packages/rpc/src/client.ts
```diff
@@ -0,0 +1,686 @@
...
+				this.#expectedNonces.set(nonce, { resolve, reject });
+				signal.addEventListener(
+					'abort',
+					(_: Event) => {
+						this.#expectedNonces.delete(nonce);
+						const reason = signal.reason instanceof Error ? signal.reason : new Error(String(signal.reason));
+						reject(reason);
+					},
+					{ once: true },
+				);
```
kyranet: Even if `{ once: true }` is set, `signal.removeEventListener` should always be called upon resolving. Maybe you could move the definition of the function outside the `new Promise` constructor, and update the code so once the `finally` branch is triggered, `signal.removeEventListener('abort', onAbort);` is called regardless of whether or not the promise resolved or rejected.

If you use `try {} finally {}`, you will probably want to do `return await new Promise(...)` in the `try` branch, and the listener cleanup in the `finally` branch.

packages/rpc/src/client.ts
```diff
@@ -0,0 +1,686 @@
...
+				const nonce = randomUUID();
+				const payload: { args: MappedRPCCommandsArgs[Cmd]; cmd: Cmd; evt?: RPCEvents; nonce: string } = {
+					cmd,
+					args,
+					nonce,
+				};
+				if (cmd === RPCCommands.Subscribe || cmd === RPCCommands.Unsubscribe) {
+					// eslint-disable-next-line @typescript-eslint/no-non-null-asserted-optional-chain
+					payload.evt = options?.evt!;
+				}
```
kyranet: This is a nitpick, but none of this code relies on the promise, maybe it can be extracted to outside the constructor. Or maybe we could even use `Promise.withResolvers` here and do `return await promise;` at the end of the `try` branch.

kyranet (changes requested):
Also regarding @Qjuh's suggestion on packages/rpc/examples/boops/package.json, I think the pnpm option would be better as we're using pnpm everywhere else.

apps/guide/content/docs/rpc/index.mdx
```diff
@@ -0,0 +1,60 @@
+---
+title: Usage
+---
+
+### Legend
+
+- `client` is a placeholder for the `Client` object, such as const `client = new RPCClient({ scopes: [Oauth2Scopes.MessagesRead] });`.
```
Qjuh, line 7:
```suggestion
- `client` is a placeholder for the `Client` object, such as `const client = new RPCClient({ scopes: [Oauth2Scopes.MessagesRead] });`.
```

apps/guide/content/docs/rpc/index.mdx
```diff
@@ -0,0 +1,60 @@
...
+	updatesDoneSinceStart++;
+}, 15_000);
+```
+
+### RPC Events
+
+Listen for messages on the current user's discord client:
+
+<Callout>
+	Listening for messages requires the scope `messages.read` or `Oauth2Scopes.MessagesRead` from `discord-api-types`
```
Qjuh, line 53: The "or" here is irritating, since those are the same value.

packages/rpc/package.json
```diff
@@ -0,0 +1,88 @@
...
+		"eslint-config-neon": "^0.2.7",
+		"eslint-formatter-compact": "^8.40.0",
+		"eslint-formatter-pretty": "^6.0.1",
+		"prettier": "^3.6.2",
+		"tsup": "^8.5.0",
+		"turbo": "^2.5.6",
+		"typescript": "~5.9.2"
+	},
+	"engines": {
+		"node": ">=22.12.0"
```
Qjuh, line 82: all other packages have `24.17.0` as min version. Is there a reason why rpc needs to support an older version?
Also don't need a double build, ESM alone should suffice since require() of ESM is supported in Node v22+. This will be changed for all other packages in #11589 anyway.

packages/rpc/src/client.ts
```diff
@@ -0,0 +1,690 @@
...
+		return promise;
+	}
+
+	/**
+	 * Performs authentication flow. Automatically calls Client#connect if needed.
+	 *
+	 * @param options - login options.
+	 * @param options.clientId - client Id.
+	 * @param options.clientSecret - client Secret (required if scopes are specified in constructor).
+	 * @param options.accessToken - Access Token (skips authorization steps).
```
Qjuh, line 225: This is wrong, because there is a parameter named `options` but it isn't the one you document here. The properties don't need to be documented here at all, since they are documented on `RPCLoginOptions` interface already.

packages/rpc/src/client.ts
```diff
@@ -0,0 +1,690 @@
...
+		);
+	}
+
+	/**
+	 * Move the user to a voice channel
+	 *
+	 * @param id - Id of the voice channel
+	 * @param options - Options
+	 * @param options.timeout - Timeout for the command
+	 * @param options.force - Force this move. This should only be done if you
```
Qjuh, line 495: Ditto

packages/rpc/src/client.ts
```diff
@@ -0,0 +1,690 @@
...
+
+		return this.request(RPCCommands.SelectVoiceChannel, args, options);
+	}
+
+	/**
+	 * Move the user to a text channel
+	 *
+	 * @param id - Id of the text channel
+	 * @param options - Options
+	 * @param options.timeout - Timeout for the command
```
Qjuh, line 516: Ditto

Qjuh (changes requested):

Qjuh:
And the lockfile has issues and needs to be regenerated.

packages/rpc/src/client.ts
OfficialSirH: Eslint doesn't take kindly to me removing the documentation for the properties
image

Qjuh: Yes, because you're now documenting a parameter by a name while it has no name... Remove that last `@param` too. It doesn't provide any useful information anyway.

Qjuh: Alternatively have it take `loginOptions: RPCLoginOptions` and do the destructuring first thing in function body instead. Else this is bad for documentation purposes since it just "hopes" that the first parameter is the one you mean without any way to check. Your previous name `options` showed that issue quite clearly.

packages/rpc/package.json
```diff
@@ -0,0 +1,83 @@
...
+		"release": "cliff-jumper"
+	},
+	"exports": {
+		".": {
+			"import": {
+				"types": "./dist/index.d.ts",
+				"default": "./dist/index.js"
+			}
+		}
+	},
```
Qjuh, line 22:
```suggestion
		".": {
			"types": "./dist/index.d.ts",
			"default": "./dist/index.js"
		}
	},
	"type": "module"
```
if this is going to be esm

Qjuh (approved):
Apart from that LGTM

packages/rpc/package.json
```diff
@@ -0,0 +1,82 @@
...
+		"changelog": "git cliff --prepend ./CHANGELOG.md -u -c ./cliff.toml -r ../../ --include-path 'packages/rpc/*'",
+		"release": "cliff-jumper"
+	},
+	"exports": {
+		".": {
+			"types": "./dist/index.d.ts",
+			"default": "./dist/index.js"
+		}
+	},
+	"type": "module"
```
Qjuh:
```suggestion
	"type": "module",
```
My bad...

kyranet (approved):

sdanialraza (approved):

packages/rpc/src/ipc.ts
```diff
@@ -0,0 +1,228 @@
...
+			try {
+				socket = await getIPC(id);
+			} catch (error: unknown) {
+				if (id === 9) throw error;
+			}
+		}
+
+		// NOTE: I wouldn't need to do this if TS' type-checker could understand that socket IS being asigned to before use,
+		// but unfortunately, `undefined` has to be included in the variable's type.
+		this.socket = socket as Socket;
```
ckohen: You can assign this.socket within the try and then the variable exists entirely within the try block which avoids this ts issue entirely.
Also don't you want to break the for loop once a known good ipc path s found?

ckohen (changes requested):

ckohen (approved):

packages/rpc/src/client.ts
```diff
@@ -0,0 +1,677 @@
...
+
+	/**
+	 * Map of nonces being expected from the transport
+	 */
+	readonly #expectedNonces: Map<
+		string,
+		{ reject(this: void, reason?: unknown): void; resolve(this: void, value: unknown): void }
+	>;
+
+	public constructor(options?: RPCClientOptions);
```
vladfrangu: This is a useless secondary constructor

packages/rpc/src/client.ts
```diff
@@ -0,0 +1,677 @@
...
+
+		if (this.ready) {
+			resolve(this);
+			return promise;
+		}
+
+		const signal = options?.signal ?? AbortSignal.timeout(10e3);
+		// NOTE: I have no clue why it thinks this value is never reassigned
+		// eslint-disable-next-line prefer-const
+		let onAbort: (_: Event) => void;
```
vladfrangu: Define onAbort after onReady and onClose as const and it'll work just fine

vladfrangu (changes requested):

packages/rpc/src/client.ts
vladfrangu: Or even before, it'll work either way

Qjuh: just needs a different `//eslint-disable-next-line` directive to make eslint shut up about use before define

vladfrangu: 🫠

OfficialSirH: Is it fine then if I disable the no-use-before-define rule in this case?

vladfrangu (approved):

iCrawl (approved):

PR #11028, opened 2025-08-04
feat(ContainerBuilder): add isEmpty method

thorminate:
Adds an `isEmpty` method to the `ContainerBuilder` class, which checks if the container has no child components.  
This provides a simple way for developers to verify container state without manually inspecting its contents.

### DIFF
--- a/packages/builders/src/components/v2/Container.ts
+++ b/packages/builders/src/components/v2/Container.ts
@@ -237,6 +237,15 @@ export class ContainerBuilder extends ComponentBuilder<APIContainerComponent> {
 	}
 
 	/**
+	 * Checks if the container is empty
+	 *
+	 * @returns Whether the container is empty
+	 */
+	public isEmpty(): boolean {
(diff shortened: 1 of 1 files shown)


didinele:
Considering this doesn't even serve as a type-guard of some sort, may I ask why you can't just access `<ContainerBuilder>.components.length` yourself? I see no need for this method considering it exposes no new information and doesn't do anything actually complex.

thorminate:
I see your point; my intention was to provide a convenience/QoL helper for readability and to make the intent clearer in code (e.g., `if (<ContainerBuilder>.isEmpty())` reads more naturally than `if (<ContainerBuilder>.components.length === 0))`.
That said, I understand if this doesn’t fit the library’s design philosophy, and I’m fine with closing the PR if you decide its unnecessary.

Jiralite:
Thank you for your efforts, however, we will not be moving forward with this change.

PR #11038, opened 2025-08-11
ci(release): use app user

ckohen:

Still need to set a user...since we have an app now we'll use that instead

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -54,6 +54,11 @@ jobs:
           ref: ${{ inputs.ref || '' }}
           ssh-key: ${{ secrets.DEPLOY_KEY_CI_RELEASE_TAGS }}
 
+      - name: Set Git User and Email
+        run: |
+          git config user.name "github-actions[bot]"
+          git config user.email "41898282+github-actions[bot]@users.noreply.github.com"
+
       - name: Install Node.js v22
         uses: actions/setup-node@v4
         with:
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -48,16 +48,17 @@ jobs:
       TURBO_TEAM: ${{ secrets.TURBO_TEAM }}
     if: github.repository_owner == 'discordjs'
     steps:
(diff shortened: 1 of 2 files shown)


vladfrangu (dismissed):

almeidx (approved):

.github/workflows/release.yml
Jiralite, line 1: Is `contents: write` still needed for permissions since the application has `permission-contents: write` and will be doing the work?

.github/workflows/release.yml
```diff
@@ -48,11 +48,17 @@ jobs:
...
+        id: app-token
+        with:
+          app-id: ${{ vars.DISCORDJS_APP_ID }}
+          private-key: ${{ secrets.DISCORDJS_APP_KEY_RELEASE }}
+          permission-contents: write
+
       - name: Checkout repository
         uses: actions/checkout@v4
         with:
           ref: ${{ inputs.ref || '' }}
```
Jiralite: We could have this use the application token too!

.github/workflows/release.yml
ckohen: Can probably become read as long as provenance doesn't need write

.github/workflows/release.yml
ckohen: Don't know if provenance supports that, we could try though?

Jiralite: We can try!

.github/workflows/release.yml
Jiralite: I believe it does need it, but since `GITHUB_TOKEN` will be that from actions/create-github-app-token, it should be fine since `permission-contents: write` is present.

Just have to be sure the application has those permissions configured here!

.github/workflows/release.yml
ckohen: provenance appears to be unrelated to the token itself, but merely requesting the perm will set the required env vars

vladfrangu (approved):

iCrawl (approved):

PR #11041, opened 2025-08-16
fix(Rest): handle webhook with token routes correctly

daenamnu:
Currently, making requests to any `/webhooks/:id/:token*` route which result in http 401 without the option `{ auth: false }` causes the manager's token to be set to `null`. Additionally, requests without a manager token and without the flag set throw an error.

While using `{ auth: false }` can already prevent this, it would be a useful quality-of-life improvement because:

1. Documentation is vague:
> Alternate authorization data to use for this request only, or `false` to disable the Authorization header

2. Forgetting to pass the flag will no longer bring down the app.

---

**Relevant links, addressing the issue:**
- [Stale Discord Thread](https://discord.com/channels/222078108977594368/1397419784991281182)
- [Stale PR](https://github.com/discordjs/discord.js/pull/10798)

(this is my first ever PR. Please be kind :3)


almeidx:
Superceded by #11066

PR #11049, opened 2025-08-20
chore(ClientEvents): add ClientEvents interface & fix it

Tristan-GPT:

This PR adds ClientEvents to recent versions as it seems to have disappeared.

Minor change but it is very useful for all developers ( for handlers for example. )


almeidx:
It was renamed to `ClientEventTypes`. See #10710

PR #11050, opened 2025-08-21
feat: add `email` and `phoneNumber` formatters

almeidx:
- https://github.com/discord/discord-api-docs/pull/7657

### DIFF
--- a/packages/formatters/src/formatters.ts
+++ b/packages/formatters/src/formatters.ts
@@ -659,6 +659,60 @@ export function applicationDirectory<ApplicationId extends Snowflake, SKUId exte
 }
 
 /**
+ * Formats an email address into an email mention.
+ *
+ * @typeParam Email - This is inferred by the supplied email address
+ * @param email - The email address to format
+ */
+export function email<Email extends string>(email: Email): `<${Email}>`;
+
+/**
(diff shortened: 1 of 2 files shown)


packages/formatters/src/formatters.ts
```diff
@@ -658,6 +658,60 @@ export function applicationDirectory<ApplicationId extends Snowflake, SKUId exte
...
+ * @typeParam PhoneNumber - This is inferred by the supplied phone number
+ * @param phoneNumber - The phone number to format. Must start with a `+` sign.
+ */
+export function phoneNumber<PhoneNumber extends `+${string}`>(phoneNumber: PhoneNumber) {
+	if (!phoneNumber.startsWith('+')) {
+		throw new Error('Phone number must start with a "+" sign.');
+	}
+
+	return `<${phoneNumber}>` as const;
+}
```
Jiralite, line 713: Thoughts on adding the `+` ourselves? Whether that would be requiring users to add it or adding it in case they didn't pass it?

almeidx: That would potentially just make the number invalid, as AFAIK a country code is expected after the `+` symbol.

PR #11051, opened 2025-08-21
chore: bump dependencies

almeidx:


packages/structures/__tests__/invite.test.ts
```diff
@@ -5,20 +5,22 @@ import { Invite } from '../src/index.js';
 import { kPatch } from '../src/utils/symbols.js';
 
 describe('Invite', () => {
-	const dataNoCode: Omit<APIInvite, 'code'> = {
+	// TODO: Check if omitting `expires_at` is appropriate
```
almeidx, line 8: This change has broken these tests:
- https://github.com/discordjs/discord-api-types/pull/1328

Qjuh: Probably should include an expires_at in the test payload instead.

packages/api-extractor-model/package.json
```diff
@@ -36,16 +36,16 @@
...
-		"eslint": "^9.30.1",
+		"eslint": "^9.33.0",
 		"eslint-config-neon": "^0.2.7",
 		"eslint-formatter-compact": "^8.40.0",
 		"eslint-formatter-pretty": "^6.0.1",
 		"prettier": "^3.6.2",
 		"terser": "^5.43.1",
 		"tsup": "^8.5.0",
-		"turbo": "^2.5.4",
+		"turbo": "^2.5.6",
```
Qjuh, line 48: Any reason for not bumping typescript here, in -utils and voice?

almeidx: My filter was too wide for filtering out api-extractor it seems
As for voice, it's stuck on that version due to array buffer changes. Would need to make changes to source

Jiralite (approved):

packages/structures/__tests__/invite.test.ts
almeidx: Including `expires_at: null` fixes the types, but fails at runtime, as `.toJSON()` doesn't include it. At least that's what it appeared to be at a glance, didn't look into it

ckohen: well it's a good thing this test indicates a failure based on that api change then, since it isn't technically properly recreating the payload

almeidx: This can be fixed in a follow-up pull request to /structures

didinele (approved):

Issue #10735, opened 2025-01-30
support e2ee in voice


nyapat:
### Which application or package is this feature request for?

voice

### Feature

> Starting September 2024, Discord is migrating voice and video in DMs, Group DMs, voice channels, and Go Live streams to use end-to-end encryption (E2EE).
> In 2025, all official Discord clients will support the protocol and it will be an enforced requirement to connect to the end-to-end encryption-eligible audio/video session types listed above.
> https://discord.com/developers/docs/change-log#voice-endtoend-encryption-dave-protocol

also up gateway version, gateway v4 is currently used 
https://github.com/discordjs/discord.js/blob/5f463eb9e927232521442b424acbf6f9f4a78540/packages/voice/src/networking/Networking.ts#L303

### Ideal solution or implementation

Use [discord/libdave.js](https://github.com/discord/libdave/tree/main/js)
Implement in /voice following the [whitepaper](https://daveprotocol.com/)

KagChi:
related library: https://github.com/Snazzah/davey

PR #11061, opened 2025-08-22
Voice readme dave protocol

jNullj:

Support for DAVE protocol was added at https://github.com/discordjs/discord.js/issues/10735
Discord mention this in a past blog post [here](https://discord.com/developers/docs/change-log#voice-endtoend-encryption-dave-protocol) and it seems like this is about to be a requirement.
As I discovered while I got repeated crashes due to discord not downgrading my connection to a non e2ee connection!
It could be very useful if i known about this dependency while searching for a solution.
It could be also a good idea to inform users in the release notes/changelog so they know it's a good idea to add this dependency.

Made another PR for the guide as well https://github.com/discordjs/guide/pull/1630

### DIFF
--- a/packages/voice/README.md
+++ b/packages/voice/README.md
@@ -61,10 +61,11 @@ try installing another.
 - `@noble/ciphers`: ^1.0.0
 - `libsodium-wrappers`: ^0.7.9
 
+**DAVE Protocol Libraries (e2ee)**
+
 > [!NOTE]
 > Some Discord clients may require the DAVE protocol for end-to-end encryption in voice chat and refuse to downgrade the connection in the future. Ensure you have `@snazzah/davey` installed to avoid compatibility issues, as `@snazzah/davey` is only set as a dev-dependency for `@discordjs/voice` and may not install in production environments without explicit installation.
 
-**DAVE Protocol Libraries (e2ee)**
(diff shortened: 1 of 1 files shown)


packages/voice/README.md
```diff
@@ -61,6 +61,13 @@ try installing another.
 - `@noble/ciphers`: ^1.0.0
 - `libsodium-wrappers`: ^0.7.9
 
+**DAVE Protocol Libraries (e2ee)**
+
+> [!NOTE]
+> Some Discord clients may require the DAVE protocol for end-to-end encryption in voice chat and refuse to downgrade the connection in the future. Ensure you have `@snazzah/davey` installed to avoid compatibility issues, as `@snazzah/davey` is only set as a dev-dependency for `@discordjs/voice` and may not install in production environments without explicit installation.
```
Jiralite: https://github.com/discordjs/discord.js/pull/11145#discussion_r2411686112

```suggestion
> Some Discord clients may require the DAVE protocol for end-to-end encryption in voice chat and refuse to downgrade the connection in the future. Ensure you have `@snazzah/davey` installed to avoid compatibility issues.
```

jNullj: Applied

PR #11064, opened 2025-08-26
fix(formatters): don't escape * in links

KTibow:
The underscore matcher already has `(?<!<a?:.+|https?:\/\/\S+)`, which prevents escaping underscores in emojis, animated emojis, and links (from the https to the first whitespace character).
This PR makes sure asterisks aren't escaped in links either, fixing #11603.

- Code changes but no Discord API related changes
- Typings don't need updating

### DIFF
--- a/packages/formatters/src/escapers.ts
+++ b/packages/formatters/src/escapers.ts
@@ -212,7 +212,7 @@ export function escapeInlineCode(text: string): string {
  */
 export function escapeItalic(text: string): string {
 	let idx = 0;
-	const newText = text.replaceAll(/(?<=^|[^*])\*([^*]|\*\*|$)/g, (_, match) => {
+	const newText = text.replaceAll(/(?<=^|[^*])(?<!https?:\/\/\S+)\*([^*]|\*\*|$)/g, (_, match) => {
 		if (match === '**') return ++idx % 2 ? `\\*${match}` : `${match}\\*`;
 		return `\\*${match}`;
 	});


KTibow:
Added

Edit: I forgot to test locally, give me a moment to set up my env

KTibow:
Patched

vladfrangu (approved):

Qjuh:
This fixes that but opens another can of worms (which also is an issue with underscore formatter then). The following markdown `<http://*.example.com/globber/*>*Hi*` is valid and renders as 
IMG_6871
But because there is no space between the > and the * your PR would escape it to not be italic Hi anymore.

KTibow:
>  But because there is no space between the > and the * your PR would escape it to not be italic Hi anymore.

Do you mean that this PR would fail to escape the italics, introducing new problematic behavior? That's fair, I'll see what I can do about that.

packages/formatters/src/escapers.ts
```diff
@@ -212,15 +212,21 @@ export function escapeInlineCode(text: string): string {
  */
 export function escapeItalic(text: string): string {
 	let idx = 0;
-	const newText = text.replaceAll(/(?<=^|[^*])\*([^*]|\*\*|$)/g, (_, match) => {
-		if (match === '**') return ++idx % 2 ? `\\*${match}` : `${match}\\*`;
-		return `\\*${match}`;
-	});
+	const newText = text.replaceAll(
+		/(?<=^|[^*])(?<!(?<!<)https?:\/\/\S*|<https?:\/\/[^\s>]*)\*([^*]|\*\*|$)/g,
```
almeidx: Looks like this can actually be any protocol at all. Not limited to http/https.

packages/formatters/src/escapers.ts
```diff
@@ -212,15 +212,21 @@ export function escapeInlineCode(text: string): string {
...
+			return `\\*${match}`;
+		},
+	);
 	idx = 0;
-	return newText.replaceAll(/(?<=^|[^_])(?<!<a?:.+|https?:\/\/\S+)_(?!:\d+>)([^_]|__|$)/g, (_, match) => {
-		if (match === '__') return ++idx % 2 ? `\\_${match}` : `${match}\\_`;
-		return `\\_${match}`;
-	});
+	return newText.replaceAll(
+		/(?<=^|[^_])(?<!<a?:.+|(?<!<)https?:\/\/\S*|<https?:\/\/[^\s>]*)_(?!:\d+>)([^_]|__|$)/g,
```
almeidx: Ditto

almeidx (changes requested):

Qjuh (approved):

PR #11066, opened 2025-08-27
fix(rest): emit warning on HTTP 401 with non-zero code

ckohen:

The risk of actually handling these errors differently is not worth it when the fix is to use the opt-out options we already provide.
All djs libraries already handle this properly and thus it's only direct users and external lib users that encounter this.

This should hopefully be enough information, with both the docs update and the warning, to get these users across the line with an easy fix to their code.

Supersedes #10798 and #11041 

### DIFF
--- a/packages/rest/src/lib/handlers/Shared.ts
+++ b/packages/rest/src/lib/handlers/Shared.ts
@@ -7,6 +7,8 @@ import { RESTEvents } from '../utils/constants.js';
 import type { ResponseLike, HandlerRequestData, RouteData } from '../utils/types.js';
 import { parseResponse, shouldRetry } from '../utils/utils.js';
 
+let authFalseWarningEmitted = false;
+
 /**
  * Invalid request limiting is done on a per-IP basis, not a per-token basis.
  * The best we can do is track invalid counts process-wide (on the theory that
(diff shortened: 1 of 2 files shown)


packages/rest/src/lib/utils/types.ts
```diff
@@ -292,7 +292,9 @@ export interface RequestData {
 	 */
 	appendToFormData?: boolean;
 	/**
-	 * Alternate authorization data to use for this request only, or `false` to disable the Authorization header
+	 * Alternate authorization data to use for this request only, or `false` to disable the Authorization header.
+	 * When making a request to a ropute that includes a token (such as interactions or webhooks), set to `false`
```
almeidx:
```suggestion
	 * When making a request to a route that includes a token (such as interactions or webhooks), set to `false`
```

packages/rest/src/lib/handlers/Shared.ts
```diff
@@ -137,15 +139,29 @@ export async function handleErrors(
...
+			// The request will not succeed for some reason, parse the error returned from the api
+			const data = (await parseResponse(res)) as DiscordErrorData | OAuthErrorData;
+			const isDiscordError = 'code' in data;
+
 			// If we receive this status code, it means the token we had is no longer valid.
 			if (status === 401 && requestData.auth === true) {
+				if (isDiscordError && data.code !== 0 && !authFalseWarningEmitted) {
+					const errorText = `Encountered HTTP 401 with error ${data.code}: ${data.message}. Your token will be removed from this REST instance. If you are using @discordjs/rest directly, consider adding 'auth: false' to the request. Open an issue with your library if not.`;
+					// Use emitWarning if possible, probably not available in edge / web
+					if (typeof process?.emitWarning === 'function') {
```
almeidx: Seems more likely that `process` wouldn't be declared at all rather than being nullish

```suggestion
					if (typeof process !== 'undefined' && typeof process.emitWarning === 'function') {
```

almeidx (dismissed):

packages/rest/src/lib/handlers/Shared.ts
Jiralite: Should this be `globalThis`?

vladfrangu (approved):

Issue #10571, opened 2024-10-21
Adding back-off during retries for REST handler


LJNeon:
### Which application or package is this feature request for?

rest

### Feature

I want to implement exponential back-off when a request is retried due to a 5xx or timeout error. However currently retries are always immediate, so I would need to set `RESTOptions#retries` to 0 and then create a function like this:
```js
await handleRetry(() => client.users.send(/* .. */));
// instead of
await client.users.send(/* .. */);
```
This is far from ideal, and it seems reasonable to implement this in the library instead.

### Ideal solution or implementation

Adding to the `RESTOptions` seems like the obvious choice.

### Other context

This was sparked by a [question](https://discord.com/channels/222078108977594368/1298001168772890738) in the discord server.

imnaiyar:
[Relevant reasoning](https://discord.com/channels/222078108977594368/1298001168772890738/1298012943736111187)

aroman:
strongly agree. we've discovered that sometimes, discord's API calls take 10s+, despite not being ratelimited (i.e. they do not return 429, just take a long time).

our workaround is lowering the `timeout` from 15s (default) to 3s. however, it'd be much nicer to be able to do some sort of exponential backoff.

As for an API, I'd propose that `timeout` could take a `number` or a function, like:

`timeout: (retryCount: number) => number`
so you could let API users implement whatever timeout logic they want

PR #11067, opened 2025-08-28
feat(rest): callbacks for timeout and retry backoff

ckohen:

Adds two related utility options for managing rest requests more dynamically.
Retry Backoff is completely new, allows backing off on 5xx and aborted requests. Closes: #10571 
timeout gets a callback utility, allows setting the timeout per request Closes: #10927 and supersedes #11013


packages/rest/__tests__/utils.test.ts
```diff
@@ -86,3 +89,94 @@ describe('makeURLSearchParams', () => {
 		});
 	});
 });
+
+describe('option normalizaztion functions', () => {
```
Qjuh:
```suggestion
describe('option normalization functions', () => {
```

Jiralite:
Does this address #11000?

ckohen:
No, not at all. That requires a change in discord.js not rest (which I'll get to later today)

packages/rest/src/lib/utils/utils.ts
```diff
@@ -157,3 +163,44 @@ export function normalizeRateLimitOffset(offset: GetRateLimitOffsetFunction | nu
...
+	route: string,
+	statusCode: number | null,
+	retryCount: number,
+	requestBody: unknown,
+): number | null {
+	if (typeof retryBackoff === 'number') {
+		let backoff = Math.max(0, retryBackoff);
+		for (let exponent = 0; exponent < retryCount; exponent++) {
+			backoff += backoff;
+		}
```
kyranet: If my math doesn't fail today, this should be equivalent:
```suggestion
		backoff += backoff * retryCount;
```

ckohen: It's unfortunately not, but there is something that I can do like this

didinele (approved):
you mentioned changing a thing in kyra's thread, other than that, LGTM

SpaceEEC (approved):

PR #10651, opened 2024-12-09
feat: fetch gateway information without requiring rest in ws

vladfrangu:
**Please describe the changes this PR makes and why it should be merged:**

Currently the only way to override the gateway URL that `/ws` will connect to is by intercepting the rest request and overriding it. This PR lets you pick a different ws url, if you may be proxying it or have custom implementations or what not.

**Status and versioning classification:**

- Code changes have been tested against the Discord API, or there are no code changes
- This PR changes the library's interface (methods or parameters added)

PR #11070, opened 2025-08-31
feat(ws): add optional URL property to WebSocketManager options

Digital39999:

Added new optional `url` parameter in `OptionalWebSocketManagerOptions` (`WebSocketManagerOptions`).
When provided, this overrides the default gateway URL returned from `fetchGatewayInformation()`.

This is really needed in order to allow usage of projects such as gateway proxies, instead of requiring developers to hack around the HTTP client response for `/gateway/bot`.

- This PR changes the library's interface (new optional parameter added)

### DIFF
--- a/packages/ws/src/utils/constants.ts
+++ b/packages/ws/src/utils/constants.ts
@@ -49,6 +49,7 @@ export const DefaultWebSocketManagerOptions = {
 		device: DefaultDeviceProperty,
 		os: process.platform,
 	},
+	url: null,
 	version: APIVersion,
 	encoding: Encoding.JSON,
 	compression: null,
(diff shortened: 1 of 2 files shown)


almeidx:
The reasoning behind the #10651 pull request was to enable these kinds of use cases. You can pass a `fetchGatewayInformation` function via the options that will make the /gateway/bot request, where you can also change the resulting gateway `url`.

Digital39999:
Unfortunately `fetchGatewayInformation` isnt typed anywhere, would it better to instead of `url` add `fetchGatewayInformation` to `OptionalWebSocketManagerOptions`?

almeidx:
It is typed in the `RequiredWebSocketManagerOptions` interface

Digital39999:
oh then this is entirely useless, it never showed me that in types, and in djs server question got ignored

PR #11072, opened 2025-09-01
fix(guildChannel): better channel editing

ckohen:

Changes how channel edits are handled to allow sublimits to be handled properly within /rest.
Byproduct is channel editing is much more robust, allows raw api data, and can edit uncached channels now.

### DIFF
--- a/packages/discord.js/src/managers/GuildChannelManager.js
+++ b/packages/discord.js/src/managers/GuildChannelManager.js
@@ -12,6 +12,7 @@ const { ChannelFlagsBitField } = require('../util/ChannelFlagsBitField.js');
 const { transformGuildForumTag, transformGuildDefaultReaction } = require('../util/Channels.js');
 const { ThreadChannelTypes } = require('../util/Constants.js');
 const { resolveImage } = require('../util/DataResolver.js');
+const { toSnakeCase } = require('../util/Transformers.js');
 const { setPosition } = require('../util/Util.js');
 const { CachedManager } = require('./CachedManager.js');
 const { GuildTextThreadManager } = require('./GuildTextThreadManager.js');
@@ -298,7 +299,7 @@ class GuildChannelManager extends CachedManager {
   /**
    * Edits the channel.
    *
-   * @param {GuildChannelResolvable} channel The channel to edit
+   * @param {GuildChannelResolvable} channel The channel to edit. Can be the id of an uncached channel if not editing position or locking permissions
    * @param {GuildChannelEditOptions} options Options for editing the channel
    * @returns {Promise<GuildChannel>}
    * @example
@@ -308,13 +309,17 @@ class GuildChannelManager extends CachedManager {
    *   .catch(console.error);
    */
   async edit(channel, options) {
-    const resolvedChannel = this.resolve(channel);
-    if (!resolvedChannel) throw new DiscordjsTypeError(ErrorCodes.InvalidType, 'channel', 'GuildChannelResolvable');
+    // Let uncached channels be edited if they don't need to lock to a parent without passing parent_id
+    const resolvedChannelId = this.resolveId(channel);
+    if (!resolvedChannelId) throw new DiscordjsTypeError(ErrorCodes.InvalidType, 'channel', 'GuildChannelResolvable');
 
     const parentId = options.parent && this.client.channels.resolveId(options.parent);
 
     if (options.position !== undefined) {
-      await this.setPosition(resolvedChannel, options.position, { position: options.position, reason: options.reason });
+      await this.setPosition(resolvedChannelId, options.position, {
+        position: options.position,
+        reason: options.reason,
+      });
     }
 
     let permission_overwrites = options.permissionOverwrites?.map(overwrite =>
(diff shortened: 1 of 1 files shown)


packages/discord.js/src/managers/GuildChannelManager.js
```diff
@@ -308,13 +309,17 @@ class GuildChannelManager extends CachedManager {
...
 
     if (options.position !== undefined) {
-      await this.setPosition(resolvedChannel, options.position, { position: options.position, reason: options.reason });
+      await this.setPosition(resolvedChannelId, options.position, {
+        position: options.position,
+        reason: options.reason,
+      });
     }
 
     let permission_overwrites = options.permissionOverwrites?.map(overwrite =>
```
SpaceEEC: I suppose we could rename this variable to `permissionOverwrite` (now).
Only tangentially related to this PR however.

SpaceEEC (approved):
I don't know whether I like the setters in the object, but I suppose it works for this use-case here.

kyranet (approved):
Aside from @SpaceEEC's comment, the rest looks good to me

packages/discord.js/src/managers/GuildChannelManager.js
```diff
@@ -308,57 +309,78 @@ class GuildChannelManager extends CachedManager {
...
+      snakeCaseBody.permission_overwrites = permissionOverwrites;
+    }
+
+    if (snakeCaseBody.available_tags) {
+      snakeCaseBody.available_tags = snakeCaseBody.available_tags.map(availableTag =>
+        'emoji' in availableTag ? transformGuildForumTag(availableTag) : availableTag,
+      );
+    }
+
+    if (snakeCaseBody.default_reaction_emoji?.id) {
```
almeidx: Any reason for checking `id` specifically here? Passing `name` wouldn't work this way.
```suggestion
    if (snakeCaseBody.default_reaction_emoji) {
```

If there is a reason for the `id` check then we must also check for `name`.

packages/discord.js/src/managers/GuildChannelManager.js
```diff
@@ -308,57 +309,78 @@ class GuildChannelManager extends CachedManager {
...
+    }
+
+    // This overwrites a passed snake_case version if a camelCase version is passed
+    if (permissionOverwrites) {
+      snakeCaseBody.permission_overwrites = permissionOverwrites;
+    }
+
+    if (snakeCaseBody.available_tags) {
+      snakeCaseBody.available_tags = snakeCaseBody.available_tags.map(availableTag =>
+        'emoji' in availableTag ? transformGuildForumTag(availableTag) : availableTag,
```
almeidx, line 370: While this is not problematic currently, it could potentially be if discord adds a property to available tags that has 2 words. Since `transformGuildForumTag` expects a discord.js object (aka camel cased) and we're passing the snake cased version here. 

Might be worth doing something along the lines of `options.availableTags?.map(x => toSnakeCase(blah...)) ?? options.available_tags` instead to prevent future headaches.

packages/discord.js/src/managers/GuildChannelManager.js
```diff
@@ -308,57 +309,78 @@ class GuildChannelManager extends CachedManager {
...
+      );
+    }
+
+    if (snakeCaseBody.default_reaction_emoji?.id) {
+      snakeCaseBody.default_reaction_emoji = transformGuildDefaultReaction(snakeCaseBody.default_reaction_emoji);
```
almeidx: The talk for `available_tags` also applies here technically, but is even more unlikely to cause issues.

packages/discord.js/src/managers/GuildChannelManager.js
```diff
@@ -308,57 +309,78 @@ class GuildChannelManager extends CachedManager {
...
+    }
+
+    if (snakeCaseBody.flags) {
+      snakeCaseBody.flags = ChannelFlagsBitField.resolve(snakeCaseBody.flags);
+    }
```
almeidx: Ditto. I don't know if this one would cause issues but I also wouldn't want to find out.

```suggestion
    if (options.flags) {
      snakeCaseBody.flags = ChannelFlagsBitField.resolve(options.flags);
    }
```

almeidx (dismissed):

packages/discord.js/src/managers/GuildChannelManager.js
didinele: sorry, I could be completely off base here since I haven't read everything/the types, but wouldn't your diff cause trouble if `{}` was passed?

packages/discord.js/src/managers/GuildChannelManager.js
ckohen: emoji is only a property if it's a djs object, we only call transformGuildForumTag if the caller is passing non-api data (it's emoji_id and emoji_name to discord)

packages/discord.js/src/managers/GuildChannelManager.js
ckohen: good callout, I forgot emoji's can be name only. I'll add || name to it.

and @didinele no, {} would pass through in the body. Though speaking of types, I didn't add a | RESTPatchAPIChannelJSONBody to the types for this method, but I theoretically could at this point....should I?

packages/discord.js/src/managers/GuildChannelManager.js
ckohen: same thing, never done with a raw api object

packages/discord.js/src/managers/GuildChannelManager.js
ckohen: the only valid value to pass as raw api data is properly resolved by the resolve function, I could realistically do either, want opinions

packages/discord.js/src/managers/GuildChannelManager.js
almeidx: Is that the direction we foresee the discord.js library going in (i.e., allowing passing raw payloads everywhere appropriate)?

ckohen: I think we should allow it (to better support new features that don't have real support yet), but I don't know that we should actually document it that way. It might be something we mention somewhere once, but I don't think we should specifically call it out except *maybe* types...my opinion will probably change once the lib is actually ts native and the underlying support for raw data is way better.

packages/discord.js/src/managers/GuildChannelManager.js
Qjuh: That doesn't change the fact that this will fail as soon as discord (and subsequently we) introduce a two-word property in tags, because that will be snake_cased at this point.

packages/discord.js/src/managers/GuildChannelManager.js
Qjuh: I'm also for using the options, not the snake_cased value here. And the JSDoc and typings currently only accept camelCased data, so I don't see the reasoning for accounting for raw data

packages/discord.js/src/managers/GuildChannelManager.js
ckohen: There are already two-word properties in tags....emoji_id and emoji_name, and that's literally the entire point of the transformGuildForumTag util

packages/discord.js/src/managers/GuildChannelManager.js
ckohen: see https://github.com/discordjs/discord.js/pull/11072#discussion_r2347635732 for why accepting raw data is the choice I made and why it's not *currently* documented that way

packages/discord.js/src/managers/GuildChannelManager.js
Qjuh: Ah, I see. My bad, ignore what I said, guess this conversation can be resolved then.

packages/discord.js/src/managers/GuildChannelManager.js
Qjuh: Fair enough. My first statement still stands, ignore the second.

packages/discord.js/src/managers/GuildChannelManager.js
almeidx: > There are already two-word properties in tags....emoji_id and emoji_name, and that's literally the entire point of the transformGuildForumTag util

No, there isn't because we didn't make it a two word property. It just so happens that specific case would work.

The problem is the input to `transformGuildForumTag`, not the output of it. Again... we're making that function with the assumption that a djs-like value is passed in, but now we're always passing in a snake cased value to it here. Like, `two_word: tag.twoWord` would not work anymore with this change

ckohen: > but now we're always passing in a snake cased value to it here

This is wrong, we only pass a value to that function if it contains emoji in it, which means it's a djs-like value

packages/discord.js/src/managers/GuildChannelManager.js
ckohen: I think I'll just go for options then

packages/discord.js/src/managers/GuildChannelManager.js
almeidx: It's a djs-like value, except it has been snake cased. Making it not a djs-like value if there were any two word props

ckohen: ah, i see what you mean, good catch

packages/discord.js/src/managers/GuildChannelManager.js
```diff
@@ -308,57 +309,78 @@ class GuildChannelManager extends CachedManager {
...
+      // Remove properties we don't want to pass to the API in the body
+      {
+        set parent(_) {},
+        set position(_) {},
+        set reason(_) {},
       },
+      toSnakeCase(options),
+    );
+
+    // This overwrites a passed snake_case version if a camelCase version OR `parent` is passed
```
almeidx, line 358: BTW, seems that in `toSnakeCase()`, if there are conflicting keys, the last one (by key order) takes precedence. i.e., if both snake case and camel case versions of a key are passed, the later one overwrites the earlier.
This won't be the case for these ones, where camel case will take precedence. Not sure if we should do something about it (maybe tweaking toSnakeCase to prefer the camel case version?..) or just let it be

ckohen: I think I had set a precedent somewhere to prefer camel case when passsed which is why I did it this way.

I'm not exactly sure if I want to change that function in this PR, but I think it would make sense to because it would be more deterministic that way.

vladfrangu: I thought the snake case had priority, not the camel case

ckohen: It doesn't currently (in the existing cases)

ckohen:
they're not *always* cached, just in 99.99% of cases

PR #11089, opened 2025-09-04
feat(GuildMemberManager)!: Add new modify self fields

Jiralite:
Adds new fields for modifying the current guild member.

Adding these into the existing edit method is problematic as this implies one may use these on any guild member. The method has been split up into 2: one for you, one for others!

This also means if you want to change your nickname without needing manage nicknames, `editMe()` should be used.

Upstream:

- https://github.com/discord/discord-api-docs/pull/7807


vakiliner:
Will the edit() method be removed?
Should it be left as deprecated?

tipakA:
@vakiliner Why would it be removed? How would you edit members without it?

vakiliner:
> @vakiliner Why would it be removed? How would you edit members without it?

Sorry, I meant that it was already removed, or rather it was renamed to editMember()

monbrey:
I'm conflicted because I hate inconsistency in the interface. I was going to point out that we don't do this for the `User` class before remembering we have a whole extended `ClientUser` to achieve this.

Not sure if `ClientGuildMember` makes sense to create and manage though.

Jiralite:
Per https://github.com/discord/discord-api-docs/pull/7807#issuecomment-3271904854, the response of this is not changing. Therefore, there are no new properties for the guild member object, so the hypothetical `ClientGuildMember` would not have anything over `GuildMember`.

You could also argue that almost every route in managers (if not, all else) are using a single route and this one is actually 2 routes in 1, so it's becoming consistent to split them out.

packages/discord.js/src/managers/GuildMemberManager.js
```diff
@@ -396,18 +396,39 @@ class GuildMemberManager extends CachedManager {
...
+   * Edits the current application's guild member in a guild.
+   *
+   * @param {GuildMemberEditCurrentOptions} options The options to provide
+   * @returns {Promise<GuildMember>}
+   */
+  async editCurrent({ reason, ...options }) {
+    const data = await this.client.rest.patch(Routes.guildMember(this.guild.id, '@me'), {
+      body: {
+        ...options,
+        icon: options.icon && (await resolveImage(options.icon)),
```
GoldenAngel2: Shouldn't this be `avatar` instead of `icon`? 
The pending Discord API docs points to it being `avatar`: 
https://github.com/discord/discord-api-docs/pull/7807/files#diff-b048a45b363d01dbbaeaf7409be45b8faec5386ce0cc3e889208c737c534855fR1022

packages/discord.js/typings/index.d.ts
```diff
@@ -6178,6 +6179,13 @@ export interface GuildMemberEditOptions {
   roles?: ReadonlyCollection<Snowflake, Role> | readonly RoleResolvable[];
 }
 
+export interface GuildMemberEditCurrentMemberOptions {
+  banner?: Base64Resolvable | BufferResolvable | null;
+  icon?: Base64Resolvable | BufferResolvable | null;
```
GoldenAngel2: Same as the previous comment

vakiliner:
Should the editCurrent() method be named to editMe()?
_Referring to the fetchMe() method_

packages/discord.js/src/managers/GuildMemberManager.js
```diff
@@ -358,14 +359,13 @@ class GuildMemberManager extends CachedManager {
...
-   * <info>The user must be a member of the guild</info>
+   * Edits a member of a guild.
    *
    * @param {UserResolvable} user The member to edit
-   * @param {GuildMemberEditOptions} options The options to provide
+   * @param {GuildMemberEditMemberOptions} options The options to provide
    * @returns {Promise<GuildMember>}
    */
-  async edit(user, { reason, ...options }) {
+  async editMember(user, { reason, ...options }) {
```
Qjuh: Wouldn't it be more in line with current naming to keep this as `edit()` and name the other method `editMe()`?

packages/discord.js/src/managers/GuildMemberManager.js
```diff
@@ -396,18 +396,40 @@ class GuildMemberManager extends CachedManager {
...
+   */
 
-    const data = await this.client.rest.patch(endpoint, { body: options, reason });
+  /**
+   * Edits the current application's guild member in a guild.
+   *
+   * @param {GuildMemberEditMeOptions} options The options to provide
+   * @returns {Promise<GuildMember>}
+   */
+  async editMe({ reason, ...options }) {
```
vladfrangu, line 422: I don't get the purpose of this function. Doesn't the normal edit function support all these fields too? Why do you need to call editMe to edit some things and then edit for others (like wanting to change nickname _and_ channel in one call [if that actually works])

GoldenAngel2: The normal edit function doesn't support these fields.
`/guilds/:id/members/:id` doesn't support `avatar, bio, banner`, bots can only change their own (hence why this function uses `/guilds/:id/members/@me` endpoint)
the `/members/@me` endpoint only supports what is shown in the `GuildMemberEditMeOptions` interface, if you want to edit the channel_id or anything else that is part of the `/members/:id` endpoint you would have to use the `edit` function. (don't know why Discord didn't add support for the other fields in the `/members/@me` endpoint)

vakiliner: > I don't get the purpose of this function. Doesn't the normal edit function support all these fields too? Why do you need to call editMe to edit some things and then edit for others (like wanting to change nickname _and_ channel in one call [if that actually works])

The endpoint `/guilds/{guild.id}/members/{user.id}` does not have the `avatar`, `banner`, and `bio` fields. They are only found in `/guilds/{guild.id}/members/@me`
You can't change, say, `avatar` and `channel_id` in a single call
And by adding this to edit method, developers might get confused and think that these fields can be edited for not-self members

vladfrangu: Lovely. So the `/@me` route doesn't support things that make sense 🙃.

vakiliner: And in the editMe method, you don't need to enter your app ID

SpaceEEC (approved):

sdanialraza (approved):

PR #11099, opened 2025-09-10
fix(users): `RESTPatchAPICurrentGuildMemberJSONBody` for editing current guild member

Jiralite:
`RESTPatchAPICurrentGuildMemberJSONBody` is the correct type for this method. The route is using `@me`.

### DIFF
--- a/packages/core/src/api/user.ts
+++ b/packages/core/src/api/user.ts
@@ -1,7 +1,8 @@
 /* eslint-disable jsdoc/check-param-names */
 
 import { makeURLSearchParams, type RequestData, type REST } from '@discordjs/rest';
-import {
+import type {
+	RESTPatchAPICurrentGuildMemberJSONBody,
 	Routes,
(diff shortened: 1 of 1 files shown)


vladfrangu (approved):

packages/core/src/api/user.ts
```diff
@@ -111,7 +111,7 @@ export class UsersAPI {
 	 */
 	public async editCurrentGuildMember(
 		guildId: Snowflake,
-		body: RESTPatchAPIGuildMemberJSONBody = {},
+		body: RESTPatchAPICurrentGuildMemberJSONBody = {},
```
almeidx, line 114: Is there no matching return type in discord-api-types for this?

Jiralite: This does not return anything different, does it?

Jiralite: https://github.com/discordjs/discord-api-types/issues/1367

PR #11108, opened 2025-09-17
feat: File upload support

### LINKED ISSUE #11106: File upload support for modals for @discordks/builders
- https://canary.discord.com/channels/1317206872763404478/1317206872763404481/1417987236267561132

Jiralite:
Resolves #11106.

### DIFF
--- a/apps/guide/content/docs/legacy/index.mdx
+++ b/apps/guide/content/docs/legacy/index.mdx
@@ -11,7 +11,7 @@ This guide will teach you things such as:
 
 - How to get a bot [up and running](./legacy/preparations/app-setup) from scratch;
 - How to properly [create](./legacy/app-creation/project-setup), [organize](./legacy/app-creation/handling-commands), and expand on your commands;
-- In-depth explanations and examples regarding popular topics (e.g. [components](./legacy/popular-topics/display-components) ,[reactions](./legacy/popular-topics/reactions), [embeds](./legacy/popular-topics/embeds), [canvas](./legacy/popular-topics/canvas));
+- In-depth explanations and examples regarding popular topics (e.g. [components](./legacy/popular-topics/display-components), [reactions](./legacy/popular-topics/reactions), [embeds](./legacy/popular-topics/embeds), [canvas](./legacy/popular-topics/canvas));
 - Working with databases (e.g. [sequelize](./legacy/sequelize/) and [keyv](./legacy/keyv/keyv));
 - Getting started with [sharding](./legacy/sharding/);
 - And much more.
--- a/apps/guide/content/docs/legacy/popular-topics/meta.json
+++ b/apps/guide/content/docs/legacy/popular-topics/meta.json
@@ -1,5 +1,21 @@
 {
 	"title": "Popular Topics",
 	"defaultOpen": true,
-	"pages": ["!faq", "...", "!display-components"]
+	"pages": [
+		"!faq",
+		"!display-components",
+		"audit-logs",
+		"canvas",
+		"collectors",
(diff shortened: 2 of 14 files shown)


packages/builders/src/components/Components.ts
```diff
@@ -162,6 +167,10 @@ export interface MappedComponentTypes {
 	 * The label component type is associated with a {@link LabelBuilder}.
 	 */
 	[ComponentType.Label]: LabelBuilder;
+	/**
+	 * The label component type is associated with a {@link FileUploadBuilder}.
```
advaith1:
```suggestion
	 * The file upload component type is associated with a {@link FileUploadBuilder}.
```

packages/builders/src/components/fileUpload/FileUpload.ts
```diff
@@ -0,0 +1,107 @@
...
+	 *
+	 * @param data - The API data to create this file upload with
+	 * @example
+	 * Creating a file upload from an API data object:
+	 * ```ts
+	 * const fileUpload = new FileUploadBuilder({
+	 * 	custom_id: "file_upload",
+	 *  min_values: 2,
+	 * });
+	 * ```
```
advaith1, line 28: This example actually isn't valid because `max_values` defaults to 1. min_values must always be <= max_values.

Jiralite: Thank you!

packages/builders/src/components/fileUpload/FileUpload.ts
```diff
@@ -0,0 +1,109 @@
...
+	}
+
+	/**
+	 * Clears the minimum values.
+	 */
+	public clearMinValues() {
+		this.data.min_values = undefined;
+		return this;
+	}
+
```
Qjuh, line 71:
```suggestion
```

I don't think this should be here. If you want to revert your change to minValues it's simply `setMinValues(1)` and should be explicit.

packages/builders/src/components/fileUpload/FileUpload.ts
```diff
@@ -0,0 +1,109 @@
...
+	/**
+	 * Clears the maximum values.
+	 */
+	public clearMaxValues() {
+		this.data.max_values = undefined;
+		return this;
+	}
+
```
Qjuh, line 89:
```suggestion
```

Ditto here. We also don't do this on SelectMenuBuilders.

Qjuh (changes requested):
Apart from my comments LGTM

packages/builders/src/components/fileUpload/FileUpload.ts
vladfrangu: We have clear methods in other builders for similar things, no?

vladfrangu (approved):

packages/builders/src/components/fileUpload/FileUpload.ts
Qjuh: See my other comment:
> We also don't do this on SelectMenuBuilders.

so no, we don't. We have it for things where there's an actual difference between providing a specific value or no value at all (`null`/`undefined`).

Jiralite: Text inputs can be used in modals and they may clear their minimum and maximum length. Should that be removed then?

vladfrangu: No, we should add clear methods to everything that can be undefined 🤷

Qjuh (approved):

PR #11120, opened 2025-09-28
ci: implement workflow to publish dev versions

didinele:
The last piece of the puzzle for our GitHub bot.

### DIFF
--- a/.github/workflows/publish-dev.yml
+++ b/.github/workflows/publish-dev.yml
@@ -32,7 +32,6 @@ jobs:
         with:
           app-id: ${{ vars.DISCORDJS_APP_ID }}
           private-key: ${{ secrets.DISCORDJS_APP_KEY_RELEASE }}
-          permission-contents: write
 
       - name: Decide ref
         id: ref


.github/workflows/release-tagged-dev.yml
```diff
@@ -0,0 +1,66 @@
...
+      ref:
+        description: 'The branch, tag or SHA to checkout'
+        required: true
+        default: 'main'
+      package:
+        description: 'Comma separated list of packages to release'
+        required: true
+        type: string
+      tag:
+        description: 'The tag to use, generally a feature name'
```
vladfrangu: I'd argue generally a pull request number (and maybe we should even assert it but that's not a blocker)

didinele: I guess my main issue with that is like, I don't think I can use the PR number as a ref and resolving it in that CI context seems.. hell-ish? The workflow doesn't really need to be the most human friendly given the goal is to call it via bot.

And also, there's maybe some advantages to the general `tag` when we're trying to test something special and maybe don't even have a PR.

vladfrangu: And I'm guessing the bot will just use `pr-<num>` as the tag then?

Either way its fine, I'm not too fussed about this, maybe just for us we should use some tags we can easily reference back

Oh, one thing to add: a flow to deprecate all releases with a tag, so once we're done they can be marked as deprecated on npms end 🙏

didinele: > And I'm guessing the bot will just use pr-<num> as the tag then?

I more-so envisioned the tag as an input so we make sure it's coherent & easy to understand what you're actually using in logs that reference `VERSION` or package.json info, e.g. `@bot pack this new-ws-compression`, but I'm OK with PR num too.

> Oh, one thing to add: a flow to deprecate all releases with a tag, so once we're done they can be marked as deprecated on npms end 🙏

Errrr, right. I'm not really sure when that would run, though? On PR close I suppose would be good? (i.e. assuming the tags aren't `pr-<num>` like I was saying above, scan if the PR triggered any packs & with what tag(s), deprecate everything on NPM with that tag)

vladfrangu: It doesn't have to be automatic necessarily/if its hard to do, but sure, whatever works and lets us mark a tag (that is not dev or latest lol) as deprecated

packages/actions/src/releasePackages/releasePackage.ts
```diff
@@ -76,7 +76,7 @@ export async function releasePackage(release: ReleaseEntry, dev: boolean, dry: b
 		}, 15_000);
 	});
 
-	if (dev) {
+	if (devTag) {
 		// Send and forget, deprecations are less important than releasing other dev versions and can be done manually
 		void $`pnpm exec npm-deprecate --name "*dev*" --message "This version is deprecated. Please use a newer version." --package ${release.name}`
```
ckohen: I think you want to use the tag name in the deprecate name

.github/workflows/release-tagged-dev.yml
ckohen: I would argue that the tag name should be something readable, and we pass a ref as well (which can absolutely just be based on the pr number)

ckohen:
Overall looks pretty good, I really don't think this is a separate workflow though. The only reason normal dev releases are is to run on cron.

.github/workflows/release-tagged-dev.yml
```diff
@@ -0,0 +1,66 @@
...
+          permission-contents: write
+
+      - name: Checkout repository
+        uses: actions/checkout@v5
+        with:
+          token: ${{ steps.app-token.outputs.token }}
+          ref: ${{ inputs.ref || '' }}
+
+      - name: Install Node.js v22
+        uses: actions/setup-node@v4
```
Jiralite: Might as well use v5 here?

Jiralite (approved):

iCrawl (approved):

.github/workflows/publish-dev.yml
```diff
@@ -4,6 +4,14 @@ on:
     - cron: '0 */12 * * *'
   workflow_dispatch:
     inputs:
+      pull:
```
vladfrangu, line 7: I would've called this ref, not pull, but sure

.github/workflows/publish-dev.yml
```diff
@@ -19,10 +27,28 @@ jobs:
...
+          permission-contents: write
+
+      - name: Decide ref
+        id: ref
+        run: |
+          if [ -n "${{ github.event.inputs.pull }}" ]; then
+            echo "ref=refs/pull/${{ github.event.inputs.pull }}/head" >> $GITHUB_OUTPUT
+          else
+            echo "ref=refs/heads/main" >> $GITHUB_OUTPUT
+          fi
```
vladfrangu, line 43: But this forces the ref to be a PR, so which is it 👀

.github/workflows/publish-dev.yml
Jiralite: Actually, I agree with that. The description matches that better (tag + SHA).

didinele: I forgot to update the name when taking it from elsewhere, it is a pull number, not an arbitrary ref.

.github/workflows/publish-dev.yml
```diff
@@ -19,10 +27,28 @@ jobs:
       TURBO_TEAM: ${{ secrets.TURBO_TEAM }}
     if: github.repository_owner == 'discordjs'
     steps:
+      - uses: actions/create-github-app-token@v2
+        id: app-token
+        with:
+          app-id: ${{ vars.DISCORDJS_APP_ID }}
+          private-key: ${{ secrets.DISCORDJS_APP_KEY_RELEASE }}
+          permission-contents: write
```
ckohen: don't need this perm

PR #11121, opened 2025-09-29
types(ClientEventTypes): use Events as keys

Pavel-Boyazov:
Allow to use native Events as key for ClientEventTypes (for examle `ClientEventTypes[Events]` instead of `ClientEventTypes[keyof ClientEventTypes]`)


Jiralite (dismissed):

Jiralite (changes requested):
index.test-d.ts is full of errors due to this change.

It seems this stops strings from being used. Unsure if this is desired. @vladfrangu, any ideas?

Pavel-Boyazov:
I agree that removing the ability to use strings directly is a bad idea. But I would also like to add compatibility of the type with the Events enum, since in fact they are completely identical.

almeidx (approved):

Pavel-Boyazov:
As I see it, it comes down to the [async_event_emitter](https://github.com/vladfrangu/async_event_emitter)'s typing. I would like to understand more precisely what can and should be changed there in order to keep support for raw strings

Pavel-Boyazov:
Another problem. `voiceServerUpdate` event not exists in `ClientEventTypes `

vladfrangu:
> Unsure if this is desired

I wouldn't want this to break normal strings, so if it does... I am -1 on this

Qjuh:
The issues that caused this to be made has been fixed by #11414 so this is no longer relevant.

PR #11123, opened 2025-09-29
types(Message): specify `rawData` arg type

Pavel-Boyazov:
Function allow usage without rawData arg and has type, but in d.ts file it required and with unknown type

### DIFF
--- a/packages/discord.js/src/structures/Message.js
+++ b/packages/discord.js/src/structures/Message.js
@@ -1106,7 +1106,7 @@ class Message extends Base {
    * method allows you to see if there are differences in content, embeds, attachments, nonce and tts properties.
    *
    * @param {Message} message The message to compare it to
-   * @param {APIMessage} rawData Raw data passed through the WebSocket about this message
+   * @param {APIMessage} [rawData] Raw data passed through the WebSocket about this message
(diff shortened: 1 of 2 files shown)


packages/discord.js/src/structures/Message.js
```diff
@@ -1106,7 +1106,7 @@ class Message extends Base {
    * method allows you to see if there are differences in content, embeds, attachments, nonce and tts properties.
    *
    * @param {Message} message The message to compare it to
-   * @param {APIMessage} rawData Raw data passed through the WebSocket about this message
+   * @param {APIMessage} [rawData] Raw data passed through the WebSocket about this message
```
Jiralite, line 1112: Why was this made optional?

Pavel-Boyazov: Based on the function body, this parameter may not exist. See line 1128
If make param not optional, this condition have no sence because every object returns true in it

Pavel-Boyazov:
I looked again at the code executed under the condition. What's the point of `rawData` anyway? The first condition does not require this parameter at all, the second and third can be replaced by the properties of the `Message` object

vladfrangu (approved):

PR #11124, opened 2025-09-30
feat(builders): refinements

didinele:
Got the idea from a random convo I saw, I'd say it's a valuable addition. Method name inspired from zod, we can change if anyone has a better idea

### DIFF
--- a/packages/builders/src/mixins/Refineable.ts
+++ b/packages/builders/src/mixins/Refineable.ts
@@ -1,0 +1,26 @@
+/**
+ * Mixin used to provide {@link Refineable.refine}
+ */
+export class Refineable {
+	/**
+	 * Refines this builder by applying the provided function to itself.
+	 * Useful for conditionally modifying the builder without breaking the method chain, while also isolating
+	 * the scope of temporary variables.
+	 *
+	 * @example
+	 * ```ts
+	 * const builder = new EmbedBuilder()
+	 *  .setTitle('Hello World')
+	 *  .setDescription('This is a description')
+	 *  .refine((b) => {
+	 *      if (externalVariable) {
+	 *         b.setColor('Red');
+	 *     }
+	 *  });
+	 * ```
+	 */
+	public refine(fn: (builder: this) => void): this {
+		fn(this);
+		return this;
+	}
+}
--- a/packages/builders/src/components/Component.ts
+++ b/packages/builders/src/components/Component.ts
@@ -1,5 +1,6 @@
 import type { JSONEncodable } from '@discordjs/util';
 import type { APIBaseComponent, ComponentType } from 'discord-api-types/v10';
+import { Refineable } from '../mixins/Refineable.js';
 
 export interface ComponentBuilderBaseData {
 	id?: number | undefined;
@@ -11,6 +12,7 @@ export interface ComponentBuilderBaseData {
  * @typeParam Component - The type of API data that is stored within the builder
  */
 export abstract class ComponentBuilder<Component extends APIBaseComponent<ComponentType>>
+	extends Refineable
 	implements JSONEncodable<Component>
 {
 	/**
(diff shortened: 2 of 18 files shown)


packages/builders/src/mixins/Refineable.ts
```diff
@@ -0,0 +1,26 @@
...
+	 * the scope of temporary variables.
+	 *
+	 * @example
+	 * ```ts
+	 * const builder = new EmbedBuilder()
+	 *  .setTitle('Hello World')
+	 *  .setDescription('This is a description')
+	 *  .refine((b) => {
+	 *      if (externalVariable) {
+	 *         b.setColor('Red');
```
Jiralite: This should take a number, unless this is showcasing something?

Jiralite (commented):
> I'd say it's a valuable addition.

Maybe add a test case of a simple & useful scenario? That could be nice to see! Reason I ask is because the documentation given for refining doesn't seem to solve a problem?

packages/builders/src/mixins/Refineable.ts
didinele: oh yeah. I hit tab on that completion without thinking about it.

didinele:
> Maybe add a test case of a simple & useful scenario? That could be nice to see! Reason I ask is because the documentation given for refining doesn't seem to solve a problem?

Will do. But the idea is that when you have a large block of logic that does a lot of conditional work on a builder, it can feel a bit "disconnected" and other logic can sneak in the middle, you also can end up introducing a lot of variables.

With `refine` you can get rid of a lot of that:
```ts
<TextBasedChannel>.send({
  embeds: [new EmbedBuilder().refine(builder => { /* ... complex logic here */ });
});
```

faceboy1392:
> > I'd say it's a valuable addition.
> 
> Maybe add a test case of a simple & useful scenario? That could be nice to see! Reason I ask is because the documentation given for refining doesn't seem to solve a problem?

I proposed this idea originally in that "random convo", (and I think I first saw the idea somewhere else). The main reason is I just otherwise find it pretty inconvenient when I'm working with a long chain of builder method calls.

In my experience, it's sometimes annoying to conditionally modify the builder with external if statements and all that, rather than being able to put everything in one long expression. Especially when dealing with components nested inside other components/objects, and typescript gets mad at me without a bunch of type guards. I just much prefer being able to throw a code block in the middle of it rather than dealing with a bunch of variables, which feels more cluttered

idk if I'm making sense. I made my own custom container builder and added this function and I like it, and I wish djs had it (and I'm annoyed javascript has no way to do this sort of thing natively).

packages/builders/src/mixins/Refineable.ts
```diff
@@ -0,0 +1,26 @@
+/**
+ * Mixin used to provide {@link Refineable.refine}
```
vladfrangu, line 2: But this is not applied as a mixin but as a class extension

didinele: There's no difference between `class X extends Refineable` and `class X extends Mixin(Refineable)`... it's applied as a mixin where multi-inheritance is needed, there's other mixins we do this with throughout the codebase

packages/builders/src/mixins/Refineable.ts
```diff
@@ -0,0 +1,28 @@
...
+	 * the scope of temporary variables.
+	 *
+	 * @example
+	 * ```ts
+	 * const builder = new EmbedBuilder()
+	 *  .setTitle('Hello World')
+	 *  .setDescription('This is a description')
+	 *  .refine((b) => {
+	 * 		const intermediateValue = computeSomething();
+	 * 		b.setFooter(`Computed value: ${intermediateValue}`);
```
Jiralite, line 17: This is an object!

packages/builders/src/mixins/Refineable.ts
Qjuh: Got any example of those "other mixins" we do that with? Because to me that is wrong too. Calling a base class a mixin I mean

PR #11125, opened 2025-10-01
ci(release): use main action always

ckohen:

Makes the Release CI always use the main workflow which allows us to release from branches that don't have the workflow. Doing this in the dev workflow adds an extra layer of security, that will be done in or after #11120

(yes, no build step for the action is correct, bun negates that)

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -81,13 +81,12 @@ jobs:
       - name: Install action deps (non-main ref only)
         if: ${{ inputs.ref && inputs.ref != 'main' }}
         shell: bash
+        working-directory: ./main
         env:
           COREPACK_ENABLE_STRICT: 0
         run: |
-          cd main
           pnpm self-update 10
           pnpm install --filter @discordjs/actions --frozen-lockfile --prefer-offline --loglevel error
-          cd ..
 
       - name: Release packages (non-main ref)
         if: ${{ inputs.ref && inputs.ref != 'main' }}
@@ -96,7 +95,7 @@ jobs:
           package: ${{ inputs.package }}
           exclude: ${{ inputs.exclude }}
           dry: ${{ inputs.dry_run }}
-          main: true
+          main: false
         env:
           NODE_AUTH_TOKEN: ${{ secrets.NPM_PUBLISH_TOKEN }}
           GITHUB_TOKEN: ${{ steps.app-token.outputs.token }}
@@ -108,7 +107,7 @@ jobs:
           package: ${{ inputs.package }}
           exclude: ${{ inputs.exclude }}
           dry: ${{ inputs.dry_run }}
-          main: false
+          main: true
         env:
           NODE_AUTH_TOKEN: ${{ secrets.NPM_PUBLISH_TOKEN }}
           GITHUB_TOKEN: ${{ steps.app-token.outputs.token }}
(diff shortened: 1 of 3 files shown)


.github/workflows/release.yml
```diff
@@ -72,12 +72,32 @@ jobs:
       - name: Build dependencies
         run: pnpm run build
 
+      - name: Checkout main repository
+        if: ${{ inputs.ref && inputs.ref != 'main' }}
+        uses: actions/checkout@v5
+        with:
+          path: 'main'
+
+      - name: Release packages
```
almeidx: Skipped steps are still shown on the UI on action runs, so this and the following steps should probably have different names (maybe a " (main)" suffix)

ckohen: Yeah I went by what we had done for the docs runs, but I do think we should name it better. 5e0c8d51c

packages/actions/src/releasePackages/action.yml
```diff
@@ -11,6 +11,9 @@ inputs:
     description: 'The published name of a single package to release'
   exclude:
     description: 'Comma separated list of packages to exclude from release (if not depended upon)'
+  main:
+    description: 'Is this running on the main branch?'
```
Jiralite: Is it me, or this description doesn't match the usage?

> `if: ${{ !inputs.ref || inputs.ref == 'main' }}`
> ...
> `main: false`

.github/workflows/release.yml
```diff
@@ -72,12 +72,43 @@ jobs:
...
+        with:
+          path: 'main'
+
+      - name: Install action deps (non-main ref only)
+        if: ${{ inputs.ref && inputs.ref != 'main' }}
+        shell: bash
+        env:
+          COREPACK_ENABLE_STRICT: 0
+        run: |
+          cd main
```
almeidx: Shouldn't we use `working-directory`[^1] here instead of manually `cd`'ing?

[^1]: https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#jobsjob_idstepsworking-directory

.github/workflows/release.yml
```diff
@@ -72,12 +72,43 @@ jobs:
...
+          pnpm self-update 10
+          pnpm install --filter @discordjs/actions --frozen-lockfile --prefer-offline --loglevel error
+          cd ..
+
+      - name: Release packages (non-main ref)
+        if: ${{ inputs.ref && inputs.ref != 'main' }}
+        uses: ./main/packages/actions/src/releasePackages
```
almeidx, line 93: Doesn't this also need to use `working-directory: ./main` for it to use the correct script and for the `$.cwd('../');` to actually change the working dir to the correct place?
https://github.com/discordjs/discord.js/blob/f510b5ffabfb22f634876e27e5316db36f9c80b4/packages/actions/src/releasePackages/action.yml#L18

packages/actions/src/releasePackages/action.yml
ckohen: flipped the logic, thanks

.github/workflows/release.yml
ckohen: that makes sense, again was following what we had done, but this is cleaner

.github/workflows/release.yml
ckohen: working directory only works with run, but this brings up I need to check if the working directory is actually what I think it is, will do that shortly, blocking until I do

Jiralite (approved):

.github/workflows/release.yml
ckohen: I should've remembered it from when I was doing the initial testing, but I see what you are saying now. It actually works out really well that the working directory is always the workspace root, cause github gives me all the information I need and I don't actually need to pass down whether we're running from main or not, it just works ™️

PR #11137, opened 2025-10-05
fix: Add missing required flag for getTextInputValue (QoL)

Axodouble:

This adds (what I believe) is a missing flag from getTextInputValue, I noticed that other methods do have a required flag yet getTextInputValue does not which I believe would benefit from having it, also for consistency reasons.

Would love to hear if there are any comments or questions!


imnaiyar:
This is by choice; if there is a text input component, then the `value` will always be present. So there's no point in having `required` there (notice the return type doesn't say it returns `null`). Same for `getStringSelectValues`.

axodouble:
Ah I am sorry about that, completely overlooked that

PR #11140, opened 2025-10-05
types(core): use result types instead of direct types

vladfrangu:

Just fixes some type references. WIll PR to DTypes to fix the thread return type being a generic channel instead of a specific type

### DIFF
--- a/packages/core/scripts/find-returns-not-return-type.mts
+++ b/packages/core/scripts/find-returns-not-return-type.mts
@@ -1,0 +1,28 @@
+import { glob, readFile } from 'node:fs/promises';
+
+const cwd = new URL('../src/api/', import.meta.url);
+const results: string[] = [];
+
+for await (const file of glob('**/*.ts', { cwd })) {
+	const content = await readFile(new URL(file, cwd), { encoding: 'utf-8' });
+
+	const matches = content.matchAll(/as Promise<(?<returnType>\w+)>/g);
+
+	for (const match of matches) {
+		const returnType = match.groups!.returnType!;
+
+		if (!returnType.startsWith('REST')) {
+			results.push(`in file core/src/api/${file}: ${returnType}`);
+		}
+	}
+}
+
+if (results.length > 0) {
+	console.warn('Found return types that are not REST return types:');
+
+	for (const result of results) {
+		console.warn(`  - ${result}`);
+	}
+} else {
+	console.log('No return types that are not REST return types found');
+}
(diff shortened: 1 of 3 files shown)


Jiralite:
Drafting until the discord-api-types release.

packages/core/src/api/thread.ts
```diff
@@ -1,12 +1,11 @@
...
 import type { RequestData, REST } from '@discordjs/rest';
-import {
+import type {
+	RESTGetAPIChannelThreadMemberResult,
 	Routes,
-	type APIThreadMember,
 	type RESTGetAPIChannelThreadMembersResult,
-	type Snowflake,
-} from 'discord-api-types/v10';
+	type Snowflake} from 'discord-api-types/v10';
```
Jiralite: Imports messed up here!

Jiralite (dismissed):

Jiralite:
Please resolve conflicts!

packages/core/scripts/find-returns-not-return-type.mts
```diff
@@ -0,0 +1,28 @@
...
+
+for await (const file of glob('**/*.ts', { cwd })) {
+	const content = await readFile(new URL(file, cwd), { encoding: 'utf-8' });
+
+	const matches = content.matchAll(/as Promise<(?<returnType>\w+)>/g);
+
+	for (const match of matches) {
+		const returnType = match.groups!.returnType!;
+
+		if (!returnType.startsWith('REST')) {
```
almeidx: I think this makes sense too
```suggestion
		if (!returnType.startsWith('REST') || !returnType.includes("Result")) {
```

packages/core/src/api/thread.ts
```diff
@@ -108,7 +108,7 @@ export class ThreadsAPI {
 			auth,
 			signal,
 			query: makeURLSearchParams(query),
-		});
+		}) as Promise<RESTGetAPIChannelThreadMemberResult>;
```
Jiralite: This isn't needed due to the above overloads!

Jiralite (approved):

packages/core/scripts/find-returns-not-return-type.mts
Jiralite, line 1: Thoughts on renaming the other script to .mts?

vladfrangu:
```
The following routes are not implemented:
 - guildMemberVerification
 ```

**Comment by Jiralite** [comment]
That's an unstable/undocumented route so that's fine.

File: packages/core/scripts/check-routes.mts
Comment (almeidx) [inline packages/core/scripts/check-routes.mts]: Since you're making this a typescript file
```suggestion
const usedRoutes = new Set<string>();
```

sdanialraza (approved):

SpaceEEC (approved):

PR #11158, opened 2025-10-08
fix(ModalSubmitInteraction): Resolve crash on handling populated select menus

Jiralite:
- `this` in the static method doesn't resolve the way we want to
- Members do not have a user object in resolved data

### DIFF
--- a/packages/discord.js/src/structures/Poll.js
+++ b/packages/discord.js/src/structures/Poll.js
@@ -164,7 +164,7 @@ class Poll extends Base {
    * @returns {Promise<Message>}
    */
   async end() {
-    if (Date.now() > this.expiresTimestamp) {
+    if (this.expiresTimestamp !== null && Date.now() > this.expiresTimestamp) {
       throw new DiscordjsError(ErrorCodes.PollAlreadyExpired);
     }
 


packages/discord.js/src/structures/ModalSubmitInteraction.js
```diff
@@ -121,22 +124,25 @@ class ModalSubmitInteraction extends BaseInteraction {
    * Transforms component data to discord.js-compatible data
    * @param {*} rawComponent The data to transform
    * @param {APIInteractionDataResolved} [resolved] The resolved data for the interaction
+   * @param {*} [extra] Extra data required for the transformation
    * @returns {ModalData[]}
    */
-  static transformComponent(rawComponent, resolved) {
+  static transformComponent(rawComponent, resolved, { client, guild } = {}) {
```
almeidx, line 130: Is this meant to be a private function? Because it's not marked as such in the JSDoc, and it's also not typed.

Additionally, why is this a static method anyways? It's only being used in this class once (besides the recursive calls) in the constructor

Jiralite: It is meant to be private. Making it private now or converting it to an instance would sadly be a breaking change.

vladfrangu, line 130:
```suggestion
  static transformComponent(rawComponent, resolved, { client, guild }) {
```

Jiralite: Could we leave it as-is for ease? Note that this is a public method so a user may just not pass it (rightly so if they're not using the new select menus).

PR #11164, opened 2025-10-09
fix(ApplicationCommandManager): set should use a different type

didinele:
in case those types ever diverge

### DIFF
--- a/packages/discord.js/typings/index.d.ts
+++ b/packages/discord.js/typings/index.d.ts
@@ -193,6 +193,7 @@ import {
   RESTPostAPIInteractionCallbackWithResponseResult,
   RESTPostAPIInteractionFollowupJSONBody,
   RESTPostAPIWebhookWithTokenJSONBody,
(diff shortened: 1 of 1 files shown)


packages/discord.js/typings/index.d.ts
```diff
@@ -4162,6 +4163,11 @@ export type ApplicationCommandDataResolvable =
   | JSONEncodable<RESTPostAPIApplicationCommandsJSONBody>
   | RESTPostAPIApplicationCommandsJSONBody;
 
+export type ApplicationEditCommandDataResolvable =
+  | ApplicationCommandData
+  | JSONEncodable<RESTPutAPIApplicationCommandPermissionsJSONBody>
```
Qjuh, line 4168: What do permissions have to do with any of this? And if you meant RESTPutAPIApplicationCommandsJSONBody then that is already an array of commands, but this type is used as an element of an array, so that's wrong too

didinele:
yea whatever lol

PR #11165, opened 2025-10-09
fix(ApplicationCommandPermissionsManager): truly optional arg

### LINKED ISSUE #8096: `ApplicationCommandPermissionsManager#fetch` typings issue

discord.js

### Issue description

1. Have a guild and try to fetch application command permissions using provided code sample.
2. According to documentation, `options` parameter should be optional in the code sample below.
3. However, typescript complains that `options` parameter wasn't provided.

### Code sample

```typescript
client.on('messageCreate', message => {
  if (!message.inGuild()) return;
  
  const permissions = message.guild.commands.permissions.fetch()
})
```

### Package version

discord.js@14.0.0-dev.1655165434-b4e28a8

### Node.js version

16.15.1, typescript 4.7.3

### Operating system

Windows

Medium (should be fixed soon)

No Partials

Guilds, GuildMembers, GuildMessages

b4e28a8

didinele:
Closes #8096

### DIFF
--- a/packages/discord.js/typings/index.d.ts
+++ b/packages/discord.js/typings/index.d.ts
@@ -4296,7 +4296,9 @@ export class ApplicationCommandPermissionsManager<
     },
   ): Promise<boolean>;
   public fetch(options: FetchSingleOptions): Promise<ApplicationCommandPermissions[]>;
-  public fetch(options: BaseOptions): Promise<Collection<Snowflake, ApplicationCommandPermissions[]>>;
+  public fetch(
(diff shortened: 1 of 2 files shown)


didinele:
we discussed it internally and while everyone applauded my genius, we have instead decided to just do some good 'ole boilerplate and make 2 separate classes, essentially considering that highly abstracted monster with 4 generics is only used in *2* places

didinele:
not sure if this approach is substantially better on the docs side, open to other suggestions. there's another layer of class-split we could do here to make it even more explicit

PR #11170, opened 2025-10-11
refactor(guide): redirect old urls to /legacy

imnaiyar:
Search engines may still show old urls, this makes it so they are properly redirected to /legacy
This is not needed for /voice since it is already valid

This assumes that the current guide pathname is 1:1 with the old one, which it currently is. If it is changed in the future, this needs to be revisited, at which point this may not even be needed and can be removed.


apps/guide/src/middleware.ts
```diff
@@ -8,9 +8,16 @@ export function middleware(request: NextRequest) {
 		return NextResponse.redirect(new URL(newUrl.pathname, request.url));
 	}
 
-	return NextResponse.redirect(new URL('/legacy', request.url));
+	// Redirect old urls to /legacy
+	if (!request.nextUrl.pathname.startsWith('/legacy') && !request.nextUrl.pathname.startsWith('/voice')) {
+		const newUrl = request.nextUrl.clone();
+		newUrl.pathname = `/legacy${newUrl.pathname}`;
+		return NextResponse.redirect(new URL(newUrl.pathname, request.url));
```
almeidx: This should be enough (i know i was the one that introduced this above... 😅)
```suggestion
		return NextResponse.redirect(newUrl);
```

apps/guide/src/middleware.ts
```diff
@@ -8,9 +8,16 @@ export function middleware(request: NextRequest) {
 		return NextResponse.redirect(new URL(newUrl.pathname, request.url));
```
almeidx:
```suggestion
		return NextResponse.redirect(newUrl);
```

PR #11184, opened 2025-10-16
fix(Client): Move client initialization code

### LINKED ISSUE #11180: ClientReady is called without user and application

discord.js

### Issue description

An error occurs at startup:
TypeError: Cannot read properties of null (reading 'displayName')

**The most important part**: bot should not be on any server

### Code sample

```typescript
const { Client, Events } = require('discord.js')
const { token } = require('./config.json')

const client = new Client({
	intents: []
})

client.once(Events.ClientReady, (client) => console.log(`${client.user.displayName} is ready!`))

client.login(token)
```

### Versions

discord.js - 15.0.0-dev.1760486511-25b22c67a
node.js  - v23.10.0

### Issue priority

Medium (should be fixed soon)

No Partials

No Intents

25b22c67a

vakiliner:

Closes #11180


packages/discord.js/src/client/Client.js
```diff
@@ -44,6 +45,8 @@ const BeforeReadyWhitelist = [
   GatewayDispatchEvents.GuildMemberRemove,
 ];
 
+let ClientUser;
```
vakiliner: Should ClientUser be lazy loaded?

Qjuh:
I'd say the better fix would be to move all the READY event handling into READY.js, not the other way around. Or at least the checkReady() call

vakiliner:
> I'd say the better fix would be to move all the READY event handling into READY.js, not the other way around. Or at least the checkReady() call

I just noticed that the Ready event is async
None of the packet handlers are async, so calling this function won't be awaitable
While this can easily be fixed by adding "await" before "PacketHandlers[packet.t](this, packet, shardId);", but is it necessary?

vakiliner:
> I'd say the better fix would be to move all the READY event handling into READY.js, not the other way around. Or at least the checkReady() call

I believe this was not added to READY.js on purpose because the handler is async

Qjuh:
> > I'd say the better fix would be to move all the READY event handling into READY.js, not the other way around. Or at least the checkReady() call
> 
> I believe this was not added to READY.js on purpose because the handler is async

Since I was the one who wrote that I can assure you that's not the reason. In v14 the checkReady check is in READY.js; this was simply an oversight when adapting to the direct use of /ws and should be reverted to check in READY.js

adding await would be a good idea however, and then the checkReady check for READY event could be added to handlePacket instead. Similar to how it checks for GUILD_CREATE and _DELETE.

vakiliner:
> In v14 the checkReady check is in READY.js

Then checkReady() was not async

packages/discord.js/src/client/websocket/handlers/READY.js
```diff
@@ -23,4 +25,7 @@ module.exports = (client, { d: data }, shardId) => {
   } else {
     client.application = new ClientApplication(client, data.application);
   }
+
+  client.status = Status.WaitingForGuilds;
+  await client._checkReady();
```
Qjuh: If we await the checkReady it should be done in `Client#handleMessage()`; the READY handler should not be async.

vakiliner: I can't find `Client#handleMessage()`, did you mean `Client#_handlePacket()`?

didinele:
was a little later than "tonight", but I got to it. we should be good :)

PR #11185, opened 2025-10-17
chore: bump dependencies

almeidx:
- eslint-config-neon patch no longer needed
- `tseslint.config()` -> `defineConfig()`: https://typescript-eslint.io/packages/typescript-eslint/#config-deprecated

### DIFF
--- a/apps/guide/next.config.ts
+++ b/apps/guide/next.config.ts
@@ -4,7 +4,6 @@ import type { NextConfig } from 'next';
 const withMDX = createMDX();
 
 export default withMDX({
-	reactStrictMode: true,
 	serverExternalPackages: ['typescript', 'twoslash'],
 	images: {
 		dangerouslyAllowSVG: true,
--- a/apps/website/next.config.ts
+++ b/apps/website/next.config.ts
@@ -1,7 +1,6 @@
 import type { NextConfig } from 'next';
 
 export default {
-	reactStrictMode: true,
 	images: {
 		dangerouslyAllowSVG: true,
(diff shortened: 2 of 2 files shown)


vladfrangu:
Google Chrome Beta - 2025-10-18 at 11-37-02-UTC@2x

packages/discord.js/src/structures/ClientApplication.js
```diff
@@ -98,6 +98,7 @@ class ClientApplication extends Application {
      * Scopes and permissions regarding the installation context
      */
 
+    /* eslint-disable jsdoc/valid-types */
```
Qjuh: Considering these docs aren't even shown in new docs site this should probably be rewritten. Same for the other case you disabled this rule.

packages/discord.js/src/util/Constants.js
```diff
@@ -202,6 +202,7 @@ exports.UndeletableMessageTypes = [
   MessageType.ThreadStarterMessage,
 ];
 
+/* eslint-disable jsdoc/valid-types */
```
Qjuh: Ditto

packages/discord.js/src/structures/ClientApplication.js
almeidx: What would you suggest?

Qjuh: In this particular case I'd actually propose a breaking change to make it a Map/Collection. But guess for the purpose of this PR disabling the lint rule and a todo comment suffice.

PR #11189, opened 2025-10-21
fix(Message): update `pinnable` to check for migrated guilds

### LINKED ISSUE #11188: Change message.pinnable to use the new PinMessages permission

discord.js

### Feature

[message.pinnable](https://github.com/discordjs/discord.js/blob/main/packages/discord.js/src/structures/Message.js#L800) now checks if client user has the ManageMessages permission. Now that Discord added a seperate permission for PinMessages it should use this instead of ManageMessages

Change ManageMessages to PinMessages 

_No response_

### Other context

_No response_

sdanialraza:
Resolves #11188

### DIFF
--- a/packages/discord.js/src/structures/Message.js
+++ b/packages/discord.js/src/structures/Message.js
@@ -781,16 +781,17 @@ class Message extends Base {
    */
   get pinnable() {
     const { channel } = this;
-    return Boolean(
-      !this.system &&
-        (!this.guild ||
-          (channel?.viewable &&
-            (channel
-              ?.permissionsFor(this.client.user)
-              ?.has([PermissionFlagsBits.ReadMessageHistory, PermissionFlagsBits.PinMessages], false) ||
-              channel
-                ?.permissionsFor(this.client.user)
-                ?.has([PermissionFlagsBits.ReadMessageHistory, PermissionFlagsBits.ManageMessages], false)))),
+
+    if (this.system) return false;
+    if (!this.guild) return true;
+    if (!channel?.viewable) return false;
+
+    const permissions = channel?.permissionsFor(this.client.user);
+    if (!permissions) return false;
+
+    return (
+      permissions.has([PermissionFlagsBits.ReadMessageHistory, PermissionFlagsBits.PinMessages]) ||
+      permissions.has([PermissionFlagsBits.ReadMessageHistory, PermissionFlagsBits.ManageMessages])
     );
   }
 


Jiralite:
For main, I would say change this completely. I doubt we will finalise our major release before 12th January 2026.

This exact change could be done for v14, however.

packages/discord.js/src/structures/Message.js
```diff
@@ -803,7 +803,14 @@ class Message extends Base {
       !this.system &&
         (!this.guild ||
           (channel?.viewable &&
-            channel?.permissionsFor(this.client.user)?.has(PermissionFlagsBits.ManageMessages, false))),
+            channel
+              ?.permissionsFor(this.client.user)
+              ?.has(
+                this.guild?.features.includes('PIN_PERMISSION_MIGRATION_COMPLETE')
+                  ? PermissionFlagsBits.PinMessages
+                  : PermissionFlagsBits.ManageMessages,
```
vakiliner: First, why check if a guild has this feature?
At the moment, even without this feature, PinMessages gives the right to pin messages
Secondly, pinning messages also requires the ReadMessageHistory permission

sdanialraza: - Because ManageMessages still works for guilds that haven't been migrated to PinMessages
- Thank you for mentioning `ReadMessageHistory`, added that in the check

packages/discord.js/src/structures/Message.js
```diff
@@ -780,12 +780,16 @@ class Message extends Base {
    * @readonly
    */
   get pinnable() {
-    const { channel } = this;
```
Qjuh, line 783: This should stay, as it was done for performance to not use the getter twice.

packages/discord.js/src/structures/Message.js
```diff
@@ -780,12 +780,16 @@ class Message extends Base {
...
-            channel?.permissionsFor(this.client.user)?.has(PermissionFlagsBits.ManageMessages, false))),
+    if (this.system || !this.channel?.viewable) return false;
+
+    const permissions = this.channel.permissionsFor(this.client.user);
+    if (!permissions) return false;
+
+    return (
+      !this.guild ||
+      permissions.has([PermissionFlagsBits.ReadMessageHistory, PermissionFlagsBits.PinMessages], false) ||
+      (!this.guild.features.includes('PIN_PERMISSION_MIGRATION_COMPLETE') &&
```
Jiralite: As of 0ff239a602aed71c74c931c975c84900cf671312, we can use `GuildFeature.PinPermissionMigrationComplete`!

packages/discord.js/src/structures/Message.js
```diff
@@ -780,12 +780,16 @@ class Message extends Base {
...
-        (!this.guild ||
-          (channel?.viewable &&
-            channel?.permissionsFor(this.client.user)?.has(PermissionFlagsBits.ManageMessages, false))),
+    if (this.system || !this.channel?.viewable) return false;
+
+    const permissions = this.channel.permissionsFor(this.client.user);
+    if (!permissions) return false;
+
+    return (
+      !this.guild ||
```
Qjuh: This check is too late, it needs to be done before we do permissionsFor(), since that makes no sense for DMChannels

packages/discord.js/src/structures/Message.js
```diff
@@ -780,12 +780,16 @@ class Message extends Base {
...
+      permissions.has([PermissionFlagsBits.ReadMessageHistory, PermissionFlagsBits.PinMessages], false) ||
+      (!this.guild.features.includes('PIN_PERMISSION_MIGRATION_COMPLETE') &&
+        permissions.has([PermissionFlagsBits.ReadMessageHistory, PermissionFlagsBits.ManageMessages], false))
```
Qjuh: Did you test that ReadMessageHistory is explicitly required here and not just implicitly granted by ManageMessages?

Qjuh (dismissed):

packages/discord.js/src/structures/Message.js
sdanialraza: Yes, it errors with Missing Access if you don't have `ReadMessageHistory` (even if you have `ManageMessages`)

Qjuh:
> The migration has nothing to do with which permissions are usable - if you have either Manage Messages or Pin Messages you can pin. (as long as you also have read message history)

The notice in permissions UI in the client seems to state differently. That's the whole point of that new separate permission afaict. 
IMG_7355

advaith1:
on that date manage messages will stop allowing pinning. that is separate from the guild feature

Jiralite (changes requested):
From internal discussions,[^1] we're going to require both permissions and make a release after removing `MANAGE_MESSAGES`.

[^1]: https://canary.discord.com/channels/222078108977594368/999796393810067607/1431221049915215902

Jiralite (approved):

packages/discord.js/src/structures/Message.js
```diff
@@ -785,7 +785,12 @@ class Message extends Base {
       !this.system &&
         (!this.guild ||
           (channel?.viewable &&
-            channel?.permissionsFor(this.client.user)?.has(PermissionFlagsBits.ManageMessages, false))),
+            (channel
+              ?.permissionsFor(this.client.user)
+              ?.has([PermissionFlagsBits.ReadMessageHistory, PermissionFlagsBits.PinMessages], false) ||
+              channel
+                ?.permissionsFor(this.client.user)
+                ?.has([PermissionFlagsBits.ReadMessageHistory, PermissionFlagsBits.ManageMessages], false)))),
```
vladfrangu: That false should be removed, admin obv allows it.

Also please for the love of god split this into multiple lines :D

Qjuh (approved):
Apart from vlad's comment LGTM

PR #11200, opened 2025-10-24
feat: Annotate result of releasing packages

Jiralite:
We will use summaries for this workflow!


ckohen:
If we're doing this we should probably just...[do job summary](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-commands#adding-a-job-summary)

Issue #11192, opened 2025-10-23
Some permission check getters do not check timeout and other properties


vakiliner:
### Which package is this bug report for?

discord.js

### Issue description

Timeout not checked (Related #9730):
- `GuildEmoji.deletable`
- `GuildInvite.deletable`
- `GuildMember.moderatable`
- `GuildMember.kickable`
- `GuildMember.bannable`
- `Message.editable`
- `Message.pinnable`
- `Message.crosspostable`

Timeout is not checked with additional conditions (Related #9730):
- `Message.bulkDeletable` When the message was sent by the client
- `ThreadChannel.editable` When the thread was created by the client
- `ThreadChannel.joinable` When the thread is private

Permission errors:
- [x] `Message.pinnable` **ReadMessageHistory** is not checked
- [x] `Message.pinnable` Doesn't check if the channel supports pins
- [ ] `ThreadChannel.editable` When a thread is created by a client, **ViewChannel** and **SendMessagesInThreads** are not checked
- [ ] `ThreadChannel.joinable` For private threads, **ManageThreads** is checked, but **ViewChannel** is not (A similar case with `VoiceChannel.manageable` #11116)
- [ ] `ThreadChannel.manageable` **ViewChannel** and **SendMessagesInThreads** are not checked
- [ ] `ThreadChannel.sendable` **ViewChannel** is not checked
- [ ] `ThreadChannel.sendable` When the thread is locked and archived **ManageThreads** is checked, but not when the thread is **only locked**
- [ ] `ThreadMember.manageable` When a **public** thread is created by a client, **ManageThreads** is not checked

### Code sample

There is automated check code, but it is large
See https://github.com/vakiliner/discordjs-error-tests
If you want, you can use it, or reproduce the issues manually

### Versions

discord.js - 15.0.0-dev.1761134503-697556fd1
nodejs - v25.0.0
TypeScript - Version 5.9.3
Windows 10

### Issue priority

Low (slightly annoying)

### Which partials do you have configured?

No Partials

### Which gateway intents are you subscribing to?

Guilds, GuildMembers, GuildPresences, GuildMessages, MessageContent

### I have tested this issue on a development release

697556fd1

Jiralite:
Note: communication disabled checks are a duplicate of #9730.

There is discussion in there on paths forward, but nothing has come of it (yet).

vakiliner:
[...]

PR #11215, opened 2025-10-26
fix(Message): check if in voice based channel for `pinnable`

sdanialraza:
You can't pin messages in voice based channels: https://github.com/discordjs/discord.js/issues/11192#issuecomment-3446575607

### DIFF
--- a/packages/discord.js/src/structures/Message.js
+++ b/packages/discord.js/src/structures/Message.js
@@ -24,12 +24,7 @@ const { Sticker } = require('./Sticker');
 const { DiscordjsError, ErrorCodes } = require('../errors');
 const ReactionManager = require('../managers/ReactionManager');
 const { createComponent, findComponentByCustomId } = require('../util/Components');
-const {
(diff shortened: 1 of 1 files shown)


Jiralite (approved):

packages/discord.js/src/structures/Message.js
```diff
@@ -784,14 +789,14 @@ class Message extends Base {
 
     if (this.system) return false;
     if (!this.guild) return true;
-    if (!channel?.viewable) return false;
+    if (VoiceBasedChannelTypes.includes(channel?.type) || !channel?.viewable) return false;
```
Qjuh:
```suggestion
    if (!channel?.viewable || VoiceBasedChannelTypes.includes(channel.type)) return false;
```
Would prefer to not have an `includes(undefined)` check

almeidx: This should be changed to match the check done in the main branch

PR #11236, opened 2025-11-01
refactor(publish-dev): take ref instead

didinele:


.github/workflows/publish-dev.yml
```diff
@@ -63,13 +54,13 @@ jobs:
         run: pnpm run build
 
       - name: Checkout main repository (non-main ref)
-        if: ${{ steps.ref.outputs.ref != 'refs/heads/main' }}
+        if: ${{ inputs.ref != 'refs/heads/main' }}
```
Jiralite: This will always be `false`, right? The default is `"main"` from above.

Jiralite (changes requested):

Jiralite (approved):

.github/workflows/publish-dev.yml
```diff
@@ -4,8 +4,8 @@ on:
     - cron: '0 */12 * * *'
   workflow_dispatch:
     inputs:
-      pull:
-        description: 'The pull number to check out'
+      ref:
+        description: 'The ref to check out'
```
almeidx: Might be useful to include examples here

didinele: will do

PR #11239, opened 2025-11-02
perf: optimize role manager cache getter

WhatCats:
I believe a `Set` would be better suited for the private member `_roles` collection.
I think the role cache getters will perform better by iterating over the `_roles` subset instead of every guild role.

Context: Trying to do stuff with the roles of every guild member in a guild with 20000 members and 200 roles.

### DIFF
--- a/.github/ISSUE_TEMPLATE/01-package_bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/01-package_bug_report.yml
@@ -62,9 +62,9 @@ body:
       label: Versions
       description: List necessary versions here. This includes your package version, runtime version, operating system etc.
       placeholder: |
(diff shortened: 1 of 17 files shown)


packages/discord.js/src/structures/GuildMember.js
```diff
@@ -59,7 +59,7 @@ class GuildMember extends Base {
      * @type {Snowflake[]}
```
vladfrangu, line 59: If you move this to a set, you'll need to adjust the jsdoc too

kyranet (approved):

almeidx (approved):

Jiralite (approved):

PR #11240, opened 2025-11-02
Improve stack trace handling and error code validation

reversalxx:
Refactor stack trace capture for consistency and improve error code validation.

This pull request improves how stack traces are handled and makes error code validation safer. It does not change how the library works, only makes the internal error system more reliable and consistent across environments.

Why it should be merged:
- Works the same as before, no breaking changes
- Cleaner and safer internal code
- Better stack trace handling for debugging
Status:
- No APi or behavior changes
- Tested and stable
- Typings not affected


Jiralite:
Can we see the before & after examples then?

PR #11248, opened 2025-11-05
feat(builders): multipart form data output support

didinele:


packages/builders/src/messages/Attachment.ts
```diff
@@ -12,21 +13,33 @@ export class AttachmentBuilder implements JSONEncodable<RESTAPIAttachment> {
 	 */
 	private readonly data: Partial<RESTAPIAttachment>;
 
+	/**
+	 * This data not included in the output of `toJSON()`. For this class specifically, this refers to binary data
```
Qjuh:
```suggestion
	 * This data is not included in the output of `toJSON()`. For this class specifically, this refers to binary data
```

packages/builders/src/messages/Attachment.ts
```diff
@@ -12,21 +13,33 @@ export class AttachmentBuilder implements JSONEncodable<RESTAPIAttachment> {
...
+	 * that will wind up being included in the multipart/form-data request, if using with the `MessageBuilder`.
```
Qjuh:
```suggestion
	 * that will wind up being included in the multipart/form-data request, if used with the `MessageBuilder`.
```

packages/builders/src/messages/Message.ts
```diff
@@ -56,7 +56,9 @@ export interface MessageBuilderData
 /**
  * A builder that creates API-compatible JSON data for messages.
  */
-export class MessageBuilder implements JSONEncodable<RESTPostAPIChannelMessageJSONBody> {
+export class MessageBuilder
+	implements JSONEncodable<RESTPostAPIChannelMessageJSONBody>, FileBodyEncodeable<RESTPostAPIChannelMessageJSONBody>
```
Qjuh:
```suggestion
	implements FileBodyEncodeable<RESTPostAPIChannelMessageJSONBody>
```

This is redundant considering `FileBodyEncodeable` extends `JSONEncodable`

packages/builders/src/messages/Attachment.ts
```diff
@@ -85,6 +98,60 @@ export class AttachmentBuilder implements JSONEncodable<RESTAPIAttachment> {
 		return this;
 	}
 
+	/**
+	 * Sets the file data to upload with this attachment.
+	 *
+	 * @param data - The file data (Buffer, Uint8Array, or string)
+	 * @remarks Note that this data is NOT included in the `toJSON()` output. To retrieve, use {@link getRawFile}.
+	 */
+	public setFileData(data: Buffer | Uint8Array | boolean | number | string): this {
```
Qjuh: From what I can tell from docs it seems like support for `boolean | number` in RawFile in `@discordjs/rest` is for when not using `payload_json` but instead sending all properties as form data. So this should probably not be supported as parameters here.
```suggestion
	public setFileData(data: Buffer | Uint8Array | string): this {
```

packages/builders/src/messages/Assertions.ts
```diff
@@ -1,9 +1,20 @@
+import { Buffer } from 'node:buffer';
 import { AllowedMentionsTypes, ComponentType, MessageFlags, MessageReferenceType } from 'discord-api-types/v10';
 import { z } from 'zod';
 import { embedPredicate } from './embed/Assertions.js';
 import { pollPredicate } from './poll/Assertions.js';
 
+const fileKeyRegex = /^files\[(?<placeholder>.+?)]$/;
+
+export const rawFilePredicate = z.object({
+	data: z.union([z.instanceof(Buffer), z.instanceof(Uint8Array), z.boolean(), z.number(), z.string()]),
```
Qjuh:
```suggestion
	data: z.union([z.instanceof(Buffer), z.instanceof(Uint8Array), z.string()]),
```

packages/util/src/JSONEncodable.ts
```diff
@@ -18,3 +20,49 @@ export interface JSONEncodable<Value> {
 export function isJSONEncodable(maybeEncodable: unknown): maybeEncodable is JSONEncodable<unknown> {
 	return maybeEncodable !== null && typeof maybeEncodable === 'object' && 'toJSON' in maybeEncodable;
 }
+
+/**
+ * Result of encoding an object that includes file attachments
+ *
+ * @typeParam BodyValue - The JSON body type
+ */
+export interface FileBodyEncodeableResult<BodyValue> {
```
Qjuh:
```suggestion
export interface FileBodyEncodableResult<BodyValue> {
```

packages/util/src/JSONEncodable.ts
```diff
@@ -18,3 +20,49 @@ export interface JSONEncodable<Value> {
...
+}
+
+/**
+ * Represents an object capable of representing itself as a request body with file attachments.
+ * Objects implementing this interface can separate JSON body data from binary file data,
+ * which is necessary for multipart/form-data requests.
+ *
+ * @typeParam BodyValue - The JSON body type
+ */
+export interface FileBodyEncodeable<BodyValue> extends JSONEncodable<BodyValue> {
```
Qjuh:
```suggestion
export interface FileBodyEncodable<BodyValue> extends JSONEncodable<BodyValue> {
```
And everywhere else.

packages/builders/src/messages/Message.ts
didinele: I'm aware, but I'm honestly for leaving it in just for extra readability without knowing what a FileBodyEncodeable is (arguably beneficial for docs too?). Thoughts @vladfrangu

Qjuh: For docs it would (should) show the toJSON() on the extended interface already, so not sure of the benefit

packages/builders/src/messages/Attachment.ts
```diff
@@ -85,6 +98,60 @@ export class AttachmentBuilder implements JSONEncodable<RESTAPIAttachment> {
...
+		return this;
+	}
+
+	/**
+	 * Converts this attachment to a RawFile for uploading.
+	 *
+	 * @returns A RawFile object, or undefined if no file data is set
+	 */
+	public getRawFile(): Partial<RawFile> | undefined {
+		if (!this.fileData) {
```
ckohen: This should probably check if fileData.data is undefined as well, since you aren't deleting keys when you clear.

(I do see you check this in the message builder but why not do it here)

ckohen (dismissed):

packages/builders/src/messages/Assertions.ts
```diff
@@ -1,9 +1,20 @@
+import { Buffer } from 'node:buffer';
 import { AllowedMentionsTypes, ComponentType, MessageFlags, MessageReferenceType } from 'discord-api-types/v10';
 import { z } from 'zod';
 import { embedPredicate } from './embed/Assertions.js';
 import { pollPredicate } from './poll/Assertions.js';
 
+const fileKeyRegex = /^files\[(?<placeholder>.+?)]$/;
```
vladfrangu: isn't the placeholder supposed to be \d+ only?

didinele: I'm actually a bit unclear from the docs. It seems like no generally, but *maybe* for messages specifically?

packages/util/src/RawFile.ts
```diff
@@ -0,0 +1,29 @@
...
+	 * An explicit key to use for the formdata field for this file.
+	 * When not provided, the index of the file in the files array is used in the form `files[${index}]`.
+	 * If you wish to alter the placeholder snowflake, you must provide this property in the same form (`files[${placeholder}]`)
+	 */
+	key?: string;
+	/**
+	 * The name of the file. This is the actual filename that will be used when uploading to Discord.
+	 * This is also the name you'll use to reference the file with attachment:// URLs.
+	 *
+	 * @example 'image.png', 'document.pdf', 'SPOILER_secret.jpg'
```
Jiralite:
```suggestion
	 * @example 'image.png'
	 * @example 'document.pdf'
	 * @example 'SPOILER_secret.jpeg'
```

packages/builders/src/messages/Assertions.ts
Jiralite: From here:

https://github.com/discord/discord-api-spec/blob/28ef48483b4745bcf176b2b53b931a27142e2f55/specs/openapi.json#L3267-L3306

It should be numbers, it seems. I wouldn't limit it to just those because the documentation says `files[42]` can be used.[^1]

[^1]: https://discord.com/developers/docs/reference#uploading-files

packages/util/src/encodables.ts
```diff
@@ -0,0 +1,68 @@
...
+ *
+ * @param maybeEncodable - The object to check against
+ */
+export function isFileBodyEncodable(maybeEncodable: unknown): maybeEncodable is FileBodyEncodable<unknown> {
+	return (
+		maybeEncodable !== null &&
+		typeof maybeEncodable === 'object' &&
+		'toJSON' in maybeEncodable &&
+		'toFileBody' in maybeEncodable
+	);
```
almeidx:
```suggestion
	return isJSONEncodable(maybeEncodable) && 'toFileBody' in maybeEncodable;
```

packages/util/src/RawFile.ts
```diff
@@ -0,0 +1,31 @@
+import type { Buffer } from 'node:buffer';
+
+/**
+ * Represents a file to be added to a request with multipart/form-data encoding
+ */
+export interface RawFile {
+	/**
+	 * Content-Type of the file (e.g., 'image/png', 'application/pdf')
+	 * If not provided, it will be inferred from the file data when possible
```
almeidx:
```suggestion
	 * Content-Type of the file.
	 * If not provided, it will be inferred from the file data when possible
	 *
	 * @example 'image/png'
	 * @example 'application/pdf'
```

packages/rest/src/lib/utils/types.ts
```diff
@@ -276,29 +276,7 @@ export interface InvalidRequestWarningData {
...
-	 * When not provided, the index of the file in the files array is used in the form `files[${index}]`.
-	 * If you wish to alter the placeholder snowflake, you must provide this property in the same form (`files[${placeholder}]`)
-	 */
-	key?: string;
-	/**
-	 * The name of the file
-	 */
-	name: string;
-}
+export { type RawFile } from '@discordjs/util';
```
almeidx:
```suggestion
export type { RawFile } from '@discordjs/util';
```

packages/builders/src/messages/Attachment.ts
```diff
@@ -85,6 +98,60 @@ export class AttachmentBuilder implements JSONEncodable<RESTAPIAttachment> {
 		return this;
 	}
 
+	/**
+	 * Sets the file data to upload with this attachment.
+	 *
+	 * @param data - The file data (Buffer, Uint8Array, or string)
```
almeidx:
```suggestion
	 * @param data - The file data
```

packages/builders/src/messages/Attachment.ts
```diff
@@ -85,6 +98,60 @@ export class AttachmentBuilder implements JSONEncodable<RESTAPIAttachment> {
...
+	 * @remarks Note that this data is NOT included in the `toJSON()` output. To retrieve, use {@link getRawFile}.
```
almeidx:
```suggestion
	 * @remarks Note that this data is NOT included in the {@link toJSON} output. To retrieve it, use {@link getRawFile}.
```

packages/builders/src/messages/Attachment.ts
```diff
@@ -85,6 +98,60 @@ export class AttachmentBuilder implements JSONEncodable<RESTAPIAttachment> {
...
+	 */
+	public clearFileData(): this {
+		this.fileData.data = undefined;
+		return this;
+	}
+
+	/**
+	 * Sets the content type of the file data to upload with this attachment.
+	 *
+	 * @remarks Note that this data is NOT included in the `toJSON()` output. To retrieve, use {@link getRawFile}.
```
almeidx:
```suggestion
	 * @remarks Note that this data is NOT included in the {@link toJSON} output. To retrieve it, use {@link getRawFile}.
```

packages/builders/src/messages/Attachment.ts
```diff
@@ -85,6 +98,60 @@ export class AttachmentBuilder implements JSONEncodable<RESTAPIAttachment> {
...
+	 */
+	public clearFileContentType(): this {
+		this.fileData.contentType = undefined;
+		return this;
+	}
+
+	/**
+	 * Converts this attachment to a RawFile for uploading.
+	 *
+	 * @returns A RawFile object, or undefined if no file data is set
```
almeidx:
```suggestion
	 * Converts this attachment to a {@link RawFile} for uploading.
	 *
	 * @returns A {@link RawFile} object, or `undefined` if no file data is set
```

packages/builders/src/messages/Assertions.ts
```diff
@@ -125,3 +136,11 @@ const messageComponentsV2Predicate = baseMessagePredicate.extend({
 });
 
 export const messagePredicate = z.union([messageNoComponentsV2Predicate, messageComponentsV2Predicate]);
+
+// This validator does not assert file.key <-> attachment.id coherence. This is fine, because the builders
+// should effectively guarantee that.
+export const fileBodyMessagePredicate = z.object({
+	body: messagePredicate,
+	// No min length to support message edits
+	files: rawFilePredicate.array(),
```
almeidx:
```suggestion
	files: rawFilePredicate.array().max(10),
```

packages/builders/src/messages/Assertions.ts
```diff
@@ -1,9 +1,20 @@
+import { Buffer } from 'node:buffer';
 import { AllowedMentionsTypes, ComponentType, MessageFlags, MessageReferenceType } from 'discord-api-types/v10';
 import { z } from 'zod';
 import { embedPredicate } from './embed/Assertions.js';
 import { pollPredicate } from './poll/Assertions.js';
 
+const fileKeyRegex = /^files\[(?<placeholder>\d+?)]$/;
```
almeidx, line 7: According to the spec, you can only use keys 0-9 here
https://github.com/discord/discord-api-spec/blob/28ef48483b4745bcf176b2b53b931a27142e2f55/specs/openapi.json#L2815-L2858


```suggestion
const fileKeyRegex = /^files\[(?<placeholder>\d)]$/;
```

Jiralite:
> I'm assuming there will be a follow-up pull request to add support for this in mainlib

Yeah, especially because #11273.

packages/builders/src/messages/Assertions.ts
vladfrangu: Technically true, you can only upload 10 images at a time, but i thought you can provide any value in the `files[<key>]` part, up to 10 elements tho

Jiralite: Gonna link what I said here: https://github.com/discordjs/discord.js/pull/11248#discussion_r2528955051

Their documentation says you can even pass `files[42]`, so...

didinele: Yeah, I'm not sure what's correct here, spec or docs.

didinele: See https://discord.com/channels/222078108977594368/999796393810067607/1439881773172527125 (internals)

We're leaving this as is

packages/builders/src/messages/Message.ts
almeidx: I think we should get rid of it. Especially considering the implementation of `isFileBodyEncodable()`.
In the end it doesn't matter all that much regardless

packages/builders/src/messages/Message.ts
```diff
@@ -56,7 +56,7 @@ export interface MessageBuilderData
 /**
  * A builder that creates API-compatible JSON data for messages.
  */
-export class MessageBuilder implements JSONEncodable<RESTPostAPIChannelMessageJSONBody> {
+export class MessageBuilder implements FileBodyEncodable<RESTPostAPIChannelMessageJSONBody> {
```
vladfrangu: Should implement both I'd say

packages/util/src/encodables.ts
```diff
@@ -0,0 +1,63 @@
...
+}
+
+/**
+ * Represents an object capable of representing itself as a request body with file attachments.
+ * Objects implementing this interface can separate JSON body data from binary file data,
+ * which is necessary for multipart/form-data requests.
+ *
+ * @typeParam BodyValue - The JSON body type
+ */
+export interface FileBodyEncodable<BodyValue> extends JSONEncodable<BodyValue> {
```
vladfrangu: I would remove the extends JSONEncodable from here since imo these two interfaces serve two separate purposes

packages/builders/src/messages/Message.ts
didinele: https://github.com/discordjs/discord.js/pull/11248#discussion_r2534490568

packages/util/src/encodables.ts
didinele: I can't think of any multipart form data endpoint that doesn't also take in json properties... they don't serve "different purposes", FileBodyEncodable fundamentally relies on `this.toJSON()`, the implementation being made in this PR behaves in this fashion too.

vladfrangu (approved):

PR #11250, opened 2025-11-05
fix: Remove data conditional check

Jiralite:
These should never be constructed without data...

The blame seems to go years and years back. Was likely ancient logic made redundant over time.


vakiliner:
Even though these constructors are private, typings needs to be updated
https://github.com/discordjs/discord.js/blob/7e5accfdb0233f58a7cd4ae9efc1b66a0f5769fb/packages/discord.js/typings/index.d.ts#L3527-L3528
https://github.com/discordjs/discord.js/blob/7e5accfdb0233f58a7cd4ae9efc1b66a0f5769fb/packages/discord.js/typings/index.d.ts#L3823-L3824

PR #11251, opened 2025-11-05
feat(core): handle request all guild members rate limit

didinele:


packages/util/src/gatewayRateLimitError.ts
```diff
@@ -0,0 +1,31 @@
+import type { GatewayOpcodeRateLimitMetadataMap, GatewayRateLimitedDispatchData } from 'discord-api-types/v10';
+
+/**
+ * Represents the error thrown when the gateway emits a RATE_LIMITED event after a certain request.
+ */
+export class GatewayRateLimitError extends Error {
+	public override readonly name = GatewayRateLimitError.name;
```
Jiralite, line 7: I don't think this override is necessary?

Image

didinele: image

packages/core/src/util/gatewayRateLimitError.ts
```diff
@@ -0,0 +1,17 @@
+import type { GatewayOpcodeRateLimitMetadataMap, GatewayRateLimitedDispatchData } from 'discord-api-types/v10';
+
+/**
+ * Represents the error thrown when the gateway emits a RATE_LIMITED event after a certain request.
```
Jiralite:
```suggestion
 * Represents the error thrown when the gateway emits a `RATE_LIMITED` event after a certain request.
```

packages/core/src/client.ts
```diff
@@ -244,6 +246,22 @@ export class Client extends AsyncEventEmitter<MappedEvents> {
 
 		let timer: NodeJS.Timeout | undefined = createTimer(controller, timeout);
 
+		const onRatelimit = (payload: GatewayDispatchPayload) => {
+			// We could verify meta.guild_id === options.guild_id as well, but really, the nonce check is enough
+			if (payload.t === GatewayDispatchEvents.RateLimited && payload.d.meta.nonce === nonce) {
+				controller.abort(new GatewayRateLimitError(payload.d));
```
Jiralite: One thing I'd like to see is what our request payload is. We do this with `DiscordAPIError`:

https://github.com/discordjs/discord.js/blob/5fa92a1ea90c6e83e6a49bc29bf89ee9f61f6d60/packages/rest/src/lib/errors/DiscordAPIError.ts#L66

So that would mean adding another parameter that exposes the data we sent, which would be `d` below (`{ ...options, nonce }`).

If you do agree with this, then I could argue further and say it could be added to @discordjs/ws so #11252 could make use of it too.

didinele: I don't really follow what you think should be added to ws? ws has no mechanism for any of this, other than just sending the request, there's no proper place to throw this error.
as such, exporting it from there makes little sense. at best we can maybe shove it in util?

Jiralite: I mean add this to the package:

https://github.com/discordjs/discord.js/blob/5fa92a1ea90c6e83e6a49bc29bf89ee9f61f6d60/packages/core/src/util/gatewayRateLimitError.ts#L6-L17

We can import that here and in discord.js and throw with it.

didinele: See https://discord.com/channels/222078108977594368/1406709301476851864/1436869888785387596

didinele: Pasting from the above for record-keeping. Relevant #11252 #11253

> I get where you're coming from though. we have 3 options:
a. current state of my PRs (generic DiscordjsError is thrown that doesn't contain the full information)
b. we duplicate the class I implemented in core in mainlib
c. we move my error implementation into @discordjs/util and mainlib imports it as well

We've agreed on c.

packages/util/package.json
```diff
@@ -68,6 +68,7 @@
 		"@types/node": "^22.18.13",
 		"@vitest/coverage-v8": "^3.2.4",
 		"cross-env": "^10.1.0",
+		"discord-api-types": "^0.38.31",
```
almeidx: Pretty sure this should be a regular dependency

didinele: why? I'm only importing a type and it wasn't used for anything previously

almeidx: Because `GatewayRateLimitedDispatchData` and `GatewayOpcodeRateLimitMetadataMap` are being used in the constructor, which will be in dist/index.d.ts

packages/util/src/gatewayRateLimitError.ts
```diff
@@ -0,0 +1,23 @@
...
+	public constructor(
+		/**
+		 * The data associated with the rate limit event
+		 */
+		public readonly data: GatewayRateLimitedDispatchData<keyof GatewayOpcodeRateLimitMetadataMap>,
+		/**
+		 * The payload data that lead to this rate limit
+		 */
+		// Too complicated to type properly here (i.e. extract the ['data']
+		// of event payloads that have t = keyof GatewayOpcodeRateLimitMetadataMap)
```
almeidx:
```suggestion
		/**
		 * The payload data that lead to this rate limit
		 *
		 * @privateRemarks
		 * Too complicated to type properly here (i.e. extract the ['data']
		 * of event payloads that have t = keyof GatewayOpcodeRateLimitMetadataMap)
		 */
```

packages/core/src/client.ts
```diff
@@ -244,6 +246,19 @@ export class Client extends AsyncEventEmitter<MappedEvents> {
 
 		let timer: NodeJS.Timeout | undefined = createTimer(controller, timeout);
 
+		const onRatelimit = ({ data }: ToEventProps<GatewayRateLimitedDispatchData>) => {
+			controller.abort(new GatewayRateLimitError(data, options));
+		};
```
vladfrangu, line 254: Pretty sure this listener is wrong, you should check for the data (nonce, etc)

almeidx: That check used to be there... not sure why, but it was removed in the last commit
https://github.com/discordjs/discord.js/pull/11251/commits/b4c4fffadb584c9713cc492ca4824b05585376d6

didinele: on accident, good catch

PR #11252, opened 2025-11-05
feat(GuildMemberManager): handle gateway request rate limit

didinele:


packages/discord.js/src/errors/ErrorCodes.js
```diff
@@ -135,6 +135,8 @@
  *
  * @property {'PermissionOverwritesTypeMandatory'} PermissionOverwritesTypeMandatory
  * @property {'PermissionOverwritesTypeMismatch'} PermissionOverwritesTypeMismatch
+ *
+ * @property {'GatewayRequestRateLimited'} GatewayRequestRateLimited
```
Jiralite: For visibility: https://github.com/discordjs/discord.js/pull/11251#discussion_r2501211302

packages/discord.js/src/managers/GuildMemberManager.js
```diff
@@ -246,46 +247,64 @@ class GuildMemberManager extends CachedManager {
...
+        if (payload.t === GatewayDispatchEvents.RateLimited && payload.d.meta.nonce === nonce) {
+          clearTimeout(timeout);
+          this.client.ws.off(WebSocketShardEvents.Dispatch, rateLimitHandler);
+          this.client.removeListener(Events.GuildMembersChunk, handler);
+          this.client.decrementMaxListeners();
+          reject(new DiscordjsError(ErrorCodes.GatewayRequestRateLimited, payload.d));
+        }
+      };
+
+      this.client.ws.on(WebSocketShardEvents.Dispatch, rateLimitHandler);
```
Qjuh, line 300: Shouldn't there be a `/src/client/websocket/handlers/RATE_LIMITED.js` for this websocket event so that
- end users can listen to it directly and
- you can use it here directly instead of having to filter the full dispatch event again 

Same applies for the backport where it would replace the usage of listening to `Events.Raw`

didinele: I considered it, my objection being
> end users can listen to it directly

should they (esp. if the behavior is a thrown error anyway when it happens)? would they want to? why?

Qjuh: Same reason we have `REST#on('rateLimited',...)` while also `rejectOnRatelimit` setting. So users can decide themselves, use it for debugging, and in general expose all events to the end user, which we do in next major anyway.

Jiralite:
Now that #11251 is merged, time to use the new class from the shared package!

packages/discord.js/src/managers/GuildMemberManager.js
didinele: In #11251 we decided to share an error class between core and util. Naturally, that expects raw API data.

While ultimately I agree with adding the rate limit event for end users, constructing the error class becomes non-trivial after the conversion step in the RATE_LIMITED.js handler. As such, the handling here will continue listening to `Dispatch`/`Raw` (v14)

packages/discord.js/src/util/Events.js
```diff
@@ -156,6 +157,7 @@ exports.Events = {
   MessageReactionRemoveEmoji: 'messageReactionRemoveEmoji',
   MessageUpdate: 'messageUpdate',
   PresenceUpdate: 'presenceUpdate',
+  RateLimited: 'rateLimited',
```
Qjuh: This needs to be added in typings too.

packages/discord.js/src/managers/GuildMemberManager.js
```diff
@@ -246,46 +248,65 @@ class GuildMemberManager extends CachedManager {
...
-        },
-      });
       const fetchedMembers = new Collection();
       let index = 0;
+
+      const cleanup = () => {
+        /* eslint-disable no-use-before-define */
+        clearTimeout(timeout);
+
+        this.client.ws.off(WebSocketShardEvents.Dispatch, rateLimitHandler);
```
Qjuh:
```suggestion
        this.client.ws.removeListener(WebSocketShardEvents.Dispatch, rateLimitHandler);
```
Consistency I guess...

packages/discord.js/src/client/websocket/handlers/RATE_LIMITED.js
```diff
@@ -0,0 +1,65 @@
...
+        nonce: data.nonce,
+      };
+
+      break;
+    }
+
+    default: {
+      if (!emittedWarning) {
+        process.emitWarning(`Received a rate limit for an unknown opcode: ${data.opcode}`);
+        emittedWarning = true;
```
Qjuh: Wouldn't we want this warning to emit again if another unknown ratelimit happens?

packages/discord.js/src/managers/GuildMemberManager.js
didinele: I wanted both to be off actually, but with a quick search through the codebase, it seems we're doing removeListener a bunch.

packages/discord.js/src/client/websocket/handlers/RATE_LIMITED.js
didinele: My thought process was that if we have a type that's unimplemented, that will lead to the request hanging (see how request guild members currently behaves right now without this PR merged), so warning at least once will make sure the user is at least aware this could be happening if they don't have a listener.

Now that you ask, I'd go even further and do `!emittedWarning && !listenerCount(...)`. Thoughts?

didinele:
Important internal convo https://discord.com/channels/222078108977594368/999796393810067607/1438120490920054784

We're currently torn on introducing the ratelimited event or not.

didinele:
We've decided to ultimately not introduce Client#ratelimited.

packages/discord.js/src/client/websocket/handlers/RATE_LIMITED.js
```diff
@@ -0,0 +1,27 @@
...
+      break;
+    }
+
+    default: {
+      if (!emittedFor.has(data.opcode)) {
+        const warning = new GatewayRateLimitError(data, '<unknown>');
+        process.emitWarning(
+          'Received a rate limit for an unknown opcode. The error below contains information about the rate limit. Please open an issue on GitHub',
+        );
+        process.emitWarning(warning);
```
Jiralite: This is only going to log the string (i.e. `GatewayRateLimitError: Request with opcode X was rate limited. Retry after Y seconds.`), is that intended? It sounds like we'd want to log the object (we lose logging `meta`).

packages/discord.js/src/client/websocket/handlers/RATE_LIMITED.js
```diff
@@ -0,0 +1,24 @@
...
+module.exports = (client, { d: data }) => {
+  switch (data.opcode) {
+    case GatewayOpcodes.RequestGuildMembers: {
+      break;
+    }
+
+    default: {
+      if (!emittedFor.has(data.opcode)) {
+        process.emitWarning(
+          `Hit a gateway rate limit on opcode ${data.opcode}. If your library is up-to-date, please open an issue on GitHub.`,
```
vladfrangu:
```suggestion
          `Hit a gateway rate limit on opcode ${data.opcode} (${GatewayOpcodes[data.opcode]}). If the discord.js version you're using is up-to-date, please open an issue on GitHub.`,
```

Issue #8733, opened 2022-10-11
WebookClient/REST memory leak


gc:
### Which package is this bug report for?

discord.js

### Issue description

I was inspecting my bots memory dump to track down whats using memory, and noticed I have 77,000 instances of `REST` in memory, even though I don't make them myself, nor hold any references to them or webhookclients in my code.  (fyi: D.JS makes a REST for the main client, and then a new one for every new WebhookClient). These seem to never be freed from memory

1. Instantiate some WebhookClient's
2. Let them fall out of scope, and no longer be referenced
3. Observe they still use memory.

These images are referenced in the code sample section below
![chrome_iqXzVWA90S](https://user-images.githubusercontent.com/30398469/195053716-bec49849-5e5b-40bf-be0f-37c6d3754d57.png)

![image](https://user-images.githubusercontent.com/30398469/195053842-b2fd513f-4aac-46bc-bb94-508a17d18e9f.png)

### Code sample

```typescript
Repo/sample: https://github.com/gc/memleakrepo
`git clone git@github.com:gc/memleakrepo.git`
`yarn && yarn start`


After running this, you'll get a heap dump. Open it in chrome, and observe exactly 1000x REST objects in memory being retained, and 0x SomeClass.
See above for images showing this.
```

### Package version

14.6.0

### Node.js version

16.9.0

### Operating system

Windows

### Priority this issue should have

Medium (should be fixed soon)

### Which partials do you have configured?

No Partials

### Which gateway intents are you subscribing to?

Not applicable (subpackage bug)

gc:
Commenting discussion with kyra so its documented here:
kyra says the cause is this: https://github.com/discordjs/discord.js/blob/main/packages/rest/src/lib/RequestManager.ts#L240
and the onus is on the user to call `.destroy()` on webhookclients to free the memory because theres no way to fix it.
I say, surely theres a way to fix it if its just a setInterval - but otherwise lets document it clearly so its known you have to destroy it when you're done to prevent a silent memory leak.

kyranet:
To give more context, `WebhookClient` (which extends `BaseClient`) keeps a `REST` instance, which sets up those two timers:

https://github.com/discordjs/discord.js/blob/f92be4fb944e69ae1b7cf667d99e434ad2ed7491/packages/rest/src/lib/RequestManager.ts#L240
[...]

PR #11266, opened 2025-11-10
refactor(Webhooks)!: remove WebhookClient

ckohen:
BREAKING CHANGE: WebhookClient has been removed, use @discordjs/core instead or fetch webhooks. Alternative solutions are in the works

WebhookClient is an incredibly inefficient way to interact with webhooks (creates a rest instance per webhook, an entire client per webhook etc..). We are working on developing an alternative to this that will mimic the simplicity of WebhookClient. In the meantime, we recommend using @discordjs/core if you are not connected to gateway, and fetching webhooks from the client if you are. We know making an extra API call is not ideal, the alternative will solve this and will land as soon as we finish internal discussions.

Closes: #8733 


packages/discord.js/src/structures/MessagePayload.js
```diff
@@ -47,15 +47,14 @@ class MessagePayload {
...
 
   /**
-   * Whether or not the target is a {@link Webhook} or a {@link WebhookClient}
+   * Whether or not the target is a {@link Webhook}
    *
    * @type {boolean}
    * @readonly
    */
   get isWebhook() {
     const { Webhook } = require('./Webhook.js');
```
vladfrangu, line 56: todo: use lazy here

ckohen:
I decided not to unify the client class in this PR. I think we should in a PR following this one though.

packages/discord.js/src/client/WebhookClient.js
```diff
@@ -1,119 +0,0 @@
...
-   */
-  constructor(data, options) {
-    super(options);
-    Object.defineProperty(this, 'client', { value: this });
-    let { id, token } = data;
-
-    if ('url' in data) {
-      const parsed = parseWebhookURL(data.url);
-      if (!parsed) {
-        throw new DiscordjsError(ErrorCodes.WebhookURLInvalid);
```
Qjuh, line 56: This error is used nowhere else, should probably be removed from `/errors/ErrorCodes.js` and `/errors/Messages.js` and typings now.

packages/discord.js/src/util/Util.js
```diff
@@ -469,11 +469,19 @@ function cleanCodeBlockContent(text) {
...
+ * @property {Snowflake} id The webhook's id
+ * @property {string} token The webhook's token
+ */
+
 /**
  * Parses a webhook URL for the id and token.
  *
  * @param {string} url The URL to parse
- * @returns {?WebhookClientDataIdWithToken} `null` if the URL is invalid, otherwise the id and the token
+ * @returns {?WebhookDataIdWithToken} `null` if the URL is invalid, otherwise the id and the token
```
Qjuh, line 484: This utility helper was only used for `WebhookClient`, do we want to keep it anyway for future use/user convenience?

Jiralite: I'd say we can keep this for user convenience.

PR #11274, opened 2025-11-12
refactor(client)!: remove `BaseClient`

### LINKED ISSUE #11271: Remove BaseClient from discord.js
Getting a bit hard to track stuff in the internals channel, so I'm making a few issues

After #11266 `Client` will become the only class extending `BaseClient`. We should be merging them

sdanialraza:
Resolves #11271

BREAKING CHANGE: `BaseClient` has been removed and its functionality has been merged into `Client`.

Depends on:
- #11266

### DIFF
--- a/packages/discord.js/src/client/websocket/handlers/RATE_LIMITED.js
+++ b/packages/discord.js/src/client/websocket/handlers/RATE_LIMITED.js
@@ -1,0 +1,24 @@
+'use strict';
+
+const process = require('node:process');
+const { GatewayOpcodes } = require('discord-api-types/v10');
+
+const emittedFor = new Set();
+
+module.exports = (client, { d: data }) => {
+  switch (data.opcode) {
+    case GatewayOpcodes.RequestGuildMembers: {
+      break;
+    }
+
+    default: {
+      if (!emittedFor.has(data.opcode)) {
+        process.emitWarning(
+          `Hit a gateway rate limit on opcode ${data.opcode} (${GatewayOpcodes[data.opcode]}). If the discord.js version you're using is up-to-date, please open an issue on GitHub.`,
+        );
+
+        emittedFor.add(data.opcode);
+      }
+    }
+  }
+};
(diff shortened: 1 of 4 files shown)


packages/discord.js/src/client/Client.js
```diff
@@ -47,14 +48,57 @@ const BeforeReadyWhitelist = [
...
-  constructor(options) {
-    super(options);
+  constructor(options = {}) {
+    super();
+
+    if (typeof options !== 'object' || options === null) {
+      throw new DiscordjsTypeError(ErrorCodes.InvalidType, 'options', 'object', true);
+    }
+
+    const defaultOptions = Options.createDefault();
```
Qjuh, line 64: This is now a duplicate of the variable `defaults` defined later on and should be done only once.

packages/discord.js/src/client/Client.js
```diff
@@ -700,11 +788,8 @@ class Client extends BaseClient {
     return `${this.options.rest.api}${Routes.oauth2Authorization()}?${query}`;
   }
 
-  toJSON() {
-    return super.toJSON({
-      actions: false,
-      presence: false,
-    });
+  toJSON(...props) {
```
Qjuh: Considering we don't need nor want to be able to pass more properties to `toJSON()` this can probably be simplified by removing `...props’ here and on next line

packages/discord.js/typings/index.d.ts
```diff
@@ -968,7 +962,8 @@ export class Client<Ready extends boolean = boolean> extends BaseClient<ClientEv
   public generateInvite(options?: InviteGenerationOptions): string;
   public login(token?: string): Promise<string>;
   public isReady(): this is Client<true>;
-  public toJSON(): unknown;
+  public toJSON(...props: Record<string, boolean | string>[]): unknown;
```
Qjuh: Ditto

packages/discord.js/src/client/Client.js
```diff
@@ -47,14 +48,57 @@ const BeforeReadyWhitelist = [
...
+ * @extends {AsyncEventEmitter}
  */
-class Client extends BaseClient {
+class Client extends AsyncEventEmitter {
   /**
    * @param {ClientOptions} options Options for the client
    */
-  constructor(options) {
-    super(options);
+  constructor(options = {}) {
```
Qjuh: `options` are a required parameter for Client constructor (in particular the `intents` are), so why have a default value defined here?

Qjuh (changes requested):

PR #11278, opened 2025-11-17
refactor!: remove AttachmentBuilder and support new file body encodables

### LINKED ISSUE #11273: Remove AttachmentBuilder from discord.js
Especially after #11248, we need to ensure discord.js is well-inline with the current state of builders.

didinele:
- [x] Depends on #11248 

Closes #11273


didinele:
Ready for review, but blocked until the other PR is merged.

packages/discord.js/typings/index.test-d.ts
```diff
@@ -3034,14 +3028,11 @@ await guildScheduledEventManager.edit(snowflake, { recurrenceRule: null });
...
 
-await textChannel.send({
-  files: [
-    new AttachmentBuilder('https://example.com/voice-message.ogg')
-      .setDuration(2)
-      .setWaveform('AFUqPDw3Eg2hh4+gopOYj4xthU4='),
-  ],
-  flags: MessageFlags.IsVoiceMessage,
-});
+await textChannel.send(
```
Qjuh, line 3031: Shouldn't we keep a test for sending a voice message instead of replacing this with a plain text file?

packages/discord.js/src/managers/MessageManager.js
```diff
@@ -223,21 +224,25 @@ class MessageManager extends CachedManager {
...
-    const data = await this.client.rest.patch(Routes.channelMessage(this.channel.id, messageId), { body, files });
+    let payload;
+    if (options instanceof MessagePayload) {
+      payload = await options.resolveBody().resolveFiles();
+    } else if (isFileBodyEncodable(options)) {
+      payload = options.toFileBody();
+    } else if (isJSONEncodable(options)) {
+      payload = { body: options.toJSON(), files: [] };
+    } else {
+      payload = await MessagePayload.create(this, options).resolveBody().resolveFiles();
```
vladfrangu: Unless the new code is correct

```suggestion
      payload = await MessagePayload.create(message instanceof Message ? message : this, options).resolveBody().resolveFiles();
```

packages/discord.js/typings/index.test-d.ts
vladfrangu: Both ideally (tho this is a type test not a runtime test)

If possible, updating one of the runtime test files would be appreciated

didinele: yea that was an accident, ty

> If possible, updating one of the runtime test files would be appreciated

are those actually.. maintained at all?

packages/discord.js/typings/index.d.ts
```diff
@@ -6688,14 +6690,7 @@ export interface BaseMessageOptions {
...
   embeds?: readonly (APIEmbed | JSONEncodable<APIEmbed>)[];
-  files?: readonly (
-    | Attachment
-    | AttachmentBuilder
-    | AttachmentPayload
-    | BufferResolvable
-    | JSONEncodable<APIAttachment>
-    | Stream
-  )[];
+  files?: readonly (Attachment | AttachmentPayload | BufferResolvable | JSONEncodable<APIAttachment> | Stream)[];
```
Jiralite: `JSONEncodable<APIAttachment>` doesn't work:

https://github.com/discordjs/discord.js/blob/77241ad7c59d622428f825ba6e4d862f46d2adbb/packages/discord.js/src/structures/MessagePayload.js#L271-L279

`attachment = fileLike.attachment` resolves to `undefined` since there's no `attachment` property. This then follows:

https://github.com/discordjs/discord.js/blob/77241ad7c59d622428f825ba6e4d862f46d2adbb/packages/discord.js/src/util/Util.js#L413-L415

Which crashes:

```
TypeError: Cannot read properties of undefined (reading 'path')
    at findName (/Users/jiralite/Documents/GitHub/discord.js/discord.js/packages/discord.js/src/util/Util.js:413:13)
    at MessagePayload.resolveFile (/Users/jiralite/Documents/GitHub/discord.js/discord.js/packages/discord.js/src/structures/MessagePayload.js:278:31)
    at /Users/jiralite/Documents/GitHub/discord.js/discord.js/packages/discord.js/src/structures/MessagePayload.js:257:85
    at Array.map (<anonymous>)
    at MessagePayload.resolveFiles (/Users/jiralite/Documents/GitHub/discord.js/discord.js/packages/discord.js/src/structures/MessagePayload.js:257:56)
    at ChannelManager.createMessage (/Users/jiralite/Documents/GitHub/discord.js/discord.js/packages/discord.js/src/managers/ChannelManager.js:186:74)
    at TextChannel.send (/Users/jiralite/Documents/GitHub/discord.js/discord.js/packages/discord.js/src/structures/interfaces/TextBasedChannel.js:186:33)
    at Client.<anonymous> (/Users/jiralite/Documents/GitHub/Soulobby/Soulobby/source/test.ts:34:83)
    at Client.emit (/Users/jiralite/Documents/GitHub/discord.js/discord.js/node_modules/.pnpm/@vladfrangu+async_event_emitter@2.4.7/node_modules/@vladfrangu/async_event_emitter/src/index.ts:524:28)
    at Client._triggerClientReady (/Users/jiralite/Documents/GitHub/discord.js/discord.js/packages/discord.js/src/client/Client.js:464:10)
```

Jiralite (changes requested):

packages/discord.js/typings/index.d.ts
didinele: well, I didn't add that :), but you're right

packages/discord.js/typings/index.test-d.ts
didinele: See https://discord.com/channels/222078108977594368/1209960895141388389/1441019125437694055

This didn't work at runtime prior to this PR, the types were incorrect. With the current state of things, it's impossible to make it work either, so goodbye.

The one below (not included in the diff, since I didn't touch it) still has sending but without the builder

packages/discord.js/src/managers/ChannelManager.js
```diff
@@ -174,18 +174,21 @@ class ChannelManager extends CachedManager {
...
-    let messagePayload;
+    let payload;
 
     if (options instanceof MessagePayload) {
-      messagePayload = options.resolveBody();
+      payload = await options.resolveBody().resolveFiles();
+    } else if (isFileBodyEncodable(options)) {
+      payload = options.toFileBody();
+    } else if (isJSONEncodable(options)) {
+      payload = { body: options.toJSON(), files: [] };
```
Jiralite:
```suggestion
      payload = { body: options.toJSON() };
```

packages/discord.js/src/managers/MessageManager.js
```diff
@@ -223,21 +224,27 @@ class MessageManager extends CachedManager {
...
-      .resolveBody()
-      .resolveFiles();
-    const data = await this.client.rest.patch(Routes.channelMessage(this.channel.id, messageId), { body, files });
+    let payload;
+    if (options instanceof MessagePayload) {
+      payload = await options.resolveBody().resolveFiles();
+    } else if (isFileBodyEncodable(options)) {
+      payload = options.toFileBody();
+    } else if (isJSONEncodable(options)) {
+      payload = { body: options.toJSON(), files: [] };
```
Jiralite:
```suggestion
      payload = { body: options.toJSON() };
```

packages/discord.js/typings/index.test-d.ts
```diff
@@ -3034,14 +3028,11 @@ await guildScheduledEventManager.edit(snowflake, { recurrenceRule: null });
...
+  new MessageBuilder()
+    .setContent(':)')
+    .addAttachments(attachment => attachment.setFileData(':)').setFilename('smiley.txt')),
+);
```
Jiralite, line 3035: Please add `setId()` for completeness!

packages/builders/src/messages/Attachment.ts
```diff
@@ -28,6 +28,11 @@ export class AttachmentBuilder implements JSONEncodable<RESTAPIAttachment> {
 	 * Creates a new attachment builder.
 	 *
 	 * @param data - The API data to create this attachment with
+	 * @example
+	 * ```ts
+	 * const attachment = new AttachmentBuilder().setId(1).setFileData(':)').setFilename('smiley.txt')
+	 * ```
+	 * @remarks <warn>Please note that the `id` field is required, it's rather easy to miss!</warn>
```
Jiralite: This only works for the legacy website, which this package does not have documentation for.

```suggestion
	 * @remarks Please note that the `id` field is required, it's rather easy to miss!
```

didinele: Oh. I searched the codebase and noticed there was one even in the util package so I assumed it worked...

packages/discord.js/src/structures/interfaces/TextBasedChannel.js
```diff
@@ -88,7 +88,7 @@ class TextBasedChannel {
    * @property {Array<(EmbedBuilder|Embed|APIEmbed)>} [embeds] The embeds for the message
    * @property {MessageMentionOptions} [allowedMentions] Which mentions should be parsed from the message content
    * (see {@link https://discord.com/developers/docs/resources/message#allowed-mentions-object here} for more details)
-   * @property {Array<(AttachmentBuilder|Attachment|AttachmentPayload|BufferResolvable)>} [files]
+   * @property {Array<(Attachment|AttachmentPayload|BufferResolvable)>} [files]
```
Qjuh: I'd prefer if there was still a way to pass an AttachmentBuilder into here, albeit not sure what type would make sense to put for that here. 

Another helper type like `RawFileEncodable` which extends `JSONEncodable` with the additional `getRawFile()` method? Seems clunky but 🤷‍♂️

didinele: The main issue I have with this is that AttachmentBuilder includes data tied to.. well, the `attachments` array.

What do we do when the user sets that data and they also provide `attachments` manually?

Qjuh: The same we did before:
https://github.com/discordjs/discord.js/blob/02fc101069aed994f7ee4579d114706047025ddd/packages/discord.js/src/structures/MessagePayload.js#L193-L204

packages/builders/src/messages/Attachment.ts
```diff
@@ -28,6 +28,11 @@ export class AttachmentBuilder implements JSONEncodable<RESTAPIAttachment> {
 	 * Creates a new attachment builder.
 	 *
 	 * @param data - The API data to create this attachment with
+	 * @example
+	 * ```ts
+	 * const attachment = new AttachmentBuilder().setId(1).setFileData(':)').setFilename('smiley.txt')
+	 * ```
+	 * @remarks Please note that the `id` field is required, it's rather easy to miss!
```
vladfrangu, line 35: If it's required, make it throw on toJSON

vladfrangu: and if it already does, SORRY

didinele: of course it does in the validator already, the remark is there because it's super not obvious considering older APIs, I spoke about this with Jira briefly in internals

packages/discord.js/src/structures/interfaces/TextBasedChannel.js
didinele: That is absolutely awful in terms of library behavior. That snippet effectively means that if you pass both `files` and `attachments` at the same time, it's a guaranteed API error, unless it's a message edit and the elements the user passed to the `attachments` array specifically refer to existing attachments.

In essence, this renders the `attachments` field as edit-only, we're just completely taking over it with this behavior, largely hiding how the API payload works too (which I am not fond of at all, no matter how rich the the abstraction).

Qjuh: But that's what AttachmentBuilder does too... it hides the same things the same way.

didinele: No. This isn't about hiding things, it's allowing the user to easily pass garbage. AttachmentBuilder on its own is fundamentally different from what you want, it removes your granular control of the files/attachments fields, it ties them together completely. There is no way when using it to get non-sense miss-matched arrays. All the more so when you use MessageBuilder

didinele: Circling back on this, the only way I'd accept it is making it a union type that forbids `files` if `attachments` is `AttachmentBuilder[]`, I'm not compromising otherwise

Qjuh: Thing is that all the other types allowed for `files` here will also fill the `attachments`. So if anything it would make sense to completely disallow `attachments` in create Message options.

And you got it backwards: it's an accepted type for the `files` property and will disallow the `attachments` property.

didinele: Sure. I'm okay with designing the API that way, but we do also have to acknowledge the complexity it adds

packages/discord.js/src/structures/MessagePayload.js
```diff
@@ -197,8 +197,13 @@ class MessagePayload {
...
       duration_secs: file.duration,
     }));
+
+    // Only passable during edits
     if (Array.isArray(this.options.attachments)) {
-      this.options.attachments.push(...(attachments ?? []));
+      this.options.attachments = this.options.attachments
+        // Note how we don't check for file body encodable, since we aren't expecting file data here
+        .map(attachment => (isJSONEncodable(attachment) ? attachment.toJSON() : attachment))
+        .concat(attachments ?? []);
```
Qjuh: I feel like now that we overwrite `this.options.attachments` either way it should probably not be overwriting the options at all and instead be a const we assign or make `attachments` a let and modify in the `if` case.

Feels wrong to use `this.options.attachments` in the returned payload when it never will be the actual passed value at that point.

packages/discord.js/src/util/APITypes.js
```diff
@@ -683,3 +683,13 @@
...
+
+/**
+ * @external RESTPostAPIChannelMessageJSONBody
+ * @see {@link https://discord-api-types.dev/api/discord-api-types-v10/interface/RESTPostAPIChannelMessageJSONBody}
+ */
+
+/**
+ * @external RESTPatchAPIChannelMessageJSONBody
+ * @see {@link https://discord-api-types.dev/api/discord-api-types-v10/interface/RESTPatchAPIChannelMessageJSONBody}
+ */
```
Jiralite, line 695: Can you alphabetise this? This file is alphabetised.

PR #11279, opened 2025-11-17
chore: remove usage of the prefer-global rules

### LINKED ISSUE #11272: Remove usage of n/prefer-global rules across the monorepo
For runtime compatibility purposes.

didinele:
should I have ignored api-extractor?

Closes #11272


packages/create-discord-bot/bin/index.ts
```diff
@@ -1,7 +1,3 @@
-#!/usr/bin/env node
```
Qjuh, line 1: why was the shebang removed?

packages/scripts/src/generateIndex.ts
```diff
@@ -1,6 +1,5 @@
 import { stat, mkdir, writeFile } from 'node:fs/promises';
 import { join } from 'node:path';
-import { cwd } from 'node:process';
```
Qjuh, line 3: I'd keep this one here

packages/scripts/src/generateSplitDocumentation.ts
```diff
@@ -1,6 +1,5 @@
 import { mkdir, stat, writeFile } from 'node:fs/promises';
 import { join } from 'node:path';
-import { cwd } from 'node:process';
```
Qjuh, line 3: ditto

packages/scripts/src/populateDevDatabaseBranch.ts
```diff
@@ -1,5 +1,4 @@
 import { readFile } from 'node:fs/promises';
-import process, { cwd } from 'node:process';
```
Qjuh, line 2: ditto, for the destructure

Qjuh (commented):
including api-extractor in this change is correct

packages/create-discord-bot/bin/index.ts
didinele: I'm not sure, an eslint rule fired on it. Gonna look into it

packages/scripts/src/generateIndex.ts
didinele: Requires disabling the rule for that line, are we okay with that? The alternative is
```ts
const { cwd } = process;
```

but that also requires a rule disable because of unbound methods

packages/create-discord-bot/bin/index.ts
didinele: Looks like those files had a bunch of lint errors but eslint never really ran on them. Fixed it up

packages/scripts/src/generateIndex.ts
Qjuh: In that case I retract my comment, your change seems fine.

Jiralite (approved):

almeidx (changes requested):
When I suggested removing the imports from built-ins, I was referring to standard globals (`URL`, `URLSearchParams`, `Blob`, etc.).

I think we should keep the node-specific things imported still. While they will (probably) work on other runtimes without the explicit import, I think it's better in this case to have an explicit import so it's extremely clear that we are expecting the implementation from/for node.

***

Additionally, I found some other globals which are still being imported while skimming over: `File`, `Blob`, and `performance`.

didinele:
> Additionally, I found some other globals which are still being imported while skimming over: File, Blob, and performance.

Unfortunately, there's no ESLint rules for those. Even so

- I could not find any instances of `File` being imported. Maybe my search was improper?
- `Blob` in only imported within `REST` once for undici request, removing it poses the following problem:

```ts
Argument of type 'Blob' is not assignable to parameter of type 'BodyInit | undefined'.
  Property 'slice' is missing in type 'Blob' but required in type 'import("buffer").Blob'.
```

where `BodyInit` is a type referenced from undici. Probably for the best we keep that import

- `performance` ✅

packages/ws/src/ws/WebSocketShard.ts
```diff
@@ -2,8 +2,6 @@ import { Buffer } from 'node:buffer';
 import { once } from 'node:events';
 import { clearInterval, clearTimeout, setInterval, setTimeout } from 'node:timers';
```
almeidx: Let's remove these too. The exception being imports from `node:timers/promises`

packages/voice/__mocks__/rtp.ts
```diff
@@ -1,5 +1,3 @@
-import { Buffer } from 'node:buffer';
```
almeidx, line 1: This should stay

packages/create-discord-bot/template/JavaScript/src/index.js
```diff
@@ -1,5 +1,3 @@
-import process from 'node:process';
```
almeidx, line 1: This should stay

packages/create-discord-bot/template/JavaScript/src/util/deploy.js
```diff
@@ -1,5 +1,3 @@
-import process from 'node:process';
```
almeidx, line 1: This should stay

packages/create-discord-bot/bin/index.ts
```diff
@@ -1,7 +1,5 @@
 #!/usr/bin/env node
 
-// eslint-disable-next-line n/shebang
-import process from 'node:process';
```
almeidx, line 4: This should stay

packages/discord.js/test/messages.js
```diff
@@ -1,10 +1,8 @@
 'use strict';
 
-const { Buffer } = require('node:buffer');
 const { createReadStream } = require('node:fs');
 const { readFile } = require('node:fs/promises');
 const path = require('node:path');
-const process = require('node:process');
```
almeidx, line 7: These should stay

packages/discord.js/test/sendtest.js
```diff
@@ -1,15 +1,13 @@
 'use strict';
 
-const { Buffer } = require('node:buffer');
 const fs = require('node:fs');
 const path = require('node:path');
-const process = require('node:process');
```
almeidx, line 6: These should stay

packages/discord.js/test/shard.js
```diff
@@ -1,10 +1,9 @@
 'use strict';
 
-const process = require('node:process');
```
almeidx, line 3: This should stay

packages/discord.js/test/sharder.js
```diff
@@ -1,6 +1,5 @@
 'use strict';
 
-const process = require('node:process');
```
almeidx, line 3: This should stay

packages/discord.js/test/tester1000.js
```diff
@@ -1,6 +1,5 @@
 'use strict';
 
-const process = require('node:process');
```
almeidx, line 3: This should stay

packages/discord.js/test/tester2000.js
```diff
@@ -1,6 +1,5 @@
 'use strict';
 
-const process = require('node:process');
```
almeidx, line 3: This should stay

packages/discord.js/test/webhooktest.js
```diff
@@ -1,6 +1,5 @@
 'use strict';
 
-const { Buffer } = require('node:buffer');
```
almeidx, line 3: This should stay

packages/voice/__mocks__/rtp.ts
```diff
@@ -1,9 +1,9 @@
...
-
 // The following constants are silence packets collected from various platforms because Discord did not previously send header extensions
 // The header extension (extra data in decrypted vs opusFrame) can be detected in the position of {encrypted.subarray(12,14)} if it is equal to 0xbe,0xde
 // The header extension length will then follow as an integer and can be removed from the decrypted data (see ../src/receive/VoiceReceiver.ts:parsePacket)
 
+import { Buffer } from 'node:buffer';
```
almeidx: Should be at the top of the file, where it used to be. Otherwise LGTM

didinele:
@coderabbitai review

iCrawl (approved):

PR #11287, opened 2025-11-20
fix: typing mismatch for empty strings being treated as undefined

WhatCats:
## Problem
At the moment methods like `GuildMemberManager:fetch` that have an optional string parameter are doing a falsy check for if the parameter was omitted. But consider the following typescript code:
```ts
const id: string = ""
const member: GuildMember = await guild.members.fetch(id)
```
According to the typings `member` should be a `GuildMember` but its actually a `Collection<Snowflake, GuildMember>` because the empty string is falsy (this is why we love JS).

## My Suggestion
Do more exact undefined/null checks.
(I might not have found every place this is an issue)


almeidx:
Personally, I think this is a step in the right direction. However, I don't think this change alone is enough, as it will potentially bring different issues. For instance, passing an empty string to `RoleManager#fetch()` will now make a request to the Get Guild Role**s** endpoint, but through the Get Guild Role path internally. This will mean the endpoint will return a different body than expected, which might create worse errors that are harder to pinpoint.

To fix that, we should make further changes here. For instance, modifying the `resolveId()` method of managers to return `null` for empty strings. This solves the issue I mentioned in most places, but there are still some places that do not make use of this method (e.g., the RoleManager), so those places would have to be modified to make use of it.

WhatCats:
> For instance, passing an empty string to `RoleManager#fetch()` will now make a request to the Get Guild Role**s** endpoint, but through the Get Guild Role path internally. 

I think I understand so the original idea was that you would query the empty string with Discord but your saying since the endpoint is `/guild_id/roles/role_id` if you pass an empty string it would result  in `/guild_id/roles/` which is a different endpoint.

*Side Note: This makes me think the REST package shouldn't allow empty strings so you don't accidentally call a different endpoint.*

> To fix that, we should make further changes here. For instance, modifying the `resolveId()` method of managers to return `null` for empty strings.

I did consider this but then you would still end up explicitly calling the fetch-all endpoint for empty strings which is a problem as I originally described. 

I'm thinking maybe it would make sense to throw a Discord like error for empty strings manually (like you would get if you passed any other non-snowflake string to the API because we can't actually pass the empty string because it would be interpreted as a different endpoint).

almeidx:
> I'm thinking maybe it would make sense to throw a Discord like error for empty strings manually (like you would get if you passed any other non-snowflake string to the API because we can't actually pass the empty string because it would be interpreted as a different endpoint).

We *could* throw an error in `resolveId()` if the input is a primitive value that is not a non-empty string

WhatCats:
~Also, I just realized it would be very easy to accidentally pass unsanitized user input into the URL since IDs are represented as strings. For example, if you want to check whether a user exists, you might naturally write `users.fetch(input)`. The fact that input is ultimately interpolated into the URL is essentially an implementation detail, so it’s easy for someone using the API not to realize they’re introducing unsanitized input into a URL path. So I am thinking the routes should also do `encodeURIComponent` or some sort of snowflake test and fast fail.~

Correction:
- The discord-api-types Routes already handle path sanitization.
- Passing empty strings to the Routes is fine because:
```ts
console.log(Routes.guildRole("123", ""))
```
resolves to `https://discord.com/api/v10/guilds/123/roles/`
instead of `https://discord.com/api/v10/guilds/123/roles`
which gets the expected Not Found result (if a valid Guild ID is supplied)  instead of fetching all the guild roles.

WhatCats:
> For instance, passing an empty string to RoleManager#fetch() will now make a request to the Get Guild Roles endpoint, but through the Get Guild Role path internally. This will mean the endpoint will return a different body than expected, which might create worse errors that are harder to pinpoint.

@almeidx  I believe this is resolved by:
```ts
console.log(Routes.guildRole("123", ""))
```
Resolves to `https://discord.com/api/v10/guilds/123/roles/`
instead of `https://discord.com/api/v10/guilds/123/roles`
which gets the expected Not Found result (if a valid Guild ID is supplied)  instead of fetching all the guild roles.

### Final Changes:
- Argument guards (options / id) now use strict === undefined / !== undefined. The typings only allow the parameter to be omitted (never null), so undefined is the exact "no argument" signal, and a falsy ID like "" correctly routes to a single fetch.
- Resolved-id checks use strict  !== null, since resolveId returns the string as-is (including "") and only null on failure.
- All affected managers have been updated.

PR #11290, opened 2025-11-21
fix(builders): add proper snowflake validation

### LINKED ISSUE #11289: In multiple instances, builders just uses z.string() for Snowflakes

builders

### Issue description

One of many examples

https://github.com/discordjs/discord.js/blob/0b122633717bfa9f3a0ffe71a33c32f7afc319df/packages/builders/src/components/Assertions.ts#L67

### Versions

Current main branch of builders

### Issue priority

Medium (should be fixed soon)

Not applicable

Not applicable

0b122633717bfa9f3a0ffe71a33c32f7afc319df

rie03p:
close #11289

This PR adds proper Snowflake validation to the builders package. Previously, multiple instances were using plain `z.string()` for Snowflake IDs, which didn't validate that the values are numeric strings as required by Discord's Snowflake format.

Changes:

- Added `snowflakePredicate` (`z.string().regex(/^\d{17,20}$/)`) to validate Snowflake format
- Replaced `z.string()` with `snowflakePredicate` in:
  - `emojiPredicate.id`
  - `buttonPremiumPredicate.sku_id`
  - `default_values.id` for channel, mentionable, role, and user select menu predicates

### DIFF
--- a/packages/builders/src/Assertions.ts
+++ b/packages/builders/src/Assertions.ts
@@ -3,7 +3,7 @@ import { z } from 'zod';
 
 export const idPredicate = z.int().min(0).max(2_147_483_647).optional();
 export const customIdPredicate = z.string().min(1).max(100);
-export const snowflakePredicate = z.string().regex(/^(?:0|[1-9][0-9]*)$/);
+export const snowflakePredicate = z.string().regex(/^(?:0|[1-9]\d*)$/);
 
 export const memberPermissionsPredicate = z.coerce.bigint();
 


Jiralite:
@rie03p looks like something is not quite right because tests are failing, may need to update them?

rie03p:
@Jiralite You're right! The validation is now working correctly and rejecting invalid IDs like 'hi!' and '0'. I've updated the existing tests to use valid IDs (numbers or proper snowflakes) so they pass with the new validation.

packages/builders/src/Assertions.ts
```diff
@@ -3,6 +3,7 @@ import { z } from 'zod';
 
 export const idPredicate = z.int().min(0).max(2_147_483_647).optional();
 export const customIdPredicate = z.string().min(1).max(100);
+export const snowflakePredicate = z.string().regex(/^\d{17,20}$/);
```
almeidx:
```suggestion
export const snowflakePredicate = z.string().regex(/^[1-9]\d{16,18}$/);
```

almeidx (dismissed):

rie03p:
I’ve fixed the tests that were failing.

packages/builders/src/Assertions.ts
```diff
@@ -3,6 +3,7 @@ import { z } from 'zod';
 
 export const idPredicate = z.int().min(0).max(2_147_483_647).optional();
 export const customIdPredicate = z.string().min(1).max(100);
+export const snowflakePredicate = z.string().regex(/^[1-9]\d{16,18}$/);
```
vladfrangu:
```suggestion
export const snowflakePredicate = z.string().regex(/^[0-9]\d{16,19}$/);
```

Jiralite: 0 is perfectly valid at the start?

Also, disagree on increasing the limit. The API doesn't even accept snowflakes that large...

vladfrangu: > 0 is perfectly valid at the start?

I do not like that we have that at the start, I completely missed that its two parts... But also why not? _technically_ even `0` is a valid snowflake (and can be used in pagination to use as a starting point)

Jiralite: This is builders though. Snowflakes here are not used for pagination. They're used for emoji ids, select menus (role ids, channel ids, and user ids)... I'd argue someone is building something not at the start of Discord's time, so realistically, `0` is invalid.

If in the future, we ever make builders for methods that use pagination, we can revisit this?

vladfrangu: 🤷, my two cents is that this should've been `\d{17,20}` and not overly complex but w/e (plus for attachments it can literally be just `\d+` when creating)...

And before you argue for attachments that it cannot be `\d+` bc the spec says so -> API cares about the limit, not the ID last I checked

Jiralite: Creating attachments will allow `0`:

https://github.com/discordjs/discord.js/blob/02fc101069aed994f7ee4579d114706047025ddd/packages/builders/src/messages/Assertions.ts#L18

When you edit attachments, you include the attachment id to keep, which again, can't be `0`.

vladfrangu: This PR alters that too, soo maybe that needs to be undone? Also you miss the point, the string `"0"` will work too

Jiralite: It alters it because we're being strict on input: snowflakes for editing and numbers for uploading, which is the documented way.

> Also you miss the point, the string "0" will work too

`0` actually gets set to `undefined`:

https://github.com/discordjs/discord.js/blob/02fc101069aed994f7ee4579d114706047025ddd/packages/builders/src/messages/Attachment.ts#L151

Is... is that an issue?

Someone _shouldn't_ be passing `"0"` though, right? That seems like a mistake...

didinele: > Someone shouldn't be passing "0" though, right? That seems like a mistake...

yup. it should be `0` (number)

IMO the meaning of the ID is either snowflake (as in, literal, actual snowflake when editing, already uploaded attachment) or index, as in "im uploading this now, its the nth file"

re the code, yes. that's a bad falsy check

Jiralite: Opened #11314 to track that.

almeidx: I'm down to change this regex to the one from the spec.

https://github.com/discord/discord-api-spec/blob/28ef48483b4745bcf176b2b53b931a27142e2f55/specs/openapi.json#L33311-L33315

Jiralite: @vladfrangu If you are fine with the above, mind applying one of these?

```suggestion
export const snowflakePredicate = z.string().regex(/^(?:0|[1-9][0-9]*)$/);
```

```suggestion
export const snowflakePredicate = z.string().regex(/^(?:0|[1-9]\d{16,18})$/);
```

PR #11291, opened 2025-11-21
fix(builders): improve component validation predicates

hakkiai:
Fixed the incorrect Zod usage in the component predicates. Replaced the invalid
`z.literal([...])` patterns with proper unions, switched to `z.nativeEnum` for
runtime enums, and moved the old `.check()` logic to `superRefine()`.

This cleans up the schemas and fixes the issue with select menu default values.

- fixes:#11289

### DIFF
--- a/packages/builders/src/components/Assertions.ts
+++ b/packages/builders/src/components/Assertions.ts
@@ -62,7 +62,7 @@ const selectMenuBasePredicate = z.strictObject({
 
 export const selectMenuChannelPredicate = selectMenuBasePredicate.extend({
 	type: z.literal(ComponentType.ChannelSelect),
-	channel_types: z.nativeEnum(ChannelType).array().optional(),
+	channel_types: z.enum(ChannelType).array().optional(),
 	default_values: z
 		.strictObject({ id: snowflakePredicate, type: z.literal(SelectMenuDefaultValueType.Channel) })
 		.array()
@@ -75,10 +75,7 @@ export const selectMenuMentionablePredicate = selectMenuBasePredicate.extend({
 	default_values: z
 		.strictObject({
 			id: snowflakePredicate,
-			type: z.union([
-				z.literal(SelectMenuDefaultValueType.Role),
-				z.literal(SelectMenuDefaultValueType.User),
-			]),
+			type: z.literal([SelectMenuDefaultValueType.Role, SelectMenuDefaultValueType.User]),
 		})
 		.array()
 		.max(25)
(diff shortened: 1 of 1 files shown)


packages/builders/src/components/Assertions.ts
```diff
@@ -104,68 +107,54 @@ export const selectMenuStringPredicate = selectMenuBasePredicate
...
 	components: z.union([
 		z
-			.object({ type: z.literal(ComponentType.Button) })
+			.strictObject({ type: z.literal(ComponentType.Button) })
 			.array()
 			.min(1)
 			.max(5),
 		z
-			.object({
-				type: z.literal([
```
Qjuh, line 162: The `literal([...])` syntax is fine and encouraged in zod 4.

packages/builders/src/components/Assertions.ts
```diff
@@ -62,9 +62,9 @@ const selectMenuBasePredicate = z.object({
 
 export const selectMenuChannelPredicate = selectMenuBasePredicate.extend({
 	type: z.literal(ComponentType.ChannelSelect),
-	channel_types: z.enum(ChannelType).array().optional(),
+	channel_types: z.nativeEnum(ChannelType).array().optional(),
```
Qjuh: `nativeEnum()` is deprecated in zod v4, `enum()` was correct here, as it is overloaded to also do this in v4.

packages/builders/src/components/Assertions.ts
```diff
@@ -73,9 +73,12 @@ export const selectMenuChannelPredicate = selectMenuBasePredicate.extend({
 export const selectMenuMentionablePredicate = selectMenuBasePredicate.extend({
 	type: z.literal(ComponentType.MentionableSelect),
 	default_values: z
-		.object({
-			id: z.string(),
-			type: z.literal([SelectMenuDefaultValueType.Role, SelectMenuDefaultValueType.User]),
```
Qjuh, line 78: Also was fine with `literal([...])` before

hakkiai:
Thank you for the clarification @Qjuh 

- I actually see now that z.literal([...]) and z.enum() were already working correctly in v4. I've reverted those changes.

- I've kept the .strictObject() fixes (which was the original bug) and the snowflakePredicate usage for proper Discord ID validation. Is my current applied approach correct now? 

- Fixes:#11289

Qjuh:
As far as I can see the main aspect of this PR has been done and merged by #11290 already. The change from `check()` to `superRefine()` is merely syntactical sugar (since `superRefine` calls `check` internally to do the exact same thing). So the only thing this really changes are the `strictObject` assertions, but I don't see what issue they fix, as that would only make it throw instead of stripping extraneous properties.

PR #11301, opened 2025-11-25
chore: Delete the pull request template

Jiralite:
Before the monorepo, discord.js was only the discord.js package. The pull request template made sense (test your API changes, shout if your changes are breaking, and more).

Now, it contains packages like @discordjs/brokers, @discordjs/actions, the api-extractor packages, as well as the website and guide applications. Semantic versioning doesn't really apply to these, and testing API changes certainly doesn't apply for the websites.

I'm moving towards removing the file. At the very least, the status and versioning classification should be removed as we double-check that anyway (it duplicates work). This results in only describing changes, which is common sense.

Open to ideas to generalise this, but I don't think that's possible unlike issue forms.

### DIFF
--- a/package.json
+++ b/package.json
@@ -53,7 +53,7 @@
 	"devDependencies": {
 		"@commitlint/cli": "^20.1.0",
 		"@commitlint/config-angular": "^20.0.0",
-		"@favware/cliff-jumper": "^4.1.0",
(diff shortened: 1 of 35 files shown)


vladfrangu (approved):
I feel like we could've kept at least the first header but I know even we struggle to write proper content in there 😄

sdanialraza (approved):

PR #11304, opened 2025-11-26
refactor: move full validation to ChatInputCommandBuilder

### LINKED ISSUE #11308: Refactor chat input command validation overrides
Investigate whether the validation override can be changed to `false`.

https://github.com/discordjs/discord.js/blob/02fc101069aed994f7ee4579d114706047025ddd/packages/builders/src/interactions/commands/chatInput/ChatInputCommand.ts#L33

almeidx:
Made the `chatInputCommandPredicate` validate the full structure, including options, so users that may be using this predicate can actually validate Chat Input Commands.

Closes #11308


packages/builders/src/interactions/commands/chatInput/Assertions.ts
```diff
@@ -47,38 +48,55 @@ const choiceBasePredicate = z.object({
 	name: choiceValueStringPredicate,
 	name_localizations: localeMapPredicate.optional(),
 });
-const choiceStringPredicate = choiceBasePredicate.extend({
+const choiceStringPredicate = z.object({
+	...choiceBasePredicate.shape,
```
Qjuh, line 52: what's the reason for not using extend() here and in the following predicates?

almeidx: From <https://zod.dev/api?id=extend>

> This approach has a few advantages:
> 
> - It uses language-level features ([spread syntax](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax)) instead of library-specific APIs
> - The same syntax works in Zod and Zod Mini
> - It's more tsc-efficient — the .extend() method can be expensive on large schemas, and due to [a TypeScript limitation](https://github.com/microsoft/TypeScript/pull/61505) it gets quadratically more expensive when calls are chained
> - If you wish, you can change the strictness level of the resulting schema by using z.strictObject() or z.looseObject()

***

I'm not strictly committed to this approach. Down to revert the change if it's unwanted

didinele: I like it better honestly, probably makes sense to use it in some other places too

Jiralite (approved):

packages/builders/src/interactions/commands/chatInput/options/ApplicationCommandOptionBase.ts
```diff
@@ -24,7 +23,7 @@ export abstract class ApplicationCommandOptionBase
 	/**
 	 * @internal
 	 */
-	protected static readonly predicate: z.ZodType = basicOptionPredicate;
+	protected static readonly predicate: z.ZodType;
```
vladfrangu, line 26: Maybe mark as abstract?

almeidx: Cannot have abstract static props/methods
- https://github.com/microsoft/TypeScript/issues/34516

didinele: No, you [can't](https://www.typescriptlang.org/play/?#code/IYIwzgLgTsDGEAJYBthjAgKgU0gg3gFACQADlAPYTbzYAmCokM8CkwEAlrAlNsHQoA7ZAE8E5etw7YAXG2ichAcwDchAL5A)

vladfrangu: then lets mark this as optional please, in case some class forgets to define it

PR #11317, opened 2025-11-28
fix: Adjust label predicates and fix buttons emoji/label behaviour

Jiralite:
- Resolves #11316
- Resolves buttons being enforced to have labels (they may have either an emoji or a label)

PR #11318, opened 2025-11-28
fix: Adjust label predicates

Jiralite:
Backports #11317.

Buttons are correctly validated in this version.

### DIFF
--- a/packages/builders/src/components/v2/Container.ts
+++ b/packages/builders/src/components/v2/Container.ts
@@ -71,7 +71,8 @@ export class ContainerBuilder extends ComponentBuilder<APIContainerComponent> {
 	 * 		},
 	 * 	],
 	 * })
-	 * 	.addComponents(separator, section);
+	 *  .addSeparatorComponents(separator)
+	 *  .addSectionComponents(section);
 	 * ```
 	 */
 	public constructor({ components, ...data }: Partial<APIContainerComponent> = {}) {


packages/builders/__tests__/components/selectMenu.test.ts
```diff
@@ -53,12 +56,12 @@ describe('Select Menu Components', () => {
 			expect(() => selectMenu().setDisabled()).not.toThrowError();
 			expect(() => selectMenu().setPlaceholder('description')).not.toThrowError();
 			const option = selectMenuOption()
-				.setLabel('test')
+				.setLabel('a'.repeat(100))
```
Qjuh: Shouldn't this be defined as a const on top too, for the case the limit changes this might otherwise easily be missed.

packages/builders/src/components/selectMenu/StringSelectMenu.ts
```diff
@@ -1,7 +1,8 @@
 import { ComponentType } from 'discord-api-types/v10';
 import type { APIStringSelectComponent, APISelectMenuOption } from 'discord-api-types/v10';
 import { normalizeArray, type RestOrArray } from '../../util/normalizeArray.js';
-import { jsonOptionValidator, optionsLengthValidator, validateRequiredSelectMenuParameters } from '../Assertions.js';
```
Qjuh, line 4: jsonOptionValidator is not used anywhere anymore, but since we export it it would be breaking to remove it. deprecate maybe?

Jiralite: I noticed this too, but also no other validators have documentation. I'm sure there's more instances... just didn't think it's worth to do it. But if we really want to, we can?

packages/builders/__tests__/components/selectMenu.test.ts
Jiralite: Done with 5b8d5265e8bd56507ce0cab1995a34a17c9b666b.

packages/builders/src/components/selectMenu/StringSelectMenu.ts
Jiralite: Just did it anyway: b94c5fc03e68a06abf84af224133cba46da7d72c

packages/builders/src/components/Assertions.ts
```diff
@@ -50,6 +50,9 @@ export const labelValueDescriptionValidator = s
 	.lengthLessThanOrEqual(100)
 	.setValidationEnabled(isValidationEnabled);
 
+/**
+ * @deprecated Replaced with selectMenuStringOptionPredicate.
```
Qjuh, line 54: While testing this out I found we don't export the SelectMenuAssertions at all, so this replacement can't be used unless we add that export.

packages/builders/src/index.ts
```diff
@@ -31,6 +31,7 @@ export {
 	 */
 	StringSelectMenuOptionBuilder as SelectMenuOptionBuilder,
 } from './components/selectMenu/StringSelectMenuOption.js';
+export { selectMenuStringOptionPredicate } from './components/selectMenu/Assertions.js';
```
Qjuh:
```suggestion
export * as SelectMenuAssertions from './components/selectMenu/Assertions.js';
```

packages/builders/src/components/Assertions.ts
Qjuh, line 54:
```suggestion
 * @deprecated Replaced with {@link SelectMenuAssertions.selectMenuStringOptionPredicate}.
```

Jiralite: This doesn't show up in the documentation and this doesn't show a link in VSCode, so what's the point of this?

Qjuh: I wasn't able to check whether IDEs would show this correctly. If VSCode doesn't then I guess it's pointless indeed.

PR #11320, opened 2025-11-28
fix(undiciRequest): file uploading

almeidx:
Prevents the following error when uploading files.

```
node:internal/deps/undici/undici:6191
            if (buffer.byteLength) {
                       ^

TypeError: Cannot read properties of null (reading 'byteLength')
```

Refs:
- https://discordapp.com/channels/222078108977594368/1371875040563433534
- http://github.com/nodejs/node/issues/59012


didinele (approved):

Jiralite (approved):

ckohen:
Does setting the global dispatcher not work? I'd rather not run this codepath on every request

vladfrangu:
> Does setting the global dispatcher not work? I'd rather not run this codepath on every request

We should not override the global dispatcher in any way-shape-form. I'd rather we cache this on a client that uses undici, and pass it down

ckohen:
Tbh the global dispatcher being a problem is a major issue, but yeah. Why not just instantiate Agent once outside the function call and do ??= like the original commit?

Yes there's the potential to set the global dispatcher later than importing the strategy though...

Qjuh:
Any code running into this issue of getGlobalDispatcher() being not an instance of Dispatcher will be broken, so doing `setGlobalDispatcher(new Agent())` once (in a lazy way) seems to be the best approach to me. After all this is what node does internally which causes this issue in the first place

almeidx:
> Tbh the global dispatcher being a problem is a major issue, but yeah. Why not just instantiate Agent once outside the function call and do ??= like the original commit?

Done.

> Yes there's the potential to set the global dispatcher later than importing the strategy though...

Honestly, don't think this is a problem at all. The fact that undici is being used here is an implementation detail, and we have an official way of providing an agent if the user wants (via the `agent` option or `setAgent()`).

> Any code running into this issue of getGlobalDispatcher() being not an instance of Dispatcher will be broken, so doing `setGlobalDispatcher(new Agent())` once (in a lazy way) seems to be the best approach to me.

I'm with @vladfrangu here. It's not discord.js' responsibility to modify the global dispatcher of undici ever.

PR #11323, opened 2025-11-28
docs(discord.js): less omit types for better docs

Qjuh:

improves docs of [MessageEditOptions](https://discord.js.org/docs/packages/discord.js/main/MessageEditOptions:Interface), [GuildScheduledEventEditOptions](https://discord.js.org/docs/packages/discord.js/main/GuildScheduledEventEditOptions:Interface), [GuildChannelCreateOptions](https://discord.js.org/docs/packages/discord.js/main/GuildChannelCreateOptions:Interface) and [GuildChannelCloneOptions](https://discord.js.org/docs/packages/discord.js/main/GuildChannelCloneOptions:Interface) by turning them into pure interface inheritance instead of Omit<>  types.


packages/discord.js/typings/index.d.ts
```diff
@@ -5398,7 +5398,35 @@ export type CacheWithLimitsOptions = {
...
+  rtcRegion?: string;
+  topic?: string;
+  type?: Exclude<
+    ChannelType,
+    | ChannelType.AnnouncementThread
+    | ChannelType.DM
+    | ChannelType.GroupDM
+    | ChannelType.PrivateThread
+    | ChannelType.PublicThread
+  >;
```
vladfrangu: wouldn't it be better for the docs to make this a specific type? `type CategoryCreateChannelType = ChannelType.X | ...`?

Qjuh: The "real" type already has such a name: `GuildChannelTypes`. This here is the original type which I didn't want to alter in case there was a reason why it includes `GuildDirectory` channel type. But I guess that is an oversight and `GuildChannelTypes` would be correct here?

packages/discord.js/typings/index.d.ts
```diff
@@ -6759,9 +6767,9 @@ export interface MessageCreateOptions extends BaseMessageCreateOptions {
 }
 
 export interface GuildForumThreadMessageCreateOptions
-  extends BaseMessageOptions, MessageOptionsFlags, MessageOptionsStickers {}
+  extends BaseMessageSendOptions, MessageOptionsFlags, MessageOptionsStickers {}
 
-export interface MessageEditOptions extends Omit<BaseMessageOptions, 'content'> {
+export interface MessageEditOptions extends BaseMessageOptions {
   attachments?: readonly (Attachment | JSONEncodable<APIAttachment>)[];
   content?: string | null;
```
almeidx: This is how it is in `BaseMessageOptions` now too. No need for override
```suggestion
```

PR #11327, opened 2025-11-30
refactor!: move `embedLength` function to util package

almeidx:
BREAKING CHANGE: Builders no longer exports a `embedLength` function. Import it from `@discordjs/util` instead.

***

This unblocks us from removing the dependency of `@discordjs/builders` from the `discord.js` package completely.

### DIFF
--- a/packages/voice/README.md
+++ b/packages/voice/README.md
@@ -62,6 +62,13 @@ try installing another.
 - `@noble/ciphers`: ^1.0.0
 - `libsodium-wrappers`: ^0.7.9
 
+**DAVE Protocol Libraries (e2ee)**
+
+> [!NOTE]
+> Some Discord clients may require the DAVE protocol for end-to-end encryption in voice chat and refuse to downgrade the connection in the future. Ensure you have `@snazzah/davey` installed to avoid compatibility issues.
+
+- `@snazzah/davey`: ^0.1.6
+
 **Opus Libraries (npm install):**
 
 - `@discordjs/opus`: ^0.4.0


Jiralite:
> BREAKING CHANGE: Builders no longer exports a `embedLength` function. Import it from `@discordjs/util` instead.

Unsure about telling users to download a package we use primarily for sharing stuff across the other packages. Shouldn't we export this in builders still?

packages/builders/__tests__/messages/embed.test.ts
```diff
@@ -14,26 +12,6 @@ const base = {
...
-			expect(embedLength(embed.toJSON())).toEqual(alpha.length * 6);
-		});
-
-		test('GIVEN an embed with zero characters THEN returns amount of characters', () => {
-			const embed = new EmbedBuilder();
-
-			expect(embedLength(embed.toJSON(false))).toEqual(0);
-		});
-	});
-
```
Qjuh, line 36: I'd say we should keep this test, import embedLength from /util. Since this doesn't (only) test the embedLength function but also that it works correctly with EmbedBuilder

Qjuh (changes requested):

didinele (changes requested):
> Unsure about telling users to download a package we use primarily for sharing stuff across the other packages. Shouldn't we export this in builders still?

I actually agree with this. I'm a bit uncertain if it's builders that should be re-exporting it, though. Probably our "entrypoint" high-level libs? core, discord.js

almeidx:
Discord.js continues to export it.
https://github.com/discordjs/discord.js/blob/0aaba0305fb29d25e4caac333fbb5be8a05fb271/packages/discord.js/src/index.js#L259

Core does not currently, and I don't think it should either

PR #11330, opened 2025-12-04
fix(builders): handle 0 as valid attachment ID

### LINKED ISSUE #11314: Attachment builder does not handle `0` for ids
`0` actually gets set to `undefined`:

https://github.com/discordjs/discord.js/blob/02fc101069aed994f7ee4579d114706047025ddd/packages/builders/src/messages/Attachment.ts#L151

_Originally posted by @Jiralite in https://github.com/discordjs/discord.js/pull/11290#discussion_r2568537025_

Exodo0:
## Description
Fixes #11314

`AttachmentBuilder.getRawFile()` was using a falsy check (`this.data.id ? ...`) to determine whether the attachment ID existed.  
Because `0` is a valid ID but is falsy in JavaScript, the method incorrectly treated it as missing, producing an invalid file reference.

This PR updates the condition to properly detect the presence of the `id` property, allowing `0` to be handled correctly.

## Changes
- Replaced the falsy check `this.data.id? ...` with a property-existence check: `'id' in this.data? ...`.
- Added a test to confirm that an attachment with ID `0` generates the expected raw file key (`files[0]`).

## Testing
- All existing tests pass.
- Added test: **"AttachmentBuilder handles 0 as a valid ID"**.
- Verified that `getRawFile()` returns `'files[0]'` when the ID is `0`.

## Type of Change
- [x] Bug fix (non-breaking change which fixes an issue)


packages/builders/__tests__/messages/fileBody.test.ts
```diff
@@ -24,6 +24,17 @@ test('AttachmentBuilder stores and exposes file data', () => {
 	expect(attachment.getRawFile()).toBe(undefined);
 });
 
+test('AttachmentBuilder handles 0 as a valid ID', () => {
```
Jiralite: We prefer "id":

```suggestion
test('AttachmentBuilder handles 0 as a valid id', () => {
```

packages/builders/src/messages/Attachment.ts
```diff
@@ -148,7 +148,7 @@ export class AttachmentBuilder implements JSONEncodable<RESTAPIAttachment> {
 		return {
 			...this.fileData,
 			name: this.data.filename,
-			key: this.data.id ? `files[${this.data.id}]` : undefined,
+			key: 'id' in this.data ? `files[${this.data.id}]` : undefined,
```
almeidx:
```suggestion
			key: this.data.id !== undefined ? `files[${this.data.id}]` : undefined,
```

PR #11337, opened 2025-12-08
types(UserManager): fix send() return type to Promise<Message<false>>

### LINKED ISSUE #11331: UserManager.send() returns Promise<Message> instead of Promise<Message<false>>

discord.js

### Issue description

UserManager.send() always sends a direct message

### Code sample

```typescript
public send(user: UserResolvable, options: string | MessagePayload | MessageCreateOptions): Promise<Message>;
```

### Versions

- discord.js@14.25.1
- Node.js v24.11.0
- typescript@5.9.3

### Issue priority

Low (slightly annoying)

Not applicable

Not applicable

_No response_

AryanBagade:
Fixes the return type of `UserManager.send()` to correctly reflect that it always returns a DM message.

### What changed
- Updated return type from `Promise<Message>` to `Promise<Message<false>>` in both TypeScript definitions and JSDoc
- Added type test to prevent regression

### Why
`UserManager.send()` always creates a DM channel and sends the message there, so the return type should indicate it's a non-guild message (`Message<false>`). This makes the types more accurate and helps developers using TypeScript.

Closes #11331


packages/discord.js/src/managers/UserManager.js
```diff
@@ -107,7 +107,7 @@ class UserManager extends CachedManager {
    *
    * @param {UserResolvable} user The UserResolvable to identify
    * @param {string|MessagePayload|MessageCreateOptions} options The options to provide
-   * @returns {Promise<Message>}
+   * @returns {Promise<Message<false>>}
```
almeidx: We don't use this generic anywhere else on the jsdocs. Not sure if there's an actual reason or if it was just missed.


```suggestion
   * @returns {Promise<Message>}
```

packages/discord.js/typings/index.test-d.ts
```diff
@@ -1532,6 +1532,12 @@ expectType<SendMethod<true>['send']>(voiceChannel.send);
 expectAssignable<SendMethod>(user);
 expectAssignable<SendMethod>(guildMember);
 
+// Test UserManager.send() returns Promise<Message<false>> (DM)
+{
+  const readyClient: Client<true> = {} as Client<true>;
+  expectType<Promise<Message<false>>>(readyClient.users.send(user, 'test'));
+}
```
almeidx: The client being ready isn't really relevant for types here, so there's no need for all that extract boilerplate
```suggestion
expectType<Promise<Message<false>>>(client.users.send(user, 'test'));
```

sdanialraza (approved):

PR #11361, opened 2025-12-12
refactor!: remove builders and formatters re-export

didinele:
~~people are gonna love me for this one~~

We've glossed over this briefly in internals, and #11327 also mentioned that this is now possible. The primary argument here is just that there's absolutely no internal dependency on builders anymore (my PR leaves it as a devDep since it is used in tests to ensure things correctly accept `JSONEncodable`s of the right type), and this actually simplifies usage in a sense. There's no real versioning issues anymore, the major of builders you're on (or even implementation? anyone can write a class that implements `JSONEncodable`, which discord.js will happily accept, it just needs to have a `toJSON` method) is not significant.

We're also generally somewhat discouraging builders, it should be an option, but it's more of an extension that you can opt into.

"But we still re-export all the other packages" yeah. We critically depend on those for internals and version miss-matches there from a user also installing them can be pretty damaging.

On the topic of formatters, I'm not dying on the hill that we can't re-export them, but I don't necessarily see why we should for the same rationale: it's not a core dependency. We do depend on it to implement certain methods in internals, but that's ok.

Note: Requires discord-api-types bump.

### DIFF
--- a/packages/builders/src/interactions/commands/chatInput/ChatInputCommandSubcommands.ts
+++ b/packages/builders/src/interactions/commands/chatInput/ChatInputCommandSubcommands.ts
@@ -74,7 +74,7 @@ export class ChatInputCommandSubcommandGroupBuilder
 		const data = {
 			...(structuredClone(rest) as Omit<APIApplicationCommandSubcommandGroupOption, 'type'>),
 			type: ApplicationCommandOptionType.SubcommandGroup as const,
-			options: options?.map((option) => option.toJSON(validationOverride)) ?? [],
+			options: options?.map((option) => option.toJSON(false)) ?? [],
 		};
 
 		validate(chatInputCommandSubcommandGroupPredicate, data, validationOverride);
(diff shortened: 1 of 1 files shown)


didinele:
I need a hand here, I'm not sure how to deal with this:

https://github.com/discordjs/discord.js/blob/8573d8b070c5a21c7ccc9ea3d5c3bd8d8f2ab02f/packages/discord.js/typings/index.d.ts#L5

Only usage

https://github.com/discordjs/discord.js/blob/8573d8b070c5a21c7ccc9ea3d5c3bd8d8f2ab02f/packages/discord.js/typings/index.d.ts#L4996-L5005

Do we just move that type to `util` and call it a day? "Re-implement" it in discord.js? builders has it written like so:

https://github.com/discordjs/discord.js/blob/df0c28afc2d0449982622073c69799c3cd8d3548/packages/builders/src/interactions/commands/chatInput/mixins/ApplicationCommandOptionChannelTypesMixin.ts#L4-L20

didinele:
@coderabbitai review

didinele:
@coderabbitai review

didinele:
@coderabbitai review

Jiralite (approved):

PR #11363, opened 2025-12-14
types(ModalSubmitInteraction): make attachments optional

### LINKED ISSUE #11359: ModalSubmitInteraction field component.attachments may be undefined despite typings

discord.js

### Issue description

```ts
    // do something
  }
}
```

### Code sample

```typescript
  const comp = field.component;
    // do something
}

// workaround:
```

### Versions

- discord.js: `14.25.1`
- bun: `1.3.2`
- typescript: `5.9.2`
- OS:
  - Prod: Linux Ubuntu 24.04

### Issue priority

Medium (should be fixed soon)

ThreadMember

_No response_

rahuld109:
## Summary

Makes `FileUploadModalData.attachments` optional to match runtime behavior.

## Problem

When a modal has an optional `FileUpload` component and the user doesn't upload a file, `component.attachments` is `undefined` at runtime. However, the TypeScript types marked it as required, causing no compile-time warning but runtime errors when accessing the property.

## Solution

Add `?` to the `attachments` field in `FileUploadModalData` interface:

```typescript
// Before
attachments: ReadonlyCollection<Snowflake, Attachment>;

// After  
attachments?: ReadonlyCollection<Snowflake, Attachment>;
```

Closes #11359

### DIFF
--- a/packages/discord.js/src/index.js
+++ b/packages/discord.js/src/index.js
@@ -105,6 +105,8 @@ exports.ActionRow = require('./structures/ActionRow.js').ActionRow;
 exports.Activity = require('./structures/Presence.js').Activity;
 exports.AnnouncementChannel = require('./structures/AnnouncementChannel.js').AnnouncementChannel;
 exports.AnonymousGuild = require('./structures/AnonymousGuild.js').AnonymousGuild;
+exports.AuthorizingIntegrationOwners =
+  require('./structures/AuthorizingIntegrationOwners.js').AuthorizingIntegrationOwners;
 exports.Application = require('./structures/interfaces/Application.js').Application;
 exports.ApplicationCommand = require('./structures/ApplicationCommand.js').ApplicationCommand;
 exports.ApplicationEmoji = require('./structures/ApplicationEmoji.js').ApplicationEmoji;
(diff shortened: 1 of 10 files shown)


almeidx (dismissed):

vladfrangu:
Is there a reason we don't just always create a collection? Thoughts?

Jiralite:
If we receive an empty array, we should make an empty collection.

rahuld109:
Just checking - would you like me to switch to the alternate approach (always creating an empty collection at runtime)? Let me know and I can make that change.

Jiralite:
> When a modal has an optional `FileUpload` component and the user doesn't upload a file, `component.attachments` is `undefined` at runtime. However, the TypeScript types marked it as required, causing no compile-time warning but runtime errors when accessing the property.

If Discord sends an empty array, we should be setting `components.attachments` to an empty collection. Can you verify that behaviour?

Jiralite (changes requested):
I have just verified this behaviour:

```JSON
{
  "type": 18,
  "id": 9,
  "component": {
    "values": [],
    "type": 19,
    "id": 10,
    "custom_id": "164"
  }
}
```

An empty array is sent for `values`. The types are fine, it's JavaScript that needs to be fixed–an empty collection should be used.

PR #11364, opened 2025-12-14
fix(discord.js): make FileUploadModalData.attachments optional

### LINKED ISSUE #11359: ModalSubmitInteraction field component.attachments may be undefined despite typings

majiayu000:
## Summary

- Makes the `attachments` property optional in `FileUploadModalData` interface
- Aligns TypeScript types with runtime behavior where attachments may be undefined
- Adds type test to verify the optional property behavior

## Issue

Fixes #11359

## Test Plan

- [x] Added type test in `typings/index.test-d.ts` to verify `attachments` can be undefined
- [x] All existing tests pass (`pnpm run test --filter=discord.js`)
- [x] Type definitions are valid (tsd passes)

## Changes

- `packages/discord.js/typings/index.d.ts`: Changed `attachments` from required to optional in `FileUploadModalData`
- `packages/discord.js/typings/index.test-d.ts`: Added type assertion test for optional attachments

### DIFF
--- a/packages/discord.js/typings/index.d.ts
+++ b/packages/discord.js/typings/index.d.ts
@@ -2562,7 +2562,7 @@ export interface SelectMenuModalData<Cached extends CacheType = CacheType> exten
 }
 
 export interface FileUploadModalData extends BaseModalData<ComponentType.FileUpload> {
-  attachments: ReadonlyCollection<Snowflake, Attachment>;
+  attachments?: ReadonlyCollection<Snowflake, Attachment>;
   customId: string;
   values: readonly Snowflake[];
 }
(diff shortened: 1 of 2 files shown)


vakiliner:
Duplicate of #11363

PR #11368, opened 2025-12-21
feat(create-discord-bot): allow certain files/folders to be included in the target directory 

jacobhumston:
This pull request allows the template to be created inside directories that contain a list of common folders and files, such as `.git` and `LICENSE`. This is achieved by the utility function found in https://github.com/jacobhumston/discord.js/blob/2c6dffc2cc74e88a19000e20ebe86361b96b8cee/packages/create-discord-bot/src/util/isFolderEmpty.ts

A `.gitignore` is also included in each template for convenience.

### DIFF
--- a/packages/structures/src/emojis/Emoji.ts
+++ b/packages/structures/src/emojis/Emoji.ts
@@ -1,0 +1,88 @@
+import { DiscordSnowflake } from '@sapphire/snowflake';
+import type { APIEmoji } from 'discord-api-types/v10';
+import { Structure } from '../Structure.js';
+import { kData } from '../utils/symbols.js';
+import { isIdSet } from '../utils/type-guards.js';
+import type { Partialize } from '../utils/types.js';
+
+/**
+ * Represents any emoji on Discord.
+ *
+ * @typeParam Omitted - Specify the properties that will not be stored in the raw data field as a union, implement via `DataTemplate`
+ * @remarks has substructure `User` which needs to be instantiated and stored by an extending class using it
+ * @remarks intentionally does not export `roles` so that extending classes can resolve `Snowflake[]` to `Role[]`
+ */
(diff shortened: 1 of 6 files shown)


jacobhumston:
> We should just use the `isFolderEmpty` helper from the create-next-app package.
> 
> https://github.com/vercel/next.js/blob/7fb2aa908216fb3a910c7fa6d24524412b5af6e5/packages/create-next-app/helpers/is-folder-empty.ts#L6
> 
> We'd need to ensure their license is included in the file

I've implemented your requested change; however, I wasn't really sure how to go about adding a .gitignore. Should we include a .gitignore in each template instead? Or just avoid creating one altogether.

almeidx:
I think we can just include a .gitignore in each template

jacobhumston:
> I think we can just include a .gitignore in each template

I added https://github.com/github/gitignore/blob/main/Node.gitignore to each template, `bun init` actually includes a `.gitignore` but it seems to just be a simplified version of the node one.

packages/create-discord-bot/src/util/isFolderEmpty.ts
```diff
@@ -0,0 +1,76 @@
+/**
+ * The MIT License (MIT)
+ *
+ * Copyright (c) 2025 Vercel, Inc.
```
almeidx, line 4: As is, this comment will not be included in `dist/index.js`. The question is, should it? Seemingly, changing the start of this comment to `/*!` seems to make it be included at the bottom of the minified bundle

cc @iCrawl @ckohen 

Also, I think we should include a permalink to the source file, so it's easier to find in the future

jacobhumston: Adding `@license` also places the comment to the bottom as well.

almeidx: Let's go with `@license` then

Jiralite (approved):

Issue #10981, opened 2025-07-12
Progress Tracker for @discordjs/structures


ckohen:
With the recent merge of #10900 we are now ready to implement the rest of the structures. The following is a progress tracker that incldues all structures.  Anyone may PR any one of these structures (or a group that makes sense together).

Some general guidelines (please take a look at the existing structures before working on new ones):
- These base structures do not construct sub-structures, that will be delegated to the caching library
- Because we do not construct sub-structures, the base structures must not expose these via a getter to make it possible to expose the constructed classes via that getter in a subclass.
- Things that are ineffecient to store the way Discord sends them may be optimized (like timestamps), they *must* be reintroduced the way Discord sends them in `toJSON()`

 
- [ ] Application
- [ ] Audit Log
- [ ] Auto Moderation
- [ ] Channel
  - [x] Channels
  - [x] PermissionOverwrite
  - [x] ForumTag
  - [x] ThreadMetadata
  - [x] Mixins
    - [x] DM
    - [x] Guild
    - [x] Text
    - [x] Owner
    - [x] Parent
    - [x] Permissions
    - [x] Pin
    - [x] Slowmode
    - [x] Topic
    - [x] Webhooks
    - [x] Thread
    - [x] ThreadOnly
    - [x] Voice
  - [ ] Message
- [ ] Emoji
- [ ] Guild
  - [ ] Guild
  - [ ] Guild Preview
  - [ ] Guild Widget
  - [ ] Guild Member
  - [ ] Integration
  - [ ] Guild Ban
  - [ ] ~~Membership Screening~~ Oh wait...this is *still* undocumented
- [ ] Guild Scheduled Event
- [ ] Guild Template
- [ ] Interactions (Need more thoughts on this)
- [x] Invite
- [ ] Stage Instance
- [ ] Sticker
  - [ ] Sticker
  - [ ] Sticker Pack
- [ ] Teams
- [x] User
  - [x] User
    - [x] AvatarDecorationData
  - [x] Connection
- [ ] Voice
  - [ ] Voice State
  - [ ] Voice Region
- [ ] Webhook

PR #11369, opened 2025-12-21
feat(structures): add emoji structure

AsadHumayun:

This PR will add the `Emoji` structure. I have done my best to follow the practices / patterns laid out in other structures. This PR should be merged to add the `Emoji` structure on to the structures package in the monorepo.

Please note that this PR does not include tests for the `Emoji` structure at the moment.

This PR makes contributions towards #10981. Mentioning for visibility.

### DIFF
--- a/packages/structures/src/emojis/Emoji.ts
+++ b/packages/structures/src/emojis/Emoji.ts
@@ -10,6 +10,7 @@ import type { Partialize } from '../utils/types.js';
  *
  * @typeParam Omitted - Specify the properties that will not be stored in the raw data field as a union, implement via `DataTemplate`
  * @remarks has substructure `User` which needs to be instantiated and stored by an extending class using it
+ * @remarks intentionally does not export `roles` so that extending classes can resolve `Snowflake[]` to `Role[]`
  */
 export class Emoji<Omitted extends keyof APIEmoji | '' = ''> extends Structure<APIEmoji, Omitted> {
 	/**


packages/structures/package.json
```diff
@@ -63,7 +63,7 @@
 	"dependencies": {
 		"@discordjs/formatters": "workspace:^",
 		"@sapphire/snowflake": "^3.5.5",
-		"discord-api-types": "^0.38.36"
+		"discord-api-types": "^0.38.37"
```
Qjuh: Ignore the AI review and revert this. Bumps like that should happen for the whole monorepo at once and are usually done as separate PRs.

packages/structures/src/emojis/Emoji.ts
Jiralite, line 1: There's a mix of "Emoji" and "emoji" here (capitalisation), should probably prefer the latter.

packages/structures/package.json
Jiralite: This update isn't required for anything in this pull request, I would recommend to remove it to simplify the diff (usually, we update dependencies in one go).

packages/structures/src/emojis/Emoji.ts
```diff
@@ -0,0 +1,100 @@
...
+		return this[kData].roles;
+	}
+
+	/**
+	 * User that created this emoji
+	 */
+	public get user() {
+		return this[kData].user;
+	}
+
```
Qjuh:
```suggestion
```

These are substructures and should not be exposed as getters, so the library extending these can replace them with their corresponding structures.

Qjuh (changes requested):

packages/structures/src/emojis/Emoji.ts
AsadHumayun: Thanks for catching that - I have gone over the structure and changed all instances to "emoji" instead of "Emoji". Please let me know if I've missed any

packages/structures/package.json
AsadHumayun: Thanks for letting me know - I've reverted the commits that do this

AsadHumayun: Reverted 👍

packages/structures/src/emojis/Emoji.ts
AsadHumayun: Removed as per your suggestion

packages/structures/src/emojis/Emoji.ts
```diff
@@ -0,0 +1,86 @@
...
+import type { APIEmoji } from 'discord-api-types/v10';
+import { Structure } from '../Structure.js';
+import { kData } from '../utils/symbols.js';
+import { isIdSet } from '../utils/type-guards.js';
+import type { Partialize } from '../utils/types.js';
+
+/**
+ * Represents any emoji on Discord.
+ *
+ * @typeParam Omitted - Specify the properties that will not be stored in the raw data field as a union, implement via `DataTemplate`
```
Qjuh, line 11: Please add a remark about the substructures that need to be handled by downstream implementers.

AsadHumayun: Sorry for missing that, I should've clocked that earlier. I've gone ahead and added a remarks comment, however I'm not too sure if we should be using `Role` or `Roles` here (singular or pluralised). I've gone ahead and used `Roles` for now in the comment. 

```ts
/**
 * Represents any emoji on Discord.
 * ...
 * @remarks has substructures `User`, `Roles` which need to be instantiated and stored by an extending class using it
 */
```

packages/structures/src/emojis/Emoji.ts
almeidx: Should `roles` actually be removed? It's an array of snowflakes, not an array of role structures

Qjuh: Extending libraries might want to expose a getter for them as array of roles though, so it can't be hardcoded as getter for ids.

packages/structures/src/emojis/Emoji.ts
```diff
@@ -0,0 +1,87 @@
...
+ * @remarks has substructures `User`, `Roles` which need to be instantiated and stored by an extending class using it
```
Qjuh:
```suggestion
 * @remarks has a substructure `User` which needs to be instantiated and stored by an extending class using it
```
as remarked correctly by Almeida roles are not a substructure, as they are simply ids.

Qjuh (changes requested):

packages/structures/src/emojis/Emoji.ts
AsadHumayun: Is it worth potentially mentioning this in another remark - roles was intentionally not exposed as a getter such that the extending library can resolve it to Role[] instead of just role ids?

packages/structures/src/emojis/Emoji.ts
AsadHumayun: Updated

packages/structures/src/emojis/Emoji.ts
almeidx: It should definitely be documented. Not sure where/how

almeidx (approved):

PR #11375, opened 2026-01-02
fix: Enforce name limit of 2 again for custom emojis

Jiralite:
Fixes a regression introduced with #11371 that allowed custom emojis to have a name of length 1.

Also fixes the dishonest emoji tests where a name was specified with a string instead of actual emojis.

### DIFF
--- a/packages/builders/src/components/Assertions.ts
+++ b/packages/builders/src/components/Assertions.ts
@@ -5,7 +5,7 @@ import { idPredicate, customIdPredicate, snowflakePredicate } from '../Assertion
 export const emojiPredicate = z
 	.strictObject({
 		id: snowflakePredicate.optional(),
-		name: z.string().max(32).optional(),
+		name: z.string().min(1).max(32).optional(),
 		animated: z.boolean().optional(),
 	})
 	.refine((data) => data.id !== undefined || data.name !== undefined, {


packages/builders/__tests__/components/button.test.ts
```diff
@@ -66,7 +76,17 @@ describe('Button Components', () => {
...
+				const button = new PrimaryButtonBuilder().setEmoji('🩵');
+				button.toJSON();
+			}).toThrowError();
+
+			expect(() => {
+				const button = new PrimaryButtonBuilder()
+					.setCustomId('custom')
+					.setLabel('test')
+					.setDisabled(true)
+					.setEmoji({ id: '1234567890123456', name: '1' });
```
vladfrangu, line 88: For custom emojis, does discord actually validate their name? Because when sending a message, we can still do `<:_:[id]>` no?

tipakA: They don't validate, but (some) clients viewing the emote will grab the name from cache.

Still, they are documented as 2-32 long in name, you cannot name them as single chars, and our previous hacky workaround of ignoring this and using `_` would break markdown in places.

almeidx:
FYI the `name` field doesn't seem to be validated at all by the API for custom emojis. You can send numbers, booleans, empty strings, very large strings, or null. Seemingly anything but an object or an array

Jiralite:
Did you try uploading an emoji?

```JSON
{
    "message": "Invalid Form Body",
    "code": 50035,
    "errors": {
        "name": {
            "_errors": [
                {
                    "code": "BASE_TYPE_BAD_LENGTH",
                    "message": "Must be between 2 and 32 in length."
                }
            ]
        }
    }
}
```

almeidx:
I sent a primary button component with a custom emoji, which is one of the predicates this pull request is modifying

Jiralite:
Right, and how did you get that custom emoji? It was uploaded somewhere.

vladfrangu:
@Jiralite i think they're referring to the API validating the emoji name when creating the message, not the emoji structure itself...

Issue #11382, opened 2026-01-16
Uncaught error: `this.client.user` was `null` when `GuildMemberManager.edit` was called.


doanba13:
### Which package is this bug report for?

discord.js

### Issue description

Steps to reproduces

1. Initialize client
```ts
await this.client.login(token);
console.log(this.client.user) // Clearly see the `user` object
```

2. Add roles to a member
```ts
const guild = await this.discordClient.guilds.fetch(guildId);
const member = await guild.members.fetch(userId);

// Work like a charm!
// for (let role of uniqueRoles) {
//     await member.roles.add(role);
// }

// Did work! Error provide below
await member.roles.add(uniqueRoles); // Snowflake[]
```

This is what happen when I add multple roles at once:
```
at GuildMemberManager.edit (/xxx/node_modules/discord.js/src/managers/GuildMemberManager.js:405:33)
at GuildMember.edit (/xxx/node_modules/discord.js/src/structures/GuildMember.js:399:31)
at GuildMemberRoleManager.set (/xxx/node_modules/discord.js/src/managers/GuildMemberRoleManager.js:207:24)
at GuildMemberRoleManager.add (/xxx/node_modules/discord.js/src/managers/GuildMemberRoleManager.js:131:19)
```

This is `GuildMemberManager` `edit` method, error was throw while access `id` in null `this.client.user`
```ts
/**
   * Edits a member of a guild.
   *
   * @param {UserResolvable} user The member to edit
   * @param {GuildMemberEditOptions} options The options to provide
   * @returns {Promise<GuildMember>}
   */
  async edit(user, { reason, ...options }) {
    const id = this.client.users.resolveId(user);
    if (!id) throw new DiscordjsTypeError(ErrorCodes.InvalidType, 'user', 'UserResolvable');

    if (options.channel) {
      options.channel = this.guild.channels.resolve(options.channel);
      if (!(options.channel instanceof BaseGuildVoiceChannel)) {
        throw new DiscordjsError(ErrorCodes.GuildVoiceChannelResolve);
      }
      options.channel_id = options.channel.id;
      options.channel = undefined;
    } else if (options.channel === null) {
      options.channel_id = null;
      options.channel = undefined;
    }
    options.roles &&= options.roles.map(role => (role instanceof Role ? role.id : role));

    if (options.communicationDisabledUntil !== undefined) {
      options.communication_disabled_until =
        // eslint-disable-next-line eqeqeq
        options.communicationDisabledUntil != null
[...]

PR #11383, opened 2026-01-16
fix(GuildMemberManager): handle potential null user ID in member resolution

doanba13:
I encountered an error when trying to add multiple roles to a guild member. The issue comes from `client.user` being null. The code runs smoothly after adding a null check for the user object.

I’ve reported the issue here: #11382


Jiralite:
https://github.com/discordjs/discord.js/issues/11382#issuecomment-3759525779

PR #11385, opened 2026-01-18
fix(voice): always install Davey as DAVE is becoming required

vladfrangu:
E2EE for voice connections is becoming mandatory anyways, so might as well mark the dependency as mandatory. And, should we, in the future, get more variants for dave implementations, we can adjust the code to support them

### DIFF
--- a/packages/collection/src/collection.ts
+++ b/packages/collection/src/collection.ts
@@ -247,7 +247,7 @@ export class Collection<Key, Value> extends Map<Key, Value> {
 	public reverse() {
 		const entries = [...this.entries()].reverse();
 		this.clear();
-		for (const [key, value] of entries) this.set(key, value);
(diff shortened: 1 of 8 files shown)


packages/voice/src/networking/DAVESession.ts
```diff
@@ -70,7 +59,7 @@ export interface DAVESessionOptions {
  * The maximum DAVE protocol version supported.
  */
 export function getMaxProtocolVersion(): number | null {
```
almeidx: There is also more redudant code involving `getMaxProtocolVersion()` in the Networking class, since this no longer returns `null` (it actually never returned null, that was typed incorrectly, it used to return `undefined`)

```suggestion
export function getMaxProtocolVersion(): number {
```

PR #11388, opened 2026-01-20
docs(structures): update DataTemplate docs

AsadHumayun:
This PR will correct comments on constructors of structures that contain defaults.

No code changes are introduced in this PR.

### DIFF
--- a/packages/structures/src/users/User.ts
+++ b/packages/structures/src/users/User.ts
@@ -14,9 +14,6 @@ import type { Partialize } from '../utils/types.js';
 export class User<Omitted extends keyof APIUser | '' = ''> extends Structure<APIUser, Omitted> {
 	/**
 	 * The template used for removing data from the raw data stored for each User.
-	 *
-	 * @remarks This template has defaults, if you want to remove additional data and keep the defaults,
-	 * use `Object.defineProperties`. To override the defaults, set this value directly.
 	 */
 	public static override readonly DataTemplate: Partial<APIUser> = {};
 


packages/structures/src/users/User.ts
```diff
@@ -13,7 +13,10 @@ import type { Partialize } from '../utils/types.js';
  */
 export class User<Omitted extends keyof APIUser | '' = ''> extends Structure<APIUser, Omitted> {
 	/**
-	 * The template used for removing data from the raw data stored for each User
+	 * The template used for removing data from the raw data stored for each User.
+	 *
+	 * @remarks This template has defaults, if you want to remove additional data and keep the defaults,
+	 * use `Object.defineProperties`. To override the defaults, set this value directly.
 	 */
 	public static override readonly DataTemplate: Partial<APIUser> = {};
```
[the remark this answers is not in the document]
AsadHumayun: TY for catching

ckohen:
DataTemplate is readonly by design, meaning you cannot directly set the property to another object, to avoid accidentally overwriting the defaults.

AsadHumayun:
Thanks for getting back to me. Considering what you've pointed out, I see it wouldn't be a good idea to suggest setting the value directly as defaults should realistically not be overridden. 

I'll go ahead and remove this part of the comment:

```
To override the defaults, set this value directly.
```

Is this something that I should bring up with the wider structures team? It seems that this is/was the convention and is present in other files as well, at the moment (eg the Invite structure). 

Let me know if you think this looks good or other teams should be notified.

ckohen:
I missed the change to readonly in the initial PR. I had originally not made them readonly so that consumers could actually overwrite completely, but overwritting individual properties in the same way you add properties seems better. The docs change was missed during this switch, you could update the rest of the docs as well.

AsadHumayun:
Just to confirm we're both on the same page, is this the kind of final doc you're expecting?

```ts
	/**
	 * The template used for removing data from the raw data stored for each team.
	 * 
	 * @remarks DataTemplate is readonly by design, meaning you cannot directly set the property to another object,
	 * to avoid accidentally overwriting the defaults.
	 */
	public static override readonly DataTemplate: Partial<APITeam> = {};
```

(along with removing any mentions of "To override the defaults, set this value directly.")

Since there's quite a few structures PRs in progress at the moment, I'm thinking of keeping this PR as draft and just setting them here at once, once they get merged. This saves from having to update each PR individually, though do let me know if you would prefer an alternative approach. Happy to adapt.

ckohen:
I think keeping mention of Object.defineProperties is a good idea, but otherwise yes

packages/structures/src/channels/mixins/ChannelPinMixin.ts
```diff
@@ -16,6 +16,9 @@ export class ChannelPinMixin<
 
 	/**
 	 * The template used for removing data from the raw data stored for each Channel.
+	 *
+	 * @remarks This template has defaults, if you want to remove additional data and keep the defaults,
+	 * use `Object.defineProperties`. To override the defaults, set this value directly.
```
Qjuh, line 21:
```suggestion
	 * use `Object.defineProperties`.
```

packages/structures/src/messages/embeds/Embed.ts
```diff
@@ -13,6 +13,9 @@ import type { Partialize } from '../../utils/types.js';
 export class Embed<Omitted extends keyof APIEmbed | '' = ''> extends Structure<APIEmbed, Omitted> {
 	/**
 	 * The template used for removing data from the raw data stored for each Embed.
+	 *
+	 * @remarks This template has defaults, if you want to remove additional data and keep the defaults,
+	 * use `Object.defineProperties`. To override the defaults, set this value directly.
```
Qjuh, line 18:
```suggestion
	 * use `Object.defineProperties`.
```

packages/structures/src/messages/Message.ts
```diff
@@ -20,7 +20,10 @@ export class Message<Omitted extends keyof APIMessage | '' = 'edited_timestamp'
 	Omitted
 > {
 	/**
-	 * The template used for removing data from the raw data stored for each Message
+	 * The template used for removing data from the raw data stored for each Message.
+	 *
+	 * @remarks This template has defaults, if you want to remove additional data and keep the defaults,
+	 * use `Object.defineProperties`. To override the defaults, set this value directly.
```
Qjuh, line 26:
```suggestion
	 * use `Object.defineProperties`.
```

packages/structures/src/messages/MessageCall.ts
```diff
@@ -9,7 +9,10 @@ export class MessageCall<Omitted extends keyof APIMessageCall | '' = 'ended_time
 	Omitted
 > {
 	/**
-	 * The template used for removing data from the raw data stored for each MessageCall
+	 * The template used for removing data from the raw data stored for each MessageCall.
+	 *
+	 * @remarks This template has defaults, if you want to remove additional data and keep the defaults,
+	 * use `Object.defineProperties`. To override the defaults, set this value directly.
```
Qjuh, line 15:
```suggestion
	 * use `Object.defineProperties`.
```

packages/structures/src/messages/Reaction.ts
```diff
@@ -12,6 +12,9 @@ import type { Partialize } from '../utils/types.js';
 export class Reaction<Omitted extends keyof APIReaction | '' = ''> extends Structure<APIReaction, Omitted> {
 	/**
 	 * The template used for removing data from the raw data stored for each Reaction.
+	 *
+	 * @remarks This template has defaults, if you want to remove additional data and keep the defaults,
+	 * use `Object.defineProperties`. To override the defaults, set this value directly.
```
Qjuh, line 17:
```suggestion
	 * use `Object.defineProperties`.
```

PR #11389, opened 2026-01-20
feat(structures): add SKU structure

AsadHumayun:
This PR will add the SKU structure, based on the documentation of the structure on the Discord API, found [here](https://discord.com/developers/docs/resources/sku#sku-object-sku-structure).

This PR does not add tests for the associated structure.

This PR also adds a new `SKUFlagsBitField` to `/bitfields`.


packages/structures/src/index.ts
```diff
@@ -8,6 +8,7 @@ export * from './messages/index.js';
 export * from './polls/index.js';
 export * from './stickers/index.js';
 export * from './users/index.js';
+export * from './skus/SKU.js';
```
almeidx:
```suggestion
export * from './skus/index.js';
```

almeidx (changes requested):

packages/structures/src/index.ts
AsadHumayun: TY for catching :)

almeidx (approved):

packages/structures/src/bitfields/SKUFlagsBitField.ts
Qjuh, line 1: This needs to be exported in `bitfields/index.ts`

Qjuh (dismissed):

packages/structures/src/bitfields/SKUFlagsBitField.ts
AsadHumayun: Exported

packages/structures/src/skus/SKU.ts
```diff
@@ -0,0 +1,68 @@
...
+
+	/**
+	 * Id of the SKU
+	 */
+	public get id() {
+		return this[kData].id;
+	}
+
+	/**
+	 * {@link https://discord.com/developers/docs/resources/sku#sku-object-sku-types | Type of SKU}
```
Jiralite: Shouldn't this be wrapped in `@see` (and below too)?

AsadHumayun: The reason I did this was to mirror API docs (as they are "masked" links there) and this seems directly relevant.

I feel like this might be a bit easier to consume since it is more so 'defining' what type is... perhaps it may be better to link the discord-api-types definition through `@see` as well? Since that way it's possible to 'jump' to the definition and see the enums there?

 I've seen a mix of `@see` and 'masked' links in structures IIRC. Let me know your thoughts.

Qjuh: There are no masked links to URLs in structures, only to other classes/interfaces in structures. URL links should indeed be in `@see` blocks.

Qjuh: To be more exact: there are others now, but those slipped through my review of AutoModeration structures and should be changed too.

AsadHumayun: No probs, I'll go ahead and change those as well in this PR

AsadHumayun: Addressed

packages/structures/src/skus/SKU.ts
```diff
@@ -0,0 +1,73 @@
...
+	 */
+	public get slug() {
+		return this[kData].slug;
+	}
+
+	/**
+	 * SKU flags combined as a bitfield
+	 *
+	 * @see {@link SKUFlags}
+	 * @see {@link https://en.wikipedia.org/wiki/Bit_field}
```
Qjuh: I'd say both of these links shouldn't be here. Explanation of what a bit field is is already provided by the BitField classes. And what flags those are is also in docs of the SKUFlagsBitField class.

AsadHumayun: Addressed

Qjuh (approved):

packages/structures/src/bitfields/AttachmentFlagsBitField.ts
```diff
@@ -4,7 +4,7 @@ import { BitField } from './BitField.js';
 /**
  * Data structure that makes it easy to interact with a {@link Attachment#flags} bitfield.
  */
-export class AttachmentFlagsBitField extends BitField<keyof AttachmentFlags> {
+export class AttachmentFlagsBitField extends BitField<keyof typeof AttachmentFlags> {
```
vladfrangu, line 7: Is this
1. Intended
2. Correct?

vladfrangu (dismissed):

packages/structures/src/skus/SKU.ts
```diff
@@ -0,0 +1,70 @@
...
+		return this[kData].slug;
+	}
+
+	/**
+	 * SKU flags combined as a bitfield
+	 */
+	public get flags() {
+		const flags = this[kData].flags;
+
+		return flags ? new SKUFlagsBitField(this[kData].flags as SKUFlags) : null;
```
almeidx: This needs updating as of:
- #11404

almeidx (changes requested):

packages/structures/src/bitfields/AttachmentFlagsBitField.ts
AsadHumayun: Intended - this was a correction that I decided to add to all bitfields as part of this PR. See https://github.com/discordjs/discord.js/pull/11393#discussion_r2724631125


> ```suggestion
> export class ApplicationFlagsBitField extends BitField<keyof typeof ApplicationFlags> {
> ```
> This is indeed wrong in the other bitfields and needs fixing there too.
> 

_Originally posted by @Qjuh in https://github.com/discordjs/discord.js/pull/11393#discussion_r2724631125_

packages/structures/src/skus/SKU.ts
AsadHumayun: Addressed (25d86fa)

almeidx (approved):

Jiralite (approved):

PR #11390, opened 2026-01-22
feat(structures): add SoundboardSound structure

AsadHumayun:
This PR will add the `SoundboardSound` API structure, as documented on the Discord API documentation [here](https://discord.com/developers/docs/resources/soundboard#soundboard-sound-object-soundboard-sound-structure).

This PR does not add tests for this structure.

Thanks to @edocsil47 for helping on how to determine if a guild id is present within the structure.

### DIFF
--- a/packages/structures/src/soundboards/SoundboardSound.ts
+++ b/packages/structures/src/soundboards/SoundboardSound.ts
@@ -20,7 +20,7 @@ export class SoundboardSound<Omitted extends keyof APISoundboardSound | '' = ''>
 	public static override readonly DataTemplate: Partial<APISoundboardSound> = {};
 
 	/**
-	 * @param data - The raw data received from the API for the emoji
+	 * @param data - The raw data received from the API for the soundboard sound
 	 */
 	public constructor(data: Partialize<APISoundboardSound, Omitted>) {
 		super(data);
--- a/packages/structures/src/soundboards/SoundboardSound.ts
+++ b/packages/structures/src/soundboards/SoundboardSound.ts
@@ -78,6 +78,8 @@ export class SoundboardSound<Omitted extends keyof APISoundboardSound | '' = ''>
 
 	/**
 	 * The timestamp this sound was created at
+	 *
+	 * @remarks only available for guild soundboard sounds
 	 */
 	public get createdTimestamp() {
 		return isIdSet(this[kData].sound_id) && isIdSet(this[kData].guild_id)
@@ -87,6 +89,8 @@ export class SoundboardSound<Omitted extends keyof APISoundboardSound | '' = ''>
 
 	/**
 	 * The time this sound was created at
+	 *
+	 * @remarks only available for guild soundboard sounds
 	 */
 	public get createdAt() {
 		const createdTimestamp = this.createdTimestamp;
--- a/packages/structures/src/soundboards/SoundboardSound.ts
+++ b/packages/structures/src/soundboards/SoundboardSound.ts
@@ -79,10 +79,9 @@ export class SoundboardSound<Omitted extends keyof APISoundboardSound | '' = ''>
 	 * The timestamp this sound was created at
 	 */
 	public get createdTimestamp() {
-		const id = this[kData].sound_id;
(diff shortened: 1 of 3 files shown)


packages/structures/src/soundboards/SoundboardSound.ts
```diff
@@ -0,0 +1,96 @@
...
+	 * The template used for removing data from the raw data stored for each soundboard sound.
+	 */
+	public static override readonly DataTemplate: Partial<APISoundboardSound> = {};
+
+	/**
+	 * @param data - The raw data received from the API for the emoji
+	 */
+	public constructor(data: Partialize<APISoundboardSound, Omitted>) {
+		super(data);
+	}
```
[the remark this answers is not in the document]
AsadHumayun: Corrected, TY for catching

packages/structures/src/soundboards/SoundboardSound.ts
```diff
@@ -0,0 +1,96 @@
...
+	 */
+	public get available() {
+		return this[kData].available;
+	}
+
+	/**
+	 * The timestamp this sound was created at
+	 */
+	public get createdTimestamp() {
+		const id = this[kData].sound_id;
```
Qjuh: This variable is no longer needed, just put the value directly into the `timestampFrom()` call.

Qjuh (dismissed):

packages/structures/src/soundboards/SoundboardSound.ts
AsadHumayun: Addressed

packages/structures/src/soundboards/SoundboardSound.ts
```diff
@@ -0,0 +1,95 @@
...
+		return this[kData].available;
+	}
+
+	/**
+	 * The timestamp this sound was created at
+	 */
+	public get createdTimestamp() {
+		const inGuild = this[kData].guild_id && typeof this[kData].guild_id === 'string';
+
+		return inGuild ? DiscordSnowflake.timestampFrom(this[kData].sound_id as string) : null;
```
almeidx: We should use `isIdSet()` here instead

```suggestion
		return isIdSet(this[kData].guild_id) && isIdSet(this[kData].sound_id) ? DiscordSnowflake.timestampFrom(this[kData].sound_id) : null;
```

almeidx (changes requested):

packages/structures/src/soundboards/SoundboardSound.ts
```diff
@@ -0,0 +1,95 @@
...
+	}
+
+	/**
+	 * The time this sound was created at
+	 */
+	public get createdAt() {
+		const timestamp = this.createdTimestamp;
+
+		return timestamp ? new Date(timestamp) : null;
```
almeidx: Consistency with 
https://github.com/discordjs/discord.js/blob/838cd2da34d99dadeae1438a019fc1b4edc703bd/packages/structures/src/channels/Channel.ts#L100-L103

```suggestion
		const createdTimestamp = this.createdTimestamp;
		return createdTimestamp ? new Date(createdTimestamp) : null;
```

packages/structures/src/soundboards/SoundboardSound.ts
Qjuh: Why is the cast needed at all here? If the sound_id could be something else then the conditional should check it with isIdSet() too. If not then the cast is redundant

almeidx: Edited the suggestion. Also, I think we should add a remark to this and `createdAt` informing that this will only be available for guild soundboard sounds

packages/structures/src/soundboards/SoundboardSound.ts
AsadHumayun: Addressed

packages/structures/src/soundboards/SoundboardSound.ts
AsadHumayun: Both addressed

almeidx (approved):

Qjuh (approved):

vladfrangu (approved):

PR #11392, opened 2026-01-22
feat(invite): new endpoints and target users file and roles

### LINKED ISSUE #11587: support `role_ids` on `GuildInviteManager#create` (Community Invites auto-role assignment)

discord.js

### Feature

Discord's REST API supports Community Invites, which include a `role_ids` field on invite creation (`POST /channels/{channel.id}/invites`), an array of role IDs automatically granted to a user when they accept the invite.
See: https://docs.discord.com/developers/tutorials/using-community-invites

`discord-api-types` already types this (`role_ids` on `RESTPostAPIChannelInviteJSONBody`), but `GuildInviteManager#create` doesn't expose it. The options currently destructured/sent are limited to `temporary, maxAge, maxUses, unique, targetUser, targetApplication, targetType`, with no `roleIds` passthrough.

Add a `roleIds` option to `InviteCreateOptions` that maps to `role_ids` in the JSON body sent by `GuildInviteManager#create`.

None; this is a straightforward passthrough of an already-typed REST field.

### Other context

`target_users_file` (CSV upload restricting who can accept) is also part of Community Invites but would need multipart/form-data support that `GuildInviteManager` doesn't currently have, so probably out of scope for a first pass.

imnaiyar:
Depends on 
- https://github.com/discordjs/discord-api-types/pull/1493
- #11555


packages/discord.js/src/managers/GuildInviteManager.js
```diff
@@ -202,23 +214,38 @@ class GuildInviteManager extends CachedManager {
...
-
+    const options = {
+      temporary,
+      max_age: maxAge,
+      max_uses: maxUses,
+      unique,
+      target_user_id: this.client.users.resolveId(targetUser),
+      target_application_id: targetApplication?.id ?? targetApplication?.applicationId ?? targetApplication,
+      role_ids: roles?.map(role => this.guild.roles.resolveId(role)),
+      target_type: targetType,
```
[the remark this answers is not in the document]
imnaiyar: If invalid role is provided, api will return errors appropriately. Imo we don't need to do anything else here ourselves

packages/discord.js/typings/index.d.ts
```diff
@@ -3026,32 +3031,40 @@ export interface RoleColorsEditResolvable {
   tertiaryColor?: ColorResolvable | null;
 }
 
-export class Role extends Base {
-  private constructor(client: Client<true>, data: APIRole, guild: Guild);
+export class BaseRole extends Base {
+  protected constructor(client: Client<true>, data: APIRole);
```
Qjuh: Technically the expected type is only a partial of APIRole here.

packages/discord.js/src/managers/GuildInviteManager.js
```diff
@@ -234,6 +262,49 @@ class GuildInviteManager extends CachedManager {
...
+    const arrayBuff = await this.client.rest.get(Routes.inviteTargetUsers(code));
+
+    return Buffer.from(arrayBuff);
+  }
+
+  /**
+   * Updates target users for an invite
+   *
+   * @param {InviteResolvable} invite The invite to update the target users
+   * @param {UserResolvable[]|BufferResolvable} targetUsersFile An array of users or a csv file with a single column of user IDs
```
Jiralite:
```suggestion
   * @param {UserResolvable[]|BufferResolvable} targetUsersFile An array of users or a CSV file with a single column of user ids
```

packages/discord.js/src/structures/BaseGuildTextChannel.js
```diff
@@ -161,6 +161,9 @@ class BaseGuildTextChannel extends GuildChannel {
    * required if `targetType` is {@link InviteTargetType.Stream}, the application must have the
    * {@link InviteTargetType.EmbeddedApplication} flag
    * @property {InviteTargetType} [targetType] The type of the target for this voice channel invite
+   * @property {UserResolvable[]|BufferResolvable} [targetUsersFile] An array of users or a csv file with a single column of user IDs
```
Jiralite:
```suggestion
   * @property {UserResolvable[]|BufferResolvable} [targetUsersFile] An array of users or a CSV file with a single column of user ids
```

packages/discord.js/src/structures/GuildInvite.js
```diff
@@ -206,6 +229,42 @@ class GuildInvite extends BaseInvite {
     await this.client.rest.delete(Routes.invite(this.code), { reason });
   }
 
+  /**
+   * Update target users of this invite.
+   *
+   * @param {UserResolvable[]|BufferResolvable} targetUsersFile An array of users or a csv file with a single column of user IDs
```
Jiralite:
```suggestion
   * @param {UserResolvable[]|BufferResolvable} targetUsersFile An array of users or a CSV file with a single column of user ids
```

Jiralite (changes requested):

packages/discord.js/src/structures/GuildInvite.js
```diff
@@ -178,6 +185,22 @@ class GuildInvite extends BaseInvite {
...
     }
+
+    if ('roles' in data) {
+      /**
+       * The roles assigned to the user upon accepting the invite.
+       *
+       * @type {?Collection<Snowflake, Role|InviteRole>}
+       */
+      this.roles = new Collection(
+        data.roles.map(role => [
```
Jiralite: `data.roles` may contain `null`:

https://github.com/discord/discord-api-spec/blob/c9423c8bf43f5fb9d1d63a6c37a316302bbc1a84/specs/openapi.json#L27180

packages/discord.js/src/structures/BaseRole.js
```diff
@@ -0,0 +1,144 @@
...
+   *
+   * @returns {string}
+   * @example
+   * // Logs: Role: <@&123456789012345678>
+   * console.log(`Role: ${role}`);
+   */
+  toString() {
+    if (this.id === this.guild.id) return '@everyone';
+    return roleMention(this.id);
+  }
```
Jiralite, line 136: Stringifying an invite role will crash. There's no guild there.

Jiralite (changes requested):
`roles` is detected via REST, but not `role_ids` from the gateway, which means support for that is missing.

packages/discord.js/src/structures/GuildInvite.js
```diff
@@ -216,6 +277,7 @@ class GuildInvite extends BaseInvite {
       channel: 'channelId',
       inviter: 'inviterId',
       guild: 'guildId',
+      roles: 'roles',
```
Qjuh: This is not needed, as those properties are for remapping to different keys. But here the key is the same so it would work directly.
But it seems this whole call is wrong since https://github.com/discordjs/discord.js/pull/10888 because `BaseInvite#toJSON()` doesn't accept any parameters. Could you please fix that by either making `BaseInvite#toJSON()` accept a rest parameter that gets spread into the object passed to `Base#toJSON(...)` or by spreading the additional properties into the result of `BaseInvite#toJSON()` similar to how `Role#toJSON()` does it?

Qjuh (changes requested):

packages/discord.js/src/structures/GuildInvite.js
imnaiyar: I had removed that earlier, it seems my rebase messed of some of the commits. I will take a look at this later today

Qjuh (approved):

Qjuh (approved):

packages/discord.js/src/managers/GuildInviteManager.js
```diff
@@ -191,7 +204,7 @@ class GuildInviteManager extends CachedManager {
   /**
    * Create an invite to the guild from the provided channel.
    *
-   * @param {GuildInvitableChannelResolvable} channel The options for creating the invite from a channel.
+   * @param {GuildInvitableChannelResolvable} channel The channel where invite should be created.
```
Qjuh:
```suggestion
   * @param {GuildInvitableChannelResolvable} channel The channel where the invite should be created.
```

Qjuh (changes requested):

Qjuh (approved):

packages/discord.js/src/managers/GuildInviteManager.js
```diff
@@ -234,6 +262,49 @@ class GuildInviteManager extends CachedManager {
...
+    return Buffer.from(arrayBuff);
+  }
+
+  /**
+   * Updates target users for an invite
+   *
+   * @param {InviteResolvable} invite The invite to update the target users
+   * @param {UserResolvable[]|BufferResolvable} targetUsersFile An array of users or a CSV file with a single column of user ids
+   * for all the users able to accept this invite
+   * @returns {Promise<unknown>}
```
Qjuh: just saw this, same question as on v14 PR: why unknown?

imnaiyar: i maybe misremembering, but pretty sure at the time this pr was made, all the accepted response that didn't return anything in dtypes was typed as `unknown` (or `never` ?) before it was changed to `undefined`. Hence that, I just forgot to update it after the change

Qjuh (approved):

packages/core/src/api/channel.ts
```diff
@@ -492,7 +493,7 @@ export class ChannelsAPI {
 	 */
 	public async createInvite(
 		channelId: Snowflake,
-		body: RESTPostAPIChannelInviteJSONBody,
```
Jiralite, line 495: Did you test this? Does not work.

In addition, should this not be using `files`?

packages/discord.js/typings/index.d.ts
```diff
@@ -2084,7 +2085,7 @@ export class BaseInvite<WithCounts extends boolean = boolean> extends Base {
   public get url(): string;
   public static InvitesPattern: RegExp;
   public toString(): string;
-  public toJSON(): unknown;
+  public toJSON(props?: Record<string, boolean | string>): unknown;
```
Jiralite, line 2089: Doing this means whatever extends it will also have this because it is not overridden there. So now `GuildInvite.toJSON({ roles: false });` is permitted in TypeScript but does not change the behaviour.

packages/discord.js/src/util/DataResolver.js
```diff
@@ -11,6 +11,30 @@ const { BaseInvite } = require('../structures/BaseInvite.js');
...
+    usersCsv = targetUsersFile.map(user => client.users.resolveId(user)).join('\n');
+  } else {
+    const resolved = await resolveFile(targetUsersFile);
+    usersCsv = resolved.data.toString('utf8');
+  }
+
+  formData.append('target_users_file', new Blob([usersCsv], { type: 'text/csv' }), 'users.csv');
+  formData.append('payload_json', JSON.stringify(rest));
+  return formData;
+}
```
Jiralite: This is just a duplicate of `files`, no? It does not look like this is needed?

packages/discord.js/src/structures/InviteRole.js
```diff
@@ -0,0 +1,12 @@
+'use strict';
+
+const { BaseRole } = require('./BaseRole');
```
Jiralite: This is introducing an inconsistency:


```suggestion
const { BaseRole } = require('./BaseRole.js');
```

packages/discord.js/typings/index.d.ts
```diff
@@ -3029,32 +3034,40 @@ export interface RoleColorsEditResolvable {
   tertiaryColor?: ColorResolvable | null;
 }
 
-export class Role extends Base {
-  private constructor(client: Client<true>, data: APIRole, guild: Guild);
+export class BaseRole extends Base {
+  protected constructor(client: Client<true>, data: Partial<APIRole>);
```
Jiralite: Pretty sure this should be `APIInviteRole | APIRole`, no?

packages/core/src/api/invite.ts
Jiralite, line 1: No `auth`?

Jiralite (dismissed):

packages/discord.js/src/util/DataResolver.js
imnaiyar: you are right, looking at the REST source code, this seems to be unnecessary

packages/discord.js/typings/index.d.ts
```diff
@@ -2103,6 +2106,9 @@ export class GuildInvite<WithCounts extends boolean = boolean> extends BaseInvit
   public approximatePresenceCount: WithCounts extends true ? number : null;
   public get deletable(): boolean;
   public delete(reason?: string): Promise<void>;
+  public updateTargetUsers(targetUsersFile: BufferResolvable | readonly UserResolvable[]): Promise<void>;
+  public fetchTargetUsersJobStatus(): Promise<TargetUsersJobStatusForInvite>;
+  public fetchTargetUsers(): Promise<Buffer>;
```
Qjuh, line 2111:
```suggestion
  public fetchTargetUsers(): Promise<Buffer>;
  public toJSON(): unknown;
```

Qjuh (changes requested):

Qjuh (approved):

PR #11393, opened 2026-01-23
feat(structures): add Application structure

keston-dev:
This PR will add the `Application` API structure, per [Discord's API documentation](https://discord.com/developers/docs/resources/application), as well as the `ApplicationFlagsBitField` to `/bitfields`.

Mentioning #10981 for visibility.

PR #11396, opened 2026-01-24
feat(structures): add Team structure

AsadHumayun:
This PR will add the `Team`s structure and `TeamMember`s substructure.

Note: if #11393 is merged first, then the TODO comment re. Teams structure should be removed.

### DIFF
--- a/packages/structures/src/teams/TeamMember.ts
+++ b/packages/structures/src/teams/TeamMember.ts
@@ -1,0 +1,43 @@
+import type { APITeamMember, TeamMemberMembershipState, TeamMemberRole } from 'discord-api-types/v10';
+import { Structure } from '../Structure.js';
+import { kData } from '../utils/symbols.js';
+import type { Partialize } from '../utils/types.js';
+
+/**
+ * Represents any team member on Discord.
+ *
+ * @typeParam Omitted - Specify the properties that will not be stored in the raw data field as a union, implement via `DataTemplate`
+ * @remarks has substructure `User` which needs to be instantiated and stored by an extending class using it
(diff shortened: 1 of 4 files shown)


almeidx (approved):

packages/structures/src/teams/TeamMember.ts
```diff
@@ -0,0 +1,49 @@
...
+
+	/**
+	 * Id of the parent team of which they are a member
+	 */
+	public get teamId() {
+		return this[kData].team_id;
+	}
+
+	/**
+	 * {@link https://discord.com/developers/docs/topics/teams#team-member-roles-team-member-role-types | Role} of the team member
```
Qjuh: This link should be in a `@see` block

packages/structures/src/teams/TeamMember.ts
```diff
@@ -0,0 +1,49 @@
...
+	 *
+	 * @see {@link TeamMemberRole}
```
Qjuh: This is redundant, since that is already the return type of this getter.

Qjuh (changes requested):

packages/structures/src/teams/TeamMember.ts
AsadHumayun: Sorry, I did a lot of these. Addressed.

packages/structures/src/teams/TeamMember.ts
AsadHumayun: Removed

Qjuh (approved):

vladfrangu (approved):

PR #11397, opened 2026-01-24
feat(structures): add Stage Instance structure

AsadHumayun:
This PR will add the Stage Instance structure.

### DIFF
--- a/packages/structures/src/stageInstances/StageInstance.ts
+++ b/packages/structures/src/stageInstances/StageInstance.ts
@@ -1,0 +1,97 @@
+import { DiscordSnowflake } from '@sapphire/snowflake';
+import type { APIStageInstance, StageInstancePrivacyLevel } from 'discord-api-types/v10';
+import { Structure } from '../Structure.js';
+import { kData } from '../utils/symbols.js';
+import { isIdSet } from '../utils/type-guards.js';
+import type { Partialize } from '../utils/types.js';
+
+/**
+ * Represents any stage instance on Discord.
+ *
+ * @typeParam Omitted - Specify the properties that will not be stored in the raw data field as a union, implement via `DataTemplate`
+ */
+export class StageInstance<Omitted extends keyof APIStageInstance | '' = ''> extends Structure<
+	APIStageInstance,
+	Omitted
+> {
+	/**
+	 * The template used for removing data from the raw data stored for each stage instance
+	 */
+	public static override readonly DataTemplate: Partial<APIStageInstance> = {};
+
+	/**
+	 * @param data - The raw data received from the API for the stage instance
+	 */
+	public constructor(data: Partialize<APIStageInstance, Omitted>) {
+		super(data);
+	}
+
+	/**
+	 * The stage instance's id
+	 */
+	public get id() {
+		return this[kData].id;
+	}
+
+	/**
+	 * The guild id of the associated stage channel
+	 */
+	public get guildId() {
+		return this[kData].guild_id;
+	}
+
+	/**
+	 * The id of the associated stage channel
+	 */
+	public get channelId() {
+		return this[kData].channel_id;
+	}
+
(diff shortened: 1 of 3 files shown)


almeidx (approved):

packages/structures/src/stageInstances/StageInstance.ts
```diff
@@ -0,0 +1,97 @@
...
+	/**
+	 * Whether the stage discovery is disabled (deprecated)
+	 *
+	 * @deprecated
+	 * {@link https://github.com/discord/discord-api-docs/pull/4296 | discord-api-docs#4296}
+	 */
+	public get discoverableDisabled() {
+		return this[kData].discoverable_disabled;
+	}
+
```
Qjuh:
```suggestion
```
We shouldn't add a getter for a deprecated field.

packages/structures/src/stageInstances/StageInstance.ts
```diff
@@ -0,0 +1,97 @@
...
+	/**
+	 * The topic of the stage instance (1-120 characters)
+	 */
+	public get topic() {
+		return this[kData].topic;
+	}
+
+	/**
+	 * The {@link https://discord.com/developers/docs/resources/stage-instance#stage-instance-object-privacy-level | privacy level} of the stage instance
```
Qjuh:
```suggestion
	 * The {@link StageInstancePrivacyLevel} of the stage instance
```

URL links should only be in `@see` blocks.

packages/structures/src/stageInstances/StageInstance.ts
```diff
@@ -0,0 +1,97 @@
...
+	 *
+	 * @see {@link StageInstancePrivacyLevel}
```
Qjuh: redundant, since this is already the return type

Qjuh (dismissed):

packages/structures/src/stageInstances/StageInstance.ts
AsadHumayun: Might it be beneficial to add a remark on the class stating the deprecated fields (`discoverableDisabled `) are not exposed?

packages/structures/src/stageInstances/StageInstance.ts
AsadHumayun: Is there any preference/convention as to whether links to discord-api-types should be used, or Discord API URLs?

Qjuh: Links to discord-api-types are not needed at all, as soon as the type is part of the return type or parameter type, since they get picked up by tsdoc automatically then. And discord-api-types links to the Discord API URLs where appropriate afaik.

packages/structures/src/stageInstances/StageInstance.ts
Qjuh: I'd raise that question to "should deprecated fields be omitted and be part of DataTemplate". But that's a question for core maintainers and @ckohen I guess.

packages/structures/src/stageInstances/StageInstance.ts
AsadHumayun: Got it, thank you :)

packages/structures/src/stageInstances/StageInstance.ts
AsadHumayun: Addressed

packages/structures/src/stageInstances/StageInstance.ts
AsadHumayun: Addressed

packages/structures/src/stageInstances/StageInstance.ts
AsadHumayun: Removed

Qjuh (approved):

vladfrangu (approved):

PR #11398, opened 2026-01-24
feat(structures): add StickerPack structure

AsadHumayun:
This PR will add the `StickerPack` structure.

`createdAt` and `createdTimestamp` were missing on the `Sticker` structure, so I went ahead and added that as part of this PR as well. We should be able to derive these properties as a snowflake id is present for that structure.

This PR does not add tests.


packages/structures/src/stickers/StickerPack.ts
```diff
@@ -0,0 +1,83 @@
...
+
+	/**
+	 * The description of the sticker pack
+	 */
+	public get description() {
+		return this[kData].description;
+	}
+
+	/**
+	 * The id of the sticker pack's {@link https://discord.com/developers/docs/reference#image-formatting | banner image}
```
Qjuh: URL links should be in `@see {@link ...}` blocks

Qjuh (changes requested):

packages/structures/src/stickers/StickerPack.ts
AsadHumayun: Addressed

Qjuh (approved):

almeidx (approved):

vladfrangu (approved):

PR #11399, opened 2026-01-25
feat(structures): add Subscription structure

AsadHumayun:
This PR will add the `Subscription` structure.

This PR does not include tests (will try to help write tests once it gets to that stage)

Note that I did add a couple new symbols for this structure - we _could_ have just used the `kEndsTimestamp` and `kStartsTimestamp` symbols, but those were not 100% clear regarding its domain and regardless I would have had to add an extra one anyway (`kCanceledTimestamp`).

### DIFF
--- a/packages/structures/src/subscriptions/Subscription.ts
+++ b/packages/structures/src/subscriptions/Subscription.ts
@@ -17,7 +17,9 @@ import type { Partialize } from '../utils/types.js';
  * @remarks intentionally does not export `skuIds`, `entitlementIds`, and `renewalSkuIds` so that extending classes can resolve
  * `Snowflake[]` to `SKU[]`, `Entitlement[]`, and `?SKU[]`, respectively
  */
-export class Subscription<Omitted extends keyof APISubscription | '' = ''> extends Structure<APISubscription, Omitted> {
+export class Subscription<
+	Omitted extends keyof APISubscription | '' = 'current_period_end' | 'current_period_start',
+> extends Structure<APISubscription, Omitted> {
 	/**
 	 * The template used for removing data from the raw data stored for each subscription
 	 */
--- a/packages/structures/src/subscriptions/Subscription.ts
+++ b/packages/structures/src/subscriptions/Subscription.ts
@@ -146,7 +146,7 @@ export class Subscription<
 	/**
 	 * The time when the subscription was canceled
 	 *
- * @remarks This is populated when the {@link Subscription#status} transitions to {@link SubscriptionStatus.Ending}.
+	 * @remarks This is populated when the {@link Subscription#status} transitions to {@link SubscriptionStatus.Ending}.
 	 */
 	public get canceledAt() {
 		const canceledTimestamp = this.canceledTimestamp;
--- a/packages/structures/src/subscriptions/Subscription.ts
+++ b/packages/structures/src/subscriptions/Subscription.ts
@@ -1,0 +1,155 @@
+import { DiscordSnowflake } from '@sapphire/snowflake';
+import type { APISubscription, SubscriptionStatus } from 'discord-api-types/v10';
+import { Structure } from '../Structure.js';
+import {
+	kData,
+	kCurrentPeriodStartTimestamp,
+	kCurrentPeriodEndTimestamp,
+	kCanceledTimestamp,
+} from '../utils/symbols.js';
+import { isIdSet } from '../utils/type-guards.js';
+import type { Partialize } from '../utils/types.js';
+
+/**
+ * Represents any subscription on Discord.
+ *
+ * @typeParam Omitted - Specify the properties that will not be stored in the raw data field as a union, implement via `DataTemplate`
+ * @remarks intentionally does not export `skuIds`, `entitlementIds`, and `renewalSkuIds` so that extending classes can resolve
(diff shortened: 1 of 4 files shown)


packages/structures/src/subscriptions/Subscription.ts
```diff
@@ -0,0 +1,155 @@
...
+	/**
+	 * Id of the user who is subscribed
+	 */
+	public get userId() {
+		return this[kData].user_id;
+	}
+
+	/**
+	 * Tmestamp of start of the current subscription period
+	 */
```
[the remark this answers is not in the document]
AsadHumayun: TY for catching, addressed.

packages/structures/src/subscriptions/Subscription.ts
```diff
@@ -0,0 +1,157 @@
...
+} from '../utils/symbols.js';
+import { isIdSet } from '../utils/type-guards.js';
+import type { Partialize } from '../utils/types.js';
+
+/**
+ * Represents any subscription on Discord.
+ *
+ * @typeParam Omitted - Specify the properties that will not be stored in the raw data field as a union, implement via `DataTemplate`
+ * @remarks intentionally does not export `skuIds`, `entitlementIds`, and `renewalSkuIds` so that extending classes can resolve
+ * `Snowflake[]` to `SKU[]`, `Entitlement[]`, and `?SKU[]`, respectively
```
Qjuh: Considering those fields are named with the `Ids` suffix I'd say we should expose them as getters returning readonly arrays.  Extending classes would then expose the full structures as `skus`, `entitlements` and `renewalSkus`.

packages/structures/src/subscriptions/Subscription.ts
```diff
@@ -0,0 +1,155 @@
...
+			this[kCurrentPeriodStartTimestamp] = Date.parse(currentPeriodStartTimestamp);
+		}
+
+		if (currentPeriodEndTimestamp) {
+			this[kCurrentPeriodEndTimestamp] = Date.parse(currentPeriodEndTimestamp);
+		}
+
+		if (canceledTimestamp) {
+			this[kCanceledTimestamp] = Date.parse(canceledTimestamp);
+		}
```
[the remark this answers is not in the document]
Qjuh: The fields get checked against being `null` in the if conditions, so this is wrong.

Qjuh (changes requested):

packages/structures/src/subscriptions/Subscription.ts
AsadHumayun: I agree, Addressed

packages/structures/src/subscriptions/Subscription.ts
```diff
@@ -0,0 +1,176 @@
...
+export class Subscription<
+	Omitted extends keyof APISubscription | '' = 'current_period_end' | 'current_period_start',
+> extends Structure<APISubscription, Omitted> {
+	/**
+	 * The template used for removing data from the raw data stored for each subscription
+	 */
+	public static override readonly DataTemplate: Partial<APISubscription> = {
+		set current_period_start(_: string) {},
+		set current_period_end(_: string) {},
+	};
```
almeidx, line 28: Missing `canceled_at`. Also applies to the default value of the type parameter on the class

almeidx (changes requested):

packages/structures/src/subscriptions/Subscription.ts
AsadHumayun: `canceled_at` can be null so I thought it shouldn't be on the defaults?

almeidx: The DataTemplate is enforced on patch too, not just on initial creation. It should be omitted from the structure's data since we're optimising it

Qjuh: It should still be optimized as timestamp and thus omitted from the internal data. The other timestamps also already check for being nullish and act accordingly.

AsadHumayun: Sorry about that, yeah that makes sense. Still trying to understand everything haha.. addressed.

Qjuh (approved):

packages/structures/src/subscriptions/Subscription.ts
```diff
@@ -0,0 +1,177 @@
...
+	 * Timestamp when the subscription was canceled
+	 */
+	public get canceledTimestamp() {
+		return this[kCanceledTimestamp];
+	}
+
+	/**
+	 * The time when the subscription was canceled
+	 *
+	 * @remarks If the user cancels the subscription, the subscription will enter the {@link SubscriptionStatus.Ending} status and the `canceled_at` timestamp will reflect the time of the cancellation.
```
almeidx: This remark seems all over the place. This is already on `canceledAt` getter, so perhaps something more like this?

```suggestion
 * @remarks This is populated when the {@link Subscription#status} transitions to {@link SubscriptionStatus.Ending}.
```

almeidx (approved):

AsadHumayun:
Thanks for that, I wanted to commit your suggestion so you got credit for that (I guess it didn't like the formatting); thank you for fixing.

vladfrangu (approved):

PR #11410, opened 2026-01-30
feat(builders): add checkbox, checkboxgroup, and radiogroup builders

faceboy1392:
these Discord features are (as of writing this) still in early access, also needs the newest version of discord-api-types

this is my first time contributing to anything so sorry if idk what I'm doing

### DIFF
--- a/.github/workflows/documentation.yml
+++ b/.github/workflows/documentation.yml
@@ -1,8 +1,6 @@
 name: Documentation
 on:
   push:
-    branches:
-      - 'main'
     paths:
       - 'packages/*/src/**'
       - '!packages/create-discord-bot/**'
@@ -13,14 +11,7 @@ on:
   workflow_dispatch:
     inputs:
       ref:
-        description: 'The branch, tag or SHA to checkout'
-        required: true
-      ref_type:
-        type: choice
-        description: 'Branch or tag'
-        options:
-          - branch
-          - tag
+        description: 'The tag to checkout'
         required: true
 concurrency:
   group: ${{ github.workflow }}-${{ github.head_ref || github.ref }}
@@ -32,7 +23,6 @@ jobs:
     env:
       TURBO_TOKEN: ${{ secrets.TURBO_TOKEN }}
       TURBO_TEAM: ${{ secrets.TURBO_TEAM }}
-      REF_TYPE: ${{ inputs.ref_type || github.ref_type }}
     if: github.repository_owner == 'discordjs'
     steps:
       - name: Checkout repository
@@ -52,13 +42,11 @@ jobs:
         run: pnpm run build
 
       - name: Checkout main repository
-        if: ${{ inputs.ref && inputs.ref != 'main' }}
         uses: actions/checkout@v4
         with:
           path: 'main'
 
       - name: Build main
-        if: ${{ inputs.ref && inputs.ref != 'main' }}
         shell: bash
         run: |
           cd main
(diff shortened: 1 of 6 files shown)


packages/builders/src/components/checkbox/Assertions.ts
```diff
@@ -0,0 +1,112 @@
...
+		custom_id: customIdValidator,
+		id: idValidator.optional(),
+		options: s.array(radioGroupOptionPredicate).lengthGreaterThanOrEqual(2).lengthLessThanOrEqual(10),
+		required: s.boolean().optional(),
+	})
+	.reshape((data) => {
+		// Ensure there is exactly one default option
+		const defaultCount = data.options.filter((option) => option.default === true).length;
+		if (defaultCount > 1) {
+			throw new RangeError('There must be exactly one default option in a radio group');
```
Qjuh:
```suggestion
			throw new RangeError('There can be at most one default option in a radio group');
```
There doesn't have to be a default set.

packages/builders/src/components/checkbox/RadioGroup.ts
```diff
@@ -0,0 +1,154 @@
...
+
+	/**
+	 * Adds options to this radio group.
+	 *
+	 * @param options - The options to add
+	 */
+	public addOptions(...options: RestOrArray<APIRadioGroupOption | RadioGroupOptionBuilder>) {
+		const normalizedOptions = normalizeArray(options);
+
+		radioGroupOptionsLengthValidator.parse(this.options.length + normalizedOptions.length);
```
Qjuh: This shouldn't validate on `addOptions` call already, because then it'll throw as soon as you add options one by one, because having only one options is invalid. Same applies to checkbox group but less severe. But even there a builder should be allowed to be in a state where it has more or less options than allowed, as long as that changes before the builder gets turned into JSON.

faceboy1392: Should I also remove that validation from spliceOptions? I'm going to assume so

Qjuh: Yes

imaai:
Is there any ETA for this one?

Smidul:
> Is there any ETA for this one?

DJS never provides ETAs. Feature PRs usually sit for about two weeks to a month after release before maintainers decide to merge them.

packages/builders/__tests__/components/checkbox.test.ts
Jiralite, line 1: Why are radio components in here? Seems the filename is misleading?

faceboy1392: I kinda refer to the whole set as "checkbox components", even if technically only two of them are actually checkboxes. I'm not really sure what else to call this group of components as a whole. I similarly have the five classes (3 components and 2 option types) in src/components/checkbox/, so the test file is named consistently with that folder. Should I rename the file/maybe split it up into 2-3 more precisely-named files?

Jiralite: Actually, on second thought, since it's v14, I'm not too worried about it.

packages/builders/src/components/checkbox/Checkbox.ts
```diff
@@ -0,0 +1,63 @@
...
+	 * 	.setCustomId('subscribe_newsletter')
+	 * 	.setDefault(true);
+	 * ```
+	 */
+	public constructor(data?: Partial<APICheckboxComponent>) {
+		super({ type: ComponentType.Checkbox, ...data });
+	}
+
+	/**
+	 * Sets the custom ID of this checkbox.
```
Jiralite: `ID` should be lowercase everywhere!

packages/builders/src/components/checkbox/CheckboxGroup.ts
```diff
@@ -0,0 +1,174 @@
...
+	public setOptions(options: RestOrArray<APICheckboxGroupOption | CheckboxGroupOptionBuilder>) {
+		return this.spliceOptions(0, this.options.length, ...options);
+	}
+
+	/**
+	 * Removes, replaces, or inserts options for this checkbox group.
+	 *
+	 * @remarks
+	 * This method behaves similarly
+	 * to {@link https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/slice | Array.prototype.splice()}.
```
Jiralite: Everywhere else too:

```suggestion
	 * to {@link https://developer.mozilla.org/docs/Web/JavaScript/Reference/Global_Objects/Array/slice | Array.prototype.splice()}.
```

faceboy1392: I copied that from the StringSelectMenuBuilder, though now that I look at it, that links to slice, not splice, shouldn't it be splice? if so then its also wrong in StringSelectMenuBuilder

Jiralite: Yeah, looks like a documentation issue. Seems to be present on main too...

faceboy1392: Should I fix that now for StringSelectMenuBuilder too or is that outside of the scope of a checkbox components PR? Im still learning OSS contribution etiquette

Jiralite: Main pull request to fix it there, then it's backported into v14.

In this pull request, you can fix it for radios/checkboxes.

packages/builders/src/components/checkbox/Assertions.ts
```diff
@@ -0,0 +1,98 @@
...
+		id: idValidator.optional(),
+		options: s.array(checkboxGroupOptionPredicate).lengthGreaterThanOrEqual(1).lengthLessThanOrEqual(10),
+		min_values: s.number().int().greaterThanOrEqual(0).lessThanOrEqual(10).optional(),
+		max_values: s.number().int().greaterThanOrEqual(1).lessThanOrEqual(10).optional(),
+		required: s.boolean().optional(),
+	})
+	.reshape((data) => {
+		// Ensure min_values is not greater than max_values
+		if (data.min_values !== undefined && data.max_values !== undefined && data.min_values > data.max_values) {
+			throw new RangeError('min_values cannot be greater than max_values');
```
vladfrangu: Aren't you supposed to return Result.err? https://github.com/sapphiredev/shapeshift#reshapervalue-t--resultr-error--iconstraint-nopvalidatorr-adds-a-constraint-able-to-both-validate

vladfrangu (changes requested):

packages/builders/src/components/checkbox/Assertions.ts
faceboy1392: Oh yea that makes sense
ngl I'm kinda dumb and somehow couldn't find the docs for this, I kept googling sapphire instead of shapeshift and thought that the bot framework I saw was completely unrelated, and then I misread some of the other builders code that I was half copying from

PR #11406, opened 2026-01-27
feat(structures): add url getters for image assets

Qjuh:
Adds `url` getters or `*URL(format)` methods for getting the URLs to image assets associated with image hashs or ids of structures.

PR #11411, opened 2026-01-31
feat(structures): add Presences structures 

AsadHumayun:
This PR will aim to implement the `GatewayPresence` structures.


AsadHumayun:
Note to self: PR CI failed due to missing dependent files. This should be fixed when the PR is in a finished/ready to review state.

AsadHumayun:
Note / CI for latest commit here was cancelled due to GitHub Actions not being picked up by a runner.

> The job was not acquired by Runner of type hosted even after multiple attempts

Jiralite:
Don't worry about that.

https://www.githubstatus.com/incidents/xwn6hjps36ty

AsadHumayun:
I've gone ahead and added the tests for these structures. While adding the tests, they uncovered a mistake that I had previously made with one of the structures (and corrected it) [[acad472](https://github.com/discordjs/discord.js/pull/11411/commits/acad472acc0bc6bb31b4e32f54368e6a109a056e)].

For `GatewayActivity`, some properties were not exposed as getters as they were either not documented on the Discord API docs, or were marked `@unstable` in discord-api-tyes, e.g:

https://github.com/discordjs/discord-api-types/blob/592503aa18a7b970bd6275ca59369b3b4c86b7be/payloads/v10/gateway.ts#L141-L146

I did consider adding this as a remark on the class, but these are undocumented fields so I decided not to in the end.

As we are already deciding not to expose deprecated API fields, I figured it made sense to also not expose these (IIRC discord.js as a library do not encourage use of undocumented API fields).

One additional note I'd like to highlight:

https://github.com/AsadHumayun/discord.js/blob/a22076a0d644443a0328212fc0f54c42f2a372ff/packages/structures/src/gatewayPresences/Activity/GatewayPresenceActivityAssets.ts#L70-L82

In case it doesn't embed, the `GatewayPresenceActivityAssets` structure's [documentation](https://discord.com/developers/docs/events/gateway-events#activity-object-activity-assets) shows a field `invite_cover_image`, which is not exposed by its corresponding discord-api-types structure; see

https://github.com/discordjs/discord-api-types/blob/592503aa18a7b970bd6275ca59369b3b4c86b7be/payloads/v10/gateway.ts#L353-L355 

I will also append this comment to the topmost comment on this PR so that it does not get missed. Apologies for this taking a while to get through, not been feeling great as of late.

packages/structures/src/gatewayPresences/Activity/GatewayPresenceActivityParty.ts
```diff
@@ -0,0 +1,59 @@
...
+		return isIdSet(this.id) ? DiscordSnowflake.timestampFrom(this.id) : null;
+	}
+
+	/**
+	 * The time the activity party was created at
+	 */
+	public get createdAt() {
+		const createdTimestamp = this.createdTimestamp;
+		return createdTimestamp ? new Date(createdTimestamp) : null;
+	}
```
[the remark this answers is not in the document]
AsadHumayun: Where on the Discord API does it say that? All I see for party is id: id of the party on the API.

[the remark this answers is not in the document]
AsadHumayun: I will leave this as a remark on the getter and remove the created[At/Timestamp] fields.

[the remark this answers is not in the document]
AsadHumayun: Can you send the link to the docs you're referencing please?

packages/structures/src/gatewayPresences/index.ts
```diff
@@ -0,0 +1,12 @@
...
+export * from './GatewayPresenceClientStatus.js';
+export * from './GatewayPresenceUpdate.js';
+
+/**
+ * @todo ://
+ *
+ * https://github.com/discordjs/discord.js/blob/c71228aab3b8859610be3b16042988cbacf1860e/packages/structures/src/voice/VoiceState.ts#L102-L107
+ *
+ * I think I missed requestToSpeak[At] getter for the timestamp value for that.
+ */
```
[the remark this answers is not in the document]
AsadHumayun: Addressed

packages/structures/src/presences/Presence.ts
```diff
@@ -0,0 +1,61 @@
...
+	 * Time of when the client went idle, or null if the client is not idle
+	 *
+	 * @todo [!!!!] - does this name make sense? I didn't think `sinceAt` would sound right. Is it
+	 * clear enough that this returns a Date object?
+	 */
+	public get since() {
+		const timestamp = this.sinceTimestamp;
+
+		return timestamp ? new Date(timestamp as number) : null;
+	}
```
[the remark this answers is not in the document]
AsadHumayun: Addressed

packages/structures/src/presences/Activity/ActivityTimestamps.ts
```diff
@@ -0,0 +1,57 @@
...
+	}
+
+	/**
+	 * Time of when the activity ended.
+	 */
+	public get endAt() {
+		const timestamp = this[kData].end;
+
+		return timestamp ? new Date(timestamp as string) : null;
+	}
```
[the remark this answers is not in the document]
AsadHumayun: Addressed

packages/structures/__tests__/presences.test.ts
```diff
@@ -0,0 +1,364 @@
...
+		});
+
+		describe('GatewayPresenceActivityButton sub-structure', () => {
+			const data = gatewayPresenceActivityButtonData;
+			const instance = new GatewayPresenceActivityButton(data);
+
+			test('correct value for all getters', () => {
+				expect(instance.label).toBe(data.label);
+				expect(instance.url).toBe(instance.url);
+			});
```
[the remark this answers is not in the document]
AsadHumayun: Addressed... not sure what I was thinking on that one

packages/structures/src/gatewayPresences/Activity/GatewayPresenceActivityParty.ts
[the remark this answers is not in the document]
AsadHumayun: Resolved

packages/structures/src/gatewayPresences/GatewayPresenceClientStatus.ts
```diff
@@ -0,0 +1,46 @@
...
+import { Structure } from '../Structure.js';
+import { kData } from '../utils/symbols.js';
+import type { Partialize } from '../utils/types.js';
+
+/**
+ * Represents any client status on Discord.
+ *
+ * @typeParam Omitted - Specify the properties that will not be stored in the raw data field as a union, implement via `DataTemplate`
+ */
+export class GatewayPresenceClientStatus<
```
Qjuh: While I read your reasoning for naming this I disagree on the `Gateway`prefix. We don't name the other structures with an `API` or `REST` prefix either. For the structure itself it doesn't matter where we received it from.
```suggestion
export class ClientStatus<
```

packages/structures/src/gatewayPresences/Activity/GatewayPresenceActivity.ts
```diff
@@ -0,0 +1,133 @@
...
+/**
+ * Represents any activity on Discord.
+ *
+ * @typeParam Omitted - Specify the properties that will not be stored in the raw data field as a union, implement via `DataTemplate`.
+ * @remarks has substructures `GatewayPresenceActivityTimestamps`, `GatewayPresenceActivityEmoji`, `GatewayPresenceActivityParty`,
+ * `GatewayPresenceActivityAssets`, `GatewayPresenceActivitySecrets` which need to be instantiated and stored by an extending class
+ * using it.
+ * @remarks intentionally does not export `buttons` so that extending classes can resolve `string[]` to `GatewayPresenceActivityButton[]`
+ */
+export class GatewayPresenceActivity<Omitted extends keyof GatewayActivity | '' = ''> extends Structure<
```
Qjuh:
```suggestion
export class Activity<Omitted extends keyof GatewayActivity | '' = ''> extends Structure<
```

packages/structures/src/gatewayPresences/GatewayPresenceUpdate.ts
```diff
@@ -0,0 +1,58 @@
...
+import { Structure } from '../Structure.js';
+import { kData } from '../utils/symbols.js';
+import type { Partialize } from '../utils/types.js';
+
+/**
+ * Represents any presence update on Discord.
+ *
+ * @typeParam Omitted - Specify the properties that will not be stored in the raw data field as a union, implement via `DataTemplate`
+ */
+export class GatewayPresenceUpdate<Omitted extends keyof APIGatewayPresenceUpdate | '' = ''> extends Structure<
```
Qjuh: This is a name of a payload we receive, not a structure. The structure would just be `Presence` and would (probably) represent a `GatewayGuildMembersChunkPresence` from `discord-api-types`.

Also would need a `@remark` about the substructures `Activity` and `ClientStatus`.

packages/structures/src/gatewayPresences/GatewayPresenceUpdate.ts
```diff
@@ -0,0 +1,58 @@
...
+	 * Time of when the client went idle, or null if the client is not idle
+	 */
+	public get since() {
+		const timestamp = this.sinceTimestamp;
+
+		return timestamp ? new Date(timestamp as number) : null;
+	}
+
+	/**
+	 * User's new status
```
Qjuh:
```suggestion
	 * The user's status
```

packages/structures/src/presences/Presence.ts
```diff
@@ -0,0 +1,58 @@
...
+
+	/**
+	 * User's new status
+	 *
+	 * @see {@link https://discord.com/developers/docs/events/gateway-events#update-presence-status-types}
```
Qjuh: I'd probably remove this here.

packages/structures/src/gatewayPresences/GatewayPresenceUpdate.ts
```diff
@@ -0,0 +1,58 @@
+import type { GatewayPresenceUpdateData as APIGatewayPresenceUpdate } from 'discord-api-types/v10';
```
Qjuh: This rename is just wrong and confusing. Only REST payloads are named with `API` prefix in `discord-api-types`. Just a general remark, considering this type shouldn't be used at all here.

packages/structures/src/gatewayPresences/Activity/GatewayPresenceActivityTimestamps.ts
```diff
@@ -0,0 +1,57 @@
...
+		return this[kData].start;
+	}
+
+	/**
+	 * Time of when the activity started.
+	 */
+	public get startAt() {
+		const timestamp = this[kData].start;
+
+		return timestamp ? new Date(timestamp as number) : null;
```
Qjuh: This should use the `isFieldSet(this[kData], 'start', 'number')` typeguard instead of the unsafe cast.

Use it like here, just that you shouldn't need a cast at all: https://github.com/discordjs/discord.js/blob/c460a920ac4f08ab9f3d2ae6be927b8e1de3810c/packages/structures/src/messages/Message.ts#L113-L115

packages/structures/src/gatewayPresences/Activity/GatewayPresenceActivityTimestamps.ts
```diff
@@ -0,0 +1,57 @@
...
+		return this[kData].end;
+	}
+
+	/**
+	 * Time of when the activity ended.
+	 */
+	public get endAt() {
+		const timestamp = this[kData].end;
+
+		return timestamp ? new Date(timestamp as number) : null;
```
Qjuh: Ditto

packages/structures/src/gatewayPresences/Activity/GatewayPresenceActivityTimestamps.ts
```diff
@@ -0,0 +1,57 @@
...
+import { Structure } from '../../Structure.js';
+import { kData } from '../../utils/symbols.js';
+import type { Partialize } from '../../utils/types.js';
+
+/**
+ * Represents any activity timestamp on Discord.
+ *
+ * @typeParam Omitted - Specify the properties that will not be stored in the raw data field as a union, implement via `DataTemplate`
+ */
+export class GatewayPresenceActivityTimestamps<
```
Qjuh:
```suggestion
export class ActivityTimestamps<
```

packages/structures/src/gatewayPresences/Activity/GatewayPresenceActivitySecrets.ts
```diff
@@ -0,0 +1,47 @@
...
+import { Structure } from '../../Structure.js';
+import { kData } from '../../utils/symbols.js';
+import type { Partialize } from '../../utils/types.js';
+
+/**
+ * Represents any activity secrets on Discord.
+ *
+ * @typeParam Omitted - Specify the properties that will not be stored in the raw data field as a union, implement via `DataTemplate`
+ */
+export class GatewayPresenceActivitySecrets<Omitted extends keyof GatewayActivitySecrets | '' = ''> extends Structure<
```
Qjuh:
```suggestion
export class ActivitySecrets<Omitted extends keyof GatewayActivitySecrets | '' = ''> extends Structure<
```

packages/structures/src/gatewayPresences/Activity/GatewayPresenceActivityParty.ts
```diff
@@ -0,0 +1,44 @@
...
+import { Structure } from '../../Structure.js';
+import { kData } from '../../utils/symbols.js';
+import type { Partialize } from '../../utils/types.js';
+
+/**
+ * Represents any activity party on Discord.
+ *
+ * @typeParam Omitted - Specify the properties that will not be stored in the raw data field as a union, implement via `DataTemplate`
+ */
+export class GatewayPresenceActivityParty<Omitted extends keyof GatewayActivityParty | '' = ''> extends Structure<
```
Qjuh:
```suggestion
export class ActivityParty<Omitted extends keyof GatewayActivityParty | '' = ''> extends Structure<
```

packages/structures/src/gatewayPresences/Activity/GatewayPresenceActivityEmoji.ts
```diff
@@ -0,0 +1,74 @@
...
+import { kData } from '../../utils/symbols.js';
+import { isIdSet } from '../../utils/type-guards.js';
+import type { Partialize } from '../../utils/types.js';
+
+/**
+ * Represents any activity emoji on Discord.
+ *
+ * @typeParam Omitted - Specify the properties that will not be stored in the raw data field as a union, implement via `DataTemplate`
+ */
+export class GatewayPresenceActivityEmoji<Omitted extends keyof GatewayActivityEmoji | '' = ''> extends Structure<
```
Qjuh: This class is probably not needed at all, since it can just be stored as a partial `Emoji`.

packages/structures/src/gatewayPresences/Activity/GatewayPresenceActivityButton.ts
```diff
@@ -0,0 +1,42 @@
...
+import type { Partialize } from '../../utils/types.js';
+
+/**
+ * Represents any activity button on Discord.
+ *
+ * @typeParam Omitted - Specify the properties that will not be stored in the raw data field as a union, implement via `DataTemplate`.
+ * @remarks When received over the gateway, the `buttons` field is an array of strings, which are the `button` labels. Bots cannot
+ * access a user's activity button URLs. When sending, the buttons field must be an array of this structure.
+ */
+export class GatewayPresenceActivityButton<Omitted extends keyof GatewayActivityButton | '' = ''> extends Structure<
```
Qjuh:
```suggestion
export class ActivityButton<Omitted extends keyof GatewayActivityButton | '' = ''> extends Structure<
```

packages/structures/src/gatewayPresences/Activity/GatewayPresenceActivityAssets.ts
```diff
@@ -0,0 +1,90 @@
...
+import { Structure } from '../../Structure.js';
+import { kData } from '../../utils/symbols.js';
+import type { Partialize } from '../../utils/types.js';
+
+/**
+ * Represents any activity assets on Discord.
+ *
+ * @typeParam Omitted - Specify the properties that will not be stored in the raw data field as a union, implement via `DataTemplate`
+ */
+export class GatewayPresenceActivityAssets<Omitted extends keyof GatewayActivityAssets | '' = ''> extends Structure<
```
Qjuh: Not so happy about the stored data being a typedef for a Record rather than an interface, but that's a `discord-api-types` issue. One that probably should be addressed when adding the key missing from it.

```suggestion
export class ActivityAssets<Omitted extends keyof GatewayActivityAssets | '' = ''> extends Structure<
```

packages/structures/src/bitfields/GatewayPresenceActivityFlagsBitField.ts
```diff
@@ -0,0 +1,16 @@
+import { ActivityFlags } from 'discord-api-types/v10';
+import { BitField } from './BitField.js';
+
+/**
+ * Data structure that makes it easy to interact with an {@link ActivityFlags} bitfield.
+ */
+export class GatewayPresenceActivityFlagsBitField extends BitField<keyof typeof ActivityFlags> {
```
Qjuh:
```suggestion
export class ActivityFlagsBitField extends BitField<keyof typeof ActivityFlags> {
```

packages/structures/src/presences/Presence.ts
```diff
@@ -0,0 +1,58 @@
...
+	public get status() {
+		return this[kData].status;
+	}
+
+	/**
+	 * Whether the user is AFK
+	 */
+	public get afk() {
+		return this[kData].afk;
+	}
```
Qjuh: I'm confused where this is coming from. The `discord-api-types` have it, but it's neither documented by discord nor is it implemented in current mainlib `discord.js`.

packages/structures/src/presences/Presence.ts
```diff
@@ -0,0 +1,58 @@
...
+	 */
+	public constructor(data: Partialize<APIGatewayPresenceUpdate, Omitted>) {
+		super(data);
+	}
+
+	/**
+	 * Unix timestamp (in milliseconds) of when the client went idle, or null if the client is not idle
+	 */
+	public get sinceTimestamp() {
+		return this[kData].since;
```
Qjuh: This is going to be very confusing for users. The field in the payload is named `since` as  a timestamp, but the value of that field is exposed under `sinceTimestamp` while there is also a `since` getter that has a different type. Unsure how to resolve this dilemma, maybe @ckohen has an idea?

packages/structures/src/gatewayPresences/GatewayPresenceClientStatus.ts
```diff
@@ -0,0 +1,46 @@
+import type { GatewayPresenceClientStatus as GatewayPresenceClientStatusTypedef } from 'discord-api-types/v10';
```
Qjuh: This rename is confusing (and not needed anymore with the rename of the class) because what you import here is actually an interface, not a typedef.

packages/structures/src/gatewayPresences/GatewayPresenceClientStatus.ts
```diff
@@ -0,0 +1,46 @@
...
+
+	/**
+	 * @param data - The raw data received from the API for the client status
+	 */
+	public constructor(data: Partialize<GatewayPresenceClientStatusTypedef, Omitted>) {
+		super(data);
+	}
+
+	/**
+	 * User's status set for an active desktop (Windows, Linux, Mac) application session
```
Qjuh:
```suggestion
	 * The user's status set for an active desktop (Windows, Linux, Mac) application session
```

packages/structures/src/gatewayPresences/GatewayPresenceClientStatus.ts
```diff
@@ -0,0 +1,46 @@
...
+	 */
+	public get desktop() {
+		return this[kData].desktop;
+	}
+
+	/**
+	 * User's status set for an active mobile (iOS, Android) application session
```
Qjuh:
```suggestion
	 * The user's status set for an active mobile (iOS, Android) application session
```

packages/structures/src/gatewayPresences/GatewayPresenceClientStatus.ts
```diff
@@ -0,0 +1,46 @@
...
+	 */
+	public get mobile() {
+		return this[kData].mobile;
+	}
+
+	/**
+	 * User's status set for an active web (browser, bot user) application session
```
Qjuh:
```suggestion
	 * The user's status set for an active web (browser, bot user) application session
```

packages/structures/src/presences/activities/ActivityParty.ts
```diff
@@ -0,0 +1,44 @@
...
+	}
+
+	/**
+	 * Used to show the party's current and maximum size.
+	 *
+	 * @remarks array of two integers (`current_size`, `max_size`)
+	 */
+	public get size() {
+		return this[kData].size;
+	}
```
Qjuh, line 53: Since this is actually a tuple of two values I'd probably lean towards exposing these two values as two separate getters `currentSize` and `maxSize` instead.

packages/structures/src/presences/Activity/ActivityAssets.ts
```diff
@@ -0,0 +1,90 @@
...
+
+	/**
+	 * URL that is opened when clicking on the small image
+	 */
+	public get smallURL() {
+		return this[kData].small_url;
+	}
+
+	/**
+	 * `discord-api-types` does not provide `invite_cover_image`.
```
Qjuh: Then this is an issue of `discord-api-types` and should be fixed there. The current state of this huge comment looks like a TODO, so this PR should be a draft or blocked until this is addressed.

packages/structures/src/gatewayPresences/Activity/GatewayPresenceActivity.ts
```diff
@@ -0,0 +1,133 @@
...
+		return this[kData].url;
+	}
+
+	/**
+	 * Time of when the activity was added to the user's session
+	 */
+	public get createdAt() {
+		const timestamp = this[kData].created_at;
+
+		return timestamp ? new Date(timestamp as number) : null;
```
Qjuh: Use `isFieldSet` here too.

packages/structures/src/presences/activities/Activity.ts
```diff
@@ -0,0 +1,133 @@
...
+		const timestamp = this[kData].created_at;
+
+		return timestamp ? new Date(timestamp as number) : null;
+	}
+
+	/**
+	 * Unix timestamp (in milliseconds) of when the activity was added to the user's session
+	 */
+	public get createdTimestamp() {
+		return this[kData].created_at;
```
Qjuh, line 63: This is also very confusing to users, having the field in the payload be named `created_at` and a number but the `createdAt` getter giving a Date. Not your fault, but still worrying and should probably find a better solution...

Qjuh (changes requested):

packages/structures/src/presences/Activity/ActivityAssets.ts
AsadHumayun: I'll open an issue on discord-api-types.
How can I set this PR as blocked? Not sure if I can add a label

packages/structures/src/presences/Presence.ts
AsadHumayun: I can see [presence update event fields (which seems like the type you recommended me from dapi-types)](https://discord.com/developers/docs/events/gateway-events#presence-update-presence-update-event-fields) and then the [structure (which is what I used)](https://discord.com/developers/docs/events/gateway-events#update-presence-gateway-presence-update-structure). I am a little confused! I will implement yours in a new commit.

packages/structures/src/presences/Presence.ts
AsadHumayun: I also renamed `start` and `end` on `ActivityTimestamps` to `startTimestamp` and `endTimestamp` for now. At least the way I see it, people are probably going to go and look at discord.js' docs as opposed to going straight to the Discord API, and as `xTimestamp` and `xAt` seem to be a recurring pattern in this library, I think at least to me it kind of makes sense. This was done in commit [[13404e9](https://github.com/discordjs/discord.js/pull/11411/commits/13404e9dd32c2cf7036c6611b23c3ac3ed38df07)] and thus can be easily reverted.

packages/structures/src/gatewayPresences/GatewayPresenceClientStatus.ts
AsadHumayun: Sorry about that, wasn't really thinking when I imported that. Ty for pointing out. Resolved.

packages/structures/src/gatewayPresences/Activity/GatewayPresenceActivity.ts
AsadHumayun: Resolved (very sorry if this is flooding your email)

packages/structures/src/gatewayPresences/GatewayPresenceClientStatus.ts
AsadHumayun: Resolved

packages/structures/src/gatewayPresences/GatewayPresenceUpdate.ts
AsadHumayun: Resolved

packages/structures/src/gatewayPresences/Activity/GatewayPresenceActivity.ts
AsadHumayun: These files have been renamed.

packages/structures/src/gatewayPresences/GatewayPresenceUpdate.ts
AsadHumayun: Resolved (import no longer renamed)

packages/structures/src/gatewayPresences/Activity/GatewayPresenceActivityTimestamps.ts
AsadHumayun: Resolved

packages/structures/src/gatewayPresences/Activity/GatewayPresenceActivityEmoji.ts
AsadHumayun: Removed (not sure why I didn't think of that)

packages/structures/src/gatewayPresences/GatewayPresenceUpdate.ts
AsadHumayun: I'm very confused! I imported the type that you suggested `GatewayGuildMembersChunkPresence`. The `@see` link was: https://discord.com/developers/docs/topics/gateway-events#update-presence and it took me to the same structure that I implemented here, with the same fields.

The fields of the structure that it links to do not match the fields that are actually defined on the interface. 

See:

https://github.com/discordjs/discord-api-types/blob/6df197ac4774cc386be383e5a33222c605c5f84d/payloads/v10/gateway.ts#L62-L68

packages/structures/src/gatewayPresences/Activity/GatewayPresenceActivityAssets.ts
AsadHumayun: An issue has been created: https://github.com/discordjs/discord-api-types/issues/1517

packages/structures/src/gatewayPresences/GatewayPresenceUpdate.ts
Qjuh: Because that `@see` link is indeed pointing to the wrong place. The "structure" this links to is for sending to the Gateway. Our structures are things we receive and store. So it's the update fields rather, not the send structure.

AsadHumayun: Thanks for clarifying that. Resolved.

packages/structures/src/presences/Presence.ts
AsadHumayun: Resolved

packages/structures/src/presences/activities/ActivityParty.ts
AsadHumayun: Resolved

packages/structures/src/presences/activities/Activity.ts
ckohen: Maybe we change this all now, make it explicit, createdDate, createdTimestamp (etc...), the "at" convention is already a little weird, and might be generally confusing.

AsadHumayun: I've applied this convention change. Tests have been updated accordingly and are passing. 

[refactor(structures)!: use xDate convention in favour of xAt](https://github.com/discordjs/discord.js/pull/11411/commits/5d529bf8bd9adfc01649ab1d9bb64b03f8a9750c)

Let me know how this feels. I'll let other contributors know otherwise I can revert these changes.

packages/structures/src/presences/Activity/ActivityAssets.ts
AsadHumayun: Resolved (PR merged)

AsadHumayun:
Dropping a quick update on here before I forget

https://github.com/discordjs/discord-api-types/pull/1519 was merged to dapi-types today and the subsequent missing types have been added. 

However I imagine that it will still take some time for the new version of dapi-types to get released, and then additional time as the monorepo gets updated to use the new version of dapi-types once that does get released. 

We also need a definitive decision regarding moving away from the `xAt` convention in favour of `xDate`.

The changes were applied in this commit: 

[refactor(structures)!: use xDate convention in favour of xAt](https://github.com/discordjs/discord.js/pull/11411/commits/eb1787aef20396ac42a76e90f109acca91d59185)

packages/structures/src/presences/Activity/ActivityAssets.ts
AsadHumayun: Resolved

packages/structures/__tests__/presences.test.ts
```diff
@@ -0,0 +1,330 @@
...
+	ActivityParty,
+	ActivitySecrets,
+	ActivityTimestamps,
+	ClientStatus,
+	Presence,
+} from '../src/index.js';
+import { kPatch } from '../src/utils/symbols.js';
+
+const user: APIUser = {
+	username: 'djs://username',
```
Qjuh: Most of these are not valid values for the fields they appear in. I'd prefer to have placeholder values that would also pass if we had validators for the fields.

packages/structures/__tests__/presences.test.ts
```diff
@@ -0,0 +1,330 @@
...
+		test('toJSON() returns expected values', () => {
+			expect(instance.toJSON()).toStrictEqual(data);
+		});
+
+		test('patching the structure works in-place', () => {
+			const patched = instance[kPatch]({
+				status: PresenceUpdateStatus.Online,
+			});
+
+			expect(patched.toJSON()).toStrictEqual(instance.toJSON());
```
Qjuh: This should check for those two being the same instance, not just returning the same JSON.

packages/structures/__tests__/presences.test.ts
```diff
@@ -0,0 +1,330 @@
...
+
+			expect(patched.status).not.toBe(data.status);
```
Qjuh: Should explicitly matching the new value, not just being different from the old one.

packages/structures/__tests__/presences.test.ts
```diff
@@ -0,0 +1,330 @@
...
+				expect(instance.statusDisplayType).toBe(data.status_display_type);
+				expect(instance.details).toBe(data.details);
+				expect(instance.detailsURL).toBe(data.details_url);
+				expect(instance.state).toBe(data.state);
+				expect(instance.stateURL).toBe(data.state_url);
+				expect(instance.instance).toBe(data.instance);
+				expect(instance.flags!.bitField).toBe(BigInt(data.flags!));
+				expect(instance.createdTimestamp).toBe(data.created_at);
+
+				expect(instance.createdDate).toEqual(new Date(instance.createdTimestamp));
```
Qjuh:
```suggestion
				expect(instance.createdDate.valueOf()).toEqual(data.created_at);
```

packages/structures/__tests__/presences.test.ts
```diff
@@ -0,0 +1,330 @@
...
+	small_text: 'djs://activity-asset/smallText',
+	invite_cover_image: 'djs://activity-asset/invite-cover-image',
+};
+
+const gatewayPresenceActivityData: GatewayActivity = {
+	id: '1',
+	name: 'djs://activity-name',
+	type: ActivityType.Playing,
+	url: 'djs://activity-url',
+	created_at: 1,
```
Qjuh: Like the other values this is not a valid timestamp in context of discord, this should be a valid value to check against.

packages/structures/__tests__/presences.test.ts
```diff
@@ -0,0 +1,330 @@
...
+			});
+		});
+
+		describe('GatewayPresenceActivityTimestamps sub-structure', () => {
+			const data = gatewayPresenceActivityTimestampsData;
+			const instance = new ActivityTimestamps(data);
+
+			test('correct value for all getters', () => {
+				expect(instance.startTimestamp).toBe(data.start);
+				expect(instance.startDate).toStrictEqual(new Date(data.start as number));
```
Qjuh:
```suggestion
				expect(instance.startDate.valueOf()).toBe(data.start);
```
It's never a good sign when a test needs to cast its value.

packages/structures/__tests__/presences.test.ts
```diff
@@ -0,0 +1,330 @@
...
+			test('toJSON() returns expected values', () => {
+				expect(instance.toJSON()).toStrictEqual(data);
+			});
+
+			test('patching the structure works in-place', () => {
+				const patched = instance[kPatch]({
+					end: 10_000_000,
+				});
+
+				expect(patched.endDate).toStrictEqual(new Date(patched.endTimestamp as number));
```
Qjuh:
```suggestion
				expect(patched.endDate.valueOf()).toBe(newEndTimestamp);
```

packages/structures/__tests__/presences.test.ts
```diff
@@ -0,0 +1,330 @@
...
+				expect(instance.endDate).toBeNull();
+			});
+
+			test('toJSON() returns expected values', () => {
+				expect(instance.toJSON()).toStrictEqual(data);
+			});
+
+			test('patching the structure works in-place', () => {
+				const patched = instance[kPatch]({
+					end: 10_000_000,
```
Qjuh:
```suggestion
			test('patching the structure works in-place', () => {
				const newEndTimestampe = 1771670132;
				const patched = instance[kPatch]({
					end: newEndTimestamp,
```

packages/structures/__tests__/channels.test.ts
```diff
@@ -77,7 +77,7 @@ describe('text channel', () => {
 		expect(instance.guildId).toBe(data.guild_id);
 		expect(instance.lastMessageId).toBe(data.last_message_id);
 		expect(instance.lastPinTimestamp).toBe(Date.parse(data.last_pin_timestamp!));
-		expect(instance.lastPinAt?.toISOString()).toBe(data.last_pin_timestamp);
+		expect(instance.lastPinDate?.toISOString()).toBe(data.last_pin_timestamp);
```
Qjuh:
```suggestion
		expect(dateToDiscordISOTimestamp(instance.lastPinDate!)).toBe(data.last_pin_timestamp);
```
using
```ts
import { dateToDiscordISOTimestamp } from '../src/utils/optimization.js';
```

packages/structures/__tests__/channels.test.ts
```diff
@@ -138,7 +138,7 @@ describe('announcement channel', () => {
 		expect(instance.guildId).toBe(data.guild_id);
 		expect(instance.lastMessageId).toBe(data.last_message_id);
 		expect(instance.lastPinTimestamp).toBe(null);
-		expect(instance.lastPinAt).toBe(data.last_pin_timestamp);
+		expect(instance.lastPinDate).toBe(data.last_pin_timestamp);
```
Qjuh:
```suggestion
		expect(instance.lastPinTimestamp).toBeNull();
		expect(instance.lastPinAt).toBeNull();
```

packages/structures/__tests__/channels.test.ts
```diff
@@ -234,7 +234,7 @@ describe('DM channel', () => {
 		expect(instance.flags?.toJSON()).toBe(data.flags);
 		expect(instance.lastMessageId).toBe(data.last_message_id);
 		expect(instance.lastPinTimestamp).toBe(Date.parse(data.last_pin_timestamp!));
-		expect(instance.lastPinAt?.toISOString()).toBe(data.last_pin_timestamp);
+		expect(instance.lastPinDate?.toISOString()).toBe(data.last_pin_timestamp);
```
Qjuh: Should also use dateTodiscordISOTimestamp.

packages/structures/__tests__/channels.test.ts
```diff
@@ -730,9 +730,9 @@ describe('thread channels', () => {
 		const instance = new ThreadMetadata(dataPrivate.thread_metadata!);
 		expect(instance.toJSON()).toEqual(dataPrivate.thread_metadata);
 		expect(instance.archived).toBe(dataPrivate.thread_metadata?.archived);
-		expect(instance.archivedAt?.toISOString()).toBe(dataPrivate.thread_metadata?.archive_timestamp);
+		expect(instance.archivedDate?.toISOString()).toBe(dataPrivate.thread_metadata?.archive_timestamp);
```
Qjuh: Ditto

packages/structures/__tests__/channels.test.ts
```diff
@@ -730,9 +730,9 @@ describe('thread channels', () => {
...
 		expect(instance.archivedTimestamp).toBe(Date.parse(dataPrivate.thread_metadata!.archive_timestamp));
-		expect(instance.createdAt?.toISOString()).toBe(dataPrivate.thread_metadata?.create_timestamp);
+		expect(instance.createdDate?.toISOString()).toBe(dataPrivate.thread_metadata?.create_timestamp);
```
Qjuh: Ditto

packages/structures/__tests__/invite.test.ts
```diff
@@ -34,7 +34,7 @@ describe('Invite', () => {
 		const instance = new Invite(data);
 		expect(instance.type).toBe(data.type);
 		expect(instance.code).toBe(data.code);
-		expect(instance.createdAt).toBe(null);
+		expect(instance.createdDate).toBe(null);
```
Qjuh:
```suggestion
		expect(instance.createdDate).toBeNull();
```

packages/structures/__tests__/invite.test.ts
```diff
@@ -44,7 +44,7 @@ describe('Invite', () => {
 		expect(instance.temporary).toBe(undefined);
 		expect(instance.uses).toBe(undefined);
 		expect(instance.expiresTimestamp).toBe(null);
-		expect(instance.expiresAt).toBe(null);
+		expect(instance.expiresDate).toBe(null);
```
Qjuh:
```suggestion
		expect(instance.expiresDate).toBeNull();
```

packages/structures/__tests__/invite.test.ts
```diff
@@ -65,7 +65,7 @@ describe('Invite', () => {
 		expect(instance.temporary).toBe(dataExtended.temporary);
 		expect(instance.uses).toBe(dataExtended.uses);
 		expect(instance.expiresTimestamp).toStrictEqual(Date.parse('2020-10-10T13:50:29.209000+00:00'));
-		expect(instance.expiresAt).toStrictEqual(new Date('2020-10-10T13:50:29.209000+00:00'));
+		expect(instance.expiresDate).toStrictEqual(new Date('2020-10-10T13:50:29.209000+00:00'));
```
Qjuh, line 71: adapt to how the other date tests work

packages/structures/src/presences/Activity/ActivityParty.ts
```diff
@@ -0,0 +1,54 @@
...
+		return this[kData].id;
+	}
+
+	/**
+	 * The current size of the party
+	 */
+	public get currentSize() {
+		const sizeArray = this[kData].size;
+
+		return sizeArray && isFieldSet(sizeArray, '0', 'number') ? (sizeArray[0] as number) : null;
```
Qjuh: While it is true that arrays are internally realized as objects, this is abusing the `isFieldSet` typeguard, which is also clear by the need to still cast as `number`. Instead check `Array.isArray(sizeArray)`, `sizeArray.length == 2` and `typeof sizeArray[0] === 'number'`. That check can probably be turned into its own typeguard function in `util/type-guards.js` if done right.

packages/structures/src/presences/Activity/ActivityParty.ts
```diff
@@ -0,0 +1,54 @@
...
+	}
+
+	/**
+	 * The maximum size of the party
+	 */
+	public get maximumSize() {
+		const sizeArray = this[kData].size;
+
+		return sizeArray && isFieldSet(sizeArray, '1', 'number') ? (sizeArray[1] as number) : null;
```
Qjuh: Ditto

packages/structures/src/presences/index.ts
```diff
@@ -0,0 +1,4 @@
+export * from './Activity/index.js';
```
Qjuh:
```suggestion
export * from './activities/index.js';
```
Rename the folder accordingly.

packages/structures/src/presences/Presence.ts
Qjuh: It's still there though.

packages/structures/src/index.ts
```diff
@@ -3,6 +3,7 @@ export * from './bitfields/index.js';
 export * from './channels/index.js';
 export * from './emojis/index.js';
 export * from './entitlements/index.js';
+export * from './presences/index.js';
```
Qjuh: exports are ordered alphabetically, please move this between `polls` and `skus`.

packages/structures/src/presences/Activity/index.ts
```diff
@@ -0,0 +1,6 @@
+export * from './ActivityAssets.js';
+export * from './ActivityButton.js';
+export * from './ActivityParty.js';
+export * from './ActivitySecrets.js';
+export * from './ActivityTimestamps.js';
+export * from './Activity.js';
```
Qjuh:
```suggestion
export * from './Activity.js';
export * from './ActivityAssets.js';
export * from './ActivityButton.js';
export * from './ActivityParty.js';
export * from './ActivitySecrets.js';
export * from './ActivityTimestamps.js';
```

Qjuh (changes requested):

packages/structures/__tests__/presences.test.ts
```diff
@@ -0,0 +1,353 @@
...
+	ActivityTimestamps,
+	ClientStatus,
+	Presence,
+} from '../src/index.js';
+import { kPatch } from '../src/utils/symbols.js';
+
+const user: APIUser = {
+	username: 'username',
+	avatar:
+		'https://images-ext-1.discordapp.net/external/sdfjkhsdf/%3Fsize%3D4096/https/cdn.discordapp.com/guilds/1/users/1/avatars/54a38112404a550eab14e01fb7f77c9c.png?format=webp&quality=lossless&width=1356&height=1356',
```
Qjuh: `avatar` in `APIUser` is only the hash, not the full URL.

```suggestion
		'54a38112404a550eab14e01fb7f77c9c',
```

packages/structures/__tests__/presences.test.ts
```diff
@@ -0,0 +1,353 @@
...
+};
+
+const gatewayPresenceActivitySecretsData: GatewayActivitySecrets = {
+	join: 'djs://join',
+	match: 'djs://match',
+};
+
+const gatewayPresenceActivityPartyData: GatewayActivityParty = {
+	id: '1',
+	size: [11, null as unknown as number],
```
Qjuh:
```suggestion
	size: [11, 40],
```
Forcing an invalid value by casting is the opposite of what tests should do

packages/structures/__tests__/presences.test.ts
```diff
@@ -0,0 +1,353 @@
...
+	animated: false,
+};
+
+const gatewayPresenceActivityButtonData: GatewayActivityButton = {
+	label: 'label',
+	url: 'https://github.com//discordjs/discord.js',
+};
+
+const gatewayPresenceActivityAssetsData: GatewayActivityAssets = {
+	large_image: 'djs://some-real-big-image-init',
```
Qjuh:
```suggestion
	large_image: '123456789012345678',
```

packages/structures/__tests__/presences.test.ts
```diff
@@ -0,0 +1,353 @@
...
+	large_text: 'djs://large-text',
+	large_url: 'activity-asset/large-url/',
```
Qjuh: Should be a valid URL, for example `https://discord.js.org`

packages/structures/__tests__/presences.test.ts
```diff
@@ -0,0 +1,353 @@
...
+	small_image: 'djs://small-image',
```
Qjuh:
```suggestion
	small_image: '123456789012345679',
```

packages/structures/__tests__/presences.test.ts
```diff
@@ -0,0 +1,353 @@
...
+	small_text: 'djs://activity-asset/smallText',
+	invite_cover_image: 'djs://activity-asset/invite-cover-image',
```
Qjuh:
```suggestion
	invite_cover_image: '123456789012345670',
```

packages/structures/__tests__/presences.test.ts
```diff
@@ -0,0 +1,353 @@
...
+};
+
+const gatewayPresenceActivityData: GatewayActivity = {
+	id: '1',
+	name: 'activity-name',
+	type: ActivityType.Playing,
+	url: 'https://github.com//discordjs/discord.js',
```
Qjuh:
```suggestion
	url: 'https://github.com/discordjs/discord.js',
```

packages/structures/__tests__/presences.test.ts
```diff
@@ -0,0 +1,353 @@
...
+	created_at: 1_540_381_143_572,
+	timestamps: gatewayPresenceActivityTimestampsData,
+	application_id: '121212',
+	status_display_type: StatusDisplayType.Details,
+	details: 'activity-details',
+	details_url: 'activity-details-url',
```
Qjuh: Not a valid URL, same for state_url

packages/structures/__tests__/presences.test.ts
```diff
@@ -0,0 +1,353 @@
...
+			});
+
+			test('toJSON() returns expected values', () => {
+				expect(instance.toJSON()).toStrictEqual(data);
+			});
+
+			test('patching the structure works in-place', () => {
+				const patched = instance[kPatch]({
+					large_text: 'djs://PATCHED-LARGE-TEXT',
+					small_url: 'djs://small-url',
```
Qjuh: Needs to be a valid URL

packages/structures/src/presences/activities/ActivityAssets.ts
```diff
@@ -0,0 +1,75 @@
...
+	 * @param data - The raw data received from the API for the activity assets
+	 */
+	public constructor(data: Partialize<GatewayActivityAssets, Omitted>) {
+		super(data);
+	}
+
+	/**
+	 * @see {@link https://discord.com/developers/docs/events/gateway-events#activity-object-activity-asset-image}
+	 */
+	public get largeImage() {
```
Qjuh, line 37: This should receive URL getters, similar to those done in #11406

packages/structures/src/presences/activities/ActivityParty.ts
```diff
@@ -0,0 +1,54 @@
...
+	 */
+	public get id() {
+		return this[kData].id;
+	}
+
+	/**
+	 * The current size of the party
+	 */
+	public get currentSize() {
+		const array = this[kData].size as number[];
```
Qjuh: The cast here should not be done, instead the typeguard should assert this

packages/structures/src/presences/activities/ActivityParty.ts
```diff
@@ -0,0 +1,54 @@
...
+
+		return array ? (isArrayFieldSet(array, 0, 'number') ? array[0] : null) : null;
+	}
+
+	/**
+	 * The maximum size of the party
+	 */
+	public get maximumSize() {
+		const array = this[kData].size as number[];
```
Qjuh: Ditto

packages/structures/src/utils/type-guards.ts
```diff
@@ -30,3 +30,21 @@ export function isFieldSet<Value extends object, Key extends string, Type extend
...
+/**
+ * This typeguard will determine whether a specified array field is the specified
+ * type.
+ *
+ * @internal
+ * @param array - The target array from which the target index will be narrowed
+ * @param targetIndex - The target index of the element whose type to narrow
+ * @param type - The type to compare against
+ */
+export function isArrayFieldSet(array: unknown[], targetIndex: number, type: TypeofType): boolean {
```
Qjuh: This isn't a typeguard yet. Writing typeguards is hard though, I'm trying to draft something up.

Qjuh (changes requested):

packages/structures/src/utils/type-guards.ts
Qjuh:
```suggestion
export function isArrayFieldSet<Type extends TypeofType, Index extends number>(array: unknown, targetIndex: Index, type: Type): array is unknown[] & Record<Index, Type> {
```

Qjuh (changes requested):

packages/structures/src/utils/type-guards.ts
AsadHumayun: I've never actually written one before, so this is definitely a great learning moment for me. Thanks for the help.

packages/structures/__tests__/presences.test.ts
```diff
@@ -0,0 +1,355 @@
...
+				expect(instance.toJSON()).toStrictEqual(data);
+			});
+
+			test('patching the structure works in-place', () => {
+				const newTimestamp = 1_771_670_132;
+				const patched = instance[kPatch]({
+					end: newTimestamp,
+				});
+
+				expect(patched.endDate?.valueOf()).toStrictEqual(patched.endTimestamp);
```
Qjuh:
```suggestion
				expect(patched.endDate?.valueOf()).toStrictEqual(newTimestamp);
```

Qjuh (changes requested):

packages/structures/__tests__/presences.test.ts
AsadHumayun: Ty for catching, resolved in [a765576](https://github.com/discordjs/discord.js/pull/11411/commits/a765576053c1f63c559fd520079b37ab273b2ccb).

packages/structures/src/subscriptions/Subscription.ts
```diff
@@ -125,7 +125,7 @@ export class Subscription<
 	/**
 	 * The time at which the current subscription period will end
 	 */
-	public get currentPeriodEndsAt() {
+	public get currentPeriodendsDate() {
```
almeidx:
```suggestion
	public get currentPeriodEndDate() {
```

packages/structures/src/subscriptions/Subscription.ts
```diff
@@ -110,7 +110,7 @@ export class Subscription<
 	/**
 	 * The time at which the current subscription period will start
 	 */
-	public get currentPeriodStartAt() {
+	public get currentPeriodstartDate() {
```
almeidx:
```suggestion
	public get currentPeriodStartDate() {
```

packages/structures/src/bitfields/ActivityFlagsBitField.ts
almeidx, line 1: Missing export for this file in `packages/structures/src/bitfields/index.ts`

packages/structures/src/messages/Message.ts
```diff
@@ -169,9 +169,8 @@ export class Message<Omitted extends keyof APIMessage | '' = 'edited_timestamp'
 			clone.edited_timestamp = dateToDiscordISOTimestamp(new Date(this[kEditedTimestamp]));
 		}
 
-		const createdAt = this.createdAt;
-		if (createdAt) {
-			clone.timestamp = dateToDiscordISOTimestamp(createdAt);
+		if (this.createdDate) {
+			clone.timestamp = dateToDiscordISOTimestamp(this.createdDate);
 		}
```
almeidx, line 178: Why was the variable removed? Now this calls the getter twice, creating 2 `Date` instances in the happy path.

packages/structures/src/presences/activities/ActivityAssets.ts
```diff
@@ -0,0 +1,106 @@
...
+	}
+
+	/**
+	 * Returns the large image URL of this `largeImage`
+	 *
+	 * @param applicationId - The id of the application to whom this asset belongs to
+	 * @param imageFormat - The desired format of the returned image URL
+	 */
+	public largeImageURL(applicationId: string, imageFormat: ApplicationAssetFormat = ImageFormat.PNG) {
+		return isFieldSet(this, 'largeImage', 'string')
```
almeidx: Any reason to use this helper here instead of just `this[kData].large_image`?

packages/structures/src/utils/type-guards.ts
```diff
@@ -30,3 +30,25 @@ export function isFieldSet<Value extends object, Key extends string, Type extend
...
+ * @param targetIndex - The target index of the element whose type to narrow
+ * @param type - The type to compare against
+ */
+export function isArrayFieldSet<Type extends TypeofType, Index extends number>(
+	array: unknown,
+	targetIndex: Index,
+	type: Type,
+): array is Record<Index, Type> & unknown[] {
+	return Array.isArray(array)
+		? array.length >= targetIndex
```
almeidx: This is wrong.
```suggestion
		? array.length > targetIndex
```

Although in practice it wouldn't make matter too much, since it checks the `typeof` anyways, and it's unlikely we'd actually have an array of holes/`undefined`s

almeidx (dismissed):
This pull request has a lot of "out of scope" changes that makes it harder to review and to reason about. I suppose it's fine since it's already done anyways but in the future I would personally prefer if a separate pull request was made for the renames and stuff.

packages/structures/src/presences/activities/ActivityAssets.ts
AsadHumayun: It errors without it.

```ts
// Argument of type 'NonNullable<"large_image" extends Omitted ? unknown : string | undefined>' is not assignable to parameter of type 'string'.ts(2345)

return this[kData].large_image
			? `${RouteBases.cdn}${CDNRoutes.applicationAsset(applicationId, this[kData].large_image, imageFormat)}`
			: null;

```
```
@discordjs/structures:build: src/presences/activities/ActivityAssets.ts:49:68 - error TS2345: Argument of type '("large_image" extends Omitted ? unknown : string | undefined) | undefined' is not assignable to parameter of type 'string'.
@discordjs/structures:build:   Type 'undefined' is not assignable to type 'string'.
```

It produces an error as well if I extract `this[kData].large_image` as a const as well. Qjuh actually also used this helper in #11406 when they were adding URL getters of a similar nature.

AsadHumayun: Everything else resolved

almeidx: let's use the internal data key `this[kData].large_image` instead of the getter then.
both for the `isFieldSet` call (`isFieldSet(this[kData], 'large_image', 'string')`) and in the cdn route call below too

AsadHumayun: Done :)
image

Edit - reading it back the above screenshot is wrong, I have updated the tests to catch this in the future & have corrected it to use the proper key.
image

packages/structures/src/bitfields/ActivityFlagsBitField.ts
```diff
@@ -0,0 +1,16 @@
+import { ActivityFlags } from 'discord-api-types/v10';
+import { BitField } from './BitField.js';
+
+/**
+ * Data structure that makes it easy to interact with an {@link ActivityFlags} bitfield.
```
kshitijanurag, line 5:
```suggestion
 * Data structure that makes it easy to interact with an {@link Activity#flags} bitfield.
```

packages/structures/__tests__/presences.test.ts
```diff
@@ -0,0 +1,365 @@
...
+			const patched = instance[kPatch]({
+				web: PresenceUpdateStatus.DoNotDisturb,
+				mobile: PresenceUpdateStatus.Idle,
+			});
+
+			expect(patched.toJSON()).toStrictEqual(instance.toJSON());
+			expect(patched.toJSON()).not.toEqual(data);
+
+			expect(instance.web).toEqual(patched.web);
+			expect(instance.mobile).toEqual(instance.mobile);
```
Qjuh: Since `patched` and `instance` are the same object here you're comparison doesn't test much. Its methods and properties will always equal each other. Instead you should compare the properties to the actual expected hardcoded values, which does an actual test, not just that it differs but that the new value is the expected one.

packages/structures/__tests__/presences.test.ts
```diff
@@ -0,0 +1,365 @@
...
+					large_text: 'djs://PATCHED-LARGE-TEXT',
+					small_url: 'https://discord.js.org/docs/packages/structures/main',
+				});
+
+				expect(patched.largeText).toEqual('djs://PATCHED-LARGE-TEXT');
+				expect(patched.smallURL).toEqual('https://discord.js.org/docs/packages/structures/main');
+
+				expect(patched).toBe(instance);
+				expect(patched.toJSON()).not.toEqual(data);
+				expect(patched.toJSON()).toStrictEqual(instance.toJSON());
```
Qjuh: Ditto

packages/structures/__tests__/presences.test.ts
```diff
@@ -0,0 +1,365 @@
...
+				});
+
+				expect(patched.statusDisplayType).toEqual(StatusDisplayType.Name);
+				expect(patched.state).toEqual('[PATCHED]-activity-state');
+				expect(patched.stateURL).toBeNull();
+				expect(patched.type).toEqual(ActivityType.Custom);
+
+				expect(patched).toBe(instance);
+				expect(patched.toJSON()).not.toEqual(data);
+				expect(patched.toJSON()).toStrictEqual(instance.toJSON());
```
Qjuh: redundant, please remove

packages/structures/__tests__/presences.test.ts
```diff
@@ -0,0 +1,365 @@
...
+			expect(instance.toJSON()).toStrictEqual(data);
+		});
+
+		test('patching the structure works in-place', () => {
+			const patched = instance[kPatch]({
+				status: PresenceUpdateStatus.Online,
+			});
+
+			expect(patched).toBe(instance);
+			expect(patched.toJSON()).toStrictEqual(instance.toJSON());
```
Qjuh: redundant, please remove

packages/structures/__tests__/presences.test.ts
```diff
@@ -0,0 +1,365 @@
...
+			test('patching the structure works in-place', () => {
+				const patched = instance[kPatch]({
+					label: '[PATCHED]-button-label',
+				});
+
+				expect(patched.label).toEqual('[PATCHED]-button-label');
+
+				expect(patched).toEqual(instance);
+				expect(patched.toJSON()).not.toEqual(data);
+				expect(patched.toJSON()).toStrictEqual(instance.toJSON());
```
Qjuh: Ditto

packages/structures/__tests__/presences.test.ts
```diff
@@ -0,0 +1,365 @@
...
+				const patched = instance[kPatch]({
+					size: [1, 999],
+				});
+
+				expect(instance.maximumSize).toBe(999);
+				expect(instance.currentSize).toBe(1);
+
+				expect(patched).toBe(instance);
+				expect(patched.toJSON()).not.toEqual(data);
+				expect(patched.toJSON()).toStrictEqual(instance.toJSON());
```
Qjuh: Ditto

packages/structures/__tests__/presences.test.ts
```diff
@@ -0,0 +1,365 @@
...
+					match,
+					spectate,
+				});
+
+				expect(patched.match).toEqual(match);
+				expect(patched.spectate).toEqual(spectate);
+
+				expect(patched).toBe(instance);
+				expect(patched.toJSON()).not.toEqual(data);
+				expect(patched.toJSON()).toStrictEqual(instance.toJSON());
```
Qjuh: Ditto

packages/structures/__tests__/presences.test.ts
```diff
@@ -0,0 +1,365 @@
...
+				const newTimestamp = 1_771_670_132;
+				const patched = instance[kPatch]({
+					end: newTimestamp,
+				});
+
+				expect(patched.endDate?.valueOf()).toStrictEqual(newTimestamp);
+
+				expect(patched).toEqual(instance);
+				expect(patched.toJSON()).not.toEqual(data);
+				expect(patched.toJSON()).toStrictEqual(instance.toJSON());
```
Qjuh: Ditto

Qjuh (changes requested):

packages/structures/__tests__/presences.test.ts
AsadHumayun: Hi, thanks for explaining. I've gone ahead and made the requested changes, also going to look back at other tests I've written to try and pre-emptively catch & resolve this.

packages/structures/src/presences/activities/Activity.ts
```diff
@@ -0,0 +1,127 @@
...
+import { isFieldSet } from '../../utils/type-guards.js';
+import type { Partialize } from '../../utils/types.js';
+
+/**
+ * Represents any activity on Discord.
+ *
+ * @typeParam Omitted - Specify the properties that will not be stored in the raw data field as a union, implement via `DataTemplate`.
+ * @remarks has substructures `ActivityTimestamps`, `ActivityParty`,`ActivityAssets`, `ActivitySecrets` which need to be instantiated
+ * and stored by an extending class using it.
+ * @remarks intentionally does not export `buttons` so that extending classes can resolve `string[]` to `ActivityButton[]`
```
ckohen, line 14: Is there a reason for this being a separate remark?

ckohen (changes requested):

packages/structures/src/presences/activities/Activity.ts
AsadHumayun: Honestly I think that's just how it's always been done?
Eg
https://github.com/discordjs/discord.js/blob/3d6121589f9c0d91f7cf4976307e8be07053a277/packages/structures/src/automoderation/AutoModerationRule.ts#L9-L11

See the conversation at https://github.com/discordjs/discord.js/pull/11369#discussion_r2666472947 for more info on this. 

The deep linking doesn't seem to be working for me - here is a screenshot:
image

and ever since that AFAIK it's just lived as an additional remark on the structure's class:

https://github.com/discordjs/discord.js/blob/3d6121589f9c0d91f7cf4976307e8be07053a277/packages/structures/src/emojis/Emoji.ts#L11-L13

Open to suggestions, changes would need to be made in multiple places (including yet-to-be-merged/reviewed structures PRs).

ckohen (approved):

Jiralite (dismissed):
For all https://discord.com links, please use Discord's new documentation site!

AsadHumayun:
> For all https://discord.com links, please use Discord's new documentation site!

@Jiralite Do these deep links work for you?

https://docs.discord.com/developers/events/gateway-events#activity-object-activity-types

It should deep-link and scroll down to the Activity Types in the Activity sections, however it does not. This link was directly copied from the "Activity Types" button on the Discord docs:

image


What I think's happening is that because of the docs site changes, the deep links straight from URLs aren't working or being seen in some way by the docs sites. Since the deep linking doesn't work, what do you wanna do here? I can keep the links in but the deep linking won't work so it could potentially leave consumers of the library a little confused about the jsdoc comments.

Jiralite:
Yes, it works for me, whether it's a new tab or I'm on the page already.

AsadHumayun:
> Yes, it works for me, whether it's a new tab or I'm on the page already.

Thanks, turns out it takes a second for it to scroll. 
I have updated to the new links and tested them all.

packages/structures/__tests__/presences.test.ts
```diff
@@ -0,0 +1,363 @@
...
+	Presence,
+} from '../src/index.js';
+import { kPatch } from '../src/utils/symbols.js';
+
+const user: APIUser = {
+	username: 'username',
+	avatar: '54a38112404a550eab14e01fb7f77c9c',
+	global_name: 'User',
+	discriminator: '0000',
+	id: '3',
```
Qjuh: please use realistic snowflakes for id (e.g. any random 19 digit number)

packages/structures/__tests__/presences.test.ts
```diff
@@ -0,0 +1,363 @@
...
+	large_image: '123456789012345678',
+	large_text: 'large-text',
+	large_url: 'https://discord.js.org',
+	small_image: '123456789012345678',
+	small_text: 'activity-asset/smallText',
+	invite_cover_image: '123456789012345670',
+};
+
+const gatewayPresenceActivityData: GatewayActivity = {
+	id: '1',
```
Qjuh: ditto

packages/structures/__tests__/presences.test.ts
```diff
@@ -0,0 +1,363 @@
...
+	name: 'activity-name',
+	type: ActivityType.Playing,
+	url: 'https://github.com/discordjs/discord.js',
+	created_at: 1_540_381_143_572,
+	timestamps: gatewayPresenceActivityTimestampsData,
+	application_id: '121212',
```
Qjuh: ditto

packages/structures/__tests__/presences.test.ts
```diff
@@ -0,0 +1,363 @@
...
+
+			const patched = instance[kPatch]({
+				web,
+				mobile,
+			});
+
+			expect(instance.web).toEqual(web);
+			expect(instance.mobile).toEqual(mobile);
+
+			expect(patched.toJSON()).not.toEqual(data);
```
Qjuh: check for the expected patched result, not just that it changed at all

packages/structures/__tests__/presences.test.ts
```diff
@@ -0,0 +1,363 @@
...
+		});
+
+		test('patching the structure works in-place', () => {
+			const patched = instance[kPatch]({
+				status: PresenceUpdateStatus.Online,
+			});
+
+			expect(patched.status).toEqual(PresenceUpdateStatus.Online);
+
+			expect(patched.toJSON()).not.toEqual(data);
```
Qjuh: ditto

packages/structures/__tests__/presences.test.ts
```diff
@@ -0,0 +1,363 @@
...
+					state: '[PATCHED]-activity-state',
+					state_url: null,
+					type: ActivityType.Custom,
+				});
+
+				expect(patched.statusDisplayType).toEqual(StatusDisplayType.Name);
+				expect(patched.state).toEqual('[PATCHED]-activity-state');
+				expect(patched.stateURL).toBeNull();
+				expect(patched.type).toEqual(ActivityType.Custom);
+				expect(patched.toJSON()).not.toEqual(data);
```
Qjuh: ditto

packages/structures/__tests__/presences.test.ts
```diff
@@ -0,0 +1,363 @@
...
+				const patched = instance[kPatch]({
+					large_text,
+					small_url,
+				});
+
+				expect(patched.largeText).toEqual(large_text);
+				expect(patched.smallURL).toEqual(small_url);
+
+				expect(patched).toBe(instance);
+				expect(patched.toJSON()).not.toEqual(data);
```
Qjuh: ditto

packages/structures/__tests__/presences.test.ts
```diff
@@ -0,0 +1,363 @@
...
+
+			test('patching the structure works in-place', () => {
+				const patched = instance[kPatch]({
+					label: '[PATCHED]-button-label',
+				});
+
+				expect(patched.label).toEqual('[PATCHED]-button-label');
+
+				expect(patched).toEqual(instance);
+				expect(patched.toJSON()).not.toEqual(data);
```
Qjuh: ditto

packages/structures/__tests__/presences.test.ts
```diff
@@ -0,0 +1,363 @@
...
+			test('patching the structure works in-place', () => {
+				const patched = instance[kPatch]({
+					size: [1, 999],
+				});
+
+				expect(instance.maximumSize).toBe(999);
+				expect(instance.currentSize).toBe(1);
+
+				expect(patched).toBe(instance);
+				expect(patched.toJSON()).not.toEqual(data);
```
Qjuh: ditto

packages/structures/__tests__/presences.test.ts
```diff
@@ -0,0 +1,363 @@
...
+				const patched = instance[kPatch]({
+					match,
+					spectate,
+				});
+
+				expect(patched.match).toEqual(match);
+				expect(patched.spectate).toEqual(spectate);
+
+				expect(patched).toBe(instance);
+				expect(patched.toJSON()).not.toEqual(data);
```
Qjuh: ditto

packages/structures/__tests__/presences.test.ts
```diff
@@ -0,0 +1,363 @@
...
+			test('patching the structure works in-place', () => {
+				const newTimestamp = 1_771_670_132;
+				const patched = instance[kPatch]({
+					end: newTimestamp,
+				});
+
+				expect(patched.endDate?.valueOf()).toStrictEqual(newTimestamp);
+
+				expect(patched).toEqual(instance);
+				expect(patched.toJSON()).not.toEqual(data);
```
Qjuh: ditto

Qjuh (changes requested):
Apart from these minor test changes LGTM

AsadHumayun:
@Qjuh I have done all of the requested changes in your review - I also went and added a beforeEach on each of the tests so that they used a fresh instance each time, this was to keep it consistent with feedback from tests in other PRs.

Qjuh (approved):

SpaceEEC (approved):

PR #11395, opened 2026-01-23
feat: modal radio group and checkbox components

imnaiyar:
Depends on https://github.com/discordjs/discord-api-types/pull/1495/

PR #10876, opened 2025-05-02
refactor: using eslint-config-neon on mainlib

Qjuh:
**Please describe the changes this PR makes and why it should be merged:**

Removes the old-style `.eslintrc.json` and the custom configuration for mainlib `discord.js` to instead use the global `eslint.config.js` using `eslint-config-neon` style with only few of our customizations.

~~This is semver minor because it moved the anonymous `findName` function to the named `findName()` in `Util.js` and we export all methods from that file.~~ The export of all methods of `Util.js` has been changed to only export a handpicked selection now, so no longer semver minor.

**Status and versioning classification:**

- Code changes have been tested against the Discord API, or there are no code changes
- I know how to update typings and have done so, or typings don't need updating
- This PR changes the library's interface (methods or parameters added)

PR #11420, opened 2026-02-13
build: Upgrade dependencies and pnpm

Jiralite:
- Upgrades dependencies
- Upgrades pnpm to 10.29.3
- Fixes an emerging voice test issue
  - If the fix isn't desirable, can revert what introduced it
- Fixes issues reported by ESLint in packages/api-extractor
- Regenerated pnpm-lock.yaml
- Unblocks #11395
- Unblocks #11411
- Unblocks builders for https://github.com/discord/discord-api-docs/pull/8123

### DIFF
--- a/packages/discord.js/typings/index.d.ts
+++ b/packages/discord.js/typings/index.d.ts
@@ -221,40 +221,26 @@ import {
 export class Activity {
   private constructor(presence: Presence, data?: GatewayActivity);
   public readonly presence: Presence;
-
   public applicationId: Snowflake | null;
-
   public assets: RichPresenceAssets | null;
-
   public buttons: string[];
-
   public get createdAt(): Date;
   public createdTimestamp: number;
-
   public details: string | null;
-
   public emoji: Emoji | null;
-
   public flags: Readonly<ActivityFlagsBitField>;
-
   public name: string;
-
   public party: {
     id: string | null;
     size: [number, number];
   } | null;
-
   public state: string | null;
-
   public syncId: string | null;
-
   public timestamps: {
     end: Date | null;
     start: Date | null;
   } | null;
-
   public type: ActivityType;
-
   public url: string | null;
   public equals(activity: Activity): boolean;
   public toString(): string;
@@ -325,17 +311,11 @@ export class ActivityFlagsBitField extends BitField<ActivityFlagsString> {
 export abstract class AnonymousGuild extends BaseGuild {
   protected constructor(client: Client<true>, data: unknown, immediatePatch?: boolean);
   public banner: string | null;
-
   public description: string | null;
-
   public nsfwLevel: GuildNSFWLevel;
-
   public premiumSubscriptionCount: number | null;
-
   public splash: string | null;
-
   public vanityURLCode: string | null;
-
   public verificationLevel: GuildVerificationLevel;
   public bannerURL(options?: ImageURLOptions): string | null;
   public splashURL(options?: ImageURLOptions): string | null;
@@ -344,28 +324,18 @@ export abstract class AnonymousGuild extends BaseGuild {
 export class AutoModerationActionExecution {
   private constructor(data: GatewayAutoModerationActionExecutionDispatchData, guild: Guild);
   public guild: Guild;
-
   public action: AutoModerationAction;
-
   public ruleId: Snowflake;
-
   public ruleTriggerType: AutoModerationRuleTriggerType;
-
   public get user(): User | null;
   public userId: Snowflake;
-
   public get channel(): ForumChannel | GuildTextBasedChannel | MediaChannel | null;
   public channelId: Snowflake | null;
-
   public get member(): GuildMember | null;
   public messageId: Snowflake | null;
-
   public alertSystemMessageId: Snowflake | null;
-
   public content: string;
-
   public matchedKeyword: string | null;
-
   public matchedContent: string | null;
   public get autoModerationRule(): AutoModerationRule | null;
 }
@@ -373,25 +343,15 @@ export class AutoModerationActionExecution {
 export class AutoModerationRule extends Base {
   private constructor(client: Client<true>, data: APIAutoModerationRule, guild: Guild);
   public id: Snowflake;
-
   public guild: Guild;
-
   public name: string;
-
   public creatorId: Snowflake;
-
   public eventType: AutoModerationRuleEventType;
-
   public triggerType: AutoModerationRuleTriggerType;
-
   public triggerMetadata: AutoModerationTriggerMetadata;
-
   public actions: AutoModerationAction[];
-
   public enabled: boolean;
-
   public exemptRoles: Collection<Snowflake, Role>;
-
   public exemptChannels: Collection<Snowflake, GuildBasedChannel>;
   public edit(options: AutoModerationRuleEditOptions): Promise<AutoModerationRule>;
   public delete(reason?: string): Promise<void>;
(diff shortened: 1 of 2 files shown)


packages/builders/src/components/label/Label.ts
```diff
@@ -70,6 +70,7 @@ export class LabelBuilder extends ComponentBuilder<APILabelComponent> {
 
 		this.data = {
 			...structuredClone(rest),
+			// @ts-expect-error Upcoming components update.
```
vladfrangu, line 73: Lol

packages/api-extractor-model/package.json
```diff
@@ -33,19 +33,19 @@
 	"dependencies": {
 		"@microsoft/tsdoc": "~0.15.1",
 		"@microsoft/tsdoc-config": "~0.17.1",
-		"@rushstack/node-core-library": "5.13.1"
+		"@rushstack/node-core-library": "5.19.1"
```
Qjuh: Please don't do version bumps on these pinned versions. We follow upstream here and do the bumps when we merge upstream changes bumping dependency versions.

packages/api-extractor/package.json
```diff
@@ -53,29 +53,29 @@
 		"@discordjs/api-extractor-model": "workspace:^",
 		"@microsoft/tsdoc": "~0.15.1",
 		"@microsoft/tsdoc-config": "~0.17.1",
-		"@rushstack/node-core-library": "5.13.1",
-		"@rushstack/ts-command-line": "5.0.1",
+		"@rushstack/node-core-library": "5.19.1",
```
Qjuh: Ditto

packages/api-extractor/package.json
```diff
@@ -53,29 +53,29 @@
...
+		"@rushstack/ts-command-line": "5.2.0",
```
Qjuh: Ditto

api-extractor.json
```diff
@@ -107,7 +107,7 @@
 				"allowArbitraryExtensions": false,
 				"allowImportingTsExtensions": false,
 				"module": "ESNext",
-				"moduleResolution": "nodenext",
+				"moduleResolution": "Bundler",
```
Qjuh, line 110: Why? Apart from it not being lowercase there is no reason to change this for ae runs.

Qjuh (changes requested):

api-extractor.json
Jiralite: Caused packages to fail with errors like this: https://github.com/discordjs/discord.js/actions/runs/21985135665/job/63517319955?pr=11420

packages/api-extractor/package.json
```diff
@@ -56,26 +56,26 @@
...
 		"@rushstack/ts-command-line": "5.0.1",
 		"colors": "~1.4.0",
-		"lodash": "~4.17.21",
+		"lodash": "~4.17.23",
 		"resolve": "~1.22.11",
-		"semver": "~7.6.3",
+		"semver": "~7.7.4",
 		"source-map": "~0.6.1",
-		"typescript": "~5.5.4"
+		"typescript": "~5.9.3"
```
almeidx: I avoided upgrading these dependencies too (not just the `@rusthstack/*` ones) on my bump prs. Especially `typescript`. Not sure if this is wanted
cc @Qjuh

Qjuh: You're right, dependencies should follow upstream. Only devDeps should be bumped here.

packages/api-extractor/src/analyzer/AstSymbol.ts
```diff
@@ -52,6 +52,7 @@ export interface IAstSymbolOptions {
  * ```
  */
 export class AstSymbol extends AstEntity {
+	// eslint-disable-next-line jsdoc/valid-types
```
almeidx: Maybe just disable this rule completely on 
https://github.com/discordjs/discord.js/blob/ccce987fa1ada0b3791127fad0c0ebb43ac6c524/eslint.config.js#L125-L132

Maybe `@typescript-eslint/no-duplicate-type-constituents` too

packages/discord.js/typings/index.d.ts
```diff
@@ -221,26 +221,40 @@ import {
 export class Activity {
   private constructor(presence: Presence, data?: GatewayActivity);
   public readonly presence: Presence;
+
```
almeidx: Any chance we can prevent these changes?

packages/scripts/turbo/generators/templates/default/package.json.hbs
```diff
@@ -59,19 +59,19 @@
 	"devDependencies": {
 		"@discordjs/api-extractor": "workspace:^",
 		"@favware/cliff-jumper": "^6.0.0",
-		"@types/node": "^22.19.2",
-		"@vitest/coverage-v8": "^4.0.15",
+		"@types/node": "^24.10.13",
+		"@vitest/coverage-v8": "^4.0.18",
 		"cross-env": "^10.1.0",
 		"esbuild-plugin-version-injector": "^1.2.1",
 		"eslint": "^9.39.1",
```
almeidx:
```suggestion
		"eslint": "^9.39.2",
```

packages/discord.js/package.json
```diff
@@ -74,27 +74,27 @@
...
     "@vladfrangu/async_event_emitter": "^2.4.7",
-    "discord-api-types": "^0.38.36",
+    "discord-api-types": "^0.38.40",
     "fast-deep-equal": "3.1.3",
     "lodash.snakecase": "4.1.1",
-    "magic-bytes.js": "^1.12.1",
+    "magic-bytes.js": "^1.13.0",
     "tslib": "^2.8.1",
-    "undici": "7.16.0"
+    "undici": "7.22.0"
```
Qjuh, line 82: Did you test this? There's a reason why we pin undici version and I'd hate to find out they messed up once again only after we release...

packages/discord.js/typings/index.d.ts
Qjuh: We already did once, I wonder why they are here at all.
https://github.com/discordjs/discord.js/pull/10876#discussion_r2072477146
https://github.com/discordjs/discord.js/blob/ccce987fa1ada0b3791127fad0c0ebb43ac6c524/eslint.config.js#L223 should prevent this, unless that rule was renamed

packages/builders/src/components/Components.ts
```diff
@@ -237,7 +238,7 @@ export function createComponentBuilder(
 		case ComponentType.FileUpload:
 			return new FileUploadBuilder(data);
 		default:
-			// @ts-expect-error This case can still occur if we get a newer unsupported component type
+			// This directive should be here: @ts-expect-error This case can still occur if we get a newer unsupported component type
```
Qjuh:
```suggestion
			// TODO: add back @ts-expect-error This case can still occur if we get a newer unsupported component type
```
This will at least make eslint warn about it still being there, in case it gets forgotten.

packages/api-extractor/src/analyzer/AstSymbol.ts
Qjuh: Disable the latter, fix the former. Or let me fix them. Because they indeed *are* invalid. And the irony here is that AE knows it is: https://github.com/discordjs/discord.js/blob/4876bb310096a6140eb3acec5e3da79f4c4af534/packages/api-extractor/src/api/ExtractorMessageId.ts#L101

PR #11432, opened 2026-02-27
feat!: make voice receivers return AudioPacket classes instead of just Audio

petergeneric:
Currently, RTP packet headers are stripped and clients are only delivered Opus frames as `Buffer`s. The lack of timestamp makes jitter/drift hard to combat for clients, who must compute wall-clock delivery time.

This change introduces an `AudioPacket` interface which extends `Buffer` with the following read-only fields:
- `sequence` (16-bit uint monotonically increasing counter to allow for identifying out-of-order RTP packets)
- `timestamp` (32-bit uint that counts encoder-side timestamps; [RFC 7587, Opus in RTP](https://datatracker.ietf.org/doc/html/rfc7587) requires this be expressed at 48kHz no matter the audio sample rate)
- `ssrc` (to allow consumers to detect a change and reset their Opus decoder state; updates on this value can already be received via SSRCMap events, however I think it also belongs on AudioPacket because there's a risk of delayed SSRCMap event delivery, and the client needs to know as soon as they receive a packet that changes it so they can reset their decoder state to correctly parse the packet)

This change deliberately does not attempt to provide a generalised parser for all the fields in the RTP Header to keep this PR small and focused just on the essential fields.

### DIFF
--- a/packages/voice/__tests__/VoiceReceiver.test.ts
+++ b/packages/voice/__tests__/VoiceReceiver.test.ts
@@ -68,6 +68,47 @@ describe('VoiceReceiver', () => {
 		expect(stream.read()).toEqual(RTP_PACKET.opusFrame);
 	});
 
+	test.each([
+		['Desktop', RTP_PACKET_DESKTOP, 10_217, 4_157_324_497],
+		['Chrome', RTP_PACKET_CHROME, 18_143, 660_155_095],
+		['Android', RTP_PACKET_ANDROID, 14_800, 3_763_991_879],
+	])('onUdpMessage: RTP header metadata from %s', async (_testName, RTP_PACKET, expectedSeq, expectedTs) => {
+		receiver['decrypt'] = vitest.fn().mockImplementationOnce(() => RTP_PACKET.decrypted);
+
+		const spy = vitest.spyOn(receiver.ssrcMap, 'get');
+		spy.mockImplementation(() => ({
+			audioSSRC: RTP_PACKET.ssrc,
+			userId: '123',
+		}));
+
+		const stream = receiver.subscribe('123');
+
+		receiver['onUdpMessage'](RTP_PACKET.packet);
+		await nextTick();
+		const packet = stream.read();
+		expect(packet.sequence).toEqual(expectedSeq);
+		expect(packet.timestamp).toEqual(expectedTs);
+		expect(packet.ssrc).toEqual(RTP_PACKET.ssrc);
+	});
+
+	test('onUdpMessage: AudioPacket is backwards compatible', async () => {
+		receiver['decrypt'] = vitest.fn().mockImplementationOnce(() => RTP_PACKET_DESKTOP.decrypted);
+
+		const spy = vitest.spyOn(receiver.ssrcMap, 'get');
+		spy.mockImplementation(() => ({
+			audioSSRC: RTP_PACKET_DESKTOP.ssrc,
+			userId: '123',
+		}));
+
+		const stream = receiver.subscribe('123');
+
+		receiver['onUdpMessage'](RTP_PACKET_DESKTOP.packet);
+		await nextTick();
+		const packet = stream.read();
+		expect(Buffer.isBuffer(packet)).toBe(true);
+		expect(packet).toEqual(RTP_PACKET_DESKTOP.opusFrame);
+	});
+
 	test('onUdpMessage: <8 bytes packet', () => {
 		expect(() => receiver['onUdpMessage'](Buffer.alloc(4))).not.toThrow();
 	});
--- a/packages/voice/__tests__/VoiceReceiver.test.ts
+++ b/packages/voice/__tests__/VoiceReceiver.test.ts
@@ -172,15 +172,17 @@ describe('VoiceReceiver', () => {
 		});
 	});
 
-	describe('decrypt', () => {
-		test('decrypt: aead_xchacha20_poly1305_rtpsize', () => {
+	describe('parsePacket', () => {
+		test('parsePacket: aead_xchacha20_poly1305_rtpsize', () => {
 			const nonceSpace = Buffer.alloc(24);
 
-			const decrypted = receiver['decrypt'](
+			const packet = receiver['parsePacket'](
 				XCHACHA20_SAMPLE.encrypted,
 				'aead_xchacha20_poly1305_rtpsize',
 				nonceSpace,
 				XCHACHA20_SAMPLE.key,
+				'123',
+				48_921,
 			);
 
 			const expectedNonce = Buffer.concat([
@@ -189,17 +191,24 @@ describe('VoiceReceiver', () => {
 			]);
 
 			expect(nonceSpace.equals(expectedNonce)).toEqual(true);
-			expect(decrypted.equals(XCHACHA20_SAMPLE.decrypted)).toEqual(true);
+			// Extension data (8 bytes) should be stripped from the 61-byte decrypted payload
+			expect(packet).toHaveLength(53);
+			expect(packet!.equals(XCHACHA20_SAMPLE.decrypted.subarray(8))).toEqual(true);
+			expect(packet!.sequence).toEqual(22_118);
+			expect(packet!.timestamp).toEqual(3_220_386_864);
+			expect(packet!.ssrc).toEqual(48_921);
 		});
 
-		test('decrypt: aead_aes256gcm_rtpsize', () => {
+		test('parsePacket: aead_aes256gcm_rtpsize', () => {
 			const nonceSpace = Buffer.alloc(12);
 
-			const decrypted = receiver['decrypt'](
+			const packet = receiver['parsePacket'](
 				AES256GCM_SAMPLE.encrypted,
 				'aead_aes256_gcm_rtpsize',
 				nonceSpace,
 				AES256GCM_SAMPLE.key,
+				'123',
+				50_615,
 			);
 
 			const expectedNonce = Buffer.concat([
@@ -208,7 +217,11 @@ describe('VoiceReceiver', () => {
 			]);
 
 			expect(nonceSpace.equals(expectedNonce)).toEqual(true);
-			expect(decrypted.equals(AES256GCM_SAMPLE.decrypted)).toEqual(true);
+			// No extension (X=0), so decrypted payload is the opus frame directly
+			expect(packet!.equals(AES256GCM_SAMPLE.decrypted)).toEqual(true);
+			expect(packet!.sequence).toEqual(41_884);
+			expect(packet!.timestamp).toEqual(2_668_332_016);
+			expect(packet!.ssrc).toEqual(50_615);
 		});
 	});
 });
--- a/packages/voice/__tests__/VoiceReceiver.test.ts
+++ b/packages/voice/__tests__/VoiceReceiver.test.ts
@@ -65,7 +65,8 @@ describe('VoiceReceiver', () => {
 
 		receiver['onUdpMessage'](RTP_PACKET.packet);
 		await nextTick();
-		expect(stream.read()).toEqual(RTP_PACKET.opusFrame);
+		const packet = stream.read();
+		expect(packet.payload).toEqual(RTP_PACKET.opusFrame);
 	});
 
 	test.each([
@@ -91,7 +92,7 @@ describe('VoiceReceiver', () => {
 		expect(packet.ssrc).toEqual(RTP_PACKET.ssrc);
 	});
 
-	test('onUdpMessage: AudioPacket is backwards compatible', async () => {
+	test('onUdpMessage: AudioPacket has payload and header fields', async () => {
 		receiver['decrypt'] = vitest.fn().mockImplementationOnce(() => RTP_PACKET_DESKTOP.decrypted);
 
 		const spy = vitest.spyOn(receiver.ssrcMap, 'get');
@@ -105,8 +106,11 @@ describe('VoiceReceiver', () => {
 		receiver['onUdpMessage'](RTP_PACKET_DESKTOP.packet);
 		await nextTick();
 		const packet = stream.read();
-		expect(Buffer.isBuffer(packet)).toBe(true);
-		expect(packet).toEqual(RTP_PACKET_DESKTOP.opusFrame);
+		expect(Buffer.isBuffer(packet.payload)).toBe(true);
+		expect(packet.payload).toEqual(RTP_PACKET_DESKTOP.opusFrame);
+		expect(typeof packet.sequence).toBe('number');
+		expect(typeof packet.timestamp).toBe('number');
+		expect(typeof packet.ssrc).toBe('number');
 	});
 
 	test('onUdpMessage: <8 bytes packet', () => {
@@ -192,8 +196,8 @@ describe('VoiceReceiver', () => {
 
 			expect(nonceSpace.equals(expectedNonce)).toEqual(true);
 			// Extension data (8 bytes) should be stripped from the 61-byte decrypted payload
-			expect(packet).toHaveLength(53);
-			expect(packet!.equals(XCHACHA20_SAMPLE.decrypted.subarray(8))).toEqual(true);
+			expect(packet!.payload).toHaveLength(53);
+			expect(packet!.payload.equals(XCHACHA20_SAMPLE.decrypted.subarray(8))).toEqual(true);
 			expect(packet!.sequence).toEqual(22_118);
 			expect(packet!.timestamp).toEqual(3_220_386_864);
 			expect(packet!.ssrc).toEqual(48_921);
@@ -218,7 +222,7 @@ describe('VoiceReceiver', () => {
 
 			expect(nonceSpace.equals(expectedNonce)).toEqual(true);
 			// No extension (X=0), so decrypted payload is the opus frame directly
-			expect(packet!.equals(AES256GCM_SAMPLE.decrypted)).toEqual(true);
+			expect(packet!.payload.equals(AES256GCM_SAMPLE.decrypted)).toEqual(true);
 			expect(packet!.sequence).toEqual(41_884);
 			expect(packet!.timestamp).toEqual(2_668_332_016);
 			expect(packet!.ssrc).toEqual(50_615);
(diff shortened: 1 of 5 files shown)


packages/voice/src/receive/AudioReceiveStream.ts
```diff
@@ -36,6 +36,28 @@ export interface AudioReceiveStreamOptions extends ReadableOptions {
 	end: EndBehavior;
 }
 
+/**
+ * A Buffer containing a decoded Opus packet with RTP header metadata.
+ */
```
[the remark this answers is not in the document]
petergeneric: fixed in 5046f133dd238523263fc17df293fffa3dc6b2a1

packages/voice/src/receive/VoiceReceiver.ts
```diff
@@ -165,6 +175,8 @@ export class VoiceReceiver {
 	 */
 	public onUdpMessage(msg: Buffer) {
 		if (msg.length <= 8) return;
+		const sequence = msg.readUInt16BE(2);
+		const timestamp = msg.readUInt32BE(4);
 		const ssrc = msg.readUInt32BE(8);
```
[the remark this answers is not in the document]
petergeneric: fixed in 5046f133dd238523263fc17df293fffa3dc6b2a1

packages/voice/src/receive/VoiceReceiver.ts
Qjuh: Since you're already touching RTP headers I noticed that this is no longer correct. We have the full packet here, so this should check the eXtension bit (4th bit of the first byte) of the header instead of looking for magic bytes of the header extension data. And it even looks like this is stripping the wrong amount of header data, unless I'm missing something here.

packages/voice/src/receive/VoiceReceiver.ts
```diff
@@ -154,7 +167,30 @@ export class VoiceReceiver {
...
+		} else {
+			return null;
+		}
+	}
+
+	/**
+	 * Extends the Buffer for Opus audio data with RTP Header information
+	 *
+	 * @param buffer the opus packet data to extend
+	 * @param sequence the sequence number of the packet
```
Qjuh:
```suggestion
	 * @param buffer - the opus packet data to extend
	 * @param sequence - the sequence number of the packet
```
linting should've caught this...

packages/voice/src/receive/VoiceReceiver.ts
```diff
@@ -154,7 +167,30 @@ export class VoiceReceiver {
...
+	 * @param timestamp see definition in
+	 * @param ssrc x
```
Qjuh: this is missing half the description

packages/voice/src/receive/VoiceReceiver.ts
```diff
@@ -154,7 +167,30 @@ export class VoiceReceiver {
...
+	 * @returns the input buffer, with RTP header information added
+	 */
+	private static addPacketHeaders(buffer: Buffer, sequence: number, timestamp: number, ssrc: number): AudioPacket {
```
Qjuh: why is this a private static method instead of simply a function?

packages/voice/src/receive/VoiceReceiver.ts
```diff
@@ -154,7 +167,30 @@ export class VoiceReceiver {
...
+		Object.defineProperties(buffer, {
+			sequence: { value: sequence, writable: false, enumerable: false, configurable: false },
+			timestamp: { value: timestamp, writable: false, enumerable: false, configurable: false },
+			ssrc: { value: ssrc, writable: false, enumerable: false, configurable: false },
+		});
+		return buffer as AudioPacket;
```
Qjuh: I don't think adding these as hidden properties is what we'd want for users of this library. They should be exposed in a way that is helpful and visible from documentation and the interface.

packages/voice/src/receive/VoiceReceiver.ts
petergeneric: Keeping it within the class seemed more in-keeping with style of class, I've changed it to a function below the class, fixed in 57034c1bdb4c075b4dd3b450c838d3d615f342ac

packages/voice/src/receive/VoiceReceiver.ts
petergeneric: fixed in 57034c1bdb4c075b4dd3b450c838d3d615f342ac

packages/voice/src/receive/VoiceReceiver.ts
petergeneric: Are you happy with the general approach of using `Object.defineProperties` and just think they should be enumerable, or do you think it should be a more general change?
I was trying to keep it backwards-compatible by extending the original Buffer, and making it documentable via the `AudioPacket` interface - happy to change approach here

packages/voice/src/receive/VoiceReceiver.ts
```diff
@@ -154,7 +167,12 @@ export class VoiceReceiver {
 			if (daveSession) packet = daveSession.decrypt(packet, userId)!;
 		}
 
-		return packet;
+		// Extend packet with RTP header information
+		if (packet) {
+			return addPacketHeaders(packet, sequence, timestamp, ssrc);
+		} else {
+			return null;
+		}
```
vladfrangu: How can packet be non-truthy at this point?

packages/voice/src/receive/AudioReceiveStream.ts
```diff
@@ -36,6 +36,28 @@ export interface AudioReceiveStreamOptions extends ReadableOptions {
 	end: EndBehavior;
 }
 
+/**
+ * A Buffer containing encoded Opus packet data and key RTP Header metadata.
+ */
+export interface AudioPacket extends Buffer {
```
vladfrangu: Why not make it a proper class instead? Adding fields to a buffer feels...eeh

packages/voice/src/receive/VoiceReceiver.ts
petergeneric: This is because `daveSession.decrypt` returns `Buffer|null`

vladfrangu: But thats handled at line 151 no?

vladfrangu: oh, missed line 160-168, sorry

packages/voice/src/receive/AudioReceiveStream.ts
petergeneric: I agree, I went into this change assuming backwards-compatibility with client API was a priority.

It sounds like from this feedback and from @Qjuh that you'd be happy to break backwards compatibility by having the stream send a structured `AudioPacket` that has fields for headers and a payload `Buffer`?

packages/voice/src/receive/VoiceReceiver.ts
petergeneric: @vladfrangu has also added feedback on this, suggested a proper class, so sounds like strict backwards compatibility isn't essential

packages/voice/src/receive/VoiceReceiver.ts
petergeneric: Yes that's right, it'll need to consider the case where the CC field is > 0.

I've been working on this. Started as using the X bit but it's now a larger change (`parsePacket` now parses RTP Header fields, and passes payload start pos into `decrypt` rather than it recomputing. As a result the unit tests call `parsePacket` rather than directly calling `decrypt`; I also took the opportunity to change some buffer variable names for better clarity). See 7575a2554991ede3cd1860d7e9000d2a1ab139bd

Are you happy with that being part of this PR, or should it be split out into a separate PR?

packages/voice/src/receive/VoiceReceiver.ts
petergeneric: Addressed in 5d344b6c8651eb302bb7372f9082b6c15850a261, AudioPacket is an interface that carries the `Buffer` as a field

packages/voice/src/receive/AudioReceiveStream.ts
petergeneric: Addressed in 5d344b6c8651eb302bb7372f9082b6c15850a261, AudioPacket is an interface that carries the `Buffer` as a field

packages/voice/src/receive/VoiceReceiver.ts
petergeneric: btw have addressed this by having the null DAVE decrypt result be handled as an early return of null close to the dave decrypt call in a similar pattern to the throw guard after rtp decrypt method call, to make reasoning about the truthiness easier (although let me know if single return is preferred and I can undo that...)

packages/voice/src/receive/VoiceReceiver.ts
```diff
@@ -126,35 +132,61 @@ export class VoiceReceiver {
...
+			if (daveSession) {
+				payload = daveSession.decrypt(payload, userId)!;
+
+				if (!payload) return null; // decryption failed but should be ignored
+			}
 		}
 
-		return packet;
+		// Construct AudioPacket with Opus payload and RTP header information
+		return { payload, sequence, timestamp, ssrc };
```
Qjuh: This should probably be an actual class instead of a plain object.

petergeneric: Ok - are you happy to keep interface and make class private, or would you rather make class public? If public class, what's your preferred way to deal with constructor being visible to end-users but not for their use?

(Asking because I can't find any other instances of private class implementing just an interface in the code, so not sure if you try to avoid this approach)

Qjuh: I don't really see a reason to make the class private. Why do you want to keep users from using it?

petergeneric: So that changing the constructor signature in the future doesn't break user code

petergeneric: Have made it a public class in 8dd7fc50bf578d9a7168c1edc18d8181513a1701 with a docstring on the constructor telling users it's not intended as a stable api

packages/voice/src/receive/AudioReceiveStream.ts
```diff
@@ -36,6 +36,49 @@ export interface AudioReceiveStreamOptions extends ReadableOptions {
...
+	public readonly ssrc: number;
+
+	/**
+	 * RTP timestamp of this packet (32-bit, wraps at 2^32 - 1, 48kHz clock).
+	 */
+	public readonly timestamp: number;
+
+	/**
+	 * Construct a new AudioPacket.
+	 * **This is not a stable public API.**
```
Qjuh: I don't see why it's not. voice follows 0ver currently, most releases of it are major, so just because we might change this in the future doesn't mean it's not stable. I fail to see your repeated issue with just making this a regular, public class usable by consumers however they see fit.

packages/voice/src/receive/VoiceReceiver.ts
Qjuh, line 119: Also just now noticed that this is bad in terms of performance, maybe should be changed too.
The data needed here already exists in `buffer` and should be extracted as subarray instead of reconstructed from the two subarrays.

Not sure if you want to touch this in this PR though.

packages/voice/src/receive/AudioReceiveStream.ts
petergeneric: I just have an instinctive dislike of projects breaking backwards compatibility, but point taken on your versioning approach, will remove comment

packages/voice/src/receive/VoiceReceiver.ts
petergeneric: Well, a small can of worms 🙂 fairly straight-forward change on the face of it, but I uncovered an issue in Secretbox.ts type decls... My inclination is to raise it as a small follow-on PR given this - but up to you.

Basic version of the change is here: https://github.com/petergeneric/discord.js/commit/1d6e5082419c69929fa15de773a3a06832bd52fc

In trying to figure out why the `aead_xchacha20_poly1305_rtpsize` codepath wraps `return Buffer.from` around a `Buffer`-returning method, I spotted that the Secretbox.ts decl of the return as `Buffer` is wrong, only sodium-native returns a Buffer. The other backends return a Uint8Array (e.g. `@noble/ciphers/chacha.js`), so I think that needs fixed too (although the good news is that by doing that, we can save another allocation when using sodium-native)


P.S. the change also improves the `aead_aes256_gcm_rtpsize` path:
1. The previous code didn't actually verify the auth tag
2. Avoids Buffer.concat in return happy path. `final()` in GCM mode should never return any data

vladfrangu: Separate PR please but if we can drop needless buffer.from's, i am all for it

Qjuh (approved):

petergeneric:
FYI: code coverage fail is misleading, this is existing lack of coverage (these lines are just reorganisations of the code that was already there to make it easier to understand). It looks like there's no test data for DAVE

packages/voice/src/receive/VoiceReceiver.ts
petergeneric: working on that in https://github.com/petergeneric/discord.js/commits/voice-decrypt-buffer-allocation-optimise/ and will create a PR after this is merged so going to resolve this thread. Found a minor mirrored improvement in Networking.ts packet encrypt flow

vladfrangu:
For testing sake, can you draft up a quick code sample that can be used to see how it works now? (bonus point, we can document the breaking change)

petergeneric:
@vladfrangu Sure, here's a sample snippet that'll join a voice channel and subscribe to all speakers, is this what you were thinking of?

```typescript
import type { VoiceBasedChannel } from 'discord.js';
import {
	joinVoiceChannel,
	entersState,
	VoiceConnectionStatus,
	EndBehaviorType,
	type AudioPacket,
} from '@discordjs/voice';

async function joinAndRecordChannel(channel: VoiceBasedChannel) {
	// Join the voice channel
	const connection = joinVoiceChannel({
		channelId: channel.id,
		guildId: channel.guild.id,
		adapterCreator: channel.guild.voiceAdapterCreator,
		selfDeaf: false, // We want to receive audio
		selfMute: true, // We don't want to send audio
	});

	// Wait for the connection to be ready
	await entersState(connection, VoiceConnectionStatus.Ready, 30_000);

	const receiver = connection.receiver;

	// Subscribe to all users when they speak
	const subscribedUsers = new Set<string>();
	receiver.speaking.on('start', (userId) => {
		if (subscribedUsers.has(userId)) return;
		subscribedUsers.add(userId);

		// Subscribe to the user's audio stream
		const audioStream = receiver.subscribe(userId, {
			end: { behavior: EndBehaviorType.Manual },
		});

		// New API: 'data' event now emits an AudioPacket rather than a Buffer
		audioStream.on('data', (packet: AudioPacket) => handleOpusPacket(userId, packet));

		audioStream.on('error', (err) => {
			console.error(`Audio stream error for ${userId}:`, err);
		});
	});
}

function handleOpusPacket(userId: string, packet: AudioPacket) {
	console.log(
		`User ${userId}: rtp.ssrc=${packet.ssrc} rtp.seq=${packet.sequence} rtp.ts=${packet.timestamp}@48kHz opus=${packet.payload.length}B`,
	);
}
```

(a better example might show how to use the new fields in opus decoding to produce a fully synchronised PCM from all speakers, but I've extracted this from my transcription bot which defers opus decode to an offline process to minimise load while recording)


Here's an initial draft of the breaking change doc entry (have tried to match general style from the Changes in v14 .mdx):

### AudioReceiveStream

`AudioReceiveStream` now emits `AudioPacket` instances instead of raw `Buffer`s. Each `AudioPacket` contains the Opus payload along with relevant RTP header metadata (`sequence`, `timestamp`, and `ssrc`).

The Opus audio data can be found in the `payload` property:

```diff
- // Previously, each chunk was a raw Buffer of Opus data:
- receiver.subscribe(userId).on('data', (chunk) => {
- 	processOpusData(userId, chunk); // chunk was a Buffer
- });

+ // Now, each chunk is an AudioPacket with payload and RTP metadata:
+ receiver.subscribe(userId).on('data', (packet: AudioPacket) => {
+ 	processOpusData(userId, packet.payload); // Opus data is on .payload
+ 	console.log(packet.sequence);    // RTP sequence number
+ 	console.log(packet.timestamp);   // RTP timestamp (48kHz clock)
+ 	console.log(packet.ssrc);        // RTP synchronization source
+ });
```

vladfrangu:
That is perfect, thanks!

vladfrangu:
Can you rebase your PR please?

petergeneric:
@vladfrangu Looks like the other PR was created after the last comment on this PR, and (given I've been doing cleanup in this PR, and other changes as requested), the conflict is a bit annoying to me... I am willing to look into this but only if there's a path to actually getting this merged in quickly, it feels like it's just sitting in limbo

petergeneric:
@vladfrangu  / @Qjuh what is happening with this PR?

dominik1106:
Hey, sorry for bumping, but I'm also interested in the state of the PR since I'm facing similar issues with desync and as far as I can see, this should solve it

petergeneric:
@vladfrangu any update?

vladfrangu:
Sorry for the dead silence, can you get this PR rebased please? From my end these changes look good, I will take another quick look but should be all good

vladfrangu (approved):

petergeneric:
rebased with no conflicts, pr tests showing awaiting maintainer approval

packages/voice/__tests__/VoiceReceiver.test.ts
```diff
@@ -65,7 +65,52 @@ describe('VoiceReceiver', () => {
...
+		await nextTick();
+		const packet = stream.read();
+		expect(Buffer.isBuffer(packet.payload)).toBe(true);
+		expect(packet.payload).toEqual(RTP_PACKET_DESKTOP.opusFrame);
+		expect(typeof packet.sequence).toBe('number');
+		expect(typeof packet.timestamp).toBe('number');
+		expect(typeof packet.ssrc).toBe('number');
 	});
 
 	test('onUdpMessage: <8 bytes packet', () => {
```
almeidx, line 116: technically still correct, but "<=12" would be more accurate now.

almeidx (approved):
🤠

PR #11435, opened 2026-03-02
chore: move proxy-container to apps

almeidx:
It is not a package. It's mostly a deployable app, just like the website and the guide. 
This also allows us to remove the special logic from `generateReleaseTree` that was handling this special "package"

### DIFF
--- a/packages/discord.js/src/client/websocket/handlers/APPLICATION_COMMAND_PERMISSIONS_UPDATE.js
+++ b/packages/discord.js/src/client/websocket/handlers/APPLICATION_COMMAND_PERMISSIONS_UPDATE.js
@@ -4,6 +4,16 @@ const { Events } = require('../../../util/Events.js');
 
 module.exports = (client, { d: data }) => {
   /**
+   * Represents the properties of an application command permissions update
+   *
+   * @typedef {Object} ApplicationCommandPermissionsUpdateData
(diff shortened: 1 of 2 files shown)


apps/proxy-container/src/index.ts
```diff
@@ -1,5 +1,4 @@
 import { createServer } from 'node:http';
-import process from 'node:process';
```
Qjuh, line 2: This looks like a symptom of eslint.config.json needing  adjustments because it currently assumes all apps are nextjs applications.

.github/ISSUE_TEMPLATE/01-package_bug_report.yml
```diff
@@ -14,6 +14,7 @@ body:
       label: Which package is this bug report for?
       options:
         - discord.js
+        - Proxy Container
```
almeidx: We could change this to `discordjs/proxy` since in practice that's the actual identifier of the container.
Thoughts? @Jiralite

Jiralite: Could do, sure.

Could also create a new issue template for this since this is a package bug report. The other apps are called "Websites bug report". I personally see no problem with having another issue template for this, even if it is majorly unused. It's good for organisation and is different enough to warrant it?

PR #11437, opened 2026-03-03
feat: modal radio group and checkbox components for v14

imnaiyar:
~~I have only cherry-picked it, need to test if it works properly for v14 before marking it ready (though i imagine it should)~~

Backports #11395

### DIFF
--- a/.github/workflows/documentation.yml
+++ b/.github/workflows/documentation.yml
@@ -1,8 +1,6 @@
 name: Documentation
 on:
   push:
-    branches:
-      - 'main'
     paths:
       - 'packages/*/src/**'
       - '!packages/create-discord-bot/**'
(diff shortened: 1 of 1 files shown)


Qjuh:
Reviewing this I was wondering whether v14 would need `XComponent` classes for the new modal components. But looking through where they get used I found out that they wouldn't be used anywhere because `LabelComponent` isn't used anywhere either, so probably not needed.

imnaiyar:
> Reviewing this I was wondering whether v14 would need `XComponent` classes for the new modal components. But looking through where they get used I found out that they wouldn't be used anywhere because `LabelComponent` isn't used anywhere either, so probably not needed.

Yeah it doesn't get used anywhere, i raised this internally here https://discord.com/channels/222078108977594368/1402934237358985246/1412389550503891065 back when label component was introduced. The only reason i kept `LabelComponent` was because jiralite had some ideas on how we can use them in the future at the time (you'll have to scroll a bit though). I don't think any of that is needed anymore

vladfrangu (approved):

PR #11443, opened 2026-03-06
feat(DMChannel)!: allow partial DMChannel without client user

### LINKED ISSUE #11429: Interactions fail to check if DMs are with another user before updating cache

discord.js

### Issue description

1. User A installs Bot to their account.
2. User A runs `/command` from Bot in their DMs with User B.
3. Bot tries to send a message to User A's DMs (via `user.send()`).
4. Fails! 50001 Missing Access error because Bot tried to send message to User B's DMs with User A instead of Bot's DMs with User A.

The bot sends reminders and has been struggling with seemingly spontaneous 50001 Missing Access errors for a while now which is causing reminders to have massive delays from the intended reminder date.

### Code sample

```typescript

```

### Versions

- discord.js 14.25.1
- Node.js 24.14.0

### Issue priority

Medium (should be fixed soon)

Not applicable

No Intents

_No response_

Qjuh:
Since user installed applications can run in DMChannels not containing the client user we need to account for that in case of partial DMChannels being cached.

This PR now stores the ids of all known recipients, not only one. That way we can discern between a DMChannel between the client user and another user or a DMChannel between two non-client users.

Also properly resolves the issues fixed by #8950 and #9774 instead of just fixing the symptoms caused by us only storing one of the recipient's ids.

Fixes #11429

BREAKING CHANGE: `DMChannel#recipientId` is now nullable.


packages/discord.js/src/structures/DMChannel.js
```diff
@@ -78,7 +80,21 @@ class DMChannel extends BaseChannel {
...
+   * @type {?Snowflake}
+   * @readonly
+   */
+  get recipientId() {
+    if (this.recipientIds.includes(this.client.user.id)) {
+      return this.recipientIds.find(recipientId => recipientId !== this.client.user.id) ?? null;
+    }
+
+    return null;
+  }
```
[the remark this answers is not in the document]
Qjuh: aware of both of these, but both are intended. If a user of the lib uses those methods on a DMChannel not containing the ClientUser then the resulting errors are on them.

packages/discord.js/src/client/actions/Action.js
```diff
@@ -33,10 +33,15 @@ class Action {
...
     const id = data.channel_id ?? data.id;
 
-    if (!('recipients' in data)) {
-      // Try to resolve the recipient, but do not add the client user.
+    if ('recipients' in data) {
+      // Try to resolve the recipient, but do not add if already existing in recipients.
       const recipient = data.author ?? data.user ?? { id: data.user_id };
-      if (recipient.id !== this.client.user.id) payloadData.recipients = [recipient];
+      if (data.recipients.every(existingRecipient => recipient.id !== existingRecipient.id))
+        payloadData.recipients = [...data.recipients, recipient];
```
vladfrangu: 1. formatting pls, {} it up
2. the every feels off, if every recipient is not the message author, add the message author?

vladfrangu (approved):
Backport too

packages/discord.js/src/client/actions/Action.js
Qjuh: "Every entry is not the recipient" means "the recipient is not yet in the array, so needs to be added"

vladfrangu: _Personally_ i'd use a `some` to make it clearer (and iirc it also exits early when a match is found) but its not a blocker

Jiralite (approved):

didinele (approved):

PR #11459, opened 2026-03-20
feat(Guild): add message search

kshitijanurag:
Add guild messages search functionality.

- Discord API Docs: https://github.com/discord/discord-api-docs/pull/8010
- Depends on: https://github.com/discordjs/discord-api-types/pull/1583

### DIFF
--- a/packages/discord.js/src/structures/Guild.js
+++ b/packages/discord.js/src/structures/Guild.js
@@ -1,9 +1,10 @@
 'use strict';
 
+const { setTimeout: sleep } = require('node:timers/promises');
 const { Collection } = require('@discordjs/collection');
 const { makeURLSearchParams } = require('@discordjs/rest');
 const { DiscordSnowflake } = require('@sapphire/snowflake');
-const { ChannelType, GuildPremiumTier, Routes, GuildFeature } = require('discord-api-types/v10');
+const { ChannelType, GuildFeature, GuildPremiumTier, Routes, RESTJSONErrorCodes } = require('discord-api-types/v10');
 const { DiscordjsError, DiscordjsTypeError, ErrorCodes } = require('../errors/index.js');
 const { AutoModerationRuleManager } = require('../managers/AutoModerationRuleManager.js');
 const { GuildApplicationCommandManager } = require('../managers/GuildApplicationCommandManager.js');
@@ -29,6 +30,7 @@ const { GuildOnboarding } = require('./GuildOnboarding.js');
 const { GuildPreview } = require('./GuildPreview.js');
 const { GuildTemplate } = require('./GuildTemplate.js');
 const { Integration } = require('./Integration.js');
+const { Message } = require('./Message.js');
 const { Webhook } = require('./Webhook.js');
 const { WelcomeScreen } = require('./WelcomeScreen.js');
 
@@ -1076,6 +1078,182 @@ class Guild extends AnonymousGuild {
     });
 
     return new GuildOnboarding(this.client, newData);
+  }
+
+  /**
+   * Options used to search messages in a guild.
+   *
+   * @typedef {Object} GuildSearchMessagesOptions
+   * @property {string} [content] Filter messages by given content (max 1024 characters)
+   * @property {Snowflake} [maxId] Get messages before a given message ID
+   * @property {Snowflake} [minId] Get messages after a given message ID
+   * @property {Snowflake[]} [channelId] Filter messages by given channel IDs (max 500)
+   * @property {MessageSearchAuthorType[]} [authorType] Filter messages by author type.
+   * @property {Snowflake[]} [authorId] Filter messages by given author IDs (max 100)
+   * @property {Snowflake[]} [mentions] Filter messages that mention given user IDs (max 100)
+   * @property {Snowflake[]} [mentionsRoleId] Filter messages that mention given role IDs (max 100)
+   * @property {boolean} [mentionEveryone] Filter messages by whether they mention @everyone
+   * @property {Snowflake[]} [repliedToUserId] Filter messages that reply to given user IDs (max 100)
+   * @property {Snowflake[]} [repliedToMessageId] Filter messages that reply to given message IDs (max 100)
+   * @property {boolean} [pinned] Filter messages by whether they are pinned
+   * @property {MessageSearchHasType[]} [has] Filter messages by whether they contain specific content types.
+   * @property {MessageSearchEmbedType[]} [embedType] Filter messages by embed type.
+   * @property {string[]} [embedProvider] Filter messages by embed provider, case-sensitive (max 100)
+   * @property {string[]} [linkHostname] Filter messages by link hostname (e.g. `'discord.com'`) (max 100)
+   * @property {string[]} [attachmentFilename] Filter messages by attachment filename (max 100)
+   * @property {string[]} [attachmentExtension] Filter messages by attachment extension (e.g. `'png'`) (max 100)
+   * @property {MessageSearchSortMode} [sortBy] The sorting algorithm to use.
+   * @property {string} [sortOrder] The direction to sort. One of `'asc'` or `'desc'` (default).
+   * <info>Sort order is not respected when sorting by relevance.</info>
+   * @property {boolean} [includeNsfw=false] Whether to include results from age-restricted channels
+   * @property {number} [limit=25] Max number of messages to return (1–25)
+   * @property {number} [offset] Number to offset the returned messages by (max 9975)
+   * @property {number} [slop=2] Max number of words to skip between matching tokens in the `content` filter (max 100)
+   * @property {boolean} [cache=true] Whether to cache the fetched messages
+   * @property {AbortSignal} [signal] An {@link AbortSignal} to cancel the request, including any automatic retries
+   * while the guild's messages are still being indexed
+   */
+
+  /**
+   * Result returned from searching guild messages.
+   *
+   * @typedef {Object} GuildSearchMessagesResult
+   * @property {number} [documentsIndexed] The number of documents that have been indexed during the current
+   * index operation, if any
+   * @property {boolean} doingDeepHistoricalIndex Whether the guild is undergoing a deep historical indexing
+   * operation
+   * @property {Collection<Snowflake, Message>} messages A collection of the matched messages
+   * @property {Collection<Snowflake, ThreadChannel>} threads A collection of the threads that contain the
+   * returned messages
+   * @property {Collection<Snowflake, Collection<Snowflake, ThreadMember>>} threadMembers A collection of thread
+   * members for any threads the matched messages belong to that the current user is a member of, keyed by
+   * thread ID, then by user ID
+   * @property {number} totalResults The total number of messages that match the query
+   */
+
+  /**
+   * Returns a collection of messages without the `reactions` property that match a search query in the guild. Requires the {@link PermissionFlagsBits.ReadMessageHistory} permission.
+   * <warn>This is restricted according to whether the {@link GatewayIntentBits.MessageContent} Privileged Intent is enabled for your application.</warn>
+   * <info>Due to speed optimizations, search may return slightly fewer results than the limit specified when messages have not been accessed for a long time.
+   * Clients should not rely on the length of the `messages` array to paginate results.
+   *
+   * Additionally, when messages are actively being created or deleted, the `totalResults` property may not be accurate.
+   *
+   * If the entity you are searching is not yet indexed, Discord will return a 202 accepted response.
+   * This method automatically waits and retries the request until the index is ready. Pass a `signal` if you would
+   * rather cancel the operation than wait for automatic retries.</info>
+   *
+   * @param {GuildSearchMessagesOptions} [options={}] Options for searching messages
+   * @returns {Promise<GuildSearchMessagesResult>}
+   * @example
+   * // Search for messages containing 'Hello'
+   * guild.searchMessages({ content: 'Hello' })
+   *   .then(result => console.log(`Found ${result.totalResults} results.`))
+   *   .catch(console.error);
+   * @example
+   * // Search for messages with images, excluding bots, sorted by oldest first
+   * guild
+   *   .searchMessages({ has: [MessageSearchHasType.Image], authorType: [MessageSearchAuthorType.NotBot], sortOrder: 'asc' })
+   *   .then(result => {
+   *     for (const message of result.messages) console.log(message.url);
+   *   })
+   *   .catch(console.error);
+   */
+  async searchMessages({
+    cache = true,
+    signal,
+    content,
+    maxId,
+    minId,
+    channelId,
+    authorType,
+    authorId,
+    mentions,
+    mentionsRoleId,
+    mentionEveryone,
+    repliedToUserId,
+    repliedToMessageId,
+    pinned,
+    has,
+    embedType,
+    embedProvider,
+    linkHostname,
+    attachmentFilename,
+    attachmentExtension,
+    sortBy,
+    sortOrder,
+    includeNsfw,
+    limit,
+    offset,
+    slop,
+  } = {}) {
+    const query = makeURLSearchParams({
+      content,
+      max_id: maxId,
+      min_id: minId,
+      channel_id: channelId,
+      author_type: authorType,
+      author_id: authorId,
+      mentions,
+      mentions_role_id: mentionsRoleId,
+      mention_everyone: mentionEveryone,
+      replied_to_user_id: repliedToUserId,
+      replied_to_message_id: repliedToMessageId,
+      pinned,
+      has,
+      embed_type: embedType,
+      embed_provider: embedProvider,
+      link_hostname: linkHostname,
+      attachment_filename: attachmentFilename,
+      attachment_extension: attachmentExtension,
+      sort_by: sortBy,
+      sort_order: sortOrder,
+      include_nsfw: includeNsfw,
+      limit,
+      offset,
+      slop,
+    });
+
+    let data = await this.client.rest.get(Routes.guildMessagesSearch(this.id), { query, signal });
+
+    while (data.code === RESTJSONErrorCodes.IndexNotYetAvailable) {
+      const retryAfter = data.retry_after || 1;
+
+      await sleep(retryAfter * 1_000, undefined, { signal });
+      data = await this.client.rest.get(Routes.guildMessagesSearch(this.id), { query, signal });
+    }
+
+    const threads = (data.threads ?? []).reduce(
+      (collection, thread) => collection.set(thread.id, this.client.channels._add(thread, this, { cache })),
+      new Collection(),
+    );
+
+    const messages = data.messages.flat().reduce((collection, message) => {
+      const channel = threads.get(message.channel_id) ?? this.client.channels.cache.get(message.channel_id);
+      return collection.set(
+        message.id,
+        channel?.messages._add(message, cache) ?? new Message(this.client, { ...message, guild_id: this.id }),
+      );
+    }, new Collection());
+
+    const threadMembers = (data.members ?? []).reduce((collection, member) => {
+      const thread = threads.get(member.id) ?? this.client.channels.cache.get(member.id);
+      const threadMember = thread?.members._add(member, cache);
+      if (!threadMember) return collection;
+
+      const membersForThread = collection.ensure(member.id, () => new Collection());
+      membersForThread.set(member.user_id, threadMember);
+      return collection;
+    }, new Collection());
+
+    return {
+      doingDeepHistoricalIndex: data.doing_deep_historical_index,
+      ...('documents_indexed' in data && { documentsIndexed: data.documents_indexed }),
+      messages,
+      threads,
+      threadMembers,
+      totalResults: data.total_results,
+    };
   }
 
   /**
--- a/packages/discord.js/typings/index.d.ts
+++ b/packages/discord.js/typings/index.d.ts
@@ -175,6 +175,10 @@ import {
   MessageFlags,
   MessageReferenceType,
   MessageType,
+  MessageSearchAuthorType,
+  MessageSearchEmbedType,
+  MessageSearchHasType,
+  MessageSearchSortMode,
   NameplatePalette,
   OAuth2Scopes,
   OverwriteType,
@@ -1507,6 +1511,7 @@ export class Guild extends AnonymousGuild {
   public widgetImageURL(style?: GuildWidgetStyle): string;
   public leave(): Promise<Guild>;
   public disableInvites(disabled?: boolean): Promise<Guild>;
+  public searchMessages(options?: GuildSearchMessagesOptions): Promise<GuildSearchMessagesResult>;
   public setIncidentActions(incidentActions: IncidentActionsEditOptions): Promise<IncidentActions>;
   public setAFKChannel(afkChannel: VoiceChannelResolvable | null, reason?: string): Promise<Guild>;
   public setAFKTimeout(afkTimeout: number, reason?: string): Promise<Guild>;
@@ -6443,6 +6448,44 @@ export interface GuildSearchMembersOptions {
   cache?: boolean;
   limit?: number;
   query: string;
+}
+
+export interface GuildSearchMessagesOptions {
+  attachmentExtension?: readonly string[];
+  attachmentFilename?: readonly string[];
+  authorId?: readonly Snowflake[];
+  authorType?: readonly MessageSearchAuthorType[];
+  cache?: boolean;
+  channelId?: readonly Snowflake[];
+  content?: string;
+  embedProvider?: readonly string[];
+  embedType?: readonly MessageSearchEmbedType[];
+  has?: readonly MessageSearchHasType[];
+  includeNsfw?: boolean;
+  limit?: number;
+  linkHostname?: readonly string[];
+  maxId?: Snowflake;
+  mentionEveryone?: boolean;
+  mentions?: readonly Snowflake[];
+  mentionsRoleId?: readonly Snowflake[];
+  minId?: Snowflake;
+  offset?: number;
+  pinned?: boolean;
+  repliedToMessageId?: readonly Snowflake[];
+  repliedToUserId?: readonly Snowflake[];
+  signal?: AbortSignal;
+  slop?: number;
+  sortBy?: MessageSearchSortMode;
+  sortOrder?: 'asc' | 'desc';
+}
+
+export interface GuildSearchMessagesResult {
+  documentsIndexed?: number;
+  doingDeepHistoricalIndex: boolean;
+  messages: ReadonlyCollection<Snowflake, Message<true>>;
+  threadMembers: ReadonlyCollection<Snowflake, ReadonlyCollection<Snowflake, ThreadMember>>;
+  threads: ReadonlyCollection<Snowflake, ThreadChannel>;
+  totalResults: number;
 }
 
 export interface GuildListMembersOptions {
--- a/packages/discord.js/src/util/APITypes.js
+++ b/packages/discord.js/src/util/APITypes.js
@@ -565,6 +565,26 @@
  */
 
 /**
+ * @external MessageSearchAuthorType
+ * @see {@link https://discord-api-types.dev/api/discord-api-types-v10/enum/MessageSearchAuthorType}
+ */
+
+/**
+ * @external MessageSearchEmbedType
+ * @see {@link https://discord-api-types.dev/api/discord-api-types-v10/enum/MessageSearchEmbedType}
+ */
+
+/**
+ * @external MessageSearchHasType
+ * @see {@link https://discord-api-types.dev/api/discord-api-types-v10/enum/MessageSearchHasType}
+ */
+
+/**
+ * @external MessageSearchSortMode
+ * @see {@link https://discord-api-types.dev/api/discord-api-types-v10/enum/MessageSearchSortMode}
+ */
+
+/**
  * @external MessageType
  * @see {@link https://discord-api-types.dev/api/discord-api-types-v10/enum/MessageType}
  */


packages/discord.js/src/structures/Guild.js
```diff
@@ -1078,6 +1088,282 @@ class Guild extends AnonymousGuild {
...
+    includeNsfw,
+    limit,
+    offset,
+    slop,
+  } = {}) {
+    if (content !== undefined && typeof content !== 'string') {
+      throw new DiscordjsTypeError(ErrorCodes.InvalidType, 'content', 'string');
+    }
+
+    if (!Number.isInteger(limit)) {
```
almeidx: We typically do not do this sort of in-depth input validation

packages/discord.js/src/structures/Guild.js
```diff
@@ -1078,6 +1088,282 @@ class Guild extends AnonymousGuild {
...
+  /**
+   * Result returned from searching guild messages.
+   *
+   * @typedef {Object} GuildSearchMessagesResult
+   * @property {number} totalResults The total number of messages that match the query
+   * @property {Collection<Snowflake, Message>} messages A collection of the matched messages
+   */
+
+  /**
+   * Returns a collection of messages without the `reactions` property that match a search query in the guild. Requires the `READ_MESSAGE_HISTORY` permission.
```
almeidx: The permission reference should be a link to the equivalent on `PermissionFlagsBits` (e.g., like the intent link below)

packages/discord.js/typings/index.d.ts
```diff
@@ -6388,6 +6393,39 @@ export interface GuildSearchMembersOptions {
   query: string;
 }
 
+export interface GuildSearchMessagesOptions {
+  attachmentExtension?: string[];
+  attachmentFilename?: string[];
+  authorId?: Snowflake[];
+  authorType?: (MessageSearchAuthorType | `-${MessageSearchAuthorType}`)[];
```
almeidx: This enum includes the negated conditions now. 

```suggestion
  authorType?: MessageSearchAuthorType[];
```

packages/discord.js/typings/index.d.ts
```diff
@@ -6388,6 +6393,39 @@ export interface GuildSearchMembersOptions {
...
+  cache?: boolean;
+  channelId?: Snowflake[];
+  content?: string;
+  embedProvider?: string[];
+  embedType?: MessageSearchEmbedType[];
+  has?: (MessageSearchHasType | `-${MessageSearchHasType}`)[];
```
```suggestion
  has?: MessageSearchHasType[];
```

packages/discord.js/src/structures/Guild.js
```diff
@@ -1078,6 +1088,282 @@ class Guild extends AnonymousGuild {
...
+
+    for (const [key, values] of arrayParams) {
+      if (values) {
+        for (const value of values) {
+          query.append(key, value);
+        }
+      }
+    }
+
+    const data = await this.client.rest.get(Routes.guildMessagesSearch(this.id), { query });
```
almeidx: This is not handling the 202 Accepted response.
TBD if we want to handle retries automatically here or just throw

packages/discord.js/src/structures/Guild.js
```diff
@@ -1078,6 +1088,282 @@ class Guild extends AnonymousGuild {
...
+      pinned,
+      sort_by: sortBy,
+      sort_order: sortOrder,
+      include_nsfw: includeNsfw,
+      limit,
+      offset,
+      slop,
+    });
+
+    const arrayParams = [
```
almeidx: Personally, I'd rather modify the `makeURLSearchParams()` function to accept arrays for object values. Preferably in a separate pull request, which would have to be merged before this one.

packages/discord.js/src/structures/Guild.js
```diff
@@ -1078,6 +1088,282 @@ class Guild extends AnonymousGuild {
...
+   * Options used to search messages in a guild.
+   *
+   * @typedef {Object} GuildSearchMessagesOptions
+   * @property {string} [content] Filter messages by given content (max 1024 characters)
+   * @property {Snowflake} [maxId] Get messages before a given message ID
+   * @property {Snowflake} [minId] Get messages after a given message ID
+   * @property {Snowflake[]} [channelId] Filter messages by given channel IDs (max 500)
+   * @property {MessageSearchAuthorType[]} [authorType] Filter messages by author type.
+   * Valid values are from {@link MessageSearchAuthorType}.
+   * All types can be negated by prefixing them with `-`, which means results will not include messages that match the type.
```
almeidx: The former is inferable from the type. The latter is included in the enum now too
```suggestion
```

packages/discord.js/src/structures/Guild.js
```diff
@@ -1078,6 +1088,282 @@ class Guild extends AnonymousGuild {
...
+   * @property {Snowflake[]} [authorId] Filter messages by given author IDs (max 100)
+   * @property {Snowflake[]} [mentions] Filter messages that mention given user IDs (max 100)
+   * @property {Snowflake[]} [mentionsRoleId] Filter messages that mention given role IDs (max 100)
+   * @property {boolean} [mentionEveryone] Filter messages by whether they mention @everyone
+   * @property {Snowflake[]} [repliedToUserId] Filter messages that reply to given user IDs (max 100)
+   * @property {Snowflake[]} [repliedToMessageId] Filter messages that reply to given message IDs (max 100)
+   * @property {boolean} [pinned] Filter messages by whether they are pinned
+   * @property {MessageSearchHasType[]} [has] Filter messages by whether they contain specific content types.
+   * Valid values are from {@link MessageSearchHasType}.
+   * All types can be negated by prefixing them with `-`, which means results will not include messages that match the type.
```
almeidx:
```suggestion
```

packages/discord.js/src/structures/Guild.js
```diff
@@ -1078,6 +1088,282 @@ class Guild extends AnonymousGuild {
...
+   * @property {MessageSearchEmbedType[]} [embedType] Filter messages by embed type.
+   * Valid values are from {@link MessageSearchEmbedType}.
```
almeidx:
```suggestion
```

packages/discord.js/src/structures/Guild.js
```diff
@@ -1078,6 +1088,282 @@ class Guild extends AnonymousGuild {
...
+   * @property {string[]} [embedProvider] Filter messages by embed provider, case-sensitive (e.g. `'Tenor'`) (max 100)
+   * @property {string[]} [linkHostname] Filter messages by link hostname (e.g. `'discord.com'`) (max 100)
+   * @property {string[]} [attachmentFilename] Filter messages by attachment filename (max 100)
+   * @property {string[]} [attachmentExtension] Filter messages by attachment extension (e.g. `'png'`) (max 100)
+   * @property {MessageSearchSortMode} [sortBy] The sorting algorithm to use.
+   * Valid values are from {@link MessageSearchSortMode}.
```
almeidx:
```suggestion
```

packages/discord.js/src/structures/Guild.js
```diff
@@ -1078,6 +1088,282 @@ class Guild extends AnonymousGuild {
...
+   *
+   * @typedef {Object} GuildSearchMessagesResult
+   * @property {number} totalResults The total number of messages that match the query
+   * @property {Collection<Snowflake, Message>} messages A collection of the matched messages
+   */
+
+  /**
+   * Returns a collection of messages without the `reactions` property that match a search query in the guild. Requires the `READ_MESSAGE_HISTORY` permission.
+   * <warn>This is restricted according to whether the {@link GatewayIntentBits.MessageContent} Privileged Intent is enabled for your application.</warn>
+   * <Info>Due to speed optimizations, search may return slightly fewer results than the limit specified when messages have not been accessed for a long time.
```
almeidx:
```suggestion
   * <info>Due to speed optimizations, search may return slightly fewer results than the limit specified when messages have not been accessed for a long time.
```

packages/discord.js/src/structures/Guild.js
```diff
@@ -1078,6 +1088,282 @@ class Guild extends AnonymousGuild {
...
+   * Clients should not rely on the length of the `messages` array to paginate results.
+   *
+   * Additionally, when messages are actively being created or deleted, the `totalResults` property may not be accurate.</Info>
```
almeidx:
```suggestion
   * Additionally, when messages are actively being created or deleted, the `totalResults` property may not be accurate.</info>
```

almeidx (dismissed):

packages/discord.js/src/structures/Guild.js
```diff
@@ -1078,6 +1088,282 @@ class Guild extends AnonymousGuild {
...
+    attachmentFilename,
+    attachmentExtension,
+    sortBy,
+    sortOrder,
+    includeNsfw,
+    limit,
+    offset,
+    slop,
+  } = {}) {
+    if (content !== undefined && typeof content !== 'string') {
```
imnaiyar: Do we really need all these checks here? personally, I think this should just be left to the api to return the error, unless the error is too broad (i have not checked).

Perhaps we can even delegate most of these to a MessagsSearchParamBuilder, which I think makes sense to have for this.

Ig it depends on what others will say 🤷

packages/discord.js/src/structures/Guild.js
kshitijanurag: Removed all validation.

packages/discord.js/src/structures/Guild.js
kshitijanurag: Removed validation. Not sure about the builder though, will wait for input from other reviewers.

packages/discord.js/src/structures/Guild.js
kshitijanurag: I am fine with doing that, should I make a PR or wait for other reviewers to give their opinion on this?

packages/discord.js/typings/index.d.ts
```diff
@@ -6388,6 +6393,39 @@ export interface GuildSearchMembersOptions {
...
+  attachmentFilename?: readonly string[];
+  authorId?: readonly Snowflake[];
+  authorType?: readonly MessageSearchAuthorType[];
+  cache?: boolean;
+  channelId?: readonly Snowflake[];
+  content?: string;
+  embedProvider?: readonly string[];
+  embedType?: readonly MessageSearchEmbedType[];
+  has?: readonly MessageSearchHasType[];
+  includeNsfw?: boolean;
```
Jiralite, line 6466: `includeNSFW`? @almeidx @vladfrangu

almeidx: I brought up this topic here https://github.com/discordjs/discord.js/pull/11393#discussion_r2725534088 but didn't get a response. I also voiced my opinion there

vladfrangu: @discordjs/core thoughts? I lean more towards `includeNSFW`

Jiralite: `includeNsfw` wins.[^1]

[^1]: https://canary.discord.com/channels/222078108977594368/1209960895141388389/1488140130181976135

packages/discord.js/src/structures/Guild.js
```diff
@@ -1078,6 +1079,146 @@ class Guild extends AnonymousGuild {
...
+   * @property {Snowflake[]} [authorId] Filter messages by given author IDs (max 100)
+   * @property {Snowflake[]} [mentions] Filter messages that mention given user IDs (max 100)
+   * @property {Snowflake[]} [mentionsRoleId] Filter messages that mention given role IDs (max 100)
+   * @property {boolean} [mentionEveryone] Filter messages by whether they mention @everyone
+   * @property {Snowflake[]} [repliedToUserId] Filter messages that reply to given user IDs (max 100)
+   * @property {Snowflake[]} [repliedToMessageId] Filter messages that reply to given message IDs (max 100)
+   * @property {boolean} [pinned] Filter messages by whether they are pinned
+   * @property {MessageSearchHasType[]} [has] Filter messages by whether they contain specific content types.
+   * @property {MessageSearchEmbedType[]} [embedType] Filter messages by embed type.
+   * @property {string[]} [embedProvider] Filter messages by embed provider, case-sensitive (e.g. `'Tenor'`) (max 100)
```
Jiralite: Tenor is out of commission very, very soon. Might be worth dropping the remark of it.

packages/discord.js/src/structures/Guild.js
Jiralite: I opened #11578 as this would need backporting and fixing right now for core.

packages/discord.js/src/structures/Guild.js
```diff
@@ -1078,6 +1079,135 @@ class Guild extends AnonymousGuild {
...
+      offset,
+      slop,
+    });
+
+    const data = await this.client.rest.get(Routes.guildMessagesSearch(this.id), { query });
+
+    if (data.code === RESTJSONErrorCodes.IndexNotYetAvailable) {
+      const messages = data.messages.flat().reduce((collection, message) => {
+        const channel = this.client.channels.cache.get(message.channel_id);
+        return collection.set(message.id, channel?.messages._add(message, cache) ?? new Message(this.client, message));
```
Jiralite: Not sure `channel` can ever be `undefined`, but in the case that it is, `message` is missing the guild id when being constructed. Might as well pass that!

packages/discord.js/src/structures/Guild.js
```diff
@@ -1078,6 +1079,135 @@ class Guild extends AnonymousGuild {
...
+      sort_order: sortOrder,
+      include_nsfw: includeNsfw,
+      limit,
+      offset,
+      slop,
+    });
+
+    const data = await this.client.rest.get(Routes.guildMessagesSearch(this.id), { query });
+
+    if (data.code === RESTJSONErrorCodes.IndexNotYetAvailable) {
```
Jiralite: From the specification, a normal response has no `code`. This means they reach `return { totalResults: 0, messages: [] };`, no? So this does not even work?

packages/discord.js/src/structures/Guild.js
```diff
@@ -1078,6 +1079,135 @@ class Guild extends AnonymousGuild {
...
+      const messages = data.messages.flat().reduce((collection, message) => {
+        const channel = this.client.channels.cache.get(message.channel_id);
+        return collection.set(message.id, channel?.messages._add(message, cache) ?? new Message(this.client, message));
+      }, new Collection());
```
Jiralite: Missing support for threads.

Jiralite (dismissed):

packages/discord.js/src/structures/Guild.js
Jiralite: In addition to this, `retry_after` and `documents_indexed` are not in this pull request.

I believe we should handle retries automatically–it makes it easier for users. Otherwise, everyone would need to try and catch _all_ their message searches.

We can add `signal?: AbortSignal` to `GuildSearchMessagesOptions` to compensate for people who want to reject.

packages/discord.js/src/structures/Guild.js
```diff
@@ -1078,6 +1080,150 @@ class Guild extends AnonymousGuild {
...
+      include_nsfw: includeNsfw,
+      limit,
+      offset,
+      slop,
+    });
+
+    let data = await this.client.rest.get(Routes.guildMessagesSearch(this.id), { query, signal });
+
+    while (data.code === RESTJSONErrorCodes.IndexNotYetAvailable) {
+      await sleep((data.retry_after ?? 1) * 1_000, undefined, { signal });
```
almeidx: Documentation says we should still retry "after a short delay" if `retry_after` is `0`[^1]. 

[^1]: https://docs.discord.com/developers/resources/message#search-guild-messages:~:text=If%20the%20retry_after%20field%20is%200%2C%20you%20should%20retry%20the%20request%20after%20a%20short%20delay.

packages/discord.js/src/structures/Guild.js
```diff
@@ -1078,6 +1080,150 @@ class Guild extends AnonymousGuild {
...
+
+    const messages = data.messages.flat().reduce((collection, message) => {
+      const channel = this.client.channels.cache.get(message.channel_id);
+      return collection.set(
+        message.id,
+        channel?.messages._add(message, cache) ?? new Message(this.client, { ...message, guild_id: this.id }),
+      );
+    }, new Collection());
+
+    return { totalResults: data.total_results, messages };
```
almeidx: What about the rest of the fields this endpoint returns? Like `doing_deep_historical_index` and `documents_indexed`

packages/discord.js/src/structures/Guild.js
```diff
@@ -1078,6 +1080,150 @@ class Guild extends AnonymousGuild {
...
+      attachment_extension: attachmentExtension,
+      sort_by: sortBy,
+      sort_order: sortOrder,
+      include_nsfw: includeNsfw,
+      limit,
+      offset,
+      slop,
+    });
+
+    let data = await this.client.rest.get(Routes.guildMessagesSearch(this.id), { query, signal });
```
almeidx, line 1217: This also includes `members`, which seemingly are not being used for anything?

almeidx (dismissed):

packages/discord.js/src/structures/Guild.js
```diff
@@ -1078,6 +1080,176 @@ class Guild extends AnonymousGuild {
...
+      if (!threadMember) return collection;
+
+      const membersForThread = collection.ensure(member.id, () => new Collection());
+      membersForThread.set(member.user_id, threadMember);
+      return collection;
+    }, new Collection());
+
+    return {
+      doingDeepHistoricalIndex: data.doing_deep_historical_index,
+      documentsIndexed: data.documents_indexed ?? 0,
```
Qjuh: Why defaulting to 0 here instead of returning the actual value? This is optional in the API response so it should be an optional property in the return value too. Because 0 indexed messages is different from "we don't know how many".

Qjuh (changes requested):

packages/discord.js/src/structures/Guild.js
```diff
@@ -1078,6 +1080,176 @@ class Guild extends AnonymousGuild {
...
+   * @property {AbortSignal} [signal] An {@link AbortSignal} to cancel the request, including any automatic retries
+   * while the guild's messages are still being indexed
+   */
+
+  /**
+   * Result returned from searching guild messages.
+   *
+   * @typedef {Object} GuildSearchMessagesResult
+   * @property {boolean} doingDeepHistoricalIndex Whether the guild is undergoing a deep historical indexing operation
+   * @property {number} documentsIndexed The number of documents that have been indexed during the current index operation, if any
```
Qjuh: This doesn't match the implementation anymore. And the implementation returns the value `undefined` instead of it being optional (i.e. being omitted from the object if it doesn't exist on data).

packages/discord.js/src/structures/Guild.js
```diff
@@ -1078,6 +1080,176 @@ class Guild extends AnonymousGuild {
...
+      if (!threadMember) return collection;
+
+      const membersForThread = collection.ensure(member.id, () => new Collection());
+      membersForThread.set(member.user_id, threadMember);
+      return collection;
+    }, new Collection());
+
+    return {
+      doingDeepHistoricalIndex: data.doing_deep_historical_index,
+      documentsIndexed: data.documents_indexed,
```
Qjuh:
```suggestion
      ...('documents_indexed' in data ? {documentsIndexed: data.documents_indexed} : {}),
```

packages/discord.js/typings/index.d.ts
```diff
@@ -6446,6 +6451,44 @@ export interface GuildSearchMembersOptions {
...
+  repliedToMessageId?: readonly Snowflake[];
+  repliedToUserId?: readonly Snowflake[];
+  signal?: AbortSignal;
+  slop?: number;
+  sortBy?: MessageSearchSortMode;
+  sortOrder?: 'asc' | 'desc';
+}
+
+export interface GuildSearchMessagesResult {
+  documentsIndexed: number;
```
```suggestion
  documentsIndexed?: number;
```

Qjuh (changes requested):

packages/discord.js/src/structures/Guild.js
almeidx: This doesn't work if Discord sends the property as `0`. Use `in` instead

Qjuh (approved):

packages/discord.js/src/structures/Guild.js
```diff
@@ -1078,6 +1080,179 @@ class Guild extends AnonymousGuild {
...
+      data = await this.client.rest.get(Routes.guildMessagesSearch(this.id), { query, signal });
+    }
+
+    const threads = (data.threads ?? []).reduce(
+      (collection, thread) => collection.set(thread.id, this.client.channels._add(thread, this, { cache })),
+      new Collection(),
+    );
+
+    const messages = data.messages.flat().reduce((collection, message) => {
+      const channel = this.client.channels.cache.get(message.channel_id);
```
Jiralite:
```suggestion
      const channel = threads.get(message.channel_id) ?? this this.client.channels.cache.get(message.channel_id);
```

packages/discord.js/src/structures/Guild.js
```diff
@@ -1078,6 +1080,179 @@ class Guild extends AnonymousGuild {
...
+      return collection.set(
+        message.id,
+        channel?.messages._add(message, cache) ?? new Message(this.client, { ...message, guild_id: this.id }),
+      );
+    }, new Collection());
+
+    const threadMembers = (data.members ?? []).reduce((collection, member) => {
+      const thread = this.client.channels.cache.get(member.id);
```
Jiralite:
```suggestion
      const thread = threads.get(member.id) ?? this.client.channels.cache.get(member.id);
```

packages/discord.js/src/structures/Guild.js
```diff
@@ -1078,6 +1080,179 @@ class Guild extends AnonymousGuild {
...
+   * @property {boolean} [cache=true] Whether to cache the fetched messages
+   * @property {AbortSignal} [signal] An {@link AbortSignal} to cancel the request, including any automatic retries
+   * while the guild's messages are still being indexed
+   */
+
+  /**
+   * Result returned from searching guild messages.
+   *
+   * @typedef {Object} GuildSearchMessagesResult
+   * @property {number} [documentsIndexed] The number of documents that have been indexed during the current
```
Jiralite, line 1121:
```suggestion
   * @property {?number} [documentsIndexed] The number of documents that have been indexed during the current
```

packages/discord.js/typings/index.d.ts
```diff
@@ -6446,6 +6451,44 @@ export interface GuildSearchMembersOptions {
...
+  repliedToMessageId?: readonly Snowflake[];
+  repliedToUserId?: readonly Snowflake[];
+  signal?: AbortSignal;
+  slop?: number;
+  sortBy?: MessageSearchSortMode;
+  sortOrder?: 'asc' | 'desc';
+}
+
+export interface GuildSearchMessagesResult {
+  documentsIndexed?: number;
```
Jiralite, line 6484:
```suggestion
  documentsIndexed?: number | null;
```

packages/discord.js/typings/index.d.ts
```diff
@@ -6446,6 +6451,44 @@ export interface GuildSearchMembersOptions {
...
+  doingDeepHistoricalIndex: boolean;
+  messages: ReadonlyCollection<Snowflake, Message>;
```
Jiralite:
```suggestion
  messages: ReadonlyCollection<Snowflake, Message<true>>;
```

packages/discord.js/src/structures/Guild.js
```diff
@@ -1078,6 +1080,179 @@ class Guild extends AnonymousGuild {
...
+   * @param {GuildSearchMessagesOptions} [options={}] Options for searching messages
+   * @returns {Promise<GuildSearchMessagesResult>}
+   * @example
+   * // Search for messages containing 'Hello'
+   * guild.searchMessages({ content: 'Hello' })
+   *   .then(result => console.log(`Found ${result.totalResults} results.`))
+   *   .catch(console.error);
+   * @example
+   * // Search for messages with images, excluding bots, sorted by oldest first
+   * guild.searchMessages({ has: ['image'], authorType: ['-bot'], sortOrder: 'asc' })
```
Jiralite: This is invalid TypeScript–a generic string is not assignable to the enum. Use the enum instead.

Jiralite (dismissed):

packages/discord.js/src/structures/Guild.js
Qjuh: That property isn't nullable though, it's optional. So this change is wrong and should be reverted.

packages/discord.js/typings/index.d.ts
Qjuh: Same here

packages/discord.js/src/structures/Message.js
```diff
@@ -536,7 +536,11 @@ class Message extends Base {
    * @readonly
    */
   get channel() {
-    return this.client.channels.resolve(this.channelId);
+    return (
+      this.client.channels.cache.get(this.channelId) ??
+      this.client.guilds.cache.get(this.guildId)?.channels.cache.get(this.channelId) ??
+      null
+    );
```
Qjuh, line 543: Why did you change this? resolve() should already do what's needed, especially since the Guild#channels#cache check is redundant here

Qjuh (changes requested):
The squash of all commits and force-pushing means more work for reviewers because the "new" commit can't be reviewed separately so we need to look over everything again. Since merging of a PR is already done as "Squash & Merge" it's also not neccessary to squash at all.

kshitijanurag:
> The squash of all commits and force-pushing means more work for reviewers because the "new" commit can't be reviewed separately so we need to look over everything again. Since merging of a PR is already done as "Squash & Merge" it's also not neccessary to squash at all.

Alright, I will not do it moving forward.

Qjuh (approved):

packages/discord.js/typings/index.d.ts
```diff
@@ -6445,6 +6450,44 @@ export interface GuildSearchMembersOptions {
...
+  sortBy?: MessageSearchSortMode;
+  sortOrder?: 'asc' | 'desc';
+}
+
+export interface GuildSearchMessagesResult {
+  documentsIndexed?: number;
+  doingDeepHistoricalIndex: boolean;
+  messages: ReadonlyCollection<Snowflake, Message<true>>;
+  threadMembers: ReadonlyCollection<Snowflake, ReadonlyCollection<Snowflake, ThreadMember>>;
+  threads: ReadonlyCollection<Snowflake, ThreadChannel>;
```
almeidx, line 6487:
```suggestion
  threads: ReadonlyCollection<Snowflake, AnyThreadChannel>;
```

packages/discord.js/src/structures/Guild.js
```diff
@@ -1078,6 +1080,182 @@ class Guild extends AnonymousGuild {
...
+    });
+
+    let data = await this.client.rest.get(Routes.guildMessagesSearch(this.id), { query, signal });
+
+    while (data.code === RESTJSONErrorCodes.IndexNotYetAvailable) {
+      const retryAfter = data.retry_after || 1;
+
+      await sleep(retryAfter * 1_000, undefined, { signal });
+      data = await this.client.rest.get(Routes.guildMessagesSearch(this.id), { query, signal });
+    }
```
vladfrangu, line 1224: Users should be able to opt out of waiting if the index is missing

vladfrangu (dismissed):
small thing but see if others disagree

packages/discord.js/src/structures/Guild.js
```diff
@@ -1078,6 +1080,182 @@ class Guild extends AnonymousGuild {
...
+   * @property {Snowflake} [maxId] Get messages before a given message ID
+   * @property {Snowflake} [minId] Get messages after a given message ID
+   * @property {Snowflake[]} [channelId] Filter messages by given channel IDs (max 500)
+   * @property {MessageSearchAuthorType[]} [authorType] Filter messages by author type.
+   * @property {Snowflake[]} [authorId] Filter messages by given author IDs (max 100)
+   * @property {Snowflake[]} [mentions] Filter messages that mention given user IDs (max 100)
+   * @property {Snowflake[]} [mentionsRoleId] Filter messages that mention given role IDs (max 100)
+   * @property {boolean} [mentionEveryone] Filter messages by whether they mention @everyone
+   * @property {Snowflake[]} [repliedToUserId] Filter messages that reply to given user IDs (max 100)
+   * @property {Snowflake[]} [repliedToMessageId] Filter messages that reply to given message IDs (max 100)
```
almeidx, line 1097: I think these should support resolvables instead of just Snowflakes. i.e., ChannelResolvable, UserResolvable, etc.

packages/discord.js/src/structures/Guild.js
```diff
@@ -1078,6 +1080,186 @@ class Guild extends AnonymousGuild {
...
+      include_nsfw: includeNsfw,
+      limit,
+      offset,
+      slop,
+    });
+
+    let data = await this.client.rest.get(Routes.guildMessagesSearch(this.id), { query, signal });
+
+    if (retryOnMissingIndex) {
+      while (data.code === RESTJSONErrorCodes.IndexNotYetAvailable) {
```
almeidx: this isn't right. if you passed `retryOnMissingIndex: false` and received IndexNotYetAvailable it would go through the success path below, which would throw on `data.messages.flat()`. 

i think we need to make a new error code and throw earlier for this case instead

packages/discord.js/src/structures/Guild.js
```diff
@@ -1078,6 +1080,186 @@ class Guild extends AnonymousGuild {
...
+    offset,
+    slop,
+  } = {}) {
+    const query = makeURLSearchParams({
+      content,
+      max_id: maxId,
+      min_id: minId,
+      channel_id: channelId?.map(channel => this.channels.resolveId(channel)),
+      author_type: authorType,
+      author_id: authorId,
```
almeidx: missing resolve

almeidx (changes requested):

PR #11460, opened 2026-03-20
fix: clear existing idle timeout in Collector.resetTimer

Sim-hu:
## Summary
- Fix `Collector.resetTimer({ idle: newValue })` not properly resetting the idle timer
- The old code only reset the idle timeout when `this._idletimeout` already existed, and kept the clear and set coupled inside the same conditional
- Now the existing idle timeout is always cleared first, then a new one is created if an idle value is provided (either via the argument or from the original options)

## Test plan
- [ ] Call `resetTimer({ idle: 5000 })` on a collector that was created with a different idle value and verify the new idle duration takes effect
- [ ] Call `resetTimer({ idle: 5000 })` on a collector that was created without an idle option and verify an idle timeout is now active
- [ ] Call `resetTimer()` with no arguments on a collector with an existing idle timeout and verify it resets to the original idle value

### DIFF
--- a/packages/discord.js/src/structures/interfaces/Collector.js
+++ b/packages/discord.js/src/structures/interfaces/Collector.js
@@ -279,10 +279,12 @@ class Collector extends AsyncEventEmitter {
 
     if (this._idletimeout) {
       clearTimeout(this._idletimeout);
+      this._idletimeout = null;
     }
 
-    if (idle ?? this.options.idle) {
-      this._idletimeout = setTimeout(() => this.stop('idle'), idle ?? this.options.idle).unref();
+    const idleValue = idle ?? this.options.idle;
+    if (idleValue) {
+      this._idletimeout = setTimeout(() => this.stop('idle'), idleValue).unref();
     }
   }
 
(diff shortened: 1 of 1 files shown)


packages/discord.js/src/structures/interfaces/Collector.js
```diff
@@ -279,6 +279,9 @@ class Collector extends AsyncEventEmitter {
 
     if (this._idletimeout) {
       clearTimeout(this._idletimeout);
+    }
+
+    if (idle ?? this.options.idle) {
       this._idletimeout = setTimeout(() => this.stop('idle'), idle ?? this.options.idle).unref();
     }
```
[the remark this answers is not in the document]
Sim-hu: Good catch — added the null assignment after clearTimeout and pulled the idle value into a local variable to avoid the double evaluation.

packages/discord.js/src/structures/interfaces/Collector.js
Syjalo: I think you can make the same here and reuse `resetTimer` where we reset them

Sim-hu: Mirrored the null/conditional pattern onto the time timeout. Left the per-collect reset idle-only — resetTimer also resets the overall `time` limit, which shouldn't restart on each collect.

PR #11461, opened 2026-03-20
fix(rest): prevent AbortSignal event listener leak

Sim-hu:
## Summary

- Add `{ once: true }` to the `abort` event listener on the user-provided `AbortSignal` in `makeNetworkRequest` (`packages/rest/src/lib/handlers/Shared.ts`)
- Without this, every request adds a new listener that is never removed. When users reuse the same `AbortSignal` across multiple requests, listeners accumulate, each capturing a reference to the local `AbortController` and preventing it from being garbage collected
- The `once` option ensures the listener is automatically removed after firing and allows the captured `controller` to be garbage collected when it is no longer needed

## Test plan

- Verify the `@discordjs/rest` package compiles without type errors (`tsc --noEmit` passes)
- Confirm that aborting a request via a user-provided signal still correctly aborts the internal controller
- Confirm that after a request completes, the listener is no longer present on the user signal (can be verified by checking `getEventListeners` count in Node.js or by observing memory profiles under repeated requests with a shared signal)

### DIFF
--- a/packages/rest/src/lib/handlers/Shared.ts
+++ b/packages/rest/src/lib/handlers/Shared.ts
@@ -76,7 +76,7 @@ export async function makeNetworkRequest(
 		// The reason why we don't re-use the user's signal, is because users may use the same signal for multiple
 		// requests, and we do not want to cause unexpected side-effects.
 		if (requestData.signal.aborted) controller.abort();
-		else requestData.signal.addEventListener('abort', () => controller.abort());
+		else requestData.signal.addEventListener('abort', () => controller.abort(), { once: true });
 	}
 
 	let res: ResponseLike;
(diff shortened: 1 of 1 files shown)


packages/rest/src/lib/handlers/Shared.ts
```diff
@@ -76,7 +76,7 @@ export async function makeNetworkRequest(
 		// The reason why we don't re-use the user's signal, is because users may use the same signal for multiple
 		// requests, and we do not want to cause unexpected side-effects.
 		if (requestData.signal.aborted) controller.abort();
-		else requestData.signal.addEventListener('abort', () => controller.abort());
+		else requestData.signal.addEventListener('abort', () => controller.abort(), { once: true });
```
almeidx: This doesn't remove the listener if the signal is not aborted.
We'd need to extract the listener to a function and remove it [potentially] on the `finally` block below instead.

almeidx (changes requested):

packages/rest/src/lib/handlers/Shared.ts
Sim-hu: Extracted the listener and remove it in `finally`, dropped `{ once: true }`.

ckohen (changes requested):
Lets not create the function on every request if it's not necessary.

PR #11462, opened 2026-03-20
feat: allow partial DMChannel without client user

Qjuh:
Backport of #11443

To not be breaking `DMChannel#recipientId` can still return the wrong id. But `UserManager#dmChannel` and all methods relying on it won't return those wrong DMChannels anymore.

### DIFF
--- a/packages/builders/CHANGELOG.md
+++ b/packages/builders/CHANGELOG.md
@@ -2,6 +2,16 @@
 
 All notable changes to this project will be documented in this file.
 
+# [@discordjs/builders@1.14.0](https://github.com/discordjs/discord.js/compare/@discordjs/builders@1.13.1...@discordjs/builders@1.14.0) - (2026-03-22)
+
+## Documentation
+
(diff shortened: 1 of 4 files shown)


packages/discord.js/src/managers/UserManager.js
```diff
@@ -42,8 +42,11 @@ class UserManager extends CachedManager {
    */
   dmChannel(userId) {
     return (
-      this.client.channels.cache.find(channel => channel.type === ChannelType.DM && channel.recipientId === userId) ??
-      null
+      this.client.channels.cache.find(
+        channel =>
+          channel.type === ChannelType.DM &&
+          channel.recipientIds.every(id => [userId, this.client.user.id].includes(id)),
```
vladfrangu: Move the `[userId, this.client.user.id]` outside the callback

packages/discord.js/src/structures/DMChannel.js
```diff
@@ -71,6 +73,16 @@ class DMChannel extends BaseChannel {
     return this.lastMessageId === undefined;
   }
 
+  /**
+   * The recipient's id.
```
vladfrangu, line 77: Lets add a callout that this might be wrong and will return null in the next major

PR #11423, opened 2026-02-16
fix(MessagePayload): allow AttachmentBuilder in files payload

Qjuh:
Fixes a regression of #11278 which causes issues when passing AttachmentBuilder in `MessagePayload#options#files` in any send method

PR #11474, opened 2026-04-03
fix(discord.js): resolve MessagePayload crash on edit

GreenHolland:
`MessagePayload.resolveBody()` crashes with `TypeError: Cannot read properties of undefined (reading 'push')` when `options.attachments` is provided without `options.files`, because `files?.map()` returns `undefined` and `.push()` is then called on it.

This affects any `message.edit({ attachments: [...] })` call that keeps existing attachments without uploading new files.

Fix: default the `files?.map()` result to `[]` with `?? []`.

### DIFF
--- a/packages/discord.js/src/structures/MessagePayload.js
+++ b/packages/discord.js/src/structures/MessagePayload.js
@@ -190,13 +190,14 @@ class MessagePayload {
       }
     }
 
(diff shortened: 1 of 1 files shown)


didinele (approved):

almeidx:
This is not the correct fix. It would remove attachments on every message update unintentionally.
The appropriate fix is included in #11423

PR #11475, opened 2026-04-03
Fix/invite handlers null guild

GreenHolland:
`INVITE_CREATE` and `INVITE_DELETE` gateway handlers crash with `TypeError: Cannot read properties of undefined (reading 'invites')` when the guild is not in cache.

Other handlers like `GUILD_BAN_ADD` correctly guard with `if (!guild) return;`.

### DIFF
--- a/packages/discord.js/src/structures/MessagePayload.js
+++ b/packages/discord.js/src/structures/MessagePayload.js
@@ -190,13 +190,14 @@ class MessagePayload {
       }
     }
 
-    const attachments = this.options.files?.map((file, index) => ({
-      id: index.toString(),
-      description: file.description,
-      title: file.title,
-      waveform: file.waveform,
-      duration_secs: file.duration,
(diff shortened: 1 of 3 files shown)


packages/discord.js/src/structures/MessagePayload.js
almeidx, line 1: The changes to this file should be reverted

almeidx (changes requested):

Qjuh:
Did you actually test this and received that error? Or are you assuming it would? Because there should be no way to have a guild channel in cache when the Guild itself isn't cached. So the `if (!channel) return false;` should already assure the Guild is cached.

GreenHolland:
No I am sorry I should have, I didn't test it. I started reading the repo and tried to find a few things that need fixing or help out, as I started using it. Will make sure to create a dummy app and test it in the future.

PR #11476, opened 2026-04-03
fix(discord.js): add missing GuildUncachedMe checks

GreenHolland:
Several getters access `guild.members.me` or `guild.members.resolve()` without guarding against null, causing `TypeError` when the bot's own member is not cached (e.g. without GuildMembers intent).

Add `GuildUncachedMe` checks consistent with the existing pattern in `GuildMember.manageable`, `GuildEmoji.deletable`, and `GuildInvite.deletable`.

Affected getters:
- `Role.editable`
- `GuildChannel.manageable`
- `Message.deletable`
- `VoiceChannel.speakable`
- `BaseGuildVoiceChannel.joinable`
- `ThreadChannel.manageable`
- `ThreadChannel.sendable`

### DIFF
--- a/packages/discord.js/src/structures/BaseGuildVoiceChannel.js
+++ b/packages/discord.js/src/structures/BaseGuildVoiceChannel.js
@@ -2,6 +2,7 @@
 
 const { Collection } = require('@discordjs/collection');
 const { PermissionFlagsBits } = require('discord-api-types/v10');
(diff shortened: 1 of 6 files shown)


Jiralite (changes requested):
I do not think this is correct. Checking for permissions already checks for a member being cached. Did you run into an issue?

PR #11479, opened 2026-04-03
fix(Action): don't add recipients to guild channels

Qjuh:
Fixes GuildChannels from another shard being cached as DMChannel on shard 0 when sending ephemeral interaaction replies with Channel partial enabled.

What happened before this fix:
- Interaction with an ephemeral reply is sent in a guild channel on shard != 0
- messageCreate emits on shard 0 because of the ephemeral message
- the Channel partial causes the guild channel to be cached as DMChannel because data.guild_id is null in the messageCreate and recipients gets added by the getChannel() call (and the channel doesn't belong to shard 0, so isn't in cache yet)

### DIFF
--- a/packages/discord.js/src/client/actions/Action.js
+++ b/packages/discord.js/src/client/actions/Action.js
@@ -40,7 +40,7 @@ class GenericAction {
       if (!data.recipients.some(existingRecipient => recipient.id === existingRecipient.id)) {
         payloadData.recipients = [...data.recipients, recipient];
       }
-    } else if ([ChannelType.DM, ChannelType.GroupDM].includes(data.type)) {
+    } else if (data.type === ChannelType.DM || data.type === ChannelType.GroupDM) {
       // Try to resolve the recipient.
       const recipient = data.author ?? data.user ?? { id: data.user_id };
       payloadData.recipients = [recipient];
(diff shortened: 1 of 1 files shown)


SpaceEEC (approved):

packages/discord.js/src/client/actions/Action.js
```diff
@@ -39,7 +40,7 @@ class GenericAction {
       if (!data.recipients.some(existingRecipient => recipient.id === existingRecipient.id)) {
         payloadData.recipients = [...data.recipients, recipient];
       }
-    } else {
+    } else if ([ChannelType.DM, ChannelType.GroupDM].includes(data.type)) {
```
vladfrangu: Do what jira said

vladfrangu (changes requested):

packages/discord.js/src/client/actions/Action.js
Jiralite:
```suggestion
    } else if (data.type === ChannelType.DM || data.type === ChannelType.GroupDM) {
```

vladfrangu (approved):

Jiralite (approved):

PR #11480, opened 2026-04-04
feat: more attachment and embed fields

kshitijanurag:
Add more attachment and embeds fields

Discord API Docs reference:
- https://github.com/discord/discord-api-docs/pull/7353

Depends on:
- https://github.com/discordjs/discord-api-types/pull/1609


packages/discord.js/src/structures/Attachment.js
```diff
@@ -161,6 +161,61 @@ class Attachment {
...
+      this.placeholderVersion ??= null;
+    }
+
+    if ('clip_participants' in data) {
+      /**
+       * For clips, the users who were in the stream
+       *
+       * @type {?Array<APIUser>}
+       */
+      this.clipParticipants = data.clip_participants;
```
Qjuh: This should use `UserManager#_add` instead to add/patch those users into the cache, not raw API data. So I assume Attachment should just store the clipParticipantIds and `clipParticipants` would be a getter, not a property.

packages/discord.js/src/structures/Attachment.js
```diff
@@ -161,6 +161,61 @@ class Attachment {
...
+      this.clipParticipants ??= null;
+    }
+
+    if ('clip_created_at' in data) {
+      /**
+       * For clips, when the clip was created
+       *
+       * @type {?Date}
+       */
+      this.clipCreatedAt = data.clip_created_at ? new Date(data.clip_created_at) : null;
```
Qjuh: If it's present then it should be a date. The field is not documented as being nullable, so the ternary here is weird.

packages/discord.js/src/structures/Attachment.js
```diff
@@ -161,6 +161,61 @@ class Attachment {
...
+      this.clipCreatedAt ??= null;
+    }
+
+    if ('application' in data) {
+      /**
+       * For clips, the application in the stream, if recognized
+       *
+       * @type {?APIApplication}
+       */
+      this.application = data.application;
```
Qjuh: Also shouldn't be raw API data.

packages/structures/src/messages/components/UnfurledMediaItem.ts
```diff
@@ -41,12 +41,35 @@ export class UnfurledMediaItem<Omitted extends keyof APIUnfurledMediaItem | '' =
...
+	 * The height of the media item (if image or video)
 	 */
 	public get height() {
 		return this[kData].height;
 	}
 
+	/**
+	 * ThumbHash placeholder (if image or video)
+	 */
+	public get placeholder() {
```
Qjuh, line 62: Should probably add these as getters to mainlib `packages/discord.js/src/structures/UnfurledMediaItem.js` too.

packages/structures/src/messages/embeds/EmbedThumbnail.ts
```diff
@@ -1,49 +1,85 @@
...
 /**
  * Represents thumbnail data in an embed on a message.
  *
  * @typeParam Omitted - Specify the properties that will not be stored in the raw data field as a union, implement via `DataTemplate`
  */
-export class EmbedThumbnail<Omitted extends keyof APIEmbedThumbnail | '' = ''> extends Structure<
-	APIEmbedThumbnail,
-	Omitted
-> {
+export class EmbedThumbnail<Omitted extends keyof APIEmbedImage | '' = ''> extends Structure<APIEmbedImage, Omitted> {
```
Qjuh: This whole class is obsolete now and should be deleted. EmbedImage already exists.

packages/structures/src/messages/Attachment.ts
```diff
@@ -117,4 +152,25 @@ export class Attachment<Omitted extends keyof APIAttachment | '' = ''> extends S
 			? new AttachmentFlagsBitField(this[kData].flags as AttachmentFlags)
 			: null;
 	}
+
+	/**
+	 * For Clips, array of users who were in the stream
+	 */
+	public get clipParticipants() {
```
Qjuh: These are substructures and shouldn't have a getter. Instead add a remark on the JSDoc for the class. Or rather add a JSDoc to the class at all, guess that was missed when writing this class.

packages/structures/src/messages/Attachment.ts
```diff
@@ -117,4 +152,25 @@ export class Attachment<Omitted extends keyof APIAttachment | '' = ''> extends S
...
+	 * For Clips, when the clip was created
+	 */
+	public get clipCreatedAt() {
+		return this[kClipCreatedAt];
+	}
+
+	/**
+	 * For Clips, the application in the stream, if recognized
+	 */
+	public get application() {
```
Qjuh: Also a substructure, no getter, add a remark to class' JSDoc.

Qjuh (changes requested):

packages/structures/src/messages/Attachment.ts
```diff
@@ -1,21 +1,42 @@
 import type { APIAttachment, AttachmentFlags } from 'discord-api-types/v10';
 import { Structure } from '../Structure.js';
 import { AttachmentFlagsBitField } from '../bitfields/AttachmentFlagsBitField.js';
-import { kData } from '../utils/symbols.js';
+import { kClipCreatedAt, kData } from '../utils/symbols.js';
 import { isFieldSet } from '../utils/type-guards.js';
 import type { Partialize } from '../utils/types.js';
 
 export class Attachment<Omitted extends keyof APIAttachment | '' = ''> extends Structure<APIAttachment, Omitted> {
```
almeidx:
```suggestion
export class Attachment<Omitted extends keyof APIAttachment | '' = 'clip_created_at'> extends Structure<APIAttachment, Omitted> {
```

packages/structures/src/messages/Attachment.ts
```diff
@@ -1,21 +1,42 @@
...
 	/**
 	 * The template used for removing data from the raw data stored for each Attachment.
 	 */
-	public static override readonly DataTemplate: Partial<APIAttachment> = {};
+	public static override readonly DataTemplate: Partial<APIAttachment> = {
+		set clip_created_at(_: string) {},
```
almeidx, line 24: Since we're optimizing this prop out, we need to override the toJSON method to add it back

packages/discord.js/typings/index.d.ts
```diff
@@ -2288,6 +2307,9 @@ export class AuthorizingIntegrationOwners extends Base {
 export class Attachment {
   private constructor(data: APIAttachment);
   private readonly attachment: BufferResolvable | Stream;
+  public application: APIApplication | null;
+  public clipCreatedAt: number | null;
```
```suggestion
  public clipCreatedAt: Date | null;
```

packages/discord.js/typings/index.d.ts
```diff
@@ -831,6 +849,7 @@ export class Embed {
   public get provider(): APIEmbedProvider | null;
   public get video(): EmbedAssetData | null;
   public get length(): number;
+  public get flags(): EmbedFlagsBitField | null;
```
almeidx: Never returns null
```suggestion
  public get flags(): EmbedFlagsBitField;
```

packages/discord.js/typings/index.d.ts
```diff
@@ -808,7 +821,12 @@ export interface EmbedFooterData extends IconData {
 }
 
 export interface EmbedAssetData {
+  contentType?: string;
+  description?: string;
+  flags?: Readonly<EmbedMediaFlagsBitField>;
```
almeidx: Always returned
```suggestion
  flags: Readonly<EmbedMediaFlagsBitField>;
```

almeidx (changes requested):

packages/discord.js/src/structures/Attachment.js
```diff
@@ -161,6 +166,74 @@ class Attachment {
...
+      this.clipCreatedAt ??= null;
+    }
+
+    if ('application' in data) {
+      /**
+       * For clips, the application in the stream, if recognized
+       *
+       * @type {?ClientApplication}
+       */
+      this.application = new ClientApplication(this.client, data.application);
```
Qjuh: This should not be a `ClientApplication` because that's only for the application instance of the logged in Client. But this application could be from any arbitrary application, similar to `IntegrationApplication` it should be another class extending `Application`.

packages/discord.js/src/structures/Attachment.js
```diff
@@ -161,6 +166,74 @@ class Attachment {
...
+      this.placeholderVersion ??= null;
+    }
+
+    if ('clip_participants' in data) {
+      /**
+       * For clips, the ids of the users who were in the stream
+       *
+       * @type {?Snowflake[]}
+       */
+      this.clipParticipantIds = data.clip_participants.map(user => user.id);
```
Qjuh: This won't add/patch the user data in cache though, missing `UserManager#_add` call here.

packages/discord.js/src/structures/Attachment.js
```diff
@@ -161,6 +166,74 @@ class Attachment {
...
+
+  /**
+   * For clips, the users who were in the stream
+   *
+   * @type {?Collection<Snowflake, User>}
+   * @readonly
+   */
+  get clipParticipants() {
+    return this.clipParticipantIds?.reduce((collection, participantId) => {
+      const user = this.client.users._add({ id: participantId });
```
Qjuh: This is only adding the id to the cache, the full user data is lost at this point. The user should be gotten from cache here, not added.

packages/discord.js/src/structures/UnfurledMediaItem.js
```diff
@@ -13,6 +15,36 @@ class UnfurledMediaItem {
...
+  /**
+   * Version of the placeholder (if image or video)
+   *
+   * @type {?number}
+   * @readonly
+   */
+  get placeholderVersion() {
+    return this.data.placeholder_version ?? null;
+  }
+
```
Qjuh, line 79: still missing `proxyURL`, `height`, `width`, `contentType` and `attachmentId` getters.

Qjuh (changes requested):

packages/discord.js/src/structures/UnfurledMediaItem.js
kshitijanurag: Added the remaining getters.

packages/discord.js/src/structures/Attachment.js
kshitijanurag: How about a new class named `ClipApplication`?

Qjuh (approved):

PR #11491, opened 2026-04-13
ci(deprecateVersion): Fix empty `message`

Jiralite:
Since moving to packages/actions, calls always pass `message`:

https://github.com/discordjs/discord.js/blob/a39fb54c78fb4617c12b334216fa6245a5f12f83/.github/workflows/deprecate-version.yml#L50-L56

https://github.com/discordjs/discord.js/blob/a39fb54c78fb4617c12b334216fa6245a5f12f83/.github/workflows/remove-tag.yml#L54-L60

This means the default message will not be used as a value is present. Moved the handling of the default to the command instead.

Also moved other variables to `env`.

### DIFF
--- a/packages/actions/src/deprecateVersion/action.yml
+++ b/packages/actions/src/deprecateVersion/action.yml
@@ -10,7 +10,7 @@ inputs:
   message:
     description: Deprecation message
     required: false
-    default: This version is deprecated. Please use a newer version.
+    # Default message lives in the command to handle empty strings.
   node-auth-token:
     description: npm authentication token
     required: true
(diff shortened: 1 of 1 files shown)


packages/actions/src/deprecateVersion/action.yml
```diff
@@ -19,7 +19,8 @@ runs:
   steps:
     - name: Deprecate version
       shell: bash
-      run: |
-        pnpm exec npm-deprecate --name "${{ inputs.version }}" --message "${{ inputs.message }}" --package "${{ inputs.package }}"
       env:
+        MESSAGE: ${{ inputs.message }}
```
almeidx: Can we move the default here instead?

almeidx (dismissed):

almeidx (approved):

vladfrangu (approved):

Qjuh (approved):

PR #10462, opened 2024-08-27
feat: send voice messages

almeidx:
**Please describe the changes this PR makes and why it should be merged:**

Adds the ability to send voice messages
Resolves #10298

Depends on:
- https://github.com/discord/discord-api-docs/pull/7267

**Status and versioning classification:**
- Code changes have been tested against the Discord API, or there are no code changes
- I know how to update typings and have done so, or typings don't need updating

PR #11493, opened 2026-04-16
feat: send voice messages

ducktrshessami:
Backports #10462

This is fully just a cherry pick, so I figured it was fine to ignore, but just a heads up that there were some linting errors on some of the docs comments for `waveform`

```
.../discord.js/packages/discord.js/src/structures/Attachment.js
  12:1  error  This line has a length of 127. Maximum allowed is 120  max-len

.../discord.js/packages/discord.js/src/structures/AttachmentBuilder.js
  172:1  error  This line has a length of 126. Maximum allowed is 120  max-len
```

### DIFF
--- a/packages/discord.js/src/structures/AttachmentBuilder.js
+++ b/packages/discord.js/src/structures/AttachmentBuilder.js
@@ -169,6 +169,6 @@ module.exports = AttachmentBuilder;
  * @property {string} [name] The name of the attachment
  * @property {string} [description] The description of the attachment
  * @property {string} [title] The title of the attachment
- * @property {string} [waveform] The base64 encoded byte array representing a sampled waveform (for voice message attachments)
- * @property {number} [duration] The duration of the attachment in seconds (for voice message attachments)
+ * @property {string} [waveform] The base64 encoded bytearray representing a sampled waveform (for voice messages)
+ * @property {number} [duration] The duration of the attachment in seconds (for voice messages)
  */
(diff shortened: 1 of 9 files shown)


almeidx:
> there were some linting errors on some of the docs comments for `waveform`

we should fix them for this branch, otherwise CI will fail (as it is now in this pull request)

packages/discord.js/src/structures/AttachmentBuilder.js
```diff
@@ -16,16 +16,41 @@ class AttachmentBuilder {
      * @type {BufferResolvable|Stream}
      */
     this.attachment = attachment;
+
     /**
      * The name of this attachment
      * @type {?string}
```
Qjuh, line 22: This has been wrong ever since. `?string` means nullable, but if `name` isn't set in `data` then it'll be `undefined`, not `null`. Same applies to `description` and the new fields. Both here and in the typings.
Either need an additional `?? null` to match the documented type or fix the type to match the implementation.

Qjuh (changes requested):

packages/discord.js/src/structures/AttachmentBuilder.js
ducktrshessami: done the former, lmk if the latter is preferred

Jiralite (dismissed):

Jiralite (approved):

PR #11495, opened 2026-04-18
fix(MessageCreateAction): receive DMs in uncached DMChannels again

### LINKED ISSUE #11486: Changes to getChannel method in 14.26.2 leading to Bots not reacting to DMs anymore

Qjuh:
While fixing #11429 with the combination of #11462 and #11479 I accidentally broke receiving of messages sent to the bot in uncached DMChannels, because they aren't recognized as such anymore.

Fixes #11486
Depends on
- https://github.com/discord/discord-api-docs/pull/8311
- https://github.com/discordjs/discord-api-types/pull/1638

Needs further testing if there still is a fixable issue for reactions being added/removed on a message in such an uncached DMChannel.

### DIFF
--- a/packages/discord.js/src/client/actions/MessageCreate.js
+++ b/packages/discord.js/src/client/actions/MessageCreate.js
@@ -10,6 +10,7 @@ class MessageCreateAction extends Action {
       id: data.channel_id,
       author: data.author,
       ...('guild_id' in data && { guild_id: data.guild_id }),
+      ...('channel_type' in data && { type: data.channel_type }),
     });
     if (channel) {
       if (!channel.isTextBased()) return {};


almeidx (changes requested):
Waiting on staff response on whether we can use this field

Jared-Is-Coding:
Just dropping by to say that I'm also noticing/running into this regression on 14.26.3. @Qjuh 

As a temporary fix I hooked a handler onto the raw gateway event handler that fetches the channel and message and re-sends the event to my messageCreate handler. Will keep an eye here to see when this is fixed.

almeidx (approved):

vladfrangu (approved):

PR #11500, opened 2026-04-19
feat(Message): add threadable getter

### LINKED ISSUE #11264: Message/channel threadable

discord.js

### Feature

I really like djs for its compact features such as; to check if a message is deletable or if a role is editable but I cant find anything similar for threads creation. To check if if the client user can create a thread in this channel or message. I dont think this exists yet or maybe there is another easy way to check if a message is threadable that i dont know of. I now check if i have the needed permission flags for the channel and if not exceeded the active thread limit size

message.threadable and channel.threadable

_No response_

### Other context

_No response_

Dramex:
## Summary

- Adds `Message#threadable`, mirroring the existing `deletable` / `editable` / `crosspostable` / `pinnable` getters. It returns whether the client user can start a thread from the message (parent channel type supports threads, no pre-existing thread, channel viewable, `CreatePublicThreads` permission).
- Updates the TypeScript typings accordingly.

## Why

Bots frequently filter messages to decide whether to offer a "start thread" affordance, and the library already ships the same sort of `-able` helpers for deleting, editing, and pinning. This closes the gap reported in the linked issue without introducing new intents or requests, since every check is based on already-cached state.

The channel-level variant that was in an earlier push of this PR was dropped per review: `TextChannel` supports both public and private threads behind separate permissions (`CreatePublicThreads` / `CreatePrivateThreads`), so a single boolean would have been misleading.

## Test plan

- [x] `pnpm --filter=discord.js test`
- [x] `pnpm --filter=discord.js lint`

Closes #11264

### DIFF
--- a/packages/discord.js/src/structures/BaseGuildTextChannel.js
+++ b/packages/discord.js/src/structures/BaseGuildTextChannel.js
@@ -1,6 +1,5 @@
 'use strict';
 
-const { PermissionFlagsBits } = require('discord-api-types/v10');
 const { GuildMessageManager } = require('../managers/GuildMessageManager.js');
 const { GuildTextThreadManager } = require('../managers/GuildTextThreadManager.js');
 const { GuildChannel } = require('./GuildChannel.js');
@@ -97,18 +96,6 @@ class BaseGuildTextChannel extends GuildChannel {
     if ('messages' in data) {
       for (const message of data.messages) this.messages._add(message);
     }
-  }
-
-  /**
-   * Whether the client user can start a new thread in this channel.
-   *
-   * @type {boolean}
-   * @readonly
-   */
-  get threadable() {
-    if (!this.viewable) return false;
-    const permissions = this.permissionsFor(this.client.user);
-    return permissions?.has(PermissionFlagsBits.CreatePublicThreads, false) ?? false;
   }
 
   /**
--- a/packages/discord.js/src/structures/BaseGuildTextChannel.js
+++ b/packages/discord.js/src/structures/BaseGuildTextChannel.js
@@ -1,5 +1,6 @@
 'use strict';
 
+const { PermissionFlagsBits } = require('discord-api-types/v10');
 const { GuildMessageManager } = require('../managers/GuildMessageManager.js');
 const { GuildTextThreadManager } = require('../managers/GuildTextThreadManager.js');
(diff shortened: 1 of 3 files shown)


packages/discord.js/src/structures/BaseGuildTextChannel.js
```diff
@@ -98,6 +99,18 @@ class BaseGuildTextChannel extends GuildChannel {
...
+  /**
+   * Whether the client user can start a new thread in this channel.
+   *
+   * @type {boolean}
+   * @readonly
+   */
+  get threadable() {
+    if (!this.viewable) return false;
+    const permissions = this.permissionsFor(this.client.user);
+    return permissions?.has(PermissionFlagsBits.CreatePublicThreads, false) ?? false;
```
vakiliner: Creating private threads requires the `CreatePrivateThreads` permission.
Maybe should remove this getter and leave it only in `Message`?

vakiliner (changes requested):

packages/discord.js/src/structures/BaseGuildTextChannel.js
Dramex: Good point, dropped the channel-level getter in [57ede0621](https://github.com/discordjs/discord.js/pull/11500/commits/57ede0621). `TextChannel` can host both public and private threads (two separate permissions), so the boolean was ambiguous. `Message#threadable` stays because `startThread` on a message always creates a public thread, which makes the single `CreatePublicThreads` check unambiguous there. The PR body + description are also updated in scope accordingly.

packages/discord.js/src/structures/Message.js
```diff
@@ -810,6 +810,23 @@ class Message extends Base {
...
+  /**
+   * Whether a thread can be started from this message by the client user.
+   *
+   * @type {boolean}
+   * @readonly
+   */
+  get threadable() {
+    const { channel } = this;
+    if (!channel) return false;
+    if (![ChannelType.GuildText, ChannelType.GuildAnnouncement].includes(channel.type)) return false;
```
Jiralite, line 822:
```suggestion
    if (channel.type !== ChannelType.GuildText && channel.type !== ChannelType.GuildAnnouncement) return false;
```

Jiralite (changes requested):

PR #11501, opened 2026-04-19
fix(collection): preserve ReadonlyCollection through tap/each

### LINKED ISSUE #10514: Able to modify ReadonlyCollection

collection

### Issue description

I doubt this matters too much, but here it is anyway.

Steps to reproduce:
1. Call any method which has the `this` return type upon the Collection.
2. The method will return a Collection which is no longer of the `ReadonlyCollection` type...
3. ...thus, you can call methods such as `.set()` and `.delete()` on the Collection.

### Code sample

```typescript
const testColl: ReadonlyCollection<string, number> = new Collection<string, number>([
    ['id', 123]
]);

testColl.tap(() => {}).set('unexpected', 0);
testColl.each(() => {}).delete('id');

console.log(testColl); // Collection(1) [Map] { 'unexpected' => 0 }
```

### Versions

- @discordjs/collection 2.1.1
- discord.js 14.16.2
- typescript 5.3.3
- NodeJS 20.14.0
- Windows 22H2

### Issue priority

Low (slightly annoying)

Not applicable

Not applicable

_No response_

Dramex:
## Summary

- `Collection#each` and `Collection#tap` return polymorphic `this`, but TypeScript resolves `this` against the `Omit<Collection, ...>` half of the `ReadonlyCollection` intersection and drops the readonly narrowing. That let callers reach `set` / `delete` on the result of a chain started from a `ReadonlyCollection`.
- Fix: exclude `each` and `tap` from the base `Omit`, then re-declare them on the `ReadonlyCollection` side of the intersection with a return type of `ReadonlyCollection`, so chaining stays readonly.
- Adds a regression test that asserts `set` / `delete` on a tapped/eached `ReadonlyCollection` are compile errors (via `@ts-expect-error`, checked by vitest's typecheck pass).

## Why

The original report shows the exact repro, paraphrasing:

```ts
const ro: ReadonlyCollection<string, number> = new Collection([['id', 123]]);
ro.tap(() => {}).set('unexpected', 0); // compiled, mutated the underlying Map
ro.each(() => {}).delete('id');
```

So the readonly marker was structurally ineffective once any chainable helper was involved. This brings the intended semantics back without changing runtime behavior.

## Test plan

- [x] `pnpm --filter=@discordjs/collection test` (168/168 passing, including the new regression test with its type-level assertions)
- [x] `pnpm --filter=@discordjs/collection lint`

Closes #10514

### DIFF
--- a/packages/collection/__tests__/collection.test.ts
+++ b/packages/collection/__tests__/collection.test.ts
@@ -1,7 +1,7 @@
 /* eslint-disable unicorn/no-array-method-this-argument */
 /* eslint-disable id-length */
 import { describe, test, expect } from 'vitest';
-import { Collection, type ReadonlyCollection } from '../src/index.js';
+import { Collection } from '../src/index.js';
 
 type TestCollection<Value> = Collection<string, Value>;
 
@@ -933,25 +933,6 @@ describe('tap() tests', () => {
 
 	test('the collection should be the same', () => {
 		coll.tap((c) => expect(c).toStrictEqual(coll));
-	});
-
-	test('preserves ReadonlyCollection through chainable methods', () => {
-		const readonly: ReadonlyCollection<string, number> = createCollectionFrom(['a', 1]);
-
-		// Compile-time-only check: the chained result of tap/each on a ReadonlyCollection
-		// must not expose Map mutation methods. The `if (false)` guard keeps the assertions
-		// purely type-level so Vitest does not actually mutate the backing collection at
-		// runtime (otherwise `set` + `delete` would silently cancel out numerically).
-
-		if (false as boolean) {
-			// @ts-expect-error `set` is not exposed on ReadonlyCollection.
-			readonly.tap(() => {}).set('b', 2);
-			// @ts-expect-error `delete` is not exposed on ReadonlyCollection.
-			readonly.each(() => {}).delete('a');
-		}
-
-		expect(readonly.size).toBe(1);
-		expect(readonly.get('a')).toBe(1);
 	});
 });
 
--- a/packages/collection/__tests__/collection.test.ts
+++ b/packages/collection/__tests__/collection.test.ts
@@ -938,13 +938,20 @@ describe('tap() tests', () => {
 	test('preserves ReadonlyCollection through chainable methods', () => {
 		const readonly: ReadonlyCollection<string, number> = createCollectionFrom(['a', 1]);
 
-		// Chaining on a ReadonlyCollection must not expose Map mutation methods.
-		// @ts-expect-error `set` is not exposed on ReadonlyCollection.
-		readonly.tap(() => {}).set('b', 2);
-		// @ts-expect-error `delete` is not exposed on ReadonlyCollection.
-		readonly.each(() => {}).delete('a');
+		// Compile-time-only check: the chained result of tap/each on a ReadonlyCollection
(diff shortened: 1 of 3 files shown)


packages/collection/__tests__/collection.test.ts
```diff
@@ -934,6 +934,18 @@ describe('tap() tests', () => {
...
+		const readonly: ReadonlyCollection<string, number> = createCollectionFrom(['a', 1]);
+
+		// Chaining on a ReadonlyCollection must not expose Map mutation methods.
+		// @ts-expect-error `set` is not exposed on ReadonlyCollection.
+		readonly.tap(() => {}).set('b', 2);
+		// @ts-expect-error `delete` is not exposed on ReadonlyCollection.
+		readonly.each(() => {}).delete('a');
+
+		expect(readonly.size).toBe(1);
+	});
```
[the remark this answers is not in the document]
Dramex: Right, the `set` / `delete` calls still ran at runtime and the `size === 1` assertion only held because they happened to cancel out numerically. Pushed [df49a23f3](https://github.com/discordjs/discord.js/pull/11501/commits/df49a23f3) that wraps the offending lines in `if (false)` so the `@ts-expect-error` pair remains a pure compile-time check, and added `get('a') === 1` as a belt to lock in key identity too.

vladfrangu (approved):

sdanialraza (approved):

packages/collection/__tests__/collection.test.ts
```diff
@@ -934,6 +934,25 @@ describe('tap() tests', () => {
 	test('the collection should be the same', () => {
 		coll.tap((c) => expect(c).toStrictEqual(coll));
 	});
+
+	test('preserves ReadonlyCollection through chainable methods', () => {
```
almeidx: Type tests should go in a `collection.test-d.ts` file instead[^1]

[^1]: https://vitest.dev/guide/testing-types.html#testing-types

packages/collection/__tests__/collection.test.ts
```diff
@@ -934,6 +934,25 @@ describe('tap() tests', () => {
...
+		const readonly: ReadonlyCollection<string, number> = createCollectionFrom(['a', 1]);
+
+		// Compile-time-only check: the chained result of tap/each on a ReadonlyCollection
+		// must not expose Map mutation methods. The `if (false)` guard keeps the assertions
+		// purely type-level so Vitest does not actually mutate the backing collection at
+		// runtime (otherwise `set` + `delete` would silently cancel out numerically).
+
+		if (false as boolean) {
```
almeidx: This will *probably* not be needed when in a *.test-d.ts file

almeidx (changes requested):

packages/collection/__tests__/collection.test.ts
Dramex: > Type tests should go in a `collection.test-d.ts` file instead

Moved in c3439d1b4. New file has `expectTypeOf` assertions for both overloads of `tap` and `each`.

packages/collection/__tests__/collection.test.ts
Dramex: > This will *probably* not be needed when in a *.test-d.ts file

Right, the `if(false)` guard and the runtime size assertion are both gone in the new file. The typecheck pass on `.test-d.ts` does the work now.

almeidx (approved):

PR #11503, opened 2026-04-19
fix(Collector): resetTimer now updates options and handles null timers

### LINKED ISSUE #11286: Collector#resetTimer does not properly update idle and time options

discord.js

### Issue description

The `Collector#resetTimer` method does not properly update the collector's timers after changing the `idle` value. Even after calling `resetTimer` with a new `idle` or `time`, the collector continues to use the old timers. This makes the method ineffective in scenarios where the idle timeout needs to be extended dynamically.

**Steps to reproduce**

1. Create a collector with an initial `idle` time.
2. Wait some time and then call `collector.resetTimer({ idle: newIdleValue })`.
3. Observe that the old idle timer still triggers, and the collector does not reflect the new `idle` value in `collector.options.idle`.

**Expected behavior**

* The old timers should be cleared.
* New timers should start based on the updated `idle` and `time` values.
* `collector.options.idle` and `collector.options.time` should reflect the new values.

### Code sample

```typescript
const collector = msg.createMessageComponentCollector({
    filter: (i) => i.user.id === author,
    time: 5 * 60_000,  // 5 minutes
    idle: 60_000,      // 1 minute
});

collector.on("collect", async (i: ButtonInteraction | StringSelectMenuInteraction) => {
    // Example: when the "start" button is pressed, attempt to reset timer
    if (i.isButton() && i.customId === "start") {
        const newTime = 15 * 60_000; // 15 minutes
        collector.resetTimer({
            time: newTime,
            idle: newTime,
        });

        // Debug logs
        console.log("Collector timer reset called:");
        console.log("Options.time:", collector.options.time);
        console.log("Options.idle:", collector.options.idle);
    }
});

collector.on("end", (collected, reason) => {
    console.log(`Collector ended: ${reason}`);
});
```

### Versions

* discord.js version: 14.22.1
* Node.js version: v20.18.1
* OS: windows 10 

### Issue priority

Low (slightly annoying)

Channel, Message, GuildMember, User, ThreadMember

No Intents, Not applicable

_No response_

Dramex:
Closes #11286.

### Root cause

\`Collector#resetTimer\` rebuilt \`_timeout\` / \`_idletimeout\` from its arguments but never wrote the new values back onto \`this.options\`. Any later code path that reads \`this.options.idle\` or \`this.options.time\` (the one in \`handleCollect\` resetting the idle timer after each collect is the loud one) kept using the old values, so the reset only held until the next collect came in.

It also returned early when the collector was started without a \`time\` or \`idle\` option at all, because \`this._timeout\` / \`this._idletimeout\` were \`null\`. That made dynamically adding a timeout to an existing collector impossible, which was the secondary complaint in the issue.

### Change

\`resetTimer\` now:

1. Writes any provided \`time\` / \`idle\` to \`this.options\` first.
2. Clears the existing timer(s) unconditionally (harmless if they were \`null\`).
3. Creates new timers from \`this.options.time\` / \`this.options.idle\` when those are truthy.

Step 3 is driven by the (possibly just updated) option value, not by the presence of an old timer, so passing a new \`idle\` on a collector that had none now activates idle tracking from that point forward.

### Verification

Manual repro from the issue no longer triggers the stale-timer behaviour, and follow-up \`handleCollect\` calls use the new \`options.idle\`.

### DIFF
--- a/packages/discord.js/src/structures/interfaces/Collector.js
+++ b/packages/discord.js/src/structures/interfaces/Collector.js
@@ -272,14 +272,25 @@ class Collector extends AsyncEventEmitter {
    * @param {CollectorResetTimerOptions} [options] Options for resetting
    */
   resetTimer({ time, idle } = {}) {
+    if (time !== undefined) this.options.time = time;
+    if (idle !== undefined) this.options.idle = idle;
+
     if (this._timeout) {
       clearTimeout(this._timeout);
-      this._timeout = setTimeout(() => this.stop('time'), time ?? this.options.time).unref();
+      this._timeout = null;
+    }
+
+    if (this.options.time) {
+      this._timeout = setTimeout(() => this.stop('time'), this.options.time).unref();
     }
 
     if (this._idletimeout) {
       clearTimeout(this._idletimeout);
-      this._idletimeout = setTimeout(() => this.stop('idle'), idle ?? this.options.idle).unref();
+      this._idletimeout = null;
+    }
+
+    if (this.options.idle) {
+      this._idletimeout = setTimeout(() => this.stop('idle'), this.options.idle).unref();
     }
   }
 


Dramex:
duplicate of #11460, my bad

PR #11504, opened 2026-04-21
fix(Action): restore DM recipient fallback for untyped channel payloads

### LINKED ISSUE #11486: Changes to getChannel method in 14.26.2 leading to Bots not reacting to DMs anymore

### LINKED ISSUE #11497: Bot Fails to Receive DM (Direct Message) Text From v14.23.2 But Working in Dev Build

discord.js

### Issue description

I'm new to Discord bots and I think I've discovered a bug that prevents the bot from properly receiving DMs (direct messages) sent from a user to the bot. I rolled back to version 14.23.1 and it works fine, but it's broken on v14.23.2 and v14.23.3. However, it's working in the latest dev build which I tested for this GitHub issue (discord.js@15.0.0-dev.1776125601-58c5ebdd0).

I've distilled this down to minimal code and a potential workaround, but it's flawed because the bot won't respond to the first message sent (you can modify the code to have it perform work in a separate function to get around this, though).

When a DM is received, the following debug event fires with this message:

> Failed to find guild, or unknown type for channel [channel_id_here] undefined

We also get a raw event and if we call `client.channels.fetch(channelId)`, we can "cache" (I think) the channel ID for future DMs. Once this runs, we get subsequent messages from that user, but this is lost when the script restarts.

Here is sample output from my console when running the code sample, then sending two DMs to my bot:

> Ready! Logged in as [my_bot_name]
> [WARNING] Detected MESSAGE_CREATE in uncached DM channel [channel_id_here] ...
> [DEBUG LOG]: Failed to find guild, or unknown type for channel [channel_id_here] undefined
> [WORKAROUND] Cached DM channel [channel_id_here] via MESSAGE_CREATE.
> Received message: hello from [my_discord_username]

If I can offer any additional help, I'd be happy to assist.
Thank you!

Edit: forgot to add client.login() to the sample code

### Code sample

```typescript
const { Client, Events, GatewayIntentBits, Partials } = require('discord.js');
const { token } = require('./config.json');

const client = new Client({
	intents: [
		GatewayIntentBits.DirectMessages
	],
	partials: [
		Partials.Channel
	],
});

// login
client.once(Events.ClientReady, (readyClient) => {
	console.log(`Ready! Logged in as ${readyClient.user.tag}`);
});

// show the error when the bot receives a DM
client.on('debug', (info) => {
	if (info.includes('Failed to find')) {
		console.log(`[DEBUG LOG]: ${info}`);
	}
});

// log a message on successful DM
client.on('messageCreate', async (message) => {
	console.log(`Received message: ${message.content} from ${message.author.tag}`);
});

// workaround: catch the bug and manually cache the channel ID
client.on('raw', async (packet) => {
	if (packet.t != 'MESSAGE_CREATE') return;

	const channelId = packet.d.channel_id;
	if (!client.channels.cache.has(channelId)) {
		console.log(`[WARNING] Detected ${packet.t} in uncached DM channel ${channelId} ...`);
		try {
			await client.channels.fetch(channelId);
			console.log(`[WORKAROUND] Cached DM channel ${channelId} via ${packet.t}.`);
		}
		catch (error) {
			console.log(`[ERROR] Failed to cache DM channel ${channelId} via ${packet.t}:`, error);
		}
	}
});

// login to Discord with our token from config.json
client.login(token);
```

### Versions

- discord.js v14.26.3
- node v24.15.0 (tested also back on v22)
- VS Code v1.115.0 (2026-04-07T22:04:57Z)
- Windows 11 x64

### Issue priority

Medium (should be fixed soon)

Channel

DirectMessages

58c5ebdd0

esterne:
Please describe the changes this PR makes and why it should be merged:

Fixes #11486. Also closes #11497 (duplicate report).

## The bug

On `v14`, uncached DM channels never receive a `messageCreate` event. The first DM sent to a freshly-started bot is silently dropped; subsequent DMs from the same user hit the same path because the cache is never populated.

Reproducible with the minimal code sample from #11486 / #11497 against `discord.js@14.26.3`.

## Root cause

Introduced in #11479. That PR narrowed `Action.getChannel()`'s recipient-injection branch from an unconditional `else` to `else if (data.type === ChannelType.DM || data.type === ChannelType.GroupDM)`.

The problem: the callers that route through `getChannel()` don't all carry a channel `type`. In particular, `MessageCreateAction.handle()` (`src/client/actions/MessageCreate.js`) builds a stripped object:

```js
const channel = this.getChannel({
  id: data.channel_id,
  author: data.author,
  ...('guild_id' in data && { guild_id: data.guild_id }),
});
```

That object has no `type` field at all. `data.type` in `getChannel` is therefore `undefined`, neither the DM nor the GroupDM arm matches, and no `recipients` key is injected into `payloadData`.

Downstream, `createChannel()` (`src/util/Channels.js`) reaches the DM detection:

```js
if ((data.recipients && data.type !== ChannelType.GroupDM) || data.type === ChannelType.DM) {
  channel = new DMChannel(client, data);
}
```

Both conditions are false without `recipients` and without `type`, so `createChannel()` returns `undefined`, `ChannelManager._add` returns `null`, and `MessageCreateAction.handle`'s `if (channel)` guard silently short-circuits. The `messageCreate` event never fires.

The cache never gets populated, so every subsequent DM hits the same path.

## The fix

Restore the pre-#11462 fallback pattern that `main` already ships: when there's no existing `recipients` array on the payload, inject the derived recipient unconditionally, guarded by the client-user check so the bot doesn't add itself as a recipient for its own outbound events.

```js
if ('recipients' in data) {
  // existing dedupe logic, unchanged
} else {
  if (recipient.id !== this.client.user.id) payloadData.recipients = [recipient];
}
```

The `ChannelType` import is no longer needed.

## Why this is safe for guild channels (i.e. why #11479's concern doesn't regress)

Guild channels enter `createChannel()` via the `else` branch (`data.guild_id` is present), which ignores `recipients` entirely — it selects the channel constructor from `data.type` via the switch and passes the guild in. Nothing reads `payloadData.recipients` on that path. The `!data.guild_id && !guild` gate at the top of the DM arm in `createChannel()` already prevents any recipients-bearing guild payload from being misrouted to `new DMChannel()`.

The guard `recipient.id !== this.client.user.id` — already on `main` — additionally prevents self-recipient injection for cases where the event author is the bot itself.

## Equivalence with `main`

`main` already has this code. It landed as part of #11462's pre-image on `main` and was never touched by the subsequent `v14` PRs. The dev build `15.0.0-dev.1776125601-58c5ebdd0` has been reported working (see #11497). This PR effectively cherry-picks `main`'s `getChannel` implementation back into `v14`.

## Status

- [x] Documentation added (if applicable) — n/a, no API change
- [x] Tests added (if applicable) — the `discord.js` package currently ships only two integration-style tests (`test/reactionCollectorCreated.test.js`, `test/resolveGuildTemplateCode.test.js`) which require live Discord credentials. There is no vitest harness set up at the package level for pure unit tests of `Action.getChannel`. Manual reproduction is documented below and matches the original bug reports.
- [x] I have run `pnpm run lint` (passes, no new warnings) and `pnpm run test:typescript` (passes).

## Manual reproduction

1. Bot with `intents: [DirectMessages, MessageContent]` and `partials: [Partials.Channel]`.
2. Send the bot a DM from a user account, cold-started, with no prior DM history cached.
3. **Before this patch:** gateway delivers `MESSAGE_CREATE` (verifiable via a `raw` listener) but no `messageCreate` event fires. `client.on('debug', …)` shows `Failed to find guild, or unknown type for channel <id> undefined`.
4. **After this patch:** `messageCreate` fires with `channel.type === 1` (DM) and downstream handlers run normally.

This same pattern was validated independently in a working downstream workaround in https://github.com/esterne/simianclaw/commit/0265e35 — the workaround is no longer needed once this fix lands.

### DIFF
--- a/packages/discord.js/src/client/actions/Action.js
+++ b/packages/discord.js/src/client/actions/Action.js
@@ -1,6 +1,5 @@
 'use strict';
 
-const { ChannelType } = require('discord-api-types/v10');
 const { Poll } = require('../../structures/Poll.js');
 const { PollAnswer } = require('../../structures/PollAnswer.js');
 const Partials = require('../../util/Partials.js');
@@ -40,10 +39,15 @@ class GenericAction {
       if (!data.recipients.some(existingRecipient => recipient.id === existingRecipient.id)) {
         payloadData.recipients = [...data.recipients, recipient];
       }
-    } else if (data.type === ChannelType.DM || data.type === ChannelType.GroupDM) {
-      // Try to resolve the recipient.
+    } else {
+      // Try to resolve the recipient, but do not add the client user.
+      // This branch fires for events where the channel type isn't carried
+      // in the payload (MESSAGE_CREATE, MESSAGE_UPDATE, etc. all strip
+      // `type` before reaching here). Without the recipient being injected
+      // the downstream createChannel() can't construct an uncached
+      // DMChannel, which silently drops the event — see #11486.
       const recipient = data.author ?? data.user ?? { id: data.user_id };
-      payloadData.recipients = [recipient];
+      if (recipient.id !== this.client.user.id) payloadData.recipients = [recipient];
     }
 
     if (id !== undefined) payloadData.id = id;


Qjuh:
Is this AI? The reasoning is completely wrong. `main` doesn't "already have this" nor was a fix introduced to `main` by #11462 since that was a backport into `v14` branch and didn't even touch the else branch this is about.

`main` branch didn't merge the PR that introduced this yet because while it fixes a bug (that this PR would reintroduce) it also adds another bug (the one this PR claims to fix).

#11495 is the real fix for the issue without reintroducing any previous bugs that got fixed by prior PRs.

Qjuh:
Superseded by #11495

PR #11514, opened 2026-04-27
fix(scripts): correctly mark external docs on non-main tags

Qjuh:
This should fix the issue where the source of re-exported external members links to a `/dist/index.d.ts` file instead of the documentation of the external package.

Example: https://discord.js.org/docs/packages/discord.js/14.25.0/SlashCommandBuilder:Class

### DIFF
--- a/packages/scripts/src/generateSplitDocumentation.ts
+++ b/packages/scripts/src/generateSplitDocumentation.ts
@@ -540,10 +540,12 @@ function resolveFileUrl(item: ApiDeclaredItem) {
 	} else if (fileUrl?.includes('/dist/') && fileUrl.includes('/packages/')) {
 		const [dir, pkg] = fileUrl.split('/packages/');
 		const pkgName = pkg!.split('/')[0];
+		const prefix = pkgName === 'discord.js' ? '' : '@discordjs/';
 		const version =
 			dir?.split('/').at(-1) === 'main'
 				? 'main'
-				: item.getAssociatedPackage()?.dependencies?.[pkgName!]?.replace(/[~^]/, '');
+				: // eslint-disable-next-line unicorn/better-regex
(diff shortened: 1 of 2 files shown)


packages/scripts/src/generateSplitDocumentation.ts
```diff
@@ -537,10 +537,13 @@ function resolveFileUrl(item: ApiDeclaredItem) {
...
-	} else if (fileUrl?.includes('/dist/') && fileUrl.includes('/main/packages/')) {
-		const [, pkg] = fileUrl.split('/main/packages/');
+	} else if (fileUrl?.includes('/dist/') && fileUrl.includes('/packages/')) {
+		const [dir, pkg] = fileUrl.split('/packages/');
 		const pkgName = pkg!.split('/')[0];
-		const version = 'main';
+		const version =
+			dir?.split('/').at(-1) === 'main'
+				? 'main'
+				: item.getAssociatedPackage()?.dependencies?.[pkgName!]?.replace(/[~^]/, '');
```
almeidx: Does this actually work? From what it looks like, `pkgName` here would only be e.g. `collection`, not `@discordjs/collection`

almeidx: Also, this line still shows a warning (even though it doesn't show up in the CI here, because of... reasons...) for the unicorn/better-regex rule
I'd rather we disable the rule here with a eslint-disable comment, since it's clearly a false-positive.

Qjuh: Thanks for catching both of these, you're right. It's hard to test this since it only applies to non-main tags.

almeidx (approved):

Jiralite (approved):

ckohen (approved):

iCrawl (approved):

PR #11516, opened 2026-04-28
fix: clean up mixin error text and switch Next middleware to proxy

AADI0009:
corrected the copied error message in ApiParameterListMixin so it points to the right mixin & renamed the guide and website Nextjs middleware entrypoints to proxy to match the current convention and remove the deprecation warning

### DIFF
--- a/apps/guide/src/proxy.ts
+++ b/apps/guide/src/proxy.ts
@@ -1,6 +1,6 @@
 import { NextResponse, type NextRequest } from 'next/server';
 
-export function middleware(request: NextRequest) {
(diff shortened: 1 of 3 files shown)


almeidx:
The api-extractor change should be made upstream instead:
https://github.com/microsoft/rushstack/blob/3c15b4dc66188e6f014a2da6bda7c3e0069353b3/libraries/api-extractor-model/src/mixins/ApiParameterListMixin.ts#L135

As for the middleware -> proxy changes, `@opennextjs/cloudflare` doesn't support it, which is what we are using.

PR #11519, opened 2026-04-30
chore: use tsdown

Qjuh:
This PR does 3 things:
- bump typescript to v6
- move from tsup to tsdown as bundler, since tsup is no longer maintained
- remove legacy CJS builds, since node support require of ESM builds in all supported versions

This is relevant because tsup doesn't support TS 6.0.


packages/actions/tsdown.config.ts
```diff
@@ -12,6 +12,5 @@ export default createTsupConfig({
 	],
 	dts: false,
 	format: 'esm',
```
almeidx:
```suggestion
```

PR #11527, opened 2026-05-07
fix(ModalSubmitFields): respect `required` option value (v14)

RiskyMH:
Fixes modal fields unconditionally throwing when required is false/undefined. 

>    `   * @param {boolean} [required=false] Whether to throw an error if the component value is not found or empty`

An example jsdoc on most of the getters where it explicitly mentions the throwing on "not found", this implies that it shouldn't throw unconditionally.

This brings it inline with the command interaction options:

https://github.com/discordjs/discord.js/blob/40ce0791a8ed348189d9aefe0669cee29bf59920/packages/discord.js/src/structures/CommandInteractionOptionResolver.js#L75-L116

---

**Note:** To try and minimize breaking changes, the public `getField` function defaults to required but the individual getters still fulfill their jsdoc promise (should prob fix properly in v15)

### DIFF
--- a/packages/discord.js/src/structures/ModalSubmitFields.js
+++ b/packages/discord.js/src/structures/ModalSubmitFields.js
@@ -54,11 +54,17 @@ class ModalSubmitFields {
    * Gets a field given a custom id from a component
    * @param {string} customId The custom id of the component
    * @param {ComponentType} [type] The type of the component
-   * @returns {ModalData}
+   * @param {boolean} [required=false] Whether to throw an error if the field is not found.
+   * @returns {?ModalData} The field, if found.
    */
-  getField(customId, type) {
+  getField(customId, type, required = false) {
     const field = this.fields.get(customId);
-    if (!field) throw new DiscordjsTypeError(ErrorCodes.ModalSubmitInteractionFieldNotFound, customId);
+    if (!field) {
+      if (required) {
+        throw new DiscordjsTypeError(ErrorCodes.ModalSubmitInteractionFieldNotFound, customId);
+      }
+      return null;
+    }
 
     if (type !== undefined && type !== field.type) {
       throw new DiscordjsTypeError(ErrorCodes.ModalSubmitInteractionFieldType, customId, field.type, type);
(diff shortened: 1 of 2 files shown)


imnaiyar:
Wouldn't this be a breaking change? And the behaviour is correct. The docs says it'll will throw an error if the values don't exist, which implies the component should exist first. 

Historically modal components have always thrown an error if it was not found. 

Though I do think its behaviour should be aligned to match command options, but perhaps on main since I think it would be breaking on v14

RiskyMH:
> Wouldn't this be a breaking change

Yeah probably on the .getField

> The docs says it'll will throw an error if the values don't exist, which implies the component should exist first.

At very least in the getters it documents the "not found" part. I could in theory make a private .getField instead so the public api stays the same. In current state some of those getters have no reason to even be able to return null?

https://github.com/discordjs/discord.js/blob/0516dc7862c71ffd2b49bde4e0679088d258606b/packages/discord.js/src/structures/ModalSubmitFields.js#L231-L240

imnaiyar:
> At very least in the getters it documents the "not found" part. I could in theory make a private .getField instead so the public api stays the same. In current state some of those getters have no reason to even be able to return null?

You are right,  that method shouldn't need required field, because I believe discord always returns a string (just empty). It should be like other fields with `value` as prop

RiskyMH:
@imnaiyar If I make the default value for .getField required true (so current code works the same), would that be a fair compromise. Like I'd be willing to say that people like me saw the jsdoc and understood it as the same as slash command one. But as this didn't mention it at all, it indeed shouldn't be changed

Or are you saying none of them can be seen as a bug fix as its still going to change peoples behavior? Like I'd argue it would cause more issues if we remove the "useless" param as people could get type errors.

Qjuh:
There's something inherently different about what "required" means for Modal fields and for chatinput commands though.
- If a command has an option as `required: false` then that option might be completely missing from the command interaction, if not used. So the check needs to account for that.
- If a Modal has a component in it, then that component will also appear in the ModalSubmitInteraction's fields data. The values might be empty, but the field will always be present.

What you are trying to "fix" here seems to me like trying to abuse the methods in a way of "I don't know what fields I actually included in the modal, so I need a fallback if I didn't include the field". I don't think that's the intention of these methods.

RiskyMH:
@Qjuh that is a reasonable answer. However it does feel a little confusing considering consistency of knowing slash commands and the jsdoc mentioned. If you believe it's correct though, then I'll be happy to at very least fix that in a diff pr!

But required does make sense to me being actually need to be present vs just not empty. Logically it does make sense the reasoning behind slash commands are easy to have optional, however when making stateless options you may have many options again and in current way you would need be smart (but its prob too much of a breaking change to touch here in v14 anyway).

Also yeah after looking at more of the functions I can indeed see that not all even have that as an param, thus further proving your point that its just communication issue.

PR #11530, opened 2026-05-09
feat(UsersAPI): add `deleteApplicationRoleConnection()` method

kshitijanurag:
Add method for deleting current user application role connection to `UsersAPI`

Discord API Docs reference:
- https://github.com/discord/discord-api-docs/pull/8348


packages/core/src/api/user.ts
```diff
@@ -183,4 +183,21 @@ export class UsersAPI {
 			signal,
 		}) as Promise<RESTPutAPICurrentUserApplicationRoleConnectionResult>;
 	}
+
+	/**
+	 * Deletes the current user's application role connection
+	 *
+	 * @see {@link https://docs.discord.com/developers/resources/user#delete-user-application-role-connection}
```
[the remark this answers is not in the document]
almeidx: Discord's documentation moved from https://discord.com/developers/docs to https://docs.discord.com/developers. Going forward we should use the new domain, so this is correct as-is

almeidx: The anchor is actually not working. And neither do the ones for the other methods above related to application role connection.

Seems like the anchor names changed from the docs migration. Perhaps we should correct them

Jiralite (changes requested):
Regarding https://github.com/discordjs/discord.js/pull/11530#discussion_r3264885613, at the very least, please address the documentation link introduced with this pull request.

packages/core/src/api/user.ts
kshitijanurag: Updated the documentation links.

almeidx (approved):

Jiralite (approved):

vladfrangu (approved):

PR #11534, opened 2026-05-17
fix: update collector options.time and options.idle in resetTimer

### LINKED ISSUE #11286: Collector#resetTimer does not properly update idle and time options

z9z:
Closes #11286.
resetTimer was passing the new time/idle values to setTimeout but never persisting them in this.options. After calling resetTimer({ time: X }), collector.options.time still showed the old value.


packages/discord.js/src/structures/interfaces/Collector.js
```diff
@@ -272,14 +272,22 @@ class Collector extends AsyncEventEmitter {
    * @param {CollectorResetTimerOptions} [options] Options for resetting
    */
   resetTimer({ time, idle } = {}) {
+    if (time !== undefined) {
```
Qjuh:
```suggestion
    if (time) {
```

packages/discord.js/src/structures/interfaces/Collector.js
```diff
@@ -272,14 +272,22 @@ class Collector extends AsyncEventEmitter {
...
+      this.options.time = time;
+    }
+
+    if (idle !== undefined) {
```
Qjuh:
```suggestion
    if (idle) {
```
This makes it consistent with how `time` and `idle` get handled in constructor.

Qjuh (dismissed):

almeidx (approved):

Qjuh (approved):

vladfrangu (approved):

PR #11536, opened 2026-05-18
fix(RoleManager): allow null in RoleColorsResolvable to clear gradient colors

camdzic:
## Summary

`RoleManager#edit` already supports clearing gradient colors by passing `null` for `secondaryColor`/`tertiaryColor` — the runtime short-circuits on falsy values and sends `secondary_color: null` / `tertiary_color: null` in the PATCH body, which Discord interprets as "clear this color". However, `RoleColorsResolvable` typed those fields as `ColorResolvable | undefined`, so consumers couldn't pass `null` without a type error.

The output type `RoleColors` already exposes these as `number | null`, so this PR brings the input type in line with the output.

## Example

Previously this was a type error despite working at runtime:

```js
await guild.roles.edit(roleId, {
  colors: {
    primaryColor: 0xff0000,
    secondaryColor: null, // ← type error before, now allowed
    tertiaryColor: null,  // ← type error before, now allowed
  },
});
```

## Changes

- `packages/discord.js/typings/index.d.ts`: widen `secondaryColor` / `tertiaryColor` in `RoleColorsResolvable` to `ColorResolvable | null`.
- `packages/discord.js/src/managers/RoleManager.js`: update the `RoleColorsResolvable` JSDoc to mark both fields nullable and document the `null` behavior.

## Status and versioning classification

- Code changes have been tested against the Discord API: yes (runtime already supported this).
- This PR changes the library's interface (typings): yes, but only widens an existing type. Non-breaking.
- This PR includes breaking changes: no.

### DIFF
--- a/packages/discord.js/src/managers/RoleManager.js
+++ b/packages/discord.js/src/managers/RoleManager.js
@@ -123,9 +123,11 @@ class RoleManager extends CachedManager {
   /**
    * @typedef {Object} RoleColorsResolvable
    * @property {ColorResolvable} primaryColor The primary color of the role
-   * @property {ColorResolvable} [secondaryColor] The secondary color of the role.
+   * @property {?ColorResolvable} [secondaryColor] The secondary color of the role.
+   * Pass `null` to clear it.
    * This will make the role a gradient between the other provided colors
-   * @property {ColorResolvable} [tertiaryColor] The tertiary color of the role.
+   * @property {?ColorResolvable} [tertiaryColor] The tertiary color of the role.
+   * Pass `null` to clear it.
    * When sending `tertiaryColor` the API enforces the role color to be a holographic style with values of `primaryColor = 11127295`, `secondaryColor = 16759788`, and `tertiaryColor = 16761760`.
    * These values are available as a constant: `Constants.HolographicStyle`
    */
--- a/packages/discord.js/src/managers/RoleManager.js
+++ b/packages/discord.js/src/managers/RoleManager.js
@@ -123,11 +123,19 @@ class RoleManager extends CachedManager {
   /**
    * @typedef {Object} RoleColorsResolvable
    * @property {ColorResolvable} primaryColor The primary color of the role
-   * @property {?ColorResolvable} [secondaryColor] The secondary color of the role.
-   * Pass `null` to clear it.
+   * @property {ColorResolvable} [secondaryColor] The secondary color of the role.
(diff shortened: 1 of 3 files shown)


SpaceEEC (approved):

packages/discord.js/src/managers/RoleManager.js
```diff
@@ -123,9 +123,11 @@ class RoleManager extends CachedManager {
   /**
    * @typedef {Object} RoleColorsResolvable
    * @property {ColorResolvable} primaryColor The primary color of the role
-   * @property {ColorResolvable} [secondaryColor] The secondary color of the role.
+   * @property {?ColorResolvable} [secondaryColor] The secondary color of the role.
+   * Pass `null` to clear it.
```
Qjuh: This type (both JSDoc and ts typings) is used for both create and edit of a Role. It should only be nullable for edit though. On create colors should either be omitted or valid colors, not null. So this would need a new type I guess.

Qjuh (changes requested):

packages/discord.js/src/managers/RoleManager.js
camdzic: Good catch — addressed. Split `RoleColorsResolvable` (unchanged, used for create) from a new `RoleColorsEditResolvable` that allows `null` on `secondaryColor`/`tertiaryColor` to clear gradient colors. `RoleEditOptions` now uses `Omit<RoleData, 'colors'>` to swap in the editable variant. JSDoc updated accordingly.

packages/discord.js/typings/index.d.ts
```diff
@@ -5861,7 +5867,8 @@ export interface RoleCreateOptions extends RoleData {
   reason?: string;
 }
 
-export interface RoleEditOptions extends RoleData {
+export interface RoleEditOptions extends Omit<RoleData, 'colors'> {
```
Qjuh: Extending an Omit<> type is bad for documentation. The inheritance hierarchy needs to be fixed here.

packages/discord.js/typings/index.d.ts
```suggestion
export interface RoleColorsResolvable extends RoleColorsEditResolvable {
```
to avoid duplication of the same properties. Just narrowing the type of secondary and tertiary to not include null here.

Qjuh (changes requested):

Qjuh (approved):

sdanialraza (approved):

PR #11537, opened 2026-05-22
feat: voice channel status

kshitijanurag:
Add support for setting voice channel status and gateway events.

Discord API Docs reference:
- https://github.com/discord/discord-api-docs/pull/6400

Depends on:
- https://github.com/discordjs/discord-api-types/pull/1631


packages/discord.js/src/structures/Guild.js
```diff
@@ -1451,6 +1452,59 @@ class Guild extends AnonymousGuild {
...
+      const handler = (channelInfo, guild) => {
+        if (guild.id !== this.id) return;
+
+        // eslint-disable-next-line no-use-before-define
+        clearTimeout(timeout);
+        this.client.removeListener(Events.ChannelInfo, handler);
+        this.client.decrementMaxListeners();
+
+        resolve(channelInfo);
+      };
```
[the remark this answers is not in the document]
kshitijanurag: I don't think Discord supports sending a nonce in the payload for this.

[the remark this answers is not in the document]
Qjuh: This is not an issue in the first place, since the response would be valid for all requests waiting on a response for that specific guild_id. Resolved.

Qjuh:
`status` and `voiceStartTime` should be properties on `VoiceChannel` and handled in `VoiceChannel#_patch(...)`. The gateway events would then emit with just the patched channel(s), no wrapping plain object needed, nor is the `Guild` as extra parameter needed to emit, since it's already present as getter on the channel.

packages/discord.js/src/client/websocket/handlers/VOICE_CHANNEL_START_TIME_UPDATE.js
```diff
@@ -0,0 +1,21 @@
...
+
+  channel._patch(data);
+
+  /**
+   * Emitted whenever the voice session start time of a voice channel changes.
+   *
+   * @event Client#voiceChannelStartTimeUpdate
+   * @param {VoiceChannel} channel The voice channel whose voice session start time changed
+   */
+  client.emit(Events.VoiceChannelStartTimeUpdate, channel);
```
Qjuh: This and the other update event should also emit the old channel (before patching) as a clone similar to other *Update events.

Qjuh (changes requested):

Qjuh (approved):

Qjuh:
Needs merge, bump and release of the dtypes PR, but other than that LGTM

packages/core/src/api/voice.ts
```diff
@@ -98,4 +99,25 @@ export class VoiceAPI {
...
+	 * Sets a voice channel's status
+	 *
+	 * @see {@link https://docs.discord.com/developers/resources/channel#set-voice-channel-status}
+	 * @param channelId - The id of the voice channel
+	 * @param body - The data for setting the voice channel status
+	 * @param options - The options for setting the voice channel status
+	 */
+	public async setVoiceChannelStatus(
+		channelId: Snowflake,
+		body: RESTPutAPIChannelVoiceStatusJSONBody = {},
```
almeidx:
```suggestion
		body: RESTPutAPIChannelVoiceStatusJSONBody,
```

packages/discord.js/src/structures/Guild.js
```diff
@@ -1451,6 +1452,66 @@ class Guild extends AnonymousGuild {
...
+   *   if (channel.status && channel.voiceStartTimestamp) {
+   *     console.log(
+   *       `Voice channel ${channel.name} has status set to ${channel.status} and the voice session started at ${channel.voiceStartTimestamp}.`,
+   *     );
+   *   }
+   * }
+   */
+  requestChannelInfo({ fields, time = 10_000 }) {
+    return new Promise((resolve, reject) => {
+      const handler = (channels, guild) => {
```
almeidx, line 1484: Right now, this event is only emitting `(channels)`. I think it would be better to include guild as well, as you've used here

packages/discord.js/src/structures/VoiceChannel.js
```diff
@@ -9,6 +9,32 @@ const { BaseGuildVoiceChannel } = require('./BaseGuildVoiceChannel.js');
...
+      this.status ??= null;
+    }
+
+    if ('voice_start_time' in data) {
+      /**
+       * The timestamp when the current voice session started, or `null` if no session is active.
+       *
+       * @type {?number}
+       */
+      this.voiceStartTimestamp = data.voice_start_time;
```
almeidx: This value comes in seconds from discord, but generally we expose *Timestamp in milliseconds, with a `*At` getter which returns a Date.

packages/discord.js/src/structures/Guild.js
```diff
@@ -1451,6 +1452,66 @@ class Guild extends AnonymousGuild {
...
+   * @param {RequestChannelInfoOptions} options The options for requesting channel info for this guild
+   * @returns {Promise<Collection<Snowflake, VoiceChannel>>}
+   * @example
+   * // Request voice channel status and start time for all channels in a guild
+   * const channels = await guild.requestChannelInfo({
+   *   fields: [GatewayRequestChannelInfoField.Status, GatewayRequestChannelInfoField.VoiceStartTime],
+   * });
+   *
+   * for (const channel of channels) {
+   *   if (channel.status && channel.voiceStartTimestamp) {
```
almeidx: This example is not correct, since `channel` here is actually an entry from the collection (`[snowflake, channel]`). One way to fix it would be to make the iterable side of the loop be `channels.values()` instead

packages/discord.js/src/structures/VoiceChannel.js
```diff
@@ -58,6 +84,30 @@ class VoiceChannel extends BaseGuildVoiceChannel {
...
+   * channel.setStatus('Playing music!')
+   *   .then(() => console.log('Voice channel status set'))
+   *   .catch(console.error);
+   * @example
+   * // Clear a voice channel status
+   * channel.setStatus(null)
+   *   .then(() => console.log('Voice channel status cleared'))
+   *   .catch(console.error);
+   */
+  async setStatus(status) {
```
almeidx: Missing `reason` parameter

packages/discord.js/src/util/Events.js
```diff
@@ -171,6 +175,8 @@ exports.Events = {
   ThreadUpdate: 'threadUpdate',
   TypingStart: 'typingStart',
   UserUpdate: 'userUpdate',
+  VoiceChannelStartTimeUpdate: 'voiceChannelStartTimeUpdate',
```
almeidx, line 179: Can we keep this alphabetically sorted?

almeidx (dismissed):
Can you also add the gateway request to the `Client` on the core package?

Qjuh (approved):

Qjuh (approved):

packages/discord.js/src/structures/Guild.js
almeidx: Not resolved. The emit of this event in packages/discord.js/src/client/websocket/handlers/CHANNEL_INFO.js still only includes `channels`, not `guild`.

kshitijanurag: Resolved.

packages/discord.js/src/structures/VoiceChannel.js
```diff
@@ -9,6 +9,42 @@ const { BaseGuildVoiceChannel } = require('./BaseGuildVoiceChannel.js');
...
+    if ('voice_start_time' in data) {
+      /**
+       * The timestamp when the current voice session started, or `null` if no session is active.
+       *
+       * @type {?number}
+       */
+      this.voiceStartTimestamp = data.voice_start_time && data.voice_start_time * 1_000;
+    } else {
+      this.voiceStartTimestamp ??= null;
+    }
```
almeidx, line 35: `voice_start_time` is documented as optional and nullable for the Voice Channel Start Time Update event. If the event is triggered without this field, nothing will be changed on the voice channel because of the `??=` here, but we also can't just use `=` I don't think. 

Not sure of the difference between `null` and optional on this event. Perhaps something to investigate or ask discord before moving on

Qjuh: After investigating it seems `null` and optional should both be handled the same way, so we'd probably need to special handle it. Possibly by artificially adding it as `null` to the data in VOICE_CHANNEL_START_TIME_UPDATE if not present (not perfect but the best I can think of atm).

Qjuh (approved):

packages/core/src/api/voice.ts
```diff
@@ -98,4 +99,25 @@ export class VoiceAPI {
...
+
+	/**
+	 * Sets a voice channel's status
+	 *
+	 * @see {@link https://docs.discord.com/developers/resources/channel#set-voice-channel-status}
+	 * @param channelId - The id of the voice channel
+	 * @param body - The data for setting the voice channel status
+	 * @param options - The options for setting the voice channel status
+	 */
+	public async setVoiceChannelStatus(
```
almeidx: This is a `/channels/*` route and is under `resources/channel` on the documentation, so it should be on the `ChannelsAPI`

packages/discord.js/src/structures/VoiceChannel.js
```diff
@@ -9,6 +9,42 @@ const { BaseGuildVoiceChannel } = require('./BaseGuildVoiceChannel.js');
  * @extends {BaseGuildVoiceChannel}
  */
 class VoiceChannel extends BaseGuildVoiceChannel {
+  _patch(data) {
+    super._patch(data);
+
+    if ('status' in data) {
+      /**
+       * The voice channel status, or `null` if none is set.
```
almeidx: Since these new fields rely on cache, as they don't get sent on bot init (supposedly, did not test: [changelog entry](<https://docs.discord.com/developers/change-log#:~:text=Voice%20channel%20status%20and%20voice,for%20channels%20in%20the%20guild.>)), I think we should drop "or `null` if none is set"

Maybe include an `<info>` too that this is only set by the new events or by manually calling `requestChannelInfo`

PR #11539, opened 2026-05-24
chore: bump dependencies

almeidx:
highlights
- [pnpm v11](<https://pnpm.io/blog/releases/11.0>) (comes minimumReleaseAge set to 1 day)
- [vite v8](<https://vite.dev/blog/announcing-vite8>) (rollup -> rolldown)
- [undici v8](<https://undici.nodejs.org/best-practices/migrating-from-v7-to-v8>) (HTTP/2 by default)
- fixed rpc's build and docs


Jiralite (approved):

didinele (approved):

didinele (approved):

packages/rest/src/strategies/undiciRequest.ts
```diff
@@ -40,7 +40,9 @@ export async function makeRequest(url: string, init: RequestInit): Promise<Respo
 		get bodyUsed() {
 			return res.body.bodyUsed;
 		},
-		headers: new Headers(res.headers as HeaderRecord),
+		// Over HTTP/2, Node's http2 attaches a `sensitiveHeaders` symbol key to the response headers,
```
ckohen: Taking a look at how undici itself handles this reveals an internal path and a fallback path that we could use.

I think this was never technically correct so we should probably add a util that does what undici does.
```ts
for (const [name, value] of Object.entries(headers ?? {})) {
		if (Array.isArray(value)) {
			for (const entry of value) {
				headersList.append(name, `${entry}`, true);
			}
		} else {
			headersList.append(name, `${value}`, true);
		}
	}
```

packages/rest/src/lib/handlers/Shared.ts
```diff
@@ -121,7 +121,8 @@ export async function makeNetworkRequest(
 				data: requestData,
 				retries,
 			},
-			res instanceof Response ? res.clone() : { ...res },
+			// This cast is needed because of a mismatch between the version of undici-types provided by @types/node and undici
```
ckohen: This actually points out a little bit of a mistake in the strategy-agnostic sense.

Response is global and because we're using instanceof it's not going to match anything but the global version (this is usually fine as in most cases that will match up).

I think the best option here is to change the ResponseLike type to include 
clone?(): ResponseLike and then we can `res.clone?.() ?? { ...res }`.
Also I think the only reason this cast is necessary is because of the headers type in ResponseLike?
If so we should overwrite that similarly to body (can Pick<Headers, 'properties' | 'we' | 'use'> as the type

packages/rest/src/lib/REST.ts
```diff
@@ -382,8 +382,9 @@ export class REST extends AsyncEventEmitter<RestEvents> {
 				}
 			}
 
-			// Set the final body to the form data
-			finalBody = formData;
+			// Set the final body to the form data.
```
ckohen, line 385: This PR may not be the right place to tackle this, but I think maybe we need to replace RequestInit with RequestInitLike (Like ResponseLike) given we actually populate fetchOptions with very specific fields and *always* pass global formData when we have formData. If it had been this way in the first place we likely would have caught that undici formdata needed seperate handling in the strategy earlier.

ckohen (dismissed):

packages/rest/src/lib/REST.ts
almeidx: I'd rather leave this refactor to a follow-up pull request.

packages/rest/src/strategies/undiciRequest.ts
almeidx: Done in d3e6ab15e1d6a536252a736f6efc490b5c8c74ca

packages/rest/src/lib/handlers/Shared.ts
almeidx: Done in d3e6ab15e1d6a536252a736f6efc490b5c8c74ca

almeidx:
Tests should be fixed with the release of:
- https://github.com/nodejs/undici/pull/5650 (still pending)

iCrawl (approved):

packages/rest/src/lib/REST.ts
almeidx: https://github.com/discordjs/discord.js/issues/11603

PR #11544, opened 2026-05-26
feat(ws): add useNativeWebSocket option to WebSocketManager

### LINKED ISSUE #10109: Add support for native NodeJS WebSocket

discord.js

### Feature

NodeJS will be enabling the `WebSocket` class by default in NodeJS 22: https://github.com/nodejs/node/pull/51594 and is already available in NodeJS 21 as experimental. This would make the dependency on `ws` package optional or allow removing entirely. 

I will say that in my own usage of it, there are some discrepancies/missing properties between the ws package and the native, so I'm waiting for the @types/node update so that it is very clear what the differences are.

Similar to how you can choose to use ffmpeg/avconv, tweetnacl/libsodium, etc. it would be nice to be able to use the `WebSocket` provided by node instead of a 3rd party version.

Something like: 

```js
const ws = options.useNativeWebSocket && WebSocket ? WebSocket : require('ws');

 ...
 let wsInstance = new ws('wss://...');
```
or just
```js
let ws = null;
try {
    ws = require('ws');
 }
 catch (ex) {
     if (typeof WebSocket !== 'undefined') { 
         ws = WebSocket;
     }
     else {
         throw new Error("No Websocket class found");
     }
 }
 
 ...
 let wsInstance = new ws('wss://...');
 ```

_No response_

### Other context

_No response_

### PR DESCRIPTION
**Pull request by freepinkg**
## Summary

Adds a `useNativeWebSocket` option to `WebSocketManagerOptions` (default `false`) that lets users opt into using `globalThis.WebSocket` (native Node.js WebSocket, stable since Node 22.12.0) instead of the `ws` package.

When set to `true` and the native WebSocket is not available, a clear error is thrown explaining the requirement.

## Related Issue

Closes #10109

## Changes

- `packages/ws/src/ws/WebSocketManager.ts` — added `useNativeWebSocket` option to `OptionalWebSocketManagerOptions`
- `packages/ws/src/utils/constants.ts` — added default `useNativeWebSocket: false`
- `packages/ws/src/strategies/context/IContextFetchingStrategy.ts` — added `useNativeWebSocket` to the fetching strategy options
- `packages/ws/src/ws/WebSocketShard.ts` — uses the option to choose between native WebSocket and `ws` package; throws descriptive error if native WebSocket is not available

### DIFF
--- a/packages/ws/src/utils/constants.ts
+++ b/packages/ws/src/utils/constants.ts
@@ -58,7 +58,6 @@ export const DefaultWebSocketManagerOptions = {
 	get useNativeWebSocket() {
 		return shouldUseGlobalFetchAndWebSocket();
 	},
-	buildWebSocket: undefined as any,
 	retrieveSessionInfo(shardId) {
 		const store = getDefaultSessionStore();
 		return store.get(shardId) ?? null;
--- a/packages/ws/src/utils/constants.ts
+++ b/packages/ws/src/utils/constants.ts
@@ -1,6 +1,6 @@
 import process from 'node:process';
 import { Collection } from '@discordjs/collection';
-import { lazy } from '@discordjs/util';
+import { lazy, shouldUseGlobalFetchAndWebSocket } from '@discordjs/util';
 import { APIVersion, GatewayOpcodes } from 'discord-api-types/v10';
 import { SimpleShardingStrategy } from '../strategies/sharding/SimpleShardingStrategy.js';
 import { SimpleIdentifyThrottler } from '../throttling/SimpleIdentifyThrottler.js';
@@ -55,7 +55,9 @@ export const DefaultWebSocketManagerOptions = {
 	encoding: Encoding.JSON,
 	compression: null,
 	useIdentifyCompression: false,
-	useNativeWebSocket: false,
+	get useNativeWebSocket() {
+		return shouldUseGlobalFetchAndWebSocket();
+	},
 	retrieveSessionInfo(shardId) {
 		const store = getDefaultSessionStore();
 		return store.get(shardId) ?? null;
@@ -71,7 +73,7 @@ export const DefaultWebSocketManagerOptions = {
 	handshakeTimeout: 30_000,
 	helloTimeout: 60_000,
 	readyTimeout: 15_000,
-} as const satisfies Omit<OptionalWebSocketManagerOptions, 'fetchGatewayInformation' | 'token'>;
+} satisfies Omit<OptionalWebSocketManagerOptions, 'fetchGatewayInformation' | 'token'>;
 
 export const ImportantGatewayOpcodes = new Set([
 	GatewayOpcodes.Heartbeat,
--- a/packages/ws/src/utils/constants.ts
+++ b/packages/ws/src/utils/constants.ts
@@ -55,6 +55,7 @@ export const DefaultWebSocketManagerOptions = {
 	encoding: Encoding.JSON,
 	compression: null,
 	useIdentifyCompression: false,
+	useNativeWebSocket: false,
 	retrieveSessionInfo(shardId) {
 		const store = getDefaultSessionStore();
 		return store.get(shardId) ?? null;
--- a/packages/ws/src/utils/constants.ts
+++ b/packages/ws/src/utils/constants.ts
@@ -73,7 +73,7 @@ export const DefaultWebSocketManagerOptions = {
 	handshakeTimeout: 30_000,
 	helloTimeout: 60_000,
 	readyTimeout: 15_000,
-} satisfies Omit<OptionalWebSocketManagerOptions, 'fetchGatewayInformation' | 'token'>;
+} as const satisfies Omit<OptionalWebSocketManagerOptions, 'fetchGatewayInformation' | 'token'>;
 
 export const ImportantGatewayOpcodes = new Set([
 	GatewayOpcodes.Heartbeat,
--- a/packages/ws/src/utils/constants.ts
+++ b/packages/ws/src/utils/constants.ts
@@ -58,6 +58,7 @@ export const DefaultWebSocketManagerOptions = {
 	get useNativeWebSocket() {
 		return shouldUseGlobalFetchAndWebSocket();
 	},
+	buildWebSocket: undefined as any,
 	retrieveSessionInfo(shardId) {
 		const store = getDefaultSessionStore();
 		return store.get(shardId) ?? null;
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

**Comment by Qjuh** [comment]
This is a regression for all non-node runtimes that already use native websocket without any need for a flag. There's already a roadmap to use native websocket in node once it's proven to be stable enough.

**Review by didinele — CHANGES_REQUESTED** [review]
The Coderabbit review does need to be addressed in some way, we cannot leave the magic constant there.

Additionally, while I suppose I am generally in favor of being able to opt in or out of this behavior regardless of your environment, as Qjuh pointed out, the brute force approach of defaulting to `false` is detrimental. Fortunately, this is rather easy to address. You have 2 options:

1. Don't provide a default for `useNativeWebSocket`. This requires a bit of annoying refactoring due to how options are typed and passed around. I'd advise against this approach.
2. `useNativeWebSocket` can just be a getter in the `DefaultWebSocketManager` object - using `shouldUseGlobalFetchAndWebSocket` - I believe this will work as intended when the spread occurs:

https://github.com/discordjs/discord.js/blob/3d6121589f9c0d91f7cf4976307e8be07053a277/packages/ws/src/ws/WebSocketManager.ts#L274-L277

Image

File: packages/ws/src/ws/WebSocketShard.ts [hunk 1]
```diff
@@ -84,9 +84,20 @@ export interface SendRateLimitState {
...
-	? (globalThis as any).WebSocket
-	: WebSocket;
+function getWebSocketConstructor(useNative: boolean): typeof WebSocket {
+	if (useNative) {
+		if (typeof globalThis.WebSocket === 'undefined') {
+			throw new Error(
+				'useNativeWebSocket is set to true, but the native WebSocket is not available in this environment. ' +
+					'Use Node.js 22.12.0 or later, or set the --experimental-websocket flag on Node.js 21.',
+			);
+		}
```
vladfrangu, line 94:
```suggestion
		if (typeof globalThis.WebSocket === 'undefined') {
			throw new Error(
				'useNativeWebSocket is set to true, but the native WebSocket is not available in this environment. ' +
					'Use Node.js 22.12.0 or later, or set the "--experimental-websocket" flag on previous Node.js versions.',
			);
		}
```

packages/ws/src/ws/WebSocketManager.ts
```diff
@@ -194,6 +194,12 @@ export interface OptionalWebSocketManagerOptions {
 	 * @defaultValue `false`
 	 */
 	useIdentifyCompression: boolean;
+	/**
+	 * Whether to use the native Node.js WebSocket (`globalThis.WebSocket`) instead of the `ws` package
+	 *
+	 * @defaultValue `false`
```
didinele: needs updating

packages/ws/src/ws/WebSocketShard.ts
vladfrangu: Also discussion topic: should we fallbackk to the npm ws module if globalThis.WebSocket is somehow missing?

packages/ws/src/utils/constants.ts
```diff
@@ -70,7 +73,7 @@ export const DefaultWebSocketManagerOptions = {
 	handshakeTimeout: 30_000,
 	helloTimeout: 60_000,
 	readyTimeout: 15_000,
-} as const satisfies Omit<OptionalWebSocketManagerOptions, 'fetchGatewayInformation' | 'token'>;
+} satisfies Omit<OptionalWebSocketManagerOptions, 'fetchGatewayInformation' | 'token'>;
```
didinele: Perhaps I'm missing something but I don't get why the `as const` was removed?

From a [quick test](https://www.typescriptlang.org/play/?#code/MYewdgzgLgBAIgUwGYEMCuAbKEYF4YDeAUAJABOCA5pimQFwwDkokUKYUjANKZQrHyhQEZAMLho7KAAoAlIVLl+aMmBgBZFFAAWAOjLsAJiAC2cgNykAvkSswUOFtEtA) it seems that getter props don't automatically disallow you from using `as const`.

didinele (changes requested):

packages/ws/src/ws/WebSocketShard.ts
didinele: If by somehow you mean something else other than bad node version, seems a bit out of reach to account for.. I don't know, `delete globalThis.WebSocket`.

packages/ws/src/utils/constants.ts
```diff
@@ -55,6 +55,9 @@ export const DefaultWebSocketManagerOptions = {
 	encoding: Encoding.JSON,
 	compression: null,
 	useIdentifyCompression: false,
+	get useNativeWebSocket() {
```
almeidx, line 58: thoughts on making an option to build the websocket instance instead? i think it would be more flexible
like `buildWebSocket: (url: string) => WebSocket`

for the default, we could keep the same logic as we currently have i guess

freepinkg:
@almeidx great suggestion! I've added `buildWebSocket` as a factory option — when provided, it takes precedence over `useNativeWebSocket`. Users can pass any custom WebSocket implementation:

```ts
const manager = new WebSocketManager({
  token: process.env.DISCORD_TOKEN,
  intents: 0,
  fetchGatewayInformation() { /* ... */ },
  buildWebSocket(url) {
    return new WebSocket(url); // custom or native
  },
});
```

The `useNativeWebSocket` boolean still works as a simpler opt-in for native WebSocket when you don't need a custom implementation. Thoughts?

packages/ws/src/utils/constants.ts
didinele: I'm not against it necessarily, but it's bit problematic on the semantics. The PR author (more like the LLM they're using) also completely failed to consider that the whole purpose of this code they've been messing with that "converts" the manager options is specifically so IPC-friendly objects can be created. Passing around callbacks is not possible in most of those contexts, including the one we support natively (worker_threads).

As such, this option specifically would need to be a strategy option. I guess plainly in the constructor of the default simple strategy, while for workers, it's something you pass into the bootstrapper.

Do we feel strongly that this is particularly useful? If not, I'd rather we didn't go through with it.

almeidx: sounds much more complex. so i'd say no, it is not worth it

freepinkg:
@didinele @almeidx fair point on the serialization issue with worker_threads — reverted the `buildWebSocket` option. The PR now keeps just `useNativeWebSocket: boolean` as originally proposed, with auto-detection via `shouldUseGlobalFetchAndWebSocket()` as the default.

PR #11547, opened 2026-05-28
fix: Check left guilds & clear expected guilds

### LINKED ISSUE #11546: Left guilds remain in the cache during sharding reconnection

discord.js

### Issue description

During sharding reconnection, if the bot leaves the guild, the guild will remain in the cache when the session is resumed.
I discovered this bug back on February 27, 2026, but because I was using my library, I only fixed this bug for myself.

### Steps:
1. Wait for the ClientReady event
2. Turn off the Internet
3. Remove the app from the guild
4. Turn on the Internet
5. Wait for the ClientReady event

### Code sample

```typescript
client.on(Events.ClientReady, (client) => {
    console.log(`Ready. ${client.guilds.cache.size} guilds`)
})
client.on(Events.GuildCreate, (guild) => {
    console.log(`Guild create ${guild.name}`)
})
client.on(Events.GuildDelete, (guild) => {
    console.log(`Guild delete ${guild.name}`)
})
```

### Versions

Nodejs v24.15.0
Discord.js dev, see below

### Issue priority

Medium (should be fixed soon)

No Partials

Guilds

3d6121589

vakiliner:
This PR adds a check for cached guilds when receiving a gateway READY event.
If a cached guild is no longer available (i. e., has left) and is not sent in a gateway READY event, it is removed from the cache.
The `expectedGuilds` has been refactored and is now split into shards.

Fix #11546


Qjuh (commented):
The logic for the calling the GuildDelete action if missing from READY event looks fine, but I disagree with the other change since I don't see a benefit from separating the expectedGuilds by shardId.

vakiliner:
> but I disagree with the other change since I don't see a benefit from separating the expectedGuilds by shardId

This is necessary if the connection suddenly breaks when not all expected guilds have been received in `GUILD_CREATE`.
If the bot leaves the guilds it's expecting in `GUILD_CREATE`, the next time it reconnects, the expected guild ID won't be cleared from the cache, which is not expected.

vakiliner:
> but I disagree with the other change since I don't see a benefit from separating the expectedGuilds by shardId.

This bug doesn't occur in v14, as the cache is cleared in `WebSocketShard.js`.
https://github.com/discordjs/discord.js/blob/8047fe7c55d99ed01dfe82b4a260be45081baf26/packages/discord.js/src/client/websocket/WebSocketShard.js#L146

Qjuh:
> This bug doesn't occur in v14, as the cache is cleared in `WebSocketShard.js`.

Ah, that explains the need for the split, thank you. But that sounds like the other half either can't happen or would exist on v14 anyway, as v14 doesn't do those artificial GUILD_DELETE events on reconnect.

Guess it needs someone to test if the issue even happens on v14 / if only the split is implemented in v15.

vakiliner:
> Ah, that explains the need for the split, thank you. But that sounds like the other half either can't happen or would exist on v14 anyway, as v14 doesn't do those artificial GUILD_DELETE events on reconnect.
> 
> Guess it needs someone to test if the issue even happens on v14 / if only the split is implemented in v15.

I can do this.
To more easily reproduce the issue, you can ignore the following events after `READY` when you first connect, leave the guild that was expected, then reconnect (not resume), and observe the results.
Regarding v14, I meant that there is no error with the set of expected guilds, but there is an error with cached left guilds.

vakiliner:
So, I ran some testing.
In v14, expectedGuilds are ok, cached left guilds are not.
In v15, there are errors in both expectedGuilds and left guilds.
There are no errors in this pull request.

~~However, I did find one not-so-good behavior.
https://github.com/vakiliner/discord.js/blob/9e1dd2e522e54d24de923d55f0d57cec10e2053a/packages/discord.js/src/client/websocket/handlers/READY.js#L32
There's no check here that the guild was actually retrieved in `GUILD_CREATE`.
Should I emit partial guilds or omit them?~~
Okay, I mistakenly assumed that this was a partial guild, when in fact it is an unavailable guild.

Qjuh:
> There's no check here that the guild was actually retrieved in `GUILD_CREATE`.

Afaict the only case that would be relevant is if Client is used without `Guilds` intent. That's already an issue as many of discord.js methods rely on that intent. But you could check for the intent being present before deleting.

vakiliner:
> But you could check for the intent being present before deleting.

If connect without the `Guilds` intent, then the `GuildCreate` and `GuildDelete` events are not provided, therefore I believe that it is possible not to emit guild delete event.

image

packages/discord.js/src/client/actions/GuildDelete.js
```diff
@@ -12,13 +12,15 @@ class GuildDeleteAction extends Action {
...
     client.guilds.cache.delete(guild.id);
 
-    /**
-     * Emitted whenever a guild kicks the client or the guild is deleted/left.
-     *
-     * @event Client#guildDelete
-     * @param {Guild} guild The guild that was deleted
-     */
-    client.emit(Events.GuildDelete, guild);
+    if (guild.name !== undefined) {
```
Qjuh: This check makes no sense. It would prevent guilds to be removed that were unavailable before but now the bot is not in it anymore.

Check `client.options.intents.has(GatewayIntentBits.Guilds)` in READY.js instead to only call this action if we had the intent that could add it over gateway before. If not then they were manually fetched and the developer is responsible for keeping cache accurate.

Qjuh (changes requested):

packages/discord.js/src/client/actions/GuildDelete.js
vakiliner: > It would prevent guilds to be removed that were unavailable before but now the bot is not in it anymore.

That's right, that's exactly why I added a check to prevent partial guilds from emitting.

Qjuh: But you're doing the opposite: you still remove them but stop the event from emitting so the end-user wouldn't be informed about it.

Qjuh: Anything you write in GuildDeleteAction applies to real GUILD_DELETE payloads too. So changing that behavior is risky to not introduce new bugs. Especially if there's no need to do it

vakiliner: I'm blocking event generation for partial guilds. This applies to both the presence and absence of the `Guilds` intent.
I still don't understand the answer. Can I generate a delete event for partial guilds?

Qjuh: There are no partial guilds. There are unavailable guilds though, but the bot can still be removed from those and they'd emit a GUILD_DELETE and that should propagate fully to cache remove and event emit, yes.

vakiliner: > There are no partial guilds. There are unavailable guilds though

Ah, sorry.
I mixed up the wording.
I meant unavailable guilds.

packages/discord.js/src/client/websocket/handlers/READY.js
```diff
@@ -14,12 +14,25 @@ module.exports = (client, { d: data }, shardId) => {
...
   }
 
+  if (expectedGuilds.size) {
+    client.expectedGuilds.set(shardId, expectedGuilds);
+  } else {
+    client.expectedGuilds.delete(shardId);
+  }
+
+  for (const guild of client.guilds.cache.values()) {
+    if (guild.shardId === shardId && !expectedGuilds.has(guild.id)) {
```
Qjuh, line 31:
```suggestion
    if (guild.shardId === shardId && !expectedGuilds.has(guild.id) && client.options.intents.has(GatewayIntentBits.Guilds)) {
```
only remove guilds from cache if the client could have received them over gateway in the first place. Else this would clear the cache on every reconnect.

Qjuh (changes requested):

packages/discord.js/src/client/websocket/handlers/READY.js
vakiliner: > Else this would clear the cache on every reconnect.

The cache won't be cleared on every reconnection. Although the absence of the `Guilds` intent doesn't provide `GUILD_CREATE`, the `READY` event provides a list of unavailable guild IDs.

Qjuh: I see, you're right. I had the misconception that `guilds` won't be provided in READY event without GUILDS intent, but that's apparently wrong.

Qjuh (approved):

PR #11548, opened 2026-05-30
ci: harden actions

hadez8877:
## What's Added?

- This PR adds [zizmor](https://docs.zizmor.sh/) to our CI to check for incorrect usage of our workflows. As for now, I tuned the tool to target high confidence and high severity


Qjuh:
I'm sorry, but in my opinion adding additional workflow dependencies proposed by a first-time contributor is the opposite of "hardening actions".

PR #11541, opened 2026-05-25
fix(ShardingManager): default api url

almeidx:
`RouteBases.api` already includes the API version in the URL, which was resulting in `https://discord.com/api/v10/v10`

Regression from #11471

PR #11549, opened 2026-05-31
fix: remove duplicate API version in fetchRecommendedShardCount URL

ErnestHysa:
Fixes #11541

RouteBases.api already includes the API version (e.g., https://discord.com/api/v10), so appending /v${version} was causing a double version in the URL. This fixes the URL construction in fetchRecommendedShardCount.


Qjuh:
Duplicate of #11541

PR #11555, opened 2026-06-17
build: Bump discord-api-types to 0.38.50

Jiralite:


Jiralite:
Not sure how that structures test passed in the past, but fixed it now.

almeidx (approved):

sdanialraza (approved):

packages/structures/__tests__/message.test.ts
```diff
@@ -197,7 +196,7 @@ describe('message with components', () => {
 			url: 'attachment://file.txt',
 			attachment_id: '0',
 			content_type: 'text/plain',
-			flags: 0,
+			flags: UnfurledMediaItemFlags.IsAnimated,
```
Qjuh: The test should pass with `flags: 0` too. If it doesn't that's a bug and shouldn't be fixed by changing the test.

Qjuh (dismissed):

packages/structures/__tests__/message.test.ts
Jiralite: From investigating, I believe this is the correct fix. discord-api-types changed this from `number` to `UnfurledMediaItemFlags` in https://github.com/discordjs/discord-api-types/pull/1609.

Jiralite: Ignored.

vladfrangu (approved):

PR #11560, opened 2026-06-24
fix: detect spoiler attachments with flag

advaith1:
Spoiler attachments do not necessarily have the `SPOILER_` filename prefix, and most will not going forward. The `IS_SPOILER` flag should be used to detect spoilers instead.


Jiralite:
Is this flag present on all prior messages to signify they are spoilers?

advaith1:
yes, it's returned on all attachments that have the SPOILER_ prefix so you don't need to manually check for the prefix.

Jiralite (approved):

almeidx (approved):

vladfrangu (approved):

PR #11564, opened 2026-07-02
fix(util): cache nullish values in lazy()

spokodev:
## Please describe the changes this PR makes and why it should be merged:

`lazy()` in `@discordjs/util` did `defaultValue ??= cb()`. Since `??=` only assigns when the current value is nullish, a callback that returns `null` or `undefined` is never cached and re-runs on every call, breaking the utility's documented compute-once contract. Any side effect in the callback repeats, and the memoization is defeated for exactly those return values.

Non-nullish falsy values (`0`, `''`, `false`) cache correctly; only `null` and `undefined` trigger the bug.

The fix tracks a `called` boolean flag instead of relying on the cached value being non-nullish, so the callback runs at most once regardless of what it returns.

## Test

Added two cases to `packages/util/__tests__/lazy.test.ts` for callbacks returning `undefined` and `null`. Both fail before the change (`vi.fn()` called 2 times, expected once) and pass after; the `lazy` unit suite stays green.

## Status and versioning classification:
- Code changes have been tested, or there are no code changes


Qjuh:
This is a non-issue. `lazy()` is explicitly meant for computing values once that take some resources to compute. Those will never result in `null` or `undefined`, even less in the cases this gets used in djs: import/require or creating a Collection.

spokodev:
Fair, and you are right that inside discord.js (import/require, Collections) the callback never returns nullish, so it is a non-issue there. My angle was the published `@discordjs/util`: `lazy<Value>` accepts any callback, so a consumer doing `lazy(() => maybeUndefined())` gets it re-run every call, breaking compute-once and repeating side effects. The fix is a one-boolean change. If you consider `lazy()` intended only for non-nullish values, that is a reasonable line and I am fine closing. Your call.

almeidx (approved):
sure

Qjuh (approved):

Jiralite (approved):

PR #11568, opened 2026-07-04
fix(website): fall back to main docs version when version lookup fails

### LINKED ISSUE #11558: discordjs.dev "Docs" button leads to a 404 page
### Which application is this bug report for?

Documentation

### Issue description

Clicking the **Docs** button on the https://discordjs.dev homepage opens: https://discordjs.dev/docs/packages/discord.js

This page returns a **404**.

The documentation is available when a version is specified, for example: https://discordjs.dev/docs/packages/discord.js/main

However, the same navigation works correctly on https://discord.js.org, where clicking **Docs** resolves to the documentation without returning a 404.

I'm not sure whether discordjs.dev is still intended to be fully supported or whether it should redirect to discord.js.org, but the current behavior between the two official domains is inconsistent.

### Steps to Reproduce

1. Visit https://discordjs.dev
2. Click the **Docs** button.
3. Observe that https://discordjs.dev/docs/packages/discord.js returns a 404.

For comparison:
1. Visit https://discord.js.org
2. Click the **Docs** button.
3. The documentation opens correctly.

### Versions

- Google Chrome 149.0.7827.156
- Windows 11 25H2

### Issue priority

Medium (should be fixed soon)

PineFruitDev:

Fixes #11558.

The Docs button on the homepage links to `/docs`. The middleware redirects that to `/docs/packages/discord.js/${latestVersion}`, where the version comes from a D1 query in `fetchLatestVersion`. When that query fails, the catch block returns an empty string, so the redirect lands on the version-less URL `/docs/packages/discord.js`. No route exists at that level (only `[packageName]/[version]`), so the result is a guaranteed 404. This matches the exact URL in the report.

It also explains why the two domains behave differently: both sites run the same code, but the discordjs.dev deployment appears unable to reach the versions database while discord.js.org can. That part is infra side and not something this PR can address.

This change makes the failure path fall back to `main`, mirroring the existing `IS_LOCAL_DEV` branch and the success path's own `?? 'main'` default. The reporter confirmed the `/main` URL resolves correctly. `fetchLatestVersion` has exactly one caller (the middleware) and nothing depends on the empty string return, so this also fixes `/stable` links breaking under the same failure condition. With this fallback in place the Docs button can never produce a version-less 404 regardless of deployment state.

- [x] Code changes have been tested against the Discord API, or there are no code changes (website app only, no Discord API interaction)
- [x] This PR changes `apps/website` only, which is not a published package, so no semver classification applies


PineFruitDev:
Deliberately kept the diff minimal since this fixes a live 404; the pattern also exists in the pre-existing IS_LOCAL_DEV branch, so I'd rather not expand scope here. Happy to extract a helper in this PR if maintainers prefer.

niridya:
Thanks for looking into this and for the detailed investigation! That explanation makes perfect sense, and the fallback to `main` looks like a solid fix.

PineFruitDev:
@Jiralite Anything needed from my side to move this forward?

Qjuh:
This is not really a fix to the issue. The real fix is one of two things:
- not use discordjs.dev at all and use discord.js.org instead
- make sure discordjs.dev can access the production db the same way discord.js.org does

I'm pretty sure we're going the first route, discordjs.dev was never meant as production URL and is just for beta version of docs site.

PR #11569, opened 2026-07-06
fix(modal-conflict): don't leak resolved data across component types

### LINKED ISSUE #11535: Conflict between `FileUploadModalData` and `SelectMenuModalData`

discord.js

### Issue description

> *The examples below reference to the data of a `UserSelect` menu for visualization but also apply to the data of `RoleSelect`, `ChannelSelect` and `MentionableSelect` (especially its individual map properties).*

If [`SelectMenuModalData`](https://discord.js.org/docs/packages/discord.js/14.26.4/SelectMenuModalData:Interface) has values, [`FileUploadModalData`](https://discord.js.org/docs/packages/discord.js/14.26.4/FileUploadModalData:Interface) will have an empty `users` and `members` map supplied. If `FileUploadModalData` has values, `SelectMenuModalData` will have an empty `attachments` map supplied. If any of them has no values at all, then the maps won't the present although docs state the `attachment` property is always present for `FileUploadModalData`, even if empty. This currently causes issues when trying to access the property but results in being `undefined`.

---

The data is retrieved using the `getField()` method on `ModalSubmitFields`:
```ts
const field = ctx.interaction.fields.getField(id)
```

## Scenario 1 (`FileUpload` without values)
`SelectMenu` response with values:
<sub>(Expected)</sub>
```js
{
  type: 5,
  id: 2,
  customId: 'user',
  values: [ '729343563401265193' ],
  users: Collection(1) [Map] {...},
  members: Collection(1) [Map] {...}
}
```

`FileUpload` response without values: 
<sub>(No empty `attachments` map present, although marked as always present even if empty. Suddenly has an empty `users` and `members` map, not documented for `FileUploadModalData`.)</sub>
```js
{
  type: 19,
  id: 4,
  customId: 'upload',
  values: [],
  users: Collection(0) [Map] {},
  members: Collection(0) [Map] {}
}
```

## Scenario 2 (`FileUpload` **with** values)
`SelectMenu` response with values:
<sub>(Suddenly has an empty `attachments` map, not documented for `SelectMenuModalData`.)</sub>
```js
{
  type: 5,
  id: 2,
  customId: 'user',
  values: [ '729343563401265193' ],
  users: Collection(1) [Map] {...},
  members: Collection(1) [Map] {...},
  attachments: Collection(0) [Map] {}
}
```

`FileUpload` response with values:
<sub>(Now `attachments` map is present, but has an empty `users` and `members` map which is not documented for `FileUploadModalData`.)</sub>
```js
{
  type: 19,
  id: 4,
  customId: 'upload',
  values: [ '1505975523967369267' ],
  users: Collection(0) [Map] {},
  members: Collection(0) [Map] {},
  attachments: Collection(1) [Map] {...}
}
```

## Scenario 3 (no values)
`SelectMenu` response:
<sub>(`users` and `members` map missing)</sub>
```js
{ type: 5, id: 2, customId: 'user', values: [] }
```

`FileUpload` response:
<sub>(`attachments` map missing)</sub>
```js
{ type: 19, id: 4, customId: 'upload', values: [] }
```

---

## Steps To Reproduce
- Create a modal with a file upload and any select menu component as fields.
- Log the returned raw data to the console.
- Submit the modal with or without specified values. 
- Take a look at the console output.
- Repeat by submitting the modal again with different values.

### Code sample

```typescript

```

### Versions

- discord.js `v14.26.4`

### Issue priority

Low (slightly annoying)

Not applicable

Not applicable

_No response_

tiagodefendi:
## Problem

Modals with multiple resolved component types (e.g. `FileUpload` + `UserSelect`) leaked properties into each other: a `FileUploadModalData` would show up with empty `users`/`members`, and a `SelectMenuModalData` would show up with an empty `attachments`. On top of that, `attachments` disappeared entirely from `FileUploadModalData` when nothing was selected in the modal, even though the typings already promised this property as always present.

Fixes #11535

## What was done

- `ModalSubmitInteraction#transformComponent` now gates `users`/`members`/`roles`/`channels`/`attachments` by each component's actual `type`, instead of blindly applying whatever exists on the interaction's shared `resolved` object.
- `attachments` is now always initialized as a `Collection` (even empty) on `FileUpload` components, even when `resolved` is absent.
- `ModalComponentResolver#getUploadedFiles` updated to keep honoring its `required`/`null` contract now that `attachments` is never `undefined`.

## How it was tested

- `tsc --noEmit`, `tsd`, `eslint`, and `prettier` all passing.
- Manual runtime verification reproducing the issue's scenarios: mixed components in the same modal, `FileUpload` with nothing selected, and all 4 `required`/empty cases for `getUploadedFiles`. Also validated `MentionableSelect`/`RoleSelect`/`ChannelSelect`/`StringSelect` in isolation.
- No new automated tests added — the `discord.js` package has no runtime test suite, only `tsc`/`tsd`.

### DIFF
--- a/packages/discord.js/src/structures/ModalSubmitInteraction.js
+++ b/packages/discord.js/src/structures/ModalSubmitInteraction.js
@@ -2,6 +2,7 @@
 
 const { Collection } = require('@discordjs/collection');
 const { lazy } = require('@discordjs/util');
+const { ComponentType } = require('discord-api-types/v10');
 const { transformResolved } = require('../util/Util.js');
 const { BaseInteraction } = require('./BaseInteraction.js');
 const { InteractionWebhook } = require('./InteractionWebhook.js');
@@ -31,7 +32,7 @@ const getAttachment = lazy(() => require('./Attachment.js').Attachment);
  * @typedef {BaseModalData} FileUploadModalData
  * @property {string} customId The custom id of the file upload
  * @property {Snowflake[]} values The values of the file upload
- * @property {Collection<Snowflake, Attachment>} [attachments] The resolved attachments
+ * @property {Collection<Snowflake, Attachment>} attachments The resolved attachments
  */
 
 /**
@@ -180,11 +181,20 @@ class ModalSubmitInteraction extends BaseInteraction {
 
     if (rawComponent.values) {
       data.values = rawComponent.values;
+
+      const isUserSelect = [ComponentType.UserSelect, ComponentType.MentionableSelect].includes(rawComponent.type);
+      const isRoleSelect = [ComponentType.RoleSelect, ComponentType.MentionableSelect].includes(rawComponent.type);
+      const isChannelSelect = rawComponent.type === ComponentType.ChannelSelect;
+      const isFileUpload = rawComponent.type === ComponentType.FileUpload;
+
+      // Attachments must always be present (even empty) on FileUpload components.
+      if (isFileUpload) data.attachments = new Collection();
+
       if (resolved) {
         const { members, users, channels, roles, attachments } = resolved;
         const valueSet = new Set(rawComponent.values);
 
-        if (users) {
+        if (isUserSelect && users) {
           data.users = new Collection();
 
           for (const [id, user] of Object.entries(users)) {
@@ -194,7 +204,7 @@ class ModalSubmitInteraction extends BaseInteraction {
           }
         }
 
-        if (channels) {
+        if (isChannelSelect && channels) {
           data.channels = new Collection();
 
           for (const [id, apiChannel] of Object.entries(channels)) {
@@ -204,7 +214,7 @@ class ModalSubmitInteraction extends BaseInteraction {
           }
         }
 
-        if (members) {
+        if (isUserSelect && members) {
           data.members = new Collection();
 
           for (const [id, member] of Object.entries(members)) {
@@ -215,7 +225,7 @@ class ModalSubmitInteraction extends BaseInteraction {
           }
         }
 
-        if (roles) {
+        if (isRoleSelect && roles) {
           data.roles = new Collection();
 
           for (const [id, role] of Object.entries(roles)) {
@@ -225,8 +235,7 @@ class ModalSubmitInteraction extends BaseInteraction {
           }
         }
 
-        if (attachments) {
-          data.attachments = new Collection();
+        if (isFileUpload && attachments) {
           for (const [id, attachment] of Object.entries(attachments)) {
             if (valueSet.has(id)) {
               data.attachments.set(id, new (getAttachment())(attachment));
--- a/packages/discord.js/src/structures/ModalSubmitInteraction.js
+++ b/packages/discord.js/src/structures/ModalSubmitInteraction.js
@@ -32,7 +32,7 @@ const getAttachment = lazy(() => require('./Attachment.js').Attachment);
  * @typedef {BaseModalData} FileUploadModalData
  * @property {string} customId The custom id of the file upload
  * @property {Snowflake[]} values The values of the file upload
- * @property {Collection<Snowflake, Attachment>} attachments The resolved attachments
+ * @property {Collection<Snowflake, Attachment>} [attachments] The resolved attachments
  */
 
 /**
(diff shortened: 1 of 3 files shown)


packages/discord.js/src/structures/ModalSubmitInteraction.js
```diff
@@ -31,7 +32,7 @@ const getAttachment = lazy(() => require('./Attachment.js').Attachment);
  * @typedef {BaseModalData} FileUploadModalData
  * @property {string} customId The custom id of the file upload
  * @property {Snowflake[]} values The values of the file upload
- * @property {Collection<Snowflake, Attachment>} [attachments] The resolved attachments
+ * @property {Collection<Snowflake, Attachment>} attachments The resolved attachments
```
Qjuh: While the fix for the selectmenu components is correct, the attachments were correct in code and JSDoc and the typings aren't. So you're fixing the wrong side of it. This way it would be a breaking change which is not correct.

packages/discord.js/src/structures/ModalSubmitInteraction.js
```diff
@@ -180,11 +181,20 @@ class ModalSubmitInteraction extends BaseInteraction {
...
     if (rawComponent.values) {
       data.values = rawComponent.values;
+
+      const isUserSelect = [ComponentType.UserSelect, ComponentType.MentionableSelect].includes(rawComponent.type);
+      const isRoleSelect = [ComponentType.RoleSelect, ComponentType.MentionableSelect].includes(rawComponent.type);
+      const isChannelSelect = rawComponent.type === ComponentType.ChannelSelect;
+      const isFileUpload = rawComponent.type === ComponentType.FileUpload;
+
+      // Attachments must always be present (even empty) on FileUpload components.
+      if (isFileUpload) data.attachments = new Collection();
```
Qjuh: Also not correct, see comment on the issue this is supposed to fix.

packages/discord.js/src/structures/ModalSubmitInteraction.js
```diff
@@ -225,8 +235,7 @@ class ModalSubmitInteraction extends BaseInteraction {
           }
         }
 
-        if (attachments) {
-          data.attachments = new Collection();
```
Qjuh, line 229: This needs to stay here.

Qjuh (changes requested):

packages/discord.js/src/structures/ModalSubmitInteraction.js
tiagodefendi: Restored, just added the isFileUpload check to it so it still fixes the original leak issue without changing the presence/absence behavior

packages/discord.js/src/structures/ModalSubmitInteraction.js
tiagodefendi: Removed, that block wasn't correct as you pointed out. attachments is no longer force-populated regardless of resolved

packages/discord.js/src/structures/ModalSubmitInteraction.js
tiagodefendi: You're right, my mistake, reverted the JSDoc back to [attachments] (optional) and fixed typings/index.d.ts instead, since that was the side that was actually wrong. Thanks for catching that

tiagodefendi:
@Qjuh It's everything ok now?

Qjuh (approved):

packages/discord.js/src/structures/ModalSubmitInteraction.js
```diff
@@ -180,11 +181,17 @@ class ModalSubmitInteraction extends BaseInteraction {
 
     if (rawComponent.values) {
       data.values = rawComponent.values;
+
+      const isUserSelect = [ComponentType.UserSelect, ComponentType.MentionableSelect].includes(rawComponent.type);
+      const isRoleSelect = [ComponentType.RoleSelect, ComponentType.MentionableSelect].includes(rawComponent.type);
```
almeidx, line 186: These variable names are misleading. `resolvesUsers` / `resolvesRoles` or something along those lines would be more appropriate

packages/discord.js/src/structures/ModalSubmitInteraction.js
```diff
@@ -180,11 +181,17 @@ class ModalSubmitInteraction extends BaseInteraction {
 
     if (rawComponent.values) {
       data.values = rawComponent.values;
+
+      const isUserSelect = [ComponentType.UserSelect, ComponentType.MentionableSelect].includes(rawComponent.type);
```
almeidx, line 185: Lets move all these into the `if (resolved)` block

almeidx (changes requested):

Issue #11571, opened 2026-07-08
Wrong Deprecation Info


Claudiocli:
### Which application is this bug report for?

Documentation

### Issue description

While browsing the docs, I stumbled into the deprecation of [`ClientEvents.ready`](https://discord.js.org/docs/packages/discord.js/14.26.4/ClientEvents:Interface#ready), but its infographic isn't correct:

<img width="1285" height="172" alt="Image" src="https://github.com/user-attachments/assets/9f4b4624-7a78-4fd4-9b2c-99a451176fa8" />

It should state "Use `ClientEvents.clientReady` instead". It's slightly annoying just because it's not stated anywere if not in the source code

### Steps to Reproduce

1. visit [ClientEvent.ready](https://discord.js.org/docs/packages/discord.js/14.26.4/ClientEvents:Interface#ready)

### Versions

- Firefox 152.0.4

### Issue priority

Low (slightly annoying)

PR #11574, opened 2026-07-14
docs: fix incorrect casing in the ready event deprecation

wasteofwifi:
Found the bug in `packages/discord.js/typings/index.d.ts`. The deprecation comment on the `ready` event links to `ClientEvents.ClientReady` but the actual property is `clientReady` (lowercase c). Since that link target doesn't exist, the docs site's `{@link}` fails to resolve and just falls back to showing "ClientEvents" instead of the real message, which is the bug reported in #11571


Jiralite (approved):

Jiralite:
> which is the bug reported in #14575.

Can you double-check the issue?

jaxsonbunch:
Sorry, I linked the wrong issue. I double checked and this is for #11571

PR #11581, opened 2026-07-21
fix(voice): handle server update race condition

### LINKED ISSUE #11553: Bot gets stuck in signalling sometimes

voice

### Issue description

My discord bot works most of the time, but sometimes, it cant get past voice connection status signalling. 

Some context: its a tts bot, has a bullmq queue (just a fault tolerant queue) and one worker per guild

I deal with it right now by not serving the request (I am going to requeue the request with a high priority so it retries immediately) but wanted to know if there's something I am doing wrong or its just that sometimes connections won't move past signalling and I will have to retry the connection after the timeout. I want to make the tts instantaneous and retrying makes the ux worse because of the delay. This is not a one off issue and happens fairly regularly, making the bot unresponsive. 

If this is a bug I am happy to contribute to solve this issue as this affects a considerable amount of my users and is resulting in poorer user experience. On the other hand, If I'm doing something wrong, please let me know. 

Exact line which throws: https://github.com/parnavh/koala/blob/a03e7a8de90c6e07cc42d56219b6cbb57f6c711b/src/lib/voice/handler.ts#L139

Source code: https://github.com/parnavh/koala/

Error log sample:

```
Unable to connect to **redacted** in guild **redacted** voiceState: signalling
app-1      | AbortError: The operation was aborted
app-1      |     at EventTarget.abortListener (node:events:999:14)
app-1      |     at [nodejs.internal.kHybridDispatch] (node:internal/event_target:786:20)
app-1      |     at EventTarget.dispatchEvent (node:internal/event_target:721:26)
app-1      |     at abortSignal (node:internal/abort_controller:369:10)
app-1      |     at AbortController.abort (node:internal/abort_controller:403:5)
app-1      |     at Timeout.<anonymous> (file:///app/node_modules/.pnpm/@discordjs+voice@0.19.0_@discordjs+opus@0.10.0/node_modules/@discordjs/voice/dist/index.mjs:3056:39)
app-1      |     at listOnTimeout (node:internal/timers:569:17)
app-1      |     at process.processTimers (node:internal/timers:512:7) {
app-1      |   code: 'ABORT_ERR',
app-1      |   [cause]: DOMException [AbortError]: This operation was aborted
app-1      |       at new DOMException (node:internal/per_context/domexception:53:5)
app-1      |       at AbortController.abort (node:internal/abort_controller:401:18)
app-1      |       at Timeout.<anonymous> (file:///app/node_modules/.pnpm/@discordjs+voice@0.19.0_@discordjs+opus@0.10.0/node_modules/@discordjs/voice/dist/index.mjs:3056:39)
app-1      |       at listOnTimeout (node:internal/timers:569:17)
app-1      |       at process.processTimers (node:internal/timers:512:7)
app-1      | }
```

### Code sample

```typescript
if (connection.state.status !== VoiceConnectionStatus.Ready) {
  try {
    await entersState(connection, VoiceConnectionStatus.Ready, 15_000); // this line throws
  } catch (error) {
    console.log(
      `Unable to connect to ${options.channelId} in guild ${options.guildId} voiceState: ${connection.state.status}`,
    );
    console.error(error);
    connection.disconnect();
    connection.destroy();
    return;
  }
}
```

### Versions

- discord.js: 14.18.0
- discordjs/voice: 0.19.0
- hosted using docker node:18-alpine

### Issue priority

Medium (should be fixed soon)

No Partials

Guilds, GuildVoiceStates

_No response_

parnavh:
`VOICE_SERVER_UPDATE` can arrive before `VOICE_STATE_UPDATE` which causes the voice connection to stay stuck in signalling since `configureNetworking()` returns early without `this.packets.state` and is not called again after state is valid.

This pr fixes this by calling `configureNetowrking()` when `this.packets.server.endpoint` is set (implying that server update was received before state update)

Resolves #11553

### DIFF
--- a/packages/voice/src/VoiceConnection.ts
+++ b/packages/voice/src/VoiceConnection.ts
@@ -384,7 +384,7 @@ export class VoiceConnection extends EventEmitter {
 			stuck in signalling state
 			gh issue #11553
 		*/
-		if (packet.channel_id && this.packets.server?.endpoint) {
+		if (packet.channel_id && this.state.status === VoiceConnectionStatus.Signalling) {
 			this.configureNetworking();
 		}
 	}
--- a/packages/voice/src/VoiceConnection.ts
+++ b/packages/voice/src/VoiceConnection.ts
@@ -384,7 +384,7 @@ export class VoiceConnection extends EventEmitter {
 			stuck in signalling state
 			gh issue #11553
 		*/
-		if (this.packets.server?.endpoint) {
+		if (packet.channel_id && this.packets.server?.endpoint) {
 			this.configureNetworking();
 		}
 	}
--- a/packages/voice/src/VoiceConnection.ts
+++ b/packages/voice/src/VoiceConnection.ts
@@ -378,6 +378,15 @@ export class VoiceConnection extends EventEmitter {
 			as it may have disconnected due to network failure. This will be gracefully handled once the voice websocket
 			dies, and then it is up to the user to decide how they wish to handle this.
 		*/
+
+		/*
+		  deals with race condition when VOICE_SERVER_UPDATE arrives before VOICE_STATE_UPDATE and connection stays
+			stuck in signalling state
+			gh issue #11553
+		*/
+		if (this.packets.server?.endpoint) {
+			this.configureNetworking();
+		}
 	}
 
 	/**


parnavh:
The `this.packets.server?.endpoint` might be redundant as a check (since its checked for in `configureNetworking()`) but makes the purpose of the call explicit which is why I have kept it

packages/voice/src/VoiceConnection.ts
```diff
@@ -378,6 +378,15 @@ export class VoiceConnection extends EventEmitter {
 			as it may have disconnected due to network failure. This will be gracefully handled once the voice websocket
 			dies, and then it is up to the user to decide how they wish to handle this.
 		*/
+
+		/*
+		  deals with race condition when VOICE_SERVER_UPDATE arrives before VOICE_STATE_UPDATE and connection stays
+			stuck in signalling state
+			gh issue #11553
+		*/
+		if (packet.channel_id && this.packets.server?.endpoint) {
```
Qjuh:
```suggestion
		if (packet.channel_id && this.state.status === VoiceConnectionState.Signalling) {
```
maybe better check for state here. `configureNetworking()` already checks for endpoint property and will return early if not present anyway.

parnavh: Yeah good point. But won't the status be signalling always? since you can't transition further without the state packet? would it serve to guard against disconnects? but a check for channel_id is present which would be falsy if disconnected right?

Qjuh: State packets get sent for other cases too, not just when connecting. For example if the bot gets server muted by a moderator. And channel_id being `null` can mean several things, compare the comment directly above your changes in the existing code.
On the other hand state will always be Signalling after a new connection was made or a new joinVoiceChannel packet was sent.

parnavh: Oh alright! Sorry got confused

parnavh:
I'm not sure if this can be tested but can be a beneficial test. perhaps can be added to `packages/voice/__tests__/VoiceConnection.test.ts`. Willing to write some test and try to reproduce the bug without the patch, and also add tests to make sure this race condition doesn't show up again. would need a little guidance on how to control when server and state updates are sent

Qjuh (approved):

parnavh:
Hey! Is there something to be improved or something that's just wrong that needs to be fixed? @vladfrangu @iCrawl @kyranet @SpaceEEC 

Sorry for the bump, its been a while and I figured that this just slipped through

sdanialraza (approved):

PR #11584, opened 2026-07-23
feat(ws): add http agent support

Angelok963:


packages/ws/src/strategies/context/IContextFetchingStrategy.ts
```diff
@@ -38,6 +39,7 @@ export interface IContextFetchingStrategy {
 
 export async function managerToFetchingStrategyOptions(manager: WebSocketManager): Promise<FetchingStrategyOptions> {
 	return {
+		agent: manager.options.agent,
```
didinele, line 42: This doesn't work. Even the built in Worker strategy hard crashes on this, because agent cannot be serialized over the communication channels.

Unfortunately for this to be supported it needs to be a strategy-level option for the default one, and for workers it's probably something you pass into the bootstrap class.

I'm not guaranteeing that the PR will be merged in that state, but it will at least be correct.

didinele (changes requested):

Qjuh:
@Angelok963 do you intend to change this PR to work or can this be closed?

PR #11585, opened 2026-07-24
types(GuildScheduledEvent): allow scheduledEndTime to be null

PHV08:
Allow `scheduledEndTime` to accept `null` in both `setScheduledEndTime()` and `BaseGuildScheduledEventOptions` to support resetting/removing the end time of a scheduled event. 

Also updates JSDocs to document this behavior.

### DIFF
--- a/packages/discord.js/src/managers/GuildScheduledEventManager.js
+++ b/packages/discord.js/src/managers/GuildScheduledEventManager.js
@@ -193,7 +193,7 @@ class GuildScheduledEventManager extends CachedManager {
    * @typedef {Object} GuildScheduledEventEditOptions
    * @property {string} [name] The name of the guild scheduled event
    * @property {DateResolvable} [scheduledStartTime] The time to schedule the event at
-   * @property {DateResolvable} [scheduledEndTime] The time to end the event at
+   * @property {DateResolvable|null} [scheduledEndTime] The time to end the event at
    * @property {GuildScheduledEventPrivacyLevel} [privacyLevel] The privacy level of the guild scheduled event
    * @property {GuildScheduledEventEntityType} [entityType] The scheduled entity type of the event
    * @property {string} [description] The description of the guild scheduled event
--- a/packages/discord.js/src/managers/GuildScheduledEventManager.js
+++ b/packages/discord.js/src/managers/GuildScheduledEventManager.js
@@ -193,7 +193,8 @@ class GuildScheduledEventManager extends CachedManager {
    * @typedef {Object} GuildScheduledEventEditOptions
    * @property {string} [name] The name of the guild scheduled event
    * @property {DateResolvable} [scheduledStartTime] The time to schedule the event at
-   * @property {DateResolvable|null} [scheduledEndTime] The time to end the event at
+   * @property {?DateResolvable} [scheduledEndTime] The time to end the event at
+   * <info>This cannot be `null` if `entityType` is {@link GuildScheduledEventEntityType.External}</info>
    * @property {GuildScheduledEventPrivacyLevel} [privacyLevel] The privacy level of the guild scheduled event
    * @property {GuildScheduledEventEntityType} [entityType] The scheduled entity type of the event
    * @property {string} [description] The description of the guild scheduled event
--- a/packages/discord.js/src/structures/GuildScheduledEvent.js
+++ b/packages/discord.js/src/structures/GuildScheduledEvent.js
@@ -450,11 +450,10 @@ class GuildScheduledEvent extends Base {
     return this.edit({ scheduledStartTime, reason });
   }
 
-  // TODO: scheduledEndTime gets reset on passing null but it hasn't been documented
   /**
    * Sets a new time to end the event at.
    *
-   * @param {DateResolvable} scheduledEndTime The time to end the event at
+   * @param {DateResolvable|null} scheduledEndTime The time to end the event at, or null to reset/remove the end time
    * @param {string} [reason] The reason for changing the scheduled end time
    * @returns {Promise<GuildScheduledEvent>}
    * @example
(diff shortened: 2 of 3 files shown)


packages/discord.js/src/structures/GuildScheduledEvent.js
```diff
@@ -450,11 +450,10 @@ class GuildScheduledEvent extends Base {
     return this.edit({ scheduledStartTime, reason });
   }
 
-  // TODO: scheduledEndTime gets reset on passing null but it hasn't been documented
```
Jiralite, line 453: Has it been documented then?

PHV08: Yes, I updated the JSDoc right below the deleted TODO comment to reflect that it accepts null to reset/remove the end time, and updated the TypeScript types in typings/index.d.ts as well

Jiralite: It would be nice to see a source!

PHV08: According to the Discord Developer Portal's [Guild Scheduled Event Object documentation](https://discord.com/developers/docs/resources/guild-scheduled-event#guild-scheduled-event-object):

* `scheduled_end_time` is defined as:  the time the guild scheduled event will end, or **null** if the event does not have a scheduled time to end.
* Under the [Modify Guild Scheduled Event](https://discord.com/developers/docs/resources/guild-scheduled-event#modify-guild-scheduled-event) endpoint, it accepts a nullable timestamp (`?ISO8601 timestamp`) to allow resetting/clearing it.

packages/discord.js/typings/index.d.ts
```diff
@@ -6462,7 +6465,7 @@ export interface BaseGuildScheduledEventOptions {
   privacyLevel?: GuildScheduledEventPrivacyLevel;
   reason?: string;
   recurrenceRule?: GuildScheduledEventRecurrenceRuleOptions | null;
-  scheduledEndTime?: DateResolvable;
+  scheduledEndTime?: DateResolvable | null;
```
almeidx, line 6468: This change was missed on the JSDoc side.
However, this is not correct as-is since this is also being used by the create options, on which the property is not documented[^1] as nullable, just optional.

[^1]: <https://docs.discord.com/developers/resources/guild-scheduled-event#create-guild-scheduled-event>

packages/discord.js/src/structures/GuildScheduledEvent.js
almeidx: It can only be null for non-external scheduled events, might be worth noting this in the jsdocs.

almeidx (changes requested):

packages/discord.js/src/managers/GuildScheduledEventManager.js
```diff
@@ -193,7 +193,7 @@ class GuildScheduledEventManager extends CachedManager {
    * @typedef {Object} GuildScheduledEventEditOptions
    * @property {string} [name] The name of the guild scheduled event
    * @property {DateResolvable} [scheduledStartTime] The time to schedule the event at
-   * @property {DateResolvable} [scheduledEndTime] The time to end the event at
+   * @property {DateResolvable|null} [scheduledEndTime] The time to end the event at
```
almeidx:
```suggestion
   * @property {?DateResolvable} [scheduledEndTime] The time to end the event at
   * <info>This cannot be `null` if `entityType` is {@link GuildScheduledEventEntityType.External}</info>
```

packages/discord.js/src/structures/GuildScheduledEvent.js
```diff
@@ -450,11 +450,10 @@ class GuildScheduledEvent extends Base {
...
   /**
    * Sets a new time to end the event at.
    *
-   * @param {DateResolvable} scheduledEndTime The time to end the event at
+   * @param {DateResolvable|null} scheduledEndTime The time to end the event at, or null to reset/remove the end time (only for non-external events)
```
almeidx: I think this makes it more clear 
```suggestion
   * @param {?DateResolvable} scheduledEndTime The time to end the event at, or `null` to remove the end time
   * <info>This cannot be `null` if this event's `entityType` is {@link GuildScheduledEventEntityType.External}</info>
```

almeidx (changes requested):

almeidx (approved):

packages/discord.js/src/structures/GuildScheduledEvent.js
Qjuh: > Under the [Modify Guild Scheduled Event](https://discord.com/developers/docs/resources/guild-scheduled-event#modify-guild-scheduled-event) endpoint, it accepts a nullable timestamp (`?ISO8601 timestamp`) to allow resetting/clearing it.

It shows as `ISO8601 timestamp` so non-nullable on that page though...

PHV08: you're right that the Modify endpoint's table lists it as 

`ISO8601 timestamp` without the `?`. 
However, the API does accept `null` to clear the end time for Stage/Voice events. In fact, `discord.js`'s internal payload formatter in `GuildScheduledEventManager` explicitly forwards `null` directly to the API:
```js
scheduled_end_time: scheduledEndTime ? new Date(scheduledEndTime).toISOString() : scheduledEndTime,

Qjuh: So the TODO remark is still valid, we don't document it accepting null since discord doesn't either.

the line with the ternary you mentioned is for when scheduledEndTime is not provided (so `undefined`) and hence gets omitted from the JSON in the request. Not for passing `null`.

PR #11588, opened 2026-07-29
fix(builders): preserve empty attachment files

ShiroKSH:
## Summary

- Preserve empty string attachment data in `AttachmentBuilder`.
- Include the resulting zero-byte file in `MessageBuilder` multipart output.

## Root cause

Truthiness checks treated `''` as if no file data had been supplied. The REST layer accepts it and creates a valid zero-byte `Blob`, so builders silently dropped a requested attachment before the request was made.

## Validation

- `vitest run --config ../../vitest.config.ts` — 25 suites, 268 tests passed; type errors: none
- `prettier --check packages/builders`
- `eslint --format=pretty packages/builders/src packages/builders/__tests__`
- Built `@discordjs/util` and `@discordjs/builders` with Node 24

### DIFF
--- a/packages/builders/__tests__/messages/fileBody.test.ts
+++ b/packages/builders/__tests__/messages/fileBody.test.ts
@@ -35,6 +35,25 @@ test('AttachmentBuilder handles 0 as a valid id', () => {
 	});
 });
 
+test('AttachmentBuilder preserves zero-byte string files', () => {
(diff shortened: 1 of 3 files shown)


almeidx (approved):

Jiralite:
This does not seem right... does this actually send to Discord?

almeidx:
Yes

Jiralite (approved):

SpaceEEC (approved):

PR #11589, opened 2026-07-29
chore: replace tsup with tsc

Qjuh:
This is an alternative proposal to #11519. This PR does (also) 3 things:
- bump typescript API to v6 while tsc is v7 (because v7 has no API yet)
- no longer use a bundler but use tsc directly instead
- remove legacy CJS builds, since node support require of ESM builds in all supported versions

This is relevant because tsup doesn't support TS 6.0.

This PR needs to be upgraded once TS v7.1 with API drops:
- https://github.com/microsoft/TypeScript/issues/63703


packages/actions/tsconfig.json
```diff
@@ -1,9 +1,13 @@
 {
 	"$schema": "https://json.schemastore.org/tsconfig.json",
-	"extends": "../../tsconfig.json",
+	"extends": "../../tsconfig.base.json",
 	"compilerOptions": {
 		"types": ["node"],
-		"skipLibCheck": true
+		"lib": ["DOM", "ESNext"],
```
vladfrangu: I don't think we want DOM types in this project

packages/api-extractor/bin/api-extractor
```diff
@@ -1,2 +1,2 @@
 #!/usr/bin/env node
-require('../dist/start.js');
+import('../dist/start.js');
```
vladfrangu, line 2: Are we sure this _works_? I'm guessing it WILL because of the pkg.json saying its ESM but pls double check

packages/brokers/tsconfig.json
```diff
@@ -1,6 +1,10 @@
 {
 	"$schema": "https://json.schemastore.org/tsconfig.json",
-	"extends": "../../tsconfig.json",
+	"extends": "../../tsconfig.base.json",
+	"compilerOptions": {
+		"rootDir": "./src",
+		"lib": ["DOM", "ESNext"]
```
vladfrangu, line 6: Ditto

packages/docgen/bin/index.js
```diff
@@ -1,2 +1,2 @@
 #!/usr/bin/env node
-require('../dist/bin/index.js');
+import '../dist/bin/index.js';
```
vladfrangu: so here we do `import './path'` but above in my previous review we do `import()`. Please uniform it to one type

packages/scripts/package.json
```diff
@@ -5,30 +5,27 @@
...
 		".": {
-			"require": {
+			"import": {
 				"types": "./dist/src/index.d.ts",
 				"default": "./dist/src/index.js"
-			},
-			"import": {
-				"types": "./dist/src/index.d.mts",
-				"default": "./dist/src/index.mjs"
 			}
```
vladfrangu: Everywhere else you collapsed these to 1 layer higher but not here 👀?

packages/scripts/tsconfig.json
```diff
@@ -1,6 +1,10 @@
 {
 	"$schema": "https://json.schemastore.org/tsconfig.json",
-	"extends": "../../tsconfig.json",
+	"extends": "../../tsconfig.base.json",
+	"compilerOptions": {
+		"lib": ["DOM", "ESNext"],
```
vladfrangu, line 5: Ditto

packages/ui/src/lib/components/discord/MessageAuthor.tsx
```diff
@@ -1,4 +1,4 @@
-import { FiCheck } from '@react-icons/all-files/fi/FiCheck';
+import { FiCheck } from '@react-icons/all-files/fi/FiCheck.js';
```
vladfrangu: Ah ah ah, double check this is actually valid!

tsconfig.base.json
```diff
@@ -18,32 +18,32 @@
...
 		"allowSyntheticDefaultImports": true,
-		"esModuleInterop": false,
 		"forceConsistentCasingInFileNames": true,
 		"isolatedModules": true,
 
 		// Language and Environment
 		"experimentalDecorators": true,
 		"lib": ["ESNext", "esnext.disposable"],
-		"target": "ESNext",
+		"target": "ES2024",
```
vladfrangu: Why not 2025?

vladfrangu (dismissed):

packages/api-extractor/bin/api-extractor
Qjuh: it does, else the Docs workflow would fail since it uses this executable.

tsconfig.base.json
Qjuh: Because we have Node24 as minimum engine atm.

vladfrangu: Which it does support: https://node.green/#ES2025

packages/brokers/tsconfig.json
Qjuh: Here we need them for `@msgpack/msgpack` because that uses `BufferSource` as a type which is a DOM type. Alternatively we could `skipLibCheck: true` if that's better in your opinion.

packages/scripts/tsconfig.json
Qjuh: Here `@vercel/blob` uses `HeadersInit` which is a DOM type.

packages/ui/src/lib/components/discord/MessageAuthor.tsx
Qjuh: It's not ideal because it imports the CJS file which vite then transpiles to ESM but it's the only solution I found to `tsc` complaining about not finding the package

Qjuh: Ah, nvm, found a better fix

tsconfig.base.json
Qjuh: ah, fair enough.

packages/scripts/tsconfig.json
vladfrangu: Riiight, and node/undici types don't define those globals?

Qjuh: undici does as export, but not as global. and since `@vercel/blob` doesn't import from undici the type is missing.

tsconfig.base.json
```diff
@@ -18,32 +18,32 @@
...
+		"module": "NodeNext",
+		"moduleResolution": "NodeNext",
 		"resolveJsonModule": true,
 		"resolvePackageJsonExports": true,
 		"resolvePackageJsonImports": true,
+		"rootDir": "${configDir}/src",
 
 		// Emit
 		"declaration": true,
-		"declarationMap": true,
```
ckohen, line 29: Why turn declaration map off now?

packages/ws/tsconfig.json
```diff
@@ -1,6 +1,10 @@
 {
 	"$schema": "https://json.schemastore.org/tsconfig.json",
-	"extends": "../../tsconfig.json",
+	"extends": "../../tsconfig.base.json",
+	"compilerOptions": {
+		"types": ["node"],
+		"rootDir": "./src"
```
ckohen: doesn't the root config do this for us? (across packages)

packages/ws/tsconfig.docs.json
```diff
@@ -2,7 +2,9 @@
 	"$schema": "https://json.schemastore.org/tsconfig.json",
 	"extends": "../../tsconfig.docs.json",
 	"compilerOptions": {
-		"outDir": "dist-docs"
+		"types": ["node"],
+		"outDir": "dist-docs",
+		"rootDir": "./src"
```
ckohen: ditto (across packages)

packages/scripts/bin/injectVersion.ts
```diff
@@ -0,0 +1,16 @@
+import { readFile, stat, writeFile } from 'node:fs/promises';
+import { join } from 'node:path';
+import { cwd } from 'node:process';
+import { pathToFileURL } from 'node:url';
+
+const packageConfig = await import(pathToFileURL(join(cwd(), 'package.json')).href, { with: { type: 'json' } });
+
+const filenames = ['index.js', 'index.mjs', 'index.cjs'];
```
ckohen: this would only work if we bundled, need to either have a config file for version injection or walk the entire tree

(or have a specific file in each package that needs it that we import from elsewhere, eg. version.ts that is just `export const packageVersion = '[[<version>]]'`)

packages/discord.js/typings/index.tst.ts
```diff
@@ -231,16 +239,13 @@ import {
...
-  UserSelectMenuComponent,
-  UserSelectMenuInteraction,
-  Webhook,
 } from './index.js';
 
+const expectType = <Type>(source: Type) => expect(source).type.toBe<Type>();
+const expectNotType = <Type>(source: unknown) => expect(source).type.not.toBe<Type>();
+const expectAssignable = <Type>(source: Type) => expect(source).type.toBeAssignableTo<Type>();
+const expectNotAssignable = <Type>(source: unknown) => expect(source).type.not.toBeAssignableTo<Type>();
+
```
mrazauskas: Glad to see you picked TSTyche! I am its author.

Just want to draw your attention that these four helpers do not work as intended. Type tests are only statically analysed, and this means the arguments will not be passed from wrappers to assertions.

Try it out by adding the following lines and running `tstyche`. Only the second line does actually error:

```ts
expectNotAssignable<string>('abc');
expect('abc').type.not.toBeAssignableTo<string>();
```

tsconfig.base.json
Qjuh: We don't "turn it off now", we never emited them. Previously we used tsup for emitting and that doesn't support declaration maps at all. So turning the setting off now actually means *not* changing what we emit.

packages/discord.js/typings/index.tst.ts
Qjuh: Thanks for that hint. Those helpers were indeed just meant temporarily while testing if tstyche does what we need/want. I didn't realize they didn't actually achieve what I planned. I now find&replace'd all occurences of those calls with the respective tstyche calls but now it seems typescript is overwhelmed.

Our typings *are* quite a lot and the test file also isn't small by any means, so I can understand why it's overwhelmed, but that means we can't just convert the tests 1:1 over to tstyche atm...

```ts
runs ./typings/index.tst.ts
/Users/qjuh/VSCode/discord.js/node_modules/.pnpm/typescript@6.0.3/node_modules/typescript/lib/typescript.js:68310
  function instantiateType(type, mapper) {
                          ^

RangeError: Maximum call stack size exceeded
    at instantiateType (/Users/qjuh/VSCode/discord.js/node_modules/.pnpm/typescript@6.0.3/node_modules/typescript/lib/typescript.js:68310:27)
    at instantiateList (/Users/qjuh/VSCode/discord.js/node_modules/.pnpm/typescript@6.0.3/node_modules/typescript/lib/typescript.js:67937:24)
    at instantiateTypes (/Users/qjuh/VSCode/discord.js/node_modules/.pnpm/typescript@6.0.3/node_modules/typescript/lib/typescript.js:67951:12)
    at instantiateTypeWorker (/Users/qjuh/VSCode/discord.js/node_modules/.pnpm/typescript@6.0.3/node_modules/typescript/lib/typescript.js:68362:36)
    at instantiateTypeWithAlias (/Users/qjuh/VSCode/discord.js/node_modules/.pnpm/typescript@6.0.3/node_modules/typescript/lib/typescript.js:68343:20)
    at instantiateType (/Users/qjuh/VSCode/discord.js/node_modules/.pnpm/typescript@6.0.3/node_modules/typescript/lib/typescript.js:68311:29)
    at getMappedType (/Users/qjuh/VSCode/discord.js/node_modules/.pnpm/typescript@6.0.3/node_modules/typescript/lib/typescript.js:67991:67)
    at /Users/qjuh/VSCode/discord.js/node_modules/.pnpm/typescript@6.0.3/node_modules/typescript/lib/typescript.js:68120:56
    at map (/Users/qjuh/VSCode/discord.js/node_modules/.pnpm/typescript@6.0.3/node_modules/typescript/lib/typescript.js:2495:19)
    at getObjectTypeInstantiation (/Users/qjuh/VSCode/discord.js/node_modules/.pnpm/typescript@6.0.3/node_modules/typescript/lib/typescript.js:68120:29)
```

packages/scripts/bin/injectVersion.ts
Qjuh: You're right, I didn't realize we had three packages that used it in other files than index.ts

packages/discord.js/typings/index.tst.ts
```diff
@@ -0,0 +1,3193 @@
...
+  const defaultCollector = message.createMessageComponentCollector();
+  expect(message.awaitMessageComponent()).type.toBeAssignableTo<Promise<MessageComponentInteraction>>();
+  expect(channel.awaitMessageComponent()).type.toBeAssignableTo<Promise<MessageComponentInteraction>>();
+  expect(defaultCollector).type.toBeAssignableTo<InteractionCollector<CollectedMessageInteraction>>();
+
+  defaultCollector.on('collect', (...args) => expect(args).type.toBe<[ButtonInteraction | SelectMenuInteraction]>());
+  defaultCollector.on('dispose', (...args) => expect(args).type.toBe<[ButtonInteraction | SelectMenuInteraction]>());
+  defaultCollector.on('end', (...args) =>
+    expect(args).type.toBe<[ReadonlyCollection<Snowflake, ButtonInteraction | SelectMenuInteraction>, string]>(),
+  );
```
mrazauskas: @Qjuh Please try commenting out this assertion. TSTyche will run without crashing, so that must be a bug. I will have a closer look.

By the way, few assertions have errors. The one from `message.channel` is known limitation related to reduced types, let me think about it. Other seems to be correct. (I guess that tests of `GuildScheduledEventRecurrenceRuleOptions` are missing `as const` somewhere, but might be something else too.)

packages/discord.js/typings/index.tst.ts
```diff
@@ -0,0 +1,3193 @@
...
+    interval: 1,
+    byMonth: [GuildScheduledEventRecurrenceRuleMonth.May],
+    byMonthDay: [4],
+    // Invalid property
+    byNWeekday: [{ n: 1, day: GuildScheduledEventRecurrenceRuleWeekday.Monday }],
+  }).type.not.toBeAssignableTo<GuildScheduledEventRecurrenceRuleOptions>();
+
+  expect({
+    startAt: new Date(),
+    frequency: GuildScheduledEventRecurrenceRuleFrequency.Yearly,
```
mrazauskas:
```suggestion
    frequency: GuildScheduledEventRecurrenceRuleFrequency.Yearly as const,
```

Hover on the `frequency` property and you will see that the inferred type is `GuildScheduledEventRecurrenceRuleFrequency`. `as const` turns it into `GuildScheduledEventRecurrenceRuleFrequency.Yearly` and the test will pass.

`as const` must be added to negative assertions as well.

---

This worked with `tsd` because it uses a different type inference mechanism for positive and negative assertions (https://github.com/tsdjs/tsd/issues/141). Hover on the same `frequency` property and you will see that positive assertions infer the type as `GuildScheduledEventRecurrenceRuleFrequency.Yearly`, but in negative assertions it becomes `GuildScheduledEventRecurrenceRuleFrequency`.

packages/discord.js/typings/index.tst.ts
```diff
@@ -0,0 +1,3193 @@
...
+    expect(message.channel).type.toBe<GuildTextBasedChannel>();
+    expect(message.guild).type.toBe<Guild>();
+    expect(message.member).type.toBe<GuildMember | null>();
+
+    expect(message.mentions).type.toBe<MessageMentions<true>>();
+    expect(message.guild).type.toBe<Guild>();
+    expect(message.mentions.members).type.toBe<Collection<Snowflake, GuildMember>>();
+  }
+
+  expect(message.channel).type.toBe<Exclude<TextBasedChannel, PartialGroupDMChannel>>();
```
mrazauskas, line 492:
```suggestion
  // TODO expect(message.channel).type.toBe<Exclude<TextBasedChannel, PartialGroupDMChannel>>();
  expect(message.channel).type.toBeAssignableFrom<Exclude<TextBasedChannel, PartialGroupDMChannel>>();
  expect(message.channel).type.toBeAssignableTo<Exclude<TextBasedChannel, PartialGroupDMChannel>>();
```

Above is the workaround for this problem. It is not very good, but can be used temporarily. The problem is that TypeScript does not have an API to get the reduced type. I found a somewhat dirty solution for that and did not dare to ship it, but it is worth having a second thought.

Hope this unblocks the PR. I will work on the two left issues this / next week.

ckohen (approved):

packages/discord.js/typings/index.tst.ts
```diff
@@ -0,0 +1,3197 @@
...
+
+    expect(message.mentions).type.toBe<MessageMentions<true>>();
+    expect(message.guild).type.toBe<Guild>();
+    expect(message.mentions.members).type.toBe<Collection<Snowflake, GuildMember>>();
+  }
+
+  // TODO: uncomment once tstyche supports the test this way
+  // expect(message.channel).type.toBe<Exclude<TextBasedChannel, PartialGroupDMChannel>>();
+  expect(message.channel).type.toBeAssignableFrom<Exclude<TextBasedChannel, PartialGroupDMChannel>>();
+  expect(message.channel).type.toBeAssignableTo<Exclude<TextBasedChannel, PartialGroupDMChannel>>();
```
mrazauskas: Must be fixed in the freshly released `tstyche@7.2.3`. The other issue remains, working on it.

Qjuh: Thank you for the fix. What other issue are you referring to, as it seems you fixed both issues we had already.

mrazauskas: There is one more commented out test on lines 545-547. Seems like it still causes "Maximum call stack size exceeded".

packages/discord.js/typings/index.tst.ts
```diff
@@ -0,0 +1,3194 @@
...
+  expect(message.awaitMessageComponent()).type.toBeAssignableTo<Promise<MessageComponentInteraction>>();
+  expect(channel.awaitMessageComponent()).type.toBeAssignableTo<Promise<MessageComponentInteraction>>();
+  expect(defaultCollector).type.toBeAssignableTo<InteractionCollector<CollectedMessageInteraction>>();
+
+  defaultCollector.on('collect', (...args) => expect(args).type.toBe<[ButtonInteraction | SelectMenuInteraction]>());
+  defaultCollector.on('dispose', (...args) => expect(args).type.toBe<[ButtonInteraction | SelectMenuInteraction]>());
+  // TODO: uncomment once tstyche supports this test
+  /* defaultCollector.on('end', (...args) =>
+    expect(args).type.toBe<[ReadonlyCollection<Snowflake, ButtonInteraction | SelectMenuInteraction>, string]>(),
+  ); */
```
mrazauskas: @Qjuh This is resolved in freshly published `tstyche@7.2.4`. The cause was `ReadonlyCollection` and `Collection` types referencing each other in their methods, and this way creating an infinite recursion. Now such cases are identified and handled properly.

PR #11590, opened 2026-08-01
feat(invites): backport invite target users and role

imnaiyar:
Backports #11392


packages/discord.js/src/managers/GuildInviteManager.js
```diff
@@ -168,7 +182,7 @@ class GuildInviteManager extends CachedManager {
 
   /**
    * Create an invite to the guild from the provided channel.
-   * @param {GuildInvitableChannelResolvable} channel The options for creating the invite from a channel.
+   * @param {GuildInvitableChannelResolvable} channel The channel where invite should be created.
```
Qjuh:
```suggestion
   * @param {GuildInvitableChannelResolvable} channel The channel where the invite should be created.
```

packages/discord.js/src/managers/GuildInviteManager.js
```diff
@@ -210,6 +238,49 @@ class GuildInviteManager extends CachedManager {
 
     await this.client.rest.delete(Routes.invite(code), { reason });
   }
+
+  /**
+   * Get target users for an invite
```
Qjuh:
```suggestion
   * Get target users for an invite.
```

packages/discord.js/src/managers/GuildInviteManager.js
```diff
@@ -210,6 +238,49 @@ class GuildInviteManager extends CachedManager {
...
+   */
+  async fetchTargetUsers(invite) {
+    const code = resolveInviteCode(invite);
+    const arrayBuff = await this.client.rest.get(Routes.inviteTargetUsers(code));
+
+    return Buffer.from(arrayBuff);
+  }
+
+  /**
+   * Updates target users for an invite
```
Qjuh:
```suggestion
   * Updates target users for an invite.
```

packages/discord.js/src/managers/GuildInviteManager.js
```diff
@@ -210,6 +238,49 @@ class GuildInviteManager extends CachedManager {
...
+
+    return this.client.rest.put(Routes.inviteTargetUsers(code), {
+      body: await createInviteFormData(this.client, { targetUsersFile }),
+      // This is necessary otherwise rest stringifies the body
+      passThroughBody: true,
+    });
+  }
+
+  /**
+   * Get status of the job processing target users of an invite
```
Qjuh:
```suggestion
   * Get status of the job processing target users of an invite.
```

packages/discord.js/src/structures/Invite.js
```diff
@@ -309,6 +334,42 @@ class Invite extends Base {
     return this;
   }
 
+  /**
+   * Update target users of this invite.
+   *
+   * @param {UserResolvable[]|BufferResolvable} targetUsersFile An array of users or a CSV file
+   * with a single column of user ids for all the users able to accept this invite
+   * @returns {Promise<unknown>}
```
Qjuh: Why is this `unknown`? Don't we know the result type? `RESTPutAPIInviteTargetUsersResult` is `undefined`, so why return anything at all and not do `await this.client.rest.put(...)` resulting in `Promise<void>` as return.

packages/discord.js/src/structures/Role.js
```diff
@@ -1,89 +1,28 @@
...
-
-    if ('color' in data) {
-      /**
-       * The base 10 color of the role
-       *
-       * @type {number}
-       * @deprecated Use {@link Role#colors} instead.
-       */
-      this.color = data.color;
-    }
```
Qjuh, line 62: This is a breaking change, because `BaseRole` doesn't have this.

packages/discord.js/src/structures/Role.js
```diff
@@ -495,21 +386,14 @@ class Role extends Base {
     );
   }
 
-  /**
-   * When concatenated with a string, this automatically returns the role's mention instead of the Role object.
-   * @returns {string}
-   * @example
-   * // Logs: Role: <@&123456789012345678>
-   * console.log(`Role: ${role}`);
-   */
```
Qjuh, line 504: why remove this JSDoc comment?

packages/discord.js/typings/index.d.ts
```diff
@@ -3343,34 +3348,40 @@ export interface RoleColorsEditResolvable {
   tertiaryColor?: ColorResolvable | null;
 }
 
-export class Role extends Base {
-  private constructor(client: Client<true>, data: RawRoleData, guild: Guild);
-  /** @deprecated Use {@link Role.colors} instead. */
-  public color: number;
```
Qjuh, line 3349: Ditto, breaking change.

packages/discord.js/typings/index.d.ts
```diff
@@ -7277,9 +7300,10 @@ export interface InviteCreateOptions {
   maxUses?: number;
   unique?: boolean;
   reason?: string;
+  roles?: readonly RoleResolvable[];
   targetApplication?: ApplicationResolvable;
   targetUser?: UserResolvable;
-  targetType?: InviteTargetType;
```
Qjuh, line 7282: Why was this removed?

Qjuh (changes requested):

packages/discord.js/typings/index.d.ts
imnaiyar: seems to be removed accidentally  while cherry-picking

packages/discord.js/src/structures/Role.js
imnaiyar: Right, I missed that

packages/discord.js/src/structures/Role.js
imnaiyar: Because it is already defined for its parent method in the `BaseRole`. Don't that carry-over? Idk if it will cause issues for api-extractor but at least for intellisense, it does work

Qjuh: While that might be true I'm not that keen to remove a JSDoc comment that was previously there in a semver minor...

imnaiyar: I'll revert it for this pr then

packages/discord.js/src/managers/GuildInviteManager.js
```diff
@@ -179,21 +193,35 @@ class GuildInviteManager extends CachedManager {
...
-      body: {
-        temporary,
-        max_age: maxAge,
-        max_uses: maxUses,
-        unique,
-        target_user_id: this.client.users.resolveId(targetUser),
-        target_application_id: targetApplication?.id ?? targetApplication?.applicationId ?? targetApplication,
-        target_type: targetType,
-      },
+      body: targetUsersFile ? await createInviteFormData(this.client, { targetUsersFile, ...options }) : options,
```
Qjuh: Please adapt this the way the way main PR does. /rest on v14 branch should be able to handle file.key too.

Qjuh (changes requested):

packages/core/src/api/channel.ts
```diff
@@ -419,12 +426,13 @@ export class ChannelsAPI {
 	 */
 	public async createInvite(
 		channelId: Snowflake,
-		body: RESTPostAPIChannelInviteJSONBody,
+		{ targetUsersFile, ...body }: CreateInviteOptions,
```
Qjuh, line 429: Technically not a breaking change but shouldn't the file be part of the third parameter instead? Since it isn't part of the body, same way reason isn't. Probably also applies to the original PR this is a backport of.

Qjuh (approved):
Apart from the question regarding /core LGTM

packages/core/src/api/channel.ts
imnaiyar: Made it so because of this comment by almeida
https://discord.com/channels/222078108977594368/1209960895141388389/1535378529129209956

Qjuh: fair enough

PR #11593, opened 2026-08-04
fix(Collector): persist time and idle options on resetTimer

69poorpotato:
This PR fixes issue #11553.

Previously, `resetTimer` would use the provided `time` or `idle` to reset the active timeout, but it failed to persist the new values to `this.options`. As a result, the next time the collector fired a `collect` event and automatically reset its idle timer, it would revert back to the originally configured `options.idle`, ignoring the user's manual timer extension.

This patch updates `this.options` if new `time` or `idle` options are explicitly passed to `resetTimer()`.


almeidx:
Duplicate of #11534

PR #11598, opened 2026-08-08
fix(voice): remove player from datastore during buffering

### LINKED ISSUE #11039: AudioPlayer does not get removed from DataStore in some situations

voice

### Issue description

For some reason, `audioPlayer.stop()` seems to cause the node process to never exit when called after destroying voice connection(s?) or not letting the event loop run for a bit before destroying voice connection(s?).

When ran in a debugger and pressing pause after the "cleanup" code has run, it will pause somewhere inside DataStore, and inspecting the variables shows that the AudioPlayer does not get removed from the list of active AudioPlayers even though `stop()` was called.

After seeing some `setTimeout` and `setIntervals`, I guessed that stopping audio player and destroying a voice connection back-to-back might cause issues, so by putting a sleep in-between, it does solve the problem and no longer hangs after "cleanup". It does look like a workaround so it should be fixed or at the very least be documented somewhere.

[audioPlayerHangSample.zip](https://github.com/user-attachments/files/21716711/audioPlayerHangSample.zip)

### Code sample

```typescript
// Please see the zip file for complete reproducible sample
```

### Versions

- `@discordjs/voice` v0.18.0
- `discord.js` v14.21.0
- Node v24.2.0
- Typescript v5.8.3

### Issue priority

Medium (should be fixed soon)

No Partials

GuildVoiceStates

_No response_

Umit-Ulusoy:
This fix updates the AudioPlayer stop method to allow direct transition to the Idle state when the player is currently in the Buffering status.

Previously, calling stop() during Buffering would hang or fail to clear the player due to silence padding constraints, leaving it lingering in the DataStore.

This ensures the player is properly cleaned up and removed from the store.

Fixes #11039


Qjuh:
(How) did you verify that the Player actually hangs/stays in DataStore before this fix? Because looking at the code the AudioResource should keep on playing 5 (by default) silence packets before the Player going to Idle state (and removing the Player from the DataStore in the process). Independent of it the resource had started or is still buffering. If users wanted to force the immediate stop then that parameter already exists.

Umit-Ulusoy:
Hello there @Qjuh, thanks for taking a closer look.
Good question.

The silence-padding mechanism (`silenceRemaining` being decremented, `_preparePacket` being called, and the eventual transition to `Idle`) is entirely handled inside `_stepPrepare()`.
That function starts with this guard:

private _stepPrepare() {
	const state = this._state;
	// Guard against the Idle state
	if (state.status === AudioPlayerStatus.Idle || state.status === AudioPlayerStatus.Buffering) return;
	...

So when the player is in `Buffering`, `_stepPrepare()` returns immediately and never processes the silence countdown.
In the original `stop()`:

} else if (this.state.resource.silenceRemaining === -1) {
	this.state.resource.silenceRemaining = this.state.resource.silencePaddingFrames;
}

this line does set `resource.silenceRemaining`, but nothing ever consumes it while the player stays in `Buffering`.
If the underlying stream never becomes readable (e.g. the scenario in the linked issue, where a voice connection is destroyed before the stream starts emitting data), the player has no path out of `Buffering`, it never reaches `Idle`, so `deleteAudioPlayer` is never called, and the player lingers in the global timer's list forever.

To confirm this isn't just a theoretical reading of the code, I ran the existing test suite against the unpatched `stop()`:

FAIL  should remove AudioPlayer from DataStore when stopped during Buffering state
AssertionError: expected 'buffering' to deeply equal 'idle'
Expected: "idle"
Received: "buffering"

and against the patched version, the same test (plus a second regression test I added for the "stream never becomes readable" case) passes cleanly, with no other test files affected.

On the `force` parameter point, I agree that's the mechanism for an intentional immediate stop.
The issue here is different, it's not that `stop()` is slow without `force`, it's that it can silently never complete at all when called during `Buffering`.
Silence padding doesn't really have meaning yet at that point anyway, since no packets have been played, so treating `Buffering` like `force` for the purposes of this transition seemed like the correct semantics rather than a workaround.

Qjuh:
Okay, I see your point now. Your test however doesn't actually test for what it should. It would also fail if the 5 silence packets were sucessfully sent and then end the resource and thus the player, because your 2 wait() calls don't take long enough for five cycles of the audio loop (which takes ~100 ms). But I verified that the player does indeed keep staying stuck like you said even if waiting longer and thank you for pointing out the reason being the guard against Idle/Buffering state in _stepPrepare().

Qjuh (approved):

Umit-Ulusoy:
Thanks for verifying it and for the feedback, @Qjuh!

PR #11600, opened 2026-08-10
feat(rest)!: add rejectOnRateLimit on a per-request basis

didinele:
because I need this feature :)

One thing of note, I removed the `string[]` variant of this option. Not only is it not that useful, but it also posed some issues:
```ts
options.rejectOnRateLimit.some((route) => rateLimitData.route.startsWith(route.toLowerCase()))
```

Notice how we `toLowerCase()` `route` (what the user provided at `REST` instantiation), which makes `rateLimitData.route` strings that contain uppercase characters unmatachable - oops!

One such easy case to spot is:
```ts
if (Date.now() - timestamp > 1_000 * 60 * 60 * 24 * 14) {
	exceptions += '/Delete Old Message';
}
```
---

BREAKING CHANGE: `RESTOptions#rejectOnRateLimit` now only accepts `boolean` or `RateLimitQueueFilter`


packages/rest/src/lib/utils/types.ts
```diff
@@ -99,14 +99,16 @@ export interface RESTOptions {
...
+	 *
+	 * Pass `true` to throw a {@link RateLimitError} on every rate limit, `false` or `null` to wait every
+	 * rate limit out, or a filter to decide per rate limit.
+	 *
+	 * This is only consulted for requests that do not provide their own {@link RequestData.rejectOnRateLimit | rejectOnRateLimit}.
 	 *
 	 * @defaultValue `null`
 	 */
-	rejectOnRateLimit: RateLimitQueueFilter | string[] | null;
+	rejectOnRateLimit: RateLimitQueueFilter | boolean | null;
```
almeidx: Since this is major anyways, I think we drop `null` and default to `false` instead. Feels more idiomatic to me, and there's no practical difference between null and false here

almeidx (changes requested):

packages/rest/src/lib/utils/utils.ts
```diff
@@ -150,15 +151,14 @@ export function shouldRetry(error: Error | NodeJS.ErrnoException) {
...
-			: options.rejectOnRateLimit.some((route) => rateLimitData.route.startsWith(route.toLowerCase()));
-	if (shouldThrow) {
+export async function onRateLimit(manager: REST, rateLimitData: RateLimitData, requestData: HandlerRequestData) {
+	// Explicit false/null opts out of `REST` level `rejectOnRateLimit`, so we compare with `undefined`.
+	const policy =
+		requestData.rejectOnRateLimit === undefined ? manager.options.rejectOnRateLimit : requestData.rejectOnRateLimit;
+
+	if (!policy) return;
+
+	if ((typeof policy === 'boolean' && policy) || (await policy(rateLimitData))) {
```
Qjuh:
```suggestion
	if (typeof policy === 'boolean' || (await policy(rateLimitData))) {
```
if it's false it would've returned early already, the double check is redundant

packages/rest/src/lib/utils/types.ts
vladfrangu: +1

vladfrangu:
I've added backport-candidate, it's easy enough to backport (and throw a deprecation warning for string[])

packages/rest/src/lib/utils/types.ts
```diff
@@ -339,6 +341,32 @@ export interface RequestData {
 	 * Reason to show in the audit logs
 	 */
 	reason?: string | undefined;
+	/**
+	 * Determines how a rate limit encountered while making this request should be handled.
+	 *
+	 * Pass `true` to throw a {@link RateLimitError} rather than wait, `false` (or `null`) to wait it out, or
+	 * a filter to decide per rate limit. Takes precedence over {@link RESTOptions.rejectOnRateLimit}, so
```
ckohen:
```suggestion
	 * a filter to decide based on rate limit data. Takes precedence over {@link RESTOptions.rejectOnRateLimit}, so
```

packages/rest/src/lib/utils/types.ts
```diff
@@ -99,14 +99,16 @@ export interface RESTOptions {
...
-	 * Determines how rate limiting and pre-emptive throttling should be handled.
-	 * When an array of strings, each element is treated as a prefix for the request route
-	 * (e.g. `/channels` to match any route starting with `/channels` such as `/channels/:id/messages`)
-	 * for which to throw {@link RateLimitError}s. All other request routes will be queued normally
+	 * The default policy determining how rate limiting and pre-emptive throttling should be handled.
+	 *
+	 * Pass `true` to throw a {@link RateLimitError} on every rate limit, `false` or `null` to wait every
+	 * rate limit out, or a filter to decide per rate limit.
+	 *
+	 * This is only consulted for requests that do not provide their own {@link RequestData.rejectOnRateLimit | rejectOnRateLimit}.
```
ckohen:
```suggestion
	 * This can be overridden per request via the {@link RequestData.rejectOnRateLimit | rejectOnRateLimit} request option.
```

ckohen (commented):
I feel like this reads better, but I'm bad at writing docs so...

Otherwise looks good (+1 on the other requested changes)

Qjuh (approved):

packages/rest/src/lib/utils/utils.ts
```diff
@@ -150,15 +151,13 @@ export function shouldRetry(error: Error | NodeJS.ErrnoException) {
...
-	const { options } = manager;
-	if (!options.rejectOnRateLimit) return;
-
-	const shouldThrow =
-		typeof options.rejectOnRateLimit === 'function'
-			? await options.rejectOnRateLimit(rateLimitData)
-			: options.rejectOnRateLimit.some((route) => rateLimitData.route.startsWith(route.toLowerCase()));
-	if (shouldThrow) {
+export async function onRateLimit(manager: REST, rateLimitData: RateLimitData, requestData: HandlerRequestData) {
+	// Explicit false opts out of `REST` level `rejectOnRateLimit`, so we compare with `undefined`.
```
ckohen:
```suggestion
	// Explicit false opts out of `REST` level `rejectOnRateLimit`, only `undefined` falls back.
```

ckohen (changes requested):

ckohen (approved):

almeidx (approved):
This could be added to `/core` per-request options in a follow-up pull request

iCrawl (approved):

PR #11601, opened 2026-08-13
refactor(WebSocketShard): clear ws session only when explicitly reconnecting

didinele:
The rationale is simple. If you call `manager.destroy()` to teardown gracefully, the if in its previous state would clear the session info to prevent a resume (i.e. the `connect()` function of the shard class decides if it should resume simply if session information exists and its known shard count matches the currently configured one).

This is arguably incorrect behavior. It assumes that a `manager.destroy()` call means "I am absolutely not coming back soon", when it could be a good-faith effort to close the connection before returning to resume (like a proper process teardown on SIGINT).

One drawback with this approach is it *really* might mean "see ya", in which case for a properly persistent session (i.e. off-process), that the user has not correctly configured to expire in any meaningful way, may be used on `connect`, leading to Discord immediately refusing a `resume` before the shard attempts to `identify`. There are 2 things we could do here, neither are a must, and they are not mutually exclusive.

1. Make a better attempt in documenting this behavior and how session info should always be set to expire if it sees no writes (e.g. `EX` in Redis `SET`)
2. Perhaps instead make the check `options.recover !== WebSocketShardDestroyRecovery.Resume && options.code !== CloseCodes.Resuming`) - though how about if the user provides `WebSocketShardDestroyRecovery.Reconnect` with `CloseCodes.Resuming`? I suppose then the better proposal would be `options.recover !== WebSocketShardDestroyRecovery.Resume || options.code !== CloseCodes.Resuming`

I'm open to feedback based on the above.


packages/ws/src/ws/WebSocketShard.ts
```diff
@@ -360,7 +360,7 @@ export class WebSocketShard extends AsyncEventEmitter<WebSocketShardEventsMap> {
 		this.failedToConnectDueToNetworkError = false;
 
 		// Clear session state if applicable
-		if (options.recover !== WebSocketShardDestroyRecovery.Resume) {
+		if (options.recover === WebSocketShardDestroyRecovery.Reconnect) {
```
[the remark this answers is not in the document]
didinele: [On this page](https://docs.discord.com/developers/topics/opcodes-and-status-codes#gateway), seems to me that under the "Gateway Close Event Codes" table `4009`/Session timed out is documented as a close code that we should attempt to resume off of, so the behavior is correct.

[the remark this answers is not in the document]
didinele: See https://docs.discord.com/developers/events/gateway#resuming

Quote, "It’s disconnected with a **close code that indicates it can reconnect**. A list of close codes is in the Opcodes and Status Codesdocumentation."

vladfrangu (approved):
I think this is fine as is, but do some testing. It sounds like
1. I can call destroy with some args to tell it "don't 'destroy destroy', i will be back"
	- Trade off: if you dont resume fast enough you just gotta identify, its w/e
2. I can call destroy with some other args it clears session info

I think thats fine as is... Maybe we can document it, sure

didinele:
Oh yeah, you made me remember that if you close with `1000` they actually reject the resume, since it tells them to teardown internal state. So I think the `options.recover !== WebSocketShardDestroyRecovery.Resume || options.code === CloseCodes.Normal` is the correct answer here (NOT `!== CloseCodes.Resuming`, since that'd filter other resumable codes, and for the usecase I described you *have* to do `destroy({ code: CloseCodes.Resuming })`

Qjuh (approved):

packages/ws/src/ws/WebSocketManager.ts
```diff
@@ -379,6 +379,9 @@ export class WebSocketManager extends AsyncEventEmitter<ManagerShardEventsMap> i
 		this.#token = token;
 	}
 
+	/**
+	 * To keep the shard(s) resumable, use the {@link CloseCodes.Resuming} code.
```
almeidx, line 386: This seems like a `@remarks` rather than the description of the method

packages/ws/src/ws/WebSocketShard.ts
```diff
@@ -360,7 +360,7 @@ export class WebSocketShard extends AsyncEventEmitter<WebSocketShardEventsMap> {
 		this.failedToConnectDueToNetworkError = false;
 
 		// Clear session state if applicable
-		if (options.recover !== WebSocketShardDestroyRecovery.Resume) {
+		if (options.recover !== WebSocketShardDestroyRecovery.Resume || options.code === CloseCodes.Normal) {
```
almeidx: This looks a bit weird. Passing `{ code: CloseCodes.Normal }` to `destroy()` would still enter this block, because `recover` would not be present (since it's different than `WebSocketShardDestroyRecovery.Resume`). Is this expected?

almeidx (changes requested):

packages/ws/src/ws/WebSocketShard.ts
didinele: Yes, see https://docs.discord.com/developers/events/gateway#initiating-a-disconnect

> When you close the connection to the gateway with close code 1000 or 1001, **your session will be invalidated** and your bot will appear offline.
If you simply close the TCP connection or **use a different close code** [...]

If we were to keep the session, the next `connect()` call would attempt to use it and get rejected.

almeidx: I said the wrong thing, I meant to use `{ code: CloseCodes.Resuming }` as the example. This still enters the block since `recover` is missing

didinele: Mmmmh, yeah you're right. In trying to fix the other problem I mentioned earlier in the thread I ended up breaking it in this other way. ~~I think the real answer then is `.recover !== Reconnecting`~~, but will need to double check if that's correct.

Figured it out at last.

packages/ws/src/ws/WebSocketShard.ts
```diff
@@ -360,7 +360,12 @@ export class WebSocketShard extends AsyncEventEmitter<WebSocketShardEventsMap> {
 		this.failedToConnectDueToNetworkError = false;
 
 		// Clear session state if applicable
-		if (options.recover !== WebSocketShardDestroyRecovery.Resume) {
+		if (
+			!(
+				(options.recover === WebSocketShardDestroyRecovery.Resume || options.code === CloseCodes.Resuming) &&
+				options.code !== CloseCodes.Normal
+			)
+		) {
```
almeidx, line 366: I think this reads better.
```suggestion
		if (
			(options.recover !== WebSocketShardDestroyRecovery.Resume && options.code !== CloseCodes.Resuming) ||
			options.code === CloseCodes.Normal
		) {
```

almeidx (approved):

PR #11602, opened 2026-08-13
refactor(ws)!: take /gateway/bot data in `WebSocketManager#connect`

didinele:
`WebSocketManager` never really refetched the data anyways, it relied on cache. If there are any staleness concerns, we can 100% address them by instead taking the data inside `connect`, which should be on-par with old behavior.

BREAKING CHANGE: `WebSocketManager` no longer accepts `fetchGatewayInformation`, pass `gatewayInformation` to `connect` instead.

### DIFF
--- a/packages/ws/src/ws/WebSocketManager.ts
+++ b/packages/ws/src/ws/WebSocketManager.ts
@@ -54,20 +54,6 @@ export interface SessionInfo {
  */
 export interface RequiredWebSocketManagerOptions {
 	/**
-	 * Information retrieved from the `/gateway/bot` endpoint.
-	 * We recommend using a REST client that respects Discord's rate limits, such as `@discordjs/rest`.
-	 *
-	 * @example
-	 * ```ts
-	 * const rest = new REST().setToken(process.env.DISCORD_TOKEN);
-	 * const manager = new WebSocketManager({
-	 *  token: process.env.DISCORD_TOKEN,
-	 *  gatewayInformation: (await rest.get(Routes.gatewayBot())) as RESTGetAPIGatewayBotResult,
-	 * });
-	 * ```
-	 */
-	gatewayInformation: RESTGetAPIGatewayBotResult;
-	/**
 	 * The intents to request
 	 */
 	intents: GatewayIntentBits | 0;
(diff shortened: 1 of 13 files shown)


didinele:
This comment https://github.com/discordjs/discord.js/pull/11602#discussion_r3778396547 combined with the complexity introduced in discord.js has made me change my mind. I'll make it an option in `connect` instead.

didinele:
@coderabbitai review

packages/ws/src/ws/WebSocketManager.ts
```diff
@@ -280,61 +264,54 @@ export class WebSocketManager extends AsyncEventEmitter<ManagerShardEventsMap> i
...
-		if (this.gatewayInformation) {
-			if (this.gatewayInformation.expiresAt <= Date.now()) {
-				this.gatewayInformation = null;
-			} else if (!force) {
-				return this.gatewayInformation.data;
-			}
+	public getGatewayInformation(): RESTGetAPIGatewayBotResult {
+		if (!this.#gatewayInformation) {
+			throw new Error('Gateway information has not been set. Invoke `connect()` first.');
 		}
```
vladfrangu, line 273: I don't like that you need to actually connect to set this data... Imo this should have a setter method instead that needs to be called before connect

vladfrangu (changes requested):

packages/ws/src/ws/WebSocketManager.ts
Qjuh: But the information about gateway like "remaining" session starts could get stale before the connect() call then. I'd say providing it at the time it gets used sounds like the correct approach to me.

Qjuh (approved):

packages/ws/src/ws/WebSocketManager.ts
didinele: Yeah, I disagree with a setter method here as well, also because it makes the natural usage of the package particularly awkward (being forced to add another method call before `connect()`), while the use case of having a manager with the data set that has not invoked `connect()` are... none, as far as I can tell.

vladfrangu (approved):

sdanialraza (approved):

didinele:
@discord-js-bot pack this

almeidx (approved):

